PR #12706, opened 2022-09-06
[Node][Metaschedule] Allow ignoring NDArray raw data in StructuralEqual / Hash

masahi:
Motivation: Currently, `StructuralEqual/Hash` take into account NDArray raw data to determine equality / compute hash. This is problematic for `link-params = True` case, because meta schedule task extraction and database look up will treat the following subgraphs as distinct, even though the only difference is the content of NDArray constants. 
```
def @fused_nn.conv2d_add(%p0: Tensor[(1, 32, 32, 256), float32] /* ty=Tensor[(1, 32, 32, 256), float32] */, Primitive=1) -> Tensor[(1, 32, 32, 256), fl
oat32] {                                                                                                                      
  %0 = nn.conv2d(%p0, meta[relay.Constant][0] /* ty=Tensor[(3, 3, 256, 256), float32] */, padding=[1, 1, 1, 1], channels=256, kernel_size=[3, 3], data_
layout="NHWC", kernel_layout="HWIO") /* ty=Tensor[(1, 32, 32, 256), float32] */;                                                                       
  add(%0, meta[relay.Constant][1] /* ty=Tensor[(1, 1, 1, 256), float32] */) /* ty=Tensor[(1, 32, 32, 256), float32] */        
}

def @fused_nn.conv2d_add(%p0: Tensor[(1, 32, 32, 256), float32] /* ty=Tensor[(1, 32, 32, 256), float32] */, Primitive=1) -> Tensor[(1, 32, 32, 256), fl
oat32] {                                                                                                                      
  %0 = nn.conv2d(%p0, meta[relay.Constant][0] /* ty=Tensor[(3, 3, 256, 256), float32] */, padding=[1, 1, 1, 1], channels=256, kernel_size=[3, 3], data_
layout="NHWC", kernel_layout="HWIO") /* ty=Tensor[(1, 32, 32, 256), float32] */;                                                                       
  add(%0, meta[relay.Constant][1] /* ty=Tensor[(1, 1, 1, 256), float32] */) /* ty=Tensor[(1, 32, 32, 256), float32] */  
} 
```  

This PR adds options to `StructuralEqual/Hash` to allow ignoring NDArray raw data when comparing / hashing. MS task extraction and database look up are now using these options, since NDArray raw data doesn't matter to them.

cc @Hzfengsy @junrushao1994

PR #13001, opened 2022-10-06
[Node] Expose StructuralEqual/Hash handler implemenation to header

masahi:
When we use `StructuralEqual/Hash`, the actual equality testing / hashing work are deferred to `RemapVarSEqualHandler` and `VarCountingSHashHandler` class respectively, defined in `.cc` files. To customize equality / hash behavior, I want to be able to derive from these classes outside of `structural_equal.cc` or `structural_hash.cc`. This is the first step toward replacing my previous attempt https://github.com/apache/tvm/pull/12706 to allow ignoring NDArray raw data in `StructuralEqual/Hash`.

So I propose to expose them as the "default" handler, while still hiding their implementation details. This lets me define a custom hasher that ignores NDArray content simply by:

```
class SHashHandlerIgnoreNDArray : public SHashHandlerDefault {
 protected:
  void DispatchSHash(const ObjectRef& object, bool map_free_vars) override {
    ICHECK(object.defined());
    if (auto ndarray = object.as<runtime::NDArray::Container>()) {
      SHashReducer hash_reduce(this, map_free_vars);
      NDArrayHash(ndarray, &hash_reduce, false);
    } else {
      SHashHandlerDefault::DispatchSHash(object, map_free_vars);
    }
  }
};
```

cc @tqchen @junrushao

### DIFF
--- a/include/tvm/node/structural_equal.h
+++ b/include/tvm/node/structural_equal.h
@@ -334,11 +334,11 @@ class SEqualHandlerDefault : public SEqualReducer::Handler {
   SEqualHandlerDefault(bool assert_mode, Optional<ObjectPathPair>* first_mismatch);
   virtual ~SEqualHandlerDefault();
 
-  virtual bool SEqualReduce(const ObjectRef& lhs, const ObjectRef& rhs, bool map_free_vars,
-                            const Optional<ObjectPathPair>& current_paths) override;
-  virtual void DeferFail(const ObjectPathPair& mismatch_paths) override;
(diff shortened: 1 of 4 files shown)


junrushao:
Would love to have @tqchen's thoughts on this :-)

tqchen:
seems to be a fine change, would be good to get another pair of eyes for reviewing

junrushao (approved):
LGTM

masahi:
@tqchen good to merge?

junrushao:
Sorry for the delay!

PR #13004, opened 2022-10-06
[Hexagon] disable cache_write schedule type from sw pipeline test

adstraw:
Disabling the "cache_write" schedule from this test as it is failing on device.  The root cause of the failure will be addressed with additional buffer dependency analysis in the InjectSoftwarePipeline pass to come in a subsequent PR.

### DIFF
--- a/tests/python/contrib/test_hexagon/test_software_pipeline_async.py
+++ b/tests/python/contrib/test_hexagon/test_software_pipeline_async.py
@@ -28,7 +28,10 @@ outer = tvm.testing.parameter(8, 16)
 inner = tvm.testing.parameter(64, 128)
 dtype = tvm.testing.parameter("uint8", "float16")
 scope = tvm.testing.parameter("global", "global.vtcm")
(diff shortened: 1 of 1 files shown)


masahi:
I'm seeing "1 workflow awaiting approval (First-time contributors need a maintainer to approve running workflows)" in this PR, while @adstraw is obviously not a first time contributor. Why is that? cc @driazati

driazati:
> I'm seeing "1 workflow awaiting approval (First-time contributors need a maintainer to approve running workflows)" in this PR, while @adstraw is obviously not a first time contributor. Why is that? cc @driazati
> 
> UPDATE: #13005 has the same issue.

Not sure why (seems like a bug from GitHub?), since even on the hover on @adstraw's profile it says there are recent contributions. ASF Infra might have changed something in the repo settings but it seems unlikely that they would do that without telling us / it would even cause this kind of behavior. I've approved both PRs, if you notice anything else like this please tag me on it

tmoreau89 (approved):
LGTM

PR #12959, opened 2022-09-30
[LOGGING] Optionally print backtrace on segfault

tkonolige:
Add BACKTRACE_ON_SEGFAULT cmake option to install a signal handler to print a backtrace on segfault. Disabled by default because existing signal handlers may be overriden by this signal handler. Creating a backtrace may allocate, which should not happen in a signal handler, but the program is crashing so it shouldn't make any difference.

Hopefully this will help diagnose segfaults like in #12955.

@areusch @driazati (not sure who else to tag)

Issue #12955, opened 2022-09-30
[Flaky Test] `tests/python/unittest/test_roofline.py::test_estimate_peak_flops[llvm]`


driazati:
### Test(s)

- `tests/python/unittest/test_roofline.py::test_estimate_peak_flops[llvm]`

### Jenkins Links

- https://ci.tlcpack.ai/blue/organizations/jenkins/tvm/detail/main/4402/pipeline

tkonolige:
Given that #12959 is merged, can we reenable the roofline test to get a stack trace for debugging?

PR #13007, opened 2022-10-07
[ci] Re-enable roofline test

driazati:
After #12959, this re-enables the test disabled in #12955 to get a
backtrace next time it fails.


tkonolige (approved):
@driazati thanks!

Issue #12362, opened 2022-08-10
[Bug][Relay] TVM generates input name that cannot be imported with the parser


mehrdadh:
I'm compiling an ONNX model which I realized the input name has this format `input:0`. When I import the model and run it, it works fine, however if I try to import the same Relay text with the parser it will result in error.

cc @gigiblender

slyubomirsky:
Can you write a simple pass that will change ("sanitize") the variable names? I don't know if a colon is treated in any special way by the parser, but you could try replacing them with underscores. I thought the parser did some sanitization of its own and am surprised that this still happens.

mehrdadh:
@slyubomirsky I added a cleanup step in the ONNX frontend to replace `:` with `_`.

driazati:
This seems like it's fixed by #13046, if that's not the case we can re-open this issue

PR #13011, opened 2022-10-07
[Frontend][ONNX] Replace `:` in input name with `_`

mehrdadh:
This PR changes onnx frontend to rename input with `:` character. This will fix this issue:

https://github.com/apache/tvm/issues/12362

### DIFF
--- a/python/tvm/relay/frontend/onnx.py
+++ b/python/tvm/relay/frontend/onnx.py
@@ -5629,6 +5629,7 @@ class GraphProto:
 
     def _parse_graph_input(self, graph):
         for i in graph.input:
+            i_name_compatible = None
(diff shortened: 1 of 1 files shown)


AndrewZhaoLuo (commented):
Hmm, I think the current code is a bit confusing, can you try to clean it up? 

Regardless, I suspect this might not be the best approach. This PR is designed to fix a problem in the parser, why not fix the problem in the parser? Or is the grammar of relay fundamentally incompatible with ":"? 

I only say this because you can make a relay variable with ":". If this is allowed in the relay language it should also be supported by the parser.

mehrdadh:
The issue is fixed with this PR:
https://github.com/apache/tvm/pull/13046

PR #13016, opened 2022-10-09
[TIR] Fix the stride calculation in InjectRollingBuffer

liangW-intellif:
The InjectRollingBuffer pass seems to miscalculate rolling_buffer strides when the bound overlap is large and fractional strides exist.
The original test case `test_upscale` does not expose this problem because there is no positive bound overlap in the intermediate buffer, so I modified the pass and the test case as well.

### DIFF
--- a/src/tir/transforms/inject_rolling_buffer.cc
+++ b/src/tir/transforms/inject_rolling_buffer.cc
@@ -126,15 +126,16 @@ class RollingBufferInjector : public StmtExprMutator {
         // We use the bound information of the BufferRealize to calculate
         // how we can legally roll
         auto stride{0};
+        auto divisor{1};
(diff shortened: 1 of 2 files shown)


wrongtest-intellif:
cc @mbaret Hi, could you kindly take a look at this :) ? 

We are also trying to migrate the primitive to TensorIR, it is quite useful for our NPU workloads.

junrushao:
@wrongtest-intellif @Hzfengsy would you guys do a round of review? Thanks!

Hzfengsy (approved):

mbaret:
Apologies for missing this, I've been away for a few weeks. Thanks for the catch @liangW-intellif and happy to review any migration attempt to TensorIR!

PR #13017, opened 2022-10-09
Mingw support

TSCCompiler:
Add support for MinGW compiler on windows platform, so You can use std c++17 or LLVM higher version(13.x ...) without upgrade MSVC compilers, It's also easy to deploy on windows platform


junrushao:
This is indeed an interesting change and it would be great if we could get this in. On the other hand, I would love to see the PR being organized a bit more :-)

sunhaohui:
Thank you! I 've already fixed conflicts I will learning how to make a better PR

junrushao:
Thank you so much! I'm happy to shepherd this PR :-)

.gitignore
```diff
@@ -268,3 +268,7 @@ gallery/how_to/work_with_microtvm/micro_tvmc.py
 
 # Used in CI to communicate between Python and Jenkins
 .docker-image-names/
+/python/images/
+/python/models/
+/python/depoly/
+/python/build/
```
junrushao, line 274: These lines may not be relevant - let's get them removed

src/relay/op/vm/vm.cc
```diff
@@ -107,7 +107,7 @@ TVM_REGISTER_GLOBAL("relay.op.vm.invoke_tvm_op")
     .set_body_typed([](Expr func, Expr inputs, Expr outputs, DictAttrs attrs) {
       return InvokeTVMOp(std::move(func), std::move(inputs), std::move(outputs), std::move(attrs));
     });
-
+TVM_REGISTER_NODE_TYPE(DictAttrsNode);
```
junrushao, line 110: it's been registered in `src/ir/attrs.cc`

src/relay/op/annotation/annotation.cc
```diff
@@ -201,5 +201,7 @@ TVM_REGISTER_GLOBAL("relay.op.annotation._make.compiler_end")
       return Call(op, {expr}, Attrs(attrs), {});
     });
 
+      TVM_DLL extern  const Op stop_fusion_op = Op::Get("annotation.stop_fusion");
+
```
junrushao, line 205: remove this?

.gitignore
sunhaohui: OK

src/relay/op/vm/vm.cc
sunhaohui: I know it ,but the init order of static variable is different from MinGW.  When registering static variable “vm.invoke_tvm_op” , It needs DictAttrs. Howerer DictAttrs has not been registered (静态全局变量的初始化顺序是编译器控制的，当注册vm.invoke_tvm_op的时候，DictAttrs还没有被注册，反正是Map，重复注册也没事，但是更安全)

src/relay/op/annotation/annotation.cc
sunhaohui: This is for MinGW register order , the same as vm.invoke_tvm_op

src/relay/op/vm/vm.cc
junrushao: It's definitely interesting to hear! As you mentioned, there exists a global static variable `vm.invoke_tvm_op` that depends on `DictAttrs` - which sounds to me is a design failure, because it is always assumed that there isn't such compiler-dependent behavior (static initialization order). Would love to see if there is an option to fix such behavior instead.

Chinese translation: 这个行为确实挺有趣的。不过我们设计这个系统的时候，希望不要有这种依赖于编译器的行为发生 (全局静态的初始化顺序)，所以不知道能不能帮忙指出 `vm.invoke_tvm_op` 这个全局变量在哪里呢？我想看看能不能修掉。

CMakeLists.txt
```diff
@@ -1,6 +1,27 @@
 cmake_minimum_required(VERSION 3.18)
 project(tvm C CXX)
 
+MESSAGE(STATUS "c compiler :  ${CMAKE_COMPILER_IS_GNUCC}")
+message(STATUS "system name : ${CMAKE_SYSTEM_NAME}")
```
driazati, line 5: thanks for the PR! just wanted to point out that we have `Summary.cmake`: https://github.com/apache/tvm/blob/main/cmake/utils/Summary.cmake#L47 (which you can enable with `-DSUMMARIZE=1` or `set(SUMMARIZE ON)` in `config.cmake`), so these info messages could be moved there

src/relay/op/vm/vm.cc
sunhaohui: src/relay/op/vm/vm.cc:111 (My branch ) Notice .set_attrs_type_key("DictAttrs")
RELAY_REGISTER_OP("vm.invoke_tvm_op")
    .describe(R"code(Invoke an operation compiled by TVM.)code" TVM_ADD_FILELINE)
    .set_num_inputs(3)
    .add_argument("op", "Function", "The operation to call")
    .add_argument("ins", "Tuple", "The input tensors.")
    .add_argument("outs", "Tuple", "The output tensors.")
    .add_type_rel("InvokeTVMOp", InvokeTVMOpRel)
    **_.set_attrs_type_key("DictAttrs")_**

junrushao: Got it! Thanks for your example. It makes sense to me!

PR #13022, opened 2022-10-10
[ETHOSN] Throw error message when inference fails

lhutton1:
Previously the runtime would silently skip interence failures and return random values as the result. This can make spotting inference failures challenging. The runtime now throws a fatal error when inference did not complete successfully along with an error message that gives some details about the error that occurred.

cc @Leo-arm @ashutosh-arm @leandron


src/runtime/contrib/ethosn/ethosn_device.cc
```diff
@@ -123,21 +133,26 @@ bool Inference(tvm::runtime::TVMArgs args, dl::Network* npu,
...
-    for (size_t i = 0; i < n_outputs; i++) {
-      DLTensor* tensor = outputs[i];
-      dl::Buffer* source_buffer = ofm_raw[i];
-      uint8_t* dest_buffer = static_cast<uint8_t*>(tensor->data);
-      size_t size = source_buffer->GetSize();
-      uint8_t* source_buffer_data = source_buffer->Map();
-      std::copy(source_buffer_data, source_buffer_data + size, dest_buffer);
-      source_buffer->Unmap();
-    }
+  WaitStatus result = WaitForInference(inference.get(), 60);
```
ashutosh-arm, line 138: nit: seems like something that should be a knob (in future) - wdyt?

tests/cpp/runtime/contrib/ethosn/inference_test.cc
```diff
@@ -0,0 +1,60 @@
...
+
+  ASSERT_EQ(result.GetErrorCode(), WaitErrorCode::Error);
+  ICHECK_EQ(result.GetErrorDescription(),
+            "Failed to read inference result status (No such file or directory)");
+}
+
+TEST(WaitForInference, InferenceTimeout) {
+  const int inference_scheduled = 0;
+  const int timeout = 0;
+  dl::Inference inference = dl::Inference(inference_scheduled);
```
ashutosh-arm, line 49: just curious why we don't have tests for other inputs except 0 and 3?

tests/cpp/runtime/contrib/ethosn/inference_test.cc
```diff
@@ -0,0 +1,60 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/*!
+ * \file tests/cpp/runtime/contrib/ethosn/inference_test.cc
+ * \brief Tests to check runtime components used during inference.
```
ashutosh-arm, line 22: nit: this is specific to npu, so it would be worthwhile noting down the name.

ashutosh-arm (commented):
Thanks Luke. Indeed a great change. A nit, a question and a suggestion below :sweat_smile:

src/runtime/contrib/ethosn/ethosn_device.cc
lhutton1: Absolutely :)

tests/cpp/runtime/contrib/ethosn/inference_test.cc
lhutton1: No particular reason other than being focussed on checking the error messages, will add those!

lhutton1: Added - 2 would be *success* which is tricky to test at this point due to requiring a valid `Inference` object

lhutton1:
@tvm-bot rerun

ashutosh-arm (approved):
@lhutton1 LGTM! Thanks for extending the test coverage :+1:

Leo-arm (approved):
LGTM

leandron (approved):
LGTM, thanks @lhutton1 @Leo-arm

Issue #12261, opened 2022-08-01
[Tracking Issue] Padded Layout Transformations


Lunderberg:
### This issue is to track progress for [padded layout transformations](https://github.com/apache/tvm-rfcs/pull/77)

- [x] New TIR Op, `tir::builtin::assume` (#12267)
- [x] New lowering transform, remove `tir::builtin::assume` (#12267)
- [x] New TIR Op, `tir::builtin::undef` (#12266)
- [x] New lowering transform, remove `tir::builtin::undef` (#12266)
- [ ] Enhancements to `transform_layout`
  - [ ] `tir.schedule.transform_layout`, rewrite loop structure if possible, mimicking TE functionality. (Implemented for padded transforms in #12720, still to implement for un-padded transforms)
  - [x] Insert pad value into generated TIR, using `tir::if_then_else`, `builtin::assume`, and `builtin::undef` (#12720)
- [ ] New utility, reorder loops according to buffer layout

  Should be implemented as wrapper function around `Schedule.transform_block_layout`.

- [ ] Enhancement, DomainTouched

  Must track a predicate for which this touch occurs, and any known values for this touch.

- [ ] Enhancement, Remove No Op

  Should identify writes that are later overwritten or deallocated without affecting the observable outputs.

- [x] Enhancement, Simplify (#13130)

  Should remove conditional statements that are provable using known contents of a buffer.

- [x] New transform, Hoist Expression. (#11592)

- [ ] New primitive, remove branching through overcompute

- [ ] New primitive, remove overcompute through branching

cc @Hzfengsy @junrushao

PR #13023, opened 2022-10-10
[Arith][TIR] Check for constant offsets of known literal constraints

Lunderberg:
Previously, the checks for a literal constraint would find exact matches for an inequality, but any alterations to the conditional would break this exact matching.  This commit introduces checks for constant offsets relative to a known value.  These checks are not always expressible using the existing `ConstIntSetAnalyzer`, which represents allowed values using a single contiguous region.  (e.g. `i!=5` is not representable, because it requires a region for `i<5` and another for `i>5`.)

This implementation reuses the internal representation for inequalities introduced in https://github.com/apache/tvm/pull/12863, along with much of its implementation.  However, the indirect comparisons (e.g. using `a < b` and `b < c` to prove that `a < c`) introduced in that PR still require an explicit flag to be used.

### DIFF
--- a/include/tvm/arith/analyzer.h
+++ b/include/tvm/arith/analyzer.h
@@ -409,10 +409,19 @@ class TransitiveComparisonAnalyzer {
    *
    * \param rhs The right-hand side of the comparison
    *
+   * \param propagate_inequalities If true, attempt to find a sequence
+   * of transitive inequalities that allow the lhs and rhs to be
+   * compared.  If false, only use the known comparison that have been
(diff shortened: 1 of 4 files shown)


Lunderberg:
This is part of a series of improved simplifications, as part of https://github.com/apache/tvm/issues/12261.

jwfromm (approved):
Very cool! LGTM.

Lunderberg:
Thank you!  (Rebasing onto main to resolve conflicts.)

csullivan:
@tvm-bot rerun

PR #13024, opened 2022-10-10
[TIR][Arith] Use TryCompare to narrow inequalities if possible

Lunderberg:
Prior to this commit, the result of TryCompare would only be used if it could definitively prove a conditional to be either true or false. For example, if it is known that `0 <= i`, a conditional of `i <= 0` would be left as-is.

This commit introduces rewrite rules to preferentially simplify into more restrictive conditions.  Using the same example, if it is known that `0 <= i`, a conditional of `i <= 0` would be simplified into `i == 0`.  Similarly, if it is known that `0 <= i`, a conditional of `i != 0` would be simplified into `0 < i`.

Because this change does not introduce significant overhead, as the results of `RewriteSimplifier::Impl::TryCompare` are already available, this change is enabled for all use cases and does not require a call to `RewriteSimplifier::SetEnabledExtensions`.

### DIFF
--- a/src/arith/rewrite_simplify.cc
+++ b/src/arith/rewrite_simplify.cc
@@ -29,6 +29,7 @@
 #include <tvm/tir/op.h>
 
 #include <algorithm>
+#include <utility>
 
 #include "../target/datatype/registry.h"
 #include "conjunctive_normal_form.h"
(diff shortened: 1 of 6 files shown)


Lunderberg:
This is part of a series of improved simplifications, as part of https://github.com/apache/tvm/issues/12261.  Currently marked as draft PR, as it depends on https://github.com/apache/tvm/pull/13023.

Lunderberg:
Last failing unit test requires #13081 to resolve.  An unrolled loop produced a `T.if_then_else(floormod(-i,2)==0, ...)`, which resulted in an error thrown from ModularSet.  Previously, this was `T.if_then_else(floormod(-i,2) <= 0, ...)`, which was ignored by the ModularSet.  Because `floormod` with a positive denominator is never negative, this could be simplified to `T.if_then_else(floormod(-i,2) == 0, ...)`, triggering that bug.

Lunderberg:
Rebased onto main to resolve conflicts.  Because this dev branch includes and depends on the changes made in https://github.com/apache/tvm/pull/13023, that PR should be landed before this one.  Therefore, marking this PR as a draft to avoid a weird git history from incorrect merge order.

Lunderberg:
Rebased onto main, which now includes #13023 to resolve the last failing unit test, and this PR is now ready for review.

Lunderberg:
@tvm-bot rerun

Lunderberg:
@tvm-bot rerun

tmoreau89 (approved):
PR looks sound to me; thank you Eric for the contribution

PR #13027, opened 2022-10-10
[Hexagon] fix bug with async dma queue id tracking

adstraw:
The `LowerAsyncDMA` pass is designed to replace TIR `async_wait_queue_scope` attributes only if it has already relaced a TIR `async_commit_queue_scope` attribute with memory copy semantics on the same queue.  Fixing a bug where the queue ID is saved in the "commit" transform regardless of whether a transformation takes place.  Meaning, the "wait" transform, at present, would operate on all queues.


adstraw:
@driazati could you start CI on this PR and help me understand why I keep bumping into the "first-time contributor" issue?

adstraw:
Decided to open a new PR that encompasses this one + adds a few more changes.

PR #13028, opened 2022-10-11
[Hexagon] [runtime] Manage RPC and runtime buffers separately

janetsc:
Creates HexagonPageAllocator to manage allocations for the RPC server

Adds new Device APIs for RPC buffer management

Allocations are tracked by two separate buffer managers:

- rpc_hexbuffs is used exclusively for RPC buffers
- runtime_hexbuffs is used exclusively for runtime buffers

This will fix the throw on shutdown in the simulator.

cc: @csullivan, @kparzysz-quic, @adstraw


src/runtime/hexagon/hexagon_device_api.cc
```diff
@@ -115,13 +118,37 @@ void* HexagonDeviceAPI::AllocDataSpace(Device dev, size_t nbytes, size_t alignme
   if (alignment < kHexagonAllocAlignment) {
     alignment = kHexagonAllocAlignment;
   }
-  return mgr->AllocateHexagonBuffer(nbytes, alignment, String("global"));
+  CHECK(runtime_hexbuffs) << "runtime_hexbuffs is not initalized";
```
Lunderberg, line 121: Same comment here, regarding error message.

src/runtime/hexagon/hexagon_device_api.cc
```diff
@@ -115,13 +118,37 @@ void* HexagonDeviceAPI::AllocDataSpace(Device dev, size_t nbytes, size_t alignme
...
 }
 
 void HexagonDeviceAPI::FreeDataSpace(Device dev, void* ptr) {
   CHECK(ptr) << "buffer pointer is null";
   CHECK(IsValidDevice(dev)) << "dev.device_type: " << dev.device_type;
-  mgr->FreeHexagonBuffer(ptr);
+  if (runtime_hexbuffs) {
+    runtime_hexbuffs->FreeHexagonBuffer(ptr);
+  } else {
+    LOG(INFO) << "FreeDataSpace called when runtime_hexbuffs is not initialized";
```
Lunderberg, line 131: Why is this only a `LOG(INFO)`, where an uninitialized `runtime_hexbuffs` is an error elsewhere?

src/runtime/hexagon/hexagon_device_api.h
```diff
@@ -190,16 +194,15 @@ class HexagonDeviceAPI final : public DeviceAPI {
            (DLDeviceType(dev.device_type) == kDLCPU);
   }
 
-  //! \brief Manages underlying HexagonBuffer allocations
+  //! \brief Manages RPC HexagonBuffer allocations
+  // rpc_hexbuffs is used only in Alloc/FreeRpcDataSpace
+  std::unique_ptr<HexagonBufferManager> rpc_hexbuffs;
```
Lunderberg, line 199: Since `rpc_hexbuffs` is initialized in the constructor and is never freed, does it need to be a `std::unique_ptr<HexagonBufferManager>` instead of `HexagonBufferManager`?

src/runtime/hexagon/hexagon_device_api.h
```diff
@@ -190,16 +194,15 @@ class HexagonDeviceAPI final : public DeviceAPI {
...
+
+  //! \brief Manages runtime HexagonBuffer allocations
```
Lunderberg, line 201: It looks like the primary purpose of the separation between `runtime_hexbuffs` and `rpc_hexbuffs` is to allow scoped allocations that can be uniformly cleaned up at the end of a scope.  Can we add a comment describing the purpose, along with why it is only being implemented for Hexagon?

Lunderberg (commented):
Overall, looks good, just a few questions and locations where comments/errors could be improved.

src/runtime/hexagon/hexagon_device_api.cc
janetsc: This is not an error condition, per se, as this can happen if clients that allocated memory free their buffers on shutdown.  Since this happens as part of process shutdown, it is ok that we've already freed the memory.

One thing I could do is keep track of the buffers that were freed in ReleaseResources, and make sure this is one of those pointers.

I also had it log a message using LOG instead of DLOG, because I want it to be visible when we "leak" memory at the end of a session.  Ideally, these should all be freed before the session ends, and ReleaseResources() is called from the RPC client.

src/runtime/hexagon/hexagon_device_api.h
janetsc: It could be static, yes.  I'll make that change.

src/runtime/hexagon/hexagon_device_api.h
janetsc: Yes, I'll add another comment.

src/runtime/hexagon/hexagon_device_api.cc
```diff
@@ -137,7 +166,7 @@ void* HexagonDeviceAPI::AllocWorkspace(Device dev, size_t size, DLDataType type_
 
 void HexagonDeviceAPI::FreeWorkspace(Device dev, void* data) {
   CHECK(IsValidDevice(dev)) << "dev.device_type: " << dev.device_type;
-  CHECK(mgr->count(data) != 0)
+  CHECK(runtime_hexbuffs == nullptr || runtime_hexbuffs->count(data) != 0)
```
janetsc, line 169: Add check for released_runtime_buffers if runtime_hexbuffs is nullptr

src/runtime/hexagon/hexagon_device_api.cc
```diff
@@ -115,13 +120,37 @@ void* HexagonDeviceAPI::AllocDataSpace(Device dev, size_t nbytes, size_t alignme
...
+    auto it = std::find(released_runtime_buffers.begin(), released_runtime_buffers.end(), ptr);
+    CHECK(it != released_runtime_buffers.end()) << "Attempted to free Hexagon data with "
+                                                << "HexagonDeviceAPI::FreeDataSpace that was not "
+                                                << "allocated during the session.";
+  }
+}
+
+void* HexagonDeviceAPI::AllocRpcBuffer(size_t nbytes, size_t alignment) {
+  CHECK(nbytes) << "number of bytes is zero";
+  CHECK(alignment) << "alignment is zero";
```
janetsc, line 144: Remove - allow 8 bit alignment for RPC data?

src/runtime/hexagon/hexagon_buffer_manager.h
```diff
@@ -85,6 +86,18 @@ class HexagonBufferManager {
     return hexagon_buffer_map_.empty();
   }
 
+  //! \brief Returns a list of allocated pointers.
```
Lunderberg, line 89: Nit: Should specify that this is a non-owning set of pointers.

src/runtime/hexagon/hexagon_buffer_manager.h
```diff
@@ -85,6 +86,18 @@ class HexagonBufferManager {
...
+  // Note - this should only be used by the device API to keep track of what
+  // was in the manager when HexagonDeviceAPI::ReleaseResources is called.
+  std::vector<void*> vector() {
```
Lunderberg, line 92: Nit: Renaming to `current_allocations` would tell a reader what is contained in the return value.

src/runtime/hexagon/hexagon_buffer_manager.h
```diff
@@ -85,6 +86,18 @@ class HexagonBufferManager {
...
+    std::vector<void*> allocated;
+    std::lock_guard<std::mutex> lock(map_mutex_);
+    for (auto it = hexagon_buffer_map_.begin(); it != hexagon_buffer_map_.end(); it++) {
```
Lunderberg, line 95: FYI, since we're on C++17 now, structured bindings with a ranged-based for loop reduce a lot of the iterator-based boilerplate.  This is more personal style, but it would let the loop be re-written as below.

```c++
for(const auto& [data_ptr, buffer] : hexagon_buffer_map_) {
    allocated.push_back(data_ptr);
}
```

Lunderberg (approved):
Changes look good to me, thank you!  A couple additional nitpicks that can go either way.

src/runtime/hexagon/hexagon_buffer_manager.h
janetsc: I like that much better, thanks!

src/runtime/hexagon/hexagon_buffer_manager.h
```diff
@@ -85,6 +86,18 @@ class HexagonBufferManager {
     return hexagon_buffer_map_.empty();
   }
 
+  //! \brief Returns a vector of currently allocated pointers, owned by the manager.
+  // Note - this should only be used by the device API to keep track of what
+  // was in the manager when HexagonDeviceAPI::ReleaseResources is called.
+  std::vector<void*> current_allocations() {
```
adstraw, line 92: I apologize for the post-merge feedback.  But ... having this method seems like a break in the abstraction of the `HexagonBufferManager` class, to me.  If we have to do this it feels like something else is amiss.

src/runtime/hexagon/hexagon_device_api.cc
```diff
@@ -115,13 +120,34 @@ void* HexagonDeviceAPI::AllocDataSpace(Device dev, size_t nbytes, size_t alignme
...
+  return runtime_hexbuffs->AllocateHexagonBuffer(nbytes, alignment, String("global"));
 }
 
 void HexagonDeviceAPI::FreeDataSpace(Device dev, void* ptr) {
   CHECK(ptr) << "buffer pointer is null";
   CHECK(IsValidDevice(dev)) << "dev.device_type: " << dev.device_type;
-  mgr->FreeHexagonBuffer(ptr);
+  if (runtime_hexbuffs) {
+    runtime_hexbuffs->FreeHexagonBuffer(ptr);
+  } else {
```
adstraw, line 134: Not sure on the value-add for this `else` case.  The only thing that happens is that we might fail in the `CHECK` if we free a buffer after we release resources that was not in the `released_runtime_buffers`.  Do we want this behavior?  What happens on-system if we just get rid of this `else` case?  We can then get rid of `released_runtime_buffers` and the need for `HexagonBufferManager::current_allocations`.  We can simply ...

```
  if (runtime_hexbuffs) {
    runtime_hexbuffs->FreeHexagonBuffer(ptr);
  }
```

With some comments to explain why the "free" may happen after the "release".

src/runtime/hexagon/hexagon_buffer_manager.h
janetsc: Thanks for the review.

I agree, this isn't ideal.  However, I wanted to make sure that I don't give a "free pass" to all allocations that come after ReleaseResources.  The scenario I was trying to avoid was:

User allocates buffer B
Session ends, ReleaseResources is called and B is freed
User frees buffer B as objects are being torn down, which should not throw as it is somewhat expected
User frees buffer C, which SHOULD throw, as it was never allocated in the first place

I added this so that I can check if there's no runtime_hexbuffs, then I make sure that what is being attempted to be freed is something that we owned before ReleaseResources was called.  In which case, that Free is a no-op.  I also want to detect spurious calls to Free that wouldn't have been legit before ReleaseResources was called.

I'm happy to discuss more, and brainstorm other ways to accomplish this.

src/runtime/hexagon/hexagon_device_api.cc
janetsc: My previous reply is related.

If there's a manager, I can just use it to call Free.  If there isn't (it has been released, or maybe acquire was never called), I can check what I had allocated before Release.  If that buffer is in there, I will not throw.

src/runtime/hexagon/hexagon_buffer_manager.h
Lunderberg: As an alternative, what if the `HexagonBufferManager` allows three states for buffers instead of two.

* Valid to use, valid to free (Initial state after allocating memory)
* Invalid to use, valid to free (State after closing a session)
* Invalid to use, invalid to free (State of unknown pointer)

That way, instead of discarding the `HexagonBufferManager runtime_hexbuffs` altogether when the session closes, it instead moves all keys from the internal `std::unordered_map<void*, std::unique_ptr<HexagonBuffer>> hexagon_buffer_map_;` to an internal `std::unordered_set<void*>`.  This avoids exposing any of the buffers to the outside, but does allow the `HexagonBufferManager` to identify a previously freed pointer.

adstraw: Thanks @Lunderberg, I think your suggestion would satisfy my concern.  What do you think @janetsc?  This would allow us to make valid frees a `nop` and to catch invalid frees as well.  If I am understanding correctly `runtime_hexbuffs` would be instantiated by the Hexagon Device API in the same way that we instantiate the `rpc_hexbuffs` ... `HexagonBufferManager runtime_hexbuffs;`?  If that is the case, we could potentially have just one HeagonBufferManager that implements the API and internal state that @Lunderberg suggests.

janetsc: There's also another state not captured - the buffer manager is in an invalid state to allocate more memory.

There needs to be two buffer managers, because on ReleaseResources how do we know which buffers need to be freed?

I'm thinking that a better model is to enforce that all frees should take place before ReleaseResources.  That's cleaner.  I will investigate which ones are outstanding.  All objects should be cleanly torn down by the time the session ends.  I will file a task to investigate.

src/runtime/hexagon/hexagon_device_api.cc
adstraw: Making sure all "temporal" frees occur before Release is a good idea and worthwhile endeavor. ✅ 

I am not convinced that two `HexagonBufferManager` objects are required. Instead, you could have separate interfaces in a single `HexagonBufferManager` class to track "permanent" (for RPC e.g.) versus "temporal" allocations.  And, in this way, alleviate the need for the `current_allocations` method which exposes the internals of the `HexagonBufferManager` class to its users.  Something to consider for a future cleanup, perhaps.

PR #13034, opened 2022-10-11
[FIX,LOWERING] Add attrs from Relay Functions to PrimFuncs in Metaschedule Lowering

tkonolige:
Attrs were not propogated from Relay functions to the corresponding PrimFunc when lowering with MetaSchedule enabled. These attrs are not seperately copied in the MetaSchedule lowering flow.

@junrushao


junrushao (approved):
LGTM. Is there any specific attribute in your mind that needs cross-IR propagation? hash?

tkonolige:
Hash is the most important one, but layout information is also useful.

junrushao:
Does Relay do any follow-up layout optimization after being lowered to TIR?

PR #13038, opened 2022-10-11
[TIR] Update ir_comparator message to be more clear about what is being com…

nverke:
…pared.

This would be more useful when debugging tensorize mismatches.


junrushao (approved):
LGTM!

PR #13040, opened 2022-10-11
[Hexagon] Remove setting CMAKE_CXX_COMPILER for Hexagon

janetsc:
Remove the setting for CMAKE_CXX_COMPILER for Hexagon as it is causing errors on docker builds on shared systems.

### DIFF
--- a/tests/scripts/task_config_build_hexagon.sh
+++ b/tests/scripts/task_config_build_hexagon.sh
@@ -28,7 +28,6 @@ echo set\(USE_RPC ON\) >> config.cmake
 echo set\(USE_MICRO ON\) >> config.cmake
 echo set\(USE_MICRO_STANDALONE_RUNTIME ON\) >> config.cmake
 echo set\(USE_LLVM "${CLANG_LLVM_HOME}/bin/llvm-config"\) >> config.cmake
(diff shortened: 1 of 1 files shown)


tests/scripts/task_config_build_hexagon.sh
```diff
@@ -28,7 +28,6 @@ echo set\(USE_RPC ON\) >> config.cmake
 echo set\(USE_MICRO ON\) >> config.cmake
 echo set\(USE_MICRO_STANDALONE_RUNTIME ON\) >> config.cmake
 echo set\(USE_LLVM "${CLANG_LLVM_HOME}/bin/llvm-config"\) >> config.cmake
-echo set\(CMAKE_CXX_COMPILER "/opt/sccache/clang++"\) >> config.cmake
```
driazati, line 31: It'd be fine to take this out, a bunch of people have had problems with it when running locally, but it should still stay on in CI. AFAIK all scripts that run in Jenkins during CI have the env variable `CI` set to `true`, could you switch on that and set `CMAKE_CXX_COMPILER` if that's the case?

janetsc: I’m going to cancel this - I was just going an experiment.  Thanks!

PR #13041, opened 2022-10-11
[Arith] Add internal NarrowPredicateExpression utility

Lunderberg:
Implements `tvm::arith::NarrowPredicateExpression`, a utility that removes free parameters from a boolean expression, such that the resulting expression being true implies that the original expression is true.  For example, the predicate `(0 <= i+f) && (i+f < 16)`, where `f` is a free parameter on the range `0 <= f < 2)`, can be narrowed to the expression `(0 <= i+0) && (i+2 < 16)`.

In effect, `NarrowPredicateExpression` functions as a context-sentive `tvm::tir::Substitute`, where the value substituted is selected such that the resulting expression errs on the side of being false.  This is an internal utility used as part of the simplifications for layout transformations ([tracking issue link](https://github.com/apache/tvm/issues/12261)).

### DIFF
--- a/src/arith/narrow_predicate_expression.cc
+++ b/src/arith/narrow_predicate_expression.cc
@@ -1,0 +1,219 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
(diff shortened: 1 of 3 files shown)


Lunderberg:
Marking as a draft PR until it passes CI.  This utility could also be moved to be internal to an upcoming data-flow analysis utility.

Lunderberg:
Rebasing onto main to get the bug fix from #13067

wrongtest-intellif:
> `0 <= f < 2)`, can be narrowed to the expression `(0 <= i+0) && (i+2 < 16)`

Hi~ I feel a bit confused about here: when `f == 0`, the `i` could take value in range `[0, 16)`, thus what is the purpose to get narrowed predicate `i + 2 < 16`?

Lunderberg:
Hi @wrongtest-intellif !  The goal is to produce an expression that, when true, implies that the original expression is true.  For this specific example, suppose we know that `i + 2 < 16`.  Because we also know that the free parameter `f` satisfies `2 < f`, we know that `i + 2 < i + f`.  Therefore, since `i + f < i + 2 < 16`, `i+f < 16`.

The `NarrowPredicateExpression` is part of a larger data-flow analysis that I'm preparing as part of handling padded buffer transforms.  When analyzing whether a specific index corresponds to padding, conditions such as `pad_left <= i+f < original_width + pad_left` (identifies a non-padding data value) frequently show up.  By converting these into expressions that are true regardless of the value of `f`, or expressions that are false regardless of the value of `f`, we can remove reduction axes from the overall predicate expressions.

jwfromm (approved):
Very cool idea and a nice implementation. Thanks @Lunderberg!

PR #13043, opened 2022-10-11
[MetaSchedule, Testing] Allow generate default design space for given target

vinx13:
This allows usage like `generate_design_space('llvm', workload, target)` without specifying types, which is useful for some debugging purpose.


junrushao:
i'm not sure if it's the right way. alternatively, can we do:

```python
ms.TuneContext(mod=mod, target=target, space_generator="post-order-apply").generate_design_space()
```

vinx13:
The alternative also works, closing this for now

PR #13046, opened 2022-10-11
[AOT] Sanitize input/output name in runtime

mehrdadh:
This PR adds name sanitization to input/output at AOT runtime module. This means when we use set_input/set_output with AOT, even if we use the original name of those input/outputs which were sanitized and changed in the codegen, it will map them to the correct input/output.

For example if your model has an input with name `input:0`, the AOT codegen would change this to `input_0` but at runtime if we try to set this input it does not exist. Now, we use the same sanitization at runtime

cc @areusch


src/runtime/aot_executor/aot_executor.cc
```diff
@@ -107,7 +108,7 @@ PackedFunc AotExecutor::GetFunction(const std::string& name,
   } else if (name == "set_input_zero_copy") {
     return PackedFunc([sptr_to_self, this](TVMArgs args, TVMRetValue* rv) {
       if (String::CanConvertFrom(args[0])) {
-        int in_idx = this->GetInputIndex(args[0].operator String());
+        int in_idx = this->GetInputIndex(tvm::runtime::SanitizeName(args[0].operator String()));
```
alanmacd, line 111: Does it also make sense to add a python test that passes in names like `input:0` to test that this functionality doesn't regress for the various AoT commands?

mehrdadh: yeah it makes sense. I will add a test.

mehrdadh: I added a test which exams both mangled name and the original name

include/tvm/runtime/name_mangling.h
```diff
@@ -0,0 +1,43 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/*!
+ * \file tvm/runtime/name_mangling.h
+ * \brief Utility functions for name mangling which are used
```
Mousius, line 22: Given the name transforms are fairly standalone, can we move them all rather than having some here and some in the original file?

Mousius (changes requested):

include/tvm/runtime/name_mangling.h
mehrdadh: I moved this one since it is shared between runtime library and tvm library. I think it's better to keep the rest in src/backend since they are not used in runtime currently. wdyt?

tests/python/relay/aot/test_cpp_aot.py
```diff
@@ -203,5 +203,39 @@ def test_pass_wrong_device_arg():
...
+    test_so_path = temp_dir / "test.so"
+    mod.export_library(test_so_path, cc="c++", options=["-std=gnu++17", "-g3", "-O0"])
+    # test both original name and mangled name
+    for name in ["input_0", input_name]:
+        loaded_mod = tvm.runtime.load_module(test_so_path)
+        runner = tvm.runtime.executor.AotModule(loaded_mod["default"](tvm.cpu(0)))
+        inputs = {name: input_data}
+        runner.set_input(**inputs)
+        runner.run()
+        assert (runner.get_output(0).asnumpy() == expected_output).all()
```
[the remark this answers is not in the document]
mehrdadh: added

alanmacd (approved):

include/tvm/runtime/name_mangling.h
Mousius: I don't see much harm moving them all across, they're small, self contained and it's likely if you need one you need others. Either way, it'd be good not to have `name_mangling.h` and `name_transforms.h` as it confuses people, if you don't want to move it together could we at least name them consistently? "Mangle" is also an unnecessarily aggressive word, we should probably not be using it in our codebase.

mehrdadh: Makes sense, I'll rename it to `name_transforms`

tests/python/relay/aot/test_cpp_aot.py
```diff
@@ -203,5 +203,44 @@ def test_pass_wrong_device_arg():
...
+            tvm.IRModule.from_expr(func),
+            target=target_kind,
+            params=params,
+            executor=tvm.relay.backend.Executor("aot", {"interface-api": "packed"}),
+        )
+    temp_dir = tvm.contrib.utils.TempDirectory()
+    test_so_path = temp_dir / "test.so"
+    mod.export_library(test_so_path, cc="c++", options=["-std=gnu++17", "-g3", "-O0"])
+    # test both original name and transformed name
+    for name in ["input_0", input_name]:
```
lhutton1, line 232: This `for` seems redundant

lhutton1 (commented):
Thanks for the fix @mehrdadh, I ran into this issue recently so thought I'd have a quick look. Just found a few nits in the tests, otherwise LGTM!

tests/python/relay/aot/test_cpp_aot.py
```diff
@@ -203,5 +203,44 @@ def test_pass_wrong_device_arg():
...
+        loaded_mod = tvm.runtime.load_module(test_so_path)
+        runner = tvm.runtime.executor.AotModule(loaded_mod["default"](tvm.cpu(0)))
+        inputs = {name: input_data}
+        runner.set_input(**inputs)
+
+        input_ind = runner.get_input_index(name)
+        assert (runner.get_input(input_ind).asnumpy() == input_data).all
```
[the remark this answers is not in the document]
mehrdadh: yep, thanks for catching that!

tests/python/relay/aot/test_cpp_aot.py
lhutton1: Apologies, ignore - I see what this is doing now

mehrdadh: I added both cases to check both sanitized name and original name works. The reason for checking sanitized name is that eventually we will add a check in relay.Var to not support special characters from the first place. That means in future if you have a model with input `input:0`, the generated relay module would convert it to `input_0` from the first place and if user take a look at the generated relay they might use `input_0` from the first place instead of the original name.
let me what you think

mehrdadh:
@Mousius PTAL. thanks!

mehrdadh:
@tvm-bot rerun

tests/python/relay/aot/test_cpp_aot.py
lhutton1: Thanks, will follow the discussion in #13063

lhutton1 (approved):

Mousius (approved):

PR #13050, opened 2022-10-12
[MetaSchedule] Consolidate module hashing and equality testing

masahi:
Currently, MS uses `StructuralEqual/Hash` in task extraction / evo search / database. Sometimes, we want to use different hashing and equality testing methods, for example (1) to ignore NDArray (https://github.com/apache/tvm/pull/12706) or (2) to enable anchor-op only tuning (identify `conv2d` and `conv2d -> add` subgraphs as equal). 

To enable such flexibility, this PR consolidate raw calls to `StructuralEqual/Hash` into one place, which for now is named `ModuleEquality`. Since hashing is also done for equality testing, I think it is appropriate to call the component responsible for hashing / equality test that way. But other suggestions are welcome.

Importantly, task extraction and database are now using the same hashing / equal method based on TIR mod, while previously task extraction was using a cache key-ed on relay mod.    

cc @junrushao @zxybazh @tqchen


src/meta_schedule/database/database.cc
```diff
@@ -25,8 +26,8 @@ namespace meta_schedule {
 
 Workload::Workload(IRModule mod) {
   ObjectPtr<WorkloadNode> n = runtime::make_object<WorkloadNode>();
-  n->shash = tvm::StructuralHash()(mod);
   n->mod = mod;
+  n->shash = ModuleEquality::Create("structural")->Hash(mod);
```
masahi, line 30: Since this constructor is not used anywhere, I'm hard coding the instance here to avoid passing the `mod_eq_name` string. I want to remove this constructor if there is no objection.

src/meta_schedule/database/database.cc
```diff
@@ -59,12 +60,9 @@ Workload Workload::FromJSON(const ObjectRef& json_obj) {
...
       std::string json_mod = Base64Decode(b64_mod);
       mod = Downcast<IRModule>(LoadJSON(json_mod));
+      std::stringstream(str_shash) >> shash;
     }
-    // Verify SHash(mod) == shash
-    shash = tvm::StructuralHash()(mod);
-    String recalc_shash = SHash2Str(shash);
-    CHECK_EQ(recalc_shash, str_shash) << "ValueError: Structural hash changed. Given: " << str_shash
-                                      << "; Recalculated: " << recalc_shash;
+
```
masahi, line 65: Since `ModuleEquality` information is not written to json (`Workload` doesn't have `ModuleEquality` instance), we cannot determine the appropriate hashing method here. So I had to remove this verification.

src/meta_schedule/database/json_database.cc
```diff
@@ -67,6 +68,10 @@ void JSONFileAppendLine(const String& path, const std::string& line) {
 /*! \brief The default database implementation, which mimics two database tables with two files. */
 class JSONDatabaseNode : public DatabaseNode {
  public:
+  explicit JSONDatabaseNode(String mod_eq_name = "structural")
```
masahi, line 71: The default value is required, since we need to be able to default-construct `JSONDatabaseNode` and other database node due to `TVM_REGISTER_NODE_TYPE` macro.

src/meta_schedule/database/json_database.cc
```diff
@@ -67,6 +68,10 @@ void JSONFileAppendLine(const String& path, const std::string& line) {
...
+      : DatabaseNode(mod_eq_name),
+        workloads2idx_(/*bucket_count*/ 0, WorkloadHash(), WorkloadEqual(GetModuleEquality())) {}
```
masahi, line 73: Since `WorkloadEqual` needs to be aware of the module hashing / equality testing being used, we need to construct this `unordered_map` after the `ModuleEquality` instance is constructed in `DatabaseNode` by L72.

src/relay/backend/task_extraction.cc
```diff
@@ -42,26 +46,35 @@ Array<meta_schedule::ExtractedTask> ExtractTask(IRModule mod, Target target,
...
-        return;
-      }
+
       auto [inputs_outputs, constants, fused_name] =
           tec::LowerTECompute(relay_func, target, /*return_inputs=*/true);
+
       if (Optional<tir::PrimFunc> f = tir_converter(inputs_outputs, constants)) {
+        IRModule tir_mod = PrimFuncToIRModule(f.value());
+
+        auto it = cache.find(tir_mod);
```
masahi, line 68: Note that the cache is now key-ed on the tir mod, rather than the relay mod (which `tec::CCacheKey` uses).

src/meta_schedule/database/database.cc
```diff
@@ -59,12 +60,8 @@ Workload Workload::FromJSON(const ObjectRef& json_obj) {
       String b64_mod = Downcast<String>(json_array->at(1));
       std::string json_mod = Base64Decode(b64_mod);
       mod = Downcast<IRModule>(LoadJSON(json_mod));
+      std::stringstream(str_shash) >> shash;
     }
-    // Verify SHash(mod) == shash
-    shash = tvm::StructuralHash()(mod);
-    String recalc_shash = SHash2Str(shash);
-    CHECK_EQ(recalc_shash, str_shash) << "ValueError: Structural hash changed. Given: " << str_shash
-                                      << "; Recalculated: " << recalc_shash;
```
masahi, line 67: Since ModuleEquality information is not written to json (Workload doesn't have ModuleEquality instance), we cannot determine the appropriate hashing method here. So I had to remove this verification.

junrushao: Perhaps we could leave a TODO here in case of future changes

src/meta_schedule/database/json_database.cc
junrushao: Is it possible if we bypass it via some tricks like unique ptr? (Sorry I don’t think very deeply so feel free to object)

junrushao (approved):
Otherwise LGTM!

include/tvm/meta_schedule/database.h
```diff
@@ -168,8 +176,16 @@ class TuningRecord : public runtime::ObjectRef {
...
+  /*!
+   * \brief Constructor
+   * \param mod_eq_name A string to specify the module equality testing and hashing method.
+   *  It must be one of the followings:
+   *    - "structural": Use StructuralEqual/Hash
+   */
+  explicit DatabaseNode(String mod_eq_name = "structural");
+
   /*! \brief Default destructor */
-  virtual ~DatabaseNode() = default;
```
zxybazh, line 172: May I ask why this is removed to `database.cc` instead?

zxybazh (commented):
Thanks for sending the new feature! It's very exciting to see such improvements!

Generally it's looking good. I have some minor suggestions over the `Workload` usage and question towards the hash map of workloads.

Other than these, currently I can see the only option of module equality functions defined on the c++ side, do you think we can also allow easier customization of the hash and equal function via providing FFI functions or by overridding the class from python side (i.e., providing a `PyModuleEquality` class)?

src/relay/backend/task_extraction.cc
masahi: Yeah, when people see `std::unordered_map<IRModule, ExtractedTask> cache`, it's not obvious if `IRModule` is supposed to be relay or TIR mod...

src/meta_schedule/database/database.cc
masahi: To keep things clean, I want to minimize the number of unique `ModuleEquality` instances scattered throughout MS infra. Currently, only task extraction and `Database` "own" a `ModuleEquality` instance.

Given this goal, I think having `Database` compute a hash of `Workload` using its `ModuleEquality` is more natural than passing `mod_eq_name` to `Workload` and let it compute hash. Moreover, all `Workload` would end up having the same `mod_eq_name`, so it feels redundant.

masahi: One way to restore hash verification is to do verification inside `DatabaseNode` after `Workload::FromJSON` is called, since the database does know what the `ModuleEquality` instance is supposed to be.

src/meta_schedule/database/database.cc
```diff
@@ -25,8 +26,8 @@ namespace meta_schedule {
 
 Workload::Workload(IRModule mod) {
```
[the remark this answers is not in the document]
masahi: This constructor is not used anywhere in practice, so I want to delete this. I explained why I don't want to add `mod_eq_name` to `Workload` above.

include/tvm/meta_schedule/database.h
masahi: This is necessary to be able to hold `std::unique_ptr<ModuleEquality>` in `DatabaseNode` where `ModuleEquality` is forward declared. You can google "unique_ptr incomplete type".

src/relay/backend/task_extraction.cc
```diff
@@ -42,26 +46,35 @@ Array<meta_schedule::ExtractedTask> ExtractTask(IRModule mod, Target target,
...
+          it->second->weight += 1;
+          return;
+        }
+
         IRModule relay_mod({{GlobalVar(fused_name), relay_func}});
-        ExtractedTask task(fused_name, relay_mod, target, {PrimFuncToIRModule(f.value())}, 1);
+        ExtractedTask task(fused_name, relay_mod, target, {tir_mod}, 1);
         tasks.push_back(task);
-        cache.emplace(cache_key, task);
+        cache.emplace(tir_mod, task);
```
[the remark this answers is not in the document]
masahi: This is a great point. In my branch where I'm working on anchor block tuning, I do need to be careful here. When there are both `conv2d` and `conv2d + add`, we need to extract `conv2d`. To do that, before I put modules into the cache, I sort them based on the fused function name, so that `conv2d` comes before `conv2d + add`.

masahi:
>  currently I can see the only option of module equality functions defined on the c++ side, do you think we can also allow easier customization of the hash and equal function via providing FFI functions or by overridding the class from python side (i.e., providing a PyModuleEquality class)?

This can be considered in future, but I don't expect that people would want to add a custom `ModuleEquality` often. So for now, `ModuleEquality` is not exposed to python, this lets me spare from boilerplates and make `ModuleEquality` creation as simple as
```
std::unique_ptr<ModuleEquality> ModuleEquality::Create(const std::string& mod_eq_name) {
  if (mod_eq_name == "structural") {
    return std::make_unique<ModuleEqualityStructural>();
  }
  LOG(FATAL) << "Unknown module equality " << mod_eq_name;
  return nullptr;
}
```

src/meta_schedule/database/json_database.cc
masahi: I didn't get what you meant here.

src/meta_schedule/database/database.cc
masahi: Restored hash verification in `JSONDatabase`, the only place where `Workload::FromJSON` is used currently.

src/meta_schedule/database/json_database.cc
junrushao: just noticed that `DatabaseNode::mod_eq_` has been already a `unique_ptr` - so we don't need an explicit default constructor `DatabaseNode::DatabaseNode`, but instead we could do inline initialization:

```diff
-  std::unique_ptr<ModuleEquality> mod_eq_;
+  std::unique_ptr<ModuleEquality> mod_eq_{nullptr};
```

BTW, just to nitpick, shall we change `mod_eq_` from `private` to `protected` so it could be accessed by derived classes?

masahi: Since `JSONDatabaseNode` is constructed by `Database` class, 
https://github.com/apache/tvm/blob/eecb7fd494052ca941f3d123daa2e887f14b7e75/src/meta_schedule/database/json_database.cc#L147-L150

and `Database` is not a subclass of `DatabaseNode`, setting `mod_eq` requires making it a public member of introduce a setter method in `DatabaseNode`. 

Moreover, it would also introduce complication to initializing `workloads2idx_` (see https://github.com/apache/tvm/pull/13050#discussion_r992887025).

masahi: Just tried: Removing the construction of `workloads2idx_`, `workloads2idx_(/*bucket_count*/ 0, WorkloadHash(), WorkloadEqual(GetModuleEquality()))` leads to compilation failure since `workloads2idx_` cannot be default-constructed (due to `WorkloadEqual` which requires `mod_eq` to be initialized).

zxybazh (approved):
Thanks @masahi for addressing my questions and adding the verification, the PR looks good to me! Looking forward to the introduction of new module equality functions!

junrushao (approved):
LGTM

PR #13054, opened 2022-10-12
[TIR][FIX] Fix crash when using 'if' without 'else' in TVMScript

WuSiYu:
This commit fix the tvm's crash during auto-tune when having a 'if' statement which doesn't have a 'else' part in TVMScript.
It's cause in tvm's tir processing, in `VisitStmt_()` for IfThenElseNode its trying to visit the 'else' part (in TVMScript) whether it is exist.
So in this commit, an if logic is added in `VisitStmt_()` for IfThenElseNode, it will not visit the 'else' part if it is not exist.

### DIFF
--- a/src/tir/analysis/estimate_flops.cc
+++ b/src/tir/analysis/estimate_flops.cc
@@ -148,7 +148,11 @@ class FlopEstimator : private ExprFunctor<TResult(const PrimExpr& n)>,
 
   TResult VisitStmt_(const IfThenElseNode* branch) override {
     TResult cond = VisitExpr(branch->condition);
-    cond += VisitStmt(branch->then_case).MaxWith(VisitStmt(branch->else_case));
+    if (branch->else_case.defined()) {
+      cond += VisitStmt(branch->then_case).MaxWith(VisitStmt(branch->else_case));
+    } else {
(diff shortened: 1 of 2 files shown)


shingjan (changes requested):
Thanks for sending this! Looks fine to me. Would that be possible that some tests can be added to this PR so we will know if future changes will break this?

junrushao (approved):
LGTM

Hzfengsy (approved):

WuSiYu:
> Would that be possible that some tests can be added to this PR so we will know if future changes will break this?

Yes, a test has been added in test_tir_analysis_estimate_tir_flops.py

shingjan (approved):
Thanks for adding the test!

WuSiYu:
@tvm-bot rerun

WuSiYu:
I am not sure but it seem's like the CI is failed by the other ramdom reason, should I continue to try to rerun it?

shingjan:
@WuSiYu It seems that the `@tvm-bot rerun` doesn't go through for some reason. The last time CI failed for this PR is for this [test](https://github.com/apache/tvm/issues/13089), which has been reported. Feel free to retrigger the CI.

PR #13056, opened 2022-10-12
[Relay] Fix simplifying consecutive casts when intermediate type has fewer bits

### LINKED ISSUE #13048: [Bug] Wrong results of `cast<int32>( cast<bool>(-1i64) )`
### Expected behavior

`cast<int32>( cast<bool>(-1i64) )` should be `1`.

### Actual behavior

`cast<int32>( cast<bool>(-1i64) )` returns wrong results (i.e., `-1`) in optimized graph runtime.

### Environment

https://github.com/apache/tvm/commit/fa17da22c73fb9e95c27e4c28130835b628caf6b on Ubuntu 20.04.

### Steps to reproduce

Preferably a minimal script to cause the issue to occur.

```python
import tvm
from tvm import relay
import numpy as np

i0 = relay.var("i0", shape=(1,), dtype="int64")
cast_bool = relay.cast(i0, "bool")
cast_int32 = relay.cast(cast_bool, "int32")

f = relay.Function([i0], cast_int32)
input = np.array([-1], dtype="int64")
opt_res = relay.create_executor("graph", device=tvm.cpu()).evaluate(f)(input)
debug_res = relay.create_executor("debug", device=tvm.cpu()).evaluate(f)(input)
print(f"debug_res: \t{debug_res}") # [-1]
print(f"opt_res: \t{opt_res}")     # [1]
```

### Triage

Please refer to the list of label tags linked above to find the relevant tags and add them here in a bullet format (example below).

* needs-triage

vinx13:
Fixed #13048 

cc @AndrewZhaoLuo @ganler

### DIFF
--- a/src/relay/transforms/simplify_expr.cc
+++ b/src/relay/transforms/simplify_expr.cc
@@ -146,7 +146,7 @@ class SimplifyConsecutiveCast : public DFPatternRewrite {
       // BFloat cast cannot be omitted
       return false;
     }
-    if (origin.code() < cast.code()) {
+    if (origin.code() < cast.code() && origin.bits() <= cast.bits()) {
       // Loosely have a hiearchy to datatypes
       // e.g. int --> uint --> float has increasing range of numbers they can represent
       return true;
--- a/tests/python/relay/aot/test_crt_aot_usmp.py
+++ b/tests/python/relay/aot/test_crt_aot_usmp.py
@@ -303,7 +303,7 @@ MOBILENET_V2_URL = (
     "model_url, usmp_algo, workspace_size, constant_size",
     [
         (MOBILENET_V1_URL, "greedy_by_size", 4845696, 8468008),
-        (MOBILENET_V1_URL, "greedy_by_conflicts", 4444288, 8468008),
+        (MOBILENET_V1_URL, "greedy_by_conflicts", 4845696, 8468008),
         (MOBILENET_V1_URL, "hill_climb", 3240064, 8468008),
     ],
 )
--- a/tests/python/relay/test_pass_simplify_expr.py
+++ b/tests/python/relay/test_pass_simplify_expr.py
@@ -445,6 +445,13 @@ def test_simplify_consecutive_cast():
     expected = run_infer_type(expr2)
     assert tvm.ir.structural_equal(actual, expected)
 
+    x = relay.var("x", shape=(3, 4), dtype="int64")
+    expr1 = relay.cast(x, "bool")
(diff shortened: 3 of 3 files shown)


ganler (approved):
LGTM!

src/relay/transforms/simplify_expr.cc
```diff
@@ -146,7 +146,7 @@ class SimplifyConsecutiveCast : public DFPatternRewrite {
       // BFloat cast cannot be omitted
       return false;
     }
-    if (origin.code() < cast.code()) {
+    if (origin.code() < cast.code() && origin.bits() <= cast.bits()) {
```
AndrewZhaoLuo, line 149: More general solution would be to calculate range of values dtypes represent. If moving to new dtype has range which is subset of old range it is a widen cast.

This perhaps is too harsh as it would block casting with int64 --> float32 or int32 --> float16 which may be relevant for quantization.

AndrewZhaoLuo (commented):
While this solves the issue technically, I think bool should just be a special case of how the semantics work.

Casting bool if I understand correctly makes 0 --> 0 and everything else --> 1.  It has no mapping back to the real numbers whereas all other types (int8, float32) due have a mapping to real numbers. 

Therefore I believe the correct change is simply block this cast simplification for bool.

src/relay/transforms/simplify_expr.cc
vinx13: That make sense. Shall we also block conversions like int64 -> uint32? So the condition will be only checking bits when the cast is between int/uint

AndrewZhaoLuo: Hmm, Reading the comment of hierarchy int --> uint --> float does not make sense to me as uint can not represent negative numbers so it should be lowest on hiearchy in a sense.

I think this simplification pass is a bit hacky or at least it's assumptions are not super clear.  int --> uint hierarchy makes sense as long as we assume all uint casts are done on int types we know are only positive, or the cast clips negatives values to 0 and that is what we want. 

Not sure.

AndrewZhaoLuo: If the point of this pass is making sure the output has the exact same output as old IRModule but with some casts remove, it actually doesn't do that job.

AndrewZhaoLuo: But I think it's a bigger issue. I think PR is at least slightly better and safer than before so will approve. Will make a reminder to open issue here.

AndrewZhaoLuo (approved):

PR #13058, opened 2022-10-12
[ci] Lint for trailing newlines and spaces 

driazati:
This adds a short lint to ensure that all files have a single trailing newline and no trailing whitespaces. This PR is in two commits, one to add the check and another to fix currently offending files in the repo. See https://github.com/apache/tvm/pull/13058/commits/ba2c2e235e2a16d62fdeed959044c65012d9f942 for just the significant changes. Auto-corrections applied with

```
pre-commit run --all-files
```


.pre-commit-config.yaml
```diff
@@ -32,55 +32,55 @@
...
-            pass_filenames: false
-        -   id: run-mypy
-            name: Type Checking with MyPY ...
-            entry: docker/lint.sh mypy
-            language: system
-            always_run: true
-            pass_filenames: false
+        - id: end-of-file-fixer
+          # stages: [push]
+    # - repo: local
```
echuraev, line 49: Why did you comment all hooks?

ci/scripts/git_skip_ci_globs.py
```diff
@@ -35,6 +35,8 @@
     # microTVM
     "apps/microtvm/poetry.lock",
     "apps/microtvm/pyproject.toml",
+    "tests/lint/*",
+    "tests/scripts/task_lint.sh",
```
echuraev, line 39: Why you added these files to the `globs`?

echuraev (commented):
Thank you for this PR. It's very good checking. Sorry, I'm not a fully understand some of the changes and I asked questions about them.

ci/scripts/git_skip_ci_globs.py
driazati: These are used to determine in Jenkins whether to run the full suite of tests or not, so if all the files in a PR match one of these globs then unit tests won't be run (lint is always run though). These files only affect lint so they don't need to run tests

echuraev: Got it. Thank you! I just thought that lint also won't be run. And I was confused about that.

echuraev (approved):
LGTM. Thanks

tests/lint/trailing_newlines.py
```diff
@@ -0,0 +1,51 @@
...
+        elif start_bytes == 1:
+            # 1 byte file
+            return False
+        else:
+            # skip to the end
+            f.seek(-2, os.SEEK_END)
+            end_bytes = f.read(2)
+
+            # should be a non-newline followed by a newline
+            return end_bytes[0] != ord("\n") and end_bytes[1] == ord("\n")
```
areusch, line 41: can this be `end_bytes != b'\n\n'` since you're opening file in binary mode?

areusch (approved):

tests/lint/trailing_newlines.py
driazati: i don’t think so since that would match things like `b’aa’`

Issue #10950, opened 2022-04-08
[Tracking Issue] Handle ONNX contrib operators


AndrewZhaoLuo:
Right now ONNX converter to Relay works by using operator name to match up with proper converter. 

The issue is we increasingly have to deal with operators from different onnx modules/domains. Onnx nodes also have this sort of namespace https://github.com/onnx/onnx/blob/main/docs/IR.md#nodes too. In the future, we might have name collisions between like names in different namespaces. We should handle this properly.

Not a big deal until it happens.

### List of operators not under default onnx operator domain (please add):
- [ ] Gelu (https://com.microsoft/)
- [ ] EmbedLayerNormalization (https://com.microsoft/)
- [ ] SkipLayerNormalization (https://com.microsoft/)
- [ ] Attention (https://com.microsoft/)

PR #13065, opened 2022-10-13
[ONNX] Handle multiple imports

AndrewZhaoLuo:
Ref: https://github.com/onnx/onnx/blob/main/docs/IR.md

Right now we take the first imported op set as the operator version for rest of conversion. This assumes the first imported op set is the default (ai.onnx), however this is not necessarily the case.

In the future, we need to support multiple operator sets well (see https://github.com/apache/tvm/issues/10950). For now, we just try to find the default (ai.onnx) namespace and use that for the opset version.

### DIFF
--- a/python/tvm/relay/frontend/onnx.py
+++ b/python/tvm/relay/frontend/onnx.py
@@ -5906,7 +5906,16 @@ def from_onnx(
     graph = model.graph
 
     try:
-        opset_in_model = model.opset_import[0].version if model.opset_import else 1
+        opset_in_model = 1
+        if model.opset_import:
+            # TODO: for now we only really support ai.onnx op set
+            # TODO: handle other namespaces well see https://github.com/apache/tvm/issues/10950
+            for opset_identifier in model.opset_import:
+                # As per https://github.com/onnx/onnx/blob/main/docs/IR.md
+                # All operator sets except the default one must specify the operator version
+                if str(opset_identifier.name) in ["ai.onnx", ""]:
+                    opset_in_model = opset_identifier.version
+                    break
     except AttributeError:
         opset_in_model = 1
 
--- a/python/tvm/relay/frontend/onnx.py
+++ b/python/tvm/relay/frontend/onnx.py
@@ -5913,7 +5913,7 @@ def from_onnx(
             for opset_identifier in model.opset_import:
                 # As per https://github.com/onnx/onnx/blob/main/docs/IR.md
                 # All operator sets except the default one must specify the operator version
-                if str(opset_identifier.name) in ["ai.onnx", ""]:
+                if str(opset_identifier.domain) in ["ai.onnx", ""]:
                     opset_in_model = opset_identifier.version
                     break
     except AttributeError:


AndrewZhaoLuo:
cc @sfvaroglu

AndrewZhaoLuo:
@jwfromm

jwfromm (approved):
This is a good starting point, thanks @AndrewZhaoLuo!

python/tvm/relay/frontend/onnx.py
```diff
@@ -5906,7 +5906,16 @@ def from_onnx(
...
     try:
-        opset_in_model = model.opset_import[0].version if model.opset_import else 1
+        opset_in_model = 1
+        if model.opset_import:
+            # TODO: for now we only really support ai.onnx op set
+            # TODO: handle other namespaces well see https://github.com/apache/tvm/issues/10950
+            for opset_identifier in model.opset_import:
+                # As per https://github.com/onnx/onnx/blob/main/docs/IR.md
+                # All operator sets except the default one must specify the operator version
+                if str(opset_identifier.name) in ["ai.onnx", ""]:
```
jwfromm, line 5916: I'm not sure if its inconsistent but the models I'm looking at use `opset_identifer.domain` rather than `name`. Also, it seems like if there is only one import there is no domain name, just the version and its assumed the domain is ai.onnx. This is at least the case for `resnet50-v1-7` which is used in the tests.

jwfromm (changes requested):

python/tvm/relay/frontend/onnx.py
AndrewZhaoLuo: It should be domai yeah, fixed.

python/tvm/relay/frontend/onnx.py
```diff
@@ -5906,7 +5906,16 @@ def from_onnx(
...
     try:
-        opset_in_model = model.opset_import[0].version if model.opset_import else 1
+        opset_in_model = 1
+        if model.opset_import:
+            # TODO: for now we only really support ai.onnx op set
+            # TODO: handle other namespaces well see https://github.com/apache/tvm/issues/10950
+            for opset_identifier in model.opset_import:
+                # As per https://github.com/onnx/onnx/blob/main/docs/IR.md
+                # All operator sets except the default one must specify the operator version
+                if str(opset_identifier.domain) in ["ai.onnx", ""]:
```
jwfromm, line 5916: I think this will still run into some issues. When there is only one `opset_import`, it doesnt have a domain name, just a version. We'll need to add a special case when the length of `model.opset_import` is 1 and just pull out the version directly.

AndrewZhaoLuo: Do you mean opset_identifier won't have the attribute "domain", or that it will return the empty string?

AndrewZhaoLuo: It could be an earlier ONNX spec which we do not support.

jwfromm: It returns an empty string despite having a version that needs to be respected. If we dont handle this case we'll fail the `import_model` test.

AndrewZhaoLuo: Hmm in the code, it should be fine with the "", I'll investigate this test.

AndrewZhaoLuo:
@tvm-bot rerun

python/tvm/relay/frontend/onnx.py
AndrewZhaoLuo: The test fail was due to known error https://github.com/apache/tvm/pull/13067.

AndrewZhaoLuo:
PTAL @jwfromm

jwfromm (approved):

jwfromm:
LGTM, thank you @AndrewZhaoLuo

PR #13066, opened 2022-10-13
[Hexagon] [runtime] VTCM bugfix, runtime buffer clarification

janetsc:
Quick bugfix and clarification for runtime buffers

### DIFF
--- a/src/runtime/hexagon/hexagon_vtcm_pool.cc
+++ b/src/runtime/hexagon/hexagon_vtcm_pool.cc
@@ -69,12 +69,16 @@ void* HexagonVtcmPool::Allocate(size_t nbytes) {
   // If this is not aligned on a 2k block, allocate from the end to avoid fragmentation
   if (nbytes & size_t(0x7FF)) {
     DLOG(INFO) << "VTCM nbytes requested: " << nbytes << " allocate from the end";
-    auto last_free_entry = free_.rbegin();
+    auto last_free_entry = free_.end();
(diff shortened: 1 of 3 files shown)


mehrdadh (commented):
LGTM, I'll wait for @adstraw to take a look as well

src/runtime/hexagon/hexagon_vtcm_pool.cc
```diff
@@ -69,12 +69,16 @@ void* HexagonVtcmPool::Allocate(size_t nbytes) {
   // If this is not aligned on a 2k block, allocate from the end to avoid fragmentation
   if (nbytes & size_t(0x7FF)) {
     DLOG(INFO) << "VTCM nbytes requested: " << nbytes << " allocate from the end";
-    auto last_free_entry = free_.rbegin();
+    auto last_free_entry = free_.end();
+    last_free_entry--;
```
adstraw, line 73: I think this would be more clear, but it's a NIT:
```
std::vector<std::pair<char*, size_t>>::iterator last_free_entry = free_.back();
```

adstraw (approved):

src/runtime/hexagon/hexagon_vtcm_pool.cc
adstraw: Never mind, this doesn't work.  Resolving. :)

AndrewZhaoLuo (approved):

PR #13073, opened 2022-10-14
[microTVM] enable building microTVM components by default

alanmacd:
- Enable building microTVM components by default in build except for Hexagon and iOS.
- Disable USE_MICRO for CPU minimal pending addition change to include StandaloneCRT with CPU minimal unit tests.


elvin-n:
@tvm-bot rerun

mehrdadh:
Looks like there's only one issue withe hexagon build with crt_standalone. Great progress!

CMakeLists.txt
```diff
@@ -58,7 +58,6 @@ tvm_option(USE_OPENMP "Build with OpenMP thread pool implementation" OFF)
 tvm_option(USE_RELAY_DEBUG "Building Relay in debug mode..." OFF)
 tvm_option(USE_RTTI "Build with RTTI" ON)
 tvm_option(USE_MSVC_MT "Build with MT" OFF)
-tvm_option(USE_MICRO "Build with Micro TVM support" OFF)
```
mehrdadh, line 61: Too keep the `USE_MICRO` option, we can have it in the CMakelist and make it default ON. Then we can remove it from `cmake/config.cmake` file

driazati (commented):
just noting something we discussed offline, but the CI failures in https://ci.tlcpack.ai/blue/organizations/jenkins/tvm/detail/PR-13073/13/pipeline/542/#step-1557-log-1949 are difficult to fix in this PR due to Jenkins issues, so we should disable them for the `CPU: minimal` build (which has the env var `PLATFORM` set to `minimal`) temporarily, then I can follow up with a fix to re-enable them

CMakeLists.txt
alanmacd: thanks, done.

mehrdadh (approved):
LGTM, thanks!

srkreddy1238:
Cross compilations fails my making microtvm default.

echuraev:
@alanmacd, @mehrdadh could you please clarify why we have enabled building microTVM by default?

areusch:
ah @srkreddy1238 @echuraev sorry for the breakage. just to document the intent here: assuming you already compile with `USE_RPC ON`, the actual net effect of successfully compiling with `USE_MICRO ON` is to add a minimal transport layer. I think these breakages are likely due to `standalone_crt` being built via a Make-based system rather than cmake.  

let's revert this one for now, and we'll retry at a later time with a cmake-based standalone_crt build, which should properly pick up `CXX` from the config.cmake.

Issue #12977, opened 2022-10-03
[Bug] FQ2I Pass May Duplicates Operations


AndrewZhaoLuo:
FQ2I can duplicate operations. More specifically, if an operation dominates the quantize in two QDQ regions, the operation can be duplicated.

### Expected behavior

See this [script](https://github.com/AndrewZhaoLuo/TVM-Sandbox/blob/main/quantization/issues_with_fq2i_pass.py#L93)  (use layers=1) where we want to convert the following fake quantized graph:

```
def @main(%input_var: Tensor[(1, 512, 512), float32] /* ty=Tensor[(1, 512, 512), float32] */) -> Tensor[(1, 512, 512), int8] {
  %0 = qnn.quantize(%input_var, 0.025f /* ty=float32 */, 1 /* ty=int32 */, out_dtype="int8") /* ty=Tensor[(1, 512, 512), int8] */;
  %1 = qnn.dequantize(%0, 0.025f /* ty=float32 */, 1 /* ty=int32 */) /* ty=Tensor[(1, 512, 512), float32] */;
  %2 = mean(%1, axis=[-1], keepdims=True) /* ty=Tensor[(1, 512, 1), float32] */;
  %3 = subtract(%2, %1) /* ty=Tensor[(1, 512, 512), float32] */;
  %4 = sqrt(%3) /* ty=Tensor[(1, 512, 512), float32] */;
  %5 = qnn.quantize(%4, 0.025f /* ty=float32 */, 1 /* ty=int32 */, out_dtype="int8") /* ty=Tensor[(1, 512, 512), int8] */;
  %6 = qnn.dequantize(%5, 0.025f /* ty=float32 */, 1 /* ty=int32 */) /* ty=Tensor[(1, 512, 512), float32] */;
  %7 = add(%6, %4) /* ty=Tensor[(1, 512, 512), float32] */;
  qnn.quantize(%7, 0.025f /* ty=float32 */, 1 /* ty=int32 */, out_dtype="int8") /* ty=Tensor[(1, 512, 512), int8] */
}
```
We expect a graph like:
```
def @main(%input_var: Tensor[(1, 512, 512), float32] /* ty=Tensor[(1, 512, 512), float32] */) -> Tensor[(1, 512, 512), int8] {
  %0 = qnn.quantize(%input_var, 0.025f /* ty=float32 */, 1 /* ty=int32 */, out_dtype="int8") /* ty=Tensor[(1, 512, 512), int8] */;
[...]

PR #13074, opened 2022-10-14
[Relay][Frontend][ONNX] Add LayerNormalization operator

ehsanmok:
This PR adds the LayerNormalization operator for relay onnx frontend that was added in onnx in opset 17 [here](https://github.com/onnx/onnx/blob/main/docs/Changelog.md#layernormalization-17).

@sfvaroglu @AndrewZhaoLuo


python/tvm/relay/frontend/onnx.py
```diff
@@ -944,6 +946,36 @@ def _impl_v1(cls, inputs, attr, params):
         return Gelu._impl_v1([inp], attr, params)
 
 
+class LayerNormalization(OnnxOpConverter):
```
sfvaroglu, line 949: Is there a reason for not using `nn.layer_norm`?

ehsanmok: The main reason is `nn.layer_norm` doesn't output the intermediate expr such as `mean` and `inv_stdev` needed for onnx output which are about half of the computation. Besides, I first tried with the existing utility `layer_norm` in onnx.py used in `EmbedLayerNormalization` and `SkipLayerNormalization` but that wasn't correct for the `LayerNormalization` because of the wide axes setup so had to reimplement based on onnx spec. I think there should be refactoring opportunity that I'll make a note for later.

AndrewZhaoLuo: Do the tests themselves test for these outputs? In the spec they are listed as optional

LayerNorm, much like BatchNorm has different behavior between training and inference time. As ONNX frontend is pretty much for inference, I would say it's ok to drop the other two outputs if it just simplifies the code a lot.

python/tvm/relay/frontend/onnx.py
```diff
@@ -944,6 +946,36 @@ def _impl_v1(cls, inputs, attr, params):
...
+        axis = attr.get("axis", -1)
+        eps = attr.get("epsilon", 1e-5)
+        # according to the onnx doc, given the int axis (default -1)
+        # to compute the mean and inv_stdev which are of dim [d[0], ..., d[axis-1], 1, ..., 1]
+        # the actual computation is over (axis, ..., rank(x) - 1) axes
+        # see https://github.com/onnx/onnx/blob/main/docs/Changelog.md#layernormalization-17
+        rank = len(infer_shape(x))
+        axis = tuple(range(axis, rank)) if axis >= 0 else tuple(range(rank + axis, rank))
+        dtype = infer_type(x).checked_type.dtype
+        mean = _op.mean(x, axis, keepdims=True)
```
AndrewZhaoLuo, line 966: I believe this is inefficient as mean will do a reduction along the axis and variance will do two reductions (one to find mean, one to find the mean of (X - E[X])^2. Giving three total reductions across the axis.

Instead just manually do E[(X - E[X])^2] to get two total reductions.

AndrewZhaoLuo (commented):
Thanks @ehsanmok! A few comments

python/tvm/relay/frontend/onnx.py
ehsanmok: Yes the tests, test for these outputs and can't be set as a placeholder

python/tvm/relay/frontend/onnx.py
ehsanmok: Not sure if I understand completely! the test cases test for actual match between all outputs (with dims included) and the "wide" axes takes care of over what axes the mean and variance are computed which conform to what onnx spec is referring [here](https://github.com/onnx/onnx/blob/main/docs/Changelog.md#layernormalization-17) where 

> the shape of Mean and InvStdDev is `[d[0], ..., d[axis-1], 1, ..., 1]`

AndrewZhaoLuo: If we look at `python/tvm/relay/op/reduce.py` we see _op.variance calls `mean` which is superfluous. You can probably change it by giving it an option to pass in the mean.

ehsanmok: Isn't that (the point of rely) gets optimized over the simplification passes that rely does so we don't need to worry too much about it?

AndrewZhaoLuo: Yes, common subexpression elimination pass should eliminate this. However, CSE can't do magic, and has limitations. You can probably check if it will be removed here by running this generated relay through the pass. Perhaps it will. 

In general, unless it adds a great amount of overhead in the reasoning of a piece of code, I would not duplicate ops however.

There are cases where it should but does not: https://github.com/apache/tvm/issues/12977

ehsanmok: Comment addressed. Please check again. Thanks!

AndrewZhaoLuo (approved):

AndrewZhaoLuo:
@ehsanmok lgtm! Just fix the linting and make sure tests are passing.

PR #12911, opened 2022-09-26
[Hexagon] vrmpy tensorization for e2e compilation of int8 models

masahi:
This PR adds TE compute and schedule definitions for int8 conv2d and dense using vrmpy tensorization, and Relay alter layout / legalize to enable using them in e2e settings. Since `vrmpy` is very similar to x86 VNNI or ARM `sdot/udot` instructions, lots of code are shared with existing x86 / ARM backend implementations. 

This lets us run int8 resnet50 in 146 msec on SD888. All convolutions and the final dense op are tensorized. The current bottleneck is `requantize`-related operations. The test script and model files to run int8 resnet50 are attached below.

[test_qresnet50.zip](https://github.com/apache/tvm/files/9651888/test_qresnet50.zip)

@kparzysz-quic @tkonolige @nverke @ibsidorenko

PR #13080, opened 2022-10-14
[Relay][Hexagon] Added per-channel FixedPointMultiply operation.

ibsidorenko:
Main goal of this commit is to improve performance for Hexagon target and preserve performance/accuracy for x86, GPU and etc. targets.

"qnn.requantize" operation is lowered into the sequence of multiply, add, shift during QNN canonicalization pass if scale quantization parameter is vector of scalars. This commit adds new Relay per-channel/per-axis FixedPointMultiply operation and is used in "qnn.requantize" operation lowering.

per-channel/per-axis FixedPointMultiply is implemented through tir.q_multiply_shift_per_axis intrinsic. For Hexagon target it overrides default implementation and generate HVX vmpye/vmpyo instruction (see _q_multiply_shift_per_axis_hexagon). For all other targets it uses default implementation (64 bits arithmetic).

**Performance/accuracy measurement:**

* _CPU(x86) target_: accuracy and performance are the same. For other targets should be the same (otherwise it is bug).

* _Hexagon target_: speedup of qnn.requantize 7x-9x times (Snapdragon 888, 4.4 ms -> 0.5 ms) Unfortunately, in some cases output of requantize can differ by "1" compared to reference output. We need to consider this tradeoff between performance and accuracy when we choose this optimized implementation.


masahi:
@tvm-bot rerun

python/tvm/topi/hexagon/tensor_intrin.py
```diff
@@ -73,6 +73,58 @@ def _q_multiply_shift_hexagon(op):
...
+def _q_multiply_shift_per_axis_hexagon(op):
+    """
+    Implementation of q_multiply_shift_per_axis through hexagon intrinsics vmpyewuh and vmpyowh when
+    q == 31.
+
+    Please note that this is introducing a small round-up error for some corner cases when we do
+    right shift. This is because we are rounding twice instead than only once. I.e.:
+
+        * original q_multiply_shift: round(x*y*2^-s)
+        * hexagon q_multiply_shift: round(round(x*y)*2^-s)
```
masahi, line 85: cc @kparzysz-quic @jverma-quic on this HVX implementation.

ibsidorenko: I will add test to demonstrate issue with accuracy.

ibsidorenko: I have fixed issue with accuracy drop.
For the case when we need both right and left shifts at the same time I use "old" approach and lower this oper to the sequence left_shift/multipy/add/right_shift (64bit arithmetic). Right now I have no idea how to implement this case through vector HVX instructions without accuracy drop.

ibsidorenko:
@tvm-bot rerun

src/target/intrin_rule.cc
```diff
@@ -194,40 +234,34 @@ TVM_REGISTER_OP("tir.q_multiply_shift")
...
+      const tir::CallNode* call = e.as<tir::CallNode>();
+      ICHECK(call != nullptr);
+
+      PrimExpr x = call->args[0];
+      PrimExpr y = call->args[1];
+      PrimExpr left_shift = call->args[2];
+      PrimExpr right_shift = call->args[3];
+      PrimExpr q = call->args[4];
+      PrimExpr is_lshift_required = call->args[5];
+      // Note, 7th argument is "is_rshift_required" flag, but we do need that here.
```
masahi, line 260: You mean "don't need"?

masahi (approved):
@kparzysz-quic This PR improves performance on int8 resnet50 from the PR https://github.com/apache/tvm/pull/12911 while preserving accuracy.

Manual schedules (no tuning): 146 msec (before) -> 92 msec. 
Tuned schedules (vrmpy auto tensorization): 105 msec -> 58 msec.

Very cool!

cc @tmoreau89 @csullivan

src/target/intrin_rule.cc
ibsidorenko: Oh... yes, exactly. My bad, this is typo in comment.

PR #13067, opened 2022-10-13
[TEST] Fix the broken VNNI MetaSchedule test

masahi:
Fix the VNNI test broken by https://github.com/apache/tvm/pull/12895

And I've just ran rocm and vulkan dp4a auto tensorization tests locally and confirmed that they still worked.

@junrushao

PR #13081, opened 2022-10-14
[TIR][Arith] Support negative coeff in ModularSet

Lunderberg:
Prior to this commit, any use of negative coefficients in `ModularSet` would result in an error.  This included cases where a constraint is being entered, such as `floormod(i, -2)==0` appearing as the condition of an if/else block.  These negative indices can also arise as intermediate simplification steps produced by `CanonicalSimplifier`, such as `floormod(-i,2)` being canonicalized to `floormod(i,-2)`.

This commit adds support for negative coefficients in `ModularSet`, using the same sign convention as is used by `CanonicalSimplifier` for negative denominators, and adds unit tests to verify that sign convention.

### DIFF
--- a/tests/python/unittest/test_arith_canonical_simplify.py
+++ b/tests/python/unittest/test_arith_canonical_simplify.py
@@ -97,6 +97,8 @@ def test_split_index_simplify():
     # cannot simplify mixed case, unless we canonicalize into one mode.
     ck.verify(tdiv(x, 6) * 2 + tmod(fld(x, 3), 2), tdiv(x, 6) * 2 + tmod(fld(x, 3), 2))
 
+    ck.verify(tmod(-x, 2), tmod(x, -2) * -1)
+
 
 def test_div_simplify():
     ck = CanonicalChecker()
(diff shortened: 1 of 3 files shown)


Lunderberg:
This was originally discovered in a failing unit test for https://github.com/apache/tvm/pull/13024.  An unrolled loop produced a `T.if_then_else(floormod(-i,2)==0, ...)`, which resulted in an error thrown from ModularSet.

Lunderberg:
Rebasing onto main to get the bug fix from #13067

tests/python/unittest/test_arith_canonical_simplify.py
```diff
@@ -97,6 +97,8 @@ def test_split_index_simplify():
     # cannot simplify mixed case, unless we canonicalize into one mode.
     ck.verify(tdiv(x, 6) * 2 + tmod(fld(x, 3), 2), tdiv(x, 6) * 2 + tmod(fld(x, 3), 2))
 
+    ck.verify(tmod(-x, 2), tmod(x, -2) * -1)
```
vinx13, line 100: does this cover modular set analysis?

Lunderberg: This test only covers the CanonicalSimplify behavior of negative indices, but the ModularSet's behavior should be consistent with the sign convention in CanonicalSimplify.  Since CanonicalSimplify's sign convention for negative indices wasn't checked by any test I could find, I wanted to add a test for its behavior.

jwfromm (approved):
Nice fix, Thanks @Lunderberg!

PR #13082, opened 2022-10-14
[Node] Fix structural equal path tracing pointer usage

vinx13:
This fixed usage of pointer to local variable `tracing_data`.

cc @masahi


masahi (approved):
oops sorry about that. So this code path is not tested at all?

vinx13:
it's tested in https://github.com/apache/tvm/blob/main/tests/python/unittest/test_tir_structural_equal_hash.py but somehow passed CI

PR #13083, opened 2022-10-14
[Hexagon]Register fast softmax schedule with default schedule

mehrdadh:
This PR registers fast softmax schedule for hexagon using the default schedule.
Also renames some schedule names to hexagon.

cc @csullivan


mehrdadh:
cc @masahi

masahi (approved):

Issue #12997, opened 2022-10-06
[Bug][MetaSchedule] Compute-At DataType Mismatch


zxybazh:
I noticed a data type issue that can be triggered when applying compute-at, causing invalid expression in TIR script with incompatible datatype, thus failure in post processors and build.

To reproduce simple run this [script](https://gist.github.com/zxybazh/265b2996e736128eb78c29c94c7a35d3) would work. It can build successfully before schedule but fails with a data type mismatch error after a compute-at scheduling.

cc @Hzfengsy @junrushao1994

masahi:
Does this issue still exist? Your test script doesn't work due to the MS API change.

zxybazh:
Yes, the issue is still there, I changed the script (remove `TuneConfig` import) and it can work now.

masahi:
The PR https://github.com/apache/tvm/pull/13327 works at `CreatePrimFunc` level and would change the generated TVMScript module. So we need to regenerate this TVMScript to verify that the PR fixes this issue.

zxybazh:
Cannot reproduce, close for now.

PR #13084, opened 2022-10-14
[MetaSchedule] Remove Exception Catch for Multithread Postproc Application

zxybazh:
The current datatype issue #12997 could cause the postproc application to fail in massive number during evolutionary search, causing tons of exception logging onto the screen via `LOG(WARN)`. This change removes the postproc application exception logging in the utility of search strategy, the reasons are:
1. The errors are triggered in multi-thread environment, the logging is hard to read anyway
2. We have a large population of candidates to mutate and apply postproc, in scenarios like the data type issue, it generates tons of logging, which is not helpful.
3. To debug we always need to run the postproc alone, which doesn’t matter in this case.
4. Without the warning log, we can still see failed postproc number count in the logging, and check for failure if necessary. If all the application fails, we would be able to see the workload fails in the performance tuning table. Example:
```c++
2022-10-13 17:03:42.383 INFO /home/apache/tvm/src/meta_schedule/search_strategy/evolutionary_search.cc:595 Evolve iter # 3 done. Summary:
Postproc #0 [meta_schedule.DisallowDynamicLoop(0x561b625b8238)]: 0 failure(s)
Postproc #1 [meta_schedule.RewriteParallelVectorizeUnroll(0x561b62606e78)]: 589 failure(s)
Postproc #2 [meta_schedule.RewriteReductionBlock(0x561b6266a358)]: 0 failure(s)
Postproc #3 [meta_schedule.RewriteLayout(0x561b625323b8)]: 0 failure(s)
```
There’s another alternate way I thought about to do this, which is to add a field called `postproc exception` counter that can output after the output of failures like
```c++
2022-10-13 17:03:42.383 INFO /home/apache/tvm/src/meta_schedule/search_strategy/evolutionary_search.cc:595 Evolve iter # 3 done. Summary:
Postproc #0 [meta_schedule.DisallowDynamicLoop(0x561b625b8238)]: 0 failure(s)
Postproc #1 [meta_schedule.RewriteParallelVectorizeUnroll(0x561b62606e78)]: 0 failure(s) 589 exception(s)
Postproc #2 [meta_schedule.RewriteReductionBlock(0x561b6266a358)]: 0 failure(s)
Postproc #3 [meta_schedule.RewriteLayout(0x561b625323b8)]: 0 failure(s)
```
I think this is unnecessary given we can always check failure reasons here and we would always expect postproc to work except verification postprocs.

There's only one other logging interface like this in Postproc implementation [here](https://github.com/apache/tvm/blob/605876e638a7303acb63ef8a9dd9ebfefc30ae24/src/meta_schedule/postproc/rewrite_tensorize.cc#L46) in `RewriteTensorize` post processor. I think this output can potentially cause the same problem, we can address it if we have consensus on the current logging interface change.

CC @junrushao @vinx13 @masahi

### DIFF
--- a/src/meta_schedule/utils.h
+++ b/src/meta_schedule/utils.h
@@ -318,14 +318,16 @@ struct ThreadedTraceApply {
 
     for (int i = 0; i < n_; ++i) {
       Item& item = items_[i];
+      bool applied = false;
       try {
-        if (!item.postproc->Apply(sch)) {
-          ++item.fail_counter;
-          return NullOpt;
+        if (item.postproc->Apply(sch)) {
+          applied = true;
         }
       } catch (const std::exception& e) {
-        // Used in multi-thread, only output to screen but failure summary sent to logging
-        LOG(WARNING) << "ThreadedTraceApply::Apply failed with error " << e.what();
+        // left blank intentionally
+      }
+      if (!applied) {
+        item.fail_counter++;
         return NullOpt;
       }
     }
--- a/src/meta_schedule/utils.h
+++ b/src/meta_schedule/utils.h
@@ -318,15 +318,7 @@ struct ThreadedTraceApply {
 
     for (int i = 0; i < n_; ++i) {
       Item& item = items_[i];
-      bool applied = false;
-      try {
-        if (item.postproc->Apply(sch)) {
-          applied = true;
-        }
-      } catch (const std::exception& e) {
-        // left blank intentionally
-      }
-      if (!applied) {
+      if (!item.postproc->Apply(sch)) {
         item.fail_counter++;
         return NullOpt;
       }


src/meta_schedule/utils.h
```diff
@@ -318,14 +318,16 @@ struct ThreadedTraceApply {
...
-        if (!item.postproc->Apply(sch)) {
-          ++item.fail_counter;
-          return NullOpt;
+        if (item.postproc->Apply(sch)) {
+          applied = true;
         }
       } catch (const std::exception& e) {
-        // Used in multi-thread, only output to screen but failure summary sent to logging
-        LOG(WARNING) << "ThreadedTraceApply::Apply failed with error " << e.what();
+        // left blank intentionally
```
7andrew7, line 327: One problem here is that we have no visibility into failures.

Would it make sense to catch and log a bounded number before giving up and re-raising?

zxybazh: This issue is not caused by a single candidate TIR running multiple times but thousands of them generated by sampler or mutator. They could all be different from each other so only log some number of exceptions is not representative. On the other hand, this is applied in multi-thread, if we keep trace of the logging numbers, it may impact the tuning speed.

Even if we don't log out the exceptions here, we keep track of the failing count, and it's reflected in the log. If the failing count doesn't make sense, it's easy to spot from the log. And we can easily reproduce with running the postprocessor separately once there's any issue. We should be able to see failed tuning of this workload or impacted performance as a result.

One thing we can do to further improve visibility is like I said create a exception counter field in the output, or simply set a check that warns user when there's unexpected postproc failure in the counter. Let me know which one you think is better.

masahi: "Benign" failures should be caught at the postproc level, e.g. this is how the error is handled for auto tensorization https://github.com/apache/tvm/blob/02c2eae510d6d6c15189427c97819f7ce05f002d/src/meta_schedule/postproc/rewrite_tensorize.cc#L43-L46.

Other errors that propagate to `ThreadedTraceApply` level should be treated as a bug in MS. So I think it is ok to remove this log or this try /catch entirely, and let it crash to reveal the error.

zxybazh:
Following up on the discussion, I agree with Masa’s idea of letting it crash if the exception of postproc is unexpected, and let the postproc itself handle the expected exceptions. In this case, I removed the try-catch here instead, and expect any crash of the postproc application to be a bug to look into.

Please take another look on the PR @7andrew7 @masahi , thanks!

7andrew7 (approved):

masahi (approved):

Issue #12442, opened 2022-08-15
[Tracking Issue] TVMScript Metaprogramming


junrushao:
This issue is to track progress for [[RFC-0079](https://github.com/apache/tvm-rfcs/blob/main/rfcs/0079-tvmscript-metaprogramming.md)] TVMScript Metaprogramming.

Main author: @cyx-6 @yongwww.

## A. Fix bugs/issues revealed by the new parser

- A1. TIR operations that are forgotten previously to be exposed to python
  - [x] call_packed_lowered, call_cpacked_lowered, call_llvm_intrin, call_llvm_pure_intrin https://github.com/apache/tvm/pull/12425
  - [x] tvm_load_matrix_sync, tvm_mma_sync, tvm_bmma_sync, tvm_fill_fragment, tvm_store_matrix_sync https://github.com/apache/tvm/pull/12589
  - [x] ptx_mma, ptx_mma_sp, mma_store, mma_fill https://github.com/apache/tvm/pull/12623
  - [x] ptx_ldmatrix, ptx_cp_async, ptx_commit_group, ptx_wait_group https://github.com/apache/tvm/pull/12611
  - [x] tvm_stack_alloca, tvm_stack_make_shape, tvm_stack_make_array https://github.com/apache/tvm/pull/12429
  - [x] tvm_tuple, tvm_struct_get, tvm_struct_set https://github.com/apache/tvm/pull/12452
  - [x] tvm_thread_allreduce, tvm_access_ptr, tvm_throw_last_error, TVMBackendAllocWorkspace, TVMBackendFreeWorkspace https://github.com/apache/tvm/pull/12468
  - [x] shift_left, shift_right https://github.com/apache/tvm/pull/12584
  - [x] vectorlow, vectorhigh, vectorcombine https://github.com/apache/tvm/pull/12571
  - [x] address_of, infinity, reinterpret, lookup_param, isnullptr https://github.com/apache/tvm/pull/12466
  - [x] assume, undef, likely https://github.com/apache/tvm/pull/12435
- [x] A2. Fix precision issue in FFI converting int/float to PrimExpr https://github.com/apache/tvm/pull/12417
- [x] A3. Expose `ScriptComplete` in header https://github.com/apache/tvm/pull/12419
- [x] A4. Conversion `BufferRegion` to vectorized `BufferLoad` https://github.com/apache/tvm/pull/12420
[...]

PR #13088, opened 2022-10-15
[TVMScript] Evaluator, core parser, var table

cyx-6:
This PR introduces Evaluator, core parser, var table as part of #12442

Co-authored-by: yongwww <yongcale@gmail.com>

### DIFF
--- a/python/tvm/script/_parser/core/dispatch.py
+++ b/python/tvm/script/_parser/core/dispatch.py
@@ -1,0 +1,155 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+from typing import TYPE_CHECKING, Any, Callable, Dict, Optional, Tuple, Type
+
+from .doc import AST
+
+if TYPE_CHECKING:
+    from .parser import Parser
+
+
+ParseMethod = Callable[["Parser", AST], None]
+ParseVTable: Dict[Tuple[str, str], ParseMethod] = {}
+
+OpMethod = Callable[..., Any]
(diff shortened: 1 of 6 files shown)


tests/python/unittest/test_tvmscript_parser_evaluator.py
```diff
@@ -0,0 +1,61 @@
...
+# specific language governing permissions and limitations
+# under the License.
+"""Unittests for tvm.script.parser.evaluator"""
+import pytest
+import tvm.testing
+from tvm.script._parser.core.diagnostics import Source
+from tvm.script._parser.core.evaluator import ExprEvaluator
+
+
+def calc(expr, extra_vars={}):
```
junrushao, line 24: 1. Start the name with an underscore
2. Use “None” as default for extra_vars

cyx-6: Fixed

MasterJH5574 (approved):
LGTM. Thanks @cyx-6 for the hard work!

python/tvm/script/_parser/core/dispatch.py
```diff
@@ -0,0 +1,156 @@
...
+OpVTable: Dict[Tuple[Type, AST, int], OpMethod] = {}
+
+
+def register(token: str, type_name: str):
+    """Register a method for a dispatch token and type name.
+
+    Parameters
+    ----------
+    token : str
+        The token for IR, e.g., T for TIR and R for Relax.
```
leandron, line 40: Hi @MasterJH5574 - it seems this PR accidentally were merged some references to Relax in `main` as opposed to `unity` - can you push a follow-up PR and change it to the appropriate branch?

leandron (commented):
A follow-up change is required, following what was discussed in https://discuss.tvm.apache.org/t/reminder-of-our-code-review-guidelines/14388

In this case it is not by the face that it was merged too early, but probably because of the comments that might be confusing to some users.

python/tvm/script/_parser/core/dispatch.py
tqchen: I feel in this case it is a comment to talk about related references and is a minor thing that is fine as it is.

leandron: I think the comment makes a lot of assumptions and is misleading, especially for new contributors that may or may not be expecting this reference to be here in this branch, so I'm still in favour of having this fixed.

As it might be situation in which the contribution is expected to be in another branch, rather than pushing myself a patch to fix the comment, I think it is best that the original contributor to push a synchronised that would fix here and in `unity`.

tqchen:
Thanks for the suggestion.

This is an API used by TVMScript internally, and the particular remark is only making an example.

If any one of the TVMScript maintainer feels it is out of place. I think they should be able to come and fix it. But my take is that they are not, as such I would hope to appreciate the volunteering time effort of the module maintainers and let them make the call.

Thank you

PR #13091, opened 2022-10-17
[MetaSchedule] Introduce a variant of ModuleEquality to enable ignoring NDArray raw data

masahi:
A follow up to https://github.com/apache/tvm/pull/13050, also builds on https://github.com/apache/tvm/pull/13001. This PR enables the functionality in https://github.com/apache/tvm/pull/12706 without changing the existing `StructuralEqual/Hash`.

A question for discussion: Should this be the default ModuleEquality used by MS? It has no effect for the `link-params = False` case, and it simplifies the MS tuning API usage for the `link-params = True` case (Hexagon etc).

cc @junrushao @zxybazh


junrushao (approved):
LGTM!

PR #13092, opened 2022-10-17
[ETHOSN] Inline non-compute-intensive partitions

lhutton1:
Adds a pass that analyzes functions partitioned for the NPU and inlines those that are deemed "non-compute-intensive" back to the main function so that they can be considered for other backends. The current heuristic for deciding a non-compute-intensive function is to collectively check all of the operations in the function have no multiply accumulate operations. This heuristic is not optimal; optimization is left for future exploration.

This pass is inspired by the "IsComputeIntensiveGraph" pass in the TensorRT integration.

cc @ashutosh-arm @leandron

### DIFF
--- a/python/tvm/relay/op/contrib/ethosn.py
+++ b/python/tvm/relay/op/contrib/ethosn.py
@@ -88,7 +88,7 @@ def InlineNonComputeIntensivePartitions() -> tvm.ir.IRModule:  # pylint: disable
     return _ethosn.InlineNonComputeIntensivePartitions()
 
 
-def inline_non_compute_intensive_partitions() -> bool:
+def is_inline_non_compute_intensive_partitions_enabled() -> bool:
     """
     Determine whether to inline none-compute-intensive partitions.
 
@@ -148,7 +148,7 @@ def partition_for_ethosn(mod, params=None, **opts):
         transform.PartitionGraph(),
         ConvertEquivalents(),
     ]
-    if inline_non_compute_intensive_partitions():
+    if is_inline_non_compute_intensive_partitions_enabled():
         passes.append(InlineNonComputeIntensivePartitions())
 
     return tvm.transform.Sequential(passes)(mod)
--- a/python/tvm/relay/op/contrib/ethosn.py
+++ b/python/tvm/relay/op/contrib/ethosn.py
@@ -64,12 +64,40 @@ def ConvertEquivalents() -> tvm.ir.IRModule:  # pylint: disable=invalid-name
     """Converts operations into a numerically equivalent form
     that can be understood by the NPU codegen.
 
-    Return
-    ------
+    Returns
+    -------
     Pass
         The module pass.
     """
     return _ethosn.ConvertEquivalents()
+
+
+def InlineNonComputeIntensivePartitions() -> tvm.ir.IRModule:  # pylint: disable=invalid-name
+    """This pass checks whether functions partitioned for the NPU are considered
+    non-compute intensive. If they are not, they will be unpartitioned and passed onto
+    other backends to consider.
+
+    A partitioned function is currently considered non-compute intensive if it contains
+    no multiply accumulate operations.
+
+    Returns
+    -------
+    Pass
+        The module pass.
+    """
+    return _ethosn.InlineNonComputeIntensivePartitions()
+
+
+def inline_non_compute_intensive_partitions() -> bool:
+    """
+    Determine whether to inline none-compute-intensive partitions.
+
+    Returns
+    -------
+    True if inlining should happen, False if not.
+    """
+    compiler_attrs = tvm.get_global_func("relay.ext.ethos-n.get_compiler_attrs")()
+    return compiler_attrs.inline_non_compute_intensive_partitions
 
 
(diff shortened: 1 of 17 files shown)


python/tvm/relay/op/contrib/ethosn.py
```diff
@@ -64,14 +64,42 @@ def ConvertEquivalents() -> tvm.ir.IRModule:  # pylint: disable=invalid-name
...
-    ------
+    Returns
+    -------
     Pass
         The module pass.
     """
     return _ethosn.ConvertEquivalents()
 
 
+def InlineNonComputeIntensivePartitions() -> tvm.ir.IRModule:  # pylint: disable=invalid-name
```
ashutosh-arm, line 75: nit: if possible, avoid ignoring pylint

python/tvm/relay/op/contrib/ethosn.py
```diff
@@ -64,14 +64,42 @@ def ConvertEquivalents() -> tvm.ir.IRModule:  # pylint: disable=invalid-name
...
+
+    Returns
+    -------
+    Pass
+        The module pass.
+    """
+    return _ethosn.InlineNonComputeIntensivePartitions()
+
+
+def inline_non_compute_intensive_partitions() -> bool:
```
ashutosh-arm, line 91: just a suggestion: this name could start with can_?

src/relay/backend/contrib/ethosn/inline_partitions.cc
```diff
@@ -0,0 +1,120 @@
...
+        is_compute_intensive = true;
+      }
+    }
+  }
+
+ private:
+  /*! \brief Whether or not the partitioned function is consdiered compute intensive. */
+  bool is_compute_intensive;
+  /*! \brief A set of operators considered compute intensive. */
+  const std::unordered_set<std::string> compute_intensive_operators{
```
ashutosh-arm, line 78: In order to make this pass more generic, would it make sense to accept this list as an input to the pass? In case of npu, this can be passed from the python partitioner.

ashutosh-arm: me overthinking :thinking: : maybe in future, this list can be accepted as a tvmc command line argument to make the pass even more generic?

src/relay/backend/contrib/ethosn/inline_partitions.cc
```diff
@@ -0,0 +1,120 @@
...
+      "ethos-n.qnn_add",        "ethos-n.qnn_conv2d",  "ethos-n.qnn_conv2d_transpose",
```
ashutosh-arm, line 79: Question: By def compute intensive ops are that involve macs, why is qnn.add included in this list? Is it because of scalar multiply and zp additions per pixel? Also, why depth_to_space?

src/relay/backend/contrib/ethosn/inline_partitions.cc
```diff
@@ -0,0 +1,120 @@
...
+    Call call = GetRef<Call>(op);
+    std::string op_name = "";
+    if (const auto* op = call->op.as<OpNode>()) {
+      op_name = op->name;
+    } else if (const auto* func = call->op.as<FunctionNode>()) {
+      op_name = func->GetAttr<String>(attr::kComposite, "").value();
+    }
+
+    if (op_name != "") {
+      if (compute_intensive_operators.find(op_name) != compute_intensive_operators.end()) {
```
ashutosh-arm, line 68: question: in this mechanism, a partitioned function containing lot of non compute intensive ops could be inlined too. is that intentional?

lhutton1: Yes currently that's possible. It's difficult to come up with a sensible limit here without being able to estimate the performance, perhaps if needed we can expose this option to the user in the future?

src/relay/backend/contrib/ethosn/inline_partitions.cc
lhutton1: I was a bit hesitant to make this pass more generic as different hardware might require a different heuristic all together (rather than just a different list of operators). That said, it does seem useful for the user to customise/tune the pass from TVMC if necessary, especially since the heuristic is not optimal. But yep, think this would be good as a followup

src/relay/backend/contrib/ethosn/inline_partitions.cc
lhutton1: Oops, good catch, thanks!

python/tvm/relay/op/contrib/ethosn.py
lhutton1: The name is like this to simply follow the style of other pass names, in general though I agree

ashutosh-arm: oh yeah I missed that as the pylint pragmas were at the top of those files at other places.

ashutosh-arm (approved):
LGTM! Thanks Luke for making it as generic as possible. It can be used by many other targets with little modifications.

leandron (approved):
LGTM, thanks @lhutton1 @ashutosh-arm

PR #13096, opened 2022-10-17
[Relay] Add set_attrs_type registry to broadcast_to op

wrongtest-intellif:
It seems that ir text is not parsable without set_attrs_type registry.

```python
%2 = broadcast_to(%1, shape=[128, 1, 28]);
```


shingjan:
Good catch. Seems that this code path is not covered by any tests. Would that be possible that we can add a test to cover the usage of this relay op?

PR #12969, opened 2022-10-03
[microTVM] Add Cortex-M DSP schedules for optimal conv2d layouts

guberti:
This pull request adds fast microTVM DSP schedules for the _optimal_ `conv2d` and `depthwise_conv2d` layouts - `NHWC/OHWI` and `NCHW/OIHW`. By letting us use the special `SMLAD` DSP instruction _without rearranging data_, **25%** of the instructions in the inner loop can be removed (for the `int16` case).

Additionally, this change allows both the `conv2d` and `depthwise_conv2d` fast DSP schedules to use the _same underlying intrinsic_, a variation of a tensordot operator. This makes the code for these schedules much more compact. This PR also:

- Adds unit tests for the new schedules
- Does not affect the old schedules, which are still used when they apply
  - The cases in which these new fast schedules apply are strictly different
- Adds support for `int8`, `int16`, and `int32` input data types in the new schedules
- Adds a `change_constant_shape` utility to `tvm.topi`

This pull request **does not** change the strategy functions to call these new fast schedules - this will occur in a follow up PR. I've also written a comment below delving into why these layouts are optimal.

PR #13099, opened 2022-10-17
[HotFix] Fix python import

echuraev:
Tuning doesn't work after #12969.
It reports the following error:

```
ImportError: cannot import name 'get_const_float' from partially initialized module 'tvm.topi.utils' (most likely due to a circular import)
```

In this commit, I moved import relay to a function which used in a test. And it helps to fix this circular import.

To reproduce this issue, you can run the following [tutorial](https://github.com/apache/tvm/blob/main/gallery/tutorial/autotvm_relay_x86.py) and will see this error.

cc: @guberti, @areusch, @ekalda

### DIFF
--- a/python/tvm/topi/utils.py
+++ b/python/tvm/topi/utils.py
@@ -22,7 +22,7 @@ from numbers import Integral
 
 import numpy as np
 import tvm
-from tvm import relay, te
+from tvm import te
 from tvm.tir import bijective_layout, layout
 from . import cpp, tag
 
@@ -452,6 +452,8 @@ def change_constant_shape(src, src_layout, dst_layout):
     dst_shape : relay.Constant
         A copy of the Constant with the new layout.
     """
+    from tvm import relay
+
     assert src_layout.isalpha() and dst_layout.isalpha()
     axis_order = [src_layout.index(c) for c in dst_layout]
     reshaped = np.transpose(src.data.numpy(), axis_order)
--- a/python/tvm/topi/utils.py
+++ b/python/tvm/topi/utils.py
@@ -452,12 +452,10 @@ def change_constant_shape(src, src_layout, dst_layout):
     dst_shape : relay.Constant
         A copy of the Constant with the new layout.
     """
-    from tvm import relay
-
     assert src_layout.isalpha() and dst_layout.isalpha()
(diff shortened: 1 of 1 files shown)


guberti:
If tuning doesn't work, that's a pretty big oversight on my part and a hotfix like this makes sense.

I can't seem to reproduce the issue though - here's a log of [what I get when I run on the latest TVM revision](https://github.com/apache/tvm/files/9801233/output.txt). 

We also have a unit test that (I believe) compiles and re-deploys the docs, with output, whenever changes are made. Here's a [link to the tutorial in question](https://tvm.apache.org/docs/tutorial/autotvm_relay_x86.html) - the output there looks fine.

Any ideas why I can't seem to reproduce this?

echuraev:
Hmmm. Looks interesting. I was able to reproduce this issue on the commit:   3f0d3f24b79c850f3415a40b594a5cdfc163dbd8
I run on Ubuntu 20.04 with python 3.8.8. [Here](https://github.com/apache/tvm/files/9801348/out_tuning_bad.log) is my output for this script from the tutorial. And you can see that tuning doesn't work. @guberti which environment do you use?

echuraev:
I see that someone else has the same issue: https://discuss.tvm.apache.org/t/error-of-import-topi-utils-during-autotune/13728

guberti:
I'm on almost the exact same environment - Ubuntu 20.04 and Python 3.8.10 [(see log)](https://github.com/apache/tvm/files/9802521/out_tuning_log.txt). I wonder if this is an issue with your Python `PATH` environment, but I don't have any way to investigate.

I suspect the error will go away once #13051 merges, but I'm very curious what the underlying problem is - we should add a test for this.

guberti (changes requested):
The changes themselves seem reasonable, but I'm hesitant to merge a hotfix for an issue we can't reproduce. Any chance this issue can be reproduced on Google Colab or something similar?

masahi:
It's not just tuning, I got this error yesterday when running https://github.com/apache/tvm/blob/main/tests/python/unittest/test_meta_schedule_schedule_rule_mlt.py by `python`. The error doesn't doesn't happen if the script is invoked by pytest.

PR #13111, opened 2022-10-17
[Hexagon] [runtime] Create objects to manage thread hardware resources

janetsc:
This is the first step towards threads dedicated to different hardware resources.

Added two objects to manage compute resources - HexagonHmx and HexagonHvx.  These can be created on the main thread for now.  The next step will be to acquire those resources on specific threads.

Tested on HDK:
pytest tests/python/contrib/test_hexagon/test_parallel_hvx_load_vtcm.py 
pytest tests/python/contrib/test_hexagon/test_parallel_hvx.py
pytest tests/python/contrib/test_hexagon/test_launcher.py
pytest tests/python/contrib/test_hexagon/test_run_unit_tests.py
pytest tests/python/contrib/test_hexagon/test_software_pipeline_async.py

cc: @csullivan, @kparzysz-quic, @JosephTheOctonaut, @adstraw


tests/cpp-runtime/hexagon/hexagon_thread_manager_tests.cc
```diff
@@ -161,7 +162,8 @@ TEST_F(HexagonThreadManagerTest, pipe_fill) {
   CHECK_EQ(answer, 42);
 }
 
-TEST_F(HexagonThreadManagerTest, pipe_overflow) {
+// TODO(HWE): Create a temporary thread manager with a smaller pipe for this test
```
janetsc, line 165: Per discussion this morning, one way around this is to split out a base class for thread management, and then a HexagonThreadManager that inherits this.  Hardware acquisition only occurs from the HexagonThreadManager.  We could use that base class to test pipe length, etc.

adstraw: Yes, would like to know the feasibility of splitting into testable base class + HexagonThreadManager before we decide whether to land this PR with this disabled test.

janetsc: I think this might have some complexity when we parameterize the thread_main for each hardware resource that will be dedicated to that thread.  So I'd like to postpone this until that task to acquire one hardware resource per thread.

csullivan (approved):
LGTM, we'll need to add `-mhmx` to the build but it looks like the power and acquire APIs don't require this compiler flag.

src/runtime/hexagon/hexagon_hvx.h
```diff
@@ -0,0 +1,55 @@
...
+  //! \brief Prevent copy assignment with HexagonHvx.
+  HexagonHvx& operator=(const HexagonHvx&) = delete;
+
+  //! \brief Prevent move construction.
+  HexagonHvx(HexagonHvx&&) = delete;
+
+  //! \brief Prevent move assignment.
+  HexagonHvx& operator=(HexagonHvx&&) = delete;
+
+ private:
```
adstraw, line 48: Looks like this can be deleted.  No private members / methods.

tests/cpp-runtime/hexagon/hexagon_thread_manager_tests.cc
```diff
@@ -28,15 +29,15 @@ using namespace tvm::runtime::hexagon;
...
     streams = htm->GetStreamHandles();
   }
-  void TearDown() override { delete htm; }
+  void TearDown() override {}
   HexagonThreadManager* htm{nullptr};
   std::vector<TVMStreamHandle> streams;
   int answer{0};
   const unsigned threads{6};
-  const unsigned pipe_size{100};
+  const unsigned pipe_size{1000};
```
adstraw, line 40: We might add a "getter" for the pipe size and stack size.

adstraw (approved):
Looks good.  A couple small things.  Is the idea that existing testing should cover this change?

PR #13112, opened 2022-10-17
[PopenPool] Enable Stdout & Stderr Redirect in PopenPool & PopenWorker

zxybazh:
This PR enables redirection of the standard output stream and standard error stream to given subprocess handler.

This function especially improves the usage of `LocalRunner` because previously a sample candidate's exception during running could print long lines of output, making console logging / notebook logging obsolete. In the meantime, any failure of the candidates are still preserved in the per-task logging files (Schedule, trace & TIR available, doesn't include the stack trace though).

Credits to @Hzfengsy's fix in https://github.com/mlc-ai/relax/pull/6.

### DIFF
--- a/python/tvm/contrib/popen_pool.py
+++ b/python/tvm/contrib/popen_pool.py
@@ -98,11 +98,11 @@ class PopenWorker:
         i.e. killed and restarted. If `None`, the process will be reused until
         an operation times out.
 
-    stdout: subprocess.PIPE or None
-        The standard output streams handler for the popen process.
+    stdout: Union[None, int, IO[Any]]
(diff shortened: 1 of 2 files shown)


zxybazh:
CC @Hzfengsy @shingjan  @junrushao @tqchen

shingjan (approved):
LGTM

Hzfengsy (approved):
LGTM, but I wonder if we can have additional testcases

zxybazh:
> LGTM, but I wonder if we can have additional testcases

Thanks, will add that in a follow up PR!

Issue #13030, opened 2022-10-11
[Bug] `trilu` not tagged with `injective` and thus miss reduce schedule


ganler:
### Expected behavior

TVM should successfully compile a model whose operators are supoprted.

### Actual behavior

For some cases, `trilu-argmin` cannot find a reduction strategy and thus throw:

https://github.com/apache/tvm/blob/813136401a11a49d6c15e6013c34dd822a5c4ff6/python/tvm/topi/x86/reduction.py#L90

The reason is that in this case `trilu`'s `operator.tag` is empty which is not regarded as `injective`:

https://github.com/apache/tvm/blob/813136401a11a49d6c15e6013c34dd822a5c4ff6/python/tvm/topi/x86/reduction.py#L84

I am not sure but maybe we need to put proper operator tag for `trilu`?

### Environment

fa17da22c73fb9e95c27e4c28130835b628caf6b on Ubuntu 20.04.

### Steps to reproduce

Preferably a minimal script to cause the issue to occur.

```python
import tvm
from tvm import relay

i0 = relay.var("i0", shape=[2, 2], dtype="int32")
i1 = relay.var("i1", shape=(), dtype="int64")
v0 = relay.trilu(i0, i1)
v1 = relay.argmin(v0, axis=[0])

f = relay.Function([i0, i1], v1)
relay.create_executor("graph", device=tvm.cpu(), target="llvm").evaluate(f)
```

### Triage

Please refer to the list of label tags linked above to find the relevant tags and add them here in a bullet format (example below).

* needs-triage

cc: @jwfromm

PR #13120, opened 2022-10-18
fix: add compute tag for trilu

ganler:
fixes https://github.com/apache/tvm/issues/13030

@masahi @jwfromm


shingjan (commented):
If the test is passing then I guess it is fine. Just wonder if we need to add `trilu` [here](https://github.com/apache/tvm/blob/48be4ff3449166a7fc1e9209bb3f165ebe769e34/python/tvm/relay/op/_transform.py#L56) as an injective op.

shingjan (approved):
Seems like adding this as an injective schedule will requires the FTVMCompute to be added, which may not be what we want as the topi trilu is implemented in python. Looks good to me with the original change!

masahi (approved):

PR #13124, opened 2022-10-18
[test][hexagon] Fix pytest fixture resolution bug

cconvey:
Fix a bug in `test_benchmark_maxpool2d.py` where pytest was unable to find a fixture defined in a different module.


cconvey:
@tvm-bot rerun

PR #13125, opened 2022-10-18
[Hexagon] [runtime] Use malloc/free for RPC buffers

janetsc:
Updated the RPC HexagonPageAllocator to use malloc/free, and removed rpc_hexbuffs.

cc: @adstraw


adstraw (approved):
I feel like this is the right thing to do in terms of simplicity of design for the Hexagon Device API.  I cannot think of a case that motivates the need to _force_ the RPC allocations to be HexagonBuffers.

jroesch (approved):

PR #13127, opened 2022-10-18
[ci] Skip sccache for local hexagon builds

driazati:
sccache chokes on the hexagon builds for some users, so this disables it
by default locally but keeps it on in CI.

cc @adstraw @janetsc


janetsc (approved):
Verified locally.  Looks great!

adstraw (approved):
LGTM

areusch (approved):

PR #13129, opened 2022-10-19
[TIR][Arith] Implement kApplyConstraintsToBooleanBranches extension

Lunderberg:
When simplifying a branch of a boolean AND or a boolean OR, the other branch may be assumed not to dominate the result of the operator.  For example, when simplifying `(A and B)`, `A` may be simplified on the assumption that `B` is true.  Similarly, when simplifying `(A or B)`, `A` may be simplified on the assumption that `B` is false.

Prior to this commit, these constraints were not used for simplifications.  This commit introduced an optional extension, `kApplyConstraintsToBooleanBranches`, which exposes these constraints for simplification.  This isn't enabled by default, as some cases require a second visit to each branch of a boolean operator.  (e.g. Simplifying the LHS of an operator, using a constraint provided by the RHS, where the sub-analyzer that uses the constraint expects it to already be simplified.)

### DIFF
--- a/src/arith/rewrite_simplify.cc
+++ b/src/arith/rewrite_simplify.cc
@@ -72,6 +72,40 @@ using namespace tir;
 // handled by CanonicalSimplifier.
 //
 
+/* Utility for rewriting only boolean portions of an expression
+ *
+ * Performs a subset of simplifications done by RewriteSimplifier,
+ * sufficient to negate a simplified expression.  Intended for
+ * application on an expression that has previously been simplified.
+ *
+ * \param expr The boolean expression to be normalized
+ *
+ * \returns The normalized boolean expression
+ */
+PrimExpr NormalizeBooleanOperators(PrimExpr expr) {
+  PVar<PrimExpr> x, y;
+
+  while (true) {
+    if ((!!x).Match(expr)) {
+      expr = x.Eval();
+    } else if ((!(x || y)).Match(expr)) {
+      return NormalizeBooleanOperators(!x.Eval()) && NormalizeBooleanOperators(!y.Eval());
+    } else if ((!(x && y)).Match(expr)) {
+      return NormalizeBooleanOperators(!x.Eval()) || NormalizeBooleanOperators(!y.Eval());
+    } else if ((x >= y).Match(expr) || (!(x < y)).Match(expr) || (!(y > x)).Match(expr)) {
+      return y.Eval() <= x.Eval();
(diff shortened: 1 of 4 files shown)


src/arith/rewrite_simplify.cc
```diff
@@ -1557,7 +1591,44 @@ PrimExpr RewriteSimplifier::Impl::VisitExpr_(const NotNode* op) {
...
+
+    PrimExpr a = op->a;
+    PrimExpr b = op->b;
+
+    // Alternate which branch is used as the constraint, and which
+    // is being simplified.  Because some sub-analyzers expect their
+    // constraints to already be simplified, each branch may require
+    // more than update.  The loop condition allows each branch to be
+    // visited up to twice, but only if performs the second visit if
+    // necessary.
```
csullivan, line 1608: nits: 

"more than one update".

"but only performs the second visit if necessary."

src/arith/rewrite_simplify.cc
```diff
@@ -1557,7 +1591,44 @@ PrimExpr RewriteSimplifier::Impl::VisitExpr_(const NotNode* op) {
...
+    for (size_t i = 0, last_updated = 0; i < last_updated + 2 && i < 4; i++) {
```
csullivan, line 1609: nit: Is there a way to make this for-loop more understandable? The multiple iterator makes it difficult to understand, possibly error prone, without detailed study.

csullivan (approved):
LGTM with only nits

src/arith/rewrite_simplify.cc
Lunderberg: Thank you, and updated!

src/arith/rewrite_simplify.cc
Lunderberg: Yeah, I wasn't sure the best way to express this one.  I only wanted to perform one update per loop iteration, to avoid the fourth simplify (a -> b -> a -> **b**) if it converged earlier.

Hmm, maybe if I rewrite it as a fixed-length loop with early bail-out instead.  I'd initially written it without the `i < 4`, so that it would iterate until convergence, but now that I know how the maximum number of loops required, that flexibility isn't needed anymore.

Lunderberg: Rewritten, and it definitely is more coherent this way.

Lunderberg:
Thank you!

(Rebased onto main to resolve conflict.)

src/arith/rewrite_simplify.cc
Lunderberg: And updated in the OrNode handler as well.

PR #13135, opened 2022-10-19
[Hexagon] Add scripts for e2e MetaSchedule tuning demonstration

masahi:
I've worked on a series of PRs to enable e2e MS tuning on Hexagon (mostly for supporting `link-params = True` in MS). Now that all the pieces have been upstreamed, I'm adding demo tuning scripts under `test_hexagon/metaschedule_e2e`. 

They are not run on CI, and running it locally requires PyTorch to generate fp16 and int8 resnet50. 

The scripts use a small number of tuning trials and `replay-trace` search strategy instead of the evolutionary search, to finish tuning quickly. Those interested in MS tuning can tweak these settings for better performance at the cost of more tuning times.

@csullivan @kparzysz-quic @farshidsp

### DIFF
--- a/src/printer/tvmscript_printer.cc
+++ b/src/printer/tvmscript_printer.cc
@@ -457,9 +457,10 @@ class TVMScriptPrinter : public StmtFunctor<Doc(const Stmt&)>,
  */
 template <typename T>
 void NDArrayToTIR(::tvm::runtime::NDArray arr, std::ostream& os) {
-  if (arr.DataType().code() == runtime::DataType::kInt ||
-      arr.DataType().code() == runtime::DataType::kUInt) {
-    // Printing integer NDArrays causes "UnicodeDecodeError: 'utf-8' codec can't decode byte"
+  if ((arr.DataType().code() == runtime::DataType::kInt ||
+       arr.DataType().code() == runtime::DataType::kUInt) &&
+      arr.DataType().bits() == 8) {
+    // Printing int8 NDArrays causes "UnicodeDecodeError: 'utf-8' codec can't decode byte"
     // error during MetaSchedule tuning on int8 models.
     return;
   }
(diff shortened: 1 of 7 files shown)


src/printer/tvmscript_printer.cc
```diff
@@ -457,6 +457,12 @@ class TVMScriptPrinter : public StmtFunctor<Doc(const Stmt&)>,
  */
 template <typename T>
 void NDArrayToTIR(::tvm::runtime::NDArray arr, std::ostream& os) {
+  if (arr.DataType().code() == runtime::DataType::kInt ||
+      arr.DataType().code() == runtime::DataType::kUInt) {
+    // Printing integer NDArrays causes "UnicodeDecodeError: 'utf-8' codec can't decode byte"
+    // error during MetaSchedule tuning on int8 models.
+    return;
+  }
```
masahi, line 465: We can't print an entire constant tensor to text, so this function is broken anyway in practice. I've update one of TVMScript parsing tests to avoid printing integer constants.

tests/python/contrib/test_hexagon/metaschedule_e2e/test_resnet50_fp16.py
```diff
@@ -0,0 +1,126 @@
...
+        graph_mod.set_input(input_name, inp.copy())
+
+        graph_mod.run()
+        hexagon_output = graph_mod.get_output(0).numpy()
+
+        print(
+            "max and mean abs difference with the reference:",
+            np.max(np.abs(ref_result - hexagon_output)),
+            np.mean(np.abs(ref_result - hexagon_output)),
+        )
```
csullivan, line 117: Are there accuracy issues with tuning the fp16 variant that prevent an assert_allclose with the reference?

csullivan (approved):
Thank you for these intrins and tests @masahi! It is invaluable to have these early demonstrations of the impact metaschedule has for Hexagon.

tests/python/contrib/test_hexagon/metaschedule_e2e/test_resnet50_fp16.py
masahi: There is non-trivial accuracy difference between x86 and HVX fp16. Add a sample output and assert check with a fairly loose bound.

csullivan: Thanks @masahi!

PR #13138, opened 2022-10-19
[Hexagon] Support fetching and building Hexagon runtime with external runtime sources

csullivan:
It can be desirable to incorporate external source libraries for Hexagon which are maintained outside of upstream TVM. This PR enables the cmake tvm option `USE_HEXAGON_EXTERNAL_LIBS` which can point to an external git repo containing runtime sources in `src/runtime/hexagon` for compilation along with the Hexagon TVM runtime. 

When a git repo url is used, a SHA is required to be specified as well. Any source controlled use of TVM for benchmarking should commit the SHA used for reproducibility in the respective benchmarking repository -- and explicitly not in TVM. 

cc @kparzysz-quic @mehrdadh @adstraw @janetsc @supersat @nverke


cmake/modules/Hexagon.cmake
```diff
@@ -182,6 +182,36 @@ if(BUILD_FOR_HEXAGON)
     "${TVMRT_SOURCE_DIR}/hexagon/ops/conv2d_fp16_hvx.cc"
     PROPERTIES COMPILE_FLAGS "-mhvx"
   )
+
+  # Include hexagon external library runtime sources
+  if(DEFINED USE_HEXAGON_EXTERNAL_LIBS)
+    if (EXISTS ${USE_HEXAGON_EXTERNAL_LIBS})
```
janetsc, line 188: I'm a bit confused here.  Should line 188 check for HEXAGON_EXTERNAL_LIBS_SHA?  If we get here, USE_HEXAGON_EXTERNAL_LIBS exists, yes?

cmake/modules/Hexagon.cmake
```diff
@@ -182,6 +182,36 @@ if(BUILD_FOR_HEXAGON)
...
+        "supports absolute paths and paths and git repository urls")
+    endif()
+
+    file_glob_append(HEXAGON_EXTERNAL_RUNTIME_SRCS
+      "${USE_HEXAGON_EXTERNAL_LIBS}/src/runtime/hexagon/*.cc"
+    )
+    list(APPEND RUNTIME_HEXAGON_SRCS "${HEXAGON_EXTERNAL_RUNTIME_SRCS}")
+    set_source_files_properties(
+      "${HEXAGON_EXTERNAL_RUNTIME_SRCS}"
+      PROPERTIES COMPILE_FLAGS "-mhvx -mhmx"
```
janetsc, line 212: Could these properties specific to the external private repro be define there, and this pulls them in from a known sub-location within an external repo?  (So that each private repo could customize flags inside itself.)

janetsc (approved):
Looks good, with a question and a comment

cmake/modules/Hexagon.cmake
csullivan: I should clear this up with a comment, thank you. EXISTS here is checking to see if USE_HEXAGON_EXTERNAL_LIBS is a path that is resolvable on the system. If it is a local path then it will use the path to the local install of external libs. EXISTS will return false when it's not a path, e.g. if it is a git url or malformed.

cmake/modules/Hexagon.cmake
csullivan: I was wondering the same also yesterday. 

One option, that I personally prefer not to rely on, is to not allow external sources and require the external libs to fully manage their own build to object files which are linked in. 

Another option that comes to mind would be to do a cmake include on a required hexagon_external.cmake in the external lib. I haven't explored the feasibility of this option but it could work.

csullivan: Thank you for the good feedback, https://github.com/apache/tvm/pull/13138/commits/18fe01c722414045fbda3642c4f1a7ed608777f9 takes your suggestion of optionally including an external library defined cmake config file which defines these compiler flags. The risk we run is keeping these synchronized across repos, but as long as the footprint stays microscopic (ie just a single shared variable name) it should be manageable.

mehrdadh (approved):
LGTM, thanks!

PR #13139, opened 2022-10-19
[Hexagon] [runtime] Remove released buffer check for post-ReleaseResources calls to FreeDataSpace

janetsc:
The motivation behind resource management is to ensure that we release all resources at the end of the session.

Given the session management on the client side, it is possible for ReleaseResources to be called before all client side objects have been torn down.  And, it is possible for frees to be called in the window between session end and process termination.  (After those buffers have been freed in ReleaseResources.)

Based on feedback from @adstraw and @Lunderberg on the usage of "released_runtime_buffers," we did an investigation into the post-release frees.  The cleanest solution is to keep the logic to free all buffers in ReleaseResources, and log a message on any frees that are executed outside a session.

### DIFF
--- a/src/runtime/hexagon/hexagon_device_api.h
+++ b/src/runtime/hexagon/hexagon_device_api.h
@@ -77,9 +77,6 @@ class HexagonDeviceAPI final : public DeviceAPI {
     runtime_threads.reset();
 
     CHECK(runtime_hexbuffs) << "runtime_hexbuffs was not created in AcquireResources";
-    if (!runtime_hexbuffs->empty()) {
-      LOG(INFO) << "runtime_hexbuffs was not empty in ReleaseResources";
-    }
     runtime_hexbuffs.reset();
 
     CHECK(runtime_vtcm) << "runtime_vtcm was not created in AcquireResources";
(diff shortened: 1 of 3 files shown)


src/runtime/hexagon/hexagon_device_api.h
```diff
@@ -79,8 +78,7 @@ class HexagonDeviceAPI final : public DeviceAPI {
 
     CHECK(runtime_hexbuffs) << "runtime_hexbuffs was not created in AcquireResources";
     if (!runtime_hexbuffs->empty()) {
-      DLOG(INFO) << "runtime_hexbuffs was not empty in ReleaseResources";
-      released_runtime_buffers = runtime_hexbuffs->current_allocations();
+      LOG(INFO) << "runtime_hexbuffs was not empty in ReleaseResources";
```
adstraw, line 81: Feels like this `LOG` could move to the destructor of the `HexagonBufferManager` and the `empty` method removed from `HexagonBufferManager` given this is the only call site.

adstraw (changes requested):

src/runtime/hexagon/hexagon_device_api.h
janetsc: The only downside to moving it is that we won't know which buffer manager had allocations when it was released.

masahi (approved):

PR #13140, opened 2022-10-19
[Hexagon] Set c++17 standard for launcher

quic-sanirudh:
The current launcher build fails because of missing c++17 flag, so this patch fixes that error


quic-sanirudh:
cc @kparzysz-quic

kparzysz-quic (approved):

PR #13145, opened 2022-10-19
[Hexagon] refactor HexagonBufferManager class

adstraw:
Refactoring `HexagonBufferManager` class with a few small changes: 1) moving mutex lock before `find` in `FreeHexagonBuffer` 2) Removing `count` method as the only caller uses it to check / log duplicate error as in `FreeHexagonBuffer` itself 3) rename to `FindHexagonBuffer` for clariry.

### DIFF
--- a/src/runtime/hexagon/hexagon_device_api.cc
+++ b/src/runtime/hexagon/hexagon_device_api.cc
@@ -154,8 +154,6 @@ void HexagonDeviceAPI::FreeWorkspace(Device dev, void* data) {
   CHECK(runtime_hexbuffs) << "Attempted to free Hexagon workspace with "
                           << "HexagonDeviceAPI::FreeWorkspace outside of a session.  "
                           << "Please call HexagonDeviceAPI::AcquireResources";
-  CHECK(runtime_hexbuffs->count(data) != 0)
-      << "Attempt made to free unknown or already freed workspace allocation";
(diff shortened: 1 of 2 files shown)


adstraw:
CC @janetsc

src/runtime/hexagon/hexagon_device_api.cc
```diff
@@ -154,8 +154,6 @@ void HexagonDeviceAPI::FreeWorkspace(Device dev, void* data) {
   CHECK(runtime_hexbuffs) << "Attempted to free Hexagon workspace with "
                           << "HexagonDeviceAPI::FreeWorkspace outside of a session.  "
                           << "Please call HexagonDeviceAPI::AcquireResources";
-  CHECK(runtime_hexbuffs->count(data) != 0)
```
janetsc, line 157: It is ok to remove count, but should be replaced with FindHexagonBuffer.  We want to make sure that the backing data buffer for this workspace is still actively allocated.

janetsc (commented):
Overall looks great - much cleaner.  I think that replacing the call to count with FindHexagonBuffer would be good.  I like the interface better.  But we do want to check to make sure the buffer for the workspace is still allocated.

janetsc (approved):
Looks great!

masahi (approved):

PR #13149, opened 2022-10-19
[Hexagon] [runtime] Clarify compute resources

janetsc:
Clarify naming for the tensor processing unit that will be power managed and acquired/released for the runtime

cc:  @csullivan, @kparzysz-quic


csullivan (approved):
LGTM, thank you!

PR #13151, opened 2022-10-20
[MetaSchedule] Support Tuning w/ No Cost Model & Fix Integration Test

zxybazh:
Previously even for replay trace & replay func strategy, an xgboost cost model will be created and retrained after each iteration of building and running. This PR introduces a `none` cost model option to avoid creating cost model. Also fixed the integration test for correctness verification by using the correct tvm device.


zxybazh:
CC @masahi @junrushao

Hzfengsy (approved):

Issue #13146, opened 2022-10-19
[Bug][Frontend][ONNX] Failure of test_forward.test_aten unit test after upgrade to PyTorch 1.12.0


oviazlo:
As described in the [discussion forum thread](https://discuss.tvm.apache.org/t/frontend-pytorch-tvm-compatibility-with-torch-1-12-0/13508/7) the test below fails after the upgrade to the PyTorch 1.12.0:
https://github.com/apache/tvm/blob/458ca81c01bd85d04ea6428e7cb4bfa021a1d37a/tests/python/frontend/onnx/test_forward.py#L5471-L5507

The test fails in the line:
https://github.com/apache/tvm/blob/458ca81c01bd85d04ea6428e7cb4bfa021a1d37a/tests/python/frontend/onnx/test_forward.py#L5488

The problem lies due to improper model conversion from PyTorch to ONNX with PyTorch 1.12.0. It looks like that argument below is ignored:
https://github.com/apache/tvm/blob/458ca81c01bd85d04ea6428e7cb4bfa021a1d37a/tests/python/frontend/onnx/test_forward.py#L5485
and ATen PyTorch operators are substituted with native ONNX operators (instead of being exported as ATen ops in ONNX). This behaviour fail an assert above.

I have created an appropriate bug report in the PyTorch repo: https://github.com/pytorch/pytorch/issues/87313

One possible quick solution will be to remove an assert above in `_convert_to_onnx` function (since the test produces correct results and satisfies all other asserts). However, it will reduce coverage of the test because `onnx_model` (in the code below) will not contain ATen operators, and the possibility to consume ONNX graph with such operators by TVM will not be tested.
https://github.com/apache/tvm/blob/458ca81c01bd85d04ea6428e7cb4bfa021a1d37a/tests/python/frontend/onnx/test_forward.py#L5497-L5503

### Environment

* TVM version: [v0.10.0rc0](https://github.com/apache/tvm/releases/tag/v0.10.0rc0) (https://github.com/apache/tvm/commit/82c4addf2f5cdf7ec142cf0378632deb6b9474e1)
* OS: macOS 12.6 (arm64)
* torch 1.12.0+cpu
* torchvision 0.13.0+cpu

### Steps to reproduce

With `Torch 1.12.0` and `Torchvision 0.13.0` environment, run:
`python -m pytest tests/python/frontend/onnx/test_forward.py`

### Triage

* frontend:onnx

oviazlo:
@masahi FYI

[...]

PR #13155, opened 2022-10-20
Remove assert to fix onnx frontent unit tests for PyTorch 1.12

oviazlo:
Issue: #13146


oviazlo:
@masahi I have removed assert as you proposed in the issue. It should fix broken onnx test.

masahi (approved):

PR #13158, opened 2022-10-20
[Bugfix][Runtime] Fix sched_setaffinity in Android

Wanger-SJTU:
On Android `sched_setaffinity ` should use `pid_t` instread of `pthread_t`,  
https://man7.org/linux/man-pages/man2/sched_setaffinity.2.html

On Android 21+, there is `pid_t pthread_gettid_np(pthread_t);` to convert the `std::thread::native_handle_type`  to `pid_t`, 

```cpp
pid_t pthread_gettid_np(pthread_t __pthread) __INTRODUCED_IN(21);
```

### DIFF
--- a/src/runtime/threading_backend.cc
+++ b/src/runtime/threading_backend.cc
@@ -28,7 +28,6 @@
 #include <pthread.h>
 #include <fstream>
 #include <sstream>
-#include <pthread.h>
 #else
 #endif
 #if defined(__linux__)
--- a/src/runtime/threading_backend.cc
+++ b/src/runtime/threading_backend.cc
@@ -26,6 +26,7 @@
 
 #if defined(__linux__) || defined(__ANDROID__)
 #include <pthread.h>
+
 #include <fstream>
 #include <sstream>
 #else
--- a/src/runtime/threading_backend.cc
+++ b/src/runtime/threading_backend.cc
@@ -25,6 +25,7 @@
 #include <tvm/runtime/threading_backend.h>
 
 #if defined(__linux__) || defined(__ANDROID__)
+#include <pthread.h>
 #include <fstream>
 #include <sstream>
 #else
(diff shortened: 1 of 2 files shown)


src/runtime/threading_backend.cc
```diff
@@ -167,7 +169,9 @@ class ThreadGroup::Impl {
       CPU_SET(id, &cpuset);
     }
 #if defined(__ANDROID__)
-    sched_setaffinity(thread, sizeof(cpu_set_t), &cpuset);
+    if (sched_setaffinity(pthread_gettid_np(thread), sizeof(cpu_set_t), &cpuset) != 0) {
```
echuraev, line 172: What if we will try to run this code on older versions than Android21+?

Wanger-SJTU: If  on older versions than Android21+, there could be compile error with no func found.  In this case, could be record the tid  
like 
```cpp
std::mutex m;
std::map<std::thread::id, pid_t> threads;
void add_tid_mapping()
{
  std::lock_guard<std::mutex> l(m);
  threads[std::this_thread::get_id()] = syscall(SYS_gettid);
}
void wrap(void (*f)())
{
  add_tid_mapping();
  f();
}
```

echuraev: Just thinking how we can avoid such compilation error. Can we [check](https://developer.android.com/ndk/reference/group/apilevels) Android API version and for API>= 21 we will use `pthread_gettid_np` and in the opposite case - another implementation?

Wanger-SJTU: I updated the with `__ANDROID_API__` macro, with api level >=21, use `pthread_gettid_np`, while < 21, get the thread tid due to the internal pthread implementation

Wanger-SJTU: @echuraev

echuraev (approved):
LGTM. Thanks!

cc: @vinx13, @FrozenGene, @masahi

masahi:
@tvm-bot rerun

Wanger-SJTU:
@echuraev maybe need you  approve review again

masahi (approved):

PR #13159, opened 2022-10-20
[ETHOSN] Consolidate target string usage

lhutton1:
Removes support for a deprecated target string. The deprecation warning has been around for a couple of releases now so it should be safe to remove. The target to use moving forward is: `ethos-n -variant=n78 ...`

Refactored direct use of a driver stack target string in the testing infrastructure to use the same string we expect users to provide. This simplified some of the code in codegen and hopefully avoids confusion in the future.

cc @Leo-arm @ashutosh-arm @leandron @Mousius


lhutton1:
This is ready for review :)

leandron (commented):
LGTM, thanks @lhutton1!

leandron (approved):

PR #13161, opened 2022-10-20
[Hexagon]: Add upload function to hexagon session

mehrdadh:
This PR adds `upload` to hexagon RPC server running on Hexagon side. This will remove the need of `adb` command for uploading/executing binary files on hexagon.

### DIFF
--- a/python/tvm/contrib/hexagon/build.py
+++ b/python/tvm/contrib/hexagon/build.py
@@ -35,6 +35,7 @@ import tvm
 from tvm.contrib.hexagon.hexagon_profiler import HexagonProfiler
 from ..._ffi import libinfo
 from .session import Session
+from .tools import HEXAGON_SIMULATOR_NAME
 
 HEXAGON_RPC_LIB_DIR = os.environ.get("HEXAGON_RPC_LIB_DIR")
 ANDROID_BASH_FILE_NAME = "android_bash.sh"
@@ -133,6 +134,7 @@ class HexagonLauncherRPC(metaclass=abc.ABCMeta):
         }
         self._rpc_info.update(rpc_info)
         self._workspace = self._create_workspace(workspace)
+        self._serial_number
 
     @abc.abstractmethod
     def start_server(self):
@@ -241,8 +243,12 @@ class HexagonLauncherRPC(metaclass=abc.ABCMeta):
             "priority": 0,
             "timeout": 0,
             "key": self._rpc_info["device_key"],
+            "serial_number": self._serial_number,
         }
         return Session(self._workspace, hexagon_remote_kw, session_name=session_name)
+
+    def is_simulator(self):
+        return self._serial_number == HEXAGON_SIMULATOR_NAME
 
 
 class HexagonLauncherAndroid(HexagonLauncherRPC):
@@ -584,7 +590,7 @@ class HexagonLauncherSimulator(HexagonLauncherRPC):
         self._toolchain = os.environ.get("HEXAGON_TOOLCHAIN")
         if not self._toolchain:
             raise RuntimeError("Please set HEXAGON_TOOLCHAIN env variable")
-        self._serial_number = "simulator"
+        self._serial_number = HEXAGON_SIMULATOR_NAME
 
         super(HexagonLauncherSimulator, self).__init__(rpc_info, workspace)
 
@@ -714,7 +720,7 @@ def HexagonLauncher(
     sysmon_profile: bool = False,
 ):
     """Creates a HexagonLauncher"""
-    if serial_number == "simulator":
+    if serial_number == HEXAGON_SIMULATOR_NAME:
         return HexagonLauncherSimulator(rpc_info, workspace)
     return HexagonLauncherAndroid(
         serial_number, rpc_info, workspace, hexagon_debug, clear_logcat, sysmon_profile
--- a/python/tvm/contrib/hexagon/build.py
+++ b/python/tvm/contrib/hexagon/build.py
@@ -264,7 +264,7 @@ class HexagonLauncherAndroid(HexagonLauncherRPC):
         if not rpc_info.get("workspace_base"):
             rpc_info["workspace_base"] = self.ANDROID_HEXAGON_TEST_BASE_DIR
         self._serial_number = serial_number
-        assert self._serial_number != ""
+        assert self._serial_number != "", "Android serial number is not set."
 
         adb_socket = rpc_info["adb_server_socket"] if rpc_info["adb_server_socket"] else "tcp:5037"
         self._adb_device_sub_cmd = ["adb", "-L", adb_socket, "-s", self._serial_number]
--- a/python/tvm/contrib/hexagon/build.py
+++ b/python/tvm/contrib/hexagon/build.py
@@ -125,7 +125,9 @@ class HexagonLauncherRPC(metaclass=abc.ABCMeta):
         used.
     """
 
-    def __init__(self, rpc_info: dict, workspace: Union[str, pathlib.Path] = None):
+    def __init__(
+        self, rpc_info: dict, workspace: Union[str, pathlib.Path] = None, serial_number: str = None
+    ):
         self._rpc_info = {
             "rpc_tracker_host": "0.0.0.0",
             "rpc_tracker_port": 9190,
@@ -134,7 +136,7 @@ class HexagonLauncherRPC(metaclass=abc.ABCMeta):
         }
         self._rpc_info.update(rpc_info)
         self._workspace = self._create_workspace(workspace)
-        self._serial_number
+        self._serial_number = serial_number
 
     @abc.abstractmethod
     def start_server(self):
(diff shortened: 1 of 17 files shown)


python/tvm/contrib/hexagon/meta_schedule.py
```diff
@@ -100,7 +101,7 @@ def run(self, runner_inputs: List[RunnerInput]) -> List[RunnerFuture]:
 
 
 def _worker_func(hexagon_launcher, evaluator_config, alloc_repeat, artifact_path, args_info):
-    with hexagon_launcher.start_session() as session:
+    with create_session(hexagon_launcher._workspace, hexagon_launcher._rpc_info) as session:
```
Lunderberg, line 104: Why is `meta_schedule` accessing internal objects owned by the `hexagon_launcher`?

python/tvm/contrib/hexagon/session.py
```diff
@@ -393,3 +394,37 @@ def _aot_executor_from_factory(
             remote_file_path = self.upload(binary_path, binary_name)
 
         return self.get_aot_executor(remote_file_path)
+
+
+def create_session(
```
Lunderberg, line 399: Currently, all calls to `create_session` pass arguments owned by a `HexagonLauncher` instance.  The `_workspace` and `_rpc_info` objects are marked as `_`, so these would be expected to be internal implementation details, not part of the external API of the `HexagonLauncher`.

Do we expect to call `create_session` based on something in a workflow that doesn't include `HexagonLauncher`?  If yes, then both the free function `hexagon.session.create_session` and the method `HexagonLauncher.create_session` should exist, with `HexagonLauncher.create_session` calling the free function instead of requiring users to access private variables.  If no, then only the `HexagonLauncher` method should exist.

python/tvm/contrib/hexagon/session.py
```diff
@@ -393,3 +394,37 @@ def _aot_executor_from_factory(
...
+    assert "rpc_tracker_port" in rpc_info.keys()
+
+    hexagon_remote_kw = {
+        "host": rpc_info["rpc_tracker_host"],
+        "port": rpc_info["rpc_tracker_port"],
+        "priority": 0,
+        "timeout": 0,
+        "key": rpc_info["device_key"],
+    }
+    return Session(workspace, hexagon_remote_kw, session_name=session_name)
```
Lunderberg, line 430: Alternatively, since the only purpose of this function is to initialize a `Session` object, why not move the free function into `Session.__init__`?

tests/python/contrib/test_hexagon/test_meta_schedule.py
```diff
@@ -62,7 +64,7 @@ def main(  # type: ignore  # pylint: disable=no-self-argument
 
 @tvm.testing.requires_hexagon
 def test_builder_runner(hexagon_launcher):
-    if hexagon_launcher._serial_number == "simulator":
+    if android_serial_number()[0] == "simulator":
```
Lunderberg, line 67: To a casual reader, there's nothing that connects the `android_serial_number` function to either the `hexagon_launcher` or the `hexagon_session`.  The current implementation also has incorrect behavior when running tests on multiple devices/simulators.  If `ANDROID_SERIAL_NUMBER=abcd1234,simulator`, the test would be erroneously run on the simulator.  If `ANDROID_SERIAL_NUMBER=simulator,abcd1234`, the test would be erroneously skipped on the device.

Instead of requiring user code to call a free function, can we instead have `HexagonLauncher.is_simulator()` and/or `HexagonSession.is_simulator()` methods?  That way, it's clear to the user that this is specific to the connected device, and not something that is globally true.

Lunderberg (changes requested):
I like the session-scoped uploads, and allowing tests to be reliant solely on the `hexagon_session` instead of the `hexagon_launcher`.  This is a strictly better usage for a testing environment, where all state should be session-scoped.  However, it doesn't allow for a production use case in which a model is uploaded once, and then run many times across many connections.

I'd propose the following design, which would allow for both use cases:

* A HexagonSession has a single workspace, which should be cleared when closing a session.  Calls to `HexagonSession.upload` go into that session's workspace.  Calls to `HexagonSession.load_module` may *only* be a `tvm.runtime.Module`, and not a `str` or `pathlib.Path`, because persistent uploads are not supported.
* A HexagonLauncher has a single workspace, which should be persistent across multiple connections to the same device.  Calls to `HexagonLauncher.upload` go into the persistent workspace.  Calls to `HexagonLauncher.load_module` may be a `tvm.runtime.Module`, or a `str`/`pathlib.Path` to access previously uploaded artifacts.

This would be simplest to implement by having a `upload` method in both `HexagonLauncher` and `HexagonSession`.  Each would call a private method of `HexagonLauncher` that takes an explicit workspace, but those workspaces would not otherwise be exposed to the user.

python/tvm/contrib/hexagon/build.py
```diff
@@ -378,13 +239,16 @@ def __init__(
...
+        assert self._serial_number != ""
+
         adb_socket = rpc_info["adb_server_socket"] if rpc_info["adb_server_socket"] else "tcp:5037"
         self._adb_device_sub_cmd = ["adb", "-L", adb_socket, "-s", self._serial_number]
         self.forwarded_ports_ = []
         self._hexagon_debug = hexagon_debug
         self._clear_logcat = clear_logcat
         self._sysmon_profile = sysmon_profile
         self._sysmon_process = None
+        rpc_info["device_key"] = HEXAGON_REMOTE_DEVICE_KEY + "." + self._serial_number
```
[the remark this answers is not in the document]
mehrdadh: Each test worker is supposed to use a pair of (android_serial_number, RPC session). But after each worker registers the RPS server to the tracker, it might pickup RPC session with an RPC server which is not related to the same android_serial_number. By adding `self._serial_number` to the rpc_key, each session is forced to use the same RPC server

python/tvm/contrib/hexagon/build.py
```diff
@@ -689,18 +556,19 @@ def start_server(self):
             self._copy_to_remote(lib_dir / item, self._workspace / item)
         # Copy libc++ from the toolchain to the workspace
         self._copy_libcxx(self._workspace)
-        self._device_key = self.HEXAGON_REMOTE_DEVICE_KEY + "." + str(os.getpid())
+        self._rpc_info["device_key"] = HEXAGON_REMOTE_DEVICE_KEY + "." + str(os.getpid())
```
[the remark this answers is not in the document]
mehrdadh: On simulator, each test initializes an RPC server. `os.getpid()` is used to distinguished between those server, otherwise one test could connect to an RPC server (by mistake) that was created before but the shutdown process for it was already started.

python/tvm/contrib/hexagon/meta_schedule.py
mehrdadh: `meta_schedule` is just creating a session and for that, we need to pass workspace and rpc_info from launcher to the session . I can make add `get` functions to HexagonLauncher instead of accessing them directory. wdyt?

python/tvm/contrib/hexagon/session.py
mehrdadh: that makes sense. I will change that

python/tvm/contrib/hexagon/session.py
mehrdadh: hexagon.session only needs a workspace on the remote device and RPC info. Considering that, it's always dependent to HexagonLauncher. It might be better to just move the create_session to HexagonLauncher?

tests/python/contrib/test_hexagon/test_meta_schedule.py
mehrdadh: Makes sense. The only issue here is that hexagon.session doesn't have access to serial number. We could pass serial number as input to the function and add is_simulator() function call

python/tvm/contrib/hexagon/build.py
```diff
@@ -102,14 +101,9 @@ class HexagonLauncherRPC(metaclass=abc.ABCMeta):
     The basic flow of interaction with the launcher is
         launcher = HexagonLauncher(...)
         launcher.start_server()
-        with launcher.start_session() as session:
+        with session.create_session() as session:
```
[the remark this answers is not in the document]
mehrdadh: I fixed this after moving the create_session to launcher.

python/tvm/contrib/hexagon/session.py
mehrdadh: I moved create_session to launcher since it is very dependent to launcher attributes

tests/python/contrib/test_hexagon/test_meta_schedule.py
mehrdadh: I added is_simulator to both Session and Launcher

Lunderberg: That makes sense, and thank you!  That looks a lot nicer in the usage.

python/tvm/contrib/hexagon/build.py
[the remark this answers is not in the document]
Lunderberg: Thank you, and that makes sense!

Lunderberg:
Thank you, and the changes made address all the specific changes requested.  I think the last item remaining is how the new design would support persistent artifacts stored on the hexagon device across multiple sessions.

PR #13163, opened 2022-10-20
[Relay][Frontend][ONNX] Add GridSample operator

ehsanmok:
This PR adds the GridSample operator introduced in [opset 16](https://github.com/onnx/onnx/blob/main/docs/Changelog.md#gridsample-16) with a couple of small fixes in the mainline topi version.

@sfvaroglu @AndrewZhaoLuo

### DIFF
--- a/python/tvm/topi/image/grid_sample.py
+++ b/python/tvm/topi/image/grid_sample.py
@@ -81,7 +81,7 @@ def _grid_sample_2d(
 
     The left-top corner (-1, -1) and right-bottom corner (1, 1) in grid will be map to
     (0, 0) and (h - 1, w - 1) of data if align_corners is "True", or
-    (-0.5, -0.5) and (h + 0.5, w + 0.5) of data if align_corners is "False".
+    (-0.5, -0.5) and (h - 0.5, w - 0.5) of data if align_corners is "False".
 
     The shape of the output will be (data.shape[0], data.shape[1], grid.shape[2], grid.shape[3]).
 
@@ -200,13 +200,13 @@ def _grid_sample_2d(
 
     def _nearest_sample(n, c, h, w):
         y, x = _compute_source_index(n, h, w)
-        y_new = te.round(y).astype("int32")
-        x_new = te.round(x).astype("int32")
+        y_new = te.nearbyint(y).astype("int32")
+        x_new = te.nearbyint(x).astype("int32")
 
         return _get_pixel_value(n, c, y_new, x_new)
 
     def _bicubic_sample(n, c, h, w):
-        A = -0.75  # 0.75 is used in pytorch, it maybe different in other frameworks
+        A = -0.75  # -0.75 is used in pytorch, it maybe different in other frameworks
 
         def cubic_weight_1(fraction):
             return ((A + 2) * fraction - (A + 3)) * fraction * fraction + 1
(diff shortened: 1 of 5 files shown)


ehsanmok:
@tvm-bot rerun

ehsanmok:
Skipping the [flaky test case](https://ci.tlcpack.ai/blue/organizations/jenkins/tvm/detail/PR-13163/2/pipeline/494)

python/tvm/topi/image/grid_sample.py
```diff
@@ -200,13 +200,13 @@ def _bilinear_sample(n, c, h, w):
 
     def _nearest_sample(n, c, h, w):
         y, x = _compute_source_index(n, h, w)
-        y_new = te.round(y).astype("int32")
-        x_new = te.round(x).astype("int32")
+        y_new = te.nearbyint(y).astype("int32")
```
AndrewZhaoLuo, line 203: Can you comment a little about this change? 

The difference is now behavior can change by setting the rounding mode. Is this needed?

tests/python/frontend/onnx/test_forward.py
```diff
@@ -5383,6 +5375,12 @@ def verify_eyelike(indata, dynamic=False):
     ]
 }
 
+flaky_target_skips = {
```
AndrewZhaoLuo, line 5378: Looking at this test: https://ci.tlcpack.ai/blue/organizations/jenkins/tvm/detail/PR-13163/2/tests

It appears the issue is the `nearbyint` rounding, which is only supported for llvm targets.  

What is the expected rounding mode when using nearbyint. Perhaps you can just use the corresponding rounding function directly.

AndrewZhaoLuo (commented):
LGTM for the most part, but have q's on the rounding.

python/tvm/topi/image/grid_sample.py
ehsanmok: The `round` fails as for `nearest` in cpu as interpolation onnx uses [`std::nearbyintf`](https://github.com/microsoft/onnxruntime/blob/62cc927f05580a8875af1fc93363d4b6d34bc2c6/onnxruntime/core/providers/cpu/tensor/grid_sample.cc#L189-L192)

tests/python/frontend/onnx/test_forward.py
ehsanmok: See my comment above

ehsanmok:
@tvm-bot rerun

tests/python/frontend/onnx/test_forward.py
AndrewZhaoLuo, line 5378: Can this be removed now?

ehsanmok: Unfortunately not, bc it fails if cuda is enabled but passes without it. Here's the minor differences between the actual onnx val and tvm val for the `gridsample_nearest` test case on nvptx

```
actual = array([[[[0., 0., 2., 2.],
         [2., 2., 5., 0.]]]], dtype=float32), 
desired = array([[[[0., 0., 2., 3.],
         [3., 2., 5., 0.]]]], dtype=float32), 
```

This could be due to the small discrepancy in the implementations noted [here](https://github.com/apache/tvm/pull/13163#discussion_r1002119311). I think, the best is to track it in an issue for now.

For the record, I'm triggering the CI now to see it without the skip.

ehsanmok: So the CI passed and it's flaky comparing with my local result. I'll have to include the skip and make a note of it.

AndrewZhaoLuo: Yeah interesting, on https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH__DOUBLE.html#group__CUDA__MATH__DOUBLE_1g2316a104cfda8362208d52238181fbfb:

`__device__​ double nearbyint ( double  x )`
`Round argument x to an integer value in double precision floating-point format. Uses round to nearest rounding, with ties rounding to even.`

So it sounds like the CUDA rounding is fixed to round to even. This is odd because this is not a mode for the std::nearbyintf

AndrewZhaoLuo: But yeah, sus, will be on the lookout for this,

PR #13172, opened 2022-10-21
[Hexagon] Adjust Hexagon pytest port range

supersat:
ADB, the Android Debug Bridge, tries to determine if an emulator is present by looking for port 5555 being open. Unfortunately, once adb thinks there's an emulator running, it will continously try to connect to it, adding an error to its log on every failed attempt. This log file quickly grows to hundreds of gigabytes before developers notice they are completely out of disk space. This problem is further compounded by the Hexagon Launcher forwarding 10 contiguous ports from the randomly-chosen RPC base port, causing this failure to happen more often than you'd otherwise suspect.

See https://stackoverflow.com/questions/3152681/android-emulator-5554-offline/10356594#10356594

cc: @csullivan @mehrdadh @cconvey


mehrdadh (approved):
great catch! thanks for the fix!

PR #13174, opened 2022-10-21
[Hexagon] [runtime] Allow creation of thread manager without hardware resources

janetsc:
This allows the creation of a thread manager with optional hardware resources.  This allows us to create a new object for unit tests without hardware resources.

This will fix the test instability we've seen recently, as all tests were using the global thread manager and things could be in an unknown state.

cc: @adstraw, @mehrdadh, @JosephTheOctonaut, @csullivan

### DIFF
--- a/tests/cpp-runtime/hexagon/hexagon_thread_manager_tests.cc
+++ b/tests/cpp-runtime/hexagon/hexagon_thread_manager_tests.cc
@@ -29,15 +29,16 @@ using namespace tvm::runtime::hexagon;
 class HexagonThreadManagerTest : public ::testing::Test {
  protected:
   void SetUp() override {
-    htm = HexagonDeviceAPI::Global()->ThreadManager();
+    // Create with no hardware resources so we don't conflict with session HexagonThreadManager
+    htm = new HexagonThreadManager(threads, stack_size, pipe_size);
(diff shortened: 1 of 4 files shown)


tests/cpp-runtime/hexagon/hexagon_thread_manager_tests.cc
```diff
@@ -54,6 +55,15 @@ TEST_F(HexagonThreadManagerTest, ctor_errors) {
   ASSERT_THROW(HexagonThreadManager(6, stack_size, 9), InternalError);
   // pipe too big
   ASSERT_THROW(HexagonThreadManager(6, stack_size, 0x10000000), InternalError);
+  // hw resources count doesn't match thread count
+  ASSERT_THROW(HexagonThreadManager(6, stack_size, pipe_size, {DMA_0}), InternalError);
+  // hw resources doesn't match specific supported configuration
```
adstraw, line 60: We should come back and revisit this.  Having this _very_ strict specification of hardware resources is probably not what we want long term.  Let's see if we can make this a little more flexible going forward.  That said, I am OK with this PR as an interim step.

adstraw (approved):

driazati (approved):

PR #13175, opened 2022-10-22
[TIR] Add the missing nearbyint intrinsic

ehsanmok:
During testing of this PR #13163, it turns out that the `nearbyint` intrinsic is missing despite being supported.

@junrushao


junrushao (approved):
LGTM!

PR #13177, opened 2022-10-23
[DLPack][runtime] Update DLPack to v0.7

cconvey:
[DLPack][runtime] Update DLPack to v0.7

- Update the `3rdparty/dlpack` git submodule from v0.5 to v0.7, so that the `DLDeviceType` enumeration has an explicitly-stated underlying storage type.  This addresses a compiler warning generated by clang 15.0.3.

- Remove `kDLHexagon` and `kDLWebGPU` from `TVMDeviceExtType`, because those enumerators are now provided by `DLDeviceType`.

- Renumber the members of `TVMDeviceExtType` to reduce the chance of unnoticed collision with members of `DLDeviceType`.

### DIFF
--- a/include/tvm/runtime/c_runtime_api.h
+++ b/include/tvm/runtime/c_runtime_api.h
@@ -80,16 +80,73 @@ extern "C" {
 /*! \brief type of array index. */
 typedef int64_t tvm_index_t;
 
-/*! \brief Extension device types in TVM */
+/*! \brief Extension device types in TVM
+ *
+ * Additional enumerators to supplement those provided by
+ * DLPack's `DLDeviceType` enumeration.
+ *
+ * MAINTAINERS NOTE #1: We need to ensure that the two devices
+ * are identified by the same integer.
+ * Currently this requires manual verification.
+ * Discussed here: https://github.com/dmlc/dlpack/issues/111
+ * As of DLPack v0.7, the highest-valued enumerator in
+ * `DLDeviceType` is kDLHexagon = 16.
+ *
+ * MAINTAINERS NOTE #2: As of DLPack v0.7, the definition for
+ * `DLDeviceType` specifies an underlying storage type of
+ * `int32_t`.  That guarantees a variable of type
+ * `DLDeviceType` is capable of holding any integers provided
+ * by *either* of these enumerations.
+ *
+ * However, the `int32_t` specification only applies when the
+ * header file is compiled as C++, and this header file is also
+ * meant to work as C code.  So the unspecified storage type
+ * could be a latent bug when compiled as C.
+ */
+#ifdef __cplusplus
+typedef enum : int32_t {
+#else
 typedef enum {
-  kDLAOCL = 5,
-  kDLSDAccel = 6,
-  kOpenGL = 11,
-  kDLMicroDev = 13,
-  kDLHexagon = 14,
-  kDLWebGPU = 15
-  // AddExtraTVMType which is not in DLPack here
+#endif
+  // To help avoid accidental conflicts between `DLDeviceType`
+  // and this enumeration, start numbering the new enumerators
+  // a little higher than (currently) seems necessary.
+  kDLAOCL = 32,
+  kDLSDAccel,
+  kOpenGL,
+  kDLMicroDev,
+  TVMDeviceExtType_End,  // sentinel value
 } TVMDeviceExtType;
+
+#ifdef __cplusplus
+// Some other parts of TVM hardcode the integer identifier for
+// some DLPack / TVM devices, rather then using the symbolic
+// enumerator.   E.g., `2` rather than `kDLCUDA`.
+// These asserts should alert us when that mapping breaks.
+#define TVM_HARCODED_INTEGER_CHANGED_MSG                                                          \
+  "Change in compile-time integer.  Make sure hardcoded uses of this integer throughout TVM are " \
+  "updated."
+static_assert(kDLCPU == 1, TVM_HARCODED_INTEGER_CHANGED_MSG);
+static_assert(kDLCUDA == 2, TVM_HARCODED_INTEGER_CHANGED_MSG);
+static_assert(kDLCUDAHost == 3, TVM_HARCODED_INTEGER_CHANGED_MSG);
+static_assert(kDLOpenCL == 4, TVM_HARCODED_INTEGER_CHANGED_MSG);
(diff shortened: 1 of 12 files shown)


cconvey:
CC: @areusch @tqchen @kparzysz-quic

kparzysz-quic (approved):
Cool!  Thanks.

tqchen (approved):

shingjan (approved):

cconvey:
@tvm-bot rerun

cconvey:
FYI CC: @Lunderberg @csullivan

python/tvm/_ffi/runtime_ctypes.py
```diff
@@ -195,44 +195,73 @@ class Device(ctypes.Structure):
...
+    kDLExtDev = 12
+    kDLCUDAManaged = 13
+    kDLOneAPI = 14
+    kDLWebGPU = 15
+    kDLHexagon = 16
+    kDLAOCL = 32
+    kDLSDAccel = 33
+    kOpenGL = 34
+    kDLMicroDev = 35
+
```
csullivan, line 220: Is there a way for us to initialize these from there corresponding DLDeviceType and TVMDeviceExtType enums potentially via a packed function call?

include/tvm/runtime/c_runtime_api.h
```diff
@@ -80,17 +80,74 @@ extern "C" {
...
+ * MAINTAINERS NOTE #2: As of DLPack v0.7, the definition for
+ * `DLDeviceType` specifies an underlying storage type of
+ * `int32_t`.  That guarantees a variable of type
+ * `DLDeviceType` is capable of holding any integers provided
+ * by *either* of these enumerations.
+ *
+ * However, the `int32_t` specification only applies when the
+ * header file is compiled as C++, and this header file is also
+ * meant to work as C code.  So the unspecified storage type
+ * could be a latent bug when compiled as C.
```
csullivan, line 104: What "unspecified" storage type is being referred to here? I didn't understand this comment and therefore couldn't gauge its severity given the reliance of the TVM C-runtime on this header.

jvm/core/src/main/java/org/apache/tvm/Device.java
```diff
@@ -17,32 +17,40 @@
...
 public class Device {
+  /**
+   * Provides the same information as the C++ enums DLDeviceType and
+   * TVMDeviceExtType.
+   */
+  static final int kDLCPU = 1, kDLCUDA = 2, kDLCUDAHost = 3, kDLOpenCL = 4, kDLVulkan = 7,
+                   kDLMetal = 8, kDLVPI = 9, kDLROCM = 10, kDLROCMHost = 11, kDLExtDev = 12,
+                   kDLCUDAManaged = 13, kDLOneAPI = 14, kDLWebGPU = 15, kDLHexagon = 16,
+                   kDLAOCL = 32, kDLSDAccel = 33, kOpenGL = 34, kDLMicroDev = 35;
+
```
csullivan, line 33: Can we initialize these with a backed function call? If we have to hard code it somewhere, better to do it in one place than in many places and make an FFI call to lookup the values.

csullivan (commented):
Many thanks @cconvey! Providing one suggestion that could help for future maintenance.

python/tvm/_ffi/runtime_ctypes.py
cconvey: > Is there a way for us to initialize these from there corresponding DLDeviceType and TVMDeviceExtType enums potentially via a packed function call?

I definitely considered that idea.  The main goal of the PR is to upgrade DLPack, and I felt a significant reworking of how these constants are handled belonged in a separate PR.

include/tvm/runtime/c_runtime_api.h
cconvey: Sorry that the comment wasn't clear!

The two header files in question, [`dlpack.h`](https://github.com/dmlc/dlpack/blob/e2bdd3bee8cb6501558042633fa59144cc8b7f5f/include/dlpack/dlpack.h) and [`include/tvm/runtime/c_runtime_api.h`](https://github.com/apache/tvm/blob/965a51de0b6e515fe0d5a34fba3b8b1ebf69aeb4/include/tvm/runtime/c_runtime_api.h), are meant to be valid C as well as valid C++.

When compiled as C++, these header files _force_ the `DLDeviceType` and `TVMDeviceExtType` enums to use `int32_t` as the underlying storage type.  (See [here](https://github.com/dmlc/dlpack/blob/e2bdd3bee8cb6501558042633fa59144cc8b7f5f/include/dlpack/dlpack.h#L44-L48) and [here](https://github.com/apache/tvm/blob/965a51de0b6e515fe0d5a34fba3b8b1ebf69aeb4/include/tvm/runtime/c_runtime_api.h#L106-L110)).

This ensures that any variable of type `DLDeviceType` or `TVMDeviceExtType` is capable of storing any of the values from those two enumerations.  (And, fortunately, any of the additional values that TVM assumes it can store there, such as `0` and `-1`.)

But when those headers are compiled as _C_ code, those enum declarations _don't_ have that `int32_t` specifier.  IIUC, C11 (the newest standard version of C, I think) [has no such mechanism for specifying the size of an enum](https://stackoverflow.com/questions/4879286/specifying-size-of-enum-type-in-c).

I believe this means the compiler is permitted to pick any underlying storage type for `DLDeviceType` vs. `TVMDeviceExtType` that can accommodate the enums' respective values.  So, for example, `int8_t` for one, `uint32_t` for the other.

Practically speaking, we'd expect the compiler(s) to pick the same storage type for both `DLDeviceType` and `TVMDeviceExtType`, because currently they both have only non-negative integers pretty close to 0.  But the compiler _could_, in principle, use one underlying storage type when compiling DLPack, and another when compiling TVM, possibly leading to an ABI mismatch.  So this is the latent bug to which I referred.

Another consideration is that TVM has code that assumes variables of (one of) these types can store the value -1.  I think this is undefined behavior, and could potentially cause problems with code that compares / tests the values of these variables.

jvm/core/src/main/java/org/apache/tvm/Device.java
cconvey: I completely agree that TVM should have _some_ mechanism to let these integers be defined in only one place.  I think a proper solution would require some discussion, so I'd prefer to treat that as outside the scope of this PR.

cconvey:
@tvm-bot rerun

csullivan:
It's great to have DLPack updated, thank you for working through the issues @cconvey! This is now merged.

PR #13180, opened 2022-10-24
[Hexagon] Add E2E test demonstrating how to apply blocked layout schedule to conv2d via metaschedule

csullivan:
Demonstrates the use of a custom search space generation from a manual written schedule. In this there are no degrees of freedom so tuning will not improve the schedule currently, but the test demonstrates the schedule-time layout transformation of the conv2d primfunc to the 8x8x32 packed layout for convolution ops.

### DIFF
--- a/tests/python/contrib/test_hexagon/metaschedule_e2e/test_resnet50_int8.py
+++ b/tests/python/contrib/test_hexagon/metaschedule_e2e/test_resnet50_int8.py
@@ -261,7 +261,7 @@ def tune_packed_8x8x32_template(mod, params, hexagon_launcher):
         _schedule_packed_8x8x32_conv2d(do_tune=True)(sch, conv2d_block)
         return [sch]
 
-    # register_func("meta_schedule.conv2d_NCHWc_int8", schedule_rule_conv2d_packed_8x8x32)
+    register_func("meta_schedule.conv2d_NCHWc_int8", schedule_rule_conv2d_packed_8x8x32)
 
     def schedule_conv2d_for_tune(sch: Schedule):
         _schedule_packed_8x8x32_conv2d(do_tune=True)(sch)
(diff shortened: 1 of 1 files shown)


masahi:
@tvm-bot rerun

tests/python/contrib/test_hexagon/metaschedule_e2e/test_resnet50_int8.py
```diff
@@ -184,3 +187,186 @@ def test_resnet50(hexagon_launcher):
...
+        np.testing.assert_allclose(ref_result, hexagon_output, atol=1e-4, rtol=1e-5)
+
+        time_ms = graph_mod.benchmark(session.device, number=1, repeat=20).mean * 1e3
+
+        print("time elapsed: ", time_ms)
+
+        debug_ex = session.get_graph_debug_executor(
+            hexagon_lowered.get_graph_json(), hexagon_lowered.lib
+        )
+        print(debug_ex.profile(input_name=inp.copy()))
```
masahi, line 372: Do we want perf measurement and detailed profiling report from this test?

csullivan: Thank you and removed. This is indeed unnecessary for this test.

tests/python/contrib/test_hexagon/metaschedule_e2e/test_resnet50_int8.py
```diff
@@ -184,3 +187,177 @@ def test_resnet50(hexagon_launcher):
...
+            target=target,
+            params=params,
+            work_dir=work_dir,
+            max_trials_global=20000,
+            max_trials_per_task=1,
+            num_trials_per_iter=1,
+            strategy="replay-trace",
+            builder=get_hexagon_local_builder(),
+            runner=get_hexagon_rpc_runner(hexagon_launcher, number=20),
+            # TODO(csullivan): Configrm the below is accurate
```
masahi, line 285: cc @zxybazh so by using `ScheduleFn` here, MS won't be looking at non conv2d blocks?

masahi (approved):
LGTM modulo the confirmation on TODO

tests/python/contrib/test_hexagon/metaschedule_e2e/test_resnet50_int8.py
zxybazh: By using `ScheduleFn` here, the search space is solely generated by the schedule function over the whole workload. In this case, if you're using `_schedule_packed_8x8x32_conv2d`'s `schedule_fn` to construct the `ScheduleFn` space generator, it will only look at conv2d blocks because the function doesn't transform any of the non-conv2d blocks right?

masahi: Yeah that's my understanding.

masahi: @csullivan so shall we remove the TODO comment (which also has a typo lol)

csullivan: :) Yes, but one more question, if one utilizes a `ScheduleFn` to define a search space, but also sets custom `sch_rules` to be applied per block as, 
```
            space=ms.space_generator.ScheduleFn(
                schedule_conv2d_for_tune,
                sch_rules=[some set of rules] , # <<<<
                postprocs=[],
                mutator_probs={},
            ),
```
what will be the interaction between the scheduling function and the rules?

zxybazh: For `ScheduleFn` I don't think there will be any interaction between the scheduling function and the rules, they will only be used for `PostOrderApply` space generator.

masahi: @csullivan can you remove the TODO comment?

tests/python/contrib/test_hexagon/metaschedule_e2e/test_resnet50_int8.py
```diff
@@ -184,3 +187,177 @@ def test_resnet50(hexagon_launcher):
...
+
+def _schedule_packed_8x8x32_conv2d(do_tune: bool):
+    """Manually schedule a conv2d block, created from TE compute op via CreatePrimFunc,
+    using 8x8x32 packed layout.
+    """
+
+    def schedule_fn(sch, conv2d_block: Optional[BlockRV] = None) -> bool:
+        if conv2d_block == None:
+            try:
+                conv2d_block = sch.get_block("conv2d_NCHWc_int8")
```
masahi, line 200: My PR https://github.com/apache/tvm/pull/13206 has `has_block(sch, block_name)` utility, which can remove this hack. I'll update this code after it is merged.

csullivan:
Thanks for patience, let us hold off on merging this one for a bit as there are a few issues that arise at the end of a tuning run with this test:
1. Lack of JSON serialization for tir::IndexMap in src/meta_schedule/database/database_utils.cc::JSONDumps
2. Metascheduler search workers hang when the search space has only one sample.

zxybazh:
Hi Chris, regarding (2) does the worker hang when the search space that has only one sample, and will not terminate by itself even after 5 iterations?

csullivan:
Thanks @zxybazh for following the thread closely. Looks like (2) was a red herring and is not reproducible for me any longer. Furthermore, (1) is coming in a follow up PR and this test is skipped so no harm merging this now.

PR #13181, opened 2022-10-24
[Hexagon] [runtime] Per-thread hardware resource management

janetsc:
Add support to lock and unlock resources on the individual threads.  Adds APIs to get a thread handle for a given resource type, and get a resource type for a given thread handle.  This will aid in asynchronous support that is sending work to dedicated threads for the hardware resources.

There are two steps to fully acquire a resource:  reserve and lock.  HTP/HVX manager objects are created on the global thread, which will reserve the resources.  The lock/unlock will occur on the thread that owns a particular instance.  The manager objects will be deconstructed when the thread manager is deconstructed.  This will release the resources.

cc:  @csullivan, @adstraw, @JosephTheOctonaut, @kparzysz-quic

### DIFF
--- a/src/runtime/hexagon/hexagon_hvx.cc
+++ b/src/runtime/hexagon/hexagon_hvx.cc
@@ -39,7 +39,6 @@ HexagonHvx::HexagonHvx() {
 }
 
 HexagonHvx::~HexagonHvx() {
-  // Release HVX.
   int rel = qurt_hvx_cancel_reserve();
   CHECK(rel == 0) << "error releasing HVX: " << rel;
 }
@@ -50,7 +49,6 @@ void HexagonHvx::Lock() {
 }
 
 void HexagonHvx::Unlock() {
-  // Unlock HVX.
   int unl = qurt_hvx_unlock();
   CHECK(unl == 0) << "error unlocking HVX: " << unl;
 }
--- a/src/runtime/hexagon/hexagon_hvx.cc
+++ b/src/runtime/hexagon/hexagon_hvx.cc
@@ -37,9 +37,7 @@ HexagonHvx::~HexagonHvx() { Release(); }
 
 void HexagonHvx::Acquire() {
   reserved_count_ = qurt_hvx_reserve(QURT_HVX_RESERVE_ALL);
-  CHECK((reserved_count_ == QURT_HVX_RESERVE_ALL) ||
-        (reserved_count_ == QURT_HVX_RESERVE_ALREADY_MADE))
-      << "error reserving HVX: " << reserved_count_;
+  CHECK(reserved_count_ == QURT_HVX_RESERVE_ALL) << "error reserving HVX: " << reserved_count_;
 }
 
 void HexagonHvx::Release() {
--- a/src/runtime/hexagon/hexagon_hvx.cc
+++ b/src/runtime/hexagon/hexagon_hvx.cc
@@ -32,24 +32,27 @@ namespace runtime {
 namespace hexagon {
 
 HexagonHvx::HexagonHvx() {
-  // Reserve HVX.
-  int res = qurt_hvx_reserve(QURT_HVX_RESERVE_ALL_AVAILABLE);
-  CHECK((res != QURT_HVX_RESERVE_NOT_SUPPORTED) && (res != QURT_HVX_RESERVE_NOT_SUCCESSFUL))
-      << "error reserving HVX: " << res;
+  reserved_count_ = qurt_hvx_reserve(QURT_HVX_RESERVE_ALL);
+  CHECK((reserved_count_ == QURT_HVX_RESERVE_ALL) ||
+        (reserved_count_ == QURT_HVX_RESERVE_ALREADY_MADE))
+      << "error reserving HVX: " << reserved_count_;
+}
 
-  // Lock HVX.
+HexagonHvx::~HexagonHvx() {
+  // Release HVX.
+  int rel = qurt_hvx_cancel_reserve();
+  CHECK(rel == 0) << "error releasing HVX: " << rel;
+}
+
+void HexagonHvx::Lock() {
   int lck = qurt_hvx_lock(QURT_HVX_MODE_128B);
   CHECK(lck == 0) << "error locking HVX: " << lck;
 }
 
-HexagonHvx::~HexagonHvx() {
+void HexagonHvx::Unlock() {
   // Unlock HVX.
   int unl = qurt_hvx_unlock();
   CHECK(unl == 0) << "error unlocking HVX: " << unl;
-
-  // Release HVX.
-  int rel = qurt_hvx_cancel_reserve();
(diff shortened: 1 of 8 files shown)


src/runtime/hexagon/hexagon_hvx.cc
```diff
@@ -31,25 +31,30 @@ namespace tvm {
...
-  int res = qurt_hvx_reserve(QURT_HVX_RESERVE_ALL_AVAILABLE);
-  CHECK((res != QURT_HVX_RESERVE_NOT_SUPPORTED) && (res != QURT_HVX_RESERVE_NOT_SUCCESSFUL))
-      << "error reserving HVX: " << res;
+HexagonHvx::HexagonHvx() { Acquire(); }
 
-  // Lock HVX.
+HexagonHvx::~HexagonHvx() { Release(); }
+
+void HexagonHvx::Acquire() {
+  reserved_count_ = qurt_hvx_reserve(QURT_HVX_RESERVE_ALL);
```
janetsc, line 39: This explicitly reserves all of the HVX units.  Depending on schedules, we may want to request fewer HVX units so a future change could include requesting only the number of HVX instances needed.  Thoughts on this approach?

src/runtime/hexagon/hexagon_hvx.cc
```diff
@@ -31,25 +31,30 @@ namespace tvm {
...
+  CHECK((reserved_count_ == QURT_HVX_RESERVE_ALL) ||
+        (reserved_count_ == QURT_HVX_RESERVE_ALREADY_MADE))
```
janetsc, line 41: I also considered making it a failure if another instance of HexagonHvx

janetsc:
@tvm-bot run

janetsc:
@tvm-bot run

src/runtime/hexagon/hexagon_hvx.cc
csullivan: I think we'll always want to own the set of available HVX coprocessors and so this is good.

src/runtime/hexagon/hexagon_hvx.cc
csullivan: Yeah I agree that this should probably be an error if `reserved_count_ == QURT_HVX_RESERVE_ALREADY_MADE`.

src/runtime/hexagon/hexagon_thread_manager.cc
```diff
@@ -168,7 +180,8 @@ void HexagonThreadManager::SpawnThreads(unsigned thread_stack_size_bytes,
     next_stack_start += thread_stack_size_bytes;
 
     // create the thread
-    contexts_[i] = new ThreadContext(&pipes_[i], i);
+    contexts_[i] = new ThreadContext(&pipes_[i], i, hw_resources_.empty() ? NONE : hw_resources_[i],
+                                     hvx_.get(), htp_.get());
```
csullivan, line 184: So if we are binding resources to the thread contexts, will they no longer be available to the global thread? Because we don't have lowering support to the HTM yet, we still need these resources availble to the global thread. Do you have thoughts on how to transition the runtime from one model to the other (global thread -> stream based backed by the HTM)?

src/runtime/hexagon/hexagon_hvx.cc
csullivan: Oh wait, does every thread main call `qurt_hvx_reserve(QURT_HVX_RESERVE_ALL);` because each owns an HexagonHvx?

src/runtime/hexagon/hexagon_thread_manager.cc
csullivan: Oh, I think I see, right now the global thread is acquiring them, and then you are passing in a pointer to the HexagonHtp or HexagonHvx. And later on you plan to acquire them (construct) directly in thread_main?

csullivan (commented):
Great! Just a few questions

src/runtime/hexagon/hexagon_hvx.cc
janetsc: No, the main thread constructs a HexagonHvx object, which calls qurt_hvx_reserve.  Each thread assigned to an HVX_x instance calls Lock();

src/runtime/hexagon/hexagon_thread_manager.cc
janetsc: I construct on the global thread.  I use the pointers to the object to then call Lock() on the actual threads

src/runtime/hexagon/hexagon_hvx.cc
janetsc: Will push an update - thanks!

src/runtime/hexagon/hexagon_thread_manager.cc
janetsc: I'll add a comment on the manager object construction which will hopefully make this more clear.

csullivan (approved):
LGTM :shipit:

src/runtime/hexagon/hexagon_hvx.cc
```diff
@@ -31,25 +31,28 @@ namespace tvm {
...
+  int rel = qurt_hvx_cancel_reserve();
+  CHECK(rel == 0) << "error releasing HVX: " << rel;
+}
+
+void HexagonHvx::Lock() {
   int lck = qurt_hvx_lock(QURT_HVX_MODE_128B);
   CHECK(lck == 0) << "error locking HVX: " << lck;
 }
 
-HexagonHvx::~HexagonHvx() {
```
nverke, line 45: This destructor unlocks and then releases HVX, but the new one just releases. Do we not need to unlock upon destruction?

src/runtime/hexagon/hexagon_thread_manager.cc
```diff
@@ -38,18 +38,15 @@ HexagonThreadManager::HexagonThreadManager(unsigned num_threads, unsigned thread
...
   hw_resources_ = hw_resources;
+  CheckResources();
 
-  if (!hw_resources_.empty()) {
+  if (create_resource_managers_) {
     DLOG(INFO) << "Initialize hardware resource managers";
-    // Acquisition/locks will be performed on specific threads
+    // This creates the manager objects, which reserves (acquires) the resources.
+    // Calls to lock/unlock will be performed on threads dedicated to instances.
+    // This must be done before spawing threads so we can pass pointers to the
```
nverke, line 48: nit: spawning*

src/runtime/hexagon/hexagon_hvx.cc
janetsc: This is by design.  Reserve/release happen in the constructor/destructors, and lock/unlock must happen on the thread that is dedicated to that resource.  In the case of HVX, it will be called 4 times.

You bring up a good point, which is what happens if one of the threads crashes before it can unlock?  I test offline.  (There's not really a good way to incorporate that test, because other tests that run in parallel depend on the resources being acquired.)

PR #12935, opened 2022-09-28
[Relay][Topi] Hook up LayerNorm to new Topi

AndrewZhaoLuo:
This makes layer_norm relay op dispatch to new topi committed in PR #12864. 

Using TIR CSE elimination pass with FP16 layernorm also necessitates the handling of FP16 type when packing args for CUDA.

PR #13182, opened 2022-10-24
[Relay] Rewrite division by constant to multiply

tkonolige:
Convert division by a scalar constant into multiplication by the inverse of the constant. Multiplication is faster than division and also allow for more optimization opportunities. Only applies to float32 and float64.

I wrote this as a separate pass. Is there a better way to do it?

@AndrewZhaoLuo

### DIFF
--- a/src/relay/transforms/div_to_mul.cc
+++ b/src/relay/transforms/div_to_mul.cc
@@ -19,6 +19,7 @@
 #include <tvm/relay/expr.h>
 #include <tvm/relay/expr_functor.h>
 #include <tvm/relay/transform.h>
+#include <tvm/runtime/builtin_fp16.h>
 
 #include "pattern_utils.h"
 
@@ -48,6 +49,14 @@ class DivToMulRewrite : public MixedModeMutator {
               return post;
             }
             static_cast<double*>(inv->data)[0] = 1. / rhs_val;
+          } else if (dtype == "float16") {
+            // Do f16 math in f32
+            float rhs_val = __gnu_h2f_ieee(static_cast<uint16_t*>(rhs->data->data)[0]);
+            // Check for division by zero
+            if (rhs_val == 0.) {
+              return post;
+            }
+            static_cast<uint16_t*>(inv->data)[0] = __gnu_f2h_ieee(1. / rhs_val);
           } else {
             // Cannot do 1/int because it will truncate
             return post;
--- a/src/relay/transforms/div_to_mul.cc
+++ b/src/relay/transforms/div_to_mul.cc
@@ -35,9 +35,19 @@ class DivToMulRewrite : public MixedModeMutator {
               runtime::NDArray::Empty(rhs->data.Shape(), rhs->data.DataType(), rhs->data->device);
           std::string dtype = DLDataType2String(rhs->data.DataType());
           if (dtype == "float32") {
-            static_cast<float*>(inv->data)[0] = 1. / static_cast<float*>(rhs->data->data)[0];
+            float rhs_val = static_cast<float*>(rhs->data->data)[0];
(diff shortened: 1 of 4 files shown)


src/relay/transforms/div_to_mul.cc
```diff
@@ -0,0 +1,67 @@
...
+namespace tvm {
+namespace relay {
+
+class DivToMulRewrite : public MixedModeMutator {
+  Expr Rewrite_(const CallNode* pre, const Expr& post) final {
+    if (const CallNode* call_node = post.as<CallNode>()) {
+      if (call_node->op == Op::Get("divide")) {
+        auto rhs = call_node->args[1].as<ConstantNode>();
+        if (rhs != nullptr) {
+          auto inv =
```
AndrewZhaoLuo, line 34: Nit: might be good to test for division by 0.

src/relay/transforms/div_to_mul.cc
```diff
@@ -0,0 +1,67 @@
...
+              runtime::NDArray::Empty(rhs->data.Shape(), rhs->data.DataType(), rhs->data->device);
+          std::string dtype = DLDataType2String(rhs->data.DataType());
+          if (dtype == "float32") {
```
AndrewZhaoLuo, line 37: Can you handle the float16 case by using `__gnu_f2h_ieee` ? e.g. see https://github.com/apache/tvm/pull/12935/files

src/relay/transforms/div_to_mul.cc
```diff
@@ -0,0 +1,67 @@
...
+#include "pattern_utils.h"
+
+namespace tvm {
+namespace relay {
+
+class DivToMulRewrite : public MixedModeMutator {
+  Expr Rewrite_(const CallNode* pre, const Expr& post) final {
+    if (const CallNode* call_node = post.as<CallNode>()) {
+      if (call_node->op == Op::Get("divide")) {
+        auto rhs = call_node->args[1].as<ConstantNode>();
```
AndrewZhaoLuo, line 32: might want to also handle the reverse case where lhs is constant (and add test)

AndrewZhaoLuo (commented):
LGTM, some comments though.

src/relay/transforms/div_to_mul.cc
tkonolige: If the divisor is not constant you have to do a division. For example, there is no way to rewrite `1.5/x` to avoid the division.

AndrewZhaoLuo: ah yes i forgot it is not commutative.

src/relay/transforms/div_to_mul.cc
AndrewZhaoLuo: Im saying do the division in FP32 and then convert to FP16 bits using the function.

AndrewZhaoLuo: Alternative would also be amenable

tkonolige: Done

AndrewZhaoLuo (approved):

AndrewZhaoLuo:
@tvm-bot rerun

tkonolige:
@tvmbot re-run

AndrewZhaoLuo:
@tvm-bot merge

Johnson9009:
@tkonolige you forgot to judge the operand of divide is "a **scalar** constant". @AndrewZhaoLuo

AndrewZhaoLuo:
Oops sorry about that. 

https://github.com/apache/tvm/pull/14382 

Fixes it by moving to using standard relay semantics so we have to think less of the two cases.

PR #13190, opened 2022-10-25
[TVMScript] TIR parser

cyx-6:
This PR introduces parser for TIR as part of #12442

Co-authored-by: yongwww <yongcale@gmail.com>

### DIFF
--- a/python/tvm/script/_parser/core/parser.py
+++ b/python/tvm/script/_parser/core/parser.py
@@ -571,6 +571,21 @@ class Parser(doc.NodeVisitor):
         """
         return _dispatch(self, "Assign")(self, node)
 
+    def visit_AnnAssign(self, node: doc.AnnAssign) -> Any:  # pylint: disable=invalid-name
+        """The general annotated assign visiting method.
+
+        Parameters
+        ----------
+        node : doc.Assign
+            The doc AST annotated assign node.
+
+        Returns
+        -------
+        res : Any
+            The visiting result.
+        """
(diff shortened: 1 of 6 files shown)


Hzfengsy (approved):
LGTM

MasterJH5574 (approved):
LGTM! Thanks @cyx-6 for the upstreaming effort!

junrushao:
Thanks for the great contribution!

I recognize that most of the TIR syntax have been covered by the IRBuilder tests accordingly, for example, `T.block` has been carefully tested in IRBuilder.

This decoupling approach is great because it largely simplifies testing effort, avoids end-to-end integration tests and favors more fine grained unittests. Therefore, given this PR is tested against all the nuanced cases not covered by the IRBuilder tests, so i would say the test coverage is good enough.

On the other hand, I do think it’s desirable for us to see some end-to-end tests to make sure the parser is functioning properly, especially when e2e tests are running super fast for in the parser. A couple of test cases I would love to see in a follow-up PR:
- `1 + 1` is not simplified to `2`, but honestly translated to `tvm.tir.Add`
- `T.Buffer()` in function signature is correctly parsed
- Parser error is properly reported in IRBuilder calls, for example, calling an non-existing intrinsic.

Would be great if we could have a follow-up PR for such tests! Thanks a lot!

MasterJH5574:
@junrushao Thanks for bring it up! I agree that it makes sense to follow up with more comprehensive test cases.

PR #13192, opened 2022-10-25
[TIR][Schedule] Add cache_index to precompute index of buffer load

multiverstack-intellif:
Buffer load index can be precomputed and saved in an index buffer, to prevent multiple index computation.
@Hzfengsy @wrongtest-intellif


Hzfengsy (approved):
LGTM

PR #13194, opened 2022-10-25
[TIR] Add utility for anchor block extraction

masahi:
I'm working on enabling "anchor block" tuning for MS. This is a utility to extract anchor blocks. 

I define the "anchor block" to be the block (1) with an init statement and (2) having the biggest flops count. The latter condition is only used when there are multiple blocks with an init statement.

For example, if the input module is conv2d + fused spatial blocks, conv2d is the anchor block. A module created from winograd convolution has multiple blocks with an init statement (input transform, batched GEMM, and output transform). We use the second condition, the flops count, to determine that the batched GEMM block is the anchor block.

cc @junrushao @Hzfengsy

### DIFF
--- a/src/tir/schedule/analysis.h
+++ b/src/tir/schedule/analysis.h
@@ -72,15 +72,6 @@ const PrimFuncNode* GetRootPrimFunc(const IRModule& mod, const StmtNode* root_bl
  */
 StmtSRef GetSRefTreeRoot(const StmtSRef& sref);
 
-/*!
- * \brief Find the entry function of the given IRModule, i.e, functions marked by
- * `tir::attr::kIsEntryFunc`, whose name is `main` or being the only PrimeFunc.
- * \param mod The IRModule to find the entry function.
- * \param result_g_var The result GlobalVar of the entry function.
- * \return The entry function.
- */
-const PrimFuncNode* FindEntryFunc(const IRModule& mod, GlobalVar* result_g_var);
-
 /******** Scope ********/
 /*!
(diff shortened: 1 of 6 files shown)


src/tir/schedule/analysis/analysis.cc
```diff
@@ -46,47 +46,6 @@ const PrimFuncNode* GetRootPrimFunc(const IRModule& mod, const StmtNode* root_bl
   throw;
 }
 
-const PrimFuncNode* FindEntryFunc(const IRModule& mod, GlobalVar* result_g_var) {
```
masahi, line 49: Moved to `src/tir/analysis/stmt_finding.cc`

src/tir/schedule/analysis.h
```diff
@@ -72,15 +72,6 @@ const PrimFuncNode* GetRootPrimFunc(const IRModule& mod, const StmtNode* root_bl
...
 StmtSRef GetSRefTreeRoot(const StmtSRef& sref);
 
-/*!
- * \brief Find the entry function of the given IRModule, i.e, functions marked by
- * `tir::attr::kIsEntryFunc`, whose name is `main` or being the only PrimeFunc.
- * \param mod The IRModule to find the entry function.
- * \param result_g_var The result GlobalVar of the entry function.
- * \return The entry function.
- */
-const PrimFuncNode* FindEntryFunc(const IRModule& mod, GlobalVar* result_g_var);
```
masahi, line 82: Moved to `include/tir/analysis.h`

junrushao (approved):
LGTM!

PR #13195, opened 2022-10-25
[MetaSchedule] Refactor ScheduleRule Attributes

junrushao:
This PR refactors the existing `schedule_rule` block annotation-based dispatch into a more organized folder structure, which follows the exact same mechanism as AutoTVM.

In the example below, when target is `Target("llvm -keys=cuda,gpu")` and the block annotation is `conv2d_nchw_winograd_inverse`:

```python
with T.block("some_block"):
  T.block_attr({
    "scehdule_rule": "conv2d_nchw_winograd_inverse",
  })
```

the system will find the following global packed functions in order:
- `meta_schedule.cuda.conv2d_nchw_winograd_inverse`
- `meta_schedule.gpu.conv2d_nchw_winograd_inverse`
whose function signatures are:

```python
def schedule_rule(
  sch: tvm.tir.Schedule,
  block: tvm.tir.schedule.BlockRV,
) -> List[tvm.tir.Schedule]:
```

In terms of code organization, for example, for target key `cuda`, the schedule functions are supposed to be defined in:
- `include/tvm/meta_schedule/schedule/cuda` for public methods
- `src/meta_schedule/schedule/cuda` for private methods
- `python/tvm/meta_schedule/schedule/cuda` for direct python side definition


spectrometerHBH (approved):

python/tvm/relay/op/strategy/cuda.py
```diff
@@ -166,9 +165,34 @@ def conv2d_strategy_cuda(attrs, inputs, out_type, target):
                     wrap_topi_schedule(topi.cuda.schedule_conv2d_nchw),
                     name="conv2d_nchw.cuda",
                 )
-            _, _, kh, kw = get_const_tuple(kernel.shape)
-            if (
-                (2 < kh < 8 and 2 < kw < 8 and kh == kw)
+            N, _, H, W = get_const_tuple(data.shape)
+            CO, CI, KH, KW = get_const_tuple(kernel.shape)
+            (_, _, judge_winograd_auto_scheduler) = judge_winograd(
```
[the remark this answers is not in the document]
junrushao: certainly - it definitely sounds better

python/tvm/topi/nn/conv2d.py
```diff
@@ -989,6 +989,119 @@ def unpack_NCHWc_to_nchw(packed_out, out_dtype):
...
+        The layout after auto-scheduler's layout rewrite pass.
+    meta_schedule_original_shape: Optional[List[PrimExpr]] = None
+        The original shape of the input tensor.
+
+    Returns
+    -------
+    output : tvm.te.Tensor
+        4-D with shape [batch, out_height, out_width, out_channel]
+    """
+    tile_size = 4
```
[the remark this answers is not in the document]
junrushao: not for simplicity but for consistency with auto-scheduler's nhwc case a couple of lines above. note that this configuration is also overridden by cuda dispatch, so we dont have to worry too much about it

python/tvm/meta_schedule/search_strategy/evolutionary_search.py
```diff
@@ -58,7 +58,7 @@ class EvolutionarySearch(SearchStrategy):
     def __init__(
         self,
         *,
-        population_size: int = 2048,
+        population_size: int = 512,
```
masahi, line 61: cc @zxybazh This change so small and looks unrelated to this PR, but it has huge implications. Was there enough discussion to make this change? Tuning time will become much shorter, which I like, but could there be a concern for perf regression due to this change?

junrushao: @masahi Ask for asking, and it's definitely a valid concern. The reason is that we found it's somehow set to an unreasonably large number (we were consistently using 512 on downstream), and now wanted to scale it back. Also note that the population calculation is slightly different than AutoScheduler, which includes invalid candidates in population, while ours always prune ahead of time.

We did make sure we have proper numbers in hand before merging, and please refer to the table below for details:

|                     | Mainline (ms) | This PR (ms) | Difference |
|---------------------|---------------|--------------|------------|
| resnet_50      |   1.829128008 |   1.73111331 |      5.66% |
| mobilenet_v2   |  0.4773168361 | 0.4806015715 |     -0.68% |
| resnet_18      |   0.678600832 | 0.6301909627 |      7.68% |
| mobilenet_v3   |  0.6649458484 |  0.668666894 |     -0.56% |
| wide_resnet_50 |   3.753781691 |  3.077549256 |     21.97% |
| densenet_121   |   2.361973117 |  2.288411393 |      3.21% |
| inception_v3   |    3.53431975 |  3.478368823 |      1.61% |
| resnet3d_18    |   7.791658449 |  7.465777971 |      4.36% |

src/meta_schedule/schedule_rule/apply_custom_rule.cc
```diff
@@ -0,0 +1,92 @@
...
+  ObjectPtr<ApplyCustomRuleNode> n = make_object<ApplyCustomRuleNode>();
+  return ScheduleRule(n);
+}
+
+bool ScheduleRule::IsApplyCustomRule(const ScheduleRule& rule) {
+  return rule->IsInstance<ApplyCustomRuleNode>();
+}
+
+TVM_REGISTER_NODE_TYPE(ApplyCustomRuleNode);
+TVM_REGISTER_GLOBAL("meta_schedule.ScheduleApplyCustomRule")
```
masahi, line 88: Is this a typo? Shouldn't it be `ScheduleRuleApplyCustomRule`.

`schedule_rule/apply_custom_rule.py` refers to `ScheduleRuleApplyCustomRule` that doesn't exist.

src/meta_schedule/space_generator/post_order_apply.cc
```diff
@@ -150,33 +141,13 @@ class PostOrderApplyNode : public SpaceGeneratorNode {
...
-        Array<tir::Schedule> applied{nullptr};
-        if (sch_rule.defined()) {
-          applied = sch_rule.value()->Apply(sch, /*block=*/block_rv);
-        } else {
-          ICHECK(custom_schedule_fn)
-              << "ValueError: Custom schedule rule not found: " << ann.value();
-          applied = (*custom_schedule_fn)(sch, block_rv);
+        if (!ScheduleRule::IsApplyCustomRule(sch_rule)) {
+          if (tir::GetAnn<String>(sch->GetSRef(block_rv), "schedule_rule").defined()) {
+            stack.emplace_back(sch, blocks);
```
masahi, line 146: It seems this change broke auto tensorization for VNNI and Hexagon `vrmpy`. They target the TE compute that happens to be annotated with `schedule_rule` (specifically, NCHWc int8 conv2d https://github.com/apache/tvm/blob/main/python/tvm/topi/nn/conv2d.py#L534-L552). After this change, `MultiLevelTilingWithIntrin` will never be applied to the block corresponding to this te compute.

So for auto tensorization, we just want to ignore this annotation.

src/meta_schedule/schedule_rule/apply_custom_rule.cc
zxybazh: I think this should be a typo, given it's part of a `ScheduleRule` and the ffi function call [here](https://github.com/apache/tvm/blob/b16a64d6edb9fd1a014fc51995dff7d0e2f4c84e/python/tvm/meta_schedule/schedule_rule/apply_custom_rule.py#L32) on the python side doesn't have corresponding registry.

junrushao: Yep this is a typo. Feel free to send a quick patch :-)

src/meta_schedule/space_generator/post_order_apply.cc
junrushao: Ah I see! I was not aware the usecase where schedule_rule is hardcoded but **intentionally** ignored. In this particular case, how about we add a line below in `ApplyCustomRule`:

```C++
sch->Unannotate(block, "schedule_rule");
```

src/meta_schedule/schedule_rule/apply_custom_rule.cc
zxybazh: Just sent a patch #13353

src/meta_schedule/space_generator/post_order_apply.cc
masahi: Thanks `Unannotate` fixed the issue.

src/te/operation/create_primfunc.cc
```diff
@@ -107,24 +108,30 @@ class LayoutFreePlaceholdersNormalizer : public StmtMutator {
...
+    if (Optional<ObjectRef> ann = n->annotations.Get(topi_attr)) {
       for (Buffer buffer : Downcast<Array<Buffer>>(ann)) {
         auto it = buffer2index_.find(buffer);
         if (it != buffer2index_.end()) {
           layout_free_buffer_indices_.insert(it->second);
-        } else {
-          new_buffers.push_back(buffer);
         }
       }
-      block.CopyOnWrite()->annotations.Set(topi_attr, new_buffers);
```
masahi, line 120: Is this change intentional? Just found that this change broke `RewriteLayout` when `link-params = True`.

junrushao: The problem is that the Buffer objects in the annotation will not be updated when doing transformation, and that’s why I removed this line. I was not aware this is useful for link-param-specific hack.

Feel free to add it back if it’s needed. Shall we add a flag in pass config which defaults to off, and hexagon-specific users could turn it on when needed?

masahi: I see, for now I reverted this change in https://github.com/apache/tvm/pull/13384 to unblock tuning with `link params`. 

I'd say the same problem applies to the current approach which specifies the layout free buffers by their indices in the parameter list, since it is also technically possible (although unlikely) that some transformation may change the signature of the prim func, which invalidates the parameter indices.

PR #13196, opened 2022-10-25
[Fix,Auto_scheduler] Default to extent of 1 if extent cannot be determined

tkonolige:
This fixes a bug in autoscheduler featurization.


junrushao (approved):
LGTM

tkonolige:
@tvm-bot rerun

merrymercy (approved):

MrJungle1:
This is a job that works fine for me, but I have a question, why is there a situation where the scope cannot be determined here? Is it a lack of consideration to return it directly to 1 here?

PR #13198, opened 2022-10-25
[MetaSchedule][Minor] Fix Memory Database Module Equality

zxybazh:
Previously the memory database assumed the workload tokens are from `CommitWorkload` function, this PR allows the use of module equality function in `GetTopK` function to compare workloads.

Thanks to @petersalas for point it out.

### DIFF
--- a/src/meta_schedule/database/memory_database.cc
+++ b/src/meta_schedule/database/memory_database.cc
@@ -71,7 +71,8 @@ class MemoryDatabaseNode : public DatabaseNode {
       if (run_secs.empty()) {
         continue;
       }
-      if (WorkloadEqual(GetModuleEquality())(record->workload, workload)) {
+      if (record->workload.same_as(workload) ||
+          WorkloadEqual(GetModuleEquality())(record->workload, workload)) {
         double sum = 0.0;
         for (const FloatImm& i : run_secs) {
           sum += i->value;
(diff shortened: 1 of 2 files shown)


zxybazh:
CC @junrushao @masahi

src/meta_schedule/database/memory_database.cc
```diff
@@ -71,7 +71,7 @@ class MemoryDatabaseNode : public DatabaseNode {
       if (run_secs.empty()) {
         continue;
       }
-      if (record->workload.same_as(workload)) {
+      if (WorkloadEqual(GetModuleEquality())(record->workload, workload)) {
```
masahi, line 74: What does it fix? They don't seem equivalent (pointer vs structural equality)

zxybazh: Here we wanted to find the tuning records for the structrually equal module, but here instead of doing WorkloadEquality comparsion, it's directly doing pointer comparison, which doesn't seem to be correct.

masahi (approved):

src/meta_schedule/database/memory_database.cc
petersalas, line 74: Would it be practical to have a common suite of `Database` tests that can be run against each implementation to test common invariants?

zxybazh: Hi Peter, we do have unit tests for MetaSchedule Database [here](https://github.com/apache/tvm/blob/23c2909f2917fe9fbccad8e440199e5e4a106ba9/tests/python/unittest/test_meta_schedule_database.py) but there's not many cases against MemoryDatabase. Will add a regression test in this PR.

junrushao (approved):
Merging this in as looks like all comments are addressed

Issue #13204, opened 2022-10-26
[Bug] MetaSchedule produce incorrect result when TensorCore is used


wllvcxz:
The `DefaultCUDATensorCore` schedule rule currently support wmma with dtype `fp16fp16fp16` and `s8s8s32`: https://github.com/apache/tvm/blob/main/src/meta_schedule/schedule_rule/schedule_rule.cc#L122-L171

When I use metaschedule to tune a matmul with DefaultCUDATensorCore, the result seems incorrect when TensorCore is used, however when I changed the input/output dtype of matmul which wmma intrinsic in DefaultCUDATensorCore do not support, the result is correct.

### Expected behavior

metaschedule produce correct result when using tensorcore

### Actual behavior

the result is wrong

### Environment

Operating System: Ubnutu-20.04
TVM version: main branch
GPU: nvidia-a100

### Steps to reproduce

```python
import tempfile
import numpy as np
import tvm
from tvm.contrib import nvcc
from tvm import meta_schedule as ms
from tvm.meta_schedule import tune_tir
from tvm.target import Target
from tvm.meta_schedule.testing import te_workload

def test_tune_tir_matmul_cuda_tensor_core(in_dtype, out_dtype, n, m, k):
    mod = tvm.te.create_prim_func(
        te_workload.matmul(n, m, k, in_dtype=in_dtype, out_dtype=out_dtype)
    )
    target = Target("nvidia/nvidia-a100")

    with tempfile.TemporaryDirectory() as work_dir:
        database = tune_tir(
            mod=mod,
            target=target,
            work_dir=work_dir,
            num_trials_per_iter=32,
            max_trials_global=32,
            strategy="replay-trace",
        )
        sch = ms.tir_integration.compile_tir(database, mod, target)
        if sch is None:
            raise RuntimeError("No valid schedule found!")
        ctx = tvm.cuda()
        if nvcc.have_tensorcore(ctx.compute_version):
            with tvm.transform.PassContext():
                func = tvm.build(sch.mod["main"], [], "cuda")
                # print(func.imported_modules[0].get_source())
                # print(sch.mod.script())
                print(sch.trace)
            a_np = np.random.uniform(-10, 10, size=(n, k)).astype(in_dtype)
            b_np = np.random.uniform(-10, 10, size=(k, m)).astype(in_dtype)
            a = tvm.nd.array(a_np, ctx)
            b = tvm.nd.array(b_np, ctx)
[...]

PR #13206, opened 2022-10-26
[MetaSchedule] Enable anchor-block tuning

masahi:
Note: Most diffs are from test cases, which are bloated due to many TVMScript modules.

Building on the notion of "module equality" introduced in https://github.com/apache/tvm/pull/13050, I'm adding a new variant of module equality based on the "anchor blocks". I defined the anchor block in https://github.com/apache/tvm/pull/13194. 

Currently, MS does tuning at the level of subgraphs. So for example in resnet where there are `conv2d -> add -> relu` and `conv2d -> add -> add -> relu` subgraphs, the two subgraphs are treated as distinct tuning tasks even if the anchor block conv2d workloads are identical. The new module equality will identify them as equal, so it reduces the number of tuning tasks if shorter tuning time is preferred over subgraph-level performance. Currently there is no dedicated API for anchor-block tuning - passing `module_equality="anchor-block"` to task extraction and `tune_relay` will enable it. 

This is particularly effective for int8 models, since each conv2d / dense (anchor) op is quantized slightly differently, ending up many similar but not identical elemwise ops fused after the anchor blocks. On the int8 resnet50 model I tested, it reduced the number of conv2d tuning tasks from 36 to 23.   

The interesting question is the performance difference between full subgraph-level tuning and anchor-block tuning. To experiment this, I tested the int8 resnet50 mentioned above, where anchor-based tuning makes the most difference in terms of the # of extracted tasks. The results are summarized below. 

`num_iter_per_task = 32`, and `max_iters_per_task = 128` in all cases.
| Target | Anchor tuning | Subgraph tuning |
| ---- | ---- | ---- |
| x86 VNNI | 4.45 | 4.58 |
| Hexagon | 58.4 | 58.1 |
|CUDA tensor core (batch size 16) | 6.0 | 6.7 (TODO: Try again)|

From the model + target combinations I tested, I didn't see much perf difference beyond those from natural tuning flaky-ness. The tensorcore result is a bit weird and needs more investigation, but I found that tuning this model using int8 tensor core auto tensorization is incredibly slow currently (for example, getting the 6.7 result took 12 hours). There is also the correctness issue discussed in https://github.com/apache/tvm/issues/13204. So I haven't done more experiments on tensorcore.

### Applying the trace tuned on an anchor block to the target block 
The tricky problem that this work address is the application of a trace, which is tuned on a "representative" anchor subgraph, to the target mod which has different post blocks. Note that, in the resnet example where there are `conv2d -> add -> relu` and `conv2d -> add -> add -> relu` subgraphs, we tune the "smaller" `conv2d -> add -> relu` subgraph, not just the pure anchor block `conv2d`. 

So while applying a trace tuned on just `conv2d` to `conv2d -> add` would be trivial (the existing `Trace::ApplyToSchedule` would just work), in practice we would be applying a trace tuned on `conv2d -> add` to `conv2d -> subtract`, for example. My proposed solution is implemented in `src/meta_schedule/trace_apply.cc` and it is tested extensively in `test_meta_schedule_trace_apply.py`.

@junrushao @zxybazh @vinx13 @tkonolige


junrushao (approved):
This is just amazing!! Thanks for the PR!

Hzfengsy (approved):
LGTM. Thanks @masahi for the great work!

src/meta_schedule/trace_apply.cc
```diff
@@ -0,0 +1,234 @@
...
+  if (sch->Get(b1)->name_hint == sch->Get(b2)->name_hint) {
+    return true;
+  }
+  for (auto prod : sch->GetProducers(b2)) {
+    if (IsAncestor(b1, prod, sch)) return true;
+  }
+  return false;
+}
+
+void InlinePostBlocks(Schedule sch, Trace anchor_trace, Target target) {
```
zxybazh, line 47: Can you please add some function description here?

src/tir/schedule/primitive/cache_read_write.cc
```diff
@@ -297,7 +297,12 @@ bool CalculateAffineFlag(const ScheduleState& self, const StmtSRef& block_sref)
  * \param stage The stage to be inserted
  * \return A SeqStmt, the result after insertion
  */
-SeqStmt InsertCacheStage(const Stmt& stmt, int pos, const Stmt& stage) {
+Stmt InsertCacheStage(const Stmt& stmt, int pos, const Stmt& stage) {
```
zxybazh, line 300: Is this change related?

src/meta_schedule/trace_apply.cc
```diff
@@ -0,0 +1,234 @@
...
+  for (auto name : block_names_orig) {
+    if (block_names_now.count(name) && name != "root" && !is_scheduled(name)) {
+      unscheduled_blocks.push_back(sch->GetBlock(name));
+    }
+  }
+
+  return unscheduled_blocks;
+}
+
+void ScheduleUsingAnchorTrace(Schedule sch, const Trace& anchor_trace, const tvm::Target& target) {
```
zxybazh, line 196: Ditto

zxybazh (approved):
Apart from a few nits looks pretty good to me, very delicate design and comprehensive tests! Thanks @masahi!

src/meta_schedule/trace_apply.cc
masahi: This is documented in the header.

src/tir/schedule/primitive/cache_read_write.cc
masahi: It is not related but I found this bug while testing this branch. It happens on CPU when `link-params = True` is used and `cache_write` is applied during tuning.

masahi: Added a test case for this fix. For some reason I cannot reproduce this bug without anchor tuning. So I'm including it here.

PR #13207, opened 2022-10-26
[Hexagon] Fix port range bug and add random.seed

mehrdadh:
This PR fix a bug with port assignment in hexagon tests.

### DIFF
--- a/python/tvm/contrib/hexagon/pytest_plugin.py
+++ b/python/tvm/contrib/hexagon/pytest_plugin.py
@@ -84,10 +84,13 @@ PREVIOUS_PORT = None
 def get_free_port() -> int:
     """Return the next port that is available to listen on"""
     global PREVIOUS_PORT
+    random.seed(0)
     if PREVIOUS_PORT is None:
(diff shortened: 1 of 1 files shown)


python/tvm/contrib/hexagon/pytest_plugin.py
```diff
@@ -84,10 +84,13 @@ def android_serial_number() -> Optional[str]:
 def get_free_port() -> int:
     """Return the next port that is available to listen on"""
     global PREVIOUS_PORT
+    random.seed(0)
```
driazati, line 87: this should only run once otherwise it resets the rng state and you get the same value every time (not sure if that's okay here or not), maybe guard this behind a global flag `RNG_SEEDED` or something

```
>>> import random
>>> random.seed(0)
>>> random.random()
0.8444218515250481
>>> random.seed(0)
>>> random.random()
0.8444218515250481
```

driazati (approved):

mehrdadh:
@driazati very good point. I fixed it. PTAL

driazati (approved):

PR #13213, opened 2022-10-27
[Relay][transform][SimplifyExpr] simplify adjacent muls and adds with constants

yangulei:
This PR enables simplification and folding of a sub graph containing adjacent `mul`s and `add`s with constant inputs.
### Motivation
Workloads like [densenet-121](https://github.com/onnx/models/blob/main/vision/classification/densenet-121/model/densenet-7.onnx) has several partitions with `conv-bn-mul-add-relu` pattern, for example:
``` go
def @main(%data_0: Tensor[(1, 3, 224, 224), float32] /* ty=Tensor[(1, 3, 224, 224), float32] */) -> Tensor[(1, 1000, 1, 1), float32] {
  %0 = nn.conv2d(%data_0, meta[relay.Constant][0] /* ty=Tensor[(64, 3, 7, 7), float32] */, strides=[2, 2], padding=[3, 3, 3, 3], channels=64, kernel_size=[7, 7]) /* ty=Tensor[(1, 64, 112, 112), float32] */;
  %1 = nn.batch_norm(%0, meta[relay.Constant][1] /* ty=Tensor[(64), float32] */, meta[relay.Constant][2] /* ty=Tensor[(64), float32] */, meta[relay.Constant][3] /* ty=Tensor[(64), float32] */, meta[relay.Constant][4] /* ty=Tensor[(64), float32] */) /* ty=(Tensor[(1, 64, 112, 112), float32], Tensor[(64), float32], Tensor[(64), float32]) */;
  %2 = %1.0 /* ty=Tensor[(1, 64, 112, 112), float32] */;
  %3 = multiply(%2, meta[relay.Constant][5] /* ty=Tensor[(64, 1, 1), float32] */) /* ty=Tensor[(1, 64, 112, 112), float32] */;
  %4 = add(%3, meta[relay.Constant][6] /* ty=Tensor[(64, 1, 1), float32] */) /* ty=Tensor[(1, 64, 112, 112), float32] */;
  %5 = nn.relu(%4) /* ty=Tensor[(1, 64, 112, 112), float32] */;
  // ...
}
```
Current transforms on this pattern are:
1. `conv-bn-mul-add-relu` as the original pattern.
2. to `conv-mul-add-mul-add-relu` as `bn` is expended to `mul-add`. 
3. to `conv-add-mul-add-relu` as the first `mul` is folded into `conv`.

As all the `mul`s and `add`s have constant second inputs, they should be folded to a single `mul-add` and a preferred transform sequence should be:

1. `conv-bn-mul-add-relu` as the original pattern.
2. to `conv-mul-add-mul-add-relu` as `bn` is expended to `mul-add`. 
3. to `conv-mul-add-relu` as `mul`s and `add`s are folded to one single `mul-add`.
4. to `conv-add-relu` as `mul` is folded into `conv`.

### Solution
Actually, any series contain `mul`s and `add`s with constant inputs could be folded to one particular `mul-add`. Three rewrite rules are added to make this happen:

1. `mul-mul` -> `mul`
2. `add-add` -> `add`
3. `add-mul` -> `mul-add`

As `SimplifyExpr` apply simplifications iteratively until no changes to the graph, any `mul` and `add` series could be rewritten to one single `mul`, `add` or `mul-add` with one of the binary inputs could evaluates to a constant in the following `FoldConstant` pass.

### DIFF
--- a/src/relay/transforms/simplify_expr.cc
+++ b/src/relay/transforms/simplify_expr.cc
@@ -672,47 +672,122 @@ class EliminateIdentityRewrite : public DFPatternRewrite {
   DFPattern const_;
 };
 
-/*! \brief Make two consecutive add able to be constant_folded.
- * This pattern matching supports commutative property for addition.
+/*!
+ * \brief Returns whether \p expr is a ConstantNode or is a Call with
+ * \p IsConstantExpr args.
  */
-class SimplifyConsecutiveAdd : public DFPatternRewrite {
+bool IsConstantExpr(const Expr& expr) {
+  if (expr.as<ConstantNode>()) {
+    return true;
+  } else if (const CallNode* call = expr.as<CallNode>()) {
+    return std::all_of(call->args.begin(), call->args.end(), IsConstantExpr);
+  } else {
+    return false;
+  }
+}
+
+/*! \brief Switch adjacent add-mul with constants to mul-add.
+ * As mul-add pattern is more friendly to FoldScaleAxis.
+ */
+class SwitchAddMultiply : public DFPatternRewrite {
  public:
-  SimplifyConsecutiveAdd() {
-    x_ = IsWildcard();
-    const1_ = IsConstant();
-    const2_ = IsConstant();
-    DFPattern add_op = IsOp("add");
-    pattern_ = add_op({add_op({x_, const1_}), const2_});
+  SwitchAddMultiply() {
+    a_ = IsWildcard();
+    b_ = IsWildcard();
+    c_ = IsWildcard();
+    pattern_ = (a_ + b_) * c_;
+  }
+
+  Expr Callback(const Expr& pre, const Expr& post,
+                const Map<DFPattern, Array<Expr>>& node_map) const override {
+    auto a = node_map[a_][0];
+    auto b = node_map[b_][0];
+    auto c = node_map[c_][0];
+
+    bool is_a_const = IsConstantExpr(a);
+    bool is_b_const = IsConstantExpr(b);
+    bool is_c_const = IsConstantExpr(c);
+
+    if (!is_c_const) {
+      return post;
+    }
+    if (is_a_const && is_b_const) {
+      return post;
+    }
+
+    Expr x, c_add, c_mul;
+    c_mul = c;
+    if (is_a_const) {
+      x = b;
+      c_add = a;
+    } else if (is_b_const) {
+      x = a;
+      c_add = b;
+    } else {
+      return post;
+    }
+
+    auto bias = Call(Op::Get("multiply"), {c_add, c_mul});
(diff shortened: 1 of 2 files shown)


yangulei:
https://github.com/apache/tvm/blob/5c9066d816408bd2858c9758b0865ca08112c78f/tests/python/relay/test_pass_partition_graph.py#L1086-L1089
@masahi The tests above failed because the pattern matcher always matches the last two `add`s in `conv-add[bias]-add[bn]-add[sum]`, but the last `add` don't have any constant inputs thus cannot be simplified. Do you have any idea to fix this?

In this PR, I generalize the `add-add` or `mul-mul` simplification from **constant inputs** to **constant expr inputs**, so that complex pattern like `mul-add-mul` could be simplified to `mul-add`. In **constant inputs** case, the information of constant inputs could be embedded into the pattern for matching, thus the last two `add`s in `conv-add[bias]-add[bn]-add[sum]` will not match. While in **constant expr inputs** case, I match all `add-add` pattern first, then check whether both `add`s have constant inputs.

masahi:
This sounds tricky, since I'd expect the pattern matcher to be operating in a "bottom up" manner. Is it possible to add "constant-ness" condition in the pattern? Otherwise I can only think of adding conv2d into the `add -> add` pattern, but this is a terrible solution...

yangulei:
The pattern with `constant inputs` is fine for the first iteration in `SimplifyExpr` pass. While starting from the second iteration, some `constant inputs` are rewritten to `constant expr inputs`.

Take `((x + c1) + c2 )+ c3` as an example, first iteration could rewrite it to `(x + (c1 + c2)) + c3`. Now the first `add` becomes `x+(c1 + c2)`, with an **input which could evaluate to a constant** instead of a **constant input**, so the pattern with `constant inputs` will not match. I think this could be solved by evaluating `c1 + c2` to a constant like `c_12` at the end of the first iteration.

Is there a way to evaluate the `constant expr` to some concrete `constant` in the `SimplifyExpr` pass? If we can do that, the input graph for the later iterations also matches the `constant inputs` pattern and the simplification will carry on.

masahi:
Since we run `FoldConstant` before `SimplifyExpr` https://github.com/apache/tvm/blob/f7f2cda6756c170755fc18cbe23f6bf4a4b0d584/src/relay/backend/utils.cc#L264-L266, aren't constant expressions that already constants when we reach SimplifyExpr?

yangulei:
`SimplifyExpr` pass do the simplifications in an iterative way.
https://github.com/apache/tvm/blob/de7f76243601e127782e530f84ab7d51115a4c14/src/relay/transforms/simplify_expr.cc#L786-L790
The iterations happen inside `SimplifyExpr`, we need to find a way to evaluate the `const expr` manually.

yangulei:
@masahi Problem solved after applying `FoldConstant` inside `SimplifyExpr`.

masahi (approved):

PR #13216, opened 2022-10-27
[TVMC] Apply constant folding when converting layout

lhutton1:
This commit ensures that constant folding is applied when a desired layout is selected during compilation. It ensures that `layout_transform` operations are removed where possible so that pattern matching for BYOC backends can work effectively. A test has been added.

### DIFF
--- a/tests/python/driver/tvmc/test_transform.py
+++ b/tests/python/driver/tvmc/test_transform.py
@@ -1,0 +1,57 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+import pytest
+import numpy as np
+
+import tvm
+from tvm import relay
+from tvm.driver.tvmc.transform import convert_graph_layout
+
+
+def test_layout_transform():
+    """
+    Test layout is correctly transformed and constant folding is applied.
+    """
+    dtype = "int8"
+    iinfo = np.iinfo(dtype)
+    data_min = iinfo.min
+    data_max = iinfo.max
+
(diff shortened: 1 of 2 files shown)


lhutton1:
also cc @leandron @ekalda

tests/python/driver/tvmc/test_transform.py
```diff
@@ -0,0 +1,57 @@
...
+# under the License.
+import pytest
+import numpy as np
+
+import tvm
+from tvm import relay
+from tvm.driver.tvmc.transform import convert_graph_layout
+
+
+def test_layout_transform():
```
Mousius, line 25: This seems to be re-testing the existing passes? Instead we could just test that it's using them? I don't know if there's any off the shelf infrastructure to get a list of passes but we could do something like:

```
mock_sequence = MagicMock()
mock_pass_runner = MagicMock()
mock_sequence.return_value = mock_pass_runner
monkeypatch.setattr(transform, 'Sequential', mock_sequence);

convert_graph_layout(mod, "NHWC")

# Might need to mock out a bunch of passes here with some
mock_sequence.assert_called_once_with([
 relay.transform.RemoveUnusedFunctions(),
 relay.transform.ConvertLayout({
        "nn.conv2d": ["NHWC", "default"],
        "nn.conv2d_transpose": ["NHWC", "default"],
        "qnn.conv2d": ["NHWC", "default"],
 }),
 relay.transform.FoldConstant(),
])
mock_pass_runner.assert_called_once_with(mod);
```

ekalda: The test in the patch makes sense to me, it checks that the constant has been folded which wouldn't happen unless we have run the `FoldConstant` pass. I suppose mocking would work as well, but don't we then have the problem of having to keep the dict of ops and preferred layouts in the codebase manually in sync with the same dict in the test?

ekalda (commented):
LGTM! :)

tests/python/driver/tvmc/test_transform.py
lhutton1: I think the issue with the current test is that it checks more than it really needs to and might become a maintenance burden going forwards. Ideally I'd like to make use of pass instruments to check that `ConvertLayout` and `FoldConstant` are run in the correct order, which means that we aren't tied to the implementation details of the passes. However, the use of `PassContext` here overwrites any context set by the user so I think we should remove this and set the context from the callee(s) - given the test works for now, could we do this in a follow up?

ekalda: Yes, I agree, that testing the implementation details of the passes themselves in not ideal. I'd be fine with merging the current patch as it is and improving the test in a follow up, once we have cleared up the nested PassContexts if @Mousius agrees?

Mousius: Agree, this will fail if we disable constant folding so that'll work for now 😸

Mousius (approved):

PR #5314, opened 2020-04-12
[RUNTIME][IR] Allow non-nullable ObjectRef, introduce Optional<T>.

tqchen:
We use ObjectRef and their sub-classes extensively throughout our codebase.
Each of ObjectRef's sub-classes are nullable, which means they can hold nullptr
as their values.

While in some places we need nullptr as an alternative value. The implicit support
for nullptr in all ObjectRef creates additional burdens for the developer
to explicitly check defined in many places of the codebase.

Moreover, it is unclear from the API's intentional point of view whether
we want a nullable object or not-null version(many cases we want the later).

Borrowing existing wisdoms from languages like Rust. We propose to
introduce non-nullable ObjectRef, and Optional<T> container that
represents a nullable variant.

To keep backward compatiblity, we will start by allowing most ObjectRef to be nullable.
However, we should start to use Optional<T> as the type in places where
we know nullable is a requirement. Gradually, we will move most of the ObjectRef
to be non-nullable and use Optional<T> in the nullable cases.

Such explicitness in typing can help reduce the potential problems
in our codebase overall.

Changes in this PR:
- Introduce _type_is_nullable attribute to ObjectRef
- Introduce Optional<T>
- Change String to be non-nullable.
- Change the API of function->GetAttr to return Optional<T>

PR #3533, opened 2019-07-11
[INFA][IR] Build and Evolve Low-level IR. Remove HalideIR dep.

tqchen:
- Node system in include/tvm/node
- IR node sin include/tvm/ir.h
- Keep the IR nodes as consistent to the HalideIR as possible, will bring later PRs for refactoring so changes can be visible.

#3474

PR #13218, opened 2022-10-27
[TIR] Use Optional<Stmt> for IfThenElseNode::else_case

Lunderberg:
This parameter is nullable for cases where the else block isn't present.  Previously, it was represented as a `Stmt` holding `nullptr`, because `IfThenElse` (https://github.com/apache/tvm/pull/3533) predates the `Optional` utility (https://github.com/apache/tvm/pull/5314).  This commit updates to use `Optional<Stmt>` instead, and updates all usages of `else_case`.


Lunderberg:
A quick little breather before jumping back into the padded layout transforms.  Low priority.

junrushao (approved):
LGTM! Thanks for updating this infra!

PR #13219, opened 2022-10-27
[IR] Remove shadowing in IRSubstituteWithDataTypeLegalization

Lunderberg:
Previously, the `IRSubstituteWithDataTypeLegalization` class implemented some virtual functions of `DataTypeLegalizer`, but not all.  As a result, some compilers gave warnings that the base class methods were being shadowed.  This commit adds the `using` declaration to avoid shadowing.


jwfromm:
Seems like a solid, change, thanks @Lunderberg!

jwfromm (approved):

Issue #13185, opened 2022-10-24
[ci] MacOS CI failing in RPC tests


driazati:
* https://github.com/apache/tvm/actions/runs/3314615063/jobs/5474129728

<details>
```
============================= test session starts ==============================
platform darwin -- Python 3.7.13, pytest-7.1.2, pluggy-1.0.0 -- /usr/local/miniconda/envs/tvm-build/bin/python
cachedir: .pytest_cache
rootdir: /Users/runner/work/tvm/tvm
collecting ... collected 12 items

tests/python/contrib/test_rpc_server_device.py::test_basic_functionality_of_rpc_session FAILED [  8%]
tests/python/contrib/test_rpc_server_device.py::test_can_call_remote_function_with_rpc_standalone PASSED [ 16%]
tests/python/contrib/test_rpc_server_device.py::test_cleanup_workspace_after_session_end XFAIL [ 25%]
tests/python/contrib/test_rpc_server_device.py::test_graph_executor_remote_run FAILED [ 33%]
tests/python/contrib/test_rpc_server_device.py::test_rpc_standalone PASSED [ 41%]
tests/python/contrib/test_rpc_server_device.py::test_can_call_remote_function_with_rpc_proxy PASSED [ 50%]
tests/python/contrib/test_rpc_server_device.py::test_rpc_proxy PASSED    [ 58%]
tests/python/contrib/test_rpc_server_device.py::test_can_call_remote_function_with_rpc_tracker PASSED [ 66%]
tests/python/contrib/test_rpc_server_device.py::test_check_auto_schedule_tuning XFAIL [ 75%]
tests/python/contrib/test_rpc_server_device.py::test_rpc_tracker PASSED  [ 83%]
tests/python/contrib/test_rpc_server_device.py::test_can_call_remote_function_with_rpc_tracker_via_proxy PASSED [ 91%]
tests/python/contrib/test_rpc_server_device.py::test_rpc_tracker_via_proxy PASSED [100%]

=================================== FAILURES ===================================
___________________ test_basic_functionality_of_rpc_session ____________________

    def wrapper():
        with server_ios_launcher.ServerIOSContextManager(
            mode=server_ios_launcher.RPCServerMode.standalone.value,
            host=HOST_URL,
            port=HOST_PORT,
            key=DEVICE_KEY,
        ) as ios_server:
>           f(host=ios_server.host, port=ios_server.port)

tests/python/contrib/test_rpc_server_device.py:81: 
[...]

PR #13229, opened 2022-10-28
[CI] Enable iOS RPC tests

KJlaccHoeUM9l:
These tests were previously disabled (see https://github.com/apache/tvm/issues/13185). The error was related to updating `MacOS` to version `12.6` and `XCode` update.
This PR fixes the error and re-enables tests for `iOS RPC`.


driazati (approved):
Thanks for the fix!

PR #13230, opened 2022-10-28
[Hexagon] Update search pattern to find .so address for on device runs

jverma-quic:
Additional changes to hexagon specific lightweight profiling. 

The PR updates the search pattern for .so address since the old pattern wasn't available for everyone. In addition, it also allows for the handling of tvm.build modules.


jverma-quic:
cc: @joshherr-quic

kparzysz-quic (approved):
LGTM

PR #13141, opened 2022-10-19
[skip ci] Add Janet and Thomas to triagers to help with Issue Triage RFC

areusch:
these two volunteered to help triage issues per [Issue Triage RFC](https://github.com/apache/tvm-rfcs/blob/main/rfcs/0093_Issue_Triage.md).

@tqchen @driazati

PR #13231, opened 2022-10-28
[skip ci] Remove naut-thomas from .asf.yaml

driazati:
We have to remove and re-add this user to re-trigger the asf invitation

Follow on to #13141


Lunderberg (approved):
LGTM!

PR #13243, opened 2022-10-31
[MetaSchedule] Fix thread bindings of MultiLevelTilingTensorCore

vinx13:
Fixed https://github.com/apache/tvm/issues/13204.

cc @junrushao

### DIFF
--- a/src/meta_schedule/schedule_rule/multi_level_tiling_tensor_core.cc
+++ b/src/meta_schedule/schedule_rule/multi_level_tiling_tensor_core.cc
@@ -556,6 +556,11 @@ ScheduleRule ScheduleRule::MultiLevelTilingTensorCore(
     Optional<Integer> max_innermost_factor, Optional<Array<Integer>> vector_load_lens,
     Optional<Map<String, ObjectRef>> reuse_read, Optional<Map<String, ObjectRef>> reuse_write,
     bool use_software_pipeline) {
+  if (tile_binds.defined()) {
+    for (const String& tile_bind : tile_binds.value()) {
+      CHECK_NE(tile_bind, "threadIdx.x") << "Cannot bind to threadIdx.x when using tensor core.";
+    }
+  }
   auto node = MultiLevelTilingInitCommon<MultiLevelTilingTensorCoreNode>(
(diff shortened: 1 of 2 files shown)


michalpiszczek (approved):

shingjan (approved):
LGTM. Thanks!

junrushao (approved):
LGTM!

junrushao:
BTW, i was curious why it's not caught by the unittests of search space generation?

vinx13:
the unit test doesn't use default config here

junrushao:
we might want to add a unittest to cover the rule, but i understand it's kind of urgent. please set a reminder to add a regression test :-)

masahi:
@tvm-bot rerun

src/meta_schedule/schedule_rule/multi_level_tiling_tensor_core.cc
```diff
@@ -556,6 +556,11 @@ ScheduleRule ScheduleRule::MultiLevelTilingTensorCore(
     Optional<Integer> max_innermost_factor, Optional<Array<Integer>> vector_load_lens,
     Optional<Map<String, ObjectRef>> reuse_read, Optional<Map<String, ObjectRef>> reuse_write,
     bool use_software_pipeline) {
+  if (tile_binds.defined()) {
+    for (const String& tile_bind : tile_binds.value()) {
+      CHECK_NE(tile_bind, "threadIdx.x") << "Cannot bind to threadIdx.x when using tensor core.";
+    }
+  }
```
csullivan, line 563: Does this check (along with existing tests) now cover the regression this fixes? If not adding a test to protect against the regression in the future will surely save you and others time!

csullivan: Oops looks like @junrushao already asked my question, I trust the your judgements :)

vinx13: Yes this check covered the broken case that incorrectly using `threadIdx.x`. As a follow up, I'm thinking also adding some e2e search space generation test for the default config for each target.

PR #13244, opened 2022-10-31
[Hexagon] Make pytest use a random port if not running in CI

nverke:
This makes it so that when running tests locally you are able to get a random port. This is needed because the ports take a bit of time to free before they can be used again so repeated tests started to fail. 

cc @masahi @mehrdadh


nverke:
@tvm-bot rerun

masahi:
@tvm-bot rerun

masahi (approved):

PR #13247, opened 2022-11-01
[Frontend][PaddlePaddle] Fix UnboundLocalError: local variable 'shape…

woobinw:
There are a local variable referenced before assignment in convert_interpolate function. I think varible 'size' is real want to be referenced.

@jiangjiajun 
@heliqi 
cc @junrushao

### DIFF
--- a/python/tvm/relay/frontend/paddlepaddle.py
+++ b/python/tvm/relay/frontend/paddlepaddle.py
@@ -778,7 +778,7 @@ def convert_interpolate(g, op, block):
         for name in input_size_tensor:
             size = g.get_node(name)
             if len(infer_shape(size)) == 0:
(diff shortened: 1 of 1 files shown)


heliqi:
@woobinw   Can you add a single test for that?

junrushao (approved):
LGTM!

woobinw:
> @woobinw Can you add a single test case for that?
> 
> https://github.com/apache/tvm/blob/2c1fecd0978d013a552ddb8d5992942acb40e5bc/tests/python/frontend/paddlepaddle/test_forward.py#L911

@heliqi 
Ok！The test case have submited #13277

heliqi:
good job!

PR #13249, opened 2022-11-01
[Docs][Rust] Add link to information on stable Rust release

bkmgit:
Releases are frequent, so documenting the current release number is not helpful.

### DIFF
--- a/rust/tvm/README.md
+++ b/rust/tvm/README.md
@@ -18,7 +18,8 @@
 # TVM
 
 This crate provides an idiomatic Rust API for [Apache TVM](https://github.com/apache/tvm).
-The code works on **Stable Rust** and is tested against `rustc 1.47`.
+The code works on **Stable Rust** and is tested against the current
+[Stable Rust release](https://releases.rs/). The MSRV is 1.60.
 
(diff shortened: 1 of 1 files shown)


bkmgit:
@jroesch Review would be appreciated.

ehsanmok:
Thanks for your contribution! Unfortunately that's not sufficient. If want to change to stable, we need to provide the exact Minimum Supported Rust Version (MSRV). You can use [cargo-msrv](https://github.com/foresterre/cargo-msrv) for example, to compile and find the version and include it in the PR.

bkmgit:
Thanks for your feedback. CI tests just stable https://github.com/apache/tvm/blob/main/docker/install/ubuntu_install_rust.sh . If MSVR is also required, maybe a CI test should also be added for this? It would probably also be helpful to automate regeneration of the README from a template whenever MSVR changes.

ehsanmok:
MSRV rarely changes and there's no need for a template.

bkmgit:
Ok. Got an MSRV of 1.60

bkmgit:
@ehsanmok Let me know if anything else is needed. If not, can rebase and squash commits.

PR #13253, opened 2022-11-01
[Adreno][Textures] Fix static memory planner 

echuraev:
- Fix memory reusage in static memory planner.
- Move token allocators to separate file
- Add test on TokenAllocator2d

cc: @elvin-n, @csullivan


src/relay/backend/graph_plan_memory.cc
```diff
@@ -513,30 +515,51 @@ class StorageAllocator : public StorageAllocaBaseVisitor {
...
+            (min_added_size_y > 0 && added_size_y < min_added_size_y) ||
+            (min_added_size_x == added_size_x && wasted_size_x < min_wasted_size_x) ||
+            (min_added_size_y == added_size_y && wasted_size_y < min_wasted_size_y)) {
+          min_added_size_x = added_size_x;
+          min_added_size_y = added_size_y;
+          min_wasted_size_x = wasted_size_x;
+          min_wasted_size_y = wasted_size_y;
           best_storage_id = free_id;
-          best_mem = new_mem;
+          best_mem = cached;
```
elvin-n: previously best_mem corresponded for the memory found in the cache to be reused, while new_mem stand for local usage only inside this loop.

By fact you substitutes new_mem by new_width and new_height local variables and started to have two variables after the loop: best_mem and new_mem which should describe the best memory, There should be only one.

You can remove best_mem and use best_storage_id to address the data from the cache and new_mem as variable storing new values of width and height.

src/relay/backend/graph_plan_memory.cc
```diff
@@ -513,30 +515,51 @@ class StorageAllocator : public StorageAllocaBaseVisitor {
...
+      if (min_added_size_x == 0 && min_added_size_y == 0) {
+        // use existing block
         free_list_.erase(best_storage_id);
+        best_mem.token_->ref_counter += prototype->ref_counter;
         return best_mem.token_;
+      } else if (min_added_size_x <= shape.width || min_added_size_y <= shape.height) {
+        // Reset the reference counter of the now live token
+        free_list_.erase(best_storage_id);
+        new_mem.token_ = prototype;
+        new_mem.token_->ref_counter += 1;
```
elvin-n: I have not observed you migrated ref_counter from previous block `blocks_[best_storage_id].token_->ref_counter` while you are deleting it by overriden of `blocks_[best_storage_id]` below

elvin-n (changes requested):

src/relay/backend/graph_plan_memory.cc
echuraev: If the memory block is in the `free_list_` that means that the reference counter of `blocks_[best_storage_id]` is equal to zero. This is why it is not really matter if I override the previous value or not.

src/relay/backend/graph_plan_memory.cc
echuraev: Done

elvin-n (approved):

echuraev:
@masahi Could you please take a look at this PR?

masahi (approved):

PR #13256, opened 2022-11-01
[Hexagon] Add HVX quant conv2d implementation

quic-sanirudh:
This patch adds a new HVX intrinsic implementation to perform quantized convolution.

It assumes that the qnn.conv2d relay op is not
canonicalized and all the quantization parameters (scales and zero points) are passed into the intrinsic implementation.

It also uses the fixed point computation function defined in hexagon topi utils to compute a fixed point (combined) scale which is used to perform the final requantization before returning the quantized output.

### DIFF
--- a/cmake/modules/Hexagon.cmake
+++ b/cmake/modules/Hexagon.cmake
@@ -178,6 +178,15 @@ if(BUILD_FOR_HEXAGON)
     "${TVMRT_SOURCE_DIR}/hexagon/ops/*.cc"
   )
 
+  include_directories(
+    "${TVMRT_SOURCE_DIR}/hexagon/ops"
+  )
+
+  set_source_files_properties(
+    "${TVMRT_SOURCE_DIR}/hexagon/ops/conv2d_quant_hvx.cc"
+    PROPERTIES COMPILE_FLAGS "-mhvx"
+  )
+
   set_source_files_properties(
     "${TVMRT_SOURCE_DIR}/hexagon/ops/conv2d_fp16_hvx.cc"
     PROPERTIES COMPILE_FLAGS "-mhvx"
--- a/src/runtime/hexagon/ops/conv2d.h
+++ b/src/runtime/hexagon/ops/conv2d.h
@@ -20,6 +20,7 @@
 #include <tvm/runtime/c_runtime_api.h>
 #include <tvm/runtime/device_api.h>
 
+#include <algorithm>
 #include <cassert>
 
 #ifndef TVM_RUNTIME_HEXAGON_OPS_CONV2D_H_
@@ -28,6 +29,7 @@
 namespace tvm {
 namespace runtime {
 namespace hexagon {
+namespace conv_utils {
 static constexpr auto hexagon_device = DLDevice{static_cast<DLDeviceType>(kDLHexagon), 0};
 
 // Standalone DLTensor: the standalone-ness means that this object owns the shape
@@ -75,7 +77,7 @@ inline void* to_ptr(uintptr_t v) { return reinterpret_cast<void*>(v); }
 
 inline uintptr_t to_uint(void* ptr) { return reinterpret_cast<uintptr_t>(ptr); }
 
-constexpr int xyc_to_sm_16b(int y, int x, int c) {
+inline constexpr int yxc_to_sm_16b(int y, int x, int c) {
   // Map y,x,c coordinates within a block to the offset (in 16-bit elements)
   // from the beginning of the block in spatial-major layout.
   // 10-bit spatial mask: yyyxcccccx
@@ -83,7 +85,23 @@ constexpr int xyc_to_sm_16b(int y, int x, int c) {
   return y << 7 | (x & 2) << 5 | c << 1 | (x & 1);
 }
 
-constexpr int hwio_to_sm_16b(int width, int y, int x, int i, int o) {
+inline constexpr int yxc_to_sm_8b(int y, int x, int c) {
+  // Map y,x,c coordinates within a block to the offset (in 8-bit elements)
+  // from the beginning of the block in spatial-major layout.
+  // 10-bit spatial mask: yyyxxxccccc
+  return y << 8 | x << 5 | c;
+}
+
+inline constexpr int hwio_to_sm_8b(int width, int y, int x, int i, int o) {
+  // Map y,x,i,o coordinates within a chunk (assuming the origin at the
+  // top-left spatial corner) to the offset (in 8-bit elements) from the
+  // beginning of the chunk in spatial-major layout.
+  // Spatial mask: p..piiioooooii, where p..p are position bits.
+  int p = y * width + (width - 1 - x);
+  return p << 10 | (i & 0x1c) << 5 | o << 2 | (i & 3);
+}
+
+inline constexpr int hwio_to_sm_16b(int width, int y, int x, int i, int o) {
   // Map y,x,i,o coordinates within a chunk (assuming the origin at the
   // top-left spatial corner) to the offset (in 16-bit elements) from the
   // beginning of the chunk in spatial-major layout.
@@ -123,6 +141,10 @@ inline uintptr_t hwio_at(const DLTensor& f, int y, int x, int i, int o) {
  * The input is mapped into the below mentioned layout (notation similar to index map used for
  * transform layout):
  *
+ * For uint8_t type
+ * lambda n, h, w, c: n, h//8, w//8, c//32, AXIS_SEPARATOR, h%8, w%8, c%32
+ *
+ * For uint16_t type
  * lambda n, h, w, c: n, h//8, w//4, c//32, AXIS_SEPARATOR, h%8, (w%4)//2, c%32, w%2
  *
  * where AXIS_SEPARATOR represents split up in the physical layout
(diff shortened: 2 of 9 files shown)


quic-sanirudh:
@csullivan @cconvey Gentle ping for a review. This is an initial quant implementation of conv2d, similar to the fp16 version that I wrote earlier.

tests/python/contrib/test_hexagon/topi/test_conv2d_quant_intrin.py
```diff
@@ -0,0 +1,262 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
mehrdadh, line 1: This file will have lint issues once https://github.com/apache/tvm/pull/13271 merged. Please fix them in the meantime

mehrdadh (changes requested):

quic-sanirudh:
@tvm-bot rerun

tests/python/contrib/test_hexagon/topi/test_conv2d_quant_intrin.py
quic-sanirudh: I've fixed the line issue, thanks.

quic-sanirudh:
@driazati @mehrdadh The CI is stuck. Could you please let me know if there's any way to restart the CI. Would pining the tvm-bot work?

driazati:
> @driazati @mehrdadh The CI is stuck. Could you please let me know if there's any way to restart the CI. Would pining the tvm-bot work?

CI is finished now, so it should be good. I've been seeing some similar queueing issues (probably Jenkins wasn't able to spin up a machine to run the jobs for some reason), I opened #13312 and will be looking into this today

driazati:
> @driazati @mehrdadh The CI is stuck. Could you please let me know if there's any way to restart the CI. Would pining the tvm-bot work?

CI is finished now, so it should be good. I've been seeing some similar queueing issues (probably Jenkins wasn't able to spin up a machine to run the jobs for some reason), I opened #13312 and will be looking into this today

quic-sanirudh:
> > @driazati @mehrdadh The CI is stuck. Could you please let me know if there's any way to restart the CI. Would pining the tvm-bot work?
> 
> CI is finished now, so it should be good. I've been seeing some similar queueing issues (probably Jenkins wasn't able to spin up a machine to run the jobs for some reason), I opened #13312 and will be looking into this today

Great, thanks a lot for the help.

tests/python/contrib/test_hexagon/topi/test_conv2d_quant_intrin.py
```diff
@@ -0,0 +1,261 @@
...
+        wgt_q = wgt_q.reshape(wgt_f.shape)
+
+        # Generate reference output
+        ref_out = conv2d_nhwc_python(act_f, wgt_f, stride=inp_stride, padding="VALID")
+
+        ref_out_q, out_scale, out_zp = quantize_np(ref_out, dtype="uint8")
+        ref_out_q = ref_out_q.reshape(ref_out.shape)
+
+        final_scale = act_scale * wgt_scale / out_scale
+        fixed_final_scale, scale_factor = get_fixed_point_value(final_scale)
```
ibsidorenko, line 228: Hi, @quic-sanirudh ! Thank you for this PR, rather interesting work. Just one small question:
As I see for Hexagon target we use int16 dtype to represent  fixed point values (`dtype `param in `get_fixed_point_value`). But in TVM we use int32 dtype for that (for example for scale parameter in requantize). Can this somehow affect the accuracy of the real life quantized models?

quic-sanirudh:
@csullivan @cconvey When you get a chance, could you review this PR. This is a quantized conv2d similar to the fp16 conv2d that I wrote earlier.

mehrdadh (commented):
LGTM! I'll wait for @cconvey and @csullivan to take a look

quic-sanirudh:
@csullivan @cconvey Could you please help in reviewing this patch when you get a chance, thanks.

quic-sanirudh:
@csullivan @cconvey Could you please review this PR or suggest someone who could be the right person to review it. Thanks.

src/runtime/hexagon/ops/conv2d_quant_hvx.cc
```diff
@@ -0,0 +1,317 @@
...
+                                   int out_code, void* res_handle);
+
+namespace tvm {
+namespace runtime {
+namespace hexagon {
+inline uint8_t* getElementPtr_int8(int block_out_y, int block_out_x, int block_out_c, int yi,
+                                   int xi, int ci, const DLTensor& block) {
+  auto block_ptr =
+      tvm::runtime::hexagon::conv_utils::nhwc_at(block, 0, block_out_y, block_out_x, block_out_c);
+  auto block_offset = yi * 256 + xi * 32 + ci;
```
janetsc, line 38: Suggest defining consts for these.  Are they derived from the supported shape?

janetsc: The same comment applies to all constants below as well.

janetsc (approved):
I meant to say "Approve" earlier today - my suggestions are only cosmetic.  (Although the bounds checking would be a good idea if you do another iteration.)

csullivan (approved):

src/runtime/hexagon/ops/conv2d.h
```diff
@@ -75,15 +77,31 @@ inline void* to_ptr(uintptr_t v) { return reinterpret_cast<void*>(v); }
...
   // 10-bit spatial mask: yyyxcccccx
   assert(y >= 0 && x >= 0 && c >= 0);
   return y << 7 | (x & 2) << 5 | c << 1 | (x & 1);
 }
 
-constexpr int hwio_to_sm_16b(int width, int y, int x, int i, int o) {
+inline constexpr int yxc_to_sm_8b(int y, int x, int c) {
+  // Map y,x,c coordinates within a block to the offset (in 8-bit elements)
+  // from the beginning of the block in spatial-major layout.
+  // 10-bit spatial mask: yyyxxxccccc
```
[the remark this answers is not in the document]
quic-sanirudh: I consciously avoided the checks here because these functions are used for indexing within the innermost loops and need to be really fast. I actually was planning to remove the check from the above `yxc_to_sm_16b` function as well.

I thought I had was to add the `assert` statements and then disable them for release builds with `#define NDEBUG`. Not sure if there's a better solution.

src/runtime/hexagon/ops/conv2d.h
```diff
@@ -75,15 +77,31 @@ inline void* to_ptr(uintptr_t v) { return reinterpret_cast<void*>(v); }
...
+  return y << 8 | x << 5 | c;
+}
+
+inline constexpr int hwio_to_sm_8b(int width, int y, int x, int i, int o) {
+  // Map y,x,i,o coordinates within a chunk (assuming the origin at the
+  // top-left spatial corner) to the offset (in 8-bit elements) from the
+  // beginning of the chunk in spatial-major layout.
+  // Spatial mask: p..piiioooooii, where p..p are position bits.
+  int p = y * width + (width - 1 - x);
+  return p << 10 | (i & 0x1c) << 5 | o << 2 | (i & 3);
```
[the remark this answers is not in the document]
quic-sanirudh: Same comment as above. I can probably add asserts if we can disable them for release builds later

src/runtime/hexagon/ops/conv2d_quant_hvx.cc
quic-sanirudh: I'll do that @janetsc thanks. Right now they're assuming the activation blocks to be `8x8x32`.

cmake/modules/Hexagon.cmake
```diff
@@ -178,6 +178,15 @@ if(BUILD_FOR_HEXAGON)
     "${TVMRT_SOURCE_DIR}/hexagon/ops/*.cc"
   )
 
+  include_directories(
+    "${TVMRT_SOURCE_DIR}/hexagon/ops"
+  )
+
+  set_source_files_properties(
+    "${TVMRT_SOURCE_DIR}/hexagon/ops/conv2d_quant_hvx.cc"
+    PROPERTIES COMPILE_FLAGS "-mhvx"
```
cconvey, line 187: Are we confident that `-mhvx` is supported by all of the compilers that might build this code?

I'm assuming that _typically_ the clang provided by Hexagon Toolchain will be used.  But I'm a little fuzzy about the intended level of support for other compilers, e.g. a user-supplied build of Clang/LLVM.

cconvey: Would it make sense to update [src/runtime/hexagon/README.md](https://github.com/apache/tvm/blob/15ee9bb5757915c73569f3ebdb5e52a4312663aa/src/runtime/hexagon/README.md) to clarify the version(s) of LLVM that support flags like `-mhvx`?

Or alternatively, use CMake's [CheckCXXCompilerFlag](https://cmake.org/cmake/help/latest/module/CheckCXXCompilerFlag.html) function to see if `-mhvx` is supported, and only use that flag if it is?

src/runtime/hexagon/ops/conv2d.h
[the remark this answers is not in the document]
cconvey: I'm pretty sure we can rely on `assert` being disabled when `CMAKE_BUILD_TYPE=Release`.  See https://stackoverflow.com/questions/34302265/does-cmake-build-type-release-imply-dndebug.

src/runtime/hexagon/ops/conv2d.h
```diff
@@ -133,7 +155,48 @@ inline uintptr_t hwio_at(const DLTensor& f, int y, int x, int i, int o) {
  * @param width
  * @param depth
  */
-void blockize_hwc_16b(void* out, void* inp_flat, int height, int width, int depth);
+template <typename T, int block_height, int block_width, int block_depth>
+void blockize_hwc(void* out, void* inp_flat, int height, int width, int depth) {
```
cconvey, line 159: Would it make sense for `inp_flat`'s type to be `const void*` rather than `void*`?

This is probably a bit of a stylistic choice; I just figured I'd ask.

src/runtime/hexagon/ops/conv2d.h
```diff
@@ -144,7 +207,42 @@ void blockize_hwc_16b(void* out, void* inp_flat, int height, int width, int dept
  * @param width
  * @param depth
  */
-void deblockize_hwc_16b(void* out_flat, void* inp, int height, int width, int depth);
+template <typename T, int block_height, int block_width, int block_depth>
+void deblockize_hwc(void* out_flat, void* inp, int height, int width, int depth) {
```
cconvey, line 211: Would it make sense for the type of `inp` to be `const void*`?

cconvey (changes requested):
Just a few minor suggestions.

cmake/modules/Hexagon.cmake
quic-sanirudh: Thanks for the review @cconvey.

I can add the details in the README or add a CMake check, but the `-mhvx` flag was added to clang all the way back in 2017 in [LLVM 6.0 release](https://releases.llvm.org/6.0.0/docs/ReleaseNotes.html#changes-to-the-hexagon-target) if not earlier, which predates the entire TVM project, so we can also probably assume safely that the `-mhvx` flag will be available for practically anyone building the TVM project now.

If you think it might still be better to add the check or the README change, please let me know which one you think makes more sense and I can make that change. Thanks

src/runtime/hexagon/ops/conv2d.h
quic-sanirudh: I agree with both, I'll add the asserts and the `const void*` for the arguments, thanks.

src/runtime/hexagon/ops/conv2d.h
quic-sanirudh: I'll add the `const void*`, it makes sense, thanks.

cconvey (changes requested):
Added a question / comment about use of "inline".

cmake/modules/Hexagon.cmake
cconvey: That makes total sense, I didn't realize `-mhvx` support went back that far.  I agree that there's no need for any additional documentation or checking.

src/runtime/hexagon/ops/conv2d.h
```diff
@@ -75,15 +77,31 @@ inline void* to_ptr(uintptr_t v) { return reinterpret_cast<void*>(v); }
 
 inline uintptr_t to_uint(void* ptr) { return reinterpret_cast<uintptr_t>(ptr); }
 
-constexpr int xyc_to_sm_16b(int y, int x, int c) {
+inline constexpr int yxc_to_sm_16b(int y, int x, int c) {
```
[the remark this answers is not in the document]
quic-sanirudh: ah okay, I did not realize this change was made. Looks like an addition that was inserted by working with the downstream repo, where this `inline` probably exists. I'll remove it, it makes sense, thanks.

src/runtime/hexagon/ops/conv2d.h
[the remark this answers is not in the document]
janetsc: Another option is to check the loop bounds to make sure y, x and c can't get bigger than can be expressed.

src/runtime/hexagon/ops/conv2d.h
```diff
@@ -133,7 +155,48 @@ inline uintptr_t hwio_at(const DLTensor& f, int y, int x, int i, int o) {
...
+  const int stride_x = depth;
+  const int stride_y = stride_x * width;
+
+  for (int cy = 0; cy < height; cy += block_height) {
+    for (int cx = 0; cx < width; cx += block_width) {
+      for (int cc = 0; cc < depth; cc += block_depth) {
+        auto block = reinterpret_cast<T*>(*out_data++);
+        int max_y = std::min(block_height, height - cy);
+        int max_x = std::min(block_width, width - cx);
+        int max_c = std::min(block_depth, depth - cc);
```
janetsc, line 179: This is an example where you could check to see that max_y, max_x and max_c are in bounds for index_func.

janetsc (approved):

cconvey (approved):
LGTM!

mehrdadh (approved):

PR #13267, opened 2022-11-02
[build][tir] suppress -Woverloaded-virtual warning

cconvey:
Suppress a spurious `-Woverloaded-virtual` clang warning for the `tvm::tir::DataTypeRewriter` class.

I think we're safe to ignore clang _version_ for this pragma, because it's been supported for a long time now.

### DIFF
--- a/src/tir/transforms/narrow_datatype.cc
+++ b/src/tir/transforms/narrow_datatype.cc
@@ -206,6 +206,15 @@ class DataTypeRewriter : public DataTypeLegalizer {
     return VisitStmt(s);
   }
 
+ protected:
+  // This class adds some overrides of `VisitStmt_` and `VisitExpr_` that
+  // are *not* present in the parent class.
+  // These `using` statements ensure that all of the *other* overrides
+  // provided by the parent class are fully visible to users of this class.
+  // (Discussed further in https://github.com/apache/tvm/pull/13267)
+  using Parent::VisitExpr_;
+  using Parent::VisitStmt_;
+
   Stmt VisitStmt_(const StoreNode* op) final {
     LOG(FATAL) << "Unexpected use of deprecated StoreNode.  Please use BufferStoreNode instead.";
     return Stmt();
--- a/include/tvm/tir/stmt_functor.h
+++ b/include/tvm/tir/stmt_functor.h
@@ -502,7 +502,7 @@ bool ContainsNode(const Stmt& stmt) {
  * base class of such passes to ensure the consistency of data types.
  */
 class DataTypeLegalizer : public StmtExprMutator {
- public:
+ protected:
   Stmt VisitStmt_(const ForNode* op) override;
 
   Stmt VisitStmt_(const AttrStmtNode* op) override;
@@ -530,7 +530,6 @@ class DataTypeLegalizer : public StmtExprMutator {
   using StmtExprMutator::VisitExpr_;
   using StmtExprMutator::VisitStmt_;
 
- protected:
   // a map from IterVar before rewrite to that after rewrite,
   // ensures one old IterVar maps to exactly one new IterVar
   std::unordered_map<const IterVarNode*, IterVar> ivmap_;


cconvey:
CC: @Lunderberg

src/tir/transforms/narrow_datatype.cc
```diff
@@ -187,6 +187,11 @@ class DataTypeVisitor final : public StmtExprVisitor {
   arith::ConstIntBoundAnalyzer::BoundMapType bound_;
 };
 
+#if __clang__
```
Lunderberg, line 190: Instead of suppressing the warning altogether, could we instead add a `using Parent::VisitStmt_;` and `using Parent::VisitExpr_`?  That would allow the parent's virtual functions to participate in overload resolution, avoiding the case that the warning is trying to catch.

cconvey: Good point.  Maybe it should hinge on what the author (@vinx13) had in mind here w.r.t. the base-class overloads: hide them, or don't care?

@vinx13 : Can you clarify?

vinx13: It only overrides some methods, for the rest it is intended to go to its parent class. I already add https://github.com/apache/tvm/blob/main/include/tvm/tir/stmt_functor.h#L530-L531 to its parent class, not sure why the warning still occurs.

cconvey: I don't think adding more overloads to the parent class changes the issue.  Here's how I understand it (but I may be missing something...)

Suppose we have code like this:
``` c++
#include <iostream>
using namespace std;

class BaseClass {
   public:
    virtual void foo(int x) {
        cout << "BaseClass::foo(int): x = " << x << endl;
    }   
};  

class DerivedClass : public BaseClass {
   public:
    virtual void foo(float x) {
        cout << "DerivedClass::foo(float): x = " << x << endl;
    }   
};  

int main() {
    DerivedClass d;
    d.foo(42);
}   
```

Many people would expect this to print `BaseClass::foo(int): x = 42` because `42` is an int.

But what _actually_ gets printed is `DerivedClass::foo(float): x = 42`.  This is unintuitive for a lot of C++ programmers, which I think is why this warning was created.  

IIUC, it adding the `using ...` statements that @Lunderberg proposed would make this code behave more like you intended originally.

vinx13: I expect it to only override virtual methods already available in `DataTypeLegalizer` or `StmtExprMutator`, instead of adding new overloads, maybe I'm missing something

cconvey: Sorry, let me try making my point better with a slightly different example:
``` c++
#include <iostream>
#include <string>
using namespace std;

class BaseClass {
   public:
    virtual void foo(int x) {
        cout << "BaseClass::foo(int): x = " << x << endl;
    }

    virtual void foo(string x) {
        cout << "BaseClass::foo(string): x = " << x << endl;
    }
};

class DerivedClass : public BaseClass {
   public:
#ifdef OVERLOAD_FOO
    virtual void foo(int x) {
         cout << "DerivedClass::foo(int): x = " << x << endl;
    }
#endif
};

int main() {
    DerivedClass d;
    d.foo("Hello, TVM!");
}
```

If I compile this with simply `clang++ -Woverloaded-virtual x.cpp`, the compilation succeeds.  When I run the resulting executable, it prints `BaseClass::foo(string): x = Hello, TVM!`.

If I try to compile this with `clang++ -Woverloaded-virtual -DOVERLOAD_FOO x.cpp`, I get this error:
```
x.cpp:27:11: error: cannot initialize a parameter of type 'int' with an lvalue of type 'const char[12]'
    d.foo("Hello, TVM!");
          ^~~~~~~~~~~~~
x.cpp:19:26: note: passing argument to parameter 'x' here
    virtual void foo(int x) {
                         ^
1 error generated.
```
 The issue, as I understand it, is that because `DerviedClass` provides _even one_ override of a base-class virtual function (`foo`), C++ will no longer look at _any_ of the `BaseClass::foo(...)` overloads when it sees a call such as `d.foo("Hello, TVM!")` in the example above.

Those base-class functions aren't _completely_ hidden.  For example, if we use `d.BaseClass("Hello, TVM!")` instead, the code compiles just fine.

vinx13: Thanks for the explanation. I think it's common in the codebase to only override visitors for some IR nodes, for example https://github.com/apache/tvm/blob/main/src/tir/transforms/storage_rewrite.cc#L63. Any idea how it works?

cconvey: Sorry it took several examples to make my point :)

> I think it's common in the codebase to only override visitors for some IR nodes, for example https://github.com/apache/tvm/blob/main/src/tir/transforms/storage_rewrite.cc#L63. Any idea how it works?

I think we'd need to look more closely at code that _calls_ one of those methods to see what's going on.  Two possibilities that come to mind are:

(a) We're doing something equivalent to this:
``` c++
StmtExprVisitor * p = new LinearAccessPatternFinder(...);
p->VisitStmt_(...)
```
Which at _compile time_ will search the `VisitStmt_` overloads provided by `StmtExprVisitor`.  (Although compile-time dispatch will still use the vtable, and thus resolve to a suitable `LinearAccessPatternFinder method if it exists.)

(b) We're unwittingly invoking the wrong overload provided by `LinearAccessPatternFinder` because C++ can automatically cast the _argument_ to one of the overloads provided by `LinearAccessPatternFinder`.  And TVM's testing just never noticed the mistake.

Lunderberg: (A little bit late to the discussion, but I hope this helps.)

I looked a bit into the warning, and it looks like part of it is because the `DataTypeLegalizer` defines the `VisitStmt_` as public, whereas the `StmtExprMutator` defines them as protected.  So the other instances of this pattern are overriding a protected method, which doesn't trigger the warning.

As example code. the warning is trying to avoid cases like the third one below, where the static type causes an unexpected overload set when calling a method on the derived type.

```c++
class Base {
public:
  virtual func(int x) {}
  virtual func(double x) {}
};

class Derived {
public:
  virtual func(double x) {}
};

int main() {
  // Static type is base, uses Base's vtable.
  std::unique_ptr<Base> base = std::make_unique<Base>();
  base->func(5); // Calls Base::func(int)

  // Static type is base, uses Derived's vtable.
  std::unique_ptr<Base> derived = std::make_unique<Dervied>();
  derived->func(5); // Calls Base::func(int)

  // Static type is Derived, uses Derived's vtable.  Overload
  // resolution selects `Derived::func(double)` because the overload
  // resolution set is based on the static type, and doesn't include
  // `Base::func(int)`.
  std::unique_ptr<Derived> derived = std::make_unique<Dervied>();
  derived->func(5); // Calls Derived::func(double)
}
```

cconvey:
@vinx13 : I've updated the PR to mark the new methods as `protected`, because IIUC that's more in line with your original intentions.

I left the `#pragma`s in place, because using `protected` wasn't sufficient to make the warning go away.

cconvey:
I've reworked this based on our discussion:
- The warnings are eliminated via `using` statements, rather than pragmas.
- I changed the methods from `public` to `protected`, to better match (AFAICT) @vinx13 's intent.

@Lunderberg @vinx13 : Ready for review imho.

Lunderberg (approved):
LGTM!  Can we apply a similar change in the [`DataTypeLegalizer`](https://github.com/apache/tvm/blob/main/include/tvm/tir/stmt_functor.h#L506)?  (Either as part of this PR or another.)  It has the same issue of the `VisitStmt_` methods being exposed, when external users should go through `StmtExprMutator::operator()` instead.

cconvey:
> Can we apply a similar change in the [DataTypeLegalizer](https://github.com/apache/tvm/blob/main/include/tvm/tir/stmt_functor.h#L506)? (Either as part of this PR or another.) It has the same issue of the VisitStmt_ methods being exposed, when external users should go through StmtExprMutator::operator() instead.

Thanks for the idea! It's not too much scope creep, so I'll add that to this PR.

vinx13 (approved):

PR #13270, opened 2022-11-02
[build][doc] Fix clang doxygen warnings

cconvey:
Fix occurrences of clang's `-Wdocumentation-unknown-command` warning.

### DIFF
--- a/src/relay/backend/annotate_used_memory.cc
+++ b/src/relay/backend/annotate_used_memory.cc
@@ -52,6 +52,7 @@ namespace backend {
  * A simple example:
  *
  * Before:
+ * \verbatim
  * def @main(%input: Tensor[(1, 2, 2, 4), int8]) -> Tensor[(1, 2, 2, 4), int8] {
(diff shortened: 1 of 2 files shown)


driazati (approved):
Changes look OK, is there a way we can turn this warning into an error + ensure its tested in CI so they don't slip back in?

cconvey:
> Changes look OK, is there a way we can turn this warning into an error + ensure its tested in CI so they don't slip back in?

Clang (and I assume gcc) lets you control this using variations of `-Werror*` and `-Wno-error*`.  (See [docs](https://clang.llvm.org/docs/UsersManual.html#options-to-control-error-and-warning-messages).)

I don't know if the community would be keen on that change though, since TVM doesn't define a list of "supported" compiler versions.

What I can say is that I'm now getting clean compilations of TVM using clang 15.0.3, and none of the warnings I fixed was really that big a deal.  I.e., I only fixed them out of preference, not necessity.

PR #13271, opened 2022-11-02
[Hexagon] Lint tests part 2

mehrdadh:
This PR adds more pylint to hexagon tests. Also this PR moves the slice OP tests to a sub directory.


mehrdadh:
@masahi please take a look.

tests/python/contrib/test_hexagon/metaschedule_e2e/test_resnet50_fp16.py
```diff
@@ -44,14 +40,19 @@ def convert_conv2d_layout(mod, desired_layouts):
...
         pytest.skip(msg="Run python export_models.py first.")
 
-    with open(model_json, "r") as fi:
-        mod = tvm.ir.load_json(fi.read())
+    with open(model_json, "r") as file:
+        mod = tvm.ir.load_json(file.read())
 
-    with open(model_params, "rb") as fi:
-        params = relay.load_param_dict(fi.read())
+    with open("resnet50_fp16.params", "rb") as file:
```
masahi, line 54: Can you restore the `model_params` variable rather than hard-coding it? The same comment for `test_resnet50_int8.py`.

tests/python/contrib/test_hexagon/metaschedule_e2e/export_models.py
```diff
@@ -34,14 +37,16 @@ def export_resnet50_fp16():
...
 
-    with open("resnet50_fp16.json", "w") as fo:
-        fo.write(tvm.ir.save_json(mod))
+    with open("resnet50_fp16.json", "w") as file:
+        file.write(tvm.ir.save_json(mod))
 
-    with open("resnet50_fp16.params", "wb") as fo:
-        fo.write(relay.save_param_dict(params))
+    with open("resnet50_fp16.params", "wb") as file:
+        file.write(relay.save_param_dict(params))
```
masahi, line 44: I think it is ok to add `# pylint: disable=invalid-name` to avoid these changes. We use that everywhere.

mehrdadh: I reversed them back to global variables, but made them capital.

tests/python/contrib/test_hexagon/metaschedule_e2e/test_resnet50_fp16.py
mehrdadh: fixed this

tests/python/contrib/test_hexagon/metaschedule_e2e/export_models.py
masahi: I meant the change from `fo` to `file`.  I'm generally not happy with `black` complaining about short names like that.

masahi (approved):

tests/python/contrib/test_hexagon/metaschedule_e2e/export_models.py
mehrdadh: If it is bothering let's disable it at the TVM level. Otherwise we're gonna have bunch of `pylint_disable` tags which kind of defeats the purpose of pylint.

PR #13277, opened 2022-11-03
[Frontend][PaddlePaddle] Add test case for interpolate op convert func…

woobinw:
Add test case for interpolate op convert function #13247 

@heliqi 
@jiangjiajun 
cc @junrushao


heliqi (approved):
LGTM

jiangjiajun (approved):
LGTM

junrushao:
would you like to fix the broken unittest so we could get it in?

junrushao (approved):
LGTM!

PR #13278, opened 2022-11-03
[BugFix][Pattern] Fixed a crash when AltPattern and FunctionPattern are used nested

liangW-intellif:
The PatternGroup doesn not check if the FunctionPattern is matched while processing the FunctionPattern, but when FunctionPattern is nested with AltPattern, the FunctionPattern may not be matched, resulting in a crash when looking up matched nodes.
This commit adds a check at handling FunctionPattern to fix this crash, same as the checks at handling other patterns(such as `dataflow_matcher.cc:667`)
cc @masahi @wrongtest-intellif


masahi:
@tvm-bot rerun

masahi (approved):

PR #13283, opened 2022-11-03
[ci] Dis-allow any non-S3 URLs in CI

driazati:
This PR makes it so any URLs accessed in tests in CI must be hosted in
S3. This improves reliability as we've seen even files on GitHub
sometimes serve 503s even when everything else is working fine. This
raises an error if any unallowed URL is detected and adds the remaining
few.


tests/scripts/request_hook/request_hook.py
```diff
@@ -56,26 +126,79 @@
...
+            msg = (
+                f"Uncaught URL found in CI: {url}. "
+                "A committer must upload the relevant file to S3 via"
+                "https://github.com/apache/tvm/actions/workflows/upload_ci_resource.yml"
+                "and add it to the mapping in tests/scripts/request_hook/request_hook.py"
+            )
+            raise RuntimeError(msg)
+
+        new_url = URL_MAP[url]
+        # Just handle the '?raw=true' query
```
areusch, line 198: why not urllib.quote?

driazati: We don’t actually want to quote the full URL here(it’d end up like https%3F%3F…) just the query params. We could separate out the different parts but this seemed like the simplest way

areusch: hm, maybe here we are just replacing the hostport with BASE and then urlencoding everything after the path?

areusch (approved):

PR #13289, opened 2022-11-04
[Relay][Op] Add support for large index fp16 mean and var

jwfromm:
Due to the maximum value of FP16 being quite low (65536), it's very common to be unable to express large reductions using it. This occurs most commonly in simple operators like `mean` and `var`, where the sum of elements is used as a divisor. Given that the actual complexity of those operators is quite low, it makes sense to just perform them in fp32 instead of fp16. This PR adds that casting and a few tests that previously would fail in FP16.


jwfromm:
@sfvaroglu can you take a look at this PR?

jwfromm:
@tvm-bot rerun

sfvaroglu:
LGTM! Thanks @jwfromm!

AndrewZhaoLuo (approved):

PR #13292, opened 2022-11-04
[TIR] Preserve loop annotation after loop partitioning

wrongtest-intellif:
Preserve loop annotations when the loop is get partitioned. Also we bind the loop region info to the analyzer for some cases 
some partition condition could not get solved due to unknown (but trivial) loop region.


Hzfengsy (approved):
LGTM

PR #13297, opened 2022-11-04
[TE] Make `elem_offset` of the buffers created by `te.extern` a variable to avoid crash

masahi:
Currently, when compiling MobileBERT on x86, we get a crash from https://github.com/apache/tvm/blob/11d22bdc1bd45d952eb140684e64f01438b7f516/src/tir/transforms/arg_binder.cc#L102-L106.

This is caused from the subgraph attached in the test. Since https://github.com/apache/tvm/pull/11341, `concatenate` is implemented as a TE extern on x86. `te.extern(...)` doesn't explicitly set the `elem_offset` parameter of input / output buffers it creates, which ends up `elem_offset` of these buffers being the default value, 0. This causes a problem from the above check in `BindBuffer`, since the subgraph does create a non-zero `elem_offset` slice from a buffer whose `elem_offset` is set to zero. 

Since we don't require buffers created in `te.extern(...)` to have zero `elem_offset`, we can avoid this problem by explicitly setting `elem_offset` to be a variable there. This fix was found by @Lunderberg and I'm sending this PR on behalf of him. He also suggested making the `strides` parameter an array of `Var` as well, but I didn't do it to keep the change minimal.

cc @Lunderberg @tkonolige @shtinsa

### DIFF
--- a/python/tvm/te/operation.py
+++ b/python/tvm/te/operation.py
@@ -326,7 +326,11 @@ def extern(
         if not isinstance(t, _tensor.Tensor):
             raise ValueError("expect inputs to be tensor")
         if in_buffers is None:
-            input_placeholders.append(tvm.tir.decl_buffer(t.shape, t.dtype, t.op.name))
+            input_placeholders.append(
+                tvm.tir.decl_buffer(
+                    t.shape, t.dtype, t.op.name, elem_offset=tvm.tir.Var("elem_offset", "int32")
+                )
+            )
         types.add(t.dtype)
 
     if dtype is None:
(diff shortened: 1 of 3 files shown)


vinx13 (approved):

shingjan (approved):
LGTM!

Lunderberg:
LGTM, and thank you for making this change!  For context on the fix, this didn't cause an issue previously, because the default `elem_offset=0` could bind to an entire buffer.  It only became necessary when the buffer being bound as an argument to the `te.extern` was a view into a larger backing allocation, which could occur after inlining.

PR #13299, opened 2022-11-04
[TIR] Introduce ReduceBranchingThroughOvercompute

Lunderberg:
This PR introduces a new transform, which attempts to reduce branching by adding overcompute, where the overcompute has provably no effect on any observable value outside the PrimFunc.

### DIFF
--- a/tests/python/unittest/test_tir_transform_simplify.py
+++ b/tests/python/unittest/test_tir_transform_simplify.py
@@ -1689,5 +1689,26 @@ class TestSimplifyBufferStore(BaseBeforeAfter):
         A[0] = 12
 
 
+# class TestTemp(BaseBeforeAfter):
+#     """Simplification using prior known"""
+
+#     transitively_prove_inequalities = False
+#     convert_boolean_to_and_of_ors = False
+#     apply_constraints_to_boolean_branches = False
+#     propagate_knowns_to_prove_conditional = False
+#     propagate_knowns_to_simplify_expressions = False
+
+#     def before(A: T.Buffer[16, "int32"]):
+#         for i in T.serial(16):
+#             if i != 0:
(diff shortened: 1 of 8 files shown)


Lunderberg:
@csullivan This PR is ready for review and passing all unit tests, if you have a chance to review.

tests/python/unittest/test_tir_transform_simplify.py
```diff
@@ -1689,5 +1689,26 @@ def expected(A: T.Buffer[1, "int32"]):
...
+#             if i != 0:
+#                 if i * i == 0:
+#                     A[i] = 5
+
+#     def expected(A: T.Buffer[16, "int32"]):
+#         for i in T.serial(16):
+#             if i != 0:
+#                 T.evaluate(0)
+
+
```
csullivan, line 1712: Will need to remove this.

Lunderberg: Thank you, and removed!

csullivan:
@tvm-bot rerun

LeiWang1999:
hi @Lunderberg , did we implement `Stmt Simplify(Stmt stmt, arith::Analyzer* analyzer);`?

Lunderberg:
@LeiWang1999 We did, yes.  It's located [here](https://github.com/apache/tvm/blob/main/src/tir/transforms/simplify.h#L38).

That said, in most cases it's better to apply [`tir::transform::Simplify`](https://github.com/apache/tvm/blob/main/src/tir/transforms/simplify.h#L38) instead.  That way, the full context is available for a single simplification call.  Simplifying a statement in isolation may miss context-dependent simplifications.  (e.g. Within an `if i==0` block, `if i + 1 < 16` could be simplified to to `True`.)

PR #13300, opened 2022-11-04
[ci] Split Jenkinsfile into platform-specific jobs

driazati:
This breaks up the Jenkinsfile into ones for GPU, CPU, etc. This removes a false dependency between the build and test steps (e.g. before the GPU tests had to wait on the Hexagon build to complete) and makes the Jenkins UI a bit better since there's not 30 tests to scroll through to find a failure. An example can be found in my fork here: https://github.com/driazati/tvm/pull/38 in the checks box. Before this is merged https://github.com/tlc-pack/ci/blob/main/jenkins/jenkins-jobs/prod/tvm.yaml will need to be updated to accept webhooks from apache/tvm instead of my fork.

See #13337 for more context


mehrdadh:
This is great!!

quic-sanirudh:
This is amazing. It'll make it so much easier to find failures, thanks a lot for this.

ci/jenkins/templates/utils/base.groovy.j2
```diff
@@ -91,41 +91,15 @@ docker_build = 'docker/build.sh'
...
-{% include "ci/jenkins/Build.groovy.j2" %}
-{% include "ci/jenkins/Test.groovy.j2" %}
-{% include "ci/jenkins/Deploy.groovy.j2" %}
+{% include "utils/Prepare.groovy.j2" %}
+{% include "utils/Build.groovy.j2" %}
 
 
 cancel_previous_build()
 
 prepare()
```
gigiblender, line 105: We could add a newline here

ci/jenkins/generate.py
```diff
@@ -173,17 +147,30 @@ def change_type(lines: List[str]) -> Change:
...
+
+if __name__ == "__main__":
+    help = "Regenerate Jenkinsfile from template"
+    parser = argparse.ArgumentParser(description=help)
+    parser.add_argument("--force", action="store_true", help="always overwrite timestamp")
+    parser.add_argument("--check", action="store_true", help="just verify the output didn't change")
+    args = parser.parse_args()
+
+    sources = TEMPLATES_DIR.glob("*jenkinsfile.groovy.j2")
+    changes = [update_jenkinsfile(source) for source in sources if source.name != "base.groovy.j2"]
```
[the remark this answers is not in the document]
driazati: its just there to keep from copying `base.groovy.j2` to the generated directory since it doesn't do anything outside of provide helper functions

ci/jenkins/generated/arm_jenkinsfile.groovy
```diff
@@ -0,0 +1,1042 @@
...
+        """, label: 'Docker image names')
+
+        is_docs_only_build = sh (
+          returnStatus: true,
+          script: './ci/scripts/git_change_docs.sh',
+          label: 'Check for docs only changes',
+        )
+        skip_ci = should_skip_ci(env.CHANGE_ID)
+        skip_slow_tests = should_skip_slow_tests(env.CHANGE_ID)
+        rebuild_docker_images = sh (
```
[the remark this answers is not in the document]
driazati: Right now the image build is just to check the build so it doesn't actually use them later on. We should definitely come in and try to clean this up some more later on since it's still a bit tangled up with the old monolithic job

ci/jenkins/generated/docker_jenkinsfile.groovy
```diff
@@ -0,0 +1,928 @@
...
+    docker tag \
+      ${ecr_image} \
+      ${hub_image}
+    retry 5 docker push ${hub_image}
+    """,
+    label: "Update ${hub_image} on Docker Hub",
+  )
+}
+
+stage('Docker Image Build') {
```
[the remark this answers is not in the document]
driazati: As long as Jenkins doesn't blow up when running it's ok (the JVM has a 64k limit on method size or something so if there's too much generated code things need to be split up)

ci/jenkins/templates/utils/base.groovy.j2
driazati: The EOF-adjacent newlines get caught by https://github.com/apache/tvm/blob/main/tests/lint/whitespace.sh, so we can't really change them without adding an exception

ci/jenkins/generated/arm_jenkinsfile.groovy
```diff
@@ -0,0 +1,933 @@
...
+
+    docker image ls --all
+    """,
+    label: 'Clean old Docker images',
+  )
+
+  if (image.contains("amazonaws.com")) {
+    // If this string is in the image name it's from ECR and needs to be pulled
+    // with the right credentials
+    // ecr_pull(image)
```
areusch, line 192: do we need to re-enable for CI rebuild?

areusch (commented):
thanks @driazati, couple questions but i agree with splitting these up

ci/jenkins/templates/utils/Build.groovy.j2
```diff
@@ -0,0 +1,57 @@
...
+}
+
+def fsim_test(image) {
+  sh (
+    script: "${docker_run} ${image} ./tests/scripts/task_python_vta_fsim.sh",
+    label: 'Run VTA tests in FSIM',
+  )
+}
+
+def make_standalone_crt(image, build_dir) {
```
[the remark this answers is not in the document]
driazati: discussed a bit offline, we'll save the refactoring stuff for a follow up

ci/jenkins/templates/arm_jenkinsfile.groovy.j2
```diff
@@ -0,0 +1,83 @@
+{% include "utils/base.groovy.j2" with context %}
```
areusch, line 1: can we still put the ASF header in this file?

docker/dev_common.sh
```diff
@@ -28,24 +28,10 @@ INVOCATION_PWD="$(pwd)"
...
-    done
-}
-
-
 function lookup_image_spec() {
-    img_line=$(cat "${GIT_TOPLEVEL}/Jenkinsfile" | filter_jenkinsfile | grep -E "^${1} = ")
+    img_line=$(python3 "${GIT_TOPLEVEL}/ci/jenkins/data.py" "$1")
     if [ -n "${img_line}" ]; then
-        img_spec=$(echo "${img_line}" | sed -E "s/${1} = '([^\"]*)'/\1/")
+        img_spec=$(echo "${img_line}" | sed -E "s/${1} = \"([^\"]*)\"/\1/")
```
areusch, line 34: i think this is unnecessary now right?

areusch (commented):
thanks @driazati a couple small last things

areusch (approved):
thanks @driazati !

shingjan:
thanks @driazati ! Great work!

alter-xp:
hi @driazati , It seems that the contents related to how to regenerate Jenkinsfile in `ci/README.md` and `ci/jenkins/README.md` are already out of date. Is there any new tutorial?

driazati:
> hi @driazati , It seems that the contents related to how to regenerate Jenkinsfile in `ci/README.md` and `ci/jenkins/README.md` are already out of date. Is there any new tutorial?

Ah good catch, I will follow up with a PR to update the docs but you can run this to re-sync everything

```bash
python3 -m pip install jinja2
python ci/jenkins/generate.py
```

PR #13301, opened 2022-11-05
[TIR] Update ReductionIterNotIndexOutputBuffer to check BlockRealizeN…

nverke:
…odes match_buffer statements when validating writes

Previously this check did not take into account any match_buffer statements and consequently would fail for tensorized schedules. Now it takes these into account when possible. 

cc @vinx13 @Hzfengsy

### DIFF
--- a/src/tir/schedule/analysis/reducer.cc
+++ b/src/tir/schedule/analysis/reducer.cc
@@ -572,9 +572,25 @@ bool ReductionIterNotIndexOutputBuffer(const Block& block) {
     if (!store) {
       return true;
     }
-    ICHECK(buffer_written.count(store->buffer.get()))
-        << "ValueError: The buffer \"" << store->buffer
-        << "\" is written in the block but is not in the block's signature";
+    const auto* body_block = block->body.as<BlockRealizeNode>();
+    if (body_block) {
+      bool write_is_covered_by_match_buffer = false;
+      for (const MatchBufferRegion& match_region : body_block->block->match_buffers) {
+        if (store->buffer.get() == match_region->buffer.get() &&
+            buffer_written.count(match_region->source->buffer.get())) {
+          write_is_covered_by_match_buffer = true;
+          break;
+        }
+      }
+      ICHECK(buffer_written.count(store->buffer.get()) || write_is_covered_by_match_buffer)
+          << "ValueError: The buffer \"" << store->buffer
+          << "\" is written in the block but is not in the block's signature nor is it covered by "
+             "a match_buffer";
+    } else {
+      ICHECK(buffer_written.count(store->buffer.get()))
+          << "ValueError: The buffer \"" << store->buffer
+          << "\" is written in the block but is not in the block's signature";
+    }
     for (const PrimExpr& index : store->indices) {
       if (f_uses_reduction_block_var(index)) {
         affected = true;
--- a/src/tir/schedule/analysis/reducer.cc
+++ b/src/tir/schedule/analysis/reducer.cc
@@ -584,11 +584,13 @@ bool ReductionIterNotIndexOutputBuffer(const Block& block) {
       return true;
     }
 
-    bool write_is_covered_by_match_buffer = match_buffer_sources.count(store->buffer.get()) && buffer_written.count(match_buffer_sources.find(store->buffer.get())->second);
+    bool write_is_covered_by_match_buffer =
+        match_buffer_sources.count(store->buffer.get()) &&
+        buffer_written.count(match_buffer_sources.find(store->buffer.get())->second);
     ICHECK(buffer_written.count(store->buffer.get()) || write_is_covered_by_match_buffer)
         << "ValueError: The buffer \"" << store->buffer
         << "\" is written in the block but is not in the block's signature nor is it covered by "
-            "a match_buffer";
+           "a match_buffer";
     for (const PrimExpr& index : store->indices) {
       if (f_uses_reduction_block_var(index)) {
         affected = true;
--- a/src/tir/schedule/analysis/reducer.cc
+++ b/src/tir/schedule/analysis/reducer.cc
@@ -563,34 +563,32 @@ bool ReductionIterNotIndexOutputBuffer(const Block& block) {
       return reduction_block_iters.count(var);
     });
   };
+
+  std::unordered_map<const BufferNode*, const BufferNode*> match_buffer_sources;
+  for (const MatchBufferRegion& region : block->match_buffers) {
+    match_buffer_sources[region->buffer.get()] = region->source->buffer.get();
+  }
   bool affected = false;
   PreOrderVisit(block->body, [&](const ObjectRef& obj) {
     if (affected) {
       return false;
     }
+    const auto* block_node = obj.as<BlockNode>();
+    if (block_node) {
+      for (const MatchBufferRegion& region : block_node->match_buffers) {
+        match_buffer_sources[region->buffer.get()] = region->source->buffer.get();
+      }
+    }
     const auto* store = obj.as<BufferStoreNode>();
     if (!store) {
       return true;
     }
-    const auto* body_block = block->body.as<BlockRealizeNode>();
-    if (body_block) {
-      bool write_is_covered_by_match_buffer = false;
-      for (const MatchBufferRegion& match_region : body_block->block->match_buffers) {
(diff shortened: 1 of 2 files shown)


Hzfengsy:
Thanks @nverke for the enhancement. Could you please add a regression test case?

nverke:
> Thanks @nverke for the enhancement. Could you please add a regression test case?

Just added!

nverke:
@tvm-bot rerun

src/tir/schedule/analysis/reducer.cc
```diff
@@ -572,9 +572,25 @@ bool ReductionIterNotIndexOutputBuffer(const Block& block) {
     if (!store) {
       return true;
     }
-    ICHECK(buffer_written.count(store->buffer.get()))
-        << "ValueError: The buffer \"" << store->buffer
-        << "\" is written in the block but is not in the block's signature";
+    const auto* body_block = block->body.as<BlockRealizeNode>();
```
vinx13, line 575: checking block->body is not sufficient. It is possible the inner block has outer loops. 
consider the case:
```
block1
  for …
     block2
        match_buffer
        buffer_store
```

In this case, we need to the parent block of the store statement

nverke: Hmm I am not sure if I understand. Are you talking about a situation like this? 
```
@T.prim_func
def nested_reduction_loop_with_match_buffers(
    in0: T.Buffer[(4, 4, 4), "int8"],
    in1: T.Buffer[(4, 4, 4), "int8"],
    out: T.Buffer[(4, 4, 4), "int8"],
) -> None:
    # body
    # with T.block("root")
    for y in T.serial(4):
        with T.block("C"):
            T.reads(in0[y, 0:4, 0:4], in1[y, 0:4, 0:4])
            T.writes(out[y, 0:4, 0:4])
            for x in T.serial(4):
                with T.block("C"):
                    T.reads(in0[y, x, 0:4], in1[y, x, 0:4])
                    T.writes(out[y, x, 0:4])
                    A = T.match_buffer(in0[y, x, 0:4], [4], dtype="int8", offset_factor=1)
                    B = T.match_buffer(in1[y, x, 0:4], [4], dtype="int8", offset_factor=1)
                    C = T.match_buffer(out[y, x, 0:4], [4], dtype="int8", offset_factor=1)
                    A_i8x4: T.int8x4 = A[0:4]
                    A_i32: T.int32 = T.reinterpret(A_i8x4, dtype="int32")
                    B_i8x4: T.int8x4 = B[0:4]
                    B_i32: T.int32 = T.reinterpret(B_i8x4, dtype="int32")
                    C[0:4] = T.reinterpret(A_i32 + B_i32, dtype="int8x4")
```
My understanding is that this check is on the level of the store statement and already has collected all of the write regions for the parent loops so just adding the regions that the match buffers check should be enough. 

Alternatively are you referring to something like this? 
```
@T.prim_func
def nested_reduction_loop_with_match_buffers(
    in0: T.Buffer[(4, 4, 4), "int8"],
    in1: T.Buffer[(4, 4, 4), "int8"],
    out: T.Buffer[(4, 4, 4), "int8"],
) -> None:
    # body
    # with T.block("root")
    for y in T.serial(4):
        with T.block("C"):
            T.reads(in0[y, 0:4, 0:4], in1[y, 0:4, 0:4])
            T.writes(out[y, 0:4, 0:4])
            A = T.match_buffer(in0[y, 0:4, 0:4], [4, 4], dtype="int8", offset_factor=1)
            B = T.match_buffer(in1[y, 0:4, 0:4], [4, 4], dtype="int8", offset_factor=1)
            C = T.match_buffer(out[y, 0:4, 0:4], [4, 4], dtype="int8", offset_factor=1)
            for x in T.serial(4):
                with T.block("C"):
                    T.reads(in0[y, x, 0:4], in1[y, x, 0:4])
                    T.writes(out[y, x, 0:4])
                    A_i8x4: T.int8x4 = A[x, 0:4]
                    A_i32: T.int32 = T.reinterpret(A_i8x4, dtype="int32")
                    B_i8x4: T.int8x4 = B[x, 0:4]
                    B_i32: T.int32 = T.reinterpret(B_i8x4, dtype="int32")
                    C[x, 0:4] = T.reinterpret(A_i32 + B_i32, dtype="int8x4")
```
Here I believe we are still able to pickup the match buffers from the body block.

vinx13: I'm referring to your former example. In this case, `if (body_block)` will always be false, and `match_buffer` will not be checked.

nverke: Alright made changes accordingly and added a test to validate this case as well as another test that on the other scenario even though that was never an issue.

vinx13 (approved):

tmoreau89:
Thank you @nverke @vinx13 - PR has been merged!

tests/python/unittest/test_tir_transform_lower_cross_thread_reduction.py
```diff
@@ -487,6 +487,149 @@ def lowered_single_reduction_loop_with_block_predicate(
...
+            for x in T.serial(4):
+                xr = T.axis.reduce(4, x)
+                with T.init():
+                    for i in T.serial(4):
+                        with T.block("C_init"):
+                            ii = T.axis.spatial(4, i)
+                            T.reads()
+                            T.writes(out[yi, ii])
+                            out[yi, ii] = 0
+                with T.block("C"):
```
junrushao, line 619: I'm late to the party! Just a quick note: this TVMScript is not valid TIR and I happened to detect it when using the new TVMScript parser which checks more carefully :-)

More specifically, both `T.init()` and `T.axis.reduce` should be placed immediately under a TIR block, while line 611 and line 612 are not :-(

junrushao: If we print out the testcase using `nested_reduction_loop_with_outer_match_buffers.show()`, then the TIR looks like:

```python
@T.prim_func
def func(in0: T.Buffer[(4, 16), "int8"], in1: T.Buffer[(4, 16), "int8"], out: T.Buffer[(4, 4), "int32"]):
    # body
    # with T.block("root")
    for y in T.serial(4):
        with T.block("C"):
            yi = T.axis.spatial(4, y)
            xr = T.axis.reduce(4, x)
            T.reads(in0[yi, 0 : 16], in1[yi, 0 : 16])
            T.writes(out[yi, 0 : 4])
            A = T.match_buffer(in0[yi, 0 : 16], [16], dtype="int8", offset_factor=1)
            B = T.match_buffer(in1[yi, 0 : 16], [16], dtype="int8", offset_factor=1)
            C = T.match_buffer(out[yi, 0 : 4], [4], dtype="int32", offset_factor=1)
            with T.init():
                for i in T.serial(4):
                    with T.block("C_init"):
                        ii = T.axis.spatial(4, i)
                        T.reads()
                        T.writes(out[yi, ii])
                        out[yi, ii] = 0
            for x in T.serial(4):
                with T.block("C"):
                    T.reads(out[yi, xr], in0[yi, yi * 4 + xr : yi * 4 + xr + 4], in1[yi, yi * 4 + xr : yi * 4 + xr + 4])
                    T.writes(out[yi, xr])
                    A_i8x4: T.int8x4 = A[yi * 4 + xr:yi * 4 + xr + 4]
                    A_i32: T.int32 = T.reinterpret(A_i8x4, dtype="int32")
                    B_i8x4: T.int8x4 = B[yi * 4 + xr:yi * 4 + xr + 4]
                    B_i32: T.int32 = T.reinterpret(B_i8x4, dtype="int32")
                    C[xr] = A_i32 + B_i32 + C[xr]
```

where we may see some use-before-def issues

junrushao: i would love to temporarily exclude this particular TVMScript from testing, but happy to merge it back if you have a follow-up PR to fix :-)

nverke: Oh interesting! Was not aware of this requirement. I can follow up with a commit in a few days or you can remove the test if its causing issues.

nverke: https://github.com/apache/tvm/pull/13373/files

PR #13302, opened 2022-11-05
[docs][tvmc] Use correct argument to reuse tuning records

StrongerXi:
Update tvmc tutorial code to use correct argument for reusing tuning records. Specifically, current code uses `tuning_records`, which is meant for saving the generated tuning results, not reusing prior results. We should use `prior_records` instead.

### DIFF
--- a/gallery/tutorial/tvmc_python.py
+++ b/gallery/tutorial/tvmc_python.py
@@ -248,7 +248,7 @@ testing.utils.install_request_hook(depth=3)
 #      ...
 #
 #      # Later run tuning and reuse tuning results
-#      tvmc.tune(model, target="llvm",tuning_records=log_file)
+#      tvmc.tune(model, target="llvm", prior_records=log_file)
 #
 # Method 2:
 #    .. code-block:: python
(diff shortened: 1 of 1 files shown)


gromero (changes requested):
@StrongerXi Thanks for the fix! LGTM.

Could you just please update the Pull Request description text? Otherwise the PR will land in TVM's tree as a commit with a body text like "As title. I found this while ... " which is not appropriate for a commit message.

Please take a look at https://github.com/apache/tvm/blob/main/docs/contribute/pull_request.rst#commit-message-guideline for more details on how to right the description.

Thank you!

StrongerXi:
> @StrongerXi Thanks for the fix! LGTM.
> 
> Could you just please update the Pull Request description text? Otherwise the PR will land in TVM's tree as a commit with a body text like "As title. I found this while ... " which is not appropriate for a commit message.
> 
> Please take a look at https://github.com/apache/tvm/blob/main/docs/contribute/pull_request.rst#commit-message-guideline for more details on how to right the description.
> 
> Thank you!

Just did, let me know if it needs more tweaking:).

gromero (approved):

gromero:
@StrongerXi Thank you. The change is merged.

PR #13118, opened 2022-10-18
[Android] Fix cpp_rpc build for android with NDK>=23

echuraev:
With NDK>= 23 when we specify explicitly linker then we got the following error:

```
ld.gold: --no-rosegment: unknown option
```

From the github repository of NDK, I found that the only one right way to configure linker is using `-DANDROID_LD` variable: https://github.com/android/ndk/issues/1426#issuecomment-760432467

Removed setting linker manually and by default `LLD` will be used.

Checked that it works on the following versions of NDK:
- 20.0.5594570
- 21.4.7075529
- 22.1.7171670
- 23.0.7599858
- 23.1.7779620
- 24.0.8215888
- 25.1.8937393

cc: @csullivan, @elvin-n

PR #13305, opened 2022-11-07
[CPP_RPC][ANDROID] Fix cpp_rpc build failure

srkreddy1238:
CPP build failure for Android devices with NDK version < 23. 
NDK below 23 require additional linker flags.


srkreddy1238:
This PR address the rpc build caused by https://github.com/apache/tvm/pull/13118

apps/cpp_rpc/CMakeLists.txt
```diff
@@ -32,6 +32,14 @@ if (OS)
    endif()
 endif()
 
+if(USE_OPENCL)
```
elvin-n, line 35: tried to build this using follow command line `cmake  -DCMAKE_TOOLCHAIN_FILE=Android/Sdk/ndk/22.1.7171670/build/cmake/android.toolchain.cmake -DANDROID_ABI=arm64-v8a -DANDROID_NATIVE_API_LEVEL=android-23 -DCMAKE_FIND_ROOT_PATH_MODE_PACKAGE=ON -DANDROID_STL=c++_static ..` and got follow error:
```
CMake Error at apps/cpp_rpc/CMakeLists.txt:37 (if):
  if given arguments:

    "VERSION_LESS" "23"

  Unknown arguments specified
```
cmake version: 3.18.4 on ubuntu 21.10

elvin-n (changes requested):

apps/cpp_rpc/CMakeLists.txt
lhez: Does CMake 3.18.4 have a bug? [`VERSION_LESS`](https://cmake.org/cmake/help/latest/command/if.html#version-comparisons) does not seem to be a new CMake feature.

elvin-n: I agree that it must not be new, on the other hand I pointed above what I observe. Have you tried to reproduce issue on your side? Not through docker that we do not have access

apps/cpp_rpc/CMakeLists.txt
```diff
@@ -32,6 +32,14 @@ if (OS)
...
+   if (ANDROID_ABI)
+     if($ENV{ANDROID_NDK_MAJOR} VERSION_LESS "23")
```
lhez, line 37: I guess this is the problem - `$ENV`.

I don't think `$ENV` is needed. The CMake toolchain file from NDK should have already defined `ANDROID_NDK_MAJOR`. Hence, when that toolchain file is specified, `ANDROID_NDK_MAJOR` can be directly used as `${ANDROID_NDK_MAJOR}`. I think certain versions of CMake don't like `$ENV` being used with `if` and `VERSION_LESS`.

apps/cpp_rpc/CMakeLists.txt
lhez: I have no issue with CMake 3.20.6 + NDK 21.4.7075529 using

```cmake
if(${ANDROID_NDK_MAJOR} VERSION_LESS "23")
```

Note that there is no `$ENV`. Could you give it try?

Nor did I have to specify `ANDROID_NDK_MAJOR` with this in the cmake config.

docker/Dockerfile.ci_adreno
```diff
@@ -25,4 +25,5 @@ COPY install/ubuntu_install_androidsdk.sh /install/ubuntu_install_androidsdk.sh
 RUN bash /install/ubuntu_install_androidsdk.sh
 ENV ANDROID_HOME=/opt/android-sdk-linux
 ENV ANDROID_NDK_HOME=/opt/android-sdk-linux/ndk/21.3.6528147
+ENV ANDROID_NDK_MAJOR=21
```
echuraev, line 28: I have doubts about explicit setting of the environment variable. As @lhez wrote, the `ANDROID_NDK_MAJOR` should be provided by NDK.

apps/cpp_rpc/CMakeLists.txt
```diff
@@ -32,6 +32,14 @@ if (OS)
...
+       set_property(TARGET tvm_rpc PROPERTY LINK_FLAGS -fuse-ld=gold)
```
echuraev, line 38: Also, I don't understand how `-fuse-ld=gold` helps to fix your [issue](https://github.com/apache/tvm/pull/13118#issuecomment-1304477305). In the log, I see that the linker cannot find `libOpenCL.so`. By manually specifying `-fuse-ld=gold` you just force compiler to use `ld.gold` and how changing linker helps you find this library? Unfortunately, I'm not able to reproduce this issue locally with docker. I don't have access to pull the `adreno_ci` docker image.

srkreddy1238:
I used ENV as in CI dockers this is the way we can supply the NDK version. 

For individual compilations  exporting in environment or prefixing NDK version should work.

```ANDROID_NDK_MAJOR=21 cmake -DCMAKE_TOOLCHAIN_FILE=Android/Sdk/ndk/22.1.7171670/build/cmake/android.toolchain.cmake -DANDROID_ABI=arm64-v8a -DANDROID_NATIVE_API_LEVEL=android-23 -DCMAKE_FIND_ROOT_PATH_MODE_PACKAGE=ON -DANDROID_STL=c++_static ..```

File: apps/cpp_rpc/CMakeLists.txt (hunk 4 above)
Comment (srkreddy1238) [reply apps/cpp_rpc/CMakeLists.txt:38]: Yes, ```-fuse-ld=gold``` helps to find the lib and link it properly.

Comment (lhez) [reply apps/cpp_rpc/CMakeLists.txt:38]: With NDK 23 and above and `ld.gold`, the host `ld` will be used. That's why we have the failure `ld.gold: --no-rosegment: unknown option` - host `ld` doesn't have this option. My guess is `ld.gold` in older NDK may be able to get a path to `libOpenCL.so` through host `ld`. I think in @srkreddy1238 's env, `libOpenCL.so` exists in the host, but not in NDK (I don't think any NDK comes with `libOpenCL.so`).

**Comment by lhez** [comment]
@srkreddy1238 You only need to set the toolchain file. `ANDROID_NDK_MAJOR` is automatically available in `CMakeLists.txt` once you specify the toolchain file. I think it is better to set the toolchain file location using `ENV`, not `ANDROID_NDK_MAJOR`.

File: docker/Dockerfile.ci_adreno (hunk 3 above)
Comment (srkreddy1238) [reply docker/Dockerfile.ci_adreno:28]: It depends on cmake support. In our case it doesn't.

**Comment by srkreddy1238** [comment]
https://github.com/apache/tvm/blob/b807613c794321bf4fa765df026b1c3b454811b3/tests/scripts/task_build_adreno_bins.sh#L42

Tool chain is set.

File: apps/cpp_rpc/CMakeLists.txt (hunk 4 above)
Comment (echuraev) [reply apps/cpp_rpc/CMakeLists.txt:38]: Yes, for me, it's the same: `libOpenCL.so` is not a part of the NDK. When I build `tvm_rpc` with OpenCL support than I pass explicitly path to OpenCL SDK: `-DUSE_OPENCL=<path/to/sdk>` and it works fine also with `ld`. As I have mentioned in #13118, I checked that build works on many versions of NDK: 20.0.5594570, 21.3.6528147, 21.4.7075529, 22.1.7171670, etc. And I was surprised that it doesn't work in the docker.
I'm curious why, in case of @srkreddy1238 `ld` cannot find the `libOpenCL.so` but `ld.gold` can do it. In case then linkers have the same paths, I would expect that they will work in the same way.
Anyway, I don't think that this discussion is a stopper for merging this PR. But I'm just trying to understand.

**Comment by srkreddy1238** [comment]
made ```ANDROID_NDK_MAJOR``` to be an optional env variable to keep higher NDK versions work default.

File: apps/cpp_rpc/CMakeLists.txt [hunk 5]
```diff
@@ -32,6 +32,16 @@ if (OS)
    endif()
 endif()
 
+if(USE_OPENCL)
+   if (ANDROID_ABI)
+     if(DEFINED ENV{ANDROID_NDK_MAJOR1})
```
echuraev, line 37: typo: `ANDROID_NDK_MAJOR1` -> `ANDROID_NDK_MAJOR`?

elvin-n (approved):

masahi (approved):

Issue #13296, opened 2022-11-04
[Bug] TVM generates wrong results for ssd300_vgg16


saurabh-shandilya:
### Expected behavior
I am trying to run pytorch's ssd300_vgg16 model via TVM's VirtualMachine. I followed the instructions in  https://github.com/apache/tvm/issues/10050 to overcome the initial problems and now the network runs fine. However the results are all wrong. They don't match the pytorch reference output and really way off. The code is based on mask-rcnn example provided here -  https://tvm.apache.org/docs/how_to/deploy_models/deploy_object_detection_pytorch.html. I've tried targets llvm and cuda both and results are bad for both. 

Pytorch reference output *************************
(tensor([[2.4417e+02, 2.2165e+02, 3.6217e+02, 2.9035e+02],
        [2.9954e+02, 2.6913e+02, 4.6809e+02, 3.8972e+02],
        [1.9778e+02, 1.9647e+02, 2.5620e+02, 3.2252e+02],
        [2.7028e+01, 2.0530e+02, 1.0533e+02, 3.5209e+02],
        [3.4073e+02, 2.0001e+02, 4.3230e+02, 3.7326e+02],
        [1.4415e+02, 2.0797e+02, 1.6326e+02, 2.6495e+02],
........],
       grad_fn=<StackBackward0>), 

tensor([0.9727, 0.9540, 0.8788, 0.8229, 0.8031, 0.7723, 0.5995, 0.5492, 0.5456,
        0.5138, 0.3374, 0.2810, 0.1931, 0.1798, 0.1736, 0.1671, 0.1607, 0.1527,
        0.1512, 0.1445, 0.1230, 0.1188, 0.1182, 0.1161, 0.1160, 0.1159, 0.1152,
        0.1151, 0.1129, 0.1125, 0.1120, 0.1094, 0.1079, 0.1076, 0.1070, 0.1059,
        0.1059, 0.1057, 0.1047, 0.1025, 0.1004, 0.0993, 0.0989, 0.0981, 0.0975,
        0.0968, 0.0963, 0.0961, 0.0941, 0.0940, 0.0936, 0.0934, 0.0930, 0.0925,
        0.0918, 0.0908, 0.0902, 0.0902, 0.0901, 0.0899, 0.0893, 0.0889, 0.0874,
...], grad_fn=<IndexBackward0>), 
[...]

Issue #12292, opened 2022-08-03
[Bug] Tensor advanced indexing with boolean mask not working


NVukobrat:
Hi,

I’m trying to index the tensor in the following fashion: 
```py
x[0, mask]
```

where x is the input tensor (e.g. (2, 3) shape), while the mask is a boolean array (e.g. (3,) shape). For a better example, here are example values: 
```py
x = [ 
  [1, 2, 3], 
  [4, 5, 6] 
] 
mask = [True, True, False]
```

Therefore, when using NumPy as a reference, the results of `x[0, mask]` should be `[1, 2]` tensor. However, as TVM output I’m getting something like `[2, 2, 1]`. Can someone provide me with more information on this topic?

Also, for easier reproduction, below is TVM test similar to the example below (test is added and run from the following test file): 
```py
tests/python/relay/test_op_level3.py
```

Unit test:
```py
def test_adv_index_with_mask(target, dev, executor_kind):
    def verify_adv_index(data_shape, index, bool_mask_shape):
        dtype = "float32"
        inputs = [relay.var("data", relay.TensorType(data_shape, dtype))]
        np_data = np.random.uniform(size=data_shape).astype(dtype)
        np_indices = np.ones(bool_mask_shape, dtype=bool)
        np_indices[-1] = False
        inputs.append(relay.var("index_{}".format(index), shape=(), dtype="int"))
        inputs.append(relay.var("mask_{}".format(index), relay.TensorType(bool_mask_shape, "bool")))
        np_out = np_data[index, np_indices]
        np_args = [np_data] + [index, np_indices]
        out = relay.op.adv_index(inputs)

        func = relay.Function(inputs, out)
        op_res = relay.create_executor(executor_kind, device=dev, target=target).evaluate(func)(
            *np_args
        )
        tvm.testing.assert_allclose(op_res.numpy(), np_out, rtol=1e-5)

    verify_adv_index((2, 3), 1, (3,))

```

Lastly, this issue is also [posted yesterday on the discussion board](https://discuss.tvm.apache.org/t/tensor-advanced-indexing-with-boolean-mask/13257)

masahi:
There were related discussion on this topic.
https://github.com/apache/tvm/issues/9237#issuecomment-992428813 
https://github.com/apache/tvm/pull/7373

NVukobrat:
I see. Thanks for the references. I'll try to find a workaround for this issue for now. 

[...]

PR #13306, opened 2022-11-07
[Torch] Fix advanced indexing with boolean mask

masahi:
Fixes https://github.com/apache/tvm/issues/12292
Fixes https://github.com/apache/tvm/issues/13296

PyTorch uses the same op, `aten::index`, to represent indexing by both integer indices and a boolean mask. In our PT converter, Relay `adv_index` op is used to convert `aten::index`, but it doesn't correctly recognize indexing by mask - it will treat True / False as indices 0 / 1. So our converter generates an invalid model if the PT model uses such indexing, see https://github.com/apache/tvm/issues/12292 and https://github.com/apache/tvm/issues/13296.

I fixed this issue by explicitly tuning the mask into integer indices using `argwhere`, which will also take care of the inherent dynamism in indexing by mask.

cc @comaniac @junrushao @shingjan @yelite

### DIFF
--- a/python/tvm/relay/frontend/pytorch.py
+++ b/python/tvm/relay/frontend/pytorch.py
@@ -2323,8 +2323,15 @@ class PyTorchOpConverter:
 
     def index(self, inputs, input_types):
         data = inputs[0]
-        indices = inputs[1]
-        return _op.adv_index([data] + indices)
+        indices_list = []
+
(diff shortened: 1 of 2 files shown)


comaniac (approved):
LGTM

comaniac:
Thanks @masahi

PR #13308, opened 2022-11-07
[TVMScript] Hide trailing return type if None

Lunderberg:
Because the majority of TIR PrimFuncs operate on buffers, write their outputs to an output parameter, and do not return a value, the `-> None` in the function signature becomes visual noise. This commit removes printing of the return type in cases where the PrimFunc has no return value.


tkonolige (approved):

junrushao (approved):
LGTM! Yeah in most of the cases the return type is None

shingjan (approved):

PR #13309, opened 2022-11-07
[TVMC] Global pass context for compile and tune

lhutton1:
Comes as a followup from conversations in #13216. By making the pass context a global value for both `compile` and `tune` commands, we can ensure the pass context is exactly as the user expected and also test components such as `convert_graph_layout` under a pass context suitable for testing (e.g. add instruments). With this change, it becomes the users responsibility to ensure the PassContext they select is suitable for the passes that will be run. By default, `opt_level` remains as 3 so current workflows that do not alter the pass context from the command line / TVMC Python API should not be affected.


lhutton1:
also cc @leandron @ekalda

ekalda (approved):
Looks good to me, thanks @lhutton1 for cleaning this up!

leandron:
Thanks @ekalda @lhutton1

PR #13311, opened 2022-11-07
[BugFix][Driver] Correctly propogate simple-mode flag in LowerSchedule

StrongerXi:
Currently one version of `tvm::LowerSchedule` doesn't pass along the input `simple_mode` flag, which causes it to default back to `false`. This commit fixes it by passing along the input flag.

### DIFF
--- a/src/driver/driver_api.cc
+++ b/src/driver/driver_api.cc
@@ -54,9 +54,6 @@ TVM_REGISTER_PASS_CONFIG_OPTION("tir.use_async_copy", Bool);
 TVM_REGISTER_PASS_CONFIG_OPTION("tir.merge_async_commit_queue_scope", Bool);
 TVM_REGISTER_PASS_CONFIG_OPTION("tir.instrument_lwp", Bool);
 
-using runtime::PackedFunc;
-using runtime::TVMArgs;
-using runtime::TVMRetValue;
 using tvm::Array;
 using tvm::transform::Pass;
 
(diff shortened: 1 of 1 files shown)


src/driver/driver_api.cc
```diff
@@ -54,9 +54,6 @@ TVM_REGISTER_PASS_CONFIG_OPTION("tir.use_async_copy", Bool);
 TVM_REGISTER_PASS_CONFIG_OPTION("tir.merge_async_commit_queue_scope", Bool);
 TVM_REGISTER_PASS_CONFIG_OPTION("tir.instrument_lwp", Bool);
 
-using runtime::PackedFunc;
-using runtime::TVMArgs;
-using runtime::TVMRetValue;
```
StrongerXi, line 59: I deleted these as well because they are not used at all in this file.

shingjan (approved):
Thanks for sending this! Would it be possible that we can add some tests with simple_mode set to true/false? Otherwise LGTM.

StrongerXi:
> Thanks for sending this! Would it be possible that we can add some tests with simple_mode set to true/false? Otherwise LGTM.

Thanks. I did look into existing tests for `LowerSchedule` or `tvm.lower`, unfortunately none of them really tests `simple_mode`, which seems to be a convenience flag for debugging/testing purposes.

I could write a simple `tvm.lower(..., simple_mode=...)` unittest, but it actually doesn't go through this problematic version of `LowerSchedule`, which is only used [here](https://github.com/apache/tvm/blob/main/src/relay/backend/te_compiler.cc#L447-L448) and [here](https://github.com/apache/tvm/blob/main/src/relay/backend/te_compiler_cache.cc#L852-L853), and they don't even use the `simple_mode` flag.

I feel like the testing effort here isn't well worth it, but let me know what you think:).

vinx13 (approved):

junrushao (approved):
LGTM

PR #13316, opened 2022-11-07
[ci] Remove Jenkinsfile for migration to platform-specific jobs

driazati:
This is a follow on to https://github.com/tlc-pack/ci/pull/58 and #13300. This removes the old `Jenkinsfile` and makes the new job statuses required for PRs to merge.


areusch (approved):
just thinking that we should place these in a single location or update our "how to add a new CI image" guide too

driazati:
> just thinking that we should place these in a single location or update our "how to add a new CI image" guide too

I wanted to make the CI images read from a text file at runtime instead of hardcoded in the jenkinsfile, that way any change of the image tags will actually be tested, we can fix up the docs along with that

PR #13318, opened 2022-11-08
[Hexagon] [runtime] Move lock/unlock to HexagonHtp temporarily

janetsc:
Temporarily moves lock/unlock for HTP to the main thread instead of its own thread, until we are fully pipelined.

cc: @csullivan


csullivan (approved):
🔥 Many thanks @janetsc for this ultra fast (temporary) change and the suggestion to move to dispatching ASAP. Let's do it.

PR #13320, opened 2022-11-08
Fixup libtorch backend build

leiwen83:
Add clang-format disable for header to prevent reorder. Torch header file need to be put at the end since torch's dlpack is a little different with tvm's.

cc @t-vi @masahi

Signed-off-by: Lei Wen <wenlei03@qiyi.com>


junrushao (approved):
LGTM

shingjan (approved):
LGTM. Could there be more context on the dlpack mismatch between PyTorch and TVM. Thanks!

PR #13326, opened 2022-11-08
[TIR][Bugfix] Fix AXIS_SEPARATORS in tir.Schedule.transform_layout

Lunderberg:
Preivously, the block SREF reuse only included a single step of changes, and would have an incorrect mapping if multiple sequential changes to the TIR block occurred.  This could happen if a `BufferStore` was updated, followed by replacement of `Block` iter vars/values.  This commit tracks the Block replacements across each usage, to ensure the SREF instances remain valid.


abhikran-quic:
Thank you @Lunderberg for this change!

vinx13 (approved):

PR #13332, opened 2022-11-09
[FQ2I] Add cast back to input data type after AvgPool2d

ibsidorenko:
**This commit fixes the following issue:**
For the sequence of `qnn.dequantize -> avg_pool2d -> conv2d -> qnn.quantize` FQ2I pass inserts qnn.requantize (or cast) to int32 unconditionally before AvgPool2d. As a result fake quantized `qnn.conv2d` gets input as int32 dtype, but it is forbidden for `qnn.conv2d` (supports only uint8/int8/int16).

**This commit adds the following:**
Add cast back to input data type after AvgPool2d. This preserve input dtype == output dtype.

### DIFF
--- a/python/tvm/relay/transform/fake_quantization_to_integer.py
+++ b/python/tvm/relay/transform/fake_quantization_to_integer.py
@@ -141,6 +141,7 @@ def avgpool2d(expr, type_map):
     arg = expr.args[0]
     t = type_map[arg]
     out_t = type_map[expr]
+    # Cast (or requantize) to int32.
     if not (
         approx_equal(t.scale, out_t.scale)
         and approx_equal(t.zero_point, out_t.zero_point)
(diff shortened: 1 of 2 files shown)


python/tvm/relay/transform/fake_quantization_to_integer.py
```diff
@@ -158,7 +159,10 @@ def avgpool2d(expr, type_map):
     else:
         arg = relay.op.cast(arg, "int32")
     out = relay.op.nn.avg_pool2d(arg, **expr.attrs)
-    return [out, TensorAffineType(out_t.scale, out_t.zero_point, "int32", out_t.axis)]
+    # Cast back to input dtype to preserve input dtype == output dtype.
+    out = relay.op.clip(out, a_min=np.iinfo(t.dtype).min, a_max=np.iinfo(t.dtype).max)
+    out = relay.op.cast(out, t.dtype)
```
apeskov, line 164: Why do you need to preserve `input dtype == output dtype`? As I understood it should match `out_t.dtype`.

I guess it should be conditional injection of cast op to `out_t.dtype` if it is not a `int32`.

apeskov (changes requested):

python/tvm/relay/transform/fake_quantization_to_integer.py
ibsidorenko: Make sense. Changed on `out_t.dtype` + if condition.

masahi (approved):

PR #13335, opened 2022-11-09
[ci] Split out C++ unittests

driazati:
This makes C++ unittests follow the normal flow of build -> upload
artifacts -> download and run tests. To simplify the changes there is a
new utility for interacting with S3.


ci/jenkins/Build.groovy.j2
```diff
@@ -208,7 +231,9 @@ stage('Build') {
       script: "${docker_run} ${ci_hexagon} ./tests/scripts/task_config_build_hexagon.sh build",
       label: 'Create Hexagon cmake config',
     )
-    make(ci_hexagon, 'build', '-j2')
+    cmake_build(ci_hexagon, 'build', '-j2')
+    make_standalone_crt(ci_hexagon, 'build')
```
alanmacd, line 235: We would want to remove make_standalone_crt() here for the hexagon build.

cc: @mehrdadh

ci/jenkins/Test.groovy.j2
```diff
@@ -14,11 +21,19 @@
...
 ) %}
   {% if shard_index == 1 %}
-  {{ m.download_artifacts(tag='gpu2', filenames=tvm_multilib) }}
-  cpp_unittest(ci_gpu)
+  {{ m.download_artifacts(tag='gpu2', filenames=tvm_multilib + crttest + standalone_crt) }}
+  sh "${docker_run} --no-gpu ${ci_gpu} ./tests/scripts/task_config_build_gpu_other.sh build"
+  // These require a GPU to finish the build (i.e. CUDA needs to be load-able)
+  make_standalone_crt(ci_gpu, 'build')
+  // make_cpp_tests(ci_gpu, 'build')
+  // cpp_unittest(ci_gpu)
```
driazati, line 29: These aren't run at all right now and are currently broken, so this leaves them commented out. The previous code uploaded the same binaries for both `gpu` and `gpu2` since there was a fixed reference to the `build` directory. This resolves the issue which causes this to fail but doesn't fix the build failure itself, which we should leave for a follow up

tests/scripts/task_microtvm_cpp_tests.sh
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env bash
```
mehrdadh, line 1: is this used anywhere?

tests/scripts/task_cpp_unittest.sh
```diff
@@ -28,38 +28,15 @@ else
...
-    --build-dir "${BUILD_DIR}"
-
-# crttest requries USE_MICRO to be enabled.
-if grep -Fq "USE_MICRO ON" ${BUILD_DIR}/TVMBuildOptions.txt; then
-  pushd "${BUILD_DIR}"
-  ninja crttest
-  popd
-fi
-
 pushd "${BUILD_DIR}"
```
mehrdadh, line 40: I'm not sure what happened here. shouldn't `ctest` also move to wherever `crttest` was built?

tests/scripts/task_microtvm_cpp_tests.sh
driazati: ah forgot to add calls to these in the relevant jobs after their C++ testing, fixed

tests/scripts/task_cpp_unittest.sh
driazati: The build stuff should all be moved into the corresponding build stage and downloaded in the relevant test step to actually be run, so this removes all the building and just runs ctest (the other stuff behind the flags is also moved to the other test script below)

mehrdadh:
@driazati can you push your branch to a TVM branch to test the jenkins changes?

driazati:
Certain GitHub forks are configured in Jenkins to be trusted (e.g. those for people on the oss team), so the extra step to push to apache/tvm isn't necessary. You can see the info here: https://ci.tlcpack.ai/job/tvm/indexing/events, when a job comes from an untrusted source it'll say something like

```
Checking pull request [#13335](https://github.com/apache/tvm/pull/13335)
    (not from a trusted source)
      ‘Jenkinsfile’ found
```

mehrdadh (approved):
LGTM, thanks for working on this!!

PR #13336, opened 2022-11-09
[Hexagon] Make allocate_hexagon_array a hexagon contrib API

csullivan:
Move this API to the hexagon contrib module so that it can be accessed by tests outside of TVM source control.


csullivan:
@tvm-bot rerun

masahi (approved):

PR #13339, opened 2022-11-09
[Minor][Testing] Consolidate IRs into corresponding functions

junrushao:
We moved most of the IR definition into the testing methods correspondingly.


Hzfengsy (approved):
LGTM

PR #13341, opened 2022-11-10
[TOPI] Fix conv2d transpose for small channel

masahi:
Fixes the issue reported in https://discuss.tvm.apache.org/t/model-compilation-fails-on-gpu-target/13866

For certain `conv2d_transpose` workloads, we get funny allocations like
```
 allocate data_pad[float32 * ((floordiv(((blockIdx.y*2) + 1), 3)*4608) + 576)], storage_scope = shared
 allocate p1.shared[float32 * ((floordiv(((blockIdx.y*2) + 1), 3)*12288) + 1536)], storage_scope = shared
```

This is apparently due to trying to split a small channel by a large factor. 

cc @vinx13


junrushao (approved):
LGTM! Thanks for the fix!

PR #13275, opened 2022-11-03
[Relay] Enhancement for fold_scale_axis and simplify_expr

leiwen83:
* add(%1, %1) convert to multiply(%1, 2f)

* enhance fold_scale_axis to fold multiply(%1, 2f) into conv

Signed-off-by: Lei Wen <wenlei03@qiyi.com>

PR #13343, opened 2022-11-10
[Relay] Refactor constant folding over expr into a utility function

StrongerXi:
Refactor duplicated and low-level logic for applying constant-folding over an `Expr` into a utility function.

### DIFF
--- a/src/relay/transforms/fold_constant.cc
+++ b/src/relay/transforms/fold_constant.cc
@@ -418,14 +418,6 @@ class ConstantFolder : public MixedModeMutator {
 
 TVM_REGISTER_GLOBAL("relay.analysis.check_constant").set_body_typed(IsComplexConstant);
 
-/*!
- * \brief Returns \p expr with any constants expressions evaluated and let-bound constants
- * inlined. Returns \p expr unchanged if no change.
- *
- * CAUTION: The importers rely on this function returning \p expr unchanged to preserve sharing
- * from their p.o.v. Furthermore, this function can be called before conversion to ANF so
- * we must avoid all recursion.
- */
 Expr FoldConstantExpr(const Expr& expr, const IRModule& mod, bool fold_qnn) {
   VLOG_CONTEXT << "FoldConstantExpr";
   VLOG(1) << "folding:" << std::endl << PrettyPrint(expr);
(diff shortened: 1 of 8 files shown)


StrongerXi:
Since it's just a refactor, CI should suffice for testing.

echuraev (approved):
LGTM. Thanks

src/relay/transforms/pattern_utils.h
```diff
@@ -120,6 +120,21 @@ inline Expr InferType(const Expr& expr) {
...
+ *
+ * \param expr The IR expression
+ * \return resulting expr
+ */
+inline Expr FoldConstantExpr(const Expr& expr) {
+  IRModule const_mod = IRModule::FromExpr(expr);
+  const_mod = transform::FoldConstant()(const_mod);
+  GlobalVar const_main = const_mod->GetGlobalVar("main");
+  return Downcast<Function>(const_mod->functions[const_main])->body;
+}
```
masahi, line 136: There is already https://github.com/apache/tvm/blob/78142ad2e4ce2c5ca096f7e80f4f4f62ca9c2bd5/src/relay/transforms/fold_constant.cc#L429. You can probably just use that.

And it's odd to add this function in `pattern_utils.h`.

StrongerXi: > You can probably just use that.
Thanks, I did see that one, but (1). currently it doesn't seem to be declared in any header, and (2). the extra `mod` argument is not ergonomic for our use case here.

> And it's odd to add this function in pattern_utils.h.
I agree, I did this for consistency because we had [InferType](https://github.com/apache/tvm/blob/main/src/relay/transforms/pattern_utils.h#L113-L121) right above from #13275.

I think they should probably both be moved to either `pass_utils.h` or a new `transform_utils.h`.

Let me know what you think, I'll update the PR accordingly:).

masahi: Okay how about this:

* Introduce `fold_constant.h` and declare `FoldConstantExpr(const Expr& expr)` and `FoldConstantExpr(const Expr& expr, const IRModule& mod, bool fold_qnn)`.
* Implement the former one in terms of the latter.
* Replace all duplicated definition of `FoldConstantExpr` with these ones.

StrongerXi: I like that, done!

Caught another duplication in `relay/quantize/realize.cc`. Let me know if it needs more tweaking.

src/relay/transforms/fold_constant.cc
```diff
@@ -434,11 +426,20 @@ Expr FoldConstantExpr(const Expr& expr, const IRModule& mod, bool fold_qnn) {
...
+Expr FoldConstantExpr(const Expr& expr, bool fold_qnn) {
+  auto mod = IRModule::FromExpr(expr);
+  return FoldConstantExpr(expr, mod, fold_qnn);
+}
+
+TVM_REGISTER_GLOBAL("relay._transform.FoldConstantExpr").set_body_typed(
+    [](const Expr& expr, const IRModule& mod, bool fold_qnn) {
+      return FoldConstantExpr(expr, mod, fold_qnn);
+    }
+);
```
StrongerXi, line 438: Needed to resolve the name overloading. I decided to keep the `FoldConstantExpr` names.

StrongerXi:
Make linter happy.

StrongerXi:
Fix build failure in `contrib`, which wasn't triggered locally from default build.

masahi (approved):

masahi:
1.

PR #13344, opened 2022-11-10
[TIR][Schedule] Make consistent implementation for GetProducers() & GetConsumers()

wrongtest-intellif:
Currently there are two versions of `GetConsumers()` and `GetProducers()` implementation. Make them consistent to avoid possible bug when there are WAR dependencies.


junrushao (approved):
LGTM! I didn't realize there is such duplication :-)

PR #13346, opened 2022-11-10
[MetaSchedule] Unannotate `schedule_rule` if corresponding schedule func is not found

masahi:
Currently, if there is a block annotated with `schedule_rule`, only the corresponding custom rule is allowed to be applied to the block. This led to the breakage of auto tensorization as discussed in https://github.com/apache/tvm/pull/13195#discussion_r1018952166.

The fix is to unannotate `schedule_rule` if the corresponding schedule func is not found, to let other schedule rules process this block.

cc @junrushao @zxybazh


junrushao:
https://github.com/apache/tvm/pull/13195#discussion_r1019469214

zxybazh (approved):

PR #13347, opened 2022-11-10
[TIR] Fix extern_primfunc buffer order bug 

### LINKED ISSUE #13330: [Bug] DomainTouchedAccessMap does not preserve buffer order
When using `te.extern_primfunc`, we need to ensure the tensors we pass in are the same size as the buffers we are loading them into. This is done by `python/tvm/te/operation.py`, which has [the following code](https://github.com/apache/tvm/blob/0e395c389ccd173cf6c1f254b47a81e715762626/python/tvm/te/operation.py#L364):

```python
def extern_primfunc(input_tensors: List[_tensor.Tensor], primfunc: tvm.tir.PrimFunc, **kwargs):
    access_map = {
        k: tuple(v) for k, v in tvm.arith._ffi_api.DomainTouchedAccessMap(primfunc).items()
    }
    in_buffers = [buf for buf, access in access_map.items() if len(access[0])]
    input_buffers = in_buffers
    # [some lines omitted for brevity]
    for obuf in out_buffers:

    assert len(input_buffers) == len(input_tensors), (
        "The number of provided input input_tensors does not match the number of ",
        "input buffers in the primfunc",
    )
    for tensor, buffer in zip(input_tensors, input_buffers):
        assert len(tensor.shape) == len(buffer.shape)
        for d1, d2 in zip(tensor.shape, buffer.shape):
            assert d1 == d2, (
                "The input input_tensors provided do not match the input buffers in the ",
                "primfunc. Please check that the order of input te.Input_Tensors and the ",
                "order of the primfunc variables in the params list agree.",
            )
```

Specifically, by using `zip` this code requires that the order of tensors in the list `input_tensors` match the order of buffers in the **dictionary** `input_buffers`. It's easy to see how this might pose a problem.

However, the issue is slightly more nuanced. For Python 3.6 and above, [dictionaries preserve order](https://www.blog.pythonlibrary.org/2018/02/27/python-3-7-dictionaries-now-ordered). Furthermore, [there are unittests](https://github.com/apache/tvm/blob/main/tests/python/unittest/test_tir_te_extern_primfunc.py) for the code in question - why doesn't it always break?

While `extern_primfunc` probably shouldn't use dictionaries to store buffers since order is important, the issue isn't here. Instead, the problem is that `tvm.arith._ffi_api.DomainTouchedAccessMap` returns a `tvm.ir.container.Map`. When used with a small number of elements (e.g. 3), `tvm.ir.container.Map` happens to preserve order, but it does not guarantee this (which is why this bug is not detected by the unit tests). However, when a larger number (e.g. 5) elements are present in the map, an error is produced:

```
E           tvm._ffi.base.TVMError: Traceback (most recent call last):
E             22: TVMFuncCall
[some lines omitted for brevity]
E             File "/home/guberti/tvm/python/tvm/te/operation.py", line 440, in extern_primfunc
E               assert len(tensor.shape) == len(buffer.shape)
E           TVMError: AssertionError
```

### Suggested fix

As far as I can tell, `extern_primfunc` needs to be given the buffers in the same order as tensors. Thus, `DomainTouchedAccessMap` should be changed to not return a map, and instead return an list of tuples. I'm blocked by this bug, so I'm just gonna fix it.

* tir:arith

CC @areusch, @Lunderberg

guberti:
This fixes #13330, which was blocking my work to write TIR schedules for microTVM.

I originally thought I'd have to change the function signature of `DomainTouchedAccessMap`, but I couldn't think of a way to do that cleanly. Instead, I changed `extern_primfunc` to use `primfunc.params` to create the buffer lists in the right order.

#13330 should have been caught by `test_tir_te_extern_primfunc.py`, but one of that test's helper functions had the same bug as `extern_primfunc`. I've thus modified `test_tir_te_extern_primfunc.py` to instantiate the input tensors a different way, allowing it to catch regressions of this issue.

Would love a look from @csullivan!

### DIFF
--- a/python/tvm/te/operation.py
+++ b/python/tvm/te/operation.py
@@ -398,11 +398,12 @@ def extern_primfunc(input_tensors: List[_tensor.Tensor], primfunc: tvm.tir.PrimF
 
         C = te.extern_primfunc([A, B], func)
     """
-    access_map = {
-        k: tuple(v) for k, v in tvm.arith._ffi_api.DomainTouchedAccessMap(primfunc).items()
-    }
-    in_buffers = [buf for buf, access in access_map.items() if len(access[0])]
-    out_buffers = [buf for buf, access in access_map.items() if len(access[1])]
+
+    # dt_access_map and primfunc.buffer_map are unordered, so use order from primfunc.params
+    dt_access_map = tvm.arith._ffi_api.DomainTouchedAccessMap(primfunc)
+    ordered_buffers = [primfunc.buffer_map[param] for param in primfunc.params]
+    in_buffers = [buf for buf in ordered_buffers if len(dt_access_map[buf][0])]
+    out_buffers = [buf for buf in ordered_buffers if len(dt_access_map[buf][1])]
     assert in_buffers, "PrimFunc has no input buffers"
     assert out_buffers, "PrimFunc has no output buffers"
 
--- a/tests/python/unittest/test_tir_te_extern_primfunc.py
+++ b/tests/python/unittest/test_tir_te_extern_primfunc.py
@@ -174,11 +174,11 @@ def verify_func_4(module):
 
 
 class TestPrimFuncs:
-    func, verify = tvm.testing.parameters(
-        [func_1, verify_func_1],
-        [func_2, verify_func_2],
-        [func_3, verify_func_3],
-        [func_4, verify_func_4],
+    func, params, verify = tvm.testing.parameters(
+        [func_1, ("A"), verify_func_1],
+        [func_2, ("C", "D"), verify_func_2],
+        [func_3, ("C", "A", "D", "E"), verify_func_3],
+        [func_4, ("C", "A", "D", "E"), verify_func_4],
     )
 
     def test_primfunc_call(self, func, verify):
@@ -186,11 +186,12 @@ class TestPrimFuncs:
         func = tvm.build(func, target=target)
         verify(func)
 
-    def test_te_extern_call(self, func, verify):
+    def test_te_extern_call(self, func, params, verify):
         ir_mod = tvm.IRModule.from_expr(func.with_attr("global_symbol", "main"))
         prim_func = ir_mod["main"]
 
-        input_tensors = create_input_tensors_for_primfunc(prim_func)
+        buf_name_map = {buf.name: buf for buf in func.buffer_map.values()}
+        input_tensors = [te.placeholder(buf_name_map[name].shape) for name in params]
         output = te.extern_primfunc(input_tensors, prim_func)
         rt_prim_func = te.create_prim_func(tensors_from_extern_op(output, prim_func))
         tvm.ir.assert_structural_equal(tvm.lower(prim_func), tvm.lower(rt_prim_func))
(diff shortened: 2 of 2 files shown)


guberti:
I'd love to add a unit test to prevent regressions - ideally in `tests/python/unittest/test_tir_te_extern_primfunc.py`. However, when running on `main` I can't get that unit test to pass. I get:
```
============================================== ERRORS ===============================================
_______________ ERROR collecting tests/python/unittest/test_tir_te_extern_primfunc.py _______________
tests/python/unittest/test_tir_te_extern_primfunc.py:36: in <module>
    def func_1(A: T.Buffer[(16,), "float32"], C: T.Buffer[(1,), "float32"]):
python/tvm/script/tir/prim_func.py:40: in prim_func
    result = from_source(input_func)
python/tvm/script/parser.py:1368: in from_source
    result = to_ast(input_func, TVMDiagnosticCtx(), parser)
../.local/lib/python3.8/site-packages/synr/compiler.py:731: in to_ast
    prog = compiler.compile_module(program_ast)
../.local/lib/python3.8/site-packages/synr/compiler.py:100: in compile_module
    new_func = self.compile_def(stmt)
../.local/lib/python3.8/site-packages/synr/compiler.py:174: in compile_def
    params = self.compile_args_to_params(args)
../.local/lib/python3.8/site-packages/synr/compiler.py:164: in compile_args_to_params
    ty = self.compile_type(arg.annotation)
../.local/lib/python3.8/site-packages/synr/compiler.py:622: in compile_type
    return self._expr2type(self.compile_expr(ty))
../.local/lib/python3.8/site-packages/synr/compiler.py:605: in _expr2type
    assert isinstance(
E   AssertionError: Expected subscript call to have lhs of Var, but it is <class 'synr.ast.Attr'>
```
Perhaps I'm building TVM wrong?

csullivan (commented):
Thanks @guberti! Yes, I think you'll want to make the same change in the test as well
```diff
diff --git a/tests/python/unittest/test_tir_te_extern_primfunc.py b/tests/python/unittest/test_tir_te_extern_primfunc.py
index 2675214562..a3c6645be7 100644
--- a/tests/python/unittest/test_tir_te_extern_primfunc.py
+++ b/tests/python/unittest/test_tir_te_extern_primfunc.py
@@ -224,8 +224,9 @@ def tensors_from_extern_op(extern, func):

 def create_input_tensors_for_primfunc(primfunc):
     access_map = {k: tuple(v) for k, v in _ffi_arith_api.DomainTouchedAccessMap(primfunc).items()}
-    in_buffers = [buf for buf, access in access_map.items() if len(access[0])]
-    out_buffers = [buf for buf, access in access_map.items() if len(access[1])]
+    ordered_buffers = [primfunc.buffer_map[param] for param in primfunc.params]
+    in_buffers = [buf for buf in ordered_buffers if len(access_map[buf][0])]
+    out_buffers = [buf for buf in ordered_buffers if len(access_map[buf][1])]
     assert in_buffers, "PrimFunc has no input buffers"
     assert out_buffers, "PrimFunc has no output buffers"
```
Please also add a failing (on HEAD) test that reproduces the issue so we can prevent the regression in the future. I admit the need for more tests in the header of that test. Unfortunately I have not had time to return to it recently, but I am happy to hear you are using it and appreciate the contribution!

guberti:
@csullivan I fixed `test_tir_te_extern_primfunc.py` so that it fails on `HEAD` (and will catch regressions of this issue). I would have done that sooner, but I was blocked for a bit by #13360. Should be good to merge now!

junrushao:
I'm happy to merge it in once it's confirmed that the TVMScript issue has been fixed.

guberti:
@junrushao I've confirmed the TVMScript issue (#13360) is fixed, can we merge this?

junrushao:
Thanks @guberti for confirming :-) Merging it in

junrushao (approved):
LGTM!

Lunderberg:
LGTM, too!  I like the improvement of using primfunc's parameter list for making the tensor to buffer mapping, since the parameter list is what actually defines the order of parameters.

PR #13349, opened 2022-11-10
[TIR][Hexagon] Add vtcm memory capacity verification for Hexagon target

Icemist:
The main items that have been added are:

tvm.tir.analysis.calculate_allocated_bytes() 
> to calculate allocated memory per memory scope

tir.transform.VerifyVTCMLimit(limit)

> to verify if the size of the allocated vtcm memory satisfies the limit

tvm.target.hexagon().vtcm_capacity 

> attribute to pass the limit

tir.vtcm_capacity 

> context configuration attribute to pass the limit alternatively

### DIFF
--- a/python/tvm/autotvm/measure/measure_methods.py
+++ b/python/tvm/autotvm/measure/measure_methods.py
@@ -330,7 +330,7 @@ class RPCRunner(Runner):
             )
 
     def get_build_kwargs(self):
-        kwargs = {}
+        kwargs = {"checks": {}}
         if (
             "cuda" in self.task.target.keys
             or "opencl" in self.task.target.keys
@@ -340,13 +340,15 @@ class RPCRunner(Runner):
             remote = request_remote(self.key, self.host, self.port)
             dev = remote.device(str(self.task.target), 0)
             max_dims = dev.max_thread_dimensions
-            kwargs["check_gpu"] = {
+            kwargs["checks"]["gpu"] = {
                 "max_shared_memory_per_block": dev.max_shared_memory_per_block,
                 "max_threads_per_block": dev.max_threads_per_block,
                 "max_thread_x": max_dims[0],
                 "max_thread_y": max_dims[1],
                 "max_thread_z": max_dims[2],
             }
+        if "hexagon" in self.task.target.keys:
+            kwargs["checks"]["hexagon"] = {"vtcm_capacity": self.task.target.vtcm_capacity}
 
         return kwargs
 
@@ -493,11 +495,11 @@ class LocalRunner(RPCRunner):
         return server, tracker
 
 
-def _build_func_common(measure_input, runtime=None, check_gpu=None, build_option=None):
+def _build_func_common(measure_input, runtime=None, checks=None, build_option=None):
     """Common part for building a configuration"""
     target, task, config = measure_input
     target, task.target_host = Target.canon_target_and_host(target, task.target_host)
-
+    checks = checks or {}
     with target:
         s, args = task.instantiate(config)
 
@@ -526,8 +528,10 @@ def _build_func_common(measure_input, runtime=None, check_gpu=None, build_option
                 current_add_lower_pass = list(current_config["tir.add_lower_pass"])
             else:
                 current_add_lower_pass = []
-            if check_gpu:
-                current_add_lower_pass.append((2, gpu_verify_pass(**check_gpu)))
+            if checks.get("gpu"):
+                current_add_lower_pass.append((2, gpu_verify_pass(**checks.get("gpu"))))
+            if checks.get("hexagon"):
+                current_add_lower_pass.append((2, vtcm_verify_pass(**checks.get("hexagon"))))
             current_config["tir.add_lower_pass"] = current_add_lower_pass
 
             with tvm.ir.transform.PassContext(
@@ -869,6 +873,23 @@ def gpu_verify_pass(**kwargs):
         valid = tvm.tir.analysis.verify_gpu_code(f, kwargs)
         if not valid:
             raise InstantiationError("Skipped because of invalid gpu kernel")
+        return f
+
+    return tvm.tir.transform.prim_func_pass(verify_pass, opt_level=0)
+
+
+def vtcm_verify_pass(**kwargs):
+    """Verify the validity of a hexagon kernel.
+    This pass will check vtcm memory usage.
+    """
+
+    def verify_pass(f, *_):
+        sizes = tvm.tir.analysis.calculate_allocated_bytes(f)
+        vtcm_capacity = kwargs.get("vtcm_capacity", 0)
+        vtcm_allocated = sizes.get("global.vtcm", 0)
+        if 0 < vtcm_capacity < vtcm_allocated:
+            raise InstantiationError("Skipped because of invalid vtcm memory usage limit")
+
         return f
 
     return tvm.tir.transform.prim_func_pass(verify_pass, opt_level=0)
--- a/python/tvm/target/target.py
+++ b/python/tvm/target/target.py
@@ -183,6 +183,10 @@ class Target(Object):
         return int(self.attrs.get("max_function_args", -1))
 
     @property
+    def vtcm_capacity(self):
+        return int(self.attrs.get("vtcm-capacity", 0))
+
+    @property
     def device_name(self):
         return str(self.attrs.get("device", ""))
 
@@ -642,6 +646,8 @@ def hexagon(cpu_ver="v66", **kwargs):
         Whether to use IEEE HVX instructions
     num_cores : int (default: 4)
         The number of HVX threads. This attribute is required by meta scheduler.
+    vtcm_capacity: int (default: 0)
+        Hexagon VTCM capacity limitation. If the value is 0, the capacity is treated as unbounded.
 
     Note: Floating point support in HVX requires LLVM 14+.
     """
@@ -675,6 +681,7 @@ def hexagon(cpu_ver="v66", **kwargs):
         "llvm_options": None,
         "use_qfloat": arch_version >= 68,
         "use_ieee_fp": False,
+        "vtcm_capacity": 0,
     }
     config.update(kwargs)
 
@@ -748,6 +755,7 @@ def hexagon(cpu_ver="v66", **kwargs):
 
     num_cores = config["num_cores"] if "num_cores" in kwargs else 4
     args_list.append("--num-cores=%d" % num_cores)
+    args_list.append("--vtcm-capacity=%d" % config["vtcm_capacity"])
 
     return Target(" ".join(["hexagon"] + args_list))
 
--- a/tests/python/unittest/test_tir_analysis_calculate_allocated_memory.py
+++ b/tests/python/unittest/test_tir_analysis_calculate_allocated_memory.py
@@ -1,0 +1,101 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+import pytest
+
+import tvm
+from tvm import tir
+from tvm.script import tir as T
+
+
+@T.prim_func
+def scale_by_two(a: T.Buffer[(128,), "int8"], c: T.Buffer[(128,), "int8"]):
+    for i in T.serial(128):
+        with T.block("C"):
+            c[i] = a[i] * T.int8(2)
+
+
+@T.prim_func
+def scale_by_two_three(a: T.Buffer[(128,), "int8"], c: T.Buffer[(128,), "int8"]):
+    B = T.alloc_buffer([128], dtype="int8", scope="global.vtcm")
+    for i in T.serial(128):
+        with T.block("B"):
+            B[i] = a[i] * T.int8(2)
+    for i in T.serial(128):
+        with T.block("C"):
+            c[i] = B[i] * T.int8(3)
+
+
+@pytest.mark.parametrize("primFunc,size", [(scale_by_two, 128), (scale_by_two_three, 256)])
+def test_scale_by(primFunc, size):
+    """Test calculate allocated bytes per scope"""
+    mod = tvm.IRModule.from_expr(primFunc.with_attr("global_symbol", "main"))
+    sch = tir.Schedule(mod, debug_mask="all")
+    block_c = sch.get_block("C")
+    (flat,) = sch.get_loops(block_c)
+    cache_block = sch.cache_read(block_c, 0, storage_scope="global.vtcm")
+    sch.compute_at(cache_block, flat)
+
+    mod = sch.mod
+    mod = tvm.tir.transform.ConvertBlocksToOpaque()(mod)
+    mod = tvm.tir.transform.LowerOpaqueBlock()(mod)
+    sizes = tvm.tir.analysis.calculate_allocated_bytes(mod["main"])
+    assert sizes.get("global.vtcm", 0) == size
+
+
+@T.prim_func
+def matmul_mix_scope(a: T.handle, b: T.handle, c: T.handle) -> None:
+    A = T.match_buffer(a, [128, 128], scope="global")
+    B = T.match_buffer(b, [128, 128], scope="global")
+    C = T.match_buffer(c, [128, 128], scope="global")
+    A_allocated = T.alloc_buffer([128, 128], dtype="float32", scope="global.texture")
+    B_allocated = T.alloc_buffer([128, 128], dtype="float32", scope="global.texture")
+    C_allocated = T.alloc_buffer([128, 128], dtype="float32", scope="global")
+
+    for i, j in T.grid(128, 128):
+        with T.block("A.allocated"):
+            A_allocated[i, j] = A[i, j]
+    for i, j in T.grid(128, 128):
+        with T.block("B.allocated"):
+            B_allocated[i, j] = B[i, j]
+
+    for i, j, k in T.grid(128, 128, 128):
+        with T.block("update"):
+            vi, vj, vk = T.axis.remap("SSR", [i, j, k])
+            with T.init():
+                C_allocated[vi, vj] = 0.0
+            C_allocated[vi, vj] = C[vi, vj] + A_allocated[vi, vk] * B_allocated[vj, vk]
+
+    for i, j in T.grid(128, 128):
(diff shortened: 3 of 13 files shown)


Icemist:
CC @masahi @csullivan

TejashShah:
cc @kparzysz-quic @janetsc

include/tvm/target/utils.h
```diff
@@ -0,0 +1,46 @@
...
+#define TVM_TARGET_UTILS_H_
+
+#include <tvm/ir/expr.h>
+#include <tvm/runtime/container/optional.h>
+#include <tvm/target/target.h>
+
+namespace tvm {
+
+/*! \brief Extract attribute from a target. */
+inline Integer Extract(const Target& target, const char* name) {
```
echuraev, line 35: The extracted attributes might have different types, not only `Integer`. Probably we can do this function as a template? Also, have some doubts about do we really need this helper function?

include/tvm/target/utils.h
```diff
@@ -0,0 +1,46 @@
...
+  ICHECK(target.defined());
+  if (Optional<Integer> v = target->GetAttr<Integer>(name)) {
+    return v.value();
+  }
+  LOG(FATAL) << "AttributedError: \"" << name << "\" is not defined in the target";
+  throw;
```
echuraev, line 41: As far as I remember, `LOG(FATAL)` should throw an exception. So, we don't need to add `throw` after that. If there is a compilation warning about return value, probably we could do something like that:
```suggestion
  Optional<Integer> v = target->GetAttr<Integer>(name);
  ICHECK(v != nullptr) << "AttributedError: \"" << name << "\" is not defined in the target";
  return v.value();
```

python/tvm/driver/build_module.py
```diff
@@ -221,29 +221,30 @@ def build(
...
-        input_mod = lower(inputs)
-    elif not isinstance(inputs, (dict, container.Map)):
-        raise ValueError(
-            f"Inputs must be te.Schedule, IRModule, PrimFunc, "
-            f"or dict of target to IRModule, "
-            f"but got {type(inputs)}."
-        )
+    target = Target.current() if target is None else target
+    target = target if target else "llvm"
+    with target:
```
echuraev, line 226: Sorry, could you please explain, why do we need this target context and these changes?

python/tvm/autotvm/measure/measure_methods.py
```diff
@@ -493,7 +495,9 @@ def set_task(self, task):
         return server, tracker
 
 
-def _build_func_common(measure_input, runtime=None, check_gpu=None, build_option=None):
+def _build_func_common(
+    measure_input, runtime=None, check_gpu=None, build_option=None, check_hexagon=None
```
echuraev, line 499: Probably we could pass all check arguments (such as `check_gpu` and `check_hexagon`) in a dictionary or list? In this case it won't be necessary to change function signature every time when we want to add new check

tests/python/contrib/test_hexagon/test_cache_read_write.py
```diff
@@ -222,5 +223,28 @@ def _visit(stmt):
...
+    """Test lowering with vtcm mem scope"""
+    mod = tvm.IRModule.from_expr(scale_by_two.with_attr("global_symbol", "main"))
+    sch = tir.Schedule(mod, debug_mask="all")
+    block_c = sch.get_block("C")
+    (flat,) = sch.get_loops(block_c)
+    outer, _, _, _ = sch.split(flat, factors=[8, 4, 2, 128])
+    cache_block = sch.cache_read(block_c, 0, storage_scope="global.vtcm")
+    sch.compute_at(cache_block, outer)
+
+    with tvm.transform.PassContext(config={"tir.vtcm_capacity": vtcm_capacity}):
```
janetsc, line 245: It would be interesting in the future if we could pass this capacity down to the device API for two reasons - to verify that it is possible, and to attempt to allocate only that size from QURT in HexagonVtcmPool

tests/python/unittest/test_tir_analysis_calculate_allocated_memory.py
```diff
@@ -0,0 +1,65 @@
...
+    for i, j in T.grid(128, 128):
+        with T.block("C"):
+            C[i, j] = C_[i, j]
+# fmt: on
+
+
+@pytest.mark.parametrize(
+    "scope,size", [("global", 131072), ("global.texture", 131072*2), ("global.texture-nhwc", 0)]
+)
+def test_allocates(scope, size):
```
janetsc, line 57: Can you add a test specifically to make sure the calculated VTCM size matches what we expect?  (To make sure it picks up "global.vtcm", as the LowerVtcmAlloc pass alters those nodes.)

include/tvm/target/utils.h
Icemist: This is not a new function, but a relocation of an existing one, to be used in other code blocks without duplication.

Icemist: deleted.

include/tvm/target/utils.h
Icemist: deleted.

python/tvm/autotvm/measure/measure_methods.py
Icemist: done.

src/meta_schedule/postproc/verify_vtcm_limit.cc
```diff
@@ -0,0 +1,102 @@
...
+    return true;
+  }
+
+  Postproc Clone() const {
+    ObjectPtr<VerifyVTCMLimitNode> n = make_object<VerifyVTCMLimitNode>(*this);
+    n->vtcm_capacity = this->vtcm_capacity;
+    return Postproc(n);
+  }
+
+  static constexpr const char* _type_key = "meta_schedule.VerifyVTCMLimit";
```
janetsc, line 88: Is this guaranteed to run after USMP, if enabled?  If not, we might be eliminating solutions that don't fit within VTCM before USMP (but do afterwards).

tests/python/unittest/test_tir_analysis_calculate_allocated_memory.py
Icemist: I have now updated the tests.
tests/python/contrib/test_hexagon/test_cache_read_write.py::test_vtcm_limit
tests/python/unittest/test_tir_analysis_calculate_allocated_memory.py::test_scale_by
tests/python/unittest/test_tir_analysis_calculate_allocated_memory.py::test_matmul_mix_scope

On your question: You mean tests that show how much memory was allocated through nd_mem_alloc_with_scope after lower, and compare this to what tir.analysis.calculate_allocated_bytes shows. Correct?

Our interest in the line - 
`let buffer_a_global.vtcm: Pointer(global.vtcm int8) = @tir.nd_mem_alloc_with_scope("global.vtcm", 1i64, @tir.tvm_stack_make_shape(1024, dtype=handle), dtype=handle)`
in the example:
```
@main = primfn(buffer_a_handle: handle, buffer_c_handle: handle) -> ()
  attr = {"global_symbol": "main", "tir.noalias": True}
  buffers = {buffer_a: Buffer(buffer_a_1: Pointer(global int8), int8, [8192], []),
             buffer_c: Buffer(buffer_c_1: Pointer(global int8), int8, [8192], [])}
  buffer_map = {buffer_a_handle: buffer_a, buffer_c_handle: buffer_c}
  preflattened_buffer_map = {buffer_c_handle: buffer_c_2: Buffer(buffer_c_1, int8, [8192], []), buffer_a_handle: buffer_a_2: Buffer(buffer_a_1, int8, [8192], [])} {
  let buffer_a_global.vtcm: Pointer(global.vtcm int8) = @tir.nd_mem_alloc_with_scope("global.vtcm", 1i64, @tir.tvm_stack_make_shape(1024, dtype=handle), dtype=handle)
  for (i_0: int32, 0, 8) {
    for (ax0: int32, 0, 1024) {
      buffer_a_global.vtcm_1: Buffer(buffer_a_global.vtcm, int8, [1024], [], scope="global.vtcm")[ax0] = buffer_a[((i_0*1024) + ax0)]
    }
    for (i_1: int32, 0, 4) {
      for (i_2: int32, 0, 2) {
        for (i_3: int32, 0, 128) {
          let cse_var_2: int32 = (i_1*256)
          let cse_var_1: int32 = (i_2*128)
          buffer_c[((((i_0*1024) + cse_var_2) + cse_var_1) + i_3)] = (buffer_a_global.vtcm_1[((cse_var_2 + cse_var_1) + i_3)]*2i8)
        }
      }
    }
  }
}
```

janetsc: Yes, that's what I meant, to make sure it works for VTCM with that lowering pass.  The new tests resolve this - thank you!

For the example you mention, I'm not sure I understand.  Is that the state after lowering, but before calculating VTCM usage?

src/meta_schedule/postproc/verify_vtcm_limit.cc
```diff
@@ -0,0 +1,101 @@
...
+
+  bool Apply(const tir::Schedule& sch) final {
+    IRModule mod = sch->mod();
+    for (const auto& kv : mod->functions) {
+      const GlobalVar& g_var = kv.first;
+      const BaseFunc& base_func = kv.second;
+      if (const auto* prim_func = base_func.as<tir::PrimFuncNode>()) {
+        IRModule lowered{nullptr};
+        try {
+          auto pass_list = Array<tvm::transform::Pass>();
```
csullivan, line 59: We'll need to include a set of passes here that reproduces the standard lowering in TIR including the new `VerifyVTCMLimit` in order to actually exclude bad samples in the search.

src/meta_schedule/postproc/verify_vtcm_limit.cc
csullivan: This will be running during a tuning session which will be prior to USMP (and other executor code responsible for model runtime)

janetsc: That means we might be discarding viable options.

csullivan: USMP is a whole program analysis, whereas the VerifyVTCMLimit pass is a single op/primfunc analysis. In principle to achieve what you are suggesting @janetsc we would want to update the MS preprocessor to be replaced with a preprocessor _generator_ which can generate a preprocessor to use based on the live VTCM as calculated by USMP. This should be quite doable in Relax. 

As a side note, with the introduction of whole program memory planning, the phase ordering of memory planning and tuning can become complex.

python/tvm/rpc/tracker.py
```diff
@@ -348,7 +348,7 @@ def close(self, conn):
         if "key" in conn._info:
             for value in conn.put_values:
                 _, _, _, key = value
-                rpc_key = key.split(":")[0]
+                rpc_key = key.rsplit(":", 1)[0]
```
echuraev, line 351: Could you please add a test on this change to highlight what is the fix here? Probably `tests/python/unittest/test_runtime_rpc.py` is the right place for such test. Also, maybe it is better to create a separate PR with this fix.

Icemist: Moved into https://github.com/apache/tvm/pull/13420

python/tvm/driver/build_module.py
Icemist: It was necessary to use with statement to be able to get the target object using Target::current().
Now I have changed this by passing the target as an optional parameter for lower function.

tests/python/unittest/test_tir_analysis_calculate_allocated_memory.py
Icemist: I meant comparing the memory counted with pass and how much memory is allocated with nd_mem_alloc_with_scope after lowering.
Something like(a sketch): 
 ```
   def calc(irmod):
        size = 0
        def _visit(stmt):
            nonlocal size
            if isinstance(stmt, tvm.tir.expr.Call) and isinstance(stmt.op, tvm.ir.op.Op) and stmt.op.name == "tir.nd_mem_alloc_with_scope":
                storage_scope, extents_size, call = stmt.args
                extents, = call.args
                dtype = call.dtype
                print("ATTRS", storage_scope, extents_size, extents, dtype)
                if storage_scope == "global.vtcm":
                    size += extents_size * dtype.bytes() #

        tvm.tir.stmt_functor.post_order_visit(irmod["main"].body, _visit)
        return size

    print(calc(lowered))
```

File: src/meta_schedule/postproc/verify_vtcm_limit.cc (hunk 8 above)
Comment (Icemist) [reply src/meta_schedule/postproc/verify_vtcm_limit.cc]: To move this RP forward. I've separated this part into another [branch](https://github.com/Deelvin/tvm/pull/new/avoronov/verify_vtcm_limit_postproc), let's look at it in more detail there.

File: include/tvm/driver/driver_api.h [hunk 10]
```diff
@@ -78,20 +78,22 @@ TVM_DLL transform::Sequential HostModulePassManager(IRModule mixed_mod, Target t
  * \brief Lower an IRModule (optimize with it with the pass list defined in CreatePassList)
  * \param mod The IRmodule to lower
  * \param simple_mode Disables the loop partition pass. Defaults to false.
+ * \param target The device target to provide additional characteristics. Defaults to not defined.
  * \return The result module.
  */
-TVM_DLL IRModule LowerModule(IRModule mod, bool simple_mode = false);
+TVM_DLL IRModule LowerModule(IRModule mod, bool simple_mode = false, Target target = {});
```
echuraev, line 84: I have some doubts about introducing the new parameter with target to many lowering functions. As far as I understood, the only one thing why we need to have this `target` parameter is for passing value of `vtcm_capacity` to the lowering functions.
From my point of view, we shouldn't modify this common API. As for me, the better way is to pass this option through `PassContext` (e.g. in the same way it works for `relay.ToMixedPrecision.keep_orig_output_dtype`).

@tqchen, @csullivan, @junrushao, @tkonolige, In file [test_vtcm.py](https://github.com/apache/tvm/pull/13349/files#diff-9380a067fbc7e640cb0257906fadaf5c400f07f46500d530af7c601b29855eb6R83-R95) @Icemist has presented possible ways for passing value of `vtcm_capacity`:
1. By passing target as a parameter to the `lower` function: 
    ```python
    _raises_exception(lambda: tvm.lower(sch.mod["main"], target=target)) == limited
    ``` 
2. By using `with` statement:
    ```python
    with target:
        assert (
            _raises_exception(lambda: tvm.lower(sch.mod["main"])) == limited
        ), "VTCM memory allocation limiter does not work correctly "
    ```
3. By passing value of `vtcm_capacity` to `PassContext`:
    ```python
    with tvm.transform.PassContext(config={"tir.vtcm_capacity": vtcm_capacity}):
        assert (
            _raises_exception(lambda: tvm.lower(sch.mod["main"])) == limited
        ), "VTCM memory allocation limiter does not work correctly "
    ```
    
Personally, I prefer the third option, because in this case, I believe, it is not necessary to modify the API of lowering functions. What do you think about that?

Icemist: In the case of a refusal to change the api and using only tir.vtcm_capacity as a transmitter - 

In tvm.build we can have this logic:
```python
vtcm_capacity = target.vtcm_capacity if target.kind.name == "hexagon" else 0
with tvm.transform.PassContext(config={"tir.vtcm_capacity": vtcm_capacity}):
	# code uses lowering
	
```
to leave the API as pleasant as possible:
```python
target = tvm.target.hexagon("v68", vtcm_capacity=128)
target = tvm.target.Target(target, host=target)
func = tvm.build(sch, [], target, name="func")
```

But also anticipate that it will be necessary to find and test/modify additional functions including C++ and python parts that use the lowering for the hexagon target.

python/tvm/target/target.py
```diff
@@ -642,6 +646,8 @@ def hexagon(cpu_ver="v66", **kwargs):
         Whether to use IEEE HVX instructions
     num_cores : int (default: 4)
         The number of HVX threads. This attribute is required by meta scheduler.
+    vtcm_capacity: int (default: 0)
+        Hexagon VTCM capacity limitation. Value of 0 disables this limitation.
```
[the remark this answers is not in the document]
Icemist: True, there is of course another way - 0 also limits, essentially forbidding the use of vtcm. In that case we could use -1 as a disable constraint as well as a default value. But, we don't have such a need right now.

python/tvm/meta_schedule/tir_integration.py
```diff
@@ -117,7 +120,7 @@ def tune_tir(
         task_weights=[1.0],
         work_dir=work_dir,
         max_trials_global=max_trials_global,
-        max_trials_per_task=max_trials_global,
+        max_trials_per_task=max_trials_per_task,
```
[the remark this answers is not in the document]
Icemist: Yes, mb it is redundant, but it unifies the interface. I will remove it from this PR.

include/tvm/driver/driver_api.h
csullivan: For now we should avoid making an API change to the compiler that would introduce `target` to lowering without some community discussion on the topic.

That said I also prefer deriving this target specific information from the target. In this case if we make the following change to move LowerVTCMAlloc into the MixedModulePassManager, we can add `tir::transform::VerifyVTCMLimit` just prior to it in the `mixed_pass_list`. As these passes are run at build time (after lowering), the target is available for use. 

```diff
diff --git a/src/driver/driver_api.cc b/src/driver/driver_api.cc
index e5e3998b1..babac2bf6 100644
--- a/src/driver/driver_api.cc
+++ b/src/driver/driver_api.cc
@@ -225,10 +225,8 @@ Array<tvm::transform::Pass> CreatePassList(bool disable_loop_partition) {
   if (!disable_storage_rewrite) {
     pass_list.push_back(tir::transform::StorageRewrite());
   }
-  // LowerVtcmAlloc must occur after any transformations that modify memory allocation locations
-  pass_list.push_back(tir::transform::LowerVtcmAlloc());
-  bool use_async_copy = pass_ctx->GetConfig<Bool>("tir.use_async_copy", Bool(false)).value();

+  bool use_async_copy = pass_ctx->GetConfig<Bool>("tir.use_async_copy", Bool(false)).value();
   if (use_async_copy) {
     pass_list.push_back(tir::transform::LowerAsyncDMA());
   }
@@ -539,6 +537,9 @@ transform::Sequential MixedModulePassManager(IRModule mixed_mod, Target target)

   Array<Pass> mixed_pass_list;
+  mixed_pass_list.push_back(tir::transform::VerifyVTCMLimit(GetVTCMCapacity(target, pass_ctx)));
+  // LowerVtcmAlloc must occur after any transformations that modify memory allocation locations
+  mixed_pass_list.push_back(tir::transform::LowerVtcmAlloc());
+
   mixed_pass_list.push_back(tir::transform::BindTarget(target));

   mixed_pass_list.push_back(tir::transform::VerifyMemory());
```

python/tvm/meta_schedule/tir_integration.py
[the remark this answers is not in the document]
Icemist: removed.

python/tvm/target/target.py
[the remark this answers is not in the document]
Icemist: changed.

echuraev (approved):
LGTM. Thanks

csullivan (approved):
LGTM! Thank you for the helpful analysis pass @Icemist

csullivan:
Thank you @Icemist, @adstraw, @echuraev @janetsc, @masahi for the great discussion and changes, this is merged.

PR #13351, opened 2022-11-10
[docs] Add some more release process docs

driazati:
This is mostly to address https://github.com/tlc-pack/tlcpack/issues/149
and make it clear how to avoid that in the future.


junrushao (approved):
LGTM

PR #13358, opened 2022-11-11
[TIR] Make syntax of AST nodes different than ops

junrushao:
Motivation example: in the current TVMScript, the snippet below:

```python
@T.prim_func
def func(...):
    A = T.match_buffer[(128 * 128), "float32"]
                       ^^^^^^^^^^^
```
is actually different from this one:

```python
@T.prim_func
def func(...):
    A = T.match_buffer[(16384, ), "float32"]
                       ^^^^^^^^^
    ...
```

This is because 128 * 128 is not immediately constant-folded into 16384, but becomes an AST node: `T.Mul(128, 128)`. This can be quite error-prone if not used carefully.

As part of effort of more formal TIR semantics, we want to more explicitly differentiate TIR AST nodes (defined in `tir/expr.h`) and TIR ops (defined in `tir/op.h`). The naming convention is that:
- Lowercased methods, for example, `tvm.tir.mul` or operator `*`, means an TIR op, which will be eagerly constant-folded, i.e. `mul(1, 2)` returns `3` immediately rather than creating an AST node.
- Capitalized callable, for example, `Mul`, means creating an AST node without constant folding.

This PR makes this behavior more explictly by printing `T.Mul(a, b)` directly when `a` and `b` are both constants, rather than sugaring it into `mul(a. b)` or `a * b`, so that the difference between an op and an AST node is clarified.

Co-authored-by: Yaxing Cai \<caiyaxing666@gmail.com\>

### DIFF
--- a/python/tvm/script/tir/intrin.py
+++ b/python/tvm/script/tir/intrin.py
@@ -17,12 +17,13 @@
 """TVM Script Parser Intrinsic Classes"""
 # pylint: disable=redefined-builtin, relative-beyond-top-level
 import builtins
-from typing import List, Any
+from typing import Any, List
 
 import tvm.tir
 from tvm.tir import FloatImm
-from ..registry import register
+
 from ...target import codegen
+from ..registry import register
 from ..utils import get_param_list, tvm_span_from_synr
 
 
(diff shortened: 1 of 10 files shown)


Hzfengsy (approved):
LGTM

PR #13361, opened 2022-11-11
[IRBuilder][Minor] Add intrinsics like `T.int32x4`

junrushao:
This PR adds all common TIR intrinsics like `T.int32x4`, `T.floatx4`.


MasterJH5574 (approved):
LGTM! Thanks @junrushao and @cyx-6!

PR #13362, opened 2022-11-11
[OpenCL] Introduce OpenCL wrapper to TVM

echuraev:
This wrapper helps dynamically loading OpenCL library. It allows us to avoid of looking for and copying OpenCL library to host, looking for OpenCL SDK.

cc: @elvin-n, @masahi, @csullivan


masahi:
@elvin-n Can you review this?

cmake/modules/OpenCL.cmake
```diff
@@ -50,9 +50,18 @@ endif(USE_AOCL)
 
 if(USE_OPENCL)
   if (NOT OpenCL_FOUND)
-    find_package(OpenCL REQUIRED)
+    find_package(OpenCL)
```
elvin-n, line 53: are these changes done only for adding definition of OpenCL_INCLUDE_DIRS? in this case let's remove everything else related to finging of opencl (we do not need it) and related to linking to opencl

echuraev: This `find_package` also set `OpenCL_LIBRARIES` this variable specifies which OpenCL library should be linked. But agree with you that the system OpenCL library should be found [here](https://github.com/apache/tvm/pull/13362/files#diff-45747f6e0c43e6e7579d2c50160a4bdd25108cf04e98dc7ed8f1b36ddcad09b1R19). 
I will remove this condition.

apps/android_camera/app/src/main/jni/Application.mk
```diff
@@ -32,9 +32,9 @@ APP_ABI ?= all
 APP_STL := c++_shared
 
 APP_CPPFLAGS += -DTVM_LOG_STACK_TRACE=0 -DTVM4J_ANDROID=1 -std=c++17 -Oz -frtti
-ifeq ($(USE_OPENCL), 1)
-    APP_CPPFLAGS += -DTVM_OPENCL_RUNTIME=1
-endif
+
+# OpenCL support
```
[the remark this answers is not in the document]
echuraev: I thought that it will be better to enable OpenCL support by default for common applications such as `android_camera` or `android_rpc`. `USE_OPENCL` is specified [here](https://github.com/apache/tvm/pull/13362/files#diff-4f7fde09bab2f87e09bd09c6e4c51c95fc1720dd6917cb239cc579fc68057e82L36-L37). It is a variable which was defined in `android_camera`. I can return this variable and enable build with OpenCL by default.

csullivan (approved):
Many thanks for this great improvement to the user experience of compiling and running code on remote OpenCL devices @echuraev!

cmake/modules/OpenCL.cmake
```diff
@@ -49,12 +49,17 @@ else()
...
 
 if(USE_OPENCL)
-  if (NOT OpenCL_FOUND)
-    find_package(OpenCL REQUIRED)
-  endif()
-  message(STATUS "Build with OpenCL support")
   tvm_file_glob(GLOB RUNTIME_OPENCL_SRCS src/runtime/opencl/*.cc)
-  list(APPEND TVM_RUNTIME_LINKER_LIBS ${OpenCL_LIBRARIES})
+  if (OpenCL_FOUND)
+    message(STATUS "Build with OpenCL support")
```
elvin-n, line 54: What are the cases when we find opencl and link vs it? Do we assume this wrapper not cover all cases and we will need to link vs original opencl?

apps/cpp_rpc/README.md
```diff
@@ -37,7 +37,15 @@ This folder contains a simple recipe to make RPC server in c++.
   # Path to the desired C++ cross compiler
   set(CMAKE_CXX_COMPILER /path/to/cross/compiler/executable)
 ```
-- If linking against a custom device OpenCL library is needed, in the config specify the path to the OpenCL SDK containing the include/CL headers and lib/ or lib64/libOpenCL.so:
+- If you need to build cpp_rpc with OpenCL support, specify variable `USE_OPENCL` in the config:
+  ```
+  set(USE_OPENCL ON)
+  ```
+  In this case [OpenCL-wrapper](../../src/runtime/opencl/opencl_wrapper) or OpenCL installed to your system will be used.
```
elvin-n, line 44: > or OpenCL installed

did you mean OpenCL SDK?
Do we consider OpenCL SDK flow at all?

src/runtime/opencl/opencl_wrapper/opencl_wrapper.cc
```diff
@@ -0,0 +1,551 @@
...
+  }
+
+ private:
+  LibOpenCLWrapper() {}
+  ~LibOpenCLWrapper() {
+    if (m_libHandler) dlclose(m_libHandler);
+  }
+  void openLibOpenCL() {
+    for (const auto it : default_so_paths) {
+      m_libHandler = dlopen(it, RTLD_LAZY);
```
elvin-n, line 82: this must not work on Windows. Need to branch and use LoadLibrary/GetProcAddress/FreeLibrary

elvin-n (changes requested):

apps/android_camera/app/src/main/jni/Application.mk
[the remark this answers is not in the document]
elvin-n: returned verification of USE_OPENCL :+1:

apps/cpp_rpc/README.md
echuraev: No, I mean that also it might be system OpenCL used. As far as I remember, you can install OpenCL on Ubuntu by this command: `sudo apt install ocl-icd-opencl-dev`. I'm not sure, can we name this package as OpenCL SDK or not. But anyway it should be enough to build the `cpp_rpc` for the host with OpenCL support. OpenCL will be found on this [line](https://github.com/apache/tvm/pull/13362/files#diff-45747f6e0c43e6e7579d2c50160a4bdd25108cf04e98dc7ed8f1b36ddcad09b1R19) and we won't use the wrapper at all.

cmake/modules/OpenCL.cmake
echuraev: This wrapper can cover all cases. I kept this search only for case which I have described in the comment to your question about OpenCL SDK. Actually, probably we can remove search of OpenCL and use wrapper in all cases. I didn't do it only because the user might want to specify a specific version of OpenCL or OpenCL SDK and I think we should give him/her this opportunity.

src/runtime/opencl/opencl_wrapper/opencl_wrapper.cc
```diff
@@ -0,0 +1,551 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/*!
+ * \file opencl_wrapper.cc
+ */
```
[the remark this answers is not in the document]
echuraev: Added information about OpenCL version. Currently, this wrapper is applicable for OpenCL 1.2. Because it is version which is used in TVM. But we can easily uplift this version, then it will be necessary.

cmake/modules/OpenCL.cmake
elvin-n: This can happen only if this specific version of OpenCL has functions which are not in the OpenCL standard. I hardly can imagine that tvm OpenCL runtime can depend on such functions. All specific can be in OpenCL extensions which can be a part of kernels, but not host API. Let's completely remove flow with static linking and depending on OpenCL SDK

apps/cpp_rpc/README.md
elvin-n: Let's stick to OpenCL wrapper only. It will definitely find OpenCL installed by apt

cmake/modules/OpenCL.cmake
elvin-n: Eventually we decided to leave static OpenCL linking but change conditions when it is selected and when wrapper should be selected. There might be cases when dlopen is prohibited due to sandbox limitations and only libraries statically selected by loader can be opened.
Modificatoin of USE_OPENCL logic: 
if USE_OPENCL has ON, always use wrapper and dynamic loading of libOpenCL.so in runtime
if USE_OPENCL points to openCL SDK, then use this sdk and static linking without wrapper

echuraev:
@tvm-bot rerun

cmake/config.cmake
```diff
@@ -65,7 +65,7 @@ set(USE_AOCL OFF)
 # Whether enable OpenCL runtime
 #
 # Possible values:
-# - ON: enable OpenCL with cmake's auto search
+# - ON: enable OpenCL with OpenCL wrapper
```
elvin-n, line 68: let's add more wording like 
"... OpenCL wrapper to remove dependency during build time and trigger dynamic search and loading of OpenCL in runtime"

elvin-n (changes requested):

cmake/config.cmake
echuraev: Done

echuraev:
@tvm-bot rerun

.gitmodules
```diff
@@ -16,3 +16,6 @@
 [submodule "3rdparty/cutlass"]
 	path = 3rdparty/cutlass
 	url = https://github.com/NVIDIA/cutlass.git
+[submodule "3rdparty/OpenCL-Headers"]
```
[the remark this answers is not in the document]
driazati: Can you add it to https://github.com/apache/tvm/blob/main/LICENSE as well?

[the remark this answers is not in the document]
echuraev: Done

elvin-n (approved):

PR #13365, opened 2022-11-14
[microNPU] Fix Cascader code generation without StorageRewrite

Alexey-Yazev:
This PR fixes Cascader code generation without StorageRewrite (when striping is enabled) on Ethos-U NPU.
The problem was that after the Replace Operator's pass, the buffers for partial results of the operation were replaced with a buffer with the results of the entire operation and it was indicated that the buffer needed to be allocated, summing up we received a larger size in the number of parts.


Alexey-Yazev:
cc @leandron @ekalda, @NicolaLancellotti

ekalda (commented):
Looks great, thanks for tracking this down! I suppose we want to wait for https://github.com/apache/tvm/pull/13369 before merging?

Alexey-Yazev:
> Looks great, thanks for tracking this down! I suppose we want to wait for #13369 before merging?
Yes, with these changes, the test will not be changed.

ekalda (approved):
LGTM!

ekalda:
Thanks @Alexey-Yazev!

PR #12496, opened 2022-08-19
[TVMScript] Reorganize the folder structure

junrushao:
This PR introduces some minor restructuring of the `python/tvm/script`
folder structure to make it more convenient for future upstreaming.

CC: @Hzfengsy @cyx-6

PR #13367, opened 2022-11-14
[Lint] Fix import in test_tvmscript_type.py

gigiblender:
#12496 does not patch an import that caused the linter to fail. This PR fixes it.

@driazati

### DIFF
--- a/tests/python/unittest/test_tvmscript_type.py
+++ b/tests/python/unittest/test_tvmscript_type.py
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 # pylint: disable=missing-function-docstring,missing-module-docstring,invalid-name,pointless-string-statement
-from tvm.script import tir as T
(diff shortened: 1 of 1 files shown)


driazati (commented):
Code looks relatively harmless but why wasn't this caught in CI / tests?

junrushao (approved):
Yeah I'm curious too, but it's good that we caught it

gigiblender:
> Would you mind sharing the error message from the linter? Thanks!

It was a python module not found error. I ran the linter locally using the docker image and, strangely am no longer able to reproduce it

junrushao:
Perhaps it could be some caching issues

junrushao:
hey what shall we do with this PR? probably we could close if the issue is not longer reproducible?

gigiblender:
> hey what shall we do with this PR? probably we could close if the issue is not longer reproducible?

Agree to close it.

PR #13373, opened 2022-11-14
[TIR] Add test to cover specific case of reducer match buffer checking

nverke:
Adding in a test that covers another case that can happen when running a check in reducer.cc [here](https://github.com/apache/tvm/blob/main/src/tir/schedule/analysis/reducer.cc#L590). 

cc @junrushao


junrushao (approved):
Thanks for fixing the unittest!

PR #13374, opened 2022-11-14
[Hexagon] Change single dma alloc buffers to be vtcm

nverke:
These need to be global.vtcm scope in order to properly test the single dma schedule. 

cc @adstraw

### DIFF
--- a/tests/python/contrib/test_hexagon/test_async_dma_pipeline.py
+++ b/tests/python/contrib/test_hexagon/test_async_dma_pipeline.py
@@ -113,124 +113,6 @@ def evaluate(
     return round(time.mean * 1000, 4)
 
 
-def get_single_dma_schedule(size_a, size_w):
-    """Generate single DMA schedule."""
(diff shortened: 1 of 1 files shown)


junrushao:
Yep I tried it out and see tests failing :-(

nverke:
Hmm which ones failed for you @junrushao?

nverke:
@tvm-bot rerun

junrushao:
The failing case is `test_loading_vtcm_for_vrmpy[1024-1]` in `TestAsyncDMAPipeline` in `test_hexagon/test_async_dma_pipeline.py`.

I spent some time last week looking into this testcase, and it turns out the the older parser cannot recognize `mem_scope="global.vtcm"` (because it's expecting `scope="global.vtcm"` as the input argument) and somehow silently falls back to using `scope="global"`.

junrushao (approved):
Thank you so much @nverke for following up!

PR #13375, opened 2022-11-14
[Metaschedule] Add partition primitive

tkonolige:
Add partition scheduling primitive. Partition splits a loop into two sequential loops based on a factor. For example:

```
for i in range(10):
  x[i] = 0
```

could be partitioned with a factor of 8 to

```
for i in range(8):
  x[i] = 0
for i_ in range(2):
  i = i_ + 8
  x[i] = 0
```

A couple of changes are necessary to support this change. Feature extraction had to be updated to handle loops with zero or negative extent and the parallel vectorize unroll postproc had to be updated to support nested blocks.

### DIFF
--- a/src/tir/schedule/analysis/analysis.cc
+++ b/src/tir/schedule/analysis/analysis.cc
@@ -890,7 +890,7 @@ IterVarType GetLoopIterType(const StmtSRef& loop_sref) {
           }
         });
       }
-      return false;
+      return true;
     }
     return true;
   };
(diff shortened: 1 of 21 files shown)


src/tir/schedule/analysis/analysis.cc
```diff
@@ -890,7 +890,7 @@ IterVarType GetLoopIterType(const StmtSRef& loop_sref) {
           }
         });
       }
-      return false;
+      return true;
```
jwfromm, line 893: What does this change do? Seems like it could cause side effects.

tkonolige: This change makes `GetLoopIterType` recurse into sub-blocks (`true` means visit sub-nodes). This makes `GetLoopIterType` correct in the presence of nested blocks.

Issue #13379, opened 2022-11-14
[Bug] no suitable user-defined conversion from "int2" to "longlong2" exists


comaniac:
PR #13317 introduces a bug when generating CUDA kernel for ops with index (e.g., gather).
With some investigations, it depends on the input shape. Taking the unit tests shown below, error only happens when the last dimension is an even number (e.g., 2).

cc @vinx13 @junrushao 

### Expected behavior

No error.

### Actual behavior

```
Testing shape  (4, 7, 5)
data type float16
idx type int64
Running on target: llvm
Running on target: cuda
Running on target: nvptx
Successed
Testing shape  (4, 2, 3)
data type float16
idx type int64
Running on target: llvm
Running on target: cuda
Running on target: nvptx
Successed
Testing shape  (4, 7, 3)
data type float16
idx type int64
Running on target: llvm
Running on target: cuda
Running on target: nvptx
Successed
Testing shape  (4, 7, 2)
data type float16
idx type int64
Running on target: llvm
Running on target: cuda
Failed: Traceback (most recent call last):
  23: TVMFuncCall
        at /home/ubuntu/raf/3rdparty/tvm/src/runtime/c_runtime_api.cc:477
  22: tvm::runtime::PackedFuncObj::CallPacked(tvm::runtime::TVMArgs, tvm::runtime::TVMRetValue*) const
        at /home/ubuntu/raf/3rdparty/tvm/include/tvm/runtime/packed_func.h:1217
  21: Call
        at /home/ubuntu/raf/3rdparty/tvm/include/tvm/runtime/packed_func.h:1213
  20: operator()
        at /home/ubuntu/raf/3rdparty/tvm/include/tvm/runtime/packed_func.h:1731
  19: unpack_call<tvm::runtime::Module, 2, tvm::<lambda(const tvm::runtime::Map<tvm::Target, tvm::IRModule>&, tvm::Target)> >
        at /home/ubuntu/raf/3rdparty/tvm/include/tvm/runtime/packed_func.h:1671
  18: run<>
        at /home/ubuntu/raf/3rdparty/tvm/include/tvm/runtime/packed_func.h:1631
  17: run<tvm::runtime::TVMMovableArgValueWithContext_>
        at /home/ubuntu/raf/3rdparty/tvm/include/tvm/runtime/packed_func.h:1631
  16: run<tvm::runtime::TVMMovableArgValueWithContext_, tvm::runtime::TVMMovableArgValueWithContext_>
[...]

PR #13382, opened 2022-11-14
[Codegen] Fix CUDA codegen for int64 Ramp

vinx13:
CUDA codegen fails to generate `T.ramp(T.int64(0), T.int64(1), 2)]` correctly.
This is a legal IR that appears after a recent commit to simplify cast and ramp. Fixed https://github.com/apache/tvm/issues/13379.

cc @comaniac @junrushao


comaniac (approved):
LGTM. Thanks for the quick fix.

junrushao:
Should fix #13379

comaniac:
Thanks @vinx13 @junrushao

PR #13383, opened 2022-11-15
[MetaSchedule] Add `from-target` Defaults for LLVM VNNI Targets

zxybazh:
This PR introduces default schedule rules, post processors and mutators for LLVM VNNI targets. The usage of default vnni classes will automatically be generated if using `from-target` as parameters.

Co-authored-by: Michal Piszczek <michalp@uw.edu>

### DIFF
--- a/include/tvm/meta_schedule/mutator.h
+++ b/include/tvm/meta_schedule/mutator.h
@@ -131,7 +131,7 @@ class Mutator : public runtime::ObjectRef {
                                    FApply f_apply, FClone f_clone, FAsString f_as_string);
   /*! \brief Create default mutators for LLVM */
   TVM_DLL static Map<Mutator, FloatImm, void> DefaultLLVM();
-  /*! \brief Create default mutators for LLVM VNNI */
+  /*! \brief Create default mutators for x86 VNNI */
   TVM_DLL static Map<Mutator, FloatImm, void> DefaultVNNI();
   /*! \brief Create default mutators for CUDA */
   TVM_DLL static Map<Mutator, FloatImm, void> DefaultCUDA();
--- a/include/tvm/meta_schedule/mutator.h
+++ b/include/tvm/meta_schedule/mutator.h
@@ -131,6 +131,8 @@ class Mutator : public runtime::ObjectRef {
                                    FApply f_apply, FClone f_clone, FAsString f_as_string);
   /*! \brief Create default mutators for LLVM */
   TVM_DLL static Map<Mutator, FloatImm, void> DefaultLLVM();
+  /*! \brief Create default mutators for LLVM VNNI */
+  TVM_DLL static Map<Mutator, FloatImm, void> DefaultVNNI();
   /*! \brief Create default mutators for CUDA */
   TVM_DLL static Map<Mutator, FloatImm, void> DefaultCUDA();
   /*! \brief Create default mutators for CUDA with TensorCore */
--- a/src/meta_schedule/postproc/postproc.cc
+++ b/src/meta_schedule/postproc/postproc.cc
@@ -59,6 +59,14 @@ Array<Postproc> Postproc::DefaultLLVM() {
   };
 }
 
+Array<Postproc> Postproc::DefaultVNNI() {
+  return Array<Postproc>{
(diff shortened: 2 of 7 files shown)


zxybazh:
CC @michalpiszczek @vinx13 @junrushao @masahi

src/meta_schedule/schedule_rule/schedule_rule.cc
```diff
@@ -85,6 +85,39 @@ Array<ScheduleRule> ScheduleRule::DefaultLLVM() {
   };
 }
 
+Array<ScheduleRule> ScheduleRule::DefaultVNNI() {
+  return {
+      ScheduleRule::AutoInline(
```
vinx13, line 90: We recently added new schedule rules `ApplyCustomRule` and `InlineConstantScalars` for llvm, let's also add them here for consistency

src/meta_schedule/schedule_rule/schedule_rule.cc
```diff
@@ -85,6 +85,39 @@ Array<ScheduleRule> ScheduleRule::DefaultLLVM() {
...
+          /*into_consumer=*/true,
+          /*inline_const_tensor=*/true,
+          /*disallow_if_then_else=*/true,
+          /*require_injective=*/true,
+          /*require_ordered=*/true,
+          /*disallow_op=*/Array<String>{"tir.exp"}),
+      ScheduleRule::AddRFactor(
+          /*max_jobs_per_core=*/16,
+          /*max_innermost_factor=*/Integer(64)),
+      ScheduleRule::MultiLevelTilingWithIntrin(
```
vinx13, line 101: need to add `MultiLevelTiling` (with the config in default llvm) as a fallback when tensorization is not applicable

src/meta_schedule/schedule_rule/schedule_rule.cc
masahi, line 90: I think you need to add `ApplyCustomRule` at the beginning, as of last week. Otherwise the fix in https://github.com/apache/tvm/pull/13346 won't be applied.

masahi: oops @vinx13 has beaten me by 30 seconds...

zxybazh:
Thanks for the comments, the previous implementation was based on an earlier commit, now the schedule rules have been added.

vinx13 (approved):

src/meta_schedule/postproc/postproc.cc
```diff
@@ -59,6 +59,14 @@ Array<Postproc> Postproc::DefaultLLVM() {
   };
 }
 
+Array<Postproc> Postproc::DefaultVNNI() {
+  return Array<Postproc>{
+      Postproc::DisallowDynamicLoop(),   Postproc::RewriteParallelVectorizeUnroll(),
+      Postproc::RewriteReductionBlock(), Postproc::RewriteTensorize(/*vectorize_init_loop=*/true),
+      Postproc::RewriteLayout(),
```
masahi, line 66: Is `RewriteLayout` useful here? Since VNNI expects the layout of the weight to be fixed.

include/tvm/meta_schedule/mutator.h
```diff
@@ -131,6 +131,8 @@ class Mutator : public runtime::ObjectRef {
                                    FApply f_apply, FClone f_clone, FAsString f_as_string);
   /*! \brief Create default mutators for LLVM */
   TVM_DLL static Map<Mutator, FloatImm, void> DefaultLLVM();
+  /*! \brief Create default mutators for LLVM VNNI */
```
masahi, line 134: Replace "LLVM VNNI" with "x86 VNNI". Same for `postproc.h` and `schedule_rule.h`

src/meta_schedule/postproc/postproc.cc
vinx13: I think it's useful as a fallback when auto tensorization can't apply

masahi: I'm cool with this as long as `RewriteLayout` doesn't break tensorization.

masahi (approved):

Hzfengsy (approved):

PR #13384, opened 2022-11-15
[MetaSchedule] Support schedules with cache read in RewriteLayout

masahi:
Currently when `CacheRead` and `RewriteLayout` are used together, the index map is derived based on the cache read block, which leads to weird result. This is because the current implementation assumes that the "layout-free" buffer is directly consumed by an "anchor" op such as conv2d or dense.

When `CacheRead` is involved, we need to find the index map for the cache-read buffer as it is consumed by an anchor op, and apply the same transformation to the layout-free buffer. My solution supports more general cases where there are multiple cache reads forming a "chain" of blocks, starting from the one that directly consumes the layout-free buffer passed as a parameter. So the layout transformation is back propagated over such chain. 

cc @vinx13 @junrushao @nverke

### DIFF
--- a/src/te/operation/create_primfunc.cc
+++ b/src/te/operation/create_primfunc.cc
@@ -110,13 +110,20 @@ class LayoutFreePlaceholdersNormalizer : public StmtMutator {
     Block block = Downcast<Block>(StmtMutator::VisitStmt_(_block));
     BlockNode* n = block.CopyOnWrite();
     if (Optional<ObjectRef> ann = n->annotations.Get(topi_attr)) {
+      Array<Buffer> new_buffers;
       for (Buffer buffer : Downcast<Array<Buffer>>(ann)) {
         auto it = buffer2index_.find(buffer);
         if (it != buffer2index_.end()) {
           layout_free_buffer_indices_.insert(it->second);
+        } else {
+          new_buffers.push_back(buffer);
         }
       }
-      n->annotations.erase(topi_attr);
+      if (new_buffers.empty()) {
+        n->annotations.erase(topi_attr);
+      } else {
+        n->annotations.Set(topi_attr, new_buffers);
+      }
     }
     for (const String& attr : this->blocklist) {
       auto it = n->annotations.find(attr);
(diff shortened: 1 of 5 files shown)


src/te/operation/create_primfunc.cc
```diff
@@ -110,13 +110,20 @@ class LayoutFreePlaceholdersNormalizer : public StmtMutator {
...
+        } else {
+          new_buffers.push_back(buffer);
         }
       }
-      n->annotations.erase(topi_attr);
+      if (new_buffers.empty()) {
+        n->annotations.erase(topi_attr);
+      } else {
+        n->annotations.Set(topi_attr, new_buffers);
+      }
```
masahi, line 126: This is a fix for `RewriteLayout` to work when `link-params = True`, the breakage discussed in https://github.com/apache/tvm/pull/13195. With this, I confirmed that the new `RewriteLayout` implementation works when there is `CacheRead` acting on `AllocateConst`. I'm not adding a test since TVMScript parsing is broken when there is `AllocateConst` with a large constant (can't print the whole array).

junrushao: Continuing the discussion from [this thread](https://github.com/apache/tvm/pull/13195#discussion_r1022226540).

I am convinced that we need to store buffers in block attributes for hexagon-specific usecases, and the only thing that I feel less natural is generally storing IR nodes in attributes. As an example, in `layout_free_placeholders`, we didn't store any IR nodes in the attribute, but instead use a list of integers which is simpler.

In the meantime, I completely understand that we need to get around this quickly, so in this particular case, how about adding or reusing a pass config flag, e.g. the one we are using "link-params", and only add enable `topi_attr` to be set when the flag is on?

```c++
else if (link_param is on) {
  n->annotations.Set(topi_attr, new_buffers);
}
```

junrushao: An alternative idea: having an extra flag in `CreatePrimFunc` that instructs the translator to retain buffers in block annotation, and the flag is on only if `link-param` is detected as "on" in TECompiler.

masahi: This `else` branch is only hit when there is a layout-free buffer that's not passed as a parameter to the prim func. Currently that can only happen when `link-params = True`. So isn't having another flag redundant?

junrushao: I see. That makes perfect sense to me :-)  As long as it's not affecting normal lowering flow, it looks like a good idea!

junrushao (approved):
LGTM!

PR #13385, opened 2022-11-15
[TOPI] Add padding for dense/batch matmul for x86 vnni

vinx13:
This added padding to make the shape of dense/batch matmul compatible with VNNI instructions.


junrushao (approved):
LGTM! This way it definitely solves many problems we have encountered when GEMM shape is less regular

junrushao:
Neat. Thanks for the contribution!

PR #13392, opened 2022-11-15
[AOT][FIX] Handle device contexts properly in CreateFunctionMetadata

mbaret:
Device contexts created handles with no corresponding buffer which would confuse CreateFunctionMetadata causing a segfault. The logic has been changed to check that a buffer exists and a test updated to include this case.


mbaret:
cc @ekalda

tmoreau89:
Thanks @mbaret @Mousius the PR has been merged

PR #13395, opened 2022-11-15
Add support for cast double to fp16

rkimball:
__truncdfhf2 can be generated in codegen but the implementation was missing. Added the missing implementation.


junrushao (approved):
LGTM!

PR #13396, opened 2022-11-15
[TVMScript] Use tir::Evaluate if expression is in statement context

Lunderberg:
For the previous version of the parser, this was special-cased for some intrinsic operators.  After the new TVMScript was enabled in https://github.com/apache/tvm/pull/12496, any `PrimExpr` that appears in the body of a statement is silently ignored.  This commit updates the parser to instead wrap the bare `PrimExpr` in a `tir::Evaluate` node.

This change effectively allows [expression statements](https://docs.python.org/3/reference/simple_stmts.html#expression-statements) in TVMScript, which are converted to `tir::Evaluate` nodes during parsing.

### DIFF
--- a/python/tvm/script/parser/tir/parser.py
+++ b/python/tvm/script/parser/tir/parser.py
@@ -20,6 +20,7 @@ import contextlib
 from functools import partial
 from typing import Any
 
+import tvm
 from tvm.ir import PrimType
 from tvm.tir import Buffer, IterVar, PrimExpr, Var
 
@@ -411,6 +412,10 @@ def visit_expr_stmt(self: Parser, node: doc.Expr) -> None:
     if isinstance(res, Frame):
         res.add_callback(partial(res.__exit__, None, None, None))
         res.__enter__()
+    elif isinstance(res, PrimExpr):
+        T.evaluate(res)
+    elif isinstance(res, (int, bool)):
+        T.evaluate(tvm.tir.const(res))
 
 
 @dispatch.register(token="tir", type_name="If")
--- a/src/printer/tvmscript_printer.cc
+++ b/src/printer/tvmscript_printer.cc
@@ -1277,9 +1277,15 @@ Doc TVMScriptPrinter::VisitStmt_(const SeqStmtNode* op) {
 Doc TVMScriptPrinter::VisitStmt_(const EvaluateNode* op) {
   // When parsing TVMScript, a PrimExpr that occurs as a statement is
   // automatically wrapped in `tir::Evaluate`.  Therefore, when
-  // printing we only need to print the value.
+  // printing, it's only necessary to print the value.  For
+  // readability, though, we still print T.evaluate() when the
+  // expression is something other than a call node.
   Doc doc;
-  doc << Print(op->value);
+  if (op->value.as<CallNode>()) {
+    doc << Print(op->value);
+  } else {
+    doc << tir_prefix_ << ".evaluate(" << Print(op->value) << ")";
+  }
   return doc;
 }
 
--- a/src/printer/tvmscript_printer.cc
+++ b/src/printer/tvmscript_printer.cc
@@ -1275,16 +1275,11 @@ Doc TVMScriptPrinter::VisitStmt_(const SeqStmtNode* op) {
 }
 
 Doc TVMScriptPrinter::VisitStmt_(const EvaluateNode* op) {
-  if (auto* call = op->value.as<CallNode>()) {
-    if (call->op.same_as(builtin::assume())) {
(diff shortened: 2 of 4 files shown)


Lunderberg:
This PR is in response to CI failures found in https://github.com/apache/tvm/pull/13130, when writing TVMScript that used the `T.assume()` builtin without wrapping in `T.evaluate()`.  I am updating that PR to use `T.evaluate()` in all new unit tests, but wanted to add this as a more general update.  With how TVMScript borrows heavily from Python's syntax, neither the pre-#12496 behavior of raising an error for expression statements, nor the post-#12496 behavior of ignoring expression statements felt expected as a user.

python/tvm/script/parser/tir/parser.py
```diff
@@ -411,6 +412,10 @@ def visit_expr_stmt(self: Parser, node: doc.Expr) -> None:
     if isinstance(res, Frame):
         res.add_callback(partial(res.__exit__, None, None, None))
         res.__enter__()
+    elif isinstance(res, PrimExpr):
+        T.evaluate(res)
+    elif isinstance(res, (int, bool)):
+        T.evaluate(tvm.tir.const(res))
```
junrushao, line 418: hey i was wondering if we want to explicitly print `T.evaluate(0)` vs `0`? The former one might look a bit clearer from my perspective

tests/python/unittest/test_tvmscript_roundtrip.py
```diff
@@ -3458,6 +3458,15 @@ def func() -> None:
     return func
 
 
+def implicit_evaluate():
+    @T.prim_func
+    def func(A: T.Buffer[1, "int32"]):
+        T.evaluate(T.assume(A[0] == 5))
+        A[0] = 10
```
junrushao, line 3465: just curious what's this line for?

junrushao (approved):
LGTM! I was thinking about the same thing but holding back a bit because I dont want to change TIR too often, but it is definitely a positive change and I really love it

src/printer/tvmscript_printer.cc
```diff
@@ -1275,16 +1275,11 @@ Doc TVMScriptPrinter::VisitStmt_(const SeqStmtNode* op) {
...
-      return doc;
-    }
-  }
-
+  // When parsing TVMScript, a PrimExpr that occurs as a statement is
+  // automatically wrapped in `tir::Evaluate`.  Therefore, when
+  // printing we only need to print the value.
   Doc doc;
-  doc << tir_prefix_ << ".evaluate(" << Print(op->value) << ")";
+  doc << Print(op->value);
```
junrushao, line 1282: How about we only sugar `CallNode`?

python/tvm/script/parser/tir/parser.py
Lunderberg: I was going back and forth on it, and I like the idea of printing `T.evaluate()` except in the case of CallNode.  The more aggressive sugaring would be to render `tir::Evaluate(0)` as `pass`, which would be even clearer to read from a casual Python reader.

tests/python/unittest/test_tvmscript_roundtrip.py
Lunderberg: Mostly because `T.assume` doesn't have any effect, and I was pulling this in from a failing case of `T.assume` in a separate PR.  This was from a unit test that validated a pass that removed `T.assume` calls from a primfunc.

src/printer/tvmscript_printer.cc
Lunderberg: I like that idea for the printing.  Maybe best of both worlds, to have the sugar apply all the time when parsing, but only `CallNode` drops the printing of `T.evaluate(...)` when generating TVMScript.

Lunderberg:
> LGTM! I was thinking about the same thing but holding back a bit because I dont want to change TIR too often, but it is definitely a positive change and I really love it

Thank you!  It's definitely a lot easier to add parsing/printing sugar to TVMScript than to change the TIR in C++.

tests/python/unittest/test_tvmscript_roundtrip.py
junrushao: Ah I see. Thanks for the clarification!

python/tvm/script/parser/tir/parser.py
junrushao: This is a good idea!

junrushao:
Thanks for the great update! This helps a lot with the usability of TVMScript :-)  Let's get it in ASAP!

Hzfengsy (approved):

PR #13397, opened 2022-11-16
[TECompiler] Replace static constant index with NameSupply

masahi:
The `static` qualifier on the constant index causes a problem, when `TECompiler` is invoked on the same module multiple times. Since the names of constants become different in each call to `TECompiler`, anything that depends on the consistency of constant names is broken. 

In practice, MetaSchedule tuning is affected by this issue, since task extraction and the final `relay.build(...)` both invoke `TECompiler`.  So if we do `get_block("fused_constant_0_global")` during tuning,  trace application after tuning will fail because there would be no longer a block with the name "fused_constant_0_global" .

### DIFF
--- a/src/relay/backend/task_extraction.cc
+++ b/src/relay/backend/task_extraction.cc
@@ -16,6 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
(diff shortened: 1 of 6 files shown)


junrushao (approved):
LGTM!

PR #13398, opened 2022-11-16
[CI] Update minor git options

mshr-h:
The changes are listed below.
- avoid git checkout after git clone

### DIFF
--- a/docker/install/ubuntu_install_ethosn_driver_stack.sh
+++ b/docker/install/ubuntu_install_ethosn_driver_stack.sh
@@ -50,10 +50,7 @@ apt-install-and-clear -y \
     wget
 
 cd "$tmpdir"
-git clone "$repo_url" "$repo_dir"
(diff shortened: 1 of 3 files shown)


junrushao (approved):
LGTM! Thanks for improving the git usage :-)

driazati:
I'm good with merging this PR but just curious about `--filter=tree:0`, it seems like from that article that `--depth=1` is still best here since it avoids downloading the whole history and we're not accessing any commits other than the checked out `HEAD`, am I missing something about it?

mshr-h:
@driazati 
Thanks for your comment!
You're right. Unless we need commit history), `--depth=1` is ok.
I guess that's more like a personal preference.

mshr-h:
I noticed that `--filter=tree:0` is supported since git 2.20 and Ubuntu 18.04 LTS has git 2.17.
We can't use `--filter=tree:0`. I'll revert it.

driazati (approved):

PR #13400, opened 2022-11-16
[OpenCL][CI] Enable OpenCL cpp tests in CI

echuraev:
Enable OpenCL cpp tests in CI


echuraev:
@driazati, @masahi please review it.

tests/scripts/task_config_build_gpu.sh
```diff
@@ -23,12 +23,28 @@ mkdir -p "$BUILD_DIR"
...
+repo_url="https://github.com/google/googletest"
+repo_revision="830fb567285c63ab5b5873e2e8b02f2249864916"
+archive_name="${repo_revision}.tar.gz"
+archive_url="${repo_url}/archive/${archive_name}"
+archive_folder="googletest-${repo_revision}"
+archive_hash="10f10ed771efc64a1d8234a7e4801838a468f8990e5d6d8fcf63e89f8d1455c4f9c5adc0bb829669f381609a9abf84e4c91a7fdd7404630f375f38fb485ef0eb"
+
+curl -sL "${archive_url}" -o "${archive_name}"
+echo "$archive_hash" ${archive_name} | sha512sum -c
+tar xf "${archive_name}"
```
echuraev, line 39: It is necessary to have google tests source code to build them for the device. In the CI we could use system GTest, but I didn't want to change the build mechanism in the CMake files, because I think it is better not to use system gtest but build it for the target device. 
This is why we need to download gtest source code in the build directory and build `tvm_runtime` with it.

driazati: The downloading is fine but this should be moved into the GPU docker image (https://github.com/apache/tvm/blob/main/docker/Dockerfile.ci_gpu) so we don't have to download it fresh each time. You'll have to send a separate PR for the docker changes, then we can update the images to use the new one, then once that's done we can merge this PR

driazati (approved):

echuraev:
@tvm-bot rerun

ci/jenkins/templates/gpu_jenkinsfile.groovy.j2
```diff
@@ -74,6 +74,14 @@
...
   )
   {% endif %}
+  sh (
+    script: "${docker_run} ${ci_gpu} python3 ./tests/scripts/task_build.py --sccache-bucket tvm-sccache-prod --cmake-target opencl-cpptest --build-dir build",
+    label: 'Make OpenCL cpp unit tests',
+  )
+  sh (
+    script: "${docker_run} ${ci_gpu} ./tests/scripts/task_opencl_cpp_unittest.sh",
+    label: 'Run OpenCL cpp unit tests',
+  )
```
driazati, line 84: Can you move this next to the `cpp_unittests` above? Right now this job is run 3 times afaik and pytest knows that each job should only run 1/3 of the tests, but this runs all the OpenCL tests 3 times

echuraev: Done

driazati (approved):

driazati:
lgtm assuming CI passes on the /branch jobs

Issue #13116, opened 2022-10-18
[Tracking Issue] TVM Explorer Infrastructure


chunit-quic:
### This issue is to track progress for [TVM Explorer Infrastructure](https://discuss.tvm.apache.org/t/pre-rfc-tvm-explorer-infrastructure/13457).

Main authors: @chunit-quic, @haowhsu-quic, @zack-ch  
RFC: [link](https://github.com/apache/tvm-rfcs/pull/92#issuecomment-1281411754)
# A. Span filling for frontends to Relay
- [x] Common API
- [x] TensorFlow 1
- [x] TFLite
- [x] PyTorch
- [x] ONNX

# B. Schedule/Stage visualization enhancement
- [x] Schedule record and API changes
- [x] TEDD modification

# C. Span propagation of Relay passes (passes within regular build flow will be covered)
- [ ] SequentialSpan and Source information builder
- [ ] Series of passes

Note that the span propagation of TIR passes is working in progress and not specifically included here. We are still refining some strategies to deal with span propagation from Relay and testing infrastructure. Beside, based on the discussion in pre-RFC about the tracing map, it might also affect the span filling design in the future.

PR #13402, opened 2022-11-16
[Relay][Frontend] Span filling common API

chunit-quic:
- Expose and add span attribute of Expr-derived types from C++ to Python
- Add common API of span filling
- Add test cases of span filling
- Add function to control whether to fill span via environment variable

### DIFF
--- a/python/tvm/relay/frontend/common.py
+++ b/python/tvm/relay/frontend/common.py
@@ -313,10 +313,9 @@ class ExprTable(object):
             shape = value.shape
         self.const_ctr += 1
         self.params[name] = value
-        tmp_var = _expr.var(name_hint=name, shape=shape, dtype=dtype)
+        self.exprs[name] = _expr.var(name_hint=name, shape=shape, dtype=dtype)
         if source_name:
-            tmp_var = set_span(tmp_var, source_name)
-        self.exprs[name] = tmp_var
+            self.exprs[name] = set_span(self.exprs[name], source_name)
         return self.exprs[name]
 
     def get_expr(self, name):
@@ -1130,7 +1129,39 @@ def _should_fill_span():
 
 
 def set_span(sym, span):
-    """Set up the sapn of relay expression(s) while converting OP"""
+    """
+    Recursively tag the span to the symbol. Stop when it encounters a span-tagged expr. Disabled
+    when setting the environment variable "TVM_SPANFILLING" as 0.
+
+    Parameters
+    ----------
+    sym :
+        A symbol is generated from the conversion of a frontend operator. Raise an error when the
+        type of the symbol is not supported.
+
+    span : String, Span, or bytes
+        The source information of the corresponding symbol.
+
+    Returns
+    -------
+    result :
+        The symbol tagged with span.
+
+    Examples
+    --------
+    .. code-block:: python
+
+      x = set_span(relay.var("x", shape=(1, 64, 56, 56)), "x_var")
+      w = relay.const(np.ones([64, 64, 3, 3]), dtype="int64")
+      y = set_span(
+          relay.nn.conv2d(x, w, channels=64, kernel_size=(3, 3), padding=(1, 1)), "conv2d"
+      )
+      print(relay.Function([x], y))
+
+      #fn (%x: Tensor[(1, 64, 56, 56), float32] /* span=x_var:0:0 */) {
+      #  nn.conv2d(%x, meta[relay.Constant][0] /* span=conv2d:0:0 */, ...) /* span=conv2d:0:0 */
+      #}
+    """
 
     if _should_fill_span():
         return _SpanFiller(span).fill(sym)
--- a/python/tvm/relay/frontend/common.py
+++ b/python/tvm/relay/frontend/common.py
@@ -18,12 +18,14 @@
 """Common utilities"""
 from __future__ import absolute_import as _abs
 import logging
+import os
 import numpy as np
 
 import tvm
 from tvm.ir import IRModule
 from tvm.topi.utils import get_const_tuple
 
+from ..expr_functor import ExprMutator
 from .. import expr as _expr
 from .. import function as _function
 from .. import transform as _transform
@@ -304,13 +306,17 @@ class ExprTable(object):
         self.const_ctr = 1
         self.in_padding = False
 
-    def new_const(self, value, shape=None, dtype="float32"):
+    def new_const(self, value, shape=None, dtype="float32", source_name=None):
+        """Construct a new var expr and add to exprs dictionary"""
         name = "_param_%d" % (self.const_ctr)
         if hasattr(value, "shape"):
             shape = value.shape
         self.const_ctr += 1
         self.params[name] = value
-        self.exprs[name] = _expr.var(name_hint=name, shape=shape, dtype=dtype)
+        tmp_var = _expr.var(name_hint=name, shape=shape, dtype=dtype)
+        if source_name:
+            tmp_var = set_span(tmp_var, source_name)
+        self.exprs[name] = tmp_var
         return self.exprs[name]
 
     def get_expr(self, name):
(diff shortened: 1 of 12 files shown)


chunit-quic:
Hi community,

This is the very first PR of the RFC, TVM Explorer Infrastructure. We spend a while to extract the code blocks, and verify them after rebasing. Feel free to give us and comment. Thank you! :D

[RFC in git](https://github.com/apache/tvm-rfcs/pull/92)
[Tracking issue](https://github.com/apache/tvm/issues/13116)

@haowhsu-quic, @zack-ch

echuraev (commented):
I have read the RFC about TVM Exprorer. It looks awesome! Thank you for your contribution. Added few minor comments.

python/tvm/relay/frontend/common.py
```diff
@@ -997,3 +1003,135 @@ def try_resolve_var_to_const(x, graph_params):
...
+        elif isinstance(sym, (float, int)):
+            return sym
+        elif isinstance(sym, np.ndarray):
+            return sym
+
+        raise RuntimeError(f"unsupported type {type(sym)}")
+
+
+def _should_fill_span():
+    should_fill_span = os.environ.get("TVM_SPANFILLING", "1")
```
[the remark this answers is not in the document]
chunit-quic: Thank you for point out. :D
I have pushed a patch with documentation below set_span(). This document includes environment variable setting, how to use and example.

python/tvm/relay/frontend/common.py
```diff
@@ -304,13 +306,17 @@ def __init__(self):
...
         name = "_param_%d" % (self.const_ctr)
         if hasattr(value, "shape"):
             shape = value.shape
         self.const_ctr += 1
         self.params[name] = value
-        self.exprs[name] = _expr.var(name_hint=name, shape=shape, dtype=dtype)
+        tmp_var = _expr.var(name_hint=name, shape=shape, dtype=dtype)
+        if source_name:
+            tmp_var = set_span(tmp_var, source_name)
+        self.exprs[name] = tmp_var
```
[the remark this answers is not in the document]
chunit-quic: Fixed.

chunit-quic:
@FrozenGene @areusch @mbs-octoml 
Tag reviewers in the forum thread and previously reverted PR.
It would be a big help to have your advices. Thank you! :D

echuraev (approved):
LGTM. Thanks

tests/python/frontend/test_common.py
```diff
@@ -27,6 +32,203 @@ def test_key_is_not_present():
...
+    _verify_call()
+    _verify_tuple()
+    _verify_tuple_getitem()
+    _verify_let()
+    _verify_if()
+    _verify_fn()
+
+
 if __name__ == "__main__":
     test_key_is_present()
```
areusch, line 232: can you change this to `tvm.testing.main()` so the test cases are split out ?

tests/python/frontend/test_common.py
```diff
@@ -27,6 +32,203 @@ def test_key_is_not_present():
     assert not attrs.has_attr("b")
 
 
+def test_set_span():
```
areusch, line 35: suggest to make this a TestCase class and promote the various e.g. `_verify_env_var_switch` functions to class methods e.g. `test_env_var_switch`, so they are separate test cases

tests/python/relay/utils/tag_span.py
```diff
@@ -0,0 +1,106 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+import tvm
+from tvm import relay, tir
+from tvm.relay.expr_functor import ExprVisitor
+
+
+def _set_span(expr, src):
```
areusch, line 22: we have an analogue `WithFields` in the C++, I think this would be better placed there. for now, see [FunctionWithFields](https://github.com/apache/tvm/blob/main/python/tvm/relay/function.py#L68).

tests/python/relay/utils/tag_span.py
```diff
@@ -0,0 +1,106 @@
...
+def _verify_span(lhs, rhs):
+    lhs_spans, rhs_spans = _collect_spans(lhs), _collect_spans(rhs)
+
+    if len(lhs_spans) != len(rhs_spans):
+        return False
+
+    for i in range(len(lhs_spans)):
+        if not tvm.ir.structural_equal(lhs_spans[i], rhs_spans[i]):
+            return False
+    return True
```
areusch, line 94: this style is a bit different than the typical pytest style e.g. `assert foo == bar` rather than `return foo == bar`.

python/tvm/relay/frontend/common.py
```diff
@@ -997,3 +1002,167 @@ def try_resolve_var_to_const(x, graph_params):
         return _op.const(value, dtype)
 
     return x
+
+
+class _SpanFiller(ExprMutator):
```
areusch, line 1007: we had been implementing a bunch of passes in C++, since the Python visitor infra involves a lot of calls that cross the FFI. I wonder if you'd be up for moving this into C++?

python/tvm/testing/utils.py
```diff
@@ -2081,3 +2081,28 @@ def pprint(name, obj):
                 f"or an instance of `tvm.tir.PrimFunc`.  "
                 f"Instead, received {type(expected)}."
             )
+
+
+class _control_span_filling:
+    def __init__(self, on=True):
+        self._old_state = os.environ["TVM_SPANFILLING"] if "TVM_SPANFILLING" in os.environ else None
```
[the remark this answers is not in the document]
chunit-quic: Because we can command like the following way to get rid of span without modifying the code. It would be a bit more convenient. 
```bash
TVM_SPANFILLING=0 python ${your_program.py}
````

File: src/relay/ir/expr.cc [hunk 9]
```diff
@@ -72,8 +72,8 @@ Constant::Constant(runtime::NDArray data, Span span) {
 
 TVM_REGISTER_NODE_TYPE(ConstantNode);
 
-TVM_REGISTER_GLOBAL("relay.ir.Constant").set_body_typed([](runtime::NDArray data) {
-  return Constant(data);
+TVM_REGISTER_GLOBAL("relay.ir.Constant").set_body_typed([](runtime::NDArray data, Span span) {
```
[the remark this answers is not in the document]
chunit-quic: Thank you for point out this concern. :D
Just a gentle reminder that we do have the default value of span for these types. Because currently the constructors in header ([include/tvm/relay/expr.h](https://github.com/apache/tvm/blob/b2058f4dd2e0ae1fc5ab51ac9f84b372a389a65a/include/tvm/relay/expr.h#L107)) have already been assigned with Span() to its attribute. We simply make span attribute be modifiable here.

tests/python/frontend/test_common.py
chunit-quic: Will do.

tests/python/frontend/test_common.py
chunit-quic: No problem!

tests/python/relay/utils/tag_span.py
chunit-quic: Thank you for let us know that we can leverage this function. :D
Yet about this _set_span, it is just a simple helper function and only used in the tests. Would you hope to add the functions like CallWithFields or TupleWithFields in C++ and expose them to python side in this patch? Or perhaps this _set_span is enough for the test cases?

tests/python/relay/utils/tag_span.py
chunit-quic: Will change.

python/tvm/relay/frontend/common.py
chunit-quic: I see. Currently we don't have plan to migrate them to a C++ pass. There are two reasons. 

1. First of all, considering the profiling result in the preRFC, the frontend conversion time is not increased too much. 
2. Second, the fill() and \_\_init\_\_() function heavily interact with different frontends. We would like to keep the flexibility to modify them via python script quickly. For example, in the [min_max_common ](https://github.com/apache/tvm/blob/b2058f4dd2e0ae1fc5ab51ac9f84b372a389a65a/python/tvm/relay/frontend/pytorch.py#L275)conversion of PyTorch, it might end up being a None, which is not common in other frontends. 

	
However, if the performance of frontend conversion is significantly decreased by SpanFiller in some cases, we can arrange a patch to move the mutator part (visitor funcions) to C++ pass. :)

chunit-quic:
> thanks @chunit-quic and sorry for the delay--i've been out on PTO for a while. i've left some comments here, overall this is a great addition!

Hi @areusch,
Thank you for reviewing! 
We have modified several parts based on your previous comments. :D 
Take your time (It's thanks giving recently.) We will wait for your reply.

src/relay/ir/expr.cc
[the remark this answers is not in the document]
areusch: okay cool--just wondering, since most of the passes are in C++, is the plan to invoke this via PackedFunc between each pass?

areusch (commented):
sorry for the delay @chunit-quic , one more question here

src/relay/ir/expr.cc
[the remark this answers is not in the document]
chunit-quic: Hi areusch,

Thanks for replying!
About the question "Is the plan to invoke this via PackedFunc between each pass?". 

No, this SpanFiller will not be triggered between each pass. Perhaps there are some misunderstandings. Please allow me to describe more. :D 

This PR, and the SpanFiller is for "frontend conversion" specifically. Take the upcomming PR of TFLite span filling for example, the function "set_span()" is invoked after a [conversion of an operator only](https://github.com/chunit-quic/tvm/blob/cee1703b4be6ac0e1ed3fb7c493432c95e7a5231/python/tvm/relay/frontend/tflite.py#L282).

About the [span filling mechanism for the passes](https://discuss.tvm.apache.org/t/pre-rfc-tvm-explorer-infrastructure/13457#pass-source-information-builder-6), we will use the SequentialSpan and source information builder to deal with. They are another set of classes and function to propagate span between passes specifically.

If there is any confusing part please feel free to let me know, I will try to make it clear. :)
Thank you!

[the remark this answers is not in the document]
areusch: ah okay that makes sense now, i think the point of this PR is very clear to me. thanks for explaining and sorry for the long delay here!

tests/python/relay/utils/tag_span.py
areusch: i think if you're going to do something like this, it would be great to just add e.g. CallWithFields and TupleWithFields properly to the FFI now. no need to migrate any other code, but keeping this code in test here seems like it's asking for two implementations to surface. lmk if this is too much of a burden.

python/tvm/relay/frontend/common.py
areusch: ok, given for now we want to invoke this after the import, we can keep it as Python for now

python/tvm/relay/frontend/common.py
```diff
@@ -997,3 +1002,167 @@ def try_resolve_var_to_const(x, graph_params):
...
+    def visit(self, expr):
+        if hasattr(expr, "span") and expr.span:
+            return expr
+
+        return super().visit(expr)
+
+    def visit_function(self, fn):
+        new_params = [self.visit(x) for x in fn.params]
+        new_body = self.visit(fn.body)
+        return _function.Function(
```
areusch, line 1030: want to use FunctionWithFields here too? (and similar below)

python/tvm/testing/utils.py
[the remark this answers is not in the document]
areusch: right--sorry perhaps I should have commented on _should_fill_span in python/tvm/relay/frontend/common.py. what I mean is--could we consider a global variable or a PassContext option which can be initialized from the env should a driver script want, rather than forcing this control to be in the environment?

areusch (commented):
sorry for the long delay here! i just have a couple of actionable comments, but otherwise I'm good with this PR landing.

python/tvm/testing/utils.py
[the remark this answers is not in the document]
chunit-quic: No problem, I change the way from using the environment variable to the passContext config. :)

python/tvm/relay/frontend/common.py
chunit-quic: Yes, I have changed them in the latest patch.

areusch (approved):
thanks @chunit-quic!

PR #13404, opened 2022-11-16
[Hexagon] Fix TIR vrmpy tensorization

masahi:
Hexagon tests using `vrmpy` tensorization were broken by the change in `tir/tensor_intrin/hexagon.py` from https://github.com/apache/tvm/pull/12496. The change is a correct one from the TIR perspective, but it breaks tensorize pattern matching. Actually `T.init()` block is not necessary in a tensorize description, so just removing it fixes the breakage. 

@mehrdadh


csullivan (approved):

junrushao:
Oh thanks for the update! In TIR, formally speaking, `T.init()` should only appear strictly under `T.block()`, so we got it removed to conform to the spec

PR #13405, opened 2022-11-16
[ETHOSN] Relax concatenate offloading requirements

lhutton1:
This appears to be a historical change due to a bug in the support library. Since the tests are working without this restriction and previously not offloaded concatenate operations are now being offloaded, it seems it is no longer necessary.


ashutosh-arm (approved):
Thanks @lhutton1 for this fix. It will help create a single partition for networks containing `concatenates`. Very excited to see this improve the correlation against the standard execution.

leandron (approved):
Thanks @lhutton1 @ashutosh-arm!

PR #13354, opened 2022-11-10
[MetaSchedule][Minor] Allow Zero Run Time In Benchmarking Result

zxybazh:
This PR introduces a check to prevent records with run time of zero into the training data of cost model. This is because when working on microTVM there're cases where the run time of certain successful runs is very tiny, such that it got recorded as zero. In such cases, the runtime of 0 would break XGBoost model because it introduces infinite running speed in GFLOPs. A regression test was also added.

PR #13406, opened 2022-11-16
[MetaSchedule][Fix] Fix Empty Run Time Issue when Benchmarking Result

Hzfengsy:
As a follow-up fix PR for https://github.com/apache/tvm/pull/13354, which introduces a bug that the tuning will crash if the run time is empty (usually because of a runtime error).

cc @junrushao @zxybazh

### DIFF
--- a/src/meta_schedule/measure_callback/update_cost_model.cc
+++ b/src/meta_schedule/measure_callback/update_cost_model.cc
@@ -44,7 +44,7 @@ class UpdateCostModelNode : public MeasureCallbackNode {
     for (int i = 0; i < n; i++) {
       if (!builder_results[i]->error_msg.defined() &&  //
           (runner_results[i]->error_msg.defined() ||   //
-          Sum(runner_results[i]->run_secs.value()) > 0)) {
+           Sum(runner_results[i]->run_secs.value()) > 0)) {
         pruned_candidate.push_back(measure_candidates[i]);
         pruned_runner_result.push_back(runner_results[i]);
       }
--- a/src/meta_schedule/measure_callback/update_cost_model.cc
+++ b/src/meta_schedule/measure_callback/update_cost_model.cc
@@ -42,7 +42,8 @@ class UpdateCostModelNode : public MeasureCallbackNode {
     pruned_candidate.reserve(n);
     pruned_runner_result.reserve(n);
     for (int i = 0; i < n; i++) {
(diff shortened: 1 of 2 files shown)


src/meta_schedule/measure_callback/update_cost_model.cc
```diff
@@ -42,7 +42,8 @@ class UpdateCostModelNode : public MeasureCallbackNode {
     pruned_candidate.reserve(n);
     pruned_runner_result.reserve(n);
     for (int i = 0; i < n; i++) {
-      if (!builder_results[i]->error_msg.defined() &&
+      if (!builder_results[i]->error_msg.defined() &&  //
+          !runner_results[i]->error_msg.defined() &&   //
```
zxybazh, line 46: We don't need to filter out failed runs because they are expected to be used as negative examples in cost models, for example see [here](https://cs.github.com/apache/tvm/blob/b4d4b82dbb9be2e4d0954f9dfd8e1c46079b66ee/python/tvm/meta_schedule/cost_model/xgb_model.py#L472) for xgboost model usage. I suggest that we change it to check `Sum(...) > 0` only when runner's error message is not defined.

src/meta_schedule/measure_callback/update_cost_model.cc
```diff
@@ -42,7 +42,8 @@ class UpdateCostModelNode : public MeasureCallbackNode {
...
           Sum(runner_results[i]->run_secs.value()) > 0) {
```
zxybazh, line 47:
```suggestion
      if (!builder_results[i]->error_msg.defined() &&  //
          (runner_results[i]->error_msg.defined() ||   //
          Sum(runner_results[i]->run_secs.value()) > 0)) {
```

zxybazh (commented):
Thanks for spotting the issue and send in the timely fix! I think we can preserve the negative sample bahavior and avoid this issue by slightly tweaking the check condition here. Otherwise LGTM.

shingjan (approved):
LGTM. Thanks for sending the fix!

src/meta_schedule/measure_callback/update_cost_model.cc
Hzfengsy: Thanks for pointing it out. It make sense to me

zxybazh (approved):
LGTM

PR #13407, opened 2022-11-16
[ACL] Enable int8 data type in QNN ADD

leandron:
This enables int8 data type to be used in QNN_ADD operator within Compute Library for the Arm(r) Architecture (ACL) BYOC integration.

cc @lhutton1 @d-smirnov for reviews

### DIFF
--- a/src/runtime/contrib/arm_compute_lib/acl_utils.cc
+++ b/src/runtime/contrib/arm_compute_lib/acl_utils.cc
@@ -130,6 +130,8 @@ arm_compute::DataType MakeACLDataType(const DLDataType& data_type) {
     return arm_compute::DataType::F32;
   } else if (data_type.code == DLDataTypeCode::kDLUInt && data_type.bits == 8) {
     return arm_compute::DataType::QASYMM8;
+  } else if (data_type.code == DLDataTypeCode::kDLInt && data_type.bits == 8) {
+    return arm_compute::DataType::QASYMM8_SIGNED;
   } else if (data_type.code == DLDataTypeCode::kDLInt && data_type.bits == 32) {
     return arm_compute::DataType::S32;
   } else {
--- a/tests/python/contrib/test_arm_compute_lib/test_add.py
+++ b/tests/python/contrib/test_arm_compute_lib/test_add.py
@@ -92,7 +92,8 @@ def test_runtime_add():
 
     for dtype, low, high, atol, rtol, op, op_params in [
         ("float32", -127, 128, 1e-7, 1e-7, relay.add, {}),
-        ("uint8", 0, 255, 0.0, 1.0, relay.qnn.op.add, _qnn_params),
+        ("uint8", 0, 255, 1.0, 0.0, relay.qnn.op.add, _qnn_params),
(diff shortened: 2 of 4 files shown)


tests/python/contrib/test_arm_compute_lib/test_add.py
```diff
@@ -93,6 +93,7 @@ def test_runtime_add():
     for dtype, low, high, atol, rtol, op, op_params in [
         ("float32", -127, 128, 1e-7, 1e-7, relay.add, {}),
         ("uint8", 0, 255, 0.0, 1.0, relay.qnn.op.add, _qnn_params),
+        ("int8", -127, 128, 0.0, 1.0, relay.qnn.op.add, _qnn_params),
```
lhutton1, line 96: Nit: looks like there has been a mixup between `atol` and `rtol` previously, I'd expect `atol=1` and `rtol=0`

lhutton1 (approved):
Thanks @leandron, LGTM!

src/runtime/contrib/arm_compute_lib/acl_utils.cc
```diff
@@ -130,6 +130,8 @@ arm_compute::DataType MakeACLDataType(const DLDataType& data_type) {
     return arm_compute::DataType::F32;
   } else if (data_type.code == DLDataTypeCode::kDLUInt && data_type.bits == 8) {
     return arm_compute::DataType::QASYMM8;
+  } else if (data_type.code == DLDataTypeCode::kDLInt && data_type.bits == 8) {
+    return arm_compute::DataType::QASYMM8_SIGNED;
```
ashutosh-arm, line 134: Just curious why only asymmetric dtype is supported?

tests/python/contrib/test_arm_compute_lib/test_add.py
leandron: Is it the same for `uint8` as well?

lhutton1: Yep should be, happy to leave to a followup though

src/runtime/contrib/arm_compute_lib/acl_utils.cc
leandron: I got it from the data types list here: https://github.com/ARM-software/ComputeLibrary/blob/aabef6c0584f06f4c0f4b61fb787d80374240619/src/core/CL/CLHelpers.cpp#L39

tests/python/contrib/test_arm_compute_lib/test_add.py
leandron: I'll update. Thanks

src/runtime/contrib/arm_compute_lib/acl_utils.cc
ashutosh-arm: After discussing this with @lhutton1, it looks like some of the buffers need to be symmetric as described on the https://www.tensorflow.org/lite/performance/quantization_spec. For Conv2D, Depthwise, Fully Connected layers, this could come handy. I wonder if the ACL errors out if the incorrect type is passed to the end APIs.

ashutosh-arm (approved):
Thanks @leandron. LGTM!

PR #13411, opened 2022-11-16
[Build] Fix MSVC compile option /bigobj

vinx13:
/bigobj was not correctly added to C++ targets. This caused error using cmake build.


junrushao (approved):
LGTM! How did you find that btw?

vinx13:
this error always happens using default cmake toolchain on windows , `conda build` works fine though that's why it was not captured by CI

PR #13414, opened 2022-11-17
[Hexagon] Enable depthwise conv2d NHWC with an HWIO kernel layout

farshidsp:
This PR adds support for `depthwise_conv2d_NHWC` with an HWIO kernel layout. 

@Lunderberg  @mehrdadh


python/tvm/topi/nn/depthwise_conv2d.py
```diff
@@ -284,9 +295,13 @@ def depthwise_conv2d_nhwc(Input, Filter, stride, padding, dilation, out_dtype=No
                     j * stride_w + dj * dilation_w,
                     idxdiv(c, channel_multiplier),
                 ].astype(out_dtype)
-                * Filter[
-                    di, dj, idxdiv(c, channel_multiplier), idxmod(c, channel_multiplier)
-                ].astype(out_dtype)
+                * Filter.__getitem__(
```
Lunderberg, line 298: The explicit `__getitem__` isn't required.  `Filter[x,y,z]` is syntactic sugar for `Filter[(x,y,z)]`.  If you already have a tuple, then `Filter[tuple( ... )]` can be called.

python/tvm/topi/nn/depthwise_conv2d.py
```diff
@@ -252,8 +253,18 @@ def depthwise_conv2d_nhwc(Input, Filter, stride, padding, dilation, out_dtype=No
...
+    dim = len(Input.shape) - 2
+
     # shape of dilated kernel
-    filter_height, filter_width, filter_channel, channel_multiplier = Filter.shape
+    if kernel_layout == "HWOI":
+        filter_height, filter_width, filter_channel, channel_multiplier = Filter.shape
+        kernel_permutation_to = [0, 1] + list(range(2, dim + 2))
+    elif kernel_layout == "HWIO":
+        filter_height, filter_width, channel_multiplier, filter_channel = Filter.shape
+        kernel_permutation_to = [dim + 1, dim] + list(range(dim))
```
Lunderberg, line 265: I'm not seeing how this permutation generates `HWIO`.  This defines `kernel_permutation_to` as `[3, 2, 0, 1]`, so `kernel_permutation_from` is `[2, 3, 1, 0]`.  With the usage below, that would permute from `[di, dj, c//channel_multiplier, c%channel_multiplier]` to `[c//channel_multiplier, c%channel_multiplier, dj, di]`, which would be `OIWH`.

Should this be `list(range(dim)) + [dim + 1, dim]` instead?

python/tvm/topi/nn/depthwise_conv2d.py
```diff
@@ -252,8 +253,18 @@ def depthwise_conv2d_nhwc(Input, Filter, stride, padding, dilation, out_dtype=No
...
+
+    kernel_permutation_from = np.argsort(kernel_permutation_to)
```
Lunderberg, line 267: Is there a benefit to defining in terms of `kernel_permutation_to` instead of directly defining `kernel_permutation_from`?

Lunderberg (commented):
I like the feature, but I'm not quite following how the permutation specified results in a `HWIO` kernel layout.

python/tvm/topi/nn/depthwise_conv2d.py
farshidsp: @Lunderberg  You are right, my bad. Not sure why I thought this would result in `HWIO`. Thanks so much for catching this bug.

python/tvm/topi/nn/depthwise_conv2d.py
farshidsp: Oh I see, I have changed this as your recommendation.

python/tvm/topi/nn/depthwise_conv2d.py
farshidsp: I tried to follow [this](https://github.com/apache/tvm/blob/6782a35018d5df63486569f17296737d75eedc55/python/tvm/topi/nn/conv2d.py#L837) as much as possible so in the future we can add more features if needed. If you think this is not the best way I could change it.

Lunderberg: Ah, I see.  It looks like that implementation is trying to be more clever, and to identify the permutation by inspecting the string.  In this case, since we're only supporting two explicitly enabled shapes, I'd lean for defining a `kernel_permutation` directly for each one.

```python
if kernel_layout == 'HWOI':
    kernel_permutaiton = [0,1,2,3]
elif kernel_layout == 'HWIO':
    kernel_permutaiton = [0,1,3,2]
else:
    raise ValueError(f'Unsupported kernel layout: {kernel_layout}')
```

Lunderberg: That said, if there are many locations where the same kernel permutation definitions are used, it may be useful to pull it out into a common utility.

farshidsp: Changed the way of permutation from your recommendation. @Lunderberg

TejashShah:
@farshidsp Is kernel layout or filter layout a better word to indicate the change in layout?

farshidsp:
> @farshidsp Is kernel layout or filter layout a better word to indicate the change in layout?

In other topi files, we have used `kernel_permutation` to indicate the change in the kernel layout and that's why I used the same term.

PR #13417, opened 2022-11-17
[bfloat16] Fixed dtype conversion in the arm_cpu injective schedule

ashutosh-arm:
Following on from the example conversion in ndarray.py,
in order to broaden the support for bfloat16 in the absence of
full support, this commit converts bfloat16 to uint16
in the injective schedule to enable full compilation
of Conv2D via arm_cpu schedules.

cc @leandron @ekalda


leandron (commented):
Is it possible to have a test case added, so that we can merge this?

PR #13420, opened 2022-11-17
[RPC] Fix tracker connection termination

Icemist:
In a wireless connection situation (android example below) we have a device key containing a colon character(**:**).

_ANDROID_SERIAL_NUMBER = 192.168.0.143:5555_ 

```
C:\Users\icemist>docker exec -it ice_tvm_container adb devices
List of devices attached
192.168.0.143:5555     device
```
In this case we get an error:
```
ERROR:asyncio:Exception in callback None()
handle: <Handle cancelled>
  File "/usr/lib/python3.8/asyncio/events.py", line 81, in _run
    self._context.run(self._callback, *self._args)
  File "/venv/apache-tvm-py3.8/lib/python3.8/site-packages/tornado/platform/asyncio.py", line 206, in _handle_events
    handler_func(fileobj, events)
  File "/git/tvm/python/tvm/rpc/tornado_util.py", line 41, in _event_handler
    self._event_handler(events)
  File "/git/tvm/python/tvm/rpc/tornado_util.py", line 79, in _event_handler
    if self._update_read() and (events & self._ioloop.WRITE):
  File "/git/tvm/python/tvm/rpc/tornado_util.py", line 121, in _update_read
    self.close()
  File "/git/tvm/python/tvm/rpc/tornado_util.py", line 67, in close
    self.on_close()
  File "/git/tvm/python/tvm/rpc/tracker.py", line 298, in on_close
    self._tracker.close(self)
  File "/git/tvm/python/tvm/rpc/tracker.py", line 353, in close
    self._scheduler_map[rpc_key].remove(value)
KeyError: '127.0.0.1'
```

For example, test_rpc_tracker_via_proxy hangs if the device key format contains a colon.

The point is that the keys before using the split have a format like:
conn.put_value has: _hexagon-dev.127.0.0.1:5555:0.513619_
_tracker_pending_puts has: _127.0.0.1:5555:0.2809967317812382_

### DIFF
--- a/python/tvm/rpc/base.py
+++ b/python/tvm/rpc/base.py
@@ -120,13 +120,16 @@ def recvjson(sock):
     return data
 
 
-def random_key(prefix, cmap=None):
+def random_key(prefix, delimiter=":", cmap=None):
     """Generate a random key
 
     Parameters
     ----------
     prefix : str
         The string prefix
+
+    delimiter : str
+        The delimiter
 
     cmap : dict
         Conflict map
@@ -136,13 +139,30 @@ def random_key(prefix, cmap=None):
     key : str
         The generated random key
     """
-    if cmap:
-        while True:
-            key = prefix + str(random.random())
-            if key not in cmap:
-                return key
-    else:
-        return prefix + str(random.random())
+    while True:
+        key = "{}{}{}".format(prefix, delimiter, random.random())
+        if not cmap or key not in cmap:
+            break
+    return key
+
+
+def split_random_key(key, delimiter=":"):
+    """Split a random key by delimiter into prefix and random part
+
+    Parameters
+    ----------
+    key : str
+        The generated random key
+
+    Returns
+    -------
+    prefix : str
+        The string prefix
+
+    random_part : str
+        The generated random
+    """
+    return key.rsplit(delimiter, 1)
 
 
 def connect_with_retry(addr, timeout=60, retry_period=5):
--- a/python/tvm/rpc/proxy.py
+++ b/python/tvm/rpc/proxy.py
@@ -322,7 +322,7 @@ class ProxyServerHandler(object):
             rpc_key, _ = base.split_random_key(key)
             handle = self._server_pool[key]
             del self._server_pool[key]
-            new_key = base.random_key(rpc_key + ":", keyset)
+            new_key = base.random_key(rpc_key, keyset)
             self._server_pool[new_key] = handle
             keyset.add(new_key)
             new_keys.append(new_key)
--- a/python/tvm/rpc/proxy.py
+++ b/python/tvm/rpc/proxy.py
@@ -319,7 +319,7 @@ class ProxyServerHandler(object):
         new_keys = []
         # re-generate the server match key, so old information is invalidated.
         for key in keys:
-            rpc_key, _ = key.split(":")
+            rpc_key, _ = base.split_random_key(key)
             handle = self._server_pool[key]
             del self._server_pool[key]
             new_key = base.random_key(rpc_key + ":", keyset)
@@ -368,7 +368,7 @@ class ProxyServerHandler(object):
             need_update_info = False
             # report new connections
             for key in self._tracker_pending_puts:
-                rpc_key = key.rsplit(":", 1)[0]
+                rpc_key, _ = base.split_random_key(key)
                 base.sendjson(
                     self._tracker_conn, [TrackerCode.PUT, rpc_key, (self._listen_port, key), None]
                 )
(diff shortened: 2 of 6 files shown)


Icemist:
@echuraev @apeskov @areusch

tests/python/unittest/test_runtime_rpc.py
```diff
@@ -448,10 +448,10 @@ def check_remote():
 
 
 @tvm.testing.requires_rpc
-def test_rpc_tracker_register():
+@pytest.mark.parametrize("device_key", ["test_device", "127.0.0.1:5555"])
```
echuraev, line 451: Is it a complete list with all possible keys which `adb devices` can return? Not clear from this added key, why previous version with split doesn't work.

Icemist: I'm not sure. I want to fix exactly the current case. 
The issue in that case is in format of key `hexagon-dev.192.168.0.143:5555:0.513619`.

Before the fix this test was falling with "127.0.0.1:5555", after it works

echuraev: Hm... It looks strange that the test has failed before with "127.0.0.1:5555". I wrote the following code:
```python
key = "127.0.0.1:5555"

print(key.split(":")[0])
print(key.rsplit(":", 1)[0])
```

And expectedly received the following result:
```
127.0.0.1
127.0.0.1
```

What is the problem to add one more key, e.g. `hexagon-dev.192.168.0.143:5555:0.513619`?

Icemist: > What is the problem to add one more key, e.g. hexagon-dev.192.168.0.143:5555:0.513619?
It is not device key, it is internal representation inside implementation.

echuraev: Thank you for clarification! In verbal discussion, decided to fix all usage of key split in TVM code, create a function which have to do this split and write simple unit tests on it. To show that the key might has three parts separated by `:`: `host:port:unique_random_key`.

python/tvm/rpc/base.py
```diff
@@ -136,13 +139,30 @@ def random_key(prefix, cmap=None):
...
-        while True:
-            key = prefix + str(random.random())
-            if key not in cmap:
-                return key
-    else:
-        return prefix + str(random.random())
+    while True:
+        key = "{}{}{}".format(prefix, delimiter, random.random())
+        if not cmap or key not in cmap:
+            break
```
echuraev, line 145: Is it expected that if the `key` is found in `cmap` then on the next iteration it contains more delimiters than only one?
E.g.:
- First iteration:
  ```
  prefix = "127.0.0.1:5555"
  key = "127.0.0.1:5555:123456"
  ```
  `cmap` has already contained this key `127.0.0.1:5555:123456`. We will iterate once again.
- Second iteration:
  ```
  key = "127.0.0.1:5555:123456:654321"
  ```
  
If we split this key after the second iteration, then we get `127.0.0.1:5555:123456` but it is not the key.
Probably this change will help us to avoid this problem?
```suggestion
        key = "{}{}{}".format(prefix, delimiter, random.random())
        if not cmap or key not in cmap:
            break
        delimiter = ""
```
Or maybe I missed something and everything should work?

python/tvm/rpc/proxy.py
```diff
@@ -319,7 +319,7 @@ def _regenerate_server_keys(self, keys):
         new_keys = []
         # re-generate the server match key, so old information is invalidated.
         for key in keys:
-            rpc_key, _ = key.split(":")
+            rpc_key, _ = base.split_random_key(key)
             handle = self._server_pool[key]
             del self._server_pool[key]
             new_key = base.random_key(rpc_key + ":", keyset)
```
echuraev, line 325: Should we also update this line? Because variable `keyset` is not on the right place in the list of arguments and I expect that after first iteration we will have something like that: `127.0.0.1:5555::123456`.

echuraev (commented):
Please carefully update all using of the updated functions.

python/tvm/rpc/base.py
Icemist: there is no saving of the variable key between iterations of the loop. there is a test for this situation. tests/python/unittest/test_rpc_base.py

python/tvm/rpc/proxy.py
Icemist: thanks. updated.

python/tvm/rpc/base.py
echuraev: Ups, sorry, my bad. Didn't notice that. Thank you.

echuraev (approved):
LGTM. Thanks!

PR #13421, opened 2022-11-17
[MetaSchedule] Fix Dynamic Loop from AutoBinding

zxybazh:
This PR fixes an issue that would cause dynamic loop when reorder is applied during auto binding incorrectly. This would cause failed cuda kernel generation. This PR provided a fix by changing the loop to fuse blocks.

Test is also added and locally tested on `nvidia/geforce-rtx-3070` target.

Thanks for @junrushao for the guidance and @yelite for reporting this issue.

### DIFF
--- a/src/meta_schedule/schedule/cuda/thread_bind.cc
+++ b/src/meta_schedule/schedule/cuda/thread_bind.cc
@@ -55,14 +55,14 @@ std::function<ExprRV(int64_t)> MakeFactorSampler(Schedule sch, Array<Integer> th
 
 Array<LoopRV> BindSpatialLoop(Schedule sch, LoopRV loop, int64_t max_threadblocks,
                               int64_t max_threads_per_block,
-                              std::function<ExprRV(int64_t)> get_factor, bool allow_reorder) {
+                              std::function<ExprRV(int64_t)> get_factor) {
   int64_t extent = -1;
   if (const int64_t* e = as_const_int(sch->Get(loop)->extent)) {
     extent = *e;
   } else {
     extent = std::numeric_limits<int64_t>::max();
   }
-  if (extent <= max_threadblocks * max_threads_per_block || !allow_reorder) {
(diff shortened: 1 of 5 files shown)


zxybazh:
CC @junrushao @Hzfengsy

src/meta_schedule/schedule/cuda/thread_bind.cc
```diff
@@ -55,14 +55,14 @@ std::function<ExprRV(int64_t)> MakeFactorSampler(Schedule sch, Array<Integer> th
...
   if (const int64_t* e = as_const_int(sch->Get(loop)->extent)) {
     extent = *e;
   } else {
     extent = std::numeric_limits<int64_t>::max();
   }
-  if (extent <= max_threadblocks * max_threads_per_block) {
+  if (extent <= max_threadblocks * max_threads_per_block || !allow_reorder) {
     if (!get_factor) {
       get_factor = MakeFactorSampler(sch, {32, 64, 128, 256, 512, 1024});
     }
```
junrushao, line 68: let's instead do this:

```c++
if (extent <= xxx) {
} else if (allow_reorder) {
  split by [None, max_threadblocks, max_threads_per_block];
} else {
  split by [max_threadblocks, max_threads_per_block, None];
}
```

tests/python/unittest/test_meta_schedule_tune_tir.py
```diff
@@ -32,6 +32,96 @@
 logging.basicConfig()
 logging.getLogger("tvm.meta_schedule").setLevel(logging.DEBUG)
 
+# fmt: off
+@tvm.script.ir_module
```
junrushao, line 36: please move this test to `tests/python/unittest/test_meta_schedule_space_cuda_winograd.py`

zxybazh:
Test integrated into the cuda winograd search space. Please take a 2nd look, thanks!

junrushao (approved):
LGTM!

PR #13423, opened 2022-11-17
[TOPI] Add handwritten matvec for dynamic cases

tkonolige:
Add a handwritten matrix-vector multiplication implementation for dynamic cases on cpu. This avoids crashing when a dynamic shape is present. Only supports vector LHS and float32 for now.

### DIFF
--- a/python/tvm/relay/op/strategy/x86.py
+++ b/python/tvm/relay/op/strategy/x86.py
@@ -518,9 +518,9 @@ def dense_strategy_cpu(attrs, inputs, out_type, target):
     strategy = _op.OpStrategy()
     # For dynamic matrix-vector multiply we use a hand written kernel.
     if (
-        inputs[0].shape[0] == 1
-        and is_dynamic_shape(inputs[0].shape)
-        or is_dynamic_shape(inputs[1].shape)
+        isinstance(inputs[0].shape[0], (int, tir.IntImm))
+        and inputs[0].shape[0] == 1
+        and (is_dynamic_shape(inputs[0].shape) or is_dynamic_shape(inputs[1].shape))
     ):
         strategy.add_implementation(
             wrap_compute_dense(topi.x86.dense_dynamic),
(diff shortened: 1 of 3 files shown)


python/tvm/relay/op/strategy/x86.py
```diff
@@ -507,10 +507,26 @@ def matmul_strategy_cpu(attrs, inputs, out_type, target):
     return strategy
 
 
+def is_any(shape):
```
jwfromm, line 510: `is_dynamic` might be a more appropriate name for this function.

python/tvm/relay/op/strategy/x86.py
```diff
@@ -507,10 +507,26 @@ def matmul_strategy_cpu(attrs, inputs, out_type, target):
...
+
+
 @dense_strategy.register("cpu")
 def dense_strategy_cpu(attrs, inputs, out_type, target):
     """dense x86 strategy"""
+
     strategy = _op.OpStrategy()
+    # For dynamic shapes we use a hand written kernel. Right now it only
+    # supports matrix-vector multiplication.
+    if is_any(inputs[0].shape) or is_any(inputs[1].shape):
```
jwfromm, line 521: should we check the size of `inputs[0]` to make sure its a vector before dispatching?

jwfromm (approved):
This all makes sense to me, I left a few nitpicks but otherwise should be good to go.

python/tvm/relay/op/strategy/x86.py
tkonolige: Added a check in the implementation.

python/tvm/relay/op/strategy/x86.py
tkonolige: renamed to `is_dynamic_shape`

tkonolige:
@jwfromm a change: the dynamic kernel usage is predicated on it being a matrix-vector multiply instead of asserting within the kernel. Apparently there are some other code paths that do dynamic matmul that I broke.

jwfromm:
Yeah that makes perfect sense to me, good change.

tkonolige:
@tvm-bot rerun

michalpiszczek:
@tvm-bot rerun

michalpiszczek:
@tvm-bot rerun

junrushao (approved):
LGTM!

PR #13429, opened 2022-11-18
[TIR][Schedule] Add condition to check buffer type

abhikran-quic:
Add condition to check buffer type while creating array for padding.
This check is helpful to ensure that `buffer_obj` data type should be similar to `pad_value` type.

Co-authored by: Anirudh Sundar Subramaniam <quic_sanirudh@quicinc.com>

cc: @Lunderberg


python/tvm/tir/schedule/schedule.py
```diff
@@ -2751,10 +2751,14 @@ def two_elementwise_transformed_intermediate_buffer(a: T.handle, c: T.handle) ->
             # buffer's type.  If the default `tvm.runtime.convert`
             # behavior is applied, these would be converted to
             # int32/float32, which may not match the buffer's type.
-            if isinstance(pad_value, int):
+            if "int" in buffer_obj.dtype and isinstance(pad_value, int):
                 pad_value = IntImm(buffer_obj.dtype, pad_value)
-            elif isinstance(pad_value, float):
+            elif "float" in buffer_obj.dtype and isinstance(pad_value, float):
                 pad_value = FloatImm(buffer_obj.dtype, pad_value)
+            else:
```
Lunderberg, line 2758: I like the change overall, to avoid constructing an `IntImm` for type `float64`, but this `else` case should be removed.  There's an existing check on the C++ side [here](https://github.com/apache/tvm/blob/main/src/tir/schedule/primitive/layout_transformation.cc#L1073), which validates the resulting type.  The python-side type coercion is only necessary to avoid ambiguous conversions.

tests/python/unittest/test_tir_schedule_transform_layout.py
```diff
@@ -533,6 +533,21 @@ def before():
...
+    pad_value = tvm.testing.parameter(0)
+
+    def before():
+        A = T.alloc_buffer(14, "float32")
+        for i in T.serial(14):
+            with T.block("block"):
+                vi = T.axis.remap("S", [i])
+                A[vi] = 0
+
+    expected = TypeError
```
Lunderberg, line 548: Does the behavior here differ from [TestErrorOnWrongPaddingType](https://github.com/apache/tvm/blob/main/tests/python/unittest/test_tir_schedule_transform_layout.py#L521)?  Other than the type of error thrown, it looks like they are testing the same behavior.

Lunderberg (changes requested):
I like the change overall, though it looks like it could raise type errors if the pad value is already a `PrimExpr`.  A couple of suggestions to improve the fix.

abhikran-quic (commented):
Thank you @Lunderberg for the review! I've added my comments as well. 
Sorry for replying back late. I was not well last week.

abhikran-quic:
@tvm-bot rerun

tests/python/unittest/test_tir_schedule_transform_layout.py
Lunderberg: Ah, got it.  A test of usage error where an entirely incompatible type is being passed, rather than just a test that requires a different Python to TIR conversion.  Thank you!

Lunderberg (approved):

PR #13431, opened 2022-11-18
[TE][TIR] Improved naming when converting TE to schedulable TIR

Lunderberg:
Prior to this commit, loop iterators were named `i0`, `i1`, and so on, while the `BlockNode::iter_vars` used the name from the TE `IterVar`. As a result, after `BlockNode::iter_vars` is lowered out, the resulting `PrimFunc` no longer contained the user-generated iterator names.  This commit updates the TIR conversion so that the loop iterators take the name of the TE `IterVar`, and the `BlockNode::iter_vars` are named `v_$IterVarName`.


junrushao:
CC: @Hzfengsy

Hzfengsy (approved):

PR #13432, opened 2022-11-18
[TIR][TVMScript] Cleaner printing of And/Or chains

Lunderberg:
Even though the operator precedence of `And` has a higher precedence than `Or`, removing parentheses based on this precedence can harm readability.  This commit adds an exception to the TVMScript rules for parentheses, to always insert parentheses between `And` and `Or` operators.

In addition, adding a rewrite rule to preferentially produce And/Or chains that may be expressed in a single left-associative chain of operators.

Between these two changes, the readability of boolean expressions can be improved.  Below is the motivating example for this change.  In each case, the output had been passed through the `black` formatter. Both expressions are equivalent, but the before-case was much more difficult to read.

```python
x = (
    AAA == 0
    and BBB < 4
    or AAA == 7
    and 6 <= BBB
    or (CCC == 0 and DDD < 4 or CCC == 7 and 6 <= DDD)
)

x = (
    (AAA == 0 and BBB < 4)
    or (AAA == 7 and 6 <= BBB)
    or (CCC == 0 and DDD < 4)
    or (CCC == 7 and 6 <= DDD)
)
```

### DIFF
--- a/src/arith/rewrite_simplify.cc
+++ b/src/arith/rewrite_simplify.cc
@@ -1732,7 +1732,7 @@ PrimExpr RewriteSimplifier::Impl::VisitExpr_(const AndNode* op) {
   }
 
   // Pattern var to match any expression
-  PVar<PrimExpr> x, y;
+  PVar<PrimExpr> x, y, z;
   // Pattern var match IntImm
   PVar<IntImm> c1, c2, c3;
   PVar<int> lanes;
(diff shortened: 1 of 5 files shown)


junrushao:
CC: @yelite

yelite:
This looks good to me! Can you also add some unit tests to the `test_tvmscript_roundtrip.py`?

yelite (approved):

Lunderberg:
> This looks good to me! Can you also add some unit tests to the `test_tvmscript_roundtrip.py`?

Certainly, and added!

junrushao (approved):

PR #13433, opened 2022-11-18
[TVMScript] Output elif where possible

Lunderberg:
When a TIR `IfThenElseNode::else_case` is itself a `IfThenElse` node, it may be printed as `elif` instead of `else: if ...`.  The `elif` syntax was already handled on the parsing side, so this commit only changes the TVMScript printer.

This change also incidentally altered the printer's output for an else block whose condition is `True`.  Previously, these were dropped from the output altogether.  Now, they are retained in the else block. Removing no-op statements should be done by a transformation pass, not a printer/parser round trip, so this change in behavior makes sense to keep.

Below are shown the same prim func, printed before and after this change.

```python
# Before
@T.prim_func
def func(i: T.int32):
    # body
    if i == 0:
        T.evaluate(0)
    else:
        if i == 1:
            T.evaluate(1)
        else:
            if i == 2:
                T.evaluate(2)
            else:
                T.evaluate(3)

# After
@T.prim_func
def func(i: T.int32):
    # body
    if i == 0:
        T.evaluate(0)
    elif i == 1:
        T.evaluate(1)
    elif i == 2:
        T.evaluate(2)
    else:
        T.evaluate(3)
```

### DIFF
--- a/src/printer/tvmscript_printer.cc
+++ b/src/printer/tvmscript_printer.cc
@@ -1259,10 +1259,22 @@ Doc TVMScriptPrinter::VisitStmt_(const IfThenElseNode* op) {
   Doc doc;
   doc << "if " << Print(op->condition) << ":";
   doc << Doc::Indent(4, Doc::NewLine() << PrintBody(op->then_case));
-  if (!is_one(op->condition) && op->else_case) {
-    doc << Doc::NewLine();
-    doc << "else:" << Doc::Indent(4, Doc::NewLine() << PrintBody(op->else_case.value()));
+
+  Optional<Stmt> else_case = op->else_case;
+  while (else_case) {
+    if (auto* else_if = else_case.value().as<IfThenElseNode>()) {
+      doc << Doc::NewLine();
+      doc << "elif " << Print(else_if->condition) << ":";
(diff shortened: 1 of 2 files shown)


yelite (approved):
This looks good! Thanks for making this improvement!

src/printer/tvmscript_printer.cc
```diff
@@ -1259,10 +1259,22 @@ Doc TVMScriptPrinter::VisitStmt_(const IfThenElseNode* op) {
   Doc doc;
   doc << "if " << Print(op->condition) << ":";
   doc << Doc::Indent(4, Doc::NewLine() << PrintBody(op->then_case));
-  if (!is_one(op->condition) && op->else_case) {
-    doc << Doc::NewLine();
-    doc << "else:" << Doc::Indent(4, Doc::NewLine() << PrintBody(op->else_case.value()));
+
+  Optional<Stmt> else_case = op->else_case;
```
shingjan, line 1263: would that be possible that we can add a comment here or a code snippet in the PR to show the elif print before and after this change. Otherwise LGTM.

Lunderberg: Certainly!  Added a quick example to the PR description.

junrushao (approved):
LGTM!

PR #13436, opened 2022-11-18
[TIR][Hexagon] Async DMA fixups

csullivan:
Bug fixes for async DMA lowering with Hexagon mode bypass:
1. Bypass is currently always on.
2. The LowerAsyncDMA doesn't visit the body of an async_wait attribute. 

cc @adstraw


adstraw (approved):
LGTM, thanks for cleaning this up!

mehrdadh (approved):

PR #13437, opened 2022-11-18
[Roofline] Allow user choice of pass for saving lowered TIR

tkonolige:
Useful if users what to use a different pass for TIR extraction.


junrushao (approved):
LGTM!

PR #13438, opened 2022-11-18
[TVMScript] Improvements tvm.script.highlight

Lunderberg:
- Automatically use "black" formatter if available.

- Allow overrides of pygmentize style based on environment variable `TVM_PYGMENTIZE_STYLE`.

### DIFF
--- a/python/tvm/ir/module.py
+++ b/python/tvm/ir/module.py
@@ -278,18 +278,24 @@ class IRModule(Node):
             self, tir_prefix, show_meta
         )  # type: ignore
 
-    def show(self, style: Optional[str] = None) -> None:
-        """
-        A sugar for print highlighted TVM script.
+    def show(self, style: Optional[str] = None, black_format: bool = True) -> None:
+        """A sugar for print highlighted TVM script.
+
         Parameters
         ----------
         style : str, optional
-            Pygments styles extended by "light" (default) and "dark", by default "light"
+
+            Pygmentize printing style, auto-detected if None.  See
+            `tvm.script.highlight.cprint` for more details.
+
+        black_format: bool
+
+            If true (default), use the formatter Black to format the TVMScript
         """
         from tvm.script.highlight import cprint  # pylint: disable=import-outside-toplevel
 
         # Use deferred import to avoid circular import while keeping cprint under tvm/script
-        cprint(self, style=style)
+        cprint(self, style=style, black_format=black_format)
 
     def get_attr(self, attr_key):
         """Get the IRModule attribute.
--- a/python/tvm/script/highlight.py
+++ b/python/tvm/script/highlight.py
@@ -17,6 +17,7 @@
 """Highlight printed TVM script.
 """
 
+import os
 import sys
 import warnings
 from typing import Optional, Union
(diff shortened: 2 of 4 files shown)


Lunderberg:
This PR started from finding that the `dracula` style shows up best in my terminal, and wanted to have a way to enable it as the default pygmentize for my environment.

junrushao:
Using black is format is a great idea! On the other hand, sometimes we want to toggle it off for huge IRs, for example, the winograd TIR has two extremely long lines which we dont really need to read much into: https://gist.github.com/zxybazh/0019023b0796a55c9fadb12593852f84. Would love to get some thoughts from you which one is better: formatting by default or not?

Lunderberg:
Oof, yeah, looks like that one becomes pretty heavily indented with `black`, with several sections starting at indentations around line 60-80.  It looks like it is specifically the nested function calls (here the chain of `T.select`), each of which adds indentation.

I think I'd still lean toward having the formatting enabled by default, since I think it improves readability on average, but it would be good to have a way to disable it when debugging cases where the readability is worsened.  Maybe having the argument default be set by an environment variable, so that the formatting could be enabled/disabled across all calls as needed.

Lunderberg:
Hmm, I'm also wondering on the representation of `T.select`.  Both it and `T.if_then_else` are useful to use in nested chains, which are what causes the poor formatting.  If a nested sequence of `N` select/if_then_else were sugared as a sequence of `2*N + 1` arguments, we could avoid having a lot of the structures that get poorly formatted.  So the winograd TIR example, line 37 would be formatted into ~40 flat lines, instead of the ~80 heavily indented lines that it currently formats into.

```python
B[i, j] = T.Select(
    i % 4 == 3 and j % 4 == 3,
    T.float32(1),
    i % 4 == 3 and j % 4 == 2,
    T.float32(0),
    i % 4 == 3 and j % 4 == 1,
    T.float32(0),
    i % 4 == 3 and j % 4 == 0,
    T.float32(0),
    i % 4 == 2 and j % 4 == 3,
    T.float32(0),
    i % 4 == 2 and j % 4 == 2,
    T.float32(1),
    i % 4 == 2 and j % 4 == 1,
    T.float32(1),
    i % 4 == 2 and j % 4 == 0,
    T.float32(-1),
    i % 4 == 1 and j % 4 == 3,
    T.float32(-1),
    i % 4 == 1 and j % 4 == 2,
    T.float32(1),
    i % 4 == 1 and j % 4 == 1,
    T.float32(-1),
    i % 4 == 1 and j % 4 == 0,
    T.float32(0),
    i % 4 == 0 and j % 4 == 3,
    T.float32(0),
    i % 4 == 0 and j % 4 == 2,
    T.float32(0),
    i % 4 == 0 and j % 4 == 1,
    T.float32(0),
    i % 4 == 0 and j % 4 == 0,
    T.float32(1),
    T.float32(0),
)
```

junrushao:
Some background: The reason that winograd conv2d has such a long chained `T.Select` is that it's used to represent A, B, G matrices, and by unrolling the corresponding loops in our scheduling, the `T.Select` is constant-folded into scalar constants. Therefore there is no performance degradation if scheduled properly.

On sugaring long chained `T.Select`: The usecases of sugaring `T.Select`, if it's only limited to constant matrices in winograd conv, then it's probably less common and I would say it may not be desirable because it requires users to understand a new syntax. 

My two-cents: as long as we have a flag to turn it off, in `.show()` methods, I don't have a strong opinion if it defaults to on or off :-)  Alternatively, when the IR is much longer (e.g. +20 extra lines after formatting), we could print out a doc, saying you may turn formatting off by tweaking the flag, etc.

Lunderberg:
Good point on the `T.Select` optimizing out during unrolling, so it wouldn't be something with semantic meaning early on.  That makes sense on having the flag to disable it the formatting.  Looks like I initially added `black_format=True` as an argument to `cprint`, but not to the methods that call into it.  All the `obj.show()` methods now also have a `black_format=True` argument.

junrushao (approved):
LGTM!

PR #13439, opened 2022-11-18
[RPC] Catch potential race-condition exception on termination

naut-thomas:
Observed a race condition on Hexagon where the rpc_key got removed from the scheduler map and thus wasn't in the remove loop anymore and throwing a KeyError. A simple try-except on the first try makes sure it is still there before removal (if we still have to).


naut-thomas:
@Lunderberg @areusch

areusch:
thanks @naut-thomas, can you confirm whether `__del__` is executing when the stacktrace is raised?

PR #13446, opened 2022-11-19
[MetaSchedule] Add MultiLevelTilingHexagon to create simple schedules…

nverke:
… that utilize async dma pipelines.

This mostly copies what is happening in MultiLevelTilingTensorCore but alters the pipelining strategy for hexagon async dma pipelining. 

I kept the intrin groups in so that this can be expanded if we want to include a synchronous dma tensorization.

### DIFF
--- a/src/meta_schedule/schedule_rule/multi_level_tiling_hexagon.cc
+++ b/src/meta_schedule/schedule_rule/multi_level_tiling_hexagon.cc
@@ -272,8 +272,8 @@ std::vector<State> MultiLevelTilingHexagonNode::AddSoftwarePipeline(HexagonState
   sch->Annotate(cache_read_loops[cache_read_loops.size() - 2],
                 tir::attr::software_pipeline_async_stages, software_pipeline_async_stages);
 
-  // TODO: Add support for nested async pipelines.
-  // TODO: Add support for async cache writes.
+  // TODO(nverke): Add support for nested async pipelines.
+  // TODO(nverke): Add support for async cache writes.
   return {state};
 }
 
--- a/src/meta_schedule/schedule_rule/multi_level_tiling_hexagon.cc
+++ b/src/meta_schedule/schedule_rule/multi_level_tiling_hexagon.cc
@@ -77,7 +77,8 @@ class HexagonStateNode : public StateNode {
 
 class HexagonState : public State {
  public:
-  explicit HexagonState(HexagonIntrinGroup intrin_group, tir::AutoTensorizeMappingInfo mapping_info, Schedule sch, BlockRV block_rv, Array<Array<tir::LoopRV>> tiles = {});
+  explicit HexagonState(HexagonIntrinGroup intrin_group, tir::AutoTensorizeMappingInfo mapping_info,
+                        Schedule sch, BlockRV block_rv, Array<Array<tir::LoopRV>> tiles = {});
 
   TVM_DEFINE_MUTABLE_OBJECT_REF_METHODS(HexagonState, State, HexagonStateNode);
 };
@@ -85,8 +86,8 @@ class HexagonState : public State {
 TVM_REGISTER_OBJECT_TYPE(HexagonStateNode);
 
 HexagonState::HexagonState(HexagonIntrinGroup intrin_group,
-                                 tir::AutoTensorizeMappingInfo mapping_info, Schedule sch,
-                                 BlockRV block_rv, Array<Array<LoopRV>> tiles) {
+                           tir::AutoTensorizeMappingInfo mapping_info, Schedule sch,
+                           BlockRV block_rv, Array<Array<LoopRV>> tiles) {
   ObjectPtr<HexagonStateNode> node = make_object<HexagonStateNode>();
   node->intrin_group = intrin_group;
   node->mapping_info = mapping_info;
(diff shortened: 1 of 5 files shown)


src/meta_schedule/schedule_rule/multi_level_tiling_hexagon.cc
```diff
@@ -0,0 +1,440 @@
...
+  if(cache_read_count == 2) {
+    software_pipeline_stage = Array<Integer>{0, 0, 1};
+    software_pipeline_order = Array<Integer>{0, 1, 2};
+    software_pipeline_async_stages = Array<Integer>{0};
+  } else {
+    software_pipeline_stage = Array<Integer>{0, 1};
+    software_pipeline_order = Array<Integer>{0, 1};
+    software_pipeline_async_stages = Array<Integer>{0};
+  }
+  sch->Annotate(cache_read_loops[cache_read_loops.size() - 2], tir::attr::software_pipeline_stage, software_pipeline_stage);
```
masahi, line 271: `cache_read_loops[cache_read_loops.size() - 2` might not be the correct reduction loop we want to pipeline. Moreover, due to many splitting, a single loop typically has a small extent. So we should fuse all reduction loops above a `cache_read` block.

masahi:
I made a new branch based on this PR, to apply `MultiLevelTilingHexagon` in practice: https://github.com/apache/tvm/compare/main...masahi:hex-ms-async?expand=1

Currently, tuning can run but the output is incorrect if async DMA is enabled. If we enable only VTCM cache read and synchronous software pipeline, the output is correct. See the test case in `test_hexagon/test_conv2d_async.py`. Here is an example of a generated IR after `LowerAsyncDMA`: https://gist.github.com/masahi/f15806cb19f5c00da10a12341aef137e

The main changes from this PR are:
* Start from `MultiLevelTilingWithIntrin`, rather than `MultiLevelTilingTensorCore`, to avoid unnecessary complications like `reindex` and `transform_layout`.
*  Using `AddReadReuse` with level `1` leads to `cache_read` block being placed under a spatial loop. The pipeline annotation is also attached to the wrong loop, which makes `DecomposeReduction` fail to place the initialization block outside of the pipeline loop. This results in an error `pipeline_stages.size() != original_order.size() (4 vs 3)` (the init block is counted as one of the blocks in the pipeline loop). Using `levels = [2]` fixes these issues.
* Moreover, rather than pipelining only over a single reduction loop, fuse all reduction loops above a cache-read block. For example, in
```
for i0_0, i1_0, i2_0, i3_0, i4_0_0, i5_0, i6_0, i7_0, i8_0, i9_0_0 in T.grid(T.int64(1), T.int64(2), T.int64(6), T.int64(3), T.int64(1), T.int64(1), T.int64(3), T.int64(1), T.int64(4), T.int64(1)):
    for ax0_ax1_ax2_ax3_ax4_fused in T.serial(T.int64(3168)):
        with T.block("p0_global.vtcm"):
           ...
    for ax0_ax1_ax2_ax3_ax4_ax5_ax6_fused in T.serial(T.int64(1536)):
        with T.block("fused_constant_global.vtcm"):
           ...
```
 `i5_0, i6_0, i7_0, i8_0, i9_0_0` are all reduction loops, which can be fused to make a bigger pipelined loop. After fusing and applying the annotations, it looks like:
```
for i0_0, i1_0, i2_0, i3_0, i4_0_0 in T.grid(T.int64(1), T.int64(1), T.int64(1), T.int64(1), T.int64(1)):
    for i5_0_i6_0_i7_0_i8_0_i9_0_0_fused in T.serial(T.int64(12), annotations={"software_pipeline_async_stages":[0], "software_pipeline_order":[0, 1, 2], "software_pipeline_stage":[0, 0, 1]}):
        for ax0_ax1_ax2_ax3_ax4_fused in T.serial(T.int64(48384)):
            with T.block("p0_global.vtcm"):
              ...
        for ax0_ax1_ax2_ax3_ax4_ax5_ax6_fused in T.serial(T.int64(3072)):
            with T.block("fused_constant_global.vtcm"):
              ...
```
* Disable parallelization, since doing `dma_copy` under a parallel loop is not supported by the current Hexagon runtime.

tests/python/unittest/test_meta_schedule_schedule_rule_mlt_hexagon.py
```diff
@@ -0,0 +1,973 @@
...
+        intrin_groups=[
+            {"compute": VRMPY_u8u8i32_VTCM_INTRIN},
+        ],
+        structure="SRSRS",
+        tile_binds=None,
+        max_innermost_factor=64,  # 64 // tensor intrin size
+        vector_load_lens=None,
+        reuse_read=ms.schedule_rule.ReuseType(
+            req="must",
+            levels=[1],
```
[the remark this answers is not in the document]
nverke: Ahh I see that now I will make changes to the test!

src/meta_schedule/schedule_rule/multi_level_tiling_hexagon.cc
nverke: Agreed, this is just the way that I have been creating these for testing purposes. We have seen experimentally that there are better configurations.

nverke:
Awesome! Ill take a look at that branch!

PR #11174, opened 2022-04-28
[fix] vec * mat in matmul in onnx converter

ganler:
Should fix https://github.com/apache/tvm/issues/10651 for ONNX converter.

cc: @masahi

PR #13448, opened 2022-11-21
[FIX][ONNX][Relay] onnx converter on matmul with scalar; bring back nn.matmul check

ganler:
This PR brings 2 bug fixes:
1. ONNX converter for matmul: ONNX matmul follows NumPy [rules](https://numpy.org/doc/stable/reference/generated/numpy.matmul.html):
> If the first argument is 1-D, it is promoted to a matrix by prepending a 1 to its dimensions. After matrix multiplication the prepended 1 is removed.
> If the second argument is 1-D, it is promoted to a matrix by appending a 1 to its dimensions. After matrix multiplication the appended 1 is removed.

The (my) previous fix https://github.com/apache/tvm/pull/11174 did not consider the second rule (append 1 dimension for the rhs vector).

2. Relay's `nn.matmul` takes 2-D matrices and the checker was removed in a recent PR https://github.com/apache/tvm/pull/13287. This PR puts the checker back to prevent process crashes (make it a readable TVMError) for readability (and also for that the CI in https://github.com/ise-uiuc/nnsmith/pull/64 won't be terminated while using TVM-10).

cc: @masahi @tkonolige

### DIFF
--- a/python/tvm/relay/frontend/onnx.py
+++ b/python/tvm/relay/frontend/onnx.py
@@ -358,8 +358,19 @@ def matmul_out_dtype(inputs, out_dtype):
         )
         return _op.reshape(output, fold_constant(final_shape))
 
-    if a_rank == 1:
-        return _op.squeeze(_op.nn.matmul(_op.expand_dims(inputs[0], axis=0), inputs[1]), axis=[0])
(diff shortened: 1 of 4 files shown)


junrushao (approved):
LGTM!

PR #13451, opened 2022-11-21
[RUNTIME][CLML] Fix Dense layer crash issue

krishnaraj36:
Fixed the bias tensor allocation failure issue for Dense layer in CLML runtime.


srkreddy1238 (changes requested):
LGTM

PR #13452, opened 2022-11-21
[MetaSchedule] TorchBench tuning script: add task extraction mode

yelite:
This PR adds the task extraction mode to the TorchBench tuning script. The task extraction mode will only import PyTorch model into TVM, extract tuning tasks from it and save them for further use. 

This helps the use case where tasks should be extracted from a machine with GPU and then be tuned on a machine with powerful CPU with rpc workers.

cc @junrushao @zxybazh

### DIFF
--- a/python/tvm/meta_schedule/testing/torchbench/run.py
+++ b/python/tvm/meta_schedule/testing/torchbench/run.py
@@ -94,6 +94,7 @@ import argparse
 import contextlib
 import logging
 import os
+import pickle
 import sys
 import warnings
 from collections import defaultdict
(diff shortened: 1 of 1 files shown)


junrushao:
CC: @zxybazh

python/tvm/meta_schedule/testing/torchbench/run.py
```diff
@@ -124,28 +125,37 @@
 class RunMode(Enum):
     """
     The running mode of this script. Available values are:
-    - tune: Only tune the model and create the tuning database.
+    - extract: Only import the model and extract tuning tasks from it.
+    - tune: Only tune the tasks and create the tuning database.
```
zxybazh, line 129: Is it possible that we need some other combination of stages, say `extract+tune` or `tune+evaluate`?

zxybazh (approved):
Generally looking good, just wondering if there's any better way to denote the input stages, creating multiple strings for combinations of the 3 stages (`extract`, `tune`, `eval`) seems not scalable when we have customized need.

shingjan (approved):
LGTM

PR #13327, opened 2022-11-09
[TIR] Unify index data type when creating prim func

vinx13:
* Added data type pass unification pass to by default promote data types of all indices and shapes to int64 when creating prim func.
* Added some fixes for lowering passes to make it compatible with int64 data type.

PR #13463, opened 2022-11-22
[TIR] Fix buffer shape and IndexMap indices dtype mismatch

masahi:
After the PR https://github.com/apache/tvm/pull/13327, I'm getting a dtype-mismatch error at https://github.com/apache/tvm/blob/b6fae9b35eff4ad1f7cc2e83d8d7da5d701d8e44/src/tir/schedule/primitive/layout_transformation.cc#L310, because `dim.dtype()` is now int64 while `virtual_var.dtype()` is int32. The dtypes of `initial_indices` in `IndexMap` are fixed to int32 (see below), this is in conflict with the above PR which made the dtypes of buffer shapes int64.

https://github.com/apache/tvm/blob/458ca81c01bd85d04ea6428e7cb4bfa021a1d37a/src/tir/ir/index_map.cc#L51
https://github.com/apache/tvm/blob/78b53221f8dd8c1d2bbeff9d34803db33ca254dd/python/tvm/tir/function.py#L395

Since `initial_indices` is used everywhere to construct `loop_var` / `iter_var` / `iter_values` of `For` and `Block` etc, I'm adding a dtype legalization at the beginning of `TransformLayout`, when the dtypes of the input buffer and `initial_indices` do not match.

@vinx13 @Lunderberg

### DIFF
--- a/src/tir/schedule/primitive/layout_transformation.cc
+++ b/src/tir/schedule/primitive/layout_transformation.cc
@@ -1055,6 +1055,8 @@ class TransformationIntroducesPaddingError : public ScheduleError {
   PrimExpr padding_predicate_;
 };
 
+// Make the dtypes of indices in IndexMap be the same as the dtype of the buffer shape, to avoid
+// dtype-mismatch issues later.
 IndexMap LegalizeIndexMapDType(const IndexMap& index_map, const Buffer& buf) {
   auto initial_indices_orig = index_map->initial_indices;
   ICHECK(buf->shape.size() == initial_indices_orig.size());
--- a/src/tir/schedule/primitive/layout_transformation.cc
+++ b/src/tir/schedule/primitive/layout_transformation.cc
@@ -1055,13 +1055,41 @@ class TransformationIntroducesPaddingError : public ScheduleError {
   PrimExpr padding_predicate_;
 };
 
+IndexMap LegalizeIndexMapDType(const IndexMap& index_map, const Buffer& buf) {
+  auto initial_indices_orig = index_map->initial_indices;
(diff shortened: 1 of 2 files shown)


Lunderberg (commented):
Good catch, and the change looks reasonable.  The `MapIndices` method already uses `SubstituteWithDataTypeLegalization`, so it makes sense for the transform layout to use it as well.

Lunderberg:
Regarding the readability of the unit test, I did some poking around, and the following PrimFunc triggers the same error on main when the layout is transformed using `lambda h: [h//8, h%8]`.

```python
@T.prim_func
def func(A: T.Buffer[T.int64(58), "int32"]):
    for i in T.serial(T.int64(58)):
        with T.block("block"):
            vi = T.axis.remap("S", [i])
            A[vi] = 0
```

masahi:
@Lunderberg Thanks for finding the simple test case! I hit this error when running this Hexagon test https://github.com/apache/tvm/blob/b29ab5c6baea8884a5b9c0702c1c93da6a830866/tests/python/contrib/test_hexagon/metaschedule_e2e/test_resnet50_int8.py#L397. The original test case in this PR, with its `cache_read`, index map etc, was directly taken from this Hexagon test.

The bug is triggered when we hit the code path https://github.com/apache/tvm/blob/b6fae9b35eff4ad1f7cc2e83d8d7da5d701d8e44/src/tir/schedule/primitive/layout_transformation.cc#L282

I was not sure what makes this code path hit and why existing tests didn't hit it.

masahi:
@Lunderberg @vinx13 PTAL, thanks.

Lunderberg (approved):

vinx13 (approved):

PR #13470, opened 2022-11-22
[MetaSchedule] Use current pass context in compile_relay, extract_tasks

tkonolige:
Adds the pass config information necessary for tuning and compiling relay with metaschedule to the existing pass context instead of overriding the existing one. Allows users to pass in their own pass instruments, required passes, and disabled passes. This also keeps the same API used to compile relay with autotvm and auto_scheduler.

### DIFF
--- a/python/tvm/meta_schedule/relay_integration.py
+++ b/python/tvm/meta_schedule/relay_integration.py
@@ -16,14 +16,14 @@
 # under the License.
 """MetaSchedule-Relay integration"""
 from contextlib import contextmanager
-from types import MappingProxyType
-from typing import TYPE_CHECKING, Any, Dict, List, Mapping, Optional, Tuple, Union, Set
+from typing import TYPE_CHECKING, Dict, List, Optional, Tuple, Union
 
 # isort: off
 from typing_extensions import Literal
 
 # isort: on
 import numpy as np  # type: ignore
+
 from tvm import nd
 from tvm._ffi import get_global_func
 from tvm.ir import IRModule, transform
--- a/python/tvm/meta_schedule/relay_integration.py
+++ b/python/tvm/meta_schedule/relay_integration.py
@@ -71,15 +71,8 @@ def _normalize_params(
     mod: IRModule,
     target: Union[Target, str],
     params: Optional[Dict[str, NDArray]],
-    pass_config: Mapping[str, Any],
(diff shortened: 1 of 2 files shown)


tkonolige:
@junrushao @masahi any thoughts on this? The change is from

```
ms.compile_relay(mod, opt_level=3)
```

to

```
with transform.PassContext(opt_level=3):
  ms.compile_relay(mod)
```

Although more verbose this mirrors the compilation process for non-metascheduled relay:

```
with transform.PassContext(opt_level=3):
  vm.compile(mod)
```

junrushao:
They are intentionally designed this way to keep the primary APIs as simple as possible, for advanced usecases like you mentioned in this PR, please do not use `compile_relay` and `extract_tasks` directly, but assemble those APIs call yourself. That is being said, I am not in favor of this change

masahi:
@junrushao Isn't it actually simplifying the API? Also considering that this changes makes the MS API consistent with other tuner APIs, I think it is a good change.

masahi:
This also makes PRs like https://github.com/apache/tvm/pull/13688 and https://github.com/apache/tvm/pull/13659 unnecessary. 

If we want to make the tuning API as simple as possible, we should remove all those default arguments (builder / runner / db etc, and even random seeds).

python/tvm/meta_schedule/relay_integration.py
```diff
@@ -390,16 +353,20 @@ def compile_relay(
     from tvm import relay
 
     # pylint: enable=import-outside-toplevel
-    mod, target, params, pass_config, executor = _normalize_params(
-        mod, target, params, pass_config, executor
-    )
+    mod, target, params, executor = _normalize_params(mod, target, params, executor)
+    pass_ctx = transform.PassContext.current()
+    pass_config = dict(pass_ctx.config)
```
masahi, line 358: To prevent breaking user's script after this change, how about setting a default ctx with opt_level = 3 if there is no current ctx?

tkonolige: Right now there is no way to determine whether or not if the current pass context is the default or not. It might be possible to add though.

I is not consistent with the rest of TVM to default the opt_level to 3 in meta schedule. Everywhere else defaults to 2. This does mean that we might break some user scripts.

Issue #13466, opened 2022-11-22
[ONNX][Relay][QNN] tvm doesn't support mix-precision inputs for qnn matmul


vvchernov:
### Expected behavior
No check for types matching is expected for qnn matmul inputs due to there is no constraint for QLinearMatMul operation in [ONNX documentation](https://github.com/onnx/onnx/blob/main/docs/Operators.md#QLinearMatMul)

### Actual behavior
Quantized [model](https://huggingface.co/philschmid/quantized-distilbert-banking77?text=I+like+you.+I+love+you) from Hugging face failed during compilation due to [matching](https://github.com/apache/tvm/blob/main/python/tvm/relay/frontend/onnx.py#L4774) of input tensor types. 

### Environment
Linux 20.04 LTE

### Steps to reproduce
Usual step for compilation and launching of the onnx-model by VirtualMachine by python front-end

### Triage
* frontend:onnx
* relay:qnn

### Notes
There is similar [issue](https://github.com/apache/tvm/issues/7258) and [discuss](https://discuss.tvm.apache.org/t/tvm-doesnt-support-mix-precision-inputs-for-qnn-conv2d/8850) for qnn conv2d. Unfortunately there is no any solution or minimal discussion. It looks like the problem is more general: tvm relays matches input types for qnn operations, but it does not assume by ONNX op description.

cc @KJlaccHoeUM9l @ehsanmok

vvchernov:
cc @masahi

PR #13469, opened 2022-11-22
[QNN] support zero points as variable scalar for QnnBatchMatMul op

vvchernov:
There is fix of [the issue](https://github.com/apache/tvm/issues/13467)

cc @ibsidorenko @apeskov

PR #13472, opened 2022-11-22
[QNN][ONNX][Relay] support mix-precision input types for QLinearMatMulfix

vvchernov:
There is fix of issue #13466. CI tests prepared earlier for this case were uncommented.

cc @masahi


AndrewZhaoLuo (commented):
LGTM, but also looks like ONNXRT may not support qlinear matmul with mixed dtypes, so might need to change up testing

vvchernov:
Hello @AndrewZhaoLuo! I've done this fix to support quantized distilbert-like model with mixed-precision in QLinearMatMut. The model is supported by ONNX RT. After this fix and fix from #13469 the model is supported by TVM with VirtualMachine. Nevertheless you are right the primitive op with mix-precision does not work on ORT and more over for TVM too. I'm working on it.

PR #13475, opened 2022-11-23
[ci] Don't merge PRs with main

### LINKED ISSUE #12657: [ci] Don't use latest `main` for PRs
Here we find the latest commit and merge with it before running a PR through CI to prevent merge conflicted PRs from taking up CI time.

https://github.com/apache/tvm/blob/caf326fab2963dac8fe03d266ea33d323f4b4470/ci/jenkins/Prepare.groovy.j2#L17-L26

This is fine except when we change the `Jenkinsfile` that always comes from the PR's `git merge-base origin/main <pr_commit>`, so if that references files that have been moved in the latest commit on `main` there will be random CI failures like https://ci.tlcpack.ai/blue/organizations/jenkins/tvm/detail/PR-12648/3/pipeline/. To fix this we should either:

1. don't merge at all and run the PR as is
2. find the last 'safe' commit (i.e. the last commit between `origin/main` and the `git merge-base origin/main <pr_commit>` that has no `Jenkinsfile` modifications) and merge to that

cc @Mousius @areusch @gigiblender

driazati:
This adds some CI load by running PRs that would merge-conflict with
main but improves UX by making it so CI is deterministic based on git
history rather than when the PR happened to be submitted.

Fixes #12657

Announced to discuss in https://discuss.tvm.apache.org/t/ci-disabling-merge-to-main-at-runtime/14075

### DIFF
--- a/ci/jenkins/generated/arm_jenkinsfile.groovy
+++ b/ci/jenkins/generated/arm_jenkinsfile.groovy
@@ -60,7 +60,7 @@
 // 'python3 jenkins/generate.py'
 // Note: This timestamp is here to ensure that updates to the Jenkinsfile are
 // always rebased on main before merging:
-// Generated at 2022-12-06T20:56:42.365592
+// Generated at 2022-12-08T16:10:54.358733
 
 import org.jenkinsci.plugins.pipeline.modeldefinition.Utils
 // These are set at runtime from data in ci/jenkins/docker-images.yml, update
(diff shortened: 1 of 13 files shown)


gigiblender (approved):
LGTM

Mousius:
@driazati sorry, this also needs a rebase 😿

driazati:
Some more context / notes around this from offline discussion with @areusch and @tqchen:

* Right now merging with main prevents some issues around merge-racing PRs (e.g. a big refactor merges but a PR is put up based on an older commit and uses the old APIs), if these happen with any frequency we should just revert this change
* We should integrate into our reporting better an explanation of what CI is doing with the code it runs (i.e. print out the commit to merge with before running unit tests to reproduce what CI actually did) if we end up not keeping this PR

driazati:
Closing for now, we may revisit this later if these kinds of problems are more frequent

PR #13485, opened 2022-11-24
[TIR] Correct type annotation for `rfactor`

junrushao:
The return type should be BlockRV rather than LoopRV


MasterJH5574 (approved):
Thanks @junrushao for catching this!

PR #13486, opened 2022-11-25
[DOCKER][ADRENO] we don't need microtvm being built for android cross…

srkreddy1238:
… compilation

PR:https://github.com/apache/tvm/pull/13073 makes microtvm built by default with x86 compilers. Cross compilation fails due to this.

### DIFF
--- a/tests/scripts/task_build_adreno_bins.sh
+++ b/tests/scripts/task_build_adreno_bins.sh
@@ -28,6 +28,7 @@ cd ${output_directory}
 
 cp ../cmake/config.cmake .
 
+echo set\(USE_MICRO OFF\) >> config.cmake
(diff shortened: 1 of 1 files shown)


echuraev (approved):

mehrdadh:
@srkreddy1238 @echuraev Is it possible to add this cross compilation case to TVM CI so we can catch issues like this in pre-commit? We could have it at least in the build stage, basically cross compile android on an x86 docker image. wdyt?

echuraev:
@mehrdadh I think it is a good idea. I'm not sure if the Adreno docker is running in the CI or not. Maybe @srkreddy1238 might know better than me. Probably, we could add this docker image to CI.

PR #13487, opened 2022-11-25
[ACL] Enable int8 data type in QNN DENSE

leandron:
This enables `QNN_DENSE` int8 data type to be used in Compute Library for the Arm(r) Architecture (ACL) BYOC integration.

cc @lhutton1 for reviews

### DIFF
--- a/tests/python/contrib/test_arm_compute_lib/test_dense.py
+++ b/tests/python/contrib/test_arm_compute_lib/test_dense.py
@@ -15,8 +15,8 @@
 # specific language governing permissions and limitations
 # under the License.
 """Arm Compute Library integration dense tests."""
-
 import numpy as np
+import pytest
 
 import tvm
 from tvm import relay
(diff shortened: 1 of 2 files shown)


tests/python/contrib/test_arm_compute_lib/test_dense.py
```diff
@@ -251,7 +252,14 @@ def test_codegen_dense():
...
 
 
-def test_qnn_dense():
+@pytest.mark.parametrize(
+    "dtype,min_range,max_range",
+    [
+        ("uint8", 0, 255),
+        ("int8", -127, 128),
+    ],
+)
```
ashutosh-arm, line 261: For future usage, could we define a function either with acl tests / under tvm testing that would provide the ranges?
Similar to https://github.com/apache/tvm/blob/2105b937cd6b4f426156813d6562f2845e5604f0/tests/python/contrib/test_cmsisnn/utils.py#L89

ashutosh-arm (commented):
It is pleasantly surprising that the codegen does not require any changes when supporting a new dtype. It shows that the ACL APIs and their integration into TVM is seamless :rocket:

These changes look good to me modulo a minor comment.

tests/python/contrib/test_arm_compute_lib/test_dense.py
leandron: That's a good idea, I can start moving that direction when implementing the next operators.

lhutton1: +1, I intended to do this for NPU testing as well but never got around to it, perhaps we can pull it out into a bigger refactoring?

lhutton1 (approved):
LGTM!

tests/python/contrib/test_arm_compute_lib/test_dense.py
ashutosh-arm: Yes, moving this into TVM testing area would be more beneficial.

PR #13489, opened 2022-11-26
[RUNTIME] Correctly handling export_module when exporting modules of different type

jinhongyii:
Previously the `export_module` function assumes that the module and all imported modules have the same `type_key`. If we have a host llvm module and some imported c modules, it may fail to set the `llvm_target_string` and `object_format` correctly, and thus fail to continue with the llvm build flow.

In this PR, I set the object format of a mixed-type module to that of the host module, and always set `llvm_target_string` when we see a host llvm module.

cc: @tqchen @YuchenJin

### DIFF
--- a/tests/python/unittest/test_runtime_module_export.py
+++ b/tests/python/unittest/test_runtime_module_export.py
@@ -21,6 +21,7 @@ from tvm import te
 import tvm.testing
 
 from tvm.contrib import utils
+import os
 
 header_file_dir_path = utils.tempdir()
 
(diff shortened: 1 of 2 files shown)


tqchen:
Thanks hongyi. This is a hidden issue as previous behavior will still go through the c export route, and not the llvm one. Which is still correct for small module size. But it will cost long export time for bigger models. cc @jwfromm

tests/python/unittest/test_runtime_module_export.py
```diff
@@ -203,7 +204,10 @@ def verify_multi_c_mod_export():
         synthetic_cpu_lib.import_module(f)
         synthetic_cpu_lib.import_module(engine_module)
         kwargs = {"options": ["-O2", "-std=c++17", "-I" + header_file_dir_path.relpath("")]}
-        synthetic_cpu_lib.export_library(path_lib, fcompile=False, **kwargs)
+        work_dir = temp.relpath("work_dir")
+        os.mkdir(work_dir)
+        synthetic_cpu_lib.export_library(path_lib, fcompile=False, workspace_dir=work_dir, **kwargs)
+        assert os.path.exists(work_dir + "/devc.o")
```
tqchen, line 210: use 

`os.path.join(work_dir, "devc.o")` so it can be portable across OS with different separators

tqchen (changes requested):

tqchen (approved):

PR #13490, opened 2022-11-26
[Relay][Pattern] Enable rewrite_once in class:DFPatternRewrite

sisleyli:
In the original `DFPatternRewrite` class, the default is to rewrite multiple times, I think it is necessary for the user to set whether to run the callback once.

### DIFF
--- a/src/relay/transforms/simplify_expr.h
+++ b/src/relay/transforms/simplify_expr.h
@@ -54,7 +54,7 @@ class DFPatternRewrite {
       Map<DFPattern, Array<Expr>> node_map = args[2];
       *rv = this->Callback(pre, post, node_map);
     };
-    return DFPatternCallback(pattern_, PackedFunc(func), require_type_);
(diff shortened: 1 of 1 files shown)


sisleyli:
Just a small change, but I think it is useful:   
```cpp
  DFPatternRewriteComposer composer;
  composer.AddRewrite<Once>();
  composer.AddRewrite<Once>();
  composer.AddRewrite<Multiple>();
  return RewritePatterns(composer.MakeCallbacks(), expr, mod);
```
What do you think? cc @masahi

masahi:
How would a user initialize this variable if he / she wants to set it to true?

sisleyli:
> How would a user initialize this variable if he / she wants to set it to true?

In general, the user inherits the class and assigns a value to the pattern_ in the constructor(we can see the code in `src/relay/transforms/simplify_expr.cc`), so I think it is enough to set true for this variable at the same time.

masahi (approved):

PR #13494, opened 2022-11-27
[Apps] add bundle deploy demo for Windows

LoSealL:
This patch creates a CMake file and ports some codes for Windows to help it can be built and tested under the Windows system with MSVC compiler.
For now only dynamic demo is supported on Windows.


LoSealL:
@MarisaKirisame @junrushao Manually @-ing possiblely related reviewers since bot didn't assign one.

PR #13498, opened 2022-11-28
[CMSIS-NN] Support int16 handling for pooling functions

neildhickey:
Support for int16 pooling functions via CMSIS-NN

-Pattern matching and RelayToTIR introduce int16 support 
-Added int16 variants to fully_connected tests

### DIFF
--- a/python/tvm/relay/op/contrib/cmsisnn.py
+++ b/python/tvm/relay/op/contrib/cmsisnn.py
@@ -287,8 +287,8 @@ def pattern_table():
         return (
             pooling.attrs.layout == "NHWC"
             and int(input_op.checked_type.shape[0]) == 1
-            and input_op.checked_type.dtype == "int8"
-            and output.checked_type.dtype == "int8"
+            and (input_op.checked_type.dtype == "int8" or input_op.checked_type.dtype == "int16")
+            and (output.checked_type.dtype == "int8" or output.checked_type.dtype == "int16")
         )
 
     def qnn_max_pool2d_pattern():
@@ -310,8 +310,8 @@ def pattern_table():
         return (
             pooling.attrs.layout == "NHWC"
             and int(input_op.checked_type.shape[0]) == 1
-            and input_op.checked_type.dtype == "int8"
-            and output.checked_type.dtype == "int8"
+            and (input_op.checked_type.dtype == "int8" or input_op.checked_type.dtype == "int16")
+            and (output.checked_type.dtype == "int8" or output.checked_type.dtype == "int16")
         )
 
     def binary_op_pattern(op):
--- a/python/tvm/relay/op/contrib/cmsisnn.py
+++ b/python/tvm/relay/op/contrib/cmsisnn.py
@@ -287,8 +287,10 @@ def pattern_table():
         return (
             pooling.attrs.layout == "NHWC"
             and int(input_op.checked_type.shape[0]) == 1
-            and (input_op.checked_type.dtype == "int8" or input_op.checked_type.dtype == "int16")
-            and (output.checked_type.dtype == "int8" or output.checked_type.dtype == "int16")
+            and (
+                (input_op.checked_type.dtype == "int8" and output.checked_type.dtype == "int8")
+                or (input_op.checked_type.dtype == "int16" and output.checked_type.dtype == "int16")
+            )
         )
 
     def qnn_max_pool2d_pattern():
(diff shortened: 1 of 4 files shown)


python/tvm/relay/op/contrib/cmsisnn.py
```diff
@@ -287,8 +287,8 @@ def check_qnn_avg_pool2d(pattern):
         return (
             pooling.attrs.layout == "NHWC"
             and int(input_op.checked_type.shape[0]) == 1
-            and input_op.checked_type.dtype == "int8"
-            and output.checked_type.dtype == "int8"
+            and (input_op.checked_type.dtype == "int8" or input_op.checked_type.dtype == "int16")
+            and (output.checked_type.dtype == "int8" or output.checked_type.dtype == "int16")
```
leandron, line 291: Minor suggestion: a slightly more pythonic comparison.

```suggestion
            and (input_op.checked_type.dtype in ("int8", "int16"))
            and (output.checked_type.dtype in ("int8", "int16"))
```

leandron (commented):
One suggestion for pythonic idioms, apart from that all good.

python/tvm/relay/op/contrib/cmsisnn.py
leandron: Just for completeness, there is a similar one in `qnn_max_pool2d_pattern`.

python/tvm/relay/op/contrib/cmsisnn.py
```diff
@@ -310,8 +310,8 @@ def check_qnn_max_pool2d(pattern):
...
```
ashutosh-arm, line 314: Consider the following check for better filtering:
```
(input_op.checked_type.dtype == "int8" and output.checked_type.dtype == "int8" ) or
(input_op.checked_type.dtype == "int16" and output.checked_type.dtype == "int16")
```

src/relay/backend/contrib/cmsisnn/relay_to_tir.cc
```diff
@@ -456,8 +467,13 @@ class RelayToTIRVisitor : public MixedModeMutator {
...
       clip_max = clip_attrs->a_max;
     } else {
-      clip_min = -128;
-      clip_max = 127;
+      if (dtype_bits == 8) {
+        clip_min = std::numeric_limits<int8_t>::min();
+        clip_max = std::numeric_limits<int8_t>::max();
+      } else {
+        clip_min = std::numeric_limits<int16_t>::min();
+        clip_max = std::numeric_limits<int16_t>::max();
```
ashutosh-arm, line 475: This looks so much better :smile:

tests/python/contrib/test_cmsisnn/test_pooling.py
```diff
@@ -103,18 +105,17 @@ def test_op_int8(
     compiler_cpu,
     cpu_flags,
 ):
-    """Tests QNN pooling op for int8 inputs"""
+    """Tests QNN pooling op for int8 and int16 inputs"""
```
ashutosh-arm, line 108: At present, we support int8, int16 input and output for the pool. Instead of `input`, we could say `pooling`. This also suggests that we should have a negative test where dtypes of the input and output to a pool op sequence do not match. For such cases, the CMSIS-NN partitioner should fail. To be clear, (int8, int16)/(int16,int8) kind of combinations should get mapped to TVM inplace of CMSIS-NN.

src/relay/backend/contrib/cmsisnn/tir_to_runtime.cc
```diff
@@ -118,7 +118,8 @@ class CodeGenCMSISNN : public codegen::CodeGenCHost {
     } else if (cmsis_func_name == "arm_fully_connected_s8" ||
                cmsis_func_name == "arm_fully_connected_s16") {
       EmitFullyConnected(op);
-    } else if (cmsis_func_name == "arm_avgpool_s8" || cmsis_func_name == "arm_max_pool_s8") {
+    } else if (cmsis_func_name == "arm_avgpool_s8" || cmsis_func_name == "arm_avgpool_s16" ||
+               cmsis_func_name == "arm_max_pool_s8" || cmsis_func_name == "arm_max_pool_s16") {
```
ashutosh-arm, line 122: I wish C++ had a single line of looking up into a list the way python allows :crying_cat_face:

ashutosh-arm (changes requested):
Thanks @neildhickey for the swift enabling of new dtype. I have added few comments below.

tests/python/contrib/test_cmsisnn/test_pooling.py
neildhickey: I tried to add a negative test for this but it wasn't possible to produce a model that fell into this condition, particularly in the case of max_pool. In the end I think it is okay to leave the check for this state in cmsisnn.py but I don't see an easy way to test this directly.

ashutosh-arm (approved):
Thanks @neildhickey for the changes. LGTM!

PR #13499, opened 2022-11-28
[Docs][Bug] Fix broken link to tvmc python

bkmgit:
Ensure correct link is generated to the next page in the documentation.

### DIFF
--- a/gallery/tutorial/tvmc_command_line_driver.py
+++ b/gallery/tutorial/tvmc_command_line_driver.py
@@ -518,6 +518,8 @@ testing.utils.install_request_hook(depth=3)
 # To see what other options are available, please have a look at ``tvmc
 # --help``.
 #
-# In the next tutorial, `Compiling and Optimizing a Model with the Python
-# Interface <auto_tuning_with_pyton>`_, we will cover the same compilation
-# and optimization steps using the Python interface.
+# In the `next tutorial <tvmc_python>`, we introduce the Python interface to TVM,
+# and in the tutorial after that,
+# `Compiling and Optimizing a Model with the Python Interface <autotvm_relay_x86>`,
+# we will cover the same compilation and optimization steps using the Python
+# interface.


leandron (approved):
LGTM, thanks @bkmgit

gromero:
@leandron Shouldn't that be pointing to https://github.com/apache/tvm/blob/main/gallery/tutorial/autotvm_relay_x86.py#L18 instead?

bkmgit:
@gromero Thanks for your feedback.  That is not the next tutorial in the sequence obtained at the [bottom of the page](https://tvm.apache.org/docs/tutorial/tvmc_command_line_driver.html) . The link you have suggested could be used but then it would be helpful to rephrase 
```
In the next tutorial, [Compiling and Optimizing a Model with the Python Interface](https://tvm.apache.org/docs/tutorial/auto_tuning_with_pyton), we will cover the same compilation and optimization steps using the Python interface.
```
to
```
In the [next tutorial](https://tvm.apache.org/docs/tutorial/tvmc_python.html), we introduce the Python interface to TVM, and in the tutorial after that, [Compiling and Optimizing a Model with the Python Interface](https://tvm.apache.org/docs/tutorial/autotvm_relay_x86), we will cover the same compilation and optimization steps using the Python interface.
```

gromero:
> @gromero Thanks for your feedback. That is not the next tutorial in the sequence obtained at the [bottom of the page](https://tvm.apache.org/docs/tutorial/tvmc_command_line_driver.html) . The link you have suggested could be used but then it would be helpful to rephrase

@bkmgit Got it!

> In the [next tutorial](https://tvm.apache.org/docs/tutorial/tvmc_python.html), we introduce the Python interface to TVM, and in the tutorial after that, [Compiling and Optimizing a Model with the Python Interface](https://tvm.apache.org/docs/tutorial/autotvm_relay_x86), we will cover the same compilation and optimization steps using the Python interface.

yeah I'm ok with that version.

Another option would be to just adjusting the text in your change, i.e. in `tvmc_command_line_driver.py`, to match the new link, so (remove the wrong tutorial title text):

"In the next tutorial, [Getting Starting using TVMC Python: a high-level API for TVM](https://tvm.apache.org/docs/tutorial/tvmc_python.html), we introduce the Python interface to TVM."

and move the second part to the bottom of tutorial [tvmc_python.py](https://tvm.apache.org/docs/tutorial/tvmc_python.html).

"In the next tutorial, [Compiling and Optimizing a Model with the Python Interface](https://tvm.apache.org/docs/tutorial/autotvm_relay_x86), we will cover the same compilation and optimization steps using the Python interface." 

But either way works for me.

gromero (changes requested):
@bkmgit The important thing here I think is to keep the text matching the fixed link, either by updating the text to match the `tvmc_pyton.py` tutorial title or by removing the stale text that became outdated by the new (fixed) link.

bkmgit:
@gromero  Thanks for the feedback. Done.

gromero (approved):
@bkmgit Thanks a lot for the new version, and specially for being careful updating the commit message accordingly , and also for taking care of the trailing spaces! ;-)

gromero:
@leandron Since there are changes since you're last review, would you like to take another look? Thanks!

leandron (approved):
LGTM

PR #13500, opened 2022-11-28
Add close method to RPC session

itmo153277:
Add close method to `RPCSession` and make sure we call it when we finish using it.

For more context, see [this discussion](https://discuss.tvm.apache.org/t/trouble-with-rpc-session/14008/1)


itmo153277:
@eqy @echuraev Please review

echuraev (commented):
Could you please also update tests and documentation which are using `request_remote`, `rpc.connect` etc.?

PR #13511, opened 2022-11-29
Add methods to get the size of VTCM on device as well as the allocated size of the HexagonVtcmPool

janetsc:
Expose methods so that callers can get both the size of VTCM on device as well as the allocated size of the pool for the session.

cc: @masahi


masahi:
Can we expose the API to python too?

masahi (approved):

PR #13515, opened 2022-11-29
[LLVM] Switch to using New Pass Manager (NPM) with LLVM 16+

kparzysz-quic:
LLVM 16 has removed support for the legacy pass manager, and the LLVM codegen in TVM will no longer even compile.


kparzysz-quic:
cc: @areusch @masahi

masahi (approved):

PR #13523, opened 2022-11-30
[CI] Add `set -x` for demo scripts

Lunderberg:
Previously, the `tests/scripts/task_demo_microtvm.sh` script prints the commands as they are run, but the
`apps/microtvm/ethosu/run_demo.sh` script did not.  As a result, it could be difficult to tell which step stalled or failed during CI.


gromero (approved):
@Lunderberg LGTM, thanks.

Issue #13508, opened 2022-11-29
[Bug] Long lowering time after #13217


comaniac:
### Expected behavior

The lowering time of the given case should be around 10 seconds.

### Actual behavior

The lowering time is more than 550 seconds.

### Environment

Any environment with commit commit 101e3a4ade226a2b9cdef6437a285af18aef9cf8 (#13217) or later.

### Steps to reproduce

The script:

```python
import time

import tvm
from tvm import topi

class Timer:
    def __init__(self, msg):
        self.msg = msg
        print(f"{msg}...", flush=True)

    def __enter__(self):
        self.start = time.time()

    def __exit__(self, *args):
        print(f"{self.msg}...{time.time() - self.start:.2f}s", flush=True)

def resize2d_dx_compute(inp, dy):
    """compute definition for resize2d_dx op"""
    size = (64, 32)
    layout = "NCHW"
    method = "cubic"
    coord_trans = "half_pixel"
    rounding_method = ""
    cubic_alpha = -0.75
    cubic_exclude = 0
    out_dtype = "float32"

    out = topi.image.resize2d(
        inp,
        (None, None, None, None),
        size,
        layout,
        method,
        coord_trans,
        rounding_method,
        bicubic_alpha=cubic_alpha,
        bicubic_exclude=cubic_exclude,
        out_dtype=out_dtype,
    )
    grads = tvm.te.gradient(out, [inp], head=dy)
    return grads

inp = tvm.te.placeholder((32, 3, 32, 32), name="inp")
dy = tvm.te.placeholder((32, 3, 64, 32), name="dy")
with Timer("te.gradient"):
    grads = resize2d_dx_compute(inp, dy)

# This problem is platform-independent.
with Timer("schedule"):
    sch = topi.x86.injective.schedule_injective(grads)

with Timer("lower"):
    print(tvm.lower(sch, [inp, dy, grads[0]], simple_mode=True))
```

1. Switch to a commit before 101e3a4ade226a2b9cdef6437a285af18aef9cf8 (#13217) and run the script.
2. Checkout the commit 101e3a4ade226a2b9cdef6437a285af18aef9cf8 (#13217) and run again.

Here are also the lowered IR without and with this commit:

Without this commit:
```
@main = primfn(inp_1: handle, dy_1: handle, resize.inp.grad_1: handle) -> ()
  attr = {"from_legacy_te_schedule": True, "global_symbol": "main", "tir.noalias": True}
[...]

PR #13524, opened 2022-11-30
[TIR] Disable RewriteSimplifier extensions in RemoveNoOp

Lunderberg:
During a `tir.Simplify` pass, these extensions were conditionally enabled based on the `PassContext`.  Prior to this commit, they were enabled by default in the `tir.RemoveNoOp` pass, as the simplified expressions were only used to prove/disprove a no-op, and did not appear in the output TIR.  However, this caused performance issues for some nested boolean expressions.

This PR disables the analyzer extensions for the analyzer used by `tir.RemoveNoOp`.  The extensions are still used internally by `ControlFlowGraph`, including during the data-flow analysis used if `tir.transform.RemoveNoOpConfig.use_dataflow_analysis` is enabled, so the opt-in data-dependent no-op removals are unaffected.

Related to issue #13508.

### DIFF
--- a/src/tir/transforms/remove_no_op.cc
+++ b/src/tir/transforms/remove_no_op.cc
@@ -306,10 +306,6 @@ Pass RemoveNoOp() {
     }
 
     arith::Analyzer analyzer;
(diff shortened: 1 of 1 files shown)


comaniac (approved):
LGTM

PR #13534, opened 2022-12-02
[TIR] Introduce Pass UnifyKernelLaunch and speedup split op

Hzfengsy:
This PR introduces a new pass UnifyKernelLaunch to fuse the nearby kernels with same thread config, which will reduce the kernel launch overhead in `split` operator.

Test on the following settings:

```python
A = te.placeholder((75, 182), "int64")
B = topi.split(A, 182, axis=1)

func = te.create_prim_func([A, *B])
```

The pass will reduce the latency from 347.03 us to 5.22 us on my machine.

cc @MasterJH5574 @vinx13 @junrushao


Hzfengsy:
@tvm-bot rerun

src/meta_schedule/schedule/cuda/split.cc
```diff
@@ -0,0 +1,34 @@
...
+namespace meta_schedule {
+
+using namespace tvm::tir;
+
+TVM_REGISTER_GLOBAL("meta_schedule.cuda.split")
+    .set_body_typed([](Schedule sch, BlockRV data_pack) -> Array<Schedule> {
+      // Do nothing, which ensure the same thread config for each block.
+      // We can fuse these kernels into only one.
+      return {sch};
+    });
```
junrushao, line 31: if we want to skip a block, use {"schedule_rule": "None"} instead

Hzfengsy: We only want to skip the block for GPU backend, but still will apply schedule on other targets.

junrushao: perhaps we should support per-device None in meta schedule :-)

vinx13 (approved):

PR #13535, opened 2022-12-02
[MetaSchedule] Make `MultiLevelTiling` apply condition customizable

masahi:
Currently, the condition for applying `MultiLevelTiling` is given by `NeedsMultiLevelTiling` function:
https://github.com/apache/tvm/blob/5364e5a39a5e33728b7f5a26ddb40543a544ea02/src/meta_schedule/schedule_rule/multi_level_tiling.cc#L95

In practice, only multi-argument reduction operations like conv2d or dense are selected. I'm proposing to make this condition customizable for use cases where we might want to apply `MultiLevelTiling` to more ops (pooling etc).

@junrushao @csullivan @vinx13


vinx13 (approved):

csullivan (approved):
Thanks @masahi, LGTM.

PR #13538, opened 2022-12-02
[MetaSchedule][Hexagon] Add postproc for verifying VTCM usage

masahi:
Building on https://github.com/apache/tvm/pull/13349, and similar to `VerifyGPUCode`, add a postproc check for the VTCM usage.

@csullivan @Icemist @nverke

### DIFF
--- a/src/tir/analysis/calculate_allocated_memory.cc
+++ b/src/tir/analysis/calculate_allocated_memory.cc
@@ -87,6 +87,15 @@ TVM_REGISTER_GLOBAL("tir.analysis.calculate_allocated_bytes").set_body_typed([](
   return CalculateAllocatedBytes(func);
 });
 
+bool VerifyVTCMLimit(const PrimFunc& func, Integer limit) {
+  auto sizes = CalculateAllocatedBytes(func);
+  const auto vtcm_allocated = sizes.Get("global.vtcm").value_or(0);
+  if (limit.IntValue() > 0 && vtcm_allocated.IntValue() > limit.IntValue()) {
+    return false;
+  }
+  return true;
+}
+
 namespace transform {
 
 Pass VerifyVTCMLimit(const Integer& limit) {
--- a/tests/python/unittest/test_meta_schedule_postproc_verify_vtcm_limit.py
+++ b/tests/python/unittest/test_meta_schedule_postproc_verify_vtcm_limit.py
@@ -1,0 +1,127 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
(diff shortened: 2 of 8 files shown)


tests/python/unittest/test_meta_schedule_postproc_verify_vtcm_limit.py
```diff
@@ -0,0 +1,127 @@
...
+                                    T.writes(conv2d_NCHWc_int8[v_n, v_oc_chunk, v_oh, v_ow, v_oc_block_i])
+                                    T.block_attr({"meta_schedule.tiling_structure":"SRSRS"})
+                                    conv2d_NCHWc_int8[v_n, v_oc_chunk, v_oh, v_ow, v_oc_block_i] = conv2d_NCHWc_int8[v_n, v_oc_chunk, v_oh, v_ow, v_oc_block_i] + T.Cast("int32", p0_global_vtcm[v_n, v_ic_outer, v_oh + v_kh, v_ow + v_kw, v_ic_f_inner * T.int64(4) + v_ic_s_inner_i]) * T.Cast("int32", p1_global_vtcm[v_oc_chunk, v_ic_outer, v_kh, v_kw, v_ic_f_inner, v_oc_block_i, v_ic_s_inner_i])
+
+#fmt on
+
+
+def test_conv2d_vtcm():
+    def get_target(vtcm_cap):
+        target = tvm.target.hexagon("v68", vtcm_capacity=vtcm_cap)
```
Icemist, line 114: You can use `get_hexagon_target` (`from .infrastructure import get_hexagon_target`) instead of writing your own version

src/tir/analysis/calculate_allocated_memory.cc
```diff
@@ -87,6 +87,15 @@ TVM_REGISTER_GLOBAL("tir.analysis.calculate_allocated_bytes").set_body_typed([](
   return CalculateAllocatedBytes(func);
 });
 
+bool VerifyVTCMLimit(const PrimFunc& func, Integer limit) {
```
Icemist, line 90: looks like `transform::VerifyVTCMLimit` also uses this logic. Can we reuse this function inside that one?

tests/python/unittest/test_meta_schedule_postproc_verify_vtcm_limit.py
masahi: I don't want to depend on a file from `contrib/test_hexagon` from this file.

src/tir/analysis/calculate_allocated_memory.cc
masahi: Yeah I wanted to do that, but the pass variant needs `vtcm_allocated` in its error message, while this one wants to return only true / false. Refactoring is possible but I thought just duplicating might be cleaner in this case.

Icemist:
LGTM

Icemist (approved):

csullivan (approved):
Thanks @masahi @Icemist!

PR #13541, opened 2022-12-02
[ci][docker] Update NRF command line tools URL

driazati:
The old one began to 404, this one is taken for 10.12.1 from
https://www.nordicsemi.com/Products/Development-tools/nrf-command-line-tools/download


mehrdadh (approved):
Thanks for the fix!

PR #11310, opened 2022-05-13
[logging] LOG(FATAL) calls [[noreturn]] functions

cconvey:
Ensure that `LOG(FATAL)` always resolves to calling
`[[noreturn]]` code.  This has two benefits:

- Helps developers more quickly understand the
  intended/required behavior for LOG(FATAL) calls.

- May eliminate spurious compiler warnings related
  based on control-flow analysis.  E.g. gcc's / clang's
  `-Wno-return` warnings.

PR #13542, opened 2022-12-02
[Logging] Mark LogFatal::~LogFatal as [[noreturn]]

Lunderberg:
Using `LOG(FATAL) << "some error message";` throws an exception when the internal `LogFatal` object is constructed.  However, the gcc warning "control reaches end of non-void function" resulted in a common pattern where a `LOG(FATAL)` statement was immediately followed by return of a default value, in order to suppress the warning.

This commit marks the `LogFatal::~LogFatal` destructor with the C++11 attribute `[[noreturn]]`, indicating that calls into it will not resume the normal control flow.  This suppresses the warning without requiring the extra `return` statement after `LOG(FATAL)`, as the compiler knows that control flow will not resume after `LOG(FATAL)`.

### DIFF
--- a/apps/dso_plugin_module/plugin_module.cc
+++ b/apps/dso_plugin_module/plugin_module.cc
@@ -43,7 +43,6 @@ class MyModuleNode : public ModuleNode {
       return TypedPackedFunc<int(int)>([sptr_to_self, this](int value) { return value_ * value; });
     } else {
       LOG(FATAL) << "unknown function " << name;
-      return PackedFunc();
     }
   }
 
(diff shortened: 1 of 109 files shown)


Lunderberg:
After conversation with @kparzysz-quic , it looks like whether `[[noreturn]]` was previously applied.  https://github.com/apache/tvm/pull/11310 also added `[[noreturn]]`, but only when the `TVM_LOG_CUSTOMIZE` macro is defined.  This PR applies `[[noreturn]]` to the other branch as well.

kparzysz-quic (approved):

Hzfengsy (approved):
LGTM. It can be beneficial to the dev experience.

junrushao (approved):
LGTM! That helps a lot!

PR #13548, opened 2022-12-04
[Fix] Use proper target in VerifyGPUCode

MasterJH5574:
Previously, the VerifyGPUCode post-processor uses hardcoded target `Target("cuda")` for applying pass LowerIntrin. This is a bit problematic since the actual target can be other GPU target (e.g., Metal). Therefore, this PR changes the hardcoded target to be the actual target.

cc @junrushao @masahi


junrushao (approved):
LGTM! Note that we do not create Target on demand, but instead always rely on users' inputs.

PR #13554, opened 2022-12-06
[TIR] Fix layout free annotation in CreatePrimFunc

vinx13:
This PR fixed missing handling of block annotations in data type rewriter.

cc @junrushao


junrushao:
Hmm I don't have a clear idea of what the issue could be. Would you like to elaborate? Thanks in advance!

vinx13:
block annotations may contain buffers that are rewritten in this pass, but it forgot to visit block annotations

junrushao:
I see. That makes sense!

Hzfengsy (approved):

PR #13556, opened 2022-12-06
Mark base64.h encode and decode API's as inline

f2013519:
The b64strlen & b64decode API's are defined in a header and will cause linker issues if included in multiple files. We ran into this issue while attempting to use this file in our project.

### DIFF
--- a/src/contrib/torch/base64.h
+++ b/src/contrib/torch/base64.h
@@ -36,7 +36,7 @@
 namespace tvm {
 namespace support {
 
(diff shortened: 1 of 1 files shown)


f2013519:
@masahi Could you pls. take a look? Thanks

f2013519:
cc @masahi

junrushao (approved):
That makes sense. Thanks for the contribution!

masahi:
@f2013519 Good catch!

f2013519:
Thanks!

PR #13557, opened 2022-12-06
[TIR] [Hexagon] Add vdmpy intrinsic and transform_layout for tests

quic-sanirudh:
This patch adds the `vdmpy` hexagon intrinsic and a sample tensorization test for the same.

This patch also modifies the test to use transform_layout instead of a packed tensor in the compute to make it obvious that this example is just matmul with a different data layout for one of the inputs


quic-sanirudh:
@masahi @nverke Could you please take a look and review this patch when you get a chance.

masahi (approved):

nverke:
Looks great!

PR #13069, opened 2022-10-13
[TIR] Fix handling of int64 extent in blockize and tensorize

vinx13:
This fixed arithmetic analysis to ensure the result of subspace division has the same dtype as the original expression.

cc @junrushao @spectrometerHBH @Hzfengsy

PR #13558, opened 2022-12-06
[TIR] Fix remaining dtype mismatch issue caused by SubspaceDivide

masahi:
Apparently a left-over from https://github.com/apache/tvm/pull/13069. I have a complex test that would hit a dtype-mismatch error because of this, but I don't know how to create a simple repro test. Hopefully it is ok as this is a follow-up to 13069?

cc @vinx13


vinx13 (approved):
LGTM, I agree a test case is not needed in this case

PR #13563, opened 2022-12-06
[OpenCL]  Introduction of weights on buffers

valmat07:
This PR introduces the use of convolution weights on buffers rather than textures. 

- For now, the weights on the buffers are used only in the fp32 case since that case is investigated 
- Also, the case of 1x1 convolutions is excluded from this approach, since the weights on the buffers work worse in this case
- This approach gives about 10% performance boost for popular models, such as: vgg-16, yolov3, etc


src/relay/transforms/annotate_texture_storage.cc
```diff
@@ -177,6 +186,17 @@ class StorageInfo : private transform::DeviceAwareExprVisitor {
             // adding info about arguments if they can be converted to texture
             for (const auto& ttype : FlattenTupleType(fn->params[i]->checked_type())) {
               std::string scope = Scope(ttype->shape, GetVirtualDevice(GetRef<Expr>(call)));
+              if (expr_attrib.as<Conv2DAttrs>() || expr_attrib.as<Conv2DWinogradAttrs>()) 
+              {
+                if ((i == 1) &&
```
echuraev, line 191: nit: Create local variable, e.g.:  `const int weights_pos = 1;`.

echuraev (commented):
In general LGTM. Several minor comments

echuraev (approved):
LGTM. Thanks

PR #13565, opened 2022-12-06
[TIR] Fix predefined inverse map in layout transform dtype legalization

vinx13:
`LegalizeIndexMapDType` previously ignore pre-defined inverse map, which causes error in `TransformLayout` when the inverse index map is predefined but the arithmetic analysis fails to compute the inverse map.

cc @masahi @Lunderberg

### DIFF
--- a/src/tir/schedule/primitive/layout_transformation.cc
+++ b/src/tir/schedule/primitive/layout_transformation.cc
@@ -1056,16 +1056,16 @@ class TransformationIntroducesPaddingError : public ScheduleError {
 
 // Make the dtypes of indices in IndexMap be the same as the dtype of the buffer shape, to avoid
 // dtype-mismatch issues later.
-IndexMap LegalizeIndexMapDType(const IndexMap& index_map, const Buffer& buf) {
+IndexMap LegalizeIndexMapDType(const IndexMap& index_map, const Array<PrimExpr>& args) {
   const auto& initial_indices_orig = index_map->initial_indices;
-  ICHECK(buf->shape.size() == initial_indices_orig.size());
+  ICHECK(args.size() == initial_indices_orig.size());
 
   Array<Var> initial_indices;
   Map<Var, PrimExpr> var_map;
 
-  for (size_t i = 0; i < buf->shape.size(); ++i) {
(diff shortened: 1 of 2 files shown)


junrushao (approved):
LGTM!

src/tir/schedule/primitive/layout_transformation.cc
```diff
@@ -1079,7 +1079,13 @@ IndexMap LegalizeIndexMapDType(const IndexMap& index_map, const Buffer& buf) {
       return SubstituteWithDataTypeLegalization(index,
                                                 [&](const Var& var) { return var_map.Get(var); });
     });
-    return IndexMap(initial_indices, final_indices);
+    Optional<IndexMap> opt_inverse_index_map =
+        Downcast<Optional<IndexMap>>(index_map->inverse_index_map);
+    if (opt_inverse_index_map.defined()) {
+      IndexMap inverse_index_map = Downcast<IndexMap>(opt_inverse_index_map.value());
+      opt_inverse_index_map = LegalizeIndexMapDType(inverse_index_map, final_indices);
```
masahi, line 1086: Probably we can remove `Downcast<IndexMap>` at L1083 and L1085?

vinx13: updated, I'm keeping L1083 as `inverse_index_map` is defined as `Optional<ObjectRef>`

src/tir/schedule/primitive/layout_transformation.cc
```diff
@@ -1079,7 +1079,12 @@ IndexMap LegalizeIndexMapDType(const IndexMap& index_map, const Buffer& buf) {
       return SubstituteWithDataTypeLegalization(index,
                                                 [&](const Var& var) { return var_map.Get(var); });
     });
-    return IndexMap(initial_indices, final_indices);
+    Optional<IndexMap> opt_inverse_index_map =
+        Downcast<Optional<IndexMap>>(index_map->inverse_index_map);
+    if (opt_inverse_index_map.defined()) {
+      opt_inverse_index_map = LegalizeIndexMapDType(opt_inverse_index_map.value(), final_indices);
```
masahi, line 1085: Can we guarantee that this recursive call would always bottom out?

Issue #13187, opened 2022-10-25
[Feature Request][CI] Add tvm-bot command to only re-run failed tests


michalpiszczek:
We have `@tvm-bot rerun`. It would be helpful to also have `@tvm-bot rerun failed`. Sometimes it's useful to just rerun tests that have failed, for example if one is believed to be flakey. Committers can already do this manually because they have the Github permissions needed. It would be nice for all contributors to have this ability. 

cc @Mousius @areusch @driazati @gigiblender

PR #13567, opened 2022-12-06
[test][do not review] test ci

driazati:
test ci.


mehrdadh:
@tvm-bot rerun

driazati:
> @tvm-bot rerun

rerun is probably broken, it doesn't know about the new job names to trigger (and also #13187 is more possible now)

PR #13568, opened 2022-12-06
[MetaSchedule] Add decompose point hint for RewriteReductionBlock

vinx13:
This PR added an annotation `meta_schedule.decompose_point` as a hint for `RewriteReductionBlock`. When the outer reduction loop has extent of 1, it won't be recognized as reduction loops and `decompose_reduction` won't be used. In some cases like tensorization, the init block need to be tensorized. In this case, `decompose_reduction` is necessary and such hint is helpful.
As part of `MultiLevelTilingTensorCore`, the outermost reduction loop is annotated with such hint.

cc @junrushao @masahi


vinx13:
Closing this as #13579 is a better solution

PR #13572, opened 2022-12-07
[ci][docker] Read docker image tags during CI runs

driazati:
Hardcoding the Docker images in the Jenkinsfiles has the downside that only trusted changes are actually run in CI. Since Docker image updates are pretty frequent, this makes for a pretty bad UX (e.g. CI runs but doesn't actually test the PR). The fix here makes it so the images are read from a different file during CI execution so updates from anyone are picked up.

This PR eliminates the need to use the batch jobs in Jenkins to update Docker images. Docker images are built as part of post-merge CI on `main` and anyone can trigger a validation run by filing a PR with changed Docker tags.

### DIFF
--- a/ci/jenkins/generated/arm_jenkinsfile.groovy
+++ b/ci/jenkins/generated/arm_jenkinsfile.groovy
@@ -60,19 +60,21 @@
 // 'python3 jenkins/generate.py'
 // Note: This timestamp is here to ensure that updates to the Jenkinsfile are
 // always rebased on main before merging:
-// Generated at 2022-12-05T14:48:42.092397
+// Generated at 2022-12-06T20:56:42.365592
 
 import org.jenkinsci.plugins.pipeline.modeldefinition.Utils
-ci_lint = 'tlcpack/ci-lint:20221013-060115-61c9742ea'
-ci_gpu = 'tlcpack/ci-gpu:20221019-060125-0b4836739'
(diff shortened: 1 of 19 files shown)


ci/jenkins/generated/arm_jenkinsfile.groovy
```diff
@@ -60,19 +60,21 @@
...
+ci_lint = ''
+ci_gpu = ''
+ci_cpu = ''
+ci_minimal = ''
+ci_wasm = ''
+ci_i386 = ''
+ci_cortexm = ''
+ci_arm = ''
+ci_hexagon = ''
+ci_riscv = ''
```
echuraev, line 77: Do we really need declaring these variables in global scope? I see that they are used only in function `prepare`, so, probably they can be declared there.

echuraev (approved):
Thank you for nice improvement! LGTM. One minor question which should not affect on merging this PR.

ci/jenkins/generated/arm_jenkinsfile.groovy
driazati: This file is stiched to a bunch of others to make the generated output, and they all use the variables all over the place so they need to be accessible there e.g. https://github.com/apache/tvm/blob/main/ci/jenkins/generated/arm_jenkinsfile.groovy#L538

Mousius:
@driazati can you rebase this? 😸

PR #13576, opened 2022-12-07
[ETHOSN] Remove inference test

lhutton1:
This test was causing cpptest to fail without reporting the test as having failed. Looking back, this test doesn't really make much sense as we are passing a file descriptor stating that inference is running when it isn't. Therefore, removing the test.

cc @ekalda @ashutosh-arm


ekalda (approved):
LGTM, thanks for doing the digging @lhutton1!

ashutosh-arm (approved):
Sounds like a good idea to remove this case as the original test was meant to look for error codes in case of failures. Thanks @lhutton1.

PR #13583, opened 2022-12-08
[Hexagon] Lookup intrinsic by name instead of using enum value

kparzysz-quic:
Don't use numeric values for LLVM intrinsics.


kparzysz-quic:
@tvm-bot rerun

masahi (approved):

PR #13588, opened 2022-12-09
[microNPU] Disable copying weights to SRAM for FullyConnected ops in CopyConstants scheduler

Alexey-Yazev:
In Ethos-U, CopyConstants scheduler currently copies weights for all operators. But in Vela, there are a number of scenarios where the weights are not buffered in SRAM, and FullyConnected case is one of them.


lhutton1 (approved):
Thanks @Alexey-Yazev LGTM! Will leave for a bit to let others have a look also

lhutton1:
Thanks @Alexey-Yazev!

PR #13589, opened 2022-12-09
Simplify Consecutive Casts

AndrewZhaoLuo:
This PR adds a new rewrite to SimplifyExpr which removes extraneous casts. An extraneous cast is one defined where after the cast, the operand is the same type as before the cast operation. Previous passes may introduce extraneous casts (e.g. AlterOpLayout) so this may be helpful.

cc @guberti


leandron (changes requested):
This looks very interesting @AndrewZhaoLuo. Can you add more context to the PR description? on what it does, which cases, etc?

PR #13591, opened 2022-12-10
[LLVM] Fix get tm allow_missing check pos

wrongtest-intellif:
It seems that `GetOrCreateTargetMachine`'s failure check would be out of `lookupTarget`' nest scope.

### DIFF
--- a/src/target/llvm/llvm_instance.cc
+++ b/src/target/llvm/llvm_instance.cc
@@ -297,9 +297,9 @@ llvm::TargetMachine* LLVMTargetInfo::GetOrCreateTargetMachine(bool allow_missing
         llvm_instance->createTargetMachine(triple_, cpu_, GetTargetFeatureString(), target_options_,
                                            reloc_model_, code_model_, opt_level_);
     target_machine_ = std::unique_ptr<llvm::TargetMachine>(tm);
-    if (!allow_missing) {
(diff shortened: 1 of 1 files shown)


wrongtest-intellif:
cc @kparzysz-quic

echuraev (approved):
LGTM. Thanks

src/target/llvm/llvm_instance.cc
```diff
@@ -297,9 +297,9 @@ llvm::TargetMachine* LLVMTargetInfo::GetOrCreateTargetMachine(bool allow_missing
         llvm_instance->createTargetMachine(triple_, cpu_, GetTargetFeatureString(), target_options_,
                                            reloc_model_, code_model_, opt_level_);
     target_machine_ = std::unique_ptr<llvm::TargetMachine>(tm);
-    if (!allow_missing) {
```
kparzysz-quic, line 300: This is the only place where `target_machine_` is assigned a value, so this check was actually sufficient.

wrongtest-intellif: Emmmm but when `llvm::TargetRegistry::lookupTarget(triple_, error)` failed，the `target_machine_`  is also nullptr,

kparzysz-quic: Ah, you're right.

PR #13593, opened 2022-12-11
[TVMScript] Refactor IRDocsifier

junrushao:
This PR refactors the TVMScript printer and includes the following
changes:
- Consolidate the logics of VarTable into IRDocsifier
- Decouple TracedObject into Object and ObjectPath for less syntactic
  noise
- Restructure the folder to ensure logics and consistency

Some tests removed because the APIs do not exist any more due to the
consolidation.


include/tvm/script/printer/printer.h
```diff
@@ -0,0 +1,82 @@
...
+namespace printer {
+
+/*! \brief Default values in the TVMScript printer */
+struct Default {
+  /*! \brief Default data type of TIR buffer */
+  DataType buffer_dtype = DataType::Float(32);
+  /*! \brief Default data type of integer literals */
+  DataType int_dtype = DataType::Int(32);
+  /*! \brief Default data type of float literals */
+  DataType float_dtype = DataType::Void();
```
Hzfengsy, line 39: Why is it `DataType::Void()`?

src/script/printer/doc_printer/python_doc_printer.cc
```diff
@@ -294,7 +295,11 @@ void PythonDocPrinter::PrintTypedDoc(const LiteralDoc& doc) {
   } else if (const auto* float_imm = value.as<FloatImmNode>()) {
     // TODO(yelite): Make float number printing roundtrippable
```
Hzfengsy, line 296: Do you know if this TODO is resolved?

tests/scripts/task_mypy.sh
```diff
@@ -32,9 +32,6 @@ mypy  --check-untyped-defs python/tvm/tir/analysis/
 echo "Checking MyPy Type defs in the transform package."
 mypy  --check-untyped-defs python/tvm/tir/transform/
 
-echo "Checking MyPy Type defs in the tvmscript printer package."
-mypy  --check-untyped-defs python/tvm/script/printer
```
Hzfengsy, line 36: Please enable mypy checks

Hzfengsy (approved):
Overall LGTM

tests/scripts/task_mypy.sh
junrushao: sure!

include/tvm/script/printer/printer.h
junrushao: I looked up the current convention and it seems that we always print T.float32/float64 for floating point numbers, so this is to be consistent with the current style

src/script/printer/doc_printer/python_doc_printer.cc
junrushao: Not yet but indeed there is a way forward

include/tvm/script/printer/printer.h
Hzfengsy: Please remove it if it is not used. All at least add a comment for better readability.

junrushao: I do think that we could consider a default dtype for float in the near future

tqchen: I think we can change the comment to "Right now we always print out the explicit type of floating point values, so setting it to Void means we do not print without T.float32/T.float64 wrapper"

junrushao: that sounds good!

tqchen (approved):

tqchen:
Thanks @junrushao @Hzfengsy !

PR #13594, opened 2022-12-12
[Torch] Stable diffusion support

masahi:
Simple updates to support the stable diffusion model from diffusers. Scripts at https://github.com/masahi/torchscript-to-tvm/tree/master/stable-diffusion can be used to verify that the conversion works and the output of the pipeline using TVM-compiled modules looks ok (takes ~10 mins on CPU running for 50 steps, untuned).

@comaniac @shingjan @jroesch


comaniac (approved):
LGTM. Nice to know that we only miss one op in stable diffusion model.

PR #13598, opened 2022-12-12
[microTVM][Zephyr] Fix TVMC test on hardware

mehrdadh:
This PR adds stack size to run test on hardware for Zephyr. It is not added for Arduino since it is not required.
4096 is derived from experiment on nucleo_l4r5zi.

### DIFF
--- a/tests/micro/common/test_tvmc.py
+++ b/tests/micro/common/test_tvmc.py
@@ -44,6 +44,21 @@ def _run_tvmc(cmd_args: list, *args, **kwargs):
     return subprocess.check_call(cmd_args_list, *args, **kwargs)
 
 
+def create_project_command(project_path: str, mlf_path: str, platform: str, board: str) -> list:
+    """Returns create project command with tvmc micro."""
+    return [
+        "micro",
+        "create-project",
+        project_path,
+        mlf_path,
+        platform,
+        "--project-option",
+        "project_type=host_driven",
+        "config_main_stack_size=4096",
+        f"board={board}",
+    ]
+
+
 @tvm.testing.requires_micro
 def test_tvmc_exist(platform, board):
     cmd_result = _run_tvmc(["micro", "-h"])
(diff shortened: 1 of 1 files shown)


mehrdadh:
@tvm-bot rerun

gromero (changes requested):
Folks, please, the PR description _has_ to be more descriptive. This will help others outside the project, which are not sitting close at the same working table, or attending the same weekly company meeting, to understand and follow how the project evolves or changes over time. It's not just a matter of fixing stuff, but also about _sharing_ information and progress.

Is that hardware a Zephyr HW, an Arduino HW or both? Which board exactly the test is failing and how? I (others) might encounter a similar error on a different board that requires more stack to run the tests in the future and I (and probably others not working close to us) would like to do the due git archaeology to find out hints about how to fix the error! I know it's very simple change but yet there is a lot to be said and shared about it, so, please, add the details to the PR description and, specially, take care to land it neatly with the commit message (also as per the [Commit Message Guideline](https://github.com/apache/tvm/blob/main/docs/contribute/pull_request.rst#commit-message-guideline).

--

Alright with adding a default main stack size to run tests on Zephyr, but won't Arduino fail on that since `config_main_stack_size` is not available on Arduino platform? Hence should a conditional be added in the test to not pass such a option when the Arduino platform is selected?

Cheers.

gromero (approved):
@mehrdadh Thanks, LGTM!

I think CI error is not related to the change, so maybe just re-trigger.

gromero:
@tvm-bot rerun

leandron:
Hi all. I think it's important to echo @gromero's comment regarding the [Commit Message Guideline](https://github.com/apache/tvm/blob/main/docs/contribute/pull_request.rst#commit-message-guideline). For next PRs, at merging time, please take a moment to adjust the commit message to be more self-explanatory and leave a good track of the changes in our git history.

As I'm involved in the ongoing release process, it complicates a lot the process to be able to track changes that have generic names or no commit message body.

PR #13600, opened 2022-12-12
[microTVM] Build standalone_crt with cmake instead of makefile

### LINKED ISSUE #13533: [Bug] StandaloneCrt.cmake limits ability to enable microTVM build on other platforms
### Expected behavior
When enabling the USE_MICRO build flag, builds complete successfully on various platforms.

### Actual behavior
Enabling USE_MICRO build flag can be problematic on various platforms, including Windows (worked-around for now), MacOS, and other cross-compilation environments.

### Details
StandaloneCrt.make builds a number of libraries when USE_MICRO is enabled. The object files and libraries are written to the host_standalone_crt directory in the TVM build directory. Only one of those libraries, libmicrotvm_rpc_common, is linked into the TVM runtime library. The other libraries in host standalone_ crt are built only to assure they are still buildable.

Additionally, StandaloneCrt.make copies a number of source files to the standalone_ crt in the build directory for use later by the microTVM Project API server as well as for building the libraries mentioned above.

The issue is that the both the makefile used to build those libraries as well as the commands to link in libmicrotvm_rpc_common to TVM in StandaloneCrt.cmake are tailored to only support GCC (or clang) and can cause issues on other platforms or in cross-compilation environments.

alanmacd:
Build standalone_crt with cmake instead of makefile to allow for better portability of microTVM code to other build environments.

fixes #13533


cmake/modules/StandaloneCrt.cmake
```diff
@@ -40,129 +40,143 @@ if(MSVC)
 
 else()
```
alanmacd, line 42: NOTE: change diff settings to hide whitespace (if not set) for this diff to make more sense since I needed to change indentation.

cmake/modules/StandaloneCrt.cmake
```diff
@@ -40,129 +40,143 @@ if(MSVC)
...
-        if("${copy_files_length}" EQUAL 0)
-          message(FATAL_ERROR "CRT copy job matched 0 files: ${job_src_base}/${copy_pattern} -> ${copy_dest}")
-        endif()
-        foreach(copy_src IN LISTS copy_files)
-          get_filename_component(dest_path "${standalone_crt_base}/${copy_dest}/${copy_src}" ABSOLUTE)
-          tvm_micro_add_copy_file(host_isolated_build_deps ${job_src_base}/${copy_src} ${dest_path})
-        endforeach()
-      endforeach()
+  function(create_crt_library CRT_LIBRARY_NAME)
+    set(cmake_crt_lib_name host_standalone_crt_${CRT_LIBRARY_NAME})
```
alanmacd, line 44: uppercase

cmake/modules/StandaloneCrt.cmake
```diff
@@ -40,129 +40,143 @@ if(MSVC)
...
+
+  add_custom_target(host_standalone_crt DEPENDS ${CMAKE_CRT_LIBRARIES} standalone_crt)
+
+  # Create the `crttest` target if we can find GTest.  If not, we create dummy
+  # targets that give the user an informative error message.
+  if(GTEST_FOUND)
+    tvm_file_glob(GLOB TEST_SRCS ${CMAKE_CURRENT_SOURCE_DIR}/tests/crt/*.cc)
+    add_executable(crttest ${TEST_SRCS})
+    target_include_directories(crttest SYSTEM PUBLIC ${CMAKE_CURRENT_BINARY_DIR}/standalone_crt/include ${CMAKE_CURRENT_SOURCE_DIR}/src/runtime/micro)
+    target_link_libraries(crttest PRIVATE ${CMAKE_CRT_LIBRARIES} GTest::GTest GTest::Main pthread dl)
```
alanmacd, line 175: rename to STANDALONE_CRT_LIBRARIES (CMAKE prefix is confusing)

cmake/modules/StandaloneCrt.cmake
```diff
@@ -40,129 +40,147 @@ if(MSVC)
...
+      foreach(copy_src IN LISTS copy_files)
+        get_filename_component(dest_path "${STANDALONE_CRT_BASE}/${copy_dest}/${copy_src}" ABSOLUTE)
+        tvm_micro_add_copy_file(host_isolated_build_deps ${job_src_base}/${copy_src} ${dest_path})
+      endforeach()
+    endforeach()
+  endforeach()
+
+  add_custom_target(standalone_crt DEPENDS ${host_isolated_build_deps})
+
+  get_filename_component(host_build_dir_abspath "${CMAKE_CURRENT_BINARY_DIR}/host_standalone_crt" ABSOLUTE)
```
alanmacd, line 140: remove, unused

cmake/modules/StandaloneCrt.cmake
```diff
@@ -40,129 +40,147 @@ if(MSVC)
...
+  create_crt_library(microtvm_rpc_server
+                     ${RUNTIME_CRT_SOURCE_DIR}/microtvm_rpc_server/rpc_server.cc)
+
+  create_crt_library(common
+                     ${RUNTIME_CRT_SOURCE_DIR}/common/crt_backend_api.c
+                     ${RUNTIME_CRT_SOURCE_DIR}/common/crt_runtime_api.c
+                     ${RUNTIME_CRT_SOURCE_DIR}/common/func_registry.c
+                     ${RUNTIME_CRT_SOURCE_DIR}/common/ndarray.c
+                     ${RUNTIME_CRT_SOURCE_DIR}/common/packed_func.c)
+
```
alanmacd, line 170: add additional libraries that are part of standalone_crt

cmake/modules/StandaloneCrt.cmake
```diff
@@ -40,129 +40,147 @@ if(MSVC)
 
 else()
```
alanmacd, line 41: try removing MSVC conditional code

leandron (commented):
Thanks @alanmacd, I think this makes the build process for `standalone_crt` much cleaner.

One comment I have is related to the python packaging mechanism. As part of packaging TVM will look for the `standalone_crt` directory (I think it is the copied one, as referred in #13533) to add the static files to the package. Can you double check whether that behaviour is also kept while you propose these changes?

The static files are needed so that users can integrated embedded projects only with the TVM installed from the package, rather than requiring the package+cloning `apache/tvm` repo.

https://github.com/apache/tvm/blob/b7015bb388a1d8f84b89e57709831e74c032d941/python/setup.py#L59-L64

alanmacd:
> The static files are needed so that users can integrate embedded projects only with the TVM installed from the package, rather than requiring the package+cloning `apache/tvm` repo.
> 
> https://github.com/apache/tvm/blob/b7015bb388a1d8f84b89e57709831e74c032d941/python/setup.py#L59-L64

Thanks @leandron I will verify that files are placed in the same locations as previous done with the makefile version (I may have stashed the library files in slightly different locations)

Is there a specific way you typically invoke setup.py to exercise this feature or an example command that I should use to test?

leandron:
> Is there a specific way you typically invoke setup.py to exercise this feature or an example command that I should use to test?

Build TVM with `USE_MICRO ON`, then `cd python; python setup.py bdist_wheel`.

Also, if that helps with context, the real use case would be following tlcpack steps https://github.com/tlc-pack/tlcpack

alanmacd:
> python setup.py bdist_wheel

@leandron ok, it looks like it's only copying the standalone_crt directory into the wheel, I've updated my PR to not put the library binaries in the standalone_crt directory (matching the current TVM main), all the other files should be the same as I didn't change the standalone_crt file copy code.

cmake/modules/StandaloneCrt.cmake
```diff
@@ -40,129 +40,154 @@ if(MSVC)
...
+    # add this library to the list of CRT libraries
+    set(CRT_LIBRARIES ${CRT_LIBRARIES} ${CRT_LIBRARY_NAME} PARENT_SCOPE)
 
-    list(APPEND crt_libraries memory graph_executor microtvm_rpc_server microtvm_rpc_common common)  # NOTE: listed in link order.
-    foreach(crt_lib_name IN LISTS crt_libraries)
-      list(APPEND crt_library_paths "host_standalone_crt/lib${crt_lib_name}.a")
-    endforeach()
+    target_include_directories(${CRT_LIBRARY_NAME}
+                               PUBLIC
+                               ${CMAKE_CURRENT_SOURCE_DIR}/src/runtime/micro/
```
alanmacd, line 61: update to use standalone_crt dir

cmake/modules/StandaloneCrt.cmake
```diff
@@ -40,129 +40,154 @@ if(MSVC)
...
-      add_dependencies(${cmake_crt_lib_name} host_standalone_crt "${cmake_crt_lib_path}")
-      set_target_properties(${cmake_crt_lib_name} PROPERTIES
-          IMPORTED_LOCATION "${cmake_crt_lib_path}"
-          IMPORTED_OBJECTS "${cmake_crt_lib_path}"
-          PUBLIC_HEADER "${crt_headers}")
-    endforeach()
+    set_target_properties(${CRT_LIBRARY_NAME}
+                          PROPERTIES
+                          ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/host_standalone_crt
+                          POSITION_INDEPENDENT_CODE ON)
```
alanmacd, line 67: add option to disable PIC if using a standalone CMake build

leandron:
> @leandron ok, it looks like it's only copying the standalone_crt directory into the wheel, I've updated my PR to not put the library binaries in the standalone_crt directory (matching the current TVM main), all the other files should be the same as I didn't change the standalone_crt file copy code.

thanks @alanmacd

cmake/modules/StandaloneCrt.cmake
alanmacd: will do this in subsequent PR that also replaced the standalone_crt/Makefile with cmake version

cmake/modules/StandaloneCrt.cmake
alanmacd: will follow-up in another PR with this

cmake/modules/StandaloneCrt.cmake
alanmacd: actually, this is only for crt_config.h so it's the same as what the previous Makefile did

areusch (approved):

areusch:
@leandron can you explicit approve if you're good w/ this?

cmake/modules/StandaloneCrt.cmake
alanmacd: done, thanks

PR #13601, opened 2022-12-12
[Fix] Task scheduler error prompt upon build/run failure

MasterJH5574:
This PR fixes one error prompt in task scheduler. Prior to this fix, the prompt assumes the error message always comes from a build failure. However, it is possible that the error message comes from a run failure.

cc @junrushao @zxybazh


zxybazh (approved):
LGTM, thanks for sending in the fix!

PR #13602, opened 2022-12-12
[Relay][Frontend][Onnx] SequenceAt and SplitToSequence Operators

jwfromm:
This PR adds support for some new sequence operators in the onnx frontend. I made a few other small tweaks to get everything working, specifically an attempt to rectify dynamic rank in `If` statements and the removal of a superfluous check in concat's shape function. With these changes, we can import and run the latest Yolo variants which is pretty cool.

### DIFF
--- a/python/tvm/relay/frontend/onnx.py
+++ b/python/tvm/relay/frontend/onnx.py
@@ -5600,11 +5600,13 @@ class SplitToSequence(OnnxOpConverter):
             if keepdims == 0:
                 output = [_op.squeeze(tensor_slice, axis=[axis]) for tensor_slice in output]
             return _expr.Tuple(list(output))
-        
+
         # Otherwise, split based on provided split value.
         else:
             # For now we only support constant valued split.
-            assert isinstance(split, _expr.Constant), "Only constant split supported for SplitToSequence"
+            assert isinstance(
+                split, _expr.Constant
+            ), "Only constant split supported for SplitToSequence"
             split = split.data.numpy()
             if len(split.shape) == 1 and split.shape[0] > 1:
                 # If split is a 1D tensor, it must be converted to indices for relay compatibility.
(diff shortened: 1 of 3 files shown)


jwfromm:
@AndrewZhaoLuo @michalpiszczek @sfvaroglu can you guys take a look at this PR?

python/tvm/relay/frontend/onnx.py
```diff
@@ -5565,6 +5581,66 @@ def _impl_v11(cls, inputs, attr, params):
...
+                split = split[:-1]
+            else:
+                # Otherwise get split as an integer.
+                split = int(split)
+
+            output = _op.split(input_tensor, split, axis=axis)
+
+            # If keepdims is set to 0 remove split axis. Note that this is
+            # an inconsistency with the onnx spec but is needed for pytorch compatibility.
+            if keepdims == 0:
```
AndrewZhaoLuo, line 5624: nit `not keepdims` throughout

python/tvm/relay/frontend/onnx.py
```diff
@@ -4008,6 +4008,22 @@ def _impl_v1(cls, inputs, attr, params):
         for var in else_free_vars:
             graph_scope._nodes.update({var.name_hint: var})
 
+        # Sometimes pytorch to onnx will insert silly if statements that produce dynamic ranks.
+        # Often these dont contribute anything. If we see a dynamic rank output, try to unify
+        # them so we can continue without breaking.
+        then_shape = infer_shape(then_expr)
```
AndrewZhaoLuo, line 4014: Very sus, but I guess this is the best we can do in this situation. The only reason I will allow this is it's in converting `If` which is inherently going to be a bit more sus due to relay limitations.

AndrewZhaoLuo (approved):

python/tvm/relay/frontend/onnx.py
jwfromm: i actually prefer comparing to 0 in this case since it has meaning distinct from None.

python/tvm/relay/frontend/onnx.py
jwfromm: @ehsanmok noted that he's working on exposing Optional types in relay that may allow a more elegant solution once it lands.

sfvaroglu:
LGTM! Thanks @jwfromm!

PR #13603, opened 2022-12-13
[microTVM] Update poetry.lock

mehrdadh:
The current poetry.lock does not work as expected with Zephyr. When I run a test outside of docker it shows error like bellow. It's probably some mismatching between the `poetry.lock` and zephyr `2.7.3` version `requirement.txt` file.

```
~/zephyrproject/zephyr/boards/arm/nucleo_l4r5zi/nucleo_l4r5zi.dts:9:10: fatal error: st/l4/stm32l4r5z(g-i)tx-pinctrl.dtsi: No such file or directory
    9 | #include <st/l4/stm32l4r5z(g-i)tx-pinctrl.dtsi>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
CMake Error at ~/zephyrproject/zephyr/cmake/dts.cmake:167 (message):
  command failed with return code: 1
Call Stack (most recent call first):
  ~/zephyrproject/zephyr/cmake/app/boilerplate.cmake:541 (include)
  ~zephyrproject/zephyr/share/zephyr-package/cmake/ZephyrConfig.cmake:24 (include)
  ~/zephyrproject/zephyr/share/zephyr-package/cmake/ZephyrConfig.cmake:40 (include_boilerplate)
  CMakeLists.txt:33 (find_package)
```

### DIFF
--- a/apps/microtvm/poetry.lock
+++ b/apps/microtvm/poetry.lock
@@ -19,7 +19,7 @@
 
 [[package]]
 name = "absl-py"
-version = "1.2.0"
+version = "1.3.0"
 description = "Abseil Python Common Libraries, see https://github.com/abseil/abseil-py."
 category = "main"
 optional = true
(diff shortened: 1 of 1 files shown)


gromero:
@mehrdadh I wonder if it was really the root cause of `st/l4/stm32l4r5z(g-i)tx-pinctrl.dtsi` not being found in your case. Have you tried to narrow down why `gcc` command supposed to take care of these includes (in `cmake/dts.cmake:167`) was failing to find this include file in your env. before refreshing your build environment?

I'm not saying I'm against this patch or against such a pragmatic approach for fixing errors like that, it's just that I could not reproduce your issue even using exactly the same `poetry.lock` you're updating -- which suggests it's not the root cause of the error. Also the patch is updating several other packages which I have not idea how they could be remotely related to the issue in question. Hence, since I can't reproduce it and I can't see how the updated packages could be related to the issue, I can't really review it too.

mkatanbaf:
@mehrdadh I had the same issue a while ago, and the solution was to install west on the poetry virtual env.

mehrdadh:
@gromero, @mkatanbaf  Thanks for pointers. 
I did more investigation, the problem was that I wasn't installing zephyr requirement.txt file:
```
poetry run pip3 install -r ~/zephyrproject/zephyr/scripts/requirements.txt
```
closing this PR as it is not necessary anymore

PR #13560, opened 2022-12-06
[TIR][Transform] Keep the allocate buffers order after update buffer allocation location

FredJia-intellif:
This PR fix indeterminate order between each allocated buffer after update buffer allocation location

PR #13605, opened 2022-12-13
[TIR] Fix PlanAndUpdateBufferAllocationLocation not visiting constant buffer

masahi:
https://github.com/apache/tvm/pull/13560 is supposed to change only the order in which buffers are visited, but the created `buffer_alloc_recorder_` misses buffers associated with `AllocateConst` nodes. These buffers are not part of `func->buffer_map`, and appear only in `read` or `match_buffer` regions, so they are not caught by existing visitors for `BufferLoadNode` and `BufferStoreNode`.

Since constant buffers are not correctly handled by `PlanAndUpdateBufferAllocationLocation` now, tuning on Hexagon is currently broken with this error. This PR fixes this.

```
E             File "/home/fparizi/src/tvm_src/src/tir/transforms/plan_update_buffer_allocation_location.cc", line 163
E           TVMError: 
E           ---------------------------------------------------------------
E           An error occurred during the execution of TVM.
E           For more information, please see: https://tvm.apache.org/docs/errors.html
E           ---------------------------------------------------------------
E             Check failed: (buffer_data_to_buffer_.count(source_var)) is false:

```

Since creating a TVMScript-based test case involving constant buffers (`AllocateConst`) is difficult, I hope we can merge this simple PR without a test case.

cc @Hzfengsy @FredJia-intellif

### DIFF
--- a/src/tir/transforms/plan_update_buffer_allocation_location.cc
+++ b/src/tir/transforms/plan_update_buffer_allocation_location.cc
@@ -61,24 +61,33 @@ class BufferAllocateOrderCollector : public StmtExprVisitor {
   }
 
  private:
+  bool find(const Buffer& buf) {
+    return std::find(buffer_alloc_recorder_.begin(), buffer_alloc_recorder_.end(), buf) !=
+           buffer_alloc_recorder_.end();
+  }
+
   void VisitStmt_(const BlockNode* op) final {
     for (const Buffer& buffer : op->alloc_buffers) {
       buffer_alloc_recorder_.push_back(buffer);
     }
+    for (const auto& region : op->match_buffers) {
+      if (!find(region->source->buffer)) {
+        buffer_alloc_recorder_.push_back(region->source->buffer);
+      }
+    }
+
     StmtExprVisitor::VisitStmt_(op);
   }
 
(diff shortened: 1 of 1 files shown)


Hzfengsy (approved):

Hzfengsy:
I'm sorry I missed this when reviewing https://github.com/apache/tvm/pull/13560. 

BTW, Could you please add regression tests for Hexagon? Then we can make sure the const tensors can be considered in every PR.

masahi:
> BTW, Could you please add regression tests for Hexagon?

Since we cannot parse a large constant via TVMScript, creating a TVMScript-based test case involving constant buffers (`AllocateConst`) is difficult or near impossible. But if there is a way to create TensorIR program directly from Relay (using `link-params=True` to generate `AllocateConst`), then I can probably create a minimum test case for `PlanAndUpdateBufferAllocationLocation`.  

I think there is no way to create TensorIR from Relay (via `CreatePrimFunc`) directly in Python currently (everything is contained in `TECompiler`). But yes, Hexagon tuning keeps breaking due to `AllocateConst` not handled properly in PRs, so I'll try to fix this situation.

masahi:
Regression test is added in https://github.com/apache/tvm/pull/13606

PR #9282, opened 2021-10-14
[CORE][Relay] Swap and remove compile_engine with te_compiler followup of #8775

mikepapadim:
This is a follow-up work for #8775 to address the pending comments to remove the `compile_engine`.

As this is an intermediate step towards completely moving away from the engine, we exposed the following legacy functions from the compile engine to ensure backwards compatibility atm:
* _TECompilerJIT 
* _TECompilerLower 
* _make_LoweredOutput
* _make_CCacheKey

Now, compile_engine.py has been replaced with te_compiler.py and deprecated funcs, such as LowerShapeFuncs kept private within the te_compiler.cc.

@jroesch @mbs-octoml @electriclilies @comaniac @junrushao1994

PR #8775, opened 2021-08-17
Remove compile_engine.h for real this time. 

jroesch:
This removes the `compile_engine.h` header enabling us to start refactoring the `te_compiler.h` interface to better match the rest of the compiler. After we land the final refactors from @mikepapadim and @mbs-octoml we can start cleaning up the interface to better match the desired end state. 

cc @electriclilies @csullivan @tmoreau89

PR #13606, opened 2022-12-13
[Relay][TIR] Add utility to lower Relay func to TIR prim func

masahi:
This is very useful for creating a TensorIR program that is otherwise hard to so via manual TVMScript or te compute + `create_prim_func`.


src/relay/backend/te_compiler_cache.cc
```diff
@@ -1088,6 +1088,30 @@ std::tuple<Array<te::Tensor>, Array<runtime::NDArray>, std::string> LowerTECompu
   return std::make_tuple(tensor_outs, constants, lower_te_compute.candidate_name_);
 }
 
+std::pair<Optional<tir::PrimFunc>, std::string> LowerToPrimFunc(const Function& relay_func,
```
csullivan, line 1091: Should we add a check in here that relay_func has attr::kPrimitive to alert the user early if they are trying to lower a non-primitive function?

csullivan (approved):
Awesome, I love it. One small suggestion if you agree.

src/relay/backend/te_compiler_cache.cc
masahi: Thanks, yes we should do that. Updated.

echuraev (approved):
Awesome and very useful improvement! Thanks.

junrushao (approved):
Looks great!

Mousius:
:wave: Hello @masahi / @csullivan / @echuraev / @junrushao,

Whilst I appreciate this is useful with the use-case demonstrated, I believe it might've needed a bit more time for others to formulate a review as it was up for less than 24 hours before merge?

One of my concerns is that this seems to be creating new public interfaces to the internals of the TECompiler which in the past we've attempted to remove:
* https://discuss.tvm.apache.org/t/rfc-relay-tecompiler-rewrite-existing-compile-engine-to-match-updated-compiler-flow/9233
* https://github.com/apache/tvm/pull/8775
* https://github.com/apache/tvm/pull/9282

This change means we have to maintain an interface which generates a single `tir::PrimFunc` from a `relay::Function` where-as it could potentially lead to multiple `tir::PrimFunc` with a single `IRModule` as a result. The next line down:

```
IRModule tir_mod = PrimFuncToIRModule(f.value());
```

Implies we wanted it in `IRModule` form anyway, so creating an `IRModule` with a `Function` and running `LowerTE(mod)` seems the best result here?

masahi:
@Mousius The new function simply composes the existing interface to the TECompiler, `LowerTECompute` and `CreatePrimFunc` that converts lowered TE compute to TensorIR. `LowerTECompute` was introduced after the compile engine refactor you mentioned completed, and made only possible thanks to that refactoring work. So I don't get your concern here.

What I wanted to add is a simple interface that converts one Relay primitive function to the corresponding TIR prim func. Then I realized that `backend/task_extraction.cc` already has code that does exactly that, so I simply extracted into a reusable function and exposed it to Python. We require the input to be a primitive function, so we never generates multiple TIR prim funcs. 

> Implies we wanted it in IRModule form anyway, so creating an IRModule with a Function and running LowerTE(mod) seems the best result here?

This is just MetaSchedule storing an IRModule in a task. I didn't change any behavior about task extraction in this PR. What we want is a `relay::Function -> tir::PrimFunc` interface, `IRModule -> IRModule` is valid but it wouldn't convey what this new API is meant for.

Mousius:
> @Mousius The new function simply composes the existing interface to the TECompiler, `LowerTECompute` and `CreatePrimFunc` that converts lowered TE compute to TensorIR. `LowerTECompute` was introduced after the compile engine refactor you mentioned completed, and made only possible thanks to that refactoring work. So I don't get your concern here.

I find this equally concerning, given this is also exposing more of the internals of the TE Compiler rather than having clear boundaries within the codebase. 

> What I wanted to add is a simple interface that converts one Relay primitive function to the corresponding TIR prim func. Then I realized that backend/task_extraction.cc already has code that does exactly that, so I simply extracted into a reusable function and exposed it to Python. We require the input to be a primitive function, so we never generates multiple TIR prim funcs.

Exposing it python and as part of the public API means others can access it, which means that we have to support it as a way of using the compiler rather than as an aspect of meta-scheduler. This is similar to the previous methods that we were unable to deprecate in https://github.com/apache/tvm/pull/13606.

> This is just a convention of MetaSchedule, storing an IRModule with one prim func as a task. I didn't change any behavior about task extraction in this PR. What we want is a `relay::Function -> tir::PrimFunc` interface, `IRModule -> IRModule` is valid but it wouldn't convey what this new API is meant for. `LowerTE` is also overkill since I don't need to lower schedules and I'm not concerned about external codegen etc.

I don't think `LowerTE` is overkill in practice, if it doesn't find any external codegen or other things to do it won't take any action - it's a standardised interface between two modules in the codebase, you've already indicated meta schedule is breaking that boundary and that leads us back to the original issues with CompileEngine.

PR #13610, opened 2022-12-13
[microTVM]Add default value to unspecified project options in project API

mehrdadh:
In current TVM, project API ignores the default values that are set for each project option in a project API server. This results in extra function handlers to specify the default value for those options in runtime.
This PR searches through available ProjectOptions for a project API server and adds default values for unspecified options before sending the `options` to API call function handler.

### DIFF
--- a/apps/microtvm/arduino/template_project/microtvm_api_server.py
+++ b/apps/microtvm/arduino/template_project/microtvm_api_server.py
@@ -69,7 +69,11 @@ PROJECT_OPTIONS = server.default_project_options(
 ) + [
     server.ProjectOption(
         "arduino_cli_cmd",
-        required=(["generate_project", "flash", "open_transport"] if not ARDUINO_CLI_CMD else None),
+        required=(
+            ["generate_project", "build", "flash", "open_transport"]
+            if not ARDUINO_CLI_CMD
+            else None
+        ),
         optional=(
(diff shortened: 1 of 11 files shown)


apps/microtvm/arduino/template_project/microtvm_api_server.py
```diff
@@ -337,7 +332,7 @@ def _copy_cmsis(self, project_path: pathlib.Path, cmsis_path: str):
         However, the latest release does not include header files that are copied in this function.
         """
         (project_path / "include" / "cmsis").mkdir()
-        cmsis_path = get_cmsis_path(cmsis_path)
+        cmsis_path = pathlib.Path(cmsis_path)
```
areusch, line 335: this one also removes the possibility of setting CMSIS_PATH through env right? can you add to PR description?

mehrdadh: not it doesn't, the default value for cmsis_path is set by environment. Check `python/tvm/micro/project_api/server.py` where we defined the cmsis_path as a projectOption

areusch: ah, gotcha, nevermind then

areusch (approved):

guberti (commented):
LGTM, sorry for the belated review!

PR #13611, opened 2022-12-14
[Metaschedule] Aligning get_top_k logic in MemoryDatabase and JSONDatabase

Icemist:
The get_top_k implementation of JSONDatabase confuses by giving out N/A results that were supposedly sorted in a non-obvious way. 
For tuning records with no measurement results, it is not possible to determine the order and identify the top results.


Icemist:
CC @csullivan

masahi (approved):

PR #13613, opened 2022-12-14
Add check for non-contiguous memory access when lowering to async dma…

nverke:
… copies.

Currently we do not support non contiguous memory copies for async dma so this adds a check to error out in the case that either load or store is non contiguous. Unfortunately if we do not throw an error in this case then we get inaccurate outputs. Added a test to verify that we are indeed throwing an error in this case that should be removed once support is added.

### DIFF
--- a/src/tir/transforms/lower_async_dma.cc
+++ b/src/tir/transforms/lower_async_dma.cc
@@ -22,6 +22,7 @@
  */
 
 #include <tvm/arith/analyzer.h>
+#include <tvm/arith/iter_affine_map.h>
 #include <tvm/tir/stmt_functor.h>
 #include <tvm/tir/transform.h>
 
@@ -33,6 +34,12 @@ namespace tir {
 class AsyncDMALowerer : public StmtExprMutator {
  public:
   explicit AsyncDMALowerer(bool dma_bypass_cache) : dma_bypass_cache_(dma_bypass_cache) {}
+
+  // Create member statement to track a mapping from iter var to iter range
+  Stmt VisitStmt_(const ForNode* op) final {
+    input_iters.Set(op->loop_var, Range(op->min, op->extent));
+    return StmtExprMutator::VisitStmt_(op);
+  }
 
   Stmt VisitStmt_(const AttrStmtNode* op) final {
     // Convert this, for example:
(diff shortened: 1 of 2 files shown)


src/tir/transforms/lower_async_dma.cc
```diff
@@ -146,13 +153,32 @@ class AsyncDMALowerer : public StmtExprMutator {
 
       // map loop variable to zero for the store index & simplify
       Array<PrimExpr> store_index = bufferstorenode->indices;
+
+      // Use DetectIterMap to detect whether store index is non-contiguous.
+      arith::Analyzer analyzer;
+      auto store_iter_map = DetectIterMap(store_index, input_iters, 1, arith::IterMapLevel::NoCheck,
+                                          &analyzer, false);
+      if (!store_iter_map->errors.empty()) {
+        LOG(FATAL) << "Unable to lower async dma for non contiguous memory access with index: "
```
adstraw, line 162: Nit:  You might differentiate the error message by including something text indicating that the "buffer store" contained the non contiguous access.

src/tir/transforms/lower_async_dma.cc
```diff
@@ -146,13 +153,32 @@ class AsyncDMALowerer : public StmtExprMutator {
...
       });
 
       // map loop variable to zero for the load index & simplify
       Array<PrimExpr> load_index = bufferloadnode->indices;
+
+      // Use DetectIterMap to detect whether load index is non-contiguous.
+      auto load_iter_map =
+          DetectIterMap(load_index, input_iters, 1, arith::IterMapLevel::NoCheck, &analyzer, false);
+      if (!load_iter_map->errors.empty()) {
+        LOG(FATAL) << "Unable to lower async dma for non contiguous memory access with index: "
```
adstraw, line 178: Nit:  Same as above.  Add text indicating "buffer load" contained non contiguous access.

adstraw (approved):
LGTM.  Just one nit.

masahi:
> once support is added.

What support is being planed? 2D DMA or breaking up into contiguous copies?

nverke:
We are still discussing the pros and cons of the potential options but the possible solutions are. 
1. Lower the copy into n x 1D dma copies. 
2. Lower to 2D dma copies.
3. Use tensorize to make synchronous dma copies then in the lower_async_dma pass add the necessary asynchronous pieces.

src/tir/transforms/lower_async_dma.cc
nverke: Added!

src/tir/transforms/lower_async_dma.cc
nverke: Added!

masahi (approved):
Need to fix the lint

PR #13615, opened 2022-12-14
[Relay] Remove overwriting of matmul shapes when they are static

jwfromm:
In the Relay Matmul shape relation, we are a little over enthusiastic about unifying dynamic shapes. If one of the shapes is static, it does not need to be unified. This change only rewrites dynamic shapes to required static constraints.

### DIFF
--- a/src/relay/op/nn/nn.h
+++ b/src/relay/op/nn/nn.h
@@ -115,20 +115,36 @@ bool MatmulRel(const Array<Type>& types, int num_inputs, const Attrs& attrs,
         auto sb = B_shape.size();
         if (transpose_a && transpose_b) {
           auto tmp = A_shape[sa - 2];
-          A_shape[sa - 2] = B_shape[sb - 1];
-          B_shape[sb - 1] = tmp;
+          if (A_shape[sa - 2].as<tir::AnyNode>()) {
+            A_shape[sa - 2] = B_shape[sb - 1];
+          }
+          if (B_shape[sb - 1].as<tir::AnyNode>()) {
+            B_shape[sb - 1] = tmp;
+          }
         } else if (transpose_a) {
           auto tmp = A_shape[sa - 2];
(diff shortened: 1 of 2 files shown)


src/relay/op/nn/nn.h
```diff
@@ -115,20 +115,36 @@ bool MatmulRel(const Array<Type>& types, int num_inputs, const Attrs& attrs,
         auto sb = B_shape.size();
         if (transpose_a && transpose_b) {
           auto tmp = A_shape[sa - 2];
-          A_shape[sa - 2] = B_shape[sb - 1];
-          B_shape[sb - 1] = tmp;
+          if (A_shape[sa - 2].as<tir::AnyNode>()) {
```
AndrewZhaoLuo, line 118: There is now a ton of nesting. To make it readable I suggest a structure like 
```
A_shape_new = ...
B_shape_new = ...
index_swap_a, index_swap_b;

if (transpose_a && transpose_b) {
   // calculate index_swap_a and index_swap_b
} ... 

if (A_shape[index_swap_a].as<Any>...) {
    // do swap
}

if (B_shape[index_swap_b].as<Any>...
```

AndrewZhaoLuo (approved):
Thanks, one nit.

src/relay/op/nn/nn.h
jwfromm: Yeah good call, updated in latest push.

slyubomirsky (changes requested):
The changes seem reasonable, but I think we should add at least some test cases that demonstrate the difference in logic to avoid possible regression.

jwfromm:
I've added a check to `test_dense` to confirm that the input shapes aren't unnecessary made dynamic that fails without this change.

slyubomirsky (approved):
Thanks for adding in the assertion. I'm surprised this quirk was lurking in an existing test case.

shingjan (approved):
LGTM. Thanks for adding the test!

PR #13617, opened 2022-12-14
[LLVM] Use std::nullopt instead of llvm::None

kparzysz-quic:
Pass `std::nullopt` to initialization of `PassBuilder` for `PGOOptions`. LLVM is moving away from its own `Optional` type to `std::optional`.


Icemist (approved):
LGTM. Looks backward compatible because `llvm::Optional<T>` has a constructor from  `std::nullopt_t`.

It is about this change in the main branch:
`LLVM::Optional<PGOOptions> PGOOpt = LLVM::None,`
=>
`std::optional<PGOOptions> PGOOpt = std::nullopt,`

masahi (approved):

PR #13622, opened 2022-12-15
[TVMScript] Fix print round-tripable multi thread env binding

wrongtest-intellif:
Fix the following script with multiple kernel launch (each take standalone thread binding) which is not round-trip-able.
```python
import tvm

@T.prim_func
def main(A: T.Buffer[128, "float32"], C: T.Buffer[128, "float32"]):
    B = T.alloc_buffer([128], "float32")
    for i in range(128):
        with T.block("B1"):
            vi = T.axis.remap("S", [i])
            B[vi] = A[vi] + 1.0
    for i in range(128):
        with T.block("B2"):
            vi = T.axis.remap("S", [i])
            C[vi] = B[vi] + 2.0

s = tvm.tir.schedule.Schedule(main)
b1 = s.get_block("B1")
b2 = s.get_block("B2")
i = s.get_loops(b1)[0]
s.bind(i, "threadIdx.x")
i = s.get_loops(b2)[0]
s.bind(i, "threadIdx.x")
f = tvm.lower(s.mod["main"])
print(f.script())
tvm.ir.assert_structural_equal(f, tvm.script.from_source(f.script()))
```

Since the env thread is always printed on the function head, `TryDeallocVar` which decrease the name uniqueness counting may not be invoked for thread binding code path.

### DIFF
--- a/src/printer/tvmscript_printer.cc
+++ b/src/printer/tvmscript_printer.cc
@@ -1045,7 +1045,6 @@ Doc TVMScriptPrinter::VisitStmt_(const AttrStmtNode* op) {
             << ")";
         doc << Doc::NewLine() << PrintBody(op->body);
       }
-      TryDeallocVar(iter_var->var);
       return doc;
     }
   }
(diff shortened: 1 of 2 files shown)


Hzfengsy (approved):
LGTM. Thanks for the fixing

PR #13623, opened 2022-12-15
[Relay][Runtime] Add `set_input/output_zero_copy` in python

shingjan:
As titled, this apis is in cpp but not python. Added this apis for GraphModule and some tests.

cc: @vinx13

### DIFF
--- a/python/tvm/contrib/graph_executor.py
+++ b/python/tvm/contrib/graph_executor.py
@@ -211,11 +211,12 @@ class GraphModule(object):
         params : dict of str to NDArray
            Additional arguments
         """
-        self._set_input_zero_copy(key, value)
+        if key is not None:
+            self._set_input_zero_copy(key, value)
+
         if params:
-            # upload big arrays first to avoid memory issue in rpc mode
             keys = list(params.keys())
-            keys.sort(key=lambda x: -np.prod(params[x].shape))
+
             for k in keys:
                 # TODO(zhiics) Skip the weights for submodule in a better way.
                 # We should use ConstLoaderModule for initialization and remove
--- a/python/tvm/contrib/graph_executor.py
+++ b/python/tvm/contrib/graph_executor.py
@@ -154,7 +154,6 @@ class GraphModule(object):
         self.module = module
         self._set_input = module["set_input"]
         self._set_input_zero_copy = module["set_input_zero_copy"]
-        self._set_output = module["set_output"]
         self._set_output_zero_copy = module["set_output_zero_copy"]
         self._run = module["run"]
         self._get_output = module["get_output"]
@@ -197,19 +196,6 @@ class GraphModule(object):
                 val = self._get_input(k)
                 if val:
                     self._get_input(k).copyfrom(params[k])
-
-    def set_output(self, key, value):
-        """Set outputs to the module
-
-        Parameters
-        ----------
-        key : int or str
-           The output key
-
-        value : the output value
-           The output value
-        """
-        self._set_output(key, value)
 
     def set_input_zero_copy(self, key, value, **params):
         """Set inputs to the module via kwargs with zero memory copy
--- a/python/tvm/contrib/graph_executor.py
+++ b/python/tvm/contrib/graph_executor.py
@@ -153,8 +153,21 @@ class GraphModule(object):
     def __init__(self, module):
         self.module = module
         self._set_input = module["set_input"]
-        self._set_input_zero_copy = module["set_input_zero_copy"]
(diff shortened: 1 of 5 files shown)


python/tvm/contrib/graph_executor.py
```diff
@@ -195,6 +198,49 @@ def set_input(self, key=None, value=None, **params):
...
+
+        Parameters
+        ----------
+        key : int or str
+           The output key
+
+        value : the output value
+           The output value
+        """
+        self._set_output(key, value)
```
vinx13, line 212: this is not needed, it doesn't make sense to copy external data to the output tensor

src/runtime/graph_executor/graph_executor.cc
```diff
@@ -160,6 +160,16 @@ void GraphExecutor::SetInput(int index, DLTensor* data_in) {
   uint32_t eid = this->entry_id(input_nodes_[index], 0);
   data_entry_[eid].CopyFrom(data_in);
 }
+/*!
+ * \brief set index-th output to the graph.
+ * \param index The output index.
+ * \param data_in The output data.
+ */
+void GraphExecutor::SetOutput(int index, DLTensor* data_in) {
```
vinx13, line 168: not needed

python/tvm/contrib/graph_executor.py
```diff
@@ -195,6 +198,49 @@ def set_input(self, key=None, value=None, **params):
...
+           The input key
+
+        value : the input value in DLPack
+           The input key
+
+        params : dict of str to NDArray
+           Additional arguments
+        """
+        self._set_input_zero_copy(key, value)
+        self.set_input(None, None, **params)
```
vinx13, line 229: not needed

vinx13 (changes requested):

python/tvm/contrib/graph_executor.py
shingjan: I was hoping that we can keep the apis consistent with `set_input` so that is the reason I kept `**params`. Should we still remove it?

vinx13: We can support `params`. However, we can't call `set_input` as it involves with copying.

shingjan: good point. Updated this method to use zero copy.

python/tvm/contrib/graph_executor.py
```diff
@@ -212,7 +212,17 @@ def set_input_zero_copy(self, key, value, **params):
            Additional arguments
         """
         self._set_input_zero_copy(key, value)
-        self.set_input(None, None, **params)
+        if params:
+            # upload big arrays first to avoid memory issue in rpc mode
+            keys = list(params.keys())
+            keys.sort(key=lambda x: -np.prod(params[x].shape))
```
vinx13, line 218: probably not needed as there is no data copy

python/tvm/contrib/graph_executor.py
```diff
@@ -212,7 +212,17 @@ def set_input_zero_copy(self, key, value, **params):
            Additional arguments
         """
         self._set_input_zero_copy(key, value)
-        self.set_input(None, None, **params)
+        if params:
```
vinx13, line 215: `key, value` vs `params` are basically two ways to specify the input. if `params` is supported, `key, value` should be by default None (similar to `set_input`)

python/tvm/contrib/graph_executor.py
```diff
@@ -195,6 +197,47 @@ def set_input(self, key=None, value=None, **params):
                 if val:
                     self._get_input(k).copyfrom(params[k])
 
+    def set_input_zero_copy(self, key, value, **params):
```
vinx13, line 200:
```suggestion
    def set_input_zero_copy(self, key=None, value=None, **params):
```

shingjan: done

shingjan:
@tvm-bot re-run

tests/python/unittest/test_runtime_module_based_interface.py
```diff
@@ -688,6 +688,36 @@ def test_num_threads():
         assert reported == hardware_threads or reported == hardware_threads // 2
 
 
+@tvm.testing.requires_llvm
+def test_graph_module_zero_copy():
```
echuraev, line 692: Probably the second test case can be added. This test case should test setting `zero_input` when the keys and values passed through `params`

shingjan: @echuraev a second test case is added for setting params input.

vinx13 (approved):

PR #13625, opened 2022-12-15
[MetaSchedule] Fix tensorcore winograd task extraction

masahi:
When both AutoTVM and AS/MS winograd are applicable, the latter is never picked by the op strategy because the AutoTVM one has a higher plevel (20 vs 15)

https://github.com/apache/tvm/blob/b16a64d6edb9fd1a014fc51995dff7d0e2f4c84e/python/tvm/relay/op/strategy/cuda.py#L271-L276

This has been fixed, and now we can properly extract FP16 winograd tasks for MS tuning.

@vinx13 @junrushao @zxybazh


junrushao (approved):
Interesting! I was completely unaware of the issue, and thanks for pointing it out!

PR #13630, opened 2022-12-16
[BugFix][TVMScript] Parser crash

lightzhan-intellif:
This PR tries to fix the crash of parser when the old value of a var is an array but the new value is not. For example:
```python
def func_wrapper(shape, dtype):
    @T.prim_func
    def test_case():
        a = T.alloc_buffer(shape, dtype=dtype)
    
    return test_case


if __name__ == "__main__":
    a = np.zeros((10, 10), dtype="int8")
    print(func_wrapper((256, 256), dtype="int8").script())
```
In the above code, there are two assignment to var 'a'. In the global scope, its value is a numpy array. But it is a Buffer in the prim function. There is a table named 'name2value' to track the value of vars like 'a' here.
When the parser wants to update its value, it will compare the value between the new and the old assignment. Here the problem comes. When we use '==' to compare an array with a value, the result is an array too, which can not be used as a condition of a if stmt directly. So, the code above will emit an error:
```shell
error: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
 --> /workspace/code_newest/tvm/private_test/test_meta_programming.py:16:9
    |  
 16 |          a = T.alloc_buffer(shape, dtype=dtype)
    |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
This PR fixes this by change "==" to "is".

### DIFF
--- a/python/tvm/script/parser/core/parser.py
+++ b/python/tvm/script/parser/core/parser.py
@@ -150,7 +150,7 @@ class VarTable:
             The options of whether variable shadowing allwed for this variable.
         """
         # Skip if the key and value are equal to those in the var_table
-        if self.name2value[var] and self.name2value[var][-1] == value:
+        if self.name2value[var] and self.name2value[var][-1] is value:
             return
         if allow_shadowing and var in self.frames[-1].vars:
             # Shadowing
--- a/python/tvm/script/parser/core/parser.py
+++ b/python/tvm/script/parser/core/parser.py
@@ -17,6 +17,7 @@
 """The core parser"""
 
 from collections import defaultdict
+import numpy as np
 from contextlib import contextmanager
 from typing import Any, Callable, Dict, List, Optional, Set, Union
 from tvm._ffi.base import TVMError
@@ -150,8 +151,11 @@ class VarTable:
             The options of whether variable shadowing allwed for this variable.
         """
         # Skip if the key and value are equal to those in the var_table
-        if self.name2value[var] and self.name2value[var][-1] is value:
-            return
+        if self.name2value[var] and type(self.name2value[var][-1]) == type(value):
+            if isinstance(value, np.ndarray) and (self.name2value[var][-1] == value).all():
+                return
+            elif self.name2value[var][-1] == value:
+                return
         if allow_shadowing and var in self.frames[-1].vars:
             # Shadowing
             self.name2value[var][-1] = value
--- a/python/tvm/script/parser/core/parser.py
+++ b/python/tvm/script/parser/core/parser.py
@@ -17,9 +17,9 @@
 """The core parser"""
 
 from collections import defaultdict
-import numpy as np
 from contextlib import contextmanager
 from typing import Any, Callable, Dict, List, Optional, Set, Union
+import numpy as np
 from tvm._ffi.base import TVMError
 
 from tvm.error import DiagnosticError
@@ -151,7 +151,7 @@ class VarTable:
             The options of whether variable shadowing allwed for this variable.
         """
         # Skip if the key and value are equal to those in the var_table
-        if self.name2value[var] and type(self.name2value[var][-1]) == type(value):
+        if self.name2value[var] and isinstance(self.name2value[var][-1], type(value)):
             if isinstance(value, np.ndarray) and (self.name2value[var][-1] == value).all():
                 return
             elif self.name2value[var][-1] == value:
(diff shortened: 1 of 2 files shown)


python/tvm/script/parser/core/parser.py
```diff
@@ -150,7 +150,7 @@ def add(self, var: str, value: Any, allow_shadowing: bool = False):
             The options of whether variable shadowing allwed for this variable.
         """
         # Skip if the key and value are equal to those in the var_table
-        if self.name2value[var] and self.name2value[var][-1] == value:
+        if self.name2value[var] and self.name2value[var][-1] is value:
             return
```
junrushao, line 154: Thanks for pointing out the issue! I believe either way might not be the most accurate, because it's possible that `self.name2valuep[var]` is a python integer or so, which cannot be compared using `is`. We might want to dispatch comparison according to different types

tests/python/unittest/test_tvmscript_regression.py
```diff
@@ -45,5 +45,20 @@ def test_multi_element_array_in_outmost_namespace():
...
+    @T.prim_func
+    def func_ref():
+        a = T.alloc_buffer([10, 10], dtype="int8")
+        T.evaluate(0)
+
+    tvm.ir.assert_structural_equal(test_case, func_ref)
+
+
 if __name__ == "__main__":
+    a = numpy.zeros((10, 10), dtype="int8")
```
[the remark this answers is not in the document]
lightzhan-intellif: I am very sorry that it can not in this case. If we do so, prim function can not capture the var 'a' because it is not a nonlocal variable of func test_case.

python/tvm/script/parser/core/parser.py
lightzhan-intellif: Thanks for your suggestion. I have done some trials in the python terminal according to your concern. Let's have a look:
```
# python3
Python 3.8.13 (default, Apr 19 2022, 00:53:22) 
[GCC 7.5.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> 1 is 1
<stdin>:1: SyntaxWarning: "is" with a literal. Did you mean "=="?
True
>>> a = 1
>>> a is 1
<stdin>:1: SyntaxWarning: "is" with a literal. Did you mean "=="?
True
>>> a = 1
>>> b = 1
>>> a is b
True
>>> a = [1, 2, 3]
>>> b = 1
>>> a[0] is b
True
>>> b = [1, 2, 3]
>>> a [0] is b[1]
False
```
According to the above output, we can find that there will be a warning if we use literal directly, but here it is a variable/list/dict which in your concern contains a literal. It looks like that python differentiates literal from variables with `literal` value. In our case, it belongs to the latter. So maybe no problem here with "is".

There might be some other scenarios I didn't cover, feel free to point out.

slyubomirsky: I believe the `is` operator checks reference equality rather than value equality. For integers, it will just check equality, so @lightzhan-intellif is correct. Whether it's preferred style is another question

junrushao: I do not have particular opinion on which style we should go for, but just wanted to point out the implication of the switch:

- `is` checks referential/pointer equality, or in python's term, identity, where it returns True only when `id(lhs) == id(rhs)`. It could depend on certain underlying implementation of the system, for example:

```python
>>> a = 257
>>> b = 257
>>> a is b
False
>>> a = 256
>>> b = 256
>>> a is b
True
```

- `==` checks equality that could be potentially overloaded, for TVM objects, it's using TVM's address comparison `.same_as()` rather than python's builtin `id()` method which is used in `is` operator. However

The implication of switching from `==` to `is` means that it bypasses TVM's `.same_as()` method, which at the moment I am not quite certain is suitable for broad usecases.

Therefore, how about we do the following: if `lhs` and `rhs` are numpy arrays, then we use numpy-specific behavior (e.g. elementwise equality), but otherwise we still use `==`.

lightzhan-intellif: Yea, I have updated the code.

junrushao (approved):
LGTM! Please fix the lint and I'm happy to get it in

lightzhan-intellif:
> LGTM! Please fix the lint and I'm happy to get it in

Done

junrushao:
Thanks @lightzhan-intellif @slyubomirsky @Hzfengsy for the discussion!

PR #12515, opened 2022-08-20
[TIR][Arith] Add more strict checking in imm construction and folding.

wrongtest-intellif:
Refer to issues https://github.com/apache/tvm/issues/12377 and https://github.com/apache/tvm/issues/12378. The behavior of compile time constant folding and general path computation differs.

The changes are:

- Add more strict value range checking when build IntImm object.

- For constant folding on integers
    - Mimic intN arithmetic on intN folding, instead of direct int64 arithmetic.

- For constant folding on floats
    - Only support float32 and float64, use float32 arithmetic on float32 folding, instead of float64 arithmetic.
    - Temporarily forbid fold float16 since we do not have implemented fp16 arithmetics yet and there may exist compatibility issues on different targets.

cc @junrushao1994

PR #13631, opened 2022-12-16
[Arith] Allow const folding on fp16 involving one and zero

masahi:
Since https://github.com/apache/tvm/pull/12515, const folding on fp16, when both lhs and rhs are constant, is disabled. This is the case even when one of the arguments are one or zero. This leads to a real problem described in https://github.com/apache/tvm/pull/13532#issuecomment-1353715208. So in practice, compiling winograd fp16 on GPU is currently broken. 

This PR enables trivial algebraic simplification for fp16 const folding when (1) both arguments are constant and (2) one or both of the inputs are one or zero. Since we already do such simplification for any dtype when only one of the inputs are constant (for example https://github.com/apache/tvm/blob/3a639a41a830ece914df10a98f807123e46cd750/src/arith/const_fold.h#L209-L216), I just removed all early exit `return NullOpt` for fp16 cases and use the code path like above to simplify `0h * 2.5h` to `0h` etc.

cc @vinx13 @wrongtest-intellif @Lunderberg

### DIFF
--- a/src/arith/const_fold.h
+++ b/src/arith/const_fold.h
@@ -142,8 +142,6 @@ inline Optional<PrimExpr> TryConstFold<tir::Add>(PrimExpr a, PrimExpr b) {
                                                        static_cast<float>(fb->value)));
       } else if (rtype.bits() == 64) {
         return FloatImm(rtype, fa->value + fb->value);
-      } else {
-        return NullOpt;
       }
     }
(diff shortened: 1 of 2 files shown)


leandron (approved):
LGTM, thanks @masahi

vinx13 (approved):

PR #13637, opened 2022-12-16
[ETHOSN] Update driver stack version to 22.11

ashutosh-arm:
NPU driver updated to 22.11.

New process memory allocator is used to create
buffers and networks. Support for 22.08 stack has
been kept intact in the sources and tests until the new
docker image is built and starts getting used. Tests
were modified to meet limitations imposed on input
zero point and kernel size by NPU software. Removed
defining ETHON_API_VERSION from cmake infra.

### DIFF
--- a/cmake/utils/FindEthosN.cmake
+++ b/cmake/utils/FindEthosN.cmake
@@ -58,18 +58,6 @@ macro(find_ethosn use_ethosn)
       PATHS ${__ethosn_stack}/lib)
     find_library(ETHOSN_COMPILER_LIBRARY NAMES EthosNSupport)
 
-    list(GET ETHOSN_INCLUDE_DIRS 0 filename)
-    set(filename "${filename}/ethosn_support_library/Support.hpp")
-    file(READ ${filename} ETHOSN_SUPPORT_H)
-    string(REGEX MATCH "VERSION_MAJOR ([0-9]*)" _ ${ETHOSN_SUPPORT_H})
-    set(ver_major ${CMAKE_MATCH_1})
-    string(REGEX MATCH "VERSION_MINOR ([0-9]*)" _ ${ETHOSN_SUPPORT_H})
-    set(ver_minor ${CMAKE_MATCH_1})
-    string(REGEX MATCH "VERSION_PATCH ([0-9]*)" _ ${ETHOSN_SUPPORT_H})
-    set(ver_patch ${CMAKE_MATCH_1})
-    set(ETHOSN_PACKAGE_VERSION "${ver_major}.${ver_minor}.${ver_patch}")
-    set(ETHOSN_DEFINITIONS -DETHOSN_API_VERSION=${USE_ETHOSN_API_VERSION})
-
     # Runtime hardware support. Driver library also needed for
     # test support.
     find_path(_DL_DIR NAMES Network.hpp
@@ -81,9 +69,7 @@ macro(find_ethosn use_ethosn)
       PATHS ${__ethosn_stack}/lib)
     find_library(ETHOSN_RUNTIME_LIBRARY NAMES EthosNDriver)
     if(${USE_ETHOSN_HW} MATCHES ${IS_TRUE_PATTERN})
-      set(ETHOSN_DEFINITIONS -DETHOSN_HW -DETHOSN_API_VERSION=${USE_ETHOSN_API_VERSION})
-    else()
-      set(ETHOSN_DEFINITIONS -DETHOSN_API_VERSION=${USE_ETHOSN_API_VERSION})
+      set(ETHOSN_DEFINITIONS -DETHOSN_HW)
     endif()
 
     if(ETHOSN_COMPILER_LIBRARY)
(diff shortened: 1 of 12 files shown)


cmake/utils/FindEthosN.cmake
```diff
@@ -58,18 +58,6 @@ macro(find_ethosn use_ethosn)
...
-    set(filename "${filename}/ethosn_support_library/Support.hpp")
-    file(READ ${filename} ETHOSN_SUPPORT_H)
-    string(REGEX MATCH "VERSION_MAJOR ([0-9]*)" _ ${ETHOSN_SUPPORT_H})
-    set(ver_major ${CMAKE_MATCH_1})
-    string(REGEX MATCH "VERSION_MINOR ([0-9]*)" _ ${ETHOSN_SUPPORT_H})
-    set(ver_minor ${CMAKE_MATCH_1})
-    string(REGEX MATCH "VERSION_PATCH ([0-9]*)" _ ${ETHOSN_SUPPORT_H})
-    set(ver_patch ${CMAKE_MATCH_1})
-    set(ETHOSN_PACKAGE_VERSION "${ver_major}.${ver_minor}.${ver_patch}")
-    set(ETHOSN_DEFINITIONS -DETHOSN_API_VERSION=${USE_ETHOSN_API_VERSION})
```
lhutton1, line 71: Thanks for clearing this up :)

tests/python/contrib/test_ethosn/test_conv2d_transpose.py
```diff
@@ -115,7 +116,7 @@ def _get_model(
     [
         ((1, 2, 2, 1), (2, 2), (1, 1), 1, False),
         ((1, 2, 2, 5), (2, 2), (3, 5), 4, False),
-        ((1, 7, 7, 4), (2, 2), (7, 9), 8, True),
+        ((1, 7, 7, 4), (2, 2), (7, 7), 8, True),
```
lhutton1, line 119: Curious, does this come from a change in the driver stack itself? It wasn't mentioned here: https://github.com/ARM-software/ethos-n-driver-stack/blob/main/CHANGELOG.md#2211 and, if so, this is wrong: https://github.com/ARM-software/ethos-n-driver-stack/blob/main/SUPPORTED.md?plain=1#L133

tests/python/contrib/test_ethosn/test_conv2d_transpose.py
```diff
@@ -169,6 +170,72 @@ def test_conv2d_transpose(ifm_shape, strides, kernel_size, out_channels, dtype,
     tei.verify(outputs, dtype, 1)
 
 
+@pytest.mark.skipif(
+    ethosn_api_version() == "3.2.0",
+    reason="Skip because NPU driver 22.11 does not support following cases.",
+)
+@requires_ethosn
```
lhutton1, line 177: I had some trouble skipping like this in the past due to the way decorators are evaluated, might need to put the skip inside the function

tests/python/contrib/test_ethosn/test_leaky_relu.py
```diff
@@ -55,9 +56,12 @@ def test_leaky_relu(dtype, shape, alpha):
     iinfo = np.iinfo(dtype)
     zp_min = iinfo.min
     zp_max = iinfo.max
-    input_zp = zp_min + 120
+    if ethosn_api_version() == "3.2.0":
+        input_zp = zp_min + 128
+    else:
+        input_zp = zp_min + 120
```
lhutton1, line 62: Probably not worth the conditional, assuming `zp_min + 120` works for both the previous and new version?

lhutton1 (commented):
Thanks @ashutosh-arm, it definitely seems like there was more work involved in this upgrade compared to previous revisions! LGTM overall, just some small things

tests/python/contrib/test_ethosn/test_leaky_relu.py
ashutosh-arm: I thought after this is fixed in 23.02, it will be worth having both the cases as a test.

tests/python/contrib/test_ethosn/test_conv2d_transpose.py
ashutosh-arm: Thanks for catching it. It cause CI failure in ci_arm.

tests/python/contrib/test_ethosn/test_conv2d_transpose.py
ashutosh-arm: An issue has been raised with the NPU software to track this.

cmake/utils/FindEthosN.cmake
ashutosh-arm: :smile:

tests/python/contrib/test_ethosn/test_leaky_relu.py
lhutton1: Ah okay make sense, let's leave it for now then :)

lhutton1 (approved):
Thanks for battling with CI to get this one green @ashutosh-arm! LGTM! Just a quick reminder that we'll also need to create an issue to track the docker image upgrade to use v22.11 :)

PR #13641, opened 2022-12-19
[Bug][CodeGen,Cuda]fix cast fp16 to int8/uint8 in cuda

ninesheep:
construct a module like below:
```python
import tvm
from tvm import relay

p0 = relay.var("p0", shape = [32,], dtype="float16")
p1 = relay.var("p1", shape = [32,], dtype="float16")

x1 = relay.multiply(p0, p1)
x2 = relay.round(x1)
x3 = relay.cast(x2, "uint8")  # or int8
func = relay.Function(relay.analysis.free_vars(x2), x2)
mod = tvm.IRModule.from_expr(func)

with tvm.transform.PassContext(
    opt_level=3,
    ):
    graph, lib, params = tvm.relay.build_module.build(
        new_mod, target="cuda --host=llvm", params=None
    )
```

and build it, it will report a error like this:
image

@wrongtest-intellif


ninesheep:
cc @wrongtest-intellif

wrongtest-intellif (approved):
LGTM

Issue #4052, opened 2019-10-03
[RFC] Tensor Core Support


Hzfengsy:
Tensor Core is a defining feature of the NVIDIA new Volta and Turing GPU Architecture, which gives a massive boost for matrix multiplication and convolution. Tensor Cores enable us to use mixed-precision to achieve higher throughput without sacrificing accuracy.

## Tensor Core Overview
Each Tensor Core provides a 4×4×4 matrix processing array that operates `D = A * B + C`, where `A`, `B`, `C` and `D` are 4×4 matrices as Figure shows. The matrix multiply inputs A and B are FP16 matrices, while the accumulation matrices C and D may be FP16 or FP32 matrices.

![image](https://user-images.githubusercontent.com/25500082/66098090-1263f680-e556-11e9-8fc3-0c97917d43c9.png)
However, CUDA programmers can only use warp-level primitive `wmma::mma_sync(acc_frag, a_frag, b_frag, acc_frag)` to perform 16×16×16 half-precision matrix multiplication on tensor cores. Before invoking the matrix multiplication, programmers must load data from memory into registers with primitive `wmma::load_matrix_sync`, explicitly. The NVCC compiler translates that primitive into multiple memory load instructions. At run time, every thread loads 16 elements from matrix A and 16 elements from B.

# Proposed Design
It can be regarded as a new hardware instruction just like gemm instruction in vta. So it is easy to use `tensorization` to replace the code. Note that unlike other accelerators, we need also to consider the shared memory when we use tensor cores. Also, `wmma::mma_sync` is a wrap-level instruction, which means it will call all threads (32 threads) in a warp. It is a brand new schedule level.

## Warp Level Schedule
Although `wmma::mma_sync` is a warp-level operator, NVIDIA doesn't change the API for kernel launch. It still uses `gridDim`, `blockDim` and `dynamic shared memory` (optional) to launch a kernel. The only thing we should do is ensuring `blockDim.x` be a multiple of warp size(32).

In tvm schedule, we can just make the extent of `threadIdx.x` equals 32. It's safe if we want to use `threadIdx.y` and `threadIdx.z`, and their extents have no extra constraint. Note that, `threadIdx.x` can be only used at memory copy or other thread-level operators.

## New Memory Scope
[...]

PR #13642, opened 2022-12-19
[Tensorize][runtime] Add support for AMX(Advanced Matrix Extensions) through Tensor intrinsics

Qianshui-Jiang:
In this PR, we add support for AMX(Advanced Matrix Extensions) through Tensor intrinsics;

we chose to enabled the AMX support and INT8 intrinsics and dense op at first, which includes:
1. Add building options and LLVM AMX intrinsic target detection ;
2. Register a global AMX init and config function;
3. LLVM AMX intrinsic integrated as Tensor intrinsic (1 for microkernel computation and 1 for accumulation tile store);
4. Add test case on u8s8s32 matmul usding the AMX Tensor intrinsics;
5. Integration of int8 dense kernel,  testcase also added;

BF16 and more op are going to be considered in the future.


python/tvm/topi/x86/dense_alter_op.py
```diff
@@ -25,13 +25,15 @@
...
 from .. import nn
 
 
-def check_vnni_applicable(x, y, allow_padding=False):
+def check_inst_applicable(x, y, allow_padding=False):
     mcpu = tvm.target.Target.current().mcpu
+    cpu_avai = target_has_vnni(mcpu) or target_has_amx(mcpu)
     return (
-        target_has_vnni(mcpu)
+        cpu_avai
```
cbalint13, line 36: * ~cpu_avai~ may sound strange, ```cpu_simd``` ?

src/runtime/contrib/amx/amx_config.cc
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  _tile_loadconfig(dst->a);
+}
+
+TVM_REGISTER_GLOBAL("runtime.amx_tileconfig").set_body([](TVMArgs args, TVMRetValue* rv) {
+  int rows = args[0];
+  int cols = args[1];
+  LOG(INFO) << "rows: " << rows << ", cols:" << cols;
+  // -----------Config for AMX tile resgister----------------------
```
cbalint13, line 83: s|resgister|register|

cbalint13 (changes requested):
Thank you @Qianshui-Jiang  for this latest x86 SIMD feature addition !

* Just two cosmetic proposal
* The [autoschedule](https://github.com/apache/tvm/blob/main/python/tvm/tir/tensor_intrin/x86.py) + [autotensorise](https://github.com/apache/tvm/blob/main/tests/python/integration/test_auto_tensorize.py) parts can be extended, maybe later, in a subsequent PR.

tests/python/contrib/test_gemm_acc32_vnni.py
```diff
@@ -115,5 +115,5 @@ def verify(target="llvm -mcpu=cascadelake"):
     # The test requires Cascade Lake and newer Intel machines to generate the
     # correct AVX512 VNNI instruction. So, disabling the test.
 
-    # test_fc_int8_acc32()
+    # test_fc_int8_acc32()``
```
cbalint13, line 118: * spurious change ?

Qianshui-Jiang: thank you for comment, already fixed.

python/tvm/topi/x86/dense_alter_op.py
Qianshui-Jiang: yep, replaced `cpu_avai` to  `simd_avai` to check if vnni or amx was availiable  in this target.

cbalint13 (approved):
Looks good to me.

@Qianshui-Jiang , just curiousity (offtopic):
* Do you have a TVM number for AMX performance (e.g. vs. VNNI) ?

Qianshui-Jiang:
@cbalint13 Thanks for your curiousity!
In our test, Compared to TVM testcase `test_fc_int8_acc32()`, AMX int8 dense performes about 8X faster in single physical core.
(Since AMX machine not officially lunched yet, I can only gives a little perf data, but it's comming soon~)

masahi:
@Qianshui-Jiang Can we test TVM-generated AMX code using the emulator (SDE)?

Qianshui-Jiang:
@masahi  Sure, you can use SDE by assign platform info like `-spr`, 
But if you are not built on 5.16 kernel, you have to comment all the system related code `runtime.amx_init` , 
and don't need to init amx before the testcase when using SDE;

Qianshui-Jiang:
@masahi, BTW, SDE would be very slow, the repeat times of time_evaluator should be small.

python/tvm/topi/x86/dense.py
```diff
@@ -373,6 +375,153 @@ def _callback(op):
     return s
 
 
+def dense_amx_int8_compute(cfg, data, packed_w, bias=None):
```
masahi, line 378: This can be shared with the vnni compute.

python/tvm/topi/x86/dense.py
```diff
@@ -373,6 +375,153 @@ def _callback(op):
...
+def schedule_dense_amx_int8(cfg, outs):
+    """Create a schedule for dense_amx_int8"""
+    s = te.create_schedule([x.op for x in outs])
+
+    def _callback(op):
+        if "dense_amx_int8" in op.tag:
+            dense_amx_int8_schedule(cfg, s, op.output(0), outs[0])
+
+    traverse_inline(s, outs[0].op, _callback)
+    return s
```
masahi, line 522: For the two functions above, try removing dup with VNNI ones.

python/tvm/topi/x86/tensor_intrin.py
```diff
@@ -348,3 +348,227 @@ def _instr(index):
...
+    (Physically, to efficiently leveraging the tile register, we constructing a 2x2 tiles
+    matmul which performs 32x128x32 in total)
+
+    The pseudo code is as follows:
+        for(k=0; k<2; k++){
+            for(n=0; n<2; n++){
+                tileload64(tmm_b, B)
+                for(m=0; m<2; m++){
+                    if(n==0)
+                        tileload64(tmm_a, A)
```
masahi, line 368: Have you considered tensorizing at finer granularity? Meaning, instead of tensoring the whole load + compute as one tensor intrin, tensorize each load and compute separately. That could make the hard coded outer-loop factors (2 in this intrin) tunable.

masahi (changes requested):

python/tvm/topi/x86/tensor_intrin.py
cbalint13: @masahi ,

I have a question related to your proposed here:

* Is there a TIR _desc equivalent for TOPI's body, reset, update ?
     topi intrin decl:  https://github.com/apache/tvm/blob/main/python/tvm/topi/x86/tensor_intrin.py#L140
* How the ```accum``` steps (very inner) can be succeeded by ```update``` (outer) or preceded by ```reset``` ?

I am looking for any TVM reference (code, example) for this, trying even finer granularity control of tensorization.

masahi: If you are asking about tensorization in TIR, there is no need to have separate init, update etc. See https://github.com/apache/tvm/blob/main/python/tvm/tir/tensor_intrin/x86.py. The accum buffer is implicitly zero-initialized and we can direct do update via `+=`.

cbalint13: > If you are asking about tensorization in TIR, there is no need to have separate init, update etc. See https://github.com/apache/tvm/blob/main/python/tvm/tir/tensor_intrin/x86.py. The accum buffer is implicitly zero-initialized and we can direct do update via `+=`.

@masahi ,
Thanks a lot ! I've noticed now "+="  for ```update``` purpose in TIR description.
Will also look @ MultiLevelTiling() (cuda only) idea for custom ISA having special ```init```, ```load_*```, ```store```.

Qianshui-Jiang: > Have you considered tensorizing at finer granularity? Meaning, instead of tensoring the whole load + compute as one tensor intrin, tensorize each load and compute separately. That could make the hard coded outer-loop factors (2 in this intrin) tunable.

Yes we considered to make it works like tensor core which mentioned here: https://github.com/apache/tvm/issues/4052, but the problem is AMX intrinsics in LLVM need user to assigned register(tmm) number by hand craft. If we seperate the load and compute, we have to leave the register assginment to schedule method, that would bring much redundancy work. So finnally we decide to follow BRGEMM and leave the whole smallest comute unit for tensorizing.

python/tvm/topi/x86/dense.py
Qianshui-Jiang: yep, already unified the AMX and VNNI interface in x86 strategy.

python/tvm/topi/x86/dense.py
Qianshui-Jiang: already unified AMX and VNNI to `dense_int8`

python/tvm/topi/x86/tensor_intrin.py
cbalint13: @masahi ,
@Qianshui-Jiang ,

Don't want hold up the PR, I think it is fine to merge with this start AMX implementation.
I  leave here a possible way to workaround the rigidity of LLVM's AMX related intrinsic declaration:

* Use ```tvm.tir.call_extern``` insead of ```tvm.tir.call_llvm_intrin``` like [here](https://github.com/apache/tvm/blob/8551a5c71faaa2eedc8b1c49a39be0fe4688c021/python/tvm/topi/arm_cpu/mprofile/dsp/micro_kernel/gemm.py#L107-L109) .
* You can than pass argument (e.g. register number, etc ...) into the custom called function.
* You can use the argument inside that function and emit the AMX ISA as bytes: [ __asm(.word)](https://github.com/cbalint13/OLIMP/blob/main/demo/icebreaker/src/rv32-custom.h#L52-L53)

This is how it is done to emit custom part of RISC-V ISA, that is also too rigid using just the LLVM declarations.

Qianshui-Jiang: @cbalint13, Thanks a lot for your advices. 
Regarding that in TVM we already have DNNL codegen and some external call to DNNL, which is an off-the -shelf compute lib leverages SIMD ISA transparently.
The main focus in this PR is to add more native capability for TVM to generate AMX instructions. 
AFAIK, the LLVM features for AMX are also moving forward such as register allocation for intrinsics and auto vectorization.
Hence we choose to call LLVM intrinsics directly and hope to have more flexibility in the future.

python/tvm/topi/x86/dense.py
```diff
@@ -293,16 +324,13 @@ def dense_vnni_compute(cfg, X, packed_w, bias=None):
             ),
             axis=ak,
         ),
-        tag="dense_vnni",
-        attrs={"schedule_rule": "dense_vnni"},
```
masahi, line 297: This shouldn't be removed, it is used here https://github.com/apache/tvm/blob/b16a64d6edb9fd1a014fc51995dff7d0e2f4c84e/tests/python/unittest/test_meta_schedule_vnni_integration.py#L198.

Since this only affects MetaSchedule, you don't have to provide this value for AMX. So only when `dense_int8_compute` is called for VNNI, you need to provide this attribute.

masahi (changes requested):
Also please verify that the VNNI test https://github.com/apache/tvm/blob/b16a64d6edb9fd1a014fc51995dff7d0e2f4c84e/tests/python/unittest/test_meta_schedule_vnni_integration.py is still functional after this change.

python/tvm/topi/x86/dense.py
Qianshui-Jiang: @masahi Is this case only use the x86 int8 compute method and inject a particular TIR scheduling? Can we just change the attribute `dense_vnni` to `dense_int8` which used here?

masahi: Yes, but it is important that we specify that we use this compute for VNNI. If the schedule rule annotation only says "dense_int8", we don't know which intrinsic to tensorize this compute.

Qianshui-Jiang: > This shouldn't be removed, it is used here
> 
> https://github.com/apache/tvm/blob/b16a64d6edb9fd1a014fc51995dff7d0e2f4c84e/tests/python/unittest/test_meta_schedule_vnni_integration.py#L198
> 
> .
> Since this only affects MetaSchedule, you don't have to provide this value for AMX. So only when `dense_int8_compute` is called for VNNI, you need to provide this attribute.

@masahi Sorry, may given the misunderstanding, I mean that can we use the `dense_int8` in this test case?  cuz here inject the VNNI intrinsic explicitly. 
And by default in relay build flow, it will check if the VNNI or AMX is availiable and chose different schedulling.
I've verified that this test case still functional after the little modification [in this commit](https://github.com/apache/tvm/pull/13642/commits/5718a059c69972cf71ea082a3303b5c29fa2d21f)  bellow

masahi: Of course the test still works, because it was already written for VNNI. The point is that the name `dense_vnni` tells that the particular TE expression is meant for VNNI tensorization, in particular the weight is pre-packed appropriately. But a generic name like `dense_int8` doesn't provide such information. 

So please revert that commit and restore and pass `attrs={"schedule_rule": "meta_schedule.x86.dense_vnni"}` when you create a compute expression for VNNI. `schedule_rule` is not relevant for AMX for now.

Qianshui-Jiang: @masahi  yep, got it, the schedule rule for ms dense_vnni are restored.  
Keep it remained but AMX not use it for now. 
( A smalll question: How do we know inside of this compute expression if it's created for VNNI not AMX? )

masahi: I suggest checking the target string in the op strategy, and create separate compute for VNNI or AMX (rather than using the same function, `dense_int8`, for both).

python/tvm/relay/op/strategy/x86.py
```diff
@@ -591,18 +591,17 @@ def dense_strategy_cpu(attrs, inputs, out_type, target):
 def dense_pack_strategy_cpu(attrs, inputs, out_type, target):
     """dense_pack x86 strategy"""
     strategy = _op.OpStrategy()
-
     if (
         inputs[0].dtype == "uint8"
         and inputs[1].dtype == "int8"
         and out_type.dtype == "int32"
         and attrs["weight_layout"] == "NC16n4c"
     ):
```
masahi, line 599: Let's check the target mcpu here and pass VNNI or AMX attribute to `dense_int8`.

Qianshui-Jiang: @masahi already added the target checking, seems generic compute wrapper didn't have the interface. but it's the same if checking inside of `dense_int8`?

apps/microtvm/zephyr/template_project/qemu-hack/qemu-system-arm
```diff
@@ -1 +0,0 @@
-qemu-system-i386
```
masahi, line 1: Is this change relevant? If not remove it.

python/tvm/relay/op/strategy/x86.py
masahi: I see, that works too.

masahi (changes requested):
Are the changes in `app/` directory intended?

apps/microtvm/zephyr/template_project/qemu-hack/qemu-system-arm
Qianshui-Jiang: @masahi  by misoperation, already removed.

masahi (approved):

PR #13643, opened 2022-12-19
[CMSIS-NN] Add a runtime error message

NicolaLancellotti:
This pr generates the code to catch and save, through TVMAPISetLastError, a CMSIS-NN error. 
In case an error is raised, it is retrieved with TVMGetLastError in the AOT test runner, and then it is printed.

@ashutosh-arm

### DIFF
--- a/python/tvm/testing/aot.py
+++ b/python/tvm/testing/aot.py
@@ -23,7 +23,7 @@ import re
 import subprocess
 import tarfile
 import logging
-from typing import Any, NamedTuple, Union, Tuple, Optional, List, Dict
+from typing import Any, NamedTuple, Union, Tuple, Optional, List, Dict, Callable
 import numpy as np
 
 import tvm
@@ -728,6 +728,7 @@ def run_and_check(
     test_dir: str = None,
     verbose: bool = False,
     use_workspace_io: bool = False,
+    checker: Optional[Callable[[str], bool]] = None,
 ):
     """
     This method uses the original test data and compiled runtime.Modules
@@ -812,6 +813,9 @@ def run_and_check(
             debug_last_error,
         )
 
+        if checker and (not checker(base_path)):
+            return False
+
         # Verify that compiles fine
         file_dir = os.path.dirname(os.path.abspath(__file__))
         makefile_dir = os.path.join(file_dir, "../../../tests/python/relay/aot")
@@ -859,11 +863,13 @@ def run_and_check(
         with open(run_log_path) as run_log:
             assert AOT_SUCCESS_TOKEN in run_log.read()
 
+        return True
+
     if test_dir is None:
         tmpdir = utils.tempdir()
-        run_and_check_body(os.path.join(tmpdir.path, "test"))
+        return run_and_check_body(os.path.join(tmpdir.path, "test"))
     else:
-        run_and_check_body(test_dir)
+        return run_and_check_body(test_dir)
 
 
 def compile_and_run(
@@ -882,7 +888,8 @@ def compile_and_run(
     test_dir: str = None,
     verbose: bool = False,
     schedule_name: str = None,
-):
+    checker: Optional[Callable[[str], bool]] = None,
+) -> bool:
     """This is a wrapper API to compile and run models as test for AoT
 
     Parameters
@@ -913,7 +920,7 @@ def compile_and_run(
         schedule_name=schedule_name,
     )
 
-    run_and_check(
+    return run_and_check(
         models=compiled_test_mods,
         runner=runner,
         interface_api=interface_api,
@@ -923,6 +930,7 @@ def compile_and_run(
         data_linkage=data_linkage,
         test_dir=test_dir,
         verbose=verbose,
+        checker=checker,
     )
 
 
--- a/python/tvm/testing/aot.py
+++ b/python/tvm/testing/aot.py
@@ -200,6 +200,7 @@ def _emit_main_prologue(
     compiled_models,
     interface_api,
     use_stack_allocator=True,
+    debug_last_error=False,
 ):
     if use_stack_allocator:
         workspace_define = f"#define WORKSPACE_SIZE ({workspace_bytes}"
@@ -243,11 +244,28 @@ void TVMLogf(const char* msg, ...) {
   va_start(args, msg);
   vfprintf(stdout, msg, args);
   va_end(args);
-}\n
-TVM_DLL int TVMFuncRegisterGlobal(const char* name, TVMFunctionHandle f, int override) {}
-int main(){\n
+}
     """
     )
+    if debug_last_error:
+        main_file.write(
+            """\n
+tvm_crt_error_t TVMPlatformTimerStart() {
+  return kTvmErrorFunctionCallNotImplemented;
+}
+tvm_crt_error_t TVMPlatformTimerStop(double* elapsed_time_seconds) {
+  return kTvmErrorFunctionCallNotImplemented;
+}
+const TVMModule* TVMSystemLibEntryPoint(void) { return NULL; }
+"""
+        )
+    else:
+        main_file.write(
+            """\n
+TVM_DLL int TVMFuncRegisterGlobal(const char* name, TVMFunctionHandle f, int override) {}
+"""
+        )
+    main_file.write("\nint main(){\n")
     main_file.write(custom_prologue)
 
 
@@ -332,10 +350,10 @@ def _emit_main_data_setup(main_file, input_map, output_map, mod_name):
 
 
 def _emit_main_c_interface_call(
-    main_file, devices, workspace_pool_names, mod_name, use_workspace_io
+    main_file, devices, workspace_pool_names, mod_name, use_workspace_io, debug_last_error
 ):
     sub_strings = list()
-    sub_strings.append(f'{_mangle_name(mod_name,"run")}(')
+    sub_strings.append(f'if ({_mangle_name(mod_name,"run")}(')
     if not use_workspace_io:
         sub_strings.append(f'&{_mangle_name(mod_name,"inputs")}, ')
         sub_strings.append(f'&{_mangle_name(mod_name,"outputs")}, ')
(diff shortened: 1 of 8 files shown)


python/tvm/testing/aot.py
```diff
@@ -244,8 +244,15 @@ def _emit_main_prologue(
...
-TVM_DLL int TVMFuncRegisterGlobal(const char* name, TVMFunctionHandle f, int override) {}
+}
+tvm_crt_error_t TVMPlatformTimerStart() {
+    return kTvmErrorFunctionCallNotImplemented;
+}
+tvm_crt_error_t TVMPlatformTimerStop(double* elapsed_time_seconds) {
+  return kTvmErrorFunctionCallNotImplemented;
+}
+const TVMModule* TVMSystemLibEntryPoint(void) { return NULL; }
+
```
ashutosh-arm, line 255: Should we be adding error set/get apis instead?

python/tvm/testing/aot.py
```diff
@@ -455,6 +465,7 @@ def _emit_main_common_includes(main_file, custom_includes):
     main_file.write("#include <math.h>\n")
     main_file.write('#include "tvm/runtime/c_runtime_api.h"\n')
     main_file.write('#include "tvm/runtime/crt/stack_allocator.h"\n')
+    main_file.write('#include "tvm/runtime/crt/module.h"\n')
```
ashutosh-arm, line 468: interesting, I thought `c_runtime_api.h` would be enough :smile_cat:

src/relay/backend/contrib/cmsisnn/tir_to_runtime.cc
```diff
@@ -355,13 +355,7 @@ class CodeGenCMSISNN : public codegen::CodeGenCHost {
...
     stream << "&" << bias_dim << ", " << bias_data << ", ";
     stream << "&" << output_dim << ", " << output_data << ");\n";
-    PrintIndent();
-    stream << "if (status != ARM_CMSIS_NN_SUCCESS) {\n";
-    PrintIndent();
-    PrintIndent();
-    stream << "return -1;\n";
-    PrintIndent();
-    stream << "}\n";
+    EmitCheckError();
```
ashutosh-arm, line 358: nit: does `EmitErrorCheck` sound better? your call.

ashutosh-arm (changes requested):
Thanks @NicolaLancellotti for these changes. I am sure they will bring transparency to the error raised by the CMSIS-NN library. I've left a few comments below.

python/tvm/testing/aot.py
NicolaLancellotti: `TVMGetLastError` and `TVMAPISetLastError` are defined in `crt_runtime_api.c`.

src/relay/backend/contrib/cmsisnn/tir_to_runtime.cc
NicolaLancellotti: I agree, I'll update the pr.

python/tvm/testing/aot.py
NicolaLancellotti: `tvm/runtime/crt/module.h` defines `TVMModule` needed by
`TVMModule* TVMSystemLibEntryPoint(void)`.

ashutosh-arm:
Thanks @NicolaLancellotti for addressing the comments. Let's also hear from @Mousius. Alternative, here is to emit Get/Set APIs from the AOT Test runner infra.

apps/microtvm/cmsisnn/Makefile
```diff
@@ -69,6 +69,18 @@ $(BUILD_DIR)/crt_backend_api.o: $(STANDALONE_CRT_PATH)/src/runtime/crt/common/cr
...
+	$(QUIET)$(CC) -c $(PKG_CFLAGS) -o $@  $^
+
+$(BUILD_DIR)/func_registry.o: $(STANDALONE_CRT_PATH)/src/runtime/crt/common/func_registry.c
+	$(QUIET)mkdir -p $(@D)
+	$(QUIET)$(CC) -c $(PKG_CFLAGS) -o $@  $^
+
+$(BUILD_DIR)/ndarray.o: $(STANDALONE_CRT_PATH)/src/runtime/crt/common/ndarray.c
+	$(QUIET)mkdir -p $(@D)
+	$(QUIET)$(CC) -c $(PKG_CFLAGS) -o $@  $^
+
```
Mousius, line 83: I'm assuming these are dependencies for `crt_runtime_api.c`? We should raise an issue to remove that dependency if we can?

@areusch had some thoughts awhile back on this https://discuss.tvm.apache.org/t/pre-rfc-api-change-formalizing-c-backend-api/10380

src/relay/backend/contrib/cmsisnn/tir_to_runtime.cc
```diff
@@ -478,10 +466,28 @@ class CodeGenCMSISNN : public codegen::CodeGenCHost {
...
+    };
+
     PrintIndent();
-    stream << "if (status != ARM_CMSIS_NN_SUCCESS) {\n";
+    stream << "switch (status) {\n";
+    PrintIndent();
+    PrintIndent();
+    stream << "case ARM_CMSIS_NN_SUCCESS: break;\n";
+    PrintIndent();
+    PrintIndent();
```
Mousius, line 483: Potential fun exercise for the future, add a parameter to `PrintIndent` for the number of indents 😸

Mousius (changes requested):
This is the right thing to be doing, but we should be doing it without impacting any existing users or adding unnecessary bloat to use-cases that don't want it. Can we add a flag to enable/disable this functionality and check that we don't need to add any extra files when it's not enabled?

apps/microtvm/cmsisnn/Makefile
NicolaLancellotti: Yes, they are dependencies for `crt_runtime_api.c`  which defines `TVMAPISetLastError` and `TVMGetLastError`.

Instead of depending on `crt_runtime_api.c` we can generate the definitions of these functions as @ashutosh-arm  thought: https://github.com/apache/tvm/pull/13643#discussion_r1052140042, what do you think?

Mousius: If we don't emit the lines that set them, we shouldn't need the definitions at all?

NicolaLancellotti: I am not sure I understood, can you elaborate?

In this pr, we save the error through `TVMAPISetLastError` and then we retrieve it through `TVMGetLastError`, so we need the definitions of these two functions.
In the current implementation the functions are defined in `crt_runtime_api.c`, the problem is that this file has other functions, which we do not use, that depend on functions defined in `func_registry.c` and `ndarray.c`.
Ashutosh proposed not to use the definitions in `crt_runtime_api.c`, but to just emit the definitions of these two functions in `aot.py`.

Mousius: There are two separate issues here:
1. The C Runtime doesn't make it easy to use only the bits you want so you suddenly need array handling if you want error handling - this is a future problem if we want to tackle it
2. If we emit calls to these functions we're breaking anyone who currently compiles their model with CMSIS-NN - we need to have a way for users to specify that they don't want the debugging behaviour and then we don't emit the calls to the `TVMAPISetLastError`/`TVMAPIGetLastError` apis.

NicolaLancellotti: I've added an option `debug_last_error` to `CMSISNNCompilerConfigNode`.
The calls to `TVMAPISetLastError` and `TVMAPIGetLastError` are generated only if this option is true.
When the option is false we no longer need the dependencies: `crt_runtime_api.c`, `func_registry.c` and `ndarray.c`. We need them only when the option is true.
I've also added a test, that checks if a model compiles with both values of this option.

NicolaLancellotti: Ping for review @Mousius. :)

tests/python/contrib/test_cmsisnn/test_last_error.py
```diff
@@ -0,0 +1,119 @@
...
+        relay.const(out_scale, "float32"),
+        relay.const(out_zero_point, "int32"),
+    )
+    return make_qnn_relu(binary_op, relu_type, out_scale, out_zero_point, "int8")
+
+
+@skip_if_no_reference_system
+@tvm.testing.requires_cmsisnn
+@pytest.mark.parametrize("debug_last_error", [True, False])
+def test_last_error(debug_last_error):
```
Mousius, line 71: This checks it compiles with last error enabled but doesn't actually check the last error which is the most important functionality in this patch - can you check the error you're expecting is printed?

Mousius (changes requested):

tests/python/contrib/test_cmsisnn/test_last_error.py
NicolaLancellotti: I've added a test.

python/tvm/testing/aot.py
```diff
@@ -765,6 +793,8 @@ def run_and_check_body(base_path):
                 )
 
         use_usmp = runner.pass_config.get("tir.usmp.enable", False)
+        options = runner.pass_config.get("relay.ext.cmsisnn.options")
+        debug_last_error = options.get("debug_last_error", False) if options else False
```
Mousius, line 797: The AOT test runner shouldn't have knowledge of `CMSIS-NN` specific arguments, can we pass `debug_last_error` as a parameter?

Mousius (changes requested):
Nice @NicolaLancellotti, I like the `checker` pattern for AOT test utils 😸 would be a nice way to refactor some of our other tests.

Just one minor change requested to not add CMSIS-NN logic into the AOT test utils then I think we can land this and move forwards - sorry it's been a long one!

python/tvm/testing/aot.py
NicolaLancellotti: Done. Thanks for the review!

Mousius (approved):
LGTM! @ashutosh-arm can you take another look please?

ashutosh-arm (commented):
Thanks @Mousius and @NicolaLancellotti for following through. I just wanted to clarify about making changes to the test runner interface.

tests/python/contrib/test_cmsisnn/test_last_error.py
```diff
@@ -0,0 +1,164 @@
...
+            inputs=inputs,
+            outputs=output_list,
+            params=None,
+            output_tolerance=1,
+        ),
+        create_test_runner(compiler_cpu, cpu_flags, debug_last_error=debug_last_error),
+        interface_api,
+        use_unpacked_api,
+        debug_last_error=debug_last_error,
+        checker=checker,
```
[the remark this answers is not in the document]
ashutosh-arm: From offline discussion, I understand that the `checker` provides consistency to the CMSIS-NN testing and also allows for conditional execution if required in future.

ashutosh-arm (approved):

PR #13644, opened 2022-12-19
[CMSIS-NN] Add Cortex-M85 support

NicolaLancellotti:
This pr adds support for Cortex-M85.

Currently, there are no Cortex-M85 tests because another reference system is needed.

@ashutosh-arm


ashutosh-arm (approved):
Thanks @NicolaLancellotti for this support. It was long due :smiley_cat:

PR #13645, opened 2022-12-19
[microNPU] Add relu6 relu_n1_to_1 test cases for Ethos-U

Alexey-Yazev:
Tests are extended with cases with activations relu6 and relu_n1_to_1. Test cases contain conv2d operation + activation because separate activation is not offloaded to NPU.

### DIFF
--- a/python/tvm/relay/op/contrib/ethosu.py
+++ b/python/tvm/relay/op/contrib/ethosu.py
@@ -688,13 +688,15 @@ class BinaryElementwiseParams:
         clip = None
         requantize = None
 
-        if str(current_call.op) == "clip":
-            clip = current_call
-            current_call = clip.args[0]
-        elif str(current_call.op) == "qnn.requantize":
-            requantize = current_call
-            clip = current_call.args[0]
-            current_call = clip.args[0]
+        if is_quantized_operation:
+            if str(current_call.op) == "clip":
+                clip = current_call
+                current_call = clip.args[0]
+        else:
+            if str(current_call.op) == "qnn.requantize":
+                requantize = current_call
+                clip = current_call.args[0]
+                current_call = clip.args[0]
         binary_op = current_call
 
         layout = "NHWC"
@@ -927,9 +929,6 @@ class MinParams(BinaryElementwiseParams):
             [self.ifm, self.ifm2, self.ofm], supported_dtypes=[np.uint8, np.int8]
         ):
             return False
-        # MIN with different scales is not supported on NPU.
-        if self.ifm.q_params.scale_f32 != self.ofm.q_params.scale_f32:
-            return False
         return True
 
 
@@ -939,19 +938,10 @@ def minimum_pattern() -> tvm.relay.dataflow_pattern.DFPattern:
     """
     minimum = is_op("minimum")(wildcard(), wildcard())
     optional_min_clip = is_op("clip")(minimum)
-    return minimum | optional_min_clip
-
-
-def minimum_clip_requantize_pattern() -> tvm.relay.dataflow_pattern.DFPattern:
-    """
-    This function creates the pattern for minimum with fused RELU activation.
-    """
-    pattern = is_op("minimum")(wildcard(), wildcard())
-    pattern = is_op("clip")(pattern)
-    pattern = is_op("qnn.requantize")(
-        pattern, is_constant(), is_constant(), is_constant(), is_constant()
+    optional_min_clip = is_op("qnn.requantize")(
+        optional_min_clip, is_constant(), is_constant(), is_constant(), is_constant()
     )
-    return pattern
+    return minimum | optional_min_clip
 
 
 class MaxParams(BinaryElementwiseParams):
@@ -977,9 +967,6 @@ class MaxParams(BinaryElementwiseParams):
             [self.ifm, self.ifm2, self.ofm], supported_dtypes=[np.uint8, np.int8]
         ):
             return False
-        # MAX with different scales is not supported on NPU.
-        if self.ifm.q_params.scale_f32 != self.ofm.q_params.scale_f32:
-            return False
         return True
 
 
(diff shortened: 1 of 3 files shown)


ekalda (changes requested):
Thanks @Alexey-Yazev for the PR! Looks all good in terms of functionality, my only concern is that this PR adds more than 800 FVP based codegen tests that are notoriously slow and would probably make upstream CI take significantly longer. The RELUs are all quite simple activation functions, so I don't think we get much benefit from testing them with each setting of shape, padding, dilation etc of e.g. conv2d or depthwise2d. I think it would be better to have separate codegen tests for the activation functions that operate on preceding ops with fixed attributes.

tests/python/contrib/test_ethosu/test_codegen.py
```diff
@@ -1109,6 +1109,45 @@ def leaky_relu_func(x):
...
+    np.random.seed(0)
+    accel_type = "ethos-u55-128"
+    ifm_shape = (1, 12, 16, 8)
+
+    @tf.function
+    def relu_n1_to_1(x):
+        """
+        The specific pattern will be replaced into RELU_N1_TO_1 by tflite.
+        """
+        return tf.math.maximum(-1.0, tf.math.minimum(x, 1.0))
```
ekalda, line 1140: Out of interest, what Relay does this get lowered to? I always thought all the different kinds of RELUs were lowered to the same Relay clip with just different min and max and then this clip was pattern matched as a part of all the different op pattern matchers and added to the primitive op's `clip_min` and `clip_max`, but as we don't pattern match for standalone clip and this test here passes, I suppose it gets lowered to something different?

Alexey-Yazev: All relu operations are converted to clip operations in tflite frontend. It seems these tests are unnecessary perhaps it's worth adding a test for maximum operation when there are different scales and it's uploaded to the npu not by one operation as in other cases, but by two.

python/tvm/relay/op/contrib/ethosu.py
```diff
@@ -938,12 +939,21 @@ def minimum_pattern() -> tvm.relay.dataflow_pattern.DFPattern:
...
+def minimum_clip_requantize_pattern() -> tvm.relay.dataflow_pattern.DFPattern:
+    """
+    This function creates the pattern for minimum with fused RELU activation.
+    """
+    pattern = is_op("minimum")(wildcard(), wildcard())
+    pattern = is_op("clip")(pattern)
+    pattern = is_op("qnn.requantize")(
+        pattern, is_constant(), is_constant(), is_constant(), is_constant()
+    )
+    return pattern
```
ekalda, line 954: What's the motivation behind having two patterns instead of one with optional requantize if we use same params extractor (`MinParams`) with them?

ekalda (commented):
Sorry for the delay on this, I took another look and I'm still quite confused about what is going on, so maybe you can elaborate more in the commit body? Two things:
(1) We should still check whether TFLite's RELU6 is correctly executed on NPU, some basic conv2d + RELU6 tests should do
(2) I'm confused about the min/max changes - is it an unrelated bugfix or necessary for RELU_N1_TO_1 implementation?

python/tvm/relay/op/contrib/ethosu.py
Alexey-Yazev: There are two patterns to support two cases:
	a) when we can offload minimum + clip + qnn.requantize to NPU with one operation ethosu_binary_elementwise if there are same scales
	b) when we can offload minimum + clip + qnn.requantize to NPU with two operations ethosu_binary_elementwise + ethosu_identity if there are different scales
	
Since there is a hardware limitation, we cannot perform min or max operation fused with requantize (please look at NPU_SET_OFM_SCALE https://developer.arm.com/documentation/102420/0200/Programmers-model/Command-stream/cmd1-commands-) when we have different scales.

Alexey-Yazev:
> (1) We should still check whether TFLite's RELU6 is correctly executed on NPU, some basic conv2d + RELU6 tests should do

in my opinion it will be enough to add separate tests for RELU6 and RELU_N1_TO_1 since RELU, RELU6, RELU_N1_TO_1 
activations are converted to clip operation in the tflite frontend https://github.com/apache/tvm/blob/main/python/tvm/relay/frontend/tflite.py#L534 and other cases with basic operations will be covered by cases with relu activation or is it necessary to add tests for RELU6, RELU_N1_TO_1 with other operations?
> (2) I'm confused about the min/max changes - is it an unrelated bugfix or necessary for RELU_N1_TO_1 implementation?

this is adding restrictions according to the hardware

ekalda (changes requested):

Alexey-Yazev:
> so I suspect that if the graph consists only RELU and nothing else, it doesn't get offloaded to NPU at all.

in this case requantize operation can be offloaded to NPU

>  I think so far the approach has been to match all the variations of the operator and then output the right combination of Ethos-U ops, e.g. Resize2d. If there is a reason it can't be done that way, it would be good to document it there since it is not immediately obvious from the code why we need two patterns.

reason that two patterns are used is that we can understand whether this pattern is suitable after checking scales match that are performed in function is_valid

Alexey-Yazev:
> To me this looks like fixing what was an oversight in the initial implementation of binary elementwise and is not related to adding support to two new activation functions (which is what the title of this PR claims) and in general should be a separate PR, but I don't think it worth the trouble to split up this PR since it is quite small.

I left in this PR only adding tests for RELU6 and RELU_N1_TO_1 and I'm going to create another PR with changes to binary elementwise implementation

tests/python/contrib/test_ethosu/test_codegen.py
```diff
@@ -1109,6 +1109,77 @@ def leaky_relu_func(x):
...
+    padding = (1, 0, 1, 1)
+
+    @tf.function
+    def conv2d_relu_n1_to_1(x):
+        tf_strides = [1, strides[0], strides[1], 1]
+        op = tf.pad(
+            x,
+            [[0, 0], [padding[0], padding[2]], [padding[1], padding[3]], [0, 0]],
+            "CONSTANT",
+        )
```
ekalda, line 1128: Why the preceding pad?

ekalda (commented):
Thanks @Alexey-Yazev, I think patch is nearing to be ready, got one question about the operation.

tests/python/contrib/test_ethosu/test_codegen.py
Alexey-Yazev: There is no specific reason, it can be removed.

ekalda (approved):
Thanks @Alexey-Yazev, LGTM!

Issue #13647, opened 2022-12-20
[CI Problem] Android build failing with numpy error


lhutton1:
This is failing due to the recently deprecated numpy value `np.bool` in the `1.9.1` release of mxnet. From the 1.24.0 release notes (https://numpy.org/doc/stable/release/1.24.0-notes.html#expired-deprecations):
```
The deprecation for the aliases np.object, np.bool, np.float, np.complex, np.str, and np.int is expired 
(introduces NumPy 1.20). Some of these will now give a FutureWarning in addition to raising an error 
since they will be mapped to the NumPy scalars in the future.
```

ashutosh-arm:
Affected PRs:
https://github.com/apache/tvm/pull/13637
https://github.com/apache/tvm/pull/13644
https://github.com/apache/tvm/pull/13643

lhutton1:
#13648 was merged as a temporary workaround, there likely needs to be a fix in MXNet to fix the compatibility issue. We can revert #13648 when it is resolved.

lhutton1:
Closing now that mxnet is archived and being removed as a dependency from TVM.

PR #13648, opened 2022-12-20
[CI] Fix android build by constraining numpy version

lhutton1:
Temporarily constrain the version of numpy to workaround the deprecated value used in mxnet. See #13647.


ashutosh-arm (approved):
LGTM @lhutton1. Thanks! It will unblock a lot of PRs.

PR #13649, opened 2022-12-21
[CLML][RELAY] Enable Pad and Conv2d layer fusion

krishnaraj36:
Enabled clml supported nn.pad+nn.conv2d fusion pattern in clml pattern table

### DIFF
--- a/python/tvm/relay/op/contrib/clml.py
+++ b/python/tvm/relay/op/contrib/clml.py
@@ -22,7 +22,7 @@ from tvm import relay
 from tvm._ffi import register_func
 from tvm.relay import transform
 from tvm.relay.build_module import bind_params_by_name
-from tvm.relay.expr import Call, Var, Constant
+from tvm.relay.expr import Var, Constant
 
 from ...dataflow_pattern import wildcard, is_op, is_constant, is_tuple_get_item, is_tuple
 from .register import register_pattern_table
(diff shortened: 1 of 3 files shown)


python/tvm/relay/op/contrib/clml.py
```diff
@@ -200,9 +218,18 @@ def check_conv(extract):
...
             call = call.args[0]
+
         attrs, args = call.attrs, call.args
         if attrs.data_layout != "NCHW":
             return False
+
+        if (
+            (not isinstance(args[0], (Var, Constant)))
+            and (args[0].op.name == "nn.pad")
+            and (len(args[0].attrs["pad_width"]) != 4)
```
echuraev, line 229: Why there is such limitation on padding width?

srkreddy1238: Ideally, this check is not required. nn.pad implementation ensures pad_width length to be in sync with incoming tensor rank.

srkreddy1238 (changes requested):

python/tvm/relay/op/contrib/clml.py
krishnaraj36: CLML module is expected to have input to 4D tensor shape for Pad layer, since conv2d layer input is always 4D tensor so Pad+conv2d fusion input is expected have 4D tensor always. It seems condition is redundant. Removed in updated commit.
@srkreddy1238 @echuraev

echuraev (approved):
LGTM. Thanks

srkreddy1238 (approved):
LGTM

PR #13650, opened 2022-12-21
[ETHOSN] Fix for the mock inference after NPU driver update

ashutosh-arm:
NPU driver 22.11 supports buffer and network creation
with only process memory allocator. This change however
should be limited to the cases when the HW is available.

NPU driver update: https://github.com/apache/tvm/pull/13637

cc @lhutton1


lhutton1 (approved):
Thanks @ashutosh-arm, LGTM!

PR #13652, opened 2022-12-21
[CMSIS-NN] Global function that provides range based on dtype

ashutosh-arm:
Follow up from https://github.com/apache/tvm/pull/13522#discussion_r1040157137.
Range for dtype was being sought differently from inside various CMSIS-NN tests. This commit provides
a common global function inside aot.py under tvm.testing that can provide (min, max) values based on the dtype.
In future, other AOT based targets can make use of this function to obtain the range.

cc @Mousius @lhutton1 @NicolaLancellotti

### DIFF
--- a/python/tvm/testing/aot.py
+++ b/python/tvm/testing/aot.py
@@ -24,7 +24,7 @@ import shutil
 import subprocess
 import tarfile
 import logging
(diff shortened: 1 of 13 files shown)


NicolaLancellotti (approved):
LGTM

Mousius (approved):

PR #12789, opened 2022-09-15
Added macro generation in MLF export

fPecc:
For description see this [forum topic](https://discuss.tvm.apache.org/t/exporting-generated-functions-input-and-output-sizes-in-the-model-library-format/13532).

Tagged reviewers: @areusch (don't know who else can review microTVM-related code)

PR #13655, opened 2022-12-23
[AOT] Added a test for detecting output size post MLF export

ashutosh-arm:
Follow up from: https://github.com/apache/tvm/pull/12789

-Fixed IO size map preparation in MLF export
-Added a separate test to detect output size from MLF codegen
-MLF size is not used in aot.py because its unavailable in case of packed apis

cc @Mousius

### DIFF
--- a/python/tvm/micro/model_library_format.py
+++ b/python/tvm/micro/model_library_format.py
@@ -485,6 +485,12 @@ def _export_graph_model_library_format(
                     "functions"
                 ]["main"][0]["outputs"][key]
 
+            input_name_to_size_map = {
+                name: property_map["size"] for name, property_map in inputs_sizes.items()
+            }
+            output_name_to_size_map = {
+                name: property_map["size"] for name, property_map in output_sizes.items()
+            }
             generate_c_interface_header(
                 mod.libmod_name,
                 inputs,
@@ -494,8 +500,8 @@ def _export_graph_model_library_format(
                 devices,
                 workspace_size,
                 include_path,
-                inputs_sizes,
-                output_sizes,
+                input_name_to_size_map,
+                output_name_to_size_map,
             )
 
         is_aot = isinstance(mod, executor_factory.AOTExecutorFactoryModule)
--- a/tests/python/relay/aot/test_crt_aot.py
+++ b/tests/python/relay/aot/test_crt_aot.py
@@ -225,6 +225,64 @@ def test_packed_global_variables():
             assert f"{func}_packed" not in tvmgen_names
 
 
+def test_io_size_definition():
+    """Check network IO size definitions in the codegen output."""
+    dtype = "float32"
+    ishape = (1, 32, 14, 14)
+    wshape = (32, 32, 3, 3)
+    interface_api = "c"
+    use_unpacked_api = True
+
+    data0 = relay.var("data", shape=ishape, dtype=dtype)
+    weight0 = relay.var("weight", shape=wshape, dtype=dtype)
+    out = relay.nn.conv2d(data0, weight0, kernel_size=(3, 3), padding=(1, 1), groups=1)
+    main_f = relay.Function([data0, weight0], out)
+    mod = tvm.IRModule()
+    mod["main"] = main_f
+    mod = transform.InferType()(mod)
+
+    i_data = np.random.uniform(0, 1, ishape).astype(dtype)
+    w1_data = np.random.uniform(0, 1, wshape).astype(dtype)
+
+    inputs = OrderedDict([("data", i_data), ("weight", w1_data)])
+
+    output_list = generate_ref_data(mod, inputs)
+    compiled_models_list = compile_models(
+        models=AOTTestModel(module=mod, inputs=inputs, outputs=output_list),
+        interface_api=interface_api,
+        use_unpacked_api=use_unpacked_api,
+        workspace_byte_alignment=8,
+        enable_op_fusion=True,
+        pass_config=AOT_DEFAULT_RUNNER.pass_config,
+        use_runtime_executor=True,
+        target=tvm.target.Target("c"),
+    )
+    ref_output_size = output_list["output"].size * np.dtype(dtype).itemsize
+    compiled_model = compiled_models_list[0]
+
+    tmp_path = utils.tempdir()
+    base_path = tmp_path.temp_dir
+
+    model = compiled_model.model
+    tar_file = os.path.join(base_path, f"{model.name}.tar")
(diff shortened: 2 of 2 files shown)


python/tvm/micro/model_library_format.py
```diff
@@ -485,6 +485,12 @@ def _export_graph_model_library_format(
                     "functions"
                 ]["main"][0]["outputs"][key]
 
+            input_name_to_size_map = {}
+            output_name_to_size_map = {}
+            for name, property_map in inputs_sizes.items():
+                input_name_to_size_map.update({name: property_map["size"]})
+            for name, property_map in output_sizes.items():
+                output_name_to_size_map.update({name: property_map["size"]})
```
Mousius, line 493:
```suggestion
            input_name_to_size_map = {
               name: property_map["size"]
               for name, property_map in inputs_sizes.items()
            }
            output_name_to_size_map = {
              name: property_map["size"]
              for name, property_map in output_sizes.items()
            }
```

python/tvm/micro/model_library_format.py
```diff
@@ -494,8 +500,10 @@ def _export_graph_model_library_format(
                 devices,
                 workspace_size,
                 include_path,
-                inputs_sizes,
-                output_sizes,
+                # inputs_sizes,
+                # output_sizes,
```
Mousius, line 504:
```suggestion
```

(Just delete these, no need to leave them here)

python/tvm/testing/aot.py
```diff
@@ -415,24 +415,23 @@ def _emit_main_compare(main_file, outputs, output_tolerance, mod_name, use_inter
             comparison_function = "fabs"
             tolerance = output_tolerance or 0.001
 
-        data_length_var_name = (
-            _mangle_name(mod_name, f"output_data_{sanitized_tensor_name}") + "_len"
+        actual_data_name = _mangle_name(mod_name, f"output_data_{sanitized_tensor_name}")
+        data_len_var = actual_data_name + "_len"
+        main_file.write(
+            f"const size_t {data_len_var}"
+            f"= sizeof({actual_data_name})/sizeof({actual_data_name}[0]);\n"
```
Mousius, line 422: We generate `actual_data_name` and `data_length_var_name` in the inputs/outputs already, why are we recalculating it here based on types we've defined within the AOT test harness?

We probably need to add a `_LEN` macro to the MLF header?

Mousius (changes requested):

python/tvm/testing/aot.py
Mousius: `_LENGTH` to be clearer 😸

python/tvm/micro/model_library_format.py
ashutosh-arm: ACK

python/tvm/micro/model_library_format.py
ashutosh-arm: ACK

python/tvm/testing/aot.py
ashutosh-arm: As discussed offline, the two variables `LEN` and `SIZE` serve different purposes. `SIZE` will come from MLF export whereas `LEN` will be part of the AOT test harness.

ashutosh-arm:
@Mousius could you please take a look again after the LEN was included from AOT test harness?

tests/python/relay/aot/test_crt_aot.py
```diff
@@ -225,6 +225,64 @@ def test_packed_global_variables():
...
+    model = compiled_model.model
+    tar_file = os.path.join(base_path, f"{model.name}.tar")
+    export_model_library_format(compiled_model.executor_factory, tar_file)
+    t = tarfile.open(tar_file)
+    t.extractall(base_path)
+
+    file_list = []
+    for path in (pathlib.Path(base_path) / "codegen" / "host" / "include").iterdir():
+        if path.is_file():
+            file_list.append(path)
```
Mousius, line 275: Given we know the `model_name` can we not just look for `tvmgen_{model_name}.h` rather than looping?

tests/python/relay/aot/test_crt_aot.py
```diff
@@ -225,6 +225,64 @@ def test_packed_global_variables():
...
+    assert len(file_list) > 0
+
+    for path in file_list:
+        with open(path, "r") as header:
+            contents = header.readlines()
+            contents = "".join(map(str, contents))
+            assert contents.count("_SIZE") == 4
+            assert str(ref_output_size) in contents
```
Mousius, line 283: We should probably check the `_SIZE` values match with the appropriate constants rather than them just appearing in the same file together?

tests/python/relay/aot/test_crt_aot.py
Mousius: Can we also do this for the input sizes?

Mousius (changes requested):

tests/python/relay/aot/test_crt_aot.py
ashutosh-arm: I tried doing that initially. Any short cuts to do that?

tests/python/relay/aot/test_crt_aot.py
Mousius: I assume this just requires looking for both headers?

tests/python/relay/aot/test_crt_aot.py
Mousius: Something like:
```
assert contents.count("_SIZE") == 4
assert f"INPUT_1_SIZE {ref_output_size}" in contents
```
?

ashutosh-arm: Ah ok. I misunderstood what you were asking for. This makes sense. Thanks for the help.

tests/python/relay/aot/test_crt_aot.py
ashutosh-arm: Oh, there is just one header in those cases too. Both models' sizes appear in a single header. So, need not be tested additionally.

Mousius: That sounds wrong, shouldn't there be a `tvmgen_model1.h` and a `tvmgen_model2.h` ?

ashutosh-arm: My bad. I confused this with the multi-model test which it is not. In case of multi model test, I do see two separate headers being produced.

Mousius (approved):

Mousius:
LGTM! Thanks @ashutosh-arm !

PR #13656, opened 2022-12-23
[CMSIS-NN] Support CMSIS NN from new GitHub location

NicolaLancellotti:
CMSIS NN has been moved out of the CMSIS project into a new GitHub project.
This pr adds support for CMSIS NN from this new GitHub location.
Both [CMSIS](https://github.com/ARM-software/CMSIS_5) and [CMSIS NN](https://github.com/ARM-software/CMSIS-NN) are now downloaded in `/opt/arm/ethosu/cmsis`


ashutosh-arm:
Thanks @NicolaLancellotti for this changes. Looks good to me overall. A small request though - is it possible to checkout the repository with the name `CMSIS-NN` instead of `NN` under `/opt/arm/ethosu/cmsis`?

NicolaLancellotti:
> Thanks @NicolaLancellotti for this changes. Looks good to me overall. A small request though - is it possible to checkout the repository with the name `CMSIS-NN` instead of `NN` under `/opt/arm/ethosu/cmsis`?

Done.

mehrdadh (changes requested):
@NicolaLancellotti thanks for this PR!
Could you please add this change to docker script installation?
https://github.com/apache/tvm/blob/main/docker/install/ubuntu_install_cmsis.sh

NicolaLancellotti:
I updated the pr with three commits:
- First, I supported CMSIS NN from the new GitHub location, as previously done.
- Second, I prevented the definition of the `cmsis_path` project API option from an environment variable. Before, thanks to the environment variable, the `cmsis_path` option was always enabled. That was not a problem, but now CMSIS NN uses a new header (`arm_acle.h`) which is not always present, so we need to explicitly enable CMSIS when we need it.
- In the end, I re-added support for the old location of CMSIS NN because the docker image is not yet updated, and we need the tests to pass to accept this pr. In this way, tvm will use the new CMSIS NN project when we will update the docker image, but for now, it uses the old one. I'll create a pr that reverts this last commit when the docker image is updated.

tests/micro/zephyr/test_zephyr.py
```diff
@@ -623,7 +624,7 @@ def test_schedule_build_with_cmsis_dependency(workspace_dir, board, microtvm_deb
     assert "CMSIS/DSP/Include" in cmake_content
     assert "CMSIS/DSP/Include/dsp" in cmake_content
     assert "CMSIS/DSP/Include" in cmake_content
-    assert "CMSIS/NN/Include" in cmake_content
+    # assert "CMSIS-NN/Include" in cmake_content
```
ashutosh-arm, line 627: If comment is not useful for future, it can be removed?

tests/python/relay/aot/corstone300.mk
```diff
@@ -48,6 +48,11 @@ CC = arm-none-eabi-gcc
 AR = arm-none-eabi-ar
 RANLIB = arm-none-eabi-ranlib
 CC_OPTS = CC=$(CC) AR=$(AR) RANLIB=$(RANLIB)
+ifeq ($(shell [ -d ${CMSIS_PATH}/CMSIS-NN ]; echo $$?), 0)
+	CMSIS_NN_PATH = ${CMSIS_PATH}/CMSIS-NN
+else
+	CMSIS_NN_PATH = ${CMSIS_PATH}/CMSIS/NN
+endif
```
ashutosh-arm, line 55: doesn't simply looking for the directory work `if [ -d " ${CMSIS_PATH}/CMSIS-NN" ]; ` for simplicity - just curious?

tests/micro/zephyr/test_zephyr.py
NicolaLancellotti: In the future, after the docker image will be updated, we want this assert.

docker/install/ubuntu_install_ethosu_driver_stack.sh
```diff
@@ -85,6 +85,9 @@ cmake -DCMAKE_TOOLCHAIN_FILE=${ethosu_dir}/core_platform/cmake/toolchain/arm-non
 make
 
 # Build NN Library
+mkdir ${CMSIS_PATH}/CMSIS-NN/build/ && cd ${CMSIS_PATH}/CMSIS-NN/build/
+cmake .. -DCMAKE_TOOLCHAIN_FILE=${ethosu_dir}/core_platform/cmake/toolchain/arm-none-eabi-gcc.cmake -DTARGET_CPU=cortex-m55 -DBUILD_CMSIS_NN_FUNCTIONS=YES -DCMSIS_PATH=${CMSIS_PATH}
+
 mkdir ${CMSIS_PATH}/CMSIS/NN/build/ && cd ${CMSIS_PATH}/CMSIS/NN/build/
 cmake .. -DCMAKE_TOOLCHAIN_FILE=${ethosu_dir}/core_platform/cmake/toolchain/arm-none-eabi-gcc.cmake -DTARGET_CPU=cortex-m55 -DBUILD_CMSIS_NN_FUNCTIONS=YES
 make
```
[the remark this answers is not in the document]
NicolaLancellotti: The first one builds the new `CMSIS-NN`, and the second one builds the old `CMSIS/NN`.
When the docker image will be updated I'll remove support for `CMSIS/NN`, that is, I''ll revert this commit: https://github.com/apache/tvm/pull/13656/commits/ff57097856d7e100fbace341c3ac528ffe52acfa.

tests/python/relay/aot/corstone300.mk
NicolaLancellotti: If you want I can check and improve that, but this code will be reverted when the docker image will be updated.

apps/microtvm/zephyr/template_project/CMakeLists.txt.template
```diff
@@ -31,21 +31,27 @@ find_package(Zephyr HINTS $ENV{ZEPHYR_BASE})
 project(microtvm_autogenerated_project)
 
 if(DEFINED CMSIS_PATH)
+  if (EXISTS ${CMSIS_PATH}/CMSIS-NN)
```
mehrdadh, line 34: Is this a temporary fix until we update the docker image?

mehrdadh:
It will be hard to keep track of all the follow up changes that are required after we update the image. Could you please add comment in all of those places so we can track them?

apps/microtvm/zephyr/template_project/CMakeLists.txt.template
NicolaLancellotti: Yes it is.

NicolaLancellotti:
> It will be hard to keep track of all the follow up changes that are required after we update the image. Could you please add comment in all of those places so we can track them?

Hi @mehrdadh, thanks for the review.
I wrote the commits in this pr in such a way that it is easy to track the required changes, we just have to revert this commit https://github.com/apache/tvm/pull/13656/commits/ff57097856d7e100fbace341c3ac528ffe52acfa when the image will be updated.

ashutosh-arm:
After discussing offline about the remaining issues with @NicolaLancellotti, it looks like the support for the old CMSIS-NN location will be removed in future. Overall the changes LGTM! Thanks @NicolaLancellotti for this critical update.

mehrdadh (approved):
LGTM, thanks!

PR #13659, opened 2022-12-26
[MetaSchedule] Add "disabled_pass" option in tuning API

ibsidorenko:
Now there is no way to disable passes in MetaShedule tuner. This commit adds new parameter `disabled_pass` in tuning API (`tune_relay`/`compile_relay`). It can be used for different experiments and non default behavoir.

### DIFF
--- a/python/tvm/meta_schedule/relay_integration.py
+++ b/python/tvm/meta_schedule/relay_integration.py
@@ -17,7 +17,7 @@
 """MetaSchedule-Relay integration"""
 from contextlib import contextmanager
 from types import MappingProxyType
-from typing import TYPE_CHECKING, Any, Dict, List, Mapping, Optional, Tuple, Union
+from typing import TYPE_CHECKING, Any, Dict, List, Mapping, Optional, Tuple, Union, Set
 
 # isort: off
 from typing_extensions import Literal
@@ -120,6 +120,7 @@ def extract_tasks(
     ),
     executor: Optional["relay.backend.Executor"] = None,
     module_equality: str = "structural",
+    disabled_pass: Optional[Union[List[str], Set[str], Tuple[str]]] = None,
 ) -> List[ExtractedTask]:
     """Extract tuning tasks from a relay program.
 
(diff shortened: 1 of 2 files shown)


ibsidorenko:
cc @masahi @junrushao

masahi (approved):

python/tvm/meta_schedule/relay_integration.py
```diff
@@ -250,6 +252,7 @@ def tune_relay(
     seed: Optional[int] = None,
     module_equality: str = "structural",
     num_tuning_cores: Union[Literal["physical", "logical"], int] = "physical",
+    disabled_pass: Optional[Union[List[str], Set[str], Tuple[str]]] = None,
```
echuraev, line 255: Could you please add information about this parameter to the docstring?

python/tvm/meta_schedule/relay_integration.py
```diff
@@ -345,6 +350,7 @@ def compile_relay(
         }
     ),
     executor: Optional["relay.backend.Executor"] = None,
+    disabled_pass: Optional[Union[List[str], Set[str], Tuple[str]]] = None,
```
echuraev, line 353: Add information to doc

echuraev (commented):
Also, I have a question. Should we add a test for this option or not?

ibsidorenko:
> Also, I have a question. Should we add a test for this option or not?

I thought about that... since this is rather trivial change, not a bug fixing + it can be checked with tuning only (time consuming operation) I decided not to add unit test.

python/tvm/meta_schedule/relay_integration.py
ibsidorenko: Done

python/tvm/meta_schedule/relay_integration.py
ibsidorenko: Done

echuraev (approved):
LGTM. Thanks

PR #13660, opened 2022-12-26
[QNN] Change in Pass Context for lookup table calculation

ibsidorenko:
**Motivation:**
It is possible to disable specific passes through the "disabled_pass" parameter in the Pass Context. These "disabled" passes can be optional for one target and mandatory for another one.
Since lookup table for some QNN operations (tanh, round and etc.) is calculated on the host and some of disabled passes can be required for the host, no need to disable these passes. This constant calculation/ evaluation is orthogonal to the compilation process for specific target.

**What was changed:**
This commit creates its own compilation Pass Context for lookup table
calculation and evaluation (for elemwise QNN ops: tanh, sqrt ...).

### DIFF
--- a/python/tvm/relay/qnn/op/canonicalizations.py
+++ b/python/tvm/relay/qnn/op/canonicalizations.py
@@ -23,10 +23,25 @@ from tvm import relay
 
 
 def run_const_expr(expr: "relay.Expr") -> np.ndarray:
-    """Evaluate a const expression, receiving result as np array."""
-    mod = tvm.IRModule.from_expr(expr)
-    vm_exe = relay.create_executor("vm", mod=mod)
-    return vm_exe.evaluate()().asnumpy()
+    """Evaluate a const expression, receiving result as np array.
+
+    If a number of passes are disabled in the current Pass Context, then there is no need to disable
+    these passes for const expression evaluation as well. That's why we use empty list
+    "disabled_pass=[]", all other arguments are inherited from the current Pass Context.
+    """
+    curr_pass_ctx = tvm.ir.transform.PassContext.current()
+    with tvm.ir.transform.PassContext(
+        opt_level=curr_pass_ctx.opt_level,
+        required_pass=curr_pass_ctx.required_pass,
+        disabled_pass=[],
+        instruments=curr_pass_ctx.instruments,
+        config=curr_pass_ctx.config,
+    ):
+        mod = tvm.IRModule.from_expr(expr)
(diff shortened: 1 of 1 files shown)


ibsidorenko:
@tvm-bot rerun

ibsidorenko:
cc @masahi @AndrewZhaoLuo

python/tvm/relay/qnn/op/canonicalizations.py
```diff
@@ -23,10 +23,25 @@
...
 def run_const_expr(expr: "relay.Expr") -> np.ndarray:
-    """Evaluate a const expression, receiving result as np array."""
-    mod = tvm.IRModule.from_expr(expr)
-    vm_exe = relay.create_executor("vm", mod=mod)
-    return vm_exe.evaluate()().asnumpy()
+    """Evaluate a const expression, receiving result as np array.
+
+    If a number of passes are disabled in the current Pass Context, then there is no need to disable
+    these passes for const expression evaluation as well. That's why we use empty list
+    "disabled_pass=[]", all other arguments are inherited from the current Pass Context.
```
masahi, line 30: Do you mean, without `disabled_pass=[]`, the original `disabled_pass` list in the current context will be ignored during const folding?

ibsidorenko: I put here `disabled_pass=[]` to ignore the original `disabled_pass` list from the current context. Otherwise VM inherits non-empty `disabled_pass` list and fails.
Note, we call this code only for several qnn ops: tanh, sqrt, erf, exp + some other ops. It does not affect global constant folding.

For clarification, here is my example to illustrate this issue:
I am trying to compile subgraph with `qnn.tanh` operation for Hexagon target **without** QNN canonicalization but **with** QNN legalization (this is acceptable for Hexagon). For `qnn.tanh` we compose lookup table by means of Virtual Machine and compute it on the host. For most of users/developers host is x86 cpu. But for x86 QNN canonicalization is mandatory pass, otherwise it fails.
Looks like the simplest way to fix this issue is to create new Pass Context and ignore the original `disabled_pass` list for lookup table calculation and constant evaluation. Other arguments (opt_level, instruments, config ...) I inherit from the current context.

P.S. Corresponding unit test will be added in the next PR.

masahi: I see, you do want to enable QNN legalization (which is disabled in the original ctx) here.

masahi (approved):

PR #13670, opened 2022-12-28
[CLML] Version compatibility and various test cases 

srkreddy1238:
CLML version compatibility changes to support future versions.

Codegen verification test cases for all the ops (convolution, concat, pad, pool ..etc.) that are supported by clml BYOC path.

Fix depthwise conv2d issue with layout

Adreno container is now free from CLML SDK dependency. We can use it for texture only builds now.


src/relay/backend/contrib/clml/codegen.cc
```diff
@@ -332,6 +332,10 @@ class CLMLJSONSerializer : public backend::contrib::JSONSerializer {
       bias = dense;
       dense = dense->args[0].as<CallNode>();
     }
+    if (backend::IsOp(dense, "nn.bias_add")) {
```
echuraev, line 335: Maybe unify `add` and `nn.bias_add`?

src/runtime/contrib/clml/clml_runtime.cc
```diff
@@ -153,13 +153,24 @@ class CLMLRuntime : public JSONRuntimeBase {
...
+      if (majorVersions[i] == 3) {
+        h_ClmlIntf = clGetMLInterfaceV3QCOM(0);
+        LOG(WARNING) << "CLML Target version:" << majorVersions[i];
+        break;
+      }
+#endif
     }
+    ICHECK(h_ClmlIntf != NULL)
+        << "clGetMLInterfaceVxQCOM:" << result
+        << " Perhaps there is mispatch between CLML SDK version to target supported version";
```
echuraev, line 173: Let's print target supported version

tests/python/contrib/test_clml/infrastructure.py
```diff
@@ -155,9 +173,9 @@ def build_and_run(
     for _ in range(no_runs):
         gen_module.run()
         out.append([gen_module.get_output(i) for i in range(outputs)])
-    time_f = gen_module.module.time_evaluator("run", device.device.cl(0), number=1)
-    cost = time_f().mean
-    print("%g secs/iteration\n" % cost)
+    # time_f = gen_module.module.time_evaluator("run", device.device.cl(0), number=1)
+    # cost = time_f().mean
+    # print("%g secs/iteration\n" % cost)
```
echuraev, line 178: Why did you remove it?

tests/scripts/task_build_adreno_bins.sh
```diff
@@ -29,7 +29,7 @@ cd ${output_directory}
 cp ../cmake/config.cmake .
 
 echo set\(USE_MICRO OFF\) >> config.cmake
-echo set\(USE_CLML ON\) >> config.cmake
+echo set\(USE_CLML "${ADRENO_OPENCL}"\) >> config.cmake
```
echuraev, line 32: Is it not necessary to do this check (`[[ -z "${ADRENO_OPENCL}" ]] && CLML_PATH='ON' || CLML_PATH="${ADRENO_OPENCL}"`) here?

tests/python/contrib/test_clml/infrastructure.py
```diff
@@ -181,16 +199,34 @@ def extract_clml_modules(module):
...
 ):
     """Check clml codegen against a known good output."""
-    module = build_module(module, target, tvm_ops=tvm_ops, clml_partitions=num_clml_modules)
-    clml_modules = extract_clml_modules(module)
+    if isinstance(mod, tvm.relay.expr.Call):
+        mod = tvm.IRModule.from_expr(mod)
+    with tvm.transform.PassContext(opt_level=3, disabled_pass=["AlterOpLayout"]):
+        mod = clml.partition_for_clml(mod, params)
+        tvm_op_count = get_cpu_op_count(mod)
+        assert tvm_op_count == tvm_ops, "Got {} TVM operators, expected {}".format(
```
echuraev, line 215: What if the user didn't pass value of `tvm_ops`? Probably `tvm_ops` should be required parameter?

tests/python/contrib/test_clml/test_ops.py
```diff
@@ -54,11 +62,8 @@ def _get_conv_model(
         shape = (shape[0], shape[1], shape[2] + padding[0] * 2, shape[3] + padding[1] * 2)
     is_depthwise = shape[1] == channels == groups
 
-    weight_format = "OIHW" if is_depthwise else "OIHW"
-    if weight_format == "IOHW":
-        weight_shape = (shape[1] // groups, channels, kernel_h, kernel_w)
-    else:
-        weight_shape = (channels, shape[1] // groups, kernel_h, kernel_w)
+    weight_format = "OIHW"
```
echuraev, line 65: Why did you remove check on the dephtwise convolution?

tests/python/contrib/test_clml/test_ops.py
```diff
@@ -86,31 +91,122 @@ def _get_conv_model(
...
+    output_shape = (1, channels, int(output_height), int(output_width))
+    out_dtype = dtype
+    is_depthwise = shape[1] == channels == groups
+
+    weight_format = "IOHW" if is_depthwise else "OIHW"
+    if weight_format == "OIHW":
+        weight_shape = (channels, shape[1] // groups, kernel_h, kernel_w)
+    else:
+        weight_shape = (shape[1] // groups, channels, kernel_h, kernel_w)
+    # weight_shape = (channels, shape[1] // groups, kernel_h, kernel_w)
```
echuraev, line 123: Redundant comment

echuraev (commented):
Thanks for the PR. Several comments

cmake/modules/contrib/CLML.cmake
```diff
@@ -22,7 +22,21 @@ if(USE_CLML)
...
         list(APPEND COMPILER_SRCS ${CLML_RUNTIME_MODULE})
     endif()
-    message(STATUS "Build with CLML support...")
+    message(STATUS "Build with CLML support : " ${USE_CLML})
+    if (NOT USE_CLML STREQUAL "ON")
+        set(CLML_VERSION_HEADER "${USE_CLML}/CL/cl_qcom_ml_ops.h")
+        if(EXISTS ${CLML_VERSION_HEADER})
+            file(READ ${CLML_VERSION_HEADER} ver)
+            string(REGEX MATCH "CL_QCOM_ML_OPS_H_MAJOR_VERSION ([0-9]*)" _ ${ver})
+            set(CLML_VERSION_MAJOR ${CMAKE_MATCH_1})
```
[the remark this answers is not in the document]
srkreddy1238: For codegen, OpenCL1.2 restriction is good as we don't have any requirement of using features from higher versions.

CLML has newer versions with new operators support and is essential to support to get perf improvements on new gerations.

tests/python/contrib/test_clml/infrastructure.py
srkreddy1238: This affects CI time and also to get ideal numbers we need to set ```number``` to a higher value too. 
For now, I am leaving it for developers to uncomment and use.

tests/scripts/task_build_adreno_bins.sh
srkreddy1238: We don't need this check in other script too as with in CI its always available as ```/adreno-opencl```.

tests/python/contrib/test_clml/infrastructure.py
srkreddy1238: Is set to default ```0``` if not passed.

tests/python/contrib/test_clml/test_ops.py
srkreddy1238: This was the issue we had earlier with depthwise.

Weight layout transform for depthwise is handled internally by CLML compilation process internally. https://github.com/apache/tvm/blob/6f6bf1912e08f5113c1ac438ea600270b479055d/python/tvm/relay/op/contrib/clml.py#L79

masahi (approved):

PR #13675, opened 2022-12-30
[BENCHMARK][ADRENO] Adreno Benchmarks with texture

srkreddy1238:
Benchmarks for various networks listed below with fp16 and fp32. resnet-18, resnet-34, resnet-50, vgg-16, vgg-19, densenet-121, inception_v3, mobilenetv1, squeezenet_v1.0, squeezenet_v1.1

### DIFF
--- a/apps/benchmark/adreno/adreno_gpu_bench_texture.py
+++ b/apps/benchmark/adreno/adreno_gpu_bench_texture.py
@@ -1,0 +1,277 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+"""Benchmark script for various models on Adreno GPU.
+"""
+import argparse
+
+import numpy as np
+
+import os
+import sys
+import tvm
+from tvm import te
+from tvm.relay import testing
+from tvm.contrib.utils import tempdir
+import tvm.contrib.graph_executor as runtime
+from tvm import relay
+from tvm import autotvm
+from tvm.contrib import utils, ndk
+
+
+def get_network(name, batch_size, dtype="float32"):
+    """Get the symbol definition and random weight of a network
+
+    Parameters
+    ----------
+    name: str
+        The name of the network, can be 'resnet-18', 'resnet-50', 'vgg-16', 'inception_v3', 'mobilenet', ...
+    batch_size: int
+        batch size
+    dtype: str
+        Data type
+
+    Returns
+    -------
+    net: tvm.IRModule
+        The relay function of network definition
+    params: dict
+        The random parameters for benchmark
+    input_shape: tuple
+        The shape of input tensor
+    output_shape: tuple
+        The shape of output tensor
+    """
+    input_shape = (batch_size, 3, 224, 224)
+    output_shape = (batch_size, 1000)
+
+    if name == "mobilenet":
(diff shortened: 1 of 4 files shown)


apps/benchmark/adreno/bench.py
```diff
@@ -0,0 +1,61 @@
...
+DEVICE_PID=$!
+sleep 5 # Wait for the device connections
+trap "{ kill ${TRACKER_PID}; kill ${DEVICE_PID}; }" 0
+
+# cleanup pycache
+find . -type f -path "*.pyc" | xargs rm -f
+# Test TVM
+make cython3
+
+if [ "texture" == $1 ] ; then
```
echuraev, line 56: Are there any reasons to run this script without `texture` parameter? If no, probably we can remove this argument.

tests/scripts/ci.py
```diff
@@ -727,6 +727,13 @@ def add_subparser(
                     "./tests/scripts/task_python_adreno.sh " + os.environ.get("ANDROID_SERIAL", ""),
                 ],
             ),
+            "benchmarks": (
```
echuraev, line 730: Are there any plans to run these benchmarks in the CI? @driazati probably you know, do we have any opportunities to run different performance tests e.g. once per week to check that no performance regressions were introduced to TVM mainline.

apps/benchmark/adreno/adreno_gpu_bench_texture.py
```diff
@@ -0,0 +1,277 @@
...
+    Parameters
+    ----------
+    msg: str
+        The message to print
+    """
+    sys.stdout.write(msg + "\r")
+    sys.stdout.flush()
+
+
+def tune_tasks(
```
echuraev, line 115: Just as an idea. If we want to use such benchmarks in the CI we can use statistic from tophub to measure performance w/o tuning. If someone changes schedules for Adreno then he/she will upload updated statistic to the tophub. In this way, we can avoid performance regressions.

apps/benchmark/adreno/adreno_gpu_bench_texture.py
```diff
@@ -0,0 +1,277 @@
...
+
+    # Build the tuning log
+    if os.path.exists(tune_log):
+        with autotvm.apply_history_best(tune_log):
+            with tvm.transform.PassContext(opt_level=3):
+                lib = relay.build(
+                    net, target=tvm.target.Target(target, host=target_host), params=params
+                )
+    else:
+        print("WARNING: Benchmark running with out tuning cache file - ", tune_log)
```
echuraev, line 181:
```suggestion
        print("WARNING: Benchmark running without tuning cache file - ", tune_log)
```

echuraev (commented):
Thanks for the PR! Several comments to discuss

tests/scripts/ci.py
tqchen: Ideally this should be separate and as part of nightly instead of every push

echuraev: Yes, totally agree. Not necessary to run it for every push. Probably, a new command to @tvm-bot can be added to run performance test for a PR if it is needed.
I think it would be nice, and it is very important to have such tests in the CI.

driazati: There's no concrete plans right now (we'd need to do some designing to figure out how to provision machines for benchmarks and where to store results) but I agree having benchmarks run on some cadence would be good for tracking our progress.

apps/benchmark/adreno/bench.py
srkreddy1238: Have plans to support CLML benchmarks too.

tests/scripts/ci.py
srkreddy1238: Script is ready to be used with CI provided we have Snapdragon devices on CI. For now we are using it internally.

apps/benchmark/adreno/adreno_gpu_bench_texture.py
srkreddy1238: Already done. https://github.com/tlc-pack/tophub/pull/27

srkreddy1238: Tuning option is left for tuning future Adreno variants if required.

apps/benchmark/adreno/adreno_gpu_bench_texture.py
srkreddy1238: Technically we have tuning cache implicate now from tophub. Let me remove this line.

masahi (approved):

PR #13679, opened 2023-01-02
[BENCHMARKS][ADRENO] Documentation for Adreno (Texture) benchmarks

srkreddy1238:
Steps to reproduce Adreno (Texture) benchmarks via adreno docker on given device.

### DIFF
--- a/apps/benchmark/README.md
+++ b/apps/benchmark/README.md
@@ -37,7 +37,8 @@ It is recommended that you run tuning by yourself if you have your customized ne
 Please follow the tutorial for
 [NVIDIA GPU](https://tvm.apache.org/docs/tutorials/autotvm/tune_conv2d_cuda.html),
 [ARM CPU](https://tvm.apache.org/docs/tutorials/autotvm/tune_relay_arm.html),
-[Mobile GPU](https://tvm.apache.org/docs/tutorials/autotvm/tune_relay_mobile_gpu.html).
+[Mobile GPU](https://tvm.apache.org/docs/tutorials/autotvm/tune_relay_mobile_gpu.html) and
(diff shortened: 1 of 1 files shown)


masahi (approved):

apps/benchmark/README.md
```diff
@@ -127,3 +128,16 @@ Build TVM with LLVM and ROCm enabled. [Help](https://tvm.apache.org/docs/install
...
+
+### Adreno GPU
+
+Adreno benchmarks are automated over the docker - [ci_adreno](https://github.com/apache/tvm/blob/main/docker/Dockerfile.ci_adreno).
+Adreno docker share the Android devices from host. It is adviced to have host adb version same as docker, which is ```1.0.41```
+
+Below command runs all the benchmarks over given Android device.
+```bash
+export ANDROID_SERIAL=<ADB ID>
+./tests/scripts/ci.py adreno -b
```
echuraev, line 140: @srkreddy1238, @driazati Do you have any ideas, why after running this command I got the following error?

```
Running 'bash .ci-py-scripts/ci-adreno-prhyz.sh' inside ci_adreno...
Unable to find image 'ci_adreno:latest' locally
docker: Error response from daemon: pull access denied for ci_adreno, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
See 'docker run --help'.
``` 

It is not the first time when I don't have access to the `adreno_ci` docker image. I would like to have this access.

driazati: It's not an access issue, the adreno Docker image was never added to the [tlcpack docker hub](https://hub.docker.com/u/tlcpack) so there is nothing to download. You'll need to build the image locally (`bash docker/build.sh ci_adreno`) or fix it so adreno is uploaded along with the other images.

PR #13683, opened 2023-01-03
[CMake][OpenCL] Remove warning for OpenCL wrapper

Lunderberg:
Previously, setting `set(USE_OPENCL ON)` would result in a warning, stating that the runtime wrapper for OpenCL would be used.  Since this is the desired behavior when OpenCL support is enabled, and is not something that a user should fix, this commit removes the warning.

### DIFF
--- a/cmake/modules/OpenCL.cmake
+++ b/cmake/modules/OpenCL.cmake
@@ -43,7 +43,6 @@ if(USE_OPENCL)
   tvm_file_glob(GLOB RUNTIME_OPENCL_SRCS src/runtime/opencl/*.cc)
 
   if(${USE_OPENCL} MATCHES ${IS_TRUE_PATTERN})
-    message(WARNING "Build with OpenCL wrapper")
     file_glob_append(RUNTIME_OPENCL_SRCS
       "src/runtime/opencl/opencl_wrapper/opencl_wrapper.cc"
     )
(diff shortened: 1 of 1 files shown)


Lunderberg:
@tvm-bot rerun

echuraev (commented):
I have added this warning message especially to notify the user that the OpenCL wrapper will be used, and it might lead to a runtime error if the OpenCL library doesn't exist on a target device.
If warning message might embarrass, let's keep just status message?

Lunderberg:
> I have added this warning message especially to notify the user that the OpenCL wrapper will be used, and it might lead to a runtime error if the OpenCL library doesn't exist on a target device.

Good point, as that is a runtime error that would occur on first usage, whereas all the other library dependencies would cause a failure when importing `libtvm_runtime.so`.

> If warning message might embarrass, let's keep just status message?

I like this idea, and have the message back in, but with `message(STATUS ...)` instead of `message(WARNING ...)`.  I also adjusted the wording of the message to make it clear what the wrapper does (runtime search for library) and not just that a wrapper exists.  Could you check that the updated wording is accurate?

cmake/modules/OpenCL.cmake
```diff
@@ -43,7 +43,7 @@ if(USE_OPENCL)
   tvm_file_glob(GLOB RUNTIME_OPENCL_SRCS src/runtime/opencl/*.cc)
 
   if(${USE_OPENCL} MATCHES ${IS_TRUE_PATTERN})
-    message(WARNING "Build with OpenCL wrapper")
+    message(STATUS "Enabled runtime search for OpenCL installation location")
```
echuraev, line 46: I would say `library location` but installation is also ok for me.
```suggestion
    message(STATUS "Enabled runtime search for OpenCL library location")
```

echuraev (approved):
LGTM. Thanks!

PR #13685, opened 2023-01-03
[AOT]Aot module post-test error workaround

janetsc:
The deletion of the AotModule python object results in the device shutting down RPC too early, causing an assertion in the DSP.  This is a temporary workaround while this is still being investigated.  The workaround is to keep a reference to the AotModule object in the HexagonSession object, so it stays until the session has been shut down.


python/tvm/contrib/hexagon/session.py
```diff
@@ -248,8 +248,8 @@ def get_aot_executor(
         GraphModule :
             Runtime graph module that can be used to execute the graph.
         """
-        aot_mod = self.load_module(module_file)
-        return tvm.runtime.executor.AotModule(aot_mod["default"](self.device))
```
mehrdadh, line 252: Please add a github issue describing the problem and link it here. Also mention that this is a temporary fix

mehrdadh (changes requested):

python/tvm/contrib/hexagon/session.py
janetsc: Filed https://github.com/apache/tvm/issues/13741

mehrdadh (approved):

PR #13689, opened 2023-01-03
[docs] Remove empty code blocks

driazati:
This removes the linter that checks that each sphinx gallery tutorial begins with the proper boilerplate by extracting it out to sphinx gallery reset functions. This runs automatically so no code in the tutorials nor associated linters are necessary.


tests/lint/check_request_hook.py
```diff
@@ -41,19 +41,24 @@ def find_code_block_line(lines: List[str]) -> Optional[int]:
     """
     in_multiline_string = False
     in_sphinx_directive = False
+    in_standalone_comment = True
 
     i = 0
     lines = lines[APACHE_HEADER_LINES:]
     while i < len(lines):
         line = lines[i].strip()
+        in_comment = False
```
guberti, line 50: delete?
```suggestion
```

guberti (approved):
LGTM - the empty code blocks thing has bugged me for a while.

guberti:
For what it's worth, I've never been a fan of how we need to add the following to _every_ tutorial:
```python
# sphinx_gallery_start_ignore
from tvm import testing

testing.utils.install_request_hook(depth=3)
# sphinx_gallery_end_ignore
```
Instead, we could use [`reset_modules`](https://sphinx-gallery.github.io/stable/advanced.html#custom-reset) from sphinx-gallery's configuration to run code before every example. It would be pretty straightforward:

```python
from tvm import testing
def add_hook(gallery_conf, fname):
    testing.utils.install_request_hook(depth=3)

sphinx_gallery_conf = {
    ...
    'reset_modules': (add_hook),
}
```

If we needed the hook for `pytest` as well, we could add a `conf.py` file to the `gallery` directory as well. We shouldn't change this here, but it might be worth considering in the future @driazati.

driazati:
@guberti that makes a lot more sense, I updated the PR to use that and delete the linter which wasn't really correct anyways (some of the preamble was misplaced in a few files).

guberti:
I love this PR.

<p align="center">
  
</p>

guberti (changes requested):
This is a good change, but there's one nuance to highlight - per-file config options like `sphinx_gallery_requires_cuda`. If we take them out of the ignore block, the flag renders in HTML, which we do not want:

image

---

I can see two ways to fix this behavior. The easier, uglier way is to continue wrapping all flags in ignore blocks:

```python
# sphinx_gallery_start_ignore
# sphinx_gallery_requires_cuda = True
# sphinx_gallery_end_ignore
```

This is what's done right now, and changing this may be out of scope for this pull request. 

A better alternative would be updating `sphinx-gallery` to version [4484d35](https://github.com/sphinx-gallery/sphinx-gallery/tree/4484d35bb556bfe081156f210904abce8732ffbf) or beyond, and then setting `remove_config_comments=True` in the config file. Unfortunately, [4484d35](https://github.com/sphinx-gallery/sphinx-gallery/tree/4484d35bb556bfe081156f210904abce8732ffbf) is not part of any PyPI releases right now, making this a little more annoying.

What do you think @driazati?

driazati:
I updated the PR to wrap the configs with ignores and made sure they are inside a code block so they don't show up on their own, so now this PR should make everything look fine from the frontend. We could merge this then update the sphinx gallery version (installing from the latest commit isn't a problem with Docker) and remove the `_ignore` tags.

guberti (approved):
Sounds good - we'll merge this, and can consider updating `sphinx-gallery` later. LGTM!

Issue #13674, opened 2022-12-30
[Bug] [TVMScript] T.power or T.pow?


lightzhan-intellif:
Hi all,
When I use TVMScript, I find a problem about the roundtripability of the intrinsic `tir.pow`. Currently, we use `T.power` to express that we want to call the intrinsic `tir.pow`, but when we print the script, it becomes "T.pow" which can not be parsed by the parser any more.

So, for the roundtripability of the intrinsic `tir.pow`, which name should we use?

(For the printer, it will use the intrinsic name `pow` to print, so if we use `T.power`, it maybe means that we need to change the intrinsic name.)

cc @junrushao

junrushao:
Thanks for bringing this up! The naming inconsistency is definitely an issue we should fix. Ideally, if the intrinsic name is `pow`, we should probably stick with `pow` everywhere. Would you mind sending a fix?

lightzhan-intellif:
Of course not.

PR #13692, opened 2023-01-04
[BugFix][TVMScript] Fix the roundtripability of  intrinsic pow

lightzhan-intellif:
Hi,
This is a bug fix discussed [here](https://github.com/apache/tvm/issues/13674). Any further suggestion is welcome.

cc @junrushao

### DIFF
--- a/python/tvm/script/ir_builder/tir/ir.py
+++ b/python/tvm/script/ir_builder/tir/ir.py
@@ -1532,7 +1532,7 @@ min_value = _op_wrapper(_tir_op.min_value)
 nearbyint = _op_wrapper(_tir_op.nearbyint)
 nextafter = _op_wrapper(_tir_op.nextafter)
 popcount = _op_wrapper(_tir_op.popcount)
-pow = _op_wrapper(_tir_op.power)  # pylint: disable=redefined-builtin
+pow = _op_wrapper(_tir_op.pow)  # pylint: disable=redefined-builtin
 q_multiply_shift = _op_wrapper(_tir_op.q_multiply_shift)
 q_multiply_shift_per_axis = _op_wrapper(_tir_op.q_multiply_shift_per_axis)
 ret = _op_wrapper(_tir_op.ret)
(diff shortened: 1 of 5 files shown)


lightzhan-intellif:
@tvm-bot rerun

tests/python/unittest/test_tvmscript_roundtrip.py
```diff
@@ -3622,5 +3622,14 @@ def test_return_none_no_trailing_type():
     assert "-> None" not in script
 
 
+def test_pow_roundtripable():
```
Hzfengsy, line 3625: Could you please use the `ir_generator` to make tests consistent in this file?

junrushao (approved):
LGTM. Please address comments from @Hzfengsy

junrushao:
BTW, there are some existing logics that depend on the python API `tir.power`, so please keep a copy as it is for backward compatibility

tests/python/unittest/test_tvmscript_roundtrip.py
lightzhan-intellif: Okay.

lightzhan-intellif:
Do you mean that users can use both `tir.pow` and `tir.power`?  I think it looks like a little bit confusing and redundant. How about we do not make any change to `tir.power`?

python/tvm/script/ir_builder/tir/ir.py
```diff
@@ -1532,7 +1532,7 @@ def wrapped(*args, **kwargs):
 nearbyint = _op_wrapper(_tir_op.nearbyint)
 nextafter = _op_wrapper(_tir_op.nextafter)
 popcount = _op_wrapper(_tir_op.popcount)
-power = _op_wrapper(_tir_op.power)
+pow = _op_wrapper(_tir_op.power)  # pylint: disable=redefined-builtin
```
junrushao, line 1535: I was thinking if we could add a `pow` method to `tvm/tir/op.py` so that it could be consistent

lightzhan-intellif: Done.

Hzfengsy (approved):

PR #12839, opened 2022-09-19
[BUILD] Re-enable ccache by default

tkonolige:
Previously ccache was disabled because of possible issues with hexagon. Re-enabling it to provide a best effort attempt at using it.

@driazati @cconvey

PR #13697, opened 2023-01-04
[Build][Bugfix] Use CMAKE_ prefix for <LANG>_COMPILER_LAUNCHER

Lunderberg:
Previously, when using `set(USE_CCACHE AUTO)`, the cmake config would set variables `CXX_COMPILER_LAUNCHER` and `C_COMPILER_LAUNCHER`. While there are the target-specific properties named [`<LANG>_COMPILER_LAUNCHER`](https://cmake.org/cmake/help/latest/prop_tgt/LANG_COMPILER_LAUNCHER.html), cmake doesn't check for their use as global variables.

This commit updates the build file to instead set the [`CMAKE_<LANG>_COMPILER_LAUNCHER`](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_LAUNCHER.html) variables, which are used as the default for the `<LANG>_COMPILER_LAUNCHER` property.


Lunderberg:
@tkonolige This is a follow-up to #12839.  I noticed that ccache wasn't being used in my dev environment, and it looks like there was a different variable to be used.

tkonolige (approved):
From my reading of the cmake documentation it seems like `CXX_COMPILER_LAUNCHER` should work. But using `CMAKE_CXX_COMPILER_LAUNCHER` should also work, so I'm fine switching to it.

PR #13700, opened 2023-01-05
[TIR][Schedule] Improve cache_index to cache common subexpressions

multiverstack-intellif:
Much of the index computation is duplicated, so this PR is to do common subexpression analyze when performing cache_index.
And because all block var related computation could be cached, buffer-based cache is somehow too restrictive. So here make it block-based.

A threshold number is needed when identifying a common sub expr, since we only want to cache the most frequently appeared sub expressions.  A common use case may be like this: calling _tvm.arith.detect_common_subexpr()_ and determining based on its result whether to perform cache_index and its corresponding threshold.
cc @wrongtest-intellif @Hzfengsy

### DIFF
--- a/python/tvm/tir/schedule/schedule.py
+++ b/python/tvm/tir/schedule/schedule.py
@@ -1294,7 +1294,10 @@ class Schedule(Object):
 
     @type_checked
     def cache_index(
-        self, block: Union[BlockRV, str], buffer_index: Union[int, str, Buffer]
+        self,
+        block: Union[BlockRV, str],
+        storage_scope: str,
+        cse_thresh: int = 0,
     ) -> List[BlockRV]:
         """Create a block to cache precomputed index for later use.
         if there is no index computation, keep unchanged.
@@ -1304,8 +1307,12 @@ class Schedule(Object):
         block : Union[BlockRV, str]
             The target block operates on the target buffer.
 
-        buffer_index: int
-            The index of the target buffer in block's read region
+        storage_scope: str
+            The storage scope of cached block.
+
+        cse_thresh: int
+            The repeat threshold that determines a common sub expr,
+            default 0 means cache all index computation.
 
 
         Returns
@@ -1334,7 +1341,7 @@ class Schedule(Object):
 
             sch = tir.Schedule(resize)
             block_a = sch.get_block("A")
-            sch.cache_index(block_a, 0)
+            sch.cache_index(block_a, "global", 1)
             print(sch.mod["main"].script())
 
         After applying cache_index, the IR becomes:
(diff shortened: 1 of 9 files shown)


python/tvm/tir/schedule/schedule.py
```diff
@@ -1289,7 +1289,10 @@ def cache_inplace(data_io: T.Buffer[64, "int32"]) -> None:
 
     @type_checked
     def cache_index(
-        self, block: Union[BlockRV, str], buffer_index: Union[int, str, Buffer]
+        self,
+        block: Union[BlockRV, str],
+        storage_scope: str,
+        cse_thresh: int,
```
Hzfengsy, line 1295: Can we set a default value for `cse_thresh`? e.g. 0/-1 for not enabling cse?

tests/python/unittest/test_tir_schedule_cache_index.py
```diff
@@ -66,13 +65,408 @@ def resize_cache_index(
...
+):
+    for i0, i1, i2, i3 in T.grid(1, 3, 80, 80):
+        with T.block("resize"):
+            i0_1, i1_1, i2_1, i3_1 = T.axis.remap("SSSS", [i0, i1, i2, i3])
+            T.reads(x[i0_1, i1_1, 0:40, 0:40])
+            T.writes(resize[i0_1, i1_1, i2_1, i3_1])
+            resize[i0_1, i1_1, i2_1, i3_1] = T.Cast(
+                "float16",
+                (
+                    T.Cast(
```
Hzfengsy, line 80: Can we have a much simpler case for better readability? Also, please use `#fmt: off` to turn off formatter for complex TVMScript

python/tvm/tir/schedule/schedule.py
multiverstack-intellif: > Can we set a default value for `cse_thresh`? e.g. 0/-1 for not enabling cse?

OK, I'll set 0 as caching all the index computation for the target block, should be useful for simple cases and expensive in complex cases.

tests/python/unittest/test_tir_schedule_cache_index.py
multiverstack-intellif: > Can we have a much simpler case for better readability? Also, please use `#fmt: off` to turn off formatter for complex TVMScript

The other test case in the same file is a simpler case, is that too simple? 
What formatter suggest here takes too many lines really, but actually I think it's easier to read in this case.

Hzfengsy: That's fine

python/tvm/tir/schedule/schedule.py
multiverstack-intellif: It's done. @Hzfengsy

Hzfengsy (approved):

wrongtest-intellif (approved):

multiverstack-intellif:
@tvm-bot rerun

multiverstack-intellif:
@tvm-bot rerun

PR #13706, opened 2023-01-05
Add Name Transforms for Rust style

Mousius:
Adding transformations to Rust-style alongside the existing C-style transforms.

Co-authored-by: Ashutosh Parkhi <ashutosh.parkhi@arm.com>

Issue #13705, opened 2023-01-05
[Tracking Issue] Rust Embedded API


Mousius:
### This issue is to track progress for Rust Embedded API
- [ ] P1. Add Rust name transforms - https://github.com/apache/tvm/pull/13706
- [ ] P2. Add Rust interface generator - https://github.com/apache/tvm/pull/13707
- [ ] P3. Initial Rust MLF generation and testing
- [ ] P4. Add Rust Traits Crate
- [ ] P5. Rust Crate MLF generation
- [ ] P6. Rust API alignment with RFC
- [ ] P7. Rust Device API

PR #13707, opened 2023-01-05
[Rust] Initial generator for Embedded Interface

Mousius:
(Builds on #13706)

This is based on the older API in the RFC, we're landing this now so as to upstream it iteratively and allow others to see it in action 😸

There'll be a follow up, as noted in https://github.com/apache/tvm/issues/13705 to update this to the fully idiomatic Rust API

Co-authored-by: Ashutosh Parkhi <ashutosh.parkhi@arm.com>

### DIFF
--- a/src/target/source/interface_rust.cc
+++ b/src/target/source/interface_rust.cc
@@ -1,0 +1,380 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/*!
+ * \file interface_rust.cc
+ * \brief Generates a Rust interface header for a given modules inputs and outputs
+ * which works on top of the C interface API
+ */
+
+#include <tvm/runtime/container/array.h>
+#include <tvm/runtime/container/string.h>
+#include <tvm/runtime/module.h>
+#include <tvm/runtime/name_transforms.h>
+#include <tvm/runtime/packed_func.h>
+#include <tvm/runtime/registry.h>
+#include <tvm/tir/usmp/utils.h>
+
+#include <string>
+
+#include "../../relay/backend/name_transforms.h"
+#include "codegen_params.h"
+
+namespace tvm {
+namespace codegen {
+
+using runtime::PackedFunc;
+using namespace tvm::relay::backend;
+using namespace tvm::runtime;
+
+class InterfaceRustNode : public runtime::ModuleNode {
+ public:
+  InterfaceRustNode(std::string module_name, Array<String> input_names, Array<String> output_names,
+                    Array<tir::usmp::AllocatedPoolInfo> pools,
+                    Map<String, tir::usmp::PoolAllocation> io_pool_allocations,
+                    Array<String> devices, int workspace_size,
+                    Map<String, Map<String, ObjectRef>> inputs,
+                    Map<String, Map<String, ObjectRef>> outputs)
+      : module_name_(module_name),
+        inputs_(inputs),
+        outputs_(outputs),
+        devices_(devices),
+        input_names_(input_names),
+        output_names_(output_names),
+        pools_(FilterExternalPools(pools)),
+        io_pool_allocations_(io_pool_allocations),
+        workspace_size_(workspace_size) {
+    ICHECK(io_pool_allocations_.empty()) << "Workspace Memory Pools IO unsupported";
+  }
+  const char* type_key() const final { return "h"; }
+
+  std::string GetSource(const std::string& format) final {
+    std::stringstream code;
+
+    EmitBrief(code, "Input tensors");
+    EmitDataStruct(code, "inputs", inputs_, input_names_);
+    EmitBrief(code, "Output tensors");
+    EmitDataStruct(code, "outputs", outputs_, output_names_);
+
+    if (!devices_.empty()) {
+      EmitBrief(code, "Device context pointers");
+      EmitDeviceStruct(code, "devices", devices_);
+    }
+
+    if (!pools_.empty()) {
+      EmitBrief(code, "Workspace pools");
+      EmitWorkspacePoolsStruct(code);
+    }
+
+    EmitRustRunFunction(code);
+    EmitCRunFunction(code);
+
+    EmitIntegerValueConst(code, "Workspace size", "WORKSPACE_SIZE", workspace_size_);
+    EmitMemoryPools(code);
+
+    return code.str();
+  }
+
+  PackedFunc GetFunction(const std::string& name, const ObjectPtr<Object>& sptr_to_self) final {
+    return PackedFunc();
+  }
+
+ private:
+  constexpr static const char* _macro_workspace_pool_size_postfix = "_WORKSPACE_POOL_SIZE";
+  constexpr static const char* _macro_constant_pool_size_postfix = "_CONSTANT_POOL_SIZE";
+  constexpr static const char* _macro_constant_pool_data_postfix = "_constant_pool_data";
+
+  void EmitBrief(std::stringstream& code_stream, const std::string& description) {
+    code_stream << "/// " << description << " for TVM module \"" << module_name_ << "\"\n";
+  }
+
+  void EmitMemoryPools(std::stringstream& code) {
+    for (const tir::usmp::AllocatedPoolInfo pool : pools_) {
+      String pool_name = pool->pool_info->pool_name;
+      Integer pool_size = pool->allocated_size;
+      if (const auto* pool_info = pool->pool_info.as<ConstantPoolInfoNode>()) {
+        EmitConstantPool(code, SanitizeName(pool_name) + " initialization data", pool_info);
+      } else {
+        EmitIntegerValueConst(code, SanitizeName(pool_name) + " size",
+                              SanitizeName(pool_name) + _macro_workspace_pool_size_postfix,
+                              pool_size->value);
+      }
+    }
+  }
+
+  void EmitConstantPool(std::stringstream& code_, const std::string& brief_description,
+                        const ConstantPoolInfoNode* pool_info) {
+    EmitBrief(code_, brief_description);
+    std::string const_name_prefixed = ToRustConstantStyle(SanitizeName(pool_info->pool_name));
+    std::string macro_name_prefixed = ToRustMacroStyle(SanitizeName(pool_info->pool_name));
+
+    if (pool_info->constant_info_array.size() > 0) {
+      std::vector<ConstantInfo> const_info_vec(pool_info->constant_info_array.begin(),
+                                               pool_info->constant_info_array.end());
+      std::sort(const_info_vec.begin(), const_info_vec.end(),
+                [](const ConstantInfo& a, const ConstantInfo& b) {
+                  return a->byte_offset->value < b->byte_offset->value;
+                });
+      int64_t accumulated_pool_len =
+          const_info_vec.back()->byte_offset.IntValue() +
+          runtime::GetDataSize(*const_info_vec.back()->data.operator->());
+      const auto& accumulated_pool = runtime::NDArray::Empty(
+          {accumulated_pool_len}, DataType::UInt(8), const_info_vec.back()->data->device);
+      for (const auto& const_info : const_info_vec) {
+        const auto& data = const_info->data;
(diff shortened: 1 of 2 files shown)


src/target/source/interface_rust.cc
```diff
@@ -0,0 +1,380 @@
...
+    code_stream << "/// Entrypoint function for TVM module \"" << module_name_ << "\"\n"
+                << "/// # Arguments\n";
+    if (io_pool_allocations_.empty()) {
+      code_stream << "/// * `inputs` - Input tensors for the module\n";
+      code_stream << "/// * `outputs` - Output tensors for the module\n";
+    }
+
+    if (!pools_.empty()) {
+      code_stream << "/// * `workspace_pools` - Workspace memory pools for the module\n";
+    }
```
tqchen, line 303: Given that this is mainly for micro TVM, it would be great to keep it as target/source/microtvm_interface_rust.cc

tqchen (dismissed):

src/target/source/interface_rust.cc
Mousius: I'd suggest this is out of scope of this PR given it sits parallel with `interface_c.cc`?

tqchen: That would mean that probably we missed that point when reviewing interface_c.cc :)

It would be a great motivator to also scoping that as well

Mousius: I'm not disagreeing with the sentiment, I'm suggesting it's out of scope within this PR as the PR works within the existing framework.

tqchen: This PR proposes a change to generate rust code, and existing framework contains rust code that interacts with tvm in a different way. 

As a result, it would useful to apply proper approaches to clarify the particular file. For example, `target/source/interface_embedded_rust.cc`  would be a fine approach.

Since it clearly signals that there can be other TVM rust interfaces.

Mousius: > As a result, it would useful to apply proper approaches to clarify the intent of the particular file. For example, one possible approach would be target/source/interface_embedded_rust.cc and comments that it intends to generate files that interfaces with embedded rust interface.

This would be different from the existing `interface_c.cc` which is also designed for the embedded use-case, which would mean this patch introduces inconsistency. If we want consistent naming around this we should figure out what we want separately and apply it to both APIs.

Mousius: As the original author of `interface_c.cc` and the Embedded APIs, I will ensure that this is cleaned up properly in line with the other Embedded APIs.

src/target/source/interface_rust.cc
```diff
@@ -0,0 +1,381 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
```
tqchen, line 3: Would be great to rename to interface_embedded_rust.cc, or place it in a separate namespace.

Both approach would enable deferred changes in followups if necessary while not disrupting existing understanding of the code base in the meantime.

tqchen (changes requested):
Hi @Mousius , it would still be great to take this action now mainly to have a clear scope of intent (for other developers), not necessarily on the particular code content itself

src/target/source/interface_rust.cc
tqchen: if desirable, you are also more than welcome to bring the set of changes to a branch, in which case feel free to continue and complete the work along with followups, and we can take a look again when you feel they are ready.

tqchen:
Thanks @Mousius i outlined the goal, and possible approaches. Technically I think they are all feasible and actionable. By listing these suggestions, of course that is my best intent of collaboration since we also need to reconcile the needs and overall architectural decisions in the project as a whole.

Mousius:
> Thanks @Mousius i outlined the goal, and possible approaches. Technically they are all feasible and actionable.
> 
> By listing these suggestions, of course that is my best intent of collaboration since we also need to reconcile the needs and overall architectural decisions in the project as a whole. Looking at your reply, you probably might could use W2 (so the changes are properly fixed together as a sequeqnce plus any followup fixes that you might prefer).
> 
> Please feel free to also suggest other possible actions that can address G0

I have put forward suggestions, you have not included them:
* This is a non-issue, both Rust and C are general-purpose languages with embedded specific engineers, there is no need to discriminate.
* If this is an issue, I have offered to fix this more broadly and fix both interface APIs which will likely result in a much broader discussion on how to make this clearer in future.

I was more than happy to work on something to fix the Embedded APIs, as many C engineers work in system environments which aren't embedded; this is a broken window in our overall architecture that doesn't seem to be recognised and thus there appears to be no value in such an RFC. 

> For example use the term `interface_embedded_rust` would leave room and allow developers to not have confusion given there can be other interfaces of rust, as per discussion of rationales in this RFC [apache/tvm-rfcs#96](https://github.com/apache/tvm-rfcs/pull/96)

Yip, I was hopeful that given responses elsewhere, there would be some interest in moving this forwards with an open mindset and encouragement to follow up with a broader discussion on the proper fix rather than being blocked on a file name. And we have fair precedent for moving forwards without an RFC so thought I would test the water here.

> W1: We can try to open a thread in the community to get a broad sense of what everyone's inputs are (wrt to the overall naming and scoping), which i would also be interested in hearing from.
> W2: If we feel a sequence of changes is needed that goes beyond a single PR to bring the set of things into a ready state, we can also try to open a branch, in which case the change is also reasonably isolate, and we are more than welcome to commit the change to branch, continue followup developments, and we can re-access when we think things are at a ready state to address G0.

* W1 is just another place for the same cyclic discussion demonstrated both here and in the RFC
* W2 moves the issue to the future, which is a shame for such a small change which could benefit real users - given the discussions here I don't see this ever landing in `main` unless I fulfil W0.

> W0: Do the renaming directly, since I think it won't affect the functionality of the proposed feature and related users, but would also benefit other users/developers who use rust API in different ways. Feel free to bring followup changes to also align the c if necessary, although given the context of C(more in embedded i do not feel that is a requirement)

Overall, I do not want to introduce these inconsistencies in the codebase nor do I wish to endorse the view that both C and Rust are not general purpose languages, yet I feel forced to W0 in order to progress.

tqchen:
Might indeed is true that G0 is non-issue for the intended audience of this PR. Again G0 is not only about intended audiences of the PR, but also the general TVM project and non-embedded uses(and differentiate them from the embedded interface), and how to make our overall namespacing structure clear so we can empower both cases. G0 exist because rust is a general purpose language.

If there is a suggestion to clarifying the scope of embedded API in the project for both C and rust, that could indeed be a better outcome, W2  likely offer a way to do so. Personally I also think the name of embedded c interface could be further clarified along the same line, but we can take one step at a time and given all the background mentioned, I do not take that ask as a blocker. But if there is a feeling that such sync should be done, either first make the c change in main then move on to the new style, or doing them in arbitrary order in W2 are both feasible. Taking W0 then updating embedded C would also achieve that goal as well.

When W2 path is taken, there is of course not blocked on whether we have to take W0, as long as G0 is achieved. For example, moving both C and rust generator under a different namespace would be a reasonable choice that is not consistent with W0 but still achieve the goal. 

Usually it is indeed OK to defer some decisions of a module, provided that they are reasonabley isolated and clarified wrt to the overall scope of the project. Such requirements usually can be easily satiesfied with proper namespacing and isolation, to allow developer to own that part of submodule/subfolder, as long as their relation is satisified.  Again considering the small impact of W0 to the intended users of the PR, and the clarity it offers to the wider community. I do not see it as a necessary bad. But there are also other paths available like I outlined.

Mousius:
> If there is a suggestion to clarifying the scope of embedded API in the project for both C and rust, that could indeed be a better outcome, W2 likely offer a way to do so. Personally I also think the name of embedded c interface could be further clarified along the same line, but we can take one step at a time and given all the background mentioned, I do not take that ask as a blocker. But if there is a feeling that such sync should be done, either first make the c change in main then move on to the new style, or doing them in arbitrary order in W2 are both feasible. Taking W0 then updating embedded C would also achieve that goal as well.
> 
> When W2 path is taken, there is of course not blocked on whether we have to take W0, as long as G0 is achieved. For example, moving both C and rust generator under a different namespace would be a reasonable choice that is not consistent with W0 but still achieve the goal.
> 
> Usually it is indeed OK to defer some decisions of a module, provided that they are reasonabley isolated and clarified wrt to the overall scope of the project. Such requirements usually can be easily satiesfied with proper namespacing and isolation, to allow developer to own that part of submodule/subfolder, as long as their relation is satisified. Again considering the small impact of W0 to the intended users of the PR, and the clarity it offers to the wider community. I do not see it as a necessary bad. But there are also other paths available like I outlined as well for G0 as well.

Thank you for your clarification @tqchen, I've articulated why I do not want to introduce inconsistencies by implementing W0, and you've effectively summarised that this work is blocked until either the rename you want is done or completion of all the additional rework of the existing API. I can understand why others choose to fork TVM, this workflow makes it difficult to deliver value to end users with the limited resources I have as a contributor.

tqchen:
hi @Mousius . As a project, we do want to empower different kinds of developers and diverse sets of users. In order to do so, we need multiple modules and standard software eng practices to provide such clarity among things, especially when it comes to user-facing interfaces when multiple variants co-exist. Naming spacing or proper naming is not a hard requirement, and we do it every day to both gain clarity and provide space for other developers/users. With proper name-spacing, likely most of us would be OK with partial, incomplete implementation first and continued evolution of the module -- this is a very flexible requirement that most of our developers follow.

As such G0 being listed is an important consideration for the project as a whole. I am not insisting on one only way to achieve that goal but to provide suggestions on what is being possible. As I said, I am also open to other thoughts on achieving G0.

Mousius:
> hi @Mousius . As a project, we do want to empower different kinds of developers and diverse sets of users. In order to do so, we need multiple modules and standard software eng practices to provide such clarity among things, especially when it comes to user-facing interfaces when multiple variants co-exist. Naming spacing or proper naming is not a hard requirement, and we do it every day to both gain clarity and provide space for other developers/users. With proper name-spacing, likely most of us would be OK with partial, incomplete implementation first and continued evolution of the module -- this is a very flexible requirement that most of our developers follow.
> 
> As such G0 being listed is an important consideration for the project as a whole. 

@tqchen, this has now returned to the cyclic discussion seen both here and on the RFC, your reply suggests that it is not a hard requirement, yet you are continuing to block the work proceeding with insistence on the same outcomes which push towards the one you want (see https://github.com/apache/tvm/pull/13707#issuecomment-1424336022 for where I've summarised this previously).

tqchen:
Sorry about my choice of terms as that causes confusion, what i meant is that adding name spacing is not hard to fullfill, updated to clarified the point. Again I am pointing out the goal G0, which I think have been quite clear through out the thread.

Mousius:
> Sorry about my choice of terms as that causes confusion, what i meant is that adding name spacing is not hard to fullfill, updated to clarified the point. Again I am pointing out the goal G0 is something we need to consider(and fulfill as part of changes), which I think have been quite clear through out the thread.

I have accepted it is not hard to fulfil, it is just the wrong thing to do, please see https://github.com/apache/tvm/pull/13707#issuecomment-1424336022 as to why it is the wrong thing to do based on the reality of both languages (see also: [Rust](https://en.wikipedia.org/wiki/Rust_(programming_language)) and [C](https://en.wikipedia.org/wiki/C_(programming_language))).

You have been quite clear on the goal you've created for this work and how you would like it carried out, there is no further need for you to attempt to push your agenda. I will contemplate where best to continue this development.

tqchen:
A technical decision being right or wrong depends on the context of the project. For something that sits under global namespace that context would be the broader project. Hopefully through out this thread I have explained the rationales under the context of the broader project, and existence of other API interfaces than that are not for embedded purposes as a result a need for better clarity.  I have no agenda to push, and merely stating an overall principle that can enable clarity and empowerment of different users/developers, along with suggestions, that is all.

Mousius:
> A technical decision being depends on the context. There is always a tradeoff being considered taking the project as a whole into consideration, rather than having everything outright being right or wrong. For something that sits under global namespace that context would be the broader project. Hopefully through out this thread I have explained the rationales under the context of the broader project, and existence of other API interfaces than that are not for embedded purposes as a result a need for better clarity. I have no agenda to push, and merely provide guidance about an overall principle that can enable clarity and empowerment of different users/developers, along with suggestions, that is all.

This is not merely provide guidance, you have clearly requested changes in lines with the goal you've created for this work and how you would like it carried out, there is no further need for you to attempt to push your agenda. I will contemplate where best to continue this development.

tqchen:
I have outlined the goals that are derived from the general principle, and provide possible suggestions of options(rather than insisting on one single approach). Additionally, like being stated, we also welcome other possible suggestions that aligns with G0. Hopefully the rationales are pretty clear and I would encourage us to take the overall project into context here.

Mousius:
> I have outlined the goals that are derived from the general principle, and provide possible suggestions of options(rather than insisting on one single approach). Additionally, like being stated, we also welcome other possible suggestions that aligns with G0. Hopefully the rationales are pretty clear and I would encourage us to take the overall project into context here.

@tqchen, please see https://github.com/apache/tvm/pull/13707#issuecomment-1424336022 to re-visit further options which are outside of the goal you have chosen. Please also take time to read the illustration of how the options you have provided are clearly pushing towards the outcome you would want or moving development elsewhere.

As shown in https://github.com/apache/tvm/pull/13707#issuecomment-1424336022, I did take the overall project into context, as a general machine learning compiler we do not need to stereotype specific kinds of engineers to work in specific types of environments outside of the reality of those languages.

Again, this is cyclic, I'm now just referring back to the original comment outlining the issues with the approach you've given - you have made it clear what you want.

PR #13708, opened 2023-01-05
[Arith] Simplify to positive numerators in floordiv/floormod

Lunderberg:
Negative numerators to modulo/remainder operations are not supported by the Vulkan API.  While the SPIR-V instructions
[`OpSRem`](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpSRem) and [`OpSMod`](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpSMod) have identical semantics to `tir::Mod` and `tir::FloorMod`, respectively, use of either instruction within Vulkan results in undefined behavior.  From the [Vulkan spec](https://registry.khronos.org/vulkan/specs/1.3/html/chap37.html#spirvenv-op-prec):

> For the OpSRem and OpSMod instructions, if either operand is negative the result is undefined.
>
> Note: While the OpSRem and OpSMod instructions are supported by the Vulkan environment, they require non-negative values and thus do not enable additional functionality beyond what OpUMod provides.

This issue was first noticed in https://github.com/apache/tvm/pull/13530, where use of integer arithmetic resulted in negative numerators.  This hadn't caused issues previously, because most use of div/mod use a denominator that is a power of two.  In these cases, `tir.LowerIntrin` implements floordiv and floormod using only bitwise operations.  When the denominator isn't a power of two, both `tir::FloorDiv` and `tir::FloorMod` are implemented in terms of `tir::Mod`, which triggers the undefined behavior for negative numerators.

This commit implements additional simplification rules that preferentially removes negative values from the numerators.  For example, simplifying `floormod(i - 2, 8)` to `floormod(i + 6, 8)`, and simplifying `floordiv(i - 2, 8)` to `floordiv(i + 6, 8) - 1`.  These handle the most common case, where some index variable is being offset by a negative constant.

### DIFF
--- a/src/arith/rewrite_simplify.cc
+++ b/src/arith/rewrite_simplify.cc
@@ -253,12 +253,19 @@ PrimExpr RewriteSimplifier::Impl::VisitExpr_(const AddNode* op) {
     TVM_TRY_REWRITE((c1 - y) + c2, (c1 + c2) - y);
     TVM_TRY_REWRITE((y - c1) + c2, y + (c2 - c1));
 
+    TVM_TRY_REWRITE(c1 + (x + c2), x + (c1 + c2));
+    TVM_TRY_REWRITE((x + c1) + c2, x + (c1 + c2));
+    TVM_TRY_REWRITE(c1 + (x - c2), x + (c1 - c2));
+    TVM_TRY_REWRITE((x - c1) + c2, x + (c2 - c1));
+    TVM_TRY_REWRITE(c1 + (c2 - x), (c1 + c2) - x);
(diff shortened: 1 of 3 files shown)


Lunderberg:
When implementing this change, I had also considered making a Vulkan-specific pass, rather than changing the simplification.  However, the effect of such a pass could be undone by a later simplification, so I believe this to be the better solution.

Lunderberg:
The latest change to the simplification caused more breakages than I think it is worth.  Closing this PR in favor of https://github.com/apache/tvm/pull/13724, which has a more limited scope.

PR #13710, opened 2023-01-05
[TIR] Fix dtype mismatch error due to LetStmt

masahi:
Previously, even if the dtype of the rhs of `LetStmt` is casted to int64, the bound variable doesn't get updated. This results in the following invalid IR to be generated:
```
p0_red_temp_v0 = T.alloc_buffer([T.int64(1), T.int64(56), T.int64(56)], dtype="int32")
...
for ax0, ax1, ax2, k3 in T.grid(T.int64(1), T.int64(56), T.int64(56), T.int64(64)):
   with T.block("p0_red_temp"):
        ...
        v_p0_red_temp_v0: T.int32 = T.Select(p0_red_temp_v1[v_ax0, v_ax1, v_ax2] > p0[v_ax0, v_ax1, v_ax2, v_k3] or (p0_red_temp_v1[v_ax0, v_ax1, v_ax2] == p0[v_ax0, v_ax1, v_ax2, v_k3] and T.Cast("int64", p0_red_temp_v0[v_ax0, v_ax1, v_ax2]) < v_k3), T.Cast("int64", p0_red_temp_v0[v_ax0, v_ax1, v_ax2]), v_k3)
        p0_red_temp_v0[v_ax0, v_ax1, v_ax2] = v_p0_red_temp_v0
```

cc @vinx13

### DIFF
--- a/src/tir/ir/data_type_rewriter.cc
+++ b/src/tir/ir/data_type_rewriter.cc
@@ -107,6 +107,25 @@ Stmt DataTypeLegalizer::VisitStmt_(const AttrStmtNode* op) {
   return StmtExprMutator::VisitStmt_(op);
 }
 
+Stmt DataTypeLegalizer::VisitStmt_(const LetStmtNode* op) {
+  PrimExpr value = this->VisitExpr(op->value);
+  Stmt body = this->VisitStmt(op->body);
+  if (value.same_as(op->value) && body.same_as(op->body)) {
+    return GetRef<Stmt>(op);
+  } else if (value.dtype() == op->var->dtype) {
+    auto n = CopyOnWrite(op);
+    n->value = std::move(value);
+    n->body = std::move(body);
+    return Stmt(n);
+  } else {
+    auto new_var = op->var.copy_with_dtype(value.dtype());
+    Map<Var, PrimExpr> vmap{{op->var, new_var}};
+    auto new_body = SubstituteWithDataTypeLegalization(
+        std::move(body), [&](const Var& var) { return vmap.Get(var); });
+    return LetStmt(new_var, value, this->VisitStmt(new_body), op->span);
+  }
+}
+
 PrimExpr DataTypeLegalizer::VisitExpr_(const SelectNode* op) {
   PrimExpr condition = this->VisitExpr(op->condition);
   PrimExpr true_value = this->VisitExpr(op->true_value);
@@ -397,6 +416,9 @@ Stmt IndexDataTypeRewriter::VisitStmt_(const BufferStoreNode* op) {
 
   Buffer new_buffer = GetRemappedBuffer(op->buffer);
   auto value = this->VisitExpr(op->value);
+  if (new_buffer->dtype != value->dtype) {
+    value = cast(new_buffer->dtype, value);
+  }
   auto indices = VisitIndices(op->indices);
 
   if (!new_buffer.same_as(op->buffer) || !value.same_as(op->value) ||
--- a/src/tir/ir/data_type_rewriter.cc
+++ b/src/tir/ir/data_type_rewriter.cc
@@ -131,7 +131,7 @@ Stmt DataTypeLegalizer::VisitStmt_(const LetStmtNode* op) {
 
 PrimExpr DataTypeLegalizer::VisitExpr_(const VarNode* op) {
   if (auto it = var_remap_.find(op); it != var_remap_.end()) {
-    return (*it).second;
+    return it->second;
   }
   return GetRef<Var>(op);
 }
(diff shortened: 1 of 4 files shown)


src/tir/ir/data_type_rewriter.cc
```diff
@@ -397,6 +417,9 @@ Stmt IndexDataTypeRewriter::VisitStmt_(const BufferStoreNode* op) {
 
   Buffer new_buffer = GetRemappedBuffer(op->buffer);
   auto value = this->VisitExpr(op->value);
+  if (new_buffer->dtype != value->dtype && value->dtype.lanes() == 1) {
+    value = cast(new_buffer->dtype, value);
```
masahi, line 421: `new_buffer->dtype != value->dtype` is true if `new_buffer->dtype == float32` and `value->dtype == float32x16`, and the indices is a vector. `value->dtype.lanes() == 1` condition is added to prevent adding a cast in such cases.

For argmax, this cast can result in a redundant chain of casts of the form int32(select(cond, int64(tru), int64(false))), but hopefully LLVM or other backend can clean them up.

Hzfengsy (approved):

src/tir/ir/data_type_rewriter.cc
```diff
@@ -107,6 +107,26 @@ Stmt DataTypeLegalizer::VisitStmt_(const AttrStmtNode* op) {
...
+    auto n = CopyOnWrite(op);
+    n->value = std::move(value);
+    n->body = std::move(body);
+    return Stmt(n);
+  } else {
+    auto new_var = op->var.copy_with_dtype(value.dtype());
+    Map<Var, PrimExpr> vmap{{op->var, new_var}};
+    auto new_body = SubstituteWithDataTypeLegalization(
+        std::move(body), [&](const Var& var) { return vmap.Get(var); });
+    // We need to visit the body again to insert additional casts
```
vinx13, line 125: it's possible to do this without revisiting. We can only visit value at the beginning, and update the vmap before visiting body

masahi: We can pass `op->body` to `SubstituteWithDataTypeLegalization`, but we still need to visit `new_body` after `SubstituteWithDataTypeLegalization`. In that case, the output of the first visit,`this->VisitStmt(op->body)` at the top is only used for the first two `if / else` branches.

So this doesn't work, because the body of the Let has outdated variable dtype. Is this your understanding as well, or are you suggesting something else?
```
    auto new_body = SubstituteWithDataTypeLegalization(
        std::move(op->body), [&](const Var& var) { return vmap.Get(var); });
    // We need to visit the body again to insert additional casts
    return LetStmt(new_var, value, new_body, op->span);
```

masahi: I'm experimenting with the following definition of `LetStmt` visitor, which looks better but it fails to pass the test for some reason (the cast before buffer store is not inserted). Looking.

```
Stmt DataTypeLegalizer::VisitStmt_(const LetStmtNode* op) {
  PrimExpr value = this->VisitExpr(op->value);
  auto body = op->body;
  auto new_var = op->var.copy_with_dtype(value.dtype());

  if (value.dtype() != op->var->dtype) {
    Map<Var, PrimExpr> vmap{{op->var, new_var}};
    body =
        SubstituteWithDataTypeLegalization(op->body, [&](const Var& var) { return vmap.Get(var); });
  }

  Stmt new_body = this->VisitStmt(body);

  if (value.same_as(op->value) && new_body.same_as(op->body)) {
    return GetRef<Stmt>(op);
  } else if (value.dtype() == op->var->dtype) {
    auto n = CopyOnWrite(op);
    n->value = std::move(value);
    n->body = std::move(body);
    return Stmt(n);
  } else {
    // We need to visit the body again to insert additional casts
    return LetStmt(new_var, value, new_body, op->span);
  }
}
```

vinx13: `SubstituteWithDataTypeLegalization` is implemented with `DataTypeLegalizer`. So I'm suggesting instead of calling `SubstituteWithDataTypeLegalization`, we just update the vmap before recursive visiting. Something like:
```
PrimExpr VisitExpr_(const VarNode* op) {
  if (auto it = vmap.find(op); it != vmap.end()) return (*it).second;
  return GetRef<Var>(op);
}

Stmt DataTypeLegalizer::VisitStmt_(const LetStmtNode* op) {
  PrimExpr new_value = this->VisitExpr(op->value);
  if (op->var->dtype != new_value.dtype()) {
    vmap[op->var] = new_var;
    new_body = VisitStmt(op->body);
    return LetStmt(new_var, new_value, new_body);
  }
}
```

masahi: ok fixed, PTAL @vinx13

masahi: sorry just saw your comment, looking now.

masahi: @vinx13 Not sure where that `vmap` is coming from?

vinx13: oh I'm suggesting adding a new `vmap_` field

masahi: I see, there is also `var_remap_` in the derived class, `IndexDataTypeNormalizer`. We may be able to unify

masahi: done

vinx13 (approved):

PR #13713, opened 2023-01-06
[TVMScript] Linter-friendly function definitions

cyx-6:
Initially, the functions like `T.int8`, `T.uint32` and `T.float64x64` were generated by loops and use `globals()` to add symbols globally to reduce code complexity.
But for linters like Pylint may not be able to look into `globals()`, where the functions defined implicitly.
This pr refactors the definitions of these functions explicitly for better linter experience.


junrushao:
Thanks for the PR! I will verify it myself before approving :-)

junrushao (approved):
LGTM!

PR #13717, opened 2023-01-06
[Tests] Replace pytest.main with tvm.testing.main

Mousius:
This ensures that if you want to run a specific test script then at least it's reasonably consistent and as people copy test files they'll use the new function :smile_cat:

### DIFF
--- a/tests/micro/stm32/test_code_emitter.py
+++ b/tests/micro/stm32/test_code_emitter.py
@@ -392,4 +392,4 @@ def test_mnist():
 
 
 if __name__ == "__main__":
-    sys.exit(pytest.main([os.path.dirname(__file__)] + sys.argv[1:]))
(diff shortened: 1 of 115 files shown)


Mousius:
As promised @mehrdadh 😸  I'm fairly sure this is going to fail with some `undefined tvm`. Out of interest do we know why we don't encourage people running tests directly using the `pytest` CLI?

junrushao (approved):
LGTM!

mehrdadh:
@Mousius I'm not sure what's the reason behind it. @driazati could help here.
Thanks for sending the fix!

driazati:
I’ve always used pytest directly (and that’s what is tested in CI), I don’t think it really makes a huge difference but it’s nice to have the test files be directly runnable

PR #13719, opened 2023-01-06
[Hexagon] Add hexagon user DMA intrins for tensorization

nverke:
Added some intrins for user dma on hexagon. Currently these seem to perform worse than all other options used in the test. 

@adstraw

### DIFF
--- a/python/tvm/tir/tensor_intrin/hexagon.py
+++ b/python/tvm/tir/tensor_intrin/hexagon.py
@@ -48,20 +48,20 @@ def generate_dma_load_intrin(
             T.evaluate(
                 T.tvm_call_packed(
                     "device_api.hexagon.dma_copy",
-                    -1, #Use QueueId of -1 to not interfere with async copies. 
+                    -1,  # Use QueueId of -1 to not interfere with async copies.
                     T.address_of(C[0], dtype="handle"),
                     T.address_of(A[0], dtype="handle"),
                     size,
-                    0, #Do not use experimental bypass mode. 
+                    0,  # Do not use experimental bypass mode.
                     dtype="int32",
                 )
             )
             T.evaluate(
                 T.tvm_call_packed(
                     "device_api.hexagon.dma_wait",
-                    -1, 
-                    0, #Wait for the sync queue (-1) to have 0 messages. 
-                    dtype="int32"
+                    -1,
+                    0,  # Wait for the sync queue (-1) to have 0 messages.
+                    dtype="int32",
                 )
             )
 
--- a/python/tvm/tir/tensor_intrin/hexagon.py
+++ b/python/tvm/tir/tensor_intrin/hexagon.py
@@ -27,7 +27,7 @@ def generate_dma_load_intrin(
     """Generator of dma_load intrins"""
 
     @T.prim_func
-    def dma_load_desc(a: T.handle, c: T.handle) -> None:
+    def sync_dma_load_desc(a: T.handle, c: T.handle) -> None:
         A = T.match_buffer(a, (size), dtype, offset_factor=1, scope="global")
         C = T.match_buffer(c, (size), dtype, offset_factor=1, scope="global.vtcm")
         with T.block("root"):
@@ -39,7 +39,7 @@ def generate_dma_load_intrin(
                     C[vii] = A[vii]
 
     @T.prim_func
-    def dma_load_impl(a: T.handle, c: T.handle) -> None:
+    def sync_dma_load_impl(a: T.handle, c: T.handle) -> None:
         A = T.match_buffer(a, (size), dtype, offset_factor=1, scope="global")
         C = T.match_buffer(c, (size), dtype, offset_factor=1, scope="global.vtcm")
         with T.block("root"):
@@ -48,17 +48,24 @@ def generate_dma_load_intrin(
             T.evaluate(
                 T.tvm_call_packed(
                     "device_api.hexagon.dma_copy",
-                    0,
+                    -1, #Use QueueId of -1 to not interfere with async copies. 
                     T.address_of(C[0], dtype="handle"),
                     T.address_of(A[0], dtype="handle"),
                     size,
-                    0,
+                    0, #Do not use experimental bypass mode. 
                     dtype="int32",
                 )
             )
-            T.evaluate(T.tvm_call_packed("device_api.hexagon.dma_wait", 0, 0, dtype="int32"))
+            T.evaluate(
+                T.tvm_call_packed(
+                    "device_api.hexagon.dma_wait",
+                    -1, 
+                    0, #Wait for the sync queue (-1) to have 0 messages. 
+                    dtype="int32"
+                )
+            )
 
-    return dma_load_desc, dma_load_impl
+    return sync_dma_load_desc, sync_dma_load_impl
 
 
 def generate_dot_product_32x4_u8u8i32(mem_scope="global"):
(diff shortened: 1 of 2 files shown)


python/tvm/tir/tensor_intrin/hexagon.py
```diff
@@ -20,6 +20,47 @@
...
+    def dma_load_impl(a: T.handle, c: T.handle) -> None:
+        A = T.match_buffer(a, (size), dtype, offset_factor=1, scope="global")
+        C = T.match_buffer(c, (size), dtype, offset_factor=1, scope="global.vtcm")
+        with T.block("root"):
+            T.reads(A[0:size])
+            T.writes(C[0:size])
+            T.evaluate(
+                T.tvm_call_packed(
+                    "device_api.hexagon.dma_copy",
+                    0,
```
adstraw, line 51: Synchronous DMA uses queue ID -1.  See [here](https://github.com/apache/tvm/blob/60c723ec267ee5095a35add5f8259e650b8ddd7b/src/runtime/hexagon/hexagon_user_dma.h#L37).  This is so as not to interfere with async DMA flow which uses queue IDs starting with 0.  Please use queue -1 and add some comments here.

python/tvm/tir/tensor_intrin/hexagon.py
```diff
@@ -20,6 +20,47 @@
...
 
 
+def generate_dma_load_intrin(
+    size: int,
+    dtype: str,
+):
+    """Generator of dma_load intrins"""
+
+    @T.prim_func
+    def dma_load_desc(a: T.handle, c: T.handle) -> None:
```
adstraw, line 30: Would like this to be called "sync_dma_load_desc" with some comments to distinguish between async and sync (copy and immediate wait) flow.

python/tvm/tir/tensor_intrin/hexagon.py
```diff
@@ -20,6 +20,47 @@
...
+        with T.block("root"):
+            T.reads(A[0:size])
+            T.writes(C[0:size])
+            for i in T.serial(size):
+                with T.block("load"):
+                    vii = T.axis.remap("S", [i])
+                    C[vii] = A[vii]
+
+    @T.prim_func
+    def dma_load_impl(a: T.handle, c: T.handle) -> None:
```
adstraw, line 42: sync_dma_load_impl

python/tvm/tir/tensor_intrin/hexagon.py
```diff
@@ -20,6 +20,47 @@
...
+            T.reads(A[0:size])
+            T.writes(C[0:size])
+            T.evaluate(
+                T.tvm_call_packed(
+                    "device_api.hexagon.dma_copy",
+                    0,
+                    T.address_of(C[0], dtype="handle"),
+                    T.address_of(A[0], dtype="handle"),
+                    size,
+                    0,
```
adstraw, line 55: Need comments, at least to indicate that this is for bypass.  Better would be to tie the setting of this bit to `tir.experimental_dma_bypass_cache` annotation.

python/tvm/tir/tensor_intrin/hexagon.py
```diff
@@ -20,6 +20,47 @@
...
+                    dtype="int32",
+                )
+            )
+            T.evaluate(T.tvm_call_packed("device_api.hexagon.dma_wait", 0, 0, dtype="int32"))
```
adstraw, line 59: Queue = -1.  Comments that Wait(queue, 0) means to wait for the queue to drain which is the sum total of the previous dma_copy.

tests/python/contrib/test_hexagon/test_vtcm_bandwidth.py
```diff
@@ -123,15 +125,18 @@ class TestMatMulVec:
...
+        1024,
+        10 * KB,
         # 20 * KB,
         # 40 * KB,
         # 80 * KB,
         # 160 * KB,
         # 320 * KB,
         640 * KB,
         # MB,
-        # 2 * MB,
```
adstraw, line 134: Did you mean to uncomment this?  Makes the test run longer in CI.

tests/python/contrib/test_hexagon/test_vtcm_bandwidth.py
```diff
@@ -104,8 +106,8 @@ def evaluate(hexagon_session, sch, size):
     )
 
     # These are reduced for CI but number=100 and repeat=10 does a good job of removing noise.
-    number = 1
-    repeat = 1
+    number = 10
+    repeat = 10
```
adstraw, line 110: Did you mean to change this?  Makes the test run longer in CI.

python/tvm/tir/tensor_intrin/hexagon.py
```diff
@@ -163,3 +204,27 @@ def dot_product_32x2_i16i16i32_vdmpy(a: T.handle, b: T.handle, c: T.handle) -> N
 
 VRMPY_u8i8i32_VTCM_INTRIN = "dot_32x4_u8i8i32_vtcm_vrmpy"
 TensorIntrin.register(VRMPY_u8i8i32_VTCM_INTRIN, *generate_dot_product_32x4_u8i8i32("global.vtcm"))
+
+DMA_READ_1_u8 = "dma_read_1_u8"
```
adstraw, line 208: I don't see users for most of these.  Seems like it might be better to delete and allow users to create what is needed based on the test case or schedule?

adstraw (changes requested):

adstraw:
Nice work.  Just wanted to note that we will need to (and have plans to) put our heads together on a unified approach to DMA lowering.  Right now, the async DMA flow stems from TIR annotations added by the InjectSoftwarePipe pass.  This PR suggests a tensorization approach for synchronous DMA.  I feel we could benefit from a unified approach.  I recently deleted an old (TE based) tensorization approach for sync DMA scheduling.  Glad to have this new TIR based version and it may in fact be the path forward for a unified approach to DMA lowering.  Let's put our heads together and get the right design.

tests/python/contrib/test_hexagon/test_vtcm_bandwidth.py
nverke: Added a check if running in CI.

tests/python/contrib/test_hexagon/test_vtcm_bandwidth.py
nverke: Same as above, should have added a check for CI awhile ago.

python/tvm/tir/tensor_intrin/hexagon.py
nverke: Good point, removed ✅

python/tvm/tir/tensor_intrin/hexagon.py
nverke: ✅

python/tvm/tir/tensor_intrin/hexagon.py
nverke: ✅

python/tvm/tir/tensor_intrin/hexagon.py
nverke: ✅

python/tvm/tir/tensor_intrin/hexagon.py
nverke: Going to just add a comment about this for now. These intrins don't currently have any planned use so figure if we find one we can add increased functionality.

python/tvm/tir/tensor_intrin/hexagon.py
nverke: ✅

mehrdadh (approved):

PR #13722, opened 2023-01-07
[CI][Docker][Cortex-M]Update ci_cortexm to Ubuntu 20.04 and update nrfjprog version to 10.18.1

mehrdadh:
This PR updates nrfjprog to latest (10.18.1) to fix issues related to running microTVM tests inside docker. The main issue was that with older version commands like `nrfjprog --com` was not working even with `docker --privileged`.

As a result of this update, we need to update ubuntu version from 18.04 to 20.04.

cc @driazati @gromero

### DIFF
--- a/docker/install/ubuntu2004_install_llvm.sh
+++ b/docker/install/ubuntu2004_install_llvm.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+set -euxo pipefail
+
+wget https://apt.llvm.org/llvm.sh
+chmod +x llvm.sh
+
+./llvm.sh 9
+./llvm.sh 10
+./llvm.sh 11
+./llvm.sh 12
+./llvm.sh 13
--- a/docker/install/ubuntu2004_install_llvm.sh
+++ b/docker/install/ubuntu2004_install_llvm.sh
@@ -18,13 +18,16 @@
 
 set -euxo pipefail
 
-wget https://apt.llvm.org/llvm.sh
-chmod +x llvm.sh
+echo deb http://apt.llvm.org/focal/ llvm-toolchain-focal main\
+    >> /etc/apt/sources.list.d/llvm.list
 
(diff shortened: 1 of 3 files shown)


docker/install/ubuntu2004_install_llvm.sh
```diff
@@ -0,0 +1,30 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+set -euxo pipefail
+
+wget https://apt.llvm.org/llvm.sh
```
Mousius, line 21: Is there any way we can validate this script before executing it? Or can we revert back to the previous way of adding the apt repos? Executing arbitrary scripts from the internet is generally a bad thing 😸

docker/install/ubuntu_install_nrfjprog.sh
```diff
@@ -21,19 +21,24 @@ set -u
...
 set -x
 
+TEMP_PATH_NAME="nrfjprog_tmp_path"
+
+# Install JLink
 NRF_COMMANDLINE_TOOLS_FILE=nRFCommandLineToolsLinuxamd64.tar.gz
-NRF_COMMANDLINE_TOOLS_URL=https://nsscprodmedia.blob.core.windows.net/prod/software-and-other-downloads/desktop-software/nrf-command-line-tools/sw/versions-10-x-x/10-12-1/nrfcommandlinetools10121linuxamd64.tar.gz
-NRF_COMMANDLINE_TOOLS_INSTALLER=nRF-Command-Line-Tools_10_12_1_Linux-amd64.deb
-JLINK_LINUX_INSTALLER=JLink_Linux_V688a_x86_64.deb
+NRF_COMMANDLINE_TOOLS_URL=https://nsscprodmedia.blob.core.windows.net/prod/software-and-other-downloads/desktop-software/nrf-command-line-tools/sw/versions-10-x-x/10-18-1/nrf-command-line-tools-10.18.1_linux-amd64.tar.gz
```
Mousius, line 28: Similarly here, there should be a hash we can check to ensure it's the file we expect?

Mousius (changes requested):

docker/install/ubuntu_install_nrfjprog.sh
mehrdadh: I got this file from NRF website and they don't provide any hash based on my understanding
https://www.nordicsemi.com/Products/Development-tools/nrf-command-line-tools/download
I'm not sure if there's other way to find a hash for this file

docker/install/ubuntu2004_install_llvm.sh
mehrdadh: From their website (https://apt.llvm.org/) I tried this:
```
deb http://apt.llvm.org/focal/ llvm-toolchain-focal main
```
But the issue was that this one is missing LLVM-13. We are not using llvm-13 in cortexm image but it was installing before. 
cc @driazati

mehrdadh: I fixed this, thanks to @driazati pointers.

docker/install/ubuntu_install_nrfjprog.sh
mehrdadh: Add SHA to verify the files

PR #13723, opened 2023-01-07
[microTVM][Zephyr] Fix flash command for nrfjprog

mehrdadh:
This PR adds serial number to flash command (nrfjprog) to fix cases where multiple devices are available.

cc @gromero

### DIFF
--- a/apps/microtvm/zephyr/template_project/microtvm_api_server.py
+++ b/apps/microtvm/zephyr/template_project/microtvm_api_server.py
@@ -265,7 +265,7 @@ def _get_openocd_device_args(serial_number: str = None):
 def _get_nrf_device_args(serial_number: str = None) -> list:
     # iSerial has string type which could mistmatch with
     # the output of `nrfjprog --ids`. Example: 001050007848 vs 1050007848
-    serial_number = str(int(serial_number))
+    serial_number = serial_number.lstrip("0")
 
     nrfjprog_args = ["nrfjprog", "--ids"]
     nrfjprog_ids = subprocess.check_output(nrfjprog_args, encoding="utf-8")
--- a/apps/microtvm/zephyr/template_project/microtvm_api_server.py
+++ b/apps/microtvm/zephyr/template_project/microtvm_api_server.py
@@ -262,7 +262,11 @@ def _get_openocd_device_args(serial_number: str = None):
     return ["--serial", generic_find_serial_port(serial_number)]
 
 
-def _get_nrf_device_args(serial_number: str = None):
+def _get_nrf_device_args(serial_number: str = None) -> list:
+    # iSerial has string type which could mistmatch with
+    # the output of `nrfjprog --ids`. Example: 001050007848 vs 1050007848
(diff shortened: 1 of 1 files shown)


apps/microtvm/zephyr/template_project/microtvm_api_server.py
```diff
@@ -714,23 +716,27 @@ def flash(self, options):
...
+        if zephyr_board.startswith("nrf5340dk") and flash_runner == "nrfjprog":
             recover_args = ["nrfjprog", "--recover"]
             recover_args.extend(_get_nrf_device_args(serial_number))
             check_call(recover_args, cwd=API_SERVER_DIR / "build")
 
         flash_extra_args = []
-        if _get_flash_runner() == "openocd" and serial_number:
-            flash_extra_args = ["--cmd-pre-init", f"""hla_serial {serial_number}"""]
+        if flash_runner == "openocd" and serial_number:
+            flash_extra_args += ["--cmd-pre-init", f"""hla_serial {serial_number}"""]
```
guberti, line 733: nit: 
```suggestion
            flash_extra_args += ["--cmd-pre-init", f"hla_serial {serial_number}"]
```

apps/microtvm/zephyr/template_project/microtvm_api_server.py
```diff
@@ -262,7 +262,11 @@ def _get_openocd_device_args(serial_number: str = None):
     return ["--serial", generic_find_serial_port(serial_number)]
 
 
-def _get_nrf_device_args(serial_number: str = None):
+def _get_nrf_device_args(serial_number: str = None) -> list:
+    # iSerial has string type which could mistmatch with
+    # the output of `nrfjprog --ids`. Example: 001050007848 vs 1050007848
+    serial_number = str(int(serial_number))
```
guberti, line 268: Would it be cleaner to do this? Not sure.
```suggestion
    serial_number = serial_number.lstrip("0")
```

guberti (approved):
LGTM, just two nits.

apps/microtvm/zephyr/template_project/microtvm_api_server.py
mehrdadh: done

apps/microtvm/zephyr/template_project/microtvm_api_server.py
gromero: @mehrdadh I wonder if it would be `f'"hla_serial {serial_number}"'`, because  afaics the triple double-quote in f-string won't expand to an additional double quote, but I think it's necessary one double-quite in the final string for the command as you said  / experimented.

gromero (approved):
LGTM.

apps/microtvm/zephyr/template_project/microtvm_api_server.py
mehrdadh: I will test this again and fix it in a follow up PR if that was the case

PR #13724, opened 2023-01-07
[Arith] Use ConstIntBound to remove negative numerator when lowering

Lunderberg:
Negative numerators to modulo/remainder operations are not supported by the Vulkan API.  While the SPIR-V instructions
[`OpSRem`](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpSRem) and [`OpSMod`](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpSMod) have identical semantics to `tir::Mod` and `tir::FloorMod`, respectively, use of either instruction within Vulkan results in undefined behavior.  From the [Vulkan spec](https://registry.khronos.org/vulkan/specs/1.3/html/chap37.html#spirvenv-op-prec):

> For the OpSRem and OpSMod instructions, if either operand is negative the result is undefined.
>
> Note: While the OpSRem and OpSMod instructions are supported by the Vulkan environment, they require non-negative values and thus do not enable additional functionality beyond what OpUMod provides.

This issue was first noticed in https://github.com/apache/tvm/pull/13530, where use of integer arithmetic resulted in negative numerators.  This hadn't caused issues previously, because most use of div/mod use a denominator that is a power of two.  In these cases, `tir.LowerIntrin` implements floordiv and floormod using only bitwise operations.  When the denominator isn't a power of two, both `tir::FloorDiv` and `tir::FloorMod` are implemented in terms of `tir::Mod`, which triggers the undefined behavior for negative numerators.

This commit alters the lowering of FloorDiv/FloorMod to TruncDiv/TruncMod, in cases where the denominator is positive, the numerator is sometimes negative, and the range of the numerator is known.  In these cases, the FloorDiv/FloorMod is now implemented by offsetting the numerator such that it is always positive.

### DIFF
--- a/src/tir/transforms/lower_intrin.cc
+++ b/src/tir/transforms/lower_intrin.cc
@@ -112,20 +112,31 @@ class IntrinInjecter : public tvm::arith::IRMutatorWithAnalyzer {
       // Common path, positive divisor
       if (analyzer_->CanProveGreaterEqual(op->a, 0) || analyzer_->CanProveGreaterEqual(e, 0)) {
         return truncdiv(op->a, op->b);
-      } else {
-        DLOG(INFO) << "LowerFloorDiv: Cannot decide the sign of divident";
-        PrimExpr rdiv = truncdiv(op->a, op->b);
-        PrimExpr rmod = truncmod(op->a, op->b);
-        // condition on b >= 0.
-        // truncmod(a, b) < 0 will implies ceildiv,
-        // So we need to correct these cases.
-        if ((dtype == DataType::Int(32) || dtype == DataType::Int(64)) && support_bitwise_op_) {
-          // equivalent to rdiv + (rmod >= 0 ? 0: -1);
-          return rdiv + (rmod >> make_const(dtype, dtype.bits() - 1));
-        } else {
-          return tir::Select(rmod >= 0, rdiv, rdiv - make_const(dtype, 1));
-        }
       }
+
+      // If the numerator's lower bound is known, express the floordiv
+      // in terms of truncdiv using only positive operands.
+      arith::ConstIntBound const_int_bound = analyzer_->const_int_bound(op->a);
+      if (const_int_bound->min_value != arith::ConstIntBound::kNegInf &&
+          const_int_bound->min_value < 0) {
+        IntImm min(op->a->dtype, const_int_bound->min_value);
+        PrimExpr ceildiv = truncdiv(-min + (op->b - 1), op->b);
+        return truncdiv(op->a + op->b * ceildiv, op->b) - ceildiv;
+      }
+
+      DLOG(INFO) << "LowerFloorDiv: Cannot decide the sign of divident";
+      PrimExpr rdiv = truncdiv(op->a, op->b);
+      PrimExpr rmod = truncmod(op->a, op->b);
+      // condition on b >= 0.
+      // truncmod(a, b) < 0 will implies ceildiv,
+      // So we need to correct these cases.
+      if ((dtype == DataType::Int(32) || dtype == DataType::Int(64)) && support_bitwise_op_) {
+        // equivalent to rdiv + (rmod >= 0 ? 0: -1);
+        return rdiv + (rmod >> make_const(dtype, dtype.bits() - 1));
+      } else {
+        return tir::Select(rmod >= 0, rdiv, rdiv - make_const(dtype, 1));
+      }
+
     } else {
       if (dtype.is_float()) {
         // floor(a / b)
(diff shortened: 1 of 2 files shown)


src/tir/transforms/lower_intrin.cc
```diff
@@ -112,20 +113,33 @@ class IntrinInjecter : public tvm::arith::IRMutatorWithAnalyzer {
...
+      // If the numerator's lower bound is known, express the floordiv
+      // in terms of truncdiv using only positive operands.
+      arith::ConstIntBound const_int_bound = analyzer_->const_int_bound(op->a);
+      if (const_int_bound->min_value != arith::ConstIntBound::kNegInf &&
+          const_int_bound->min_value < 0 &&
+          const_int_bound->min_value > -(1LL << (op->a->dtype.bits() - 1))) {
+        IntImm min(op->a->dtype, const_int_bound->min_value);
+        PrimExpr ceildiv = truncdiv((op->b - 1) - min, op->b);
+        PrimExpr offset_numerator = analyzer_->Simplify(op->a + op->b * ceildiv);
+        return truncdiv(offset_numerator, op->b) - ceildiv;
```
wrongtest-intellif, line 127: IIUC, the rationale is:
```
a // b =>
(a + -min - -min) // b => 
(a + (x*b-y) - (x*b-y)) // b =>  ( where x is ceildiv(-min, b), x >= 0, y >= 0 )
(a + x*b) // b - x =>
(a + x*b) / b - x ( since a + x*b >= a + -min >= 0 )
```
Right?

src/tir/transforms/lower_intrin.cc
```diff
@@ -165,21 +179,34 @@ class IntrinInjecter : public tvm::arith::IRMutatorWithAnalyzer {
...
+      // If the numerator's lower bound is known, express the floormod
+      // in terms of truncmod using only positive operands.
+      arith::ConstIntBound const_int_bound = analyzer_->const_int_bound(op->a);
+      if (const_int_bound->min_value != arith::ConstIntBound::kNegInf &&
+          const_int_bound->min_value < 0 &&
+          const_int_bound->min_value > -(1LL << (op->a->dtype.bits() - 1))) {
+        IntImm min(op->a->dtype, const_int_bound->min_value);
+        PrimExpr ceildiv = truncdiv(-min + (op->b - 1), op->b);
+        PrimExpr offset_numerator = analyzer_->Simplify(op->a + op->b * ceildiv);
+        return truncmod(offset_numerator, op->b);
```
wrongtest-intellif, line 193:
```
floormod(a, b) =>
floormod(a + -min - -min, b) => 
floormod(a + (x*b-y) - (x*b-y)), b) =>  ( where x is ceildiv(-min, b), x >= 0, y >= 0 )
floormod(a + x*b, b) =>
(a + x*b) % b  ( since a + x*b >= a + -min >= 0 )
```
Right?

wrongtest-intellif (approved):

src/tir/transforms/lower_intrin.cc
```diff
@@ -112,20 +113,33 @@ class IntrinInjecter : public tvm::arith::IRMutatorWithAnalyzer {
...
       if (analyzer_->CanProveGreaterEqual(op->a, 0) || analyzer_->CanProveGreaterEqual(e, 0)) {
         return truncdiv(op->a, op->b);
+      }
+
+      // If the numerator's lower bound is known, express the floordiv
+      // in terms of truncdiv using only positive operands.
+      arith::ConstIntBound const_int_bound = analyzer_->const_int_bound(op->a);
+      if (const_int_bound->min_value != arith::ConstIntBound::kNegInf &&
+          const_int_bound->min_value < 0 &&
+          const_int_bound->min_value > -(1LL << (op->a->dtype.bits() - 1))) {
```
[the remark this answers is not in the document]
Lunderberg: Thank you, and op.h's `min_value` is exactly what I had been looking for.  I had checked whether `ConstIntBound::Everything(dtype)` was public, but wasn't aware of `min_value`.  PR updated to use `min_value` instead of doing the calculation here.

src/tir/transforms/lower_intrin.cc
Lunderberg: Yup, that is the rationale.  I've added the derivation in a comment, which I probably should have done from the start.

src/tir/transforms/lower_intrin.cc
Lunderberg: Yup, that derivation is correct, and I've added a comment here as well.

PR #13725, opened 2023-01-08
[HotFix][docs] Use correct Colab button URL

guberti:
PR #13627 added Google Colab support to the TVM tutorials, allowing them to be opened interactively with one click. This mostly works - [try it yourself](https://tvm.apache.org/docs/how_to/extend_tvm/low_level_custom_pass.html)!

However, the button to open the tutorial in Google Colab does not render, as #13627 used the wrong URL:

image

The file is stored in the repo tlc-pack/web-data, not apache/web-data. This PR fixes the bug and makes the button render correctly.

cc @mehrdadh

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -153,7 +153,7 @@ COLAB_URL_BASE = "https://colab.research.google.com/github"
 IPYTHON_GITHUB_BASE = "apache/tvm-site/blob/asf-site/docs/_downloads/"
 
 # The SVG image of the "Open in Colab" button.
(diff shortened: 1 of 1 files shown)


guberti:
Before merging this, I need to confirm this change will trigger the docs to be re-built and re-deployed. I _believe_ that happens after every PR - if this is not the case, let me know.

driazati (approved):
Confirmed it works on https://pr-docs.tlcpack.ai/PR-13725/1/docs/tutorial/install.html#sphx-glr-tutorial-install-py, the deploy should happen automatically once this is merged

PR #9553, opened 2021-11-23
[PROFILING] Add ability to profile a single function_profiling

tkonolige:
Add a new function `tvm.runtime.profiling.profile_function` which collects performance metrics for a single function in an IRModule. For example, collecting performance counters using `PAPIMetricCollector`. This is helpful for optimizing kernels and schedules for a single operator.

PR #13726, opened 2023-01-08
[Profiler] Allow user to flush L2 cache in `time_evalutor` function for profiling CUDA kernels

yzh119:
# Motivation
Currently, our default profiler (`time_evaluator`) does not flush the L2 cache per execution, this might lead to incorrect time measurement because the input data last run might reside in L2 cache and reduce the data fetching time in the next run. Both [Triton](https://github.com/openai/triton/blob/ff399fbc2059a6f35cb93534dc29398f7b82dbc7/python/triton/testing.py#L156-L181) and [nvbench](https://github.com/NVIDIA/nvbench/blob/1a13a2e724b8aa8aee27649ac6878babb63862a6/nvbench/detail/measure_cold.cuh#L123) consider this effect thus reporting more accurate measurements.

# Solution
`time_evalutor` has an argument `f_preproc` where user can specify a pre-processing function per execution of the kernel being evaluated. Currently, TVM supports `cache_flush_cpu_non_first_arg` which flushes CPU cache. But similar functionality for GPU is missing.

This PR completely borrows the design of nvbench's [l2flush](https://github.com/NVIDIA/nvbench/blob/1a13a2e724b8aa8aee27649ac6878babb63862a6/nvbench/detail/l2flush.cuh) struct and allow the user to specify `"l2_cache_flush_cuda"` as a preprocessing function which flushes NVIDIA GPU's L2 cache. `l2_cache_flush_cuda` is not a default value so existing program's behavior would not be influenced.

<del>
Note that this PR also changes the location where `f_preproc` being triggered: previously `f_preproc` is triggered per repeat but that doesn't sound correct to me because most users specify `repeat=1` and `f_preproc` need to be triggered once per run.</del>

cc @masahi @tkonolige @junrushao @spectrometerHBH @tqchen


src/runtime/profiling.cc
```diff
@@ -894,15 +891,21 @@ PackedFunc WrapTimeEvaluator(PackedFunc pf, Device dev, int number, int repeat,
               std::max((min_repeat_ms / (duration_ms / number) + 1), number * golden_ratio));
         }
 
+        int64_t accum_t_nanos = 0;
         // start timing
-        Timer t = Timer::Start(dev);
         for (int j = 0; j < number; ++j) {
+          // call preprocessing function
+          if (f_preproc != nullptr) {
+            f_preproc.CallPacked(args, &temp);
```
Icemist, line 899: From experience that this change needs to be reversed.
The **repeat** and **number** parameters are designed for different scenarios. For your scenario, you need to increase the **repeat** and use **number=1**.
Scenario with increasing **number** (or using **min_repeat_ms**) is used when it is problematic to measure single runs - "_The timer is specifically in the outer loop so that we can time functions with a very small runtime or use less precise timers._"

yzh119: Thanks for the clarification, I'll fix it soon.
Unfortunately it looks like there are a lot of misuse of `number` in the codebase, maybe we should create another PR fixing them.

Icemist: > Thanks for the clarification, I'll fix it soon. Unfortunately it looks like there are a lot of misuse of `number` in the codebase, maybe we should create another PR fixing them.
I don't think it matters that much if we are satisfied with the average number and don't need to call f_preproc. So don't be too eager about it.

yzh119: I have reverted `profiling.cc` and updated unittest accordingly.

yzh119: Another concern is whether we should have a `warmup` argument for `time_evaluator` function, I think it's common in all frameworks' profiling tools.

Icemist: > Another concern is whether we should have a `warmup` argument for `time_evaluator` function, I think it's common in all frameworks' profiling tools. 

There is a pre-start call of the function, before starting the logic of measurements.  Do you think that one is not enough? Or warm it up with something else?
```
// skip first time call, to activate lazy compilation components.
pf.CallPacked(args, &temp);
```

yzh119: PyTorch profiler has a `warmup` argument: https://pytorch.org/docs/stable/profiler.html#torch.profiler.schedule, where users can specify a custom number of steps to warm up.

yzh119: I also noticed `ProfileFunction` defined in `profiling.cc` which has a `warmup_iters` argument, which somehow makes me confused: in what case should user select `ProfileFunction` instead of `WrapTimeEvaluator`?

tests/python/unittest/test_evaluator_flush_l2_cache.py
```diff
@@ -0,0 +1,63 @@
...
+    a = tvm.nd.array(np.random.rand(128, 128).astype("float32"), device=dev)
+    b = tvm.nd.array(np.random.rand(128, 128).astype("float32"), device=dev)
+    c = tvm.nd.array(np.zeros((128, 128)).astype("float32"), device=dev)
+    args = [a, b, c]
+    print("Evaluator (w/o L2 flush):\t{:.5f}ms".format(evaluator_no_flush(*args).mean * 1000))
+
+    evaluator_with_flush = f.time_evaluator(
+        f.entry_name, dev, repeat=1000, number=1, f_preproc="l2_cache_flush_cuda"
+    )
+    print("Evaluator (w/ L2 flush):\t{:.5f}ms".format(evaluator_with_flush(*args).mean * 1000))
```
echuraev, line 59: Are there any criteria where the test should be failed?

src/runtime/cuda/l2_cache_flush.cc
```diff
@@ -0,0 +1,81 @@
...
+namespace tvm {
+
+namespace runtime {
+
+class L2Flush {
+ public:
+  L2Flush() : initialized_(false), l2_size_(0), l2_buffer_(nullptr) {}
+
+  ~L2Flush() {
+    if (l2_size_ > 0) {
```
echuraev, line 39: Do we really need `l2_size_`? Probably we just can check that `l2_buffer_` in not a `nullptr`.

tests/python/unittest/test_evaluator_flush_l2_cache.py
yzh119: There isn't but I'd love to have one, any ideas on that?

src/runtime/cuda/l2_cache_flush.cc
yzh119: Yes, we need it, every time we call the flush function, we need `l2_size_` for `cudaMemsetAsync`.

tests/python/unittest/test_evaluator_flush_l2_cache.py
Icemist: > Are there any criteria where the test should be failed?

This test tests for the presence of **l2_cache_flush_cuda**. It's normal integration test :)

Not sure if we need to do the first run: evaluator_no_flush.
In some tests, the base scenario is hidden with an **if False** block.
For example you can look at do_tune block in  _tests/python/contrib/test_hexagon/metaschedule_e2e/test_resnet50_int8.py::test_packed_8x8x32_resnet50_
Or use **@pytest.mark.parametrize** for **f_preproc**

src/runtime/profiling.cc
Icemist: Maybe warmup has a place, also we already have skip_first=1 similar to pytorch. 
But this would require changes not only in the C++ part, but also in the C and webasm part. 

For ProfileFunction  plz look original PR with https://github.com/apache/tvm/pull/9553 there are a couple of words about the motivation for adding.

yzh119: Got it, I prefer supporting the `warmup` argument in another PR. Thank you for the discussion!

tests/python/unittest/test_evaluator_flush_l2_cache.py
yzh119: `pytest.mark.parametrize` sounds good to me

src/runtime/cuda/l2_cache_flush.cc
```diff
@@ -0,0 +1,81 @@
...
+      CUDA_CALL(cudaFree(l2_buffer_));
+    }
+  }
+
+  void Flush() {
+    if (!initialized_) {
```
Icemist, line 45: we expect that after cudaMalloc the variable l2_buffer_ will not be nullptr? then can we use this instead of initialized_?

src/runtime/cuda/l2_cache_flush.cc
```diff
@@ -0,0 +1,81 @@
...
+      // initialize l2_buffer_ and l2_size_
+      initialized_ = true;
+      int device_id;
+      CUDA_CALL(cudaGetDevice(&device_id));
+      CUDA_CALL(cudaDeviceGetAttribute(&l2_size_, cudaDevAttrL2CacheSize, device_id));
+      if (l2_size_ > 0) {
+        void* buffer = l2_buffer_;
```
Icemist, line 52: I'm not good at CUDA, but can't we do without the buffer temporary variable? 
`CUDA_CALL(cudaMalloc((void**)&l2_buffer_, l2_size_))`

Icemist (approved):
LGTM, except for a couple of minor non-functional issues.

src/runtime/cuda/l2_cache_flush.cc
yzh119: 1. Yes
2. `initialized` was used to determine whether it was the first run. We don't want to call this malloc function if `Flush` was never triggered so we don't put them under initialization function.

src/runtime/cuda/l2_cache_flush.cc
yzh119: I suppose they are equivalent.

yzh119: I adopted your suggestion because it looks better:)

src/runtime/cuda/l2_cache_flush.cc
```diff
@@ -0,0 +1,79 @@
+/*
```
tkonolige, line 1: We don't like to include code from other projects in TVM unless it lives in the `3rdparty` directory. As this code is basically the same as https://github.com/NVIDIA/nvbench/blob/1a13a2e724b8aa8aee27649ac6878babb63862a6/nvbench/detail/l2flush.cuh, it should be moved into `3rdparty/nvbench` and the license header at the top should be changed to the one in the nvbench repository.

@areusch is this acceptable?

tkonolige (changes requested):
Thanks @yzh119! This is a great addition to benchmarking. Maybe we should consider flushing the cache by default instead of requiring the user to specify a preprocessor. Maybe something like a cold-start mode.

There's one change I need to you make around moving this code to third party as it is almost directly from nvbench.

src/runtime/cuda/l2_cache_flush.cc
echuraev: I think that @Icemist wanted to propose removing `initialized_` variable and check `l2_buffer_` instead:
```c++
  void Flush() {
    if (l2_buffer_ == nullptr) {
      // ...
```

With the same idea, I have asked my question about `l2_size_` variable in the destructor.

src/runtime/cuda/l2_cache_flush.cc
echuraev: My bad. I have missed it.

tests/python/unittest/test_evaluator_flush_l2_cache.py
echuraev: > This test tests for the presence of **l2_cache_flush_cuda**. It's normal integration test :)

Yes, but I thought if we can check that `l2_cache_flush_cuda` works. Previously, it was two runs of `time_evaluator`. And I thought if we can compare the execution time, then we probably can say if `l2_cache_flush_cuda` works or not.

Probably the better test for such functionality is to check L2 cache after `Flush` function. But probably such test should be implemented by using gtest, and I'm not sure that such infrastructure exists for cuda. Anyway, I don't think that we should spend a lot of time on implementing such test in this PR and I don't want to block this PR only because of such test. The existing test must be enough.

src/runtime/cuda/l2_cache_flush.cc
tqchen: @yzh119 Let us move to `3rdparty/nvbench` then we can include it from there

tqchen: Please see an example of compiler-rt header 
- original file(with fixes) https://github.com/apache/tvm/blob/main/3rdparty/compiler-rt/builtin_fp16.h
- mention license here https://github.com/apache/tvm/blob/main/LICENSE#L235
- add licenses file to this folder https://github.com/apache/tvm/blob/main/licenses/LICENSE.builtin_fp16.txt

yzh119: Thank @tkonolige @tqchen for the comments, I'll refactor the code to include nvbench.

src/runtime/cuda/l2_cache_flush.cc
yzh119: Per [CUDA runtime documentation](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html): 
>[cudaDevAttrL2CacheSize](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1gg49e2f8c2c0bd6fe264f2fc970912e5cd1b0342682d15910022ba3f383a851ad7): Size of L2 cache in bytes. 0 if the device doesn't have L2 cache.

`l2_size_` would be set to 0 if the device does not have `l2`, in which case `l2_buffer_` is still a nullptr I suppose.

echuraev: Thank you for the clarification! Yes, in this case I think `initialized_` is the better solution.

src/runtime/cuda/l2_cache_flush.cc
yzh119: @tkonolige @tqchen  I have refactored the code and added license files correspondingly, let me know if they look good to you:)

tqchen (approved):

tkonolige (approved):

PR #13729, opened 2023-01-09
[TE][PrimFunc] Fix create primfunc from te extern with explicit buffer load

wrongtest-intellif:
Fix the issue when `te.extern` use explicit buffer load, create_primfunc would return a func with aliased buffer decls before. Also fix some minor typo bug in `specialize`.


junrushao:
CC: @Hzfengsy

Hzfengsy (approved):
LGTM

PR #11468, opened 2022-05-26
[microNPU] Optimize separate padding operation for conv2d

lhutton1:
Optimizes a case where padding appears as a separate nn.pad operation followed by a qnn.conv2d. If possible, the nn.pad will be partitioned and offloaded together with the qnn.conv2d operation, as opposed to separately. As a fallback, both operations will be considered separately.

cc @Mousius @NicolaLancellotti @ekalda @manupa-arm

PR #13732, opened 2023-01-09
[microNPU] Add support for TFLite PAD

Alexey-Yazev:
A separate nn.pad relay operator is legalized to an Ethos-U depthwise_conv2d operator. For ethosu_depthwise_conv2d the hardware only supports padding up to 31, 31, 32, 32, 32, so the pad size for legalization on the NPU is within these limits.

### DIFF
--- a/python/tvm/relay/backend/contrib/ethosu/legalize.py
+++ b/python/tvm/relay/backend/contrib/ethosu/legalize.py
@@ -1355,6 +1355,62 @@ class FullyConnectedRewriter(DFPatternCallback):
         return ethosu_fc
 
 
+class PadRewriter(DFPatternCallback):
+    """Convert ethos-u.pad2d composite function to ethosu_depthwise_conv2d
+    operator"""
+
+    def __init__(self):
+        super().__init__(require_type=True)
+        self.pattern = (
+            wildcard().has_attr({"Composite": ethosu_patterns.PadParams.composite_name})
+        )(wildcard())
+
+    def callback(
+        self, pre: tvm.relay.Expr, post: tvm.relay.Expr, node_map: tvm.ir.container.Map
+    ) -> tvm.relay.Expr:
+        params = ethosu_patterns.PadParams(post.op.body)
+        params.ifm.tensor = post.args[0]
+        channels_map = {
+            "NHWC": 3,
+        }
+        w_h, w_w = (1, 1)
+        # OHWI format for the ethosu_depthwise_conv2d kernel weights
+        weight_shape = (params.ifm.shape[-1], w_h, w_w, 1)
+        weights = relay.const(np.full(weight_shape, 1), params.ifm.dtype)
+        scale_bias = vela_api.pack_biases(
+            biases=np.zeros(params.ifm.shape[-1]),
+            ifm_scale=params.ifm.q_params.scale_f32,
+            ifm_dtype=np.dtype(params.ifm.dtype),
+            weight_scales=np.array(1.0, dtype=np.float32),
+            ofm_scale=params.ofm.q_params.scale_f32,
+            is_activation_tanh_or_sigmoid=False,
+        )
+
+        return ethosu_ops.ethosu_depthwise_conv2d(
+            ifm=post.args[0],
+            weight=weights,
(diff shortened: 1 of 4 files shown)


python/tvm/relay/backend/contrib/ethosu/legalize.py
```diff
@@ -1355,6 +1355,62 @@ def callback(self, pre, post, node_map):
...
+        )(wildcard())
+
+    def callback(
+        self, pre: tvm.relay.Expr, post: tvm.relay.Expr, node_map: tvm.ir.container.Map
+    ) -> tvm.relay.Expr:
+        params = ethosu_patterns.PadParams(post.op.body)
+        params.ifm.tensor = post.args[0]
+        channels_map = {
+            "NHWC": 3,
+        }
```
ekalda, line 1375: IIRC this one entry `channels_map` is a historic relic that can go and so simplify the code little bit, but as it is present in other operators, it's probably a clean up task for some other time.

ekalda (approved):
Thanks @Alexey-Yazev , I think it's good to go, but I'll leave it open for little longer in case anybody else wants to have a look.

lhutton1 (approved):
Thanks @Alexey-Yazev, LGTM! After looking at how padding is lowered in [Vela](https://git.mlplatform.org/ml/ethos-u/ethos-u-vela.git/about/) I think there might be a couple more opportunities for optimization, although it seems out of scope for this PR. Just a couple of things to consider in the future:
- In some cases its possible to fuse a `nn.pad` with the following operation. As an example we currently fuse `nn.pad`->`qnn.conv2d` (https://github.com/apache/tvm/pull/11468), however, it seems a similar approach is also possible for average pooling (see: https://git.mlplatform.org/ml/ethos-u/ethos-u-vela.git/tree/ethosu/vela/tflite_graph_optimiser.py#n1413)
- With the current implementation `nn.pad` does not get offloaded if the provided padding exceeds [31, 31, 32, 32]. If these dimensions are exceeded, we might be able to use multiple average pooling operations similar to https://git.mlplatform.org/ml/ethos-u/ethos-u-vela.git/tree/ethosu/vela/tflite_graph_optimiser.py#n1500

lhutton1:
Thanks @Alexey-Yazev, @ekalda!

lhutton1:
Oops I forgot to ask, would it be possible to add a legalization test under: `tests/python/contrib/test_ethosu/test_legalize.py` in a follow-up? Apologies for missing this before!

arina-grovety:
Hello @lhutton1, thanks for the review!
> * With the current implementation `nn.pad` does not get offloaded if the provided padding exceeds [31, 31, 32, 32]. If these dimensions are exceeded, we might be able to use multiple average pooling operations similar to https://git.mlplatform.org/ml/ethos-u/ethos-u-vela.git/tree/ethosu/vela/tflite_graph_optimiser.py#n1500

Yes, this was the first option that we tried to implement. But in the Vela implementation, this is done by **"copies IFM to the right place inside the OFM"** using **write_offset** attribute of the created AvgPool operation.

In the TVM, VelaAPI operations are derived from the NpuOperation class, which does not have a write_offset attribute, so we cannot replicate Vela convert_pad() function. 

We tried to implement PAD legalization using the Concatenate operation but encountered an error. Seems like Cascader must be turned off for Concatenate to work. For example, Cascader is disabled in test_tflite_concat() (if Cascader is enabled, there is the same error as we have with the Concatenate)

So far, the most feasible option seems to use several depthwise_conv2d operators, if padding exceeds [31, 31, 32, 32].

But of course, I do not have all the knowledge about this, maybe there are other options?

Alexey-Yazev:
> Oops I forgot to ask, would it be possible to add a legalization test under: tests/python/contrib/test_ethosu/test_legalize.py in a follow-up? Apologies for missing this before!

Test was added in [PR](https://github.com/apache/tvm/pull/13750)

lhutton1:
Thanks @arina-grovety for the explanation, just following up on some of the questions...

I suspect this is a case of needing to expose this functionality from within Vela, I'll see if we can make this happen for a future Vela release.

The concatenate error does indeed sound like a separate issue in itself. It might be worth investigating the reason for that at some point.

PR #13733, opened 2023-01-09
[TOPI][OP] Support grouped conv2d_NCHWc

liangW-intellif:
This PR supports conv2d_NCHWc with groups.
cc @masahi @wrongtest-intellif

### DIFF
--- a/python/tvm/topi/nn/conv2d.py
+++ b/python/tvm/topi/nn/conv2d.py
@@ -388,9 +388,11 @@ def conv2d_NCHWc(data, kernel, stride, padding, dilation, layout, out_layout, ou
     n, ic_chunk, ih, iw, ic_bn = get_const_tuple(data.shape)
     in_channel = ic_chunk * ic_bn
     target = tvm.target.Target.current(allow_none=False)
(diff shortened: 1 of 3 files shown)


python/tvm/topi/nn/conv2d.py
```diff
@@ -424,17 +424,15 @@ def conv2d_NCHWc(data, kernel, stride, padding, dilation, layout, out_layout, ou
 
     return te.compute(
         oshape,
-        lambda n, oc_chunk, oh, ow, oc_block: te.sum(
+        lambda n, occ, oh, ow, oc_block: te.sum(
```
wrongtest-intellif, line 427: Could we preserve a group=1 code path like `conv2d_NCHWc_int8` below?

wrongtest-intellif (changes requested):

python/tvm/topi/nn/conv2d.py
liangW-intellif: Done, please take a look again.

wrongtest-intellif (approved):
LGTM

masahi (approved):

PR #13734, opened 2023-01-09
[Hexagon][QNN] Improve performance wo QNN canonicalization

ibsidorenko:
This commit improves performance of different models tuned with MetaScheduler for Hexagon target and without QNN canonicalization.

Benchmarking of several models on Snapdragon 8gen1 and tuned with MS:

Model               | QNN canon enabled, ms | QNN canon disabled, ms |   speedup   |
-----------------|-----------------------------|----------------------------|--------------|
ResNet, int8      |                   50                   |                     48                |     +4.2%     |
Inception, int8  |                  103                  |                    106               |      -2.8%     |
SRGAN, int8      |                 348                   |                    431               |     -19.3%    |

**What was done:**

1. Added 2 new passes: `QnnLegalize `and `QnnCanonicalize`. But this is just wrappers for Legalize("FTVMQnnLegalize") and Legalize("FTVMQnnCanonicalize").

2. Added ability to disable inline for specific blocks in MetaSchedule AutoInline rule. For example, it can be done through the `T.block_attr({"meta_schedule.inline_rule": "disable"}).`

3. Implemented compute, alter op and legalization functions for `qnn.conv2d` operation (for Hexagon target).

### DIFF
--- a/python/tvm/relay/qnn/op/legalizations.py
+++ b/python/tvm/relay/qnn/op/legalizations.py
@@ -405,6 +405,11 @@ def is_fast_int8_on_intel():
     return target_has_sse42(target.mcpu)
 
 
+# Helper function to align up given value.
+def helper_align_up(value, aligner):
+    return ((value + aligner) // aligner) * aligner
+
+
 ########################
 # ARM CPU legalizations.
 ########################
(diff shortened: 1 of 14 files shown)


include/tvm/meta_schedule/schedule_rule.h
```diff
@@ -124,7 +125,8 @@ class ScheduleRule : public runtime::ObjectRef {
                                          bool disallow_if_then_else,  //
                                          bool require_injective,      //
                                          bool require_ordered,        //
-                                         Optional<Array<String>> disallow_op);
+                                         Optional<Array<String>> disallow_op,
+                                         Optional<Array<String>> disallow_block = {});
```
junrushao, line 129: It is less stable to specify blocks by their names, instead, please consider using other approaches to suppress inline, for example, `T.block_attr`

src/meta_schedule/trace_apply.cc
```diff
@@ -175,16 +175,30 @@ std::vector<BlockRV> ApplyAnchorTrace(Schedule sch, Trace anchor_trace) {
...
-      // new_outputs.size(). We workaround this problem by assuming that the prefix of the "new"
-      // outputs matches with the "old" outputs, and truncating the new outputs accordingly.
-      ICHECK(inst->outputs.size() <= outputs.size());
-      TranslateAddOutputRVs(
-          inst->outputs, Array<ObjectRef>(outputs.begin(), outputs.begin() + inst->outputs.size()),
-          &rv_map);
+      if (inst->outputs.size() == outputs.size()) {
+        TranslateAddOutputRVs(inst->outputs, outputs, &rv_map);
+      } else if (inst->outputs.size() < outputs.size()) {
+        // We want to allow a trace generated for a single conv2d block to be applied to
```
junrushao, line 181: MetaSchedule requires structural stability - that is being said, it does not support the case where the number of children varies. The workaround is probably too hacky to get it work...Shall we consider other alternatives?

junrushao (changes requested):

include/tvm/meta_schedule/schedule_rule.h
ibsidorenko: hmm... Interesting idea, I didn't think of it before. Need to check

ibsidorenko: Done. now it can be disabled through the `T.block_attr({"meta_schedule.inline_rule": "disable"})`

src/meta_schedule/trace_apply.cc
ibsidorenko: I have reverted all my changes here. But anyway this "hack" was introduced before my changes.

ibsidorenko:
cc @junrushao @masahi

src/meta_schedule/trace_apply.cc
masahi: Yeah, this is necessary since `get_child_blocks`, `get_consumers` etc can return different number of results depending on the post ops, even if the anchor op is the same.

python/tvm/relay/qnn/op/legalizations.py
```diff
@@ -483,3 +488,68 @@ def _qnn_dense_legalize_cuda(attrs, inputs, types):
...
+IN_CHANNEL_VECTOR_LENGTH = 4
+OUT_CHANNEL_VECTOR_LENGTH = 32
+
+
+@qnn_conv2d_legalize.register("hexagon")
+def _qnn_conv2d_legalize_hexagon(attrs, inputs, types):
+    """Legalize qnn.conv2d op for vrmpy tensorization.
+
+    If the inputs are signed or unsigned int8 and data/kernel layouts are NCHW/OIHW, then the input
+    and output channels are padded to be a multiple of 4 and 32 respectively.
```
masahi, line 506: Is it ok not to legalize dtype as well?

masahi (approved):
Interesting to see that we can directly tune QNN models and performance is reasonable or even better than canonicalization-enabled.

python/tvm/relay/qnn/op/legalizations.py
ibsidorenko: agree... dtype legalization will be the next step here.

PR #13736, opened 2023-01-09
[CI][Docker][Cortex-M]Update scripts to update ci_cortexm to Ubuntu 20.04

mehrdadh:
This PR makes changed to update ci_cortexm to 20.04. Here is a list of changes:
- Updates nrfjprog to latest (10.18.1) to fix issues related to running microTVM tests inside docker. The main issue was that with older version commands like nrfjprog --com was not working even with docker --privileged.
- Update python install script to support python 3.7 in 20.04 -> we keep using python 3.7 in ubuntu 20.04 until its EOL
- Add script to install LLVM in ubuntu 20.04

cc @driazati @gromero

### DIFF
--- a/docker/Dockerfile.ci_cortexm
+++ b/docker/Dockerfile.ci_cortexm
@@ -38,7 +38,7 @@ RUN bash /install/ubuntu_install_googletest.sh
 ENV TVM_VENV /venv/apache-tvm-py3.7
 COPY python/bootstrap/lockfiles /install/python/bootstrap/lockfiles
 COPY install/ubuntu_install_python.sh /install/ubuntu_install_python.sh
-RUN bash /install/ubuntu_install_python.sh
+RUN bash /install/ubuntu_install_python.sh 3.7
 ENV PATH ${TVM_VENV}/bin:$PATH
 ENV PYTHONNOUSERSITE 1  # Disable .local directory from affecting CI.
 
--- a/docker/Dockerfile.ci_cortexm
+++ b/docker/Dockerfile.ci_cortexm
@@ -17,7 +17,7 @@
 
 # CI docker CPU env
 # tag: v0.62
-FROM ubuntu:18.04
+FROM ubuntu:20.04
 
 COPY utils/apt-install-and-clear.sh /usr/local/bin/apt-install-and-clear
 
@@ -45,8 +45,8 @@ ENV PYTHONNOUSERSITE 1  # Disable .local directory from affecting CI.
 COPY install/ubuntu_install_python_package.sh /install/ubuntu_install_python_package.sh
 RUN bash /install/ubuntu_install_python_package.sh
 
-COPY install/ubuntu1804_install_llvm.sh /install/ubuntu1804_install_llvm.sh
-RUN bash /install/ubuntu1804_install_llvm.sh
+COPY install/ubuntu2004_install_llvm.sh /install/ubuntu2004_install_llvm.sh
+RUN bash /install/ubuntu2004_install_llvm.sh
 
 # Rust env (build early; takes a while)
 COPY install/ubuntu_install_rust.sh /install/ubuntu_install_rust.sh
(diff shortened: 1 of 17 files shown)


mehrdadh:
@tvm-bot rerun

mehrdadh:
cc @Mousius

mehrdadh:
@tvm-bot rerun

Mousius (changes requested):
Looks a lot more secure @mehrdadh 😸 one last change and then I think it's good to go.

docker/Dockerfile.ci_cortexm
```diff
@@ -45,8 +45,8 @@ ENV PYTHONNOUSERSITE 1  # Disable .local directory from affecting CI.
 COPY install/ubuntu_install_python_package.sh /install/ubuntu_install_python_package.sh
 RUN bash /install/ubuntu_install_python_package.sh
 
-COPY install/ubuntu1804_install_llvm.sh /install/ubuntu1804_install_llvm.sh
-RUN bash /install/ubuntu1804_install_llvm.sh
+COPY install/ubuntu2004_install_llvm.sh /install/ubuntu2004_install_llvm.sh
+RUN bash /install/ubuntu2004_install_llvm.sh
```
ashutosh-arm, line 49: @mehrdadh, thanks for this upgrade. Since Ubuntu20.04 has default installation of `python3.8` and `python3.7` EOL is June 2023 as per https://peps.python.org/pep-0537/#lifespan. Since I have not seen any other discussions, my assumption is that we still do wish to support older version until its EOL. We most likely would require old installation from https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa.

mehrdadh: @ashutosh-arm do we need python3.7 for cortexm CI image? If so, we should add it in a separate PR.
This PR doesn't update the image yet.

docker/install/ubuntu2004_install_llvm.sh
```diff
@@ -0,0 +1,35 @@
...
+
+set -euxo pipefail
+
+echo deb http://apt.llvm.org/focal/ llvm-toolchain-focal main\
+    >> /etc/apt/sources.list.d/llvm.list
+
+echo deb http://apt.llvm.org/focal/ llvm-toolchain-focal-13 main\
+    >> /etc/apt/sources.list.d/llvm.list
+
+wget -q -O - http://apt.llvm.org/llvm-snapshot.gpg.key|sudo apt-key add -
```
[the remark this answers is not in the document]
mehrdadh: done

mehrdadh:
@Mousius PTAL and explicitly approve if there's no other suggestions. thanks!

mehrdadh:
@tvm-bot rerun

docker/Dockerfile.ci_cortexm
ashutosh-arm: Yes, we will need to support `python3.7` until its EOL. Considering this, the next thing that pops up in my brain is when the `ci_cortexm` gets updated eventually corresponding to this change, will the default venv still remain at python3.7 or we would move cortex-m docker image to use `install/ubuntu2004_install_python.sh`? https://github.com/apache/tvm/blob/92da138bcb18c76604763785d3177de3f9f9b296/docker/Dockerfile.ci_cortexm#L38

mehrdadh: `install/ubuntu2004_install_python.sh` installs python3.8 as well. We have to add another script to install 3.7 in ubuntu20.04 if we need it.

Switching the ci image to 3.8 doesn't mean we're not supporting 3.7 anymore. Could you elaborate if there is a specific reason that this image has to run 3.7? We have switched other images like hexagon to 3.8 already.

mehrdadh: I see your point, I will change the scripts to install python3.7 in ci_cortexm.

mehrdadh: https://github.com/apache/tvm/pull/13736/commits/08ce7f8086d6fa6ec3281eab9b28663d269f0cbb addresses your concern.

ashutosh-arm: Thanks @mehrdadh

ashutosh-arm (approved):
I have had a look at the changes done for `python3.7` support. LGTM! Thanks @mehrdadh for the update.

Mousius (approved):
This looks ok to me, thanks for your diligence on the security points here @mehrdadh 😺

PR #13746, opened 2023-01-10
[Arith] Support eq in detect_clip_bound

wrongtest-intellif:
Hi~ this is a patch to handle EqNode in `detect_clip_bound`. We got some regression failure when using this methodology as the power of upstream arithmetic simplifier grows.

Many trivial cases like `x >= 4, x in [0, 4]` seems to get well simplified to `x==4`  and certain assumptions on some condition be non-eq comparation do not hold then.

### DIFF
--- a/src/arith/detect_linear_equation.cc
+++ b/src/arith/detect_linear_equation.cc
@@ -204,6 +204,7 @@ bool DetectClipBound(const PrimExpr& cond,
     if (!op->a.dtype().is_int()) return false;
     canonical = op->a - op->b;
   } else if (const EQNode* op = cond.as<EQNode>()) {
+    if (!op->a.dtype().is_int()) return false;
     canonical = op->a - op->b;
     is_eq = true;
   } else {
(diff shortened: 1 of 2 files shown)


wrongtest-intellif:
Could you kindly take a look at this? @Lunderberg

src/arith/detect_linear_equation.cc
```diff
@@ -217,6 +221,9 @@ bool DetectClipBound(const PrimExpr& cond,
     } else {
       p.min_value = -ret.base;
     }
+    if (is_eq) {
```
Lunderberg, line 224: Should this also be conditioned on `p.min_value.defined()`?  Otherwise, if `p.max_value` has already been set by another condition, this would overwrite that earlier bound.  This would be a repetition of the condition inside the `if(is_const_int(ret.coeff, -1))`.

Lunderberg: Now that I think of it, maybe it would be better to collect an `Optional<PrimExpr> min_value` and `Optional<PrimExpr> max_value`.  For inequalities, only one of the two is non-empty.  For equality expressions, both are non-empty.  That way, the logic of how to adjust `p.min_value` and `p.max_value` wouldn't need to be duplicated.  I'm thinking something like the following:

```c++
Optional<PrimExpr> min_value = NullOpt;
Optional<PrimExpr> max_value = NullOpt;

if(is_const_int(ret.coeff, 1)) {
  min_value = -ret.base;
  if(is_eq) {
    max_value = min_value;
  }
} else if(is_const_int(ret.coeff, -1)) {
  max_value = ret.base;
  if(is_eq) {
    min_value = max_value;
  }
}
```

src/arith/detect_linear_equation.cc
```diff
@@ -202,6 +203,9 @@ bool DetectClipBound(const PrimExpr& cond,
   } else if (const GENode* op = cond.as<GENode>()) {
     if (!op->a.dtype().is_int()) return false;
     canonical = op->a - op->b;
+  } else if (const EQNode* op = cond.as<EQNode>()) {
```
Lunderberg, line 206: For consistency, should this have the same check for `if(!op->a.dtype().is_int())` as the other cases?

Lunderberg (changes requested):
I like it overall, especially as you pointed out that simplification preferentially changes inequalities into equalities where possible.  I have a couple of suggestions, mostly to prevent letting conditions overwrite each other, but overall looks good!

Lunderberg (approved):
Thank you for the changes, and LGTM!

Also, the current CI failure has appeared on a few other PRs as well, and should be fixable by rebasing onto main.

PR #13750, opened 2023-01-10
[microNPU] Add a legalization test for TFLite PAD

Alexey-Yazev:
Added a legalization test for stand-alone pad operation which is legalized to depthwise operation on the NPU.


lhutton1:
Thanks @Alexey-Yazev!

PR #13752, opened 2023-01-10
[microTVM] Use QNN schedules to give SOTA performance

guberti:
In #13242, I rewrote microTVM's convolution schedules to give a major improvement in performance. While I demonstrated in tests that my changes worked, they could not be used with `relay.build`.

This pull request expands the functionality of #13242 and adds new `legalize` and `alter_op` passes to take advantage of the quantized schedules. This dramatically improves performance on some models, dramatically cuts RAM usage, and removes the _need_ for autotuning on microTVM. More specifically, for the `vww` model from MLPerf Tiny running on the `nucleo_l4r5zi`, this pull request:

- Improves **untuned** performance from `1741 ms` to `137 ms` - a **6.8x** improvement!
- Improves **tuned** performance from `337 ms` to `137 ms`.
- Sets a new state-of-the-art for MLPerf Tiny, [beating Plumerai's previous 208 ms record](https://blog.plumerai.com/)
- Reduces RAM consumption by **73 KB** (a large amount on microcontrollers!) by eliminating intermediate buffers.
- Reduces flash consumption for model weights by **5x**
- Slightly improves accuracy

@mehrdadh has kindly tested these changes himself, and has confirmed my `137 ms` figure.
 
To enable the schedules that grant these performance improvements, this pull request:
- Adds `out_layout` support to the regular and depthwise conv2d schedules from #13242.
- Generalizes the schedules from #13242 to be more widely applicable.
- Adds a layout alternation pass to ensure regular and depthwise conv2d schedules always get their desired input formats.
- Adds a `conv2d -> depthwise conv2d -> unpadded conv2d` rewrite step to remove empty channels from `conv2d` operators.
- Adds a `conv2d -> average pool -> dense` rewrite step to remove empty channels from `conv2d` operators.
- Adds an `alter_op` pass to fold padding into a separate Relay operator.


guberti:
Note - to use these changes, you will need to disable QNN legalization. This can be done by calling `relay.build` as follows:
```python
with tvm.transform.PassContext(
    opt_level=3,
    config={
        "tir.disable_vectorize": True,
        "relay.backend.use_meta_schedule": True,
        "relay.backend.tir_converter": "allow_extern",
    },
    disabled_pass=["qnn.Legalize"],
), meta_schedule.database.ScheduleFnDatabase(schedule_fn):
    lowered = tvm.relay.build(
        mod,
        target=target,
        params=params,
        runtime=crt_runtime,
        executor=executor,
    )
```

guberti:
## Next steps

The `137 ms` performance for the `vww` model is impressive, and beats the current state-of-the-art by a good margin. However, there is still a lot of room to improve our MLPerf Tiny performance even further:

- Measure performance for the `ic` and `kws` MLPerf Tiny models. The changes in this pull request should dramatically improve performance on these as well, and we should be able to use them with only minor tinkering. I'm currently working on a follow-up PR to add this functionality.
- Add support for **autotuning** to my tensordot schedules. Specifically, tuning `num_outputs` will give a substantial performance improvement with very little work. If we're willing to be a little more ambitious, we could use tuning to do reordering of the assembly code in the generated `tensordot` functions (this would especially help Cortex-M7 performance).
- Add a second, word-unaligned copy of convolution kernels when it would help, and add support for this to `tensordot.py`.
- Skip padding steps by folding padding into previous operators (this should be enabled by Relax).
- See if we can use floor instead of rounding in `tensordot.py`'s requantization implementation. This should shave a couple of `ms` off the runtimes of `ic`, `kws`, and `vww`, but it might hurt accuracy slightly.
- Write a Cortex-M schedule for `qnn_dense`. This will improve performance for `ic`, `kws`, and `vww` by a tiny amount, but it will dramatically improve `ad` performance (which currently still sucks).
- Generalize `tensordot.py` to support Cortex-M CPUs _without_ the DSP extension. This would allow us to give good performance for Cortex M0, M0+, M1, and M3 devices (this PR only improves performance for M4 and M7).
- Fix the bug with Arduino Cortex-M performance. Currently, this bug makes the Arduino implementation comically slow.

_Note: adding proper Helium support would require rewriting our `tensordot` implementation, and re-writing our legalization and alter_op passes as well. Helium is very cool, but proper support would take a lot of effort._

## Generalization of changes

Some of the `legalization` and `alter_op` changes would be useful very broadly in TVM, but are currently only enabled for Arm Cortex-M. This includes our output layout rewriting for `conv -> depthwise` convolution patterns, our stripping of empty channels from `conv2d` operator, and stripping `pad` out into a separate Relay operator (the last one only helps _in some cases_). However, I would want to write more general passes before doing this, and I'm not sure how these would interact with Relax. I'll hold off on this for now.

python/tvm/topi/arm_cpu/qnn.py
```diff
@@ -221,23 +261,21 @@ def qnn_conv2d(attrs, inputs, out_type):
...
     out_channels, kernel_h, kernel_w, _ = get_const_tuple(kernel.shape)
-    y_stride, x_stride = get_const_tuple(attrs.strides)
 
+    y_stride, x_stride = get_const_tuple(attrs.strides)
     out_height = _compute_output_dim(height, kernel_h, y_stride)
     out_width = _compute_output_dim(width, kernel_w, x_stride)
 
     # Decide how many sums our function should have running at the same time. Doing
     # this lets us do "more work" for each memory load, but doing too many of them causes us to run
     # out of registers. Currently this is set to either 1 or 2, but autotuning this value would
```
alanmacd, line 276: Can this sentence in the comment be removed since (I think?) it is no longer true?
```suggestion
    # out of registers. 
```

python/tvm/topi/arm_cpu/qnn.py
```diff
@@ -368,3 +389,136 @@ def kernel_ptr(buffer, c, offset=0):
...
+    function_names: Dict[Tuple, str],
+    function_code: str,
+    ptr_gens: Tuple[Callable, Callable],
+    output_layout="NHWC",
+):
+    out_height, out_width, out_channels = output_dimensions
+    data_shape, kernel_shape, bias_shape, scale_shape, output_shape = buffer_shapes
+    data_ptr, kernel_ptr = ptr_gens
+
+    def output_ptr(output, y, c):
```
alanmacd, line 406: nit: inconsistent use of leading '_' for internal functions

src/target/source/codegen_c.cc
```diff
@@ -631,8 +632,11 @@ void CodeGenC::PrintVecBinaryOp(const std::string& op, DataType t, PrimExpr lhs,
   }
 }
 
+unsigned int random_seed = 0;
 void CodeGenC::VisitStmt_(const AllocateConstNode* op) {
-  std::string symbol_name = op->buffer_var->name_hint;
+  // Add a random suffix to eliminate duplicate global variables.
+  int suffix = rand_r(&random_seed) % (2 << 24);
```
alanmacd, line 638: any reason to not just use a counter and add that as suffix to each global var name?

tests/python/relay/strategy/arm_cpu/test_quantized_convolution.py
```diff
@@ -95,6 +95,56 @@ def _get_mobilenet_v1_layer_attributes(layer_num):
     return ((1, 1, 1, 1), (1, 1), True)
 
 
+@pytest.mark.parametrize("layer", range(2, 27, 2))
+def test_infinite_bias_detection(interpreter, layer):
+    """Some models (mainly MobileNetV1) have kernels with many output channels full entirely of
+    zeroes. The VWW mdoel is one of these. This test confirms that the outputs of these channels,
```
alanmacd, line 101: nit: model

tests/python/relay/strategy/arm_cpu/test_quantized_convolution.py
```diff
@@ -95,6 +95,56 @@ def _get_mobilenet_v1_layer_attributes(layer_num):
...
+
+    # We now need to compute values for the following depthwise layer
+    if layer == 26:
+        return
+
+    _, kernel, bias, output = _load_tflite_layer(interpreter, layer + 1)
+    kernel_data, kernel_quant = kernel
+    bias_data, bias_quant = bias
+    output_data, output_quant = output
+    is_depthwise = _get_mobilenet_v1_layer_attributes(layer + 1)[2]
```
alanmacd, line 141:
```suggestion
    _, _, _, output = _load_tflite_layer(interpreter, layer + 1)
    output_data, output_quant = output
    is_depthwise = _get_mobilenet_v1_layer_attributes(layer + 1)[2]
```

python/tvm/topi/arm_cpu/qnn.py
guberti: Reworded the sentence to be correct.

src/target/source/codegen_c.cc
guberti: A counter would work fine - this is just a hack to fix the duplicate global variables problem. Would love @areusch's take on the right way to fix this long term.

python/tvm/topi/arm_cpu/qnn.py
guberti: My understanding was that `_` should be used for global internal functions, but that `_` should not be used for local functions. Correct me if I'm wrong, though.

alanmacd: I just noticed that inside the `def _make_unrolled_conv2d_primfunc` block there is the following:
```
    def output_ptr
    def _make_row_call
    def biased_quantized_conv2d
```

But I may be missing something and I am by no means an expert on the subject 😆

guberti: Oops, you're right - `_make_row_call` should be `make_row_call`. Thanks!

src/target/source/codegen_c.cc
mkatanbaf: I believe a similar fix is needed for the case usmp is enabled here: https://github.com/apache/tvm/blob/2877c5a3cf126637f0968bb9090454410c426cd0/src/target/source/source_module.cc#L295

guberti:
Would love a look from @areusch and @mehrdadh!

python/tvm/topi/arm_cpu/qnn.py
```diff
@@ -368,3 +389,139 @@ def kernel_ptr(buffer, c, offset=0):
...
+        __3 = bias[0, 0, 0, 0]
+        __4 = scale[0]
+        # pylint: enable=unused-variable
+
+        for c_ax in T.grid(out_channels // 2):
+            with T.block("conv2ds"):
+                T.block_attr({"pragma_import_c": function_code})
+                c = T.axis.remap("S", [c_ax]) * 2
+
+                # TODO how can I programatically make the right number of
```
mehrdadh, line 456: where you planing to change this part?

python/tvm/topi/arm_cpu/qnn_legalize.py
```diff
@@ -0,0 +1,349 @@
...
+
+    conv_input_zero_point = depthwise_op.args[2].data.numpy().item()
+    rq_input_scale = requantize_op.args[1].data.numpy()
+    rq_output_scale = requantize_op.args[3].data.numpy().item()
+    rq_output_zero_point = requantize_op.args[4].data.numpy().item()
+    bias_data = bias_add_op.args[1].data.numpy()
+
+    kernel_size = get_const_tuple(depthwise_op.attrs.kernel_size)
+
+    # Make a kernel_size x kernel_size array of fixed_input
```
mehrdadh, line 94: remove or if needed for explanation, make it more polished?

python/tvm/topi/arm_cpu/qnn_legalize.py
```diff
@@ -0,0 +1,349 @@
...
+        # a replacement. This is excessive - we only have to check for the padding that will
+        # actually be performed on the depthwise convolution, which is often less. If we felt even
+        # more ambitious, we could do the replacement for "close enough" looking convolution
+        # outputs, which in theory could reduce accuracy but in practice does not. Doing this would
+        # yield a ~0.5% speed gain on MobileNetV1, and nothing on other models.
+
+        if np.all(clipped == clipped[0, 0]):
+            fixed_outputs[i] = clipped[0, 0]
+
+    # TODO look for all-zero entries in the depthwise kernel. I don't think these really occur in
```
mehrdadh, line 119: same here?
also for TODOs please add your github handler in ()

python/tvm/topi/arm_cpu/qnn_legalize.py
```diff
@@ -0,0 +1,349 @@
...
+import numpy as np
+from scipy.signal import convolve2d
+from tvm.topi.utils import get_const_tuple
+from tvm import nd, relay
+from .qnn_alter_op import prev_ops_match, edit_attrs
+from ..nn import qnn_bias_add_legalize
+
+
+def _compute_fixed_conv2d_outputs(requantize_op):
+    """Compute all conv2d output values that do not depend on the layer input."""
```
areusch, line 41: could you document the return value, here and below?

src/relay/transforms/simplify_expr.cc
```diff
@@ -979,7 +992,16 @@ Pass SimplifyExpr() {
...
+Pass SimplifyExprPostAlterOp() {
+  runtime::TypedPackedFunc<Function(Function, IRModule, PassContext)> pass_func =
+      [=](Function f, IRModule m, PassContext pc) {
+        return Downcast<Function>(SimplifyExprPostAlterOp(f, m));
+      };
+  return CreateFunctionPass(pass_func, 0, "SimplifyExprPostAlterOp", {"InferType"});
+}
+
 TVM_REGISTER_GLOBAL("relay._transform.SimplifyExpr").set_body_typed(SimplifyExpr);
+// Don't globally register SimplifyExprPostAlterOp
```
mehrdadh, line 1004: why is that?

mehrdadh (changes requested):
Overall looks good! added few comments

src/target/source/codegen_c.cc
```diff
@@ -631,8 +632,10 @@ void CodeGenC::PrintVecBinaryOp(const std::string& op, DataType t, PrimExpr lhs,
   }
 }
 
+NameSupply global_name_supply = NameSupply("");
 void CodeGenC::VisitStmt_(const AllocateConstNode* op) {
-  std::string symbol_name = op->buffer_var->name_hint;
+  std::string symbol_name = global_name_supply->FreshName(op->buffer_var->name_hint);
```
areusch, line 637: why was this needed? i think need to initialize the NameSupply from the IRModule

tests/python/relay/strategy/arm_cpu/test_quantized_convolution.py
```diff
@@ -95,6 +95,53 @@ def _get_mobilenet_v1_layer_attributes(layer_num):
...
+        scale = bias_quant["scales"][i] / output_quant["scales"][0]
+        channel_constant = round(bias_data[i] * scale + out_zero_point)
+        clipped = min(127, max(-128, channel_constant))
+
+        out_channel_values = output_data[0, :, :, i].flatten()
+        assert all(x == clipped for x in out_channel_values)
+        fixed_channels[i] = clipped
+
+    # We now need to compute values for the following depthwise layer
+    if layer == 26:
```
areusch, line 134: anything else you can assert here other than `== 26`?

python/tvm/topi/arm_cpu/qnn.py
```diff
@@ -368,3 +389,139 @@ def kernel_ptr(buffer, c, offset=0):
 def schedule_qnn_depthwise_conv2d(_attrs, _outs, _target):
     """Schedule function for qnn.depthwise_conv2d."""
     return None
+
+
+def _make_unrolled_conv2d_primfunc(
```
areusch, line 394: can you document the parameters here?

areusch (commented):
thanks @guberti ! this is very exciting! i have some asks around documentation mostly

python/tvm/topi/arm_cpu/qnn.py
guberti: Fixed!

src/relay/transforms/simplify_expr.cc
guberti: You're right, we should register this as well. It doesn't seem useful elsewhere, but it is weird to not register just this one.

src/target/source/codegen_c.cc
guberti: Unfortunately, just calling `name_supply_->FreshName(op->buffer_var->name_hint);` does not work - we need to have a global name generator.

python/tvm/topi/arm_cpu/qnn.py
guberti: Done, for both `_make_unrolled_conv2d_primfunc` and `_make_conv2d_primfunc`.

python/tvm/relay/qnn/op/_qnn.py
```diff
@@ -85,12 +91,72 @@ def simulated_dequantize_compute(attrs, inputs, output_type):
...
+    if dtype.startswith("int"):
+        return (-(2 ** (bits - 1)), 2 ** (bits - 1) - 1)
+    elif dtype.startswith("uint"):
+        return (0, 2**bits - 1)
+    else:
+        raise TVMError(f"Clip legalization is not supported for data type '{dtype}'!")
+
+
+@register_legalize("clip")
+def legalize_clip(attrs, inputs, tinfos):
```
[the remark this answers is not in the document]
guberti: Added some tests to `tests/python/relay/qnn/test_clip_legalization`!

python/tvm/topi/arm_cpu/qnn.py
```diff
@@ -17,25 +17,40 @@
...
+
+
+def _pick_num_outputs(out_width):
+    """Guess a good value for num_outputs."""
+
+    assert out_width > 1
+
+    # num_outputs is capped at 8
+    for i in range(2, min(out_width + 1, 8)):
+        if out_width % i == 0:
```
mkatanbaf, line 50: Is this a requirement? i.e. does the `out_width` have to be divisible by `num_outputs`?

src/relay/qnn/op/dense.cc
```diff
@@ -242,7 +242,7 @@ RELAY_REGISTER_OP("qnn.dense")
                   "The quantization zero_point of the weight tensor.")
     .set_support_level(11)
     .add_type_rel("QDense", QnnDenseRel)
-    .set_attr<FInferCorrectLayout>("FInferCorrectLayout", QnnDenseInferCorrectLayout)
+//    .set_attr<FInferCorrectLayout>("FInferCorrectLayout", QnnDenseInferCorrectLayout)
```
areusch, line 245: revert?

src/target/source/codegen_c.cc
areusch: @gigiblender could you advise what you think is best to do here?

areusch (commented):
thansk @guberti ! left a couple more, but this is looking pretty good. @masahi could you have a look over the relay pass addition?

masahi (commented):
I suggest sending a standalone PR for the Relay pass addition with test cases. It's not clear what transformation the pass is intended for.

Also cc @ibsidorenko for topi QNN stuff.

python/tvm/topi/nn/qnn.py
```diff
@@ -212,6 +212,48 @@ def qnn_requantize_alter_layout(_attrs, _inputs, _tinfos, _out_type):
...
+        Grouped input symbols
+    tinfos : list
+        Input shape and dtype
+
+    """
+    return None
+
+
+@tvm.target.generic_func
+def qnn_clip_legalize(_attrs, inputs, _tinfos, _out_type):
```
ibsidorenko, line 236: Do we need this? Can not find place where we call `qnn_clip_legalize`

python/tvm/relay/qnn/op/_qnn.py
```diff
@@ -85,12 +91,72 @@ def simulated_dequantize_compute(attrs, inputs, output_type):
 register_strategy("qnn.conv2d", strategy.qnn_conv2d_strategy)
 
 
+def _get_clip_dtype_bounds(dtype):
+    """Returns the minimum and maximum values of a C integer data type."""
+    assert "int" in dtype
+    bits = int(dtype[dtype.find("int") + 3 :])
+
+    if dtype.startswith("int"):
+        return (-(2 ** (bits - 1)), 2 ** (bits - 1) - 1)
```
ibsidorenko, line 100: Just a nit comment... np.iinfo(dtype).min / np.iinfo(dtype).max are not suitable?

python/tvm/topi/nn/qnn.py
```diff
@@ -212,6 +212,48 @@ def qnn_requantize_alter_layout(_attrs, _inputs, _tinfos, _out_type):
...
+
+    Note
+    ----
+    Unlike other TOPI functions, this function operates on both graph level and operator level.
+    """
+    return inputs[0]
+
+
 @tvm.target.generic_func
 def qnn_add_alter_layout(_attrs, _inputs, _tinfos, _out_type):
```
ibsidorenko, line 258: Here is the same. How about `qnn_add_alter_layout `--> `add_alter_layout`. Since we do it for nn.add (not qnn.add)

python/tvm/relay/qnn/strategy/arm_cpu.py
```diff
@@ -21,9 +21,55 @@
...
+    return strategy
+
+
+@qnn_dense_strategy.register("arm_cpu")
+def qnn_dense_strategy_arm_cpu(_attrs, _inputs, _out_type, _target):
+    """qnn.dense strategy for arm_cpu"""
+    strategy = _op.OpStrategy()
+    strategy.add_implementation(
+        wrap_topi_qnn_dense(topi.hexagon.qnn_dense),
+        wrap_topi_schedule(topi.hexagon.schedule_qnn_dense),
```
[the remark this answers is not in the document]
guberti: It's fine for the time being. I know @mkatanbaf is working on a Cortex-M schedule for `dense`, but these operations do not take very much time on convolutional models.

python/tvm/topi/nn/qnn.py
guberti: You're right, removed.

python/tvm/relay/qnn/op/_qnn.py
guberti: That's much cleaner - fixed!

python/tvm/topi/arm_cpu/qnn.py
guberti: Yep! Otherwise we would have to implement a "special case" for left-over values. At some point, this will have to be done.

python/tvm/topi/arm_cpu/qnn.py
```diff
@@ -118,38 +133,89 @@ def _make_tscript_ptr(buffer, offset, length, dtype="int16"):
...
         aligned_channels = out_channels // 2
         offset_channels = out_channels // 2
-        c_step = tvm.tir.const(2)
-
-    def bias_ptr(bias, c):
-        return _make_tscript_ptr(bias, c, 1, dtype="int32")
-
-    def scale_ptr(scale, c):
-        return _make_tscript_ptr(scale, c, 1, dtype="int32")
+        c_step = const(2)
```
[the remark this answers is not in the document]
guberti: Ideally yes, but it makes the implementation harder in the depthwise_conv2d case. Luckily, no MLPerf models use an odd number of output channels.

mehrdadh (commented):
LGTM, I'll wait for others to finish their review

areusch:
@masahi could you have another look?

@guberti i think the main comment of mine that still needs resolving is https://github.com/apache/tvm/pull/13752#discussion_r1096416812

masahi:
My comment hasn't been addressed, but if the Relay pass addition is a minor one, I'm cool with this.

src/target/source/codegen_c.cc
areusch: i discussed with @gigiblender and in doing so we realized that there is already `var_idmap_` which uses another NameSupply. Does `AllocVarID(op->buffer_var.get())` work here? this parallels what was done in [AllocateNode](https://github.com/apache/tvm/blob/bf86d9f8c25e105a12c1b96ab0a4cc977ee088b7/src/target/source/codegen_c.cc#L857).

areusch (approved):
thanks @guberti !

PR #13758, opened 2023-01-11
[ONNX,FIX] onnx Pad operator `constant_value` omit use default value 0

fantasyRqg:
ONNX Pad input define: https://github.com/onnx/onnx/blob/main/docs/Operators.md#inputs-2---4-2


fantasyRqg:
@mbrookhart

mbrookhart (commented):
Can you please add a test that exercises the incorrect behavior?

AndrewZhaoLuo (approved):
Thanks!

PR #13767, opened 2023-01-12
[Relay][Frontend] Span Filling ONNX

chunit-quic:
- Set node name as the source name of span during the conversion of ONNX model.
- Assign node name to a node based on op type when it is empty.
- To get the reference of renamed nodes. Add a function to export the ONNX model after conversion.
- Add structural_equal comparisons with and without set_span to the existing test cases.
- Add span test case for exporting model parameter.

### DIFF
--- a/python/tvm/relay/frontend/onnx.py
+++ b/python/tvm/relay/frontend/onnx.py
@@ -57,6 +57,7 @@ from .common import (
     shape_of,
     try_resolve_var_to_const,
     unbind,
+    set_span,
 )
 
 __all__ = ["from_onnx"]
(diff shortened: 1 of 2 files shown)


chunit-quic:
Hi community,

This is a part of the RFC, TVM Explorer Infrastructure. After this PR, the source node information of ONNX will be tagged to the converted relay. 
Note that the node names might not exist in the some model. Therefore we did the following changes for span mapping:

1. Assign new node names based on their op types when they are missing during the conversion.
2. Add a parameter to from_ONNX to export the model after the conversion.

Node names of the exported model can be a reference to the source_names of spans. Feel free to give us any comment. Thank you! :D

[apache/tvm-rfcs#92](https://github.com/apache/tvm-rfcs/pull/92)
[#13116](https://github.com/apache/tvm/issues/13116)

masahi (approved):

PR #13769, opened 2023-01-12
[testing][py_converter] Enhance py_converter to better support entire modules

slyubomirsky:
This PR makes a few improvements to `py_converter` to make it more useful for fuzz testing, especially when running larger modules.

In particular, these changes are to support returning the definition of a global var directly (e.g., if you do `run_as_python(main_var, mod=mod)`, the result will be a function corresponding to mod["main"]) and to correct two bugs in the previous implementation:
* Previously, it was not possible to insert a function into a runtime container object like an ADT. This was because the converter was simply compiling Relay functions into Python functions. This change solves this problem by registering the functions into `PackedFunc`s. However, another fix was also needed: Even though `PackedFunc` is an `ObjectRef` in C++, the Python bindings do not recognize `PackedFunc`s as `Object`s, so a debug wrapper function was registered to enable this case. Hopefully that will not be destructive.
* The implementation relied on `IRModule.from_expr` to wrap passed in expressions in a module. However, `from_expr` will not overwrite the `main` function if one is passed in via the `functions` argument. Thus, if the user passed in a module that already had a `main` function defined, the wrapping would be done incorrectly and result in the `main` being copied many times. This PR corrects this error by not assuming that the name `main` will be available and instead constructing a new module with a reserved name for the target.

None of the cases above had been tested before (there are now tests included)


slyubomirsky:
Please review @junrushao (or tag other reviewers)

slyubomirsky:
Hm, this test worked in my fork. Wonder why it fails in mainline

slyubomirsky:
The problem seems to be that this cast on mainline will not convince Python that the result is an opaque Object rather than a PackedFunc. Not sure how to change it:
```python
TVM_REGISTER_GLOBAL("runtime.CastPackedFuncToObject").set_body_typed([](const PackedFunc& pf) {
  return Downcast<ObjectRef>(pf);
});
```
Any advice?

slyubomirsky:
@tqchen helpfully advised that the issue stems from passing a list containing a PackedFunc to the constructor. Calling the `ADT` constructor via the FFI API directly (while a little strange) gets around this issue. That seems like a good fix that does not require disrupting other sections of the codebase, so I will implement it

slyubomirsky:
The CI error appears to be spurious.

slyubomirsky:
@tvm-bot rerun

slyubomirsky:
I think there's been another spurious failure.

slyubomirsky:
@tvm-bot rerun

junrushao:
CC @Hzfengsy would you like to review this PR? Thanks!

Hzfengsy (approved):
LGTM. Thanks a lot, @slyubomirsky!

PR #13773, opened 2023-01-12
[Test][Topi] Avoid depending on f32 rounding behavior for crop_and_divide tests

Lunderberg:
The `crop_and_resize` operator uses floating-point arithmetic to determine whether an index is within a view-box.  This can cause the use of `extrapolation_value` to depend on target-dependent rounding differences.

For example, this issue was initially noticed on Vulkan during debugging of https://github.com/apache/tvm/pull/13530, and was the result of computing `0.2*223.0 + 0.8*223.0 < 223.0`.  If all intermediates are cast to float32, the left-hand side evaluates to `223.00002`.  If intermediates are kept at a higher precision, the left-hand side evaluates to `223.0`.

The floating-point indexing can't be removed, because the operator must match the API defined by TensorFlow's operator implementation. The TensorFlow documentation for [`CropAndResize`](https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/crop-and-resize) does not specify behavior in these cases, nor do the current TensorFlow unit tests check cases of rounding error.  Since the TensorFlow unit tests only use binary fractions for the `boxes` argument, which largely avoids the rounding issue, this commit updates the TVM unit tests to avoid depending on floating-point precision.

### DIFF
--- a/tests/python/relay/test_op_level5.py
+++ b/tests/python/relay/test_op_level5.py
@@ -238,12 +238,14 @@ class TestCropAndResize:
 
         if layout == "NHWC":
             img_shape = (10, 224, 224, 3)
-            boxes = np.array([[0.1, 0.2, 0.8, 0.7], [0.2, 0, 1, 0.6]]).astype("float32")
+            boxes = np.array([[0.125, 0.25, 0.8125, 0.71875], [0.25, 0, 1, 0.625]]).astype(
+                "float32"
+            )
             box_indices = np.array([1, 0]).astype("int32")
             crop_size = np.array([20, 30]).astype("int32")
         elif layout == "NCHW":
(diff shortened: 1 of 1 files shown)


Lunderberg:
Failure wasn't reproducible locally with 1k repetitions, each with a different random input.  Re-running CI with a fixed seed for random input data.

Lunderberg:
The errors in the i386 test look like they are due to the same type of rounding issue that this PR was intended to avoid.  That is, when an input pixel falls exactly on the edge of a bounding box, whether the `extrapolation_value` is used depends entirely on the numerical rounding error.  Rather than trying to reproduce exact errors, updating the PR such that the tests have some epsilon different applied to the bounding boxes.

masahi (approved):

PR #13780, opened 2023-01-13
[Adreno] fix injective schedule with textures on 4d output

elvin-n:
in case of 4d buffer output and texture input, tvm tir does not recognize that it needs to treat texture read data as vector. split and vectorization by 4 triggers proper work and usage of correct data types in opencl kernel


echuraev:
Close this PR because the original problem which tried to be solved in this PR should be solved in #14397.

PR #13788, opened 2023-01-15
[TIR][Fix] Buffer slicing using index dtype as extent

MasterJH5574:
This PR fixes a bug of TIR Buffer `__getitem__` on Python side.

Whenever the `__getitem__` parameter is a multi-dim slice, on non-slice dimension a Range will be created. Prior to this PR, that range will have int32 dtype in regardless of the `start`, while the proper behavior is to generate the extent according to the dtype of `start` when `start` is a PrimExpr.

This PR fixes the issue and provides a regression test.

### DIFF
--- a/tests/python/unittest/test_tvmscript_regression.py
+++ b/tests/python/unittest/test_tvmscript_regression.py
@@ -17,6 +17,7 @@
 import numpy
 
 import tvm
+import tvm.testing
 from tvm.script import tir as T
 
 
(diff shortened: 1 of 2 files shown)


MasterJH5574:
cc @cyx-6 @junrushao @tqchen

cyx-6 (approved):
LGTM. Thanks for reporting and fixing this bug.

junrushao (approved):
LGTM, and would you like to revert the change to `3rdparty/cutlass`? Seems like irrelevant

MasterJH5574:
> LGTM, and would you like to revert the change to `3rdparty/cutlass`? Seems like irrelevant

@junrushao Sorry my bad. I didn’t notice that... Will do later.

tests/python/unittest/test_tvmscript_regression.py
```diff
@@ -73,9 +74,17 @@ def func_ref():
...
+    assert func.body.block.body.body.block.reads[0].region[0].extent.dtype == "int64"
+
+
 if __name__ == "__main__":
-    a = numpy.zeros((10, 10), dtype="int8")
-    test_multi_element_array_in_outmost_namespace()
-    test_different_dtype_assignment_to_var()
-    b = 1
-    test_var_capturing_order()
+    tvm.testing.main()
```
Hzfengsy, line 90: `a = numpy.zeros((10, 10), dtype="int8")` here is to test var capturing in https://github.com/apache/tvm/pull/13640. I appreciate it if you could revert it and add a comment here.

PR #13789, opened 2023-01-15
[TIR][Fix] IndexDataTypeNormalizer not unwrapping float casting

MasterJH5574:
This PR fixes a bug of IndexDataTypeNormalizer that is used by CreatePrimFunc.

#13449 enhanced the normalizer by unwrapping unnecessary casts, by the rule that “if the value to be casted already has the target index dtype`, the cast will be omitted. However, there are cases where a cast here is for other purpose. Like in image resize operators, there are sometimes rounding operation for indices, like
```python
 T.Cast("int64", T.round(T.float32(128) / T.Cast("float32", oh) * T.Cast("float32", v_i2), dtype="float32"))
```
is used as a final index. Note that this PrimExpr contains casts that directly operated on indices, such as `T.Cast("float32", oh)`. This cast is for division purpose, in order to have floating point division result. As a result, it is not expected to unwrap such casts, while the previous implementation did remove the casts.

Therefore, this PR patches this issue by only unwrapping the casts whose target dtype is integer dtype.

cc @vinx13 @tqchen

### DIFF
--- a/python/tvm/tir/buffer.py
+++ b/python/tvm/tir/buffer.py
@@ -179,7 +179,7 @@ class Buffer(Object):
 
     def __getitem__(self, indices):
         from ..arith import Analyzer  # pylint: disable=import-outside-toplevel
-        from .expr import BufferLoad, Ramp  # pylint: disable=import-outside-toplevel
+        from .expr import BufferLoad, Ramp, const  # pylint: disable=import-outside-toplevel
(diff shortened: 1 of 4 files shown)


MasterJH5574:
Depends on #13788

junrushao (approved):
LGTM. Same here that you might want to revert the change to `3rdparty/cutlass`

PR #13790, opened 2023-01-16
[MeteSchedule] Bugfix: Add checks for nullable `run_secs`

spectrometerHBH:
Runner sec is an Optional ObjectRef, so we need to check it's defined before we access its value.


junrushao (approved):
LGTM! Please add some commit message :-)

PR #13791, opened 2023-01-16
[TOOL][NATIVE] Android native application for deploy and run

srkreddy1238:
This application helps as a reference for verifying and integration of TVM compiled models on Android targets natively independent of RPC setup.

tvmc will be used to for compiling tuning and to test run it before deployment.

This PR also covers
 * Enabling clml for tvmc compilation tool.
 * Graph runtime api "get_output_info" to return output tensor specification similar to "get_input_into"
 * This tool adds and enabled 3rdparty dependency "cnpy" to deal with npz files.


.gitmodules
```diff
@@ -19,3 +19,6 @@
 [submodule "3rdparty/OpenCL-Headers"]
 	path = 3rdparty/OpenCL-Headers
 	url = https://github.com/KhronosGroup/OpenCL-Headers.git
+[submodule "3rdparty/cnpy"]
+	path = 3rdparty/cnpy
+	url = https://github.com/rogersce/cnpy.git
```
junrushao, line 24: I was generally less aggressive when introducing new dependency. It doesn't mean I don't like it in this case, but do you think we could use TVM's native NDArray saving methods instead?

srkreddy1238: I considered it initially to reuse what is available in TVM. But, this require additional effort from the developer/user to parse NDArray dump. Or we need to give a python utility that converts it to a numpy dump.

apps/cpp_rtvm/main.cc
```diff
@@ -0,0 +1,260 @@
...
+/*!
+ * \brief ParseCmdArgs parses the command line arguments.
+ * \param argc arg counter
+ * \param argv arg values
+ * \param args the output structure which holds the parsed values
+ */
+void ParseCmdArgs(int argc, char* argv[], struct ToolArgs& args) {
+  const string model = GetCmdOption(argc, argv, "--model=");
+  if (!model.empty()) {
+    args.model = model;
```
echuraev, line 146: What if the `model` is empty? In this case, you won't update `args.model` and which model should be executed by TVM?

apps/cpp_rtvm/main.cc
```diff
@@ -0,0 +1,260 @@
...
+
+  if (args.input.empty() || args.output.empty()) {
+    LOG(INFO) << "Executing dry run ... ";
+    // Set random input for all inputs
+    for (auto& elem : mInfo.input_info) {
+      LOG(INFO) << "Set Random Input for :" << elem.first;
+      auto shape = elem.second.first;
+      size_t ssize = runner.GetInputMemSize(elem.first);
+      char* data = (char*)malloc(ssize);
+      // TODO: Ramdom initilalization
```
echuraev, line 203: When this TODO comment should be resolved?

apps/cpp_rtvm/scripts/download_models.py
```diff
@@ -0,0 +1,36 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+tmp_dir = "./model_data/"
+dload_models = []
+
+# Keras : Resnet50
```
echuraev, line 24: In the `README.md` was written that this script can be used to `download varius well known models from different frameworks`. Now it looks like this script can download only ResNet50 from Keras. Probably the `README.md` should be updated, and we should mention that this script can be used as a template, and it should be extended by the user to download other models?

apps/cpp_rtvm/tvm_runner.cc
```diff
@@ -0,0 +1,304 @@
...
+namespace runtime {
+
+/*!
+ * \brief Get the TVM device id corresponding to device string.
+ * \param device the target device in string format.
+ * \return cl_device corresponding to the device string.
+ */
+int GetTVMDevice(std::string device) {
+  if (!device.compare("cl")) {
+    return static_cast<int>(kDLOpenCL);
```
echuraev, line 43: For now `rtvm` can be used only for running model on OpenCL device. Should it be extended to other devices (e.g. cpu or others)?

apps/cpp_rtvm/tvm_runner.cc
```diff
@@ -0,0 +1,304 @@
...
+/*!
+ * \brief Get the output alloc mem size.
+ * \param input_id The output id to query the mem size.
+ * \return The memory size.
+ */
+size_t TVMRunner::GetOutputMemSize(std::string input_id) {
+  LOG(INFO) << "TVMRunner::GetOutputMemSize:" << input_id;
+
+  NDArray in_arr = r_graph_handle.GetFunction("get_output")(input_id);
+  auto ssize = GetMemSize(in_arr);
```
echuraev, line 134:
```suggestion
size_t TVMRunner::GetOutputMemSize(std::string output_id) {
  LOG(INFO) << "TVMRunner::GetOutputMemSize:" << output_id;

  NDArray out_arr = r_graph_handle.GetFunction("get_output")(output_id);
  auto ssize = GetMemSize(out_arr);
```

apps/cpp_rtvm/tvm_runner.cc
```diff
@@ -0,0 +1,304 @@
...
+ * \param 0 on success else error code.
+ */
+int TVMRunner::SetInput(std::string inputfile) {
+  LOG(INFO) << "TVMRunner::SetInput (Numpy):" << inputfile;
+  cnpy::npz_t npz_input = cnpy::npz_load(inputfile);
+
+  for (auto& elem : mInfo.input_info) {
+    LOG(INFO) << "Set Numpy Input for :" << elem.first;
+    NDArray in_arr = r_graph_handle.GetFunction("get_input")(elem.first);
+    auto ssize = GetMemSize(in_arr);
```
echuraev, line 151: It is the same, isn't it?
```suggestion
    auto ssize = GetInputMemSize(elem.first);
```

apps/cpp_rtvm/tvm_runner.cc
```diff
@@ -0,0 +1,304 @@
...
+/*!
+ * \brief Set the model input from the given binary buffer.
+ * \param input_id input node name.
+ * \param raw_input binary input buffer to copy over input NDArray.
+ * \param 0 on success else error code.
+ */
+int TVMRunner::SetInput(std::string input_id, char* raw_input) {
+  LOG(INFO) << "TVMRunner::SetInput (Raw)";
+  NDArray in_arr = r_graph_handle.GetFunction("get_input")(input_id);
+  auto ssize = GetMemSize(in_arr);
```
echuraev, line 172:
```suggestion
auto ssize = GetInputMemSize(input_id);
```

echuraev: The same comments for `GetOutput` methods.

.gitmodules
junrushao: Thanks for the context! If so, I do not object the idea as long as it's not part of default build

apps/cpp_rtvm/tvm_runner.cc
srkreddy1238: But, I need in_arr for data copy below.

apps/cpp_rtvm/main.cc
srkreddy1238: We could have a random initialization but doesn't matter for TVM as it can dry run with garbage data. Removing todo for now.

apps/cpp_rtvm/tvm_runner.cc
echuraev: Ups... Sorry. My bad. Thank you!

.gitmodules
tqchen: Personally I still think having an util to be able to create such a dump can be helpful. Mainly because the numpy format can be complicated and there will be questions of supporting other formats, such as protobuf. So converting to a standard one that we commonly use can be better.

echuraev (approved):
LGTM. Thanks

junrushao (approved):
Please go ahead and get it merged as @echuraev approved it already :-)

tqchen:
One minor note: submodule inclusion would need to include license info, specifically,

- add a section here https://github.com/apache/tvm/blob/main/LICENSE#L232
- add license to https://github.com/apache/tvm/tree/main/licenses

Please send a followup PR :)

PR #13792, opened 2023-01-16
[TIR] Fix cache_write bug with allocate const node

quic-sanirudh:
Applying `cache_write`, when there's a allocate_const node in the TIR causes the cache write block to be inserted at the incorrect location. This patch tries to fix that error.

Co-authored-by: Abhikrant Sharma <abhikran-quic@quicinc.com>


Hzfengsy (approved):
LGTM

masahi (approved):

PR #13802, opened 2023-01-18
[ONNX] Support Bernoulli op on ONNX front-end

vvchernov:
Support Bernoulli op on ONNX front-end

### DIFF
--- a/python/tvm/relay/frontend/onnx.py
+++ b/python/tvm/relay/frontend/onnx.py
@@ -5582,6 +5582,35 @@ class GridSample(OnnxOpConverter):
         )
 
 
+class Bernoulli(OnnxOpConverter):
+    """Operator converter for Bernoulli"""
+
+    @classmethod
+    def _impl_v15(cls, inputs, attr, params):
+        in_dtype = infer_type(inputs[0]).checked_type.dtype
+        assert in_dtype in [
+            "float32",
+            "float64",
+        ], "Only float input tensor is currently supported."
+        # The data type for the elements of the output tensor.
+        # if not specified, we will use the data type of the input tensor
+        out_dtype = attr.get("dtype", None)
+        if out_dtype is None:
+            out_dtype = in_dtype
+        else:
+            out_dtype = get_type(out_dtype)
+
+        seed = attr.get("seed", None)
+        if seed is None:
+            seed = np.random.randint(1e6)
+        else:
+            seed = int(seed)
+
+        key = _random.threefry_key(seed)
+        _, uniform_nums = _op.random.uniform(key, infer_shape(inputs[0]), in_dtype)
+        return _op.cast(_op.less(uniform_nums, inputs[0]), out_dtype)
+
+
 class RandomNormal(OnnxOpConverter):
     """Operator converter for random_normal"""
 
@@ -6348,6 +6377,7 @@ def _get_convert_map(opset):
         "QLinearGlobalAveragePool": QLinearGlobalAveragePool.get_converter(opset),
         "QLinearLeakyRelu": QLinearLeakyRelu.get_converter(opset),
         # Random number generation.
+        "Bernoulli": Bernoulli.get_converter(opset),
         "RandomNormal": RandomNormal.get_converter(opset),
         "RandomNormalLike": RandomNormalLike.get_converter(opset),
         "RandomUniform": RandomUniform.get_converter(opset),
--- a/python/tvm/relay/frontend/onnx.py
+++ b/python/tvm/relay/frontend/onnx.py
@@ -5607,7 +5607,8 @@ class Bernoulli(OnnxOpConverter):
             seed = int(seed)
 
         key = _random.threefry_key(seed)
-        _, uniform_nums = _op.random.uniform(key, infer_shape(inputs[0]), in_dtype)
+        inter_outputs = _op.random.uniform(key, infer_shape(inputs[0]), in_dtype)
+        _, uniform_nums = _expr.TupleWrapper(inter_outputs, 2)
         return _op.cast(_op.less(uniform_nums, inputs[0]), out_dtype)
 
 
--- a/tests/python/frontend/onnx/test_forward.py
+++ b/tests/python/frontend/onnx/test_forward.py
@@ -6786,19 +6786,19 @@ def test_random_bernoulli(target, dev):
                 tvm.testing.assert_allclose(ideal_mean, tvm_val.mean(), rtol=rtol, atol=atol)
 
     # Simple test
-    verify_bernoulli([1000])
+    verify_bernoulli(shape=[1000])
 
     # Floating output type
-    verify_bernoulli([1000], out_dtype="float32")
+    verify_bernoulli(shape=[1000], out_dtype="float32")
 
     # Double input type
-    verify_bernoulli([1000], in_dtype="float64")
+    verify_bernoulli(shape=[1000], in_dtype="float64")
 
     # Test N-D tensor generation
-    verify_bernoulli([2, 4, 100, 100])
+    verify_bernoulli(shape=[2, 4, 100, 100])
 
     # Test with seed
-    verify_bernoulli([1000], seed=np.random.randint(1e6))
+    verify_bernoulli(shape=[1000], seed=np.random.randint(1e6))
 
     # Test input sequence of 0 and 1
     inputs = np.random.randint(2, size=[2, 4, 100, 100])
--- a/tests/python/frontend/onnx/test_forward.py
+++ b/tests/python/frontend/onnx/test_forward.py
@@ -6792,7 +6792,6 @@ def test_random_bernoulli(target, dev):
                 assert tvm_flat_val[i] == 0 or tvm_flat_val[i] == 1
         for tvm_val, ort_val in zip(tvm_out, ort_out):
             tvm.testing.assert_allclose(ort_val.mean(), tvm_val.mean(), rtol=rtol, atol=atol)
-            tvm.testing.assert_allclose(np.std(ort_val), np.std(tvm_val), rtol=rtol, atol=atol)
             assert ort_val.dtype == tvm_val.dtype
 
     # Simple test
--- a/tests/python/frontend/onnx/test_forward.py
+++ b/tests/python/frontend/onnx/test_forward.py
@@ -6776,8 +6776,7 @@ def test_random_bernoulli(target, dev):
         ideal_mean = np.mean(inputs)
         # check that values are 0 or 1
         tvm_flat = tvm_out.flatten()
-        for i in range(len(tvm_flat)):
-            assert tvm_flat[i] == 0 or tvm_flat[i] == 1
+        assert np.array_equal(tvm_flat, tvm_flat.astype("bool"))
         if in_out_equal:
             tvm.testing.assert_allclose(inputs, tvm_out)
         else:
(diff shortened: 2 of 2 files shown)


python/tvm/relay/frontend/onnx.py
```diff
@@ -5451,6 +5451,36 @@ def _impl_v16(cls, inputs, attr, params):
...
+class Bernoulli(OnnxOpConverter):
+    """Operator converter for Bernoulli"""
+
+    @classmethod
+    def _impl_v15(cls, inputs, attr, params):
+        in_dtype = infer_type(inputs[0]).checked_type.dtype
+        assert in_dtype in [
+            "float32",
+            "float64",
+        ], "Only float input tensor is currently supported."
```
KJlaccHoeUM9l, line 5463: ONNX has support for `float16`. However, this type is not supported here. Maybe it's worth pointing out the reason (`TODO`) why this data type is not currently supported at the ONNX front-end level?

KJlaccHoeUM9l (approved):
LGTM

vvchernov:
Hello @AndrewZhaoLuo! Could you review this PR? One more question: do you now what means `expended` part in test names ("test_bernoulli_expanded", "test_bernoulli_double_expanded", "test_bernoulli_seed_expanded")? I'm ready to support these tests, but do not know what should they do.

echuraev (approved):
LGTM. Thanks

AndrewZhaoLuo:
IIRC the expanded tests simply expand the node into things that can be written in terms of existing ops.

E.g.
If we had an “FMA” node in ONNX which does output = in1 * in2 + in3 , the FMA_test would be just an FMA node but the FMA_expanded_test will be a multiplication node and an addition node.

So in FMA_test we would hit the FMA node converter, while for FMA_expanded_test we would hit the multiply and addition node converters

https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/test/onnx/main.cc#L699 If we look here, the bernoulli tests are listed as broken so :shrug:

tests/python/frontend/onnx/test_forward.py
```diff
@@ -6663,6 +6663,105 @@ def verify_qlinearsigmoid(a_shape):
...
+                convert_config=convert_config,
+            )
+
+        if not isinstance(tvm_out, list):
+            tvm_out = [tvm_out]
+        if not isinstance(ort_out, list):
+            ort_out = [ort_out]
+        for tvm_val, ort_val in zip(tvm_out, ort_out):
+            tvm.testing.assert_allclose(ort_val.mean(), tvm_val.mean(), rtol=rtol, atol=atol)
+            tvm.testing.assert_allclose(np.std(ort_val), np.std(tvm_val), rtol=rtol, atol=atol)
```
AndrewZhaoLuo, line 6746: stdev is meaningless if our output is only 1 or 0, all information is encoded in mean.

tests/python/frontend/onnx/test_forward.py
```diff
@@ -6663,6 +6663,105 @@ def verify_qlinearsigmoid(a_shape):
...
+                opt_level=opt_level,
+                convert_config=convert_config,
+            )
+
+        if not isinstance(tvm_out, list):
+            tvm_out = [tvm_out]
+        if not isinstance(ort_out, list):
+            ort_out = [ort_out]
+        for tvm_val, ort_val in zip(tvm_out, ort_out):
+            tvm.testing.assert_allclose(ort_val.mean(), tvm_val.mean(), rtol=rtol, atol=atol)
```
AndrewZhaoLuo, line 6745: Hmm I think you need to do two things:
1. verify all outputs are 0 or 1
2. verify the mean of tvm_val ~= input probability, no need to compare against ort_val (or at least we gain no additional confidence in result from doing so)

2. also has an issue that it is inherently flaky. We can do student-t test or something but for example if you set p-value to 0.05, 5% of the time the test will fail.

cc @octoJon for consultation

AndrewZhaoLuo (commented):
LGTM for most part, just some comments on testing

AndrewZhaoLuo (changes requested):
Actually going to block for now, due to concern about flakes. at least until @octoJon chimes in.

tests/python/frontend/onnx/test_forward.py
octoJon: Statistician here! Here are the things I would test:

1. Verify all outputs are 0 or 1.
2. Verify that if you feed in an array of input probabilities which is a mix of zeroes and ones, then the outputs you get out match your inputs. That is, a Bernoulli distribution with p=1 should always return a 1, and a Bernoulli distribution with p=0 should always return a 0. And if you have a mix of those in the same input tensor, then you still get the expected mix of zeroes and ones out.
3. Pick a couple fixed probabilities to test. (I would recommend p=0.5 and p=0.1.) Verify that if you feed in a large-ish input array of all, say, p=0.5, then the mean value of your outputs is "sufficiently" close to 0.5. The best way to test for that is with a binomial proportion test, like `scipy.stats.binomtest`. Normally we'd run a binomial proportion test and confirm that the p-value from that test is larger than 0.05, but, as Andrew says, that would give you quite a flaky test. For a non-flaky test, I would recommend having the test fail only if the p-value from the binomial proportion test is <1e-6, and just making sure that you have a large enough sample that it's possible to detect small deviations from the intended binomial distribution. With a sample of size 10,000, there's roughly a 1e-6 chance that your sample mean will be more than 2.5 percentage points away from the true mean, so an input tensor with 10,000 entries should get the job done.

So to recap on #3: If it's not too slow to generate a tensor of 10,000 Bernoulli values, my recommendation would be to generate 10,000 Bernoulli values with p=0.5, run a binomial proportion test, and have the unit test fail if the binomial proportion test has a p-value of <1e-6. That test should only have a flaky failure once per million times that it's run. ... And then also repeat that unit test again with p=0.1 rather than p=0.5.

tests/python/frontend/onnx/test_forward.py
octoJon: I agree that there's no need to check standard deviations. Standard deviation isn't meaningless here, it's just redundant in the sense that if you know the sample mean then you also know the sample standard deviation.

vvchernov: done

tests/python/frontend/onnx/test_forward.py
vvchernov: all points have been done

vvchernov:
Hello @AndrewZhaoLuo and @octoJon! I think CI will be green soon, Could you recheck updated test part?

tests/python/frontend/onnx/test_forward.py
```diff
@@ -6707,6 +6707,117 @@ def verify_qlinearsigmoid(a_shape):
...
+                target,
+                dev,
+            )
+
+        if isinstance(tvm_out, list):
+            tvm_out = tvm_out[0]
+        ideal_mean = np.mean(inputs)
+        # check that values are 0 or 1
+        tvm_flat = tvm_out.flatten()
+        for i in range(len(tvm_flat)):
```
AndrewZhaoLuo, line 6779: you can make this vectorized by doing something like 
`assert np.array_equal(arr, arr.astype('bool'))`
which will be faster

tests/python/frontend/onnx/test_forward.py
```diff
@@ -6707,6 +6707,117 @@ def verify_qlinearsigmoid(a_shape):
...
+            )
+            assert bnm_test_res.pvalue >= 1e-6
+
+    # Test input sequence of 0 and 1
+    inputs = np.random.randint(2, size=[10000]).astype("float32")
+    verify_bernoulli(inputs, in_out_equal=True)
+
+    # Binomial test input with 0.5 values
+    val_num = 10000
+    arr = [0.5] * val_num
```
AndrewZhaoLuo, line 6796: nit: recommend np.ones([10000]) * 0.5 for numpy style

tests/python/frontend/onnx/test_forward.py
```diff
@@ -6707,6 +6707,117 @@ def verify_qlinearsigmoid(a_shape):
...
+    verify_bernoulli(shape=[1000], out_dtype="float32")
+
+    # Double input type
+    verify_bernoulli(shape=[1000], in_dtype="float64")
+
+    # Test N-D tensor generation
+    verify_bernoulli(shape=[2, 4, 100, 100])
+
+    # Test with seed
+    verify_bernoulli(shape=[1000], seed=np.random.randint(1e6))
```
AndrewZhaoLuo, line 6818: For seed value, I believe you need to check you get the same values every time you invoke new bernoulli func with the same seed,

AndrewZhaoLuo:
LGTM, just a few final nits.

tests/python/frontend/onnx/test_forward.py
```diff
@@ -6707,6 +6707,117 @@ def verify_qlinearsigmoid(a_shape):
...
+        ideal_mean = np.mean(inputs)
+        # check that values are 0 or 1
+        tvm_flat = tvm_out.flatten()
+        for i in range(len(tvm_flat)):
+            assert tvm_flat[i] == 0 or tvm_flat[i] == 1
+        if in_out_equal:
+            tvm.testing.assert_allclose(inputs, tvm_out)
+        else:
+            # check that mean value is close to the theoretical one by binomial test
+            bnm_test_res = scipy.stats.binomtest(
```
octoJon, line 6785: Nit: Strictly speaking, this test is only appropriate when the input probabilities are all identical. I *think* it's guaranteed to be over-conservative in cases where the input probabilities are not identical, like when you call `verify_bernoulli(shape=[1000])`. (By "over-conservative", I mean that the true rate of flaky failures will be less than the nominal 1e-6 implied by the p-value threshold in the test.) Reference for that claim: https://stats.stackexchange.com/a/510798

I would recommend just documenting this with a comment in the code. Something like:
> This test is strictly appropriate when input probabilities are all identical. In that case, it should lead to flaky failures in only one run in a million.
> The test should be over-conservative when input probabilities are not identical. (i.e., It should have a rate of flaky failures lower than one run in a million.) If this test starts repeatedly throwing flaky failures, consult a statistician in addition to your regular debugging.

octoJon: Oh, god. I didn't read that reference link closely enough. It contains a really blatantly misogynistic comment. Apologies to anyone I exposed to that. Please don't include that reference link in the code.

(I think the math is right, though.)

tests/python/frontend/onnx/test_forward.py
vvchernov: done

tests/python/frontend/onnx/test_forward.py
vvchernov: done

tests/python/frontend/onnx/test_forward.py
vvchernov: done

tests/python/frontend/onnx/test_forward.py
vvchernov: Hello @octoJon! I've modified the test due to it was already "over-conservative" with p-value threshold = 1e-6. I've increased threshold to 0.05 as more classical approach. If test condition failed there are two cases: something wrong in the operation or we have gotten "bad" output sequence on the tail of distribution. Due to the last is rare case and should be rechecked I repeat the test again (and third time if need) with new seed for internal distribution (input is the same).
P.S. As you know RandomUniform and RandomNormal already were implemented on TVM side. Possibly their CI tests should be also updated to testing stability and avoiding of flaky failures.

AndrewZhaoLuo (approved):
Thanks!

PR #13808, opened 2023-01-19
[TIR]Fix the crash of the pass RemoveNoOp

lightzhan-intellif:
This PR tries to fix the crash of the pass `RemoveNoOp` when the condition of IfThenElseNode can be statically determined. For example:
```python
  @L.prim_func
  def remove_no_op():
    if True:
      L.evaluate(0)
    else:
      L.evaluate(0)
```
This case will cause `Segment fault` error because `RemoveNoOp` don't cover the case the return of `Parent::VisitStmt_` is not a `IfThenElseNode`, which will result in the `stmt.as<IfThenElseNode>()`  being `null`.


Hzfengsy (approved):
LGTM. Thanks for the fixing

PR #13809, opened 2023-01-19
[TIR][FIX] check args size when creating prim_func by runtime::Registry

Ever-Kid:
default value for `index_dtype_override` is null, so it would be better if the registry block of `te.CreatePrimFunc` could check args size.

### DIFF
--- a/src/te/operation/create_primfunc.cc
+++ b/src/te/operation/create_primfunc.cc
@@ -574,6 +574,7 @@ TVM_REGISTER_GLOBAL("te.CreatePrimFunc").set_body([](TVMArgs args, TVMRetValue* 
   Array<te::Tensor> arg_list = args[0];
   std::optional<DataType> index_dtype_override{std::nullopt};
   // Add conversion to make std::optional compatible with FFI.
+  ICHECK_EQ(args.size(), 2);
   if (args[1].type_code() != kTVMNullptr) {
     index_dtype_override = args[1].operator DataType();
   }


Ever-Kid:
cc. @vinx13

Hzfengsy:

Ever-Kid:
> Could you please add a regression test?

Sure, I guess I should test `te.CreatePrimFunc` by `runtime::Registry`. Any hints where to put this testcase, like `tests/cpp/data_type_rewriter_test.cc`, `tests/cpp/tensor_test.cc` or others ?

Hzfengsy:
After looking at the codebase again, it would be great if we can ensure this PackedFunc has two args. Another approach is adding `ICHECK_EQ(args.size(), 2)`. 

Keeping PackedFunction stable and minimal will benefit the registry system and user experience. If you want to call it from cpp side, we have [API](https://github.com/apache/tvm/blob/main/src/te/operation/create_primfunc.h#L34) here, and no need to call PackedFunction

junrushao:
That sounds reasonable to me :-)

tqchen (approved):

tqchen:
sorry i approved pre-maturely, after seeing @Hzfengsy 's comment, i agree with the accessment

Ever-Kid:
I followed @Hzfengsy 's suggestion. And since no need to call PackedFunc, also testcases like `test_tensor_layout_attr` in `tests/python/unittest/test_te_create_primfunc.py` are quite inclusive, so I added no test. 
Any other suggestions would be welcomed~ :smile:

PR #13813, opened 2023-01-20
[TVMScript] `T.allocate` with `T.decl_buffer` syntax sugar for TVMScript printer

cyx-6:
This PR implements the syntax sugar of `T.allocate` with `T.decl_buffer` for new TVMScript printer. This syntax sugar will skip the `T.allocate`, when its body is a matched `T.decl_buffer`, and the `Var` defined in `T.allocate` is only used in that `T.decl_buffer`. For example, it will change

```python
buffer_data = T.allocate([128, 128])
buffer = T.decl_buffer([128, 128], data=buffer_data)
```
into

```python
buffer = T.decl_buffer([128, 128])
```
but keep the following `T.allocate` unchanged:
```python
buffer_data = T.allocate([128, 128])
buffer_0 = T.decl_buffer([128, 128], data=buffer_data)
buffer_1 = T.decl_buffer([128, 128], data=buffer_data)
```
and
```python
buffer_data = T.allocate([128, 128])
buffer = T.decl_buffer([256, 256], data=buffer_data)
```

### DIFF
--- a/tests/python/unittest/test_tvmscript_printer_tir.py
+++ b/tests/python/unittest/test_tvmscript_printer_tir.py
@@ -22,6 +22,7 @@ from tvm.ir import Range
 from tvm.script.ir_builder import IRBuilder
 from tvm.script.ir_builder import tir as T
 from tvm.script.printer import default
+import tvm.testing
 
 
 @contextmanager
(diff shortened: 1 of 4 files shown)


tests/python/unittest/test_tvmscript_printer_tir.py
```diff
@@ -327,6 +328,53 @@ def test_allocate():
...
+        with T.allocate([128, 128], "float32") as buffer_data:
+            with T.decl_buffer([256, 256], "float32", data=buffer_data) as buffer:
+                T.evaluate(buffer_data)
+    obj = ib.get()
+    _assert_print(
+        obj,
+        """
+with T.allocate([128, 128], "float32", "global") as v:
+    buffer = T.decl_buffer((256, 256), data=v)
+    T.evaluate(v)
```
junrushao, line 373: i think we could still do sugaring in this case, i.e.:

```python

with T.decl_buffer((256, 256), data=v) as buffer:
    T.evaluate(buffer.data)
```

junrushao (approved):
LGTM! Please enhance the sugaring in a separate PR :-)

PR #12846, opened 2022-09-20
[CMSIS-NN] Reduction in code size of AOT test runner binary

ashutosh-arm:
This change reduced the code size of AOT test runner binary for CMSIS-NN flow.

Presently, integration tests for CMSIS-NN generate
binaries which includes all the CMSIS-NN kernels sources.
This fix introduces additional compilation and runtime
knobs to pick kernel sources for which TVM has generated
the code.

PR #13815, opened 2023-01-20
[CMSIS-NN] Reduction in code size of AOT test runner binary

NicolaLancellotti:
This pr has the same changes of https://github.com/apache/tvm/pull/12846.

### DIFF
--- a/python/tvm/topi/arm_cpu/mprofile/dsp/micro_kernel/multi_channel_convolve.py
+++ b/python/tvm/topi/arm_cpu/mprofile/dsp/micro_kernel/multi_channel_convolve.py
@@ -179,20 +179,26 @@ def _dual_int16_channel_convolve_impl(_tensor_h, tensor_w, channels, kernel_h, k
         extern "C"
         #endif
         int32_t {_get_func_name("int16", tensor_w, channels, kernel_h, kernel_w, suffix)}(
-            uint32_t *out,
-            uint32_t *tensor,
-            uint32_t *kernel) {{
+            int32_t *out,
+            int16_t *tensor,
+            int16_t *kernel) {{
 
-          uint32_t sum_c0 = 0;
-          uint32_t sum_c1 = 0;
+          int32_t sum_c0 = 0;
+          int32_t sum_c1 = 0;
+
+          int32_t kernel_i32[{kernel_h} * {kernel_w}];
+          memcpy(kernel_i32, kernel, {kernel_h} * {kernel_w} * 4);
+
+          int32_t tensor_length = {((kernel_w - 1) * (channels // 2) + (kernel_h - 1) * tensor_w * (channels // 2)) + 1};
+          int32_t tensor_i32[tensor_length];
+          memcpy(tensor_i32, tensor, tensor_length * 4);
 
           #pragma GCC unroll 3
           for (int i = 0; i < {kernel_h}; i++) {{
             #pragma GCC unroll 3
             for (int j = 0; j < {kernel_w}; j++) {{
-              uint32_t tensor_c10 = *(tensor + j * {channels // 2}
-                + i * {tensor_w * (channels // 2)});
-              uint32_t kernel_c10 = *kernel++;
+              int32_t tensor_c10 = tensor_i32[j * {channels // 2} + i * {tensor_w * (channels // 2)}];
+              int32_t kernel_c10 = kernel_i32[{kernel_w} * i + j];
               sum_c0 = __builtin_arm_smlabb(tensor_c10, kernel_c10, sum_c0);
               sum_c1 = __builtin_arm_smlatt(tensor_c10, kernel_c10, sum_c1);
             }}
--- a/python/tvm/topi/arm_cpu/mprofile/dsp/micro_kernel/multi_channel_convolve.py
+++ b/python/tvm/topi/arm_cpu/mprofile/dsp/micro_kernel/multi_channel_convolve.py
@@ -116,15 +116,15 @@ def _quad_int8_channel_convolve_impl(_tensor_h, tensor_w, channels, kernel_h, ke
             tensor_c3210, \
             sum_c0, sum_c1, sum_c2, sum_c3) {{ \
           \
-          uint32_t kernel_c3210 = *arranged_kernel++; \
+          int32_t kernel_c3210 = *arranged_kernel++; \
           \
-          uint32_t tensor_c20 = __sxtb16(tensor_c3210); \
-          uint32_t kernel_c20 = __sxtb16(kernel_c3210); \
+          int32_t tensor_c20 = __sxtb16(tensor_c3210); \
+          int32_t kernel_c20 = __sxtb16(kernel_c3210); \
           sum_c0 = __builtin_arm_smlabb(tensor_c20, kernel_c20, sum_c0); \
           sum_c2 = __builtin_arm_smlatt(tensor_c20, kernel_c20, sum_c2); \
           \
-          uint32_t tensor_c31 = __sxtb16(__ror(tensor_c3210, 8)); \
-          uint32_t kernel_c31 = __sxtb16(__ror(kernel_c3210, 8)); \
+          int32_t tensor_c31 = __sxtb16(__ror(tensor_c3210, 8)); \
+          int32_t kernel_c31 = __sxtb16(__ror(kernel_c3210, 8)); \
           sum_c1 = __builtin_arm_smlabb(tensor_c31, kernel_c31, sum_c1); \
           sum_c3 = __builtin_arm_smlatt(tensor_c31, kernel_c31, sum_c3); \
         }}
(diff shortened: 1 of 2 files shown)


NicolaLancellotti:
Ci failed because of a type of punning error, which has been addressed in the second commit: https://github.com/apache/tvm/pull/13815/commits/9febf6554907338324724fcb9d214cb4d06352c2.
Now the parameters of the function have the correct type `(int32_t, int16_t, int16_t)` instead of `(uint32_t, uint32_t, uint32_t)`, and `memcpy`s have been introduced to have a view of the memory of type `int32_t`, which is the type of the `__builtin_arm_smla*` builtins.

NicolaLancellotti:
@guberti can you review it, please?

NicolaLancellotti:
@ashutosh-arm shall we close https://github.com/apache/tvm/pull/12846?

python/tvm/topi/arm_cpu/mprofile/dsp/micro_kernel/multi_channel_convolve.py
```diff
@@ -179,20 +179,26 @@ def _dual_int16_channel_convolve_impl(_tensor_h, tensor_w, channels, kernel_h, k
...
+            int16_t *tensor,
+            int16_t *kernel) {{
 
-          uint32_t sum_c0 = 0;
-          uint32_t sum_c1 = 0;
+          int32_t sum_c0 = 0;
+          int32_t sum_c1 = 0;
+
+          int32_t kernel_i32[{kernel_h} * {kernel_w}];
+          memcpy(kernel_i32, kernel, {kernel_h} * {kernel_w} * 4);
```
ashutosh-arm, line 190: nit: can we please add `sizeof()` instead of using 4 directly?

ashutosh-arm: Same applies to the code below where 2 and 4 are used.

python/tvm/topi/arm_cpu/mprofile/dsp/micro_kernel/multi_channel_convolve.py
```diff
@@ -179,20 +179,26 @@ def _dual_int16_channel_convolve_impl(_tensor_h, tensor_w, channels, kernel_h, k
...
+
+          int32_t tensor_length = {((kernel_w - 1) * (channels // 2) + (kernel_h - 1) * tensor_w * (channels // 2)) + 1};
```
ashutosh-arm, line 192: Curious why are braces used in this piece of logic?

tests/python/relay/aot/corstone300.mk
```diff
@@ -64,7 +64,7 @@ PKG_CFLAGS = ${PKG_COMPILE_OPTS} \
 CMAKE_FLAGS = -DCMAKE_TOOLCHAIN_FILE=${TVM_ROOT}/tests/python/contrib/test_ethosu/reference_system/arm-none-eabi-gcc.cmake \
 	-DCMAKE_SYSTEM_PROCESSOR=${MCPU}
 
-PKG_LDFLAGS = -lm -specs=nosys.specs -static -T ${AOT_TEST_ROOT}/corstone300.ld
+PKG_LDFLAGS = -lm -specs=nosys.specs -static -Wl,--gc-sections -T ${AOT_TEST_ROOT}/corstone300.ld
```
ashutosh-arm, line 67: maybe add a comment why `gc-section` and `fdata-sections` were used for the build?

ashutosh-arm (changes requested):
Overall LGTM! Thanks @NicolaLancellotti

python/tvm/topi/arm_cpu/mprofile/dsp/micro_kernel/multi_channel_convolve.py
NicolaLancellotti: Because here the `tensor_length` value is computed in Python, do we prefer to do it in C?

python/tvm/topi/arm_cpu/mprofile/dsp/micro_kernel/multi_channel_convolve.py
NicolaLancellotti: `sizeof(int32_t)` instead of `4` makes sense, but why `2`?

ashutosh-arm: I guess the same reason applies. For clarity, we could show the derivation of constants here.

python/tvm/topi/arm_cpu/mprofile/dsp/micro_kernel/multi_channel_convolve.py
ashutosh-arm: The formulae don't look wrong. But it does not look like a C syntax. So, if possible imho, we should drop braces.

python/tvm/topi/arm_cpu/mprofile/dsp/micro_kernel/multi_channel_convolve.py
ashutosh-arm: my bad - `2` is part of the compute itself as clarified by @NicolaLancellotti offline.

python/tvm/topi/arm_cpu/mprofile/dsp/micro_kernel/multi_channel_convolve.py
ashutosh-arm: As discussion offline, I understand this piece now. The code is still in python.

tests/python/relay/aot/corstone300.mk
NicolaLancellotti: Done

ashutosh-arm (approved):
LGTM. Thanks for addressing concerns. This change will align TVM's tests with TFLM strategy of using `ofast` and only files / functions that are needed to be part of the final binary.

python/tvm/topi/arm_cpu/mprofile/dsp/micro_kernel/multi_channel_convolve.py
guberti: Yep, I agree this formula is correct.

guberti (approved):
LGTM! Have we confirmed these changes improve performance?

python/tvm/topi/arm_cpu/mprofile/dsp/micro_kernel/multi_channel_convolve.py
```diff
@@ -179,20 +179,26 @@ def _dual_int16_channel_convolve_impl(_tensor_h, tensor_w, channels, kernel_h, k
...
+            int16_t *tensor,
+            int16_t *kernel) {{
 
-          uint32_t sum_c0 = 0;
-          uint32_t sum_c1 = 0;
+          int32_t sum_c0 = 0;
+          int32_t sum_c1 = 0;
+
+          int32_t kernel_i32[{kernel_h} * {kernel_w}];
+          memcpy(kernel_i32, kernel, {kernel_h} * {kernel_w} * sizeof(int32_t));
```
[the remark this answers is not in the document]
NicolaLancellotti: Hi Gavin, this is a functional change to remove the type punning error.  So performance can be worse.
But, it is not right to rely on undefined behaviours to have better performance.
Do you know any other way to address this undefined behaviour, without the `memcpy`s?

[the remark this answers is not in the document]
ashutosh-arm: If there is no alternative way to fix type punning without memcpy, we need to update other schedules too  :smile_cat:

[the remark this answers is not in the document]
guberti: Perhaps I'm not understanding - why can't we just do `static_cast`?

python/tvm/topi/arm_cpu/mprofile/dsp/micro_kernel/multi_channel_convolve.py
```diff
@@ -179,20 +179,26 @@ def _dual_int16_channel_convolve_impl(_tensor_h, tensor_w, channels, kernel_h, k
         extern "C"
         #endif
         int32_t {_get_func_name("int16", tensor_w, channels, kernel_h, kernel_w, suffix)}(
```
guberti, line 181: Can we also make these changes for the `int8` version of this function in the lines above?

python/tvm/topi/arm_cpu/mprofile/dsp/micro_kernel/multi_channel_convolve.py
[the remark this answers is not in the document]
gromero: @NicolaLancellotti Is it the `size` of `memcpy` correct here? It seems that `kernel` is type `int16_t`, not `int32_t`, so seems copying beyond `kernel` end?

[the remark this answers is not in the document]
Mousius: @guberti these schedules are compiled as C not C++ aren't they? Also, last time I checked the output of one of these type punning fixes, the memcpy was optimised into two loads which can work unaligned, rather than one load of two words which only works if properly aligned - still faster than a copy 😸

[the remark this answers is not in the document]
guberti: Oops, I'm being silly here. This seems fine to me - thanks @Mousius for checking.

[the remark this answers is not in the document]
NicolaLancellotti: @gromero I've checked and `kernel_i32` has `kernel_h * kernel_w` elements of `int32_t`, the `int16_t` kernel has twice as many elements. So the size is correct, (with a lower value it doesn't work, I checked it too).

NicolaLancellotti:
@tvm-bot rerun

NicolaLancellotti:
I've addressed the comments. Ping for a review.

PR #13816, opened 2023-01-20
Extend the USE_LIBBACKTRACE option

yelite:
This PR extends the value accepted by USE_LIBBACKTRACE cmake variable. It adds ability to search pre-compile libbacktrace from system path or user-provided path. The behavior of this variable will be:
- OFF: Don't use libbacktrace
- ON: Find libbacktrace from system paths. Fail the build generation if libbacktrace is not found.
- COMPILE: Build and link to libbacktrace from 3rdparty/libbacktrace. 
- \<PATH\>: Looking for the libbacktrace header and static lib from a user-provided path. Fail the build generation if libbacktrace is not found.
- AUTO: 
  - Find libbacktrace from system paths.
  - If not found, fallback to COMPILE on Linux or MacOS, fallback to OFF on Windows or other platforms.

cc @junrushao


junrushao:
CC: @Hzfengsy

tests/scripts/task_config_build_cpu.sh
```diff
@@ -54,7 +54,7 @@ echo set\(USE_ETHOSN_HW OFF\) >> config.cmake
 echo set\(USE_CMSISNN OFF\) >> config.cmake
 echo set\(USE_VITIS_AI ON\) >> config.cmake
 echo set\(USE_VERILATOR ON\) >> config.cmake
-echo set\(USE_LIBBACKTRACE ON\) >> config.cmake
+echo set\(USE_LIBBACKTRACE COMPILE\) >> config.cmake
```
Hzfengsy, line 57: Could you please update the comments in `config.cmake`?

yelite: It's updated. Also removed the comments from Logging.cmake to avoid duplication.

Hzfengsy (approved):

PR #13818, opened 2023-01-20
[microTVM] Update Zephyr version and Zephyr SDK version

mehrdadh:
This PR updates the microTVM code to use Zephyr 3.2 and SDK 0.15.2. Az a result of this change, there are few other changes that are included in this PR:

- this PR adds a launch script for Zephyr and Arduino template project to run the Project API server from a different Python ENV than TVM Python ENV.
- For Zephyr, launch script uses global python3.8 which is where `west` is registered. However, for Arduino it uses a separate virtual ENV with python3 version that exists in the host.
- removes `tests/micro/project_api/test_arduino_microtvm_api_server.py` since these tests where using arduino microtvm api server by importing it to TVM. We no longer support Arduino/Zephyr dependencies in TVM testing python ENV.
- Disables a demo and test due to a CMSIS-NN bug: https://github.com/apache/tvm/issues/13856

There would be a follow up work to move zephyr to a completely separate virtual ENV.


mehrdadh:
@tvm-bot rerun

apps/microtvm/arduino/template_project/launch_microtvm_api_server.sh
```diff
@@ -0,0 +1,43 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+function show_usage() {
+    cat <<EOF
+This script is for running microtvm_api_server.
```
alanmacd, line 21: nit:
```suggestion
This script is for running microtvm_api_server with arduino.
```

apps/microtvm/zephyr/template_project/launch_microtvm_api_server.sh
```diff
@@ -0,0 +1,41 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+function show_usage() {
+    cat <<EOF
+This script is for running microtvm_api_server.
```
alanmacd, line 21: nit:
```suggestion
This script is for running microtvm_api_server with zephyr.
```

alanmacd (approved):

tests/micro/project_api/test_project_api.py
```diff
@@ -40,15 +42,36 @@
```
mkatanbaf, line 42: Is there any test that shows how to pass the west_cmd if one doesn't use the default one? If there is not, it would be nice to add one.

mkatanbaf (approved):

apps/microtvm/arduino/template_project/launch_microtvm_api_server.sh
```diff
@@ -0,0 +1,43 @@
...
+Usage: launch_microtvm_api_server.sh <microtvm_api_server.py> --read-fd <READ_FD_PATH> --write-fd <WRITE_FD_PATH>
+EOF
+}
+
+if [ "$#" -lt 5 -o "$1" == "--help" ]; then
+    show_usage
+    exit -1
+fi
+
+ARDUINO_VENV_PATH=${HOME}/.tvm/micro_arduino
```
gromero, line 31: Why Arduino code is being touched in this PR which is related to a Zephyr / Zephyr SDK update? I think it would be better to summit Arduino changes on separate PR.

apps/microtvm/zephyr/template_project/launch_microtvm_api_server.sh
```diff
@@ -0,0 +1,41 @@
...
+
+if [ "$#" -lt 5 -o "$1" == "--help" ]; then
+    show_usage
+    exit -1
+fi
+
+west_file_path=$(which west)
+
+# Remove space and extra characters
+line=$(head -n 1 ${west_file_path})
```
gromero, line 34: I don't recall different `west` script version request in its #! a Python binary which includes the minor version, like `#!/usr/bin/python3.8`, which we can't just do `which python3` here instead?

apps/microtvm/zephyr/template_project/microtvm_api_server.py
```diff
@@ -68,9 +68,14 @@
 
 # Used to check Zephyr version installed on the host.
 # We only check two levels of the version.
-ZEPHYR_VERSION = 2.7
+ZEPHYR_VERSION = 3.2
 
-WEST_CMD = default = sys.executable + " -m west" if sys.executable else None
+try:
+    PYTHON_CMD = subprocess.check_output(["which", "python3.8"], encoding="utf-8").strip("\n")
```
gromero, line 74: How about using `shutl.which()` instead of `check_output()`?

apps/microtvm/zephyr/template_project/microtvm_api_server.py
```diff
@@ -680,12 +695,7 @@ def build(self, options):
                 st.st_mode | stat.S_IEXEC,
             )
 
-        check_call(["cmake", "-GNinja", ".."], cwd=BUILD_DIR, env=env)
-
-        args = ["ninja"]
-        if verbose:
-            args.append("-v")
-        check_call(args, cwd=BUILD_DIR, env=env)
+        check_call(options["west_cmd"].split(" ") + ["build"], cwd=API_SERVER_DIR, env=env)
```
gromero, line 698: Are you using `split()` to just implicitly transform the `options["west_cmd"]` into a list or there is any case that such a option will have a string with a space in it?

gromero (changes requested):
@mehrdadh Thanks for this PR bumping Zephyr, that's great.

I did a first pass and left a few comments but I need more time to do the second pass.

Overall the Zephyr / Zephyr SDK looks good.

Question: do you really want to remove [tests/micro/project_api/test_arduino_microtvm_api_server.py](https://github.com/apache/tvm/pull/13818/files#diff-04b94386da4afb92f9197d3d59252a300abefbd46cecf551fe868dddaee08f4f)  ? If so, could you elaborate a bit more on what you meant by "removed test since it was importing microtvm_api_server to tvm" ? Also, see my comments on touching Arduino code in this PR. I think ideally the Arduino parts should be left to a separate PR.

apps/microtvm/arduino/template_project/launch_microtvm_api_server.sh
mehrdadh: Initially I didn't plan for this, but since we are testing both in the same image and the new image has different Python env I had to make this changes. Would it be okay to update the PR title and description to cover all these changes?

apps/microtvm/zephyr/template_project/launch_microtvm_api_server.sh
mehrdadh: what we are trying to do here is to reuse the python that west was installed in and separate that from TVM python ENV. The ideal solution is to create a virtual python ENV for zephyr, install west in it and use that. That seems a bit complicated, so I decided to postpone that for now. Currently, Zephyr is installed in the global python3.8 vs TVM which is in a pythoon3.7 virtual ENV

apps/microtvm/zephyr/template_project/microtvm_api_server.py
mehrdadh: done

apps/microtvm/zephyr/template_project/microtvm_api_server.py
mehrdadh: it's because west could be in this format: `python3.8 -m west`

gallery/how_to/work_with_microtvm/install_zephyr.rst
```diff
@@ -34,19 +34,18 @@ Install Zephyr
...
-        ZEPHYR_SDK_FILE="/content/zephyr-sdk-linux-setup.run"
-        wget --no-verbose -O $ZEPHYR_SDK_FILE \
-            https://github.com/zephyrproject-rtos/sdk-ng/releases/download/v${ZEPHYR_SDK_VERSION}/zephyr-sdk-${ZEPHYR_SDK_VERSION}-linux-x86_64-setup.run
-        chmod +x $ZEPHYR_SDK_FILE
-        "$ZEPHYR_SDK_FILE" -- -d /content/zephyr-sdk --quiet
+        cd /content
+        wget https://github.com/zephyrproject-rtos/sdk-ng/releases/download/v0.15.2/zephyr-sdk-0.15.2_linux-x86_64.tar.gz
+        tar xvf zephyr-sdk-0.15.2_linux-x86_64.tar.gz
+        mv zephyr-sdk-0.15.2 zephyr-sdk
+        rm zephyr-sdk-0.15.2_linux-x86_64.tar.gz
```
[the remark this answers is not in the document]
mehrdadh: done

tests/micro/project_api/test_project_api.py
mehrdadh: `west_cmd` can be passed as a project option, we also had option to pass it as pytest argument before but since project api uses the west from ENV, we removed that

apps/microtvm/zephyr/template_project/launch_microtvm_api_server.sh
gromero: I see.

So, a side effect now is that when `PYTHON_CMD` is set based on `which("python3.8")` if the user has no python3,8 installed in the env command like that use the Project API server, like `tvmc micro create-project`, will quite saying that `Error: Option 'west_cmd' is required but was not specified. Use --list-options to see all required options.`, which is misleading. 

Hence I wonder if the right env isolation should actually happen in the launcher, i.e the launcher would do the check for the right Python version -- 3.8 -- (and later the virtual env), whilst the Project API server would continue to use `sys.executable` to find out the west module.

gromero: Also, it doesn't seem right to me to check the python used in the `west` shebang and use it to start the Project API server.

apps/microtvm/arduino/template_project/launch_microtvm_api_server.sh
gromero: Ok, makes sense. I think it's ok if you just explain it in the description and leave the title as is.

apps/microtvm/arduino/template_project/microtvm_api_server.py
```diff
@@ -28,7 +28,7 @@
 from string import Template
 from packaging import version
 
-from tvm.micro.project_api import server
+import server
```
gromero, line 31: Could you please clarify why this change is necessary? As a consequence now `server.py` is being copied to the new project dir too right?

gromero (changes requested):

alanmacd (approved):

apps/microtvm/zephyr/template_project/launch_microtvm_api_server.sh
```diff
@@ -0,0 +1,41 @@
...
+line=$(echo ${line} | sed 's/ //')
```
gromero, line 35: btw, for stripping the shebang here you could simplify it with something like:

```
$ sed 's/#!//; q' ~/.local/bin/west
/usr/bin/python3
```

gromero (changes requested):

apps/microtvm/arduino/template_project/launch_microtvm_api_server.sh
gromero: ah, and why it's necessary to remove the `tests/micro/project_api/test_arduino_microtvm_api_server.py` ? Is that right? That test is going away? For what reason?

mehrdadh: This test file was fully based on importing the project api server to TVM which it doesn't work now because of dependencies. Also I checked with @guberti who wrote this test and it doesn't seem to be necessary to keep it anymore.

apps/microtvm/arduino/template_project/microtvm_api_server.py
mehrdadh: Yes, we are copying server.py as well to remove dependency of project api server to TVM.

apps/microtvm/zephyr/template_project/launch_microtvm_api_server.sh
mehrdadh: `sys.executable` doesn't work since it will return tvm python env in docker. I think it should be python3.8 specifically and later we change it to use the python ENV that we use to setup Zephyr/west. Let me know what you think.

apps/microtvm/zephyr/template_project/launch_microtvm_api_server.sh
mehrdadh: Let me try this.

apps/microtvm/zephyr/template_project/launch_microtvm_api_server.sh
gromero: Yep, I got it. I think that it would be right to adjust the python version to be used in the API server inside the launcher, instead of hardcoding the requested version inside the Project API server. Also this way `sys.executable` would return the right version, no?

apps/microtvm/arduino/template_project/microtvm_api_server.py
gromero: hm I don't follow it. Isn't the client (client.py) always use too when connecting to the server and client is actually called from the tvm.micro.project_api anyways?

apps/microtvm/zephyr/template_project/launch_microtvm_api_server.sh
mehrdadh: I changed it.

apps/microtvm/arduino/template_project/microtvm_api_server.py
mehrdadh: client starts the microtvm API server, previously it was using sys.executable to directly run microtvm_api_server.py file, however that mean it was using a global python ENV which is the same for any project api server. To distinguish between different project servers and their python requirements we have an alternative to run a bash script to start the sever. This was originally part of the project API design but never used.

apps/microtvm/zephyr/template_project/launch_microtvm_api_server.sh
mehrdadh: no, if we you use `sys.executable` it will still pickup the default global Python version. We should change it to ENV later.

gromero: If you call python3.8 from the launcher to launch the server will the server (i.e. `sys.executable` called from the server) still report the global Python version -- which I'm assuming is Python 3.7?

apps/microtvm/arduino/template_project/microtvm_api_server.py
gromero: @mehrdadh I understand all that and I agree that this was exactly the originally need for the project API client/server design. What I was questioning here is why it was necessary to use `server.py` since it was common/generic code for the server in TVM. But I see now that `microtvm_api_server.py`  imports `server.py` also for isolated env, so the python env in this case doesn't have tvm.micro in its env. ok.

apps/microtvm/zephyr/template_project/launch_microtvm_api_server.sh
mehrdadh: no, in that case it would be python3.8. I think you're right, we can keep using sys.executable since it would follow the Python ENV that called the python script

gromero (approved):
Thanks @mehrdadh LGTM now :+1:

guberti (approved):
LGTM!

gromero:
@alanmacd @mkatanbaf @guberti Thanks for the reviews. @mehrdadh Thanks a lot for the change. Kudos for the awesome PR description! :-)

PR #13819, opened 2023-01-20
[TVMScript] Implicit root block syntax sugar for TVMScript printer

cyx-6:
This PR implements the syntax sugar of implicit root block for new TVMScript printer. This syntax sugar will skip the `T.block("root")`, when the root block realize is simple and we shall reconstruct that root block via `tvm::tir::ScriptComplete` when roundtripping. For example, it will change
```python
@T.prim_func
def root_block_explicitly():
  with T.block("root"):
    a = T.alloc_buffer([128, 128])
    for i, j in T.grid(128, 128):
      with T.block():
        T.evaluate(0)
```
into
```python
@T.prim_func
def main():
  a = T.alloc_buffer((128, 128))
  for i, j in T.grid(128, 128):
    with T.block(""):
      T.evaluate(0)
```

### DIFF
--- a/tests/python/unittest/test_tvmscript_printer_tir.py
+++ b/tests/python/unittest/test_tvmscript_printer_tir.py
@@ -717,20 +717,48 @@ def test_remap():
 
     expected_output = """@T.prim_func
 def main():
-    with T.block("root"):
-        T.reads()
-        T.writes()
-        for i0, i1, i2, i3, i4, i5 in T.grid(128, 128, 128, 128, 128, 128):
-            with T.block("update"):
-                v0 = T.axis.spatial(128, i0 + 1)
-                v1, v2 = T.axis.remap("SR", [i1, i2])
-                v3 = T.axis.spatial(128, i3 - 1)
-                v4, v5 = T.axis.remap("RS", [i4, i5])
-                T.reads()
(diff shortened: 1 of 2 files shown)


junrushao (approved):
LGTM!

tests/python/unittest/test_tvmscript_printer_tir.py
```diff
@@ -717,21 +717,49 @@ def block_with_remap_explicitly():
...
+    @T.prim_func
+    def root_block_explicitly():
+        with T.block("root"):
+            a = T.alloc_buffer([128, 128])
+            for i, j in T.grid(128, 128):
+                with T.block():
+                    T.evaluate(0)
+
+    expected_output = """@T.prim_func
+def main():
```
junrushao, line 752: Could we add a line of comment saying the root block is sugared? For example:

```python
@T.prim_func
def main():
    # The root block is sugared: with T.block("root"):
    a = T.alloc_buffer((128, 128))
    for i, j in T.grid(128, 128):
        with T.block(""):
            T.reads()
            T.writes()
            T.evaluate(0)
```

cyx-6: sure

PR #13823, opened 2023-01-23
[Docker] Add script to build llvm from source

neildhickey:
This allows us to use the latest llvm and make sure it builds with the same configuration as the rest of tvm

Change-Id: I004ec89700498167c632d25f4fdf667922ba4943

### DIFF
--- a/docker/Dockerfile.ci_arm
+++ b/docker/Dockerfile.ci_arm
@@ -50,8 +50,8 @@ COPY install/ubuntu_install_sccache.sh /install/ubuntu_install_sccache.sh
 RUN bash /install/ubuntu_install_sccache.sh
 ENV PATH /opt/sccache:$PATH
 
-COPY install/ubuntu_install_llvm.sh /install/ubuntu_install_llvm.sh
-RUN bash /install/ubuntu_install_llvm.sh
+COPY install/ubuntu_install_llvm_from_source.sh /install/ubuntu_install_llvm_from_source.sh
+RUN bash /install/ubuntu_install_llvm_from_source.sh 15.0
 
 ENV TVM_VENV /venv/apache-tvm-py3.7
 COPY python/bootstrap/lockfiles /install/python/bootstrap/lockfiles
--- a/docker/Dockerfile.ci_arm
+++ b/docker/Dockerfile.ci_arm
@@ -51,7 +51,7 @@ RUN bash /install/ubuntu_install_sccache.sh
 ENV PATH /opt/sccache:$PATH
 
 COPY install/ubuntu_install_llvm_from_source.sh /install/ubuntu_install_llvm_from_source.sh
-RUN bash /install/ubuntu_install_llvm_from_source.sh 15.0.7
+RUN bash /install/ubuntu_install_llvm_from_source.sh 15.0.7 8b5fcb24b4128cf04df1b0b9410ce8b1a729cb3c544e6da885d234280dedeac6
 
 ENV TVM_VENV /venv/apache-tvm-py3.7
 COPY python/bootstrap/lockfiles /install/python/bootstrap/lockfiles
--- a/docker/Dockerfile.ci_arm
+++ b/docker/Dockerfile.ci_arm
@@ -51,7 +51,7 @@ RUN bash /install/ubuntu_install_sccache.sh
 ENV PATH /opt/sccache:$PATH
 
 COPY install/ubuntu_install_llvm_from_source.sh /install/ubuntu_install_llvm_from_source.sh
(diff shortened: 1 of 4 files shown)


ekalda:
@tvm-bot rerun

mehrdadh:
@neildhickey what is the benefit of building from source if we are using the released version anyway?

neildhickey:
@mehrdadh Newer versions of gcc and clang introduce an option to outline atomic calls which means it will look for these atomic functions calls in glibc. The atomic calls are of the form __aarch64_ldadd4_acq_rel and others.

With the version of gcc and glibc we use to build tvm, these functions aren't available, and the build fails when trying to build against prebuilt libraries which use these options. 

The released llvm packages for llvm-15 are built with these options enabled, so a workaround is to build llvm from source using the same version of the compiler we use to build tvm, those alleviating the of the undefined function references.

docker/install/ubuntu_install_llvm_from_source.sh
```diff
@@ -0,0 +1,99 @@
...
+# under the License.
+# This script builds LLVM and clang from the llvm-project tarball
+# using CMake. It is tested with LLVM from version 15.
+
+set -e
+
+LLVM_VERSION_MAJOR=$1
+
+detect_llvm_version() {
+  curl -sL "https://api.github.com/repos/llvm/llvm-project/releases?per_page=100" | \
```
Mousius, line 27: Do we need to follow redirects here?

docker/install/ubuntu_install_llvm_from_source.sh
```diff
@@ -0,0 +1,99 @@
...
+}
+
+LLVM_VERSION=$(detect_llvm_version)
+echo ${LLVM_VERSION}
+
+curl -sL \
+  https://github.com/llvm/llvm-project/releases/download/llvmorg-${LLVM_VERSION}/llvm-project-${LLVM_VERSION}.src.tar.xz \
+  -o llvm-project-${LLVM_VERSION}.src.tar.xz
+unxz llvm-project-${LLVM_VERSION}.src.tar.xz
+tar xf llvm-project-${LLVM_VERSION}.src.tar
```
Mousius, line 42: Does this not work? Surprised `tar` doesn't sniff the format.

```suggestion
tar xf llvm-project-${LLVM_VERSION}.src.tar.xz
```

Mousius (changes requested):

docker/install/ubuntu_install_llvm_from_source.sh
```diff
@@ -0,0 +1,99 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+# This script builds LLVM and clang from the llvm-project tarball
+# using CMake. It is tested with LLVM from version 15.
+
+set -e
+
+LLVM_VERSION_MAJOR=$1
```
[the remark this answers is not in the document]
leandron: I agree, and if we remove the $1 parameter, then should we also cleanup the `detect_llvm_version` function? What do you think?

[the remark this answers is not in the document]
Mousius: We could still use the parameter if we wanted to, similar to:

https://github.com/apache/tvm/blob/6c04ac52bd43065156cabb609f76cd2323449d27/docker/Dockerfile.ci_adreno#L25

Otherwise, yes, `detect_llvm_version` should be removed.

docker/Dockerfile.ci_arm
```diff
@@ -50,8 +50,8 @@ COPY install/ubuntu_install_sccache.sh /install/ubuntu_install_sccache.sh
 RUN bash /install/ubuntu_install_sccache.sh
 ENV PATH /opt/sccache:$PATH
 
-COPY install/ubuntu_install_llvm.sh /install/ubuntu_install_llvm.sh
-RUN bash /install/ubuntu_install_llvm.sh
+COPY install/ubuntu_install_llvm_from_source.sh /install/ubuntu_install_llvm_from_source.sh
+RUN bash /install/ubuntu_install_llvm_from_source.sh 15.0.7
```
Mousius, line 54: Can we add a hash parameter in here to verify the source tarball? Similar to:
https://github.com/apache/tvm/blob/main/docker/install/ubuntu_install_zephyr_sdk.sh#L45-L48

Mousius (changes requested):

Mousius (approved):

PR #13824, opened 2023-01-23
[Metaschedule] get_top_k should not return not built records

Icemist:
Avoid the situation when the records have measurements equal to 1e10, which is used as a stub for an indefinite measurement time, for example, when forming a cost models of the meta scheduler. The result of this will be extracting a record(that fails while building) as the best and an error in applying the tuning.

### DIFF
--- a/src/meta_schedule/database/memory_database.cc
+++ b/src/meta_schedule/database/memory_database.cc
@@ -68,20 +68,20 @@ class MemoryDatabaseNode : public DatabaseNode {
     std::vector<std::pair<double, TuningRecord>> results;
     results.reserve(records.size());
     for (const TuningRecord& record : records) {
-      if (!record->run_secs.defined()) {
-        continue;
-      }
-      Array<FloatImm> run_secs = record->run_secs.value();
-      if (run_secs.empty()) {
+      auto run_secs = record->run_secs;
+      if (!run_secs.defined() || run_secs.value().empty() ||
+          std::all_of(run_secs.value().begin(), run_secs.value().end(),
+                      // 1e10 is used as a stub for undefined measurement times.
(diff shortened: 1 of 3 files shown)


Icemist:
CC @masahi @echuraev

src/meta_schedule/database/memory_database.cc
```diff
@@ -68,20 +68,20 @@ class MemoryDatabaseNode : public DatabaseNode {
...
-      if (!record->run_secs.defined()) {
-        continue;
-      }
-      Array<FloatImm> run_secs = record->run_secs.value();
-      if (run_secs.empty()) {
+      auto run_secs = record->run_secs;
+      if (!run_secs.defined() || run_secs.value().empty() ||
+          std::all_of(run_secs.value().begin(), run_secs.value().end(),
+                      // 1e10 is used as a stub for undefined measurement times.
+                      [](tvm::FloatImm v) { return v.defined() && v->value == 1e10; })) {
```
echuraev, line 75: Probably we can define this constant in some common place and reuse it here and in other places?
E.g. as it was done [here](https://github.com/apache/tvm/blob/main/src/meta_schedule/utils.h#L456).

echuraev (approved):
In general LGTM. One minor question. @Icemist what do you think about that?

src/meta_schedule/database/memory_database.cc
Icemist: I thought about using SortTuningRecordByMeanRunSecs, but memory_database is not sorted unlike json_database.
This constant appears in all tuners and unfortunately there is no common definition for it.
I can suggest to make an NFC patch with refactoring this constant throughout the repository.

masahi (approved):

Issue #12922, opened 2022-09-28
[Bug] Relay/TOPI `adv_index` not supporting `None` as arguments


shingjan:
The advanced indexing implementation of relay/topi is not able to handle cases that use None as one of the indexing argument.

### My repro under `test_forward_index`
```
    class Index2(Module):
        def forward(self, x):
            return x[None, [2, 2]]

    input_data = torch.rand([3, 3, 3, 3]).float().cpu()
    verify_model_with_input(Index2().eval(), [input_data])
```

### Expected behavior

A tensor shaped `[1, 2, 3, 3, 3]` shall be returned.

### Actual behavior
```
>           raise Exception("warning unhandled case: {0}".format(type(expr)))
E           Exception: warning unhandled case: <class 'NoneType'>

python/tvm/relay/expr_functor.py:76: Exception
```
### Environment

Latest TVM main

cc: @masahi

PR #13826, opened 2023-01-23
[Torch] Fix advanced indexing with NoneType index arguments

padreofthegame:
Suppose to solve issue #12922.

### DIFF
--- a/python/tvm/relay/frontend/pytorch.py
+++ b/python/tvm/relay/frontend/pytorch.py
@@ -2330,19 +2330,49 @@ class PyTorchOpConverter:
 
     def index(self, inputs, input_types):
         data = inputs[0]
-        indices_list = []
+        data_shape = self.infer_type(data).shape
 
-        for indices in inputs[1]:
-            if self.infer_type(indices).dtype == "bool":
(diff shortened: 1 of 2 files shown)


shingjan:
Thanks for taking a look at this @padreofthegame! Will circle back once this PR is ready for review

padreofthegame:
This solution is very similar to #13306, and represents workaround for the problem mentioned in #12922 for PyTorch models (since advanced indexing does not work the same for example for numpy and for pytorch, I put solution in torch frontend, because issue was reported for torch models). 
Also, I put WIP label only because I am not sure if I managed to cover all possible test cases, so @shingjan you are welcome to try some additional scenarios, and call me out if detect something strange.

shingjan:
@padreofthegame i think this fix is fine in terms of the difference between adv_index of numpy and it of pytorch. There is no numpy frontend for TVM yet. I think overall this PR LGTM. Will be great if @masahi can take a look and approve. Thanks!

shingjan (approved):
LGTM

masahi (approved):

PR #13829, opened 2023-01-23
[docker][microTVM]Fix Zephyr 0.15.2 SDK installation and separate Zephyr python environment

mehrdadh:
This PR fixes the SDK installation and separates SDK installation from ZephyrProject installation.
Also it separates the Zephyr python environment to 3.8 which is the required version and it is different than tvm virtual env.

### DIFF
--- a/docker/Dockerfile.ci_cortexm
+++ b/docker/Dockerfile.ci_cortexm
@@ -76,12 +76,15 @@ COPY install/ubuntu_install_sccache.sh /install/ubuntu_install_sccache.sh
 RUN bash /install/ubuntu_install_sccache.sh
 ENV PATH /opt/sccache:$PATH
 
-# Zephyr SDK deps
+# Zephyr Project
 COPY install/ubuntu_install_zephyr.sh /install/ubuntu_install_zephyr.sh
 COPY install/ubuntu_init_zephyr_project.sh /install/ubuntu_init_zephyr_project.sh
-COPY install/ubuntu_install_zephyr_sdk.sh /install/ubuntu_install_zephyr_sdk.sh
 RUN bash /install/ubuntu_install_zephyr.sh
 ENV ZEPHYR_BASE=/opt/zephyrproject/zephyr
+
+#Zephyr SDK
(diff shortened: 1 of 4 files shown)


docker/Dockerfile.ci_cortexm
```diff
@@ -76,12 +76,15 @@ COPY install/ubuntu_install_sccache.sh /install/ubuntu_install_sccache.sh
...
 
-# Zephyr SDK deps
+# Zephyr Project
 COPY install/ubuntu_install_zephyr.sh /install/ubuntu_install_zephyr.sh
 COPY install/ubuntu_init_zephyr_project.sh /install/ubuntu_init_zephyr_project.sh
-COPY install/ubuntu_install_zephyr_sdk.sh /install/ubuntu_install_zephyr_sdk.sh
 RUN bash /install/ubuntu_install_zephyr.sh
 ENV ZEPHYR_BASE=/opt/zephyrproject/zephyr
+
+#Zephyr SDK
```
driazati, line 85:
```suggestion
# Zephyr SDK
```

docker/install/ubuntu_install_zephyr.sh
```diff
@@ -58,10 +78,4 @@ chmod -R o+w ${ZEPHYR_PROJECT_PATH}
 mkdir zephyr/.cache
 chmod o+rwx zephyr/.cache
 
-pip3 install -r /opt/zephyrproject/zephyr/scripts/requirements.txt
-
-# the requirements above overwrite junintparser with an older version, but it is not
-# used so overwrite it again with the correct version
-pip3 install junitparser==2.4.2
```
driazati, line 65: Isn't this just using the system Python with `python_cmd` here? This should still re-install junitparser since that was a hack to get a specific version working in CI which didn't agree with zephyr's requirements

areusch: i think i might have removed pip3 to stop ppl from installing stuff outside the tvm venv

areusch (commented):
lgtm but will allow @driazati to approve

driazati (approved):

PR #13830, opened 2023-01-24
[TVMScript] More concise `T.allocate` syntax printing

cyx-6:
This PR is the follow up of #13813. We simplify the printing output of `T.allocate` with `T.decl_buffer`. For example, we have a code snippet as 

```python
buffer_data = T.allocate(...)
buffer = T.decl_buffer(..., data=buffer_data)
T.evaluate(buffer_data)
```

Originally, we skip the `T.allocate` only if the var `buffer_data` defined by `T.allocate` is used only once by the following `T.decl_buffer`. This was due to the limitation of the old printer design. 

But in the new printer, we may automatically replace the `buffer_data` with `buffer.data` if skipping the definition of `buffer_data`. We are able to link all `buffer_data` usages together. So the new output result will be like

```python
buffer = T.decl_buffer(...)
T.evaluate(buffer.data)
```


junrushao (approved):
LGTM!

PR #13831, opened 2023-01-24
[TVMScript] Introduce `PrinterConfig`

junrushao:
This PR introduces `PrinterConfig`, a systematic way to configure
TVMScript printer without having to set global flags.

This PR enables more customization of printer behavior. More
specifically, now any TVM’s object in python, as long as it
inherits from `Scriptable`, it automatically gains two methods:
- `.script(tir_prefix=...)`
- `.show(...)`


Hzfengsy (approved):

Mousius:
@junrushao was this meant for `unity`? it contains several references to `relax` in the code introduced. This PR wasn't left open for much time else it might've been caught during review, it would've been good to leave such a large change open a while longer. Could you please follow up and remove the `relax` references?

tqchen:
Thanks for getting to this thread. I think the particular set of changes won;'t cause technical debt that breaks the any code in the current mainline

But if there is a strong desire to remove(which i think is a nit TBH), we can do a followup PR.

PR #13832, opened 2023-01-24
[RUNTIME] Fix determination of big/little cores domains

elvin-n:
In case of three or more cores domains we starts to treat non little cores as big ones and offload work to them as well by default. This fix current modern schemes like 1-3-4 with 1 huge, 3 big cores and 4 little cores and causes tvm to use huge and big cores for inference


echuraev (approved):
LGTM. Thanks

junrushao (approved):
LGTM

PR #13834, opened 2023-01-24
[CLML][RUNTIME] Enable more ops in CLML runtime

krishnaraj36:
Enable the DepthToSpace and Resize bilinear operator in CLML runtime and bug fix in concat layer.

### DIFF
--- a/src/runtime/contrib/clml/clml_runtime.cc
+++ b/src/runtime/contrib/clml/clml_runtime.cc
@@ -430,6 +430,14 @@ class CLMLRuntime : public JSONRuntimeBase {
           auto out = CreateBinaryLayer(&layer_, node);
           this->layer_.storage_map.insert({nid, std::make_pair(out, node)});
           this->layer_.func_outs.push_back(out);
+        } else if ("nn.depth_to_space" == op_name) {
+          auto out = CreateDepthToSpaceLayer(&layer_, node);
+          this->layer_.storage_map.insert({nid, std::make_pair(out, node)});
+          this->layer_.func_outs.push_back(out);
+        } else if ("nn.upsampling" == op_name) {
+          auto out = CreateResizeLayer(&layer_, node);
+          this->layer_.storage_map.insert({nid, std::make_pair(out, node)});
+          this->layer_.func_outs.push_back(out);
         } else {
           LOG(FATAL) << "Unsupported op: " << op_name;
         }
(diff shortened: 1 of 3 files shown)


tests/python/contrib/test_clml/test_ops.py
```diff
@@ -574,5 +574,107 @@ def _verify(out, params, inputs):
...
+                "op": "kernel",
+            },
+        ]
+        verify_codegen(out, exp_codegen, device, params)
+
+    _verify(*(_get_model((1, 64, 8, 8), 4)))
+    _verify(*(_get_model((1, 64, 8, 8), 8)))
+
+
+@pytest.mark.parametrize("dtype", ["float32"])
```
srkreddy1238, line 626: Can we cover "float16" also here if supported on CLML & TVM too ?

src/runtime/contrib/clml/clml_runtime.cc
```diff
@@ -1151,13 +1159,14 @@ class CLMLRuntime : public JSONRuntimeBase {
...
     auto output = MakeCLMLTensorFromJSONNode(node, CL_TENSOR_LAYOUT_OPTIMAL_QCOM, cl_dtype);
+    cl_uint axis = std::stoi(node.GetAttr<std::vector<std::string>>("axis")[0]);
     cl_ml_tensor_qcom* concatInputs = new cl_ml_tensor_qcom[inputSize];
     for (int i = 0; i < inputSize; i++) {
       auto input = MakeCLMLTensorFromJSONEntry(node.GetInputs()[i], {},
                                                CL_TENSOR_LAYOUT_OPTIMAL_QCOM, cl_dtype);
       concatInputs[i] = input->tensor;
     }
-    cl_ml_op_concat_desc_qcom concatDesc = {1, (cl_uint)inputSize, cl_arithmetic_mode};
+    cl_ml_op_concat_desc_qcom concatDesc = {axis, (cl_uint)inputSize, cl_arithmetic_mode};
```
echuraev, line 1169: Could you please update `concat` test and check how it works for different axes?

tests/python/contrib/test_clml/test_ops.py
krishnaraj36: Yes I have added float16 test case as well.

src/runtime/contrib/clml/clml_runtime.cc
srkreddy1238: Till them probably we can limit it in frontend ```clml.py``` and let it take TVM path for ```axis != 1``` instead of failing at run time.

echuraev (approved):
LGTM. Thanks

srkreddy1238 (approved):

PR #13839, opened 2023-01-25
[TVMScript] Comments and docstrings printing

cyx-6:
This PR introduces the `CommentDoc` for comments printing and `DocStringDoc` for docstring printing. It enables to add free comments and docstring as `stmt` in printing, e.g.
```python
# comment 1
# comment 2
"""
docstring 1
docstring 2
"""
```
The free here means to not be bound to any `stmt`, but acts as a single `stmt`, similar to `ExprStmtDoc` for `ExprDoc`. 

This PR also introduces an example for the `CommentDoc`, as follow up of #13819.
In the old printer, we always print a `# with T.block("root"):`, when there is an implicit root block skipped when printing. For example,
```
@T.prim_func
def main():
  # with T.block("root"):
  a = T.alloc_buffer((128, 128))
  for i, j in T.grid(128, 128):
    with T.block(""):
      ...
```
We bring this syntax reminder back in this PR.
In addition, we introduce a field of `ir_usage` and `print_headers` into the printer configuration, to support the printing of headers for `IRModule` and `PrimFunc`. For example,

```python

@I.ir_module
class Module():
  @T.prim_func
  def func():
    ...
```

### DIFF
--- a/src/script/printer/doc_printer/python_doc_printer.cc
+++ b/src/script/printer/doc_printer/python_doc_printer.cc
@@ -169,6 +169,8 @@ class PythonDocPrinter : public DocPrinter {
   void PrintTypedDoc(const ScopeDoc& doc) final;
   void PrintTypedDoc(const FunctionDoc& doc) final;
   void PrintTypedDoc(const ClassDoc& doc) final;
+  void PrintTypedDoc(const CommentDoc& doc) final;
+  void PrintTypedDoc(const DocStringDoc& doc) final;
 
  private:
   void NewLineWithoutIndent() { output_ << "\n"; }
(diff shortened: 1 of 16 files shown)


junrushao (commented):
Thanks for brining in this change!

BTW, for IRModule and PrimFunc printing, would you like to add a couple of lines of comments for clarify:

```python
# from tvm.script import ir as I      # if `IR(...)` is ever used
# from tvm.script import tir as T   # if `TIR(...)` is ever used

@I.ir_module
class Module()
```

src/script/printer/doc_printer/python_doc_printer.cc
```diff
@@ -642,11 +652,40 @@ void PythonDocPrinter::PrintTypedDoc(const ClassDoc& doc) {
...
+    }
+    if (cfg->ir_usage.count("relax")) {
+      headers.push_back(CommentDoc("from tvm.script import relax as " + cfg->relax_prefix));
+    }
+    if (headers.size()) {
+      // Add a blank line between headers and script codes
+      headers.push_back(CommentDoc(""));
+    }
+    printer.Append(StmtBlockDoc(headers));
+  }
```
junrushao, line 688: We should move this logic to the caller site :-)

junrushao (approved):
Perfect. Please address my comment above and fix the lint, and we are good to go!

PR #13844, opened 2023-01-25
[Hexagon] Detect and flush "dirty" cache prior to DMA read with cache bypass enabled

adstraw:
Add software cache management to enable DMA with cache bypass enabled.  DMA with cache bypass is an experimental feature requiring software management of the cache.   DMA with cache bypass enabled assumes that HexagonBuffer objects are not cached unless explicitly modified by the primfunc.  This PR adds cache flush and invalidation after HexagonBuffer allocation with `malloc` or copy with `memcpy` to uphold that assumption.  In addition, this PR adds logic to flush and invalidate the cache prior to a DMA with cache bypass enabled when the buffer in question has been modified by the primfunc.  The `test_matmul.py` test hits this case by performing layout transforms in global address space ahead of a DMA.  CC @csullivan with thanks for providing the test in question.

### DIFF
--- a/src/runtime/hexagon/hexagon_buffer.cc
+++ b/src/runtime/hexagon/hexagon_buffer.cc
@@ -49,6 +49,16 @@ struct Allocation {
 struct DDRAllocation : public Allocation {
   DDRAllocation(size_t nbytes, size_t alignment) : Allocation(nbytes, alignment) {
     int ret = posix_memalign(&data_, alignment, nbytes);
+
+    // The heap used by malloc on Hexagon is always mapped as cacheable. The heap manager may
+    // not perform cache flush and invalidation on a prior memory free. So, a subsequent memory
+    // allocation request to the heap manager may allocate memory that resides in part or in full in
+    // the cache. Hence, we must flush and invalidate the allocation from the cache to ensure that
+    // DMA with cache bypass enabled will function properly. DMA with cache bypass enabled assumes
+    // that HexagonBuffer objects are not cached unless explicitly modified by the primfunc. We must
+    // flush and invalidate after malloc to uphold this assumption.
+    qurt_mem_cache_clean(reinterpret_cast<qurt_addr_t>(data_), nbytes,
+                         QURT_MEM_CACHE_FLUSH_INVALIDATE, QURT_MEM_DCACHE);
     CHECK_EQ(ret, 0);
   }
   ~DDRAllocation() { free(data_); }
@@ -235,19 +245,20 @@ void hexagon_buffer_copy_across_regions(const BufferSet& dest, const BufferSet& 
 
   // Finally, do the memory copies.
   for (const auto& copy : macro_copies) {
-    // clean Hexagon cache before / after memcpy to ensure clean cache state to enable usage of DMA
-    // bypass mode for increased DMA bandwidth
     // TODO(HWE): Switch to ION Buffer to avoid need for memcpy and potentially lighten or alleviate
     // the burden of cache invalidation in this code
-    qurt_mem_cache_clean(reinterpret_cast<qurt_addr_t>(copy.dest), copy.num_bytes,
-                         QURT_MEM_CACHE_INVALIDATE, QURT_MEM_DCACHE);
-    qurt_mem_cache_clean(reinterpret_cast<qurt_addr_t>(copy.src), copy.num_bytes,
-                         QURT_MEM_CACHE_INVALIDATE, QURT_MEM_DCACHE);
     memcpy(copy.dest, copy.src, copy.num_bytes);
+
+    // We must flush and invalidate both the destination and source of the memcpy from the cache to
+    // ensure that DMA with cache bypass enabled will function properly. DMA with cache bypass
+    // enabled assumes that HexagonBuffer objects are not cached unless explicitly modified by the
+    // primfunc. We must flush and invalidate after the memcpy to uphold this assumption. Note that
+    // we need not flush and invalidate prior to the memcpy because it is coherent - issued through
+    // the cache.
     qurt_mem_cache_clean(reinterpret_cast<qurt_addr_t>(copy.dest), copy.num_bytes,
-                         QURT_MEM_CACHE_INVALIDATE, QURT_MEM_DCACHE);
+                         QURT_MEM_CACHE_FLUSH_INVALIDATE, QURT_MEM_DCACHE);
     qurt_mem_cache_clean(reinterpret_cast<qurt_addr_t>(copy.src), copy.num_bytes,
-                         QURT_MEM_CACHE_INVALIDATE, QURT_MEM_DCACHE);
+                         QURT_MEM_CACHE_FLUSH_INVALIDATE, QURT_MEM_DCACHE);
   }
 }
 
--- a/src/runtime/hexagon/hexagon_buffer.cc
+++ b/src/runtime/hexagon/hexagon_buffer.cc
@@ -234,7 +234,8 @@ std::vector<MemoryCopy> MemoryCopy::MergeAdjacent(std::vector<MemoryCopy> micro_
 }
 
 void hexagon_buffer_copy_across_regions(const BufferSet& dest, const BufferSet& src,
-                                        size_t bytes_to_copy) {
+                                        size_t bytes_to_copy, bool src_is_hexbuff,
+                                        bool dest_is_hexbuff) {
   // First, determine all copies that do not cross boundaries in
   // either source or destination region.
   auto micro_copies = BufferSet::MemoryCopies(dest, src, bytes_to_copy);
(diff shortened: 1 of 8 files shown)


src/runtime/hexagon/hexagon_buffer.cc
```diff
@@ -49,6 +49,16 @@ struct Allocation {
...
+
+    // The heap used by malloc on Hexagon is always mapped as cacheable. The heap manager may
+    // not perform cache flush and invalidation on a prior memory free. So, a subsequent memory
+    // allocation request to the heap manager may allocate memory that resides in part or in full in
+    // the cache. Hence, we must flush and invalidate the allocation from the cache to ensure that
+    // DMA with cache bypass enabled will function properly. DMA with cache bypass enabled assumes
+    // that HexagonBuffer objects are not cached unless explicitly modified by the primfunc. We must
+    // flush and invalidate after malloc to uphold this assumption.
+    qurt_mem_cache_clean(reinterpret_cast<qurt_addr_t>(data_), nbytes,
+                         QURT_MEM_CACHE_FLUSH_INVALIDATE, QURT_MEM_DCACHE);
```
janetsc, line 61: Could this be invalidate only, since there is no data copied in this buffer yet?

src/runtime/hexagon/hexagon_buffer.cc
```diff
@@ -235,19 +245,20 @@ void hexagon_buffer_copy_across_regions(const BufferSet& dest, const BufferSet&
...
+    // enabled assumes that HexagonBuffer objects are not cached unless explicitly modified by the
+    // primfunc. We must flush and invalidate after the memcpy to uphold this assumption. Note that
+    // we need not flush and invalidate prior to the memcpy because it is coherent - issued through
+    // the cache.
     qurt_mem_cache_clean(reinterpret_cast<qurt_addr_t>(copy.dest), copy.num_bytes,
-                         QURT_MEM_CACHE_INVALIDATE, QURT_MEM_DCACHE);
+                         QURT_MEM_CACHE_FLUSH_INVALIDATE, QURT_MEM_DCACHE);
     qurt_mem_cache_clean(reinterpret_cast<qurt_addr_t>(copy.src), copy.num_bytes,
-                         QURT_MEM_CACHE_INVALIDATE, QURT_MEM_DCACHE);
+                         QURT_MEM_CACHE_FLUSH_INVALIDATE, QURT_MEM_DCACHE);
```
janetsc, line 261: Can this be simplified to a flush on dst?

src/tir/transforms/lower_async_dma.cc
```diff
@@ -192,19 +198,33 @@ class AsyncDMALowerer : public StmtExprMutator {
...
+          Evaluate(Call(DataType::Int(32), builtin::dma_copy(),
+                        {queue_id,
+                         Call(DataType::Handle(), builtin::address_of(),
+                              {BufferLoad(bufferstorenode->buffer, store_index)}),
+                         Call(DataType::Handle(), builtin::address_of(),
+                              {BufferLoad(bufferloadnode->buffer, load_index)}),
+                         for_loop->extent * bufferloadnode->dtype.bytes(), dma_bypass_cache_}));
+
+      // if the buffer we are about to DMA was modified by the primfunc
+      // then we need to flush the buffer from the cache prior to the DMA
```
janetsc, line 211: This is great!  Nice general solution. Can we detect a buffer coming in the other direction?  In this direction it is a flush, coming back we need an invalidate.

src/runtime/hexagon/hexagon_buffer.cc
```diff
@@ -235,19 +245,20 @@ void hexagon_buffer_copy_across_regions(const BufferSet& dest, const BufferSet&
...
   // Finally, do the memory copies.
   for (const auto& copy : macro_copies) {
-    // clean Hexagon cache before / after memcpy to ensure clean cache state to enable usage of DMA
-    // bypass mode for increased DMA bandwidth
     // TODO(HWE): Switch to ION Buffer to avoid need for memcpy and potentially lighten or alleviate
     // the burden of cache invalidation in this code
-    qurt_mem_cache_clean(reinterpret_cast<qurt_addr_t>(copy.dest), copy.num_bytes,
-                         QURT_MEM_CACHE_INVALIDATE, QURT_MEM_DCACHE);
-    qurt_mem_cache_clean(reinterpret_cast<qurt_addr_t>(copy.src), copy.num_bytes,
-                         QURT_MEM_CACHE_INVALIDATE, QURT_MEM_DCACHE);
```
janetsc, line 245: When copying to Hexagon, did we verify that a flat src buffer always passed as a parameter to FastRPC?  It could be the object pointer (which has several allocations).

src/tir/transforms/lower_async_dma.cc
tmoreau89: Agreed that it's probably better to perform an invalidation vs. flush depending on the directionality of the data transfer:

- Upon DMA “read”, you have to flush before

- Upon DMA “write”, you have to invalidate after

A helpful example hopefully could be what was done for VTA when VTA would be implemented with non-coherent DMA, as in here: https://github.com/apache/tvm/blob/bf0607bd317a7db8eba1a91c12170934c1ad201f/vta/runtime/runtime.cc#L1319-L1329

src/runtime/hexagon/hexagon_buffer.cc
adstraw: Yes, and, by the same logic we should be able to invalidate (only) the source of the `memcpy` below given that the source is not modified by the `memcpy`.  I have made this change as well.

adstraw: Without going into too much detail, I believe that invalidate (only) maps to "flush and invalidate" when data is modified in the cache.  Before this PR we were doing invalidate (only) on the destination of the `memcpy` which is modified in cache which should not have worked unless the modified data was first flushed from the cache.  Still, using "invalidate" for unmodified data and "flush and invalidate" for modified data is semantically correct and makes the code more readable so I will follow that convention.

src/runtime/hexagon/hexagon_buffer.cc
adstraw: Prefer to keep "flush and invalidate" which is safer (if less performant) to remove the HexagonBuffer object from the cache which is the idea captured in comments behind this PR.

janetsc: Can you explain why src would need a cache operation after the copy?  I think only dst needs to be flushed, my comment was to remove the operation on src, as the data src points to does not change after the memcpy.  src should be invalidated before a memcpy where the hexagon buffer is being copied to Hexagon.

src/runtime/hexagon/hexagon_buffer.cc
janetsc: Some clarifications (from our offline discussion, saving here for posterity):

I was wondering what gets passed in the fast RPC call.  If it is a pointer to the allocation and not a pointer to the hexagon buffer object, then we shouldn't need any of these cache operations surrounding the memcpy.  (Because of Karl's comment that those have code that will make sure buffers passed as arguments are coherent.)

It sounds like DDR is going to be just one allocation in that object.  And we do have the map of allocations to buffer objects, so that makes me think we can get rid of everything except a flush on dst when copying to the device, after the memcpy operation.

The reason that is needed is just in case there is no primfunc modifying that data before a DMA to VTCM.  In that case, we want to make sure it starts out flushed to DDR.

src/tir/transforms/lower_async_dma.cc
adstraw: @tmoreau89 and @janetsc I think this is really good feedback but I am a little leery to make changes without a failing unit test to use for test driven development as with `test_matmul.py` in this PR.  I imagine that software cache management to enable DMA bypass on Hexagon will be an iterative process.  It seems like you are pointing to the next iteration based on VTA example.  I would like to let this PR move through on its own merit and then address follow on cases, if possible.  Thoughts?

janetsc: So the next PR could have:

1.  Unit tests that expose the root of the problem (if they don't, we should go back to the drawing board)
2.  Additions to the matmul test that expose the need to invalidate in the VTCM to DDR direction
3. Changes to insert flush in the right place and invalidate in the right place.

The main reason I was advocating to do both directions now is that I'd like to confirm that those two unit tests do indeed fail.  If they don't, we'll need to do more experiments to get to the root of it.  The unit tests (from our offline discussion):

DDR to VTCM
    Allocate two buffers in DDR and VTCM
    Write "0xbeefbeef" to the DDR buffer.  Flush.
    Write "0xfaceface" to the DDR buffer.  Do NOT flush.
    DMA from DDR to VTCM with BYPASS ON.
Compare DDR and VTCM buffers.  They should not match, because the real values you wanted in DDR weren't flushed to DDR.

VTCM to DDR
    Allocate two buffers in DDR and VTCM
    Write "0xbeefbeef" to the DDR buffer.  (You can flush or not - it actually won't matter.)
    Write "0xfaceface" to the VTCM buffer.
    DMA from VTCM to DDR with BYPASS ON.
Compare DDR and VTCM buffers.  They will not match because there were stale values in the L2 cache for the DDR buffer, so it won't pick up what you wrote with BYPASS ON.  DDR needed to be invalidated first.

All that said, with a flush-only operation for DMA, I'll sign off on this as a good incremental step.  Thanks!

janetsc (approved):
Thanks - looks good!

tmoreau89 (approved):
I like the changes you've applied inside of `hexagon_buffer_copy_across_regions`. This one is good to go.

PR #13845, opened 2023-01-26
[microTVM] Update tutorials

mehrdadh:
This PR updates microTVM tutorials to use updated APIs. 
It also adds an ordering to the tutorials that are useful for first time users.
RVM tutorial is also removed as it is not supported anymore.


python/tvm/micro/testing/utils.py
```diff
@@ -47,10 +47,13 @@ def get_supported_boards(platform: str):
...
 
 
-def get_target(platform: str, board: str) -> tvm.target.Target:
+def get_target(platform: str, board: str = None) -> tvm.target.Target:
     """Intentionally simple function for making Targets for microcontrollers.
     If you need more complex arguments, one should call target.micro directly. Note
     that almost all, but not all, supported microcontrollers are Arm-based."""
+    if platform == "crt":
+        return tvm.target.target.micro("host")
+
```
guberti, line 56: Can we throw a better error if `platform` is _not_ `"crt"` and `board` is `None`?

gallery/how_to/work_with_microtvm/micro_pytorch.py
```diff
@@ -93,11 +93,11 @@
...
 #
 
-
 # Simulate a microcontroller on the host machine. Uses the main() from `src/runtime/crt/host/main.cc`
-# To use physical hardware, replace "host" with another physical micro target, e.g. `nrf52840`
-# or `mps2_an521`. See more more target examples in micro_train.py and micro_tflite.py tutorials.
-target = tvm.target.target.micro("host")
+# To use physical hardware, replace `board` with another physical micro target, e.g. `nrf5340dk_nrf5340_cpuapp`
+# or `mps2_an521` and change the platform type to Zephyr.
+# See more more target examples in micro_train.py and micro_tflite.py tutorials.
```
guberti, line 99: Can we mention these tutorials in text instead of in a code comment? That way we can add clickable links.

guberti (approved):
LGTM - thanks for updating these tutorials!

gallery/how_to/work_with_microtvm/micro_pytorch.py
mehrdadh: good catch, fixed it

python/tvm/micro/testing/utils.py
mehrdadh: added, thanks!

PR #13846, opened 2023-01-26
[Relay] Convert negative axes to positive when importing ONNX Unsqueeze

balaram-cadence:
Currently, ONNX frontend imports Unsqueeze as a sequence of expand_dims in Relay. This works fine when the axes are positive. 

However, for negative axes we need to compute the output rank and add the negative axis value to get its equivalent positive axis value. 

This commit computes the output rank as the sum of the input rank of data and the number of axis in the axes array and convert negative axes to positive. 

This commit also adds testcases for unsqueeze with negative axes and verifies the result with onnx runtime.


balaram-cadence:
@AndrewZhaoLuo : This is my first pull request. Can you please take a look?

masahi (approved):

PR #13849, opened 2023-01-26
[RUNTIME] Fix the manual determination of cores in FillDataForMeasure

dsbarinov1:
**Motivation:** Assertion failed during tuning

**Error message from _thread_pool.cc:295_:**
`Check failed: num_task <= num_workers_used_ (8 vs. 1) : Request parallel sync task larger than number of threads used  workers=1 request=8`

**Main problem description:**
Tuning of the _ARM Snapdragon 888 CPU_ architecture ends with an error above.

**Suspected reason:**
Incorrect (manual) determination of the number of threads. The number of threads is determined using _MaxConcurrency_ and returns 8 threads for this architecture, but the number of actually used threads is 4. This fix urges to use automatic determination of the number of threads by passing 'zero' as 'num_threads' attribute in 'TVMBackendParallelLaunch' to avoid the abovementioned discrepancy.

### DIFF
--- a/src/runtime/contrib/random/mt_random_engine.cc
+++ b/src/runtime/contrib/random/mt_random_engine.cc
@@ -192,12 +192,12 @@ class RandomEngine {
     struct ParallelTask {
       static int RunTask(int task_id, TVMParallelGroupEnv* penv, void* cdata) {
         ParallelTask* task = static_cast<ParallelTask*>(cdata);
-        task->Run(task_id);
+        task->Run(task_id, penv->num_task);
         return 0;
       }
 
-      void Run(int i) {
-        int64_t chunk_size = size / num_threads;
+      void Run(int i, int num_tasks) {
+        int64_t chunk_size = size / num_tasks;
         int64_t st = i * chunk_size;
         int64_t ed = std::min(st + chunk_size, size);
         self->FillDataImpl(data, st, ed, dtype);
@@ -205,7 +205,6 @@ class RandomEngine {
 
       RandomEngine* self;
       void* data;
-      int num_threads;
       int64_t size;
       DLDataType dtype;
     };
@@ -220,8 +219,7 @@ class RandomEngine {
     }
     if (dtype.bits == 1 || dtype.bits == 4 || dtype.bits == 8 || dtype.bits == 16 ||
         dtype.bits == 32 || dtype.bits == 64) {
-      int num_threads = task.num_threads = runtime::threading::MaxConcurrency();
-      int res = TVMBackendParallelLaunch(ParallelTask::RunTask, &task, num_threads);
+      int res = TVMBackendParallelLaunch(ParallelTask::RunTask, &task, 0);
       ICHECK_EQ(res, 0) << "RandomFillForMeasure: TVMBackendParallelLaunch failed";
     } else {
       LOG(FATAL) << "Doesn't support dtype code " << dtype.code << " dtype bits " << dtype.bits;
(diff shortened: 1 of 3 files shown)


src/runtime/contrib/random/mt_random_engine.cc
```diff
@@ -192,12 +192,12 @@ class RandomEngine {
...
         ParallelTask* task = static_cast<ParallelTask*>(cdata);
-        task->Run(task_id);
+        task->Run(task_id, penv->num_task);
         return 0;
       }
 
-      void Run(int i) {
-        int64_t chunk_size = size / num_threads;
+      void Run(int i, int num_threads) {
+        int64_t chunk_size = ceil(size / num_threads);
```
echuraev, line 200: In case when `penv->num_task` is a pretty big number, then is it still correct to use this number as a divider for `size`? And how `prev->num_task` is correlate with number of threads?
Is it possible that `size` is less than `num_threads`?

src/runtime/contrib/random/mt_random_engine.cc
```diff
@@ -220,8 +220,7 @@ class RandomEngine {
     }
     if (dtype.bits == 1 || dtype.bits == 4 || dtype.bits == 8 || dtype.bits == 16 ||
         dtype.bits == 32 || dtype.bits == 64) {
-      int num_threads = task.num_threads = runtime::threading::MaxConcurrency();
-      int res = TVMBackendParallelLaunch(ParallelTask::RunTask, &task, num_threads);
+      int res = TVMBackendParallelLaunch(ParallelTask::RunTask, &task, 0);
```
echuraev, line 223: It looks like variable `num_threads` from `ParallelTask` structure doesn't use anymore. Should we remove it from the structure?

src/runtime/contrib/random/mt_random_engine.cc
```diff
@@ -192,12 +192,12 @@ class RandomEngine {
...
       static int RunTask(int task_id, TVMParallelGroupEnv* penv, void* cdata) {
         ParallelTask* task = static_cast<ParallelTask*>(cdata);
-        task->Run(task_id);
+        task->Run(task_id, penv->num_task);
         return 0;
       }
 
-      void Run(int i) {
-        int64_t chunk_size = size / num_threads;
+      void Run(int i, int num_threads) {
```
echuraev, line 199:
```suggestion
      void Run(int i, int num_tasks) {
```

echuraev (commented):
Could you please add a unit test?

src/runtime/contrib/random/mt_random_engine.cc
apeskov: > is it still correct to use this number as a divider for size.

Yes and no simultaneously. With current implementation it is incorrect. There is missed check at line below `int64_t st = std::min(i * chunk_size, size);`. With adding this line it will be correct.

> how prev->num_task is correlate with number of threads?

It's one and the same.  TVMBackendParallelLaunch API [reference](https://github.com/apache/tvm/blob/56771a87d1560f8963cf745ad093ffc3d83f3f6a/include/tvm/runtime/c_backend_api.h#L119-L146) 

> Is it possible that size is less than num_threads?

Yes, it's possible.

dsbarinov1:
@tvm-bot rerun

dsbarinov1:
@tvm-bot rerun

junrushao (approved):
LGTM!

echuraev (approved):
LGTM. Thanks

PR #13851, opened 2023-01-27
[MetaSchedule] Fix for RewriteLayout + AllocateConst when the rank of the rewritten weight doesn't change

masahi:
Previously, the condition `constant.Shape().size() == index_map->initial_indices.size()` was used to decide if the constant that `TECompiler` finds has its layout write already applied or not. But for 1x1 conv2d case, there seems to be a case where the rank of the weight doesn't change after layout rewriting. For example, (1, 1, 32, 16) -> (16, 4, 2, 4). 

This bug led to the same index map being applied twice on a constant during MS tuning and the final compilation.


junrushao (approved):
LGTM!

Hzfengsy (approved):

PR #13852, opened 2023-01-27
[TIR] Handle nullptr returned by FindEntryFunc

Icemist:
The FindEntryFunc function can return a null pointer. In development I got this situation, which appears as a segfault.


Icemist:
cc @masahi @echuraev

junrushao:
Quick question: do you think it’s necessary to add a unittest?

echuraev (commented):
In general LGTM. But the same question as @junrushao have already asked. Can you add a unit test?

echuraev (approved):
LGTM. Thanks

PR #13854, opened 2023-01-27
[QNN][Relay][Topi] Add qnn.dense with weight layout

ibsidorenko:
This commit adds new Relay operation `qnn.dense_pack` that supports different weights layout (`nn.dense` and `qnn.dense` do not support this attribute). This new operation is full analog of `nn.contrib_dense_pack` operation but in QNN space.

With this PR, current QNN Dense can achieve ~10x performance gain on Hexagon target without QNN canonicalization (through the use of `vrmpy` intrinsic).

Also, this PR includes slight performance improvement for `qnn.mul` (without QNN canonicalization).


TejashShah:
cc @quic-sanirudh @jverma-quic @fars

quic-sanirudh:
> cc @quic-sanirudh @jverma-quic @farshidsp

This is cool, thanks a lot. I went through this a little already earlier, but I'll go through the PR more closely.

python/tvm/topi/hexagon/qnn/nn.py
```diff
@@ -874,6 +916,98 @@ def schedule_qnn_dense(outs):
...
+    outs: Array of Tensor
+          The computation graph description of qnn.dense
+          in the format of an array of tensors.
+
+    Returns
+    -------
+    sch: Schedule
+        The computation schedule for the op.
+    """
+    return default_schedule(outs)
```
quic-sanirudh, line 1008: is there a plan to add vrmpy tensorized schedule separately, or is that going to be dependent on metaschedule?

quic-sanirudh (approved):
Looks good to me apart from the one question about schedule.

Also, one small thought and I might be wrong here, but I see that the `nn` version of this op is called `nn.contrib_dense_pack`, so should we name this `qnn.contrib_dense_pack`?

python/tvm/topi/hexagon/qnn/nn.py
ibsidorenko: I have such plans, but with low priority. The most interesting is work of MetaScheduler for now.

ibsidorenko:
> analog

Yes, I can rename this operator to be aligned with its analog `nn.contrib_dense_pack`

python/tvm/relay/qnn/op/legalizations.py
```diff
@@ -340,6 +340,62 @@ def helper_change_dtypes_to_int8(attrs, inputs, types, relay_op):
...
+    # Shift input if necessary.
+    if data_dtype == "int8":
+        # Compute (QA + 128) and (zp_a + 128)
+        data, input_zero_point = _shift(data, input_zero_point, "uint8")
+
+    # Shift kernel if necessary.
+    if kernel_dtype == "int8":
+        # Compute (QA + 128) and (zp_a + 128)
+        kernel, kernel_zero_point = _shift(kernel, kernel_zero_point, "uint8")
+
```
jverma-quic, line 391: Why do we want to convert kernel to int8?

ibsidorenko: Main goal of conversion int8 --> uint8 is to enable using of faster vrmpy u8**u8**i32 intrinsic instead of vrmpy u8**i8**i32 for `qnn.dense`/`qnn.conv2d`
Yes, you're absolutely right. We have some overhead on such conversion. That's why I have commented out this code and skip i8 -> u8 conversion for weights. Right now I do not see any performance improvement due to overhead.
I will enable this code if I will manage to get better performance.

tests/python/relay/test_pass_qnn_legalize.py
```diff
@@ -296,7 +297,98 @@ def _get_mod(data_dtype, kernel_dtype):
         assert "cast" in legalized_mod.astext() and "qnn" in legalized_mod.astext()
 
 
+def test_qnn_legalize_qnn_conv2d_non_scalar_qnn_params():
+    """
+    Test QNN legalization for qnn.dense op for Hexagon target when kernel zero point and kernel
```
jverma-quic, line 302: typo?: qnn.dense->qnn.conv2d

tests/python/relay/test_pass_qnn_legalize.py
```diff
@@ -296,7 +297,98 @@ def _get_mod(data_dtype, kernel_dtype):
...
+    scale are not scalars.
```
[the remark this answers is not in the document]
ibsidorenko: So, by "not scalar" I mean constant vector of scalars. For example:
`relay.const([val1, val2 ... valN])`
This is more interesting case for testing because on legalization we do padding of weights. And dimension of weights, vector of scale/zero point and 'channels' attribute should be aligned.

tests/python/relay/test_pass_qnn_legalize.py
ibsidorenko: Done, fixed

jverma-quic (approved):

echuraev (approved):

PR #13858, opened 2023-01-27
[microTVM]Refactor test and add skip to current failing tests/boards

mehrdadh:
This PR refactors some of the Zephyr tests and adds skip for each test/board that is currently failing.


driazati (approved):
refactors look pretty harmless, but is there any record which behaviors are supported on which boards?

mehrdadh:
@driazati on the internal device testing we see failures on these tests. Some of these failures will be fixed with better workspace memory calculation, but until then it's better to have these as skip to avoid spams in the microTVM CI.

PR #13859, opened 2023-01-28
[TVMScript] Connect `assert_structural_equal` with new TVMScript printer

cyx-6:
This PR refactors the output of `assert_structural_equal`. Different from the directly printing mismatching nodes, in the old version, the improved one will print the whole scripts, with mismatching nodes underlined. And we print the `ObjectPath` to the mismatching nodes for further better debug. For example, we have following functions

```python
@T.prim_func
def func1(a: T.handle, b: T.handle):
  A = T.match_buffer(a, (128, 128))
  B = T.match_buffer(b, (128, 128))

@T.prim_func
def func2(a: T.handle, b: T.handle):
  A = T.match_buffer(a, (128, 128))
  B = T.match_buffer(b, (128, 256))
```

the log of `assert_structural_equal(func1, func2)` will be like

```python
ValueError: StructuralEqual check failed, caused by lhs at <root>.buffer_map[b].shape[1].value:

@T.prim_func
  A = T.match_buffer(a, (128, 128))
  B = T.match_buffer(b, (128, 128))
                              ^^^
  T.evaluate(0)
and rhs at <root>.buffer_map[b].shape[1].value:

@T.prim_func
  A = T.match_buffer(a, (128, 128))
  B = T.match_buffer(b, (128, 256))
                              ^^^
  T.evaluate(0)
```

instead of

```python
ValueError: StructuralEqual check failed, caused by lhs:
128
and rhs:
256
```

which is not readable sometimes.

### DIFF
--- a/src/node/structural_equal.cc
+++ b/src/node/structural_equal.cc
@@ -202,8 +202,12 @@ bool SEqualReducer::ObjectAttrsEqual(const ObjectRef& lhs, const ObjectRef& rhs,
  */
 class SEqualHandlerDefault::Impl {
  public:
-  Impl(SEqualHandlerDefault* parent, bool assert_mode, Optional<ObjectPathPair>* first_mismatch)
-      : parent_(parent), assert_mode_(assert_mode), first_mismatch_(first_mismatch) {}
+  Impl(SEqualHandlerDefault* parent, bool assert_mode, Optional<ObjectPathPair>* first_mismatch,
+       bool defer_fails)
+      : parent_(parent),
+        assert_mode_(assert_mode),
+        first_mismatch_(first_mismatch),
+        defer_fails_(defer_fails) {}
 
   bool SEqualReduce(const ObjectRef& lhs, const ObjectRef& rhs, bool map_free_vars,
                     const Optional<ObjectPathPair>& current_paths) {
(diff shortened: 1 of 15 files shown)


src/node/structural_equal.cc
```diff
@@ -313,10 +321,38 @@ class SEqualHandlerDefault::Impl {
       *first_mismatch_ = current_paths;
     }
     if (assert_mode_ && !result) {
-      LOG(FATAL) << "ValueError: StructuralEqual check failed, caused by lhs:" << std::endl
-                 << lhs << std::endl
-                 << "and rhs:" << std::endl
-                 << rhs;
+      std::ostringstream oss;
+      oss << "ValueError: StructuralEqual check failed, caused by lhs";
```
Hzfengsy, line 325: How do we print nodes that cannot be printed as TVMScript (e.g. Relay Nodes)?

cyx-6: This is a great question. I have added the fallback mechanism to the `TVMScriptPrinter::Script` [here](https://github.com/apache/tvm/blob/0b8b8a2bab09a03258060560b487a8da5ca92f1c/src/node/script_printer.cc#L32-L34) in the latest commit. If we cannot dispatch the nodes to print (i.e. relay nodes, `Array`, `Map`, etc.), we fallback to the legacy printer.

Hzfengsy: Could you please also comment a bit in this file, saying we will fallback to Repr printer for unsupported nodes.

cyx-6: Sure, the comments have been added

Hzfengsy (approved):

junrushao:
Thanks @cyx-6 for the PR! A side note: you may note the PR description in the commit message :-)

PR #13861, opened 2023-01-29
[TVMScript] Robustify the Highlight Printer

Hzfengsy:
Current `cprint` allows `PrimFunc` and `IRModule` as input. However, after the fragment printing enabled by @junrushao, we can make it robust to support any Objects that have method `script`

cc @junrushao @cyx-6


junrushao (approved):
LGTM

junrushao:
Shall we fix the CI error and get it in?

PR #13863, opened 2023-01-29
[ONNX] Support SequenceLength op

vvchernov:
Support SequenceLength op on ONNX front-end and test for it


echuraev (approved):
LGTM. Thanks

KJlaccHoeUM9l (approved):
LGTM

PR #13865, opened 2023-01-30
[ONNX] Support SequenceErase op

vvchernov:
Support SequenceErase op in ONNX front-end and its test in CI

### DIFF
--- a/python/tvm/relay/frontend/onnx.py
+++ b/python/tvm/relay/frontend/onnx.py
@@ -6168,10 +6168,13 @@ class SequenceErase(OnnxOpConverter):
         else:
             position = -1
 
+        seq_len = len(input_sequence)
+        assert -seq_len <= position < seq_len, "Position is out of bounds"
+
         if position < 0:
-            position = len(input_sequence) + position
+            position = seq_len + position
         # Convert sequence to a list, insert tensors before erased, and repackage as Tuple.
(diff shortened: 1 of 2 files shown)


python/tvm/relay/frontend/onnx.py
```diff
@@ -6148,13 +6148,35 @@ def _impl_v11(cls, inputs, attr, params):
...
+                position = params[position.name_hint].numpy()
+            else:
+                raise NotImplementedError("Position must be a constant.")
+        else:
+            position = -1
+
+        if position < 0:
+            position = len(input_sequence) + position + 1
+        # Convert sequence to a list, insert tensors before erased, and repackage as Tuple.
+        tensor_list = [input_sequence[i] for i in range(position)]
```
KJlaccHoeUM9l, line 6174: Looks like it could be simplified like this:
```Python
tensor_list = [input_sequence[i] for i in range(len(input_sequence)) if i != position]
```
This `if` will not affect performance in any way, because everything will be done at compile time. However, this code looks more readable.

vvchernov: done

KJlaccHoeUM9l (approved):
LGTM

echuraev (approved):
LGTM. Thanks

PR #13869, opened 2023-01-30
Enable C++17 for cmake modules

echuraev:
Moved the lines which add c++17 to CXX_FLAGS before the include of modules. After this c++17 features should be also supported in source code of the modules.


echuraev:
Build in PR #13868 is failed because c++17 features are not available in OpenCL module. I want to try to enable c++17 support in all modules. For now, this is a draft PR, because I need to check that nothing has broken with these changes.

echuraev:
cc: @kparzysz-quic, @masahi, @junrushao

kparzysz-quic (approved):

PR #13874, opened 2023-01-30
[TOPHUB] use keys as a keyword for searching of existing statistics

elvin-n:
In case of ARM we might not specify -device and in this case llvm will be used while even in this case we can determin proper filename with stat since keys have architecture defined. The same situatin must with with x86

### DIFF
--- a/python/tvm/autotvm/tophub.py
+++ b/python/tvm/autotvm/tophub.py
@@ -110,8 +110,10 @@ def context(target, extra_files=None):
         device = tgt.attrs.get("device", "")
         if device != "":
             possible_names.append(_alias(device))
-        for k in tgt.keys:
-            possible_names.append(k)
+        # for cases when we do have explicitly defined -device in the target,
+        # we still might have information about it stored in keys container
+        # in other case we will load statistics for definitely irrelative stat
+        possible_names.extend(tgt.keys)
         possible_names.append(tgt.kind.name)
 
         all_packages = list(PACKAGE_VERSION.keys())
--- a/python/tvm/autotvm/tophub.py
+++ b/python/tvm/autotvm/tophub.py
@@ -106,13 +106,19 @@ def context(target, extra_files=None):
         if isinstance(tgt, str):
             tgt = Target(tgt)
 
+        # The TOPHUB file names rely on Target's device or kind. Both these types of
+        # information exist in Target.keys, but rules of filling this filed is not explicitly
+        # defined, we are afraid to rely only on Target.keys. At the same time Target.device
+        # is filled only if device was pointed explicitly in target string, that is not mandatory
+        # and in some cases we need to get information about device from Target.keys
+        # In priority order we verify:
+        # 1) Target.device
+        # 2) Target.keys
+        # 3) Target.kind
         possible_names = []
         device = tgt.attrs.get("device", "")
         if device != "":
             possible_names.append(_alias(device))
-        # for cases when we do have explicitly defined -device in the target,
-        # we still might have information about it stored in keys container
-        # in other case we will load statistics for definitely irrelative stat
         possible_names.extend(tgt.keys)
         possible_names.append(tgt.kind.name)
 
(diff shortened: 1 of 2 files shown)


python/tvm/autotvm/tophub.py
```diff
@@ -110,6 +110,8 @@ def context(target, extra_files=None):
         device = tgt.attrs.get("device", "")
         if device != "":
             possible_names.append(_alias(device))
+        for k in tgt.keys:
+            possible_names.append(k)
```
echuraev, line 114: Please add an explanatory comment to describe what you are trying to fix here.

Also, probably this would work in the same way?
```suggestion
        possible_names.extend(tgt.keys)
```

echuraev (approved):
LGTM. Thanks

python/tvm/autotvm/tophub.py
```diff
@@ -110,6 +110,10 @@ def context(target, extra_files=None):
         device = tgt.attrs.get("device", "")
         if device != "":
             possible_names.append(_alias(device))
+        # for cases when we do have explicitly defined -device in the target,
+        # we still might have information about it stored in keys container
+        # in other case we will load statistics for definitely irrelative stat
```
ekalda, line 115: Any chance this comment can be made clearer? :) I read it several times and I don't get it. 
`might have information`
What information?
`about it`
About what?
`keys container`
I guess that is Target.keys? Maybe make that more explicit
`in other case`
What case? When keys container is empty or when -device is not defined?
`irrelative stat`
What is irrelative stat?

Also, is it one or multiple sentences? Capital letters and punctuation would be helpful :)

elvin-n: :) tried to include logic of matching statistics without detailed description in this place.

1. the only mandatory field in target is `kind`. it can be llvm/opencl/cuda/...
2. it might be `device` specified, but not required. It might be "arm_cpu", or "adreno" for example
3. it might be `model` specified, but not required. It might be "riscv32-unknown-linux-gnu" or "kirin970" for example
4. the `keys` field is not specified but is filled after analysis of the Target string automatically by devices. It is done for hierarchical search in the `topi` implementations and do fallback to lower/parents layers

Before this change, dedicated tophub *file* was searched only by `device` and second by `kind` 
As you remember, `device` is not mandatory in the Target. At the same time TVM understands that target belongs to arm_cpu, it treat it as arm_cpu, but before my change, the target like `llvm -model=snapdragon835 -mtriple=arm64-linux-android -mattr=+neon` was treated only as `llvm` stat file. I cannot even imagine on which device `llvm` stat was collected, but definitely not on ARM. I.e. statistics was irrelevant.

Adding keys to the priority before `kind` we increase the probability of searching proper statistics.

Do I need to comment line by line your answer?

elvin-n: ok, will modify comment

elvin-n: As for applying of the statistics, I am not even sure that taking statistics from another model of the same device is a good idea. Fortunately it more or less works for ARM CPU currently but if we take intel graphics, the previous models of Iris had different limitation of group worksize. Statistics which collected on newer graphics cannot be applied on previous one. I.e. we risk not to have bad performance but have just not working kernels.

I.e. I agree that we need to do something with tophub and search algorithm, but do not have idea how we need to do this.

And probably it is not so important assuming that Metaschedule should be the tuner #1. On the other hand MS is not a solution for default compilation without tuning.

ekalda: Yeah I agree that applying same tuning statistics across different implementations of the same architecture is quite sketchy, especially if it happens in this non-transparent way it currently does. I'm also not sure how often the database gets updated with new stats. I haven't seen compilation crashing because of unsuitable tuning records from tophub myself, but if it happens, that's really bad, since it makes tuning compulsory for being able to compile a network with TVM. Also, sounds like a really difficult thing to debug. But I suppose as long as AutoTVM is the default tuner (as it is for tvmc) and therefore the one new potential users will use to evaluate TVM, it should be kept in a state where its behaviour is representative of TVM's capabilities. In my opinion it would be good if we didn't have 3 tuners to maintain in parallel, but I suppose that's where the community is at the moment.

ekalda (approved):
Looks great, thanks @elvin-n for updating the comment! :)

PR #13875, opened 2023-01-30
[Fix][Arith] Analyzer simplification starts with canonical

MasterJH5574:
This PR updates the order of arithmetic analyzer simplification, by adding a stage of canonical simplification at the very beginning so that every simplification always starts with a canonical round. This is because the rewrite simplification may destroy some PrimExpr property that the canonical simplification can make use of. Therefore, adding the canonical one in the front can maximize the use of canonical simplification.

### DIFF
--- a/src/arith/canonical_simplify.cc
+++ b/src/arith/canonical_simplify.cc
@@ -335,6 +335,8 @@ class SumExprNode : public CanonicalExprNode {
    * \return whether the cast can be safely pushed to children
    */
   bool CanPushCastToChildren(DataType dtype, Analyzer* analyzer) const {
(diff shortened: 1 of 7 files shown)


wrongtest-intellif (approved):
LGTM!

src/arith/canonical_simplify.cc
```diff
@@ -514,6 +514,9 @@ class SumExprNode : public CanonicalExprNode {
       }
     }
     if (base < 0) {
+      if (base == -(1LL << (dtype.bits() - 1))) {
+        base += 1;
+      }
```
MasterJH5574, line 519: cc @tqchen

MasterJH5574:
@tvm-bot rerun

tqchen:
@tvm-bot rerun

tqchen (approved):

PR #13880, opened 2023-01-31
[ARITH] support floordiv in deduce bound

wrongtest-intellif:
Hi, the patch handle floordiv `a // positive` in bound deducer. So that we could know `x >= 24` in `x // 8 >= 3`

### DIFF
--- a/tests/python/unittest/test_arith_deduce_bound.py
+++ b/tests/python/unittest/test_arith_deduce_bound.py
@@ -219,7 +219,6 @@ def test_deduce_non_support():
         res = tvm.arith.deduce_bound(a, lhs < 10, {}, {})
         assert res.is_nothing()
 
-    test_non_support(tvm.tir.floordiv(a, 16))
     test_non_support(tvm.tir.floormod(a, 16))
     test_non_support(tvm.tir.Min(a, 16))
     test_non_support(tvm.tir.Max(a, 16))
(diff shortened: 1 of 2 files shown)


tests/python/unittest/test_arith_deduce_bound.py
```diff
@@ -233,5 +232,39 @@ def test_non_support(lhs):
...
+
+    # test basic cases
+    do_test(lambda a: a // 8 > 3, {}, 32, "pos_inf")
+    do_test(lambda a: a // 8 >= 3, {}, 24, "pos_inf")
+    do_test(lambda a: a // 8 < 3, {}, "neg_inf", 23)
+    do_test(lambda a: a // 8 <= 3, {}, "neg_inf", 31)
+    do_test(lambda a: a // 8 == 3, {}, "pos_inf", "neg_inf")
+    do_test(lambda a: a // 8 > -3, {}, -16, "pos_inf")
+    do_test(lambda a: a // 8 >= -3, {}, -24, "pos_inf")
+    do_test(lambda a: a // -8 > 3, {}, "pos_inf", "neg_inf")
```
Hzfengsy, line 257: I wonder how hard to support these two cases, is it easy to have a quick patch to this PR?
`a // 8 == 3` and `a // -8 > 3`

wrongtest-intellif: I think it is not hard to cover negative divisors. But for EQ some changes seems have to be made across all op impls, currently the detector only maintain single side states.

Hzfengsy (approved):
LGTM

PR #13881, opened 2023-01-31
[TIR][TOPI][CI] Fix number of arguments in calls of llvm_pure_intrin

vvchernov:
All `call_llvm_pure_intrin(...)` methods were rechecked and argument corresponding number of input args was corrected for some of them. Corresponding tests were upstreamed


vvchernov:
@tvm-bot rerun

junrushao:
CC: @vinx13 @masahi

vinx13 (approved):

masahi (approved):
yeah, I wondered why we were passing 0 to VNNI intrins and still getting correct output.

vinx13:
@masahi I found that the number of args is used to decide the arg types of the intrin, if it is not overloaded, it doesn't matter

PR #13887, opened 2023-02-01
[Build] Fix misleading error messages

lucifer1004:
Currently, in `FindLLVM.cmake`, when the `llvm-config` commands fail, the error message might be `Fatal error executing: ON --libfiles` (since `USE_LLVM=ON`), which is misleading.

In this PR, `${use_llvm}` is changed to `${LLVM_CONFIG}`, and the error messages will be more comprehensible.


echuraev (approved):
LGTM. Thanks

junrushao (approved):

PR #13894, opened 2023-02-01
[ONNX] Support ScatterElements with reduction

vvchernov:
Support ScatterElements on ONNX front-end as described in ONNX docs. Just now Scatter op implementation is used where `reduction` attribute is not supported at all. Also CI tests are supported for the op.

P.S. In other PRs after that I plan: 1. remove scatter_add and reconnect all its using to ScatterElements(reduction="add") 2. remove scatter implementation and use ScatterElements(reduction="update") instead of it. It will remove the restriction related to input tensor rank size (just now rank <= 4)

### DIFF
--- a/python/tvm/topi/cuda/scatter_elements.py
+++ b/python/tvm/topi/cuda/scatter_elements.py
@@ -94,6 +94,7 @@ def scatter_elements(data, indices, updates, axis=0, reduction="update"):
         elif i > axis:
             ind_after_axis_range *= value
     ind_before_axis_stride = ind_axis_range * ind_after_axis_range
+    ind_full_range = ind_before_axis_range * ind_before_axis_stride
     ind_full_range_excl_axis = ind_before_axis_range * ind_after_axis_range
 
     def gen_ir(data, indices, updates, out):
@@ -117,6 +118,20 @@ def scatter_elements(data, indices, updates, axis=0, reduction="update"):
             with ib.if_scope(index < full_range):
                 out_ptr[index] = data_ptr[index]
 
+        # Check indices and shift to positive side if need
+        with ib.new_scope():
+            num_blocks_1 = ceil_div(ind_full_range, max_threads)
+            bx1 = te.thread_axis("blockIdx.x")
+            tx1 = te.thread_axis("threadIdx.x")
+            ib.scope_attr(bx1, "thread_extent", num_blocks_1)
+            ib.scope_attr(tx1, "thread_extent", max_threads)
+
+            ind_fused = bx2 * max_threads + tx2
+            with ib.if_scope(ind_fused < ind_full_range):
+                index_check = tir.LT(indices_ptr[ind_fused], tir.const(0, indices.dtype))
+                indices_ptr[ind_fused] += tir.Select(index_check, axis_range, tir.const(0, indices.dtype))
+                # TODO(vvchernov): assert for index out of bounds
+
         # TODO (vvchernov): use atomic function for special conditions (see cuda.scatter_nd)
         with ib.new_scope():
             num_blocks_2 = ceil_div(ind_full_range_excl_axis, max_threads)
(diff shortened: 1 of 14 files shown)


vvchernov:
@tvm-bot rerun

python/tvm/topi/scatter_elements.py
```diff
@@ -0,0 +1,150 @@
...
+        # pylint: disable=invalid-name
+        ib = tir.ir_builder.create()
+
+        data = ib.buffer_ptr(data_ptr)
+        indices = ib.buffer_ptr(indices_ptr)
+        updates = ib.buffer_ptr(updates_ptr)
+        out = ib.buffer_ptr(out_ptr)
+
+        # Copy initial input data to output
+        with ib.for_range(0, full_range, "i", kind="parallel") as i:
```
KJlaccHoeUM9l, line 105: It seems to be better to use `vectorize` here?

vvchernov: I tnink `parallel` is enough here. There are several reasons. 1) We can not use `vectorize` from the box: the loop should be tiled to get mostinner loop with size 4, 8, 16 or other specified degree of 2 (it is a question which one should be choosen). Another thing full_range is arbitrary, it is not limited by degre of 2, it means we should think additionally about loop tail. 2) This implementation is used by pure compilation. Of course, it does not give us the best performance, but it is assumed in TVM that the best performance can be gotten after metascheduling.

python/tvm/topi/cuda/scatter_elements.py
```diff
@@ -0,0 +1,181 @@
...
+        with ib.new_scope():
+            num_blocks_1 = ceil_div(ind_full_range, max_threads)
+            bx1 = te.thread_axis("blockIdx.x")
+            tx1 = te.thread_axis("threadIdx.x")
+            ib.scope_attr(bx1, "thread_extent", num_blocks_1)
+            ib.scope_attr(tx1, "thread_extent", max_threads)
+
+            ind_fused = bx1 * max_threads + tx1
+            with ib.if_scope(ind_fused < ind_full_range):
+                index_check = tir.LT(indices_ptr[ind_fused], tir.const(0, indices.dtype))
```
AndrewZhaoLuo, line 131: Why can this block not be fused with the below block? Is it to prevent warp divergence?

Perhaps you can do something like:

`index = index + (index < 0) * axis_range`

If that is the case.

AndrewZhaoLuo (commented):
Did a first pass, looks good for most part. One comment. 

Will take a closer look tonight.

python/tvm/topi/cuda/scatter_elements.py
vvchernov: Hello @AndrewZhaoLuo It was initially fused to the common loop. I decided that conditions is not good thing for GPU calculations aspecially wrapped to additional loops. But in this case it looks like the same (I can return it back if you think it is better way). Another thing is assert for probably shifted axis which is still out of bounds (theoretically we should do this check), but I do not know how I can do this check on ir builder side.

vvchernov: It is very convenient and clear expression that you supposed. Is it the same as I wrote or use other tir procedures? I'm slightly aware about excess multiplication in it

AndrewZhaoLuo: The arithmetic intensity per element loaded is very very low (~2) so I am not very worried about the multiplication. It's a pretty common trick to get "branch-less" programming.

Yes, I am not as familiar with IR Builder, I will say if you want to enforce indices being valid you can just do something like  modulo indices internally (for wrapping behavior). 

Personally I think it is ok to not check and assume caller will check and guarantee good inputs as we don't necessarily want the check in the base computation.

AndrewZhaoLuo (commented):
Can you compare current performance of your GPU kernel vs the current scatter_add implementation? 

Looking at it, it may be easier to reuse the existing scatter_add topi implementation and extend it with new reduction functions.

vvchernov:
@AndrewZhaoLuo I did not compare performance on GPU, but I prepared new [PR](https://github.com/apache/tvm/pull/14008) where replace scatter_add by scatter_elements and reuse code for cuda

vvchernov:
One moment for scatter_add: it is implemented for 1d, 2d, 3d and 4d input tensors, as scatter. I thought that I can reuse scatter_add approach, but as result I have done scatter_elements in general way without restrictions on input data rank

python/tvm/topi/cuda/scatter_elements.py
vvchernov: Ok. I have fused block related to index shifting with general one and use your expression.
Code related to scatter_add was refactored in another branch and PR was prepared (see below), the latter is waiting for merging of this PR.

vvchernov:
@tvm-bot rerun

AndrewZhaoLuo (approved):

Issue #13892, opened 2023-02-01
[Bug] TVMScript fails with LetStmt


liangW-intellif:
Hi there,
I encountered an error that when generating TVMScript with LetStmt, the generated script cannot be parsed if the var in LetStmt appears the first time in the IR.
```
import tvm
from tvm.script import tir as T
from tvm import te


def test_let_roundtrip():
    x = te.var("x")
    y = te.var("y")
    let0 = tvm.tir.LetStmt(x, 0, tvm.tir.Evaluate(0))
    let1 = tvm.tir.LetStmt(y, 1, tvm.tir.Evaluate(0))
    body = tvm.tir.SeqStmt([let0, let1])
    func = tvm.tir.PrimFunc([], body)
    print(func.script())
    # Parse fail
    roundtrip = tvm.script.from_source(func.script())
    tvm.ir.assert_structural_equal(func, roundtrip, True)
```

The above case generates TVMScript as below
```
@T.prim_func
def main():
    with T.let(x, 0):
        T.evaluate(0)
    y: T.int32 = 1
    T.evaluate(0)
```
But when parsing this TVMScript, it reports an error that 'x' is an undefined variable so that the parse failed. The TVMScript with LetStmt is not round-trippable in this case.

```
test_let_roundtrip.py::test_let_roundtrip # from tvm.script import tir as T

@T.prim_func
def main():
    with T.let(x, 0):
        T.evaluate(0)
    y: T.int32 = 1
    T.evaluate(0)
error: Undefined variable: x
 --> <str>:5:16
   |  
 5 |      with T.let(x, 0):
   |                 ^     
FAILED
```

For the semantic of LetStmt, I think'x' should be identified as a definition, the script should be parsed successly.

Please let me know if there is any confusion, thanks very much!

### Environment

latest TVM mainline

@junrushao  @wrongtest-intellif

junrushao:
CC @cyx-6 would you like to take a look? Thanks!

cyx-6:
Thanks for reporting this bug. I believe that this is due to the bug of printing `LetStmt`. Since we forget to print the `x = T.var("int32")` before the `with T.let(x, 0):`. This bug will be fixed in the #13900. And more details are in the description of that PR too.

junrushao:
Closed via #13900

PR #13900, opened 2023-02-01
[Fix][TVMScript] Fix `LetStmt` printing logic

cyx-6:
This PR is the bug fix reported in #13892. Initially, we mix the logic of `LetStmt` docsifying method with and without concise scoping. For example, in
```python
x = T.var("int32")
with T.let(x, 0):
```
`x` in the `LetStmt` works as a right value, while in
```python
x: T.int32 = 0
```
`x` in the `LetStmt` works as a left value as result.
Our old logic mixed them together to generate the wrong code for the first case.
Meanwhile, during the fix, we found another bug in concise scoping check. For example, we have
```python
x = T.var("int32")
y = T.var("int32")
with T.let(x, y):
  with T.let(y, 0):
```
here we should not output
```python
x = T.var("int32")
y = T.var("int32")
with T.let(x, y):
  y: int32 = 0
```
becase this will define a new `y_1: int32 = 0` indeed, due the the variable shadowing logic of the parser, which is different from the `y` we define and refer to.
Our concise scoping `v: ... = ...` should launch if and only if the `v` is never defined before.
Otherwise, we use `with T.let(v, ...):` instead.

### DIFF
--- a/src/script/printer/tir/stmt.cc
+++ b/src/script/printer/tir/stmt.cc
@@ -57,13 +57,12 @@ TVM_STATIC_IR_FUNCTOR(IRDocsifier, vtable)
 TVM_STATIC_IR_FUNCTOR(IRDocsifier, vtable)
     .set_dispatch<tir::LetStmt>("", [](tir::LetStmt stmt, ObjectPath p, IRDocsifier d) -> Doc {
       bool concise = AllowConciseScoping(d);
-      ExprDoc rhs = d->AsDoc<ExprDoc>(stmt->value, p->Attr("value"));
-      With<TIRFrame> f(d, stmt);
-      ExprDoc lhs = d->IsVarDefined(stmt->var) ? d->GetVarDoc(stmt->var).value()
-                                               : DefineVar(stmt->var, *f, d);
-      AsDocBody(stmt->body, p->Attr("body"), f->get(), d);
-      Array<StmtDoc>* stmts = &(*f)->stmts;
-      if (concise) {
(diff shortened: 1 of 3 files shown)


junrushao:
CC: @wrongtest-intellif @lightzhan-intellif

Hzfengsy (approved):

wrongtest-intellif:
> ```python
> x = T.var("int32")
> y = T.var("int32")
> with T.let(x, y):
>   with T.let(y, 0):
> ```

Thank you for the fix! Here is my small question for the example, does it shoud be 
```
with T.let(y, 0):
  with T.let(x, y):
```

junrushao:
Off the topic, in the future, we might want to change the syntax to:

```python
with T.LetStmt(a) as b:
```

cyx-6:
@wrongtest-intellif  Oh, that is just an example of the case that a `LetStmt->var` - `y` has usage before the its new value. I picked the outer `LetStmt->value` field for the usage of `y` for convenience. So never mind its logic, it makes no sense, except its format.

lightzhan-intellif:
> Off the topic, in the future, we might want to change the syntax to:
> 
> ```python
> with T.LetStmt(a) as b:
> ```

I prefer this. The current one looks a little bit confusing from the view of native python users.

junrushao:
Sounds good! I am merging this bugfix  in first given it seems non-controversial.

Right now I am focusing on implementing TVMScript printer for Relax (coming soon). And after this is done, the syntax we mentioned above will be introduced with backward compatibility guarantee.

PR #13902, opened 2023-02-02
[Unity][CI] setup CI (do not upstream to main)

YuchenJin:
Set up CI to run Relax tests.


Mousius:
@YuchenJin because you're always launching the instances and then exitting, you'd incur the minimum charge of 60 seconds for the instances you're running even if you don't use any of the compute - might be better to turn off the builds in github somewhere?

YuchenJin:
> @YuchenJin because you're always launching the instances and then exiting, you'd incur the minimum charge of 60 seconds for the instances you're running even if you don't use any of the compute - might be better to turn off the builds in github somewhere?

Thanks for the suggestion! The main reason for adding `exit 0` to a bunch of scripts is to avoid the complication when rebasing against the main branch, and we don't need to modify the jenkins while reusing the current tvm jenkins ci setting. 
Would love to hear others' thoughts: @driazati @areusch

tests/scripts/task_convert_scripts_to_python.sh
```diff
@@ -17,6 +17,9 @@
 # under the License.
 set -euxo pipefail
 
+# SKIP: DO NOT UPSTREAM TO MAIN
+exit 0
```
driazati, line 21: Can you move this logic into the Jenkins scripts themselves before any `node(...)` calls are made? Doing it here still allocates machines, runs the docker setup, etc. which is not that much but expensive to end up doing nothing

tqchen:
Thanks for the suggestions. https://github.com/apache/tvm/pull/13910 should achieve the asks that leverages a customized jenkins setup and skip most of the existing pipelines without starting a stage

YuchenJin:
> #13910 should achieve the asks that leverages a customized jenkins setup and skip most of the existing pipelines without starting a stage

Thanks!

PR #13903, opened 2023-02-02
[ci] Disable GPU unit tests

driazati:
This is a (hopefully) temporary bandaid to reduce cloud costs for OctoML
since they've spiked recently, mostly for GPU instances. This cuts down
on the GPU testing we do in order to save on those build minutes.

### DIFF
--- a/ci/jenkins/generated/gpu_jenkinsfile.groovy
+++ b/ci/jenkins/generated/gpu_jenkinsfile.groovy
@@ -60,7 +60,7 @@
 // 'python3 jenkins/generate.py'
 // Note: This timestamp is here to ensure that updates to the Jenkinsfile are
 // always rebased on main before merging:
-// Generated at 2022-12-09T15:39:24.455336
+// Generated at 2023-02-01T17:14:09.557178
 
 import org.jenkinsci.plugins.pipeline.modeldefinition.Utils
 // These are set at runtime from data in ci/jenkins/docker-images.yml, update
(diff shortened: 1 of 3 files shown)


leandron:
Just a quick question. Can you clarify what are the tests being disabled? I imagine it is the diff between `task_python_frontend_lite.sh` and `task_python_frontend.sh`, but can you clarify what in practice was running before, and it is not running now?

Mousius:
@driazati are there any plans to allow other contributors to provide nodes? I remember I asked about this a long time ago but it wasn't a priority, now it might be worth re-visiting?

driazati:
> @driazati are there any plans to allow other contributors to provide nodes? I remember I asked about this a long time ago but it wasn't a priority, now it might be worth re-visiting?

On the technical side this isn't too complicated, it's mostly just a matter of making sure the Jenkins head can communicate with the machines and installing the client daemon. I'd be open to trying it if the machines are there but we'd probably also need some level of commitment to support those machines to make sure they're at least as reliable as our current runners


> Just a quick question. Can you clarify what are the tests being disabled? I imagine it is the diff between `task_python_frontend_lite.sh` and `task_python_frontend.sh`, but can you clarify what in practice was running before, and it is not running now?

There's that and some more broad disabling of stuff in https://ci.tlcpack.ai/blue/organizations/jenkins/tvm-gpu/detail/main/244/pipeline vs https://ci.tlcpack.ai/blue/organizations/jenkins/tvm-gpu/detail/PR-13903/1/pipeline/101

PR #13909, opened 2023-02-03
[microTVM] additional refactoring for enabling USE_MICRO in more builds

alanmacd:
Additional changes that make it possible to enable USE_MICRO across nearly all builds except for hexagon or when USE_RPC is disabled. This mainly standardizes how the microTVM RPC common files are built across all platforms. There will be a follow-up PR that will enable USE_MICRO for nearly all build types once the best mechanism and behaviors for doing so is determined.


CMakeLists.txt
```diff
@@ -264,6 +264,12 @@ else(MSVC)
   endif()
 endif(MSVC)
 
+# USE_MICRO is on by default, but it is not needed for Hexagon builds nor will
+# it function properly with USE_RPC disabled.
+if(USE_HEXAGON OR NOT USE_RPC)
+  set(USE_MICRO OFF)
```
mehrdadh, line 270: I suggest for this case we just show an error and say this mode is not supported. changing the user specified configs silently is not ideal.

alanmacd: Thanks, updated PR to remove this until the best way of doing so is determined.

src/runtime/crt/common/crt_runtime_api.c
```diff
@@ -112,7 +112,7 @@ int TVMDeviceAllocDataSpaceWithScope(DLDevice dev, int ndim, const int64_t* shap
 
 int TVMDeviceFreeDataSpace(DLDevice dev, void* ptr) { return TVMPlatformMemoryFree(ptr, dev); }
 
-static bool IsContiguous(const DLTensor* arr) {
+TVM_ATTRIBUTE_UNUSED static bool IsContiguous(const DLTensor* arr) {
```
mehrdadh, line 115: Can you clarify why this was needed? was it for eliminating the warning?

alanmacd: `IsContiguous()` is only used in an `assert()` below, so in a non-debug build we get an unused function warning

mehrdadh (commented):
LGTM, just waiting for others to take a look as well

mehrdadh (approved):

cmake/modules/StandaloneCrt.cmake
```diff
@@ -15,41 +15,29 @@
...
 
-  add_library(host_standalone_crt
-              STATIC
-              3rdparty/libcrc/src/crcccitt.c
-              src/runtime/crt/microtvm_rpc_common/frame_buffer.cc
-              src/runtime/crt/microtvm_rpc_common/framing.cc
-              src/runtime/crt/microtvm_rpc_common/session.cc
-              src/runtime/crt/microtvm_rpc_common/write_stream.cc)
+  list(APPEND RUNTIME_SRCS ${TVM_CRT_SOURCES})
+  include_directories(SYSTEM src/runtime/micro)
```
driazati, line 31: Why the `SYSTEM` here?

alanmacd: it needs a header in that directory to build those files that are in TVM_CRT_SOURCES, it's the same pattern used in other .cmake and CmakeLists.txt

Issue #13911, opened 2023-02-03
[Bug & fix] gen_requirements issue with latest setup tools


mikeseven:
latest setup tools crash when doing `make cython`
The issue is in gen_requirements.py:
`    ("numpy", "<=1.23.*"),`

must be changed to:
 `   ("numpy", "<=1.23"),
`

lhutton1:
Thanks for reporting @mikeseven, I pushed #13912 to fix this using your suggestion.

mikeseven:
all good, thanks for super speedy fix ;-)

PR #13912, opened 2023-02-03
Fix numpy version constraint

lhutton1:
Issue #13911 reported that newer versions of setuptools crash with the version constraint "<=1.23.*", this commit implements the suggested fix by using "<=1.23".

cc @mikeseven @driazati @leandron @Liam-Sturge


ashutosh-arm (approved):
Thanks @lhutton1.

PR #13914, opened 2023-02-03
Fix build platform environment variable

Liam-Sturge:
Certain tests that make use of `pytest_wrapper.py`, such as those triggered by `task_python_integration.sh` will fail when a "PLATFORM" environment variable is not set within the Docker container.

When using `build.sh` to both create a container, and run a command to execute one of the tests that requires a "PLATFORM", an error will occur due to the missing environment variable.

This patch is necessary to add support for this environment variable to `build.sh` and prevent such errors.

CC @ashutosh-arm @leandron @Mousius @driazati @areusch


driazati (approved):

Liam-Sturge:
@tvm-bot rerun

Liam-Sturge:
Seems to be a transient error with the CI locating Ubuntu mirrors. Was able to reproduce locally and now the issue has cleared up for me locally, so I will re-run the CI.

Liam-Sturge:
@tvm-bot rerun

PR #13804, opened 2023-01-18
[TOPI] Making test_strided_set require a GPU for testing

neildhickey:
https://github.com/apache/tvm/pull/13724 caused an issue in ci_cpu testing when running the TOPI tests. This pull request sets the failing test to require a gpu to run

PR #13918, opened 2023-02-06
[Arith] ConstIntBound was incorrectly assuming bounds were over int64…

neildhickey:
…_t range

[#13804](https://github.com/apache/tvm/pull/13804) made a test requires gpu due to a TOPI test failure. The LLVM 15 update also caused an issue with some tests that were caused by the same section of code.

This change fixes the assumptions made in that code and reenables the test for more targets.

### DIFF
--- a/src/tir/transforms/lower_intrin.cc
+++ b/src/tir/transforms/lower_intrin.cc
@@ -118,16 +118,8 @@ class IntrinInjecter : public tvm::arith::IRMutatorWithAnalyzer {
       // If the numerator's lower bound is known, express the floordiv
       // in terms of truncdiv using only positive operands.
       arith::ConstIntBound const_int_bound = analyzer_->const_int_bound(op->a);
-      // this will depend on the size of op->a
-      int64_t kNegInf;
-      if (op->a.dtype() == DataType::Int(32)) {
-        kNegInf = -std::numeric_limits<int32_t>::max();
-      } else {
-        kNegInf = arith::ConstIntBound::kNegInf;
-      }
-
-      if (const_int_bound->min_value > kNegInf && const_int_bound->min_value < 0 &&
-          const_int_bound->min_value > Downcast<IntImm>(tvm::min_value(op->a->dtype))->value) {
+      if (const_int_bound->min_value < 0 &&
+          const_int_bound->min_value > -(Downcast<IntImm>(tvm::max_value(op->a->dtype))->value)) {
         // The goal is to write floordiv(a,b) in terms of truncdiv, without using
         // negative operands.
         //
@@ -221,16 +213,8 @@ class IntrinInjecter : public tvm::arith::IRMutatorWithAnalyzer {
       // If the numerator's lower bound is known, express the floormod
       // in terms of truncmod using only positive operands.
       arith::ConstIntBound const_int_bound = analyzer_->const_int_bound(op->a);
-      // this will depend on the size of op->a
-      int64_t kNegInf;
-      if (op->a.dtype() == DataType::Int(32)) {
-        kNegInf = -std::numeric_limits<int32_t>::max();
-      } else {
-        kNegInf = arith::ConstIntBound::kNegInf;
-      }
-
-      if (const_int_bound->min_value > kNegInf && const_int_bound->min_value < 0 &&
-          const_int_bound->min_value > Downcast<IntImm>(tvm::min_value(op->a->dtype))->value) {
+      if (const_int_bound->min_value < 0 &&
+          const_int_bound->min_value > -(Downcast<IntImm>(tvm::max_value(op->a->dtype))->value)) {
         // The goal is to write floormod(a,b) in terms of truncdiv and truncmod,
         // without using negative operands.
         //
--- a/src/tir/transforms/lower_intrin.cc
+++ b/src/tir/transforms/lower_intrin.cc
@@ -118,8 +118,15 @@ class IntrinInjecter : public tvm::arith::IRMutatorWithAnalyzer {
       // If the numerator's lower bound is known, express the floordiv
       // in terms of truncdiv using only positive operands.
       arith::ConstIntBound const_int_bound = analyzer_->const_int_bound(op->a);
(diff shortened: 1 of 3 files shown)


neildhickey:
@Lunderberg

ashutosh-arm (commented):
Thanks @neildhickey for re-enabling this. Just a small question below and a nit.

Lunderberg:
The changes look reasonable, and thank you on the catch.  The check was intended to identify cases where there is a known lower bound (i.e. not `kNegInf`) less than zero.  This didn't take into account `ConstIntBoundAnalyzer::Impl::Everything`'s use of the data-type bounds, which seems to be causing the issue here.

Long-term, this could be avoided by having `ConstIntBound` track a `std::optional<int64_t>` for the min/max value, rather than having special values `kPosInf` and `kNegInf`.  That would be a clearer way to indicate "no known bound", but which wasn't possible until C++17's `std::optional`.

Can you add a unit test that validates this behavior for int32 types?

neildhickey:
Yes, I'm working on a test to cover all integer sizes.

tests/python/relay/test_op_floordiv.py
```diff
@@ -0,0 +1,112 @@
...
+from tvm import relay
+import pytest
+from tvm.relay.testing import run_infer_type
+import tvm.topi.testing
+from tvm.contrib.nvcc import have_fp16
+import tvm.testing
+from tvm.topi.utils import get_const_tuple
+from tvm.script import tir
+
+executor_kind = tvm.testing.parameter("graph", "vm")
```
ashutosh-arm, line 15: nit: is it needed by test below?

tests/python/relay/test_op_floordiv.py
```diff
@@ -0,0 +1,112 @@
...
+from tvm import te
+import scipy
+from tvm import relay
+import pytest
+from tvm.relay.testing import run_infer_type
+import tvm.topi.testing
+from tvm.contrib.nvcc import have_fp16
+import tvm.testing
+from tvm.topi.utils import get_const_tuple
+from tvm.script import tir
```
ashutosh-arm, line 13: nit: Looks like some redundant imports here.

tests/python/relay/test_op_floordiv.py
```diff
@@ -0,0 +1,112 @@
...
+        (func_32, "int32"),
+        (func_64, "int64"),
+    ]:
+        built = tvm.build(opfunc, target=target)
+        x_data = np.random.randint(te.min_value(type), te.max_value(type), size=(100), dtype=type)
+        data = []
+        for i in range(N):
+            data.append(i)
+
+        y_data = np.asarray(data, dtype=type)
```
ashutosh-arm, line 89: this bit can be compressed into a single line?
```
y_data = np.asarray([i for i in range(N)], dtype=type)
```

tests/python/relay/test_op_floordiv.py
```diff
@@ -0,0 +1,112 @@
...
+        B: tir.Buffer[(N), "int16"],
+        C: tir.Buffer[(N), "int16"],
+    ):
+        for i in tir.serial(N):
+            with tir.block("A"):
+                v_i = tir.axis.spatial(N, i)
+                A[v_i, 0] = tir.floordiv(C[v_i] - tir.max_value("int16"), divisor)
+                A[v_i, 1] = tir.floormod(C[v_i] - tir.max_value("int16"), divisor)
+                A[v_i + 100, 0] = tir.floordiv(B[v_i], divisor)
+                A[v_i + 100, 1] = tir.floormod(B[v_i], divisor)
```
ashutosh-arm, line 63: Question: second set of checks that use `B` are not harmful, but do they test the blocks touched in this commit? To be clear, not asking to modify it :smile_cat:

src/tir/transforms/lower_intrin.cc
```diff
@@ -214,9 +213,8 @@ class IntrinInjecter : public tvm::arith::IRMutatorWithAnalyzer {
       // If the numerator's lower bound is known, express the floormod
       // in terms of truncmod using only positive operands.
       arith::ConstIntBound const_int_bound = analyzer_->const_int_bound(op->a);
-      if (const_int_bound->min_value != arith::ConstIntBound::kNegInf &&
-          const_int_bound->min_value < 0 &&
-          const_int_bound->min_value > Downcast<IntImm>(tvm::min_value(op->a->dtype))->value) {
+      if (const_int_bound->min_value < 0 &&
+          const_int_bound->min_value > -(Downcast<IntImm>(tvm::max_value(op->a->dtype))->value)) {
```
ashutosh-arm, line 217: Thanks for reusing the existing APIs to be consistent :sweat_smile:

ashutosh-arm (commented):
Thanks for this revision and tests covering dtypes. Some nits below.

tests/python/relay/test_op_floordiv.py
neildhickey: The was primarily to check the behavior of the floormod and floordiv over a random sampling of values across the whole range of integer. I think in the specific case of this test it won't be able to determine the bounds to anything other than min_value(type) max_value(type) so I suppose all it verifies is that the check isn't broken for smaller integer sizes.

tqchen:
To followup on this. Is the bound check (of Inf) necessary ? Since ConstIntBound always returns a valid bound, with special value of kPosInf and kNegInf that are symmetric. 

If we are only looking at negative values, seems that checking the bound in range is not necessary, as long as the value is in the range.

@Lunderberg

Lunderberg:
I think the check on the lower bound is required, though primarily to avoid unnecessary offsets when the bound is unknown.  If the bounds of a signed 32-bit argument `x` are unknown, then the argument would be offset to `x + 2**31`, subtracting off the minimum value of an int32.

tqchen:
get it, thanks for the explaination

ashutosh-arm (approved):
Thanks @tqchen @Lunderberg for the reviews. Congratulations on getting CI to work. It was a tough one @neildhickey.

PR #13919, opened 2023-02-06
[OpenCL] Refactor OpenCL init function

echuraev:
On the platforms with several OpenCL platforms (e.g. Intel CPU and NVidia GPU) it was possible to use OpenCL device only from one platform. And in case when Intel was the first in the platforms list than it wasn't possible to run model on NVidia GPU.

In this PR the init function was modified and now it is possible to use OpenCL devices from different platforms. In case when there are several GPU accelerators then it is possible to select one of them. You can use device id to select GPU device. On the code below the device names of two OpenCL devices are printed:
```python
import tvm

print("opencl 0: ", tvm.opencl(0).device_name)
print("opencl 1: ", tvm.opencl(1).device_name)
```

In case then the machine doesn't contain any GPUs then we will try to use CPU if OpenCL runtime is available.

### DIFF
--- a/src/runtime/opencl/opencl_common.h
+++ b/src/runtime/opencl/opencl_common.h
@@ -221,16 +221,14 @@ class OpenCLWorkspace : public DeviceAPI {
  public:
   // type key
   std::string type_key;
-  // global platform id
-  cl_platform_id platform_id;
-  // global platform name
-  std::string platform_name;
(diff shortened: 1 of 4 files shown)


echuraev:
cc: @masahi, @elvin-n, @srkreddy1238

masahi (approved):

PR #13920, opened 2023-02-06
[CMSIS-NN] Add int16 add and mul operator support

FranklandJack:
The CMSIS-NN backend will now partition quantized `int16` additions and multiplication operators and emit calls to `arm_elementwise_add_s16` and `arm_elementwise_mul_s16` respectively during codegen.

Because `arm_elementwise_mul_s16` and `arm_elementwise_add_s16` do not handle non-zero shift parameters at present, for non-zero zero point values (which map directly to shift in the CMSIS-NN backend) partitioning fails and we fall back on the regular C codegen path.

This patch also adds `int16` tests, including testing for the non-zero zero point edge case described above.

### DIFF
--- a/python/tvm/relay/op/contrib/cmsisnn.py
+++ b/python/tvm/relay/op/contrib/cmsisnn.py
@@ -340,20 +340,37 @@ def pattern_table():
 
         arg0 = binary_op.args[0]
         arg1 = binary_op.args[1]
-        both_args_scalar = False
+
+        # Check arguments are not scalar.
         if (
             isinstance(arg0, tvm.relay.expr.Constant)
             and len(arg0.checked_type.shape) == 0
             and isinstance(arg1, tvm.relay.expr.Constant)
             and len(arg1.checked_type.shape) == 0
         ):
-            both_args_scalar = True
+            return False
 
-        return (
-            arg0.checked_type.dtype == "int8"
-            and arg1.checked_type.dtype == "int8"
-            and not both_args_scalar
-        )
+        arg0_type = arg0.checked_type.dtype
+        arg1_type = arg1.checked_type.dtype
+
+        # Check arguments are of valid type.
+        if arg0_type not in ["int8", "int16"]:
+            return False
+
+        # Check arguments are the same type.
+        if arg0_type != arg1_type:
+            return False
+
+        # Check zero points are non-zero (arm_elementwise_(add|mul)_s16 does not
+        # handle non-zero zero points).
+        if arg0_type == "int16" and str(binary_op.op.name) in ["qnn.add", "qnn.mul"]:
+            arg_0_zero_point = binary_op.args[3].data.numpy()
+            arg_1_zero_point = binary_op.args[5].data.numpy()
+            output_zero_point = binary_op.args[7].data.numpy()
+            if any([arg_0_zero_point, arg_1_zero_point, output_zero_point]):
+                return False
+
+        return True
 
     return [
         ("cmsis-nn.qnn_conv2d", qnn_conv2d_pattern(with_pad=True), check_qnn_conv2d_pad),
--- a/src/relay/backend/contrib/cmsisnn/relay_to_tir.cc
+++ b/src/relay/backend/contrib/cmsisnn/relay_to_tir.cc
@@ -112,6 +112,21 @@ class RelayToTIRVisitor : public MixedModeMutator {
     ir_module_->Add(global_var, replacement_func);
   }
 
+  auto GetIntMinMax(int bit_width) {
+    const int32_t min =
+        (bit_width == 8) ? std::numeric_limits<int8_t>::min() : std::numeric_limits<int16_t>::min();
+    const int32_t max =
+        (bit_width == 8) ? std::numeric_limits<int8_t>::max() : std::numeric_limits<int16_t>::max();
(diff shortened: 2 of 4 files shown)


FranklandJack:
@ashutosh-arm
@Mousius

Would be great to get some reviews on this if/when you folk have time! 😀

python/tvm/relay/op/contrib/cmsisnn.py
```diff
@@ -340,20 +340,39 @@ def check_qnn_binary_op(pattern):
...
+        # handle non-zero zero points).
+        if arg0_type == "int16" and str(binary_op.op.name) in ["qnn.add", "qnn.mul"]:
+            arg_0_zero_point = binary_op.args[3].data.numpy()
+            arg_1_zero_point = binary_op.args[5].data.numpy()
+            output_zero_point = binary_op.args[7].data.numpy()
+            if arg_0_zero_point or arg_1_zero_point or output_zero_point:
+                return False
+
+        # If we got this far we've met all the conditions for binary op to be
+        # supported by cmsis-nn.
```
Mousius, line 374: I don't mind the other comments, although I think they could equally be expressed by variables, but this one seems a bit obvious?

tests/python/contrib/test_cmsisnn/test_binary_ops.py
```diff
@@ -153,6 +153,92 @@ def test_op_int8(
...
+    cmsisnn_mod = cmsisnn.partition_for_cmsisnn(orig_mod)
+
+    # validate pattern matching
+    if input_0_zero_point or input_1_zero_point or output_zero_point:
+        # arm_elementwise_(mul|add)_s16 does not support non-zero shifts in any
+        # argument, in this edge case we fall back on the non partitioned
+        # implementation.
+        assert_no_external_function(cmsisnn_mod)
+    else:
+        assert_partitioned_function(orig_mod, cmsisnn_mod)
```
Mousius, line 220: The branch here makes it worth a separate test case, if we check the module doesn't change we don't need to run it?

Mousius (changes requested):
Just a few minor points @FranklandJack, otherwise seems sensible to me 😸

python/tvm/relay/op/contrib/cmsisnn.py
```diff
@@ -340,20 +340,39 @@ def check_qnn_binary_op(pattern):
...
+        if arg0_type != arg1_type:
+            return False
+
+        # Check zero points are non-zero (arm_elementwise_(add|mul)_s16 does not
+        # handle non-zero zero points).
+        if arg0_type == "int16" and str(binary_op.op.name) in ["qnn.add", "qnn.mul"]:
+            arg_0_zero_point = binary_op.args[3].data.numpy()
+            arg_1_zero_point = binary_op.args[5].data.numpy()
+            output_zero_point = binary_op.args[7].data.numpy()
+            if arg_0_zero_point or arg_1_zero_point or output_zero_point:
```
Mousius, line 370:
```suggestion
            if any(arg_0_zero_point, arg_1_zero_point, output_zero_point):
```

python/tvm/relay/op/contrib/cmsisnn.py
FranklandJack: Sure seems fair, have removed.

tests/python/contrib/test_cmsisnn/test_binary_ops.py
FranklandJack: I've split this into two test cases.

python/tvm/relay/op/contrib/cmsisnn.py
FranklandJack: Nice! Have fixed.

src/relay/backend/contrib/cmsisnn/relay_to_tir.cc
```diff
@@ -587,10 +587,11 @@ class RelayToTIRVisitor : public MixedModeMutator {
   }
 
   void EmitMul(const GlobalVar& global_var, const Expr& expr) {
-    int32_t output_min = std::numeric_limits<int8_t>::min();
-    int32_t output_max = std::numeric_limits<int8_t>::max();
     const auto& pattern = ParseBinaryElementwiseOpClipPattern(expr);
     Call mul_call = pattern.binary_op;
+    const auto bit_width = mul_call->type_as<TensorTypeNode>()->dtype.bits();
+    int32_t output_min = -(1 << (bit_width - 1));
+    int32_t output_max = (1 << (bit_width - 1)) - 1;
```
ashutosh-arm, line 594: nit: Can we save`std::numeric_limits()` as part of the class and use it where needed? Looks more standard than bit based calculations.

ashutosh-arm (changes requested):
Looks good overall @FranklandJack. Just a nit below.

src/relay/backend/contrib/cmsisnn/relay_to_tir.cc
FranklandJack: Yeah good shout. I've tried to replace some code duplication throughout this file, so let me konw if that's what you had in mind :D

ashutosh-arm (approved):
LGTM @FranklandJack! Thanks @Mousius for the reviews.

Mousius (approved):

PR #13922, opened 2023-02-06
[Arith] Updated incorrect simplification rule

Lunderberg:
The rules that rewrite `min(floordiv(x + (A-1), A) * A, max(x, A))` and `min(truncdiv(x + (A-1), A) * A, max(x, A))` into `max(x, A)` did not have sufficiently tight bounds.  The `truncdiv` rule required that `x >= 0`, while the `floordiv` rule had no requirement on `x`.  In both cases, the simplification was incorrect when `x==0`, as it would result in a rewrite from `min(0, max(0, A))` into `max(0, A)`.

This commit updates the rules to require that `x >= 0` for each of these rules.


tqchen:
cc @wrongtest-intellif

wrongtest-intellif:
> `x >= 0`

Hi :) , is it a typo for `x > 0`? @Lunderberg

wrongtest-intellif (approved):
LGTM!  

Another helper view is when `x % A == 0`, `min(ceildiv(x, A), max(x, a))` ==  `x`, thus we should exclude the case `x==0` when `max(x, A) != x`

PR #13923, opened 2023-02-06
[Arith][UnitTest] Parametrize tests of RewriteSimplifier

Lunderberg:
Split out each of the simplification test cases for RewriteSimplifier into independent parametrized unit tests.


Lunderberg:
This arose from some experiments regarding `RewriteSimplifier` that didn't end up materializing.  During development, it had become difficult to see the effect of changes, because a failure in simplifying one expression preventing any further expressions from being tested in that unit test.

masahi (approved):

PR #13924, opened 2023-02-06
[TVMScript,Fix] Fix findsource when classes are indented

tkonolige:
Leaving class definitions was not correctly handled when recreating scoping information. The fix correctly pops scope whenever the indentation level becomes less than the current scope. I believe this was introduced in #12978.

@cyx-6 @junrushao


cyx-6 (approved):
Hi @tkonolige! Thanks for reporting this bug. The fix looks good to me.

PR #13925, opened 2023-02-07
[Metaschedule,Fix] Move x86 feature detection to target directory for metaschedule

tkonolige:
Metaschedule (space generation specifically) requires python functions registered in TOPI to correctly detect targets. If topi was not manually imported then we would get a crash. Fix this by moving the target feature detection to `python/tvm/target/x86.py`.

Test script:
```
import tvm
from tvm import meta_schedule as ms

@tvm.script.ir_module
class matmul:
    @T.prim_func
    def matmul_impl(
        A: T.Buffer[(12, 196, 64), "float32"],
        B: T.Buffer[(12, 64, 196), "float32"],
        C: T.Buffer[(12, 196, 196), "float32"],
    ):
        T.func_attr({"global_symbol": "main", "tir.noalias": True})
        for b in range(12):
            for i in range(196):
                for j in range(196):
                    for k in range(64):
                        with T.block("main"):
                            vb, vi, vj, vk = T.axis.remap("SSSR", [b, i, j, k])
                            with T.init():
                                C[vb, vi, vj] = 0.0
                            C[vb, vi, vj] += A[vb, vi, vk] * B[vb, vk, vj]

db = ms.tune_tir(matmul, "llvm -mcpu=znver3 -num-cores=1", "tune_dir", 512)
```

@zxybazh

### DIFF
--- a/python/tvm/meta_schedule/__init__.py
+++ b/python/tvm/meta_schedule/__init__.py
@@ -52,3 +52,7 @@ from .tir_integration import tune_tir
 from .tune import tune_tasks
 from .tune_context import TuneContext
 from .utils import derived_object
+
+# Space generator requires that vnni, avx detection is available. These
+# functions are registered in the topi module.
+from .. import topi
--- a/python/tvm/meta_schedule/__init__.py
+++ b/python/tvm/meta_schedule/__init__.py
@@ -52,7 +52,3 @@ from .tir_integration import tune_tir
 from .tune import tune_tasks
 from .tune_context import TuneContext
 from .utils import derived_object
-
-# Space generator requires that vnni, avx detection is available. These
-# functions are registered in the topi module.
-from .. import topi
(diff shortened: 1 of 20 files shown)


junrushao:
I agree that some functionalities (target detection) which now exist in TOPI are useful to MetaSchedule, and enabling this is the common goal we both share. The limitation of this particular PR is that it introduces a huge dependency change that makes one subpackage to depend on the other, which was designed to be decoupled originally. Therefore, would you like to think about potential alternative approaches that allow us to enable the particular functionality in your mind without having to create unnecessary dependency? Thanks a lot!

tkonolige:
How about if I move these functions to `python/tvm/target/x86.py`?

junrushao:
That makes sense! Thanks for updating it!

junrushao (approved):
LGTM! Please update the title accordingly

PR #13931, opened 2023-02-08
[ETHOSN] Support for addition with constant input

ashutosh-arm:
After NPU driver 22.11 update, constants can be provided as inputs to addition operator.
This commit traverses constants to update operand entries for it and adds tests to verify these cases.

cc @lhutton1

### DIFF
--- a/src/relay/backend/contrib/ethosn/codegen.cc
+++ b/src/relay/backend/contrib/ethosn/codegen.cc
@@ -192,9 +192,7 @@ void InferTensorsVisitor::VisitExpr_(const CallNode* cn) {
 
 void ConstructNetworkVisitor::VisitExpr_(const ConstantNode* cn) {
   Constant constant = GetRef<Constant>(cn);
-  std::cout << "Traversing constant..." << std::endl;
   if (tensor_table_.count(constant)) {
-    std::cout << "Found constant...." << std::endl;
     sl::TensorInfo tensor_info = tensor_table_[constant][0];
     sl::TensorAndId<sl::Constant> tensor_and_id =
         sl::AddConstant(network_, tensor_info, constant->data->data);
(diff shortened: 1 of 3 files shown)


tests/python/contrib/test_ethosn/test_addition.py
```diff
@@ -173,6 +173,61 @@ def test_addition_both_inputs_constants(
     assert "qnn.add" not in mod.astext(False)
 
 
+@requires_ethosn
+@pytest.mark.parametrize("dtype", ["uint8", "int8"])
+@pytest.mark.parametrize(
+    "lhs_shape,lhs_is_constant,rhs_shape,rhs_is_constant",
+    [
+        ((1, 4, 4, 8), False, (1, 4, 4, 8), True),
+        ((1, 16, 12, 4), True, (1, 16, 12, 4), False),
```
lhutton1, line 182: It might also be good to test broadcast shapes not covered by `qnn_add_to_depthwise` e.g. ((1, 4, 4, 8), False, (1, 1, 4, 8), True), I believe this should be supported?

src/relay/backend/contrib/ethosn/codegen.cc
```diff
@@ -190,6 +190,19 @@ void InferTensorsVisitor::VisitExpr_(const CallNode* cn) {
   }
 }
 
+void ConstructNetworkVisitor::VisitExpr_(const ConstantNode* cn) {
+  Constant constant = GetRef<Constant>(cn);
+  std::cout << "Traversing constant..." << std::endl;
```
lhutton1, line 195: Nit: remove prints

lhutton1 (commented):
Just a couple of small things :)

tests/python/contrib/test_ethosn/test_addition.py
lhutton1: Discussed this offline - it's not something that's currently supported by the support library

src/relay/backend/contrib/ethosn/codegen.cc
ashutosh-arm: Done!

lhutton1 (approved):

lhutton1:
Thanks @ashutosh-arm!

PR #12043, opened 2022-07-08
[MetaSchedule][Test] Add unittests for C2D

junrushao1994:

PR #13936, opened 2023-02-09
[Arith] Simplifications for floormod(x, 2)

Lunderberg:
Because `floormod(x,2)` has only two possible values, it can be simplified more aggressively than most FloorMod expressions.  The additional simplifications are derived from `floormod(x,2) + floormod(x+1,2) == 1`, which is true for denominator `2`, along with the usual `floordiv(x,2)*2 + floormod(x,2) == x`, which is true for all denominators.

This initially arose from an index expression `floormod(x + 1, 2) * 8192`, for `x ∈ [0, 2)`.  This commit allows the expression to be re-written as `x * (-8192) + 8192` and recognized as a strided access.


src/arith/rewrite_simplify.cc
```diff
@@ -278,10 +278,16 @@ PrimExpr RewriteSimplifier::Impl::VisitExpr_(const AddNode* op) {
     TVM_TRY_REWRITE_IF(floordiv(floormod(x, c2) + c1, c2) + floordiv(x, c2), floordiv(x + c1, c2),
                        c2.Eval()->value > 0);
 
+    TVM_TRY_REWRITE(floormod(x + 1, 2) + floormod(x, 2), OneWithTypeLike(x));
```
wrongtest-intellif, line 281: `TVM_TRY_RECURSIVE_REWRITE_IF(floormod(x + c1, 2), floormod(x, 2) * (-1) + 1, floormod(c1.Eval()->value, 2) == 1);` would lead to `(x + 1) % 2` -> `1 - x % 2` 

Then it seems that rule 281 is already covered, if we rewrite recursively?

src/arith/rewrite_simplify.cc
```diff
@@ -473,6 +487,9 @@ PrimExpr RewriteSimplifier::Impl::VisitExpr_(const SubNode* op) {
         floordiv(x - y, c1) * c3 - x * c2, (0 - floormod(x - y, c1) - y) * c2,
         c1.Eval()->value != 0 && c3.Eval()->value == c1.Eval()->value * c2.Eval()->value);
 
+    TVM_TRY_RECURSIVE_REWRITE(floordiv(x + 1, 2) - floormod(x, 2), floordiv(x, 2));
+    TVM_TRY_RECURSIVE_REWRITE(floordiv(x, 2) - floormod(x - 1, 2), floordiv(x - 1, 2));
```
wrongtest-intellif, line 491: Like above, `floormod(x - 1, 2)` -> `1 - x % 2`, then it seems apply `x // 2 + x % 2` rule could achieve `(x + 1) // 2 - 1`

src/arith/iter_affine_map.cc
```diff
@@ -898,6 +898,11 @@ class IterMapRewriter : public ExprMutator {
   PrimExpr SplitFloorModConst(IterSplitExpr lhs, PrimExpr base, PrimExpr rhs);
 
   static void AddToLhs(IterSumExprNode* lhs, IterSplitExpr rhs, int sign) {
+    if (sign < 0 && is_const_int(rhs->extent, 2)) {
```
wrongtest-intellif, line 901: Could we add a cover case for the codepath?

wrongtest-intellif (changes requested):

wrongtest-intellif:
BTW for cancellation like `f(x + c1) - f(x + c2)`, it seems generally the pair may not coincide in summation terms. So we would fail to cancel things like `f(x + c1) + g(x + c3) - f(x + c2)` even if we do have the rules. 

For my curiosity about our tradeoff in simplication, may in the future it allows introduce O(n^2) level rewriting for summation terms to fully utilize rewrite rules?  cc @tqchen @Lunderberg

src/arith/rewrite_simplify.cc
Lunderberg: Good point, and removed.  Testing, it gets the same simplification by first applying the rule on line 289, then 264.

src/arith/rewrite_simplify.cc
Lunderberg: Updated here as well.  The related rule on line 490 with `floordiv(x + 1,2)` is kept, because there isn't an equivalent change to the argument of `floordiv`.

src/arith/iter_affine_map.cc
Lunderberg: Can do, and added!  This was a case that was caught by other unit tests, since the additional `RewriteSimplifier` rules prevented `DetectIterMap` from recognizing some patterns after they had been simplfied, but a specific unit test for this case is better than needing to track it down later.

src/arith/rewrite_simplify.cc
```diff
@@ -454,6 +460,14 @@ PrimExpr RewriteSimplifier::Impl::VisitExpr_(const SubNode* op) {
     TVM_TRY_REWRITE_IF(floordiv(x - y, c1) * c1 - x, 0 - floormod(x - y, c1) - y,
                        c1.Eval()->value != 0);
 
+    TVM_TRY_RECURSIVE_REWRITE(
+        floordiv(x + c1, 2) - floordiv(x + c2, 2),
+        floormod(x, 2) * (floormod(c1, 2) - floormod(c2, 2)) + floordiv(c1, 2) - floordiv(c2, 2));
```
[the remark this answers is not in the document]
Lunderberg: Can do, and thank you for the catch!

Lunderberg:
Thank you for the review, @wrongtest-intellif , and everything should be updated now.

Regarding cases where it can fail to cancel, that definitely can occur.  I have some work on a local branch where I've been trying to make the rewrite rules more general (and to reduce duplication as a benefit), but at the moment it slows down the RewriteSimplifier by about 10x.

One thing that I think could help would be to enforce an ordering between terms for any commutative and associative operator.  For example, right now `(x + y)//4 - (y + x)//4` doesn't cancel out, because the numerators are different.  Sorting by the variable(s) used by each term would at least move related terms together, making it more likely that they'd cancel out.

wrongtest-intellif (approved):
LGTM！

wrongtest-intellif:
some expected tir in ut seems outdated

Lunderberg:
@wrongtest-intellif Thank you for pointing that out, and merged main into the dev branch to resolve.

Hzfengsy:
Could you please fix the CI and we can get it in @Lunderberg

Lunderberg:
@Hzfengsy , thank you on the reminder, and it would be good to get this PR finished and merged in.

@junrushao Could I get your assistance on the last CI error?  The failing unit tests are in [`test_cuda_c2d`](https://github.com/apache/tvm/blob/main/tests/python/unittest/test_meta_schedule_space_cuda.py#L122), added in https://github.com/apache/tvm/pull/12043, but I'm not familiar enough with the meta-schedule search space generation to tell if this is a spurious failure.

It looks like this simplification allowed the search space to go down an entirely different route than the unit test was expecting, as the generated sketch outputs the transpose (`conv2d_transpose_nhwc: T.Buffer((1, 8, 8, 256))`) with a completely different shape than the expected output defined in the unit test (`conv2d_nhwc: T.Buffer((1, 112, 112, 64))`).

Lunderberg:
Thank you, and sorry for pinging you on PTO, and this is much lower priority than PTO rest/relaxation.

I wouldn't guess that it was a parsing issue, as the parsed TIR hasn't changed as a result of the additional simplifications.  So, maybe it would be more likely to be a change in the output of `mod = create_te_workload("C2D", 0)`.

Lunderberg:
Well, first mystery solved.  The test failure was in `test_cuda_t2d`, but I had been comparing against the expected result in `test_cuda_c2d`.  The `test_cuda_t2d` does start with the transpose, so it isn't an issue of a schedule primitive erroneously inserting a transposition.  After that, I was able to track down and fix the test breakage.  The main issue in understanding the test was that I assumed that it was comparing the module in `actual[0].mod` to the expected module.  Instead, it was only using the trace from the sketch, re-applying it with the hard-coded decisions, and then comparing that result to the expected module.

After that, the only difference was resolved by replacing `(var % 2 + 1) // 2` with `(var % 2)`, exactly the simplification enabled by this PR.

Lunderberg:
@tvm-bot rerun

src/arith/iter_affine_map.cc
tqchen: NOTE: this actually is a bug, see https://github.com/apache/tvm/pull/14571

This makes me think that if other rules leads to the regression. Please do check.

tqchen:
Note that this PR introduces a bug in https://github.com/apache/tvm/pull/14571

And the bug was intended to fix a regression, which now comes back. This makes me think that we should revisit other rules here. Since there is a possibility of other issues.

Likely we should remove some of the simplification in favor of iter_map because that is more important

tqchen:
@wrongtest-intellif sorry for following up late, for term cancelations, we have canonical simplifier which should be able to do such cancelation, so likely we don't need to enhance rewrite simplifier too much here.

tqchen:
added more discussions in https://github.com/apache/tvm/pull/14571

PR #13938, opened 2023-02-09
[Hexagon] Updated MapRanges function to perform more accurate ranges calculation for large SIMD

shtinsa:
This change allow align inner dimension if it is not divisible by len(SIMD)

### DIFF
--- a/src/tir/ir/index_map.cc
+++ b/src/tir/ir/index_map.cc
@@ -209,15 +209,23 @@ Array<Range> IndexMapNode::MapRanges(const Array<Range>& ranges, arith::Analyzer
     // For example, [N] mapped through i=>[i//4,i%4] should have shape
     // [ceildiv(N,4), 4].  However, for N<4, this method instead
     // results in a shape [1, N].
+    PrimExpr total_sz(1);
     std::unordered_map<const VarNode*, arith::IntSet> dom_map;
     for (size_t i = 0; i < initial_indices.size(); i++) {
       dom_map[initial_indices[i].get()] = arith::IntSet::FromRange(ranges[i]);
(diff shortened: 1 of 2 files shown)


echuraev:
@tvm-bot rerun

masahi:
Please add a test.

cc @Lunderberg who shared a concern with this patch before.

Lunderberg:
I think this case should throw an error instead at this point, rather than a workaround at this location.  Currently, the behavior of `Inverse` throws an error whenever padding would be introduced, whereas `MapRanges` and `MapShape` allow padding.

If we make this change, the difference in padding semantics between `MapIndices` and `MapShape` would be even larger.  For the transformation `n => [n//8, n%8]` applied to a buffer of shape `[7]`, `MapIndices` would attempt to reshape this into `[1,8]` and throw an error, while `MapShape` would attempt to reshape into `[1,7]` and return a success.  I don't think the divergent semantics is appropriate here.  Instead, I think we should throw the same error that occurs in `IndexMap::Inverse`, and throw an exception in this case.  That is, this case would be an error in the calling scope rather than a use case that `IndexMap` should handle, with padded maps handled through a separate function call.

For the use case where the SIMD size is large relative to the buffer size, don't we need the padding anyways so that the SIMD instruction doesn't overrun the bounds?  I'm wondering if we could find a better place to apply a fix to explicitly allow the IndexMap to introduce padding including in the allocated buffer in all locations where it is used.

shtinsa:
Hello Eric your concern is absolutely correct and I'll try to prepare another solution to resolve the issue

Lunderberg:
Thank you, and let me know if I can help in finding a different place to apply a solution.

PR #13941, opened 2023-02-09
[TVMScript][UX] Introduce decorator for deprecation

junrushao:
This PR introduces a decorator `tvm.ir.base.deprecated`, which emits a
deprecation warning if an outdated API is used, but preserves backward
compatibility by still allowing the API to be used.

For example, currently the preferred way of TIR buffer declaration in
function signature is:

```python
def example(
  A: T.Buffer(...),  # legacy behavior is `T.Buffer[...]`
): ...
```

With this decorator, if a user writes `T.Buffer[...]`, the parser will
still function properly, but emits a warning that guides the user to
adopt `T.Buffer(...)` if possible.

While there is no breaking change at all in this PR, we believe this
is useful to help users upgrade before any breaking change eventually
takes place.

### DIFF
--- a/apps/pt_tvmdsoop/tests/test_as_torch.py
+++ b/apps/pt_tvmdsoop/tests/test_as_torch.py
@@ -52,7 +52,7 @@ def matmul(M: int, N: int, K: int, dtype: str):
 @tvm.script.ir_module
 class ModuleGPU:
     @T.prim_func
-    def main(A: T.Buffer[8, "float32"], B: T.Buffer[8, "float32"]) -> None:
+    def main(A: T.Buffer(8, "float32"), B: T.Buffer(8, "float32")) -> None:
         T.func_attr({"global_symbol": "main", "tir.noalias": True})
         for i_0 in T.thread_binding(2, thread="blockIdx.x"):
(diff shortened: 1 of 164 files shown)


junrushao:
CC: @cyx-6 @wrongtest-intellif @Hzfengsy

spectrometerHBH (approved):

python/tvm/topi/hexagon/qnn/adaptive_avg_pool1d.py
```diff
@@ -1,120 +1,120 @@
-# Licensed to the Apache Software Foundation (ASF) under one
```
Hzfengsy, line 1: Not sure what happens in these files. Seems no changes at all

junrushao: Yep idk why git thinks it's different...I'm going to revert this file anyways

junrushao: I got them reverted and will submit a separate PR

Hzfengsy (approved):

PR #13947, opened 2023-02-10
[TIR][Analysis] Implement IdentifyMemCpy analysis function

Lunderberg:
This commit adds a utility function `tir::IdentifyMemCpy`, which analyzes a `tir::For` loop and determines whether it is equivalent to a memcpy.  If it is equivalent, it will additionally return the source and destination regions of the memcpy.

This utility is initially intended for use in the `LowerAsyncDMA` pass, but may be useful in other areas in the
future.  (e.g. Identifying an entire TIR function as equivalent to memcpy would allow calls to that TIR function to be removed from end-to-end models.)


src/tir/analysis/identify_memcpy.cc
```diff
@@ -0,0 +1,317 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/*!
+ * \file tir/analysis/identify_memcpy.cc
+ * \brief Check if a loopnest is equivalent to memcpy
```
Hzfengsy, line 22:
```suggestion
 * \brief Check if a loop nest is equivalent to memcpy
```

src/tir/analysis/identify_memcpy.cc
```diff
@@ -0,0 +1,317 @@
...
+namespace tvm {
+namespace tir {
+
+std::variant<MemCpyDetails, std::string> IdentifyMemCpyImpl(const For& loop,
+                                                            arith::Analyzer* analyzer) {
+  Map<Var, arith::IntSet> loop_intervals;
+  Map<Var, Range> loop_ranges;
+  PrimExpr total_loop_iterations = 1;
+
+  // Walk through the loopnest, stopping at the first loop whose body
```
Hzfengsy, line 48:
```suggestion
  // Walk through the loop nest, stopping at the first loop whose body
```

src/tir/analysis/identify_memcpy.cc
```diff
@@ -0,0 +1,317 @@
...
+      return static_cast<const std::stringstream&>(
+                 std::stringstream()
+                 << "Mismatch between loop iterations (" << total_loop_iterations
+                 << ") and number of dst indices touched (" << dst_interval
+                 << ".  Equality to prove simplified to " << simplified)
+          .str();
+    }
+  }
+
+  // Thrid check, is there a transformation applied between the input
```
Hzfengsy, line 185:
```suggestion
  // Third check, is there a transformation applied between the input
```

src/tir/analysis/identify_memcpy.cc
```diff
@@ -0,0 +1,317 @@
...
+#include <sstream>
+#include <string>
+#include <variant>
+
+#include "../../arith/ir_visitor_with_analyzer.h"
+
+namespace tvm {
+namespace tir {
+
+std::variant<MemCpyDetails, std::string> IdentifyMemCpyImpl(const For& loop,
```
Hzfengsy, line 42: It's a bit tricky to return the result and error message at the same time. An alternate way is to raise a specific exception and catch from outside.

src/tir/analysis/identify_memcpy.cc
```diff
@@ -0,0 +1,317 @@
...
+  auto result = IdentifyMemCpyImpl(loop, analyzer);
+  if (auto* ptr = std::get_if<MemCpyDetails>(&result)) {
+    return *ptr;
+  } else {
+    return std::nullopt;
+  }
+}
+
+// Expose the IdentifyMemCpy functionality to Python API for purpose
+// of unit testing.
```
Hzfengsy, line 285:
```suggestion
// Expose the IdentifyMemCpy functionality to Python API for purpose of unit testing.
```

tests/python/unittest/test_tir_analysis_identify_memcpy.py
```diff
@@ -0,0 +1,325 @@
...
+# specific language governing permissions and limitations
+# under the License.
+
+import re
+
+import pytest
+
+import tvm
+import tvm.testing
+from tvm import te, topi
```
Hzfengsy, line 24: Please remove this line if it's not used

Hzfengsy (commented):
Otherwise looks good to me. Please update and fix the CI and we can get this in

tests/python/unittest/test_tir_analysis_identify_memcpy.py
Lunderberg: Thank you, and removed.

src/tir/analysis/identify_memcpy.cc
Lunderberg: Looks good, and updated.

src/tir/analysis/identify_memcpy.cc
Lunderberg: Thank you, and fix applied.

src/tir/analysis/identify_memcpy.cc
Lunderberg: Thank you, and fix applied.

src/tir/analysis/identify_memcpy.cc
Lunderberg: Thank you, and fix applied.

src/tir/analysis/identify_memcpy.cc
Lunderberg: Agreed, though I think the `std::variant` avoids most of the usual issues.  Rather than returning both the result and the error message at the same time, it returns either the result or the error message, but never both.  I didn't want to throw an exception, as I expect the "error" case to be much more common, which is why the public-facing API exposes a `std::optional<MemCpyDetails>` instead of the internal variant.

adstraw (approved):
LGTM

Hzfengsy (approved):

PR #13953, opened 2023-02-10
[TensorRT] Fix BiasAdd with correct axis attribute

jsheng-jian:
The original TensorRT BiasAdd converter does not use `axis` attribute from `nn.bias_add` so it only works with `NC, NCHW` and implicit mode (when batch dimension is removed). 
We recently found some NLP and Vision Transformer models are parsed with `axis=-1` which caused reshape error later during conversion.
```
[20:37:10] /home/ubuntu/apache_tvm/tvm/src/runtime/contrib/tensorrt/tensorrt_logger.h:51: Error: ERROR: 4: [graphShapeAnalyzer.cpp::analyzeShapes::1294] Error Code 4: Miscellaneous (IShuffleLayer (Unnamed Layer* 2) [Shuffle]: reshape changes volume. Reshaping [1,768,768,1] to [1,197,768].)
```
This CR adds support of `axis` attribute during conversion and related test case for `axis=-1`

### DIFF
--- a/src/runtime/contrib/tensorrt/tensorrt_ops.cc
+++ b/src/runtime/contrib/tensorrt/tensorrt_ops.cc
@@ -961,11 +961,9 @@ class BiasAddOpConverter : public TensorRTOpConverter {
     int axis = std::stoi(params->node.GetAttr<std::vector<std::string>>("axis")[0]);
     if (axis == -1) {
       // Make sure there are 2 dimensions after channel dimension,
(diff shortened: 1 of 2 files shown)


yongwww (approved):
lgtm

jsheng-jian:
Thanks @yongwww ! Can you merge this PR? It passed the CI.

masahi (approved):

PR #13955, opened 2023-02-10
[microTVM]Refactor crt_config.h header file generation

mehrdadh:
This PR:
1. refactors crt_config.h generation to a single template file and use CMake `configure_file` func to generate in build
2. Refactor Zephyr.cmake, CRT.cmake and Arduino.cmake to Micro.cmake


mehrdadh:
cc @alanmacd @mkatanbaf for review

src/runtime/crt/crt_config.h.template
```diff
@@ -17,47 +17,53 @@
  * under the License.
  */
 
-/*!
- * \file tvm/runtime/crt_config.h.template
- * \brief Template for CRT configuration, to be modified on each target.
+ /*!
```
alanmacd, line 20: nit: something happened with spacing and slashes here

.gitignore
```diff
@@ -277,3 +277,9 @@ gallery/how_to/work_with_microtvm/micro_tvmc.py
 
 # GDB history file
 .gdb_history
+
```
alanmacd, line 280: would it makes sense to output these files into the build/ directory and add that directory as an include instead of putting these files into the source tree?

cmake/modules/StandaloneCrt.cmake
```diff
@@ -15,9 +15,11 @@
 # specific language governing permissions and limitations
 # under the License.
 
-
 if(USE_MICRO)
 
+include(cmake/utils/CRTConfig.cmake)
```
alanmacd, line 20: As we mentioned earlier, this will need coordination with #13909

cmake/modules/Micro.cmake
```diff
@@ -15,8 +15,112 @@
...
+        tvm_micro_add_copy_file(platform_template_deps
+                                ${job_src_base}/${copy_src} ${dest_path})
+      endforeach()
+    endforeach()
+  endforeach()
+
+  add_custom_target(${platform} DEPENDS ${platform_template_deps})
+endfunction()
+
+if(USE_MICRO)
```
alanmacd, line 115: Maybe just one `if(USE_MICRO)` for the whole file

apps/microtvm/zephyr/template_project/src/host_driven/main.c
```diff
@@ -90,9 +89,9 @@ void uart_irq_cb(const struct device* dev, void* user_data) {
       // Write it into the ring buffer.
       g_num_bytes_in_rx_buffer += rx_size;
 
-      if (g_num_bytes_in_rx_buffer > RING_BUF_SIZE_BYTES) {
-        TVMPlatformAbort((tvm_crt_error_t)0xbeef3);
-      }
+      // if (g_num_bytes_in_rx_buffer > RING_BUF_SIZE_BYTES) {
```
alanmacd, line 92: delete commented code if not needed

apps/bundle_deploy/build_model.py
```diff
@@ -18,12 +18,15 @@
...
 import os
+import pathlib
+
 from tvm import relay
 import tvm
 from tvm import runtime as tvm_runtime
 import logging
 from tvm.relay.backend import Runtime
 from tvm.contrib import cc as _cc
+from tvm.micro import copy_crt_config_header
```
alanmacd, line 29: does this add a dependency on USE_MICRO being enabled that wasn't here previously, should this be conditional on that?

apps/microtvm/zephyr_cmsisnn/model/convert_labels.py
```diff
@@ -39,5 +43,13 @@ def create_labels_header(labels_file, output_path):
         header_file.write("};\n")
 
 
+def prepare_crt_config():
+    crt_config_output_path = pathlib.Path(__file__).parent.resolve() / ".." / "crt_config"
```
mkatanbaf, line 47: nit: use either `.parent` or `/".."`

mkatanbaf (approved):

apps/microtvm/zephyr_cmsisnn/model/convert_labels.py
mehrdadh: fixed that

src/runtime/crt/crt_config.h.template
mehrdadh: fixed this

areusch (approved):
thanks @mehrdadh !

alanmacd (approved):

Issue #13957, opened 2023-02-11
[Bug] Crash when applying transform_layout with axis separator


quic-sanirudh:
When we apply `transform_layout` on a write buffer generated by `cache_write` with axis separators, we see an error that the sref pointed by the block does not exist in stmt2ref. The below code reproduces the error:

### Expected behavior

The code should work fine without errors and apply the layout transform with axis separators

### Actual behavior

Code crashes with error `Check failed: (it != self_->stmt2ref.end()) is false: InternalError: The sref points to a statement that does not exist in stmt2ref`

### Environment

Tested on Upstream TVM main with Ubuntu 18.04

### Steps to reproduce

```python
import tvm
import tvm.contrib.hexagon
import tvm.testing
from tvm.script import tir as T


@T.prim_func
def add(
    p0: T.Buffer((T.int64(33), T.int64(128)), "float32"),
    p1: T.Buffer((T.int64(33), T.int64(128)), "float32"),
    T_add: T.Buffer((T.int64(33), T.int64(128)), "float32"),
):
    T.func_attr({"global_symbol": "main", "tir.noalias": True})
    # with T.block("root"):
    for ax0, ax1 in T.grid(T.int64(33), T.int64(128)):
        with T.block("T_add"):
            v_ax0, v_ax1 = T.axis.remap("SS", [ax0, ax1])
            T.reads(p0[v_ax0, v_ax1], p1[v_ax0, v_ax1])
            T.writes(T_add[v_ax0, v_ax1])
            T_add[v_ax0, v_ax1] = p0[v_ax0, v_ax1] + p1[v_ax0, v_ax1]


def transform_fn(x, y):
    return [x // 32, y, tvm.te.AXIS_SEPARATOR, x % 32]


def main():
    sch = tvm.tir.Schedule(add, debug_mask="all")
    block_rv = sch.get_block("T_add")
    sch.cache_write(block_rv, 0, "global")
    sch.transform_layout(block_rv, ("write", 0), transform_fn, pad_value=0.0)


if __name__ == "__main__":
    main()
```

### Triage

* [needs-triage](tir:schedule)

quic-sanirudh:
A temporary fix to this error is in #13958

quic-sanirudh:
Closing in favor of an updated issue #13974

PR #13958, opened 2023-02-11
[TIR] Fix crash on transform_layout

quic-sanirudh:
When `ScheduleTransformLayout` is called, the block corresponding to the BlockRV named `block` is potentially changed. The symbol_table_ however is not updated for changes to BlockRV and hence when the same blockRV `block` is passed to `ScheduleSetAxisSeparator`, the sref corresponding to that is invalid and that results in crash as unknown type_index_

This PR is a temporary fix and a permanent fix would probably need to identify that the BlockRV is not pointing to a valid sref and throw a proper error or update the symbol_table_ appropriately to avoid this error completely.

### DIFF
--- a/python/tvm/tir/schedule/schedule.py
+++ b/python/tvm/tir/schedule/schedule.py
@@ -2760,6 +2760,7 @@ class Schedule(Object):
 
         """
         block = self._normalize_block_arg(block)
+        block_name = self.get(block).name_hint
         buffer_index_type, buffer_index, buffer_obj = self._normalize_buffer_arg(block, buffer)
 
         ndim = len(buffer_obj.shape)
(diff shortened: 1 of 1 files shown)


quic-sanirudh:
This PR provides a temporary fix to the error seen in #13957

quic-sanirudh:
cc @vinx13 @Lunderberg

quic-sanirudh:
@tvm-bot rerun

quic-sanirudh:
> This PR is a temporary fix and a permanent fix would probably need to identify that the BlockRV is not pointing to a valid sref and throw a proper error

Just realized that if we enable `debug_mask="all"` when creating the `Schedule`, we do get a proper error message related to the sref, so maybe this fix could be considered a permanent fix?

quic-sanirudh:
@tvm-bot rerun

vinx13:
I think as a permanent fix, the schedule primitive should always guarantee correct remapping of block reference. This is achieved by setting correct `block_sref_reuse` https://github.com/apache/tvm/blob/main/src/tir/schedule/primitive/layout_transformation.cc#L1165

quic-sanirudh:
> I think as a permanent fix, the schedule primitive should always guarantee correct remapping of block reference. This is achieved by setting correct `block_sref_reuse` https://github.com/apache/tvm/blob/main/src/tir/schedule/primitive/layout_transformation.cc#L1165

Thanks @vinx13  for the suggestion. I'll try to update this PR with the permanent fix.

quic-sanirudh:
Closing in favor of #14023

PR #13959, opened 2023-02-11
[Minor] Change file mode 755 -> 644; EOL CRLF -> LF

junrushao:
During our previous PR (https://github.com/apache/tvm/pull/13941), we happened to spot that some files are mistakenly in 755 mode, or using CRLF EOL, so we decided to have a separate PR for a quick fix. In the long run, we might want to set up a linter, but those are not huge deals anyways :-)


junrushao:
CC: @Hzfengsy

tqchen (approved):

PR #13960, opened 2023-02-11
[TVMScript] Enable Safe Autocasting in BufferStore

junrushao:
Previously, when writing a BufferStore statement in TVMScript, dtype
mismatch will not be detected but silently allowed.

For example, in the snippet below:

```python
buffer[indices] = RHS
```

even if the elements of `buffer` is floating point numbers, while RHS
is a handle, there is no mechanism to warn, fix, or error out.

This PR introduces auto casting mechanism in TVMScript to ensure that
the dtype of RHS matches LHS. When there is possible precision loss, it
will emit a warning so that the users could adjust accordingly.

### DIFF
--- a/src/script/ir_builder/tir/ir.cc
+++ b/src/script/ir_builder/tir/ir.cc
@@ -492,6 +492,34 @@ Var EnvThread(String thread_tag) {
 }
 
 void BufferStore(Buffer buffer, PrimExpr value, Array<PrimExpr> indices) {
+  runtime::DataType buffer_dtype = buffer->dtype;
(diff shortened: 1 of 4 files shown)


MasterJH5574 (approved):

Hzfengsy (approved):
Please update the PR description and we can get it in

junrushao:
I will tweak it a bit by adding some more warning messages today and ping you guys back tomorrow

junrushao:
@Hzfengsy @MasterJH5574 I added an extra warning when there is potential precision loss, for example, casting floating point numbers to integers. It should be good to go

PR #13962, opened 2023-02-11
[TVMScript] Deterministic function ordering

junrushao:
This PR makes IRModule printing deterministic by fix the function ordering. The "main" function always comes last, while other functions should be ordered alphabetically.

### DIFF
--- a/src/script/printer/ir/ir.cc
+++ b/src/script/printer/ir/ir.cc
@@ -26,27 +26,47 @@ namespace printer {
 
 TVM_REGISTER_NODE_TYPE(IRFrameNode);
 
+struct SortableFunction {
+  int priority;
+  GlobalVar gv;
+  BaseFunc func;
+
+  explicit SortableFunction(const std::pair<GlobalVar, BaseFunc>& obj)
+      : priority(0), gv(obj.first), func(obj.second) {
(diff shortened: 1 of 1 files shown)


src/script/printer/ir/ir.cc
```diff
@@ -26,21 +26,31 @@ namespace printer {
...
+
 TVM_STATIC_IR_FUNCTOR(IRDocsifier, vtable)
     .set_dispatch<IRModule>("", [](IRModule mod, ObjectPath p, IRDocsifier d) -> Doc {
       std::vector<std::pair<GlobalVar, BaseFunc>> functions{mod->functions.begin(),
                                                             mod->functions.end()};
       // print "main" first
       std::sort(functions.begin(), functions.end(), [](const auto& lhs, const auto& rhs) {
-        String lhs_name = lhs.first->name_hint;
-        String rhs_name = rhs.first->name_hint;
-        if (lhs_name == "main") {
```
tqchen, line 37: nit: we can cache sort key by generating a tuple first then sort instead of doing string look up compare each time.

junrushao: Thanks! Updated accordingly

tqchen (approved):

PR #13963, opened 2023-02-12
[Fix][MetaSchedule] RPCRunner timeout when queueing up

MasterJH5574:
This PR fixes the timeout rule of MetaSchedule RPCRunner.

Prior to this PR, the RPCRunner sets a timeout threshold for jobs submitted to popen pool. As a result, the jobs are timed since the time that they are sent to the remote side.

Consider the case where there is only a single device for measurement. In this case, all jobs can only be executed serially and jobs must queue up. Therefore, the previous timeout configuration means the time spent on queueing will also be counted. This causes some jobs, in the worst cases, gets timeout without even started to execute, and has negative impacts on RPC MetaSchedule tuning, from the perspectives of both efficiency and result performance.

To address the problem, this PR
* removes the RPCRunner timeout requirement for PopenExecutor. Since now, the RPCRunner timeout will only be raised from the RPC server side.
* changes the default maximum number of RPCRunner workers to 1, which is to avoid the case where too many jobs waiting in line.

Co-authored-by: Bohan Hou <32121147+spectrometerHBH@users.noreply.github.com>

### DIFF
--- a/python/tvm/meta_schedule/runner/rpc_runner.py
+++ b/python/tvm/meta_schedule/runner/rpc_runner.py
@@ -27,7 +27,6 @@ from tvm.runtime import Device, Module
 from ..logging import get_logger
 from ..profiler import Profiler
 from ..utils import (
-    cpu_count,
     derived_object,
     get_global_func_on_rpc_session,
     get_global_func_with_default_on_worker,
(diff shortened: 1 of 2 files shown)


MasterJH5574:
cc @junrushao @tqchen

junrushao (approved):

junrushao:
please fix the lint

MasterJH5574:
Note: I reverted the removal of `timeout_sec` field of RPCRunnerFuture, as I found it has uses on RPCRunnerMicro and Hexagon side, which I’m not familiar with and does not know the potential implications.
https://github.com/apache/tvm/blob/5477802ec62e9775fdfc74e64378f316f299a41d/python/tvm/contrib/micro/meta_schedule/rpc_runner_micro.py#L107
https://github.com/apache/tvm/blob/5477802ec62e9775fdfc74e64378f316f299a41d/python/tvm/contrib/hexagon/meta_schedule.py#L103

PR #13964, opened 2023-02-12
[Relay][Pass] Separate out the graph partitioning code from fuse_ops.cc

Hzfengsy:
The current `fuse_ops.cc` contains the following parts:
1. `IndexedForwardGraph` and `DominatorTree` which are used for graph partitioning
2. A Relay Expr visitor to create the `DominatorTree`
3. A Relay Expr mutator to fuse the ops

This PR separates the graph partitioning code from `fuse_ops.cc` and moves it to the analysis folder, for:
1. Better code organization and readability as the graph partitioning code is quite long and not directly related to the fusion mutator
2. Possible reuse opportunities for other fusion passes in Relax

NOTE: we won't bring relax fusion in `main` branch for now, but this pr is still reasonable for `main`.

cc @tqchen

### DIFF
--- a/src/relay/analysis/graph_partitioner.cc
+++ b/src/relay/analysis/graph_partitioner.cc
@@ -1,0 +1,334 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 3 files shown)


junrushao (commented):
looks like the only dependency to relay is `op_attr_types`, so shall we instead move them to `src/ir/`?

tqchen:
I think it is a good positive change overall. We can think about future reorganizations. Perhaps some extra namespace for related support data structures.

tqchen (approved):

PR #13965, opened 2023-02-12
[TVMScript][Fix] Print Multi-line String as Metadata

junrushao:
Multi-line strings might make less sense to be printed out by default, as they could be LLVM snippets, CUDA source code and anything hard to comprehend but easy to mess up with the TVMScript itself. Therefore, this PR is introduced to print them as metadata by default.

### DIFF
--- a/python/tvm/script/parser/core/entry.py
+++ b/python/tvm/script/parser/core/entry.py
@@ -41,10 +41,12 @@ def parse(program: Union[doc.AST, Any, str], extra_vars: Dict[str, Any] = None) 
         The parsed TVMScript program.
     """
     if extra_vars is None:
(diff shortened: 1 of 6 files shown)


junrushao:
CC: @wrongtest-intellif @Hzfengsy @spectrometerHBH

Hzfengsy (approved):

junrushao:
The unittest below has been testing this case:

```
tests/python/unittest/test_tvmscript_roundtrip.py::test_roundtrip[string_annotation_escaping]
```

tqchen:
Thanks @junrushao , i would even count this as a bugfix, since multi-line string printing would results in script that cannot be parsed

PR #13967, opened 2023-02-12
Fix out of bound enum conversion

tqchen:
This PR fixes Wenum-constexpr-conversion which is introduced in recent version of clang. Explciitly use integer instead of enum to store the device that may contain invalid options.

### DIFF
--- a/include/tvm/runtime/ndarray.h
+++ b/include/tvm/runtime/ndarray.h
@@ -42,14 +42,6 @@ namespace tvm {
 // alias DLDevice
 using Device = DLDevice;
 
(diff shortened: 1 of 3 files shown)


tqchen:
cc @Wheest background https://discuss.tvm.apache.org/t/tvm-web-build-error-and-unrecognised-target/14324

Wheest:
> cc @Wheest background https://discuss.tvm.apache.org/t/tvm-web-build-error-and-unrecognised-target/14324

Cheers, pulling this PR allows me to run `make` in `web/` without the error being triggered.

junrushao (approved):
LGTM!

PR #13971, opened 2023-02-13
[TVMScript] Encourage using T.Buffer directly

junrushao:
Previously there are two equivalent ways of declaring a buffer in TVMScript:

```python
buffer = T.buffer_decl(...)
buffer = T.Buffer(...)
```

The two approaches are aliases to each other and are essentially the same in implementation. Therefore, this PR encourages to use `T.Buffer` as the recommended approach as it's a bit shorter. Meanwhile, `T.buffer_decl` will continue to be valid in TVMScript, but a deprecation warning will be emitted if its used.


junrushao:
CC: @cyx-6

cyx-6 (approved):
LGTM!

PR #13972, opened 2023-02-13
[CLML] Changes corresponding to OpenCL workspace refactorization

srkreddy1238:
OpenCL context accessing is changed while refactoring OpenCL backend.

### DIFF
--- a/src/runtime/contrib/clml/clml_runtime.cc
+++ b/src/runtime/contrib/clml/clml_runtime.cc
@@ -144,6 +144,8 @@ class CLMLRuntime : public JSONRuntimeBase {
     workspace = cl::OpenCLWorkspace::Global();
     workspace->Init();
     tentry = workspace->GetThreadEntry();
+    device_id = workspace->GetCLDeviceID(tentry->device.device_id);
+    platform_id = workspace->device_to_platform[device_id];
 
     if (!ExtensionStringPresent()) {
       LOG(WARNING) << "CLML Runtime Init: Qualcomm extn not present.\n";
@@ -492,7 +494,7 @@ class CLMLRuntime : public JSONRuntimeBase {
       JSONGraphNode node = it->second.second;
       void* node_data = nullptr;
 
-      allocateTensorMemory(h_ClmlIntf, workspace->context, tensor_desc);
+      allocateTensorMemory(h_ClmlIntf, workspace->contexts[platform_id], tensor_desc);
 
       if (node.GetOpType() == "const") {
         node_data = data_entry_[EntryID(it->first, 0)]->data;
@@ -581,9 +583,6 @@ class CLMLRuntime : public JSONRuntimeBase {
 
   bool ExtensionStringPresent(void) {
     cl_int result = 0;
-    if (workspace->platform_id == nullptr) {
-      return 0;
-    }
     size_t reqd_size = 0;
     cl_device_id device_id = workspace->devices[workspace->GetThreadEntry()->device.device_id];
     result = clGetDeviceInfo(device_id, CL_DEVICE_EXTENSIONS, 0, NULL, &reqd_size);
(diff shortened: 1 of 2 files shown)


src/runtime/contrib/clml/clml_runtime.cc
```diff
@@ -581,9 +583,6 @@ class CLMLRuntime : public JSONRuntimeBase {
 
   bool ExtensionStringPresent(void) {
     cl_int result = 0;
-    if (workspace->platform_id == nullptr) {
-      return 0;
-    }
     size_t reqd_size = 0;
     cl_device_id device_id = workspace->devices[workspace->GetThreadEntry()->device.device_id];
```
echuraev, line 587: For now, you don't check that the OpenCL device is available. So it might be a runtime error. I suppose that you can add a new method to `OpenCLWorkspace` e.g.:
```c++
bool IsDeviceExists(int device_id) {
    return device_id < devices.size();
}
```

In this case, you'll be able to softly check if OpenCL device available. And use this method instead of removed if condition. 

Also, you can use `GetCLDeviceID` for getting `device_id`:
```suggestion
    cl_device_id device_id = GetCLDeviceID(workspace->GetThreadEntry()->device.device_id);
```

src/runtime/contrib/clml/clml_runtime.cc
```diff
@@ -144,6 +144,8 @@ class CLMLRuntime : public JSONRuntimeBase {
     workspace = cl::OpenCLWorkspace::Global();
     workspace->Init();
     tentry = workspace->GetThreadEntry();
+    device_id = workspace->GetCLDeviceID(tentry->device.device_id);
+    platform_id = workspace->device_to_platform[device_id];
```
echuraev, line 148: If there is no OpenCL devices presented, then an exception will be generated here. Is it ok to you? Or I suggest moving these two lines after if condition `if (!ExtensionStringPresent())`.

echuraev (commented):
Sorry, I forgot to update OpenCLML part in the PR with OpenCL workspace refactoring. Thank you for the PR.

src/runtime/contrib/clml/clml_runtime.cc
srkreddy1238: It's unlikely to have multiple OpenCL devices or no OpenCL device on Snapdragon for now.

Anyway, I will keep the initialization flow clean by looking for ```ExtensionStringPresent``` followed by other assignments.

src/runtime/contrib/clml/clml_runtime.cc
```diff
@@ -581,11 +583,12 @@ class CLMLRuntime : public JSONRuntimeBase {
...
   bool ExtensionStringPresent(void) {
     cl_int result = 0;
-    if (workspace->platform_id == nullptr) {
-      return 0;
-    }
     size_t reqd_size = 0;
-    cl_device_id device_id = workspace->devices[workspace->GetThreadEntry()->device.device_id];
+    if (!workspace->IsDeviceExists(workspace->GetThreadEntry()->device.device_id)) {
+      LOG(FATAL) << "OpenCL device not found";
+    }
```
echuraev, line 589: Are you sure that you need to throw exception from this method and from `init` function? I thought that you need just check if OpenCL device is existed (or in previous code `platform_id`) and if it is not presented then continue to work.

Because in case if exception is ok for this function then this check doesn't really need because `GetCLDeviceID` will generate exception about invalid device id.

echuraev (approved):
In general LGTM. One question and after the answer or adding some changes (up to you, I'm ok with current version of code) it can be merged.

src/runtime/contrib/clml/clml_runtime.cc
srkreddy1238: There no point proceeding without a valid CL device for CLML too. I could remove this check.

PR #13975, opened 2023-02-13
[TOPI][Fix] Pool must return error if layout is tiled on H, W, D dimensions

psrivas2:
`topi.nn.poolNd` implementation does not support tiling on H, W or D dimensions. However, current implementation has a bug where no error is thrown if the layout is tiled along any of these dimensions. This PR fixes it.


psrivas2:
cc @masahi

Hzfengsy (approved):

masahi (approved):

PR #13977, opened 2023-02-13
[Diagnostic] Support constructing Diagnostic Error through ObjectRef

MasterJH5574:
This PR supports creating a diagnostic error from an arbitrary Object.

Given that we are bringing the diagnostic error for general uses in the long term, in which case
* not every Expr necessarily has a span,
* we have well-implemented elegant location-aware printer for an object itself,
* we may need to print some object other than Expr, or does even not have a span field,

we support diagnostic error with arbitrary object to denote the location.

Co-authored-by: Tianqi Chen <tianqi.tchen@gmail.com>

### DIFF
--- a/include/tvm/ir/diagnostic.h
+++ b/include/tvm/ir/diagnostic.h
@@ -56,6 +56,14 @@ class DiagnosticNode : public Object {
   DiagnosticLevel level;
   /*! \brief The span at which to report an error. */
   Span span;
+  /*!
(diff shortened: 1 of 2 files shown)


MasterJH5574:
@tvm-bot rerun

MasterJH5574:
@tvm-bot rerun

Hzfengsy (approved):

PR #13997, opened 2023-02-15
[microNPU] Sum legalization support

Alexey-Yazev:
Supports legalizing a relay sum operation to an equivalent series of NPU operations. It supports case with int8 output type and channel axis.

### DIFF
--- a/python/tvm/relay/backend/contrib/ethosu/te/pooling.py
+++ b/python/tvm/relay/backend/contrib/ethosu/te/pooling.py
@@ -110,11 +110,12 @@ def pooling_compute(
     padding = [int(v) for v in padding]
     stride_h, stride_w = [int(v) for v in strides]
     pool_shape_h, pool_shape_w = [int(v) for v in pool_shape]
+    ifm_channels = ofm_channels if pooling_type != "SUM" else ifm.shape[-1]
     upscale_factor = 2 if upscale != "NONE" else 1
 
     # Compute operation for the IFM DMA pipeline
     dmaed_ifm = dma_ifm_compute(
-        ifm, ifm_layout, ifm_zero_point, ifm_scale, ofm_channels, padding, upscale_factor
+        ifm, ifm_layout, ifm_zero_point, ifm_scale, ifm_channels, padding, upscale_factor
     )
 
     # Pooling compute operation
@@ -122,6 +123,8 @@ def pooling_compute(
     ofm_width = (dmaed_ifm.shape[2] - pool_shape_w) // stride_w + 1
     rh = te.reduce_axis((0, pool_shape_h), name="ry")
     rw = te.reduce_axis((0, pool_shape_w), name="rx")
+    rc = te.reduce_axis((0, 1 if pooling_type != "SUM" else ifm_channels), name="rc")
+    ofm_dtype = ifm.dtype if pooling_type != "SUM" else "int32"
 
     pooling_attrs = {
         "op": "ethosu_pooling",
@@ -149,10 +152,10 @@ def pooling_compute(
     pooling = te.compute(
         (1, ofm_height, ofm_width, ofm_channels),
         lambda nn, hh, ww, cc: te.max(
-            (dmaed_ifm(nn, hh * stride_h + rh, ww * stride_w + rw, cc) + lut_expr).astype(
-                ifm.dtype
+            (dmaed_ifm(nn, hh * stride_h + rh, ww * stride_w + rw, cc + rc) + lut_expr).astype(
+                ofm_dtype
             ),
-            axis=[rh, rw],
+            axis=[rh, rw, rc],
         ),
         name="ethosu_pooling",
         attrs=pooling_attrs,
--- a/python/tvm/relay/op/contrib/ethosu.py
+++ b/python/tvm/relay/op/contrib/ethosu.py
@@ -1375,6 +1375,82 @@ def mean_pattern() -> tvm.relay.dataflow_pattern.DFPattern:
     return pattern
 
 
+class SumParams:
+    """
+    This class will parse a call to ethosu.sum composite function
+    and extract the parameter information.
+    """
+
+    composite_name = "ethos-u.sum"
+
+    def __init__(self, func_body: Call):
+        from tvm.relay.backend.contrib.ethosu.util import RequantArgs
+
+        requantize = func_body
+        sum_op = requantize.args[0]
+        attrs = sum_op.attrs
+        cast = sum_op.args[0]
+
+        layout = "NHWC"
+        self.ifm = TensorParams(
+            cast.args[0],
+            layout,
+            requantize.args[RequantArgs.IFM_SCALE.value],
+            requantize.args[RequantArgs.IFM_ZERO_POINT.value],
+        )
+        self.ofm = TensorParams(
+            requantize,
+            layout,
+            requantize.args[RequantArgs.OFM_SCALE.value],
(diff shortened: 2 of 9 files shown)


src/relay/op/contrib/ethosu/pooling.cc
```diff
@@ -46,14 +46,27 @@ bool EthosuPoolingRel(const Array<Type>& types, int num_inputs, const Attrs& att
...
+      param->pooling_type != "SUM") {
     reporter->GetDiagCtx().EmitFatal(Diagnostic::Error(reporter->GetSpan())
                                      << "Invalid operator: expected " << operator_name
-                                     << " type 'AVG' or 'MAX' but was " << param->pooling_type);
+                                     << " type 'AVG', 'MAX', or 'SUM' but was "
+                                     << param->pooling_type);
     return false;
   }
 
-  CheckDataType(reporter, ifm->dtype, {DataType::UInt(8), DataType::Int(8)}, operator_name, "ifm",
```
Alexey-Yazev, line 56: Should types be {uint8, int8, int16} as specified in TRM?

lhutton1: Yes I believe they should be, its likely just a case of not having needed `int16` previously, but I think it would be good to add :)

src/relay/op/contrib/ethosu/pooling.cc
```diff
@@ -46,14 +46,27 @@ bool EthosuPoolingRel(const Array<Type>& types, int num_inputs, const Attrs& att
...
+  auto max_avg_pooling_ifm_dtypes = {DataType::UInt(8), DataType::Int(8)};
+  auto sum_pooling_ifm_dtypes = {DataType::UInt(8), DataType::Int(8), DataType::Int(16),
```
lhutton1, line 59: Nit: prefer types over `auto`

python/tvm/relay/backend/contrib/ethosu/te/pooling.py
```diff
@@ -110,18 +110,21 @@ def pooling_compute(
...
+        ifm, ifm_layout, ifm_zero_point, ifm_scale, ifm_channels, padding, upscale_factor
     )
 
     # Pooling compute operation
     ofm_height = (dmaed_ifm.shape[1] - pool_shape_h) // stride_h + 1
     ofm_width = (dmaed_ifm.shape[2] - pool_shape_w) // stride_w + 1
     rh = te.reduce_axis((0, pool_shape_h), name="ry")
     rw = te.reduce_axis((0, pool_shape_w), name="rx")
+    rc = te.reduce_axis((0, 1 if pooling_type != "SUM" else ifm_channels), name="rc")
+    ofm_dtype = ifm.dtype if pooling_type != "SUM" else "int32"
```
lhutton1, line 127: According to the TRM, the AVG pool output datatype can be any of {uint8, int8, int16} without regard to the input datatype - it's okay to leave like this for now, just noting as something that might come up later

lhutton1 (commented):
Thanks @Alex-grovety, LGTM! Since we've introduced a new pooling type, would it also be possible to add a test case to https://github.com/apache/tvm/blob/bd37515cfb1ce6a56f03b9ccda94ec9ae3f8cca9/tests/python/contrib/test_ethosu/test_replace_pooling.py

python/tvm/relay/op/contrib/ethosu.py
```diff
@@ -1375,6 +1375,82 @@ def mean_pattern() -> tvm.relay.dataflow_pattern.DFPattern:
...
+            return False
+        if ifm_shape_len == 3 and self.axis not in [[2]]:
+            return False
+        if ifm_shape_len == 4 and self.axis not in [[3]]:
+            return False
+
+        return True
+
+
+def sum_pattern() -> tvm.relay.dataflow_pattern.DFPattern:
```
sergey-grovety, line 1438: Could we add optional clip to the pattern?

python/tvm/relay/op/contrib/ethosu.py
```diff
@@ -1375,6 +1375,82 @@ def mean_pattern() -> tvm.relay.dataflow_pattern.DFPattern:
...
+    def is_valid(self) -> bool:
+        """
+        Checks whether Sum has compatible attributes with HW.
+        """
+
+        ifm_shape_len = len(self.ifm.shape)
+
+        if not check_valid_dtypes([self.ifm], [np.uint8, np.int8, np.int16, np.int32]):
+            return False
+        if not check_valid_dtypes([self.ofm], [np.int8]):
```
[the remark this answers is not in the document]
Alexey-Yazev: It is not possible to support int16 using SHR operation as it supports int8, uint8, int32, but for example I can use ADD operation instead of SHR. Is there an operation that is better to use for type conversion or would it be optimal to use the ADD operation?

src/relay/op/contrib/ethosu/pooling.cc
Alexey-Yazev: C++ code style says:

> Favor concrete type declaration over auto as long as it is short.

How short should it be?
Is it ok to use
`std::initializer_list<DataType>`
or better 
```
using DTypes = std::initializer_list<DataType>;
DTypes sum_pooling_ifm_dtypes
```
?

python/tvm/relay/op/contrib/ethosu.py
Alexey-Yazev: Yes

src/relay/op/contrib/ethosu/pooling.cc
lhutton1: I think it's okay to use `std::initializer_list<DataType>` in this instance

python/tvm/relay/op/contrib/ethosu.py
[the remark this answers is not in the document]
lhutton1: Thanks that makes sense, in this case it's likely better to use the ADD operation instead since we can support more output data types - I'll double check on this

src/relay/op/contrib/ethosu/pooling.cc
Alexey-Yazev: Done.

python/tvm/relay/op/contrib/ethosu.py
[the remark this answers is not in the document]
Alexey-Yazev: I use MUL operation but there are some problems for int16 output type that require further investigation, so now only int8 type support has been added.

src/relay/op/contrib/ethosu/pooling.cc
Alexey-Yazev: Done.

python/tvm/relay/backend/contrib/ethosu/te/pooling.py
```diff
@@ -110,18 +110,21 @@ def pooling_compute(
     padding = [int(v) for v in padding]
     stride_h, stride_w = [int(v) for v in strides]
     pool_shape_h, pool_shape_w = [int(v) for v in pool_shape]
+    ifm_channels = ofm_channels if pooling_type != "SUM" else ifm.shape[-1]
```
lhutton1, line 113: Won't block on this as the functionality remains the same, but wondering if `ofm_channels` ever made sense here and whether we can just use `ifm.shape[-1]` for all pooling types?

lhutton1 (approved):
Thanks @Alexey-Yazev, looks good to me!

lhutton1:
Thanks @Alexey-Yazev, @sergey-grovety!

PR #13998, opened 2023-02-15
[Keras] Fix importing conv2d_transpose for NHWC layout

echuraev:
The issue was in the shapes mismatch when `layout=NHWC` was passed to importer. In this case kernel layout in conv2d_transpose operation was set to HWIO but in params HWOI layout was used. And we got the following error:
```
`Tensor[(2, 2, 64, 128), float32]` does not match `Tensor[(2, 2, 128, 64), float32]`
```


vvchernov (approved):
LGTM

leandron (approved):
LGTM, thanks @echuraev!

PR #13999, opened 2023-02-15
[ONNX][TOPI] Add `DFT` operator

KJlaccHoeUM9l:
In this PR, converter for [ONNX DFT operator](https://github.com/onnx/onnx/blob/main/docs/Operators.md#DFT) is added.

P.S: 
This PR implements a naive Fourier transform algorithm. However, in order to improve computational efficiency, in the next PR I plan to add a Fast Fourier Transform ([FFT](https://en.wikipedia.org/wiki/Fast_Fourier_transform)) algorithm at the TOPI level.


KJlaccHoeUM9l:
Hello @AndrewZhaoLuo!
Could you take a look at this PR?
It looks like all functionality tests are green, however, the GPU documentation pipeline is red. I think I will solve this problem soon.

KJlaccHoeUM9l:
@tvm-bot rerun

python/tvm/relay/frontend/onnx.py
```diff
@@ -4877,6 +4877,122 @@ def _impl_v1(cls, inputs, attr, params):
...
+        re_input_tensor = cls._swap_axes(re_input_tensor, axis, swap_axis)
+        im_input_tensor = cls._swap_axes(im_input_tensor, axis, swap_axis)
+
+        if onesided:
+            re_input_tensor = cls._crop_onesided(re_input_tensor, axis)
+            im_input_tensor = cls._crop_onesided(im_input_tensor, axis)
+
+        output = cls._merge_real_and_imag_parts(re_input_tensor, im_input_tensor)
+
+        return output
```
vvchernov, line 4934: output variable is not necessary here

python/tvm/topi/cuda/stft.py
```diff
@@ -133,3 +133,98 @@ def gen_ir(
...
+            tid = bx * max_threads + tx
+            with ib.if_scope(tid < base_range):
+                base_idx = tid * n_fft
+                with ib.for_range(0, n_fft) as n:
+                    n_idx = base_idx + n
+                    re_output_ptr[n_idx] = tir.Cast(re_output_ptr.dtype, 0)
+                    im_output_ptr[n_idx] = tir.Cast(im_output_ptr.dtype, 0)
+                    with ib.for_range(0, n_fft) as k:
+                        k_idx = base_idx + k
+                        w = sign * -2 * pi * k * n / n_fft
```
vvchernov, line 206: looks like ratio `sign * -2 * pi / n_fft` could be calculated earlier and once

vvchernov: The same for generic implementation

vvchernov (changes requested):
In general LGTM, some small corrections only

python/tvm/relay/frontend/onnx.py
```diff
@@ -4877,6 +4877,122 @@ def _impl_v1(cls, inputs, attr, params):
...
+        # ************************* Parse inputs ***********************
+        t1 = ["float16", "float32", "float64"]
+        t2 = ["int32", "int64"]
+
+        # input
+        assert infer_type(input_tensor).checked_type.dtype in t1
+        input_shape = infer_shape(input_tensor)
+        assert len(input_shape) >= 3
+        if axis < 0:
+            axis = len(input_shape) - axis
```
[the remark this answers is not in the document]
KJlaccHoeUM9l: done

python/tvm/relay/frontend/onnx.py
```diff
@@ -4877,6 +4877,122 @@ def _impl_v1(cls, inputs, attr, params):
...
+        assert 1 <= axis <= len(input_shape) - 1
```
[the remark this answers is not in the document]
KJlaccHoeUM9l: done

python/tvm/relay/frontend/onnx.py
KJlaccHoeUM9l: done

python/tvm/topi/cuda/stft.py
KJlaccHoeUM9l: done

KJlaccHoeUM9l:
Hello @AndrewZhaoLuo!
I fixed a bug related to documentation generation. It looks like all tests are green. I also made edits in accordance with Valery's comments.
Could you take a look at this PR?

AndrewZhaoLuo:
Yes sorry, I will try to look at it tomorrow. It is an operator I am less familiar with.

masahi (commented):
> This PR implements a naive Fourier transform algorithm. However, in order to improve computational efficiency, in the next PR I plan to add a Fast Fourier Transform ([FFT](https://en.wikipedia.org/wiki/Fast_Fourier_transform)) algorithm at the TOPI level.

I wonder how exactly you are going to do it. It would be a significant effort if we need to support all input size, both real / complex transform, inverse transform etc, support both CPU and GPU. Not to mention the performance problem compared to MKL / cuFFT. At least I'd expect the implementation to be done by manual IR builder.

python/tvm/topi/stft.py
```diff
@@ -123,3 +123,85 @@ def gen_ir(
         name="stft_cpu",
         tag="stft_cpu",
     )
+
+
+def dft(
```
masahi, line 128: How about we introduce `topi/signal.py` (or better name) for signal processing-related ops?

KJlaccHoeUM9l: done

KJlaccHoeUM9l:
Hello @masahi! 
Thanks for the review!

First of all, my task was to support the DFT operator for ONNX front-end. During the execution of this task, I was based on the [ONNX documentation](https://github.com/onnx/onnx/blob/main/docs/Operators.md#DFT) and [general information](https://en.wikipedia.org/wiki/Discrete_Fourier_transform) about DFT. All functionality described in the documentation has been supported in this PR.
It is also worth noting that the ONNX documentation does not mention that the implementation uses the Fast Fourier Transform. However, if you look at the [source code](https://github.com/onnx/onnx/blob/72300aab201f81615885f13f0f4cb6f331320be0/onnx/reference/ops/op_dft.py#L19) of ONNX, you can see that "under hood" ONNX uses [numpy.fft](https://numpy.org/doc/stable/reference/generated/numpy.fft.fft.html#numpy.fft.fft), which implements one-dimensional n-point fast Fourier transform ([FFT](https://en.wikipedia.org/wiki/Fast_Fourier_transform)) with a computational complexity of `O(NlogN)` (where N is the length of the Fourier transform). The implementation refers to [Cooley-Tukey FFT algorithm](https://en.wikipedia.org/wiki/Cooley%E2%80%93Tukey_FFT_algorithm).

> I wonder how exactly you are going to do it. It would be a significant effort if we need to support all input size (e.g. not just a power of two)

In general, the Cooley-Tukey algorithm re-expresses DFT of an arbitrary composite size `N=N1*N2`. But the best computational efficiency is achieved when N is a power of two. As a first approximation, I wanted to implement two algorithms (direct and fast) and, depending on `N`, use the appropriate algorithm.

> both real / complex transform, inverse transform etc

All these transformations are supported right now and do not affect the future addition of FFT. 
One of the possible improvements could be a separate implementation of the algorithm for the real input, because in this case, we need to do only half of the calculations (the imaginary part of the input is identically equal to zero)

> support both CPU and GPU.

Right now, the default TOPI operators for the CPU and GPU are implemented. For FFT it will look similar.

> Not to mention the performance problem compared to MKL / cuFFT.

Yes, that's a good point, but it's outside the scope of this PR. This PR supports DFT operator. Performance issues can be taken out for future research.


In conclusion, I also want to add that the project in which this PR was created unexpectedly ended. Therefore, the addition of the FFT will be delayed for some time.

python/tvm/relay/op/transform.py
```diff
@@ -1987,6 +1987,33 @@ def stft(
     return _make.stft(data, n_fft, hop_length, win_length, window, normalized, onesided)
 
 
+def dft(re_data, im_data, inverse=False):
+    """
```
masahi, line 1991: We should avoid allocating the imaginary part when possible. In practice, I think we only need real to complex DFT and complex to real iDFT. In both cases, we can remove one of input or output tensors.

So rather than defining one big DFT op that does everything, how about making more fine-grained definition for each possible use case?

tests/python/frontend/onnx/test_forward.py
```diff
@@ -7933,6 +7933,77 @@ def verify_linear_regressor(a_shape, c_shape, i_shape, targets=1, batch=1):
...
+    D = 7
+
+    for axis in list(range(1, n)) + [-2]:
+        for inverse, onesided in [(0, 0), (0, 1), (1, 0)]:
+            for n_fft in [D, D - 1, D + 1]:
+                for c in [1, 2]:
+                    input_shape = [batch_size] + n * [D] + [c]
+                    output_shape = [batch_size] + n * [D] + [2]
+                    if onesided == 1:
+                        output_shape[axis] = output_shape[axis] // 2 + 1
```
masahi, line 8002: `output_shape` is not used

KJlaccHoeUM9l: This is used above in `helper.make_graph`

masahi: Then just pass `input_shape` and `output_shape` to `verify_dft`. This is not a good programming practice.

masahi:
Also, please make sure to make our definition of DFT compatible with other frameworks like PyTorch without big performance cost. Here is one for PyTorch https://pytorch.org/docs/stable/fft.html

python/tvm/relay/op/transform.py
masahi: Also if we define real to complex DFT, we can exploit the "onesided" optimization.

include/tvm/relay/attrs/transform.h
```diff
@@ -605,6 +605,17 @@ struct StftAttrs : public tvm::AttrsNode<StftAttrs> {
   }
 };  // struct StftAttrs
 
+/*! \brief Attributes used in DFT operator */
+struct DFTAttrs : public tvm::AttrsNode<DFTAttrs> {
```
masahi, line 609: Please document that this is 1D DFT along the last axis. 

We cannot use this op to support 2D DFT (PyTorch `fft2` op), correct?

KJlaccHoeUM9l: > Please document that this is 1D DFT along the last axis.

The description of the fact that calculations occur along the last axis is already given in the functions where the algorithm is implemented:
* `python/tvm/topi/signal.py`
* `python/tvm/topi/cuda/signal.py`
* `python/tvm/relay/op/transform.py`

> We cannot use this op to support 2D DFT (PyTorch fft2 op), correct?

We can use 1D DFT (which computes along the last axis) to calculate n-D DFT (and 2D DFT in particular). This can be done with the following pseudocode:
```Python
def swap_axes(a, axis1, axis2):
    permutation = list(range(len(shape(a))))
    permutation[axis1] = axis2
    permutation[axis2] = axis1
    return transpose(a, permutation)

def dftn(input_tensor, axes):
    output = input_tensor
    for axis in reversed(axes):
        output = swap_axes(output, axis, -1)
        output = DFT(output)
        output = swap_axes(output, axis, -1)
    return output
```
`axes` is a vector that contains the axes along which it is necessary to perform n-D Fourier transform (for 2D, the length of this vector must be 2).

The same approach is used in numpy: [here](https://github.com/numpy/numpy/blob/3b5eff00e49ed8ab1e6c9fa9345e51817a691b2f/numpy/fft/_pocketfft.py#L716) and [here](https://github.com/numpy/numpy/blob/3b5eff00e49ed8ab1e6c9fa9345e51817a691b2f/numpy/fft/_pocketfft.py#L701).

tests/python/frontend/onnx/test_forward.py
KJlaccHoeUM9l: done

masahi:
> Can we merge this PR in this form (with the possible addition of TODO), since the direction of my work has changed a lot?

Yes, I think serious DFT / FFT implementation would be a significant project on its own, that would likely take multiple PRs / iterations. This is a good start, and thanks for the effort!

masahi (approved):

PR #14000, opened 2023-02-15
[DOCKER] Configurable NDK version support

srkreddy1238:
Let the Android NDK version configurable as a command line argument

### DIFF
--- a/docker/Dockerfile.ci_adreno
+++ b/docker/Dockerfile.ci_adreno
@@ -22,10 +22,7 @@ COPY utils/apt-install-and-clear.sh /usr/local/bin/apt-install-and-clear
 
 # Android SDK
 COPY install/ubuntu_install_androidsdk.sh /install/ubuntu_install_androidsdk.sh
-RUN bash /install/ubuntu_install_androidsdk.sh
-ENV ANDROID_HOME=/opt/android-sdk-linux
-ENV ANDROID_NDK_HOME=/opt/android-sdk-linux/ndk/21.3.6528147
-ENV ANDROID_NDK_MAJOR=21
+RUN bash /install/ubuntu_install_androidsdk.sh 25.2.9519653 3.22.1 33.0.2 33
 ENV PATH /opt/android-sdk-linux/platform-tools:$PATH
 
 # Clang tool for CLML source codegen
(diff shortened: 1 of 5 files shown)


docker/Dockerfile.ci_adreno
```diff
@@ -22,10 +22,10 @@ COPY utils/apt-install-and-clear.sh /usr/local/bin/apt-install-and-clear
 
 # Android SDK
 COPY install/ubuntu_install_androidsdk.sh /install/ubuntu_install_androidsdk.sh
-RUN bash /install/ubuntu_install_androidsdk.sh
+RUN bash /install/ubuntu_install_androidsdk.sh 25.2.9519653 25
 ENV ANDROID_HOME=/opt/android-sdk-linux
-ENV ANDROID_NDK_HOME=/opt/android-sdk-linux/ndk/21.3.6528147
-ENV ANDROID_NDK_MAJOR=21
+ENV ANDROID_NDK_HOME=/opt/android-sdk-linux/ndk/25.2.9519653
+ENV ANDROID_NDK_MAJOR=25
```
echuraev, line 28: Is it necessary to specify these variables here? I see that they will be added to `/etc/profile` in `ubuntu_install_androidsdk.sh`.

docker/Dockerfile.ci_adreno
```diff
@@ -22,10 +22,7 @@ COPY utils/apt-install-and-clear.sh /usr/local/bin/apt-install-and-clear
 
 # Android SDK
 COPY install/ubuntu_install_androidsdk.sh /install/ubuntu_install_androidsdk.sh
-RUN bash /install/ubuntu_install_androidsdk.sh
-ENV ANDROID_HOME=/opt/android-sdk-linux
-ENV ANDROID_NDK_HOME=/opt/android-sdk-linux/ndk/21.3.6528147
-ENV ANDROID_NDK_MAJOR=21
+RUN bash /install/ubuntu_install_androidsdk.sh 25.2.9519653 25
```
echuraev, line 25: Sorry, one more comment. Probably we shouldn't pass `ANDROID_NDK_MAJOR`? We can parse this value from `ANDROID_NDK_VERSION`.

srkreddy1238: Done. Made few other tools versions too configurable. Cmake was causing some warnings over older version with new NDK.

echuraev (approved):
Sorry for the delay, I was on vacation. LGTM

Issue #14004, opened 2023-02-15
[Flaky Test] `tests/micro/common/test_tvmc.py::test_tvmc_model_run[zephyr-qemu_x86-output_dirX]`


ashutosh-arm:
These tests were found to be flaky (intermittently failing on `main` or failed in a PR with unrelated changes). See [the docs](https://github.com/apache/tvm/blob/main/docs/contribute/ci.rst#handling-flaky-failures) for details.

### Tests(s)

  - `tests/micro/common/test_tvmc.py::test_tvmc_model_run[zephyr-qemu_x86-output_dir0]`
  - `tests/micro/common/test_tvmc.py::test_tvmc_model_run[zephyr-qemu_x86-output_dir1]`

### Jenkins Links

This was observed multiple times in local runs.

ashutosh-arm:
cc @Mousius @gromero

PR #14005, opened 2023-02-15
[CI][Flaky] Skip zephyr_qemu-x86 tests that are part of task_python_microtvm

ashutosh-arm:
These tests fail sometimes with beef3 due to small size of ring buffer. RPC client ends with
SessionTerminatedError. Skipping these tests until a permanent solution is found to https://github.com/apache/tvm/issues/14004. 

cc @Mousius @gromero @leandron


ashutosh-arm:
@tvm-bot rerun

driazati (approved):

PR #14007, opened 2023-02-16
[Build] Fix find_include_path when using TVM python package

jinhongyii:
When we are using TVM python package instead of building TVM from source, we are not able to find 3rdparty directory under python's site-package directory, so we cannot build module using 3rdparty in that situation. A fix is to let the user clone a TVM repo as TVM_HOME and find include path in that clone.

cc: @tqchen


junrushao (approved):
This is an interesting issue…LGTM

PR #14008, opened 2023-02-16
[TOPI][Relay][ONNX] Replace scatter_add by scatter_elements(reduction="add")

vvchernov:
Operation scatter_add duplicates one of functionality of scatter_elements. It was replaced, but useful code was transferred to scatter_elements (particularly cuda implementation for 1d input)

cc @AndrewZhaoLuo


python/tvm/relay/op/_transform.py
```diff
@@ -687,7 +678,6 @@ def argwhere_shape_func(attrs, inputs, out_ndims):
 
 
 _reg.register_shape_func("scatter", False, elemwise_shape_func)
```
AndrewZhaoLuo, line 680: Do you plan to deprecate `relay.scatter` too?

python/tvm/topi/cuda/scatter_elements.py
```diff
@@ -20,6 +20,188 @@
 from tvm import te, tir
 from ..utils import ceil_div, get_const_int
 from ..math import cast
+from .nms import atomic_add
+
+
+def gen_scatter_add_1d_atomic(data, indices, updates, out, axis, _):
```
AndrewZhaoLuo, line 26: Ah this is something I think I might have missed. Within indices along a particular axis, do we expect all indices to be unique? Otherwise I believe something similar must be done with all operators.

python/tvm/relay/op/_transform.py
vvchernov: https://github.com/apache/tvm/pull/14019

python/tvm/topi/cuda/scatter_elements.py
vvchernov: Yes, potentially we can. The problem is that quick search did not give me any atomic operation like atomic_add, moreover it is not supported by some architectures ("vulkan", "metal"). I postponed this task due to it requires more time for investigation and implementation but does not have any priority.

vvchernov:
@tvm-bot rerun

vvchernov:
Hello @AndrewZhaoLuo! Could you see this PR again?

AndrewZhaoLuo (approved):
Thanks for the refactor @vvchernov !

PR #14010, opened 2023-02-16
[TVMC][TRANSFORMS] ToMixedPrecision transform support with custom options enabled

srkreddy1238:
Adds new command line options:
* `--mixed-precision` - Enable mixed precision conversion
* `--mixed-precision-ops` - List of operators to be converted to mixed precision
* `--mixed-precision-calculation-type` - Calculation precision type
* `--mixed-precision-acc-type` - Accumulator precision type
    
And: 
* `--desired-layout-ops` - The list of operators to be transformed with desired layout.

### DIFF
--- a/python/tvm/driver/tvmc/autotuner.py
+++ b/python/tvm/driver/tvmc/autotuner.py
@@ -39,7 +39,7 @@ from .main import register_parser
 from .model import TVMCModel
 from .target import target_from_cli, generate_target_args, reconstruct_target_args
 from .shape_parser import parse_shape_string
-from .transform import convert_graph_layout
+from .transform import generate_transform_args, parse_graph_transform_args, apply_graph_transforms
 
 
 # pylint: disable=invalid-name
@@ -127,12 +127,7 @@ def add_tune_parser(subparsers, _, json_params):
         metavar="PATH",
         help="path to an auto-tuning log file by AutoTVM.",
     )
-    parser.add_argument(
-        "--desired-layout",
-        choices=["NCHW", "NHWC"],
-        default=None,
-        help="change the data layout of the whole graph",
-    )
+    generate_transform_args(parser)
     parser.add_argument(
         "--enable-autoscheduler",
         help="enable tuning the graph through the AutoScheduler tuner",
@@ -269,6 +264,8 @@ def drive_tune(args):
         rpc_hostname = None
         rpc_port = None
 
+    transform_args = parse_graph_transform_args(args)
+
     tune_model(
         tvmc_model,
         args.target,
@@ -283,7 +280,7 @@ def drive_tune(args):
         tuner=args.tuner,
         min_repeat_ms=args.min_repeat_ms,
         early_stopping=args.early_stopping,
-        desired_layout=args.desired_layout,
+        transform_args=transform_args,
         timeout=args.timeout,
         repeat=args.repeat,
         number=args.number,
@@ -309,7 +306,7 @@ def tune_model(
     tuner: str = "xgb",
     min_repeat_ms: Optional[int] = None,
     early_stopping: Optional[int] = None,
-    desired_layout: Optional[str] = None,
+    transform_args: Optional[Dict[str, Any]] = None,
     timeout: int = 10,
     repeat: int = 1,
     number: int = 10,
@@ -354,10 +351,8 @@ def tune_model(
         Minimum time to run each trial. Defaults to 0 on x86 and 1000 on other targets.
     early_stopping : int, optional
         When specified, stop tuning after this number of trials if results aren't improving.
-    desired_layout : str, optional
-        Can be one of "NCHW" or "NHWC". When specified, compatible operations in the graph
-        will have their layout set to this format. Tasks will then be tuned using this
-        specified layout.
+    transform_args: dict, optional
+        Graph transformation arguments that are applied to the relay module.
     timeout : int, optional,
         If a kernel trial lasts longer than this duration in seconds, it will be
         considered a failure.
@@ -453,7 +448,7 @@ def tune_model(
                 mod=mod,
                 params=params,
                 target=target,
-                alter_layout=desired_layout,
+                transform_args=transform_args,
                 hardware_params=hardware_params,
                 include_simple_tasks=include_simple_tasks,
             )
@@ -475,7 +470,7 @@ def tune_model(
                 mod=mod,
                 params=params,
                 target=target,
-                alter_layout=desired_layout,
+                transform_args=transform_args,
             )
 
             # In autotvm, trials is specified per task. We can convert the per-model input
@@ -504,7 +499,7 @@ def autotvm_get_tuning_tasks(
     params: Dict[str, tvm.nd.NDArray],
     target: str,
     target_host: Optional[str] = None,
-    alter_layout: Optional[str] = None,
+    transform_args: Optional[Dict[str, Any]] = None,
 ):
     """Get the autotvm tuning tasks for a given relay module.
 
@@ -518,10 +513,8 @@ def autotvm_get_tuning_tasks(
         The compilation target.
     target_host : str, optional
         The compilation target for the host.
-    alter_layout : str, optional
-        The layout to convert the graph to. Note, the convert layout
-        pass doesn't currently guarantee the whole of the graph will
-        be converted to the chosen layout.
+    transform_args: dict, optional
+        Graph transformation arguments that are applied to the relay module.
 
     Returns
     -------
@@ -530,8 +523,7 @@ def autotvm_get_tuning_tasks(
     """
     target, target_host = Target.canon_target_and_host(target, target_host)
 
-    if alter_layout:
-        mod = convert_graph_layout(mod, alter_layout)
+    mod = apply_graph_transforms(mod, transform_args)
 
     tasks = autotvm.task.extract_from_program(
         mod["main"],
@@ -547,7 +539,7 @@ def autoscheduler_get_tuning_tasks(
     params: Dict[str, tvm.nd.NDArray],
     target: str,
     target_host: Optional[str] = None,
-    alter_layout: Optional[str] = None,
+    transform_args: Optional[Dict[str, Any]] = None,
     hardware_params: Optional[HardwareParams] = None,
     include_simple_tasks: bool = False,
 ):
@@ -563,10 +555,8 @@ def autoscheduler_get_tuning_tasks(
         The compilation target.
     target_host : str, optional
         The compilation target for the host.
-    alter_layout : str, optional
-        The layout to convert the graph to. Note, the convert layout
-        pass doesn't currently guarantee the whole of the graph will
-        be converted to the chosen layout.
+    transform_args: dict, optional
+        Graph transformation arguments that are applied to the relay module.
     hardware_params : Optional[HardwareParams]
         Hardware parameters used for the search tasks
 
@@ -579,8 +569,7 @@ def autoscheduler_get_tuning_tasks(
     """
     target, target_host = Target.canon_target_and_host(target, target_host)
 
-    if alter_layout:
-        mod = convert_graph_layout(mod, alter_layout)
+    mod = apply_graph_transforms(mod, transform_args)
 
     # Extract the tasks
     tasks, task_weights = auto_scheduler.extract_tasks(
--- a/python/tvm/driver/tvmc/autotuner.py
+++ b/python/tvm/driver/tvmc/autotuner.py
@@ -14,6 +14,7 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+# pylint: disable=unused-argument
 """
 Provides support to auto-tuning networks using AutoTVM.
 """
@@ -280,7 +281,6 @@ def drive_tune(args):
         tuner=args.tuner,
         min_repeat_ms=args.min_repeat_ms,
         early_stopping=args.early_stopping,
-        transform_args=transform_args,
         timeout=args.timeout,
         repeat=args.repeat,
         number=args.number,
@@ -289,6 +289,7 @@ def drive_tune(args):
         include_simple_tasks=args.include_simple_tasks,
         log_estimated_latency=args.log_estimated_latency,
         additional_target_options=reconstruct_target_args(args),
+        **transform_args,
     )
 
 
@@ -306,7 +307,6 @@ def tune_model(
     tuner: str = "xgb",
     min_repeat_ms: Optional[int] = None,
     early_stopping: Optional[int] = None,
-    transform_args: Optional[Dict[str, Any]] = None,
     timeout: int = 10,
     repeat: int = 1,
     number: int = 10,
@@ -315,6 +315,12 @@ def tune_model(
     include_simple_tasks: bool = False,
     log_estimated_latency: bool = False,
     additional_target_options: Optional[Dict[str, Dict[str, Any]]] = None,
+    desired_layout: Optional[str] = None,
+    desired_layout_ops: Optional[List[str]] = None,
+    mixed_precision: bool = False,
+    mixed_precision_ops: Optional[List[str]] = None,
+    mixed_precision_calculation_type: Optional[str] = None,
+    mixed_precision_acc_type: Optional[str] = None,
 ):
     """Use tuning to automatically optimize the functions in a model.
 
@@ -351,8 +357,6 @@ def tune_model(
         Minimum time to run each trial. Defaults to 0 on x86 and 1000 on other targets.
     early_stopping : int, optional
         When specified, stop tuning after this number of trials if results aren't improving.
-    transform_args: dict, optional
-        Graph transformation arguments that are applied to the relay module.
     timeout : int, optional,
         If a kernel trial lasts longer than this duration in seconds, it will be
         considered a failure.
@@ -371,12 +375,28 @@ def tune_model(
         If using the autoscheduler, write the estimated latency at each step of tuning to file.
     additional_target_options: Optional[Dict[str, Dict[str, Any]]]
         Additional target options in a dictionary to combine with initial Target arguments
+    desired_layout: str, optional
+        Can be one of "NCHW" or "NHWC". When specified, compatible operations in the graph
+        will have their layout set to this format. Tasks will then be tuned using this
+        specified layout.
+    desired_layout_ops: list[str], optional
+        The list of operators to be transformed with desired layout.
+    mixed_precision: bool
+        To enable mixed precision transformation.
+    mixed_precision_ops: list[str], optional
+        The list of operators to be converted to mixed precision.
+    mixed_precision_calculation_type: str
+        The calculation dtype to be used while mixed precision.
+    mixed_precision_acc_type: str
+        The accumulation data type to be used while mixed precision.
+
 
     Returns
     -------
     tuning_records : str
         The path to the produced tuning log file.
     """
+    transform_args = parse_graph_transform_args(locals())
     target, extra_targets = target_from_cli(target, additional_target_options)
     target, target_host = Target.canon_target_and_host(target, target_host)
     # TODO(jwfromm) Remove this deepcopy once AlterOpLayout bug that mutates source
--- a/python/tvm/driver/tvmc/compiler.py
+++ b/python/tvm/driver/tvmc/compiler.py
@@ -37,7 +37,7 @@ from .main import register_parser
 from .target import target_from_cli, generate_target_args, reconstruct_target_args
 from .pass_config import parse_configs
 from .pass_list import parse_pass_list_str
-from .transform import convert_graph_layout
+from .transform import generate_transform_args, parse_graph_transform_args, apply_graph_transforms
 from .shape_parser import parse_shape_string
 from .workspace_pools import generate_workspace_pools_args, workspace_pools_recombobulate
 
@@ -61,12 +61,7 @@ def add_compile_parser(subparsers, _, json_params):
         default="",
         help="the cross compiler options to generate target libraries, e.g. '-mfpu=neon-vfpv4'.",
     )
-    parser.add_argument(
-        "--desired-layout",
-        choices=["NCHW", "NHWC"],
-        default=None,
-        help="change the data layout of the whole graph.",
-    )
+    generate_transform_args(parser)
     parser.add_argument(
         "--dump-code",
         metavar="FORMAT",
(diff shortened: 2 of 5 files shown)


leandron (dismissed):
Thanks @srkreddy1238 this looks very interesting. Can you improve the PR description with an example command on how it looks like?

Can you also add a few tests to make sure this don't regress in future? Thanks.

python/tvm/driver/tvmc/compiler.py
```diff
@@ -281,6 +299,10 @@ def compile_model(
         compilation.
     instruments: Optional[Sequence[PassInstrument]]
         The list of pass instrument implementations.
+    pre_build_hooks: list[str], optional
+        List of pre build hooks to be applied before calling relay.build.
+    pre_build_hooks: list[str], optional
```
leandron, line 304:
```suggestion
    post_build_hooks: list[str], optional
```

Mousius:
@srkreddy1238 is there a reason you're not doing this via something like `RelayToTIR` or similar? it seems you could create a `Target` hook which would do this whenever it encounters specific parameters rather than having to manually add it to the `tvmc` command line?

python/tvm/driver/tvmc/composite_target.py
```diff
@@ -29,6 +29,7 @@
 from tvm.relay.op.contrib.bnns import partition_for_bnns
 from tvm.relay.op.contrib.vitis_ai import partition_for_vitis_ai
 from tvm.relay.op.contrib.clml import partition_for_clml
+from tvm.relay.op.contrib import adreno
```
leandron, line 32: What happens here is TVM is compiled without OpenCL support, which is the case for, example, tlcpack?

Reference: https://github.com/tlc-pack/tlcpack/blob/d3393f530331f28494b71dc68f60477e6b6dc9de/wheel/build_wheel_manylinux.sh#L88-L103

srkreddy1238: Shouldn't be an issue. ```tvm.relay.op.contrib.adreno``` is just a ```ToMixedPrecision``` invocation and independent of any framework or target.

srkreddy1238:
> @srkreddy1238 is there a reason you're not doing this via something like `RelayToTIR` or similar? it seems you could create a `Target` hook which would do this whenever it encounters specific parameters rather than having to manually add it to the `tvmc` command line?


@Mousius thanks for the review.

Initially I thought of this approach where I can create an external codegen and extend the ```REGISTERED_CODEGEN``` as shown below. https://github.com/apache/tvm/blob/e7ad4bc0717537807b24a27f8f9c75586b1ee047/python/tvm/driver/tvmc/composite_target.py#L50
This way tvmc calls the pass_pipeline and we can do what ever we want over the module.

These hooks are essentially small personalization over existing transforms. Like in this example we tried to customize the Mixed Precision pass for Adreno. Essentially we are helping developers with these simple reusable utilities. These hooks are optional features on the existing target and they may grow over time. Creating multiple targets for each such feature results in multiple similar targets definitions.

I thought hooks may be a better approach to keep them away from the existing real external codegens.

What do you think ?

Mousius:
> > @srkreddy1238 is there a reason you're not doing this via something like `RelayToTIR` or similar? it seems you could create a `Target` hook which would do this whenever it encounters specific parameters rather than having to manually add it to the `tvmc` command line?
> 
> @Mousius thanks for the review.
> 
> Initially I thought of this approach where I can create an external codegen and extend the `REGISTERED_CODEGEN` as shown below.
> 
> https://github.com/apache/tvm/blob/e7ad4bc0717537807b24a27f8f9c75586b1ee047/python/tvm/driver/tvmc/composite_target.py#L50
> 
> 
> This way tvmc calls the pass_pipeline and we can do what ever we want over the module.
> These hooks are essentially small personalization over existing transforms. Like in this example we tried to customize the Mixed Precision pass for Adreno. Essentially we are helping developers with these simple reusable utilities. These hooks are optional features on the existing target and they may grow over time. Creating multiple targets for each such feature results in multiple similar targets definitions.
> 
> I thought hooks may be a better approach to keep them away from the existing real external codegens.
> 
> What do you think ?

I think there's two different thoughts:

Rather than wrapping this in `tvmc`, it'd be good to use `Target` hooks and process the `Target` within the hook to decide any additional lowering passes, there's also pre-processing logic which you can apply - I gather this is harder as it's going through some GPU target? I'm a bit nervous about creating a whole hook registry for this as it seems like we should be pushing that into the `Target` rather than working around it in `tvmc`- from a users point of view it should be something like:
```
tvmc --target=opencl --target-opencl-mixed-precision
```
Else, do you agree it's a bit odd to have to essentially build your ML pipeline in CLI arguments?

The other thing is that you mentioned this being reusable utility, I like that approach more, and I'm curious whether we can add a `--mixed-precision` option so as to benefit all `Target`s?

srkreddy1238:
Adreno currently share ```OpenCL``` target, we probably should consider defining an exclusive ```Target``` for Adreno to support customizations.

I like the idea of  tvmc directly supporting ```--mixed-precision``` kind of options for these target independent preprocess. The target dependents definitely can go to target options. Current requirement of ```mixed-precision``` can definitely be used across all targets.

Not sure about ```tvmc``` accommodating preprocessing needs that arise from vendors in future which are not critical for Targets but developer friendly.

Coming to post build hooks, the intension here is to accommodate altering the compiled library module. One of the use case being importing additional modules before export/save. Any thoughts here ?

srkreddy1238:
@Mousius 

Another thought here, How about using packed functions instead of inventory of hooks ?

Hook is defined in user application or the respective contrib as

```
@tvm.register_func("adreno.mixed_precision_fp16")
def mixed_precision_fp16(mod, params, args):
      . . . . . .

     return mod
```

Invoked as ```--pre-build-hooks "adreno.mixed_precision_fp16"``` and tvmc looks for this packed function and invokes it.

This can be good starting point where ```tvmc``` can be out of vendor specific hooks and going forward when we see any API's used across many targets we could bring them into ```tvmc``` infra.

Mousius:
> Adreno currently share `OpenCL` target, we probably should consider defining an exclusive `Target` for Adreno to support customizations.
> 
> I like the idea of tvmc directly supporting `--mixed-precision` kind of options for these target independent preprocess. The target dependents definitely can go to target options. Current requirement of `mixed-precision` can definitely be used across all targets.
> 
> Not sure about `tvmc` accommodating preprocessing needs that arise from vendors in future which are not critical for Targets but developer friendly.

The advice I'd give here is that is that we don't always need a generic solution to a specific problem, the `--mixed-precision` parameter provides benefits to other `Target`s and doesn't require us creating a new plugin system within `tvmc` - we have similar functionality in the `--alter-layout` parameter.

> Coming to post build hooks, the intension here is to accommodate altering the compiled library module. One of the use case being importing additional modules before export/save. Any thoughts here ?

What is the use-case for this? I can't think of a use-case for generically bundling things into the eventual output unless I'm missing something, specific `Target`s would likely benefit from it, and they'd better understand the shape of their own dependencies which indicates it should be done in some `Target` hook? (Thinking of the example with CMSIS-NN where-in we may use this kind of hook to bundle the CMSIS kernels or API descriptions)

> @Mousius
> 
> Another thought here, How about using packed functions instead of inventory of hooks ?
> 
> Hook is defined in user application or the respective contrib as
> 
> ```
> @tvm.register_func("adreno.mixed_precision_fp16")
> def mixed_precision_fp16(mod, params, args):
>       . . . . . .
> 
>      return mod
> ```
> 
> Invoked as `--pre-build-hooks "adreno.mixed_precision_fp16"` and tvmc looks for this packed function and invokes it.
> 
> This can be good starting point where `tvmc` can be out of vendor specific hooks and going forward when we see any API's used across many targets we could bring them into `tvmc` infra.

Sounds a bit dangerous to me, generically calling a packed function against a module also means you have access to a plethora of other registered functions - do we want to expose them all to the audience of `tvmc`? I'd also still push for things that are more specific to a given `Target` to be attached to the `Target` itself, so the scripting API and CLI remain aligned.

srkreddy1238:
> The advice I'd give here is that is that we don't always need a generic solution to a specific problem, the `--mixed-precision` parameter provides benefits to other `Target`s and doesn't require us creating a new plugin system within `tvmc` - we have similar functionality in the `--alter-layout` parameter.

To implement the complete functionality of mixed precision we need passing the list of ops to be converted to mixed precision, precision type and other configurable options for mixed precision pass. All these are use case dependent. If we are good with enabling mixed-precision pass with it's configurable options, I am good to go this way.

> What is the use-case for this? I can't think of a use-case for generically bundling things into the eventual output unless I'm missing something, specific `Target`s would likely benefit from it, and they'd better understand the shape of their own dependencies which indicates it should be done in some `Target` hook? (Thinking of the example with CMSIS-NN where-in we may use this kind of hook to bundle the CMSIS kernels or API descriptions)

My use case is about embedding additional tuning cache and target precompiled  OpenCL binary source into the module itself. These tuning cache and clBinary blobs will be generated by running the compiled modules on real device via RPC. I don't think we now encourage this way from the core compilation flow itself. IMHO, such first time requirements should enabled outside core compiler and later absorbed if there are more similar use cases.

> Sounds a bit dangerous to me, generically calling a packed function against a module also means you have access to a plethora of other registered functions - do we want to expose them all to the audience of `tvmc`? I'd also still push for things that are more specific to a given `Target` to be attached to the `Target` itself, so the scripting API and CLI remain aligned.

Its a great tool for new comers with it's simplicity and functionality. In my opinion, ```tvmc``` is a good starting point for a newbie to quickly achieve the entire process from CLI. On the other side, packed function system is not some thing we want to hide from users. We encourage users to register their own packed functions into TVM compiler flow as well as use TVM's packed functions. 

From a newbie point of view,  option ```--pre-build-hooks "adreno.mixed_precision_fp16"``` is just a command line option and don't even know about  ```packed function``` or there exists a global registry. Without ```tvmc`` support, the user need to learn and switch into python interface for more functionality.

We are talking about only two options
- Add functionality directly into TVM compiler core (accessible via ```tvmc```)
- Use python interface for advanced feature access.
I am talking about a staging option. It's like a feature which require some feedback from developers/users before making it a mainline compiler feature (due to it's complexity to implement into mainline compiler or does it have more audience or not).

As I see, we have two perspectives here to consider. One being  ```tool integrity``` and other ```user experience```.
Staging helps ```tool integrity``` as we don't need to take every feature on day one into ```tvmc```.  Other hand,  these options definitely results good user experience as all the functionality can be achieved through CLI interface.

Mousius:
> To implement the complete functionality of mixed precision we need passing the list of ops to be converted to mixed precision, precision type and other configurable options for mixed precision pass. All these are use case dependent. If we are good with enabling mixed-precision pass with it's configurable options, I am good to go this way.

I think this makes sense, a naive first attempt is something like this?
```
--mixed-precision // Enables
--mixed-precision-ops=nn.conv2d,nn.dense
--mixed-precision-input=float16
--mixed-precision-output=float32
```

> My use case is about embedding additional tuning cache and target precompiled OpenCL binary source into the module itself. These tuning cache and clBinary blobs will be generated by running the compiled modules on real device via RPC. I don't think we now encourage this way from the core compilation flow itself. IMHO, such first time requirements should enabled outside core compiler and later absorbed if there are more similar use cases.

That definitely looks specific to OpenCL and loading precompiled OpenCL? Are all of these modules generated ahead of TVM compilation? I don't think it necessarily has to be deep in the core compiler, but it would make sense for such options to live near `Target` specific logic. It's a bit tricky to work through this without an example though, it'd be good to raise something outside this PR else I think this will take a lot longer.

> `tvmc` is a great tool for new comers with it's simplicity and functionality. In my opinion, `tvmc` is a good starting point for a newbie to quickly achieve the entire process from CLI. On the other side, packed function system is not some thing we want to hide from users. We encourage users to register their own packed functions into TVM compiler flow as well as use TVM's packed functions.

One of the reasons `tvmc` is simpler is because it doesn't expose many of the inner workings of TVM, such as the packed API, it abstracts many things to give a simpler user experience 😸 A large part of that is easily printing a help message and having arguments describe what they do, we may understand `--pre-build-hooks "adreno.mixed_precision_fp16"` but `--pre-build-hooks` would be hard to express to a user.

I don't think we're disagreeing on the way forwards for mixed precision though? Shall we try and get that resolved and look to the OpenCL blobs at the same time?

srkreddy1238:
Thanks for your time on this @Mousius 

Yes, we are good with ```--mixed-precision``` and I will enhance this PR accordingly.

OpenCL blobs feature will have major change in TVM compiler and TVMC changes is cherry on the top. We can revisit with an RFC and more details.

elvin-n:
> That definitely looks specific to OpenCL and loading precompiled OpenCL? Are all of these modules generated ahead of TVM compilation? I don't think it necessarily has to be deep in the core compiler, but it would make sense for such options to live near `Target` specific logic. It's a bit tricky to work through this without an example though, it'd be good to raise something outside this PR else I think this will take a lot longer.

Probably I missed the initial proposal and don't feel that I fully understand proposal `to live near `Target` specific logic.` But conversion to FP16 and execution in low precision is not specific neither to opencl nor to Adreno. This is generic for many platforms. We have to support such conversion for
* any opencl
* arm cpu
* hexagon
* cuda
*  vulkan
* metal
* rockm

If we add such feature dedicated only to Adreno, we will have to extend it to each above target. That would be very undesirable.

python/tvm/driver/tvmc/transform.py
```diff
@@ -13,15 +13,97 @@
...
+    def __exit__(self, ptype, value, trace):
+        for op_name in self.ops:
+            op = relay.op.get(op_name)
+            op.reset_attr(self.attr_key)
+            if self.older_attr[op_name]:
+                op.set_attr(self.attr_key, self.older_attr[op_name])
+
+
+def convert_to_mixed_precision(
+    mod, ops="nn.conv2d,nn.dense", input_type="float16", out_type="float16"
```
elvin-n, line 69: names of arguments `input_type` and `out_type` does not correspond to the meaning how you are using them.
there are three places where data types can be pointed
1. data type of calculation. It is pointed as a parameter of `ToMixedPrecision` pass. like `amp_mod = ToMixedPrecision(mixed_precision_dtype)(mod)`
2. data type for accumulation. This is exactly how you are using output_type. At the same time this data type will never be kept in this precision. There will be always cast  to dtype
3. data type for the latest op in a sequence or it's better to say the attribute pointing if we need to convert back to original data type. This is exactly pointed in `config={"relay.ToMixedPrecision.keep_orig_output_dtype": True}`

I propose
1. to rename input_type to something like calculatin_type and feed to `ToMixedPrecision` that is not implemented yet.
2. to rename out_type to something like acc_type
3. I do not propose to introduce any new argument for keeping of output in origin precision. It will confuse because really it does not say anything about output of the network, it will relate to the converted ops and will be very topology specific. I can assume that there are usage models for certain models and certain hardware where it might be required, but for 99,99% of people it will not be and will confuse by the non deterministic behaviour.

elvin-n (changes requested):

srkreddy1238:
> > That definitely looks specific to OpenCL and loading precompiled OpenCL? Are all of these modules generated ahead of TVM compilation? I don't think it necessarily has to be deep in the core compiler, but it would make sense for such options to live near `Target` specific logic. It's a bit tricky to work through this without an example though, it'd be good to raise something outside this PR else I think this will take a lot longer.
> 
> Probably I missed the initial proposal and don't feel that I fully understand proposal `to live near `Target` specific logic.` But conversion to FP16 and execution in low precision is not specific neither to opencl nor to Adreno. This is generic for many platforms. We have to support such conversion for
> 
> * any opencl
> * arm cpu
> * hexagon
> * cuda
> * vulkan
> * metal
> * rockm
> 
> If we add such feature dedicated only to Adreno, we will have to extend it to each above target. That would be very undesirable.

@elvin-n  The discussion on this PR had two contexts a while back. Pre compile hooks (Primarily to support FP16) & post build hook (to allow additional processing on compiled module - To be specific importing additional modules like OpenCL cache/binary programs ...etc.).  Finally, we decided to confine this PR to discuss about FP16 support and now the proposal is to generically support for any target (which is inline with your recommendation :)).

python/tvm/driver/tvmc/transform.py
```diff
@@ -13,15 +13,97 @@
...
 from tvm import relay, transform
 from tvm.driver.tvmc import TVMCException
 
+# ToMixedPrecision
+ACC_DTYPE = "float32"
 
-def convert_graph_layout(mod, desired_layout):
+
+def mixed_precision_rule(call_node: "relay.Call", mixed_precision_type: str):
+    global ACC_DTYPE
```
Mousius, line 29: Why does this need to be marked as global? Doesn't it inherit from the outer scope anyway?

python/tvm/driver/tvmc/transform.py
```diff
@@ -13,15 +13,97 @@
...
+        Output or accumulation precision to be used.
+
+    Returns
+    -------
+    mod : tvm.IRModule
+        The converted module.
+    """
+
+    global ACC_DTYPE
+    ACC_DTYPE = out_type
```
Mousius, line 91: Can we pass this into `MixedPrecision` and generate a rule from it instead of using a global? Something like:

```python
def generate_rule(acc_dtype):
    def _mixed_precision_rule(call_node: "relay.Call", mixed_precision_type: str):
        return [
            relay.transform.mixed_precision.MIXED_PRECISION_ALWAYS,
            ACC_DTYPE,
            mixed_precision_type,
        ];
    return _mixed_precision_rule
```

tests/python/driver/tvmc/test_transform.py
```diff
@@ -70,5 +70,55 @@ def test_layout_transform_convert_layout_pass_args(relay_conv2d, monkeypatch):
     )
 
 
+def test_layout_transform_to_mixed_precision_pass_args(relay_conv2d, monkeypatch):
+    """
+    Check the mixed precision arugments which are expected when
+    mixed precision arguments are provided.
+    """
+    mock_mixed_precision = MagicMock()
+    mock_mixed_precision.return_value = tvm.driver.tvmc.transform.MixedPrecision([])
```
Mousius, line 79: This seems to do a good job of checking the graph transformation function is called, do we have any tests which call into `tvmc.transform.MixedPrecision` to test it behaves as we expect?

python/tvm/driver/tvmc/transform.py
```diff
@@ -13,15 +13,97 @@
...
+
+    with MixedPrecision(ops.split(",")):
```
Mousius, line 93: Can we do this split somewhere in `drive_tune` / `drive_compile` ? 

That way the scripting interface uses the more natural `mixed_precision_ops=["nn.conv2d", "nn.woof"]` rather than `mixed_precision_ops="nn.conv2d,nn.woof"`

python/tvm/driver/tvmc/compiler.py
```diff
@@ -260,10 +256,8 @@ def compile_model(
     target_host : str, optional
         The target of the host machine if host-side code
         needs to be generated.
-    desired_layout: str, optional
```
Mousius, line 263: Similarly here, keep `desired_layout`, use array for `mixed_precision.ops` 😸

Mousius (changes requested):
This looks really cool @srkreddy1238 😸 hopefully only a few things to resolve and we can this feature in!

python/tvm/driver/tvmc/transform.py
srkreddy1238: Reworked without globals.

tests/python/driver/tvmc/test_transform.py
srkreddy1238: May be a bit tricky. We need to craft a graph and inspect the result ops. Let me try what we can do here.

srkreddy1238: Managed with a testcase that pass through ToMixedPrecision and checks the dtypes across ops.

python/tvm/driver/tvmc/transform.py
```diff
@@ -21,7 +22,88 @@
...
+def convert_to_mixed_precision(mod, ops=None, calculation_type="float16", acc_type="float16"):
+    """Converts the operator datatypes
+
+    Parameters
+    ----------
+    mod : tvm.IRModule
+        The relay module to convert.
+    ops : list
+        List of operators to be precision converted.
+    calculation_type: str
```
elvin-n, line 79: this parameter is not used. need to remove from parameters of `convert_to_mixed_precision` or pass it to `ToMixedPrecision`

python/tvm/driver/tvmc/autotuner.py
```diff
@@ -376,12 +375,28 @@ def tune_model(
         If using the autoscheduler, write the estimated latency at each step of tuning to file.
     additional_target_options: Optional[Dict[str, Dict[str, Any]]]
         Additional target options in a dictionary to combine with initial Target arguments
+    desired_layout: str, optional
+        Can be one of "NCHW" or "NHWC". When specified, compatible operations in the graph
+        will have their layout set to this format. Tasks will then be tuned using this
+        specified layout.
+    desired_layout_ops: list[str], optional
```
elvin-n, line 382: could you please point where it is used, was not able to find its usage in PR

elvin-n (changes requested):

python/tvm/driver/tvmc/autotuner.py
srkreddy1238: Handled.

python/tvm/driver/tvmc/compiler.py
```diff
@@ -281,6 +279,20 @@ def compile_model(
...
+        will have their layout set to this format. Tasks will then be tuned using this
+        specified layout.
+    desired_layout_ops: list[str], optional
+        The list of operators to be transformed with desired layout.
+    mixed_precision: bool
+        To enable mixed precision transformation.
+    mixed_precision_ops: list[str], optional
+        The list of operators to be converted to mixed precision.
+    mixed_precision_calculation_type: str
+        The calculation dtype to be used while mixed precision.
```
elvin-n, line 293: It make sense to describe default values for these new arguments which are defined in transform.py, `generate_transform_args()`

elvin-n (approved):

Mousius (approved):
I think this is fine, pending linting - @leandron could you take another look?

srkreddy1238:
@leandron can you take a look on this ? I have some dependencies for TVMCon next week.

echuraev (approved):

srkreddy1238:
Thanks @echuraev  & @Mousius . Happy to address any concerns later as a different PR.

PR #14013, opened 2023-02-16
[microTVM][CRT]Move Makefile to CMake to be cross-platform compatible

mehrdadh:
This PR converts CRT Makefile to a CMakelist file to be cross platform compatible.


mehrdadh:
cc @alanmacd for review.

src/runtime/crt/host/CMakeLists.txt.template
```diff
@@ -0,0 +1,51 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+cmake_minimum_required(VERSION 3.18)
+set(CMAKE_CXX_STANDARD 11)
+
+project(crt_autogenerated_project C CXX)
+add_executable(main)
+
+set(CRT_LIB_BASE crt/src/runtime/crt)
+set(CRT_LIBS microtvm_rpc_server microtvm_rpc_common aot_executor_module aot_executor graph_executor_module graph_executor common memory)
```
alanmacd, line 27: I would put one these targets (libs) per line to make it more readable

src/runtime/crt/host/microtvm_api_server.py
```diff
@@ -135,10 +133,11 @@ def generate_project(self, model_library_format_path, standalone_crt_dir, projec
                 shutil.copy2(src_path, dst_path)
 
         # Populate Makefile
```
alanmacd, line 135:
```suggestion
        # Populate CMake file
```

src/runtime/crt/host/CMakeLists.txt.template
mehrdadh: done

src/runtime/crt/host/microtvm_api_server.py
mehrdadh: done

alanmacd (approved):
LGTM

src/runtime/crt/host/microtvm_api_server.py
```diff
@@ -162,13 +161,10 @@ def generate_project(self, model_library_format_path, standalone_crt_dir, projec
...
-        if options.get("verbose"):
-            args.append("VERBOSE=1")
-
-        args.append(self.BUILD_TARGET)
-
-        subprocess.check_call(args, cwd=PROJECT_DIR)
+        build_dir = PROJECT_DIR / "build"
+        build_dir.mkdir()
+        subprocess.check_call(["cmake", ".."], cwd=build_dir)
+        subprocess.check_call(["make", "-j12"], cwd=build_dir)
```
areusch, line 167: why -j12? we didn't have this one before

src/runtime/crt/host/CMakeLists.txt.template
```diff
@@ -0,0 +1,59 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
areusch, line 1: since additional stuff is added beyond what's here, and nothing is really templatized, what do you think about calling this CMakeLists.txt.prelude or something?

mehrdadh: makes sense, that would also make it clear that this file is not supposed to be used in this directory.

mehrdadh: I'd say we keep it as `.template` to match with the rest. Also, it could be templated in future if we want to set <CRT_LIBS> instead of building all libraries.

src/runtime/crt/host/microtvm_api_server.py
mehrdadh: totally missed that, fixed now!

driazati (approved):

PR #14020, opened 2023-02-17
[TIR] Enable Host Func Attribute for PrimFunc

zxybazh:
This PR enables a new attribute `kIsHostFunc` to ensure certrain prim func is run on CPU, for example `shape_func` that computes shape information dynamically. With the new attribute, the primfunc will be skipped in verification pass and split host device pass. A unit test is added.

CC: @sunggg @YuchenJin @tqchen

### DIFF
--- a/include/tvm/tir/function.h
+++ b/include/tvm/tir/function.h
@@ -326,7 +326,7 @@ constexpr const char* kIsEntryFunc = "tir.is_entry_func";
 constexpr const char* kIsGlobalFunc = "tir.is_global_func";
 
 /*!
- * \brief Mark the function as run on the host.
+ * \brief Mark the function as run on the host, mutually exclusive with kTarget.
  *
  * Type: Integer
  */
--- a/include/tvm/tir/function.h
+++ b/include/tvm/tir/function.h
@@ -325,6 +325,13 @@ constexpr const char* kIsEntryFunc = "tir.is_entry_func";
  */
 constexpr const char* kIsGlobalFunc = "tir.is_global_func";
 
+/*!
+ * \brief Mark the function as run on the host.
+ *
+ * Type: Integer
+ */
+constexpr const char* kIsHostFunc = "tir.is_host_func";
+
 }  // namespace attr
 }  // namespace tir
 }  // namespace tvm
--- a/src/tir/transforms/primfunc_utils.cc
+++ b/src/tir/transforms/primfunc_utils.cc
@@ -30,6 +30,10 @@ namespace tir {
 namespace transform {
 transform::Pass BindTarget(Target target) {
   auto fpass = [target](tir::PrimFunc f, IRModule m, transform::PassContext ctx) {
+    if (f->GetAttr<Integer>(tvm::tir::attr::kIsHostFunc) == 1) {
+      ICHECK(target->host.defined());
+      return WithAttr(std::move(f), tvm::attr::kTarget, target->host.value());
+    }
     return WithAttr(std::move(f), tvm::attr::kTarget, target);
   };
   return tir::transform::CreatePrimFuncPass(fpass, 0, "tir.BindTarget", {});
--- a/src/tir/transforms/primfunc_utils.cc
+++ b/src/tir/transforms/primfunc_utils.cc
@@ -31,8 +31,10 @@ namespace transform {
 transform::Pass BindTarget(Target target) {
   auto fpass = [target](tir::PrimFunc f, IRModule m, transform::PassContext ctx) {
     if (f->GetAttr<Integer>(tvm::tir::attr::kIsHostFunc) == 1) {
-      ICHECK(target->host.defined());
-      return WithAttr(std::move(f), tvm::attr::kTarget, target->host.value());
+      return WithAttrs(std::move(f), Map<String, ObjectRef>{
+                       {tvm::attr::kTarget, target->host.value_or(Target("llvm"))},
+                       {tvm::tir::attr::kIsHostFunc, Integer(0)},
+                   });
     }
     return WithAttr(std::move(f), tvm::attr::kTarget, target);
   };
--- a/src/tir/transforms/primfunc_utils.cc
+++ b/src/tir/transforms/primfunc_utils.cc
@@ -31,11 +31,8 @@ namespace transform {
 transform::Pass BindTarget(Target target) {
   auto fpass = [target](tir::PrimFunc f, IRModule m, transform::PassContext ctx) {
     if (f->GetAttr<Integer>(tvm::tir::attr::kIsHostFunc) == 1) {
-      return WithAttrs(std::move(f),
-                       Map<String, ObjectRef>{
-                           {tvm::attr::kTarget, target->host.value_or(Target("llvm"))},
-                           {tvm::tir::attr::kIsHostFunc, Integer(0)},
-                       });
+      return WithAttr(std::move(WithoutAttr(std::move(f), tvm::tir::attr::kIsHostFunc)),
+                      tvm::attr::kTarget, target->host.value_or(Target("llvm")));
     }
     return WithAttr(std::move(f), tvm::attr::kTarget, target);
   };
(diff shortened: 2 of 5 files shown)


junrushao:
Hey Xiyou,

My understanding of `VerifyMemory` is that it checks the illegal memory access, for example, CPU code directly access GPU memory.

I probably don't have much context on this but was curious:
- Q1. in which case, `VerifyMemory` fails on a shape function that is run on CPU, given there is no GPU array as inputs of the PrimFunc in shape functions?
- Q2. in which case, given the target has been attached in PrimFunc as attributes, do we have to add a new attribute because target is not enough?

zxybazh:
Hi, thanks for checking my PR this late! Very good questions!

Let me share some of the context here. We are trying to support a dynamic shape operator on Cuda. This function is generated during a relax pass called `VMShapeLower` which is part of the relax build. And it will generate a primfunc as follows:
```
@T.prim_func
def shape_func(H: T.Buffer((T.int64(3),), "int64")):
    T.func_attr({"global_symbol": "shape_func"})
    H[T.int64(2)] = T.int64(4) * H[T.int64(0)] * H[T.int64(1)]
```
Apparently, it's supposed to be running on CPU, i.e, the host instead of the device. However, since this pass doesn't have access to the target information, when the function is generated it doesn't include the target in its attribute. Therefore, we would like to add an attribute to automatically bind it to the target host in `BindTarget` and avoid it being splited into device code.

For Q1, It does not fail because this pass is after `BindTarget`. Thanks for the tip! <s>Will remove this change in verify memory pass.</s> Has reverted the change.
For Q2, given the context, IMHO if we can access target information in the pass and do target binding, it's possible to avoid this new attribute. I'm not quite sure if it's expected to add target as argument for certain pass.

junrushao:
> For Q2, given the context, IMHO if we can access target information in the pass and do target binding, it's possible to avoid this new attribute. I'm not quite sure if it's expected to add target as argument for certain pass.

I'm not sure if I'm missing anything, but I do think the `VerifyMemory` pass assumes the target information is always available: https://github.com/apache/tvm/blob/d7253fbc7ebdfe9ab349eb85a7db3d68260ec059/src/tir/analysis/verify_memory.cc#L180-L185

zxybazh:
Yes, it's available in `VerifyMemory` pass and I've reverted the change in this pass. For Q2 I was refering to `VMShapeLower` pass in relax build, where this primfunc is generated.

src/tir/transforms/split_host_device.cc
```diff
@@ -346,6 +347,13 @@ class HostDeviceSplitter : public StmtMutator {
 };
 
 PrimFunc SplitHostDevice(PrimFunc&& func, IRModule* device_mod) {
+  // skip the check if the function is host function.
+  if (func->GetAttr<Integer>(tvm::tir::attr::kIsHostFunc) == 1) {
```
tqchen, line 351: handle the case when the attribute is not available and defaults to None

tests/python/unittest/test_tir_host_func.py
```diff
@@ -0,0 +1,58 @@
...
+
+    mod = tvm.lower(s, [A, A2], name="f")
+
+    assert len(mod.get_global_vars()) == 1, """Before split, expected 1 global function."""
+    mod = tvm.tir.transform.Apply(
+        lambda f: f.with_attr(
+            {
+                "global_symbol": "test",
+                "target": tvm.target.Target("cuda"),
+                "tir.is_host_func": True,
```
tqchen, line 49: Ideally we can make is_host_func and target to be mutually exclusive. So they only appear before BindTarget

tqchen:
Thanks @zxybazh . After looking at the discussions especially inputs from junru. I think it would be great to clarify that we want `is_host_func` and `target` attr  to be mutually exclusive to each other and only being used by BindTarget. 

With explicit target being attached to the function then such attr is no longer necessary and can be a source of duplication.

junrushao:
Thanks @zxybazh @tqchen for the clarification - this has been much clear to me now! Let's do the following change:

- In `BindTarget`, if we have `tvm::tir::attr::kIsHostFunc` flag set up, bind the target host instead and also remove the `kIsHostFunc` flag in the meantime.
- Remove the logic in `SplitHostDevice`

src/tir/transforms/primfunc_utils.cc
```diff
@@ -30,6 +30,12 @@ namespace tir {
 namespace transform {
 transform::Pass BindTarget(Target target) {
   auto fpass = [target](tir::PrimFunc f, IRModule m, transform::PassContext ctx) {
+    if (f->GetAttr<Integer>(tvm::tir::attr::kIsHostFunc) == 1) {
+      return WithAttrs(std::move(f), Map<String, ObjectRef>{
+                       {tvm::attr::kTarget, target->host.value_or(Target("llvm"))},
```
zxybazh, line 35: Not sure if this is the best option when target host is not available. This is my impression on the default target host.

zxybazh:
Thanks for the careful review and discussion. I've removed duplicate changes and created a unittest that checks target and host func attribute. Please take another look :)

tests/python/unittest/test_tir_host_func.py
```diff
@@ -0,0 +1,81 @@
...
+    func = tvm.te.create_prim_func(
+        te_workload.matmul(729, 729, 729, in_dtype="float32", out_dtype="float32")
+    )
+    mod = tvm.ir.IRModule({"main": func})
+    target = tvm.target.Target("cuda")
+    mod = tvm.tir.transform.Apply(
+        lambda f: f.with_attr(
+            {
+                "global_symbol": "test",
+                "target": target,
```
tqchen, line 68: should not contain target here

tqchen (approved):
One final nit

tests/python/unittest/test_tir_host_func.py
```diff
@@ -0,0 +1,81 @@
...
+    def main(
+        A: T.Buffer((729, 729), "float32"),
+        B: T.Buffer((729, 729), "float32"),
+        C: T.Buffer((729, 729), "float32"),
+    ):
+        T.func_attr(
+            {
+                "global_symbol": "test",
+                "target": T.target({"keys": ["cpu"], "kind": "llvm", "tag": ""}),
+                "tir.is_host_func": 0,
```
tqchen, line 37: should not contain host func attr, When target exists, we do not require is_host_func attr

include/tvm/tir/function.h
```diff
@@ -325,6 +325,13 @@ constexpr const char* kIsEntryFunc = "tir.is_entry_func";
  */
 constexpr const char* kIsGlobalFunc = "tir.is_global_func";
 
+/*!
+ * \brief Mark the function as run on the host, mutually exclusive with kTarget.
```
tqchen, line 329: When a function have kIsHostFunc attribute, it should not contain target attribute

tests/python/unittest/test_tir_host_func.py
```diff
@@ -0,0 +1,81 @@
...
+                "tir.noalias": True,
```
tqchen, line 38: So we should update BindParams to remove the is_host_func attribute here

tqchen (changes requested):

tests/python/unittest/test_tir_host_func.py
tqchen: related https://github.com/apache/tvm/blob/main/include/tvm/ir/attrs.h#L412

tqchen (approved):

src/tir/transforms/primfunc_utils.cc
```diff
@@ -30,6 +30,10 @@ namespace tir {
 namespace transform {
 transform::Pass BindTarget(Target target) {
   auto fpass = [target](tir::PrimFunc f, IRModule m, transform::PassContext ctx) {
+    if (f->GetAttr<Integer>(tvm::tir::attr::kIsHostFunc) == 1) {
+      return WithAttr(std::move(WithoutAttr(std::move(f), tvm::tir::attr::kIsHostFunc)),
+                      tvm::attr::kTarget, target->host.value_or(Target("llvm")));
```
junrushao, line 35: Is there a case where the target host is `None`?

zxybazh: Yes, when we use target tags like `nvidia/geforce-rtx-3070` the default target host is `None`.

junrushao: got it. that makes sense!

junrushao (approved):
LGTM. Feel free to merge it in once the CI is green!

PR #14021, opened 2023-02-17
[TIR] More flexible buffer compaction

wrongtest-intellif:
Hi there, the change want to enforce the power and flexiblity of `CompactBufferAllocation` pass in two aspects:
1. Free of pass order
    - It could work on both s-tir (with opaque blocks) and lowered tir. For example, now one could be able to invoke `LoopPartition` and then make partitioned tile aware buffer compactions.
    - We test existing cases to ensure that 
      `(LowerOpaqueBlock . CompactBufferAllocation) (mod) == (CompactBufferAllocation . LowerOpaqueBlock) (mod)` 

2. Allow "non-strict" compaction
    - Add an option `is_strict` defaults to `True`, to denote that during compaction we should respect the original buffer shape bound. Thus the compacted buffer region never exceed the original.
    - If set to `False`, the "compacted" shape is totally determined to cover buffer region accesses. Thus it may become larger than the original shape. This change the original semantic for out-of-bound accesses but may be helpful in certain usages. 
    - If loop domain changed (eg, align the loop dim or remove the extra predicate), the accessed buffer region may grow, the pass could provide a fallback implementation to adapt the buffer regions.

About implementation issues:
- To achieve [1]
    -  Buffer decl point:
        -  (s-tir) `T.alloc_buffer` in block form is handled without any change.
        -  (lowered) `T.decl_buffer` is newly handled. We assume it is at the proper position to dominate all accesses.
    -  Predicates
        - Predicates in `T.where` and `IfThenElse`, `T.if_then_else` are handled uniformlly now. We would try simply resolve the predicate to update loop domain as before. But on failures, now we keep them into pending stack.
        - The visit logic is very alike `IRVisitorWithAnalyzer`'s, but since  `arith::Analyzer` and `arith::IntSet` are independent components now, the change do not introduce `IRVisitorWithAnalyzer`'.
    - Buffer accesses
        - There is no difference between s-tir and lowered form. If there are pending predicates on access point, always try affine iter analysis to get more tight relaxed region.
    - Thread binding
        - To work on lowered form, `attr::thread_extent` and `attr::virtual_thread` are handled to record neccesary info for thread relaxion.
    - Buffer aliasing 
        - No change to `T.match_buffer` handling, and we explicitly disable compaction to alised buffers in lowered form.
    - Dim alignment
        - We utilize annotation field in `T.allocate` and preserve `attr::buffer_dim_align` in `LowerOpaqueBlock` pass when it convert `T.alloc_buffer` to `T.allocate`. Thus the compaction could collect the dim alignment information in lowered form.

- To achieve [2]
    - It is much direct. `SimplifyAndNarrowBufferRegionFromNDIntSet` would not intersect accessed region with original shape if the `is_strict` option is overrided to false.

### DIFF
--- a/include/tvm/tir/transform.h
+++ b/include/tvm/tir/transform.h
@@ -456,10 +456,11 @@ TVM_DLL Pass ConvertBlocksToOpaque();
  *
  *  \endcode
  *
- *
+ * \param is_strict ensure the compacted shape always smaller than the original shape.
+ *   otherwise it allows to grow the shape to match actual accessed buffer regions.
  * \return The pass.
  */
-TVM_DLL Pass CompactBufferAllocation();
+TVM_DLL Pass CompactBufferAllocation(bool is_strict = true);
 
 /*!
  * This pass legalizes packed calls by wrapping their arguments into TVMValues
--- a/python/tvm/tir/transform/transform.py
+++ b/python/tvm/tir/transform/transform.py
@@ -747,7 +747,7 @@ def ConvertBlocksToOpaque():
     return _ffi_api.ConvertBlocksToOpaque()  # type: ignore
 
 
-def CompactBufferAllocation():
+def CompactBufferAllocation(is_strict=True):
     """Compact the buffer access region. by removing the buffer regions
     that are not accessed, i.e. narrowing the buffer shape and adjust
     the access region if necessary.
@@ -783,13 +783,19 @@ def CompactBufferAllocation():
                 for j in range(0, 16):
                     C[i, j] = B[0, j] + 1
 
+    Parameters
+    ----------
+    is_strict : bool
+        Ensure the compacted shape to be always smaller than the original shape.
+        Otherwise it allows to grow the shape to match actual accessed buffer regions.
+             
     Returns
     -------
     fpass : tvm.transform.Pass
         The result pass
 
     """
-    return _ffi_api.CompactBufferAllocation()  # type: ignore
+    return _ffi_api.CompactBufferAllocation(is_strict)  # type: ignore
 
 
 def LowerMatchBuffer():
--- a/python/tvm/tir/transform/transform.py
+++ b/python/tvm/tir/transform/transform.py
@@ -788,7 +788,7 @@ def CompactBufferAllocation(is_strict=True):
     is_strict : bool
         Ensure the compacted shape to be always smaller than the original shape.
         Otherwise it allows to grow the shape to match actual accessed buffer regions.
-             
+
     Returns
     -------
     fpass : tvm.transform.Pass
--- a/python/tvm/tir/transform/transform.py
+++ b/python/tvm/tir/transform/transform.py
@@ -747,7 +747,7 @@ def ConvertBlocksToOpaque():
     return _ffi_api.ConvertBlocksToOpaque()  # type: ignore
 
 
-def CompactBufferAllocation(is_strict=True):
+def CompactBufferAllocation(is_strict: bool = True):
     """Compact the buffer access region. by removing the buffer regions
     that are not accessed, i.e. narrowing the buffer shape and adjust
     the access region if necessary.
--- a/src/tir/analysis/block_access_region_detector.cc
+++ b/src/tir/analysis/block_access_region_detector.cc
@@ -60,6 +60,8 @@ class BlockReadWriteDetector : public StmtExprVisitor {
   std::unordered_map<const VarNode*, arith::IntSet> dom_map_;
   /*! \brief Extra iteration range hint for free vars */
   std::unordered_map<const VarNode*, arith::IntSet> hint_map_;
+  /*! \brief Unresolved conditions within current scope. */
+  std::vector<PrimExpr> pending_conditions_;
   /*! \brief The buffers that the current block reads */
   std::vector<Buffer> read_buffers_;
   /*! \brief The buffers that the current block writes */
@@ -164,12 +166,12 @@ void BlockReadWriteDetector::VisitStmt_(const IfThenElseNode* op) {
   VisitExpr(op->condition);
   {
     // Visit then branch
-    With<ConditionalBoundsContext> ctx(op->condition, &dom_map_, &hint_map_, true);
+    With<ConditionalBoundsContext> ctx(op->condition, &dom_map_, &hint_map_, &pending_conditions_);
     StmtExprVisitor::VisitStmt(op->then_case);
   }
   if (op->else_case) {
     // Visit else branch
-    With<ConditionalBoundsContext> ctx(op->condition, &dom_map_, &hint_map_, false);
+    With<ConditionalBoundsContext> ctx(!op->condition, &dom_map_, &hint_map_, &pending_conditions_);
     StmtExprVisitor::VisitStmt(op->else_case.value());
   }
 }
@@ -207,12 +209,12 @@ void BlockReadWriteDetector::VisitExpr_(const CallNode* op) {
     VisitExpr(op->args[0]);
     {
       // Visit then branch
-      With<ConditionalBoundsContext> ctx(op->args[0], &dom_map_, &hint_map_, true);
+      With<ConditionalBoundsContext> ctx(op->args[0], &dom_map_, &hint_map_, &pending_conditions_);
       StmtExprVisitor::VisitExpr(op->args[1]);
     }
     {
       // Visit else branch
-      With<ConditionalBoundsContext> ctx(op->args[0], &dom_map_, &hint_map_, false);
+      With<ConditionalBoundsContext> ctx(!op->args[0], &dom_map_, &hint_map_, &pending_conditions_);
       StmtExprVisitor::VisitExpr(op->args[2]);
     }
     return;
--- a/src/tir/transforms/compact_buffer_region.cc
+++ b/src/tir/transforms/compact_buffer_region.cc
@@ -82,7 +82,7 @@ class BufferAccessRegionCollector : public StmtExprVisitor {
         : buffer(buffer), accessed_region(region) {}
   };
 
-  explicit BufferAccessRegionCollector(bool collect_inbound) : collect_inbound_(collect_inbound){};
+  explicit BufferAccessRegionCollector(bool collect_inbound) : collect_inbound_(collect_inbound) {}
 
   /**************** Visitor overload ****************/
 
--- a/src/tir/transforms/compact_buffer_region.cc
+++ b/src/tir/transforms/compact_buffer_region.cc
@@ -271,13 +271,18 @@ class BufferAccessRegionCollector : public StmtExprVisitor {
 
   void VisitStmt_(const AllocateNode* op) final {
     auto it = var2buffer_.find(op->buffer_var);
+
+    // Do not make compaction when the buffer def and
+    // the allocation is not one-to-one with the same dtype.
     if (it == var2buffer_.end() || it->second.size() > 1) {
-      // Skip alised buffer
-      StmtExprVisitor::VisitStmt_(op);
-      return;
+      return StmtExprVisitor::VisitStmt_(op);
     }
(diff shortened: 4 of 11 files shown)


wrongtest-intellif:
cc @Hzfengsy @spectrometerHBH @vinx13 @Lunderberg

python/tvm/tir/transform/transform.py
```diff
@@ -780,7 +780,7 @@ def ConvertBlocksToOpaque():
     return _ffi_api.ConvertBlocksToOpaque()  # type: ignore
 
 
-def CompactBufferAllocation():
+def CompactBufferAllocation(is_strict=True):
```
Hzfengsy, line 783:
```suggestion
def CompactBufferAllocation(is_strict: bool = True):
```

src/tir/transforms/ir_utils.cc
```diff
@@ -455,5 +489,52 @@ std::pair<PrimExpr, PrimExpr> GetAsyncWaitAttributes(const AttrStmtNode* op) {
...
+      for (const auto& storage_align_tuple : storage_align_annotation) {
+        int buffer_index = storage_align_tuple[0]->value;
+        const Buffer& buffer = op->writes[buffer_index]->buffer;
+        storage_align_[buffer->data].push_back(storage_align_tuple);
+      }
+    }
+    StmtVisitor::VisitStmt_(op);
+  }
+
+  /*! \brief For lowerred tir, the alignment annotations reside in allocate annotations. */
```
Hzfengsy, line 512:
```suggestion
  /*! \brief For lowered tir, the alignment annotations reside in allocate annotations. */
```

Hzfengsy:
I have no objections. But it would be great if we can have double-checked with other TIR maintainers, as this PR changes the fundamental flow.

src/tir/analysis/block_access_region_detector.cc
```diff
@@ -170,12 +172,12 @@ void BlockReadWriteDetector::VisitStmt_(const IfThenElseNode* op) {
   VisitExpr(op->condition);
   {
     // Visit then branch
-    With<ConditionalBoundsContext> ctx(op->condition, &dom_map_, &hint_map_, true);
+    With<ConditionalBoundsContext> ctx(op->condition, &dom_map_, &hint_map_, &pending_conditions_);
```
Lunderberg, line 175: Should these `With<ConditionalBoundsContext>` also be applied to a `BlockRealizeNode::predicate`?

src/tir/transforms/compact_buffer_region.cc
```diff
@@ -247,37 +218,61 @@ class BufferAccessRegionCollector : public StmtExprVisitor {
...
   }
 
   void VisitStmt_(const BlockRealizeNode* op) final {
-    PrimExpr cur_predicate = predicate_in_scope;
-    predicate_in_scope = op->predicate;
+    With<ConditionalBoundsContext> ctx(op->predicate, &dom_map_, &hint_map_, &pending_conditions_);
+    StmtExprVisitor::VisitStmt_(op);
+  }
+
+  void VisitStmt_(const DeclBufferNode* op) final {
```
Lunderberg, line 230: If there is no DeclBuffer present for a buffer, would the resizing work correctly if it pre-visits the stmt to collect all buffers used within the body?  That may be a way to avoid depending on DeclBuffer, which isn't used at all points in the lowering flow.

src/tir/transforms/compact_buffer_region.cc
```diff
@@ -481,47 +520,67 @@ class BufferCompactor : public StmtExprMutator {
...
-          stride = stride * shape[dim];
+  Stmt VisitStmt_(const DeclBufferNode* op) final {
+    Buffer new_buffer = RewriteAllocBuffer(op->buffer);
+    auto n = CopyOnWrite(op);
+    n->buffer = std::move(new_buffer);
+    n->body = VisitStmt(op->body);
+    return DeclBuffer(n);
+  }
+
+  Stmt VisitStmt_(const AllocateNode* op) final {
```
Lunderberg, line 531: Should we also support `AllocateConstNode`?

Lunderberg (commented):
I like the changes, especially the flexibility in order of passes.  I don't have any worries on the lowering flow, especially since this PR doesn't change the order itself, and only prepares a way for doing so in the future.  I did have a couple of questions regarding exactly how much of TIR should be supported by this pass:

1. Since this pass can now be applied to the lowered TIR, would this allow for de-duplication between the S-TIR lowering path and TE lowering path?

2. Are the `DeclBuffer` nodes sufficient to find all buffer declaration locations?  @vinx13 can correct me, but last I heard they weren't fully used across the lowering flow.

junrushao:
@wrongtest-intellif @Hzfengsy @Lunderberg any updates?

wrongtest-intellif:
> @wrongtest-intellif @Hzfengsy @Lunderberg any updates?

sorry for too late.. I am make modifications for review suggestions.

src/tir/analysis/block_access_region_detector.cc
wrongtest-intellif: yes! it is applied when we visit block realize

src/tir/transforms/compact_buffer_region.cc
wrongtest-intellif: After some thinking we decide to switch to `Allocate` as the buffer def point, thus not depend on `DeclBuffer`.

That is, if we have
```python
data = T.allocate(...)
for i in range(10):
    ...
    X = T.decl_buffer(..., data=data)
    for j in range(10):
        # access X[i, j]
```

The region is relaxed wrt scope of `i` instead of `j`, since we actually want to mutate allocation regions.

src/tir/transforms/compact_buffer_region.cc
wrongtest-intellif: I am not sure whether we are allowed to trim the const data, currently the const data is left as it is.  I am open to try it if we think it is neccesary :)

wrongtest-intellif:
> 1. Since this pass can now be applied to the lowered TIR, would this allow for de-duplication between the S-TIR lowering path and TE lowering path?

 I remove the legacy check, let's see if there are issues in testing. Generally, since the pass is after storage flatten for te codepath, I expect there would be no special IR form to cause problems.

> 2. Are the `DeclBuffer` nodes sufficient to find all buffer declaration locations?  @vinx13 can correct me, but last I heard they weren't fully used across the lowering flow.

 The change is refactored to not depend on `DeclBuffer` nodes.

src/tir/transforms/compact_buffer_region.cc
```diff
@@ -469,7 +507,8 @@ class BufferCompactor : public StmtExprMutator {
     // Step 0. Check there is no Init part.
     ICHECK(!op->init.defined());
     // Step 1. Reallocate and rewrite alloc_buffers, also update BufferAllocInfo.
-    Array<Buffer> alloc_buffers = RewriteAllocBuffer(op->alloc_buffers);
+    Array<Buffer> alloc_buffers = op->alloc_buffers.Map(
```
Lunderberg, line 548: Nit: Personal preference, but I think lambda expressions tend to be more readable than the equivalent `std::bind` expression.  Here, this could be rewritten as `op->alloc_buffers.Map([this](const Buffer& buf) { return RewriteAllocBuffer(buf); });`

src/tir/transforms/compact_buffer_region.cc
Lunderberg: Good point.  The potential issue I was picturing was an `AllocateConst`/`DeclBuffer` pair, where only the `DeclBuffer` would be rewritten.  Now that the `Allocate` is used to identify locations that should be rewritten, rather than the `DeclBuffer`, this case can no longer occur.

include/tvm/tir/transform.h
```diff
@@ -437,10 +437,11 @@ TVM_DLL Pass ConvertBlocksToOpaque();
  *
  *  \endcode
  *
- *
+ * \param is_strict ensure the compacted shape always smaller than the original shape.
+ *   otherwise it allows to grow the shape to match actual accessed buffer regions.
  * \return The pass.
  */
-TVM_DLL Pass CompactBufferAllocation();
+TVM_DLL Pass CompactBufferAllocation(bool is_strict = true);
```
Lunderberg, line 444: Should the `is_strict` argument be a parameter in the `PassContext` instead?

wrongtest-intellif: Emmmm I am preferring to use it as a parameter since we are planning to use it multiple times under the same context., like
- In phase0: standard compaction same with the mainline
- In following phases : configure it as non-strict to adjust buffer regions after we enlarge or align  loop domains to match certain hardware constraints.

wrongtest-intellif:
Sea of failures when use it with "legacy" schedule (T.T)..  We are trying to fix all of them to best improve the generality, but at the last I think we'd better to skip it by default. @Lunderberg

include/tvm/tir/transform.h
Lunderberg: That makes sense, and thank you for the clarification!

wrongtest-intellif:
Here lists lessons from te workloads:
- There may exist opaque usage to the var before we meet any buffers, so a pre-collection `var2buffer_` is added to ensure we can treat the opaque access as full region access to the corresponding buffer. If the aliased buffer count > 1, we would skip the region collection and the buffers would not get mutated.

- The itervar of thread binding from te-workloads would have empty `dom` field.

- There would be conditions with non-bool type. Thus we add some normalization from `if(x: int32)` to `if (x != 0)` for non-bool conditions.

- There would be mismatch from allocation dtype and buffer dtype, even if the buffer is the only one who refer to the allocated var.

- The inferred region extent could be data dependent (for example, dependent by the buffer value `Y = T.allocate([X[i], ...)`, in such cases, we should not compact the buffer to the dynamic extent, just like in the loop-dependent allocation cases.

Now it seems ok to the CI cases. Now we could disable the process for workloads from TE again and just leave some more covering cases for them.

wrongtest-intellif:
Hi~ Could you help to take another round of look? cc @Hzfengsy @Lunderberg

Lunderberg (commented):
I think the changes make sense.  The main question I have is whether the improved domain analysis should be merged into a more general utility.

src/tir/transforms/compact_buffer_region.cc
```diff
@@ -103,17 +59,62 @@ NDIntSet NDIntSetEval(Region region, PrimExpr predicate,
...
  * \note The param buffer regions will not be collected.
  */
 class BufferAccessRegionCollector : public StmtExprVisitor {
  public:
   static std::unordered_map<Buffer, Region, ObjectPtrHash, ObjectPtrEqual> Collect(
-      const PrimFunc& f) {
-    BufferAccessRegionCollector collector;
-    collector(f->body);
-    return std::move(collector.buffer_access_region_);
+      const PrimFunc& f, bool collect_inbound) {
```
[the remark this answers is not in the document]
wrongtest-intellif: Hi~ @Lunderberg I totally agree the `DomainTouched` utility is of the similar purpose . Here are some differences I think:
- `BufferAccessRegionCollector` collect region wrt the allocation point, that is, 
```python
for i in range(10):
    a = T.allocate([10])
    A = T.decl_buffer( [10, 10], data=a)
    for j in range(10):
         # use A[i, j]
```
would give region `A[i, 0:10]`, which is the domain "touched" under each allocation scope.

- `BufferAccessRegionCollector` take some special considerations on thread bindings.

- `BufferAccessRegionCollector` try use more arith utilities to improve intset analysis.

It would be great to share the same implementation (maybe at the cost of adding cost to where we use `DomainTouched`?). I'd like to try it in a standalone MR.

junrushao (approved):
I have no objections either given all the issues are resolved. Seems a positive change as it makes the existing logic much more clear

junrushao:
CC @Hzfengsy @Lunderberg for a second look

PR #14023, opened 2023-02-17
[TIR] [Bugfix] Pass the correct block_sref_reuse to Replace

quic-sanirudh:
A mismatch between the blocks present in the `result` vs the blocks passed in `block_sref_to_reuse` caused the bug mentioned in #13974. 

This patch tries to fix that bug by collecting only the blocks that are part of result and also present in the block replacement map `new_block_to_old_`. Since the scope block is `result`, only that block and its child blocks would be replaced, and any replaced block would be present in `rewriter.new_block_to_old_`. Thus, collecting the replaced blocks from among child blocks of `result` guarantees that the `block_sref_reuse` would contain all the replaced blocks and that they'll point to the correct block in `result` thus avoiding the missing SRef error.


src/tir/schedule/primitive/layout_transformation.cc
```diff
@@ -704,6 +704,42 @@ class TransformLayoutPlanner : private StmtExprVisitor {
   Buffer old_buffer_;
 };
 
+/**
+ * @brief Collect blocks that are part of root block to be passed to ScheduleState::Replace for SRef
```
Hzfengsy, line 708:
```suggestion
/*!
 * \brief Collect blocks that are part of root block to be passed to ScheduleState::Replace for SRef
```

Hzfengsy (commented):
Please add a regression test for it

quic-sanirudh:
> Please add a regression test for it

Yes, I realized that I did not add the test after creating the PR. I'll update with the test, thanks.

src/tir/schedule/primitive/layout_transformation.cc
quic-sanirudh: Done thanks.

Hzfengsy (approved):
LGTM. Thanks @quic-sanirudh

PR #14024, opened 2023-02-17
[FIX][RUNTIME] Convert container with function value type

MasterJH5574:
Prior to this PR, though the `convert` function is capable of converting a single Python function/lambda to TVM func, it is not able to convert a container whose values inside are functions to TVM object.

This PR adds function conversion to `convert_to_object` and redirects `convert` to `convert_to_object`, so that now the conversion is always recursive, and therefore will work well on function container value type.

Co-authored-by: Chaofan Lin <1713833595@qq.com>

### DIFF
--- a/python/tvm/runtime/object_generic.py
+++ b/python/tvm/runtime/object_generic.py
@@ -35,7 +35,7 @@ class ObjectGeneric(object):
         raise NotImplementedError()
 
 
-ObjectTypes = (ObjectBase, NDArrayBase, Module, ObjectRValueRef, PyNativeObject)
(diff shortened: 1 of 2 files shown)


MasterJH5574:
cc @SiriusNeo @tqchen

tqchen (approved):

junrushao (approved):
Interesting! It's probably something we've left over with @cyx-6's work bringing PackedFunc as an Object. Thanks for the patch!

SiriusNEO (approved):

PR #14027, opened 2023-02-17
[FIX][TOPI] Clip with IntImm/FloatImm

MasterJH5574:
Prior to this PR, TOPI clip op only accepts the min/max values with Python native float/int type, and rejects FloatImm and IntImm.

This PR enhances the clip op to allow it accept FloatImm and IntImm.

Co-authored-by: Siyuan Feng <Hzfengsy@sjtu.edu.cn>


MasterJH5574:
cc @tqchen @Hzfengsy

vinx13 (approved):

PR #14029, opened 2023-02-17
[Unity][Pass] Operator legalization

MasterJH5574:
This PR is the operator legalization pass, which transforms high-level operator calls to `call_tir`s of corresponding low-level TIR PrimFuncs.

* The legalization pass provides customizability, which enables people to pass in a customized legalization map to override the default legalization method.

* The legalization supports symbolic shape. (At this moment only pooling does not support symbolic shape, as TOPI pooling does not support. This needs to be fixed in followup PRs.)

---

The following code shows how to use this pass:
```python
# Define the pass input IRModule
@tvm.script.ir_module
class Module:
    @R.function
    def main(
        x: R.Tensor((2, 3), "float32"), y: R.Tensor((2, 3), "float32")
    ) -> R.Tensor((2, 3), "float32"):
        z: R.Tensor((2, 3), "float32") = R.add(x, y)
        r: R.Tensor((2, 3), "float32") = R.multiply(y, z)
        return r

# Define the customized legalization function for "relax.add"
def customize_legalize_add(bb: relax.BlockBuilder, call: relax.Call) -> relax.Expr:
    from tvm import topi
    return bb.call_te(topi.add, call.args[1], call.args[0])

# Apply the pass with the customized function to the module.
mod = LegalizeOps({"relax.add": customize_legalize_add})(Module)

#################################################################
# The result IRModule (note that the first binding in "main" is customized to "(y, x)"):
@tvm.script.ir_module
class Module:
    @R.function
    def main(
        x: R.Tensor((2, 3), "float32"), y: R.Tensor((2, 3), "float32")
    ) -> R.Tensor((2, 3), "float32"):
        z = R.call_tir(add, (y, x), (2, 3), dtype="float32")
        r = R.call_tir(multiply, (y, z), (2, 3), dtype="float32")
        return r

    @T.prim_func
    def add(
        A: T.Buffer[(2, 3), "float32"],
        B: T.Buffer[(2, 3), "float32"],
        T_add: T.Buffer[(2, 3), "float32"],
    ):
        T.func_attr({"tir.noalias": True})
        for ax0, ax1 in T.grid(2, 3):
            with T.block("T_add"):
                v_ax0, v_ax1 = T.axis.remap("SS", [ax0, ax1])
                T.reads(A[v_ax0, v_ax1], B[v_ax0, v_ax1])
                T.writes(T_add[v_ax0, v_ax1])
[...]

PR #14033, opened 2023-02-18
[Unity][TVMScript] Move tir/relax import in script out of __init__.py

MasterJH5574:
Prior to this PR, `python/tvm/script/__init__.py` imports both tir and relax submodules. This leads to the phenomenum that when people does
```python
```
, the relax submodule will be implicitly visited by `__init__.py` as well.

Since TIR does not rely on Relax, it is good not to import both of them at the same time. (This can prevent cyclic imports sometimes.)

This PR does this decoupling by introducing two files

* `python/tvm/script/relax.py`
* `python/tvm/script/tir.py`

and removing the imports from `python/tvm/script/__init__.py` and `python/tvm/script/parser/__init__.py`. With this change, we force people to manually do `from tvm.script import tir` and `from tvm.script import relax` to use TVMScript parser, which is right our conventional way.

---

To elabore a bit more on the cyclic import:

With #14029, submodule `relax.transform` will import topi to register the legalization functions for each operators into the op registry. Module topi imports the TVMScript parser of TIR, in `python/tvm/topi/arm_cpu.py`.

As mentioned above, the previous behavior is when importing TIR parser, the submodule `tvm.script.parser.relax` is also visited and initialized, and thereby brings the risk of cyclic import.

Since the TOPI side only needs TIR parser, we don’t need to import the Relax parser. By decoupling the import of TIR and Relax in module `tvm.script` can prevent the cyclic dependency.


MasterJH5574:
cc @tqchen @junrushao @cyx-6 @Hzfengsy

tqchen (approved):

PR #14037, opened 2023-02-18
[Unity] Initial PyTorch Frontend

MasterJH5574:
This PR introduces initial pytorch frontend components of Relax, including
- a FX translator that translates a Torch FX graph module to an TVM IRModule,
- a Relax-backend of Torch Dynamo, which brings the mechanism to build PyTorch model using Relax compilation pipeline,
- a pipeline prototype that contains the collection of pre-defined pipelines that optimizes and lower IRModule before passing to minimum build.

### DIFF
--- a/python/tvm/relax/frontend/torch/dynamo.py
+++ b/python/tvm/relax/frontend/torch/dynamo.py
@@ -1,0 +1,156 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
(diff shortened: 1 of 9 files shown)


MasterJH5574:
The Dynamo backend tests are skipped due to only the nightly build PyTorch contains dynamo, which means our CI docker image does not have dynamo.

I tested it locally before sending this PR.

tqchen (approved):

python/tvm/relax/frontend/torch/dynamo.py
```diff
@@ -0,0 +1,156 @@
...
+
+        ex = relax_build(mod, target=target)
+
+        vm = tvm.relax.vm.VirtualMachine(exec=ex.mod, device=dev)
+
+        def exec_tvm(*i_args):
+            args = [a.contiguous() for a in i_args]
+            vm_args = list()
+            for arg in args:
+                if arg.dim() != 0:
```
liquanfeng, line 105: Hi. I'm confused about this line. Actually, I'm working with a model that takes a float value as one of its inputs. So why was the value with no dim ignored? Thanks!

MasterJH5574: Hi @liquanfeng, likely this is a redundant check. You're welcome to contribute and fix it once confirmed.

PR #14038, opened 2023-02-18
[Unity][Pass] Block-level static memory planning

MasterJH5574:
This PR introduces the static memory planning pass on binding block level, as well as an analysis function that estimate the memory usage after the memory planning pass. It supports the following features: nested-tuples,  reuse memory of the input of reshape ops, an estimator that returns total memory size needed to be allocated before and after memory planning, as well as the number of tensors / memory blocks to be allocated before and after memory planning.

The estimation is static -- it does not consider control flows (such as “if” and cross-function calls). It simply accumulates the size of every alloc_tensor and alloc_storage.

We will produce “`relax.memory.alloc_tensor/storage`” as the results produced by memory planning.

### DIFF
--- a/include/tvm/relax/transform.h
+++ b/include/tvm/relax/transform.h
@@ -96,6 +96,15 @@ TVM_DLL Pass CallTIRRewrite();
 TVM_DLL Pass RewriteDataflowReshape();
 
 /*!
+ * \brief The static memory planning pass on BindingBlock level.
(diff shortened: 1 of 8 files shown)


tqchen:
the relevant tests pass, gpu ci outage, merging to save ci cycles.

tqchen (approved):

PR #14041, opened 2023-02-18
[Unity] Relax dataflow pattern language (matching)

ganler:
The dataflow pattern language for Relax (originally from https://github.com/tlc-pack/relax/pull/163).

The implementation splits patterns into two parts:

- Match an Expression: match an expression syntactically (MatchExprPattern, i.e., DFPatternMatcher);
- Match a Graph: match a graph (cross multiple VarBinding) topologically (MatchGraphPattern); 

Credit also given to relay pattern lang authors and reviewers of https://github.com/tlc-pack/relax/pull/163.

cc: @YuchenJin

### DIFF
--- a/tests/python/relax/test_dataflow_pattern.py
+++ b/tests/python/relax/test_dataflow_pattern.py
@@ -1,0 +1,867 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+import pytest
+import tvm.testing
+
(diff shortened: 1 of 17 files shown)


ganler:
When testing `pytest tests/python/relax/test_dataflow_pattern.py`, it throws a warning like "DeprecationWarning: T.var is deprecated, use T.{dtype} instead". So seems it's better to update the test-case somehow?

YuchenJin (approved):

YuchenJin:
> When testing `pytest tests/python/relax/test_dataflow_pattern.py`, it throws a warning like "DeprecationWarning: T.var is deprecated, use T.{dtype} instead". So seems it's better to update the test-case somehow?

Yes, updating the test case would be great!

tests/python/relax/test_dataflow_pattern.py
```diff
@@ -0,0 +1,867 @@
...
+    # https://developer.nvidia.com/blog/nlu-with-tensorrt-bert/
+    @tvm.script.ir_module
+    class SelfAttention:
+        @R.function
+        def main(
+            x: R.Tensor(("b", "s", "n", "h"), "float32"),
+            wq: R.Tensor(("h", "h"), "float32"),
+            wk: R.Tensor(("h", "h"), "float32"),
+            wv: R.Tensor(("h", "h"), "float32"),
+        ) -> R.Tensor:
```
ganler, line 711: This should be the function that throws the deprecation warning for `T.var`. But I can hardly tell which part is wrong tho...  It'd be great if anyone can give me a hand. :)

ganler: Oh I found a clue in https://github.com/apache/tvm/pull/14039!

tqchen:
Thanks @ganler . please fix the lint error

junrushao (approved):

PR #14046, opened 2023-02-19
[WEB] Update web runtime to support latest emcc

tqchen:
This PR provide fixes to re-enable web build for latest emcc. WASM runtime now works. WebGPU still need some catch up to the API spec. Tested locally.


junrushao (approved):
LGTM

PR #14047, opened 2023-02-19
[MetaSchedule] Fix anchor-block flow with empty design space generator

Icemist:
In the current implementation when creating an empty design space generator for anchor-block module equality, tuning records of no sense are generated (with one operation sch.enter_postproc() which is no-op) but the implementation of anchor trace applying requires meaningful instructions, so it makes sense to filter out such instructions at receiving stage (function GetTopK).


Icemist:
@tvm-bot rerun

Icemist:
cc @echuraev @masahi

Icemist:
@tvm-bot rerun

Icemist:
cc @vinx13 @junrushao

masahi (approved):

PR #14048, opened 2023-02-20
[WEB] WebGPU Codegen

tqchen:
This PR provide an implementation of WebGPU codegen. Previously we relied on SPIRV codegen for WebGPU, which is deprecated in favor of the WGSL shading language. Pass limited testing on elementwise via chrome.
Likely we will do future iterations.

### DIFF
--- a/src/target/intrin_rule.cc
+++ b/src/target/intrin_rule.cc
@@ -118,6 +118,22 @@ TVM_REGISTER_OP("tir.nearbyint")
 TVM_REGISTER_OP("tir.pow").set_attr<FLowerIntrinsic>("default.FLowerIntrinsic",
                                                      DispatchPureExtern<FloatSuffix>);
 
+PrimExpr DispatchFastErf(const PrimExpr& e) {
+  LOG(WARNING) << "fast_erf will be used instead of erf";
+  const CallNode* call = e.as<CallNode>();
+  ICHECK(call != nullptr);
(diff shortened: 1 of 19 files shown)


src/target/intrin_rule.cc
```diff
@@ -118,6 +119,22 @@ TVM_REGISTER_OP("tir.nearbyint")
...
+PrimExpr DispatchFastErf(const PrimExpr& e) {
+  LOG(WARNING) << "fast_erf will be used instead of erf";
+  const CallNode* call = e.as<CallNode>();
+  ICHECK(call != nullptr);
+  ICHECK_EQ(call->args.size(), 1);
+  PrimExpr arg = call->args[0];
+  int bits = arg.dtype().bits();
+  PrimExpr res;
+  if (arg.dtype().is_float() && (bits == 16 || bits == 32)) {
+    res = topi::fast_erf_float_expr(arg, bits);
```
junrushao, line 131: i know it's from some legacy impl of metal, but can we structure the folder slightly better to avoid depending on topi? i.e. moving this method to tir instead.

junrushao (approved):
Otherwise LGTM

src/target/intrin_rule.cc
tqchen: Thanks for the comment, reorganized to remove topi dep

PR #14050, opened 2023-02-20
[Relay][Frontend] Span Filling PyTorch

chunit-quic:
- Construct debug name of C graph instruction as the source name of span for pytorch model.
- To get the reference of renamed nodes. Add a function to export the converted C graph after conversion.
- Add structural_equal comparisons with and without set_span to the existing test cases.
- Add span test case for exporting model parameter.

### DIFF
--- a/python/tvm/relay/expr.py
+++ b/python/tvm/relay/expr.py
@@ -605,7 +605,7 @@ class TupleWrapper(object):
     def __getitem__(self, index):
         if index >= len(self):
             raise IndexError("Tuple index out of range")
-        return TupleGetItem(self.tuple_value, index)
+        return TupleGetItem(self.tuple_value, index, span=self.tuple_value.span)
 
     def __len__(self):
         return self.size
--- a/python/tvm/relay/frontend/common.py
+++ b/python/tvm/relay/frontend/common.py
@@ -1169,6 +1169,10 @@ class _SpanFiller(ExprMutator):
             return sym
         elif isinstance(sym, np.ndarray):
             return sym
+        elif not sym:
+            # some op conversion may return None
+            # e.g. op in frontend/pytorch.py: prim::device
+            return sym
 
         raise RuntimeError(f"unsupported type {type(sym)}")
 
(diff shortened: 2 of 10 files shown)


chunit-quic:
Hi community,

This is the last PR for the frontend span filling in [TVM explorer infra](https://github.com/apache/tvm/issues/13116). After this PR, the source node information of PyTorch will be tagged to the converted relay. The conversion flow of Pytorch is the most complex frontend within our PRs. To give you more context, I would try to describe some key points here.

# Renaming
We rename the debug names of C graph instructions' outputs within the model at first. Then we
can obtain them during the conversion. Rather than using the high-level IR, TopLevelTracedModule, the source IR of pytorch frontend is C graph instance. In some cases, there is no unique source name to use. Therefore we construct a unique name for each output of a C graph instruction based on their op types. Via renaming the debug name at first, we can obtain them during the conversion.

# The exported model format
We export the pure text format of C graph for users. There are two reasons for us to choose this
format.
1. Native Netron cannot obtain the modified dubeg names  
Considering the most used visulization tool, Netron. The way how it obtains the source name is
[based on the sequence of modules](https://github.com/lutzroeder/netron/blob/5d35a11d70997d3ca68235db27803282d5ea373d/source/pytorch.js#L579). It is extremely hard for us to tag the rewritten name of C graph instruction to its source. Without a patch to modify the code of Netron, we cannot check the rewritten names via Netron.

2. C Graph will be modified  
In some conversion, like qnn, C graph might be modified compared to its original format. To give
more information to users, we decide to export the modified graph for them, so that users can have more clues to check with.

Comparing with previous three frontends, Pytorch is a relative hard one. Because we have to deal the renaimg problem, exported format, and some more special cases during the conversion.

Thank you for reading. Please feel free to give us any comment. :D

#

# 
One more thing which is not related to this PR, but I would like to throw it out there. When I try to run
pytest for pytorch frontend. I found some test cases were failed even in main branch inside
CI docker. Are they expected to be failed? Or something did I do wrong (I use CPU to test only)
```
test_forward.py
    -test_forward_conv
        -group convolution
    -test_forward_conv2d_transpose_group
test_fx_quant.py
qnn_test.py
    -test_quantized_imagenet
test_object_detection.py
```

masahi (approved):

PR #14051, opened 2023-02-20
[Runtime] Add missing Type2Str for TVMByteArray

wrongtest-intellif:
To support register function return TVMByteArray.


junrushao (approved):
LGTM!

PR #14058, opened 2023-02-20
[Unity][Fix][Pass] Fix FuseOps for lack graph edges

MasterJH5574:
This PR fixes a mistake of #14044. In #14044, in VisitLeaf of graph construction of FuseOps, we first check if the input node is Leaf and then check if it is Tuple. This is not right: as Tuple is not categorized as one leaf node, when the input node is a Tuple, the function will return since the input is not a LeafNode. And the check for Tuple will thereby never holds.

It is quite interesting that our existing unit tests fail to filter this mistake out. I add a regression test for this case, which can ensure that the tuple is always visited.


junrushao:
Please rebase against the latest Unity branch

junrushao (approved):
LGTM!

PR #14059, opened 2023-02-21
[Unity] Relax op: collapse sum

SiriusNEO:
This PR brings high-level operators `relax.collapse_sum_like` and `relax.collapse_sum_to` which is useful when doing AD in Relax. To achieve this, it exposes the interface of `topi.collapse_sum`. Moreover, this PR also implements the legalization of these op and adds corresponding tests.


junrushao (approved):
LGTM!

PR #14062, opened 2023-02-21
[Unity][BYOC] Add pass to merge composite functions to offload large subgraphs

masahi:
This PR adds a pass that merges neighboring calls to composite functions offloaded to the same external backend into one function. This is important for backends that want to receive as large subgraph as possible, for example TensorRT. It plays the same role as the MergeCompilerRegion pass in Relay BYOC does, and the algorithm follows the same idea described in https://discuss.tvm.apache.org/t/relay-improved-graph-partitioning-algorithm/5830. 

Original PR
https://github.com/tlc-pack/relax/pull/372

Substantial improvement by @yelite 
https://github.com/tlc-pack/relax/pull/411

Related fix PR by @yelite 
https://github.com/tlc-pack/relax/pull/406

### DIFF
--- a/src/relax/transform/merge_composite_functions.cc
+++ b/src/relax/transform/merge_composite_functions.cc
@@ -1,0 +1,355 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/*!
+ * \file src/relax/transform/merge_composite_functions.cc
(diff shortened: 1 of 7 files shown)


junrushao (approved):

src/relax/transform/merge_composite_functions.cc
```diff
@@ -0,0 +1,355 @@
...
+  auto new_mod = MakeGroupedFunctions(mod, group_map);
+
+  CompositeInliner inliner(mod);
+  for (const auto& [gvar, func] : new_mod->functions) {
+    if (func->GetAttr<String>(attr::kCodegen)) {
+      auto new_func = inliner.Run(Downcast<Function>(func));
+      new_func = WithAttr(new_func, tvm::attr::kGlobalSymbol, gvar->name_hint);
+      new_mod->Update(gvar, new_func);
+    }
+  }
```
Ubospica, line 333: These lines seems to modify the map while iterating over it, and that makes `tests/python/relax/test_transform_merge_composite_functions.py` does not pass. Shall we write something like

```suggestion
  std::vector<std::pair<GlobalVar, BaseFunc>> to_update;
  for (const auto& [gvar, func] : new_mod->functions) {
    if (func->GetAttr<String>(attr::kCodegen)) {
      auto new_func = inliner.Run(Downcast<Function>(func));
      new_func = WithAttr(new_func, tvm::attr::kGlobalSymbol, gvar->name_hint);
      to_update.emplace_back(gvar, new_func);
    }
  }
  for (const auto& [gvar, func] : to_update) {
    new_mod->Update(gvar, func);
  }
```

@masahi

masahi: That's surprising but good find (very annoying...)

Can you PR that? I can quickly merge it.

Ubospica: Sure!

PR #14065, opened 2023-02-21
[TORCH][TOPI] Add diagonal_scatter

vvchernov:
End-to-end support of diagonal_scatter operation from pytorch on TVM side


vvchernov:
Hello @shingjan , @yelite and @AndrewZhaoLuo! Could you see this PR?

KJlaccHoeUM9l (approved):
LGTM

python/tvm/topi/cuda/diagonal_scatter.py
```diff
@@ -0,0 +1,207 @@
...
+# specific language governing permissions and limitations
+# under the License.
+# pylint: disable=invalid-name, unused-argument
+"""Diagonal scatter operator """
+import tvm
+from tvm import te, tir
+from ..utils import ceil_div, get_const_int
+
+
+def diagonal_scatter(data, src, offset=0, dim1=0, dim2=1):
```
shingjan, line 24: Per offline discussion with @AndrewZhaoLuo , it seems that having a hand-written ir builder for scatter is the way it is in TVM. I wonder if it is possible that we can simplify it by applying schedule on the generic one for cuda strategy. I also wonder what @masahi 's thoughts on this specific op.

shingjan (commented):
Overall LGTM. Can approve it once we figure out about whether to apply schedule to cuda strategy. Would appreciate extra sets of eyes on this. @masahi @AndrewZhaoLuo

python/tvm/topi/cuda/diagonal_scatter.py
vvchernov: Some note: initially I planned to implement pytorch diagonal_scatter on front-end side by scatter_elements or scatter_nd which I've refactored last weeks. But I refused this idea due to the following: 1. In both cases it requires to construct indices which is not easy task especially for lateral diagonal in asymmetric slice of multidimensional input 2. It does not work for dynamic shape at least it should be done on back-end side 3. It did not look like the solution with good performance for this op if utilize way with reusing scatter-like ops. Therefore I've decided to implement front-end and back-end

masahi (commented):
Comments are full of typo and broken English, please fix them.

Also, what's the use case of this op?

vvchernov:
Hello @masahi! In general I copy-past description from pytorch docs, but ok I will try to recheck English. It does not have any scecified use case it was done in frames of my scatter-like ops activity.
Are there any substantive questions?

masahi:
I hesitate to merge such big manual IR op unless there is a concrete use case for it. This seems like a very strange op to me, I wonder why PT added it.

The operator set is supposed to be small. We don't have to implement a new op just because PT has it.

masahi:
See https://github.com/pytorch/pytorch/pull/64430. This doesn't seem like an op that PT users use. Are you adding `slice_scatter` and `select_scatter` as well? Please give us a good reason for adding these *_scatter ops.

PR #14066, opened 2023-02-21
[Unity][Layout] Add layout transformation analysis for PrimFunc

psrivas2:
This change adds a PrimFunc level analysis to propose layout transformations to block and buffers in the PrimFunc based on the layout transformations to PrimFunc outputs. It analyzes buffer access to figure this out. It tries to preserve sequential access to buffers when it does this.

For example given the following PrimFunc and write buffer "relu" transformation `lambda n, c, h, w: (n, h, w, c // 4, c % 4)`, it will suggest to make the following transformations.

* Block transformation on "compute": `lambda n, c, h, w: (n, h, w, c // 4, c % 4)`
* Buffer transformation on  "relu": `lambda n, c, h, w: (n, h, w, c // 4, c % 4)`
* Buffer transformation on "arg":  `lambda n, c, h, w: (n, h, w, c // 4, c % 4)`
```python
@T.prim_func
  def elemwise_relu(
      arg: T.Buffer((32, 64, 224, 224), "float32"),
      relu: T.Buffer((32, 64, 224, 224), "float32"),
  ):
      for i0, i1, i2, i3 in T.grid(32, 64, 224, 224):
          with T.block("compute"):
              v_i0, v_i1, v_i2, v_i3 = T.axis.remap("SSSS", [i0, i1, i2, i3])
              T.reads(arg[v_i0, v_i1, v_i2, v_i3])
              T.writes(relu[v_i0, v_i1, v_i2, v_i3])
              relu[v_i0, v_i1, v_i2, v_i3] = T.max(arg[v_i0, v_i1, v_i2, v_i3], T.float32(0))
```
These transformations can then be applied on the PrimFunc to get the PrimFunc with new layout.

```python
@T.prim_func
def elemwise_relu(
    arg: T.Buffer((32, 224, 224, 16, 4), "float32"),
    relu: T.Buffer((32, 224, 224, 16, 4), "float32"),
):
    for ax0, ax1, ax2, ax3, ax4 in T.grid(32, 224, 224, 16, 4):
        with T.block("compute"):
            v0, v1, v2, v3, v4 = T.axis.remap("SSSSS", [ax0, ax1, ax2, ax3, ax4])
            T.reads(arg[v0, v1, v2, v3, v4])
            T.writes(relu[v0, v1, v2, v3, v4])
            relu[v0, v1, v2, v3, v4] = T.max(arg[v0, v1, v2, v3, v4], T.float32(0))
```

Note: This PR is migrated from https://github.com/tlc-pack/relax/pull/449 to Unity branch

### DIFF
--- a/src/relax/analysis/layout_transformation.cc
+++ b/src/relax/analysis/layout_transformation.cc
@@ -24,67 +24,20 @@
  */
 #include <tvm/arith/iter_affine_map.h>
 #include <tvm/relax/analysis.h>
+#include <tvm/tir/analysis.h>
 #include <tvm/tir/stmt_functor.h>
+
+#include "../../support/array.h"
 
 namespace tvm {
 namespace relax {
 
 using namespace tir;
 
-using VarSet = std::unordered_set<tir::Var, ObjectPtrHash, ObjectPtrEqual>;
-using VarToBlockIndexMap = std::unordered_map<tir::Var, int, ObjectPtrHash, ObjectPtrEqual>;
-using VarToIterTypeMap = std::unordered_map<tir::Var, IterVarType, ObjectPtrHash, ObjectPtrEqual>;
-
 /********** Helper Functions **********/
 
-/*! \brief Converts a list to an Array */
-template <typename T>
-static inline Array<T> list_to_array(std::list<T> l) {
-  Array<T> array;
-  for (auto& v : l) {
-    array.push_back(v);
-  }
-  return array;
-}
-
-/*! \brief Converts an Array to a list */
-template <typename T>
-static inline std::list<T> array_to_list(Array<T> array) {
-  std::list<T> l;
-  for (auto& v : array) {
-    l.push_back(v);
-  }
-  return l;
-}
-
-/*! \brief Checks if `expr` is dependent on any var in `defs`*/
-static inline bool IsDependentOn(const PrimExpr& expr, const VarSet& vars) {
-  bool is_dependent = false;
-  tir::PostOrderVisit(expr, [&](const ObjectRef& node) {
-    if (const tir::VarNode* op = node.as<tir::VarNode>()) {
-      if (vars.count(GetRef<tir::Var>(op)) != 0) {
-        is_dependent = true;
-      }
-    }
-  });
-  return is_dependent;
-}
-
-/*! \brief Checks the PrimExpr for any vars not defined in `defs` */
-static inline bool HasUndefinedVars(const PrimExpr& expr, const VarSet& defs) {
-  bool has_undefined_vars = false;
-  tir::PostOrderVisit(expr, [&](const ObjectRef& node) {
-    if (const tir::VarNode* op = node.as<tir::VarNode>()) {
-      if (defs.count(GetRef<tir::Var>(op)) == 0) {
-        has_undefined_vars = true;
-      }
-    }
-  });
-  return has_undefined_vars;
-}
-
-/* Checks if a transformation is bijective affine over the given ranges*/
-static inline bool IsBijectiveAffine(const IndexMap& m, const Array<Range>& ranges) {
+/*! \brief Checks if a transformation is bijective affine over the given ranges */
+static bool IsBijectiveAffine(const IndexMap& m, const Array<Range>& ranges) {
   Map<tir::Var, Range> input_iters;
   ICHECK_EQ(m->initial_indices.size(), ranges.size());
   for (size_t i = 0; i < ranges.size(); i++) {
@@ -97,9 +50,11 @@ static inline bool IsBijectiveAffine(const IndexMap& m, const Array<Range>& rang
   return !iter_map_result->indices.empty();
 }
 
-/*! \brief Analyzes the indices returned from DetectIterMap analysis. It collects the spatial
- * iterators that are used in indices. This is important to get which spatial iter vars are
- * accessed in each index of buffer access.
+/*!
+ * \brief Analyzer to collect iterators from IterSumExpr.
+ * \details Analyzes the indices from DetectIterMap analysis to collect the spatial iterators that
+ * are used in it. This is important to get which spatial iterators are accessed in each index
+ * of buffer access.
  */
 class IndexAnalyzer : public ExprVisitor {
  public:
@@ -138,19 +93,23 @@ class IndexAnalyzer : public ExprVisitor {
   Array<tir::Var> iterators_;
 };
 
-/*! \brief Analyzes IterMapResult to get the Spatial Layout of buffer access. We define Spatial
- * Layout of a buffer access as an array of length equal to the dimensions of the buffer. i-th
- * element of Spatial Layout contains spatial iter var used from the block iteration domain. For
- * indices, where no spatial iter vars are used, the spatial layout element is empty. If any of the
- * buffer access indices use multiple spatial iter vars, the spatial layout is undefined.
+/*!
+ * \brief Analyzes IterMapResult to get the Spatial Layout of buffer access.
+ * \details We define Spatial Layout of a buffer access as an array of length equal to the
+ * dimensions of the buffer. i-th element of Spatial Layout contains spatial iter var used from the
+ * block iteration domain. For indices, where no spatial iter vars are used, the spatial layout
+ * element is empty. If any of the buffer access indices use multiple spatial iter vars, the spatial
+ * layout is undefined.
+ *
  * Here are a few examples of inferred spatial layout from buffer access. si denotes i-th spatial
  * iter var, and ri denotes i-th reduction iter var.
+ *
  * SpatialLayout(A[s0*constant, s1]) = {s0, s1}
  * SpatialLayout(A[s0, constant, r0, s1]) = {s0, null, null, s1}
  * SpatialLayout(A[s0 * c + s1]) = undefined
  */
 using SpatialLayout = Array<Optional<tir::Var>>;
-static inline SpatialLayout GetSpatialLayout(const arith::IterMapResult& iter_map_result) {
+static SpatialLayout GetSpatialLayout(const arith::IterMapResult& iter_map_result) {
   ICHECK(!iter_map_result->indices.empty());
   SpatialLayout result;
   for (const arith::IterSumExpr& index : iter_map_result->indices) {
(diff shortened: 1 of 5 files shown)


psrivas2:
cc @masahi @sunggg @YuchenJin

masahi:
@psrivas2 Can you create a tracking issue for Relax layout transform? So that we can see what follow-up work can be expected (e.g. padding)?

psrivas2:
> @psrivas2 Can you create a tracking issue for Relax layout transform? So that we can see what follow-up work can be expected (e.g. padding)?

absolutely! here is the tracking issue https://github.com/apache/tvm/issues/14070 :)

src/relax/analysis/layout_transformation.cc
```diff
@@ -0,0 +1,636 @@
...
+using namespace tir;
+
+using VarSet = std::unordered_set<tir::Var, ObjectPtrHash, ObjectPtrEqual>;
+using VarToBlockIndexMap = std::unordered_map<tir::Var, int, ObjectPtrHash, ObjectPtrEqual>;
+using VarToIterTypeMap = std::unordered_map<tir::Var, IterVarType, ObjectPtrHash, ObjectPtrEqual>;
+
+/********** Helper Functions **********/
+
+/*! \brief Converts a list to an Array */
+template <typename T>
```
Hzfengsy, line 41: could you move these two helper functions to https://github.com/apache/tvm/blob/unity/src/support/array.h#L78?

src/relax/analysis/layout_transformation.cc
```diff
@@ -0,0 +1,636 @@
...
+static inline std::list<T> array_to_list(Array<T> array) {
+  std::list<T> l;
+  for (auto& v : array) {
+    l.push_back(v);
+  }
+  return l;
+}
+
+/*! \brief Checks if `expr` is dependent on any var in `defs`*/
+static inline bool IsDependentOn(const PrimExpr& expr, const VarSet& vars) {
```
Hzfengsy, line 61: Is it same as https://github.com/apache/tvm/blob/82a0d4678f66956bb7fa8038e1eedf8c75827685/include/tvm/tir/analysis.h#L126?

src/relax/analysis/layout_transformation.cc
```diff
@@ -0,0 +1,636 @@
...
+      if (vars.count(GetRef<tir::Var>(op)) != 0) {
+        is_dependent = true;
+      }
+    }
+  });
+  return is_dependent;
+}
+
+/*! \brief Checks the PrimExpr for any vars not defined in `defs` */
+static inline bool HasUndefinedVars(const PrimExpr& expr, const VarSet& defs) {
```
Hzfengsy, line 74: Please see a similar helper function at https://github.com/apache/tvm/blob/82a0d4678f66956bb7fa8038e1eedf8c75827685/include/tvm/tir/analysis.h#L102

src/relax/analysis/layout_transformation.cc
```diff
@@ -0,0 +1,636 @@
...
+    input_iters.Set(m->initial_indices[i], ranges[i]);
+  }
+  arith::Analyzer analyzer;
+  auto iter_map_result = DetectIterMap(m->final_indices, input_iters, /* predicate = */ 1,
+                                       /*check_level=*/arith::IterMapLevel::Bijective, &analyzer,
+                                       /*simplify_trivial_iterators=*/true);
+  return !iter_map_result->indices.empty();
+}
+
+/*! \brief Analyzes the indices returned from DetectIterMap analysis. It collects the spatial
```
Hzfengsy, line 100:
```suggestion
/*! 
 * \brief Analyzes the indices returned from DetectIterMap analysis. It collects the spatial
```

src/relax/analysis/layout_transformation.cc
```diff
@@ -0,0 +1,636 @@
...
+    else
+      VisitExpr(op->source);
+    VisitExpr(op->extent);
+  }
+
+ private:
+  Array<tir::Var> iterators_;
+};
+
+/*! \brief Analyzes IterMapResult to get the Spatial Layout of buffer access. We define Spatial
```
Hzfengsy, line 141: Please change to the following format for multi-line documents
```suggestion
/*! 
 * \brief Analyzes IterMapResult to get the Spatial Layout of buffer access. We define Spatial
```

src/relax/analysis/layout_transformation.cc
```diff
@@ -0,0 +1,636 @@
...
+    if (iter_vars.empty()) {
+      result.push_back({});
+      continue;
+    }
+    result.push_back(iter_vars[0]);
+  }
+  return result;
+}
+
+/*! \brief Checks if the two spatial layouts are identical. Two empty spatial layouts are treated as
```
Hzfengsy, line 173: dito

src/relax/analysis/layout_transformation.cc
```diff
@@ -0,0 +1,636 @@
...
+ * unequal.*/
+static inline bool AreIdenticalSpatialAccess(const SpatialLayout& s0, const SpatialLayout& s1) {
```
Hzfengsy, line 175: I'm not sure why these functions are marked as `static inline`, could you explain it a bit?

src/relax/analysis/layout_transformation.cc
```diff
@@ -0,0 +1,636 @@
...
+  if (s0.size() != s1.size()) return false;
+  for (size_t i = 0; i < s0.size(); ++i) {
+    if ((!s0[i].defined() && s1[i].defined()) || (s0[i].defined() && !s1[i].defined()))
+      return false;
+    if (!s0[i].same_as(s1[i])) return false;
+  }
+  return true;
+}
+
+/*! \brief Checks if the block accesses a buffer sequentially in terms of spatial dimensions
```
Hzfengsy, line 186: dito

src/relax/analysis/layout_transformation.cc
```diff
@@ -0,0 +1,636 @@
...
+      t1->initial_indices.Map([](tir::Var i) -> PrimExpr { return i; });
+  auto t0_output = t0->MapIndices(t1_initial_indices);
+  arith::Analyzer analyzer;
+  for (size_t i = 0; i < t0_output.size(); ++i) {
+    if (!analyzer.CanProveEqual(t0_output[i], t1->final_indices[i])) return false;
+  }
+  return true;
+}
+
+/*! Given the source buffer spatial layout and its transformation, this function infers the
```
Hzfengsy, line 219: dito

tests/python/relax/test_analysis_suggest_layout_transforms.py
```diff
@@ -0,0 +1,827 @@
...
+
+    suggested_transforms = relax.analysis.suggest_layout_transforms(
+        func=before,
+        write_buffer_transforms=[
+            lambda n, c, h, w: (n, h, w, c // 4, c % 4),
+            lambda n, c, h, w: (n, h, w, c // 4, c % 4),
+        ],
+    )
+    after = apply_transformations(before, suggested_transforms)
+    tvm.ir.assert_structural_equal(after, expected)
```
Hzfengsy, line 827: Please add the entry as
```python
if __name__ == "__main__":
    tvm.testing.main()
```

masahi:
cc @vinx13

masahi (commented):
This doesn't seem specific to Relax. Can we move this under `src/tir` and send the PR to `main`?

tqchen:
Thanks @masahi for the review. Given the analysis is still intended to be used in hybrid tir use-cases, we can still view it as a relax analysis tool, and keep it here for now. We did the same thing for TIR op pattern anlysis.

src/relax/analysis/layout_transformation.cc
psrivas2: `static` is used to indicate that it can't be accessed outside of the translation unit (this cc file). 
inline was used mainly to indicate intent (as the compiler does a good job inlining anyways). We can drop inline to be consistent with relax codebase though.

tests/python/relax/test_analysis_suggest_layout_transforms.py
```diff
@@ -0,0 +1,827 @@
...
+    ):
+        for ax0, ax1 in T.grid(4, 4):
+            with T.block("flatten"):
+                v_ax0, v_ax1 = T.axis.remap("SS", [ax0, ax1])
+                T.reads(arg[v_ax0, v_ax1])
+                T.writes(output[v_ax0 * 4 + v_ax1])
+                output[v_ax0 * 4 + v_ax1] = arg[v_ax0, v_ax1]
+
+    suggested_transforms = relax.analysis.suggest_layout_transforms(
+        func=before, write_buffer_transforms=[lambda a: (a // 4, a % 4)]
```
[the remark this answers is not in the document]
psrivas2: This is not supported yet. The reason being that it would be difficult to analyze the block transformation and other read buffer transformations if the write buffer does not have trivial spatial indexing i.e., single spatial iterator in each dimension.
This would also be a less common scenario as we expect most PrimFuncs through which we would want to flow layout transformations would have trivial indexing for write buffer.

psrivas2:
> This doesn't seem specific to Relax. Can we move this under `src/tir` and send the PR to `main`?

To add to what TQ mentioned above is that this analysis assumes that last `n` params of a PrimFunc are `n` outputs of the PrimFunc. This is the semantics of using `call_tir` in Relax. General TIR is more flexible than that. So the analysis is more geared towards use in a Relax IRModule.

masahi (approved):

src/relax/analysis/layout_transformation.cc
```diff
@@ -0,0 +1,621 @@
...
+    Array<tir::Var> iter_vars = index_analyzer.Analyze(index);
+    if (iter_vars.size() >= 2) {
+      LOG(WARNING) << "[LayoutInference] Unable to get spatial layout of access: "
+                   << arith::NormalizeIterMapToExpr(index);
+      return {};
+    }
+    if (iter_vars.empty()) {
+      result.push_back({});
+      continue;
+    }
```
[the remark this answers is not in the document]
psrivas2: I have a slight preference of using if-continue/return blocks instead of if-else blocks where possible. Makes the code more readable IMO. So if there is a code style recommended for TVM which prohibits this, then I would change it (grudgingly :)). Otherwise I would like to keep it as is :)

PR #14069, opened 2023-02-21
[Unity][Transform] Add LiftTransformParams pass

vinx13:
This PR added a pass `LiftTransformParams`. It allows to compile the end-to-end model without weights provided. The idea is annotate the input parameters that are weights, and identify and lift the transformations to weights, and compile it to a separate function `transform_params` that can be executed in runtime. Users can run `transform_params` with weights to get the weights for the optimized model as a prep step before the deployment. In this way, we perform the same optimizations and defer the weight transformations to the user side, while the overhead of the deferred weight transformation can be ignored as it only need to be run once.

This pass is integrated with the default `vm.build`. It is optional and only necessary when the parameters are kept as inputs when importing the model from the frontend.

cc @tqchen @jinhongyii @spectrometerHBH @MasterJH5574

### DIFF
--- a/include/tvm/relax/transform.h
+++ b/include/tvm/relax/transform.h
@@ -174,6 +174,22 @@ TVM_DLL Pass Normalize();
  */
 TVM_DLL Pass LegalizeOps(Optional<Map<String, PackedFunc>> cmap);
 
+/*
+ * \brief Lift transformation of the parameters of a function.
+ *
+ * When some inputs of the function is marked as 'parameters' (the model weights), this pass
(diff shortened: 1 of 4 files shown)


include/tvm/relax/transform.h
```diff
@@ -174,6 +174,22 @@ TVM_DLL Pass Normalize();
  */
 TVM_DLL Pass LegalizeOps(Optional<Map<String, PackedFunc>> cmap);
 
+/*
```
SiriusNEO, line 177: Find a small problem in doc: should be `/*!` @vinx13

PR #14071, opened 2023-02-21
[AoT]Add get_input_name function to AoT Module

mehrdadh:
This PR adds get_input_name function to AoT. This functionality would help with features like TVMC where we need to set the input without asking the user to specify the input name.

This PR also adds `create_local_aot_executor` for micro and refactors that across the codebase for all CRT use cases.
In addition this PR fixes a bug in python/tvm/relay/build_module.py where `relay.build` for AotExecutor was using `graph` executor instead of `aot`.


mehrdadh:
cc @alanmacd for review

masahi (approved):

PR #14072, opened 2023-02-21
[MetaSchedule] MutateTileSize skip single-candidate SampleCategorical

MasterJH5574:
When handling SampleCategorical instructions, our MutateTileSize mutator mutates it by asserting that "there should be at least one candidate other than the current decision". However, this argument fails when a SampleCategorical instruction has only one candidate, since there is no other candidate besides the unique one. On such cases, the mutator will lead to exception.

For such cases, the expected behavior is that the mutator ignores such SampleCategorical instructions. This PR updates the mutator in this way and provides a regression test.


vinx13 (approved):

tqchen (approved):
LTGM

PR #14073, opened 2023-02-21
[Codegen][CUDA] Add error message for missing fragment info

vinx13:
This PR added error message when wmma fragment info is missing. This happens when the wmma fragment memory scope is used but `tensorize` is not called. Previously, it shows error message like `tvm._ffi.base.TVMError: _Map_base: :at` that doesn't provide enough information.

cc @masahi


masahi:
@tvm-bot rerun

masahi (approved):

PR #14075, opened 2023-02-21
[Unity][Analysis] TIR pattern kind analysis for multi-buffer write block

MasterJH5574:
This PR supports TIR pattern kind analysis for TIR blocks which write to multiple buffers, which is helpful for normalization operators like layernorm, groupnorm, etc.

Prior to this PR, the analyzer does not support a blocks which write to multiple buffers. On seeing such a block, the analyzer simply sets the analysis result to "opaque". With this PR, on seeing a block which writes multiple buffers, the analyzer will check if all the BufferStores have the same indices. And it will only set the result to "opaque" when the BufferStores have different indices.

By doing this, the analysis works for common cases where a block may write to multiple buffers, like layernorm or groupnorm.

Besides the unit test for the analysis itself, this PR also adds a unit test for FuseOps pass, make sure that a "layernorm + relu" pattern can be fused together.


MasterJH5574:
**Needs review.**

Hzfengsy (commented):
Thanks @MasterJH5574 for the PR. I appreciate that you added a real-world test case `group norm`. My concern is that the testcase is a bit complex and hard to debug if it fails at any time. Could you please simply it or add a simple case for possible debugging?

MasterJH5574:
> Could you please simply it or add a simple case for possible debugging?

@Hzfengsy Sounds great. I changed the test for the analyzer to a PrimFunc which computes both the sum and the square sum of the input buffer, and changed the “groupnorm + silu” FuseOps test to “layernrom + relu” which is smaller.

Hzfengsy (approved):
LGTM

Issue #14074, opened 2023-02-21
[CI Problem] gallery/how_to/work_with_microtvm/micro_tvmc.sh fails silently


mehrdadh:
I noticed that this script is failing since `zephyr_board` project option was replaced with `board` but we haven't fixed it in this tutorial. However, failure is not caught in CI even though it is running in this script: tests/scripts/task_python_microtvm.sh
We need to do more investigation here.

cc @driazati

mehrdadh:
I can see the failure here but it doesn't fail in the CI:
https://ci.tlcpack.ai/blue/rest/organizations/jenkins/pipelines/tvm-cortexm/branches/PR-14076/runs/1/nodes/105/steps/384/log/?start=0

```
conv2d NHWC layout is not optimized for x86 with autotvm.
[2023-02-22T01:54:59.910Z] One or more operators have not been tuned. Please tune your model for better performance. Use DEBUG logging level to see more details.
[2023-02-22T01:55:00.844Z] **Error: Option 'board' is required but was not specified. Use --list-options to see all required options.**
[2023-02-22T01:55:01.102Z] + python3 gallery/how_to/work_with_microtvm/micro_mlperftiny.py
```

PR #14076, opened 2023-02-21
[microTVM] Fix tvmc tutorial

mehrdadh:
This PR applies appropriate changes to make sure the CI fails if `micro_tvmc.sh` tutorial fails. This issue was captured in https://github.com/apache/tvm/issues/14074.
This PR also makes changes to avoid this breakage in bash script tutorials in future. In addition, this PR fixes the bug in running TVMC tutorial which happened due to renaming `zephyr_board` to `board`.

### DIFF
--- a/docs/script_convert.py
+++ b/docs/script_convert.py
@@ -20,8 +20,7 @@ import pathlib
 
 BASH = "# bash"
 BASH_IGNORE = "# bash-ignore"
-BASH_MULTILINE_COMMENT_START = ": '"
-BASH_MULTILINE_COMMENT_END = "'"
+BASH_MULTILINE_COMMENT = "# bash-comment"
 
 
 def bash_to_python(src_path: pathlib.Path, dest_path: pathlib.Path):
@@ -33,6 +32,8 @@ def bash_to_python(src_path: pathlib.Path, dest_path: pathlib.Path):
             bash_detected = False
             bash_ignore_detected = False
             new_line_required = False
+            bash_multiline_comment_detected = False
+
             while line:
                 line = line.strip("\n").strip("\r")
                 if bash_detected:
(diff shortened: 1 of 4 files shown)


mehrdadh:
fixes https://github.com/apache/tvm/issues/14074

leandron (approved):
LGTM, thanks @mehrdadh

gromero (approved):

gromero (changes requested):
@mehrdadh I approved it and was about to merge but I saw in your description that there is a fix besides adding the `-e`  flag but I don't see the fix included. Am I missing something or have you missed to include the fix? :-)

mehrdadh:
@gromero sorry for confusion, you're right. I was expecting a failure and then I wanted to add the fix. But still it doesn't fail. I see the error but it doesn't fail the CI:

```
[2023-02-22T01:55:00.844Z] Error: Option 'board' is required but was not specified. Use --list-options to see all required options.
[2023-02-22T01:55:01.102Z] + python3 gallery/how_to/work_with_microtvm/micro_mlperftiny.py
[2023-02-22T01:55:01.360Z] 2023-02-22 01:55:01.282289: W tensorflow/stream_executor/platform/default/dso_loader.cc:64] Could not load dynamic library 'libcudart.so.11.0'; dlerror: libcudart.so.11.0: cannot open shared object file: No such file or directory; LD_LIBRARY_PATH: 
[2023-02-22T01:55:01.360Z] 2023-02-22 01:55:01.
```
Let's figure out why this error doesn't fail the CI and then we can merge this with the fix.

driazati (approved):
lgtm once CI passes

mehrdadh:
@gromero We finally found the CI issue. PTAL, thanks!

gromero (approved):
@mehrdadh LGTM. But I'd love it if you could improve the PR description AND merge the improved description. For instance, "This PR forces microTVM TVMC tutorial in CI in case of failure." is missing something there and also could be improved - I know what you mean now, but I bet we would need to re-read it three times in a couple of weeks to grasp what it meant.   

The title could also just be "Fix TVMC tutorial" and then explain that an additional fix in the doc/CI scripts was also made to catch other potential errors like that in the future.

mehrdadh:
@gromero thanks! Updated both title and PR description.

PR #14086, opened 2023-02-22
[WEB] Reduce memleak in web runtime

tqchen:
This PR robustifies the web runtime to reduce memory leak
and enhances the runtime with object support.

Specifically we introduce scoping and auto-release
mechanism when we exit the scope. The improvements
are helpful to deal with memory leak in wasm and webgpu settings.


MasterJH5574 (approved):
Looks good.

PR #14091, opened 2023-02-22
[microTVM][Zephyr] Remove unnecessary use of generate_c_interface_header

mehrdadh:
In standalone mode we can use the generated tvmgen_default.h in model/host/include instead of regenerating it using generate_c_interface_header. 
This PR also moves test_qemu_make_fail to test_zephyr since this test is not a standalone mode test.


driazati (approved):

guberti (approved):
LGTM!

PR #14092, opened 2023-02-22
[Unity] Relax Recursive function

yongwww:
this pull request adds:

- TVMScript local recursive function support
- Update lambda lifting pass. Removed CalledGlobalVars, it was not used anymore.
- Update well-form pass. Previously the [well-form pass](https://github.com/tlc-pack/relax/blob/relax/src/relax/analysis/well_formed.cc#L180-L182) doesn't allow un-defined vars, which doesn't work for var used for recursive call

cc: @YuchenJin @tqchen @Hzfengsy


tqchen (approved):

tqchen:
Thanks @yongwww !

PR #12300, opened 2022-08-03
[TIR] Add DeclBuffer IR node and functors

vinx13:
This PR is part of https://github.com/apache/tvm/issues/11627. It added the IR node `DeclBufferNode`, IR functors and TVMScript printer / parser.

cc @tqchen @Lunderberg @Hzfengsy @junrushao1994 @wrongtest-intellif

PR #14103, opened 2023-02-23
[LLVM] Add support for DeclBufferNode

quic-sanirudh:
`DeclBufferNode` was added recently in #12300, but support to traverse it in codegen_llvm was not added. Since this stmt just declares a new buffer for a body of TIR, we just need to traverse the body to properly generate code for LLVM.


quic-sanirudh:
cc @kparzysz-quic @vinx13

vinx13 (approved):

PR #14104, opened 2023-02-23
[Do not merge]Update docker images with Compute Library v22.11

leandron:
Using this PR to validate with a full set of tests agains Compute Library v22.11


leandron:
cc @neildhickey

PR #14106, opened 2023-02-23
[Unity][BYOC]Add relax backend pattern registry

yelite:
This PR adds pattern registry to manage patterns used by different BYOC backends. It provides API for registering and retrieving patterns. 

The main motivation of this PR is to make it easier to do analysis in backend codegen, by making the pattern behind the partitioned function accessible to codegen.

cc @vinx13 @masahi

### DIFF
--- a/python/tvm/relax/backend/contrib/cutlass.py
+++ b/python/tvm/relax/backend/contrib/cutlass.py
@@ -1,0 +1,68 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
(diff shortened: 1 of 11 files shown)


python/tvm/relax/backend/pattern_registry.py
```diff
@@ -0,0 +1,70 @@
...
+from tvm.runtime import Object
+
+from . import _ffi_api
+
+
+@tvm._ffi.register_object("relax.backend.PatternRegistryEntry")
+class PatternRegistryEntry(Object):
+    name: str
+    pattern: DFPattern
+    arg_patterns: Mapping[str, DFPattern]
```
vinx13, line 33: what's the use case of `arg_patterns`?

python/tvm/relax/backend/contrib/cutlass.py
```diff
@@ -0,0 +1,68 @@
...
+        ),
+        (
+            "cutlass.matmul_bias_gelu",
+            make_matmul_pattern(
+                with_bias=True,
+                activation="relax.nn.gelu",
+            ),
+        ),
+    ]
+)
```
vinx13, line 68: let's also define a `partition_for_cutlass` similar to https://github.com/apache/tvm/blob/main/python/tvm/relay/op/contrib/cutlass.py#L284

python/tvm/relax/backend/pattern_registry.py
yelite: It can be used in the backend codegen to identify input arg expressions in a robust way. The codegen can do something like
```python
match_result = extract_matched_expr(...)
attrs.update({f"{arg_name}_idx": find_arg_index(match_result[arg_pattern] 
              for arg_name, arg_pattern in entry.arg_patterns.items()})
```

masahi (approved):

PR #14107, opened 2023-02-23
[CRT][microTVM] Enable USMP by default for AoTExecutor + CRT runtime

mehrdadh:
This PR enables USMP by default when AoTExecutor and CRT runtime are selected. Check forum discussion about this change: https://discuss.tvm.apache.org/t/enable-usmp-by-default-in-aot-executor-with-runtime-crt/14406

As a result, the workspace memory in mlperftiny project type is removed since memory allocation is not required. If we keep this workspace, the model doesn't fit since some of the memory is allocated twice.


src/relay/backend/aot_executor_codegen.cc
```diff
@@ -44,6 +44,7 @@
 #include <string>
 #include <vector>
 
+#include "../../target/parsers/mprofile.h"
```
Mousius, line 47:
```suggestion
```

tests/micro/arduino/test_arduino_error_detection.py
```diff
@@ -19,13 +19,13 @@
 
 from tvm.micro.project_api.server import ServerError
 
-import test_utils
+from . import utils
```
Mousius, line 22: Any particular reason this is changing as part of this PR? Seems unrelated.

tests/micro/arduino/test_arduino_error_detection.py
```diff
@@ -19,13 +19,13 @@
...
 import tvm.testing
 
 
 @pytest.fixture
 def project(board, microtvm_debug, workspace_dir, serial_number):
-    return test_utils.make_kws_project(board, microtvm_debug, workspace_dir, serial_number)
```
Mousius, line 28: I think `test_utils` is clearer here, `utils` is a fairly vague term.

Mousius (changes requested):
In general this looks fine, I would also default `runtime` to `crt` though, set the new defaults ready for the next release.

Otherwise, there seems be a large unrelated refactor in this patch that confuses which files are effected by the actual USMP change - could you split that out into another PR?

mehrdadh:
@Mousius thanks for the review. I removed unrelated changes. The rest are only removing some unused packages. PTAL.

Mousius (approved):
LGTM! Thanks @mehrdadh, these changes to introduce better default code paths are vital for the overall health of our codebase 😸

PR #14108, opened 2023-02-23
[MetaSchedule] Tile and pack intermediate output for CUDA TensorCore

vinx13:
This PR changes the meta schedule rule for CUDA tensor core. It adds additional tiling and layout transformation (pack the buffer by 16x16 accumulator shape) of the output in the shared memory. By transforming the buffer to higher rank with innermost shape of 16x16, it avoids strided access that `CompactBufferRegion` and `compute_at` cannot handle and reduces shared memory usage for the output

### DIFF
--- a/src/tir/schedule/ir_comparator.cc
+++ b/src/tir/schedule/ir_comparator.cc
@@ -43,7 +43,7 @@ class TensorIntrinMismatchError : public ScheduleError {
     std::ostringstream os;
     os << "The stmt {0} doesn't match the tensor intrin\nThe pattern attempting to be matched:\n"
        << lhs_stmt_ << "\nDoes not match the tensorize description:\n"
-       << rhs_stmt_;
+       << rhs_stmt_ << '\n';
     for (const auto& msg : error_messages_) {
       os << msg << std::endl;
     }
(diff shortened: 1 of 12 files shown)


src/tir/schedule/ir_comparator.h
```diff
@@ -102,8 +102,12 @@ class TensorizeComparator : public ExprComparator, public StmtComparator {
   bool assert_mode_;
   /*! \brief Whether it is visiting the scope block (the outermost block). */
   bool is_scope_block = true;
-  /*! \brief The arithmetic analyzer. */
+  /*! \brief The arithmetic analyzer for comparing LHS and RHS */
   arith::Analyzer analyzer_;
+  /*! \brief The arithmetic analyzer for simplifying expressions on LHS.
```
Hzfengsy, line 107:
```suggestion
  /*!
   * \brief The arithmetic analyzer for simplifying expressions on LHS.
```

Hzfengsy (approved):

Hzfengsy:
Thanks @vinx13

src/tir/schedule/ir_comparator.cc
```diff
@@ -487,7 +490,8 @@ bool TensorizeComparator::CompareBufferRegion(const BufferRegion& lhs, const Buf
         }
         return false;
       }
-      PrimExpr normalized_lhs_min = (lhs->region[i + offset]->min - indices_base[i + offset]);
+      PrimExpr normalized_lhs_min =
```
LeiWang1999, line 493: do we have any specific reason for lhs simplification?

PR #14109, opened 2023-02-24
[Unity] Add callback to FuseOpsByPattern to check match result is accepted

vinx13:
This adds a callback check function to `FuseOpsByPattern` to further check whether the match result should be accepted.

cc @yelite @masahi

### DIFF
--- a/include/tvm/relax/transform.h
+++ b/include/tvm/relax/transform.h
@@ -223,6 +223,8 @@ TVM_DLL Pass FuseOps(int fuse_opt_level = -1);
  * of a fused function after successful matching.
  * \param patterns The patterns to detect. The order of the patterns determines the order
  * of priority in which they are matched. Higher-priority patterns should come earlier in the list.
+ * \param checks The callback functions that takes a match result and returns a boolean value to
+        indicate whether the match result is accepted.
  * \param annotate_codegen If true, wrap each created composite function with another function,
  * whose body consists only of a call to the composite function, and annotate the outer function
  * with kCodegen and kGlobalSymbol attributes. The kCodegen attribute is set as the prefix of the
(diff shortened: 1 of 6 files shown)


src/relax/transform/fuse_ops.cc
```diff
@@ -908,23 +908,27 @@ class PatternBasedPartitioner : ExprVisitor {
...
+    PatternBasedPartitioner part(pattern_name, pattern, check, AnalyzeVar2Value(expr));
     // Initialize each expr to have its own group
     PostOrderVisit(
         expr, [arena, &part](const Expr& e) { part.group_map_[e.get()] = arena->make<Group>(); });
     part.VisitExpr(expr);
     return part.group_map_;
   }
 
-  PatternBasedPartitioner(String pattern_name, DFPattern pattern, const Map<Var, Expr>& bindings)
+  PatternBasedPartitioner(String pattern_name, DFPattern pattern, runtime::PackedFunc check,
```
yelite, line 921: Will it be better to use stronger typing, replacing `PackedFunc` with `TypedPackedFunc`, so that it's clearer on what to pass here?

include/tvm/relax/transform.h
```diff
@@ -223,6 +223,8 @@ TVM_DLL Pass FuseOps(int fuse_opt_level = -1);
  * of a fused function after successful matching.
  * \param patterns The patterns to detect. The order of the patterns determines the order
  * of priority in which they are matched. Higher-priority patterns should come earlier in the list.
+ * \param checks The callback functions that takes a match result and returns a boolean value to
```
yelite, line 226: Taking a match result would work for simple cases, but will it still work for more complicated ones like MHA? I think it would be difficult to retrieve useful information from the match result alone, if the expr contains multiple nodes with similar nature. OTOH, It seems that only the call expr and the match result are available at the call site of checks function.

src/relax/transform/fuse_ops.cc
vinx13: Updated, there are still a few places using `Array<PackedFunc>` because `TypedPackedFunc` can't be Array element

vinx13:
Updated to make check func accept both match mapping and `Call`

masahi:
Need to resolve conflicts with https://github.com/apache/tvm/pull/14139

masahi (approved):

PR #14111, opened 2023-02-24
[TVMScript] Use op attribute to control whether to print dtype in TVMScript

liangW-intellif:
This PR adds an op attribute `TScriptDtypePrintLocation`, and modifies the dtype printing logic of the builtin op to check this attribute. So that user defined operators can use it to specify how there dtype argument are printed by appending attributes instead of appending members to `dtype_first_arg`/`dtype_last_arg`.
@junrushao @wrongtest-intellif

### DIFF
--- a/src/script/printer/tir/expr.cc
+++ b/src/script/printer/tir/expr.cc
@@ -222,23 +222,9 @@ TVM_STATIC_IR_FUNCTOR(IRDocsifier, vtable)
     .set_dispatch<tir::Call>("", [](tir::Call call, ObjectPath call_p, IRDocsifier d) -> Doc {
       static const OpAttrMap<tir::TScriptPrinterName>& op_names =
           Op::GetAttrMap<tir::TScriptPrinterName>("TScriptPrinterName");
-      static const std::unordered_set<const Object*> dtype_first_arg = {
-          tir::builtin::reinterpret().get(),
-          tir::builtin::call_extern().get(),
-          tir::builtin::call_llvm_intrin().get(),       //
-          tir::builtin::call_llvm_pure_intrin().get(),  //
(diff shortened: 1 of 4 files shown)


liangW-intellif:
@tvm-bot rerun

src/script/printer/tir/expr.cc
```diff
@@ -222,23 +222,9 @@ TVM_STATIC_IR_FUNCTOR(IRDocsifier, vtable)
...
-          tir::builtin::mma_fill().get(),
-          tir::builtin::vectorlow().get(),
-          tir::builtin::vectorhigh().get(),
-          tir::builtin::vectorcombine().get(),
-          Op::Get("tir.type_annotation").get(),
-      };
+      static const OpAttrMap<tir::TScriptDtypePrintLocation> dtype_locations =
+          Op::GetAttrMap<tir::TScriptDtypePrintLocation>("TScriptDtypePrintLocation");
+      tir::ScriptDtypePrintLocation dtype_print_location = tir::ScriptDtypePrintLocation::kNone;
       static const std::unordered_set<const Object*> dtype_last_arg = {
```
MasterJH5574, line 228: Thanks for the improvement! I suppose the `dtype_last_arg` can be smoothly removed right?

junrushao: @liangW-intellif we should remove `dtype_last_arg` given it's not used any more

liangW-intellif: Done, thanks for reminding.

MasterJH5574 (approved):

Hzfengsy (approved):

junrushao (approved):

PR #11882, opened 2022-06-24
[Relay][VirtualDevice] Expose WithFields to Python to do proper copy in ExprMutator

rafzi:
As discussed in https://discuss.tvm.apache.org/t/relay-function-virtual-device-property/12958

@mbs-octoml

PR #14114, opened 2023-02-24
[Runtime] Fix Potential DeviceAPIManager Memory Bug

Johnson9009:
This PR do 2 things:

1. Fix a potential memory bug, we meet this bug because our device type is greater than 32, so the cross-border access happened in device api manager code.

2. Remove a useless decorator, it seems there isn't any code use it, and it will cause the Python function become as a packed function, so the caller can't use the key word argument when call it.

### DIFF
--- a/python/tvm/relay/function.py
+++ b/python/tvm/relay/function.py
@@ -98,7 +98,6 @@ class Function(BaseFunc):
         return astext(self, show_meta_data, annotate)
 
 
-@tvm._ffi.register_func("relay.FunctionWithFields")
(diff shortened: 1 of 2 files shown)


Johnson9009:
@cconvey Please help to check 1, caused by your PR #13177.
@rafzi Please help to check 2, introduced by your PR #11882.

tqchen (approved):

rafzi:
@Johnson9009 Looks good, it seems like that was a misunderstanding I had about what the binding requires.

PR #14117, opened 2023-02-24
[Unity][Fix] Fix bug in MergeCompositeFunctions

Ubospica:
Currently `MergeCompositeFunctions` will modify the map while iterating over it, and that makes tests/python/relax/test_transform_merge_composite_functions.py does not pass. This PR fixes this bug.

cc @masahi

### DIFF
--- a/src/relax/transform/merge_composite_functions.cc
+++ b/src/relax/transform/merge_composite_functions.cc
@@ -324,12 +324,16 @@ IRModule MergeCompositeFunctions(IRModule mod) {
   auto new_mod = MakeGroupedFunctions(mod, group_map);
 
   CompositeInliner inliner(mod);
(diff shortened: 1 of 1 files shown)


yongwww (commented):
would you mind adding a test case for this fix?

MasterJH5574:
> would you mind adding a test case for this fix?

@yongwww Such concurrent modification issues are already covered by the test suite. They just show up under certain compilation configuration.

MasterJH5574 (approved):

yongwww:
> > would you mind adding a test case for this fix?
> 
> @yongwww Such concurrent modification issues are already covered by the test suite. They just show up under certain compilation configuration.

I see, interesting, thanks!

PR #14123, opened 2023-02-24
[Unity][TVMScript] emit_te sugar

yongwww:
Add R.emit_te, a requested support https://github.com/tlc-pack/relax/issues/197

cc: @psrivas2 @Hzfengsy @YuchenJin @tqchen

### DIFF
--- a/python/tvm/script/ir_builder/relax/ir.py
+++ b/python/tvm/script/ir_builder/relax/ir.py
@@ -20,12 +20,13 @@
 import builtins
 import functools
 import inspect
-from typing import Any, Dict, List, Optional, Tuple, Union
+from typing import Any, Dict, List, Optional, Tuple, Union, Callable
 
 import tvm
 from tvm import DataType, relax
 from tvm.ir import PrimExpr
 from tvm.relax import Call, Expr, ExternFunc, TupleGetItem, Var, VarBinding, const
+from tvm.relax.block_builder import BlockBuilder as rx_bb
 
 ############################### Operators ###############################
 from tvm.relax.op import (
@@ -304,6 +305,7 @@ invoke_closure = _sinfo_arg_wrapper(invoke_closure)  # pylint: disable=invalid-n
 
 call_builtin_with_ctx = _sinfo_arg_wrapper(call_builtin_with_ctx)  # pylint: disable=invalid-name
 
+
 ############################### Bindings ###############################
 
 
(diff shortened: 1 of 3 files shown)


tqchen (approved):

tqchen:
Thanks @yongwww , as a followup, would be great if we can also confirm if it works on TVMScript parser

python/tvm/script/ir_builder/relax/ir.py
```diff
@@ -325,6 +327,35 @@ def emit(value: Expr, annotate_struct_info: Optional[StructInfo] = None) -> Var:
...
+        to the PrimFunc that gets generated.
+
+    Returns
+    -------
+    var : Var
+        A newly created variable that gets bound to the call code.
+    """
+
+    # Levarage the util function call_te in Relax Block Blocker
+    emit_expr = rx_bb().call_te(func, *args, **kwargs)
```
psrivas2, line 355: this would emit the call binding correctly, but would it add the Primfunc to IRModule?

yongwww: Yes, the actual emit happens later. Actually the call_te is just a helper function, it is used to help prepare the expr to be emitted later.

yongwww:
> Thanks @yongwww , as a followup, would be great if we can also confirm if it works on TVMScript parser

Thanks @tqchen! Is the following test case what you meant?

```python
@tvm.script.ir_module
    class Before:
        @R.function
        def main(x: R.Tensor((10, 20), "float32")) -> R.Tensor(None, "float32", ndim=2):
            # block 0
            with R.dataflow():
                lv0 = R.emit_te(topi.add, x, relax.const(1, "float32"))
                lv1 = R.emit_te(topi.exp, lv0)
                gv = R.emit_te(topi.squeeze, lv1)
                R.output(gv)
            return gv

    Before.show()
```

output of Before.show():
```python
@I.ir_module
class Module:
    @R.function
    def main(
        x: R.Tensor((10, 20), dtype="float32")
    ) -> R.Tensor((10, 20), dtype="float32"):
        with R.dataflow():
            lv = R.call_tir(
                add,
                (x, R.const(1, "float32")),
                out_sinfo=R.Tensor((10, 20), dtype="float32"),
            )
            lv0: R.Tensor((10, 20), dtype="float32") = lv
            lv2 = R.call_tir(exp, (lv0,), out_sinfo=R.Tensor((10, 20), dtype="float32"))
            lv1: R.Tensor((10, 20), dtype="float32") = lv2
            lv4 = R.call_tir(
                squeeze, (lv1,), out_sinfo=R.Tensor((10, 20), dtype="float32")
            )
            gv: R.Tensor((10, 20), dtype="float32") = lv4
            R.output(gv)
        return gv
```

tqchen:
Ah sorry i missed it!

psrivas2:
Thanks for example @yongwww . In the output of `Before.show()` PrimFuncs for `add`, `squeeze` and `exp` are not emitted. I only see the Relax function `main` present in the module.

yongwww:
> Thanks for example @yongwww . In the output of `Before.show()` PrimFuncs for `add`, `squeeze` and `exp` are not emitted. I only see the Relax function `main` present in the module.

ohhhhhh, i see, thanks for pointing this out!!! Will fix it by end of week.

PR #14124, opened 2023-02-25
[Frontend][TFLite] Fix conv2d import bug

mehrdadh:
This PR fixes a bug on TFLite frontend import. Basically for conv2d group cases TFLite was only detecting group on only operators with specific `DEPTHWISE_CONV_2D` definition and ignoring group cases that happens with `CONV_2D`.

I also cleaned up test_forward for this importer.


mehrdadh:
cc @AndrewZhaoLuo for review

masahi (approved):

PR #14127, opened 2023-02-25
[Unity][Legalize] Fix Scalar Constant Legalization

zxybazh:
This PR fixes the issue of loss of data type during Legalization. Previously, if we use a constant scalar in operators like `multiply`, it will automatically be converted to a python data type variable, which may lose its original data type. For example, `float16` may become python `float` and be interpreted as `float32` later.

This is now fixed by avoiding scalar value conversion. The conversion could be added back once we have better support for scalar prim value.

Co-authored by @sunggg and @vinx13.

CC: @MasterJH5574 @tqchen

### DIFF
--- a/python/tvm/relax/transform/legalize_ops/common.py
+++ b/python/tvm/relax/transform/legalize_ops/common.py
@@ -20,7 +20,7 @@ from typing import Callable, Optional, Union
 import tvm
 from tvm import te
 from ...block_builder import BlockBuilder
-from ...expr import Call, Expr
+from ...expr import Call, Expr, Constant
 
 
 ##################### Types #####################
@@ -39,9 +39,11 @@ LegalizeFunc = Callable[[BlockBuilder, Call], Expr]
 ##################### Utilities #####################
 
 
-def _try_convert_to_scalar_const(expr: Expr) -> Union[Expr, bool, float, int]:
+def _try_convert_to_scalar_const(
+    expr: Expr, python_native: bool = False
+) -> Union[Expr, bool, float, int]:
     """Check if the input Expr is a scalar constant.
-    If it is, return its plain value.
+    If it is, return its plain value with the same data type or in native python type.
     If it is not, return the input expr.
 
     Parameters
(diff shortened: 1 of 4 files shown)


sunggg:
Given that some tests fail, I think it would be better to fix this. 
@zxybazh, can we track how we generate TIR function with constants? 
Also, it seems like we also have `float64`. 
In this case, if we convert it to generic float, we may lose some accuracy.

MasterJH5574:
Thanks for catching this and sorry for the late reply. Here the place where we turn scalars into Python native POD value is here
https://github.com/apache/tvm/blob/ec868ebc58175431250c5f56af10749b77347f51/python/tvm/relax/transform/legalize_ops/common.py#L60

Considering this issue, I suppose changing the logic here to “create and return IntImm/FloatImm/...” instead of directly returning Python POD value will work. Would be great if you can help try and confirm this.

zxybazh:
Thanks @MasterJH5574 and @sunggg, I think Ruihang's suggestion makes sense and we can preserve the data type via IntImm and FloatImm. I also added an extra argument of whether to return native python type, because in some cases, e.g. `full` operator, we may want to get the python native value. This should also work with most of the current tests.

For Sung's concern of preserving float type, please see the first added unit test where fp16 can be preserved. I think `fp64` can be preserved in the same way.

Please take another look at your convenience, thanks!

python/tvm/relax/transform/legalize_ops/common.py
```diff
@@ -39,9 +39,11 @@
 ##################### Utilities #####################
 
 
-def _try_convert_to_scalar_const(expr: Expr) -> Union[Expr, bool, float, int]:
+def _try_convert_to_scalar_const(
+    expr: Expr, python_native: bool = False
+) -> Union[Expr, bool, float, int]:
```
MasterJH5574, line 44: Would you mind adding IntImm/FloatImm to the return annotation here, and add some doc string for `python_native`?

MasterJH5574 (commented):
Thanks @zxybazh! It looks good to go. Just one more annotation and doc string nit. Thanks!

python/tvm/relax/transform/legalize_ops/common.py
sunggg: Also, it would be nice to add this potential problem we found with python_native approach.

zxybazh:
Thanks for the advice! Type annotation has been fixed and I added docs explaining loss of data type with use of `python_native=True`.

MasterJH5574 (approved):

PR #14130, opened 2023-02-25
[Fix][TVMScript] Fix index of metadata in printed script

Ubospica:
Currently, if the same metadata object (e.g. a multi-line `tir.StringImm`) is referenced for more than one times in an IRModule, each reference will have different indices of the metadata array. For example, this code

```
str_imm = T.StringImm("aaa\nbbb\n")
@I.ir_module
class Module:
    @T.prim_func
    def foo() -> None:
        A = str_imm
        B = str_imm

    @T.prim_func
    def foo1() -> None:
        A = str_imm
Module.show()
```

where `str_imm` is referenced three times, will generate such output:

```
@I.ir_module
class Module:
    @T.prim_func
    def foo():
        A: T.handle = metadata["tir.StringImm"][0]
        B: T.handle = metadata["tir.StringImm"][1]
        T.evaluate(0)

    @T.prim_func
    def foo1():
        A: T.handle = metadata["tir.StringImm"][2]
        T.evaluate(0)
```

Each time has a different metadata index. 

This PR fixes this problem by detecting duplicate item in `IRDocsifierNode::AddMetadata`.

cc @junrushao @Hzfengsy

### DIFF
--- a/tests/python/unittest/test_tvmscript_printer_metadata.py
+++ b/tests/python/unittest/test_tvmscript_printer_metadata.py
@@ -20,11 +20,9 @@ from tvm.script.parser import ir as I
 from tvm.script.parser import tir as T
 
 
-def _assert_print(obj, expected):
-    assert obj.script(verbose_expr=True).strip() == expected.strip()
-
-
 def test_str_metadata():
+    # This test is to check we reuse the existing metadata element for the same tir.StringImm
+    # So metadata["tir.StringImm"][0] will occur in the printed script for three times
     str_imm = T.StringImm("aaa\nbbb\n")
 
     @I.ir_module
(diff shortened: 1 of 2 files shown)


tests/python/unittest/test_tvmscript_printer_metadata.py
```diff
@@ -0,0 +1,66 @@
...
+            B = str_imm
+
+        @T.prim_func
+        def foo1() -> None:
+            A = str_imm
+
+    _assert_print(
+        Module,
+        """
+# from tvm.script import ir as I
```
Hzfengsy, line 44: Not sure if checking printed string is a good idea, one alternative approach is to check there is only one `tir.StringImm ` metadata

MasterJH5574 (approved):

junrushao:
Not sure if the failed test is relevant (very unlikely). Re-triggered

PR #14132, opened 2023-02-26
[Unity] Add Global info

jinhongyii:
As we start to build more structure into the program, there is an increasing need to incorporate globally static object that are referred by the IR itself. So we introduce "GlobalInfo" here to represent these objects. Below is an example of a possible representation of a distributed program with global info. Pay attention to "mesh", which represents the global infos in the module. It contains a list of two device meshes , and they are accessed using syntax like "mesh[0]", which is a sugar for IRModule.global_infos["mesh"][0]
```
    @I.ir_module
    class TestModule:
        I.module_attrs({"device_num": 10})
        I.module_global_infos(
            {
                "mesh": [
                    R.device_mesh((2, 2), R.range(0, 4)),  # mesh[0]
                    R.device_mesh((1,), R.range(4, 5)),  # mesh[1]
                ]
            }
        )

        @T.prim_func
        def tir_func(
            x: T.Buffer((T.int64(128), T.int64(128)), "float32"),
            y: T.Buffer((T.int64(128), T.int64(128)), "float32"),
        ):
            T.func_attr({"tir.noalias": True})
            for i, j in T.grid(T.int64(128), T.int64(128)):
                with T.block():
                    y[vi, vj] = x[vi, vj] + 1.0

        @R.function
        def foo(
            x: R.DTensor((128, 128), "float32", device_mesh="mesh[0]", placement="S[0], R"),
        ) -> R.DTensor((128, 128), "float32", device_mesh="mesh[0]", placement="S[0], R"):
            gv0 = R.dist.call_tir(
                tir_func,
                x,
                R.DTensor(
                    shape=(128, 128), dtype="float32", device_mesh="mesh[0]", placement="S[0], R"
                ),
            )
            return gv0
```
To summarize the PR, it
1. add a base class GlobalInfo
2. add a global_infos: Map[str, List[GlobalInfo]] field in IRModule
3. support parsing/printing for I.module_attrs and I.module_global_infos.\

cc: @tqchen, @junrushao


tqchen (approved):

yongwww:
Hi @jinhongyii thanks for the work! I have a follow-up question, looks the global info here is very similar to metadata section, is there any difference?  I am wondering if we can merge them

jinhongyii:
Meta data are used to store ObjectRef fragment in the IR that are not printable. However, global infos should be printable

PR #14133, opened 2023-02-26
[Fix][TIR] SampleCategorical apply-to-schedule

MasterJH5574:
This PR is another way to fix the issue described in #14118.

Since we do not have a standard for json file on the format of float numbers (for example, we cannot require a json file producer to print the "integer" float numbers with at least one decimal), and the json parser is not responsible for determining if an integer in a json file should be parsed to a float or an int, the most convenient way of fixing the SampleCategorical issue will be allowing both FloatImms and IntImms as input, and converting all IntImms to FloatImms accordingly.

This PR fixes the issue in this way.


junrushao (approved):
Thanks for the fix! Future forward, we might want to think about auto conversion from IntImm to FloatImm to avoid those cases in general

PR #14139, opened 2023-02-27
[Unity][BYOC] Assign group to unused bindings and ignroe PrimFunc

vinx13:
This PR update `PatternBasedPartitioner` in `FuseOpsByPattern` to allow unused bindings to exist. Previously we apply `PostOrderVisit` which doesn't guarantee all LHS vars in `VarBindingNode` are visited. `FuseOpsByPattern` and `RunCodegen` are also updated to ignore `PrimFunc`

cc @sunggg @masahi @yelite

### DIFF
--- a/src/relax/transform/fuse_ops.cc
+++ b/src/relax/transform/fuse_ops.cc
@@ -942,7 +942,6 @@ class PatternBasedPartitioner : ExprVisitor {
       for (const auto& [pat, match] : matches_opt.value()) {
         // Put all matching call nodes into the parent group.
         if (pat->IsInstance<CallPatternNode>() && match != GetRef<Call>(call)) {
-          // AddToGroup(match, parent_group);
           // Put the bound variable on the LHS into the same parent group.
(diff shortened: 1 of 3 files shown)


src/relax/transform/fuse_ops.cc
```diff
@@ -939,15 +936,13 @@ class PatternBasedPartitioner : ExprVisitor {
...
+      auto parent_group = GetGroupForBoundVar(binding->var);
       ICHECK(parent_group);
       parent_group->attrs.Set(attr::kComposite, pat_name_);
-
       for (const auto& [pat, match] : matches_opt.value()) {
-        ICHECK(group_map_.count(match.get()));
         // Put all matching call nodes into the parent group.
         if (pat->IsInstance<CallPatternNode>() && match != GetRef<Call>(call)) {
-          AddToGroup(match, parent_group);
+          // AddToGroup(match, parent_group);
```
masahi, line 945: Yeah I think we can remove this, since only the group for the binding LHS matters in practice.

masahi (approved):

PR #14140, opened 2023-02-27
[TVMScript][Printer] Remove relax prefix for now

tqchen:
This PR cleans up relax prefix in printer for now. While these setups are useful and do not cause any technical debts in the codebase. We remove it given requests. They can be added back to unity branch and later as part of upstream.


tqchen:
Well, we have already stated the disagreements. As objectively it does not cause any of the issues in the overall codebase and nor does it fundamentally break the printer/parser infrastructure.  It is very different from say introducing an architecture that has not been RFCed. So the argument is at most a nit.

But I feel there is no need for arguing on this for now.

junrushao (approved):

yongwww (approved):

MasterJH5574 (approved):

PR #14144, opened 2023-02-28
[TRT][BYOC] allow strided_slice ops on selected dimensions (#14142)

### LINKED ISSUE #14142: [Bug] Possible error in strided_slice_checker
Hello, I'm trying to use TensorRT for tuning my model. For that matter, I use partition_for_tensorrt before building the model. However, this function fails when processing strided_slice ops. This might be related to strided_slice ops, where fewer axes than tensor dimensions are sliced, e.g., `%435 = strided_slice(%376, begin=[0i64], end=[1i64], strides=[1i64], axes=[0i64])`. I suspect that these ops will fail since the for loop https://github.com/apache/tvm/blob/7e3dc45fed385e12f8cfd9c9284b5551ada82f95/python/tvm/relay/op/contrib/tensorrt.py#L712-L713 iterates over the number of input dimensions while the strided_slice op does not necessarily cover all dimensions, e.g., in the above call only the 0th axis.

### Expected behavior

I expected the partition_for_tensorrt function to return without error.

### Actual behavior

IndexError in strided_slice_checker: https://github.com/apache/tvm/blob/7e3dc45fed385e12f8cfd9c9284b5551ada82f95/python/tvm/relay/op/contrib/tensorrt.py#L713, error thrown at: https://github.com/apache/tvm/blob/7e3dc45fed385e12f8cfd9c9284b5551ada82f95/python/tvm/runtime/container.py#L57

### Environment

OS: Ubuntu 20.04, Kernel 5.15
TVM: built from master branch, commit 0e046daf9e51724b3910aa7ba199069b09e2707e

### Triage

* needs-triage
* byoc:ethosu

AreopagX:
Fixes #14142

### DIFF
--- a/python/tvm/relay/op/contrib/tensorrt.py
+++ b/python/tvm/relay/op/contrib/tensorrt.py
@@ -709,7 +709,8 @@ def strided_slice_checker(
     if any([x is not None and x <= 0 for x in attrs.strides]):
         logger.info(f"{op_name}: stride must be positive")
         return False
-    for i in range(0, len(args[0].checked_type.shape)):
+    length: int = len(attrs.axes) if attrs.axes is not None else len(args[0].checked_type.shape)
+    for i in range(0, length):
         begin = int(attrs.begin[i])
         if attrs.slice_mode == "end":
             end = (


python/tvm/relay/op/contrib/tensorrt.py
```diff
@@ -709,7 +709,9 @@ def strided_slice_checker(
     if any([x is not None and x <= 0 for x in attrs.strides]):
         logger.info(f"{op_name}: stride must be positive")
         return False
-    for i in range(0, len(args[0].checked_type.shape)):
+    if len(attrs.axes) > len(args[0].checked_type.shape):
+        logger.info(f"{op_name}: number of indexed axes is higher than input tensor rank")
+    for i in range(0, len(attrs.axes)):
         begin = int(attrs.begin[i])
```
masahi, line 715: it should be `attrs.begin[attrs.axes[i]]`

python/tvm/relay/op/contrib/tensorrt.py
```diff
@@ -709,7 +709,9 @@ def strided_slice_checker(
     if any([x is not None and x <= 0 for x in attrs.strides]):
         logger.info(f"{op_name}: stride must be positive")
         return False
-    for i in range(0, len(args[0].checked_type.shape)):
+    if len(attrs.axes) > len(args[0].checked_type.shape):
+        logger.info(f"{op_name}: number of indexed axes is higher than input tensor rank")
```
masahi, line 713: No need for these changes, remove L712 and L713.

python/tvm/relay/op/contrib/tensorrt.py
AreopagX: wouldn't this lead to an error if, e.g., attrs={begin: [0], end: [1], axes: [2]}?

python/tvm/relay/op/contrib/tensorrt.py
AreopagX: done

masahi (approved):

masahi:
`test_maskrcnn_resnet50` has failed, you need to handle the case when `axes` is None.

PR #14145, opened 2023-02-28
[Android] Fix using system libraries in Android apps

echuraev:
- Starting from API 31, using `uses-native-library` is required if we
  want to open system library:
  https://developer.android.com/about/versions/12/reference/compat-framework-changes#enforce_native_shared_library_dependencies

  We should specify OpenCL library in `user-native-library` in all
  applications where OpenCL backend might be used.

- Updated README files and describe how to fix synchronization issues
  in Android Studio.


echuraev:
@masahi, @junrushao  could you please take a look at this PR?

masahi (approved):

PR #14146, opened 2023-02-28
[Unity][Pass] Enhance constant folding to fold relax ops by evaluating them.

psrivas2:
Enhance constant folding to fold relax ops by evaluating them.
This uses the registered legalization function attached to the op to lower it to `call_tir` and uses the existing `call_tir` folding mechanism to fold it.

This kind of op folding is only allowed within dataflow block as ops could have side-effects.

Limitations:
* This currently does not support folding ops that could lower to multiple `call_tir` bindings.
* Folding by evaluating ops is not always beneficial. We need a heuristic to check if it is useful. This is not implemented yet and folding is always allowed by evaluating expressions.

### DIFF
--- a/python/tvm/relax/transform/legalize_ops/common.py
+++ b/python/tvm/relax/transform/legalize_ops/common.py
@@ -90,7 +90,7 @@ def _call_topi_without_attr(te_func: TEFunc, primfunc_name: Optional[str] = None
 _LEGALIZE_ATTR_NAME = "FLegalize"
 
 
-def register_legalize(op_name: str, legal_func: LegalizeFunc = None, level: int = 10):
+def register_legalize(op_name: str, legal_func: LegalizeFunc = None):
     """Register legal transformation function for a Relax op.
 
     Parameters
@@ -100,8 +100,5 @@ def register_legalize(op_name: str, legal_func: LegalizeFunc = None, level: int 
 
     legal_func: function (bb: BlockBuilder, call: Call) -> new_expr: Expr
         The function for transforming an expr to another expr.
-
-    level: int, optional
-        The priority level
     """
-    return tvm.ir.register_op_attr(op_name, _LEGALIZE_ATTR_NAME, legal_func, level)
+    return tvm.ir.register_op_attr(op_name, _LEGALIZE_ATTR_NAME, legal_func)
--- a/python/tvm/relax/transform/legalize_ops/common.py
+++ b/python/tvm/relax/transform/legalize_ops/common.py
@@ -90,7 +90,7 @@ def _call_topi_without_attr(te_func: TEFunc, primfunc_name: Optional[str] = None
 _LEGALIZE_ATTR_NAME = "FLegalize"
 
 
-def register_legalize(op_name: str, legal_func: LegalizeFunc = None):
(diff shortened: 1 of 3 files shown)


psrivas2:
cc @slyubomirsky @sunggg @jinhongyii

python/tvm/relax/transform/legalize_ops/common.py
```diff
@@ -90,7 +90,7 @@ def _call_topi_without_attr(te_func: TEFunc, primfunc_name: Optional[str] = None
 _LEGALIZE_ATTR_NAME = "FLegalize"
 
 
-def register_legalize(op_name: str, legal_func: LegalizeFunc = None):
+def register_legalize(op_name: str, legal_func: LegalizeFunc = None, level: int = 10):
```
sunggg, line 93: Is there any necessity for `level` in this PR? 
Otherwise, I would prefer not introducing `level` yet and keep the legalizer dedicated to the reference implementation.  
If we want to lower to different implementation depending on the context, we can introduce target-specialization pass. In this way, I believe we can make the lowering process composable.

src/relax/transform/fold_constant.cc
```diff
@@ -127,6 +129,19 @@ class ConstantFolder : public ExprMutator {
...
 
+  /*!
+   * \brief Checks if it is useful to fold \p expr.
+   * \details Folding an expr is a trade-off - we are materializing a constant in the IRModule and
+   * paying compile time cost to avoid the cost of executing this expr at runtime. For example,
+   * folding iota ops could result in large constants being materialized, thus increasing the size
+   * of the program.
+   */
+  bool ShouldBeFolded(Expr expr) {
+    // TODO(prakalp): Implement a heuristic to check if folding this op is actually useful or
```
sunggg, line 140: Would this include whether this `expr` is eligible or not? 
Or do we have a check before this heuristic and every `expr` at this point is eligible?

sunggg (commented):
Thank you for the quick action!
Please see my comments below.

python/tvm/relax/transform/legalize_ops/common.py
psrivas2: I introduced that change for testing purpose. reverted it.

slyubomirsky (approved):
I'm glad that this change turned out to be so compact and relatively simple. I see no issues with the implementation.

src/relax/transform/fold_constant.cc
psrivas2: This would not check if the expression is eligible, but it need not check that too. It would probably only need to check expr input and output shapes and decide if the folding would be useful. 

Another place to do this could be post matching call_tir and figured out its constant input args. We could then pass input/output shapes to figure out if folding would be useful.

yongwww (approved):
lgtm

src/relax/transform/fold_constant.cc
MasterJH5574: I’m curious about the plan of the heuristics here. As this check is standalone, it seems to me that we can introduce this function together with followup PRs.

psrivas2: I introduced it here so we don't forget about it. Indeed there are other ways to achieve the same, perhaps a github issue. Is that preferred?

PR #14148, opened 2023-02-28
[Unity][Draft][WIP] Relax language specification

slyubomirsky:
[Rendered view.](https://github.com/slyubomirsky/tvm/blob/relax-spec-draft/relax_spec.md)

(Transferred from https://github.com/tlc-pack/relax/pull/273, see the old PR for older discussions and comments.)

This document is a draft language specification for Relax. The purpose of the language specification is to serve as a technical reference describing the language's behavior in sufficient detail as to clarify the intended behavior of the compiler, hence it is by design a very detailed document rather than an accessible tutorial for the language. Its focus is the "what" and "how" of Relax, but not always the "why," though we can add more sections giving design reasons if that is desired.

Note that «double caret marks» (guillemets) are used to denote parts of the specification discussing functionality that the present prototype doesn't yet support. This notation is somewhat cumbersome, but I wasn't sure how else to proceed because Github Markdown does not support changing the text color (which was how my initial document indicated these areas). The caret marks may look strange in the text, but they have the benefit of being easy to find by text search.

Out of scope, for now, in this document is the subject of parsing: We should eventually document how we intend to parse Python into Relax, but the parser itself is being greatly reworked. We can revisit the issue of documenting its behavior once that work has been completed. Additionally, this specification is intended (for now) to focus on the user-visible behavior of Relax rather than specifying lower-level interfaces or the precise mechanisms of Relax's implementation.

# Aspects Requiring Review or Still to be Determined

Since this document is a draft, any part of it is up for review and open to revision, but certain parts of the document have proven particularly challenging to describe and could benefit the most from community discussions.

1. The `StructInfo` system has been a great challenge to specify and there are many questions as to how it should work. One question is whether "strong shaping" might lead to too many error messages for something that could be checked dynamically. Additionally, many potential shape mismatches could be eliminated using constant propagation or other transformations: If we check shapes without applying transformations, we would force users to add lots of redundant shape checks. On the other hand, if we require these transformations first, that might make the code harder for users to reason about.
2. The run-time representations of values in the language will be important for determining how `PackedFunc`s can interact with Relax values. However, embedded targets do not support the TVM object system, so describing values in terms of the TVM object system directly may not work for all settings. Additionally, it should be determined how much detail about the representations should be included in the specification.
3. Operators used for core language functions (like `call_tir`) should be described in the specification. I am not certain the descriptions presently in the last section are entirely correct, so more review of them should be appreciated. Additionally, are there operators that should be there but are presently missing?
4. Finally, there is the question of _process_: How will we permit the specification to be revised? Does any change require a fresh RFC? Is there a threshold for changes that can be done as a direct PR? I have not considered this question directly, but the language specification is an important document for the community and changes to the language specification should not be taken lightly.

There are also some more minor TODOs throughout the document.

# The Future of This Document

Eventually, we will want this document to be part of the Relax documentation, in which case it will be placed into a different location in the repo and probably be formatted as an `rst` file. Before that, we will officially RFC the spec into TVM to allow for maximum public discussion as to the design decisions underlying the specification. Hence, I am going to leave this document as a "WIP PR" until it is officially RFC'd.


relax_spec.md
```diff
@@ -0,0 +1,803 @@
...
+
+### Notes on `DataType` and Related Terminology
+
+The representation of datatypes, `DataType`, in the above AST is taken directly from TIR. However, the usage of datatypes in Relax is more restricted than in TIR.
+1. The `lanes` field for the `Int`, `UInt`, and `Float` datatypes must always be 1; we do not directly consider vectorized values in Relax.
+2. The `lanes` field for the `Handle` datatype must always be 0, indicating that it is `Void` (see below). The `bits` field for `Handle` should always be set to 64 (it will not be used by Relax).
+
+We also define the following special notation for datatypes, to be used in the rest of the specification:
+1. `Bool()`: This is shorthand for `UInt(bits=1, lanes=1)`, since TIR does not have a separate Boolean type. "True" refers to a value of 1 in this datatype and "false" refers to a value of 0. For convenience, we will refer to Boolean values as a separate datatype in the specification, due to their significance in `If` nodes.
+2. `Void()`: This is shorthand for `Handle(bits=64, lanes=0)`. TIR uses this datatype to refer to opaque objects; in Relax, it is used to denote an unknown datatype.
```
kparzysz-quic, line 115: In TIR `void` has `bits=0` and `lanes=0`.  Why the deviation in Relax?

Also, `void` in TIR doesn't mean "opaque", it means "nothing".  The use of "kHandle" for it doesn't carry any special meaning.  If `void` in Relax is intended to mean something else, then we should pick a different name to avoid confusion.

relax_spec.md
```diff
@@ -0,0 +1,803 @@
...
+11. `TupleGetItem` nodes represent tuple indexing. The `tuple_value` expression must evaluate to a tuple with at least `index + 1` items and the item with the given index will be returned.
+12. `SeqExpr` describes a sequence of binding blocks followed by a return expression. The `SeqExpr` opens a new scope. Its binding blocks are evaluated in order and add new variables to the scope. Binding blocks are either ordinary `BindingBlock`s or `DataflowBlock`s and both consist of a series of bindings. `DataflowBlock`s are the only kind allowed to introduce bindings with `DataflowVar`s and it does not permit any constructs featuring control flow (`If` nodes or recursive calls) or calls to (possibly) impure functions. There are two different kinds of bindings:
+    1. `VarBinding`s: The `value` expression (the right-hand side of the binding) of the binding is evaluated first and is bound to the `var` expression, which must be a new `Var` or `DataflowVar` (in a dataflow block). The newly bound variable will have that value for the remainder of the scope (`DataflowVar`s are scoped only to the `DataflowBlock` in which they appear; `Var`s are scoped to the entire `SeqExpr`).
```
kparzysz-quic, line 138: If "`Var`s are scoped to the entire `SeqExpr`", does it mean that a var can be used before it's defined as long as it's in the scope of the same `SeqExpr`?

relax_spec.md
```diff
@@ -0,0 +1,803 @@
...
+
+The representation of datatypes, `DataType`, in the above AST is taken directly from TIR. However, the usage of datatypes in Relax is more restricted than in TIR.
+1. The `lanes` field for the `Int`, `UInt`, and `Float` datatypes must always be 1; we do not directly consider vectorized values in Relax.
+2. The `lanes` field for the `Handle` datatype must always be 0, indicating that it is `Void` (see below). The `bits` field for `Handle` should always be set to 64 (it will not be used by Relax).
+
+We also define the following special notation for datatypes, to be used in the rest of the specification:
+1. `Bool()`: This is shorthand for `UInt(bits=1, lanes=1)`, since TIR does not have a separate Boolean type. "True" refers to a value of 1 in this datatype and "false" refers to a value of 0. For convenience, we will refer to Boolean values as a separate datatype in the specification, due to their significance in `If` nodes.
+2. `Void()`: This is shorthand for `Handle(bits=64, lanes=0)`. TIR uses this datatype to refer to opaque objects; in Relax, it is used to denote an unknown datatype.
+
+## Expression Survey
```
kparzysz-quic, line 117: We need the AST spec here.  I find the description below confusing.  For example `PrimExpr` vs `PrimValue`...

relax_spec.md
slyubomirsky: Good point. I should rephrase that to mean from the point of its definition in a binding onwards.

relax_spec.md
slyubomirsky: That was my understanding from TIR; from what I have gathered in discussions with others, void dtypes in TIR can be used to refer to objects that TIR should not operate on except through builtins/intrinsics. Whatever we use for Void in Relax should match TIR; I will check the resulting ASTs.

relax_spec.md
slyubomirsky: Would you recommend including the structure of the nodes in the below description? The AST diagram is in the section above, so I'm not sure where I would move it.

relax_spec.md
slyubomirsky: As for whether we should use a different name, that would be worth discussing. My impression is that void was chosen for convenience (reusing the same dtype AST from TIR).

relax_spec.md
kparzysz-quic: No, it was a brainfart.  I was focusing on reading this section and I forgot what I read 10 minutes prior...  :|

slyubomirsky:
Updated to account for #14394, please review.

slyubomirsky:
Specified `FuncStructInfo` for `PrimFunc`s and direct calls to `PrimFunc`s based on this PR: https://github.com/apache/tvm/pull/15026

relax_spec.md
```diff
@@ -0,0 +1,914 @@
...
+6. If `struct_info` is `FuncStructInfo(params, ret, purity, derive_func)`, then if `params` is defined, check that `value` is a closure value; if `derive_func` is defined, check that `value` is a `PackedFunc`. No further validation may be done on a `PackedFunc`. «If `value` is a closure value, then it can contain run-time structural information indicating its purity and the structural information of its intended arguments and return value that can be compared against `purity`, `params`, and `ret`.»
+
+### Checking Structural Information at the Start and End of a Function
+
+«Shape variables are bound at the start and end of a function or in `MatchCast` bindings. This checking is done similarly to `MatchCast`, though with a slight difference: per rule #6 in under the [well-formedness criteria](#well-formedness-criteria), we allow shape variables to appear in arguments in any order so long as shape variables appear in a binding position at least once. This requires us to check the shapes of arguments dimension by dimension in a specific order.
```
slyubomirsky, line 369: This approach was discussed here: https://github.com/apache/tvm/pull/15577#discussion_r1307947120

I am not sure this is the best policy, since it makes the checking of function signatures more complex. However, I am led to understand that this is more convenient in certain situations. We should be sure the implementation follows this policy.

slyubomirsky:
Added semantics for heterogenous computation per #15823. Please have a look @yongwww! Some of the proposed rules haven't been implemented yet but I think we might need to introduce new invariants to keep things consistent and address some cases that the current unit tests don't check.

slyubomirsky:
Now that Unity is in mainline I work to make this an RFC instead, so see the upcoming RFC for future updates.

PR #14149, opened 2023-02-28
[Unity][Analysis] Analysis for detecting recursion in Relax

slyubomirsky:
This PR adds an analysis that constructs a Relax dependency graph in a module (which global functions reference which other global functions) and detects recursive functions, both simple recursion and mutual recursion. Simple recursion is a self-cycle in the dependency graph and mutual recursion is a directed cycle. The analysis works by using [Johnson's algorithm for detecting elementary circuits](https://www.cs.tufts.edu/comp/150GA/homeworks/hw1/Johnson%2075.PDF) (directed cycles where each node is visited at most once) in the dependency graph and then coalescing all elementary circuits that have at least one node in common to form groups of mutually recursive functions. This can be useful for future passes that need special handling for recursion or mutual recursion.


python/tvm/relax/analysis/analysis.py
```diff
@@ -319,3 +320,35 @@ def suggest_layout_transforms(
...
+
+    (Note that functions do not have to call each other to reference each other.
+    For example, if a function returns another function, that is still a reference
+    that could potentially be recursive, even without a call.)
+
+
+    If a function is simply recursive and not mutually recursive with any other,
+    it will be reported as a group by itself.
+
+        Parameters
```
Hzfengsy, line 342:
```suggestion
    Parameters
```

Hzfengsy (approved):
Overall LGTM

psrivas2 (approved):
This looks great, thanks!
LGTM!

yongwww (approved):
Overall look great to me!

PR #14150, opened 2023-03-01
[Unity][OP] Add an operator for fused multi head attention

cyx-6:
This PR introduces the new relax operator `R.nn.attention` for fused multi head attention, and the support of fused multi head attention to relax cutlass BYOC. The input of the operator are query, key and value tensor, with `BSNH` layout, namely `[batch size, sequence length, number of heads, dimension of heads]`. And the output shares the same layout with all input tensor.

### DIFF
--- a/include/tvm/relax/attrs/nn.h
+++ b/include/tvm/relax/attrs/nn.h
@@ -184,6 +184,15 @@ struct DropoutAttrs : public tvm::AttrsNode<DropoutAttrs> {
   }
 };  // struct DropoutAttrs
 
+/*! \brief Attributes used in fuse multi head attention operator */
+struct AttentionAttrs : public tvm::AttrsNode<AttentionAttrs> {
+  DataType out_dtype;
+
+  TVM_DECLARE_ATTRS(AttentionAttrs, "relax.attrs.AttentionAttrs") {
+    TVM_ATTR_FIELD(out_dtype).describe("The data type of the output tensor");
+  }
+};  // struct AttentionAttrs
+
 }  // namespace relax
 }  // namespace tvm
 
--- a/include/tvm/relax/attrs/nn.h
+++ b/include/tvm/relax/attrs/nn.h
@@ -184,15 +184,6 @@ struct DropoutAttrs : public tvm::AttrsNode<DropoutAttrs> {
   }
 };  // struct DropoutAttrs
 
-/*! \brief Attributes used in fuse multi head attention operator */
-struct AttentionAttrs : public tvm::AttrsNode<AttentionAttrs> {
-  DataType out_dtype;
-
-  TVM_DECLARE_ATTRS(AttentionAttrs, "relax.attrs.AttentionAttrs") {
-    TVM_ATTR_FIELD(out_dtype).describe("The data type of the output tensor");
-  }
-};  // struct AttentionAttrs
-
 }  // namespace relax
 }  // namespace tvm
 
--- a/python/tvm/contrib/cutlass/attention_operation.py
+++ b/python/tvm/contrib/cutlass/attention_operation.py
@@ -23,15 +23,32 @@ def instantiate_attention_template(attrs, func_args):
     """Return CUTLASS host code for fused multi head attention
     based on a template and the provided attribute map."""
 
-    bias_template = """
+    bias_template = {
+        "B11S'": """
+  CHECK(${arg3}->ndim == 2); // B, 1, 1, S'
+
+  p.attn_bias_ptr = reinterpret_cast<T *>(${arg3}->data);
+  p.bias_strideM = 0; // 0
+  p.bias_strideH = 0; // 0
+  p.bias_strideB = p.num_keys; // S'
+""",
+        "B1SS'": """
+  CHECK(${arg3}->ndim == 3); // B, 1, S, S'
+
+  p.attn_bias_ptr = reinterpret_cast<T *>(${arg3}->data);
+  p.bias_strideM = p.num_keys; // S'
+  p.bias_strideH = 0; // 0
+  p.bias_strideB = p.bias_strideM * p.num_queries; // S' * S
+""",
+        "BNSS'": """
   CHECK(${arg3}->ndim == 4); // B, N, S, S'
 
   p.attn_bias_ptr = reinterpret_cast<T *>(${arg3}->data);
   p.bias_strideM = p.num_keys; // S'
   p.bias_strideH = p.bias_strideM * p.num_queries; // S' * S
   p.bias_strideB = p.bias_strideH * p.num_heads; // S' * S * N
-
-"""
+""",
+    }
 
     template = """
   using T = ${data_type};
@@ -107,7 +124,9 @@ def instantiate_attention_template(attrs, func_args):
   kernel_fn<<<p.getBlocksGrid(), p.getThreadsGrid(), smem_bytes>>>(p);
 """
     if attrs["kSupportsBias"]:
-        template = substitute_template(template, {"bias_template": bias_template})
+        template = substitute_template(
+            template, {"bias_template": bias_template[attrs["bias_layout"]]}
+        )
     else:
         template = substitute_template(template, {"bias_template": ""})
     for i, arg in enumerate(func_args):
--- a/python/tvm/contrib/cutlass/attention_operation.py
+++ b/python/tvm/contrib/cutlass/attention_operation.py
@@ -24,7 +24,7 @@ def instantiate_attention_template(attrs, func_args):
     based on a template and the provided attribute map."""
 
     template = """
-  using T = cutlass::half_t;
+  using T = ${data_type};
 
   CHECK(${arg0}->ndim == 4); // B, S, N, H
   CHECK(${arg1}->ndim == 4); // B, S', N, H
@@ -34,7 +34,7 @@ def instantiate_attention_template(attrs, func_args):
   using Attention =
       AttentionKernel<T,
                       /*ArchTag=*/${arch},
-                      /*is_aligned=*/true,
+                      /*is_aligned=*/${kIsAligned},
                       /*queries_per_block=*/${kQueriesPerBlock},
                       /*keys_per_block=*/${kKeysPerBlock},
                       /*single_value_iteration=*/${kSingleValueIteration}
@@ -47,8 +47,13 @@ def instantiate_attention_template(attrs, func_args):
   p.value_ptr = reinterpret_cast<T *>(${arg2}->data);
   p.logsumexp_ptr = nullptr;
   p.output_ptr = reinterpret_cast<T *>(out0->data);
-  static_assert(!Attention::kNeedsOutputAccumulatorBuffer);
   p.output_accum_ptr = nullptr;
+  if (Attention::kNeedsOutputAccumulatorBuffer) {
+    cudaMalloc(
+      &p.output_accum_ptr, 
+      ${output_size} * sizeof(Attention::output_accum_t)
+    );
+  }
 
   p.num_heads = ${num_heads}; // N
   p.num_batches = ${num_batches}; // B
@@ -57,7 +62,6 @@ def instantiate_attention_template(attrs, func_args):
   p.num_queries = ${num_queries}; // S
   p.num_keys = ${num_keys}; // S'
   p.scale = 1.0f / sqrt(float(${head_dim}));
-  // p.causal = false;
 
   // stride for N
   p.q_strideH = p.head_dim; // H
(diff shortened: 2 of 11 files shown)


tqchen:
thanks @cyx-6 , can we also add a note about existing attention operator layout in libraries(HF and PT), that would help us to write down the rationale of the operator

tests/python/relax/test_codegen_cutlass.py
```diff
@@ -438,5 +439,79 @@ def test_matmul_transposed_bias_gelu_offload(matmul_x, matmul_y, matmul_bias):
...
+                result = R.emit(R.nn.attention(q, k, v))
+                R.output(result)
+
+            R.func_ret_value(frame.output_vars[0])
+
+    func = builder.get()
+    return tvm.IRModule({"main": func})
+
+
+def get_numpy_attention_ref(q, k, v):
```
tqchen, line 495: use tvm.contrib.memoize so we don't need to run it multiple times

src/relax/op/nn/attention.h
```diff
@@ -0,0 +1,41 @@
...
+
+#include <tvm/relax/attrs/nn.h>
+
+#include "../op_common.h"
+
+namespace tvm {
+namespace relax {
+
+/*! \brief fused multi head attention */
+Expr attention(Expr query, Expr key, Expr value, DataType out_dtype);
```
masahi, line 36: what about bias?

masahi (commented):
Please add a test case for the Relax op.

python/tvm/contrib/cutlass/attention_operation.py
```diff
@@ -0,0 +1,95 @@
...
+
+  CHECK(${arg0}->ndim == 4); // B, S, N, H
+  CHECK(${arg1}->ndim == 4); // B, S', N, H
+  CHECK(${arg2}->ndim == 4); // B, S', N, H'
+  CHECK(out0->ndim == 4); // B, S, N, H'
+
+  using Attention =
+      AttentionKernel<T,
+                      /*ArchTag=*/${arch},
+                      /*is_aligned=*/true,
```
vinx13, line 37: we also need to dispatch `is_aligned` based on input shape

python/tvm/contrib/cutlass/attention_operation.py
```diff
@@ -0,0 +1,95 @@
...
+  p.q_strideM = p.q_strideH * p.num_heads; // H * N
+  p.k_strideM = p.k_strideH * p.num_heads; // H * N
+  p.v_strideM = p.v_strideH * p.num_heads; // H' * N
+  p.o_strideM = p.head_dim_value * p.num_heads; // H' * N
+
+  // stride for B
+  p.q_strideB = p.q_strideM * p.num_queries; // H * N * S
+  p.k_strideB = p.k_strideM * p.num_keys; // H * N * S'
+  p.v_strideB = p.v_strideM * p.num_keys; // H'* N * S'
+  // p.o_strideB = p.o_strideM * p.num_queries; // H'* N * S
```
vinx13, line 78: remove if not needed

python/tvm/contrib/cutlass/attention_operation.py
```diff
@@ -0,0 +1,95 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+# pylint: disable=invalid-name, unused-wildcard-import, wildcard-import
+"""Generator for CUTLASS attention kernels."""
+from .library import *
+
+
+def instantiate_attention_template(attrs, func_args):
```
sunggg, line 22: Correct me if i'm wrong, but I heard there are various ways to implement attention. 
Is there any specific form (maybe something standard) of attention we target? 
Or can we also support a set of variants?

sunggg (commented):
Thank you for exciting contribution, @cyx-6! 
Overall, it looks good. Please see my comments below.

python/tvm/contrib/cutlass/attention_operation.py
vinx13: this only implements `softmax(Q*K.T) * V` in BSNH layout. We can support other layout, and fused QKV in the future by changing the strides

spectrometerHBH:
Do we consider adding it to the Fx translator since PT has MHA op?
https://pytorch.org/docs/stable/generated/torch.nn.MultiheadAttention.html

masahi:
cc @hwu36

hwu36:
@danthe3rd @mnicely

tests/python/relax/test_codegen_cutlass.py
cyx-6: Sure! I have refactored the unittest codes and add the memoize decorator.

include/tvm/relax/attrs/nn.h
```diff
@@ -184,6 +184,15 @@ struct DropoutAttrs : public tvm::AttrsNode<DropoutAttrs> {
   }
 };  // struct DropoutAttrs
 
+/*! \brief Attributes used in fuse multi head attention operator */
+struct AttentionAttrs : public tvm::AttrsNode<AttentionAttrs> {
+  DataType out_dtype;
```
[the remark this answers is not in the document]
cyx-6: Yes. The `AttentionAttrs` is redundant indeed. I have removed them.

python/tvm/contrib/cutlass/attention_operation.py
cyx-6: Sure. The commented codes have been removed.

python/tvm/contrib/cutlass/attention_operation.py
```diff
@@ -0,0 +1,93 @@
...
+  CHECK(out0->ndim == 4); // B, S, N, H'
+
+  using Attention =
+      AttentionKernel<T,
+                      /*ArchTag=*/${arch},
+                      /*is_aligned=*/true,
+                      /*queries_per_block=*/${kQueriesPerBlock},
+                      /*keys_per_block=*/${kKeysPerBlock},
+                      /*single_value_iteration=*/${kSingleValueIteration}
+      >;
```
danthe3rd, line 41: Note: if you don't need it, you should set `kSupportsDropout` and `kSupportsBias` to false for better performance
https://github.com/NVIDIA/cutlass/blob/f396cdd15ccf873d5c92ae73b73ed680c77f4400/examples/41_fused_multi_head_attention/kernel_forward.h#L107-L108

python/tvm/contrib/cutlass/attention_operation.py
slyubomirsky: Is that the difference between Bahdanau and Luong attention? (I don't have a ton of experience with attention mechanisms)

src/relax/op/nn/attention.h
vinx13: yeah bias can be useful in some cases. @cyx-6 can we add an optional input `bias` for attention op?

cyx-6: yes, I am adding it

MasterJH5574:
Just note that 3rdparty/cutlass is changed by this PR. Could you revert this change?

python/tvm/contrib/cutlass/attention_operation.py
vinx13: it's mostly different variants of mask/bias apply to Q*K.T

python/tvm/contrib/cutlass/gen_tensor_op.py
```diff
@@ -652,4 +653,25 @@ def get_batch_stride(stride_annot, arg0_idx, arg1_idx, arg0_axis_idx, arg1_axis_
         code = instantiate_conv2d_template(attrs, func_args)
         return CodegenResult(code, headers)
 
+    elif "attention" in func_name:
+        headers.append("kernel_forward.h")
+        attrs["num_batches"] = str(int(annotations["num_batches"]))
+        attrs["num_queries"] = str(int(annotations["num_queries"]))
+        attrs["num_keys"] = str(int(annotations["num_keys"]))
+        attrs["num_heads"] = str(int(annotations["num_heads"]))
+        attrs["head_dim"] = str(int(annotations["head_dim"]))
```
masahi, line 662: Feel free to update `substitute_template(...)` to get rid off `str(int(...))` stuff. I can also do it later otherwise.

python/tvm/contrib/cutlass/attention_operation.py
cyx-6: the dispatch for alignment has been added

python/tvm/contrib/cutlass/gen_tensor_op.py
cyx-6: The `substitute_template` is now updated to accept `int`, `IntImm` and `bool`. And all the usages of `str(int(...))` have been removed, including usages in conv2d and matmul.

masahi: Thanks!

python/tvm/contrib/cutlass/gen_tensor_op.py
```diff
@@ -653,4 +655,34 @@ def get_batch_stride(stride_annot, arg0_idx, arg1_idx, arg0_axis_idx, arg1_axis_
...
+            attrs["kIsAligned"] = False
+        else:
+            raise NotImplementedError()
+        if h_v > 64:
+            attrs["kQueriesPerBlock"] = "32"
+            attrs["kKeysPerBlock"] = "128"
+            attrs["kSingleValueIteration"] = h_v <= 128
+        else:
+            attrs["kQueriesPerBlock"] = "64"
+            attrs["kKeysPerBlock"] = "64"
```
masahi, line 681: We can also remove the quotation now.

cyx-6: Oh, I missed that. Fixed already.

src/relax/op/nn/attention.h
cyx-6: The bias for `Q @ K + bias` has been added. The kernel currently does not support the bias for `softmax(Q @ K) @ V + bias`. We may find a solution in the future if needed.

python/tvm/contrib/cutlass/attention_operation.py
cyx-6: Thanks for your valuable advice! I have set the `false` for these two fields by default. :)

vinx13 (approved):

python/tvm/relax/op/nn/nn.py
```diff
@@ -577,3 +577,41 @@ def cross_entropy_with_logits(predictions: Expr, labels: Expr) -> Expr:
...
+    key: relax.Expr
+        The input key to the operator. The layout of the input key should be
+        (batch_size, seq_len_kv, num_head, head_dim).
+
+    value: relax.Expr
+        The input value to the operator. The layout of the input value should be
+        (batch_size, seq_len_kv, num_head, head_dim_v).
+
+    bias: Optional[Expr]
+        The optional attention bias to the operator. The layout of the attention bias should be
```
vinx13, line 608: it's more common to have BNSS' and it allows broadcastable shape. Let's say
```
The shape of the attention bias should be able to broadcast to (batch_size, num_head, seq_len, seq_len_kv)
```

cyx-6: Sure! I have updated the layout of the bias to `BNSS'`.

python/tvm/contrib/cutlass/attention_operation.py
```diff
@@ -24,12 +24,12 @@ def instantiate_attention_template(attrs, func_args):
...
     bias_template = """
-  CHECK(${arg3}->ndim == 4); // B, S, N, S'
+  CHECK(${arg3}->ndim == 4); // B, N, S, S'
 
   p.attn_bias_ptr = reinterpret_cast<T *>(${arg3}->data);
-  p.bias_strideH = p.num_keys; // S'
-  p.bias_strideM = p.bias_strideH * p.num_heads; // S' * N
-  p.bias_strideB = p.bias_strideM * p.num_queries; // S' * N * S
+  p.bias_strideM = p.num_keys; // S'
+  p.bias_strideH = p.bias_strideM * p.num_queries; // S' * S
```
vinx13, line 31: we should also handle broadcast case

cyx-6: The broadcast for bias is added. Now we accept the bias with layout of `BS'`, `BSS'` and `BNSS'`.

vinx13 (approved):

tqchen (approved):

masahi:
The test was added only for cutlass BYOC. Shouldn't we have a standalone test for Relax op + legalize?

cyx-6:
@masahi Yes, as long as it is a Relax op, we should have it fully tested. I will add those necessary tests in the following PR.

PR #14151, opened 2023-03-01
[Unity] Add bind_constants option to FuseOpsByPattern

vinx13:
This PR added `lift_constants` option to FuseOpsByPattern since different BYOC backends have different requirements on whether constants should be lifted.

cc @masahi

### DIFF
--- a/python/tvm/relax/backend/contrib/cutlass.py
+++ b/python/tvm/relax/backend/contrib/cutlass.py
@@ -117,6 +117,6 @@ def partition_for_cutlass(mod):
     """
 
     cutlass_patterns = get_patterns_with_prefix("cutlass")
-    return transform.FuseOpsByPattern(cutlass_patterns, lift_constants=True, annotate_codegen=True)(
+    return transform.FuseOpsByPattern(cutlass_patterns, bind_constants=True, annotate_codegen=True)(
         mod
     )
--- a/python/tvm/relax/backend/contrib/cutlass.py
+++ b/python/tvm/relax/backend/contrib/cutlass.py
@@ -117,4 +117,6 @@ def partition_for_cutlass(mod):
     """
 
     cutlass_patterns = get_patterns_with_prefix("cutlass")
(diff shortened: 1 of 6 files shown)


tests/python/relax/test_codegen_cutlass.py
```diff
@@ -226,7 +226,7 @@ def get_result_with_relax_cutlass_offload(mod, *args):
 
     seq = tvm.transform.Sequential(
         [
-            relax.transform.FuseOpsByPattern(patterns, annotate_codegen=True),
+            relax.transform.FuseOpsByPattern(patterns, lift_constants=True, annotate_codegen=True),
```
masahi, line 229: I didn't realize that we have `partition_for_cutlass` for Relax, let's use that here.

masahi (commented):
I prefer `bind_constants = True` to `lift_constants = False`. The name `lift_constants` refers to the inner working of `OperatorFusor` which shouldn't be exposed to users. From the user's perspective, what we want to say is "whether or not to keep bound constants in a grouped function". So I think `bind_constants` is more intuitive name (a more verbose option would be "keep_bound_constants"). And the default value for `bind_constants` should be True, since we shouldn't add unnecessary change to the input mod by default.

tests/python/relax/test_codegen_cutlass.py
vinx13: got some errors when using `partition_for_cutlass`, seems there are some issues with pattern registry, will fix and update it in a separate PR

python/tvm/relax/backend/contrib/cutlass.py
```diff
@@ -117,4 +117,6 @@ def partition_for_cutlass(mod):
     """
 
     cutlass_patterns = get_patterns_with_prefix("cutlass")
-    return transform.FuseOpsByPattern(cutlass_patterns, annotate_codegen=True)(mod)
+    return transform.FuseOpsByPattern(cutlass_patterns, lift_constants=True, annotate_codegen=True)(
```
masahi, line 120: needs update here

PR #14155, opened 2023-03-01
[Unity][Analysis] Checking function return struct info in well-formed check

Hzfengsy:
The current well-formed misses the check of function return struct info, which may mistakenly pass the check if there are undefined vars in the function return struct info.


yongwww (approved):
LGTM!

tqchen (approved):

PR #14156, opened 2023-03-01
【PaddlePaddle Hackathon 4】add paddle sin and where op

Asthestarsfalll:
add paddle sin and where op


Asthestarsfalll:
@jiangjiajun Hi
According to Contributor Guide, I used this command to run a single unit test but nothing happended, but it failed. I wonder how can I trigger the tests in paddlepaddle in correct way.
```
TVM_FFI=ctypes python -m pytest -v -k "test_forward_norm" tests/python/frontend/paddlepaddle/test_forward.py
```
36bde4b59926e533368de0aef5c7a675

PR #14162, opened 2023-03-01
[Unity][BYOC] Use Relax legalize + CPU build for reference in tests

masahi:
Currently BYOC tests are using Relay as reference because (1) legalize was not available before or (2) tensorcore result is needed for cutlass. This PR removes the use of Relay in CUTLASS, DNNL, and TRT tests.

For cutlass, I'm using small integer inputs to workaround the accuracy issue.

@tqchen @vinx13 @yelite


tqchen:
One minor nits, for reference data generation, please use memoize to reduce the cost of repetitively running tests locally

https://github.com/apache/tvm/blob/main/tests/python/topi/python/test_topi_conv2d_hwcn.py#L47

masahi:
> One minor nits, for reference data generation, please use memoize to reduce the cost of repetitively running tests locally

Updated DNNL and TRT tests. But cutlass tests are written in a way not compatible with `memoize`. So they are left as is.

vinx13 (approved):

PR #14165, opened 2023-03-01
[TVMC][UMA] Support using UMA with TVMC

rafzi:
This change introduces:

- A `TVMExtension` interface to extend TVMC with custom code. Currently its only use is to add `UMABackend`s.
- A command line option `--experimental-tvm-extension <dir>` to point to extension modules.
- An envvar `$TVM_EXTENSION_DIR` for the same purpose.

See discussion thread: https://discuss.tvm.apache.org/t/tvmc-uma-integration/14309

@Mousius @cgerum @MichaelJKlaiber @PhilippvK

### DIFF
--- a/python/tvm/driver/tvmc/compiler.py
+++ b/python/tvm/driver/tvmc/compiler.py
@@ -49,23 +49,31 @@ from .transform import generate_transform_args, parse_graph_transform_args, appl
 from .shape_parser import parse_shape_string
 from .workspace_pools import generate_workspace_pools_args, workspace_pools_recombobulate
 from .extensions import load_extensions, get_extensions
+from .arguments import TVMCSuppressedArgumentParser
 
 # pylint: disable=invalid-name
 logger = logging.getLogger("TVMC")
 
 
 @register_parser
-def add_compile_parser(subparsers, _, json_params):
+def add_compile_parser(subparsers, main_parser, json_params, argv):
     """Include parser for 'compile' subcommand"""
 
     parser = subparsers.add_parser("compile", help="compile a model.")
     parser.set_defaults(func=drive_compile)
+
     parser.add_argument(
         "--experimental-tvm-extension",
-        default="",
-        help="path from which to load packages named tvm_extension which implement the TVMExtension interface."
+        default=[],
+        action="append",
+        help="path from which to load packages named tvm_extension which implement the TVMExtension interface.",
     )
-    _handle_extensions()
+    disposable_parser = TVMCSuppressedArgumentParser(main_parser)
+    try:
+        known_args, _ = disposable_parser.parse_known_args(argv)
+        _handle_extensions(known_args.experimental_tvm_extension)
+    except TVMCException:
+        pass
 
     parser.add_argument(
         "--cross-compiler",
@@ -189,14 +197,12 @@ def add_compile_parser(subparsers, _, json_params):
     generate_workspace_pools_args(parser)
 
 
-def _handle_extensions():
-    # Need to manually parse this argument so that the parser options of any extension can be generated automatically.
-    extension_paths = []
-    for i, arg in enumerate(sys.argv):
-        if arg == "--experimental-tvm-extension" and i + 1 < len(sys.argv):
-            extension_paths.append(sys.argv[i + 1])
-    load_extensions(extension_paths)
+def _handle_extensions(extra_paths):
+    extension_paths = extra_paths
+    if os.environ.get("TVM_EXTENSION_DIR", None):
+        extension_paths.append(os.environ["TVM_EXTENSION_DIR"])
 
+    load_extensions(extension_paths)
     for ext in get_extensions():
         for uma_backend in ext.uma_backends():
             uma_backend.register()
--- a/python/tvm/driver/tvmc/compiler.py
+++ b/python/tvm/driver/tvmc/compiler.py
@@ -71,9 +71,13 @@ def add_compile_parser(subparsers, main_parser, json_params, argv):
     disposable_parser = TVMCSuppressedArgumentParser(main_parser)
     try:
         known_args, _ = disposable_parser.parse_known_args(argv)
-        _handle_extensions(known_args.experimental_tvm_extension)
     except TVMCException:
         pass
+    try:
+        ext_dirs = known_args.experimental_tvm_extension
+    except AttributeError:
+        ext_dirs = []
+    _handle_extensions(ext_dirs)
 
     parser.add_argument(
         "--cross-compiler",
--- a/python/tvm/driver/tvmc/compiler.py
+++ b/python/tvm/driver/tvmc/compiler.py
@@ -21,7 +21,6 @@ Provides support to compile networks both AOT and JIT.
 import logging
 import os.path
 import re
-import sys
 import itertools
 from copy import deepcopy
 from typing import Any, Optional, Dict, List, Union, Callable, Sequence
@@ -59,14 +58,15 @@ logger = logging.getLogger("TVMC")
 def add_compile_parser(subparsers, main_parser, json_params, argv):
     """Include parser for 'compile' subcommand"""
 
-    parser = subparsers.add_parser("compile", help="compile a model.")
+    parser = subparsers.add_parser("compile", help="compile a model.", add_help=False)
     parser.set_defaults(func=drive_compile)
 
     parser.add_argument(
         "--experimental-tvm-extension",
         default=[],
         action="append",
-        help="path from which to load packages named tvm_extension which implement the TVMExtension interface.",
+        help="path from which to load packages named tvm_extension which implement the "
+        "TVMExtension interface.",
     )
     disposable_parser = TVMCSuppressedArgumentParser(main_parser)
     try:
--- a/python/tvm/driver/tvmc/compiler.py
+++ b/python/tvm/driver/tvmc/compiler.py
@@ -21,6 +21,7 @@ Provides support to compile networks both AOT and JIT.
 import logging
 import os.path
 import re
+import sys
 import itertools
 from copy import deepcopy
 from typing import Any, Optional, Dict, List, Union, Callable, Sequence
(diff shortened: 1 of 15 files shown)


python/tvm/driver/tvmc/compiler.py
```diff
@@ -40,17 +40,38 @@
...
 logger = logging.getLogger("TVMC")
 
 
 @register_parser
-def add_compile_parser(subparsers, _, json_params):
+def add_compile_parser(subparsers, main_parser, json_params, argv):
     """Include parser for 'compile' subcommand"""
 
-    parser = subparsers.add_parser("compile", help="compile a model.")
+    parser = subparsers.add_parser("compile", help="compile a model.", add_help=False)
```
lhutton1, line 54: Curious, why do we need to remove help option?

python/tvm/driver/tvmc/extensions.py
```diff
@@ -0,0 +1,128 @@
...
+                            lineno=warning.lineno,
+                            file=warning.file,
+                            line=warning.line,
+                        )
+
+                if loaded_mod is not None:
+                    for _name, obj in inspect.getmembers(loaded_mod):
+                        if _is_concrete_extension_type(obj):
+                            scanned_extensions.append(obj)
+            except ImportError as err:
```
lhutton1, line 116: would be good to add a test for this case

tests/python/contrib/test_uma/test_tvmc.py
```diff
@@ -0,0 +1,58 @@
...
+
+    extension_dir = os.path.join(os.path.dirname(os.path.realpath(__file__)), "vanilla_ext")
+    compile_str = (
+        f"tvmc compile --target vanilla_accelerator,c -f mlf "
+        f"--experimental-tvm-extension={extension_dir} "
+        f"--desired-layout NCHW "
+        f"--output={package_path} {model_path}"
+    )
+    compile_args = compile_str.split(" ")[1:]
+    assert _main(compile_args) == 0
```
lhutton1, line 54: I think it would be good to check the resulting graph was partitioned correctly via `vanilla_accelerator`, it should be possible by dumping the relay with `--dump_code=relay` and checking for "vanilla_accelerator" in the output, WDYT?

python/tvm/driver/tvmc/extensions.py
```diff
@@ -0,0 +1,128 @@
...
+        for importer, modname, _ in pkgutil.walk_packages(
+            path=mdl.__path__, prefix=mdl.__name__ + ".", onerror=lambda x: None
+        ):
+            try:
+                module_name = modname.rsplit(".", 1)[-1]
+                # If module's name starts with "_", do not load the module.
+                # But if the module's name starts with a "__", then load the
+                # module.
+                if module_name.startswith("_") and not module_name.startswith("__"):
+                    continue
```
lhutton1, line 83: Curious, what's the reason for this?

lhutton1 (commented):
This looks really nice, thanks @rafzi :) I had a couple of questions and some suggestions for tests

python/tvm/driver/tvmc/extensions.py
cgerum: Not scanning in '_' allows to split interface and implementation:

`my_backend.py`:

```
class MyBackendExtension(TVMExtension):
    def __init__(self):
        from ._impl import MyBacken
        self.backend = MyBackend()

    def uma_backends(self):
        return [self.backend]
```


`_my_backend_impl.py`:

```
class MyBackend(UMABackend):
   ....
```

The exception for "__" is there to allow declaring extensions in `__init__.py` and `__main__.py` but I am not sure wether this is actually worth it. Or we should just scan all modules for now.

cgerum (commented):
Thanks @rafzi for the proposal, thanks @lhutton1 for the feedback. 

I also have a few minor requests.

python/tvm/driver/tvmc/extensions.py
rafzi: I tried adding this, but encountered an issue with race conditions. The way the `sys.path` is currently modified by extensions.py is problematic when more than one test is run in parallel. Is there already an approach for serializing tests in TVM? There doesn't seem to be an upstream solution: https://github.com/pytest-dev/pytest-xdist/issues/18

python/tvm/driver/tvmc/compiler.py
rafzi: I encountered an issue where the TVMCSuppressedArgumentParser still caused a SystemExit exception. However, I cannot seem to reproduce the issue right now. So I'll remove this again.

python/tvm/driver/tvmc/extensions.py
```diff
@@ -0,0 +1,128 @@
...
+import pkgutil
+import warnings
+import copy
+from abc import abstractmethod
+
+
+_EXTENSIONS = []
+
+
+class TVMExtension(object):
```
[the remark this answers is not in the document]
rafzi: I agree that TVMC extension makes more sense. I'll change it here, in the cmdline arg and envvar.

python/tvm/driver/tvmc/compiler.py
rafzi: Alright, I found the input again: "tvmc compile --help". When removing the help option from the subparser, this is resolved. The help option still works because it is enabled through the main parser.

python/tvm/driver/tvmc/extensions.py
lhutton1: Could we instead avoid modifying `sys.path` when loading a module? I had a naive search and wondered if [this]( https://stackoverflow.com/questions/67631/how-can-i-import-a-module-dynamically-given-the-full-path ) could help? cc @Mousius @leandron who might know more about these things

python/tvm/driver/tvmc/compiler.py
lhutton1: I tried this out locally but see some difference in the help prompt when executing `tvmc compile --help`. Before the patch we get something like:
```
$ tvmc compile --help
usage: tvmc compile [-h] [--cross-compiler CROSS_COMPILER]
                    [--cross-compiler-options CROSS_COMPILER_OPTIONS]
                    [--desired-layout {NCHW,NHWC}] [--dump-code FORMAT]
                    [--model-format {keras,onnx,pb,tflite,pytorch,paddle,relay}]
                    [-o OUTPUT] [-f {so,mlf}] [--pass-config name=value]
                    [--target TARGET]
...
positional arguments:
  FILE                  path to the input model file.

optional arguments:
  -h, --help            show this help message and exit
  --cross-compiler CROSS_COMPILER
                        the cross compiler to generate target libraries, e.g.
                        'aarch64-linux-gnu-gcc'.
  --cross-compiler-options CROSS_COMPILER_OPTIONS
                        the cross compiler options to generate target
                        libraries, e.g. '-mfpu=neon-vfpv4'.
  --desired-layout {NCHW,NHWC}
                        change the data layout of the whole graph.
```
While after seems to lead to:
```
$ tvmc compile --help
usage: tvmc compile [--experimental-tvm-extension EXPERIMENTAL_TVM_EXTENSION]
                    [--cross-compiler CROSS_COMPILER]
                    [--cross-compiler-options CROSS_COMPILER_OPTIONS]
                    [--desired-layout {NCHW,NHWC}] [--dump-code FORMAT]
                    [--model-format {keras,onnx,pb,tflite,pytorch,paddle,relay}]
                    [-o OUTPUT] [-f {so,mlf}] [--pass-config name=value]
                    [--target TARGET]
...
tvmc compile: error: the following arguments are required: FILE
```
I think it would be really helpful if we can keep the argument descriptions as before

MichaelJKlaiber (approved):
I agreed with @lhutton1 and @cgerum 's comments. 
Besides that everything LGTM.
Thanks @rafzi for the contribution

python/tvm/driver/tvmc/compiler.py
rafzi: Alright, I can confirm this issue. The underlying issue is that the `TVMCSuppressedArgumentParser` still throws `SystemExit` when given such a command line. We also cannot override `error` of `ArgumentParser`, because the failing parser is the subparser, which is not a `TVMCSuppressedArgumentParser`.

Before this patch, this problem already occurs. For example with `tvmc run --help`, `SystemExit` is thrown and a partial help message is printed (missing `--project-option`). So my suggested fix is to move the `parse_known_args` as far down as possible to yield a mostly complete help message.

python/tvm/driver/tvmc/extensions.py
rafzi: While such an approach should work for single source files, I'm not sure how that would work for packages or namespace packages.

tests/python/contrib/test_uma/test_tvmc.py
rafzi: yes, definitely!

python/tvm/relay/backend/contrib/uma/api/lower.py
```diff
@@ -71,9 +71,7 @@ def _get_tensors(te_cached_func):
         )
 
         compiler_attr = relay_prim_func.attrs["Compiler"]
-        target = tvm.target.Target.current()
-        if target.kind.name != compiler_attr:
-            target = tvm.target.Target(compiler_attr)
+        target = tvm.target.Target(compiler_attr)
```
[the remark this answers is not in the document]
rafzi: Thanks for clarifying! I implemented your suggestion.

python/tvm/driver/tvmc/compiler.py
lhutton1: Thanks for investigating! This looks much better, so I won't block on it. Perhaps we can create a separate issue to track this bug?

python/tvm/driver/tvmc/extensions.py
lhutton1: Apologies for the delay, I had a think about this but wasn't able to come up with a better solution myself, so I won't block on this

PhilippvK:
@tvm-bot rerun

rafzi:
Sorry for the late update.

The outstanding test failure is likely because of the outstanding issue that the sys.path is modified in order to support all module loading. This probably inteferres with running the tests in parallel.

https://github.com/tum-ei-eda/tvm/blob/c1c2d3ca2db1c56d3cc8d695530ab7aa8b90ea00/python/tvm/driver/tvmc/extensions.py#L24-L34

I'm not sure how to resolve this as I did not find another method to achieve the same kind of module loading. Do we need to compromise here and settle for a simple module load? I unresolved the relevant discussion with @lhutton1.

@cgerum @PhilippvK

PR #14166, opened 2023-03-01
[Unity][BYOC] Add batch matmul support to Relax CUTLASS BYOC

yelite:
This PR adds the support of batch matmul to Relax CUTLASS BYOC. This includes batch matmul with shape 2D x ND, ND x 2D and ND x ND. Additionally, this supports bias, transposed y and all supported epilogues from regular matmul.

This PR also includes:
1. Add check function to all APIs related to PatternRegistryEntry (a follow up of #14109)
2. Add check function to CUTLASS pattern table, blocking unsupported matmul from being partitioned.
3. Reorganize the tests related to CUTLASS matmul.

cc @vinx13 @masahi

### DIFF
--- a/python/tvm/contrib/cutlass/gen_tensor_op.py
+++ b/python/tvm/contrib/cutlass/gen_tensor_op.py
@@ -529,8 +529,7 @@ def instantiate_template(func_name, annotations, func_args):
         return dim1 + " * " + dim2
 
     if "dense" in func_name or "matmul" in func_name:
-        batched = "batch_matmul" in func_name
-        batched_offset = 1 if batched else 0
+        batched = "batch" in annotations
         transposed = "transposed" in func_name
         lhs_arg_idx = _get_optional_int_annotation(annotations, "lhs_arg_idx", 0)
         rhs_arg_idx = _get_optional_int_annotation(annotations, "rhs_arg_idx", 1)
@@ -539,6 +538,8 @@ def instantiate_template(func_name, annotations, func_args):
         rhs_arg = func_args[rhs_arg_idx]
         lhs_shape = annotations[f"arg{lhs_arg_idx}_shape"]
         rhs_shape = annotations[f"arg{rhs_arg_idx}_shape"]
+        lhs_batched_offset = len(lhs_shape) - 2
+        rhs_batched_offset = len(rhs_shape) - 2
 
         attrs["lhs_arg"] = lhs_arg
         attrs["rhs_arg"] = rhs_arg
@@ -548,17 +549,20 @@ def instantiate_template(func_name, annotations, func_args):
         attrs["ElementInputB"] = DataTypeTag[dtype_map[annotations[f"arg{rhs_arg_idx}_dtype"]]]
         attrs["ElementOutput"] = DataTypeTag[dtype_map[annotations["ret_dtype"]]]
 
-        attrs["K"] = str(int(lhs_shape[batched_offset + 1]))
-        attrs["M"] = get_dim(lhs_shape[batched_offset], lhs_arg, 0, batched_offset)
+        attrs["K"] = str(int(lhs_shape[lhs_batched_offset + 1]))
+        attrs["M"] = get_dim(lhs_shape[lhs_batched_offset], lhs_arg, 0, lhs_batched_offset)
 
         if transposed:
-            attrs["N"] = get_dim(rhs_shape[batched_offset], rhs_arg, 0, batched_offset)
+            attrs["N"] = get_dim(rhs_shape[rhs_batched_offset], rhs_arg, 0, rhs_batched_offset)
         else:
-            attrs["N"] = get_dim(rhs_shape[batched_offset + 1], rhs_arg, 1, batched_offset)
+            attrs["N"] = get_dim(rhs_shape[rhs_batched_offset + 1], rhs_arg, 1, rhs_batched_offset)
 
         if batched:
             headers.append("cutlass/gemm/device/gemm_batched.h")
-            attrs["batch"] = get_dim(lhs_shape[0], lhs_arg, 0)
+            # TODO: Support dynamic shape
+            # With the support of more general broadcasting in batch matmul,
+            # it needs more sophisticated code to handle dynamic shape
+            attrs["batch"] = str(int(annotations["batch"]))
             attrs["batch_stride_A"] = get_batch_stride(
                 annotations["batch_stride_A"],
                 lhs_arg_idx,
--- a/python/tvm/contrib/cutlass/gen_tensor_op.py
+++ b/python/tvm/contrib/cutlass/gen_tensor_op.py
@@ -559,11 +559,7 @@ def instantiate_template(func_name, annotations, func_args):
 
         if batched:
             headers.append("cutlass/gemm/device/gemm_batched.h")
-            # TODO: Support dynamic shape
-            # With the support of more general broadcasting in batch matmul,
-            # like matmul with shape (a, b, c, 5, 6) x (6, 8),
-            # it needs more sophisticated code to handle dynamic shape
-            attrs["batch"] = str(int(annotations["batch"]))
+            attrs["batch"] = get_dim(annotations["batch"], lhs_arg, 0)
             attrs["batch_stride_A"] = get_batch_stride(
                 annotations["batch_stride_A"],
                 lhs_arg_idx,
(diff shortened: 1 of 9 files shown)


tests/python/relax/test_codegen_cutlass.py
```diff
@@ -255,13 +177,48 @@ def test_conv2d_offload():
...
+
+    out = get_result_with_relax_cutlass_offload(Conv2dx2, data_np, weight1_np, weight2_np)
+
+    relay_expr = get_relay_conv2d_relu_x2(data_np.shape, weight1_np.shape)
+    ref = get_relay_ref(relay_expr, data_np, weight1_np, weight2_np)
+
+    tvm.testing.assert_allclose(out, ref, rtol=1e-5, atol=1e-5)
+
+
+def get_reference_matmul_result(x, y, bias=None, transposed_y=False, activation=None):
```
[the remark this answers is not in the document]
masahi: Were you able to pass accuracy check with this function?

src/relax/transform/fuse_ops.cc
```diff
@@ -900,7 +900,7 @@ class PatternBasedPartitioner : ExprVisitor {
   using Group = GraphPartitioner::Group;
   using GroupMap = OperatorFusor::GroupMap;
   using ExprVisitor::VisitExpr_;
-  using FCheckMatch = backend::PatternRegistryEntryNode::FCheckMatch;
+  using FCheckMatch = runtime::TypedPackedFunc<bool(const Map<DFPattern, Expr>&, const Expr&)>;
```
vinx13, line 903: we can remove https://github.com/apache/tvm/blob/03c5f998b37afe41257e2c37b1e95e0ef5357ce8/src/relax/transform/fuse_ops.cc#L43 if it's no longer needed

python/tvm/relax/backend/contrib/cutlass.py
```diff
@@ -17,84 +17,132 @@
...
+        raise ValueError("Cannot find call to matmul from match_result.")
+
+    lhs_shape = _get_static_shape(matmul_call.args[0].struct_info.shape)
+    rhs_shape = _get_static_shape(matmul_call.args[1].struct_info.shape)
+    if len(lhs_shape) < 2 or len(rhs_shape) < 2:
+        return False
+
+    lhs_dtype = matmul_call.args[0].struct_info.dtype
+    rhs_dtype = matmul_call.args[1].struct_info.dtype
+    if lhs_dtype != "float16" or rhs_dtype != "float16":
```
vinx13, line 59: fp32 is also supported (if not, let's revisit them later)

tests/python/relax/test_codegen_cutlass.py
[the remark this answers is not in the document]
yelite: Yes, all tests in this file pass. The tolerance it uses is 'rtol=1e-2, atol=1e-3'.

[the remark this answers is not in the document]
masahi: Nice, I changed the matmul tests to use integer inputs in https://github.com/apache/tvm/pull/14162, that might not be needed.

python/tvm/relax/backend/contrib/cutlass.py
yelite: I updated the dtype check to match what we have in Relay CUTLASS BYOC.
```
        (lhs_dtype == "float16" and rhs_dtype == "float16")
        or (lhs_dtype == "float32" and rhs_dtype == "float32")
        or (lhs_dtype in ("int8", "uint8") and rhs_dtype in ("int8", "uint8"))
```

python/tvm/contrib/cutlass/gen_tensor_op.py
```diff
@@ -548,17 +549,20 @@ def get_batch_stride(stride_annot, arg0_idx, arg1_idx, arg0_axis_idx, arg1_axis_
...
-            attrs["N"] = get_dim(rhs_shape[batched_offset + 1], rhs_arg, 1, batched_offset)
+            attrs["N"] = get_dim(rhs_shape[rhs_batched_offset + 1], rhs_arg, 1, rhs_batched_offset)
 
         if batched:
             headers.append("cutlass/gemm/device/gemm_batched.h")
-            attrs["batch"] = get_dim(lhs_shape[0], lhs_arg, 0)
+            # TODO: Support dynamic shape
+            # With the support of more general broadcasting in batch matmul,
+            # it needs more sophisticated code to handle dynamic shape
+            attrs["batch"] = str(int(annotations["batch"]))
```
masahi, line 565: What do you mean by `more general broadcasting`? Like the conceptual "batch" dimension spanning multiple axes?

tests/python/relax/test_codegen_cutlass.py
```diff
@@ -120,12 +119,27 @@ def test_conv2d_offload():
...
 
+def test_kernel_sharing():
+    data_np = np.random.randn(16, 32, 32, 16).astype("float16")
+    weight1_np = np.random.randn(16, 3, 3, 16).astype("float16")
+    weight2_np = np.random.randn(16, 3, 3, 16).astype("float16")
+
+    out = get_result_with_relax_cutlass_offload(Conv2dx2, data_np, weight1_np, weight2_np)
+
+    relay_expr = get_relay_conv2d_relu_x2(data_np.shape, weight1_np.shape)
+    ref = get_relay_ref(relay_expr, data_np, weight1_np, weight2_np)
```
masahi, line 130: This diff seems like a merge bug

python/tvm/relax/backend/pattern_registry.py
```diff
@@ -46,25 +47,59 @@ class PatternRegistryEntry(Object):
...
+
+
+_CLEANUP_REGISTERED = False
+
+
+def _ensure_cleanup_function_registered():
+    """
+    Add a cleanup function to be called on interpreter termination, to remove all
+    patterns registered on the Python side. Without cleaning up those patterns,
+    program will segfault on termination. It's because the check functiosn of pattern
```
vinx13, line 86:
```suggestion
    program will segfault on termination. It's because the check functions of pattern
```

python/tvm/relax/backend/contrib/cutlass.py
```diff
@@ -116,7 +173,6 @@ def partition_for_cutlass(mod):
         compiled by the CUTLASS backend.
     """
 
-    cutlass_patterns = get_patterns_with_prefix("cutlass")
-    return transform.FuseOpsByPattern(cutlass_patterns, bind_constants=True, annotate_codegen=True)(
-        mod
-    )
+    cutlass_pattern_entries = get_patterns_with_prefix("cutlass")
+    patterns = [(e.name, e.pattern, e.check) for e in cutlass_pattern_entries]
+    return transform.FuseOpsByPattern(patterns, bind_constants=True, annotate_codegen=True)(mod)
```
vinx13, line 178:
```suggestion
    return transform.FuseOpsByPattern(patterns, bind_constants=False, annotate_codegen=True)(mod)
```
I made a typo in the previous PR, it should be False

python/tvm/relax/backend/pattern_registry.py
vinx13: this seems a general issue of ffi
cc @tqchen @junrushao

python/tvm/relax/backend/contrib/cutlass.py
```diff
@@ -17,84 +17,141 @@
...
+    )
+
+
+def _check_matmul(
+    match_result: Mapping[DFPattern, Expr],
+    _: Expr,
+) -> bool:
+    matmul_call: Call = None
+    for _, expr in match_result.items():
+        if isinstance(expr, Call) and expr.op.name == "relax.matmul":
```
vinx13, line 56:
```suggestion
        if isinstance(expr, Call) and isinstance(expr.op, tvm.ir.Op) and expr.op.name == "relax.matmul":
```
it's possible it's calling a PrimFunc, it will be a global var

vinx13: The problem here is it will iterate over all nodes in the pattern, including wildcard that matches anything. This can actually be a problem when there are two consecutive matmuls, if the first matmul is not offloaded (e.g doesn’t pass check func), and when we are checking the match for the second matmul, the match result will have two matmuls because the wildcard matches the first matmul

python/tvm/contrib/cutlass/gen_tensor_op.py
yelite: Yes, for example, matmul with shape (a, b, c, 5, 5) x (5, 5) has batch `a * b * c`. I reverted the change on this line. Right now the check function in the cutlass pattern table requires static shape anyway. I will create a followup PR for formal dynamic shape support in cutlass matmul.

tests/python/relax/test_codegen_cutlass.py
yelite: Fixed

masahi (approved):

python/tvm/relax/backend/contrib/cutlass.py
yelite: I will fix this in my follow up PR about dynamic shape support.

PR #14176, opened 2023-03-02
[WIP][Unity][TVMScript] return from IfNode

yongwww:
I am trying to add return support from true/false branch of `IfNode` as we discussed in https://github.com/tlc-pack/relax/issues/446. 

I introduced a `RelaxReturnGlobalInfo` (similar to `DummyGlobalInfo`), which is used to save all the `Expr` (`body` of `SeqExpr` ) that need to be returned. But this causes it only works at IR_Module level, it failed to work for relax function without explicitly defining in ir_module. Options I have at this moment.

- o0: Leverage metadata to save all the return Exprs.
- o1: Introduce GlobalInfo in relax Function level, a new member like `global_info` will be added into FunctionNode
- o2: Add a new member tag `return_body` into `SeqExpr` (we don't  want to change IR at this stage, list it here for consideration in the future)

Any comments are truly appreciated!  cc: @slyubomirsky  @tqchen  @junrushao @Hzfengsy @YuchenJin

---
update:  If a pass modifies the return Expr (for example the original return expr is a CallNode, a var is created and bound to the CallNode in a pass, thus the var is the newly generated return expr), then the globally maintained return exprs should be updated in the pass as well. Looks it burdens the pass writing.


yongwww:
I close this PR since it should be fine to ask user to return in the end of the function at this moment, I can reopen and keep working on this if needed.

PR #14179, opened 2023-03-02
[CRT][microTVM]Use AoTExecutor by default for CRT runtime

mehrdadh:
This PR changes default `executor` to `None` at `relay.build` and set the default inside the function based on the runtime. The default executor is set to AOTExecutor if runtime=CRT, otherwise it is set to GraphExecutor.

### DIFF
--- a/apps/bundle_deploy/build_model.py
+++ b/apps/bundle_deploy/build_model.py
@@ -18,14 +18,12 @@
 
 import argparse
 import os
-import pathlib
 
 from tvm import relay
 import tvm
 from tvm import runtime as tvm_runtime
 import logging
-from tvm.relay.backend import Runtime
-from tvm.contrib import cc as _cc
+from tvm.relay.backend import Runtime, Executor
 
 RUNTIMES = [
     (Runtime("crt", {"system-lib": True}), "{name}_c.{ext}"),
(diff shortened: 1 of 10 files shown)


Mousius (changes requested):
This doesn't make sense to me, the executor decides the interface you use to interact with the module not the runtime - this creates scenarios where changing the runtime, which should be behind the scenes, changes the user interface to the module.  Can it be if executor is AOT then set C runtime by default?

areusch:
@Mousius i think it's two angles to the same problem: GraphExecutor isn't a great choice with the C runtime. However, you can in practice use AOT with the C++ runtime. I would think in terms of which compiler config is most obvious to an embedded user, the order would be:
1. the CPU setting
2. the runtime
3. the executor

since the executor is essentially an implementation detail to new users (they are already expecting to be shown how to interact with TVM and wouldn't know that a priori; whereas they would certainly know they can't link a C++ malloc-laden runtime and they would expect to need to tell the compiler to target their particular class of CPUs). so in this sense it seems like it's better to auto-select executor based on runtime. thoughts?

Mousius:
As with other compilers, ahead of time compilation can be applied to a much broader scope than just embedded and the graph executor does work in more constrained environments, is there any reason to limit AoT to just embedded scenarios? 

Considering a broader set of users, it makes sense to work from the interface, to the backend runtime we provide. If you have a user who starts out using the Graph executor, then decides to try using the C runtime, they'll suddenly find they've got an unwanted change in their application interface; as an end user, I would consider this unexpected behaviour and consider raising a bug given it impacted my application in a weird way. 

Having to link a different runtime by default (AOT infers C runtime) is also slightly irritating, but far less strange than a runtime change causing a users application to suddenly require changes.

As an aside, the goal of reducing the number of arguments required is a good one, do we know which functions in the C++ runtime can't be implemented in the C runtime? If the C++ runtime was just a set of C wrappers the user wouldn't even need to choose?

PR #14180, opened 2023-03-02
[Unity][Analysis] Restore Python bindings for var analyses

slyubomirsky:
This PR restores the Python bindings for several analyses related to variables, originally done in https://github.com/tlc-pack/relax/pull/265. These had been implemented in C++ but not exposed in Python.

Question: Do we also want to restore the `CalledGlobalVars` analysis? I notice the C++ implementation for that one has been removed and I am wondering if there was a reason for the removal.


yongwww (approved):
Thanks for adding this!  Looks good to me!

Let's ignore `CalledGlobalVars`, it is not used anymore.

tqchen (approved):

PR #14182, opened 2023-03-02
[Unity] Introduce Default GPU Schedule Pass

zxybazh:
This PR introduces a default schedule pass that creates default thread binding for PrimFuncs, including symbolic shape functions so that these PrimFuncs could be built and run on Cuda device. The pass picks out all the blocks inside the PrimFunc and does loop fusion, split and reorder based on the loop extent and target information (max thread block number and max thread per block).

Co-authored-by: Josh Fromm  <<jwfromm@octoml.ai>>
CC: @sunggg @vinx13 @tqchen


src/relax/transform/default_schedule.cc
```diff
@@ -0,0 +1,143 @@
...
+          } else {
+            Array<tir::LoopRV> splits = sch->Split(
+                fused, /*factors=*/{NullOpt, Integer(std::min(product, max_thread_per_block))});
+            sch->Bind(splits[0], "blockIdx.x");
+            sch->Bind(splits[1], "threadIdx.x");
+          }
+        }
+        mutator.builder_->AddFunction(sch->mod()->Lookup(gv->name_hint), gv->name_hint);
+      } else {
+        mutator.builder_->AddFunction(func, gv->name_hint);
```
vinx13, line 102: not needed since the function is already in the module and it's not changed

src/relax/transform/default_schedule.cc
```diff
@@ -0,0 +1,143 @@
...
+#include "../../tir/ir/functor_common.h"
+
+namespace tvm {
+namespace relax {
+
+/*!
+ * \brief The helper class to schedule functions and build a new module which calls the new TIR
+ * function.
+ */
+class ThreadBindMutator : public ExprMutator {
```
vinx13, line 38: since we are not visiting the function bodies probably it's not necessary to use a `ExprMutator`, we can directly update the module using `IRModule::Add`

src/relax/transform/default_schedule.cc
```diff
@@ -0,0 +1,143 @@
...
+ public:
+  static IRModule Transform(const IRModule& mod, int64_t max_thread_per_block) {
+    ThreadBindMutator mutator(mod);
+
+    for (const auto& kv : mod->functions) {
+      const GlobalVar& gv = kv.first;
+      const BaseFunc& func = kv.second;
+
+      if (func->IsInstance<tir::PrimFuncNode>()) {
+        IRModule mod = IRModule(Map<GlobalVar, BaseFunc>({kv}));
```
vinx13, line 48: it is possible to skip creating a new module if we use `Schedule::WorkOn` to directly schedule the original module (after copy on write), see https://discuss.tvm.apache.org/t/manual-scheduling-of-call-tir-functions-in-relax/14446/7?u=vinx13

src/meta_schedule/utils.h
```diff
@@ -554,6 +554,68 @@ inline double Sum(const Array<FloatImm>& arr) {
...
+    return BlockCollector(sch, f_block_filter).Run();
+  }
+
+ private:
+  /*! \brief Entry point */
+  Array<tir::BlockRV> Run() {
+    std::vector<tir::BlockRV> results;
+    for (const auto& kv : sch_->mod()->functions) {
+      const GlobalVar& gv = kv.first;         // `gv->name_hint` is the name of the function
+      const BaseFunc& base_func = kv.second;  // this can be PrimFunc or relay::Function
```
masahi, line 571: can use

```
for (const auto& [gv, base_func]: ...
```

src/relax/transform/default_schedule.cc
psrivas2: +1 we can skip using ExprMutator for this and directly use Schedule APIs to modify the PrimFunc, and then return the module with updated PrimFunc.

src/relax/transform/default_schedule.cc
```diff
@@ -0,0 +1,143 @@
...
+        mutator.builder_->AddFunction(sch->mod()->Lookup(gv->name_hint), gv->name_hint);
+      } else {
+        mutator.builder_->AddFunction(func, gv->name_hint);
+      }
+    }
+    return mutator.builder_->GetContextIRModule();
+  }
+
+ private:
+  explicit ThreadBindMutator(const IRModule& mod) : mod_(mod) {}
```
psrivas2, line 109: Calling `ExprMutator` constructor with module i.e.,`ExprMutator(mod)` here would populate the builder context IRModule with old IRModule. So we won't have to call `builder_->AddFunction` (line 102) for functions that are not modified at all, and for others we can use `builder_->UpdateFunction` (line 100).

src/relax/transform/default_schedule.cc
```diff
@@ -0,0 +1,143 @@
...
+/*!
+ * \brief The helper class to schedule functions and build a new module which calls the new TIR
+ * function.
+ */
+class ThreadBindMutator : public ExprMutator {
+ public:
+  static IRModule Transform(const IRModule& mod, int64_t max_thread_per_block) {
+    ThreadBindMutator mutator(mod);
+
+    for (const auto& kv : mod->functions) {
```
psrivas2, line 43: nit: we can use `const auto& [gv, func]: ...` to avoid`kv.first` and `kv.second` statements

psrivas2 (changes requested):
Thanks @zxybazh!

src/relax/transform/default_schedule.cc
```diff
@@ -0,0 +1,145 @@
...
+        }
+        return relax::DefaultSchedule(m, max_thread_per_block.IntValue());
+      };
+  return CreateModulePass(/*pass_function=*/pass_func,      //
+                          /*opt_level=*/0,                  //
+                          /*pass_name=*/"DefaultSchedule",  //
+                          /*required=*/{});
+}
+
+TVM_REGISTER_GLOBAL("relax.transform.DefaultSchedule").set_body_typed(DefaultSchedule);
```
yongwww, line 140: I feel it would be better to move it into `tir.transform.DefaultSchedule` (src/tir/transform), looks only PrimFunc is updated

tqchen:
cc @junrushao

vinx13 (approved):

include/tvm/tir/transform.h
```diff
@@ -709,6 +709,12 @@ TVM_DLL Pass ManifestSharedMemoryLocalStage();
  */
 TVM_DLL Pass InstrumentProfileIntrinsics();
 
+/*!
+ * \brief Create default schedule for PrimFuncs to run on cuda device.
+ * \return The Pass.
+ */
+TVM_DLL Pass DefaultSchedule();
```
jwfromm, line 716: Should we call it `DefaultCudaSchedule` or `DefaultGPUSchedule` to indicate its specifically for thread bindings?

zxybazh: I'm open to naming options, `DefaultCudaSchedule` sounds like a better version given it's only targeting cuda for now.

jwfromm: I do think that would be a better name, `DefaultSchedule~ makes it sound like it would be useful for cpu too.

tqchen:
cc @junrushao

yongwww (approved):

include/tvm/tir/transform.h
MasterJH5574: Though this pass is intended only for CUDA backend at this moment, I suppose it will actually work for some other GPU backends, as the tasks tuned by MetaSchedule on these backends are all following the same schedule rule.
https://github.com/apache/tvm/blob/main/src/meta_schedule/utils.h#L512

So I personally would prefer the name of “DefaultGPUSchedule” and list the GPU backends this pass supports in the doc string. In the pass we have already checked if the target contains attribute “maximum # of threads per block.” We can fall back with no change if the target doesn’t have this attribute.

python/tvm/tir/transform/transform.py
```diff
@@ -1040,3 +1040,13 @@ def InstallDebugSpans():
         The result pass
     """
     return _ffi_api.InstallDebugSpans()  # type: ignore
+
+
+def DefaultSchedule():
+    """Default schedule for PrimFuncs.
```
MasterJH5574, line 1046: I’m thinking about the performance. If we don’t intend to guarantee a decent performance, we probably should state in the documents that this pass only intends to make the IRModule able to be built, and has no guarantee on the performance.

BTW could we give the pass more documents? The PR description looks very good to serve as document.

src/tir/transforms/default_schedule.cc
```diff
@@ -0,0 +1,125 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+#include <tvm/tir/schedule/schedule.h>
+#include <tvm/tir/stmt_functor.h>
+
+#include "../../meta_schedule/utils.h"
+#include "../../relay/analysis/graph_partitioner.h"
+#include "../../support/arena.h"
+#include "../../tir/ir/functor_common.h"
```
MasterJH5574, line 25: These headers are redundant. They seem to come from the FuseOps pass.

tqchen:
please run the following commend to update to latest change
```
git rebase --onto upstream/unity upstream/unity-rebase-backup-2023-03-05 
```

include/tvm/tir/transform.h
tqchen: I agree DefaultGPUSchedule seems to be a good name here

python/tvm/tir/transform/transform.py
zxybazh: Thanks for the advice! Added more document according to PR description and the intention of this PR.

include/tvm/tir/transform.h
zxybazh: It appears that there is consensus among the contributors that the current name DefaultSchedule is not clear enough, and many prefer the new name DefaultGPUSchedule as it allows for potential expansion to other GPU targets in the future. As a result, I have renamed it to DefaultGPUSchedule for the time being, but I am still open to other suggestions for naming.

src/tir/transforms/default_gpu_schedule.cc
```diff
@@ -0,0 +1,116 @@
...
+ * under the License.
+ */
+
+#include "../../meta_schedule/utils.h"
+
+namespace tvm {
+namespace tir {
+namespace transform {
+/*!
+ * \brief A helper class to do default thread binding for a block.
```
psrivas2, line 26: nit: s/class/function

jwfromm:
Thanks for this excellent PR @zxybazh! This is now merged.

PR #14183, opened 2023-03-03
[Unity] Introduce call_dps_packed

yongwww:
Introduce call_dps_packed to call packed function instead tir via destination passing convention


src/relax/op/op.cc
```diff
@@ -106,6 +106,44 @@ Expr MakeCallTIR(Expr func, Tuple args, Array<TensorStructInfo> out_sinfo_list,
...
+  return call->sinfo_args[0];
+}
+
+RELAY_REGISTER_OP("relax.call_dps_packed")
+    .set_num_inputs(2)
+    .add_argument("func", "Expr", "The destination-passing-style function.")
+    .add_argument("args", "Tuple", "The input arguments.")
+    .set_attr<FInferStructInfo>("FInferStructInfo", InferStructInfoCallDPSPacked);
+
+Expr MakeCallDPSPacked(Expr func, Tuple args, Array<TensorStructInfo> out_sinfo_list) {
```
slyubomirsky, line 125: Why are the output StructInfo required to be tensors?

tests/python/relax/test_ast_printer.py
```diff
@@ -432,7 +432,7 @@ def test_call_tir():
     @R.function
     def foo(x: R.Tensor(("m", "n"), "float32")):
         m, n = T.var("int64"), T.var("int64")
-        gv0 = R.call_tir("test.op.identity", (x,), R.Tensor((m, n), dtype="float32"))
+        gv0 = R.call_dps_packed("test.op.identity", (x,), R.Tensor((m, n), dtype="float32"))
```
slyubomirsky, line 435: Maybe we should just drop this test case if it's not going to actually use `call_tir`

slyubomirsky:
Do we have any way to enforce that the function passed to `call_tir` is a `PrimFunc`? I believe making `call_tir` more specific was discussed before. (Should we still permit passing a `PrimFunc` to `call_dps_packed`?)

psrivas2:
> Do we have any way to enforce that the function passed to `call_tir` is a `PrimFunc`? I believe making `call_tir` more specific was discussed before. (Should we still permit passing a `PrimFunc` to `call_dps_packed`?)

We can enforce it in `MakeCallTIR` and `MakeCallDPSPacked`. 
For now I don't see a reason to allow `call_dps_packed` to call `Primfunc`.
cc @tqchen

src/relax/op/op.cc
```diff
@@ -106,6 +106,44 @@ Expr MakeCallTIR(Expr func, Tuple args, Array<TensorStructInfo> out_sinfo_list,
...
+  for (const TensorStructInfo& sinfo : out_sinfo_list) {
```
psrivas2, line 126: This has a lot of common code with MakeCallTIR, can we avoid code duplication here?

slyubomirsky:
> > Do we have any way to enforce that the function passed to `call_tir` is a `PrimFunc`? I believe making `call_tir` more specific was discussed before. (Should we still permit passing a `PrimFunc` to `call_dps_packed`?)
> 
> We can enforce it in `MakeCallTIR` and `MakeCallDPSPacked`. For now I don't see a reason to allow `call_dps_packed` to call `Primfunc`. cc @tqchen

It would have to be in the implementation as well, since you can do something like assign a TIR function to a local var and pass that local var elsewhere. In a higher-order function, we wouldn't be able to detect that at compile time unless we make TIR functions part of the type system (though that could be an option).

include/tvm/relax/transform.h
```diff
@@ -89,6 +89,13 @@ TVM_DLL Pass ToNonDataflow();
  */
 TVM_DLL Pass CallTIRRewrite();
 
+/*!
+ * \brief Perform explicit tensor allocation for call_dps_packed.
+ *
+ * \return The Pass.
+ */
+TVM_DLL Pass CallDPSPackedRewrite();
```
MasterJH5574, line 97: I’m wondering if we need this pass. I suppose that the lowering of `call_tir` and `call_dps_packed` are the same (the diff check  also suggests this). The reason we want to introduce two is to reduce confusion. In terms of their lowering, I suppose they can directly share a same logic, which makes things easy (say when someone wants to change CallTIRRewrite, it won’t need to change CallDPSPackedRewrite in the same way as well).

tqchen:
We can have wellformess to enforce such invariant

tqchen:
please run the following commend to update to latest change
```
git rebase --onto upstream/unity upstream/unity-rebase-backup-2023-03-05 
```

include/tvm/relax/transform.h
slyubomirsky: It might be fine to make them syntactic sugar for the same underlying operator. There are only two differences I can think of:
1. `call_tir` is generally assumed to be pure, whereas `call_dps_packed` may not be (unless we want to require it)
2. `call_tir` expects a `PrimFunc` as its operand, whereas `call_dps_packed` expects a `PackedFunc`.

slyubomirsky:
It would not always be possible to check statically that the operand to `call_tir` is a `PrimFunc`. Here is a case where it wouldn't work:

```python
@R.function
def higher_order(f: R.Object) -> R.Object:
    return R.call_tir(f, ...)

@R.function
def g() -> R.Object:
    if some_condition:
        y = some_packed_func
    else:
        y = some_primfunc
    return higher_order(y)
```

We would not be able to determine statically that the call to `higher_order` in `g` will result in a non-`PrimFunc` being passed to `call_tir`. If we care about enforcing the distinction, we would have to check for such cases dynamically.

src/relax/op/op.cc
yongwww: it has a few lines (around 10) of common code here, probably we can keep it

include/tvm/relax/transform.h
yongwww: Thanks for the comments! Actually I was struggling to add CallDPSPackedRewrite, it does duplicate lots of code from CallTIRRewrite, and modifying CallTIRRewrite to work for both call_tir and call_dps_packed brings some confusion (probably a different pass name). I removed CallDPSPackedRewrite in the latest commit, we can add it back when the differences are introduces.

yongwww:
> It would not always be possible to check statically that the operand to `call_tir` is a `PrimFunc`. Here is a case where it wouldn't work:
> 
> ```python
> @R.function
> def higher_order(f: R.Object) -> R.Object:
>     return R.call_tir(f, ...)
> 
> @R.function
> def g() -> R.Object:
>     if some_condition:
>         y = some_packed_func
>     else:
>         y = some_primfunc
>     return higher_order(y)
> ```
> 
> We would not be able to determine statically that the call to `higher_order` in `g` will result in a non-`PrimFunc` being passed to `call_tir`. If we care about enforcing the distinction, we would have to check for such cases dynamically.
> 
> Edit: Otherwise, we could ask for only global vars to be the arguments to `call_tir`, which I don't think would be the worst thing in the world, but this limits the extent to which `PrimFunc`s and `PackedFunc`s are first class. (I don't think they really need to be first class, but in past discussions, we've said that that was desirable.)


Have updated to well-form check to just allow GlobalVar as the first argument of call_tir. Actually current inline PrimFunc was disallowed in Relax IR, `y = some_primfunc` in the example has to be `y = global_var_of_some_primfunc`. I guess we won't run into issues at this moment, correct me if i am wrong

tests/python/relax/test_ast_printer.py
yongwww: Have updated the test case to use call_tir here

slyubomirsky:
Yes, the RHS of that assignment would have to be a global var; we do not permit inline PrimFunc definitions. Perhaps we should get wider agreement if we want to have a restriction of using only global vars for `call_tir` or `call_dps_packed`?

yongwww:
> Yes, the RHS of that assignment would have to be a global var; we do not permit inline PrimFunc definitions. Perhaps we should get wider agreement if we want to have a restriction of using only global vars for `call_tir` or `call_dps_packed`? I can update the spec to reflect whatever we decide.

Perhaps we can discuss this in the open dev meeting today if time allows

src/relax/analysis/well_formed.cc
```diff
@@ -260,6 +261,19 @@ class WellFormedChecker : public relax::ExprVisitor,
       }
     }
 
+    // call_tir works for tir PrimFunc, call_dps_packed works for ExternFunc
+    static const Op& call_tir_op = Op::Get("relax.call_tir");
+    static const Op& call_dps_packed_op = Op::Get("relax.call_dps_packed");
+    if (op->op == call_tir_op && !op->args[0]->IsInstance<GlobalVarNode>()) {
+      Malformed(Diagnostic::Error(op->args[0]->span)
+                << "call_tir expects a prim_func, but gets " << op->args[0]->GetTypeKey());
+    }
```
slyubomirsky, line 270: As mentioned in the community meeting, we can enforce this in the `FInferStructInfo` for `call_tir` and not need any special cases in the well-formedness check.

tests/python/relax/test_ast_printer.py
slyubomirsky: Thanks

src/relax/op/op.cc
```diff
@@ -65,6 +65,9 @@ StructInfo InferStructInfoCallTIR(const Call& call, const BlockBuilder& ctx) {
     ctx->ReportFatal(Diagnostic::Error(call)
                      << "sinfo_args should have exact 1 output struct info.");
   }
+  CHECK(call->args[0]->IsInstance<GlobalVarNode>())
+      << "call_tir expects the first argument as a GlobalVar referring tir PrimFunc. "
```
slyubomirsky, line 69:
```suggestion
      << "call_tir expects the first argument to be a GlobalVar referring to a TIR PrimFunc. "
```

python/tvm/relax/op/base.py
```diff
@@ -51,12 +51,12 @@ def call_tir(
     tir_vars: Optional[Union[ShapeExpr, Tuple[PrimExpr], List[PrimExpr]]] = None,
 ) -> Call:
     """
-    Call a destination-passing-style function and return the output.
+    Call a tir.prim_func function and return the output.
 
     Parameters
     ----------
     func : Union[str, Expr]
```
slyubomirsky, line 58: I think the signature in Python should reflect that we expect the argument to be a global var if that's how we're going to implement it.

slyubomirsky (approved):
My concerns have been addressed and everything seems to be reasonable.

src/relax/transform/call_tir_rewrite.cc
```diff
@@ -33,7 +33,7 @@ namespace relax {
```
psrivas2, line 33: nit: should we rename this file and pass to reflect that it handles both call_dps_packed and call_tir?

src/script/printer/relax/call.cc
```diff
@@ -134,7 +138,7 @@ Optional<ExprDoc> PrintCallTIR(const relax::Call& n, const ObjectPath& n_p, cons
 TVM_STATIC_IR_FUNCTOR(IRDocsifier, vtable)
     .set_dispatch<relax::Call>(  //
         "", [](relax::Call n, ObjectPath n_p, IRDocsifier d) -> Doc {
-          // Special case: call_tir
+          // Special case: call_tir, call_dps_packed
           if (Optional<ExprDoc> doc = PrintCallTIR(n, n_p, d)) {
```
psrivas2, line 142: nit: does PrintCallTIR need renaming?

psrivas2 (approved):
Overall LGTM! Thanks!
Just some minor nits which can be handled in follow up PRs

legalize_ops would throw a warning for call_dps_packed: https://github.com/psrivas2/tvm/blob/emit-te/src/relax/transform/legalize_ops.cc#L105

Would there be follow up PRs for cleaning up transform/analysis passes which earlier checked if call_tir is calling an extern func such as 
* https://github.com/psrivas2/tvm/blob/emit-te/src/relax/backend/task_extraction.cc#L77
* https://github.com/psrivas2/tvm/blob/emit-te/src/relax/transform/fold_constant.cc#L181
* https://github.com/psrivas2/tvm/blob/emit-te/src/relax/transform/fuse_ops.cc#L188
* https://github.com/psrivas2/tvm/blob/emit-te/src/relax/transform/fuse_tir.cc#L305
* https://github.com/psrivas2/tvm/blob/emit-te/src/relax/transform/rewrite_dataflow_reshape.cc#L78

src/relax/transform/call_tir_rewrite.cc
yongwww: I was thinking about this, but don't have a better name, any suggestions? if we change this, then the pass name CallTIRRewrite needs to be updated accordingly.

src/script/printer/relax/call.cc
yongwww: changed it to `PrintCallTIRDPSPacked`

yongwww:
> Overall LGTM! Thanks! Just some minor nits which can be handled in follow up PRs
> 
> legalize_ops would throw a warning for call_dps_packed: https://github.com/psrivas2/tvm/blob/emit-te/src/relax/transform/legalize_ops.cc#L105
> 
> Would there be follow up PRs for cleaning up transform/backend passes which earlier checked if call_tir is calling an extern func such as
> 
> * https://github.com/psrivas2/tvm/blob/emit-te/src/relax/backend/task_extraction.cc#L77
> * https://github.com/psrivas2/tvm/blob/emit-te/src/relax/transform/fold_constant.cc#L181
> * https://github.com/psrivas2/tvm/blob/emit-te/src/relax/transform/fuse_ops.cc#L188
> * https://github.com/psrivas2/tvm/blob/emit-te/src/relax/transform/fuse_tir.cc#L305
> * https://github.com/psrivas2/tvm/blob/emit-te/src/relax/transform/rewrite_dataflow_reshape.cc#L78

Have updated for all passes, thanks a lot for pointing them out!!!

src/relax/transform/call_tir_rewrite.cc
psrivas2: DPSRewrite would be my choice but we can discuss and change it later.

psrivas2 (approved):

src/relax/transform/call_tir_rewrite.cc
yongwww: Thanks for the suggestion! We can discuss it later, it might need to get wider agreement for pass renaming.

PR #14185, opened 2023-03-03
[Refactor] Move `VarUseDefAnalysis` to header file

yzh119:
# Motivation
`UndefinedVars` is a frequently used function in our codebase and its implementation relies on `VarUseDefAnalysis` class which is more general, currently we expose `UndefinedVars` in `analysis.h`, but both the definitions of `UndefinedVars` and `VarUseDefAnalysis` resides in `split_host_device.cc`.

This PR moves `VarUseDefAnalysis` class to `analysis.h` so that developers can use it in other files that requires use/def analysis than `split_host_devices.cc`. We create a `var_use_def_analysis.cc` under `src/src/analysis` for the implementations of both `UndefinedVars` and `VarUseDefAnalysis`.

# Notes
We rename `VarUseDefAnalysis` to `VarUseDefAnalyzer`.

cc @Hzfengsy @Lunderberg


include/tvm/tir/analysis.h
```diff
@@ -86,6 +87,58 @@ TVM_DLL double EstimateTIRFlops(const Stmt& stmt);
  */
 TVM_DLL double EstimateTIRFlops(const IRModule& mod);
 
+/*!
+ * \brief Visitor class to perform use/def analysis, also delete unreferenced lets.
+ * \sa UndefinedVars
+ */
+class VarUseDefAnalyzer : public StmtExprMutator {
```
Hzfengsy, line 94: Do we really need to move the mutator to the public header file?

yzh119: In some cases, we might need to visit structures that are neither an `Stmt` nor a `PrimExpr`, with the result type set instead of an array, existing `UndefinedVars` API is not convenient in this case.

Lunderberg: I agree, the collected variable definitions may be useful in a lot of cases.  Exposing the visitor itself also allows for analyzing multiple expressions, then examining variables that are undefined in any of them.  I'm wondering if there's a tighter API that we'd want to expose, but nothing is coming to mind.

Lunderberg: Though, could we inherit from `StmtExprVisitor` instead of `StmtExprMutator`?  The only modification done is the deletion of unreferenced lets, but that really feels like something that should be part of the `RemoveNoOp` pass.

yzh119: @Lunderberg I'll try to decouple these functionalities and make it a `StmtExprVisitor`.

tqchen: This is a case where hiding things internally still helps, for advanced use-cases, we can put it as an internal header under src/tir/analysis/var_def_analyzer.h

yzh119: Sounds good, besides the location issue. Should we decouple the Analyzer into 3 functions?
1. VarUseDefAnalyzer
2. DeviceInfoCollector
3. SimplifyingLetStmt

I suppose 2 and 3 are not necessary for the general def-use analysis case but necessary for `SplitHostDevice` pass, I prefer doing the refactor and only keeping `VarUseDefAnalyzer` in the header, and making 2 & 3 parts of `split_host_device.cc`.

yzh119:
@Hzfengsy @Lunderberg @tqchen  I have decoupled the previous `VarDefUseAnalyzer` class into three classes:
1. `VarUseDefAnalyzer` which is a `StmtExprVisitor`, we expose it to our files via a header under `src`
2. `DeviceInfoCollector` which is a `StmtVisitor` that collects program information about shared memory usage, thread axis/extents etc.
3. `UnreferencedLetRemover` which removes redundant let stmt/exprs.

I know this class has been used for long time and we don't want any changes on its behavior, so we need to be careful in this refactor, and please correct me if I made any mistakes. One thing I'm not sure is if we need to couple the three functionalities together in a single pass, or we can decouple them like what I did in the refactor.

If my refactor doesn't work, I'm totally okay if we keep the original class and create a simpler `VarDefUseAnalyzer` for light-weight analysis use (e.g. UndefinedVars).

include/tvm/tir/analysis.h
tqchen: That works

tqchen (approved):

include/tvm/tir/analysis.h
```diff
@@ -30,6 +30,7 @@
 #include <tvm/tir/function.h>
 #include <tvm/tir/op_attr_types.h>
 #include <tvm/tir/stmt.h>
+#include <tvm/tir/stmt_functor.h>
```
Lunderberg, line 33: Is this include file required, now that the visitor/mutator are in a separate header file?

src/tir/analysis/var_use_def_analysis.h
```diff
@@ -0,0 +1,91 @@
...
+ */
+class VarUseDefAnalyzer : public StmtExprVisitor {
+ public:
+  explicit VarUseDefAnalyzer(const Array<Var>& defined_vars, bool visit_thread_extent = true);
+  // The fields are publically readible to
+  // be accessible to the users.
+  bool visit_thread_extent_{true};
+  Array<Var> undefined_;
+
+  std::unordered_map<const VarNode*, int> use_count_;
```
Lunderberg, line 51: The `use_count_` and `def_count_` use `const VarNode*` as the map key, whereas `let_binding_` uses `Var`.  I know that they're used somewhat interchangeably throughout the codebase, but it seems odd to have different usage so clsoe together.  Is there a reason for the different usage, and if so, could we add a comment explaining why they were chosen?

src/tir/transforms/split_host_device.cc
```diff
@@ -104,44 +81,42 @@ class VarUseDefAnalysis : public StmtExprMutator {
...
+};
 
-  Stmt VisitStmt_(const StoreNode* op) final {
-    LOG(FATAL) << "Unexpected use of deprecated StoreNode.  Please use BufferStoreNode instead.";
-  }
+/*!
+ * \brief Mutator class to remove unrefenced let stmt/expressions.
+ * \param use_count The pre-computed variable to use count map.
+ */
+class UnreferencedLetRemover : public StmtExprMutator {
```
Lunderberg, line 95: Thank you for making this change, and it really helps in the readability to have the analysis/mutation be separated out!

Lunderberg (approved):
LGTM!  Just a couple of small nitpicks.

src/tir/analysis/var_use_def_analysis.h
yzh119: I don't think there are any specific reasons and I just kept their original type in `split_host_device.cc`. I've changed to use `const VarNode*` for all cases.

src/tir/transforms/split_host_device.cc
yzh119: Thank you for your review! I have updated the code according to your suggestions.

Hzfengsy (approved):

PR #14186, opened 2023-03-03
[TVMC] Add option to dump TIR code to file

lhutton1:
Currently dumps the TIR code after the final phase of lowering (phase 3 from https://github.com/apache/tvm/blob/665dd413bc85d14f7836324daf7cc0dd9281c85a/gallery/how_to/extend_tvm/low_level_custom_pass.py#L152) before codegen. This is done by running a pass to capture the TIR module as it is not saved during the build. The result is saved to file similar to the other code dumps.

In the future it seems good to have more granularity in the IR that is being dumped, e.g. print IR before/after named pass. I thought this might be helpful in the meantime.

cc @ekalda @ashutosh-arm @neildhickey

### DIFF
--- a/tests/python/driver/tvmc/test_compiler.py
+++ b/tests/python/driver/tvmc/test_compiler.py
@@ -40,11 +40,12 @@ from tvm.contrib import utils
 
 def test_save_dumps(tmpdir_factory):
     tmpdir = tmpdir_factory.mktemp("data")
-    dump_formats = {"relay": "fake relay", "ll": "fake llvm", "asm": "fake asm"}
+    dump_formats = {"relay": "fake relay", "tir": "fake tir", "ll": "fake llvm", "asm": "fake asm"}
     tvmc.compiler.save_dumps("fake_module", dump_formats, dump_root=tmpdir)
 
     assert path.exists("{}/{}".format(tmpdir, "fake_module.ll"))
     assert path.exists("{}/{}".format(tmpdir, "fake_module.asm"))
+    assert path.exists("{}/{}".format(tmpdir, "fake_module.tir"))
     assert path.exists("{}/{}".format(tmpdir, "fake_module.relay"))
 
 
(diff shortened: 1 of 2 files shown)


tests/python/driver/tvmc/test_compiler.py
```diff
@@ -87,6 +88,16 @@ def test_compile_tflite_module(use_vm, tflite_mobilenet_v1_1_quant):
...
 
 
+def test_tir_dump(tflite_mobilenet_v1_1_quant):
+    pytest.importorskip("tflite")
+    tvmc_model = tvmc.load(tflite_mobilenet_v1_1_quant)
+    tvmc_package = tvmc.compile(tvmc_model, target="llvm", dump_code="tir")
+    dumps_path = tvmc_package.package_path + ".tir"
+    assert os.path.exists(dumps_path)
+    with open(dumps_path) as f:
+        assert "tir" in f.read()
```
ashutosh-arm, line 98: @lhutton1, feels great to see this addition to `dump_code` :smile: Just a small question about testing: does this file already have a test for dumping multiple IRs from the same command?

lhutton1: It does not unfortunately, I'll add one

lhutton1: Added

ashutosh-arm (approved):
LGTM! Thank you :smile:

neildhickey:
LGTM too. Thanks

echuraev (approved):
LGTM. Thanks

ashutosh-arm:
Thanks @lhutton1, @neildhickey and @echuraev for the commits and the reviews :smile:

PhilippvK:
> In the future it seems good to have more granularity in the IR that is being dumped, e.g. print IR before/after named pass. I thought this might be helpful in the meantime.

@lhutton1 I really like these changes. Are you going to do a followup for this soon? If not it would be great if we could open up an issue to collect some ideas. I plan to do something similar for the RelayIR dump (allow dumping post/pre-partitioning IR) and would like to follow a similar approach.

lhutton1:
Hi @PhilippvK, thanks for the interest! I'm hoping to have a follow-up soon (~1wk) but I have another commitment I need to work on first. I'll summarise the thoughts I have in a discuss post - it would be good to see if they align to your thoughts/requirements before diving in

PR #14190, opened 2023-03-04
[TOPI] Batch Norm Training Mode

SiriusNEO:
Prior to this PR, TOPI batch_norm only supports inference. 

This PR adds `training: bool` flag and `momentum: float` argument to support training mode (update `moving_mean / var` and return), which aligns with `torch.nn.functional.batch_norm`.


SiriusNEO:
cc @MasterJH5574 @tqchen

MasterJH5574 (approved):
LGTM. Thanks for the enhancement!

PR #14192, opened 2023-03-04
[Unity][Transform] LiftTransformParams handling multiple functions

MasterJH5574:
Previously, the LiftTransformParams pass only works on function `"main"`. This is a bit restrictive as in our recent practice on stable diffusion, there are cases where multiple Relax functions inside an IRModule all need to be transformed.

Therefore, this PR enhances the LiftTransformParams pass, so that it will now transform **all** functions **with attribute `num_input`**. For functions without this attribute, the pass will simply skip them.


MasterJH5574:
cc @vinx13

vinx13 (approved):

PR #14197, opened 2023-03-04
[Unity][Frontend] FX translator returning weights with `keep_params_as_input`

MasterJH5574:
PR #14067 introduces the flag `keep_params_as_input` to the FX translator, in the purpose to handle to model weights outside of the translated Relax function.

This PR takes a further step, by returning the model weights as NDArrays when the flag `keep_params_as_input` is true. With this PR, the translator now can return back the weights upon requested. Otherwise, after the import we will lose the model weights in the given PyTorch model.

### DIFF
--- a/python/tvm/relax/frontend/__init__.py
+++ b/python/tvm/relax/frontend/__init__.py
@@ -17,3 +17,5 @@
 """
 Frontends for constructing Relax programs, with the model importers
 """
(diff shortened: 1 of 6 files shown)


MasterJH5574:
cc @vinx13

tqchen (approved):

PR #14200, opened 2023-03-05
[Fix][TVMScript]TVMScript BinOP printing refactor 

cyx-6:
This PR fixes the output for `T.Div(int, int)`. It will print `T.Div(int, int)`, instead of `int / int`, to avoid the integer division ambiguity in parser.

And this PR refactors the logic of binary operators printing in TVMScript. The updated TVMScript printer will print the binary operator to avoid constant folding when parsing back.


junrushao:
Any updates?

python/tvm/script/parser/tir/operation.py
```diff
@@ -46,12 +46,12 @@ def r(op: Type, i: int, m: OpMethod):  # pylint: disable=invalid-name
 
     for i in [0, 1]:
         # Case 1. binop
```
junrushao, line 48: i was thinking if we could remove all the binary operators here given our printer has taken good care of it

cyx-6: Sure! All those overloaded op have been removed.

junrushao (approved):
LGTM!

junrushao:
Feel free to merge it yourself as soon as it turns green

PR #14209, opened 2023-03-06
[MetaSchedule][ARM] Enable ARM CPU intrinsic for MetaSchedule

dsbarinov1:
Motivation:
The purpose of this PR is to add support for intrinsics to optimize matrix multiplication operations (e.g. matmul, convolution) during tuning with MetaScheduler.

Information about PR:
The present PR integrates the existing neon and dotprod (namely, sdot and udot) ARM CPU intrinsics into MetaScheduler, introduces a new "hybrid" dotprod intrinsic ("hdot") working with uint8, uint8 -> int32 data types, and changes the intrinsic selection and application processes for the ARM CPU case, since we operate with multiple intrinsics, rather than with a specific one.

### DIFF
--- a/include/tvm/meta_schedule/schedule_rule.h
+++ b/include/tvm/meta_schedule/schedule_rule.h
@@ -300,6 +300,8 @@ class ScheduleRule : public runtime::ObjectRef {
   TVM_DLL static Array<ScheduleRule, void> DefaultHexagon();
   /*! \brief Create default schedule rules for Micro */
   TVM_DLL static Array<ScheduleRule, void> DefaultMicro();
+  /*! \brief Create default schedule rules for ARM CPU (NEON and DOTPROD) */
+  TVM_DLL static Array<ScheduleRule, void> DefaultARM(const String& type);
 
   TVM_DEFINE_MUTABLE_OBJECT_REF_METHODS(ScheduleRule, ObjectRef, ScheduleRuleNode);
 };
--- a/include/tvm/runtime/container/array.h
+++ b/include/tvm/runtime/container/array.h
@@ -580,6 +580,36 @@ class Array : public ObjectRef {
     }
   }
 
+  template <typename... Args>
+  static size_t CalcCapacityImpl() {
+    return 0;
+  }
+
+  template <typename... Args>
+  static size_t CalcCapacityImpl(Array<T> value, Args... args) {
+    return value.size() + CalcCapacityImpl(args...);
+  }
+
+  template <typename... Args>
+  static size_t CalcCapacityImpl(T value, Args... args) {
+    return 1 + CalcCapacityImpl(args...);
+  }
+
+  template <typename... Args>
+  static void AgregateImpl(Array<T>& dest) {}  // NOLINT(*)
+
+  template <typename... Args>
+  static void AgregateImpl(Array<T>& dest, Array<T> value, Args... args) {  // NOLINT(*)
+    dest.insert(dest.end(), value.begin(), value.end());
+    AgregateImpl(dest, args...);
+  }
+
+  template <typename... Args>
+  static void AgregateImpl(Array<T>& dest, T value, Args... args) {  // NOLINT(*)
+    dest.push_back(value);
+    AgregateImpl(dest, args...);
+  }
+
  public:
   // Array's own methods
 
@@ -679,6 +709,19 @@ class Array : public ObjectRef {
 
   /*! \brief specify container node */
   using ContainerType = ArrayNode;
+
+  /*!
+   * \brief Agregate arguments into a single Array<T>
+   * \param args sequence of T or Array<T> elements
+   * \return Agregated Array<T>
+   */
+  template <typename... Args>
+  static Array<T> Agregate(Args... args) {
+    Array<T> result;
+    result.reserve(CalcCapacityImpl(args...));
+    AgregateImpl(result, args...);
+    return result;
+  }
 
  private:
   /*!
--- a/python/tvm/tir/tensor_intrin/arm_cpu.py
+++ b/python/tvm/tir/tensor_intrin/arm_cpu.py
@@ -26,7 +26,7 @@ from .dot_product_common import DP4A_INTRIN  # pylint: disable=unused-import
 
 
 @T.prim_func
-def dot_product_4x4_i8i8i32_desc(
+def neon_4x4_i8i8i32_desc(
     A: T.Buffer((4,), "int8", offset_factor=1),
     B: T.Buffer((4, 4), "int8", offset_factor=1),
     C: T.Buffer((4,), "int32", offset_factor=1),
@@ -42,7 +42,7 @@ def dot_product_4x4_i8i8i32_desc(
 
 
 @T.prim_func
-def dot_product_4x4_i8i8i32_neon(
+def neon_4x4_i8i8i32_impl(
     A: T.Buffer((4,), "int8", offset_factor=1),
     B: T.Buffer((4, 4), "int8", offset_factor=1),
     C: T.Buffer((4,), "int32", offset_factor=1),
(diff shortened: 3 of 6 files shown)


include/tvm/meta_schedule/schedule_rule.h
```diff
@@ -180,7 +180,7 @@ class ScheduleRule : public runtime::ObjectRef {
    * \return The schedule rule created
    */
   TVM_DLL static ScheduleRule MultiLevelTilingWithIntrin(
-      String intrin_name, String structure, Optional<Array<String>> tile_binds,
+      Array<String> intrin_name, String structure, Optional<Array<String>> tile_binds,
```
vvchernov, line 183: I understand why it is done (String -> Array<String>), but it should be rethink one more time due to API changing affects other places not only your own task

src/meta_schedule/schedule_rule/multi_level_tiling_with_intrin.cc
```diff
@@ -85,21 +101,23 @@ class MultiLevelTilingWithIntrinNode : public MultiLevelTilingNode {
 
  public:
   /*! \brief The name of a tensor intrinsic. */
-  String intrin_name;
+  Array<String> intrin_name;
```
vvchernov, line 104: if the field type is still be changed, I recommend to rename it to `intrin_names` for the sake of clarity

src/meta_schedule/space_generator/space_generator.cc
```diff
@@ -110,6 +155,16 @@ void SpaceGeneratorNode::InitializeWithTuneContext(const TuneContext& context) {
       default_sch_rules = ScheduleRule::DefaultMicro();
       default_postprocs = Postproc::DefaultMicro();
       default_mutator_probs = Mutator::DefaultMicro();
+    } else if (kind == "neon") {
```
vvchernov, line 158: It looks like different levels of target types are checked here. Possibly it should be "arm" type with splitting to "neon"/"dotprod" in separated method.

vvchernov (commented):
I leave comments here to think about how to make the code cleaner, but they are not required to fix

src/meta_schedule/space_generator/space_generator.cc
```diff
@@ -21,6 +21,28 @@
...
+}
+
+// static inline bool HasFlag_(Optional<String> attr, std::string flag) {
+//   if (!attr) {
+//     return false;
+//   }
+//   return HasFlag_(attr.value(), flag);
+// }
+
+static inline bool HasFlag_(Optional<Array<String>> attr, std::string flag) {
```
ibsidorenko, line 36: Looks like we have the same code in `src/target/parsers/aprofile.cc`.
Instead of code duplication, can we move it into common place?

src/meta_schedule/schedule_rule/schedule_rule.cc
```diff
@@ -295,6 +298,119 @@ Array<ScheduleRule> ScheduleRule::DefaultMicro() {
...
+          /*into_consumer=*/true,
+          /*inline_const_tensor=*/true,
+          /*disallow_if_then_else=*/true,
+          /*require_injective=*/true,
+          /*require_ordered=*/true,
+          /*disallow_op=*/Array<String>{"tir.exp"}),
+      ScheduleRule::AddRFactor(
+          /*max_jobs_per_core=*/8,
+          /*max_innermost_factor=*/Integer(32)),
+      ScheduleRule::MultiLevelTilingWithIntrin(
```
ibsidorenko, line 361: As I understand, new API in `MultiLevelTilingWithIntrin` is  not required anymore?

include/tvm/meta_schedule/schedule_rule.h
dsbarinov1: Yes, new API changes were reverted to the original ones, while keeping the same new functionality.

src/meta_schedule/space_generator/space_generator.cc
dsbarinov1: Code duplication fixed, we are now using different method to pull specific keys from the target.

src/meta_schedule/schedule_rule/schedule_rule.cc
dsbarinov1: Yes, it is not in use rn.

dsbarinov1:
@ibsidorenko could you review my changes, please ? :)

python/tvm/tir/tensor_intrin/arm_cpu.py
```diff
@@ -131,8 +163,68 @@ def dot_product_4x4_i8i8i32_sdot(
...
+        A_i32 = T.reinterpret(A_i8x4, dtype="uint32")
+        vec_ai32 = T.broadcast(A_i32, 4)
+        vec_a = T.reinterpret(vec_ai32, dtype="uint8x16")
+
+        vec_b = B.vload([0, 0], dtype="uint8x16")
+
+        vec_c = C.vload([0], dtype="int32x4")
+
+        C[T.ramp(T.int32(0), 1, 4)] = T.call_llvm_pure_intrin(
+            T.llvm_lookup_intrinsic_id("llvm.aarch64.neon.udot.v4u32.v16u8"),
```
ibsidorenko, line 215: You use the same intrinsic id "llvm.aarch64.neon.udot.v4u32.v16u8" in both cases. Is it Ok?

dsbarinov1: When experimenting with **tflite_mobilenet_v3_quant** model, we encountered convolution with multiplication of uint8 uint8 tensors into int32 accumulator, which did not allow to apply existing _sdot/udot_ intrinsics, so we had to create new _hdot_ intrinsic, which already works with such dtypes layout. From my knowledge, there is no such neon instruction to work with u8u8i32 layout of dtypes, for that reason we could try to call the closest instruction, which we did and the intrinsic succesfully applied to that type of convolution, bringing us a performance benefit.

src/meta_schedule/schedule_rule/multi_level_tiling_with_intrin.cc
dsbarinov1: Outdated.

src/meta_schedule/space_generator/space_generator.cc
dsbarinov1: Done.

dsbarinov1:
@echuraev, @masahi, could you review my changes, please :)

python/tvm/tir/tensor_intrin/arm_cpu.py
```diff
@@ -131,8 +163,68 @@ def dot_product_4x4_i8i8i32_sdot(
...
+            T.uint32(3),
+            vec_c,
+            vec_a,
+            vec_b,
+            dtype="int32x4",
+        )
```
masahi, line 221: It should be possible to clean up a lot of code duplication between different dtypes here. See `tensor_intrin/cuda.py` for examples.

src/meta_schedule/schedule_rule/schedule_rule.cc
```diff
@@ -295,6 +295,118 @@ Array<ScheduleRule> ScheduleRule::DefaultMicro() {
...
+      ScheduleRule::ParallelizeVectorizeUnroll(
+          /*max_jobs_per_core=*/8,
+          /*max_vectorize_extent=*/32,
+          /*unroll_max_steps=*/Array<Integer>{0, 8, 32, 256},
+          /*unroll_explicit=*/true),
+      ScheduleRule::RandomComputeLocation(),
+  };
+}
+
+Array<ScheduleRule> ScheduleRule::DefaultARMDotprod() {
```
masahi, line 343: Please remove the dup with `DefaultARMDotprod`, to make the difference obvious.

masahi (changes requested):

include/tvm/runtime/container/array.h
```diff
@@ -580,6 +580,36 @@ class Array : public ObjectRef {
...
+    return value.size() + CalcCapacityImpl(args...);
+  }
+
+  template <typename... Args>
+  static size_t CalcCapacityImpl(T value, Args... args) {
+    return 1 + CalcCapacityImpl(args...);
+  }
+
+  template <typename... Args>
+  static void AgregateImpl(Array<T>& dest) {}  // NOLINT(*)
```
dsbarinov1, line 599: This edit (NOLINT) was made on the consideration that *quote* _"Google C++ Style Guide seems to have allowed using non-const references as parameters"_.
Reference: https://github.innominds.com/cpplint/cpplint/issues/148

masahi:
@tvm-bot rerun

dsbarinov1:
@masahi, could you review my changes, please? :)

src/meta_schedule/schedule_rule/schedule_rule.cc
```diff
@@ -295,6 +295,94 @@ Array<ScheduleRule> ScheduleRule::DefaultMicro() {
...
+          /*vector_load_lens=*/NullOpt,
+          /*reuse_read=*/NullOpt,
+          /*reuse_write=*/
+          Map<String, ObjectRef>{{"req", String("may")},
+                                 {"levels", Array<Integer>{1, 2}},
+                                 {"scope", String("global")}}),
+  };
+}
+
+Array<ScheduleRule> GetDotprodSpecificRules() {
```
masahi, line 314: This is specific to "ARM" dot product only so the naming is not the best.  I'll merge it for now, but please fix this when you get a chance later.

masahi (approved):

masahi:
sorry I forgot to take another look

PR #14215, opened 2023-03-06
[WIP][Unity][Pass] Add a pass to alter the TIR implementation of an operator

psrivas2:
Add a pass to alter the TIR implementation of an operator (identified by operator_kind attribute on PrimFunc). It also inserts layout changes to i/o buffers at Relax level.


tests/python/relax/test_transform_alter_op_impl.py
```diff
@@ -0,0 +1,342 @@
...
+                T.writes(output[v_ax0, v_ax1])
+                output[v_ax0, v_ax1] = arg0[v_ax0, v_ax1] + arg1[v_ax0, v_ax1]
+    # fmt: on
+    index_map = lambda i: (i // 4, i % 4)
+    _check(
+        Before,
+        Expected,
+        operator_name="relax.add",
+        replacement_primfunc=add_2d,
+        layout_changes=[index_map, index_map, index_map],
```
masahi, line 94: Looking at this example, I don't understand why we need to pass `replacement_primfunc`. If we can identify the function to transform by `operator_name`, wouldn't the index maps be sufficient to realize the same transformation?

src/relax/transform/alter_op_impl.cc
```diff
@@ -0,0 +1,312 @@
...
+
+    Array<Expr> updated_inputs;
+    int index = 0;
+    for (const auto& input : inputs->fields) {
+      auto transform = transforms[index++];
+      ICHECK(IsTransformBijective(input, transform))
+          << "Non bijective transforms on input and output buffers are not supported.";
+      updated_inputs.push_back(TransformLayout(input, transform));
+    }
+    return {replacement_gv, Tuple(updated_inputs)};
```
masahi, line 213: It doesn't seem necessary to pass `replacement_gv` to this function.

tests/python/relax/test_transform_alter_op_impl.py
psrivas2: Index maps would only be able to modify the i/o buffer layouts, but not the underlying compute which could be manually scheduled PrimFunc for best performance.

Indeed this is a toy example and the underlying compute is trivial, however, we can imagine a use case where the user wants to drop in a manually scheduled PrimFunc replacement for a complex operation like conv2d, attention, etc.

src/relax/transform/alter_op_impl.cc
psrivas2: right, thanks! updated the function

tests/python/relax/test_transform_alter_op_impl.py
sunggg: So, this pass will serve general TIR->TIR replacement based on the operator?
If so, maybe good to clarify its intension in the comment. 
Also, like @masahi suggested, it might be bit redundant to pass both index_map and TIR func for layout transformation case. 
Can we add some syntax sugar to make it easier to use?

tests/python/relax/test_transform_alter_op_impl.py
```diff
@@ -0,0 +1,342 @@
...
+    tvm.ir.assert_structural_equal(after, expected)
+
+
+def test_single_output():
+    # fmt: off
+    @I.ir_module
+    class Before:
+        @T.prim_func
+        def add(arg0: T.Buffer((16,), "float32"), arg1: T.Buffer((16,), "float32"), output: T.Buffer((16,), "float32")):
+            T.func_attr({"operator_name": "relax.add"})
```
sunggg, line 41: Do we already add this attribute during lowering?

sunggg (commented):
Thank you for the step forward in layout transformation, @psrivas2!
A couple of comments.

tests/python/relax/test_transform_alter_op_impl.py
masahi: It would be great if we can automatically infer the index maps from the original and the replacement func.

psrivas2: To support automatic inference of index maps, I can think of one possible way. The original and replacement PrimFunc has layout string specified for each param buffer using function attributes. We can then infer the layout transformation on the buffers. 

The contract would be:
* name of the function attribute specifying layouts, say "layouts"
* the length of `layouts` is same as number of params on PrimFunc, so layouts[i] corresponds to param[i] layout.
* the notation used for layouts between original and replacement PrimFunc matches. So if `original_conv2d` uses "NCHW" notation it should match the `replacement_conv2d` notation, otherwise we throw an error.

I was hesitant on introducing this contract but if there is consensus on it, I am happy to support it and make index_maps optional. Let me know in comments.

```python
@T.prim_func
def original_conv2d(arg0: T.Buffer((32, 3, 224, 224), "float32"), arg1: T.Buffer((5, 5, 3, 64), "float32"), output: T.Buffer((32, 64, 224, 224), "float32")):
    T.func_attr({"operator_name": "relax.conv2d", "layouts": ["NCHW", "HWIO"]})
    ...

@T.prim_func
def replacement_conv2d(arg0: T.Buffer((32, 224, 224, 3), "float32"), arg1: T.Buffer((64, 3, 5, 5), "float32"), output: T.Buffer((32, 224, 224, 64), "float32")):
    T.func_attr({"operator_name": "relax.conv2d", "layouts": ["NHWC", "OIHW"]})
    ...
```

masahi: I was hoping that `DetectIterMap` on the original and replacement buffer shapes would just work... Is it not sufficient?

psrivas2: `DetectIterMap` maps a buffer access into it's quasi-affine representation. It is unclear to me how `DetectIterMap` would be useful here. Let's take a toy example of `add` computation. Given two implementations of `add` below how would `DetectIterMap` figure out the layout transformations on any param? 

```python
@T.prim_func
def add_2d(arg0: T.Buffer((4, 4), "float32"), arg1: T.Buffer((4, 4), "float32"), output: T.Buffer((4, 4), "float32")):
    for ax0, ax1 in T.grid(4, 4):
        with T.block("T_add"):
            v_ax0, v_ax1 = T.axis.remap("SS", [ax0, ax1])
            T.reads(arg0[v_ax0, v_ax1], arg1[v_ax0, v_ax1])
            T.writes(output[v_ax0, v_ax1])
            output[v_ax0, v_ax1] = arg0[v_ax0, v_ax1] + arg1[v_ax0, v_ax1]
```

```python
@T.prim_func
def add(arg0: T.Buffer((16,), "float32"), arg1: T.Buffer((16,), "float32"), output: T.Buffer((16,), "float32")):
    T.func_attr({"operator_name": "relax.add"})
    for ax0 in range(16):
        with T.block("T_add"):
            v_ax0 = T.axis.spatial(16, ax0)
            T.reads(arg0[v_ax0], arg1[v_ax0])
            T.writes(output[v_ax0])
            output[v_ax0] = arg0[v_ax0] + arg1[v_ax0]
```

masahi: Sorry I was naively thinking that there should be a way to figure out an index map that does the transformation `(4, 4) -> (16,)`, using `DetectIterMap`. Maybe I was confused with something else I did in the past.

masahi (approved):

psrivas2:
Thank you @masahi and @sunggg for reviews! 
This PR is ready to be merged

PR #14217, opened 2023-03-06
[TVMScript] Sugar T.env_thread + T.launch_thread

junrushao:
This PR introduces a syntactic sugar that combines T.env_thread and T.launch_thread.

Previously, an AttrStmt that specifies thread extent or virtual thread is required to be written in two steps:

```python
bx = T.env_thread("blockIdx.x")  // creates an IterVar
with T.launch_thread(bx, 128):   // specify the iter domain
  ...
```

With this PR, now this behavior can be merged in a single line:

```python
with T.launch_thread("blockIdx.x", 128) as bx:
  ...
```

### DIFF
--- a/include/tvm/script/ir_builder/tir/ir.h
+++ b/include/tvm/script/ir_builder/tir/ir.h
@@ -391,6 +391,14 @@ DeclBufferFrame DeclBuffer(Array<PrimExpr> shape, DataType dtype, String buffer_
 LaunchThreadFrame LaunchThread(Var var, PrimExpr extent);
 
 /*!
+ * \brief Launch a new thread.
+ * \param thread_tag The thread type tag.
+ * \param extent The extent of environment thread.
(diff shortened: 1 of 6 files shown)


junrushao:
CC: @spectrometerHBH @yzh119 @Hzfengsy

spectrometerHBH (approved):

Hzfengsy:
Could you please add a test case for it?

junrushao:
@Hzfengsy Existing roundtripping tests should suffice already. Do you prefer if I added a new one?

Hzfengsy:
> @Hzfengsy Existing roundtripping tests should suffice already. Do you prefer if I added a new one?

It would be great if we can have another explicit one :)

junrushao:
Got it. @Hzfengsy Just added a new testcase

Hzfengsy (approved):

PR #14218, opened 2023-03-06
[MetaSchedule] Change anchor trace apply to block-mapping based approach

jinhongyii:
The old anchor trace apply assumes that the prefix of the "new" outputs matches with the "old" outputs, which is not the case for some tasks in complex models. Also the old approach use block name to match corresponding blocks among the old module in database and the new module to apply trace on.  In some cases, the block names may differ even though the block body are identical. In this PR, I maintain a block mapping between the old IRmodule and the new IRmodule, so that we can find the matching parts in both IRmodule with no assumptions of the relationship of their locations in IRModule.

cc: @tqchen @junrushao @masahi

### DIFF
--- a/python/tvm/meta_schedule/trace_apply.py
+++ b/python/tvm/meta_schedule/trace_apply.py
@@ -16,11 +16,14 @@
 # under the License.
 """Specialized applications of trace"""
 from ..tir.schedule import Schedule, Trace
+from ..ir.module import IRModule
 from ..target import Target
 from . import _ffi_api
 
 
-def schedule_using_anchor_trace(sch: Schedule, anchor_trace: Trace, target: Target) -> None:
+def schedule_using_anchor_trace(
+    sch: Schedule, anchor_trace: Trace, record_mod: IRModule, target: Target
+) -> None:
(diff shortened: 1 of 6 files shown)


junrushao:
I have to admit I don't fully understand this...Would you like to elaborate a bit? Thanks!

masahi:
I'll can take a look tomorrow. Please fix the CI issues in the mean time.

jinhongyii:
The CI seems weird. I cannot reproduce the error locally. The error mainly results from the string difference of `v_ax2 + v_ax1 * T.int64(16)` vs `v_ax1 * T.int64(16) + v_ax2`. I don't know why it will generate different results on different machine.

jinhongyii:
To elaborate this PR, this PR solves 2 cases that the old trace apply doesn't handle:
1. struturally equal blocks with different block names are not recognized as matched blocks
2. trace apply assumes that the prefix of the "new" outputs matches with the "old" outputs for get_child_blocks inst. But it's not true for some cases. Maybe the prefix of the old outputs match the new outputs or maybe the mapping is not for an interval of blocks.

PR #14219, opened 2023-03-06
[MetaSchedule] preseve global_symbol attached to function after applying MS

jinhongyii:
Right now seems a global symbol main will be attached to every fused prim_func after Meta Schedule tuning.
Although the problem get hidden because we will re-attach global symbols, this is a behavior that need to be fixed.
Now the transformed prim func preserves the original global symbol. If it doesn't exist, attach global symbol main.

cc: @tqchen @junrushao

### DIFF
--- a/python/tvm/meta_schedule/tune_context.py
+++ b/python/tvm/meta_schedule/tune_context.py
@@ -44,7 +44,8 @@ if TYPE_CHECKING:
 def _normalize_mod(mod: Union[PrimFunc, IRModule]) -> IRModule:
     """Normalize the input to an IRModule"""
     if isinstance(mod, PrimFunc):
-        mod = mod.with_attr("global_symbol", "main")
+        if not (mod.attrs and "global_symbol" in mod.attrs):
+            mod = mod.with_attr("global_symbol", "main")
         mod = mod.with_attr("tir.noalias", True)
         mod = IRModule({"main": mod})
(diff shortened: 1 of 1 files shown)


junrushao (approved):
LGTM!

python/tvm/meta_schedule/tune_context.py
```diff
@@ -44,7 +44,8 @@
 def _normalize_mod(mod: Union[PrimFunc, IRModule]) -> IRModule:
     """Normalize the input to an IRModule"""
     if isinstance(mod, PrimFunc):
-        mod = mod.with_attr("global_symbol", "main")
+        if not (mod.attrs and "global_symbol" in mod.attrs):
+            mod = mod.with_attr("global_symbol", "main")
         mod = mod.with_attr("tir.noalias", True)
         mod = IRModule({"main": mod})
```
MasterJH5574, line 50: I’m not very sure here. One of the principle is that when a function in an IRModule has global_symbol, the global_symbol value and its global var name are supposed to equal. Here we are creating an IRModule with the only function “main”, so I think it is reasonable to attach global_symbol “main” to the function.

jinhongyii: Good thought! The inconsistency of global_symbol in the case I'm handling comes from the applying MS phase, so it seems I should change the code there and revert this change. I will send a fix PR later

PR #14224, opened 2023-03-07
[TIR] Enhance loop unroll with unroll local access

tqchen:
This PR enhances the unroller with an unroll local access option. This option will detect loop variables that access local memories and unroll them independent of other options.

A test case is added. This option is by default turned off and can be useful in certain cases to improve unroller as these local memory access have to be unrolled at some time pt to be lifted as registers.

### DIFF
--- a/src/tir/transforms/unroll_loop.cc
+++ b/src/tir/transforms/unroll_loop.cc
@@ -33,6 +33,7 @@
 #include <unordered_set>
 #include <vector>
 
+#include "../../runtime/thread_storage_scope.h"
 #include "ir_utils.h"
 
 namespace tvm {
@@ -43,6 +44,7 @@ struct UnrollLoopConfigNode : public tvm::AttrsNode<UnrollLoopConfigNode> {
   int auto_max_depth;
   int auto_max_extent;
   int explicit_unroll;
+  int unroll_local_access;
 
   TVM_DECLARE_ATTRS(UnrollLoopConfigNode, "tir.transform.UnrollLoopConfig") {
     TVM_ATTR_FIELD(auto_max_step)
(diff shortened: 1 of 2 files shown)


src/tir/transforms/unroll_loop.cc
```diff
@@ -137,14 +165,36 @@ class LoopUnroller : public StmtExprMutator {
...
+        }
+      }
+    }
+    return StmtMutator::VisitStmt_(op);
   }
 
   Stmt VisitStmt_(const EvaluateNode* op) final {
     ++step_count_;
-    return StmtExprMutator::VisitStmt_(op);
+    return StmtMutator::VisitStmt_(op);
```
yzh119, line 197: Why changed to `StmtMutator`?

Hzfengsy (approved):

src/tir/transforms/unroll_loop.cc
```diff
@@ -137,14 +165,36 @@ class LoopUnroller : public StmtExprMutator {
     LOG(FATAL) << "Unexpected use of deprecated StoreNode.  Please use BufferStoreNode instead.";
   }
 
+  PrimExpr VisitExpr_(const BufferLoadNode* op) final {
+    if (unroll_local_access_) {
+      auto storage_scope = runtime::StorageScope::Create(GetPtrStorageScope(op->buffer->data));
+      if (storage_scope.rank == runtime::StorageRank::kLocal) {
```
yzh119, line 171: Should we also consider warp-level memory (identical to local for CUDA).

tqchen: added

yzh119 (approved):
LGTM

PR #14226, opened 2023-03-07
[Unity] ensure memory.alloc_tensor/storage roundtrippable

yongwww:
- make R.memory.alloc_tensor/alloc_storage roundtrippable
- Expose relax.vm ops to python side

cc: @MasterJH5574


MasterJH5574 (approved):
Thanks for the fix!

PR #14227, opened 2023-03-07
[TVMScript] Support SizeVar Roundtripping

junrushao:
Previously the printer does not respect the difference between tir.Var and tir.SizeVar, which means SizeVar info will be lost during roundtripping, which is not ideal when dealing with dynamic shape workloads, where SizeVar provides additional bound information to the arithmetic analyzer.

This PR extends Var printing to support SizeVar:

```python
>>> a = tir.SizeVar("a", "int32")
>>> print(a.script(verbose_expr=True))
a = T.int32(is_size_var=True)
a
```

### DIFF
--- a/include/tvm/script/ir_builder/tir/ir.h
+++ b/include/tvm/script/ir_builder/tir/ir.h
@@ -430,14 +430,27 @@ void Evaluate(PrimExpr value);
  * \brief Create a TIR var that represents a pointer
  * \param dtype The data type of the pointer.
  * \param storage_scope The storage scope of the pointer.
+ * \param is_size_var Whether the pointer is a size var.
  * \return The pointer.
  */
-Var Handle(runtime::DataType dtype = runtime::DataType::Void(), String storage_scope = "global");
+inline Var Handle(runtime::DataType dtype = runtime::DataType::Void(),  //
+                  String storage_scope = "global",                      //
+                  bool is_size_var = false) {
+  Type type_annotation{nullptr};
+  if (dtype.is_void() && storage_scope == "global") {
+    type_annotation = PrimType(runtime::DataType::Handle());
+  } else {
+    type_annotation = PointerType(PrimType(dtype), storage_scope);
+  }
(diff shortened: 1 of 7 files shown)


include/tvm/script/ir_builder/tir/ir.h
```diff
@@ -430,14 +430,26 @@ void Evaluate(PrimExpr value);
...
  */
-Var Handle(runtime::DataType dtype = runtime::DataType::Void(), String storage_scope = "global");
-
-#define TVM_TIR_IR_BUILDER_DEF_DTYPE_CAST(FuncName, DType)                             \
-  inline PrimExpr FuncName(Optional<PrimExpr> expr = NullOpt) {                        \
-    DataType dtype = DType;                                                            \
-    return expr.defined() ? tvm::cast(dtype, expr.value()) : tvm::tir::Var("", dtype); \
+inline Var Handle(runtime::DataType dtype = runtime::DataType::Void(),  //
+                  String storage_scope = "global",                      //
+                  bool size_var = false) {
```
Hzfengsy, line 438: Would be great if we can rename it to `is_size_var`

python/tvm/script/ir_builder/tir/ir.py
```diff
@@ -1333,11 +1333,13 @@ def func(
             Literal["inf", "-inf", "nan"],
             int,
             float,
-        ] = None
+        ] = None,
+        *,
+        size_var: bool = False,
```
Hzfengsy, line 1338: dito

Hzfengsy (approved):
LGTM except a naming change

include/tvm/script/ir_builder/tir/ir.h
junrushao: It's a matter of naming, and I personally believe `size_var` is a more proper name for the following reasons:
- It's shorter
- It already implies the semantics of "is_"
- it's consistent with TIR's SizeVar.

Well, I don't hold strong opinion though, so open to change if you insist `is_size_var` be slightly better :-)

junrushao: Updated to `is_size_var` per discussion with @Hzfengsy

PR #14228, opened 2023-03-07
[Unity][Bugfix] Do not include `PrimFunc`s in the dependency graph when checking for recursion

slyubomirsky:
This PR fixes a small oversight in #14149. Namely, the analysis for detecting recursion in Relax adds all reference to global vars into the dependency graph, but some global vars might refer to TIR `PrimFunc`s. Those should not be in the dependency graph, and this PR ignores any that it detects.


yongwww (approved):
LGTM

Hzfengsy (approved):

PR #14230, opened 2023-03-08
[Fix]Fix function ObjectPath in IRModule SEqual

cyx-6:
This PR fixes the `IRModuleNode::SEqualReduce` to produce the correct `ObjectPath` for functions in `IRModule`.

And this PR fixes some missing `Doc->source_paths`'s, when printer transforms `Doc` locally.


junrushao (approved):
LGTM!

PR #14231, opened 2023-03-08
Enable onednn in tvm.

feng-intel:


leandron:
Hi @feng-intel thanks for the PR.

I saw you're basing this change with `v0.11.0` branch. Can you please move it to `main` as v0.11.0 was already released? Also, it would be good to add a comprehensive description on the support you are adding in the context of this PR, perhaps pointing out which operators are supported, etc. As much information as possible please.

python/tvm/__init__.py
```diff
@@ -32,6 +32,7 @@
 # tvm.runtime
 from .runtime.object import Object
 from .runtime.ndarray import device, cpu, cuda, gpu, opencl, cl, vulkan, metal, mtl
+#from .runtime.ndarray import device, cpu, cuda, sycl, gpu, opencl, cl, vulkan, metal, mtl
```
leandron, line 35: I think this is a duplicated line that needs to be removed?

CMakeLists.txt
```diff
@@ -12,6 +12,8 @@ include(cmake/utils/FindLLVM.cmake)
 include(cmake/utils/FindROCM.cmake)
 include(cmake/utils/FindEthosN.cmake)
 
+set(CMAKE_VERBOSE_MAKEFILE on)
```
leandron, line 15: Does this need removing as well?

leandron (changes requested):
I'd like to suggest for a quick review in the PR in general, to make sure:
- We have a CMAKE flag e.g. `USE_ONEDNN` to toggle compilation with and without the library
-  Debug output in the sense of `print(...)` and `std::cout << ...` is replaced with use of logging infrastructure
- Test coverage is improved to cover all supported operators.

apeskov:
I guess this PR was submitted by mistake. Looks like this dev branch has some minimal PoC with offloading SoftMax to OneDNN for GPU device. It's definitely not ready for merging.

Could you please change this branch to draft status or move to your private TVM fork to prevent other guys from spending time to review.

PR #14232, opened 2023-03-08
[TEST] Fix division by 0 in llvm codegen test

lhutton1:
In the test `test_llvm_div`, some cases can lead to a division by 0. Consider the case when `start=-12, end=-12, dstart=-2 and dend=0`, the range of values input to clipb() will be: [-12, ..., -1, 1] (with 1 being specially selected to prevent division by 0). For the input 1, clipb evaluates to: `min(/*dend=*/0, max(/*dstart=*/-12, /*i=*/1)) = 0`, thus leading to a division by 0.

LLVM11, which runs in CI, doesn't seem to throw any kind of error, while using LLVM15 leads to a "floating point exception". Regardless, since division by 0 is undefined behaviour and the compiler can do anything, the solution here is to fix the test so division by 0 doesn't occur.

### DIFF
--- a/tests/python/unittest/test_target_codegen_llvm.py
+++ b/tests/python/unittest/test_target_codegen_llvm.py
@@ -525,7 +525,7 @@ def test_llvm_div():
     ]:
         for dstart, dend in [
             (-11, -1),
-            (-11, 0),
+            (-11, 1),
             (-4, -4),
             (-2, -2),
             (1, 11),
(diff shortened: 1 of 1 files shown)


tests/python/unittest/test_target_codegen_llvm.py
```diff
@@ -534,7 +534,7 @@ def _show_info():
             (2, 2),
             (-11, 11),
         ]:
-            if end < start or dend < dstart or (dend == 0 and dstart == 0):
```
ashutosh-arm, line 537: nit: could you please tally against https://github.com/apache/tvm/blob/902c2e2db70b75c36f9bd8c253707b1e8761cc18/tests/python/unittest/test_tir_nodes.py#L300 to check the divide by zero test case is covered there?

lhutton1: This test seems to be checking for div by 0 at compile time due to the use of constants, in the case of `test_llvm_div`, the div by 0 occurs at runtime, so I don't think this is something we can check here

PR #14233, opened 2023-03-08
[Unity] Add span to the Relax ops in python

gigiblender:
- Adds span as a parameter to the Relax ops in python
- Adds a SpanContext that can be used to set the span of multiple ops


gigiblender:
@jwfromm

junrushao:
unfortunately I don’t think span is used in either TIR or Relax via TVMScript at the moment. This is a legacy feature we may or may not really want to use. The recommended and current way of rendering errors is via ObjectPath.

gigiblender:
> unfortunately I don’t think span is used in either TIR or Relax via TVMScript at the moment. This is a legacy feature we may or may not really want to use. The recommended and current way of rendering errors is via ObjectPath.

I don't think this is about rendering errros. The intent here is to use the span to support tracking of relax ops that are created by the frontend. For example, an ONNX conv2d can generated calls to `relax.op.nn.conv2d`, `relax.op.reshape` and `relax.op.add`. We want to be able to attach span information to the relax ops that points back to the ONNX layer that generated them (ONNX conv2d in this example).

junrushao:
that said, we don’t want to have span in general

jwfromm:
Thanks for the though Junru, the idea with span here is essentially just to annotate operators, in this case with which onnx node they originated from. If not span, how would you recommend doing that annotation?

junrushao:
Hey sorry for the long delay!

Sure I agree there might be usecases specific to ONNX importer, and I fully support a solution if the design is fully confined to the importer itself. Right now the primary issue with the current design is that it's been leaked to global IR structure, which is something we've made significant amount of effort last year to move away from.

To provide some more context: why is `span` considered non-ideal in TVM stack? This is because 1) preserving `span` across transformation could be notoriously challenging. For example, after some TIR scheduling transformation, eventually the `span`s left are non-existent or meaningless; 2) the IR is effectively not a tree but a DAG - for example, the very same `relax.Var` or `tir.Buffer`/`tir.Var` may appear multiple times in different parts of the IR, and so they do not have a single deterministic `span`.

One potential approach we might think of is to save some node-to-span information as a context or IR attributes, which are defined specific to ONNX. I'm not thinking a lot about details, but please let me know if we could discuss a bit further

areusch:
@junrushao looking over ObjectPath, it seems like a way to identify a particular IR node when printed. I agree it accomplishes the goal of identifying a tree path to an IR node which may otherwise have multiple.

i think that's different from the goal here--the goal is to identify which framework operator ultimately caused such IR to be included. I agree this gets murky as transforms run--scheduling, for instance, creates a TIR function in which the Span could be blanket-assigned to the generated PrimFunc. If a PrimFunc fused two graph-level operators, different fragments of TIR would have different Span information attached. Problem 1 you're describing here has been contemplated in [TVM Explorer Infra RFC](https://discuss.tvm.apache.org/t/pre-rfc-tvm-explorer-infrastructure/13457) and the proposed/accepted solution was SequentialSpan (in other words, attach both parent spans when e.g. a further optimization on a schedule pass intermixes statements with different parent spans).

we can also consider other solutions here, but I'm not sure I see how ObjectPath helps here. could you point us to an RFC or other discussion where Span was deprecated?

junrushao:
> could you point us to an RFC or other discussion where Span was deprecated?

To clarify, I am not proposing to deprecate `Span` support in current TIR as it is right now, and its potential deprecation is subject to more careful discussion. What I am implying is not to add span in Relax, which is a different topic.

> Problem 1 you're describing here has been contemplated in [TVM Explorer Infra RFC](https://discuss.tvm.apache.org/t/pre-rfc-tvm-explorer-infrastructure/13457) and the proposed/accepted solution was SequentialSpan (in other words, attach both parent spans when e.g. a further optimization on a schedule pass intermixes statements with different parent spans)

I agree that SequentialSpan might solve some problem at Relay level and I fully support the RFC. The problem I pointed out in this thread is that the IR is not a tree, but a DAG, which cannot be represented simply by a context-independent `Span` class. In this case, a node should exist with the path from root, which is ObjectPath.

junrushao:
Some further comments:

One thing that we need to consider here is the overall engineering overhead and execution time in maintaining the tracking. I agree such tracking would be useful, however, relying solely on it for error reporting would be problematic since inaccurate tracking results in garbage-in-garbage-out situation. Such information also becomes less useful (and harder to accurately track) as we goes further deeper into the pipeline.

As a reference, solutions like Torch FX do not come with span information as there is 1-1 correspondence with the current state of transformation (when they turn back into python). In our case, having TVMScript at the current point, together with ObjectPath and use object itself,  for error reporting.  See [how](https://github.com/apache/tvm/blob/unity/src/relax/op/op_common.h#L107) existing error reporting leverages this mechanism that will provide something that always works even when span is not available.

Note that it is always possible to recover `obj->span` if it is available, and there can be helpful scenarios in frontend that tracks span and early transformations. So my recommendation would be:
- Encourage all error reporting to report object which leverages ObjectPath, which by default  provide good amount of error mechanisms regardless how accurate span get tracked.
- The span information remains optional and can be used when available.

Issue #14136, opened 2023-02-27
[Bug] fatal error: llvm/ADT/Triple.h: No such file or directory


ZhaoDongyu-AK47:
### Expected behavior

Successfully completed make

### Actual behavior
After I installed llvm, I configured USE_LLVM in config.cmake. When make tvm, I encountered the following error
```
/home/Project/tvm/src/target/llvm/codegen_blob.cc:29:10: fatal error: llvm/ADT/Triple.h: No such file or directory
   29 | #include <llvm/ADT/Triple.h>
      |          ^~~~~~~~~~~~~~~~~~~
compilation terminated.
make[2]: *** [CMakeFiles/tvm_objs.dir/build.make:8280: CMakeFiles/tvm_objs.dir/src/target/llvm/codegen_blob.cc.o] Error 1
make[2]: *** Waiting for unfinished jobs....
/home/observer/Project/tvm/src/target/llvm/codegen_llvm.cc:30:10: fatal error: llvm/ADT/Triple.h: No such file or directory
   30 | #include <llvm/ADT/Triple.h>
      |          ^~~~~~~~~~~~~~~~~~~
compilation terminated.
```
So I executed the following statement:
```
~/Project$ find . -name Triple.h
./llvm-project/llvm/include/llvm/TargetParser/Triple.h
./llvm/include/llvm/TargetParser/Triple.h
```
This file exists, it's just a path problem

### Environment

- Operating System
```
Ubuntu20.04.5 LTS
```
- TVM version
```
commit aa927df6b3e33856998cb3a4938d8fb4cbdbcec9 (HEAD -> main, origin/main, origin/HEAD)
Author: Sergey <88086617+shtinsa@users.noreply.github.com>
Date: Thu Feb 16 11:12:27 2023 +0300
```
- llvm version
```
commit 1b43650c419fd7448af1a9a097e062d1f64ca485 (grafted, HEAD -> main, origin/main, origin/HEAD)
Author: Vitaly Buka <vitalybuka@google.com>
Date: Tue Feb 21 18:28:06 2023 -0800
```
### Solution
```shell
**diff --git a/src/target/llvm/codegen_blob.cc b/src/target/llvm/codegen_blob.cc**
**index b67aac480..1e9084c81 100644**
**--- a/src/target/llvm/codegen_blob.cc**
**+++ b/src/target/llvm/codegen_blob.cc**
@@ -26,7 +26,7 @@
 #include <llvm/ADT/SmallVector.h>
[...]

PR #14235, opened 2023-03-08
[LLVM][BugFix] Fix include Triplet.h bug when LLVM version>= 17

yanggg1997:
In the latest version of [llvm-project](https://github.com/llvm/llvm-project), the `llvm/ADT/Triple.h` has been moved to `llvm/TargetParser/Triple.h`. Thus, anyone who compile and install llvm from source may encounter this:
```shell
fatal error: llvm/ADT/Triple.h: No such file or directory
   30 | #include <llvm/ADT/Triple.h>
```
This error can be solved by adding `LLVM_VERSION_MAJOR` judgement to include correct head file.

#14136

### DIFF
--- a/src/target/llvm/codegen_blob.cc
+++ b/src/target/llvm/codegen_blob.cc
@@ -26,7 +26,11 @@
 
 #include <llvm/ADT/SmallVector.h>
 #include <llvm/ADT/StringRef.h>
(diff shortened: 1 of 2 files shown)


junrushao (approved):
LGTM! Thanks for the patch!

PR #14238, opened 2023-03-08
[Unity][Op] Add repeat, tile, conv2d_transpose, avg_pool2d

Ubospica:
This PR adds these operators:
- `repeat(data: Expr, repeats: int, axis: Optional[int] = None)`
- `tile(data: Expr, repeats: Union[int, Tuple[int], List[int]])`
- `relax.nn.conv2d_transpose`
	- This operator is intended to find the gradient of conv2d w.r.t its input. For details see [pytorch document](https://pytorch.org/docs/stable/generated/torch.nn.ConvTranspose2d.html), and the document in `python/tvm/relax/op/nn/nn.py`.
	- Now TOPI support of conv2d_transpose is limited. It does not support dilations; layouts other than default; symbolic `output_padding`.
- `relax.nn.avg_pool2d`

cc @tqchen @MasterJH5574


yongwww (commented):
Thanks for the great effort! Looks very good to me

tqchen (approved):

PR #14239, opened 2023-03-08
[SVE] Adding codegen tests for SVE

neildhickey:
Now that ci_arm contains LLVM 15, add tests to check that code for SVE can be correctly generated.

Add tests to cover a selection of arithmetic ops and gather loads. The gather load is currently xfailed until work is completed to enable them through TVM and LLVM.

### DIFF
--- a/tests/python/unittest/test_target_codegen_aarch64.py
+++ b/tests/python/unittest/test_target_codegen_aarch64.py
@@ -1,0 +1,545 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+import tvm
+from tvm import te
+from tvm.script import tir as TIR
+import re
+import os
+import ctypes
+import pytest
+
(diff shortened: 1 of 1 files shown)


tests/python/unittest/test_target_codegen_aarch64.py
```diff
@@ -0,0 +1,545 @@
...
+import re
+import os
+import ctypes
+import pytest
+
+from tvm.target.codegen import llvm_version_major
+
+
+@pytest.mark.skipif(
+    llvm_version_major() < 15, reason="Test requires an LLVM version of at least 15 to target SVE"
```
Mousius, line 29: What happens if we upgrade this in a non-SVE enabled environment?

tests/python/unittest/test_target_codegen_aarch64.py
```diff
@@ -0,0 +1,545 @@
...
+    check_correct_assembly("float")
+    check_correct_assembly("float16")
+    check_correct_assembly("uint8")
+    check_correct_assembly("uint16")
+    check_correct_assembly("uint32")
+    check_correct_assembly("uint64")
+    check_correct_assembly("int8")
+    check_correct_assembly("int16")
+    check_correct_assembly("int32")
+    check_correct_assembly("int64")
```
ashutosh-arm, line 62: I feel like there is a reason for not parameterizing these tests with dtype. I can't seem to recall now :thinking:

tests/python/unittest/test_target_codegen_aarch64.py
neildhickey: These are not execution tests, so as long as it have at least LLVM 15, these should still run fine

tests/python/unittest/test_target_codegen_aarch64.py
neildhickey: Do you mean using @pytest.mark.parametrize for the dtype? Yes we could actually, thanks for pointing that out. Originally I'd tried to make use of python generics to allow different types but this didn't work within TVM, but I will make this improvement to the code.

ashutosh-arm (approved):
LGTM! Thanks @neildhickey.

leandron (approved):
LGTM, thanks @neildhickey @ashutosh-arm @Mousius

PR #14242, opened 2023-03-08
[Unity][Transform] Automatic Mixed Precision

spectrometerHBH:
This PR adds a new pass ToMixedPrecision to automatically cast fp32 models to fp16 when necessary.

see https://github.com/spectrometerHBH/tvm/blob/amp/src/relax/transform/to_mixed_precision.cc#L51 on how this pass works.


src/relax/transform/to_mixed_precision.cc
```diff
@@ -0,0 +1,538 @@
...
+ * Formally, they are of the form:
+ *
+ * var = Call(Op, [args], attrs)
+ *
+ * where Op is a specific op we want to rewrite, and attrs is the attributes of the op.
+ * var and args are all exprs with type Tensor or Tuple of Tensors. They might
+ * be vars, constants, or Tuple of vars and constants.
+ * Depending on the properties of the op, we may have 3 different ways to rewrite it:
+ *
+ * 1. kAlways: Always cast the args to fp16
```
yongwww, line 65: Probably we will lose some accuracy with `kAlways`, I am wondering if it is possible to allow user to control the `MixedPrecisionPolicyKind` in python side ToMixedPrecision(kind=kalways/kfollow/auto/never)

spectrometerHBH: That's in principle possible. Here the pass is meant to serve as similar things like `torch.autocast("cuda")`, where it will always cast the inputs of matmul/conv2d to fp16 from fp32.

junrushao (approved):
LGTM! Please fix merge conflicts

tests/python/relax/test_transform_to_mixed_precision.py
```diff
@@ -0,0 +1,540 @@
...
+                    lv,
+                    lv1,
+                    strides=[1, 1],
+                    padding=[0, 0, 0, 0],
+                    dilation=[1, 1],
+                    groups=1,
+                    data_layout="NCHW",
+                    kernel_layout="OIHW",
+                    out_layout="NCHW",
+                    out_dtype="float32",
```
masahi, line 62: What's the reasoning for always fixing the output dtype to be fp32? This is different from the Relay pass behavior.

masahi: ok but `out_dtype = fp16` case should be tested, since it affects where casts are inserted.

spectrometerHBH: Sure

PR #14246, opened 2023-03-09
[CI][ETHOSN] Add ssh to the driver stack installation

lhutton1:
Some use cases of the NPU integration rely on ssh being available within the docker image in order to launch an RPC server on a remote device. This commit adds ssh as a dependency on the NPU driver stack installation.

cc @ashutosh-arm @Liam-Sturge

### DIFF
--- a/docker/install/ubuntu_install_ethosn_driver_stack.sh
+++ b/docker/install/ubuntu_install_ethosn_driver_stack.sh
@@ -47,7 +47,8 @@ apt-install-and-clear -y \
     python-dev \
     python3 \
     scons \
-    wget
+    wget \
+    openssh-client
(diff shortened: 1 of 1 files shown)


lhutton1:
There is an argument that this would be useful for development within the containers more generally. Wondering if the "[CI]" folks had any thoughts?

ashutosh-arm:
@driazati to confirm if its okay to move this to `ubuntu_install_core.sh`

ashutosh-arm:
Submitting this now. We can take up other changes in a follow up if needed.

ashutosh-arm (approved):
Thanks @lhutton1 :smile:

PR #14247, opened 2023-03-09
[Unity][Frontend] Import `tanh` and fix `layer_norm`

SiriusNEO:
This PR provides some quick fixes for fx_translator to import `tanh` and fix the error when importing `torch.nn.functional.layer_norm`.


tqchen:
Thanks @SiriusNEO can you add a unit-test about the fix?

MasterJH5574 (approved):

PR #14249, opened 2023-03-09
[Utils] Allow classmethod and staticmethod in TVMDerivedObject

Lunderberg:
Instance methods that exist in the user-defined class but not in the TVM base are forward using `__getattr__`.  However, this is only applied for attribute look of instances, and doesn't apply for attribute lookup on the class object itself, such as when calling a classmethod or staticmethod.

This commit exposes class methods and static methods in the wrapper class, if they are defined in the user-defined subclass.


junrushao (approved):
LGTM! Thanks for enhancing the utility!

PR #14250, opened 2023-03-09
[TVMScript] Use new variable frame in If/Then/Else

Lunderberg:
Previously, while TVMScript introduces a new scope for other contexts (e.g. `for`, `while`, `with`, etc), the `if` and `else` blocks did not introduce a new scope.  This caused erroneous parsing errors if the `if` and `else` blocks each contained a variable with the same name.  Added a `self.var_table.with_frame()` context resolves this issue.


Hzfengsy:
cc @junrushao @wrongtest-intellif @lightzhan-intellif

junrushao (approved):
Looks like a nice enhancement! LGTM. @wrongtest-intellif could you take a look?

junrushao:
I do believe this is good to merge. @wrongtest-intellif what do you think?

junrushao:
Thanks @Lunderberg!!

PR #14251, opened 2023-03-09
[TIR][Utility] More flexible tir::Substitute arguments

Lunderberg:
Previously, the `tir::Substitute` method had overloads that supported a few ways of providing the variable map (e.g. `const Map<Var,PrimExpr>&`, `std::unordered_map<const VarNode*, PrimExpr>&`, etc.), delegating out to the overload that uses `std::function<Optional<PrimExpr>(const Var&)>`.  However, the types supported for the variable map depended on the type being substituted (e.g. only supporting `const Map<Var,PrimExpr>&` with substituting into a `Array<Range>`), which would be unexpected to new developers.

This PR makes the `tir::Substitute` utility more uniform in the arguments that it accepts.

* For any type that is supported by `tir::Substitute`, `Array<T>` is also supported.

* Any variable mapping type can be used with any substitution type. All variable mapping types are normalized to `std::function<Optional<PrimExpr>(const Var&)>`.

* For `Map` and `std::unordered_map` arguments, the value type may be any subclass of `PrimExpr` (e.g. `Map<Var, Var>` instead of `Map<Var, PrimExpr>`).  Previously, the calling scope needed to either construct a temporary map that returned `PrimExpr`, or to use a broader value type in the map than otherwise required.

The initial and primary goal was to allow a `Map<Var, Var>` to be used as an argument to `tir::Substitute`, rather than a `Map<Var, PrimExpr>`, and making the utility more general was more straightforward than adding multiple overloads specificall for `Map<Var, Var>`.

### DIFF
--- a/include/tvm/tir/stmt_functor.h
+++ b/include/tvm/tir/stmt_functor.h
@@ -368,44 +368,134 @@ TVM_DLL PrimExpr Substitute(PrimExpr expr, std::function<Optional<PrimExpr>(cons
 
 /*!
  * \brief Substitute the var specified by vmap.
- * \param region The object whose vars are to be substituted
- * \param vmap The map of new values.
+ * \param arr The array of Stmt/PrimExpr to be substituted
+ * \param vmap returns a new value if re-mapping is needed, otherwise returns nullptr.
  * \return The result.
- */
-TVM_DLL Array<Range> Substitute(const Array<Range>& region, const Map<Var, PrimExpr>& vmap);
-
-/*!
- * \brief Sugar for substitute via a given map.
- * \param input The input to be updated.
- * \param value_map The map of new values.
- * \return The result.
- * \tparam T the input type, can be PrimExpr or Stmt.
  */
 template <typename T>
-inline auto Substitute(T input, const Map<Var, PrimExpr>& value_map) {
-  auto vmap = [&](const Var& var) -> Optional<PrimExpr> {
-    auto it = value_map.find(var);
-    if (it != value_map.end()) return (*it).second;
-    return Optional<PrimExpr>(nullptr);
-  };
-  return Substitute(std::move(input), vmap);
+Array<T> Substitute(const Array<T>& arr, std::function<Optional<PrimExpr>(const Var& var)> vmap) {
+  return arr.Map([&vmap](const auto& elem) { return Substitute(elem, vmap); });
 }
 
 /*!
- * \brief Sugar for substitute via a given map.
- * \param input The input to be updated.
- * \param value_map The map of new values.
- * \return The result.
- * \tparam T the input type, can be PrimExpr or Stmt.
+ * \brief Substitute the vars specified by vmap.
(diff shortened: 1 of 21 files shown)


include/tvm/tir/stmt_functor.h
```diff
@@ -368,44 +368,134 @@ TVM_DLL PrimExpr Substitute(PrimExpr expr, std::function<Optional<PrimExpr>(cons
...
+}
+
+/*!
+ * \brief Substitute the vars specified by vmap.
+ *
+ * Delegates to the Substitute methods that use std::function.
+ *
+ * \param obj The object in which TIR variables should be substituted
+ * \param vmap Map defining the TIR variables to be replaced
+ * \return The modified Range.
```
kparzysz-quic, line 415: \return The modified object (not Range).

include/tvm/tir/stmt_functor.h
```diff
@@ -368,44 +368,134 @@ TVM_DLL PrimExpr Substitute(PrimExpr expr, std::function<Optional<PrimExpr>(cons
...
- * \param value_map The map of new values.
- * \return The result.
- * \tparam T the input type, can be PrimExpr or Stmt.
+ * \brief Substitute the vars specified by vmap.
+ *
+ * Delegates to the Substitute methods that use std::function.
+ *
+ * \param obj The object in which TIR variables should be substituted
+ * \param vmap Map defining the TIR variables to be replaced
+ * \return The modified Range.
```
kparzysz-quic, line 437: Same here: Range -> object.

include/tvm/tir/stmt_functor.h
```diff
@@ -368,44 +368,134 @@ TVM_DLL PrimExpr Substitute(PrimExpr expr, std::function<Optional<PrimExpr>(cons
...
+}
+
+/*!
+ * \brief Substitute the vars specified by vmap.
+ *
+ * Delegates to the Substitute methods that use std::function.
+ *
+ * \param obj The object in which TIR variables should be substituted
+ * \param vmap Map defining the TIR variables to be replaced
+ * \return The modified Range.
```
kparzysz-quic, line 459: Same here.

include/tvm/tir/stmt_functor.h
```diff
@@ -368,44 +368,134 @@ TVM_DLL PrimExpr Substitute(PrimExpr expr, std::function<Optional<PrimExpr>(cons
...
+}
+
+/*!
+ * \brief Substitute the vars specified by vmap.
+ *
+ * Delegates to the Substitute methods that use std::function.
+ *
+ * \param obj The object in which TIR variables should be substituted
+ * \param iter_vmap Map defining the TIR variables to be replaced
+ * \return The modified Range.
```
kparzysz-quic, line 481: And one last time...

kparzysz-quic (approved):
LGTM, pending some comment fixes.

Lunderberg:
Thank you, and corrected the copy/paste errors in the docstring.

PR #14252, opened 2023-03-09
[Unity][BYOC] Add conv2d and residual block patterns for Relax cutlass BYOC

yelite:
This PR adds cutlass BYOC pattern for conv2d and conv2d residual block. 

In additional to the new patterns in the cutlass pattern table, this PR:
1. Remove the assumption on fused function arg order from the conv2d codegen.
2. Restructure how the cutlass pattern table is constructed, making it more readable.

The check logic at https://github.com/apache/tvm/blob/main/python/tvm/relay/op/contrib/cutlass.py#L200 isn't implemented in this PR yet because it requires non-trivial change to the interface of FuseOpsByPattern. I will send a separate PR to do this.

cc @vinx13 @masahi

### DIFF
--- a/python/tvm/relax/backend/contrib/cutlass.py
+++ b/python/tvm/relax/backend/contrib/cutlass.py
@@ -26,9 +26,10 @@ from tvm.relax.dpl import CallPattern, DFPattern
 
 from ..pattern_registry import get_patterns_with_prefix, register_patterns
 from ..patterns import (
+    make_attention_pattern,
     make_fused_bias_activation_pattern,
     make_matmul_pattern,
-    make_attention_pattern,
+    make_residual_block_pattern,
 )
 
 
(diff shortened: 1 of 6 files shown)


python/tvm/relax/backend/contrib/cutlass.py
```diff
@@ -51,21 +52,59 @@ def _is_supported_dtype(lhs_dtype, rhs_dtype):
...
+        data_layout != "NHWC"
+        or kernel_layout != "OHWI"
+        or not _is_supported_dtype(data.struct_info.dtype, weight.struct_info.dtype)
+    ):
+        return False
+
+    # pylint: disable=invalid-name
+    IC = data.struct_info.shape.values[3]
+    OC = weight.struct_info.shape.values[0]
+    # not depthwise conv2d
```
masahi, line 95: We do support depthwise now cc @vinx13

vinx13: the codegen part for depthwise is still missing

vinx13 (approved):

masahi (approved):

PR #14261, opened 2023-03-09
[Unity][TVMScript] Expose CallNode::attrs for CallTIR

Lunderberg:
Previously, the `attrs` field exists within the `relax::CallNode` object, and could be accessed through the Python FFI, but could neither be defined through TVMScript nor shown using the TVMScript printer.  This commit updates the parser and printer for a keyword argument `attrs` in `R.call_tir`.

### DIFF
--- a/python/tvm/relax/op/base.py
+++ b/python/tvm/relax/op/base.py
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # pylint: disable=redefined-builtin
 """The base Relax operators."""
-from typing import Union, List, Tuple, Optional
+from typing import Union, List, Tuple, Optional, Dict
 
 
 import tvm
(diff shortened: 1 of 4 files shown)


tqchen:
Because call_tir follows a special semantics, we do not apply attrs to call_tir in general for now. This would help to reduce the complexity in analyzing the related pieces. 

Most of the attrs are hidden under the more pythonic kwargs API that can be explicitly passed by the user

Hzfengsy:
Hi @Lunderberg, thanks for the PR.

As @tqchen mentioned, `CallTIR` is designed without attrs since it's a particular semantics. I'm curious about the use case of CallTIR attrs

Lunderberg:
Thank you, and that's what I was less certain on, in terms of the intended use of `CallNode::attrs`.

In this case, the attributes were used as a way to separate out an analysis pass which planned a series of layout changes that would be applied for each TIR PrimFunc, and an application pass which performed those changes.  While that use case is going to be obsoleted by a more general implementation that @psrivas2 is working on, I wanted to break out any useful utilities from that dev branch.

psrivas2:
Does this PR need to be closed or waiting for input from someone? @Lunderberg

Lunderberg:
@psrivas2 Thank you for checking.  If there's feedback on whether this utility would be useful in other cases, that would be helpful.  If it would be useful elsewhere, I can rebase onto the most recent unity.  Otherwise, I can re-implement my use case using the `PrimFunc::attrs` inside the callee instead of the `CallNode::attrs` of the caller.

PR #14265, opened 2023-03-09
[Unity][Transform] Enhance RewriteDataflowReshape transform

spectrometerHBH:
This pr enhances the current RewriteDataflowReshape transformation.

Originally, it will use loop vars to prove the equality of addresses of LHS and RHS, which will cause some cases to fail due to the limitation of the arith module.

Instead, we can just use block vars to do the proof, which is supposed to be equivalent and the expressions are simpler, which allow us to cover more cases.

### DIFF
--- a/tests/python/relax/test_transform_rewrite_dataflow_reshape.py
+++ b/tests/python/relax/test_transform_rewrite_dataflow_reshape.py
@@ -137,8 +137,63 @@ def test_reshape_pattern_detect():
     @tvm.script.ir_module
     class Module:
         @T.prim_func
-        def reshape5(rxplaceholder: T.Buffer((T.int64(2), T.int64(4096), T.int64(320)), "float16"), T_reshape: T.Buffer((T.int64(2), T.int64(4096), T.int64(5), T.int64(64)), "float16")):
-            T.func_attr({"global_symbol": "main", "op_pattern": 2, "tir.noalias": True})
+        def reshape(rxplaceholder: T.Buffer((T.int64(2), T.int64(4096), T.int64(320)), "float32"), T_reshape: T.Buffer((T.int64(2), T.int64(4096), T.int64(5), T.int64(64)), "float32")):
+            for ax0_ax1_ax2_ax3_fused_1 in T.thread_binding(T.int64(256), thread="blockIdx.x"):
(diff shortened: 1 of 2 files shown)


tests/python/relax/test_transform_rewrite_dataflow_reshape.py
```diff
@@ -128,6 +128,30 @@ def main(
...
+                            v_ax2 = T.axis.spatial(T.int64(5), (ax0_ax1_ax2_ax3_fused_0 * T.int64(262144) + ax0_ax1_ax2_ax3_fused_1 * T.int64(1024) + ax0_ax1_ax2_ax3_fused_2) % T.int64(320) // T.int64(64))
+                            v_ax3 = T.axis.spatial(T.int64(64), (ax0_ax1_ax2_ax3_fused_0 * T.int64(262144) + ax0_ax1_ax2_ax3_fused_1 * T.int64(1024) + ax0_ax1_ax2_ax3_fused_2) % T.int64(64))
+                            T.reads(rxplaceholder[(((v_ax2 * T.int64(64) + v_ax3) // T.int64(320) + v_ax1) // T.int64(4096) + v_ax0) % T.int64(2), ((v_ax2 * T.int64(64) + v_ax3) // T.int64(320) + v_ax1) % T.int64(4096), (v_ax2 * T.int64(64) + v_ax3) % T.int64(320)])
+                            T.writes(T_reshape[v_ax0, v_ax1, v_ax2, v_ax3])
+                            T_reshape[v_ax0, v_ax1, v_ax2, v_ax3] = rxplaceholder[(((v_ax2 * T.int64(64) + v_ax3) // T.int64(320) + v_ax1) // T.int64(4096) + v_ax0) % T.int64(2), ((v_ax2 * T.int64(64) + v_ax3) // T.int64(320) + v_ax1) % T.int64(4096), (v_ax2 * T.int64(64) + v_ax3) % T.int64(320)]
+    # fmt: on
+
+    assert relax.analysis.has_reshape_pattern(Module["reshape5"])
```
yongwww, line 152: Perhaps it needs to apply the transform after the assert:
```python
mod = relax.transform.RewriteDataflowReshape()(Module)
tvm.ir.assert_structural_equal(mod, Module)
```

spectrometerHBH: Sg, will add later.

MasterJH5574 (approved):

PR #14266, opened 2023-03-10
[Unity][Transform] DefaultSchedule pass

spectrometerHBH:
This pr enhances introduces tir.is_scheduled to allow default schedule pass to skip some PrimFuncs.

### DIFF
--- a/include/tvm/tir/function.h
+++ b/include/tvm/tir/function.h
@@ -332,6 +332,13 @@ constexpr const char* kIsGlobalFunc = "tir.is_global_func";
  */
 constexpr const char* kIsHostFunc = "tir.is_host_func";
 
+/*!
+ * \brief Mark the function as scheduled, so the default schedule will pass will skip it.
+ *
+ * Type: Integer
+ */
+constexpr const char* kIsScheduled = "tir.is_scheduled";
+
 }  // namespace attr
 }  // namespace tir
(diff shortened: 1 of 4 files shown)


MasterJH5574 (approved):

python/tvm/relax/transform/transform.py
```diff
@@ -475,18 +475,20 @@ def multiply(
...
        (it will be ignored when an user passes database)
+    default_schedule_fallback : bool
+       whether to fallback to default schedule when no schedule is found in database
     Returns
     -------
     ret : tvm.transform.Pass
         The registered pass
     """
-    return _ffi_api.MetaScheduleApplyDatabase(work_dir)  # type: ignore
+    return _ffi_api.MetaScheduleApplyDatabase(work_dir, default_schedule_fallback)  # type: ignore
```
tqchen, line 491: Let us instead do a compositional pass, so in this case.  We can instead have a flag `warn_missing_entry`.
Which can be set to false.

Then we can compose the two pass together without having to injecting fallback logic into MSApplyDB

tqchen (changes requested):

junrushao (approved):
LGTM!

spectrometerHBH:
@tqchen Would you mind taking another look?

tqchen (approved):

PR #14269, opened 2023-03-10
[Unity] Consider target context for Relay to Relax conversion

ibsidorenko:
Some of Relay passes are target specific (for example, AlterOpLayout). This commit adds target information for Relay -> Relax conversion.


ibsidorenko:
cc @csullivan 
This PR is a part of work that was done for demo preparation. It will be great to merge it, if possible.

PR #14270, opened 2023-03-10
[Relay][Op] Connect existing arm_cpu schedule to relay strategy for concat

ashutosh-arm:
NPU used generic implementation for concatenate
before cpu schedules were made the default fallback
schedules. This leads to performance degradation
as this blocks fusion with nearby ops. This commit
adds Relay op strategy for arm_cpu implementation
which makes it use arm_cpu schedule before cpu one.
Reference: https://github.com/apache/tvm/pull/13775

cc @lhutton1


tests/test_select_implementation.py
```diff
@@ -0,0 +1,56 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
lhutton1, line 1: Nit: perhaps this test could go under `tests/python/relay/strategy/arm_cpu/`?

lhutton1 (commented):
Nice catch @ashutosh-arm! Just one small nit about the test location..

tests/test_select_implementation.py
ashutosh-arm: Since the test checks for cpu and arm_cpu selections, I thought it was best to keep it under relay strategy so that the same test can be reused for multiple targets. Also, in future, the same can be expanded to cover other ops too. wdyt?

lhutton1: Makes sense :)

lhutton1 (approved):
LGTM!

PR #14272, opened 2023-03-11
[TVMC] Improve --desired-layouts functionality

PhilippvK:
this aims to make the `--desired-layout` argument more powerful based on the previously merged changes from #14010 (@srkreddy1238) by introducing two new features:

1. Allow passing multiple arguments to `--desired-layout` instead of only one, to specify one layout per transformed operator specified in `--desired-layout-ops`. (Number of arguments has to bei either 1 or match the number of transformed operators)
2. Optionally, you can now specify a non-default kernel layout as follows: `NHWC:HWIO`

Example Usage: `tvmc compile … --desired-layout nn.max_pool2d qnn.conv2d --desired-layout-ops NCHW NHWC:HWIO`

I also added unit tests for the new use-cases.

### Known Limitations:
* It would make sense to specify individual kernel layouts for regular convolutions and depthwise ones. However since both are usually implemented as generalized `nn.conv2d`, we can not transform them individually. Are there any good workarounds for this?
* The arguments of `--desired-layouts` have previously been checked for validity during cmdline parsing (e.g. only NCHW and NHWC are allowed) which is not possible anymore. Should I add a regular expression for that?

### DIFF
--- a/python/tvm/driver/tvmc/transform.py
+++ b/python/tvm/driver/tvmc/transform.py
@@ -130,8 +130,6 @@ def convert_graph_layout(mod, desired_layouts, ops=None):
         assert isinstance(desired_layouts, str)
         desired_layouts = [desired_layouts]
 
-    assert len(desired_layouts) > 0
-
     if len(desired_layouts) != len(ops):
         if len(desired_layouts) != 1:
             raise TVMCException(
--- a/python/tvm/driver/tvmc/transform.py
+++ b/python/tvm/driver/tvmc/transform.py
@@ -236,7 +236,7 @@ def generate_transform_args(parser):
         help="Change the data/kernel layout of the graph. (i.e. NCHW or NHWC:HWIO)"
         "This option can be provided multiple times to specify per-operator layouts, "
         "e.g. '--desired-layout NHWC:HWIO' (Apply same layout for every operator)."
-        "e.g. '--desired-layout-ops nn.conv2d nn.avg_pool2d --desired-layout NCHW NHWC'."
+        "e.g. '--desired-layout-ops nn.conv2d nn.avg_pool2d --desired-layout NCHW NHWC'.",
     )
     parser.add_argument(
         "--desired-layout-ops",
--- a/python/tvm/driver/tvmc/transform.py
+++ b/python/tvm/driver/tvmc/transform.py
@@ -235,7 +235,7 @@ def generate_transform_args(parser):
     parser.add_argument(
         "--desired-layout",
         nargs="+",
-        help="Change the data/kernel layout of the graph. (i.e. NCHW or NHWC:HWIO)",
+        help="Change the data/kernel layout of the graph. (i.e. NCHW or NHWC:HWIO)"
         "This option can be provided multiple times to specify per-operator layouts, "
         "e.g. '--desired-layout NHWC:HWIO' (Apply same layout for every operator)."
         "e.g. '--desired-layout-ops nn.conv2d nn.avg_pool2d --desired-layout NCHW NHWC'."
--- a/python/tvm/driver/tvmc/transform.py
+++ b/python/tvm/driver/tvmc/transform.py
@@ -113,6 +113,7 @@ def convert_graph_layout(mod, desired_layouts, ops=None):
     desired_layouts : list[str]
         The layouts to convert to.
         Expects either a single element or one str per operator.
+        Can be only data layouts or combination of both, e.g. NHWC:HWIO
     ops : list
         List of operators to be layout converted.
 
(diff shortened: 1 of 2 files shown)


python/tvm/driver/tvmc/transform.py
```diff
@@ -123,7 +125,24 @@ def convert_graph_layout(mod, desired_layout, ops=None):
     if ops is None:
         ops = ["nn.conv2d", "nn.conv2d_transpose", "qnn.conv2d"]
 
-    desired_layouts = {op: [desired_layout, "default"] for op in ops}
+    assert isinstance(desired_layouts, list) and len(desired_layouts) > 0
```
lhutton1, line 128: Shall we still allow a single `str` input to keep backwards compatibility?

python/tvm/driver/tvmc/transform.py
```diff
@@ -210,9 +229,8 @@ def generate_transform_args(parser):
     # AlterLayout
     parser.add_argument(
         "--desired-layout",
-        choices=["NCHW", "NHWC"],
-        default=None,
-        help="Change the data layout of the whole graph.",
+        nargs="+",
+        help="Change the data/kernel layout of the graph. (i.e. NCHW or NHWC:HWIO)",
```
lhutton1, line 233: Would be good to also show an example of a list of layouts when `--desired-layout-ops` is provided

lhutton1 (commented):
Thanks @PhilippvK for exposing this functionality, overall LGTM! Just responding to your questions as well as a couple of nits:

> It would make sense to specify individual kernel layouts for regular convolutions and depthwise ones. However since both are usually implemented as generalized nn.conv2d, we can not transform them individually. Are there any good workarounds for this?

One way that I can think of, although it's not very elegant, is to allow a special `nn.conv2d_depthwise` parameter thorough `desired_layout_ops`, then override the `FTVMConvertOpLayout` function for nn.conv2d before running the pass similar to what is done [here](https://github.com/apache/tvm/blob/main/python/tvm/relay/op/contrib/arm_compute_lib.py#L115). Perhaps we can leave as a "TODO" for now and create a tracking issue for it?

> The arguments of --desired-layouts have previously been checked for validity during cmdline parsing (e.g. only NCHW and NHWC are allowed) which is not possible anymore. Should I add a regular expression for that?

I believe it should be okay to leave this, especially since this will get more complex when you consider the layouts of conv3d etc. Unsupported layouts should raise an error at a later stage while running the ConvertLayout pass anyway.

PhilippvK:
> One way that I can think of, although it's not very elegant, is to allow a special nn.conv2d_depthwise parameter thorough desired_layout_ops, then override the FTVMConvertOpLayout function for nn.conv2d before running the pass similar to what is done [here](https://github.com/apache/tvm/blob/main/python/tvm/relay/op/contrib/arm_compute_lib.py#L115). Perhaps we can leave as a "TODO" for now and create a tracking issue for it?

Thank you for pointing that out. If I understand it correctly, we than need to run the `ConvertLayout` pass twice since we would not be able to pass layouts for both types of convolutions at the same time:
    1. With `FTVMConvertOpLayout` being a no-op for **depthwise** convolutions and desired layout for `nn.conv2d`
    2. With `FTVMConvertOpLayout` being a no-op for **non-depthwise** convolutions and desired layout for `nn.conv2d`

I agree that this is far away from elegant and we should track this somewhere else...

python/tvm/driver/tvmc/transform.py
PhilippvK: I have updated the help message.

python/tvm/driver/tvmc/transform.py
```diff
@@ -123,7 +125,29 @@ def convert_graph_layout(mod, desired_layout, ops=None):
     if ops is None:
         ops = ["nn.conv2d", "nn.conv2d_transpose", "qnn.conv2d"]
 
-    desired_layouts = {op: [desired_layout, "default"] for op in ops}
+    if not isinstance(desired_layouts, list):
+        # For backwards compatibility
+        assert isinstance(desired_layouts, str)
+        desired_layouts = [desired_layouts]
+
+    assert len(desired_layouts) > 0
```
lhutton1, line 133: Nit: This check seems redundant to me, I think if no layouts are provided the convert layout pass won't do anything - would it be okay to check?

lhutton1:
Thanks for the updates, LGTM after fixing the test failures :) 

I believe the `ConvertLayout` pass would still only need to be applied once. The `FTVMConvertOpLayout` function can be run on each `nn.conv2d` the `ConvertLayout` pass encounters, check whether the current convolution is normal or depthwise, then use the desired layout specified for `nn.conv2d_depthwise` if the convolution is depthwise and the layout for `nn.conv2d_depthwise` has been specified. 

Before passing the `desired_layouts` dictionary to the `ConvertLayout` pass itself, we can remove the special `nn.conv2d_depthwise` value. 

Apologies for the rushed explanation, happy to help explain further in a separate conversation

python/tvm/driver/tvmc/transform.py
PhilippvK: It is in fact redundant, especially since I use `nargs='+'`.

I will drop it when fixing the CI bugs

lhutton1 (approved):
LGTM! Thanks for the work on this @PhilippvK, I'll leave open for a day or so to give others a change to comment

PhilippvK:
@lhutton1 Should I rebase/squash the commits before merge?

lhutton1:
No need to rebase for now, and no need to squash the commits as this will be done automatically when merging

lhutton1:
Thanks @PhilippvK!

PhilippvK:
>  Perhaps we can leave as a "TODO" for now and create a tracking issue for it?

@lhutton1 Followup discussion: https://discuss.tvm.apache.org/t/convertlayout-dealing-with-depthwise-convolutions/14581

PR #14273, opened 2023-03-11
[TVMC] Expose tir and relay.qnn passes to --disable-pass argument

PhilippvK:
I recently tried to disable the `qnn.Legalize` pass on the TVMC command line and learned, that the list of passes is actuallt not complete. It lacks:
* Custom relay passes which not not follow the `relay._transform.` scheme such as `relay.qnn._transform.Legalize`
* Any TIR passes. Is this intentional, i.e. to keep complexity down and not overwhelm users?

My patch fixes this by automatically mapping user-provided path names to fully-qualified TVM passes as follows:
```python
replacements = {
    "tir.": "tir.transform.",
    "qnn.": "relay.qnn._transform.",
    "": "relay._transform.",
}
```

A small unit test for this use case was added. 

**Example usage:** `tvmc compile ... --disabled-pass tir.UnrollLoop,qnn.Legalize`

### Discussion
* To not break the compatibility with the older interface, I still use an empty prefix for the `relay._transform.` passes which is inconsistent with the QNN & TIR ones. Should I allow `relay.ABC` inputs instead and deprecate the unprefixed version instead?
* Having these non-trivial mappings might lead to users struggling how to find the expected names for these passes. Allowing to specify the fully-qualified pass name (i.e. `relay._transform.RemoveUnusedFunctions`) would be help but also blow up the number of allowed passes which would be printed in case of a failed lookup in `pass_list.py`.
* A number of passes is currently not exposed automatically for two reasons:
    ```
    relay.transform.LiftConstants
    relay.transform.RecoverVirtualDeviceMap
    relay.transform.ManifestAlloc
    relay.transform.MemoryPlan
    tir.usmp.transform.ConvertPoolAllocationsToOffsets
    tir.usmp.transform.CreateAllocatesForIO
    tir.usmp.transform.AssignPoolInfo
    ```
  1. Using `relay.transform.` as prefix instead of `relay._transform.`. Can anyone explain why?
  2. Mapping `tir.usmp.transform.` -> `tir.usmp.` is currently not implemented. Should I add it?

### DIFF
--- a/python/tvm/driver/tvmc/pass_list.py
+++ b/python/tvm/driver/tvmc/pass_list.py
@@ -50,7 +50,7 @@ def parse_pass_list_str(input_string):
                 else:
                     p = p.replace(short, long, 1)
                 break
-            elif reverse and p.startswith(long):
+            if reverse and p.startswith(long):
                 p = p.replace(long, short, 1)
                 break
         return p
--- a/python/tvm/driver/tvmc/pass_list.py
+++ b/python/tvm/driver/tvmc/pass_list.py
@@ -35,16 +35,37 @@ def parse_pass_list_str(input_string):
     -------
     list: a list of existing passes.
     """
-    _prefix = "relay._transform."
+    replacements = {
+        "tir.": "tir.transform.",
+        "qnn.": "relay.qnn._transform.",
+        "": "relay._transform.",
+    }
+
+    def apply_prefix(p, reverse=False):
+        p = p.strip()
+        for short, long in replacements.items():
+            if not reverse and p.startswith(short):
+                if len(short) == 0:
+                    p = long + p
+                else:
+                    p = p.replace(short, long, 1)
+                break
+            elif reverse and p.startswith(long):
+                p = p.replace(long, short, 1)
+                break
+        return p
+
     pass_list = input_string.split(",")
     missing_list = [
-        p.strip()
+        p
         for p in pass_list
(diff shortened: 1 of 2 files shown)


lhutton1:
Thanks for looking into fixing this @PhilippvK, overall I think it would be very useful to expose additional passes to the disabled pass option. Here are my thoughts:

Allowing relay.ABC makes sense to me, but I think there could be a deeper design problem here. I would argue that we can remove the prefix all together and use the pass names exactly as they are defined in the code. Personally, as a developer wanting to disable a particular pass, I'd find the pass name in the code, then expect to use this same name with the `--disable-pass` option. 

Also, while I don't believe it's the case currently, we could have a scenario where a replaced pass name results in a collision e.g. `tir.my.namespace.ABC` and `tir.my.namespace2.ABC` -> `tir.ABC`.

Passes with odd names such as `*._transform.*` which the prefix mechanism seems to work around can always be renamed in the codebase so they as more sensible.

I think this helps solve some of the discussion questions you listed, happy to hear other thoughts and opinions on this :)

PhilippvK:
Thank you for your feedback.


Two notes:
- We already have such situation where a Pass with the same name exists in two different Namespaces: `relay._transform.Legalize` vs. `relay.qnn._transform.Legalize`. We either need to find a way to allow both of them or need rename the latter to `relay.qnn._transform.QNNLegalize` which might break downstream implementations…
- The `relay._transform.` prefix is actually the most common here while `relay.transform.` seems to be odd. I have no idea why these two namings exists but would agree that these should be patched for more consistency.

PR #14274, opened 2023-03-11
[Unity][Transform] SplitCallTIRByPattern and CUTLASS backend

spectrometerHBH:
Currently, the BYOC system is based on op-level pattern matching, this PR intends to provide primary support for TIR-level pattern matching based on backend registration and dispatching.

For now, it simply matches the first set of for loops in PrimFunc.

Co-authored-by: Hongyi Jin (@jinhongyii)

### DIFF
--- a/python/tvm/contrib/cutlass/gemm_operation.py
+++ b/python/tvm/contrib/cutlass/gemm_operation.py
@@ -369,7 +369,8 @@ def instantiate_gemm_template(attrs):
             {
                 "bias_decl": "void* ptr_bias = (void*)(${bias_arg}->data);\n",
                 "ptr_c": "ptr_bias",
-                "c_stride": "${bias_arg}->ndim == 1 ? 0 : " + attrs["ldc"],
+                "c_stride": "(${bias_arg}->ndim == 1 || ${bias_arg}->shape[0] == 1) ? 0 : "
+                + attrs["ldc"],
             }
         )
     else:
--- a/python/tvm/relax/transform/transform.py
+++ b/python/tvm/relax/transform/transform.py
@@ -693,6 +693,25 @@ def ToMixedPrecision(out_dtype="float32") -> tvm.ir.transform.Pass:
     return _ffi_api.ToMixedPrecision(out_dtype)  # type: ignore
 
 
+def SplitCallTIRByPattern(patterns, fcodegen) -> tvm.ir.transform.Pass:
+    """Split a PrimFunc into 2 parts: the first part is a TIR PrimFunc which is
+       matched with some pattern, and the second part is the rest of the original
+       PrimFunc. It will call fcodegen to generate the code for the matched pattern
+       to replace it with a ExternFunc call.
+    Parameters
+    ----------
+    patterns : List[PrimFunc]
+        The list of patterns to match.
+    fcodegen: Callable[[List[MatchResult]], List[Object]]
+        The function to generate the code for the matched patterns.
+    Returns
+    -------
+    ret : tvm.transform.Pass
+        The registered pass for splitting call_tir.
+    """
+    return _ffi_api.SplitCallTIRByPattern(patterns, fcodegen)  # type: ignore
+
+
 def _wrap_class_function_pass(pass_cls, pass_info):
     """Wrap a python class as function pass."""
 
--- a/tests/python/relax/test_codegen_tir_cutlass.py
+++ b/tests/python/relax/test_codegen_tir_cutlass.py
@@ -1,0 +1,755 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+from __future__ import annotations
+import tempfile
+
+from tvm import relax, runtime
+import tvm
+import tvm.testing
+from tvm import relax
+import scipy
+from scipy.special import erf
+import numpy as np
+from tvm.target import Target
+from tvm.relax.vm_build import build as relax_build
+from tvm.script.ir_builder import relax as R
+from tvm.script.ir_builder import ir as I
+from tvm.script.ir_builder import tir as T
+from tvm.script.ir_builder import IRBuilder
+
+from tvm.relax.backend_tir import get_tir_pattern
+from tvm.relax.backend_tir.contrib.cutlass import cutlass_fcodegen, compile_options
+
+A_TYPE = "float16"
+B_TYPE = "float16"
+C_TYPE = "float16"
+
+target = Target("cuda")
+
+
+def f_run(rt_mod: runtime.Module, device: runtime.ndarray.Device, *input):
+    vm = relax.vm.VirtualMachine(rt_mod=rt_mod, device=device)
+    return vm["main"](*input)
+
+
+def build(mod):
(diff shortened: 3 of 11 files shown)


tests/python/relax/test_codegen_tir_cutlass.py
```diff
@@ -0,0 +1,767 @@
...
+                    np.transpose(result.numpy(), (0, 3, 1, 2)),
+                    D_torch.cpu().numpy(),
+                    rtol=5e-2,
+                    atol=5e-2,
+                )
+                counter += 1
+
+
+if __name__ == "__main__":
+    test_cutlass_dense()
```
yongwww, line 754: tvm.testing.main()

python/tvm/relax/transform/transform.py
```diff
@@ -681,6 +681,25 @@ def ToMixedPrecision(out_dtype="float32") -> tvm.ir.transform.Pass:
     return _ffi_api.ToMixedPrecision(out_dtype)  # type: ignore
 
 
+def SplitCallTIRByPattern(patterns, fcodegen) -> tvm.ir.transform.Pass:
```
yongwww, line 684: looks the transformation is for tir only, it would be good to add it (and `MatchResult`) into folder like python/tvm/tir/transform

yongwww: oh i see, it should be fine in relax transform then. The distinction between the two is becoming blurred.

python/tvm/contrib/cutlass/gemm_operation.py
```diff
@@ -369,7 +369,8 @@ def instantiate_gemm_template(attrs):
             {
                 "bias_decl": "void* ptr_bias = (void*)(${bias_arg}->data);\n",
                 "ptr_c": "ptr_bias",
-                "c_stride": "${bias_arg}->ndim == 1 ? 0 : " + attrs["ldc"],
+                # "c_stride": "${bias_arg}->ndim == 1 ? 0 : " + attrs["ldc"],
```
jinhongyii, line 372:
```suggestion

```

jinhongyii (approved):
LGTM!

junrushao:
CC @yongwww please take another look :-)

python/tvm/contrib/cc.py
```diff
@@ -231,6 +233,20 @@ def _linux_compile(output, objects, options, compile_cmd, compile_shared=False):
...
+
+        compute_version = "".join(
+            get_target_compute_version(Target.current(allow_none=True)).split(".")
+        )
+        cmd += ["-gencode", f"arch=compute_{compute_version},code=sm_{compute_version}"]
+        cmd += ["-O3"]
+        cmd += ["-std=c++17"]
+        cmd += ["-Xcompiler=-fPIC"]
+        cmd += ["-Xcompiler=-fno-strict-aliasing"]
+        cmd += ["-t 24"]
```
FrozenGene, line 245: Should add the guard of CUDA version check >= 11.2 and to get the real cpu cores: refer this:https://github.com/apache/tvm/blob/main/python/tvm/contrib/cutlass/build.py#L72

python/tvm/contrib/cc.py
```diff
@@ -231,6 +233,20 @@ def _linux_compile(output, objects, options, compile_cmd, compile_shared=False):
...
+        cmd += ["-DCUTLASS_ENABLE_TENSOR_CORE_MMA=1"]
+        cmd += ["-Xcompiler=-Wconversion"]
+        cmd += ["--use_fast_math"]
```
FrozenGene, line 248: doesn't make sense we enable fast math by default. We should control it using one option and should set it false by default.

python/tvm/contrib/cc.py
spectrometerHBH: Sure. Will do.

tests/python/relax/test_codegen_tir_cutlass.py
spectrometerHBH: Sure. Will do.

MasterJH5574 (approved):

tests/python/relax/test_codegen_tir_cutlass.py
```diff
@@ -0,0 +1,754 @@
...
+    A_tvm = tvm.nd.array(A, dev)
+    B_tvm = tvm.nd.array(B, dev)
+    bias_tvm = tvm.nd.array(bias, dev)
+    result = f_run(executable, dev, A_tvm, B_tvm, bias_tvm)
+    C = A @ B + bias
+    O = 0.5 * C * (1 + erf(C / np.sqrt(2)))
+    np.testing.assert_allclose(result.numpy(), O, rtol=5e-2, atol=5e-2)
+
+
+# def constructBatchGEMM_bias2_mul(batch, M, N, K):
```
vinx13, line 403: remove this

python/tvm/_ffi/libinfo.py
```diff
@@ -205,11 +205,18 @@ def find_include_path(name=None, search_path=None, optional=False):
         tvm_include_path = [os.path.join(p, "include") for p in header_path]
         dlpack_include_path = [os.path.join(p, "dlpack/include") for p in header_path]
         dmlc_include_path = [os.path.join(p, "dmlc-core/include") for p in header_path]
-
+        if use_nvcc:
+            cutlass_include_path = [os.path.join(p, "cutlass/include") for p in header_path]
```
vinx13, line 209: is it possible to make it configurable and move this to some cutlass specific places?

python/tvm/contrib/cc.py
```diff
@@ -231,6 +234,21 @@ def _linux_compile(output, objects, options, compile_cmd, compile_shared=False):
...
+        )
+        cmd += ["-gencode", f"arch=compute_{compute_version},code=sm_{compute_version}"]
+        cmd += ["-O3"]
+        cmd += ["-std=c++17"]
+        cmd += ["-Xcompiler=-fPIC"]
+        cmd += ["-Xcompiler=-fno-strict-aliasing"]
+        cuda_ver = get_cuda_version()
+        if cuda_ver >= (11, 2):
+            cmd += ["-t " + str(multiprocessing.cpu_count())]
+        cmd += ["-DCUTLASS_ENABLE_TENSOR_CORE_MMA=1"]
```
vinx13, line 249: these should be passed through `options`

python/tvm/contrib/cutlass/gemm_operation.py
vinx13: is it needed?

vinx13:
cc @masahi

tests/python/relax/test_codegen_tir_cutlass.py
spectrometerHBH: I put it here in purpose since I expect the batch matmul+bias+residual to be supported

python/tvm/contrib/cc.py
spectrometerHBH: Sure. Will do.

python/tvm/contrib/cutlass/gemm_operation.py
spectrometerHBH: I see. So matmul+add(residual) is implemented using bias now?

tests/python/relax/test_codegen_tir_cutlass.py
```diff
@@ -0,0 +1,754 @@
...
+                    R.output(C)
+                (C,) = df.output_vars
+                R.func_ret_value(C)
+    mod = ib.get()
+    return mod
+
+
+@tvm.testing.requires_cutlass
+def test_cutlass_conv2d():
+    import torch
```
masahi, line 525: we shouldn't depend on torch for testing

spectrometerHBH: I see.

python/tvm/contrib/cutlass/gemm_operation.py
spectrometerHBH: That's kind of strange here. Why don't we use beta=0 and residual epilogue, or Am I understanding it incorrectly?

masahi: yes, otherwise there is an ambiguity in which of bias add or residual add should be used for `matmul + add` pattern. This causes a subtle issue for `FusesOpsByPattern`, which applies rewriting over a list of patterns which should be ordered by the "size" of a pattern. 

We are using a residual add pattern only for matmul + bias_add + residual add (two add).

spectrometerHBH:
cc @vinx13 @masahi

PR #14275, opened 2023-03-12
[Unity][Contrib] Introduce several features of cutlass profiler

spectrometerHBH:
- allow Conv2d using different alignment factors for input and epilogue, which can influence performance 
- store the profiler cache on disk, reducing CUTLASS profiler overhead across different runs
- use the same set of default tile configurations as CUTLASS for sm80 https://github.com/NVIDIA/cutlass/blob/master/tools/library/scripts/generator.py#L1881


junrushao (approved):
LGTM! Could you fix the linter?

Hzfengsy:
Could you please send the PR directly to `main`? since it's not unity related.

junrushao (approved):

masahi:
Just noticed that this PR was sent to `unity`, but it should have been sent to `main`. @spectrometerHBH please follow up. 

The search for the output alignments added in this PR does nothing to our CUTLASS BYOC (since the C alignment doesn't seem to be used in codegen), but increases the conv2d profiling time by 4x. I was going to fix that on `main` and found that this code only exists on `unity`.

junrushao:
Yeah we should probably send this commit to main as well. The co-existence of the branches does cause some trouble in maintenance

PR #14279, opened 2023-03-12
[TVMScript][Parser] Add more warp-level builtins and `Range`

yzh119:
# Motivation
Several builtins "tvm_storage_sync", "tvm_warp_shuffle", "tvm_warp_shuffle_up", "tvm_warp_shuffle_down", "tvm_warp_activemask" and `Range` will appear in TVMScript printer but are missing in TVMScript parser. This PR fix the behavior.
 
@cyx-6 @junrushao

PR #1050, opened 2018-03-26
[SCHEDULE][PASS] Enable Warp memory and lower to shuffle

tqchen:
This PR enables the warp memory abstraction. ```"warp"``` is a special memory scope that represents shared memory among the threads in the warp.

From the programmer's perspective, warp memory is exactly like shared memory, but it is instead lowered into local registers and its reads becomes shuffle instructions in typical GPUs. See the 

So far cuda works with a simple exmaple. to enable it for other backends

- Create a target with correct warp_size
- always set threadIdx.x's size to be warp size
- Create an intrinsic dispatching rule for tvm_warp_shuffle to dispatch to the right intrinsic

PR #14280, opened 2023-03-12
[TensorIR][Transform] Enable warp shuffling for `LowerWarpMemory`

yzh119:
# Motivation
The `LowerWarpMemory` pass cannot emit `shfl_sync` instructions because of an [internal check](https://github.com/apache/tvm/blob/e3c8f2bc2d9ab3f5ba1c0db0755e455bf59a39b8/src/tir/transforms/lower_warp_memory.cc#L334-L337) introduced in #9727 . Actually if we load value from another lane in the warp, the `local_index` would inevitably carry the warp index, and this case would be disabled by the check.

This PR fix the issue by disabling the check and add an unit test for warp shuffling.

The PR depends on #14279 , I'll rebase to upstream/main after that PR is merged.

@Lunderberg @masahi @tqchen


yzh119:
It seems the unit test still works if I add the `ICHECK`, I'll close the PR first.

junrushao:
Its fine to keep it open as a draft PR :-)

tests/python/unittest/test_tir_transform_lower_warp_memory.py
```diff
@@ -347,5 +348,105 @@ def test_lower_warp_memory_divide_by_factor():
...
+def func(a: T.handle, b: T.handle) -> None:
+    A = T.match_buffer(a, [32], "float32")
+    B = T.match_buffer(b, [32], "float32")
+    for i in range(32):
+        with T.block("warp_shuffle"):
+            vi = T.axis.spatial(32, i)
+            B[vi] = A[(vi % 4) * 8 + vi // 4] + T.float32(1)
+
+
+def test_warp_shuffle_transform():
```
Lunderberg, line 361: The test looks reasonable as-is, though there's also a `tvm.testing.CompareBeforeAfter` that you could use to further reduce the boilerplate.

```python
class TestWarpShuffleTransform(tvm.testing.CompareBeforeAfter):
    transform = tvm.tir.transform.LowerWarpMemory()

    def before(A: T.handle("float32", "global"), B: T.handle("float32", "global")):
        ...

    def expected(A: T.handle("float32", "global"), B: T.handle("float32", "global")):
        ...
```

tests/python/unittest/test_tir_transform_lower_warp_memory.py
```diff
@@ -347,5 +348,105 @@ def test_lower_warp_memory_divide_by_factor():
...
+    @tvm.script.ir_module
+    class Before:
+        @T.prim_func
+        def main(A: T.handle("float32", "global"), B: T.handle("float32", "global")):
+            blockIdx_x = T.env_thread("blockIdx.x")
+            threadIdx_x = T.env_thread("threadIdx.x")
+            T.func_attr(
```
Lunderberg, line 368: It looks like the test case only requires the `"target"` attribute, and only requires `"kind"` and `"thread_warp_size"` within that.  Can we remove the extra attributes from the unit test?

tests/python/unittest/test_tir_transform_lower_warp_memory.py
```diff
@@ -347,5 +348,105 @@ def test_lower_warp_memory_divide_by_factor():
...
+                    "tir.is_global_func": 1,
+                    "tir.noalias": 1,
+                }
+            )
+            T.launch_thread(blockIdx_x, 1)
+            A_warp = T.allocate([32], "float32", "warp")
+            B_warp = T.allocate([32], "float32", "warp")
+            T.launch_thread(threadIdx_x, 32)
+            A_warp_1 = T.Buffer((32,), data=A_warp, scope="warp")
+            A_1 = T.Buffer((32,), data=A)
```
Lunderberg, line 395: Instead of having a separate `A: T.handle` and `A_1: T.Buffer`, the buffer could be declared as a parameter `A_1: T.Buffer(32)`.  It does result in slightly different TIR, as it follows the style from before `MakePackedAPI` is applied, but for a unit test would help to emphasize the change being tested.

tests/python/unittest/test_tir_transform_lower_warp_memory.py
```diff
@@ -347,5 +348,105 @@ def test_lower_warp_memory_divide_by_factor():
...
+            A_warp_1[threadIdx_x] = A_1[threadIdx_x]
+            B_warp_1 = T.Buffer((32,), data=B_warp, scope="warp")
+            T.tvm_storage_sync("warp")
+            B_warp_1[threadIdx_x] = A_warp_1[threadIdx_x % 4 * 8 + threadIdx_x // 4] + T.float32(1)
```
Lunderberg, line 399: Could we add a comment here, indicating that this line is the one that should be updated correctly?

Lunderberg (commented):
Made a couple of comments as I was reading through, though your comment about the tests passing even with the `ICHECK` present is interesting.  It looks like `local_index` in the case of warp shuffle is `0`, and is used to build the `A_warp[0]` argument to `tvm_warp_shuffle`.

Lunderberg:
(Also, it looks like the initial check dates back to [PR#1050](https://github.com/apache/tvm/pull/1050/files#diff-2798beec4533fbfc67324a90dc70ea51b263dbeb56e48e295da335fe512656d8R206), just with different refactorings that touched that line along the way.)

PR #14281, opened 2023-03-12
[Schedule] Add an optional argument `disable_checks` for `Schedule`

yzh119:
# Motivation
Currently, some of the schedule checks are too strict, which makes it hard to schedule some workloads such as FlashAttention whose reduction is two-stage and does not strictly follows our standard.

This PR adds an optional argument `disable_checks` which mutes some checks. The argument defaults to `False` and we can enable it whenever we want to disable some `soft` checks (by `soft` we mean if we violate such checks, the schedule is not necessarily invalid, and if we violate `hard` checks the schedule step is invalid).

In the future, we should collect the `soft` and `hard` checks for all schedule primitives. This PR serves for FlashAttention and only cares `bind` and some reduction primitives for now.

@spectrometerHBH @junrushao @Hzfengsy

### DIFF
--- a/include/tvm/tir/schedule/schedule.h
+++ b/include/tvm/tir/schedule/schedule.h
@@ -779,12 +779,11 @@ class Schedule : public runtime::ObjectRef {
    * \param debug_mask Do extra correctness checking after the class creation
    * and each time after calling the Replace method.
    * \param error_render_level The level of error rendering
-   * \param enable_check Whether to enable prequisite checks for schedule primitives.
+   * \param enable_check Whether to enable some prequisite checks for schedule primitives, it's
+   *   user's duty to guarantee the schedule correctness if we disable the checks.
    * \return The concrete schedule created
    * \sa ScheduleDebugMask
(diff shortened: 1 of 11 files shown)


include/tvm/tir/schedule/schedule.h
```diff
@@ -779,29 +779,33 @@ class Schedule : public runtime::ObjectRef {
    * \param debug_mask Do extra correctness checking after the class creation
    * and each time after calling the Replace method.
    * \param error_render_level The level of error rendering
+   * \param enable_check Whether to enable prequisite checks for schedule primitives.
```
Hzfengsy, line 782: Please add comments saying that it's user's responsibility to ensure the correctness if we disable the checks

yzh119: Sounds good, will do it.

tqchen:
@tvm-bot  rerun

junrushao (approved):
LGTM

PR #14282, opened 2023-03-12
[Unity][Transform] Introduce data-dependent operation of reshape and its constant folding

sunggg:
As discussed in previous Unity open dev meeting, this PR implements data-dependent operation of reshape.

## Representative examples
This PR realizes data-dependent operation of reshape as follows.

```
# Target shape will be computed at runtime and stored in tensor
target_shape: R.Tensor(ndim=1) = ...
# With new 'tensor_to_shape`, we can convert tensor values to ShapeExpr
lv: R.Shape(ndim=2) = R.tensor_to_shape(target_shape)
# Reshape is extended to take a Var as long as it is bound with ShapeExpr
gv: R.Tensor(ndim=2, dtype="float32") = R.reshape(data, lv)
```

Also, `FoldConstant` is extended to support `tensor_to_shape`. 

```
# Before `FoldConstant`: c0, c1 are constants
lv0 = R.add(c0, c0)
target_shape = R.multiply(lv0, c1)
lv2: R.Shape(ndim=2) = R.tensor_to_shape(target_shape)
gv: R.Tensor(ndim=2, dtype="float32") = R.reshape(data, lv2)

# After `FoldConstant`
gv: R.Tensor((16, 16), dtype="float32") = R.reshape(data, R.shape([16, 16]))
```

## Summary of changes
* Introduce new ~~builtin~~ `tensor_to_shape`
* Current reshape takes the target shape in `ShapeExpr | Array[PrimExpr]`. This PR extends this to take `Var` only when it is bound to `ShapeExpr`. 
* Extend `FoldConstant` pass to support `tensor_to_shape`

## 03/18/2023 Update 
Turned out if we implement `target_to_shape` as a builtin, its lowering happens too late and we cannot lower the following `reshape`-like ops since they need at least symbolic shape to legalize. Therefore, `target_to_shape` should be lowered before we lower reshape. New change extends existing `SimplifyNormInference` pass to serve as a generic pass to decompose composite operators like `tensor_to_shape`, `attention`, `erf`, etc. 

### Follow-up PRs
* Formal introduction of composite ops
* Refactor of ConstantFold pass to handle such composite ops


sunggg:
cc @jwfromm @psrivas2 @slyubomirsky  @MasterJH5574

tests/python/relax/test_relax_operators.py
```diff
@@ -193,5 +193,21 @@ def test_op_shape_of():
...
+def test_op_tensor_to_shape():
+    out_shape = run_cpu(
+        TensorToShapeTest, "run_tensor_to_shape", tvm.nd.array(np.array([1, 2, 3]).astype("int64"))
+    )
+    assert out_shape == tvm.runtime.ShapeTuple([1, 2, 3])
+
+
 if __name__ == "__main__":
-    tvm.testing.main()
+    test_op_tensor_to_shape()
```
Hzfengsy, line 212: :-)

sunggg: oops! Thanks for the catch :)

src/relax/backend/vm/vm_builtin_lower.cc
```diff
@@ -129,9 +131,45 @@ class VMBuiltinLowerMutator : public ExprMutator {
...
+    const ShapeStructInfoNode* sinfo = GetStructInfoAs<ShapeStructInfoNode>(call_node);
+    ICHECK(sinfo);
+    // call builtin function that converts tensor to shape tuple
+    Var call = builder_->Emit(Call(ExternFunc("vm.builtin.tensor_to_shape"), {expr}, {},
+                                   {GetRef<ShapeStructInfo>(sinfo)}));
+
+    // define symbolic variables
+    Array<PrimExpr> shape_var;
+    for (int i = 0; i < sinfo->ndim; i++) {
+      shape_var.push_back(tir::Var("x", DataType::Int(64)));
```
yongwww, line 166: will unique name_hint for tir::Var be generated automatically later?

tqchen:
please update via
```
git rebase --onto  upstream/unity upstream/unity-rebase-backup-2023-03-13
```

src/runtime/relax_vm/builtin.cc
```diff
@@ -380,6 +380,40 @@ TVM_REGISTER_GLOBAL("vm.builtin.make_tuple").set_body([](TVMArgs args, TVMRetVal
   *rv = arr;
 });
 
+TVM_REGISTER_GLOBAL("vm.builtin.tensor_to_shape").set_body_typed([](NDArray data) {
+  NDArray arr = data;
+  if (data->device.device_type != kDLCPU) {
```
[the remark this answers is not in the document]
slyubomirsky: I assume it's to avoid data movement if the tensor isn't on CPU. Perhaps we should incorporate data movement if it's necessary

src/relax/backend/vm/vm_builtin_lower.cc
slyubomirsky: Generating the names sounds like it should be a BlockBuilder functionality

src/relax/op/op.cc
```diff
@@ -315,6 +315,32 @@ Expr MakeShapeOf(Expr expr) {
...
+// tensor_to_shape
+
+StructInfo ReturnTensorToShapeStructInfo(const Call& call, const BlockBuilder& ctx) {
+  ICHECK(call->args.size() == 1);
+  ICHECK(call->args[0]->struct_info_.defined());
+  const auto* tsinfo = GetStructInfoAs<TensorStructInfoNode>(call->args[0]);
+  ICHECK(tsinfo && tsinfo->shape.defined());
+  ShapeExpr shape_expr = Downcast<ShapeExpr>(tsinfo->shape.value());
+  ICHECK(shape_expr->values.size() == 1);
+  const IntImmNode* ndim = shape_expr->values[0].as<IntImmNode>();
```
jwfromm, line 327: I dont think this logic is right. Shouldn't the number of dimensions and the output shape match the incoming shape? This seems like it may have been a remnant of `Reshape`'s shape function.

tests/python/relax/test_relax_operators.py
```diff
@@ -193,5 +193,20 @@ def test_op_shape_of():
...
+class TensorToShapeTest:
+    @R.function
+    def run_tensor_to_shape(t: R.Tensor(ndim=1, dtype="int64")) -> R.Shape((1, 2, 3)):
+        gv: R.Shape(ndim=3) = R.tensor_to_shape(t)
+        return gv
+
+
+def test_op_tensor_to_shape():
+    out_shape = run_cpu(
+        TensorToShapeTest, "run_tensor_to_shape", tvm.nd.array(np.array([1, 2, 3]).astype("int64"))
```
jwfromm, line 206: Noting my comment above, I'm not sure this is the proper behavior. The input tensor has `ndim=1`. Should the produced shape really have `ndim=3`?

slyubomirsky: I think this is correct. We are converting the rank-1 tensor [1, 2, 3] into a shape (1, 2, 3). I think the problem here is that `ndim` means different things for tensors (where it means the rank) and shapes (where it means the number of dimensions denoted by the shape).

src/relax/backend/vm/vm_builtin_lower.cc
sunggg: Yeah, it will resolve naming conflict by adding incremental numbers.

src/runtime/relax_vm/builtin.cc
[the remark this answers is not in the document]
sunggg: Yeah, currently we conduct all shape-related things on the host side.

tests/python/relax/test_transform_legalize_ops_manipulate.py
```diff
@@ -537,6 +584,54 @@ def reshape(var_rxplaceholder: T.handle, var_T_reshape: T.handle):
...
     tvm.ir.assert_structural_equal(mod, Expected)
 
+    # ShapeExpr might be produced by shape computation
+    @tvm.script.ir_module
+    class Reshape2:
+        @R.function
+        def main(x: R.Tensor(("a", "b"), "float32")) -> R.Tensor(("a // 2", "b * 2"), "float32"):
+            a = T.int64()
+            b = T.int64()
+            lv: R.Shape((a // 2, b * 2)) = R.shape((a // 2, b * 2))
```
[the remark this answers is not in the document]
sunggg: What do you mean by symbolic dim? If that means rank dynamism, legalization currently supports.

zxybazh (approved):

tests/python/relax/test_transform_legalize_ops_manipulate.py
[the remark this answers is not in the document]
yongwww: somthing like:
```python
        @R.function
        def main(x: R.Tensor((10, "b"), "float32")) -> R.Tensor((5, "b * 2"), "float32"):
            b = T.int64()
            lv: R.Shape((5, b * 2)) = R.shape((5, b * 2))
```

python/tvm/relax/op/base.py
```diff
@@ -402,3 +402,17 @@ def shape_of(expr: Expr) -> Expr:
...
+
+def tensor_to_shape(expr: Expr) -> Expr:
+    """Convert tensor to shape expr.
+    Parameters
+    ----------
+    expr : Expr
+        The input Expr
+    Returns
+    -------
+    result : ShapeExpr
```
slyubomirsky, line 415: Not entirely correct, the result is an expression of `Shape` type (it's an op call).

src/relax/op/op.cc
```diff
@@ -315,6 +315,32 @@ Expr MakeShapeOf(Expr expr) {
 
 TVM_REGISTER_GLOBAL("relax.op.shape_of").set_body_typed(MakeShapeOf);
 
+// tensor_to_shape
+
+StructInfo ReturnTensorToShapeStructInfo(const Call& call, const BlockBuilder& ctx) {
+  ICHECK(call->args.size() == 1);
+  ICHECK(call->args[0]->struct_info_.defined());
+  const auto* tsinfo = GetStructInfoAs<TensorStructInfoNode>(call->args[0]);
+  ICHECK(tsinfo && tsinfo->shape.defined());
```
slyubomirsky, line 324: I think this condition is more restrictive than necessary. I think it's okay for the shape not to be known at compile time and to return `ShapeStructInfo(ndim=-1)` if the input rank/shape is unknown at compile time.

src/runtime/relax_vm/builtin.cc
```diff
@@ -380,6 +380,40 @@ TVM_REGISTER_GLOBAL("vm.builtin.make_tuple").set_body([](TVMArgs args, TVMRetVal
...
+    arr = data.CopyTo(DLDevice{kDLCPU, 0});
```
slyubomirsky, line 386: Does this copy fail if the NDArray is on another device? I'm a little hesitant to have an op that just will not work depending on the device.

zxybazh: Is it possible that we have a shape tensor not on host device?

slyubomirsky: Not sure where this might come up. I wouldn't be surprised if, say, we import an ONNX model (this is where this use-case first came up) that we mean to run on GPU and every tensor (including those that stand for shapes) is stored on GPU.

tests/python/relax/test_transform_legalize_ops_manipulate.py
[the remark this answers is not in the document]
sunggg: Added your test case

src/relax/op/op.cc
sunggg: Do you mean these lines?
```
  // define symbolic variables
  Array<PrimExpr> shape_var;
  for (int i = 0; i < sinfo->ndim; i++) {
    shape_var.push_back(tir::Var("x", DataType::Int(64)));
  }
```
Initially, I also wanted to support the unknown rank but realized it is trickier than I thought. 
The problem is you need to insert these symbolic variables at compile-time so we need this info.

src/runtime/relax_vm/builtin.cc
sunggg: Based on my current understanding, we want to keep the shape-related computation on the host side since its result could be related to the memory planning. 
Even for general GPU execution besides this case, we are running output shape computation (inserted by `VMShapeLower` pass) on the host side.

src/relax/op/op.cc
slyubomirsky: What exactly are the new shape vars needed for?

sunggg: To MatchCast and put them in the `ShapeExpr`.
```
// define symbolic variables
Array<PrimExpr> shape_var;
for (int i = 0; i < sinfo->ndim; i++) {
    shape_var.push_back(tir::Var("x", DataType::Int(64)));
 }

// bind symbolic variables to the shape tuple
relax::Var var("y", ShapeStructInfo(shape_var));
builder_->EmitNormalized(MatchCast(var, call, ShapeStructInfo(shape_var)));
return ShapeExpr(shape_var);
```

slyubomirsky: That much I understand, what I am more curious about is why we need to construct this shape expression

sunggg: So that later reshape op can take it as the second input. 
```
def reshape(x: Expr, shape: Union[Tuple[PrimExprLike], Expr]) -> Expr
```
Is there any better way to do this?

slyubomirsky: Can't we just bind the result of `tensor_to_shape` to a variable and pass that variable as the shape argument to `reshape`?

sunggg: The contract with lowering is that we need to know each shape value (could be symbolic var) for memory pre-allocation. 
So, this example wouldn't lower. 
```
@I.ir_module
class Module:
    @R.function
    def main(x: R.Tensor((3,), dtype="int64")) -> R.Tensor((3,), dtype="int64"):
        gv: R.Shape(ndim=3) = R.call_packed(
            "vm.builtin.tensor_to_shape", x, sinfo_args=(R.Shape(ndim=3),)
        )
        gv3: R.Tensor(ndim=3, dtype="int64") = R.reshape(x, gv)
        gv_1: R.Tensor(ndim=3, dtype="int64") = gv3
        return gv_1
mod = LegalizeOps()(Module)
mod.show()
```

sunggg:
This PR extends existing `SimplifyNormInference` pass to serve as a generic pass to decompose composite operators like `tensor_to_shape`, `attention`, `erf`, etc. Would you check if you are okay with this change? 
cc. @Hzfengsy @SiriusNEO

src/relax/transform/decompose_composite_ops.cc
```diff
@@ -110,21 +111,60 @@ class NormInferenceSimplifier : public ExprMutator {
   Map<Expr, Expr> batch_norm_map_;
 };
 
+class OpDecomposer : public ExprMutator {
```
Hzfengsy, line 114: The name of OpDecoposer is too generic since it's only for `tensor_to_shape`

Hzfengsy (approved):
LGTM except for a naming issue

SiriusNEO:
@sunggg Hi actually we also do some changes in `SimplifyNormInference` but it has not been upstreamed yet (https://github.com/mlc-ai/relax/pull/162). Notice that you don't change the part which simplifies `BatchNorm`, I can find some time to rebase my changes on yours.

SiriusNEO (approved):

src/relax/transform/decompose_composite_ops.cc
sunggg: We can extend this function to add other composite ops like `erf`, `attention` which will be upcoming soon.

sunggg:
@yongwww @slyubomirsky would you take another look?

yongwww:
@sunggg it would be good to rebase it via `git rebase --onto upstream/unity upstream/unity-rebase-backup-2023-03-20`

src/relax/backend/vm/vm_builtin_lower.cc
```diff
@@ -129,9 +129,23 @@ class VMBuiltinLowerMutator : public ExprMutator {
...
+    auto arg = call_node->args[1];
+    CHECK(arg->IsInstance<ShapeExprNode>() || arg->IsInstance<VarNode>())
+        << "VMBuiltinLower expects the shape arg of reshape op to be a ShapeExpr or VarNode bound "
+           "to a ShapeExpr";
+
+    if (arg->IsInstance<ShapeExprNode>()) {
+      return Call(builtin_reshape_, call_node->args, Attrs(), {GetStructInfo(call_node)});
+    }
+
+    ICHECK(arg->IsInstance<VarNode>());
```
yongwww, line 141: unnecessary ICHECK

src/relax/transform/decompose_composite_ops.cc
```diff
@@ -110,21 +111,60 @@ class NormInferenceSimplifier : public ExprMutator {
...
+ private:
+  using ExprMutator::VisitExpr_;
+  Expr TensorToShape(const Call& call_node) {
+    ICHECK(call_node->args.size() == 1);
+    ICHECK(call_node->struct_info_.defined());
+    Expr expr = call_node->args[0];
+    const ShapeStructInfoNode* sinfo = GetStructInfoAs<ShapeStructInfoNode>(call_node);
+    ICHECK(sinfo);
+    // call builtin function that converts tensor to shape tuple
+    Var call = builder_->Emit(Call(ExternFunc("vm.builtin.tensor_to_shape"), {expr}, {},
```
yongwww, line 127: I think the Emit should be : 

```c++
static const Op& tensor_to_shape_op = Op::Get("relax.builtin.tensor_to_shape");
Var call = builder_->Emit(Call(tensor_to_shape_op, {expr}, {}, ...
```

src/relax/transform/decompose_composite_ops.cc
```diff
@@ -110,21 +111,60 @@ class NormInferenceSimplifier : public ExprMutator {
...
 };
 
+class OpDecomposer : public ExprMutator {
+ public:
+  static Expr Decompose(Expr expr) { return OpDecomposer()(expr); }
+
+ private:
+  using ExprMutator::VisitExpr_;
+  Expr TensorToShape(const Call& call_node) {
+    ICHECK(call_node->args.size() == 1);
```
yongwww, line 121: seems the ICHECK is duplicate, the check in ReturnTensorToShapeStructInfo should be good to cover this

yongwww (changes requested):

src/relax/transform/decompose_composite_ops.cc
sunggg: Thanks! Applied your suggestion.

yongwww (approved):
LGTM, thanks for addressing my concerns!

src/relax/op/op.cc
yongwww: great discussion folks! Looks the limitation for shape_expr is from TE, interesting

slyubomirsky: Hmmmm, I think we should put this on the list to fix later. Let's please note this limitation in a comment, since it is not obvious from looking at the code

sunggg: Added this in the comment. Would it be good to go now?

slyubomirsky (approved):
My concerns have been addressed, thank you

src/relax/transform/decompose_composite_ops.cc
yongwww: Thanks for the update, I am okay for that.

zxybazh:
Thanks @sunggg for the thoughtful design and quick implementation!

PR #14284, opened 2023-03-13
[TIR][Arith] Introduce iter range partition as an arith utility

wrongtest-intellif:
The change aims to move iteration range partition algorithm from loop partition pass into an arithmetic utility. It helps to solve certain problems by divide and conquer the iteration region. And in each sub-region we could eliminate predicates which are unresolvable under global domain.


tqchen:
closing due to staleness, feel free to raise a new PR

PR #14291, opened 2023-03-14
[Unity][BYOC] Add cuBLAS backend

masahi:
To support dynamic shape better than cutlass BYOC


src/runtime/contrib/cublas/cublas.cc
```diff
@@ -133,6 +134,120 @@ bool CheckMixPrecisionType(DLDataType in_dtype, DLDataType out_dtype, bool int_s
...
+    int batch_count_A = get_batch_count(A->shape, batch_offset_A);
+    int batch_count_B = get_batch_count(B->shape, batch_offset_B);
+    int batch_count_C = get_batch_count(C->shape, C->ndim - 2);
+    int64_t batch_stride_A = M * K;
+    int64_t batch_stride_B = K * N;
+    int64_t batch_stride_C = M * N;
+
+    // cuBLASLt does not seem to support batched GEMM with one of matrices having
+    // one batch (with batch_stride 0).
+    ICHECK_EQ(batch_count_A, batch_count_B);
```
masahi, line 231: I was not able to get ND x 2D batch GEMM working with cuBLASLt, even though the old API perfectly supports it https://github.com/apache/tvm/blob/5cacecc0c026fcd4c4c0064a8ef740374b86a699/src/runtime/contrib/cublas/cublas.cc#L305-L313

I wonder if I'm missing something or it is a known limitation cc @mnicely

vinx13 (approved):

yelite (approved):

src/runtime/contrib/cublas/cublas.cc
mnicely: Hi @masahi, let me check with the engineering team.

mnicely: @masahi what you are trying to do should definitely be supported. There might be a bug. Do you mind filing a [bug](https://forums.developer.nvidia.com/t/how-to-report-a-bug/67911)?
Include what problems, precision, and devices the caused the issue. Lastly, providing logs from [cuBLASLt](https://docs.nvidia.com/cuda/cublas/#cublaslt-logging) would very helpful.

masahi: Thanks @mnicely ! I didn't know about the logging feature, and indeed it directly points at the problem:
```
[2023-03-16 04:57:19][cublasLt][10366][Error][cublasLtMatmul] Input matrices batch counts mismatch: input B matrix batchCount (1) must be equal to all other matrices batch counts, expected (3).
```

So cublaslt only supports batched GEMM with the same batch count on A and B matrices. I'll submit a bug report.

python/tvm/contrib/cutlass/build.py
```diff
@@ -559,22 +559,9 @@ def _extract_relax_function_signature(f):
...
-    for name, annotation_pattern in pattern_entry.annotation_patterns.items():
-        arg_expr = matched_expr[annotation_pattern]
-        if arg_expr not in func_args:
-            continue
-        arg_idx[name] = func_args.index(arg_expr)
-
-    return arg_idx
+    extract_func = tvm.get_global_func("relax.contrib.extract_arg_idx")
+    arg_indices = extract_func(pattern_name, f)
+    return {k: int(v) for k, v in arg_indices.items()}
```
masahi, line 564: cc @yelite this has been ported to cpp

masahi:
Need to wait for rebase against `main` to get https://github.com/apache/tvm/pull/14363 in the `unity` branch

vinx13:
@masahi please rebase as the other PR is merged

Hzfengsy:
any updates?

masahi:
waiting for the next rebase

src/runtime/contrib/cublas/cublas.cc
```diff
@@ -133,6 +134,120 @@ bool CheckMixPrecisionType(DLDataType in_dtype, DLDataType out_dtype, bool int_s
...
+    set_batch(B_desc, batch_count_B, batch_stride_B);
+    set_batch(C_desc, batch_count_C, batch_stride_C);
+  }
+
+  auto A_data = static_cast<char*>(A->data) + A->byte_offset;
+  auto B_data = static_cast<char*>(B->data) + B->byte_offset;
+  auto C_data = static_cast<char*>(C->data) + C->byte_offset;
+
+  CHECK_CUBLAS_ERROR(cublasLtMatmul(hdl, op_desc, alpha, B_data, A_desc, A_data, B_desc, beta,
+                                    C_data, C_desc, C_data, C_desc, nullptr, nullptr, 0, nullptr));
```
vinx13, line 243: cublas API which has a default workspace pool, it seems cublasLT always require workspace being explicit set, does passing nullptr here impact the performance? We may want to have default workspace allocated and stored as thread local in `CublasThreadEntry`

masahi: yeah there are some performance knobs that might worth exploring, in terms of memory management and algorithm selection (see also https://docs.nvidia.com/cuda/cublas/#heuristics-cache). I haven't tested any of them, those are good items for future work if cuBLAS BYOC gets more traction.

masahi:
Just realized that the `unity` branch is specifying the CI setup in a different (and old?) way https://github.com/apache/tvm/blob/unity/ci/jenkins/unity_jenkinsfile.groovy. 

In particular, the GPU image is using an outdated one https://github.com/apache/tvm/blob/unity/ci/jenkins/unity_jenkinsfile.groovy#L34. That's why I'm still getting an build error after I've updated CUDA version on `main` https://github.com/apache/tvm/pull/14363

What to do? @tqchen @driazati

PR #8119, opened 2021-05-24
[CI,DOCKER] Bump gpu image to cuda 11.0.3

tkonolige:
@areusch @tqchen @comaniac

PR #14293, opened 2023-03-14
[CI] Update CUDA to 11.7

masahi:
The last time we updated the CUDA version on CI was [two years ago](https://github.com/apache/tvm/pull/8119), I think it is a good time for another update.

v11.7 is the latest version officially supported by PyTorch (https://pytorch.org/get-started/locally/). Rather than jumping directly to v12, I think it is better to stay in the stable v11 line for now. 

This would unblock the PR https://github.com/apache/tvm/pull/14291 which uses a relatively new feature in cuBLAS introduced in v11.4.

### DIFF
--- a/docker/Dockerfile.ci_gpu
+++ b/docker/Dockerfile.ci_gpu
@@ -17,7 +17,7 @@
 
 # CI docker GPU env
 # tag: v0.60
-FROM nvidia/cuda:11.0.3-cudnn8-devel-ubuntu18.04
+FROM nvidia/cuda:11.7.1-cudnn8-devel-ubuntu18.04
 
 COPY utils/apt-install-and-clear.sh /usr/local/bin/apt-install-and-clear
 


leandron:
My experience with CUDA and the whole CUDA ecosystem is very limited, but just wanted to point out that it would be good to review also the tlcpack CUDA versions, perhaps to propose this new one and deprecate some of the old version. Can you have a look?

masahi:
Do you mean this https://github.com/tlc-pack/tlcpack? I'm not familiar with this package, but its README says how to build an image with a specific CUDA version. Is there other action required to support newer CUDA versions in tlcpack?

driazati (approved):
the tlc-pack/tlcpack repo is separate and doesn't currently have a good strategy to verify changes, so I don't think we should block this PR on that. The updates there on the Docker images used to build packages can happen as a follow up

masahi:
@driazati It seems the driver has been updated, can you help?

masahi:
It passed!

PR #14294, opened 2023-03-14
[LLVM] Add support to generate llvm.assume

quic-sanirudh:
We're adding support to generate `llvm.assume` from `tir.assume` as we currently see an error when lowering code with tir.assume in certain cases.


kparzysz-quic (approved):
This looks good, but I'll defer to @Lunderberg to comment on what should happen to the assume builtins in TIR.

quic-sanirudh:
@Lunderberg Could you please take a look at this change and provide your inputs on handling `tir.assume`, thanks

PR #14298, opened 2023-03-14
[TIR][Hexagon] Enhancement of NarrowDataType pass for binary ops

ibsidorenko:
This is enhancement of [PR#13327](https://github.com/apache/tvm/pull/13327).

**Motivation**:
Playing with MetaScheduler for Hexagon target it was found that `avg_pool2d` has rather poor performance due to lack of vectorized code. `IndexDataTypeNormalizer `pass converts all indices to "int64" format and `NarrowDataTypeRewriter` should do the opposite (back to "int32"). In case of fail, we have a lot of int64 arithmetic for average pooling that can not be vectorized.

**What was done:**
Added support of binary ops ("div", "max", "min", "+" etc.) in `NarrowDataTypeRewriter`. In case of different bitwidth of operands in binary opeation it does downcasting instead of upcasting (as it was before).

**Performance impact:**
`avg_pool2d` from quantized InceptionV3 with the shape [1, 8, 35, 35, 32] (NCHW32c layout) tuned with MetaScheduler on Snapdragon 8gen1:

shape             | Before fix, ms | After fix, ms |   speedup   |
------------------|----------------|---------------|-------------|
avg_pool2d, int32 |      6.67      |      4.41     |    +34%     |

### DIFF
--- a/src/tir/transforms/narrow_datatype.cc
+++ b/src/tir/transforms/narrow_datatype.cc
@@ -258,6 +258,44 @@ class NarrowDataTypeRewriter : public IndexDataTypeRewriter {
     return Parent::VisitExpr_(op);
   }
 
+#define TVM_DEFINE_BIOP_EXPR_MUTATE_WITH_TYPE_MATCH(OP, FUNC)             \
+  PrimExpr VisitExpr_(const OP* op) {                                     \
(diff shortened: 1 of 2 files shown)


ibsidorenko:
@tvm-bot rerun

masahi (approved):

Issue #14260, opened 2023-03-09
[VOTE] Release Apache TVM v0.11.1


leandron:
Dear TVM community,

This is a call for vote to release Apache TVM version v0.11.1. https://github.com/apache/tvm/issues/14259 for any edits you would like to see included in the release notes.

Link to release notes:
https://github.com/apache/tvm/issues/14259

Link to release candidate:
https://dist.apache.org/repos/dist/dev/tvm/tvm-v0.11.1-rc1/

Link to github pre-release page:
https://github.com/apache/tvm/releases/tag/v0.11.1

Everyone is welcomed to vote. Please vote by replying to this thread explicitly.

+1 = approve
+0 = no opinion
-1 = disapprove (provide reason)

### RC0
~Vote will close Monday, Mar. 13th at 23:59PM GMT.~
We found a problem in RC0 which was fixed by https://github.com/apache/tvm/pull/14300
So an RC1 package was published and this post was updated.

### RC1
Vote will close Thursday, Mar. 16th at 23:59PM GMT.
Start of the voting thread at https://github.com/apache/tvm/issues/14260#issuecomment-1470024791

areusch:
+1

Hzfengsy:
+1

masahi:
+1

tqchen:
+1

Johnson9009:
+1, by the way the version is still 0.11.0dev, this should be changed to 0.11.1 too.
Actually, I want 0.12.0 more, because the 0.11 branch is behind too much than main.

driazati:
Checked building with [this config](https://github.com/tlc-pack/tlcpack/blob/main/wheel/build_wheel_manylinux.sh#L88) minus `ETHOSN` and `ETHOSU` and it seems to import correctly and `typing_extensions` is installed during a `pip install`. As @Johnson9009 pointed out though the rc package will need to be rebuilt after doing https://tvm.apache.org/docs/contribute/release_process.html#update-tvm-version-on-main for `v0.11.0` -> `v0.11.1` everywhere

areusch:
-1

yeah agreed we should fix the version numbering first before releasing, sorry i forgot to check that. -1 til we do that.

leandron:
Just to clarify what is missing: is it updating (e.g. via `version.py`) the version numbers in the source package so that it reflects the right version when built?

[...]

PR #14301, opened 2023-03-14
[docs] Add details about patch releases

driazati:
This updates the docs with some of the discussion and methods from
the v0.11.1 release issue #14260.

### DIFF
--- a/docs/contribute/release_process.rst
+++ b/docs/contribute/release_process.rst
@@ -86,20 +86,30 @@ The last step is to update the KEYS file with your code signing key https://www.
 Cut a Release Candidate
 -----------------------
 
-To cut a release candidate, one needs to first cut a branch using selected version string, e.g.,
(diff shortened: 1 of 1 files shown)


docs/contribute/release_process.rst
```diff
@@ -86,20 +86,30 @@ The last step is to update the KEYS file with your code signing key https://www.
...
+	git commit -m"Bump version numbers to v0.6.0"
+
+	# Replace v0.6 with the relevant version
+	git branch v0.6
+	git push --set-upstream origin v0.6
+
+	git tag v0.6.0.rc0
+	git push origin refs/tags/v0.6.0.rc0
+
+Make sure the version numbers in the source code are correct (example: https://github.com/apache/tvm/pull/14300). Run ``python3 version.py`` to update the version. Version numbers should be updated immediately after a release candidate branch is pushed.
```
leandron, line 108: Should we add a note that PRs need to be marked as GPG "Verified" in GitHub, to be acceptable? I needed to jump quite a few hoops to get that and it was not obvious beforehand, so perhaps we could make that clear for the future Release Mgrs?

Hzfengsy (approved):

PR #14302, opened 2023-03-15
[Unity][DEBUG] Add Instrument

tqchen:
This PR adds an instrumentation option to the relax VM. The instrument will be called before/after each call instruction if specified.

We also include a testing utility that leverages uses instrument. LibCompareVMInstrument leverages the instrument to compare implementations on another backend.

Also updated a few places in web runtime to improve debugging.


yongwww (approved):
LGTM!

MasterJH5574 (approved):

PR #14304, opened 2023-03-15
[Fix][TIR] Fix tvm::arith::UnionLowerBound

Biubiubiu12:
The UnionLowerBound function does not take into account the condition that  the empty set has a special representation [+inf, -inf].

cc @wrongtest-intellif

### DIFF
--- a/tests/python/unittest/test_arith_intset.py
+++ b/tests/python/unittest/test_arith_intset.py
@@ -373,6 +373,10 @@ def test_union_lower_bound():
     result = tvm.arith.int_set.union_lower_bound([set_0, set_1])
     assert result.min_value.same_as(neg_inf)
     assert result.max_value.same_as(pos_inf)
+    set_2 = tvm.arith.IntervalSet(min_value=pos_inf, max_value=neg_inf)
+    result = tvm.arith.int_set.union_lower_bound([set_0, set_2])
+    assert result.min_value.same_as(neg_inf)
+    assert result.max_value.same_as(0)
 
 
 if __name__ == "__main__":
--- a/tests/python/unittest/test_arith_intset.py
+++ b/tests/python/unittest/test_arith_intset.py
@@ -374,9 +374,9 @@ def test_union_lower_bound():
     assert result.min_value.same_as(neg_inf)
     assert result.max_value.same_as(pos_inf)
     set_2 = tvm.arith.IntervalSet(min_value=pos_inf, max_value=neg_inf)
-    result = tvm.arith.int_set.union_lower_bound([set_0, set_2])
+    result = tvm.arith.int_set.union_lower_bound([set_0, set_1, set_2])
     assert result.min_value.same_as(neg_inf)
-    assert result.max_value.same_as(0)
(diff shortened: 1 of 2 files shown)


Biubiubiu12:
cc @wrongtest-intellif

Hzfengsy:

wrongtest-intellif:
We could add a testcase to tests/python/unittest/test_arith_intset.py::test_union_lower_bound

Biubiubiu12:
> Could you please add a regression test?

Already added, thanks for your suggestion.

Biubiubiu12:
> We could add a testcase to tests/python/unittest/test_arith_intset.py::test_union_lower_bound

Already added, thanks for your suggestion

tests/python/unittest/test_arith_intset.py
```diff
@@ -370,7 +370,8 @@ def test_union_lower_bound():
     pos_inf = tvm.arith.int_set.pos_inf()
     set_0 = tvm.arith.IntervalSet(min_value=neg_inf, max_value=0)
     set_1 = tvm.arith.IntervalSet(min_value=1, max_value=pos_inf)
-    result = tvm.arith.int_set.union_lower_bound([set_0, set_1])
+    set_2 = tvm.arith.IntervalSet(min_value=pos_inf, max_value=neg_inf)
+    result = tvm.arith.int_set.union_lower_bound([set_0, set_1, set_2])
```
wrongtest-intellif, line 374: it is better to preserve original case and just append new cases

wrongtest-intellif (changes requested):

tests/python/unittest/test_arith_intset.py
Biubiubiu12: code updated，thanks for your great advice!

tests/python/unittest/test_arith_intset.py
```diff
@@ -373,6 +373,8 @@ def test_union_lower_bound():
     result = tvm.arith.int_set.union_lower_bound([set_0, set_1])
     assert result.min_value.same_as(neg_inf)
     assert result.max_value.same_as(pos_inf)
+    set_2 = tvm.arith.IntervalSet(min_value=pos_inf, max_value=neg_inf)
+    result = tvm.arith.int_set.union_lower_bound([set_0, set_2])
```
wrongtest-intellif, line 377: missing final result assertions

wrongtest-intellif (changes requested):

tests/python/unittest/test_arith_intset.py
Biubiubiu12: code updated，thanks for your great advice!

wrongtest-intellif (approved):
Look good to me

Hzfengsy (approved):
Please make the CI green and we can get it in :-)

Biubiubiu12:
@tvm-bot rerun

Biubiubiu12:
> Please make the CI green and we can get it in :-)

Hi, CI is finished.

PR #14307, opened 2023-03-15
Add a data type check for TVMArrayCopyFromTo

hxzd5568:
Add a data type check for tvm.nd.array's API TVMArrayCopyFromTo. Lacking the type check
leads to a value error in some situations. 
For example, the user input x to a built graph, the x has a shape of [200,10] and has a dtype of 'float32'. When copying the x
to the NDarray placeholder with type 'float64' whose shape is [100,10], the program doesn't report a warning, because it can pass the size check i.e. 200\*10\*32==100\*10\*64, however, every 64 continous bits in inputx is treated a number  and copyed to the new ndarray, the values change. because of different hardcode. More importantly, this kind of error is implict.    
@masahi

### DIFF
--- a/python/tvm/runtime/ndarray.py
+++ b/python/tvm/runtime/ndarray.py
@@ -140,6 +140,9 @@ class NDArray(NDArrayBase):
             Reference to self.
         """
         if isinstance(source_array, NDArrayBase):
+            if(source_array.dtype!=self.dtype):
+                raise TypeError("source dtype: {0} is different with target dtype: {1}".format(
+                    source_array.dtype, self.dtype))
             source_array.copyto(self)
             return self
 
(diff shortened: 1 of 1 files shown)


hxzd5568:
@mbrookhart @jwfromm @Huyuwei

Hi，professors. I carefully checked all the errors given by the automated review tools and I think these errors are not due to my fixe, but because the test cases themselves are not rigorous.


image-20230316114959762

tqchen:
Thanks @hxzd5568 , can you update some of the testcases?

hxzd5568:
Yes, I can. My plan for updating test cases is 1. to remove unnecessary implicit data type conversion, such as the conversion int32 to float32 in tests/python/frontend/onnx/test_forward.py::test_reshape[llvm], 2. to add force conversion expressions cast() if necessary. And we could add a flag check_dtype in python/tvm/runtime/ndarray.py:copyfromto. How about it?

tqchen:
likely we don't needs the check_dtype in set_input, but instead do that in copyfrom, so we can first copy the data to the right dtype then call set_input

tqchen:
ideally set input can be kept as minimum, if we can start from copyfrom that would be great.

PR #14308, opened 2023-03-15
Hexagon compilation on MacOS system

apeskov:
**Short desc**
This changes allow my to compile and tune models for hexagon directly from my macOS laptop without full switching to linux environment.

**List of changes**
* Replace local linker call with call from docker container with Hexagon SDK. Yes, that is the only SDK tool used by TVM during compilation.
* Enhanced search of ADB. Not only in PATH, but also in ANDROID_HOME, ANDROID_SDK_ROOT and default sdk installation directory. Mac OS doesn't allow to easily change default PATH env var for UI application launched from dock bar. So adb is not available for IDE by default.

**Motivation**
Some engineers would like to continue work with comfortable macOS environment even if they have to play with hexagon devices. At this moment there is no official Hexagon SDK for macOS system. Alternatives are next: fully switch to remote linux, use local linux virtual machine or try to introduce required hexagon SDK functionality for macOS. The last option is more preferable to me.

### DIFF
--- a/python/tvm/contrib/hexagon/build.py
+++ b/python/tvm/contrib/hexagon/build.py
@@ -29,6 +29,7 @@ import stat
 import random
 import string
 import subprocess
+import sys
 import tempfile
 from typing import Union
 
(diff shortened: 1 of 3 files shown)


python/tvm/contrib/hexagon/tools_docker_wrp.py
kparzysz-quic, line 1: Could you either merge this code into `tools.py` or rename this file to `tools_macos.py`?

python/tvm/tir/schedule/schedule.py
```diff
@@ -2852,7 +2852,7 @@ def transform_layout(
             The buffer to be transformed, or a specification of how to
             identify the buffer to be transformed.
 
-            If `buffer` if a tuple of ``(str,int)``, the first item
+            If `buffer` is a tuple of ``(str,int)``, the first item
```
kparzysz-quic, line 2855: This is an unrelated change, please put it in a separate PR.

apeskov: Removed.

python/tvm/contrib/hexagon/tools_docker_wrp.py
apeskov: Sure, as you wish. File name "tools_macos.py" doesn't look like perfect match with content, so I moved it to "tools.py".

kparzysz-quic (approved):

FdyCN:
> **Short desc** This changes allow my to compile and tune models for hexagon directly from my macOS laptop without full switching to linux environment.
> 
> **List of changes**
> 
> * Replace local linker call with call from docker container with Hexagon SDK. Yes, that is the only SDK tool used by TVM during compilation.
> * Enhanced search of ADB. Not only in PATH, but also in ANDROID_HOME, ANDROID_SDK_ROOT and default sdk installation directory. Mac OS doesn't allow to easily change default PATH env var for UI application launched from dock bar. So adb is not available for IDE by default.
> 
> **Motivation** Some engineers would like to continue work with comfortable macOS environment even if they have to play with hexagon devices. At this moment there is no official Hexagon SDK for macOS system. Alternatives are next: fully switch to remote linux, use local linux virtual machine or try to introduce required hexagon SDK functionality for macOS. The last option is more preferable to me.

Is there any ReadME.md to follow？I am a newbee in Hexagon, and i don't know how to use them. thanks a lot.

PR #14310, opened 2023-03-15
[Unity][BYOC] Improve expressiveness of the pattern check function in FuseOpsByPattern

yelite:
This PR change the interface of the check function in FuseOpsByPattern, in order to make it possible to broader types of check during fusion. In particular,
1. The old check function signature is `bool(const Map<DFPattern, Expr>& match_result, const Expr& matched_expr)`, which isn't very useful because:
    - The `Map<DFPattern, Expr> match_result` is hard to consume. To retrieve the wanted expr from this map, one has to either keep track of the DFPattern that matches the expr, or to iterate through the map values and filter by expr type.
    - The `matched_expr` provides little help because one cannot traverse to other expressions from `matched_expr`, without the help from things like `var2val`.
2. The new signature is `bool(const PatternCheckContext&)`. A dedicated type `PatternCheckContext` is created for passing information to the check function. This prevents old check functions from being broken when new input needs to be supplied to the check function. Currently `PatternCheckContext` has 
```
  /*!
   * \brief A map which contains all expressions matched by the sub patterns in
   * FuseOpsPatternNode::annotation_patterns.
   */
  Map<String, Expr> annotated_expr;

  /*!
   * \brief A map mapping variable definitions to a set of uses.
   */
  Map<Var, Array<Var>> var_usages;

  /*!
   * \brief Map from value to its bound variable.
   */
  Map<Expr, Var> value_to_bound_var;
```
3. This PR also changes the signature of `FuseOpsByPattern`, while keeping it backward compatible.

As an example usage of the new FuseOpsByPattern API, this PR includes an improvement to the check function of cutlass conv2d residual block pattern, as a follow up of #14252.

cc @vinx13 @masahi

### DIFF
--- a/include/tvm/relax/transform.h
+++ b/include/tvm/relax/transform.h
@@ -220,7 +220,7 @@ TVM_DLL Pass FuseOps(int fuse_opt_level = -1);
  * fused, it needs to be matched with `pattern` and the `check` function needs to return
  * true.
  */
-class FuseOpsPatternNode : public Object {
+class FusionPatternNode : public Object {
  public:
   /*!
    * \brief The name of pattern. It becomes the value of the kComposite attribute
@@ -245,7 +245,7 @@ class FuseOpsPatternNode : public Object {
    * NullOpt if check function is not necessary for this pattern.
    *
    * It should have signature
-   * bool(const PatternCheckFunctionInput& input)
+   * bool(const PatternCheckContext& context)
    */
   Optional<PackedFunc> check;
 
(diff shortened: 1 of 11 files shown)


masahi:
Need to resolve the conflict.

include/tvm/relax/transform.h
```diff
@@ -215,17 +215,108 @@ TVM_DLL Pass AnnotateTIROpPattern();
  */
 TVM_DLL Pass FuseOps(int fuse_opt_level = -1);
 
+/*!
+ * \brief The pattern object used as the input of FuseOpsByPattern. For bindings to be
+ * fused, it needs to be matched with `pattern` and the `check` function needs to return
+ * true.
+ */
+class FuseOpsPatternNode : public Object {
```
masahi, line 223: The name `FuseOpsPattern` might be confusing, since it sounds more related to `FuseOps` than `FuseOpsByPattern`.

yelite: As discussed offline, `FuseOpsPattern` is renamed to `FusionPattern` and `PatternCheckFunctionInput` is renamed to `PatternCheckContext`.

masahi (approved):

PR #14311, opened 2023-03-15
[Unity][Op] Enable special dimension value 0 in reshape

jwfromm:
In most definitions of Reshape, having a `0` in the `new_shape` argument means that the shape for that dimension in the output should be copied from the input. This PR extends the shape handling of `relax.op.reshape` to support this case and updates the tests accordingly.

### DIFF
--- a/src/relax/op/tensor/manipulate.cc
+++ b/src/relax/op/tensor/manipulate.cc
@@ -522,7 +522,8 @@ Expr ConvertNewShapeToExpr(const Expr& data, const ObjectRef& shape) {
                              "Array of PrimExprs. However, the given new shape is "
                           << shape;
   int dim_to_infer = -1;
-  PrimExpr new_shape_prod = IntImm(DataType::Int(64), 1);
+  // Keep track of which dimensions should be copied from input.
+  std::vector<int> zero_dims;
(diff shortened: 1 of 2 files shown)


yongwww (approved):
LGTM

slyubomirsky:
Does the `FInferStructInfo` for reshape reflect the deduced dimensions?

jwfromm:
Yes, the way it works currently is during the construction of reshape, any special values are rewritten to positive integers. Any down the line passes or inference will see the non-special values.

slyubomirsky:
Great. Are there test cases that demonstrate that this is handled correctly during `StructInfo` checking? E.g., just have a binding like `x: R.Tensor((insert resulting shape), "float32") = R.reshape(...)`.

jwfromm:
The tests in `test_op_manipulate` do exactly that :)

MasterJH5574 (approved):
LGTM, thanks @jwfromm! It is indeed fine for us to infer 0s on the API side, as long as we don’t introduce 0 and -1 to ShapeExpr.

junrushao:
Hey I hit this semantics just now in op schema effort :-)

This is definitely an interesting usecase and is not hard to support. I'm just curious in which frameworks 0s are used to indicate copy?

Thanks a lot!

PR #14314, opened 2023-03-15
[microTVM][Target]Consolidate `stm32` target to `micro` target

mehrdadh:
This PR consolidates stm32 targets into micro targets and renames some of the STM micro targets with generalized name. In addition it adds -keys=arm_cpu,cpu by default to all ARM micro targets.

### DIFF
--- a/apps/microtvm/zephyr/template_project/boards.json
+++ b/apps/microtvm/zephyr/template_project/boards.json
@@ -43,7 +43,7 @@
     },
     "nucleo_f746zg": {
         "board": "nucleo_f746zg",
-        "model": "stm32f746xx",
+        "model": "stm32f7xx",
         "is_qemu": false,
         "fpu": true,
         "vid_hex": "0483",
@@ -51,7 +51,7 @@
     },
     "nucleo_l4r5zi": {
         "board": "nucleo_l4r5zi",
-        "model": "stm32l4r5zi",
+        "model": "stm32l4r5xx",
         "is_qemu": false,
         "fpu": true,
         "vid_hex": "0483",
@@ -93,7 +93,7 @@
     },
     "stm32f746g_disco": {
         "board": "stm32f746g_disco",
-        "model": "stm32f746xx",
+        "model": "stm32f7xx",
         "is_qemu": false,
         "fpu": true,
         "vid_hex": "0483",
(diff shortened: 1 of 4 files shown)


mehrdadh:
cc @areusch for review

apps/microtvm/zephyr/template_project/boards.json
```diff
@@ -43,15 +43,15 @@
     },
     "nucleo_f746zg": {
         "board": "nucleo_f746zg",
-        "model": "stm32f746xx",
+        "model": "stm32f7xx",
```
areusch, line 46: why these changes?

mehrdadh: `stm32f7xx` is more generic than `stm32f746xx`
https://www.st.com/en/microcontrollers-microprocessors/stm32f7-series.html

python/tvm/target/target.py
```diff
@@ -443,22 +443,45 @@ def intel_graphics(model="unknown", options=None):
...
-    "nrf5340dk": ["-mcpu=cortex-m33"],
-    "rp2040": ["-mcpu=cortex-m0"],
-    "sam3x8e": ["-mcpu=cortex-m3"],
-    "stm32f746xx": ["-mcpu=cortex-m7", "-march=armv7e-m"],
-    "stm32h7xx": ["-mcpu=cortex-m7"],
-    "stm32l4r5zi": ["-mcpu=cortex-m4"],
-    "stm32u5xx": ["-mcpu=cortex-m33"],
-    "zynq_mp_r5": ["-mcpu=cortex-r5"],
+    # NXP
+    "imxrt10xx": ["-keys=arm_cpu,cpu", "-mcpu=cortex-m7"],
```
janetsc, line 477: Does this also need "-march=armv7e-m"?

mehrdadh: I think `armv7e-m` should be on all cortex-m4 and cortex-m7 boards:
https://en.wikipedia.org/wiki/ARM_Cortex-M

mehrdadh: cc @areusch

areusch: i believe the compiler infers this from -mcpu for ARM parts. cc @Mousius if he knows more

mehrdadh: @areusch that was my expectation too, but I saw `-march=armv7e-m` in one of the targets and I got confused.

janetsc: +1 - that's why I asked, because I saw it specified above.  Maybe that can be removed?

PR #14316, opened 2023-03-16
[TensorIR] New schedule primitive `set_dtype`

yzh119:
# Motivation
Currently, we miss a schedule primitive to change the data type of allocated buffer (e.g. via `cache_read`/`cache_write`), and thus we cannot perform type conversion while loading data from global to shared memory.

This PR adds a new schedule primitive `set_dtype` that follows the interface of `set_scope` and allows users to customize the allocated buffers' data type.

# Example
Before running `set_dtype`:
```python
@T.prim_func
def before_set_dtype(
    A: T.Buffer((128, 128), "float32"), C: T.Buffer((128, 128), "float32")
) -> None:
    B = T.alloc_buffer((128, 128), dtype="float32")

    for i, j in T.grid(128, 128):
        with T.block("B"):
            B[vi, vj] = A[vi, vj] * 2.0
    for i, j in T.grid(128, 128):
        with T.block("C"):
            vi, vj = T.axis.remap("SS", [i, j]
            C[vi, vj] = B[vi, vj] + 1.0
```
then we perform the `set_dtype` schedule:
```python
sch = tir.Schedule(before_set_dtype)
sch.set_dtype("B", buffer_index=0, dtype="float16")
print(sch.mod["main"].script())
```
we get transformed code:
```python
@T.prim_func
def after_set_dtype(
    A: T.Buffer((128, 128), "float32"), C: T.Buffer((128, 128), "float32")
) -> None:
    B = T.alloc_buffer((128, 128), dtype="float16")

    for i, j in T.grid(128, 128):
        with T.block("B"):
            B[vi, vj] = T.cast(A[vi, vj] * 2.0, "float16")
    for i, j in T.grid(128, 128):
        with T.block("C"):
            vi, vj = T.axis.remap("SS", [i, j]
            C[vi, vj] = T.cast(B[vi, vj], "float32") + 1.0
```
where data type conversions are inserted automatically.

# Other Usage
Using the combination of `cache_read` + `set_dtype` can help us load data from the memory hierarchy while converting data to the desired type.

cc @Hzfengsy @vinx13 @junrushao

### DIFF
--- a/include/tvm/tir/schedule/schedule.h
+++ b/include/tvm/tir/schedule/schedule.h
@@ -594,11 +594,13 @@ class ScheduleNode : public runtime::Object {
   /*!
    * \brief Set the data type of a buffer, where the buffer is specified by a block and a
    * write-index
+   * \note This schedule primitive is unsafe and may change correctness of program because of
+   *   type conversion, please use with caution.
    * \param block_rv The producer block of the buffer
    * \param buffer_index the index of the buffer in block's write region
    * \param dtype The data type to be set
    */
-  virtual void SetDType(const BlockRV& block_rv, int buffer_index, const String& dtype) = 0;
+  virtual void UnsafeSetDType(const BlockRV& block_rv, int buffer_index, const String& dtype) = 0;
   /******** Schedule: Blockize & Tensorize ********/
   /*!
    * \brief Convert the subtree rooted at a specific loop into a block.
--- a/include/tvm/tir/schedule/schedule.h
+++ b/include/tvm/tir/schedule/schedule.h
@@ -584,13 +584,21 @@ class ScheduleNode : public runtime::Object {
   virtual void StorageAlign(const BlockRV& block_rv, int buffer_index, int axis, int factor,
                             int offset) = 0;
   /*!
-   * \brief Set the storage scope of a buffer, where the buffer is specified by the a block and a
+   * \brief Set the storage scope of a buffer, where the buffer is specified by a block and a
    * write-index
    * \param block_rv The producer block of the buffer
    * \param buffer_index The index of the buffer in block's write region
    * \param storage_scope The storage scope to be set
    */
   virtual void SetScope(const BlockRV& block_rv, int buffer_index, const String& storage_scope) = 0;
+  /*!
+   * \brief Set the data type of a buffer, where the buffer is specified by a block and a
+   * write-index
+   * \param block_rv The producer block of the buffer
+   * \param buffer_index the index of the buffer in block's write region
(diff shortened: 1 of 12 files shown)


Hzfengsy:
It's a bit tricky since it would change the behavior of the PrimFunc (aka get different result before and after the schedule)

quic-sanirudh:
> It's a bit tricky since it would change the behavior of the PrimFunc (aka get different result before and after the schedule)

There seems to be some interest in introducing schedule primitives like these that might cause different output. This certainly seems like a useful primitive to have, but since we don't normally want schedule primitives to cause the output to change, primitives like these might be hard to introduce.

Would it make sense to have some discussion on perhaps introducing a different scheduling class that allows these kinds of primitives, or explicitly setting a "danger" flag in the Schedule class before using these primitives. That way, we could still allow primitives like these and make sure that a user knows the risk of using a primitive that could cause the output to change.

yzh119:
Adding a `danger` flag sounds good to me, any ideas on how to make proper abstractions for `danger`?

tqchen:
how about `unsafe_set_dtype`, with document to users that this operation can change the model behavior.

junrushao:
@yzh119 any updates

yzh119:
`unsafe_set_dtype` sounds good to me, I have updated code accordingly.

include/tvm/tir/schedule/schedule.h
```diff
@@ -584,13 +584,21 @@ class ScheduleNode : public runtime::Object {
...
    */
   virtual void SetScope(const BlockRV& block_rv, int buffer_index, const String& storage_scope) = 0;
+  /*!
+   * \brief Set the data type of a buffer, where the buffer is specified by a block and a
+   * write-index
+   * \param block_rv The producer block of the buffer
+   * \param buffer_index the index of the buffer in block's write region
+   * \param dtype The data type to be set
+   */
+  virtual void SetDType(const BlockRV& block_rv, int buffer_index, const String& dtype) = 0;
```
junrushao, line 601: looks like we missed a couple of places lol

tqchen (approved):

tqchen:
oops, @yzh119 please also check the c++ apis @junrushao mentioned :)

include/tvm/tir/schedule/schedule.h
yzh119: Fixed :)

junrushao (approved):
LGTM!

junrushao:
I don't know why but now it requires committer approval to run github actions for existing contributors. I'm merging this in after the actions got green

PR #14319, opened 2023-03-16
[Bugfix][TVMScript] Preserve variable names in LetStmt

Lunderberg:
When the TVMScript parser generates a `tir::LetStmt` to represent assignment to a variable, the name of the TVMScript variable should be used as the name of the `tir::Var` in the `tir::LetStmt`.

This bug was introduced in https://github.com/apache/tvm/pull/14207.


junrushao:
Retriggered the CI. Should be good to merge

PR #14323, opened 2023-03-17
[LLVM] Fix registerCallbacks API after recent change

quic-sanirudh:
LLVM upstream recently changed their API for registerCallbacks in https://reviews.llvm.org/D146160 and this causes an error while building


quic-sanirudh:
cc @kparzysz-quic

kparzysz-quic (approved):
Thanks!

LRY89757:
Are you sure that this api has changed since the version 17.0.0? I used your version to compile tvm with llvm 17.0.0, but failed. When I use the fam instead of the mam, it runs smoothly.

kparzysz-quic:
Version 17 of LLVM at this moment refers to the `main` branch of it.  If you refresh your sources of LLVM, you will get the changes that this patch reflects.

PR #14324, opened 2023-03-17
[Unity][Op] Fix Strided Slice Shape Inference

zxybazh:
In the case where the begin index of strided slice is out of [-ndim, dim), the strided slice operator will yield some incorrect shape inference. This PR corrected this issue by canonicalizing the begin and end index for strided slice and calculate the symbolic shape. Added some new unit tests for out of range begin locations and changed some symbolic shape tests.

I have 2 questions that we can follow up in the design:
1. Is it expected that in the case of symbolic shape input tensor we should have symbolic shape inferenced for generated tensors after strided slice? See unit test change for details, when we have different input shape at runtime we may have corresponding symbolic output shape.
2. We do have the similar function / logic in topi strided slice implementation. Should we unify them some how? Is it expected that we use some mechanism to store the shape inference logic for both relax / topi operators?

Would be great to see other contributors opinions on this.

CC: @MasterJH5574 @sunggg @tqchen

### DIFF
--- a/src/relax/op/tensor/index.cc
+++ b/src/relax/op/tensor/index.cc
@@ -24,7 +24,6 @@
 
 #include "index.h"
 
-#include <algorithm>
 #include <utility>
 #include <vector>
 
@@ -147,7 +146,7 @@ inline PrimExpr CanonicalizeIndex(PrimExpr index, PrimExpr extent, int64_t strid
   PrimExpr begin_range = stride < 0 ? -1 : 0;
   PrimExpr end_range = stride < 0 ? extent - 1 : extent;
   index = if_then_else(index < 0, index + extent, index);
-  return min(max(index, begin_range), end_range);
+  return min(max(index, begin_range), end_range);  // NOLINT
 }
 
 PrimExpr GetLength(PrimExpr begin, PrimExpr end, const int stride, const PrimExpr& ndim) {
--- a/src/relax/op/tensor/index.cc
+++ b/src/relax/op/tensor/index.cc
@@ -24,6 +24,7 @@
 
 #include "index.h"
 
+#include <algorithm>
 #include <utility>
 #include <vector>
 
--- a/src/relax/op/tensor/index.cc
+++ b/src/relax/op/tensor/index.cc
@@ -143,20 +143,22 @@ Expr strided_slice(Expr x,                 //
 TVM_REGISTER_GLOBAL("relax.op.strided_slice").set_body_typed(strided_slice);
 
 inline PrimExpr CanonicalizeIndex(PrimExpr index, PrimExpr extent, int64_t stride) {
+  // Same as topi strided slice CanonicalizeIndex function in
+  // include/tvm/topi/detail/strided_slice.h
   PrimExpr begin_range = stride < 0 ? -1 : 0;
   PrimExpr end_range = stride < 0 ? extent - 1 : extent;
   index = if_then_else(index < 0, index + extent, index);
   return min(max(index, begin_range), end_range);  // NOLINT
 }
 
-PrimExpr GetLength(PrimExpr begin, PrimExpr end, const int stride, const PrimExpr& ndim) {
+PrimExpr GetLength(PrimExpr begin, PrimExpr end, const int64_t stride, const PrimExpr& ndim) {
   begin = CanonicalizeIndex(begin, ndim, stride);
   end = CanonicalizeIndex(end, ndim, stride);
 
   if (stride < 0) {
-    return ceildiv(begin - end, -stride);
+    return ceildiv(begin - end, IntImm(DataType::Int(64), -stride));
   } else {
-    return ceildiv(end - begin, stride);
+    return ceildiv(end - begin, IntImm(DataType::Int(64), stride));
   }
 }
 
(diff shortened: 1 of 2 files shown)


slyubomirsky:
It would be nice to have a single source of truth for shape inference, if that's what you're referring to in question 2.

What exactly is question 1 referring to? You can use `MatchCast` to bind the inferred dimensions to new variables to avoid having to copy over cumbersome expressions, but we should be able to nest PrimExprs in symbolic shapes and do further algebraic reasoning on them

src/relax/op/tensor/index.cc
```diff
@@ -24,6 +24,7 @@
 
 #include "index.h"
 
+#include <algorithm>
```
slyubomirsky, line 27: What is used from `std::algorithm` here?

tests/python/relax/test_op_index.py
```diff
@@ -376,22 +409,22 @@ def test_strided_slice_infer_struct_info_shape_symbolic():
     _check_inference(
         bb,
         relax.op.strided_slice(x0, axes=[0], begin=[1], end=[3]),
-        relax.TensorStructInfo((2, n), "float32"),
+        relax.TensorStructInfo((tir.min(3, m) - tir.min(1, m) + 1 - 1, n), "float32"),
```
yongwww, line 412: Thanks for the update.

`relax.TensorStructInfo((2, n), "float32")` makes sense to me as well, error could be raised if `m < 1 || m > 3` happens. Glad to hear more comments.

yongwww:
for question 1, I would prefer to keep it static if it is not data-dependent cases, raising error during runtime when the dim is not valid should be fine to me.

tests/python/relax/test_op_index.py
slyubomirsky: Oh I see, that's what the question was getting at. So if `m` is not 2, would the current implementation give an error? If we require that, perhaps `FInferStructInfo` should assert that `m` is 2 and require a `MatchCast` beforehand. I don't think we should allow for shape mismatch errors to happen "silently" (have the StructInfo system accept the program without any explicit casts but then fail at runtime)

sunggg: I think the problem is depending on what `m` is given at runtime, the output shape can be different due to the boundary condition. 
For example, when I play around with this in numpy:
```
>>> a= np.random.rand(4)
>>> a[1:3].shape
(2,)
>>> b=np.random.rand(3)
>>> b[1:3].shape
(2,)
>>> c=np.random.rand(2)
>>> c[1:3].shape
(1,)
>>> d=np.random.rand(1)
>>> d[1:3].shape
(0,)
```
So, I think it might be inevitable to have symbolic shape in this case and insert the runtime check (e.g., detecting zero like `d`).

yongwww: ohh, just took a look at the op implementation logic, it is valid even `m` is 1 or 2, a bit tricky...

yongwww: then it looks fine to me.

yongwww (approved):

tests/python/relax/test_op_index.py
zxybazh: Thanks for the comments. In this case, I think even when `m < 3` it's still a valid case there's no need to raise an error. And this seems to be a simple case where the input shape has a simple `m` variable. If the input is another symbolic shape we may not be able to point out some invalid range like `m < 3`.

On the other hand, indeed in most cases the result is the same as static version as `(2, n)` so I'm wondering if this could cause any side effect if we keep the symbolic shape generated here.

sunggg: @zxybazh is your concern that we may face invalid 'm' at runtime? If so, StructInfo is designed to be the best effort approach. If there are certain things we cannot identify, we can defer it to runtime checks.

sunggg:
Thank you for great contribution, @zxybazh!
Regarding your points, 
1. I left my thoughts here: https://github.com/apache/tvm/pull/14324#discussion_r1141201248
2. Like @slyubomirsky pointed out, I also want to keep single source of truth. However, this requires major refactoring to TOPI side which is also used by Relay side. So, as a temporary solution, I think it is okay to have a separate output shape computation and we can fix this when we gradually migrate to Unity topi.

tests/python/relax/test_op_index.py
zxybazh: Thanks for the clarification. Then it should be good enough.

sunggg (approved):
LGTM!

src/relax/op/tensor/index.cc
```diff
@@ -142,6 +142,26 @@ Expr strided_slice(Expr x,                 //
...
+  // include/tvm/topi/detail/strided_slice.h
+  PrimExpr begin_range = stride < 0 ? -1 : 0;
+  PrimExpr end_range = stride < 0 ? extent - 1 : extent;
+  index = if_then_else(index < 0, index + extent, index);
+  return min(max(index, begin_range), end_range);  // NOLINT
+}
+
+PrimExpr GetLength(PrimExpr begin, PrimExpr end, const int64_t stride, const PrimExpr& ndim) {
+  begin = CanonicalizeIndex(begin, ndim, stride);
+  end = CanonicalizeIndex(end, ndim, stride);
```
MasterJH5574, line 156: Here it is the dim length rather than `ndim`.
```suggestion
PrimExpr GetLength(PrimExpr begin, PrimExpr end, const int64_t stride, const PrimExpr& length) {
  begin = CanonicalizeIndex(begin, length, stride);
  end = CanonicalizeIndex(end, length, stride);
```

MasterJH5574 (commented):
Just one quick fix.

PR #14330, opened 2023-03-19
[CODEGEN][METAL] Fix ramp codegen

tqchen:
Fix ramp node codegen for the metal backend.
The default C codegen can cause problem in
vector indices assignment.

### DIFF
--- a/src/target/source/codegen_metal.cc
+++ b/src/target/source/codegen_metal.cc
@@ -299,6 +299,17 @@ void CodeGenMetal::VisitExpr_(const BroadcastNode* op, std::ostream& os) {  // N
   os << ')';
 }
 
+void CodeGenMetal::VisitExpr_(const RampNode* op, std::ostream& os) {  // NOLINT(*)
+  PrintType(op->dtype, os);
+  os << "(";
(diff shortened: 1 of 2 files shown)


tqchen:
currently ci do not support metal, confirmed on apple M2 for some cases that works on cuda,

MasterJH5574 (approved):
Thanks for the fix!

Hzfengsy (approved):

echuraev (commented):
The PR LGTM. And I don't want to block this PR by the comment, but I have one question: will it lake a long time to add a test for this change? If not, then I could run this test on my x86 Mac and check if it works here as well.
I don't think that it should take a lot of time for adding such test, because metal codegen test exists in the TVM.

junrushao (approved):

junrushao:
Agree with @echuraev that we might need a unittest. @tqchen @MasterJH5574 would you like to follow up with a new PR? Thanks a lot!

tqchen:
Thanks @echuraev . we will send in a test in a followup

tqchen:
https://github.com/apache/tvm/pull/14331

PR #14332, opened 2023-03-19
[CODEGEN][METAL] Fix unaligned vector load

tqchen:
This PR fixes the implementation of unaligned vector load. Previously vector construction was printed as (float2)(v0, v1). This will cause problem as C have comma expression, and (v0, v1) will be evaluated as v1. The final result will become float2(v1, v1). The bug affects all codegen that uses the default implementation, such as metal. We added a testcase on metal to cover this case.

### DIFF
--- a/src/target/source/codegen_c.cc
+++ b/src/target/source/codegen_c.cc
@@ -802,15 +802,16 @@ void CodeGenC::VisitExpr_(const LetNode* op, std::ostream& os) {  // NOLINT(*)
 }
 
 void CodeGenC::VisitExpr_(const RampNode* op, std::ostream& os) {  // NOLINT(*)
-  // constraint of current logic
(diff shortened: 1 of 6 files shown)


junrushao (approved):

MasterJH5574 (approved):

spectrometerHBH (approved):
Shall we somehow examine the output C source code is as expected?

tqchen:
I have thought a bit about it in the end feel test correctness coverage is better in this case if we want future edits. We can indeed followup with source checking if new related regression arises

junrushao:
it seems that this patch breaks OpenCL/Vulkan codegen..

tqchen:
interesting, will dig a bit.

tqchen:
seems that `(float2)(v0,v1)` was the right syntax for opencl, will add an overload to the opencl codegen

PR #14334, opened 2023-03-19
[Unity][Pass] Reuse prior infra to implement more complete DCE

ganler:
As a follow-up to https://github.com/apache/tvm/pull/14262, I just noticed that I previously implemented a function called `RemoveAllUnused` (https://github.com/apache/tvm/pull/14043) that can do function-wise DCE and should be more complete than https://github.com/apache/tvm/pull/14262 as `RemoveAllUnused` can also remove dead dataflow blocks (added two test-cases to show that). 

Sorry, I should have mentioned that earlier to avoid duplication... but still I am putting a patch here in case it is desired.

cc: @spectrometerHBH @tqchen

### DIFF
--- a/include/tvm/relax/transform.h
+++ b/include/tvm/relax/transform.h
@@ -387,6 +387,7 @@ TVM_DLL Pass ConvertLayout(Map<String, Array<String>> desired_layouts);
 
 /*!
  * \brief Dead code elimination.
+ * \sa RemoveAllUnused
  * Currently it removes:
  *   1. Unused local VarBindings in a DataflowBlock.
  *   2. Unused DataflowBlocks in a function.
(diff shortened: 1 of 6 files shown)


ganler:
At the high level, https://github.com/apache/tvm/pull/14043 is based on a fix-point algorithm to remove unused dataflow statements iteratively. While I think @spectrometerHBH 's impl is more efficient based on the assumption from the dataflow block.

junrushao (approved):
LGTM

ganler:
BTW, although this patch uses `RemoveAllUnused` over @spectrometerHBH 's implementation, it did not remove the prior implementation as well. Not sure if we should just keep it in case of future usage or just trim it to simplify the code...

tqchen:
I think we can simply trim it and consolidate the use

ganler:
For now:

- `tvm.relax.transform.DeadCodeElimination` is the pass for running DCE over an IRModule.
- `tvm.relax.analysis.remove_all_unused` is the pass that .... over a function.

`tvm.relax.transform.DeadCodeElimination` is implemented based on `tvm.relax.analysis.remove_all_unused`. I did not sync the function name for the two, as I saw there are other uses for `RemoveAllUnused` so just want to be conservative first.

MasterJH5574:
@ganler Thanks for the clarification!

MasterJH5574 (approved):

PR #14336, opened 2023-03-20
[Unity] Remove Python interface of RemoveUnusedFunction

MasterJH5574:
The pass RemoveUnusedFunctions was renamed and enhanced to the DCE pass in #14262. This PR removes the Python interface of RemoveUnusedFunctions in transform.py.

It is surprising that this Python API was never tested in any testing file before.

This PR also fixes `alter_op_impl.cc` and `to_mixed_precision.cc` to for Clang build warnings.

### DIFF
--- a/python/tvm/relax/transform/transform.py
+++ b/python/tvm/relax/transform/transform.py
@@ -190,24 +190,6 @@ def BindParams(
     return _ffi_api.BindParams(func_name, tvm_params)  # type: ignore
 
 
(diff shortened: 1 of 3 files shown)


sunggg (approved):
LGTM! 

> It is surprising that this Python API was never tested in any testing file before

Ooops.. Iirc, this pass was introduced as a part of BYOC mechanism and it seems like I forgot to add test cases haha My bad! Will double check next time :)

MasterJH5574:
@sunggg Never mind, thanks!

PR #14337, opened 2023-03-20
[Unity][TVMScript] Register the dispatch for `runtime::Module`

sunggg:
This PR registers the dispatch for `runtime::Module` so that we can print `IRModule` with BYOC runtime after `RunCodegen` pass. 

cc. @yelite @junrushao

### DIFF
--- a/src/script/printer/ir/ir.cc
+++ b/src/script/printer/ir/ir.cc
@@ -103,6 +103,14 @@ TVM_STATIC_IR_FUNCTOR(IRDocsifier, vtable)
     });
 
 TVM_STATIC_IR_FUNCTOR(IRDocsifier, vtable)
+    .set_dispatch<runtime::Module>("",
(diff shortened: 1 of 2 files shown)


Hzfengsy:
One generic problem: can `runtime::Module` roundtrip-able print and parse if we printed it as a string? One alternate solution is printed it as metadata. 

Love to hear more opinions from the community. cc @junrushao @cyx-6

junrushao:
I don't think it's roundtrippable, and additionally, there is no guarantee that it's serializable, which means even if we print it as metadata it's not guaranteed to be parsed back

tqchen:
I think we can close this for now, following by a more robust solution that detects binary serializable modules and only serialize those(perhaps via base64 encoding)

tqchen:
https://github.com/apache/tvm/pull/14406

sunggg:
Sounds good to me. Will create the new PR with this change https://github.com/apache/tvm/pull/14406. Thank you all for the review and suggestions!

PR #14338, opened 2023-03-20
[Unity][Frontend] FX exp and strided_slice fix

MasterJH5574:
* Add the support of `exp` for the FX translator.
* Previously the way FX translator dealt with `None` in torch tensor slice (e.g., `x[:, None, None]`) is not right. This PR fixes this issue. Specifically, the `None` here means dim expansion, and the previous impl mistakenly increases the dim counter when seeing `None`, which will lead to dim counter out-of-range issue in the end.

### DIFF
--- a/tests/python/relax/test_frontend_from_fx.py
+++ b/tests/python/relax/test_frontend_from_fx.py
@@ -19,7 +19,7 @@ import pytest
 import tvm
 from tvm import relax
 import tvm.testing
-from tvm.script.parser import relax as R, tir as T
+from tvm.script.parser import ir as I, relax as R, tir as T
 
 
 def verify_model(torch_model, input_info, binding, expected):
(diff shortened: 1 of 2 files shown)


tests/python/relax/test_frontend_from_fx.py
```diff
@@ -28,6 +28,7 @@ def verify_model(torch_model, input_info, binding, expected):
 
     graph_model = fx.symbolic_trace(torch_model)
     mod = from_fx(graph_model, input_info)
+    print(mod.script())
```
psrivas2, line 31: left over debug statement?

psrivas2 (approved):

tests/python/relax/test_frontend_from_fx.py
MasterJH5574: Yes sorry for that. Just removed it.

tqchen (approved):

PR #14339, opened 2023-03-20
[Unity][BYOC] Confirm with recent changes

sunggg:
This PR fixes the outdated test case.
It seems like CI does not check some files including this one, probably BYOC or tuning-related. 
cc. @psrivas2 @areusch


psrivas2 (approved):
LGTM, thanks!
@yongwww is working on enabling more tests in CI so these could be caught sooner :)

Issue #14326, opened 2023-03-17
[Bug] Conv3Dtranspose default kernel layout should be IODHW


rebel-jangys:
Hi, I found some bugs in TVM relay type relation of conv3dtranspose.

https://github.com/apache/tvm/blob/fe3fa9d75d6bda75cca644516b3cbcdedd0354a8/src/relay/op/nn/convolution.cc#L597
I believe the above line should be set to the IODHW layout, since the below conv3dtranspose kernel is made with the IODHW layout.

https://github.com/apache/tvm/blob/fe3fa9d75d6bda75cca644516b3cbcdedd0354a8/src/relay/op/nn/convolution.cc#L629-L630

Additionally, this line should use 'indexdiv(param->channels, param->groups)' like Conv2dTransposeRel instead of dividing the input channels by the number of groups.
https://github.com/apache/tvm/blob/fe3fa9d75d6bda75cca644516b3cbcdedd0354a8/src/relay/op/nn/convolution.cc#L663

The weight of this test is treated as 'OIDHW' layout(the default kernel layout for conv3dtranspose is "OIDHW") despite having 10 input channels and 12 output channels.
https://github.com/apache/tvm/blob/fe3fa9d75d6bda75cca644516b3cbcdedd0354a8/tests/python/relay/test_op_level2.py#L785-L787

Correct me if I'm wrong. Once again, thanks for your wonderful framework!

masahi:
> I believe the above line should be set to the IODHW 

You are right.

> Additionally, this line should use 'indexdiv(param->channels, param->groups)' 

I think this line tries to compare the input channels. So the correct fix would be to remove the indexdiv.

We should also fix the followings:
https://github.com/apache/tvm/blob/main/include/tvm/relay/attrs/nn.h#L591
https://github.com/apache/tvm/blob/main/include/tvm/relay/attrs/nn.h#L433
https://github.com/apache/tvm/blob/fd5915a098f419f1d6ac0ee8c78d3a364f56a163/python/tvm/relay/op/nn/nn.py#L439

Can you send a PR to fix them?

PR #14340, opened 2023-03-20
[Bugfix] Conv3Dtranspose default kernel layout should be IODHW

rebel-jangys:
Fix the issue #14326 
@masahi


masahi (approved):

masahi:
Can you look at the failed test?

rebel-jangys:
SInce this PR corrects the conv3d transpose kernel layout, many frontend relay converters and test cases have to be fixed as well.

For example, the below test case intentionally set wrong kernel_layout to fit in original wrong kernel layout implementation.
https://github.com/apache/tvm/blob/46fb2ff35f95321e5d76610960c626bfa7a784fc/tests/python/contrib/test_dnnl.py#L479-L490

I’ve fixed pytorch relay converter and some test cases, but I would greatly appreciate any assistance with fix.

masahi:
@rebel-jangys Thanks for your hard work, it's merged finally!

PR #14343, opened 2023-03-20
[TOPI] Fix data race of batch multibox detection

wrongtest-intellif:
- For cpu,  currently the buffer space of `box_coord` and `pred_coord` is shared across parallel batch like
```python
for i in parallel(batch_size):
    # ...
                for l in range(4):
                    box_coord[l] = anchor[0, j, l]
                    pred_coord[l] = loc_pred[i, j * 4 + l]
```

- For cuda, fix two index usage error of batch idx.

### DIFF
--- a/python/tvm/topi/vision/ssd/multibox.py
+++ b/python/tvm/topi/vision/ssd/multibox.py
@@ -137,17 +137,17 @@ def multibox_prior(data, sizes=(1,), ratios=(1,), steps=(-1, -1), offsets=(0.5, 
 
 
 @hybrid.script
-def _hybridy_transform_loc(box, pred_loc, variance, clip):
+def _hybridy_transform_loc(box, pred_loc, variance, clip, batch_idx):
     """Transform prior anchor box to output box through location predictions."""
-    al = box[0]
-    at = box[1]
-    ar = box[2]
-    ab = box[3]
+    al = box[batch_idx, 0]
+    at = box[batch_idx, 1]
+    ar = box[batch_idx, 2]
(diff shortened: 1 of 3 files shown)


wrongtest-intellif:
@tvm-bot  rerun

python/tvm/topi/vision/ssd/multibox.py
```diff
@@ -230,9 +242,9 @@ def hybrid_multibox_transform_loc(cls_prob, loc_pred, anchor, clip, threshold, v
                 out_loc[i, valid_count[i], 0] = cls_id - 1.0
                 out_loc[i, valid_count[i], 1] = score
                 for l in range(4):
-                    box_coord[l] = anchor[0, j, l]
-                    pred_coord[l] = loc_pred[i, j * 4 + l]
-                out_coord = _hybridy_transform_loc(box_coord, pred_coord, variances, clip)
+                    box_coord[i, l] = anchor[0, j, l]
```
masahi, line 245: It seems `box_coord` is the same in all threads. How about passing `anchor`, `j`, and `l` to `_hybridy_transform_loc` instead?

masahi (approved):

python/tvm/topi/vision/ssd/multibox.py
wrongtest-intellif: good idea~ updated as suggested.

masahi:
Need to fix lint.

masahi:
@tvm-bot rerun

wrongtest-intellif:
@tvm-bot rerun

PR #14344, opened 2023-03-20
[BugFix] Support rewrite_once when the number of callbacks > 1

sisleyli:
The `rewrite_once` of the current `PatternRewriter`  rewrites once only when the last callback in the callbacks list is `rewrite_once`. I don't think this is as expected.   
I think the `rewrite_once` strategy should be that the callback marked with `rewrite_once` is rewritten only once, and the rest of the callbacks are still rewritten multiple times.

### DIFF
--- a/src/relay/ir/dataflow_matcher.cc
+++ b/src/relay/ir/dataflow_matcher.cc
@@ -796,24 +796,38 @@ Expr PatternRewriter::Rewrite(const Array<DFPatternCallback>& callbacks, const E
   bool equal = true;
   static auto* structural_equal = runtime::Registry::Get("node.StructuralEqual");
   ICHECK(structural_equal) << "node.StructuralEqual is not registered.";
+  // Use the callback until the callback's attribute is rewrite_once=true and has been rewritten
+  std::unordered_map<DFPatternCallback, bool, ObjectPtrHash, ObjectPtrEqual> callbacks_map;
+  for (auto callback : callbacks) {
+    callbacks_map.insert({callback, true});
+  }
   do {
     last = post;
     for (auto callback : callbacks) {
-      callback_ = callback;
-      if (callback_->require_type) {
-        post = InferTypeWithModule(post, mod_);
+      if (callbacks_map[callback]) {
+        auto before = post;
+        callback_ = callback;
+        if (callback_->require_type) {
+          post = InferTypeWithModule(post, mod_);
+        }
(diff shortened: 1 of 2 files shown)


sisleyli:
What do you think @masahi ? Thank you!

src/relay/ir/dataflow_matcher.cc
```diff
@@ -796,24 +796,38 @@ Expr PatternRewriter::Rewrite(const Array<DFPatternCallback>& callbacks, const E
   bool equal = true;
   static auto* structural_equal = runtime::Registry::Get("node.StructuralEqual");
   ICHECK(structural_equal) << "node.StructuralEqual is not registered.";
+  // Use the callback until the callback's attribute is rewrite_once=true and has been rewritten
+  std::unordered_map<DFPatternCallback, bool, ObjectPtrHash, ObjectPtrEqual> callbacks_map;
```
masahi, line 800: `callbacks_map` -> `done`, swapping false and true

src/relay/ir/dataflow_matcher.cc
```diff
@@ -796,24 +796,38 @@ Expr PatternRewriter::Rewrite(const Array<DFPatternCallback>& callbacks, const E
   bool equal = true;
   static auto* structural_equal = runtime::Registry::Get("node.StructuralEqual");
   ICHECK(structural_equal) << "node.StructuralEqual is not registered.";
+  // Use the callback until the callback's attribute is rewrite_once=true and has been rewritten
```
masahi, line 799: // Keep track of callbacks that have finished rewriting

src/relay/ir/dataflow_matcher.cc
sisleyli: Please take another look. Thank you!

masahi (approved):

sisleyli:
> It seems that CI has passed, could you please help me merge this PR? Thanks @masahi !

I don't know whether our repo will automatically rebase when it merges PR, so I just performed a rebase operation myself, do I need to do this every time the main branch is updated?

masahi:
No you don't have to do that. You now have to wait for another 4h :)

sisleyli:
Thanks masahi, I should have consulted before performing the rebase operation :(

sisleyli:
CI passed again...I won't do redundant operations this time.hhhh

PR #14345, opened 2023-03-20
[Hexagon] Adapt some intrinsics for high vector lanes

apeskov:
Goal is to allow to apply intrinsics "q_multiply_shift" and "q_multiply_shift_per_axis" for vector type i32x128. Originally it supports only "i32x32" which is natively supported by platform (1024 bit vector).

**Motivation** 
There are situation than we have to use vector size slightly more than supported by platform. As example consider sequence of element-wise operators: add<i32> -> q_multiply_shift<i32> -> cast<i8>. To achieve performance we have to squash it into one single loop (`sch.compute_at(...)`). First two operators would like to be vectorised with using data type "int32x32". last one cast operator want to use i32x128 as src and i8x128 as dst. As result we have to adapt all this operator to accept vector size "??x128" to successfully vectorise entire loop.

This change allows to achieve significant performance speedup for tuning tasks like `conv -> add -> qnn.requantize -> cast_i8`.

### DIFF
--- a/python/tvm/topi/hexagon/tensor_intrin.py
+++ b/python/tvm/topi/hexagon/tensor_intrin.py
@@ -22,44 +22,163 @@ from tvm.ir import register_intrin_lowering
 from tvm import te
 
 
+def get_lanes(dtype: str):
+    if "x" not in dtype:
+        return 1
+
+    _, lanes = dtype.split("x")
+    return int(lanes)
+
+
+def is_vector_type(dtype: str):
+    return get_lanes(dtype) != 1
+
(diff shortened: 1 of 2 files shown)


apeskov:
@masahi @kparzysz-quic @jverma-quic Previously you reviewed patches like this. Could you please take a look on this one?

masahi (approved):

ibsidorenko:
@tvm-bot rerun

ibsidorenko:
@tvm-bot rerun

PR #14346, opened 2023-03-20
[CI][Lint] Update black

AndrewZhaoLuo:
Goes to the latest revision in the same major version (22.12.0). 

23.1.0 is released but it involves some style changes so we would need to reformat the entire codebase. 

I need 22.12.0 to properly deal with processing some files found in this PR: 

https://github.com/apache/tvm/pull/14167

Where black cannot parse the file in the current version, but can in the updated version.


AndrewZhaoLuo:
cc @driazati

tqchen (approved):

PR #14347, opened 2023-03-20
[TIR] Fix Datatype in Lower TVM Builtin

zxybazh:
This PR fixes the issue of overflow when calculating the size of buffer using product of loop extents. In certain cases it could overflow to a negative number and fail to compile. Unit test is also attached.

Co-authored with Sunghyun Park <<spark@octoml.ai>>

CC: @sunggg @jwfromm @tqchen

### DIFF
--- a/src/tir/transforms/lower_tvm_builtin.cc
+++ b/src/tir/transforms/lower_tvm_builtin.cc
@@ -239,8 +239,9 @@ class BuiltinLower : public StmtExprMutator {
         }
       }
     }
-    PrimExpr total_bytes = make_const(op->extents[0].dtype(), nbytes);
+    PrimExpr total_bytes = make_const(DataType::UInt(64), nbytes);
     for (size_t i = 0; i < op->extents.size(); ++i) {
+      // set total_bytes to uint64 to avoid overflow
       total_bytes = total_bytes * op->extents[i];
     }
     ICHECK(device_type_.defined()) << "Unknown device type in current IR";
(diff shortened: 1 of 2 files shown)


tests/python/unittest/test_tir_transform_lower_tvm_builtin.py
```diff
@@ -172,6 +172,47 @@ def build_tir():
     tvm.testing.assert_allclose(a.numpy(), expected_value)
 
 
+def test_lower_overflow_int32():
+    @T.prim_func
+    def variance4(
```
sunggg, line 177: Can we only keep the minimal reproducible parts?

sunggg (commented):
Thanks, @zxybazh! 
Just quick comment.

tests/python/unittest/test_tir_transform_lower_tvm_builtin.py
zxybazh: Thanks, have updated to the minimal version.

src/tir/transforms/lower_tvm_builtin.cc
```diff
@@ -239,8 +239,9 @@ class BuiltinLower : public StmtExprMutator {
         }
       }
     }
-    PrimExpr total_bytes = make_const(op->extents[0].dtype(), nbytes);
+    PrimExpr total_bytes = make_const(DataType::UInt(64), nbytes);
     for (size_t i = 0; i < op->extents.size(); ++i) {
+      // set total_bytes to uint64 to avoid overflow
```
areusch, line 244: possible for us to also add a guard check here to verify that total_bytes isn't negative? although...i guess it is kinda unlikely to alloc > (1 << 31) worth of space.

tqchen (approved):

src/tir/transforms/lower_tvm_builtin.cc
zxybazh: I think technically it's possible to allocate 2G of memory and also possible to have an intermediate large result between passes. On the other hand each loop extent are supposed to be positive so unless overflow it's unlikely to produce any negative results. I also tried locally to multiply uint64 with negative int32, turns out it will trigger an error. Therefore I think we are good with current change here.

echuraev (approved):

PR #14348, opened 2023-03-21
[Unity] Add More Ops For FX Translator

jinhongyii:
This PR makes 2 changes:
1. Add Relax Op Maximum and Minimum
2. Add translation function for torch function/method silu, to, ones, full, masked_fill_, mean, rsqrt, neg, max in fx translator

cc: @tqchen @MasterJH5574


tests/python/relax/test_frontend_from_fx.py
```diff
@@ -19,20 +19,51 @@
...
+    if use_dynamo:
+        args = []
+        for info in input_info:
+            args.append(torch.zeros(*info[0], dtype=_convert_data_type(info[1])))
+        graph_model = dynamo.export(torch_model, *args)[0]
+    else:
+        graph_model = fx.symbolic_trace(torch_model)
+    print(graph_model.code)
+    mod = from_fx(graph_model, input_info, unwrap_unit_return_tuple=True)
+    print(mod.script())
```
MasterJH5574, line 42: Debugging lines to remove.

tests/python/relax/test_frontend_from_fx.py
```diff
@@ -19,20 +19,51 @@
...
-    mod = from_fx(graph_model, input_info)
+    if use_dynamo:
+        args = []
+        for info in input_info:
+            args.append(torch.zeros(*info[0], dtype=_convert_data_type(info[1])))
+        graph_model = dynamo.export(torch_model, *args)[0]
+    else:
+        graph_model = fx.symbolic_trace(torch_model)
+    print(graph_model.code)
+    mod = from_fx(graph_model, input_info, unwrap_unit_return_tuple=True)
```
MasterJH5574, line 41: For `unwrap_unit_return_tuple=True` I think it depends. Usually for graph from `dynamo.export`, you want to use `unwrap_unit_return_tuple=True`. But for graphs from `fx.symbolic_trace`, it is not necessary to use True. We also want to test the case of `unwrap_unit_return_tuple=False` in this file. Seems that with this change it is untested.

jinhongyii: this is just a workaround to avoid the unit tuple printing bug. If that bug is fixed, we can change this line back later

jinhongyii: maybe I can add a param to decide whether to turn on unwrap_unit_return_tuple for now

tests/python/relax/test_frontend_from_fx.py
```diff
@@ -19,20 +19,51 @@
...
 from tvm import relax
 import tvm.testing
-from tvm.script.parser import ir as I, relax as R, tir as T
+from tvm.script import ir as I
+from tvm.script import relax as R
+from tvm.script import tir as T
 
 
-def verify_model(torch_model, input_info, binding, expected):
+def verify_model(
```
yongwww, line 27: Question (out scope of this pr) for pt converter verification:  Looks we verify the IR, not very sure if we need to verify the inference result as well?

jinhongyii: I'm not sure either. I think we don't need it here given we don't have any low-level implementation in the IR after fx translator.

jinhongyii:
seems the CI failed because its pytorch version is not over 2.0

jinhongyii:
Is there a way to skip some of the tests?

tqchen:
we can use pytest.markskipif()

https://github.com/apache/tvm/blob/unity/tests/python/relax/test_frontend_dynamo.py#L19

if the tests involves dynamo, consider move them to test_frontend_dynamo instead

MasterJH5574 (approved):

PR #14349, opened 2023-03-21
[Unity] Support model kwargs in dynamo_capture_subgraph

jinhongyii:
This PR enables user to pass the torch model's kwargs of forward function to dynamo_capture_subgraph, which makes it more flexible. 

cc: @spectrometerHBH


junrushao (approved):
LGTM

PR #14356, opened 2023-03-21
[Unity] Add support to append relay op attrs in translator

quic-sanirudh:
This PR adds an argument to the relay to relax translator to append the relay Op Attrs as function attributes to the generated TIR PrimFuncs

This information is really useful in Relax as the absence of this prevents us from being able schedule efficiently for ops that are heavily sensitive to the attributes.

For example, the `groups` attribute to `conv2d` op is needed to differentiate between regular conv2d and depthwise conv2d.


quic-sanirudh:
cc @YuchenJin @psrivas2

junrushao (approved):

PR #14357, opened 2023-03-21
[Unity][Fix] Fix block memory plan to handle bool

psrivas2:
This PR fixes the size calculation of storage tokens in Static Plan Block Memory pass. Prior to this PR the storage for a tensor  `R.Tensor((2, 3), dtype="bool")` would be calculated as `R.memory.alloc_storage(R.shape([1]), dtype="bool")` instead of `R.memory.alloc_storage(R.shape([6]), dtype="bool")`

cc @jwfromm @MasterJH5574

### DIFF
--- a/src/relax/transform/static_plan_block_memory.cc
+++ b/src/relax/transform/static_plan_block_memory.cc
@@ -107,8 +107,7 @@ class StorageToken : public ObjectRef {
     }
 
     ObjectPtr<StorageTokenNode> n = make_object<StorageTokenNode>();
-    int bytes_per_element = (dtype.bits() * dtype.lanes() + 7) / 8;
-    n->bytes = size * bytes_per_element;
+    n->bytes = size * dtype.bytes() * dtype.lanes();
     n->dtype = dtype;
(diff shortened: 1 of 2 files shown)


src/relax/transform/static_plan_block_memory.cc
```diff
@@ -107,7 +107,8 @@ class StorageToken : public ObjectRef {
     }
 
     ObjectPtr<StorageTokenNode> n = make_object<StorageTokenNode>();
-    n->bytes = (size * dtype.bits() * dtype.lanes() + 7) / 8;
+    int bytes_per_element = (dtype.bits() * dtype.lanes() + 7) / 8;
+    n->bytes = size * bytes_per_element;
```
MasterJH5574, line 111: Thanks for fixing! Credit to @Ubospica, we found that we can directly use `size * dtype.bytes() * dtype.lanes()`, which is equivalent with your fix.

MasterJH5574 (approved):
LGTM. Thanks for catching my stupid mistake 🙏

PR #14360, opened 2023-03-21
[Unity][Transform] Fix AMP tests

spectrometerHBH:
cc @yongwww


yongwww (approved):
Thanks for the fix!!!

PR #14361, opened 2023-03-21
[Unity][Transform] Common Subexpression Elimination

slyubomirsky:
This PR implements a dataflow block CSE transformation. Since we use ANF internally, the only nesting really occurs with tuples. The pass only needs to look at the RHS of bindings to determine if we have encountered any subexpressions.

Co-authored by @psrivas2  Prakalp Srivastava <prakalp@octoml.ai>


src/relax/transform/eliminate_common_subexpr.cc
```diff
@@ -0,0 +1,209 @@
...
+ public:
+  // overriding VisitExpr ensures we do this for every subexpression
+  void VisitExpr(const Expr& e) override {
+    // Cases we ignore because we will not substitute them:
+    // 1. Vars of all kinds
+    // 2. Op nodes (nothing we can do)
+    // 3. Scalar constants (not much benefit from binding to a var)
+    if (!(e->IsInstance<VarNode>() || e->IsInstance<DataflowVarNode>() ||
+          e->IsInstance<GlobalVarNode>() || e->IsInstance<tvm::OpNode>() ||
+          (e.as<ConstantNode>() && (e.as<ConstantNode>()->is_scalar())))) {
```
slyubomirsky, line 43: It's up to us if we think it's worth rebinding scalar constants. I doubt it, though.

src/relax/transform/eliminate_common_subexpr.cc
```diff
@@ -0,0 +1,209 @@
...
+      }
+      new_blocks.push_back(block);
+    }
+
+    if (all_unchanged) {
+      return GetRef<Expr>(seq);
+    }
+    // do not visit the body
+    return SeqExpr(new_blocks, seq->body, seq->span);
+  }
```
slyubomirsky, line 133: This seemed like a bit of a strange thing to have to implement in a dataflow block pass. It could be avoided if we require lambda-lifting (arguably we should for all DF block passes).

tests/python/relax/test_transform_cse.py
```diff
@@ -0,0 +1,186 @@
...
+                # using canonicalize bindings, eliminate unused bindings, and CSE again
+                lv0 = bar(x)
+                lv1 = lv0
+                lv2 = R.add(lv0, lv1)
+                lv3 = lv0
+                lv4 = lv0
+                lv5 = R.add(lv3, lv4)
+                lv6 = R.add(lv2, lv5)
+                gv = lv6
+                R.output(gv)
```
slyubomirsky, line 179: This is not ideal because there are still repeated additions. @psrivas2 once proposed having a canonicalization pass that runs until fixpoint. This is an example where running such a pass will help.

src/relax/transform/eliminate_common_subexpr.cc
slyubomirsky: Update: Based on the Unity Community Meeting discussion, it doesn't sound like there is much appetite for imposing phase orderings like this, so I would be interested instead if there is a clean way to deal with local functions in dataflow block passes (generalizing the approach shown here, for example). I am sure that other dataflow block passes don't handle the local function case and might exhibit strange bugs if given a program with local functions

spectrometerHBH:
Nice work!

In what scenarios do we want to apply this pass?

slyubomirsky:
The ideal scenario would be if some model (especially an imported one) repeatedly invokes the same operation (an expensive one). We encountered that pattern already in some imported models. CSE is a classic optimization.

spectrometerHBH:
> 

Yeah I agree. One issue that comes to my mind is that it might be critical to decide what subexprs to eliminate. If it's lightweight and inlinable to surrounding ops, then we should probably reject to eliminate the redundancy.

In classical settings, the trade-off of CSE is to enlarge the live range of some vars, which might cause performance regression due to register spill. But for DL workloads, it's likely that the model simply can not be deployed given a certain amount of GPU Memory.

slyubomirsky:
In this case, the bindings will be live only within a single dataflow block, so I don't think there will be many issues with keeping values live for much longer than they would be otherwise.

It would be easy to add a heuristic for deciding when we shouldn't deduplicate.

yongwww (approved):
Looks good to me!

tests/python/relax/test_transform_cse.py
```diff
@@ -0,0 +1,186 @@
...
+
+def test_inner_function():
+    @I.ir_module
+    class Before:
+        @R.function
+        def foo(x: R.Tensor((), dtype="int32")) -> R.Tensor((), dtype="int32"):
+            with R.dataflow():
+                # we are going to do CSE inside the local function
+                @R.function
+                def bar(y: R.Tensor((), dtype="int32")) -> R.Tensor((), dtype="int32"):
```
yongwww, line 126: Not sure if we should handle the case: duplicate gloabl/local functions. I met this case when lower jax to hlo.

                
                @R.function
                def bar1(y: R.Tensor((), dtype="int32")) -> R.Tensor((), dtype="int32"):
                    z = R.add(R.add(y, y), R.add(y, y))
                    with R.dataflow():
                        lv0 = R.add(y, y)
                        lv1 = R.add(y, y)
                        lv2 = R.add(lv0, lv1)
                        gv = lv2
                        R.output(gv)
                    return R.add(z, gv)
                
                @R.function
                def bar2(y: R.Tensor((), dtype="int32")) -> R.Tensor((), dtype="int32"):
                    z = R.add(R.add(y, y), R.add(y, y))
                    with R.dataflow():
                        lv0 = R.add(y, y)
                        lv1 = R.add(y, y)
                        lv2 = R.add(lv0, lv1)
                        gv = lv2
                        R.output(gv)
                    return R.add(z, gv)

                lv0 = bar1(x)
                lv1 = bar2(x)
                lv2 = R.add(lv0, lv1)

slyubomirsky: This actually will handle duplicate local functions. We could do a module-level analysis too for that

yongwww: wow, looks great! module-level analysis for global funcs will be helpful for my case, it could be a todo.

slyubomirsky: To be clear, it would only handle it if the bindings happen inside a `DataflowBlock` :zany_face: 

We might want to consider expanding this pass to handle non-dataflow sections as well. Purity tracking would help with that

PR #14362, opened 2023-03-22
[Unity][WEB] Support async pipeline creation

tqchen:
This PR enables async pipeline creation in webgpu module loading. This will enable us to report progress in shader compilation and leverage multi-threading on the host to compile shaders.


junrushao (approved):
LGTM!

PR #14368, opened 2023-03-22
[Unity][Op] Add stop_lift_params

jinhongyii:
In this PR, I add a "StopLifting" attribute to primfunc to signal the LiftTransformParam Pass to stop lifting a call_tir and any other var bindings that consumes it. This is useful in cases like low-bit quantization, where we add encode and decode function and we only want to lift the encode function.

cc: @vinx13 @tqchen


include/tvm/relax/expr.h
```diff
@@ -983,6 +983,8 @@ constexpr const char* kCodegen = "Codegen";
 constexpr const char* kComposite = "Composite";
 /*! \brief Indicate the function was created by the Pattern Partitioning Pass. */
 constexpr const char* kPartitionedFromPattern = "PartitionedFromPattern";
+/*! \brief Indicate the function cannot be lifted to transform_params*/
+constexpr const char* kStopLifting = "StopLifting";
```
tqchen, line 987: can we instead add a special relax op, `relax.builtin.stop_lifting`, which is an identity op, and can be used to mark the boundary

jinhongyii:
@tqchen please review again

python/tvm/relax/op/unary.py
```diff
@@ -527,3 +527,23 @@ def isnan(x: Expr) -> Expr:
         The computed result.
     """
     return _ffi_api.isnan(x)  # type: ignore
+
+
+###################### Indicator operators ######################
+
+
+def stop_lift_params(x: Expr) -> Expr:
```
MasterJH5574, line 535: I feel it might better for us to categorize this op as a builtin one? It is a nop except the indicator functionality. The “unary”/“binary” class is basically dedicated for arithmetic operations.

src/relax/transform/lift_transform_params.cc
```diff
@@ -144,6 +144,8 @@ class TransformParamsFuncBuilder : public ExprMutator {
  */
 class LiftTransformParamsPlanner : public ExprVisitor {
  public:
+  explicit LiftTransformParamsPlanner(IRModule mod) : mod_(mod) {}
+
```
MasterJH5574, line 148: Just to confirm, are we using the IRModule somewhere? I seems to find no use in both LiftTransformParamsPlanner and TransformParamsLifter.

python/tvm/relax/op/unary.py
tqchen: agree builtin is better

tqchen:
cc @vinx13 please also help and take a look

src/relax/transform/lift_transform_params.cc
jinhongyii: sorry it's some code used for attr. since we have changed to op, we can delete them

vinx13 (approved):

PR #14369, opened 2023-03-22
[Unity] Update CUTLASS submodule to match `main`

kparzysz-quic:
Having the same submodule versions makes it easier to manage branch merges and rebases.


vinx13:
It is intended as there are some features in unity branch depending on newer cutlass version. What about updating main to match unity?

kparzysz-quic:
> It is intended as there are some features in unity branch depending on newer cutlass version. What about updating main to match unity?

Sorry, I just realized that `main` has the older version...  Nevermind.

Issue #14351, opened 2023-03-21
[Bug] Raise error when call tvmc.run whose arg tvmc_package compiled by vm


lxiongh:
https://github.com/apache/tvm/blob/36b30974a9e7c5f2244c8fade392af1570db3c1d/python/tvm/driver/tvmc/runner.py#L641

should be:
```python
if not isinstance(exe_outputs, (list, tvm.runtime.container.ADT)):
```

lxiongh:
Here is error detail:
```code
2023-03-21 11:46:37.694 INFO load_module /tmp/tmpws1rwd_s/lib.so
Traceback (most recent call last):
  File "run_pkg_vm.py", line 9, in <module>
    result = tvmc.run(package, device="cuda", repeat=10, number=10, benchmark=True, end_to_end=True, inputs=inputs)
  File "/home/haoke/.local/lib/python3.7/site-packages/tvm/driver/tvmc/runner.py", line 646, in run_module
    outputs[output_name] = val.numpy()
  File "/home/haoke/.local/lib/python3.7/site-packages/tvm/runtime/object.py", line 67, in __getattr__
    raise AttributeError("%s has no attribute %s" % (str(type(self)), name)) from None
AttributeError: <class 'tvm.runtime.container.ADT'> has no attribute numpy
```

leandron:
Hello @lxiongh - would you be able to push a PR with that fix?

PR #14377, opened 2023-03-23
[TVMC] fix error when call tvmc.run for vm package

lxiongh:
fix issue #14351


lxiongh:
@leandron here is fix for #14377 , please help to review, thanks

leandron:
> @leandron here is fix for #14377 , please help to review, thanks

Thank you very much for the contribution. I just noticed we probably got this bug because we don't have any test to validate vm compilation. Would you mind adding a test perhaps turning the python API?

You can look for all existing tests in https://github.com/apache/tvm/tree/main/tests/python/driver/tvmc

PR #14380, opened 2023-03-23
[FuseOps][Relax][Fix] Do not replace ConstantNodes with Vars in FuseOps

janetsc:
Bug fix, ConstantNodes should not be replaced with Vars.

Looking at unit tests now to see how to verify this - tests/python/relax/test_transform_fuse_ops.py

cc: @csullivan @farshidsp

### DIFF
--- a/src/relay/transforms/fuse_ops.cc
+++ b/src/relay/transforms/fuse_ops.cc
@@ -520,7 +520,7 @@ class FuseMutator : private MixedModeMutator {
       auto* arg_group = gmap_.at(arg.get())->FindRoot();
       auto type = arg->checked_type();
       Expr new_arg = this->Mutate(arg);
-      if (current_group != arg_group) {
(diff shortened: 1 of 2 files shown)


Hzfengsy:
Thanks @janetsc for the PR. I acknowledge your valuable findings and use cases, however, I think it's still controversial:

- On the one side, inlining constants may help to bring more information and optimization opportunities to the callee functions.
- On the other side, it may introduce similar functions if we call the same function with different args. e.g. the following case will generate two different functions:
    ```python
    y = func(x, 1)
    z = func(x, 2)
    ```
There is a related discussion at https://github.com/apache/tvm/pull/14097

In this particular case, I recommend using another pass to inline the constant into fused functions after FuseOps.

janetsc:
Hello!  Yes, I meant to close this as it has already been addressed right below where I added my changes.  I think that change went in with the migration to tlc-pack/relax.  (Which still has the problem.)

PR #14382, opened 2023-03-23
[Rewrite] Fix rewriting division to constant to handle vector case

AndrewZhaoLuo:
Previously only handle scalar case. That is rewriting `a ./ 5 --> a .* 0.2`. Now handle vector case `a ./ [2, 3, 4] --> a .* [0.5, 0.33, 0.25]`

Also changes the pass to rely on constant folding for proper folding instead of manually constructing constant data.

cc @Johnson9009  @tkonolige

### DIFF
--- a/src/relay/transforms/div_to_mul.cc
+++ b/src/relay/transforms/div_to_mul.cc
@@ -27,10 +27,11 @@ namespace tvm {
 namespace relay {
 
 template <typename T>
-inline bool const_has_values(size_t size, const ConstantNode* const_node, const std::vector<T>&& values) {
+inline bool const_has_values(size_t size, const ConstantNode* const_node,
+                             const std::vector<T>&& values) {
   for (size_t i = 0; i < size; i++) {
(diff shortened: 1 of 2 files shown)


src/relay/transforms/div_to_mul.cc
```diff
@@ -26,42 +26,61 @@
...
+            LOG(WARNING) << "Unknown dtype not handled for div_to_mull: " << rhs->data.DataType();
             return post;
           }
-          return Multiply(call_node->args[0], Constant(inv));
+
+          if (const_has_zero_flag) {
+            return post;
+          }
+
+          // rely on constant folding to fold things
```
tkonolige, line 82: Are you sure constant folding is going to fire at the right time? I can't remember if the divisions need to be rewritten before FakeQuantization or if it doesn't matter when.

AndrewZhaoLuo: Hmm good point, I think there is a way to specify a required pass to be run after this one, so maybe will try to figure that out.

PR #14388, opened 2023-03-23
[Unity][Op] Conv1d

LeshengJin:
This PR implements `Conv1d`.

Unit tests are provided accordingly.

### DIFF
--- a/include/tvm/relax/attrs/nn.h
+++ b/include/tvm/relax/attrs/nn.h
@@ -29,6 +29,49 @@
 namespace tvm {
 namespace relax {
 
+/*! \brief Attributes used in Conv1d operator */
+struct Conv1DAttrs : public tvm::AttrsNode<Conv1DAttrs> {
+  Array<IntImm> strides;
+  Array<IntImm> padding;
+  Array<IntImm> dilation;
+  int groups;
+  String data_layout;
+  String kernel_layout;
+  String out_layout;
+  DataType out_dtype;
+
+  TVM_DECLARE_ATTRS(Conv1DAttrs, "relax.attrs.Conv1DAttrs") {
+    TVM_ATTR_FIELD(strides).describe("Specifies the strides of the convolution.");
+    TVM_ATTR_FIELD(padding).describe(
+        "If padding is non-zero, then the input is implicitly zero-padded"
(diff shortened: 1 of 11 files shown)


include/tvm/relax/attrs/nn.h
```diff
@@ -29,6 +29,49 @@
 namespace tvm {
 namespace relax {
 
+/*! \brief Attributes used in Conv1d operator */
+struct Conv1DAttrs : public tvm::AttrsNode<Conv1DAttrs> {
```
yongwww, line 33: looks it is quite similar to Conv2Dattrs, probably we can unify them into a ConvAttrs

MasterJH5574: It seems possible to unify them together. We can do sanity checks in respective struct info inference of conv1d and conv2d.

tqchen: Let us keep them separate for now

tests/python/relax/test_tvmscript_parser_op_nn.py
```diff
@@ -35,6 +35,24 @@ def _check(
         tvm.ir.assert_structural_equal(parsed, expect)
 
 
+def test_conv1d():
+    @R.function
+    def foo(
+        x: R.Tensor((2, 3, 228), "float32"), w: R.Tensor((16, 3, 5), "float32")
+    ) -> R.Tensor((2, 16, 224), "float16"):
+        gv: R.Tensor((2, 16, 224), "float16") = R.nn.conv1d(x, w, out_dtype="float16")
```
Hzfengsy, line 43: Not a big problem, but it is not a common case to take fp32 as input and fp16 as output :-)

Hzfengsy (approved):

tests/python/relax/test_tvmscript_parser_op_nn.py
tqchen: right, would be good to change x and w to float16 in here

LeshengJin: Fixed
Thank you for the review!

tqchen (approved):

PR #14389, opened 2023-03-23
[Unity] Improve error message in VerifyWellFormed

Lunderberg:
Previously, the error message gave the name of the variable that should not be used within the current block.  This commit updates the error message to also state where that variable was defined, and which block contains the erroneous use of the variable.


Hzfengsy:
Thanks @Lunderberg for the improvement. BTW, could you please send this PR to `main` instead of `unity` as it's not relax-related at all

Lunderberg:
Thank you for the catch, and the PR is now rebased onto main, and set to merge into main instead of unity.

Lunderberg:
@tvm-bot rerun

PR #14390, opened 2023-03-23
[TVMScript][Fix] Fix `bool` printing for roundtrip

cyx-6:
This PR fixes the TVMScript printer to print `T.bool(True)` or `T.bool(False)` for `IntImm(dtype=bool)`. So the parser is able to parse it back.


zxybazh (approved):
Thanks for the quick fix, LGTM!

MasterJH5574 (approved):

Hzfengsy (approved):

PR #14393, opened 2023-03-24
[CI][microTVM] Enable USE_MICRO for mac and windows CI builds

alanmacd:
Enable USE_MICRO for mac and windows CI builds to ensure code building doesn't break for those platforms;

### DIFF
--- a/conda/recipe/bld.bat
+++ b/conda/recipe/bld.bat
@@ -27,6 +27,7 @@ cmake ^
       -DUSE_LLVM=ON ^
       -DUSE_RPC=ON ^
       -DUSE_CPP_RPC=ON ^
+      -DUSE_MICRO=ON ^
       -DUSE_SORT=ON ^
       -DUSE_RANDOM=ON ^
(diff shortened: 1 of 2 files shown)


mehrdadh:
@driazati Will these changes be tested in CI or it has to be from a committer?

leandron:
> @driazati Will these changes be tested in CI or it has to be from a committer?

So the GitHub action that runs on Windows seems to have picked up the proposed change: https://github.com/apache/tvm/actions/runs/4506447598/jobs/7933495133?pr=14393#step:4:335

Based on that, I'd say it already run in CI.

leandron (approved):
LGTM, thanks @alanmacd @mehrdadh

driazati:
Yeah as @leandron saw these do get tested, the only things that need to be from a branch in apache/tvm are the jenkinsfiles and associated scripts (anything in `ci/scripts/jenkins/*` and `ci/jenkins/generated/*`)

PR #14394, opened 2023-03-24
[Unity][IR] Purity Tracking

slyubomirsky:
In this PR, I am beginning to implement the tracking of function purity as part of the `StructInfo` system. This will allow the compiler to enforce that no impure function (one that can possibly have visible side effects) can be called in a `DataflowBlock`. Tracking this requires noting which operators are pure or impure (I am presently doing this with an operator attribute called `FPurity`, a simple boolean), though dealing with calls to other Relax functions means that this information must also be tracked via the `StructInfo` system. 

Additionally, it is difficult to infer the purity of a function in the general case (when there are calls to other Relax functions), so this change does require users to annotate impure functions using a new field on functions, `is_pure` (in TVMScript, this can be done using `R.is_pure()` or `R.is_impure()`). Since most Relax functions are likely to be pure and purity is the default assumption, this will hopefully not be a large imposition on users. We can consider inferring purity in the easier cases, since those are likely to be common.

Note that `PackedFunc`s are conservatively treated as impure. However, in situations where they are needed inside a dataflow block, a call to a `PackedFunc` that is, in reality, pure can be done via the new operator `call_pure_packed` or the existing operator `call_dps_packed` (it is assumed that any `PackedFunc` used with it will be pure). (Similarly, a new operator `invoke_pure_closure` is introduced as a counterpart to `invoke_closure` for dealing with closure objects, though this really should only come up with the `LambdaLifting` pass.)

As an "escape hatch" to the purity system, one can use the attribute `relax.force_pure`, which indicates to the compiler to treat the entire function as pure even if it contains an impure call. Additionally, even though `PackedFunc`s are normally treated as impure, a user can use `call_pure_packed` or `call_dps_packed` to call `PackedFunc`s in dataflow blocks when appropriate. These can be used to deal with the following situations:
1. A function does side effects but only on a value that will not be exposed anywhere else or on a new value that will be returned. Even though the individual actions are "impure," the overall function fulfills the definition of being pure. `relax.force_pure` would be useful here.
2. A `PackedFunc` is, in reality, pure. `call_pure_packed` or `call_dps_packed` are useful in this situation.

Changes include:
* Enforcing that impure functions are not used in `DataflowBlock`s in the well-formed check.
* Enforcing that functions that are not labeled impure do not contain impure calls (unless `relax.force_pure` is set).
* Implementing the `call_pure` operator

Still to be done:
- [x] Label purity for all operators
- [x] Address certain tricky bugs

(Also address the design concerns below)

----

The process of implementing purity tracking has revealed a few issues that may require some further design discussion.

## Labeling purity or impurity

Using function attributes to label purity/impurity seems very messy. It might be worth making this part of the `Function` node's AST to avoid having to wrangle attributes in many places in the codebase.

## ~~The treatment of `call_pure`~~

~~`call_pure` presents a dilemma because many passes look for calls to certain operators, but with `call_pure`, these would be "wrapped" like so: `Call(Op("relay.call_pure"), [inner_op, arg1, ..., argn], attrs, sinfo_args)`. In the posted PR, many passes needed to be revised to look for instances of `call_pure` so that they could be treated the same as calls to ordinary operators. It might cause less disruption to passes to turn `call_pure` into a specialized AST node that literally wraps a call node. This way visitors or mutators could have an easy default case and the lower-level code generation passes could simply ignore the `call_pure` node and deal with the wrapped call. As painful as introducing a new AST node would be, it would likely be more maintainable than having every pass have to have a special case for the `call_pure` operator.~~

The more restricted `call_pure_packed`/`invoke_pure_closure` are less likely to come up in passes and so are less likely to need special handling.

## Staging

Related to the above issue, having purity tracking in the well-formedness analysis means that even low-level passes like `VMLowerBuiltin`, which replaces some Relax operators with `PackedFunc`s (treated as impure by default, as they are in principle dangerous black boxes), have to insert calls to `call_pure` to avoid triggering purity errors. Perhaps it might be useful to ignore purity by the time we reach lower stages of compilation, much as the `ToNonDataflow` pass is used in the VM build process to eliminate `DataflowBlock`s.

**Update**: Based on the TVM Unity Community Meeting on Mar. 28, 2023, I've provisionally adopted a variant of this approach thanks to a suggestion by @tqchen. Namely, there is a pass called `RemovePurityChecking` that simply adds the `ForcePure` attribute to all pure functions and unwraps all invocations of `call_pure_packed`, `call_pure_dps_packed`, and `invoke_pure_closure`. This allows lower-level passes to remain unchanged and not to have to consider purity at all. It is inserted after `ToNonDataflow` in the `build()` function in `vm_build.py`.

**Update 2**: Thanks to suggestions by @tqchen, I've made `is_pure` and `force_pure` into fields on functions rather than attributes and added syntactic sugar in TVMScript to make these easy to set.

**Update 3**: Per @tqchen's suggestion `force_pure` will be kept as an attribute instead. This is because it acts more as a direction to the compiler rather than an inherent property of a function.


tests/python/relax/test_analysis_contains_impure_call.py
```diff
@@ -0,0 +1,104 @@
...
+        body.blocks[0].bindings[0],
+        body.blocks[0].bindings[1],
+        body.blocks[0].bindings[-1],
+    ]
+    new_body = rx.SeqExpr([rx.BindingBlock(new_bindings)], body.body)
+
+    # if we didn't ignore the recursive call, the fact the var's StructInfo
+    # calls it impure would throw it off
+    assert not contains_impure_call(new_body, own_name=own_name)
+    assert contains_impure_call(new_body)
```
slyubomirsky, line 100: Somehow the second call fails. The `StructInfo` for the `GlobalVar` `recursive_impure` indicates that it is pure (which is wrong), whereas the `StructInfo` for `RecursiveTest["recursive_impure"]` (the `Function` node itself) correctly shows that it is impure. Why would this happen? Is there something in `block_builder.cc` that I would have to update?

I would greatly appreciate any help.

src/relax/transform/legalize_ops.cc
```diff
@@ -85,21 +87,46 @@ class LegalizeMutator : public ExprMutator {
...
+    // for call_pure, legalize the inner call
+    if (op == call_pure_op) {
+      auto inner_call = Call(call->args[0], Array<Expr>(call->args.begin() + 1, call->args.end()),
+                             call->attrs, call->sinfo_args);
+      auto res = VisitExpr_(inner_call.as<CallNode>());
+      if (res.as<CallNode>()) {
+        return WrapCallPure(Downcast<Call>(res));
+      }
+      return res;
+    }
```
slyubomirsky, line 111: One (of many) examples of having to have special handling for `call_pure`.

src/relax/backend/vm/vm_builtin_lower.cc
```diff
@@ -81,35 +83,59 @@ class VMBuiltinLowerMutator : public ExprMutator {
       }
       return ShapeExpr({ret});
     } else {
-      return Call(builtin_compute_alloc_shape_, {shape, DataTypeImm(dtype)}, Attrs(),
-                  {GetStructInfo(shape)});
+      // TODO(@slyubomirsky): Find a way to register these builtins as pure to avoid needing to emit
+      // call_pure each time
+      return WrapCallPure(Call(builtin_compute_alloc_shape_, {shape, DataTypeImm(dtype)}, Attrs(),
+                               {GetStructInfo(shape)}));
```
slyubomirsky, line 89: An example (one of many) of having to insert `call_pure` because of the use of `PackedFunc`s at low levels of compilation.

src/relax/op/op.cc
```diff
@@ -292,7 +349,10 @@ RELAY_REGISTER_OP("relax.invoke_closure")
     .set_num_inputs(2)
     .add_argument("closure", "Expr", "The VMClosure.")
     .add_argument("args", "Tuple", "The captured variables.")
-    .set_attr<FInferStructInfo>("FInferStructInfo", InferStructInfoInvokeClosure);
+    .set_attr<FInferStructInfo>("FInferStructInfo", InferStructInfoInvokeClosure)
+    // TODO: This might be another case where we would want a macro or even use an attr.
+    // It may depend on the particulars of the closure
+    .set_attr<Bool>("FPurity", Bool(false));
```
slyubomirsky, line 355: Here is an example where using a simple boolean for indicating the purity of an op might not work out. I don't know if there is any way to figure out whether a given VM closure is pure or not. 

Alternatively, if we take the view that we should not track purity past a certain point in compilation (e.g., by the time we get rid of dataflow blocks), maybe this is less of an issue.

tests/python/relax/test_analysis_contains_impure_call.py
```diff
@@ -0,0 +1,104 @@
...
+
+    assert not contains_impure_call(PureTest["pure_func"])
+
+
+def test_simple_impure_case():
+    @tvm.script.ir_module
+    class ImpureTest:
+        @R.function
+        def impure_func() -> R.Object:
+            R.func_attr({"IsPure": False})
```
slyubomirsky, line 42: I plan on adding syntactic sugar for the purity attributes (`IsPure` and `ForcePure`).

tests/python/relax/test_analysis_contains_impure_call.py
slyubomirsky: Bug fixed. It had to do with the parsing of function declarations (attributes were not normally parsed at that point).

python/tvm/script/parser/relax/parser.py
```diff
@@ -220,7 +220,30 @@ def visit_tvm_declare_function(self: Parser, node: doc.FunctionDef) -> GlobalVar
...
+                if isinstance(key, doc.Constant) and key.value == "IsPure":
+                    index = i
+                    break
+            if index is not None:
+                val = item.value.args[0].values[index]
+                if isinstance(val, doc.Constant):
+                    purity = bool(val.value)
+                    attrs = make_node("DictAttrs", IsPure=purity)
+
+    func_signature = relax.Function.create_empty(params, ret_sinfo, attrs=attrs)
```
slyubomirsky, line 246: This seems very messy, but it was needed to prevent a bug where a function would be parsed and the global var would have the wrong `StructInfo` (the GV would be pure even if the function was impure). It turned out to be due to the parsing of function declarations.

We could consider some alternatives, like using a decorator for purity/impurity instead of an attribute, or checking the types of global vars during normalization.

slyubomirsky:
This will probably keep failing tests because lots of them depend on syntactic matches and this has messed up a lot of it. I will start going through them, but some have proven very difficult to debug.

tests/python/relax/test_tvmscript_parser.py
```diff
@@ -1309,6 +1309,9 @@ def add(
 
         @R.function
         def main(x: R.Tensor((2, 4), dtype="float32")) -> R.Tensor((10,), dtype="float32"):
+            # slight hack: normally, we would prefer to use True, but the func attrs, when printed,
+            # will have it as 1, so it would fail roundtripping otherwise
+            R.func_attr({"ForcePure": 1})
```
slyubomirsky, line 1314: @yongwww, would you happen to know how I could avoid this (print the attribute as `True` instead of 1)? It's not super-pressing, but it would help for roundtripping

More important: Would you be able to advise how we could have a special case in the pretty-printer for `call_pure`? I would like to show it properly wrapping the "inner call"

src/relax/backend/vm/vm_builtin_lower.cc
slyubomirsky: No longer needed with `RemovePurityChecking`

slyubomirsky:
More general question (maybe one to leave to a community meeting): Where should we communicate that we expect certain passes to be run before others? We already have `ToNonDataflow` and this PR adds `RemovePurityChecking`. I also saw that a line in `VMShapeLower` indicates that it expects `LambdaLifting` to have already been used.

src/script/printer/relax/call.cc
```diff
@@ -132,13 +132,75 @@ Optional<ExprDoc> PrintCallTIRDPSPacked(const relax::Call& n, const ObjectPath&
...
+  // is the _format_ string, or else roundtripping will fail
+  // (the format string will be interpreted as an argument and there will be a new default format
+  // string given)
+  ExprDoc first_arg = d->AsDoc<ExprDoc>(n->args[0], n_p->Attr("args")->ArrayIndex(0));
+  Array<ExprDoc> args;
+  for (size_t i = 1; i < n->args.size(); i++) {
+    args.push_back(d->AsDoc<ExprDoc>(n->args[i], n_p->Attr("args")->ArrayIndex(i)));
+  }
+  return Relax(d, "print")->Call(args, {"format"}, {first_arg});
+}
```
slyubomirsky, line 183: One would think this isn't related to these changes at all, but in the process of testing roundtripping for `call_pure`, I discovered roundtripping bugs for `print` and `assert_op` (both having to do with their format strings).

tests/python/relax/test_transform_legalize_ops_manipulate.py
```diff
@@ -1080,7 +1081,8 @@ def main(x: R.Tensor(("a", "b", "c"), "float32")) -> R.Tensor(("b", 1), "float32
     # fmt: on
 
     mod = LegalizeOps()(CollapseSumTo)
-    tvm.ir.assert_structural_equal(mod, Expected)
+    # TODO(@relax-team): Uncomment when this is supported
+    # tvm.ir.assert_structural_equal(mod, Expected)
```
slyubomirsky, line 1085: These were giving false negatives for my IDE due to having an undefined variable. We probably should not keep tests that don't work

tests/python/relax/test_codegen_dnnl.py
```diff
@@ -72,6 +72,7 @@ def test_dnnl_offload():
 
     seq = tvm.transform.Sequential(
         [
+            relax.transform.RemovePurityChecking(),
```
slyubomirsky, line 75: This is another case where communicating expectations on phase ordering might be helpful. Do we want to deal with `call_pure` in codegen passes and the like? Probably not.

src/relax/transform/run_codegen.cc
```diff
@@ -102,7 +102,13 @@ class CodeGenRunner : ExprMutator {
...
           func = (*RemoveFuncAttrFunc)(func, attr::kCodegen);
           builder_->UpdateFunction(gvar, func);
-          return create_call_dps_packed(new_func, func->ret_struct_info);
+          // preserve the purity: if the func was originally pure, wrap call_pure
+          bool purity = GetStructInfoAs<FuncStructInfoNode>(gvar)->purity;
+          auto ret = create_call_dps_packed(new_func, func->ret_struct_info);
+          if (purity) {
+            return WrapCallPure(ret);
+          }
+          return ret;
```
slyubomirsky, line 111: This is another case where being clear about staging might make some of this reasoning unnecessary. We do codegen before doing the VM build in `test_codegen_dnnl.py`, so I wasn't sure if I should have eliminated dataflow blocks. The fusion passes seem to rely on having dataflow blocks, so I can't trivially get rid of purity checking and `call_pure` before fusion (unless we also disable checking purity from the well-formed check).

src/relax/op/op.cc
```diff
@@ -177,7 +227,9 @@ TVM_REGISTER_OP("relax.call_builtin_with_ctx")
     .set_num_inputs(4)
```
yongwww, line 227: just noticed this should be 2, it is not related to your change...

src/relax/op/op.cc
```diff
@@ -177,7 +227,9 @@ TVM_REGISTER_OP("relax.call_builtin_with_ctx")
...
     .add_argument("func", "Expr", "The builtin packed func.")
     .add_argument("args", "Tuple", "The input arguments.")
-    .set_attr<FInferStructInfo>("FInferStructInfo", InferStructInfoCallBuiltinWithCtx);
+    .set_attr<FInferStructInfo>("FInferStructInfo", InferStructInfoCallBuiltinWithCtx)
+    // TODO(relax-team): Please verify if these are normally impure or not
```
yongwww, line 231: seems the existing builtin packedfuncs are pure

src/relax/op/tensor/manipulate.cc
```diff
@@ -1357,7 +1370,8 @@ TVM_REGISTER_OP("relax.cumsum")
     .set_attrs_type<CumsumAttrs>()
     .set_num_inputs(1)
     .add_argument("data", "Tensor", "The input tensor.")
-    .set_attr<FInferStructInfo>("FInferStructInfo", InferStructInfoCumsum);
+    .set_attr<FInferStructInfo>("FInferStructInfo", InferStructInfoCumsum)
+    .set_attr<Bool>("FPurity", Bool(true));
```
yongwww, line 1374: will we run into issue if FPurity is not specified explicitly?

slyubomirsky: We should specify it explicitly. I have an assert in the well-formedness checker that it's defined.

python/tvm/relax/analysis/analysis.py
```diff
@@ -276,6 +276,34 @@ def has_reshape_pattern(func: tir.PrimFunc) -> bool:
     return _ffi_api.has_reshape_pattern(func)  # type: ignore
 
 
+def contains_impure_call(expr: Expr, own_name: Optional[Union[Var, GlobalVar]] = None) -> bool:
+    """
+        Check if the given expression (likely a function body) contains any impure calls.
```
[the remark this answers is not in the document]
slyubomirsky: Will do, thanks

src/relax/op/op.cc
slyubomirsky: Good observation

python/tvm/script/parser/relax/parser.py
```diff
@@ -220,7 +220,31 @@ def visit_tvm_declare_function(self: Parser, node: doc.FunctionDef) -> GlobalVar
...
+    # (we don't need any other attributes at the function declaration stage)
+    attrs = None
+    for item in node.body:
+        if (
+            isinstance(item, doc.Expr)
+            and isinstance(item.value, doc.Call)
+            and isinstance(item.value.func, doc.Attribute)
+            and item.value.func.attr == "func_attr"
+            and len(item.value.args) == 1
+            and isinstance(item.value.args[0], doc.Dict)
```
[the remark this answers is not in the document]
slyubomirsky: Is there a good place to put such a function?

slyubomirsky:
Tests failing due to the rebase (have to update some more ops/passes), will address shortly.

src/relax/op/op.cc
slyubomirsky: Are we sure all of them are?

slyubomirsky:
Points from the the April 4, 2023, community meeting, proposed by @tqchen:

1. Should we replace `call_pure` with `call_pure_packed`? Almost all uses of `call_pure` in the tests are for packed funcs. Using `call_pure_packed` would mean that passes would only need to look for `call_pure_packed` if they have anything to do with `PackedFunc`s.

There is one other case that comes up in the tests, which is calling a `PrimFunc` directly via its `GlobalVar` (not via `call_tir`). Would it be reasonable to use `call_pure_packed` for that as well? If so, then this would be a reasonable replacement.

2. Possible syntactic sugar we could use for purity/impurity: `@R.function(pure=False)`.

Having the optional `pure` argument in `function` would require a level of indirection to implement due to the way Python decorators work, but this could work.
```python
def function(pure=False):
     def inner_function(func: FType):
          # parse the function...
     return inner_function 
```
Would we be fine with this? I believe this would have to be in `python/tvm/script/parser/relax/entry.py`, for `function`. @yongwww, does that seem accurate to you, or are there other places I would have to add that parameter?

Any opinions on these options?

tests/python/relax/test_transform_legalize_ops_manipulate.py
yongwww: +1 remove the tests that don't work, not difficult to get it back from the commit history if would like to enable them

yongwww:
I feel call_pure_packed is a little weird for calling PrimFunc. It seems to me that `call_pure_external` is solely used for ExternFunc. Perhaps we can keep it as call_pure if we don't have a better name, and change it whenever we get a better name proposal. The `R.function(pure=False)` looks fine to me, and yes I think the entry is the right place for it.  Feel free to leave a todo for me if that doesn't work.

tests/python/relax/test_tvmscript_parser.py
yongwww: it related printing should be https://github.com/apache/tvm/blob/unity/src/script/printer/relax/function.cc#L54-L58. Feel free to leave a todo for me

yongwww (approved):
Overall look good to me! Thanks for adding this helpful feature!

sunggg:
Thanks for leading this effort, @slyubomirsky! 
Great work. 

> There is one other case that comes up in the tests, which is calling a PrimFunc directly via its GlobalVar (not via call_tir).

Would you share the pointer? Wondering how the case looks like.

In general, I'm with @yongwww. I also feel a little strange to use `call_pure_packed` for calling PrimFunc since it does not look consistent.

psrivas2:
> There is one other case that comes up in the tests, which is calling a PrimFunc directly via its GlobalVar

Wondering what is the use case for this? Should we just disallow this as we already have call_tir for this functionality.

psrivas2:
> Possible syntactic sugar we could use for purity/impurity: @R.function(pure=False).
syntactic sugar looks great! What is the default value?

slyubomirsky:
> > There is one other case that comes up in the tests, which is calling a PrimFunc directly via its GlobalVar
> 
> Wondering what is the use case for this? Should we just disallow this as we already have call_tir for this functionality.

I'll have to find the example. It comes up in some of the lower-level passes, or maybe it's only in the test cases. We should decide if we want to prohibit this usage.

slyubomirsky:
> > Possible syntactic sugar we could use for purity/impurity: @R.function(pure=False).
> syntactic sugar looks great! What is the default value?

I think the default should be true, as most functions people will write are going to be pure.

tests/python/relax/test_transform_legalize_ops_manipulate.py
slyubomirsky: Removed. If others object, it's easy to restore them.

slyubomirsky:
After discussion, @yongwww and I will defer adding syntactic sugar to a follow-up PR, since it's separate from implementing the core functionality.

src/relax/op/op.cc
slyubomirsky: I noticed that there are builtins for updating a key-value store (those related to the `attention_kv_cache`). Updates are not pure, so that's a counterexample

slyubomirsky:
@psrivas2 Cases where `PrimFunc`s are called directly and not via `call_tir`: It appears to be created by the `CallTIRRewrite` pass. If I have the well-formedness checker throw an error whenever it encounters a direct call to a `PrimFunc`, any test using `call_tir` and the default VM build will fail. :sweat_smile:

include/tvm/relax/utils.h
```diff
@@ -81,6 +81,42 @@ TVM_DLL bool IsBoolStructInfo(const StructInfo& sinfo, bool permit_unknown_rank
...
+/*!
+ * \brief Wrap the Call node in the call_pure op, transferring over the attributes and sinfo_args.
+ *
+ * \param call The input call
+ *
+ * \return A Call to the call_pure op that wraps the original call.
+ *
+ * \note Transfers over StructInfo from the input to the return value.
+ */
+TVM_DLL Call WrapCallPure(const Call& call);
```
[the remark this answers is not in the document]
slyubomirsky: Okay, I'll do it and we'll see if it proves sufficient.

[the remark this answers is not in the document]
slyubomirsky: @psrivas2 points out that `call_packed_dps` would also need a pure wrapper, so I will make a `call_pure_packed_dps` as well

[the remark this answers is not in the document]
slyubomirsky: Another nasty case: lambda-lifting lowers functions into closures, which may or may not be pure. This means `invoke_closure` also needs a pure variant.

slyubomirsky:
I've made the change over to `call_pure_packed` and a few other counterparts. You can review as to whether this is a simplification. It allowed me to eliminate the special handling in the parser and pretty printer and simplified a few cases. There was some additional boilerplate as a result, but ultimately it wasn't too bad.

tests/python/relax/test_analysis_contains_impure_call.py
```diff
@@ -0,0 +1,104 @@
...
+    # but if we remove the impure call...
+    body = RecursiveTest["recursive_impure"].body
+    own_name = body.blocks[0].bindings[-1].value.op
+    # skipping the call to print...
+    new_bindings = [
+        body.blocks[0].bindings[0],
+        body.blocks[0].bindings[1],
+        body.blocks[0].bindings[-1],
+    ]
+    new_body = rx.SeqExpr([rx.BindingBlock(new_bindings)], body.body)
```
psrivas2, line 95: can we create the new function using TVMScript again? Not sure if using rx.* API is making it easier to follow or saving space.

src/relax/analysis/well_formed.cc
```diff
@@ -56,6 +56,14 @@
...
  *           * Inside the fields of Tuple nodes
  *    13. Expr always has checked_type_ (with the exception of Op).
+ *    14. DataflowBlocks may not contain If nodes.
+ *    15. DataflowBlocks may not contain calls to impure functions or operators
+ *        (only checked if check_struct_info is true).
+ *    16. If a function is annotated as pure (kIsPure is true)
+ *        and purity is not forced (kForcePure is true),
+ *        the body may not contain any impure call
+ *        (only checked if check_struct_info is true).
+ *    17. If a function's purity is forced, kForcePure cannot be true
```
psrivas2, line 66: I don't understand `17`. If the function's purity is forced, wouldn't kForcePure have to be set as `true`?

src/relax/op/image/resize.cc
```diff
@@ -122,7 +122,8 @@ TVM_REGISTER_OP("relax.image.resize2d")
     .add_argument("size", "Shape", "The output image shape.")
     .set_attr<FInferStructInfo>("FInferStructInfo", InferStructInfoResize2D)
     .set_attr<FRelaxInferLayout>("FRelaxInferLayout", InferLayoutResize2d)
-    .set_attr<TMixedPrecisionPolicy>("TMixedPrecisionPolicy", MixedPrecisionPolicyKind::kFollow);
+    .set_attr<TMixedPrecisionPolicy>("TMixedPrecisionPolicy", MixedPrecisionPolicyKind::kFollow)
+    .set_attr<Bool>("FPurity", Bool(true));
```
psrivas2, line 126: seems like all ops have to be marked as pure except `call_dps_packed` and `call_packed`.
cc @junrushao  if there is any implication of this for op_schema which is WIP.

src/relax/op/op.cc
psrivas2: this concern is resolved now with ForcePure and RemoveTrackingPurity, right?

src/relax/op/op.cc
```diff
@@ -450,7 +538,9 @@ TVM_REGISTER_GLOBAL("relax.op.memory.alloc_tensor").set_body_typed(MakeMemAllocT
 RELAY_REGISTER_OP("relax.memory.kill_storage")
     .set_num_inputs(1)
     .add_argument("storage", "Expr", "The storage to be killed.")
-    .set_attr<FInferStructInfo>("FInferStructInfo", ReturnVoidStructInfo);
+    .set_attr<FInferStructInfo>("FInferStructInfo", ReturnVoidStructInfo)
+    // deallocation also isn't considered a "visible effect" as far as purity is concerned
+    .set_attr<Bool>("FPurity", Bool(false));
```
psrivas2, line 543: the comment says that deallocation is not considered a visible effect, yet it is marked as impure?

src/relax/op/op.cc
```diff
@@ -464,7 +554,9 @@ TVM_REGISTER_GLOBAL("relax.op.memory.kill_storage").set_body_typed(MakeMemKillSt
 RELAY_REGISTER_OP("relax.memory.kill_tensor")
     .set_num_inputs(1)
     .add_argument("tensor", "Expr", "The tensor to be killed.")
-    .set_attr<FInferStructInfo>("FInferStructInfo", ReturnVoidStructInfo);
+    .set_attr<FInferStructInfo>("FInferStructInfo", ReturnVoidStructInfo)
+    // memory deallocation also isn't considered a "visible effect" as far as purity is concerned
```
psrivas2, line 558: same as above

psrivas2 (approved):
Overall LGTM after addressing the pending `tensor_to_shape` bug and a few comments. Thanks for all the detailed hard work on this!

src/relax/op/op.cc
slyubomirsky: Ooh, that's a mistake

src/relax/op/op.cc
slyubomirsky: Sort of, it's resolved with `invoke_pure_closure`. See the changes to `LambdaLifting`: it reasons about whether the function that was turned into a closure is pure.

src/relax/analysis/well_formed.cc
slyubomirsky: Sorry, that was a typo. `kIsPure` cannot be false in that case. Will fix.

slyubomirsky:
FWIW, the `tensor_to_shape` issue has been "fixed" by using `call_pure_packed` directly (the old legalization used `call_packed` directly), I would just argue that this was not a great way to legalize it in the first place.

tests/python/relax/test_analysis_contains_impure_call.py
slyubomirsky: In this case, I needed to keep the old vars with the old struct info. If I had rewritten the function, it would be pure and there would be no weird behavior. The point of this setting was in case a situation like this happens while making changes to a module/rebuilding a module.

slyubomirsky: I'll add a comment to make that clear.

slyubomirsky:
@tvm-bot rerun

yongwww:
@slyubomirsky tvm-bot doesn't work for this branch, probably a new push is required to trigger the ci

slyubomirsky:
Okay, I'll push something lol

include/tvm/relax/utils.h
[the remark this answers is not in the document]
tqchen: How about we assume `call_dps_packed` to be pure? If we want to be extra careful we can name as `call_dps_pure_packed`. I cannot think of a case where impure dps call can be too useful

src/relax/utils.cc
```diff
@@ -111,6 +113,75 @@ bool IsLeafOrTuple(const Expr& expr) {
...
+
+  // transfer over struct info if we can
+  if (call->struct_info_) {
+    UpdateStructInfo(ret, GetStructInfo(call));
+  }
+  return ret;
+}
+
+Call UnwrapCallPure(const Call& call) {
+  static const Op& call_pure_packed_op = Op::Get("relax.call_pure_packed");
```
tqchen, line 156: Now that we have purity per operator, do we still need generic unwrapping? Likely we can simplify by directly detect and derive using is_pure attribute of these functions

src/relax/utils.cc
```diff
@@ -111,6 +113,75 @@ bool IsLeafOrTuple(const Expr& expr) {
...
+    static auto purity_map = Op::GetAttrMap<Bool>("FPurity");
+    ICHECK(purity_map.count(op)) << "Cannot find the registered purity of this op: " << op->name;
+    return !(purity_map[op]->value);
+  }
+  // the StructInfo must be FuncStructInfo
+  auto func_struct_info = GetStructInfoAs<FuncStructInfoNode>(call->op);
+  return !func_struct_info->purity;
+}
+
+Call WrapCallPure(const Call& call) {
```
tqchen, line 128: Same here, once we have special purity deduction, per op, likely we don;t need to rely on wrapping unwrapping in most cases

src/relax/op/op.cc
```diff
@@ -73,6 +73,40 @@ StructInfo InferStructInfoShapeOf(const Call& call, const BlockBuilder& ctx) {
...
+  auto callee = call->args[0];
+  ICHECK(!callee.as<OpNode>()) << "call_pure_packed cannot be used with an op node";
+  auto opt = MatchStructInfo<FuncStructInfo>(callee);
+  ICHECK(opt) << "Callee must have a function struct info";
+  FuncStructInfo finfo = opt.value();
+  ICHECK(finfo->IsOpaque()) << "call_pure_packed must be called with an opaque function, but "
+                            << callee << " is not opaque";
+
+  // derives the struct info of the result as it would for a call to the inner args
+  auto hypothetical_call = UnwrapCallPure(call);
```
tqchen, line 94: We can directly have a derivation rule similar to call packed instead of relying on wrapping unwrapping

tqchen (changes requested):
Thanks for making the updates. My remaining comments mainly stands on removing wrapping unwrapping and rely on the original set of deductions for each intrinsic(and perhaps we can have common funcs among them)

src/relax/op/op.cc
slyubomirsky: This is doing the exactly same thing as the StructInfo inference for packed calls, in fact. I adapted it from `BlockBuilder::InferStructInfo`. I unwrapped the call to avoid code duplication

src/relax/utils.cc
slyubomirsky: I could split this to handle the different op variants. I kept them together for convenience, but there's no inherent reason they need to be kept together. It's convenient in passes to unwrap and rewrap, but I'll see how it will look if I get rid of these helpers.

src/relax/op/op.cc
tqchen: How about we have the same common function InferStructInfoCallPacked  and register that same function for call_pure_packed and call_packed as they are similar

python/tvm/relax/transform/lazy_transform_params.py
```diff
@@ -166,11 +165,15 @@ def visit_var_binding_(self, binding: relax.VarBinding) -> None:
...
+                # Properly speaking, this pass should require ToNonDataflow first,
+                # but the liveness analysis requires dataflow blocks. This should be refactored
+                relax.call_pure_packed(
                     relax.ExternFunc("set_item"),
-                    [index, var_before_setitem],
-                    None,
-                    [relax.ObjectStructInfo()],
+                    index,
+                    var_before_setitem,
+                    sinfo_args=(relax.ObjectStructInfo(),),
```
slyubomirsky, line 176: This is a total hack to get tests to pass, see bug report #14829. I don't think we should allow this call to be inserted inside a dataflow block.

include/tvm/relax/expr.h
```diff
@@ -985,6 +985,11 @@ constexpr const char* kComposite = "Composite";
 constexpr const char* kPartitionedFromPattern = "PartitionedFromPattern";
 /*! \brief The required workspace for an external function. */
 constexpr const char* kWorkspaceSize = "WorkspaceSize";
+/*! \brief Indicate whether the function is pure (has no visible side effects for any input). */
+constexpr const char* kIsPure = "IsPure";
+/*! \brief Indicate whether the function should be considered pure even if it contains
+ * an impure call. */
+constexpr const char* kForcePure = "ForcePure";
```
tqchen, line 992: One minor nit: we now hav two kinds of patterns for attribute name

- stl_style: `global_symbol`, `tir.noalias`
- CamelStyle: as in this case.

Would be great for us to consolidate a bit to a single style. Some of the CameStyle were inherited from relay. To be consistent with some of the more frequently used annotation (like global symbol)

include/tvm/relax/expr.h
```diff
@@ -985,6 +985,11 @@ constexpr const char* kComposite = "Composite";
 constexpr const char* kPartitionedFromPattern = "PartitionedFromPattern";
 /*! \brief The required workspace for an external function. */
 constexpr const char* kWorkspaceSize = "WorkspaceSize";
+/*! \brief Indicate whether the function is pure (has no visible side effects for any input). */
+constexpr const char* kIsPure = "IsPure";
```
tqchen, line 989: my understanding is that is_pure is more like an annotation of the function, like ret_struct_info, it might make sense to make it directly as an attribution of the function (`Optional<Bool> purity_annotation`). Can be done as a followup

include/tvm/relax/expr.h
slyubomirsky: Okay, I'll rename it. I'm not a fan of CamelCase in general ;)

include/tvm/relax/expr.h
slyubomirsky: I agree, actually. I brought up the issue before and at that time we didn't want to change the constructor. Maybe it's a better idea to do it in this PR to avoid even more disruption later.

slyubomirsky: The main issue is how parsing would work. I'll see if I'll be able to do something about that

slyubomirsky: I've implemented this suggestion. Please have a look at how it's turned out.

include/tvm/relax/expr.h
slyubomirsky: I've eliminated the need to use these attributes, so this will no longer be an issue.

slyubomirsky:
@psrivas2 et al, any idea what could be causing the test failure in `test_op_gradient_numeric.py`? It works for me locally. Is this just flakiness?

include/tvm/relax/expr.h
```diff
@@ -920,10 +920,16 @@ class FunctionNode : public BaseFuncNode {
   Expr body;
   /*! \brief The return type of the function. */
   StructInfo ret_struct_info;
+  /*! \brief Whether the function is annotated as pure or not. */
+  bool is_pure;
```
tqchen, line 924: Thanks for getting these in. One thing that I think would worth distinguish is purity annotation (which is more of a UX thing) and `force_pure`. I know it is getting into preference part :) So it is debatable.

- is_pure or purity annotation is something that I think is worth to be part of function annotation, since this is what we expect most users to annotate and is consistent with ret_struct_info
- force_pure on the other hand, is closer to attributes like `tir.noalias`, which affects the compiler internal, and might makes sense to remain as attribute `relax.force_pure` for now.

On the syntax side, would be nice if we can enable things like (can serve as a followup)
```
@R.function(pure=True, private=True)
def func(...)
```

slyubomirsky: I tried to implement that but it gets really complicated because of how Python decorators work (namely, having optional arguments while still using `@R.function` without any explicit arguments). I'm sure it's probably possible but I wasn't able to get that working. Maybe it could be a separate decorator, like `@R.pure`.

slyubomirsky: So would you say that I should have `force_pure` remain as an attribute, then?

tqchen: right, would be good to keep force_pure as an attr.

interesting, didn't think about that issue(optional argument while having R.function without explicit argument), perhaps allowing also a default for R.function can work. We can do that as a followup

slyubomirsky: Implemented the change back to an attr

slyubomirsky:
Updated to reflect the fix from #14864, thanks to @jinhongyii :smile:

tqchen (approved):

tqchen (approved):

PR #14396, opened 2023-03-24
[Unity] Support simple dynamic-shape-aware fusion

Hzfengsy:
This PR adds support for simple dynamic-shape-aware fusion, which is the first step towards supporting dynamic shapes. The main changes are as follows:

- Fix FuncStructInfo in well-formed checks
- Renew symbolic var defs in fuse_ops to prevent malformed functions

cc @tqchen @MasterJH5574 @jinhongyii

### DIFF
--- a/src/relax/ir/expr_functor.cc
+++ b/src/relax/ir/expr_functor.cc
@@ -577,7 +577,10 @@ Expr ExprMutator::VisitExpr_(const FunctionNode* op) {
   for (Var param : op->params) {
     Var new_param = this->VisitVarDef(param);
     params.push_back(new_param);
-    all_params_unchanged &= param.same_as(new_param);
+    if (!param.same_as(new_param)) {
+      var_remap_[param->vid] = new_param;
+      all_params_unchanged = false;
+    }
   }
 
   Expr body = this->VisitWithNewScope(op->body, params);
--- a/src/relax/transform/fuse_ops.cc
+++ b/src/relax/transform/fuse_ops.cc
@@ -33,6 +33,7 @@
 #include <tvm/relax/expr_functor.h>
 #include <tvm/relax/struct_info.h>
 #include <tvm/relax/transform.h>
+#include <tvm/tir/expr_functor.h>
 #include <tvm/tir/function.h>
 
 #include <optional>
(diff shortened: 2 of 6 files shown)


src/relax/ir/expr_functor.cc
```diff
@@ -752,6 +752,7 @@ Var ExprMutator::VisitVarDef(const Var& var) {
   } else {
     LOG(FATAL) << "TypeError: Invalid type: " << var->GetTypeKey();
   }
+  this->var_remap_[var->vid] = ret;
```
tqchen, line 755: I am not too sure if we wantr to do directly remap change here, ideally such remap should be changed in the caller side, either function def or binding

tqchen (changes requested):

src/relax/transform/fuse_ops.cc
```diff
@@ -344,6 +345,42 @@ class GraphCreator : public ExprVisitor {
...
+
+ private:
+  SymbolicVarRenewMutator() = default;
+  using relax::ExprMutator::VisitExpr;
+  using relax::ExprMutator::VisitExpr_;
+  using tir::ExprMutator::VisitExpr_;
+
+  PrimExpr VisitPrimExpr(const PrimExpr& expr) final { return tir::ExprMutator::VisitExpr(expr); }
+
+  PrimExpr VisitExpr_(const tir::VarNode* op) final {
```
tqchen, line 368: this is the use side, let us update in the def side, then update. the use side, which is safer. It will also allow us to only renew symbolic vars that are defined within function scope, and report error for symbolic vars that are not bound

This is something that is useful in general and we can make it as part of analysis.

Hzfengsy: Thanks TQ, I agree that your approach is safer. However, in this particular case, I think both ways are the same. I would leave a TODO here for further updates.

BTW, there is not an easy/formal way to visit symbolic var at def side, we may add a method to the base visitor/mutator later.

src/relax/ir/expr_functor.cc
Hzfengsy: We only call `VisitVarDef` at the definition side, e.g. the param or binding. If we mutate the var `x` to `x'` at the definition side, all use side needs to be updated, since `x` no longer exists. 

Could you please show an example to indicate what I missed?

Hzfengsy: After offline discussion, we agree we move the logic to the `VisitExpr_(FunctionNode*)`, thanks again for pointing it out

tqchen (approved):

masahi:
Hi @Hzfengsy, it seems this change broke one of CUTLASS BYOC tests: `pytest test_codegen_cutlass.py::test_matmul_offload[float16-x_shape17-y_shape17-False-none-none] -s`.

If I remove `SymbolicVarRenewMutator::Renew`, the test works. I don't know why it creates an undefined sym var for a BYOC function. Can you take a look?

Hzfengsy:
Hi @masahi, thanks for reporting the bug. 

`SymbolicVarRenewMutator::Renew` works to re-define all symbolic vars to prevent the same symbolic var in different functions. However, it does not update the out_sinfo of the body. 

I will send a PR to fix it soon

Hzfengsy:
@masahi, After a deep dive, it's may not the issue of FuseOps, as the output is well-formed and pass the check. The error is reported at the pass `RunCodegen`

masahi:
Yes, it might be a bug in `RunCodegen`. And IRs produced by with or without `SymbolicVarRenewMutator::Renew` look identical at least textually. So I'm confused what makes `SymbolicVarRenewMutator::Renew` break `RunCodegen`.

masahi:
>  However, it does not update the out_sinfo of the body

Why not? Does this mean `out_sinfo` refer to the old sym var?

Hzfengsy:
> Why not? Does this mean out_sinfo refer to the old sym var?

Sorry, it updated. My bad

masahi:
https://github.com/apache/tvm/blob/unity/src/relax/transform/run_codegen.cc#L105

Maybe we shouldn't use the original function's `ret_struct_info` there. After `RunCodegen`, it indeed references a dubious sym var:
```
@R.function                                                                                                                                          
def main(x: R.Tensor((1, 1, 16, 15), dtype="float16"), y: R.Tensor((3, 2, "a", 15, 2), dtype="float16")) -> R.Tensor((3, 2, "a", 16, 2), dtype="float
16"):                                                                     
    a = T.int64()                                                         
    a_1 = T.int64()                                                       
    with R.dataflow():                                                                                                                               
        gv = R.call_dps_packed("fused_relax_matmul_cutlass", (x, y), out_sinfo=R.Tensor((3, 2, a_1, 16, 2), dtype="float16"))                     
      R.output(gv)                                                      
    return   
```

masahi:
@Hzfengsy Is this line in `ExprMutator` still correct?

https://github.com/apache/tvm/blob/unity/src/relax/ir/expr_functor.cc#L592

I think the mutated function would continue referring to the old sym var if the sym var is used in a return sinfo.

Hzfengsy:
> I think the mutated function would continue referring to the old sym var if the sym var is used in a return sinfo.

We need to update the `ret_struct_info`. However, I'm suprised that the undefined var in ret_sinfo is not reported by well-formed checker

masahi:
I think this is a problem unique to BYOC. `RunCodegen` replaces a call to a function with an opaque `call_dps_packed` call. After `RunCodegen` runs, the original function is gone but we still need its `ret_struct_info`.

masahi:
The fix has been added in the PR https://github.com/apache/tvm/pull/14291

PR #14399, opened 2023-03-24
[TVMScript] Improved error message for unexpected top frame

Lunderberg:
Previously, the error messasge from `FindPrimFuncFrame`, `FindBlockFrame`, and `FindIfFrame` stated that they could not find the requested frame when the top-most frame did not match the requested type.  This error message could be misinterpreted by a user as stating that the frame didn't exist at all.

This commit updates the error message to distinguish between the case of a missing frame (e.g. `T.reads()` occurring outside of any `T.block()` frame) and a frame not being top-most (e.g. `T.reads()` occurring inside a `T.block()`, but inside an `if` conditional instead of the top of the block).


cyx-6 (approved):
LGTM!

PR #14400, opened 2023-03-24
[TIR] Update LowerTVMBuiltin to use Optional<T>

Lunderberg:
The member variables `device_type_` and `device_id_` are tracked during `LowerTVMBuiltin`, are initially undefined, and are checked before use.  This is the usage pattern expected for `Optional<T>`, but these variables previously had the type `PrimExpr`.

This commit updates these to `Optional<PrimExpr>`, and uses the helper function `GetDeviceMethodName` to add more detail to the error messages when they are undefined.


csullivan:
@tvm-bot rerun

masahi (approved):

PR #14402, opened 2023-03-24
[Unity][MetaSchedule] Skip Scheduled PrimFuncs in Task Generation

zxybazh:
<s> This PR introduced a check to skip tasks that has been thread-binded in previous passes or manually so that we can process them differently and avoid failures like for `cumsum`. </s>

This PR introduced a check to skip tasks that has been scheduled either by `DefaultGPUSchedule` pass or the `MetaScheduleApplyDatabase` pass.

CC: @sunggg @jwfromm

### DIFF
--- a/python/tvm/meta_schedule/relax_integration.py
+++ b/python/tvm/meta_schedule/relax_integration.py
@@ -125,7 +125,7 @@ def extracted_tasks_to_tune_contexts(
         get_loggers_from_work_dir(work_dir, [t.task_name for t in extracted_tasks]),
         fork_seed(seed, n=len(extracted_tasks)),
     ):
-        if task.mod.attrs.get("tir.is_scheduled", False):
+        if task.mod.attrs is not None and task.mod.attrs.get("tir.is_scheduled", False):
             warnings.warn("The task {task.task_name} is already scheduled, skipping it.")
             continue
         tasks.append(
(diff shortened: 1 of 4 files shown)


junrushao:
I vaguely remembered that there was a PR a while ago adding a tir function attribute that indicates a PrimFunc has been scheduled already, but cannot find it out now :-((

zxybazh:
Yeah if we can have such attribute to mark the primfunc, it would be even more convenient and generic to targets outside cuda. Specifically we are looking into the case where we have a cumsum operator manually scheduled -- where we want to avoid doing that again in tuning pass. The is kinda new because when we design MetaSchedule we are not expecting scheduled workloads and now it's possible given MS is a pass in the TVM Unity.

tqchen:
cc @spectrometerHBH  recently introduce a tir.is_scheduled attr to mark already scheduled program in #14266

junrushao:
Oh actually this is called `tir::attr::kIsScheduled` introduced in #14266

zxybazh:
This PR has been modified to utilize `kIsScheduled` attribute to skip tasks, and will automatically mark PrimFuncs as scheduled in `DefaultGPUSchedule` pass. Would you please take another look, thanks!

psrivas2 (approved):
LGTM!

tqchen (approved):

PR #14406, opened 2023-03-26
[Runtime] Introduce runtime module property

sunggg:
Currently, we only classify whether the runtime module is DSO-exportable.

This PR further classifies each runtime module into the properties as follows:
- `kBinarySerializable`: we can serialize the module to the stream of bytes. CUDA/OpenCL/JSON runtime are representative examples.
- `kRunnable`: we can run the module directly. LLVM/CUDA/JSON runtime, executors (e.g, virtual machine) runtimes are runnable. Non-runnable modules, such as CSourceModule, requires a few extra steps (e.g,. compilation, link) to make it runnable.
- `kBinaryExportable`: when the module is kBinarySerializable and kRunnable, we consider this module as binary exportable. A binary exportable module can be integrated into final runtime artifact by being serialized as data into the artifact, then deserialzied at runtime. This class of modules must implement `SaveToBinary`, and have a matching deserializer registered as `runtime.module.loadbinary_<type_key>`.
- `kDSOExportable`: we can export the module as DSO. A DSO exportable module (e.g., a CSourceModuleNode of type_key 'c') can be incorporated into the final runtime artifact (ie shared library) by compilation and/or linking using the external compiler (llvm, nvcc, etc). DSO exportable modules must implement `SaveToFile`.

Please note that `kDSOExportable` is a mutual exclusive property with `kBinaryExportable`. 

To represent each category, the following status bits are defined.
```
namespace property {
/*! \brief Binary serializable runtime module */
constexpr const uint8_t kBinarySerializable = 0b001;
/*! \brief Runnable runtime module */
constexpr const uint8_t kRunnable = 0b010;
/*! \brief DSO exportable runtime module */
constexpr const uint8_t kDSOExportable = 0b100;
/*! \brief Binary exportable runtime module */
constexpr const uint8_t kBinaryExportable = kBinarySerializable | kRunnable;
};  // namespace property
```

## Edit based on the review

This PR further classifies each runtime module into the properties as follows:
- `kBinarySerializable`: we can serialize the module to the stream of bytes. CUDA/OpenCL/JSON runtime are representative examples. A binary exportable module can be integrated into final runtime artifact by being serialized as data into the artifact, then deserialized at runtime. This class of modules must implement `SaveToBinary`, and have a matching deserializer registered as `runtime.module.loadbinary_<type_key>`.
- `kRunnable`: we can run the module directly. LLVM/CUDA/JSON runtime, executors (e.g, virtual machine) runtimes are runnable. Non-runnable modules, such as `CSourceModule`, requires a few extra steps (e.g,. compilation, link) to make it runnable. 
- `kDSOExportable`: we can export the module as DSO. A DSO exportable module (e.g., a `CSourceModuleNode` of type_key 'c') can be incorporated into the final runtime artifact (ie shared library) by compilation and/or linking using the external compiler (llvm, nvcc, etc). DSO-exportable modules must implement `SaveToFile`. In general, DSO exportable modules are not runnable unless there is a special support like JIT for `LLVMModule`. 

To represent each category, the following status bits are defined.
```
enum ModulePropertyMask : int {
  kBinarySerializable = 0b001,
  kRunnable = 0b010,
  kDSOExportable = 0b100
};
```

cc. @tqchen

### DIFF
--- a/include/tvm/runtime/module.h
+++ b/include/tvm/runtime/module.h
@@ -45,21 +45,17 @@ namespace runtime {
  * \brief Property of runtime module
  * We classify the property of runtime module into the following categories.
  * - kBinarySerializable: we can serialize the module to the stream of bytes. CUDA/OpenCL/JSON
- * runtime are representative examples.
+ * runtime are representative examples. A binary exportable module can be integrated into final
+ * runtime artifact by being serialized as data into the artifact, then deserialzied at runtime.
+ * This class of modules must implement SaveToBinary, and have a matching deserializer registered as
+ * 'runtime.module.loadbinary_<type_key>'.
  * - kRunnable: we can run the module directly. LLVM/CUDA/JSON runtime, executors (e.g,
  * virtual machine) runtimes are runnable. Non-runnable modules, such as CSourceModule, requires a
  * few extra steps (e.g,. compilation, link) to make it runnable.
- * - kBinaryExportable: when the module is kBinarySerializable and kRunnable, we consider this
- * module as binary exportable. A binary exportable module can be integrated into final runtime
- * artifact by being serialized as data into the artifact, then deserialzied at runtime. This class
- * of modules must implement SaveToBinary, and have a matching deserializer registered as
- * 'runtime.module.loadbinary_<type_key>'.
  * - kDSOExportable: we can export the module as DSO. A DSO exportable module (e.g., a
  * CSourceModuleNode of type_key 'c') can be incorporated into the final runtime artifact (ie shared
  * library) by compilation and/or linking using the external compiler (llvm, nvcc, etc). DSO
  * exportable modules must implement SaveToFile.
- *
- * Please note that kDSOExportable is a mutual exclusive property with kBinaryExportable.
  */
 namespace property {
 /*! \brief Binary serializable runtime module */
@@ -68,8 +64,6 @@ constexpr const uint8_t kBinarySerializable = 0b001;
 constexpr const uint8_t kRunnable = 0b010;
 /*! \brief DSO exportable runtime module */
 constexpr const uint8_t kDSOExportable = 0b100;
-/*! \brief Binary exportable runtime module */
-constexpr const uint8_t kBinaryExportable = kBinarySerializable | kRunnable;
 };  // namespace property
 
 class ModuleNode;
--- a/include/tvm/runtime/module.h
+++ b/include/tvm/runtime/module.h
@@ -223,7 +223,7 @@ class TVM_DLL ModuleNode : public Object {
   virtual int GetProperty() const { return 0b000; }
 
   /*! \brief Returns true if this module is 'DSO exportable'. */
-  bool IsDSOExportable() const { return GetProperty() == ModulePropertyMask::kDSOExportable; };
+  bool IsDSOExportable() const { return GetProperty() & ModulePropertyMask::kDSOExportable; };
 
   /*!
    * \brief Returns true if this module has a definition for a function of \p name. If
--- a/include/tvm/runtime/module.h
+++ b/include/tvm/runtime/module.h
@@ -41,6 +41,37 @@
 namespace tvm {
 namespace runtime {
 
+/*!
+ * \brief Property of runtime module
+ * We classify the property of runtime module into the following categories.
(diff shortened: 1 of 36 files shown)


include/tvm/runtime/module.h
```diff
@@ -41,6 +41,37 @@
...
+ * - kDSOExportable: we can export the module as DSO. A DSO exportable module (e.g., a
+ * CSourceModuleNode of type_key 'c') can be incorporated into the final runtime artifact (ie shared
+ * library) by compilation and/or linking using the external compiler (llvm, nvcc, etc). DSO
+ * exportable modules must implement SaveToFile.
+ *
+ * Please note that kDSOExportable is a mutual exclusive property with kBinaryExportable.
+ */
+namespace property {
+/*! \brief Binary serializable runtime module */
+constexpr const uint8_t kBinarySerializable = 0b001;
```
tqchen, line 66: let us use int t keep things simple, make it an enum class

include/tvm/runtime/module.h
```diff
@@ -41,6 +41,37 @@
...
+/*! \brief Runnable runtime module */
+constexpr const uint8_t kRunnable = 0b010;
+/*! \brief DSO exportable runtime module */
+constexpr const uint8_t kDSOExportable = 0b100;
+/*! \brief Binary exportable runtime module */
+constexpr const uint8_t kBinaryExportable = kBinarySerializable | kRunnable;
```
tqchen, line 72: skip this one

src/runtime/contrib/tflite/tflite_runtime.h
```diff
@@ -60,6 +60,9 @@ class TFLiteRuntime : public ModuleNode {
    */
   const char* type_key() const { return "TFLiteRuntime"; }
 
+  /*! \brief Get the property of the runtime module .*/
+  uint8_t GetProperty() const final { return property::kRunnable; };
```
tqchen, line 64: make the return value as int as that is more common in bitmasking

sunggg: Done

include/tvm/runtime/module.h
sunggg: Used `enum ModulePropertyMask : int` instead for implicit casting.

echuraev (commented):
Thank you for the PR. I have one question. Where is it planned to use these properties? In this PR I see only adding these properties but no use-cases. Probably a simple unit test will be helpful to understand what should be achieved after enabling these properties.

sunggg:
Thank you, @echuraev for the review. As @tqchen pointed out, this PR changes the way we achieve the same functionality but with a more refined manner. This PR was motivated by the discussion in my outstanding PR https://github.com/apache/tvm/pull/14337 about roundtrip-abilty and with this change, we can offer more systematic error message when we cannot guarantee this.

junrushao:
This is a really comprehensive summary and I believe it will generally be quite useful to anyone (including me) who wants to understand runtime.Module in depth. @sunggg @tqchen shall we put it somewhere in the official tutorial?

echuraev:
@sunggg Thank you for your reply! It is a nice summary and the reason why we did it starts to be much clearer for me. I'm ok with current tests but agree with @junrushao that creating some documentation might be very useful.

include/tvm/runtime/module.h
```diff
@@ -221,10 +221,20 @@ class TVM_DLL ModuleNode : public Object {
    * By default, none of the property is set. Derived class can override this function and set its
    * own property.
    */
-  virtual int GetProperty() const { return 0b000; }
+  virtual int GetPropertyMask() const { return 0b000; }
+
+  /*! \brief Returns true if this module is 'binary serializable'. */
+  bool IsBinarySerializable() const {
```
tqchen, line 227: I feel we don;t need this method. Instead, we can just expose GetProperty and let python side use this function

sunggg:
@junrushao @echuraev Thank you for the suggestion. Addition to the official tutorial also sounds great to me.

By reflecting @tqchen's suggestion, I also added some unit tests and assertion in export_library.
Hope this helps!

mehrdadh:
@tvm-bot rerun

echuraev (approved):
LGTM. Thanks!

tqchen (approved):

PR #14408, opened 2023-03-27
[MetaSchedule][Hexagon] Improve vectorization for standalone elementwise op

ibsidorenko:
**Motivation:**
It was found that for standalone elementwise operations (add, sub, etc.) MetaScheduler generates code with poor performance due to lack of vector code on some input tensor shapes. Current implementation is not able to vectorize if innermost loops extent is not multiple of the vector length.

**What was done:**
Core changes: it checks current loops nest, if all loops are "simple", i.e. loops without annotations, bindings, reduce axis, then it does the following:
 1) Fuse all loops into single one.
 2) Split this new loop into 2 parts: inner and outer. Herewith split factor for the inner loop is equal to '_max_vectorize_extent_' MetaScheduler parameter.
 3) Parallelize outer loop and vectorize inner loop.

**Performance measurement**:
Measurement was done on Qualcomm Snapdragon 888. As it was expected, 1 and 2 got significant performance boost, 3 and 4 - without changes.

N |    op   | Dtype |      Shape       | Before fix, ms | After fix, ms | speedup |
--|---------|-------|------------------|----------------|---------------|---------|
1 | add     | uint8 | 1, 8, 56, 56, 32 |      1.264     |     0.167     |  7.5x   |
2 | qnn.add | uint8 | 1, 8, 56, 56, 32 |      2.213     |     0.336     |  6.6x   |
3 | add     | int32 | 1, 8, 56, 56, 32 |      0.161     |     0.150     |  1.07x  |
4 | seq*    | uint8 | 1, 64, 56, 56    |      2.634     |     2.679     |  0.98x  |

seq* - test of the ops sequence: qnn.conv2d + bias_add + qnn.requantize,
       weights shape = [256, 64, 1, 1]

### DIFF
--- a/src/meta_schedule/postproc/rewrite_parallel_vectorize_unroll.cc
+++ b/src/meta_schedule/postproc/rewrite_parallel_vectorize_unroll.cc
@@ -144,13 +144,40 @@ void RemoveParsedAnn(const Schedule& sch, const BlockRV& block_rv, const ParsedA
   }
 }
 
+int CalculateNumRewritableLoops(const Array<StmtSRef>& loop_srefs,
+                                const std::vector<int>& loop_types) {
+  int rw_loops_num = 0;
+  ICHECK_EQ(loop_srefs.size(), loop_types.size());
+  for (size_t i = 0; i < loop_srefs.size(); ++i) {
+    const StmtSRef& loop_sref = loop_srefs[i];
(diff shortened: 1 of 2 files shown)


ibsidorenko:
@tvm-bot rerun

masahi (approved):

PR #14410, opened 2023-03-27
[TOPI] Add instance_norm operator

vinx13:
This PR added `instance_norm` operator in TOPI. It's based on tuple reduction that computes mean and variance in one-pass, which is faster than previous implementation. The underlying algorithm is the same as layer norm so this operator is delegated to the layer norm implementation

cc @MasterJH5574 @zxybazh


vinx13:
@tvm-bot rerun

PR #14417, opened 2023-03-28
[Unity] Remove non-deterministic behavior from graph pattern matching 

masahi:
Graph pattern matching uses `std::map` and `std::set` keyed on pointers, which leads to non-deterministic matching result. This PR removed all non-determinism. 

For example, consider matching against QKV projection before attention. Given a sequence of bindings,
```
with R.dataflow():
    lv0 = R.matmul(x, w0)
    lv1 = R.matmul(x, w1)
    lv2 = R.matmul(x, w2)
    ...
```

and patterns,
```
with PatternContext() as ctx:
    inp_pat = wildcard()
    Q_weight_pat = wildcard()
    K_weight_pat = wildcard()
    V_weight_pat = wildcard()

```

intuitively I expect the following matching result:
```
Q_weight_pat - w0
K_weight_pat - w1
V_weight_pat - w2
```

But since three matmul patterns are independent of each other, each pattern can match any of three matmuls in the bindings. And since graph pattern matching processes patterns and variables in the bindings in a non-deterministic order, matching result becomes non deterministic as well. 

To make graph pattern matching deterministic, I introduced an implicit ordering constraint on the patterns: Patterns are matched in the order that they are declared in the pattern context. So in the example above, we make sure to start pattern matching from the first matmul pattern. Likewise, the variables in the bindings, which are the candidates for matching, are also considered for matching in the order they appear in the bindings. So `lv0` is matched first, followed by `lv1` etc.

@ganler @sunggg

### DIFF
--- a/src/relax/ir/dataflow_matcher.cc
+++ b/src/relax/ir/dataflow_matcher.cc
@@ -739,8 +739,8 @@ Map<DFPattern, Var> MatchGraph(const PatternContext& ctx, const DataflowBlock& d
 
   if (start_hint) {
     auto rnode_ptr = var2node.at(start_hint.value().get());
-    for (auto& [df_pattern, pattern_node] : pattern2node) {
-      if (try_match(&pattern_node, &rnode_ptr, &matcher, def2use, caller2callees)) {
+    for (auto& p_node : pattern2node) {
+      if (try_match(&p_node.second, &rnode_ptr, &matcher, def2use, caller2callees)) {
         for (const auto& [df_pattern, pattern_node] : pattern2node)
           ret.Set(GetRef<DFPattern>(df_pattern), GetRef<Var>(pattern_node.matched));
         return ret;
--- a/src/relax/ir/dataflow_matcher.cc
+++ b/src/relax/ir/dataflow_matcher.cc
@@ -555,7 +555,7 @@ static bool try_match(PNode* p, RNode* r, DFPatternMatcher* m,
       ICHECK_EQ(p->matched, r->ptr);
       return;
     }
-    ICHECK_EQ(r->matched, nullptr);
+    ICHECK(r->matched == nullptr);
     p->matched = r->ptr;
     r->matched = p->ptr;
     undo_stack.emplace(p, r);
--- a/src/relax/ir/dataflow_matcher.cc
+++ b/src/relax/ir/dataflow_matcher.cc
@@ -541,7 +541,7 @@ struct RNode {
  * \brief This method try to match a real node and a pattern node along with its neighbors.
  */
 static bool try_match(PNode* p, RNode* r, DFPatternMatcher* m,
-                      const std::map<const VarNode*, std::set<const VarNode*>>& def2use,
+                      const std::map<const VarNode*, std::vector<const VarNode*>>& def2use,
                       const std::map<const VarNode*, std::vector<const VarNode*>>& use2def) {
   if (nullptr != p->matched && p->matched == r->ptr) return true;  // matched before.
   if (!m->Match(GetRef<DFPattern>(p->ptr), GetRef<Var>(r->ptr))) return false;
@@ -550,6 +550,12 @@ static bool try_match(PNode* p, RNode* r, DFPatternMatcher* m,
 
   const auto commit = [&undo_stack](PNode* p, RNode* r) {
     // match with each other.
+    // TODO(ganler, masahi): Why commit on the same p-r pair happens more than once?
+    if (p->ptr == r->matched) {
+      ICHECK_EQ(p->matched, r->ptr);
+      return;
+    }
+    ICHECK_EQ(r->matched, nullptr);
     p->matched = r->ptr;
     r->matched = p->ptr;
     undo_stack.emplace(p, r);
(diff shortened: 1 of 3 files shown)


tests/python/relax/test_dataflow_pattern.py
```diff
@@ -1006,5 +1006,50 @@ def rewriter(_, matchings):
...
+        inp_pat = wildcard()
+        Q_weight_pat = wildcard()
+        K_weight_pat = wildcard()
+        V_weight_pat = wildcard()
+
+        matmul1 = is_op("relax.matmul")(inp_pat, Q_weight_pat)
+        matmul2 = is_op("relax.matmul")(inp_pat, K_weight_pat)
+        matmul3 = is_op("relax.matmul")(inp_pat, V_weight_pat)
+
+        # TODO(masahi): Automate addition of used_by constraints during is_op
```
masahi, line 1037: I have a follow-up PR to address this, which removes all used_by stuff below @ganler

masahi:
@ganler Is `start_hint` stuff necessary? I want to remove it, since matching is now deterministic. Some tests depend on the use of `start_hint`.

ganler:
@masahi Thanks. Start hint means the start point of graph matching. It is useful when you have multiple matchable patterns and want to explicitly match one of them which are close to / subsuming certain node.

ganler:
(Traveling) if you can wait I can file a review after 10 hours but definitely within 24 hours. :). Hope it won't get you blocked.

sunggg (commented):
Thank you for the improvement! 
Removing non-determinism should be a right way to go. 
Overall, it looks good to me. We can merge this when @ganler can confirm with the implementation details.

src/relax/ir/dataflow_matcher.cc
```diff
@@ -541,15 +541,21 @@ struct RNode {
...
                       const std::map<const VarNode*, std::vector<const VarNode*>>& use2def) {
-  if (nullptr != p->matched && p->matched == r->ptr) return true;  // matched before.
+  if (p->matched != nullptr && p->matched == r->ptr) return true;  // matched before.
   if (!m->Match(GetRef<DFPattern>(p->ptr), GetRef<Var>(r->ptr))) return false;
 
   std::stack<std::pair<PNode*, RNode*>> undo_stack{};
 
   const auto commit = [&undo_stack](PNode* p, RNode* r) {
     // match with each other.
+    // TODO(ganler, masahi): Why commit on the same p-r pair happens more than once?
```
ganler, line 553: I guess in either way the eventual results are the same. But yeah doing a pre-check could be faster (avoid the overhead of `undo_stack.emplace(p, r)`).

ganler (approved):
The changes make sense to me. I don't see any (apparent) logic flaws. Thanks for optimizing the code style and modernizing the code to C++17 standard!

src/relax/ir/dataflow_matcher.cc
masahi: Here my concern is not about performance. If the same pair is committed more than once, I think there is something odd about the matching algorithm.

Later I'll try to improve the matching algorithm implementation. In particular, I want to remove `try_match` loop on parents, https://github.com/apache/tvm/blob/278bc9bbdf7d9b12c44bb78b71f91e4872b14353/src/relax/ir/dataflow_matcher.cc#L577-L607. This bidirectional matching has been a source of confusion to me - Since the constraint graph is assumed to be a DAG, I think we can start matching from the root nodes in a topo-sorted order, and matching should be able to proceed purely in one direction.

ganler: I see. Because the constraint or relation between nodes may or may not be single-way so in the beginning I made it bidirectional such that the pattern can be matched as long as you can let any node of the matched subgraph be start hint and get the pattern matched eventually. Specifying certain matching order definitely makes the code logic and debugging easier but I am afraid it also cuts the flexibility in some way. Not sure if it is worth to have some flexibility or if we can keep them both.

masahi: I'm also going to look at how similar projects like MLIR, OpenVINO etc implements general graph pattern matching.

ganler: Sorry, I need some time to illustrate but my flight is taking off now. I will get you back probably after 8 hours. 🥲

masahi: ok :) I'll share my thought in the meantime. 

Looking at an example in https://github.com/apache/tvm/blob/unity/tests/python/relax/test_dataflow_pattern.py#L540-L545, I cannot imagine how one would use `start_hint` in practice. Here, you wrote the input mod by hand, so you know `dfb.bindings[0].var` is associated to the "left" branch of the CBRx2. But in general, we don't have such information, especially in e2e scenarios for real world models. 

Even if there was a good use case for it, I claim that it doesn't justify making API and the implementation more complicated. If one has such advanced knowledge of model structure and variables, they may as well have a different way to match & extract such subgraph.

masahi:
Just hit an interesting bug in the graph matcher.

The following fake QKV projection mod and pattern should not match. Without this PR I get segfault. With this PR, it incorrectly matches with the result      
`{*: x2, *: w1, Op(relax.matmul)(*, *): lv1, *: w2, Op(relax.matmul)(*, *): lv2, Op(relax.matmul)(*, *): lv0, *: w0}`

```
@R.function
def main(...) -> R.Tensor:
    with R.dataflow():
        lv0 = R.matmul(x1, w0)
        lv1 = R.matmul(x2, w1)
        lv2 = R.matmul(x2, w2)
        ...

with PatternContext() as ctx:
    inp_pat = wildcard()
    ...

```

I'll work on a fix in a separate PR.

tqchen (approved):

src/relax/ir/dataflow_matcher.cc
ganler: Thanks for bringing it up. When designing the usages it was my bad that I did not know enough use cases and I made it "more capable" by compromising simplicity. masa you have more professional experience with how patterns look in practice so I strongly agree that we should cut off the implementation/complexity for such long-tail use cases.

ganler: BTW, also to explain why I was using the code style of `CONSTANT logic_op VAR`. This is just a practice to avoid writing `VAR = CONSTANT` which causes silent logic errors. But yeah it is nice to normalize the code style.

PR #14418, opened 2023-03-28
[Unity][Fix] Copy over module attrs in FuseTIR

psrivas2:
This fixes a bug in FuseTIR where module attributes were dropped in the transformation.

cc: @jwfromm @sunggg


jwfromm (approved):
Awesome, thanks for the fix!

sunggg (approved):

PR #14420, opened 2023-03-28
[Unity][Fix] Annotate TIR op pattern could have no stores.

psrivas2:
Fix a bug where a block could have no stores, the pass crashes in such a case. We should mark the PrimFunc as opaque in such a case.
This happens if instead of using buffer load/stores we use `tir.call_packed` for computation.

cc @jwfromm @sunggg


sunggg (commented):
Thanks for the fix! It would be nice to add more context about when we face such cases in the comment.

psrivas2:
> Thanks for the fix! It would be nice to add more context about when we face such cases in the comment.

added a test and some context.

sunggg (approved):
@psrivas2 Thank you!

PR #14376, opened 2023-03-23
[Hexagon] Allow scalar tensors to have null shape during allocation

janetsc:
Update assert during buffer allocation.  Zero dimensional tensors (scalars) can have a null shape.

cc: @csullivan, @farshidsp

PR #14427, opened 2023-03-29
[Unity][Hexagon] Allow scalar tensors to have null shape during allocation

ibsidorenko:
Update assert during buffer allocation. Zero dimensional tensors (scalars) can have a null shape.

This PR is attempt to port [PR#14376](https://github.com/apache/tvm/pull/14376) from tvm/main that enables Hexagon tests with Relax.


janetsc (approved):
Thanks for porting this fix!

janetsc (approved):

farshidsp:
Thanks for this PR. LGTM.

masahi:
Can we wait for the next rebase instead of making the same change?

ibsidorenko:
> Can we wait for the next rebase instead of making the same change?
Sure, we can wait If rebase is planned in the near future.

tqchen (approved):

tqchen:
yes, we plan to do it sometime this weekend

tqchen:
should be in now

PR #14431, opened 2023-03-29
[Unity] Minor updates to DataFlowBlockRewrite

masahi:
* `relax/binding_rewrite.py` is not imported by default, so we get errors like `AttributeError: <class 'tvm.runtime.object.Object'> has no attribute add` if we try to pass the C++ object to python

* `DataflowBlock` member variable shouldn't be optional, but it is declared with `Optional<DataflowBlock> dfb_` and accessed via `dfb_.value()` all over the places.

* `root_fn_` **should** be optional (e.g. when `DataFlowBlockRewrite` is used inside another `ExprMutator`) , but it's always updated via `root_fn_.value()` without checking   

* Other minor style updates that hopefully improve readability.

@ganler @tqchen

### DIFF
--- a/include/tvm/relax/binding_rewrite.h
+++ b/include/tvm/relax/binding_rewrite.h
@@ -60,7 +60,7 @@ class DataflowBlockRewriteNode : public Object {
   void RemoveAllUnused();
 
   /*! \brief The rewritten dataflow block. */
-  DataflowBlock MutatedDataflowBlock() { return dfb_.value(); }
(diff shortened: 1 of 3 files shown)


masahi:
I just realized that I didn't need to use `DataflowBlockRewrite` for my purpose. But keeping this PR for now.

psrivas2 (approved):

tqchen (approved):

PR #14439, opened 2023-03-30
[Unity][Graph matching] Automatically add `used-by` constraints for `is_op` pattern

masahi:
Compared to the typical, structural matching (e.g. matching `is_op` based pattern against `CallNode`), graph pattern matching requires additionally specifying `used_by` constraints. This leads to an obvious redundancy in the usage:

```

inp_pat.used_by(matmul1, 0)
inp_pat.used_by(matmul2, 0)
inp_pat.used_by(matmul3, 0)

Q_weight_pat.only_used_by(matmul1, 1)
K_weight_pat.only_used_by(matmul2, 1)
V_weight_pat.only_used_by(matmul3, 1)
``` 

This PR hacks `is_op` constructor by automatically adding such `used_by` constraints between caller and callee patterns. See the simplified test case for QOL improvement.

@ganler @sunggg

### DIFF
--- a/include/tvm/relax/dataflow_matcher.h
+++ b/include/tvm/relax/dataflow_matcher.h
@@ -65,15 +65,6 @@ TVM_DLL tvm::runtime::Map<DFPattern, Var> MatchGraph(const PatternContext& ctx,
                                                      Optional<Var> start_hint = NullOpt,
                                                      bool must_include_hint = false);
 
-/**
- * \brief Match a graph-wise pattern with the current context (PatternContext::Current()).
- */
-inline tvm::runtime::Map<DFPattern, Var> MatchGraphDefault(const DataflowBlock& dfb,
-                                                           Optional<Var> start_hint = NullOpt,
-                                                           bool must_include_hint = false) {
-  return MatchGraph(PatternContext::Current(), dfb, start_hint, must_include_hint);
-}
-
 }  // namespace relax
 }  // namespace tvm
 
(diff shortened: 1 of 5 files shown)


include/tvm/relax/dataflow_matcher.h
```diff
@@ -65,15 +65,6 @@ TVM_DLL tvm::runtime::Map<DFPattern, Var> MatchGraph(const PatternContext& ctx,
                                                      Optional<Var> start_hint = NullOpt,
                                                      bool must_include_hint = false);
 
-/**
- * \brief Match a graph-wise pattern with the current context (PatternContext::Current()).
- */
-inline tvm::runtime::Map<DFPattern, Var> MatchGraphDefault(const DataflowBlock& dfb,
-                                                           Optional<Var> start_hint = NullOpt,
-                                                           bool must_include_hint = false) {
-  return MatchGraph(PatternContext::Current(), dfb, start_hint, must_include_hint);
```
masahi, line 74: This gets removed since it is not used anywhere but it touches `PatternContext::Current()` (which is now `Optional`)

tests/python/relax/test_dataflow_pattern.py
```diff
@@ -1034,15 +1034,6 @@ def main(
...
         matmul3 = is_op("relax.matmul")(inp_pat, V_weight_pat)
 
-        # TODO(masahi): Automate addition of used_by constraints during is_op
-        inp_pat.used_by(matmul1, 0)
-        inp_pat.used_by(matmul2, 0)
-        inp_pat.used_by(matmul3, 0)
-
-        Q_weight_pat.only_used_by(matmul1, 1)
-        K_weight_pat.only_used_by(matmul2, 1)
-        V_weight_pat.only_used_by(matmul3, 1)
```
ganler, line 1044: I understand `used_by` can be removed now as it is implied by a function call. However, `only_used_by` is a stronger constraint that cannot be implied via a fn call right?

ganler: But yeah I guess you meant oftentimes we don't fold Q, K, and V so it is fine if they are being used by others at the same time.

PR #14440, opened 2023-03-30
[Unity][Graph matching] Clean up undo stack for parent and child nodes properly

masahi:
This is a fix for the issue described in https://github.com/apache/tvm/pull/14417#issuecomment-1490049917

In graph pattern matching, backtracking behavior is implemented by undoing matchings in `undo_stack`. When an intermediate fails to match, `undo_stack`s for parent and child nodes that have matched tentatively also need to be cleaned up as well. This corner-case handling is missing in the current implementation. 

The fix is to return `undo_stack` from recursive `try_match` calls when matching succeeds, and merge parent / child `undo_stack` with the `undo_stack` for the "current" node

@ganler @sunggg @cyx-6

### DIFF
--- a/src/relax/analysis/udchain.cc
+++ b/src/relax/analysis/udchain.cc
@@ -75,7 +75,7 @@ std::pair<runtime::Map<Var, runtime::Array<Var>>, runtime::Array<Var>> FunctionU
     uses.reserve(users.size());
     for (const auto& v : users) {
       if (v == nullptr &&
-          std::find(fn_outs.begin(), fn_outs.end(), GetRef<Var>(var)) == fn_outs.end() ) {
+          std::find(fn_outs.begin(), fn_outs.end(), GetRef<Var>(var)) == fn_outs.end()) {
         fn_outs.push_back(GetRef<Var>(var));
       } else {
         uses.push_back(GetRef<Var>(v));
--- a/src/relax/analysis/udchain.cc
+++ b/src/relax/analysis/udchain.cc
@@ -65,36 +65,36 @@ class UDChain : public relax::ExprVisitor {
 std::pair<runtime::Map<Var, runtime::Array<Var>>, runtime::Array<Var>> FunctionUseDef(
     const Function& fn) {
   UDChain udchain;
-  udchain.VisitExpr_(fn.get());
+  udchain.VisitExpr(fn);
 
   Map<Var, Array<Var>> user_map;
   Array<Var> fn_outs;
 
-  for (const auto& kv : udchain.to_users) {
+  for (const auto& [var, users] : udchain.to_users) {
     Array<Var> uses{};
-    uses.reserve(kv.second.size());
-    for (const auto& v : kv.second) {
-      if (nullptr == v &&
-          fn_outs.end() == std::find(fn_outs.begin(), fn_outs.end(), GetRef<Var>(kv.first))) {
-        fn_outs.push_back(GetRef<Var>(kv.first));
+    uses.reserve(users.size());
+    for (const auto& v : users) {
+      if (v == nullptr &&
+          std::find(fn_outs.begin(), fn_outs.end(), GetRef<Var>(var)) == fn_outs.end() ) {
+        fn_outs.push_back(GetRef<Var>(var));
       } else {
         uses.push_back(GetRef<Var>(v));
       }
     }
-    user_map.Set(GetRef<Var>(kv.first), std::move(uses));
+    user_map.Set(GetRef<Var>(var), std::move(uses));
   }
   return std::make_pair(std::move(user_map), std::move(fn_outs));
 }
 
 runtime::Map<Var, Array<Var>> DataflowBlockUseDef(const DataflowBlock& dfb) {
   UDChain udchain;
-  udchain.VisitBindingBlock_(dfb.get());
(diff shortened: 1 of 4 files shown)


src/relax/ir/dataflow_matcher.cc
```diff
@@ -540,33 +541,48 @@ struct RNode {
...
-  const auto commit = [&undo_stack](PNode* p, RNode* r) {
+  const auto commit = [&undo](PNode* p, RNode* r) {
     // match with each other.
     // TODO(ganler, masahi): Why commit on the same p-r pair happens more than once?
     if (p->ptr == r->matched) {
       ICHECK_EQ(p->matched, r->ptr);
       return;
     }
-    ICHECK(r->matched == nullptr);
+    // TODO(ganler, masahi): Why this condition can fail?
```
ganler, line 561: In practice they won't, it's like a debugging check that I did not remove. Thanks for optimizing it.

src/relax/ir/dataflow_matcher.cc
```diff
@@ -540,33 +541,48 @@ struct RNode {
 /**
  * \brief This method try to match a real node and a pattern node along with its neighbors.
  */
-static bool try_match(PNode* p, RNode* r, DFPatternMatcher* m,
-                      const std::map<const VarNode*, std::vector<const VarNode*>>& def2use,
-                      const std::map<const VarNode*, std::vector<const VarNode*>>& use2def) {
-  if (p->matched != nullptr && p->matched == r->ptr) return true;  // matched before.
-  if (!m->Match(GetRef<DFPattern>(p->ptr), GetRef<Var>(r->ptr))) return false;
+using UndoStack = std::stack<std::pair<PNode*, RNode*>>;
```
ganler, line 544: Consider `std::stack<std::pair<PNode*, RNode*>, std::vector<std::pair<PNode*, RNode*>>>` for cache-friendliness.

src/relax/analysis/udchain.cc
```diff
@@ -65,36 +65,36 @@ class UDChain : public relax::ExprVisitor {
 std::pair<runtime::Map<Var, runtime::Array<Var>>, runtime::Array<Var>> FunctionUseDef(
     const Function& fn) {
```
masahi, line 66: The change in this file is unrelated, but I couldn't help updating this file after noticing clean-up opportunities.

src/relax/ir/dataflow_matcher.cc
ganler: If we have perf consideration, we can do `std::vector` and do `std::copy` in L580.

ganler (approved):
LGTM.

src/relax/analysis/udchain.cc
```diff
@@ -65,36 +65,36 @@ class UDChain : public relax::ExprVisitor {
...
     Array<Var> uses{};
-    uses.reserve(kv.second.size());
-    for (const auto& v : kv.second) {
-      if (nullptr == v &&
-          fn_outs.end() == std::find(fn_outs.begin(), fn_outs.end(), GetRef<Var>(kv.first))) {
-        fn_outs.push_back(GetRef<Var>(kv.first));
+    uses.reserve(users.size());
+    for (const auto& v : users) {
+      if (v == nullptr &&
+          std::find(fn_outs.begin(), fn_outs.end(), GetRef<Var>(var)) == fn_outs.end() ) {
```
ganler, line 78:
```suggestion
          std::find(fn_outs.begin(), fn_outs.end(), GetRef<Var>(var)) == fn_outs.end()) {
```

src/relax/ir/dataflow_matcher.cc
masahi: no I added this check in the previous pr. But the new test fails at this check.

ganler: Oh I was talking about `ICHECK_EQ(p->matched, r->ptr);` sorry for the confusion.

src/relax/ir/dataflow_matcher.cc
masahi: ok it seems it doesn't have to be a stack. I'll update.

src/relax/ir/dataflow_matcher.cc
masahi: sorry I checked again, it seems after the undo stack fix, I don't get an error from this check. I'll revert this change.

masahi:
Hey I'm happy to report that I have pattern-based rewriter for dataflow block working. Compared to the existing call node based matching & rewriting that requires a common post-dominator in a pattern, it lets us match a tree structure and replace leaf nodes or branches with new expression. 

This can be immediately used for combining *any* number of multiple matmuls sharing the same LHS into one matmul. In Relay, we have `CombineParallelDense` pass for that purpose, but we can achieve the same thing via graph (tree) matching and rewrite.   

For example, in SD UNet we have many three parallel matmul for QKV projections. In addition, there are also highly non-obvious parallel matmuls consisting of 32 or 22 of them. Those patterns can all be matched and rewritten via the following generic pattern and rewriter. I'm very excited about this! 
https://github.com/masahi/web-stable-diffusion/blob/unet-opt/test.py#L37-L68

psrivas2 (approved):
Fantastic, thanks!

Issue #14437, opened 2023-03-30
[Bug] TVMError: CodeGenVM cannot handle this intrinsic now: Op(relax.nn.conv2d)


zhaoyang-star:
I try to import ResNet18 PyTorch model to Relax. The error occurs when executing `relax.build`  and the error is:
```
File "/WORK/Dev/tvm/src/relax/backend/vm/codegen_vm.cc", line 169
TVMError: CodeGenVM cannot handle this intrinsic now:
Op(relax.nn.conv2d)
```
It hints `relax.nn.conv2d` cannot handle. But the unittest [test_op_nn_convolution.py](https://github.com/apache/tvm/blob/unity/tests/python/relax/test_op_nn_convolution.py#L32) is passed. So I am puzzled.

Test code: 
```python
import numpy as np
import torch.fx as fx
import torchvision.models as models
from tvm.relax.frontend.torch import from_fx
import tvm
from tvm import relax, tir

if __name__ == "__main__":
    torch_model = models.resnet18()

    # Use FX tracer to trace the PyTorch model.
    graph_module = fx.symbolic_trace(torch_model)

    # symbolic dimensions
    bs = tir.Var("bs", "int64")
    input_info = [((bs, 3, 224, 224), "float32")]

    # Use the importer to import the PyTorch model to Relax.
    mod = from_fx(graph_module, input_info)

    # Print out the imported model.
    mod.show()

    # build and create vm executor
    target = tvm.target.Target("llvm", host="llvm")
    ex = relax.build(mod, target)
    vm = relax.VirtualMachine(ex, tvm.cpu())

    # init parameters
    params = tvm.relax.testing.nn.init_params(mod)

    # run the mlp model on relax vm
    data = tvm.nd.array(np.random.rand(1, 3, 224, 224).astype(np.float32))
    res = vm["main"](data, *params)
    print(res)
```

### Environment
branch: unity
commit 854f2e9bc12fb8bc938db0d65f70835e129a48f9
Date:   Tue Mar 28 19:28:32 2023 +0300
OS: Linux

masahi:
You need to run `Legalize()` pass before build.

yongwww:
@zhaoyang-star  thanks for applied the LegalizeOps pass and report the follow-up error. We are discussing to add this pass into vm build these days.

Symbolic shape is supported in Relax. As for the follow-up error you reported, it is from the topi pool2d, the output shape for pool2d was converted to int32 there, I will send the fix  by end of week.

[...]

PR #14443, opened 2023-03-31
[Unity] Remove cast for output shape of pool2d

yongwww:
The patch to fix issue reported in #14437

The symbolic dim "m" becomes T.Cast(m, int32) in topi pool2d. It is expected to be int64 in ShapeExpr.

cc: @tqchen @slyubomirsky @psrivas2 

---
update: remove the cast to i32 for pool2d in topi


tqchen:
We should perhaps fix topi instead so that the symbolic dimensions remain same as input, in this case i64

tqchen:
Would be good to send topi change to main

yongwww:
Closed this. sent #14549 into main

PR #14446, opened 2023-03-31
[Unity] Pattern-based rewriting for dataflow block

masahi:
Compared to the existing call node based matching & rewriting that requires a common post-dominator in a pattern (introduced in https://github.com/apache/tvm/pull/14312), it lets us match a tree structure and replace leaf nodes or branches with new expression.

This can be immediately used for combining any number of multiple matmuls sharing the same LHS into one matmul. In Relay, we have a dedicated pass for that purpose (`CombineParallelDense`), but we can achieve the same thing via graph (tree) matching and rewrite. 

For example, in SD UNet we have many three parallel matmul for QKV projections. In addition, there are also highly non-obvious parallel matmuls consisting of 32 or 22 of them. Those patterns can all be matched and rewritten via the following generic pattern and rewriter. 
https://github.com/masahi/web-stable-diffusion/blob/unet-opt/test.py#L37-L68

I got all matmul combining for SD UNet working. It reduces the number of `R.matmul` from 200 to 116.
Original: https://gist.github.com/masahi/0dab4b8f53115da9c33f4352c9175a87
Rewritten: https://gist.github.com/masahi/57ab925a43d2343cbfdb79a31c5b9946 (look for `R.concat` followed by `R.matmul` to see where rewriting happened)

This is exactly the same reduction possible using Relay passes on an equivalent Relay mod.

@ganler @sunggg @psrivas2 @cyx-6 @vinx13

### DIFF
--- a/python/tvm/relax/dpl/__init__.py
+++ b/python/tvm/relax/dpl/__init__.py
@@ -19,3 +19,4 @@
 
 from .pattern import *
 from .context import *
+from .rewrite import rewrite_call, rewrite_bindings
--- a/python/tvm/relax/dpl/rewrite.py
+++ b/python/tvm/relax/dpl/rewrite.py
@@ -1,0 +1,115 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+"""APIs for pattern-based rewriting."""
+from typing import Dict, Callable
+from .pattern import DFPattern
+from .context import PatternContext
+
+from ..expr import Expr, Function, Var
+from . import _ffi as ffi
+
+
+def rewrite_call(
+    pattern: DFPattern, rewriter: Callable[[Expr, Dict[DFPattern, Expr]], Expr], func: Function
+) -> Function:
+    """
+    Rewrite a function with the given pattern and the rewriter function.
+
+    Parameters
+    ----------
+    pattern: DFPattern
+        The pattern to match.
+
+    rewriter: Callable[[Expr, Dict[DFPattern, Expr]], Expr]
+        The function to be called on a successful matching for rewriting. Given the matched
+        call node and the map of patterns and matched expressions, it should return a new call node
+        to replace the original one or the original matched call node as is.
+
+        For example, to replace x + x with 2 * x, we can write the rewriter as follows:
+        ```
+        x = wildcard()
+        pattern = is_op("relax.add")(x, x)
+
+        def rewriter(orig, matchings):
+            return R.multiply(matchings[x], R.const(2, "float32"))
+        ```
+
+    func: Function
+        The function to rewrite.
+
+    Returns
+    -------
+    rewritten_func: Function
+        The rewritten or the input function, depending on the pattern matching result.
+    """
+    return ffi.rewrite_call(pattern, rewriter, func)
+
+
+def rewrite_bindings(
+    ctx: PatternContext, rewriter: Callable[[Dict[DFPattern, Var]], Dict[Var, Expr]], func: Function
+) -> Function:
+    """
+    Rewrite a function with the given pattern and the rewriter function.
+
+    Parameters
+    ----------
+    ctx: PatternContext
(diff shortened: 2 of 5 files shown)


python/tvm/relax/dpl/__init__.py
```diff
@@ -19,3 +19,4 @@
 
 from .pattern import *
 from .context import *
+from .rewrite import rewrite_call, rewrite_bindings
```
masahi, line 22: The existing `rewrite` function is renamed to `rewrite_call` to make it clear that it is CallNode rewriting. And together with the new dataflow block rewriting function, it is put under the new file.

python/tvm/relax/dpl/rewrite.py
```diff
@@ -0,0 +1,115 @@
...
+            matmul = R.matmul(inp, concat)
+            Q = R.strided_slice(matmul, axes=[2], begin=[0], end=[width])
+            K = R.strided_slice(matmul, axes=[2], begin=[width], end=[width * 2])
+            V = R.strided_slice(matmul, axes=[2], begin=[width * 2], end=[width * 3])
+
+            # matchings[matmul1] gives the bound variable in the binding whose RHS matches with
+            # the matmul1 pattern. For example, lv0 in lv0 = R.matmul(x1, w0).
+            # We want to replace the RHS of this binding with Q.
+            return {matchings[matmul1]: Q, matchings[matmul2]: K, matchings[matmul3]: V}
+        ```
```
masahi, line 105: I hope the API for the rewriter makes sense and the usage is intuitive cc @ganler

src/relax/ir/dataflow_matcher.cc
```diff
@@ -780,18 +780,29 @@ Optional<Map<DFPattern, Var>> MatchGraph(const PatternContext& ctx, const Datafl
...
+      : pattern_(pat), rewriter_func_(rewriter_func), params_(params) {}
 
-  static Expr Run(DFPattern pat, PackedFunc rewriter_func, Function f) {
-    PatternRewriter rewriter(pat, rewriter_func);
+  PatternRewriter(const PatternContext& ctx, PackedFunc rewriter_func,
+                  const std::unordered_set<const VarNode*>& params)
+      : ctx_(ctx), rewriter_func_(rewriter_func), params_(params) {}
+
+  template <typename PatternType>
+  static Expr Run(PatternType pat, PackedFunc rewriter_func, Function f) {
```
masahi, line 800: Depending on the passed pattern type (`DFPattern` or `PatternContext`), it does either call node rewriting or dataflow block rewriting. It never does both in a single pass (obvious from the constructors).

src/relax/ir/dataflow_matcher.cc
```diff
@@ -807,25 +818,112 @@ class PatternRewriter : ExprMutator {
...
+        emitted_vars->insert(var_bind->var.get());
+        if (--num_unemitted == 0) {
+          return;
+        }
+      }
+    }
+  }
+
+  // Repeat until all matchable subsets of bindings are rewritten.
+  BindingBlock RewriteDataflowBlockFixedPoint(BindingBlock block) {
```
masahi, line 870: We need to apply rewriting repeatedly since, for example, the same QKV projection pattern appears a number of times in a single DFB.

src/relax/ir/dataflow_matcher.cc
```diff
@@ -807,25 +818,112 @@ class PatternRewriter : ExprMutator {
...
+    }
+
+    size_t num_unemitted = unemitted_vars.size();
+    for (size_t i = 0; i < pending_bindings.size(); ++i) {
+      const auto& binding = pending_bindings[i];
+      if (auto var_bind = binding.as<VarBindingNode>();
+          var_bind && unemitted_vars.count(var_bind->var.get())) {
+        // var_bind->value may also depend on other unemitted vars in this range
+        Array<Binding> prev_bindings(pending_bindings.begin(), pending_bindings.begin() + i);
+        EmitUsedVars(var_bind->value, prev_bindings, emitted_vars);
```
masahi, line 859: I want to get rid of this recursive call and make sure we traverse `pending_bindings` only once. The issue is that `PostOrderVisit` does not look into subexpressions when it encounters the corresponding bound variable. For example, given the contrived input bindings below, 

```
with R.dataflow():
    w0_t = R.permute_dims(w0, axes=None)
    lv0 = R.matmul(x1, w0_t)
    w1_t = R.permute_dims(w1, axes=None)
    w1_t_t = R.permute_dims(w1_t, axes=None)
    lv1 = R.matmul(x1, w1_t_t)
    w2_t = R.permute_dims(w2, axes=None)
    lv2 = R.matmul(x1, w2_t)
```   
we need to emit all `permute_dims` binding before emitting `concat` and the combined matmul, since `concat` depends on all weights some of which are defined after the first matmul. When `PostOrderVisit` is applied on `R.matmul(x1, w1_t_t)`, `w1_t` is not visited. So even though we need to emit `w1_t` before `w1_t_t`, `w1_t` is not put into the initial `unemitted_vars` set.

I think we can use `AnalyzeVar2Value` on the input function to get bindings, and recursively traverse the bound expression when we encounter a new unemitted var. But I find that a bit complicated for a simple job like this, so I'm looking for a simpler solution. For now I'm keeping this recursive solution that is not efficient but extremely simple.

vinx13 (approved):

PR #14447, opened 2023-03-31
[Unity][Op] introduce `shape_to_tensor` op

sunggg:
In Unity, we have a clear distinction between `tensor` and `shape`: we have `ShapeExpr` and `ShapeStructInfo` in AST, `ShapeTuple` in runtime container. Meanwhile, most of operators and their TOPI implementations are defined with tensors. For example, `relax.take` is defined as follows:

```
// Operator definition
TVM_REGISTER_OP("relax.take")
    .set_attrs_type<TakeAttrs>()
    .set_num_inputs(2)
    .add_argument("x", "Tensor", "The source tensor.")
    .add_argument("indices", "Tensor", "The indices of the values to extract.")
    .set_attr<FInferStructInfo>("FInferStructInfo", InferStructInfoTake);
       
StructInfo InferStructInfoTake(const Call& call, const BlockBuilder& ctx) {
  Array<TensorStructInfo> input_sinfo = GetInputTensorStructInfo(call, ctx);
   // Assume inputs are `Tensor` 
  TensorStructInfo data_sinfo = input_sinfo[0];
  TensorStructInfo indices_sinfo = input_sinfo[1];
  ...
}

// TOPI
inline Tensor take(const Tensor& a, const Tensor& indices, int batch_dims, int axis, ...)
```

To allow the shape computation, this PR introduces a `shape_to_tensor` op that converts `ShapeTuple` to `NDArray` at runtime. This enables the common shape computation patterns like following:
```
shape_var: R.Shape(ndim=4)
lv: R.Tensor((4,), dtype="int64") = R.shape_to_tensor(shape_var)
lv1: R.Tensor((1,), dtype="int64") = R.take(lv, indices, axis=0)
lv2: R.Tensor((1, 1), dtype="int64") = R.expand_dims(lv1, axis=[0])
gv: R.Tensor((1, 1), dtype="int64") = R.concat((lv2,), axis=0)
```
It's worth noting that `tensor_to_shape` op is already introduced in https://github.com/apache/tvm/pull/14282, so roundtrip between shape and tensor would be now possible. 

Currently, this op requires special handling in `FoldConstant` pass since this pass is only able to evaluate `TIR primfunc`, not `PackedFunc`. Once we extend `FoldConstant` to support `PackedFunc` evaluation, we should be able to remove these unnecessary special handling. 

cc. @jwfromm @yongwww @psrivas2 @slyubomirsky @tqchen


src/relax/op/op_common.cc
```diff
@@ -38,7 +38,7 @@ Array<TensorStructInfo> GetInputTensorStructInfo(const Call& call, const BlockBu
     if (sinfo == nullptr) {
       ctx->ReportFatal(Diagnostic::Error(call)
                        << op << " requires the input " << op->arguments[i]->name
-                       << " to be Tensor. However, the given one is "
+                       << " to be Tensor. However, the given one has a "
```
sunggg, line 41: This is unrelated to this PR, but found the current message confusing during the debugging.

python/tvm/relax/op/base.py
```diff
@@ -253,6 +254,19 @@ def render_object(val: tvm.Object) -> str:
...
+def relax_shape_to_tensor(shape_tuple: tvm.runtime.ShapeTuple) -> tvm.nd.NDArray:
+    """
+    Takes a ShapeTuple and convert it to NDArray.
+
+    Parameters
+    ----------
+    shape_tuple: tvm.runtime.ShapeTuple
+        Shape tuple that we want to convert to NDArray at runtime
+    """
+    return tvm.nd.array([int(v) for v in shape_tuple])
```
vinx13, line 267: do we assume it's always on cpu?

src/relax/transform/fold_constant.cc
```diff
@@ -279,6 +279,25 @@ class ConstantFolder : public ExprMutator {
...
           return ShapeExpr(shape_values);
         }
+      } else if (op->name == "relax.shape_to_tensor") {
+        // Special handling for "relax.shape_to_tensor" since it is implemented in PackedFunc.
+        // TODO(sunggg): revisit this when we extend ConstantFolding to fold PackedFunc.
+        Expr arg = post_call->args[0];
+        ShapeExpr shape = Downcast<ShapeExpr>(arg);
+        Array<PrimExpr> values = shape->values;
+        Array<Integer> arr;
+        bool isKnown = true;
```
vinx13, line 289: nit
```suggestion
        bool is_known = true;
```

python/tvm/relax/op/base.py
sunggg: Yes. shape tuple and shape computation happen on the CPU side.

yongwww (approved):
LGTM!

vinx13 (approved):

Issue #14364, opened 2023-03-22
[Bug] Error in tir.Schedule with dynamic shape


Ewenwan:
```py
@T.prim_func
def before_inline(a: T.handle, c: T.handle, m:T.int32, n:T.int32) -> None:
    A = T.match_buffer(a, (m, n))
    B = T.alloc_buffer((m, n))
    C = T.match_buffer(c, (m, n))
    for i, j in T.grid(m, n):
        with T.block("B"):
            vi, vj = T.axis.remap("SS", [i, j])
            B[vi, vj] = A[vi, vj] * 2.0
    for i, j in T.grid(m, n):
        with T.block("C"):
            vi, vj = T.axis.remap("SS", [i, j])
            C[vi, vj] = B[vi, vj] + 1.0
sch = tvm.tir.Schedule(before_inline)
sch.compute_inline(sch.get_block("B"))
print(sch.mod["main"].script())
```
# Error message: 
```
ScheduleError: An error occurred in the schedule primitive 'compute-inline'.
The IR with diagnostic is:
# from tvm.script import tir as T
@tvm.script.ir_module
class Module:
    @T.prim_func
    def main(a: T.handle, c: T.handle, m: T.int32, n: T.int32) -> None:
        A = T.match_buffer(a, [m, n], dtype="float32")
        C = T.match_buffer(c, [m, n], dtype="float32")
        # body
        # with T.block("root")
        B = T.alloc_buffer([m, n], dtype="float32")
        for i, j in T.grid(m, n):
            # tir.Block#0
            with T.block("B"):
            ^^^^^^^^^^^^^^^^^^
                vi, vj = T.axis.remap("SS", [i, j])
                T.reads(A[vi, vj])
                T.writes(B[vi, vj])
                B[vi, vj] = A[vi, vj] * T.float32(2)
        for i, j in T.grid(m, n):
            with T.block("C"):
                vi, vj = T.axis.remap("SS", [i, j])
                T.reads(B[vi, vj])
                T.writes(C[vi, vj])
                C[vi, vj] = B[vi, vj] + T.float32(1)
    
Error message: The body of the inlined block should be in form of
    'A[i, j, k, ...] = f(i, j, k, ...)',
where the indices on the left are distinct atomic variables,
and there should be no variables other than the index variables
```

PR #14451, opened 2023-04-02
[TIR] Fix compute inline for dynamic loop and nontrivial store indices

wrongtest-intellif:
Hi, the change wants to fix https://github.com/apache/tvm/issues/14364 where our `compute-inline` validity check failed. 
- Do not depend on undefined var counting, since dynamic loop extents may also be counted. The fast path just check the store indices should be the same with the block iter vars, which would cover the most cases.
- Add fallback implementation for store indices which are bijective wrt block iter vars, which is already utilized in `reverse-compute-inline` implementation.
- The change do not check the inlined producer's region cover the consumer. Refer to unittest `test_compute_inline_out_of_bound_consumer`, the previous behavior is not changed in this MR.

### DIFF
--- a/tests/python/unittest/test_tir_schedule_compute_inline.py
+++ b/tests/python/unittest/test_tir_schedule_compute_inline.py
@@ -894,6 +894,98 @@ def test_compute_inline_multi_consumer(use_block_name):
     verify_trace_roundtrip(sch=sch, mod=elementwise_multi_producer_consumer)
 
 
+def test_compute_inline_layout_transformed_store():
+    @T.prim_func
+    def before(X: T.Buffer[(9, 96), "float32"]):
+        A = T.alloc_buffer([6, 9, 16], "float32")
+        B = T.alloc_buffer([3, 3, 96], "float32")
+        for i, j in T.grid(9, 96):
+            with T.block("producer"):
+                vi, vj = T.axis.remap("SS", [i, j])
+                A[vj // 16, vi, vj % 16] = X[vi, vj]
+        for i0, i1, j in T.grid(3, 3, 96):
+            with T.block("consumer"):
+                vi0, vi1, vj = T.axis.remap("SSS", [i0, i1, j])
+                B[vi0, vi1, vj] = A[vj % 16, vi0 * 3 + vi1, vj // 16] + 1.0
+
+    @T.prim_func
+    def after(X: T.Buffer[(9, 96), "float32"]):
+        B = T.alloc_buffer([3, 3, 96], "float32")
+        for i0, i1, j in T.grid(3, 3, 96):
+            with T.block("consumer"):
+                vi0, vi1, vj = T.axis.remap("SSS", [i0, i1, j])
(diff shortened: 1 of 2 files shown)


yzh119:
Hi @wrongtest-intellif , this PR looks good to me, we can still use `GetNumUndefinedNonpointerVars` to check if we can avoid visiting buffer shape/strides in the `UndefinedVars` function (I have created another PR #https://github.com/apache/tvm/pull/14502 for that), what do you think?

wrongtest-intellif:
> Hi @wrongtest-intellif , this PR looks good to me, we can still use `GetNumUndefinedNonpointerVars` to check if we can avoid visiting buffer shape/strides in the `UndefinedVars` function (I have created another PR ##14502 for that), what do you think?

Thank you for the remind! It seems great to rely on #14502 to solve the dynamic buffer issue and then I could go on for the 
patch when the write indices are not in the ordered form.

tests/python/unittest/test_tir_schedule_compute_inline.py
```diff
@@ -894,6 +894,98 @@ def test_compute_inline_multi_consumer(use_block_name):
...
+            with T.block("C"):
+                vi, vj = T.axis.remap("SS", [i, j])
+                C[vi, vj] = A[vi, vj] * 2.0 + 1.0
+
+    sch = tir.Schedule(before, debug_mask="all")
+    sch.compute_inline(sch.get_block("B"))
+    tvm.ir.assert_structural_equal(after, sch.mod["main"])
+    verify_trace_roundtrip(sch=sch, mod=before)
+
+
```
yzh119, line 988: Adding another "nested block" case with variable loop extents in outer blocks is better.

yzh119 (commented):
Hi @wrongtest-intellif , after I tried more dynamic shape cases I feel it unnecessary to keep the `GetNumUndefinedNonpointerVars` check, I'll close my PR and let's move on with your changes.

tests/python/unittest/test_tir_schedule_compute_inline.py
yzh119: For example:

```python
@T.prim_func
def symbolic_shape_blockized(a: T.handle, b: T.handle):
    N = T.int32()
    A = T.match_buffer(a, (N, 32))
    B = T.match_buffer(b, (N, 32))
    for i in range(N):
        with T.block("outer"):
            vi = T.axis.spatial(N, i)
            temp = T.alloc_buffer((32), "float32", scope="local")
            for j in range(32):
                with T.block("T"):
                    vj = T.axis.spatial(32, j)
                    temp[vj] = A[vi, vj] + 1
            for j in range(32):
                with T.block("B"):
                    vj = T.axis.spatial(32, j)
                    B[vi, vj] = temp[vj] * A[vi, vj]
```

We should support computing inline the block `T`.

yzh119:
Hi @wrongtest-intellif , any updates on this?

wrongtest-intellif:
> Hi @wrongtest-intellif , any updates on this?

Sorry to be late, I would turn back to it soon.

PR #14456, opened 2023-04-03
[Target] Add Jetson AGX Xavier tag

mshr-h:
cc @junrushao @zxybazh 

I have the dev board and the change works fine with it.


mshr-h:
Ah, it's already defined. I had missed it.

mshr-h:
But why is AGX Xavier not using `TVM_REGISTER_CUDA_TAG` macros? Even though Jetson Nano is using it.

zxybazh:
I think the only difference here is definition of target host, the current `TVM_REGISTER_CUDA_TAG` macro doesn't involve target host definition. Please feel free to send a PR to add host definition to the macro if you wish to unify the target tag definitions.

PR #14457, opened 2023-04-03
[Unity][Op] Dynamic Gather

sunggg:
This PR introduces `relax.gather` op and extends its existing TOPI implementation. 
Current `topi.gather` has a strong restriction on the dimension of data to be equal to the dimension of indices. 
Thus, we relax this constraint to serve cases like [onnx gather](https://github.com/onnx/onnx/blob/main/docs/Changelog.md#gather-13). 

cc. @jwfromm @yongwww


yongwww:
it would be good to add some tests in `tests/python/topi/python/test_topi_transform.py` as well.

sunggg:
It turned out the semantic of `onnx.gather` is different with current `topi.gather` that implements `pytorch.gather`. 
Semantic of `onnx.gather` is closer to the special case of `topi.take` that implements `numpy.take` (with explicit axis and "raise" mode). Therefore, instead of introducing a separate op, I think it would be better to re-use the existing operator. So, I decided to close this PR and take the different approach: https://github.com/apache/tvm/pull/14481

sunggg:
It turned out the semantic of onnx.gather is different with current topi.gather that implements pytorch.gather.
Semantic of onnx.gather is closer to the special case of topi.take that implements numpy.take (with explicit axis and "raise" mode). Therefore, instead of introducing a separate op, I think it would be better to re-use the existing operator. So, I decided to close this PR and take the different approach: https://github.com/apache/tvm/pull/14481

PR #14458, opened 2023-04-03
[CI] Fix broken model link

lhutton1:
The path to the `cnn_s_quantized.tflite` model used in some TVMC tests was changed in https://github.com/ARM-software/ML-zoo/pull/56. The link has been updated with the new path.

### DIFF
--- a/tests/python/driver/tvmc/conftest.py
+++ b/tests/python/driver/tvmc/conftest.py
@@ -242,7 +242,7 @@ def tflite_mobilenet_v1_0_25_128(tmpdir_factory):
 
 @pytest.fixture(scope="session")
 def tflite_cnn_s_quantized(tmpdir_factory):
-    base_url = "https://github.com/ARM-software/ML-zoo/raw/master/models/keyword_spotting/cnn_small/tflite_int8/"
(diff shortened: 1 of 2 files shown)


tests/python/driver/tvmc/conftest.py
```diff
@@ -242,7 +242,7 @@ def tflite_mobilenet_v1_0_25_128(tmpdir_factory):
 
 @pytest.fixture(scope="session")
 def tflite_cnn_s_quantized(tmpdir_factory):
-    base_url = "https://github.com/ARM-software/ML-zoo/raw/master/models/keyword_spotting/cnn_small/tflite_int8/"
+    base_url = "https://github.com/ARM-software/ML-zoo/raw/master/models/keyword_spotting/cnn_small/model_package_tf/model_archive/TFLite/tflite_int8/cnn_s_quantized.tflite"
     file_to_download = "cnn_s_quantized.tflite"
     model_file = download_testdata(
         "{}/{}".format(base_url, file_to_download), file_to_download, module=["tvmc"]
```
echuraev, line 248: Not sure that it will work. After `"{}/{}".format(base_url, file_to_download)` we will get `https://.../tflite_int8/cnn_s_quantized.tflite/cnn_s_quantized.tflite`. So the url will be incorrect.

lhutton1: Oops yes you're correct, thanks!

echuraev (approved):
LGTM. Thanks!

PR #14463, opened 2023-04-03
[Unity][OP] Add `arange` op

Hzfengsy:
The current `arange` op in PyTorch is converted as constant at frontend, however, it does not work when the length is dynamic.

This PR adds `arange` op to relax, and use different rules when legalization:

- If the length is fixed, we will convert it to constant.
- If the length is dynamic, we use `topi.arange` to compute it.


yongwww (approved):
LGTM

PR #14036, opened 2023-02-18
[AutoScheduler][AutoTVM] Enable xgboost >= 1.7.x new changes

cbalint13:
This PR enables **tvm** to work with newer xgboost >= 1.7.x

- xgboost trivial changes here [dmlc/xgboost#7280](https://github.com/dmlc/xgboost/pull/7280)
-  the issue tracking this is here: [xgboost version conflict #12009](https://github.com/apache/tvm/issues/12009)
- this work is based on stalled PR here: [Upgrade autoscheduler xgboost callback #12144](https://github.com/apache/tvm/pull/12144)

Tested with booth AutoTVM and AutoSchedule.

Cc @Sunny-Island , @zxybazh @junrushao 
also Cc @driazati for CI xgboost>=1.7.x changes

Thanks,
~Cristian.

PR #4912, opened 2020-02-19
[REFACTOR] Polish ffi convention.

tqchen:
- Remove the src/api, keep registration local to the c++ function.
- Remove the api_internal as it is no longer needed.

PR #2032, opened 2018-10-30
[RUNTIME][OPENCL] Create command queue for each module

kazum:
This PR creates a command queue for each module and uses it for kernel execution.  The change improves performance when we run multiple modules concurrently.

The below is an experimental result (platform: OpenCL 1.2 CUDA 8.0.0, GPU: GeForce GTX TITAN X).
- before
  ```
  $ python apps/benchmark/gpu_imagenet_bench.py --target opencl --network resnet-18 --thread 4
  --------------------------------------------------
  Network Name         Mean Inference Time (std dev)
  --------------------------------------------------
  resnet-18            15.75 ms            (9.50 ms)
  resnet-18            15.71 ms            (9.59 ms)
  resnet-18            15.90 ms            (9.37 ms)
  resnet-18            15.78 ms            (9.40 ms)
  ```
- after
  ```
  $ python apps/benchmark/gpu_imagenet_bench.py --target opencl --network resnet-18 --thread 4
  --------------------------------------------------
  Network Name         Mean Inference Time (std dev)
  --------------------------------------------------
  resnet-18            9.76 ms             (10.56 ms)
  resnet-18            9.85 ms             (11.22 ms)
  resnet-18            9.88 ms             (11.86 ms)
  resnet-18            9.74 ms             (11.92 ms)
  ```

@tqchen @nishi-t  Can you please review?

PR #1680, opened 2018-08-31
INT8 conv operator implementation with NCHWc data layout for Intel machines

anijain2305:
This PR implements a conv operator for INT8 operations for Intel Skylake and upcoming Intel processors. Currently, this supports input of NCHWc format. Later, there will be NNVM effort to pick up this kernel for conv and perform kernel transform using "CorrectLayout". This PR tackles only the schedule for INT8 conv kernel.

# Background
**Hardware support**
Skylake provides HW support of performing a dot product of 2 4-int8 values while keeping the computational precision at INT32. For Skylake these instructions are vpamaddubsw, vpmaddwd. This support will be enhanced for VNNI instruction. More details can be found at this link (https://software.intel.com/en-us/articles/lower-numerical-precision-deep-learning-inference-and-training). 

**Why new schedule?**
These instructions require some modifications to the current FP32 schedule. The current schedule does not perform reduction across different elements of a vector register. But, Intel instructions allow reduction across 4 int8 values. Therefore, a new schedule is required. 

**Why not just rely on LLVM for codegen?**
LLVM codegen is not mature for generating these instructions. LLVM has a very restrictive pattern matching to lower down to these INT8 operations. We will need decent efforts on both LLVM and TVM side to reach an IR where LLVM can directly generate these instructions. Therefore, I am currently calling LLVM intrinsics directly from LLVM.

# Performance Speedup
These are different conv layers from resnet network. I don't have NNVM changes yet to run an end-to-end experiment. Will update these numbers when we have that. 

Workload | kernelSize | FP32 time | INT8 time | Speedup
-- | -- | -- | -- | --
Workload#0 | 3x3 | 1.27E-04 | 7.61E-05 | 1.668205764
Workload#1 | 1x1 | 1.45E-05 | 1.62E-05 | 0.8956588432
Workload#2 | 3x3 | 9.77E-05 | 4.52E-05 | 2.163945956
Workload#3 | 1x1 | 9.56E-06 | 8.55E-06 | 1.117662311
Workload#4 | 3x3 | 0.000124318896 | 9.35E-05 | 1.32998721
Workload#5 | 3x3 | 8.53E-05 | 4.26E-05 | 2.00352076
Workload#6 | 1x1 | 9.68E-06 | 8.45E-06 | 1.145552654
Workload#7 | 3x3 | 0.000101081785 | 8.07E-05 | 1.252166183
Workload#8 | 3x3 | 7.25E-05 | 4.82E-05 | 1.502493515
[...]

PR #14468, opened 2023-04-03
[AutoTVM] New rank-binary loss_type for the new xgboost >= 2.0.0 behaviour

cbalint13:
This PR fix the latest xgboost >= 2.0.0 behaviour requiring binarized labels.

* The strict behaviour was introduced by https://github.com/dmlc/xgboost/pull/8931
* A possible solution here (also [suggested](https://github.com/dmlc/xgboost/pull/9007#issuecomment-1494739265)) would be to binarize by threshold at the evaluation  calls.
* This is a continuation on https://github.com/apache/tvm/pull/14036 xgboost

This address ~~both~~  only```autotune``` (~~and ```autoscheduler```~~).

**Note:**
Unsure about TVM overall tunner impact, but we can introduce more sophisticated way of measuring AP like [PASCAL evenly spaced one](https://en.wikipedia.org/wiki/Evaluation_measures_(information_retrieval)#Average_precision), the advantages are unclear and would require extensive comparative tests.

---
The errors cought on TVM **autotune** process:
```
File "/usr/lib64/python3.11/site-packages/tvm/autotvm/tuner/xgboost_cost_model.py"
, line 538, in after_iteration
    bst_eval = model.eval_set(self.evals, epoch, feval)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/lib64/python3.11/site-packages/xgboost/core.py", line 1995, in eval_set
    _check_call(
File "/usr/lib64/python3.11/site-packages/xgboost/core.py", line 270, in _check_call
    raise XGBoostError(py_str(_LIB.XGBGetLastError()))
xgboost.core.XGBoostError: [15:59:59] /builddir/build/BUILD/xgboost/src/common/ranking_utils.h:378: 
Check failed: is_binary: MAP can only be used with binary labels.
```
---

Cc @Sunny-Island , @zxybazh , @junrushao , @vinx13 , please help with the review.

Thanks,
~Cristian.

---

Update:

 * The **autoschedule** (ansor) is not affected at all. 
 * The **autotune** ```reg``` (reg:linear)  ```loss_type``` is fine.
 * Only **autotune** with ```rank``` (rank:pairwise) ```loss_type``` is affected.

### DIFF
--- a/vta/scripts/tune_resnet.py
+++ b/vta/scripts/tune_resnet.py
@@ -183,8 +183,30 @@ def tune_tasks(
         prefix = "[Task %2d/%2d] " % (i + 1, len(tasks))
 
         # create tuner
-        if tuner == "xgb" or tuner == "xgb-rank":
+        if tuner == "xgb":
+            tuner_obj = XGBTuner(tsk, loss_type="reg")
+        elif tuner == "xgb_knob":
+            tuner_obj = XGBTuner(tsk, loss_type="reg", feature_type="knob")
+        elif tuner == "xgb_itervar":
+            tuner_obj = XGBTuner(tsk, loss_type="reg", feature_type="itervar")
+        elif tuner == "xgb_curve":
+            tuner_obj = XGBTuner(tsk, loss_type="reg", feature_type="curve")
+        elif tuner == "xgb_rank":
             tuner_obj = XGBTuner(tsk, loss_type="rank")
+        elif tuner == "xgb_rank_knob":
+            tuner_obj = XGBTuner(tsk, loss_type="rank", feature_type="knob")
+        elif tuner == "xgb_rank_itervar":
+            tuner_obj = XGBTuner(tsk, loss_type="rank", feature_type="itervar")
+        elif tuner == "xgb_rank_curve":
+            tuner_obj = XGBTuner(tsk, loss_type="rank", feature_type="curve")
+        elif tuner == "xgb_rank_binary":
+            tuner_obj = XGBTuner(tsk, loss_type="rank-binary")
+        elif tuner == "xgb_rank_binary_knob":
+            tuner_obj = XGBTuner(tsk, loss_type="rank-binary", feature_type="knob")
+        elif tuner == "xgb_rank_binary_itervar":
+            tuner_obj = XGBTuner(tsk, loss_type="rank-binary", feature_type="itervar")
+        elif tuner == "xgb_rank_binary_curve":
+            tuner_obj = XGBTuner(tsk, loss_type="rank-binary", feature_type="curve")
         elif tuner == "ga":
             tuner_obj = GATuner(tsk, pop_size=50)
         elif tuner == "random":
--- a/apps/benchmark/adreno/adreno_gpu_bench_clml.py
+++ b/apps/benchmark/adreno/adreno_gpu_bench_clml.py
@@ -116,6 +116,7 @@ def print_progress(msg):
 def tune_tasks(
     tasks,
     measure_option,
+    tuner="xgb",
     n_trial=1024,
     early_stopping=None,
     log_filename="tuning.log",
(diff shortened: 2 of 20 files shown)


tqchen:
Given that we are doing cost model. I am not sure if binarization is the best approach here. Can you dump out the labels and check the current assigned behavior?

cbalint13:
@tqchen 

> Given that we are doing cost model. I am not sure if binarization is the best approach here. 

* In a short discussion here this was suggested: https://github.com/dmlc/xgboost/pull/9007#issuecomment-1494739265
* Looking through [changes](https://github.com/dmlc/xgboost/pull/8931) , the old behaviour also clamped somehow the values (not clear for me if to pure binary).

> Can you dump out the labels and check the current assigned behavior?

* Sure, attached is a small script + dmatrix dump: [tvm-xgboost-dmatrix.zip](https://github.com/apache/tvm/files/11151803/tvm-xgboost-dmatrix.zip) with [results.txt](https://github.com/apache/tvm/files/11151857/results.txt)

* This was captured from a real tvm autotunning process targeting a rk3399 opencl device.

> 
> Likely we might want to move away from the MAP metric, and use other metric instead, either regression metric or pair-wise ranking.

* Apparently this proposal works, tunning finds good kernels, but the real impact is hard to measure (on personal side).


Another quick idea for now is to add condition of binarization to xgboost >=1.7.5  version, keeping the old behaviour.

cbalint13:
@tqchen , additionally to my response to the request in the previous message: 

> > Likely we might want to move away from the MAP metric, and use other metric instead, either regression metric or pair-wise ranking.
> 
> * Apparently this proposal works, tunning finds good kernels, but the real impact is hard to measure (on personal side).
> 
> Another quick idea for now is to add condition of binarization to xgboost >=1.7.5 version, keeping the old behaviour.

* If binarization is too **steep** , a simple (but recursive) way would be the [PASCAL](https://en.wikipedia.org/wiki/Evaluation_measures_(information_retrieval)#Average_precision) trick to avg/split into 11 levels: 

image

tqchen:
I think in this case we should change ranking loss to regression loss, use logistic regression so the values can still be used. binarization causes too much info loss

cbalint13:
@tqchen ,

> I think in this case we should change ranking loss to regression loss, use logistic regression so the values can still be used. binarization causes too much info loss

Updates:

1. The autoschedule (ansor) is not affected at all.
2. The autotune ```reg (reg:linear)``` loss_type is not affected.
3. Only autotune with ```rank (rank:pairwise)``` loss_type``` is affected.
4. Only xgboost > 2.0.0-dev (as it presents itself via py API).

I updated this PR code to do binarization **only in case**:
  - check if xgboost > 2 
  - check if rank:pairwise

Updated here the code, the title, the first comment (barred out any erroneous info).
I could imagine to leave this for **autotune** in a prepearing for what will be xgb > 2.0.0

tqchen:
I see, i think we should report error if binarization is needed, since the original intention was continuous prediction.

I know it might still work OK, but that was not the intention of the cost predictor.

Would be good to visit the default choice, i think if ranking is not possible, reg:logistic would be another good choice usually

cbalint13:
@tqchen ,

* Introduced a new ```rank-binary``` loss_type along with current ```rank``` and ```reg```.

> I see, i think we should report error if binarization is needed, since the original intention was continuous prediction.

* If loss_type is ```rank``` and xgboost >= 2.0.0 will report an error with explicit suggestion to use ```rank-binary``` instead.

> I know it might still work OK, but that was not the intention of the cost predictor.

* All remains as it was, it is user's choice (informed by tvm) to switch over with ```rank-binary``` or downgrade xgboost.

> Would be good to visit the default choice, i think if ranking is not possible, reg:logistic would be another good choice usually

* I won't touch this, all the described above would offer user a informed alternative.


Let me know if still need changes or more polishing, I stop here for now.
In future, will keep track on xgboost upcoming ```custom gain function``` , we may restore ```rank``` continuous behaviour.

tqchen:
we still need to make sure the default `loss_type` now updates to regression loss

cbalint13:
@tqchen ,

> Thanks @cbalint13 ! we still need to make sure the default `loss_type` now updates to regression loss, mainly because rank-binary do not fit into the customized loss models.

* Changed to "reg" all class init occurrences.
* Lowered test case rmse for the new "reg" default.

There are still tutorials / applications that use explicit "rank", would like to change all of them ?

tqchen:
@cbalint13 Yes, let us update and to change all to reg. cc @junrushao to double check cases in MetaSchedule.

Might be useful to use reg:logistic, if the output is scaled into [0, 1]

cbalint13:
@tqchen ,

> @cbalint13 Yes, let us update and to change all to reg.

* Updated all tutorials, apps and tests from any explicit  "rank" to "reg".

> cc @junrushao to double check cases in MetaSchedule.

* Double check is welcome, as quick info, none of my metascheduler test failed, xgboost would print:
```
xgboost.core.XGBoostError: [15:59:59] /builddir/build/BUILD/xgboost/src/common/ranking_utils.h:378: 
Check failed: is_binary: MAP can only be used with binary labels.
```
* Used xgboost 20230403 git hash 15e073ca .
 
> Might be useful to use reg:logistic, if the output is scaled into [0, 1]

My thought on the newly  introduced ```type_loss="rank-binary"```:
   * Is **alternative** to the original ```rank``` and **booth have**: ```("objective": "rank:pairwise")``` 
   * It binarize (within a cloned copy) the labels only for the xgboost eval step (being xgboost>=2 compatible).
   * The dynamics inside TVM further steps still remained with full continuous values.

I would leave it as described, if would like we can create another one ```type_loss="reg-binary" w/ reg:logistic```.
For the case reg:logistic the naming "rank-binary" loose the "rank"-ness, so it would be just another "reg" class type.

cbalint13:
@tqchen , @junrushao 

In continuation of previous comment, I  also attach here some test result.

Comparative test confirms that ```rank-binary``` (binarized only at eval step) behaves identically with ```rank``` (original):

    loss_type="rank-binary" (xgboost-2.0.0-dev 20230403 git hash 15e073ca)
    [Task  1/54] (conv2d_nchw_spatial_pack.mali) {17.75 GFLOPS / #4912 records} SKIP
    [Task  2/54] (conv2d_nchw_spatial_pack.mali) {40.74 GFLOPS / #1040 records} SKIP
    [Task  3/54] (conv2d_nchw_spatial_pack.mali) {19.63 GFLOPS / #2032 records} SKIP

    loss_type="rank" (xgboost-1.7.5 20230328 git hash 21d95f3d)
    [Task  1/54] (conv2d_nchw_spatial_pack.mali) {11.71 GFLOPS / #1680 records} SKIP
    [Task  2/54] (conv2d_nchw_spatial_pack.mali) {26.17 GFLOPS / #1024 records} SKIP
    [Task  3/54] (conv2d_nchw_spatial_pack.mali) {13.15 GFLOPS / #1040 records} SKIP

Note:
  *  In the case of rank-binary there was more steps (bit prologed / see amount of records) hence bit better results.
  * The tuned network was first three layers of yolov8s using float16 for half a day on a rk3399 board (nanopc-t4).

zxybazh:
@tvm-bot rerun

vta/scripts/tune_resnet.py
```diff
@@ -184,7 +184,7 @@ def tune_tasks(
 
         # create tuner
         if tuner == "xgb" or tuner == "xgb-rank":
```
FrozenGene, line 186: ```tuner == "xgb-rank"``` should be changed too.

cbalint13: @FrozenGene 
> `tuner == "xgb-rank"` should be changed too.

Thanks for the observation !

* fixed and exposed properly all such cases.
* exposed / updated tmvc with extended options coverage.
* extended the demo lists (taken from original ```tune_relay_arm.py``` for completeness)

cbalint13: @FrozenGene ,

Uptade

> @FrozenGene
> 
> > `tuner == "xgb-rank"` should be changed too.
> 
> Thanks for the observation !
> 
> * fixed and exposed properly all such cases.
> * exposed / updated tmvc with extended options coverage.
> * extended the demo lists (taken from original `tune_relay_arm.py` for completeness)

Plus:

* leaved "rank" types in the list for compatibility with xgboost < 2.0.0

FrozenGene (approved):

PR #14473, opened 2023-04-03
[Unity][CI] Update gpu and lint image

masahi:
I tried to update the GPU image for unity CI in https://github.com/apache/tvm/pull/14291 but it doesn't seem to be working. Trying sending the change from `apache` and also including the recent `lint` image update https://github.com/apache/tvm/pull/14373. Not sure if this is the correct protocol for updating CI images for unity.

@tqchen @driazati


tqchen:
indeed a branch in apache would be the way to go, normally we use unity-staging for that

PR #14474, opened 2023-04-03
[Unity][Hexagon] Don't pass raw scalars in hexagon/qnn/nn.py

kparzysz-quic:
Everything is expected to be either te.Tensor, or a PrimExpr, so make sure that function parameters conform to that. Also, remove incorrect type annotations from `qnn_requantize`.


yongwww (approved):

masahi:
Is this specific to Unity?

kparzysz-quic:
> Is this specific to Unity?

At least part of the problem was specific to a commit that's only in unity.

masahi (approved):

PR #14475, opened 2023-04-03
[pytest] Don't return values from test_* functions

kparzysz-quic:
Pytest expects test functions to return None, and errors to be flagged via exceptions. It actually emits warnings for "return" statements encountered in test_ functions.

### DIFF
--- a/tests/python/contrib/test_hexagon/test_maxpool2d_blocked.py
+++ b/tests/python/contrib/test_hexagon/test_maxpool2d_blocked.py
@@ -151,7 +151,7 @@ class TestMaxPooling(BaseMaxPooling):
             padding=(pad, pad, pad, pad),
             dtype=dtype,
         )
-        assert all([output, ref_output])
+        assert all([output is not None, ref_output is not None])
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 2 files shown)


Lunderberg (approved):
Thank you for the catch, and looks good to me!  If I recall correctly, these were initially intended to reduce duplication between some of the tests and the test fixtures, as a workaround for pytest's lack of subtests, but I missed the return statements in pre-PR cleanup.

tests/python/contrib/test_hexagon/test_maxpool2d_blocked.py
```diff
@@ -151,7 +151,7 @@ def test_maxpool(self, shape_nhwc, window_size, stride, pad, dtype, target):
             padding=(pad, pad, pad, pad),
             dtype=dtype,
         )
-        return output, ref_output
+        assert all([output is not None, ref_output is not None])
```
Lunderberg, line 154: If the assert is causing spurious unit test failures, it can be removed.  The test is to ensure that no exceptions are through while lowering/building, and no further assert is required.

kparzysz-quic: I did remove it at first, but then the linter complained that `output` and `ref_output` were unused values...  :|

PR #14477, opened 2023-04-03
feat: add spot instances for cpu tests

naut-brian:


naut-brian:
need to generate files using generate.py

naut-brian:
need to add input variable for tests + the actual node input. Need to remove hard code "CPU" node

ci/jenkins/templates/gpu_jenkinsfile.groovy.j2
```diff
@@ -193,7 +193,10 @@ def deploy_docs() {
   }
 }
 
-def deploy() {
```
driazati, line 196: I think this should always just be on-demand, this messes with external state (deploying the docs to github pages) and doesn't take too long anyways

driazati (approved):

PR #14478, opened 2023-04-03
[Unity][VM] Add Attention KV cache builtin

tqchen:
This PR provides a simple implementation of inplace attention kv cache for relax runtime. The main goal here is to help us enable auto-regressive decoding quickly in relax.

This is likely not the only way to support attention kv-cache. We keep the implementation private for now and will continue to evolve the relevant code.

### DIFF
--- a/src/runtime/relax_vm/attention_kv_cache.cc
+++ b/src/runtime/relax_vm/attention_kv_cache.cc
@@ -1,0 +1,162 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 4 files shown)


src/runtime/relax_vm/attention_kv_cache.cc
```diff
@@ -0,0 +1,162 @@
...
+  /*!
+   * \brief number of slots already filled.
+   */
+  int64_t fill_count{0};
+
+  /*!
+   * \brief View all current cached values as one array.
+   * \param shape The cached values.
+   */
+  NDArray View(const ShapeTuple& shape) {
```
jinhongyii, line 74: why do we need to pass shape as argument? just for check?

src/runtime/relax_vm/attention_kv_cache.cc
```diff
@@ -0,0 +1,162 @@
...
+      CHECK_EQ(shape[i], data->shape[i]) << "Dimension " << i << " mismatch";
+    }
+    return data.CreateView(shape, data->dtype);
+  }
+
+  /*!
+   * \brief Append value to the cache.
+   * \param value The value to be appended.
+   */
+  void Append(NDArray value) {
```
jinhongyii, line 86: I think we need to specify the dim to append because seq_len may not be the first dimension in kv cache

tqchen: That would be harder due to continuous memory requirement and atm we only handle the top dimension

src/runtime/relax_vm/attention_kv_cache.cc
tqchen: Yes only for safety checks

PR #14481, opened 2023-04-04
[Unity][Op] Extend `relax.op.take` op to match behavior with `topi.take`. 

sunggg:
Currently, `relax.op.take` implements semantics of [torch.take](https://pytorch.org/docs/stable/generated/torch.take.html) which has a stronger restriction than its backing `topi.take` which seems to follow the semantics of [numpy.take](https://numpy.org/doc/stable/reference/generated/numpy.take.html). 
i.e., `relax.op.take` enforces its input `indices` to be 1-d tensor. 

To match its behavior with `topi.take`, this PR extends the semantics of `relax.op.take` by removing such restriction without breaking existing test cases (except one for checking 1-d tensor restriction). 

As a result, this PR can support [onnx.gather](https://github.com/onnx/onnx/blob/main/docs/Changelog.md#Gather-13) which was the main motivation of my previous PR https://github.com/apache/tvm/pull/14457 by resolving the problem in my previous stab. 

cc. @jwfromm @yongwww @MasterJH5574

### DIFF
--- a/src/relax/op/tensor/index.cc
+++ b/src/relax/op/tensor/index.cc
@@ -51,8 +51,7 @@ StructInfo InferStructInfoTake(const Call& call, const BlockBuilder& ctx) {
   if (indices_sinfo->IsUnknownDtype()) {
     // TODO(tvm-team): Do we have an equivalent of `ctx->ReportFatal` for warning?
     LOG(WARNING) << "Data type of indice has not been specified. Assume it has an integer type.";
-  } else if (!indices_sinfo->IsUnknownDtype() &&
-             !(indices_sinfo->dtype.is_int() || indices_sinfo->dtype.is_uint())) {
+  } else if (!(indices_sinfo->dtype.is_int() || indices_sinfo->dtype.is_uint())) {
     ctx->ReportFatal(Diagnostic::Error(call)
                      << "Take op requires the input indices to have integer dtype. However, the "
                         "given indices dtype is "
--- a/src/relax/op/tensor/index.cc
+++ b/src/relax/op/tensor/index.cc
@@ -47,11 +47,10 @@ StructInfo InferStructInfoTake(const Call& call, const BlockBuilder& ctx) {
   Array<TensorStructInfo> input_sinfo = GetInputTensorStructInfo(call, ctx);
   TensorStructInfo data_sinfo = input_sinfo[0];
   TensorStructInfo indices_sinfo = input_sinfo[1];
-  if (indices_sinfo->ndim != 1) {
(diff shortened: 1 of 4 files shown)


src/relax/op/tensor/index.cc
```diff
@@ -47,11 +47,10 @@ StructInfo InferStructInfoTake(const Call& call, const BlockBuilder& ctx) {
...
-  if (indices_sinfo->ndim != 1) {
-    ctx->ReportFatal(Diagnostic::Error(call)
-                     << "Take op requires the input indices to be 1-dimensional tensor. However, "
-                        "the given indices ndim is "
-                     << indices_sinfo->ndim);
+
+  if (indices_sinfo->IsUnknownDtype()) {
+    // TODO(tvm-team): Do we have an equivalent of `ctx->ReportFatal` for warning?
+    LOG(WARNING) << "Data type of indice has not been specified. Assume it has an integer type.";
   } else if (!indices_sinfo->IsUnknownDtype() &&
```
yongwww, line 54: remove dup check `!indices_sinfo->IsUnknownDtype() &&`

yongwww (commented):
Overall looks good to me

psrivas2 (approved):

tqchen:
as long as the existing behavior are preserved, i think we can support multi-dim index

sunggg:
@tqchen I checked that existing tests are passed, so I believe the existing behavior is intact.

tqchen:
@MasterJH5574 would be great if you can help to take a quick look at this one

src/relax/op/tensor/index.cc
```diff
@@ -47,13 +47,11 @@ StructInfo InferStructInfoTake(const Call& call, const BlockBuilder& ctx) {
...
-  if (indices_sinfo->ndim != 1) {
-    ctx->ReportFatal(Diagnostic::Error(call)
-                     << "Take op requires the input indices to be 1-dimensional tensor. However, "
-                        "the given indices ndim is "
-                     << indices_sinfo->ndim);
-  } else if (!indices_sinfo->IsUnknownDtype() &&
-             !(indices_sinfo->dtype.is_int() || indices_sinfo->dtype.is_uint())) {
+
+  if (indices_sinfo->IsUnknownDtype()) {
+    // TODO(tvm-team): Do we have an equivalent of `ctx->ReportFatal` for warning?
```
MasterJH5574, line 52: Leaving as LOG(WARNING) looks fine to me.

MasterJH5574 (approved):

MasterJH5574:
@sunggg Sorry for the delay and thanks for the extension!

PR #14483, opened 2023-04-04
[TOPI][Target] Add SVE specific convolution

FranklandJack:
This commit will:
* Expose SVE as a target feature for Arm(R) Cortex(R) A-Profile CPUs.
* Update the compute definition of `conv2d_spatial_pack_nhwc` to defer to the LLVM backend for vectorization when compiling on an SVE enabled target for data tensors with unit width and height since this has been shown to be performant for wide vector architectures.

### DIFF
--- a/python/tvm/topi/arm_cpu/conv2d_spatial_pack.py
+++ b/python/tvm/topi/arm_cpu/conv2d_spatial_pack.py
@@ -20,6 +20,7 @@ from __future__ import absolute_import as _abs
 import tvm
 from tvm import te
 from tvm import autotvm
+from tvm.target import Target
 from tvm.autotvm.task.space import SplitEntity, OtherOptionEntity, AnnotateEntity, ReorderEntity
 from .. import nn
 from ..utils import get_const_tuple
@@ -316,12 +317,23 @@ def conv2d_spatial_pack_nhwc(cfg, data, kernel, strides, padding, dilation, out_
                     return candidate
             return 1
 
-        # Tile size 8 results in efficient vectorization for these schedules.
-        # If the axis is not divisible by 8, try 4
+        # For data tensors with unity height and width we can leave it to the
+        # backend to vectorize the inner loop. This has been observed to be more
+        # performant on SVE targets with a vector width > 128bits.
+        target = Target.current(allow_none=False)
+        if target.features.has_sve and OW == OH and OW == 1:
+            tile_size = [OC]
+            vectorize = "none"
+        else:
+            # Tile size 8 results in efficient vectorization for these schedules.
+            # If the axis is not divisible by 8, try 4
+            tile_size = [8, 4]
+            vectorize = "vec"
+
         cfg["tile_oh"] = SplitEntity([-1, 1])
         cfg["tile_ow"] = SplitEntity([-1, _tile_size(OW, [8, 4])])
-        cfg["tile_co"] = SplitEntity([-1, _tile_size(OC, [8, 4])])
-        cfg["ann_spatial"] = AnnotateEntity(["none", "vec"])
+        cfg["tile_co"] = SplitEntity([-1, _tile_size(OC, tile_size)])
+        cfg["ann_spatial"] = AnnotateEntity(["none", vectorize])
         cfg["ann_reduce"] = AnnotateEntity(["none", "none"])
         cfg["reorder_conv"] = ReorderEntity([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
         cfg["compat"] = OtherOptionEntity(0)
--- a/src/target/parsers/aprofile.cc
+++ b/src/target/parsers/aprofile.cc
@@ -106,29 +106,30 @@ static TargetFeatures GetFeatures(TargetJSON target) {
   Optional<String> mtriple = Downcast<Optional<String>>(target.Get("mtriple"));
   Optional<Array<String>> mattr = Downcast<Optional<Array<String>>>(target.Get("mattr"));
 
-  double arch_version = GetArchVersion(mattr);
+  const double arch_version = GetArchVersion(mattr);
 
-  bool is_aarch64 = IsAArch64(mtriple);
+  const bool is_aarch64 = IsAArch64(mtriple);
 
-  bool simd_flag = HasFlag(mcpu, mattr, "+neon") || HasFlag(mcpu, mattr, "+simd");
-  bool has_asimd = is_aarch64 || simd_flag;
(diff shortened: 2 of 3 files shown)


src/target/parsers/aprofile.cc
```diff
@@ -126,9 +127,8 @@ static TargetFeatures GetFeatures(TargetJSON target) {
   bool has_dotprod = (dotprod_default && !dotprod_disable) || (dotprod_support && dotprod_flag);
 
   return {
-      {"is_aarch64", Bool(is_aarch64)},
-      {"has_asimd", Bool(has_asimd)},
-      {"has_dotprod", Bool(has_dotprod)},
+      {"is_aarch64", Bool(is_aarch64)},  {"has_asimd", Bool(has_asimd)},
+      {"has_sve", Bool(has_sve)},        {"has_dotprod", Bool(has_dotprod)},
```
Mousius, line 131: Tests please 😸

Mousius (changes requested):

src/target/parsers/aprofile.cc
FranklandJack: Good shout, I've added a test for this [here](https://github.com/apache/tvm/pull/14483/files#diff-66ebf57fcac747725f36ace0cce2bb5d097bc6b3d3a6c225f6a367b4fb23ff08R293).

src/target/parsers/aprofile.cc
```diff
@@ -112,6 +112,7 @@ static TargetFeatures GetFeatures(TargetJSON target) {
 
   bool simd_flag = HasFlag(mcpu, mattr, "+neon") || HasFlag(mcpu, mattr, "+simd");
   bool has_asimd = is_aarch64 || simd_flag;
+  const auto has_sve = HasFlag(mcpu, mattr, "+sve");
```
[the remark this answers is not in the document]
Mousius: I don't *think* `const` adds any practical value here? The compiler likely won't do anything different and the values could be overwritten without causing concern so it's not aiding the programmer.

The repo loosely follows the [Google C++ Guidelines](https://google.github.io/styleguide/cppguide.html#Type_deduction), which advises against using `auto` unless it adds readability, in this case `auto` doesn't make it any easier to read or understand. The variable names could be used to infer what the type might be, but if it's explicit there's no further effort required on the part of the reader.

Introducing a single variation in the file also introduces a different weirdism for readers, where this line has been singled out for some unknown reason and future readers will likely find that at least interesting. Working in a shared codebase does mean working in a style which is generally consistent with the existing code so as to avoid such weirdisms.

python/tvm/topi/arm_cpu/conv2d_spatial_pack.py
```diff
@@ -316,12 +317,23 @@ def _tile_size(axis, candidates):
                     return candidate
             return 1
 
-        # Tile size 8 results in efficient vectorization for these schedules.
-        # If the axis is not divisible by 8, try 4
+        # For data tensors with unity height and width we can leave it to the
+        # backend to vectorize the inner loop. This has been observed to be more
+        # performant on SVE targets with a vector width > 128bits.
+        target = Target.current(allow_none=False)
+        if target.features.has_sve and OW == OH and OW == 1:
```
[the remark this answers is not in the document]
FranklandJack: Good question, not that has been observed. In the cases we saw the LLVM backend choose the optimal vectorization (fixed length vs. scalable) depending on the number of output channels. If we were forcing the use of scalable vectors though, I think you are right we would need a heuristic here :)

src/target/parsers/aprofile.cc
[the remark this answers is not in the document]
FranklandJack: I take your point on `auto` but I'm not convinced this shouldn't be `const` just to be consistent with the surrounding code when we could quite easily update all the variables here to be `const` as part of this PR.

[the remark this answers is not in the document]
Mousius: Sorry for the lag on this, I'm fine with `const` if you want to add it to everything 😸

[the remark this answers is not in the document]
FranklandJack: Nice, I've _constified_ all the things! 😝

tests/cpp/target/parsers/aprofile_test.cc
```diff
@@ -290,9 +290,28 @@ TEST(AProfileParser, ArchVersionInvalidLetter) {
   ASSERT_EQ(Downcast<Bool>(features.at("has_dotprod")), false);
 }
 
+using AProfileOptionalSVE = testing::TestWithParam<float>;
+TEST_P(AProfileOptionalSVE, OptionalSVESupport) {
+  const auto arch_attr = "+v" + std::to_string(GetParam()) + "a";
+
+  // Check that the "has_sve" feature is not set by default when "+sve" isn't set as an attribute.
+  auto target = ParseTargetWithAttrs("", "aarch64-arm-none-eabi", {arch_attr});
+  auto features = Downcast<TargetFeatures>(target.at("features"));
```
Mousius, line 299: Can you change these `auto`s to explicit types similar to the other tests?

Mousius (changes requested):

tests/cpp/target/parsers/aprofile_test.cc
FranklandJack: In the first and third case it avoids repeating the type name that is already on the RHS of the assignment operator. The repeated typename is redundant information to the reader and is just extra noise. Anyone reading case 1 can see that it is a `std::string` from the call to `std::to_string` and reading case 2 can see it is a `TargetFeatures` from the call to `Downcast<TargetFeatures>`. I don't have a strong opinion on this, but I do think `auto` adds value here and isn't just avoiding writing an explicit type.

FranklandJack: Any thoughts on the above, would be good to get this merged since it's been open a while :)

neildhickey: I presume the point that @Mousius is making here is that, while auto is fine to use in the way that you are using it, it doesn't match with the way these types have been specified in the rest of the file and, therefore, creates a moment of hesitation when skimming through the code and makes the reader think, "is something different being done here?". If this change was in a new file it would be fine, but being in the middle of a file, where in all other cases these types are specified differently, creates a bit of incongruousness.

FranklandJack: I've updated this to name the types.

Mousius (approved):
> @Mousius I believe your comments are addressed, do you want to take a look again?

Eek! Yep, looks good to me, apologies again for such latency - thanks for the ping @ekalda!

PR #14484, opened 2023-04-04
[ACL] Prevent offloading of per-channel quantized operators

NicolaLancellotti:
Currently, only per-layer quantization is supported in the Arm Compute Library runtime. However, there is no check that prevents the offloading of per-channel quantized operators, as a consequence, tvm fails during inference if such operators are found.

This patch adds a check that prevents offloading per-channel quantized operators.

### DIFF
--- a/tests/python/contrib/test_arm_compute_lib/test_dense.py
+++ b/tests/python/contrib/test_arm_compute_lib/test_dense.py
@@ -380,8 +380,51 @@ def test_codegen_qnn_dense(dtype):
         verify_codegen(func, exp_codegen)
 
 
+@pytest.mark.parametrize(
+    "param",
+    ["kernel_sc", "kernel_zp"],
+)
+def test_codegen_qnn_dense_per_channel_quantization(param):
+    if skip_codegen_test():
+        return
+
+    np.random.seed(0)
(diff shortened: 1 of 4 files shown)


tests/python/contrib/test_arm_compute_lib/test_dense.py
```diff
@@ -380,8 +380,51 @@ def test_codegen_qnn_dense(dtype):
...
+    exp_codegen = _get_expected_codegen(*args, has_bias=composite)
+    verify_codegen(func, exp_codegen, num_acl_modules=0, tvm_ops=3)
+
+
 if __name__ == "__main__":
     test_dense()
     test_qnn_dense()
     test_codegen_dense()
     test_codegen_qnn_dense()
+    test_codegen_qnn_dense_per_channel_quantization()
```
lhutton1, line 430: No need to add now, we should probably convert to the:
```
if __name__ == "__main__":
    tvm.testing.main()
```
format when we get a another chance

lhutton1 (approved):
Thanks @NicolaLancellotti, LGTM!

lhutton1:
Thanks @NicolaLancellotti!

PR #14485, opened 2023-04-04
[Relay] Check if the attribute "name" exists before accessing it

NicolaLancellotti:
I found that some models fail to compile with tvm when the target is: `compute-library, llvm -mtriple=aarch64-linux-gnu -mattr=+v8.2a,+neon`

The error is:  `AttributeError: <class 'tvm.ir.expr.GlobalVar'> has no attribute name`.

This patch checks if the attribute `name` exists before accessing it, and solves the compilation issues that I had, producing the expected result during the following inference.


ekalda:
Thanks @NicolaLancellotti!

PR #14486, opened 2023-04-04
[Draft][TVMScript][Unittest] Validate round-trip for each TIR lowering step

Lunderberg:
Prior to this PR, some of the IRModule transformations used during lowering can produce TIR that cannot be round-tripped through TVMScript.  Since TVMScript is the default method for printing all TIR, this can make it difficult to identify which pass has introduced a breaking change.

The first commit in this PR adds a test that checks whether a module can correctly round-trip from TIR to TVMScript and back, for each lowering pass used in `tvm.build`.  Each of the following independent commits resolves one of the issues found by the more general test, and adds a specific test for the round-trip failure that it resolves.

My plan is to break out these independent commits will be into separate PRs, for ease of review.  After all the component PRs have landed or found an alternative fix, this PR will be rebased to only consist of the new test itself, and will be opened for review.

### DIFF
--- a/include/tvm/node/structural_equal.h
+++ b/include/tvm/node/structural_equal.h
@@ -195,20 +195,40 @@ class SEqualReducer {
    * \param rhs The right operand.
    * \return the immediate check result.
    */
-  bool operator()(const double& lhs, const double& rhs) const;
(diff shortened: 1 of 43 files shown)


Lunderberg:
Closing this PR, as it is no longer relevant to the majority of TVM development.

PR #14491, opened 2023-04-04
[TIR] Use String instead of StringImm for AttrStmtNode::node

Lunderberg:
`tir::StringImm` can round-trip through TVMScript when used in a context that requires a PrimExpr, such as the arguments of a `tir::Call`.  However, contexts that only require a `ObjectRef`, such as the `AttrStmtNode::node`, use the same TVMScript representation as `"string_value"`, but are parsed `tvm::String` instances.

This commit updates `MakePackedAPI` to use `String` instead of `StringImm` in its default value for `AttrStmtNode::node`.


Lunderberg:
The test added by this PR depends on the https://github.com/apache/tvm/pull/14488.  Should re-run CI after that one lands.

kparzysz-quic (approved):
LGTM.  Is this the only place where this happens?

PR #14493, opened 2023-04-04
[Unity][Op] introduce `ScatterElement` op

sunggg:
This PR brings `ScatterElement` op that can be useful for preprocessing argument tensors in the frontend. 
For example, for `topi::dynamic_strided_slice` expects its argument tensors (e.g., `begin`, `end`) to have the same rank with the data tensor. (see [link](https://github.com/apache/tvm/blob/unity/include/tvm/topi/transform.h#L657-L661))
```C++
inline Tensor dynamic_strided_slice(const Tensor& x, const Array<PrimExpr>& begin,
                                    const Array<PrimExpr>& end, const Array<PrimExpr>& strides,
                                    std::string name = "T_dynamic_strided_slice",
                                    std::string tag = kInjective) {
  const size_t src_tensor_dim = x->shape.size();
  ICHECK_LE(begin.size(), src_tensor_dim);
  ICHECK_LE(end.size(), src_tensor_dim);
  ICHECK_LE(strides.size(), src_tensor_dim);
  ICHECK_EQ(begin.size(), end.size());
  ICHECK_EQ(begin.size(), strides.size());
  ...
```
Thus, frontend often needs to preprocess those argument tensors to make sure the later constraints are satisfied. 

For example, say if we have the following situation: 
```Python
data_rank = 4
axes = [2,0]
begin = [1,2]
# We want to preprocess `begin` to 
new_begin = [2, 0, 1, 0]  
```
With `ScatterElement`, we can conveniently do so. 
```Python
new_begin = R.scatter_elements(
           relax.const([0] * data_rank),   # data
           axes,    # indices 
           begin,   # updates
           axis=0, 
           reduction="update")
)
```

cc. @jwfromm @yongwww @psrivas2


yongwww:
Not sure if this ScatterElement will be used by some other ops as well. I feel we can handle this scattering stuff in strided_slice itself. For example, the strided_slice is able to support begin/end has less elements than the rank of input data: https://github.com/apache/tvm/blob/unity/tests/python/topi/python/test_topi_transform.py#L850

BTW, the ICHECK_LE in the posted sample expects the begin/end has equal (or less) elements

sunggg:
@yongwww thanks for taking a look! 
In TOPI, we have two versions of strided slice: One static, one dynamic. 
Static one is less constraints, you can pass arbitrarily number of begin/end elements since those values are compile-time constants. However, dynamic strided slice is bit more restricted because you need to know the "rank" of such arguments as presented in the description. 

When you look at the test for dynamic slice, they all have the equal ranks with the data rank: https://github.com/apache/tvm/blob/unity/tests/python/topi/python/test_topi_transform.py#L854

yongwww:
@sunggg  Gotcha, thanks for sharing this info!!!

src/relax/op/tensor/manipulate.cc
```diff
@@ -1359,5 +1359,119 @@ TVM_REGISTER_OP("relax.cumsum")
...
+StructInfo InferStructInfoScatterElements(const Call& call, const BlockBuilder& ctx) {
+  arith::Analyzer* analyzer = ctx->GetAnalyzer();
+  const auto* data_sinfo = GetStructInfoAs<TensorStructInfoNode>(call->args[0]);
+  const auto* indices_sinfo = GetStructInfoAs<TensorStructInfoNode>(call->args[1]);
+  const auto* updates_sinfo = GetStructInfoAs<TensorStructInfoNode>(call->args[2]);
+
+  auto diag_def = [&](const TensorStructInfoNode* sinfo, String name, String type_key) {
+    if (sinfo == nullptr) {
+      ctx->ReportFatal(Diagnostic::Error(call)
+                       << "ScatterEelemtns requires the input " << name
```
yongwww, line 1384: typo: ScatterElements

yongwww (approved):
Looks good to me! Thanks for adding this!

vinx13 (approved):

PR #14495, opened 2023-04-04
[TIR] Merged kDeviceThreadAxis and kUseDynamicSharedMemoryTag

Lunderberg:
Previously, `kDeviceThreadAxis` defined the IterVar to be used for each thread/block axis, and `kUseDynamicSharedMemoryTag` defined whether dynamic memory allocations exist, which are primarily used to produce a list of strings by `tvm::codegen::ExtractFuncInfo`.  Because `kDeviceThreadAxis` is a `Array<IterVar>`, the IterVar is used prior to its definition site at `tir::attr::thread_extent`, which results in errors when attempting to round-trip through TVMScript.

This commit replaces these attributes with
`attr::kKernelLaunchParams`, which directly contains the kernel launch parameters.  These are expressed as an `Array<String>`, allowing the generated TVMScript to successfully round-trip.

This is part of changes described in https://github.com/apache/tvm/pull/14486, to improve round-trip TVMScript failures that occur in lowering.

### DIFF
--- a/include/tvm/tir/function.h
+++ b/include/tvm/tir/function.h
@@ -272,10 +272,11 @@ PrimFunc Specialize(PrimFunc func, const Map<Var, ObjectRef>& param_map);
  * \sa tvm::attr
  */
 namespace attr {
+
 /*!
  * \brief List of thread IterVar that a DeviceLaunch function corresponds to.
  *
- * Type: Array<tir::IterVar>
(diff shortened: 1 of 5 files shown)


Lunderberg:
@tvm-bot rerun

vinx13 (approved):

PR #14496, opened 2023-04-04
[TIR] Update SplitHostDevice to post-process with ConvertSSA

Lunderberg:
Prior to this commit, the same `tir::Var` could be defined both in the body of the host function and in a parameters of the device function.  While this is legal, as TIR only requires SSA across each `PrimFunc`, this does result in failure to round-trip through TVMScript, because the round-tripped TIR uses a new `tir::Var` in each instance.  This PR adds a `ConvertSSA()` transform that can apply to an entire `IRModule`, in addition to the existing `ConvertSSA(Stmt)` utility.

### DIFF
--- a/include/tvm/tir/transform.h
+++ b/include/tvm/tir/transform.h
@@ -177,6 +177,19 @@ TVM_DLL Pass RewriteUnsafeSelect();
 TVM_DLL Pass Simplify();
 
 /*!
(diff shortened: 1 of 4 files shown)


Lunderberg:
@tvm-bot rerun

tqchen (approved):

PR #14497, opened 2023-04-04
[TIR] Improved SeqStmt::Flatten utility

Lunderberg:
Prior to this commit, `SeqStmt::Flatten` could accept an arbitrary number of arguments, where each argument was of type `const tir::Stmt&` or an iterable.  However, if `SeqStmt::Flatten` were passed a subclass of `tir::Stmt`, the templated overload was selected as the better match.  This commit rewrites `SeqStmt::Flatten` using C++17's `"constexpr if"` feature, to handle cases of `SeqStmt`, superclasses of `SeqStmt`, and other subclasses of `Stmt`.

### DIFF
--- a/include/tvm/tir/stmt.h
+++ b/include/tvm/tir/stmt.h
@@ -734,19 +734,33 @@ class SeqStmt : public Stmt {
    public:
     explicit Flattener(Array<Stmt>* seq) : seq_(seq) {}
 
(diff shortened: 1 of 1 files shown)


Lunderberg:
This issue came up during development on https://github.com/apache/tvm/pull/14486, but ended up not being needed for it.

kparzysz-quic (approved):

PR #14498, opened 2023-04-04
[Node] Utility methods for ObjectPathPair handling

Lunderberg:
This commit adds a templated overload to `SEqualReducer::operator()` that accepts a lambda function to update the path of the LHS and RHS of the comparison.

```c++
// Usage prior to this utility function
if (equal.IsPathTracingEnabled()) {
  const ObjectPathPair& self_paths = equal.GetCurrentObjectPaths();
  ObjectPathPair attr_paths = {self_paths->lhs_path->Attr("value"),
                               self_paths->rhs_path->Attr("value")};
  if (!equal(this->value, other->value, attr_paths)) return false;
} else {
  if (!equal(this->value, other->value)) return false;
}

// Usage after this utility function
if (!equal(this->value, other->value,
           [](const auto& path) { return path->Attr("value"); })) {
  return false;
}
```

### DIFF
--- a/src/ir/module.cc
+++ b/src/ir/module.cc
@@ -63,46 +63,46 @@ IRModule::IRModule(tvm::Map<GlobalVar, BaseFunc> functions,
 }
 
 bool IRModuleNode::SEqualReduce(const IRModuleNode* other, SEqualReducer equal) const {
-  if (!equal(this->attrs, other->attrs)) return false;
+  if (!equal(this->attrs, other->attrs, [](const auto& path) { return path->Attr("attrs"); })) {
+    return false;
+  }
 
-  if (functions.size() != other->functions.size()) return false;
-  // Update GlobalVar remap
+  if (equal.IsPathTracingEnabled()) {
+    if ((functions.size() != other->functions.size()) ||
+        (type_definitions.size() != other->type_definitions.size())) {
+      return false;
+    }
+  }
+
+  // Define remaps for GlobalVar and GlobalTypeVar based on their
+  // string name.  Early bail-out is only performed when path-tracing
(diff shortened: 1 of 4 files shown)


Lunderberg:
This arose during debugging for https://github.com/apache/tvm/pull/14486, but wasn't strictly necessary for it.

src/ir/module.cc
```diff
@@ -63,46 +63,46 @@ IRModule::IRModule(tvm::Map<GlobalVar, BaseFunc> functions,
...
   }
 
-  if (type_definitions.size() != other->type_definitions.size()) return false;
-  // Update GlobalTypeVar remap
-  for (const auto& gtv : this->GetGlobalTypeVars()) {
-    if (!other->ContainGlobalTypeVar(gtv->name_hint)) return false;
-    if (!equal.DefEqual(gtv, other->GetGlobalTypeVar(gtv->name_hint))) return false;
+  // Checking functions and type definitions
+  if (!equal(this->functions, other->functions,
+             [](const auto& path) { return path->Attr("functions"); })) {
```
vinx13, line 98: Is `func_path` (adding function name to the path) handled like before?

Lunderberg: It is, yes.  The `IRModule`'s handler only needs to handle appending `.functions` to the path, with the function name added by the equality check on `Map`.  This results in a path of `<root>.functions[I.GlobalVar("func")].body...`.

I've added a unit test that verifies that the error message includes the function name, as that's definitely a good thing to verify for user-friendliness.

Lunderberg:
@tvm-bot rerun

vinx13 (approved):

PR #14501, opened 2023-04-04
[Unity][Graph matching] Improved matching algorithm and implementation

masahi:
* As discussed and agreed in https://github.com/apache/tvm/pull/14417#discussion_r1152443315, remove `start_hint` feature to simplify the API and the implementation

* To support the `start_hint` feature, the current algorithm has "bidirectional" nature: Matching can start from any node. This introduces recursive matching on both child and parent nodes, which makes the implementation complicated and reasoning about the algorithm almost impossible. Since this PR removes `start_hint`, we can now implement the algorithm in a purely one-directional manner, by starting matching from root nodes (those without parent nodes). Note that the pattern constraint graph is assumed to be a DAG. 

* The current implementation relies on side-effecting operations to maintain matchings. So when matching fails in the middle, we need to carefully "undo" matchings in the intermediate state (see https://github.com/apache/tvm/pull/14440). This, again, makes reasoning about the algorithm more difficult. I've reworked the data structure so that the matching algorithm is now "purely functional". Also, we can use `const` everywhere now.

Overall, this PR makes the matching algorithm a fairly standard one based on backtracking search. The implementation is also easier to understand and safer.

@ganler @sunggg @vinx13 @cyx-6 @psrivas2

### DIFF
--- a/src/relax/ir/dataflow_matcher.cc
+++ b/src/relax/ir/dataflow_matcher.cc
@@ -615,6 +615,7 @@ static std::optional<MatchState> TryMatch(const PNode& p, const RNode& r, DFPatt
     auto p_node_parent = p.parents[i].first;
     if (p_node_parent->ptr->IsInstance<WildcardPatternNode>()) {
       if (auto v = result.matched(p_node_parent); v && v != r.parents[i]->ptr) {
+        // A parent wildcard pattern is already matched to other variable.
         return std::nullopt;
       }
       result.add(p_node_parent, r.parents[i]);
@@ -667,6 +668,7 @@ static std::optional<MatchState> MatchTree(
     const std::unordered_map<const VarNode*, RNode>& var2node, DFPatternMatcher* matcher,
     const std::vector<DFPattern>& roots, const MatcherUseDefAnalysis& ud_analysis) {
   auto get_next_root = [&](int root_idx) -> const PNode* {
+    // Look for the next unmatched root node.
     for (; root_idx < roots.size(); ++root_idx) {
       const auto& root = pattern2node.at(roots[root_idx].get());
       if (!current_matches.matched(root)) {
@@ -678,7 +680,8 @@ static std::optional<MatchState> MatchTree(
 
   const auto root = get_next_root(current_root_idx);
 
-  if (!root || current_matches.matched(root)) {
+  if (!root) {
+    // All root nodes have been matched
     return current_matches;
   }
 
@@ -686,11 +689,14 @@ static std::optional<MatchState> MatchTree(
     const RNode& r_node = var2node.at(var);
     if (current_matches.matched(r_node)) continue;
     if (auto matches = TryMatch(*root, r_node, matcher, ud_analysis)) {
+      // Recursivly try to match the next subtree.
       if (auto matches_rec = MatchTree(*matches, current_root_idx + 1, pattern2node, var2node,
                                        matcher, roots, ud_analysis)) {
         matches->add(*matches_rec);
         return matches;
       }
+      // Recursive matching has failed, backtrack.
+      continue;
     }
   }
 
--- a/src/relax/ir/dataflow_matcher.cc
+++ b/src/relax/ir/dataflow_matcher.cc
@@ -578,9 +578,9 @@ struct MatchState {
     match_r_p[r] = p;
   }
 
-  void add(const MatchState& other) {
-    match_p_r.insert(other.match_p_r.cbegin(), other.match_p_r.cend());
-    match_r_p.insert(other.match_r_p.cbegin(), other.match_r_p.cend());
+  void add(MatchState&& other) {
+    match_p_r.merge(std::move(other.match_p_r));
+    match_r_p.merge(std::move(other.match_r_p));
   }
 
   const VarNode* matched(const PNode* p) const {
@@ -655,7 +655,7 @@ static std::optional<MatchState> TryMatch(const PNode& p, const RNode& r, DFPatt
 
       if (auto match_rec = TryMatch(*pchild, *rchild, m, ud_analysis)) {
         result.add(pchild, rchild);
-        result.add(*match_rec);
+        result.add(std::move(*match_rec));
       }
     }
     if (!result.matched(pchild) || !any_cons_sat) return std::nullopt;
@@ -694,7 +694,7 @@ static std::optional<MatchState> MatchTree(
       // Recursivly try to match the next subtree.
       if (auto matches_rec = MatchTree(*matches, current_root_idx + 1, pattern2node, var2node,
                                        matcher, roots, ud_analysis)) {
-        matches->add(*matches_rec);
+        matches->add(std::move(*matches_rec));
         return matches;
       }
       // Recursive matching has failed, backtrack.
--- a/src/relax/ir/dataflow_matcher.cc
+++ b/src/relax/ir/dataflow_matcher.cc
@@ -590,8 +590,8 @@ struct MatchState {
   }
 
   const DFPatternNode* matched(const RNode* r) const {
-    if (!match_r_p.count(r)) return nullptr;
-    return match_r_p.at(r)->ptr;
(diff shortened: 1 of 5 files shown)


src/relax/ir/dataflow_matcher.cc
```diff
@@ -523,109 +523,117 @@ bool MatchExpr(DFPattern pattern, Expr expr, Optional<Map<Var, Expr>> bindings_o
...
+    match_p_r[p] = r;
+    match_r_p[r] = p;
+  }
 
-  const auto try_match_update_undo = [&](PNode* p, RNode* r) {
-    if (auto undo_more = try_match(p, r, m, def2use, use2def)) {
-      undo.insert(undo.end(), undo_more->begin(), undo_more->end());
-      return true;
+  void add(const MatchState& other) {
+    for (const auto& [p, r] : other.match_p_r) {
```
ganler, line 582: May consider making it more cache-friendly:

```c++
match_p_r.insert(other.match_p_r.cbegin(), other.match_p_r.cend());
match_r_p.insert(other.match_r_p.cbegin(), other.match_r_p.cend());
```

ganler: Here I assume the incoming key-value pairs won't conflict with that in `*this`. But if it is intended to overriding values of existing keys, then use [`insert_or_assign`](https://en.cppreference.com/w/cpp/container/unordered_map/insert_or_assign) over `insert`.

src/relax/ir/dataflow_matcher.cc
```diff
@@ -523,109 +523,117 @@ bool MatchExpr(DFPattern pattern, Expr expr, Optional<Map<Var, Expr>> bindings_o
...
+      add(p, r);
     }
-    return false;
-  };
+  }
 
-  commit(p, r);
+  const VarNode* matched(const PNode* p) const {
+    if (!match_p_r.count(p)) return nullptr;
+    return match_p_r.at(p)->ptr;
```
ganler, line 589: minor perf to get things down with only one look-up (feel free to ignore it if you believe it hurts readbility):

```suggestion
    if (auto it = match_p_r.find(p); it != match_p_r.end()) return it->ptr;
    return nullptr;
```

src/relax/ir/dataflow_matcher.cc
```diff
@@ -523,109 +523,117 @@ bool MatchExpr(DFPattern pattern, Expr expr, Optional<Map<Var, Expr>> bindings_o
...
 
-  // match parent patterns.
-  for (auto& [pparent, constraints] : p->parents) {
-    bool any_cons_sat = false;
-    for (auto& rparent : r->parents) {
-      // skip if mismatch.
-      if (rparent->matched && rparent->matched != pparent->ptr) continue;
+  const DFPatternNode* matched(const RNode* r) const {
+    if (!match_r_p.count(r)) return nullptr;
+    return match_r_p.at(r)->ptr;
```
ganler, line 594: Same:

```suggestion
    if (auto it = match_r_p.find(p); it != match_r_p.end()) return it->ptr;
    return nullptr;
```

src/relax/ir/dataflow_matcher.cc
```diff
@@ -523,109 +523,117 @@ bool MatchExpr(DFPattern pattern, Expr expr, Optional<Map<Var, Expr>> bindings_o
...
+
+  MatchState result;
+
+  for (size_t i = 0; i < p.parents.size(); ++i) {
+    const auto p_node_parent = p.parents[i].first;
+    if (p_node_parent->ptr->IsInstance<WildcardPatternNode>()) {
+      ICHECK_EQ(p.parents.size(), r.parents.size());
+      if (auto v = result.matched(p_node_parent); v && v != r.parents[i]->ptr) {
+        // A parent wildcard pattern is already matched to other variable.
+        return std::nullopt;
```
ganler, line 620: Sorry if I missed anything here. But could you kindly help explain why we need to add special handlings for nodes whose parents (aka arguments) are wildcard?

ganler: I guess you meant that if it is wildcard, we don't have to obey "one-one" mapping (originally suggested here https://github.com/tlc-pack/relax/issues/160#issue-1268184762 in the `One-One Match` section). If that's the case, I felt it might still make some sense to optionally allow one-one mapping even for wildcards?

ganler (approved):
LGTM. Apprecaite the consective efforts from @masahi to make the pattern matcher better and better!

src/relax/ir/dataflow_matcher.cc
masahi: I wonder why C++ doesn't define `get_or()` function to express this idiom easily.

src/relax/ir/dataflow_matcher.cc
masahi: This is for preventing false matching below:

```
 lv0 = R.matmul(x1, w0)
 lv1 = R.matmul(x2, w1)
 lv2 = R.matmul(x2, w2)
```

```
inp_pat = wildcard()
Q_weight_pat = wildcard()
K_weight_pat = wildcard()
V_weight_pat = wildcard()

```

The pattern adds a constraint that the LHS of three matmuls be the same. After matching the first matmul, we learn that `inp_pat` matches with `x1`. On the second matmul pattern + the binding (`lv1`), `m->Match(...)` would return true , since it simply matches a matmul pattern with two wildcards. This additional check tells us that `inp_pat` has already been matched to `x1`, so the second matching where `inp_pat` would match to `x2` must be disallowed.

ganler: > The pattern adds a constraint that the LHS of three matmuls be the same.

Correct as we assume one DFPattern should only match one actual/real node.

> On the second matmul pattern + the binding (lv1), m->Match(...) would return true , since it simply matches a matmul pattern with two wildcards.

In theory, we want it to return `false` due to one-one mapping. IIRC, I thought when matching the 2nd mm pattern to `R.matmul(x2, w1)`, we will check if `x2` is the the matched RNode for `inp_pat` (which is `x1`) => so false. That being said, I don't understand why we need to specifically check the `wildcard` pattern. Because in theory if a pattern is matched to an RNode, and when re-matching the same pattern we shall assert that the incoming RNode must be the matched one.

src/relax/ir/dataflow_matcher.cc
```diff
@@ -523,109 +523,116 @@ bool MatchExpr(DFPattern pattern, Expr expr, Optional<Map<Var, Expr>> bindings_o
...
-  };
+struct MatchState {
+  void add(const PNode* p, const RNode* r) {
+    match_p_r[p] = r;
+    match_r_p[r] = p;
+  }
 
-  commit(p, r);
+  void add(const MatchState& other) {
+    match_p_r.insert(other.match_p_r.cbegin(), other.match_p_r.cend());
```
ganler, line 582: If you are interested, as TVM now supports C++ 17, we might just do:
```c++
match_p_r.merge(other.match_p_r);
match_r_p.merge(other.match_r_p);
```
https://en.cppreference.com/w/cpp/container/unordered_map/merge

src/relax/ir/dataflow_matcher.cc
masahi: > In theory, we want it to return false due to one-one mapping

Note that by `m->Match(...)` I meant a pure DF pattern matcher. It is not aware of RNode etc.

> That being said, I don't understand why we need to specifically check the wildcard pattern

Right, this additional condition is not specific to wildcard. I lifted this check out of `if (p_node_parent->ptr->IsInstance<WildcardPatternNode>()) {` and I'm trying to see if I can remove this if block (L619-L622) altogether.

src/relax/ir/dataflow_matcher.cc
masahi: nice I didn't know about this!

src/relax/ir/dataflow_matcher.cc
masahi: Good news, I found two bugs in my implementation. They are fixed in the latest commit, and this entire `if` block has been removed.

PR #14502, opened 2023-04-05
[TensorIR][Schedule] Bugfix on `compute_inline` when buffer has symbolic shape

yzh119:
# The bug
Currently, the `compute_inline` function would report an error if we try to `compute_inline` a block where the buffer has a symbolic shape, below is a minimal reproduction:

```python
import tvm
import numpy as np

@T.prim_func
def failed_compute_inline(a: T.handle, b: T.handle):
    N = T.int32()
    A = T.match_buffer(a, (N, 32), "float32")
    temp = T.alloc_buffer((N, 32), "float32", scope="local")
    B = T.match_buffer(b, (N, 32), "float32")
    for i in range(N):
        for j in range(32):
            with T.block("T"):
                temp[vi, vj] = A[vi, vj] + 1
        for j in range(32):
            with T.block("B"):
                B[vi, vj] = temp[vi, vj] * A[vi, vj]

sch = tvm.tir.Schedule(failed_compute_inline)
T = sch.get_block("T")
B = sch.get_block("B")
j = sch.get_loops(B)[-1]
sch.compute_inline(T)
```

The reason is that the validity check analysis would use `UndefinedVars` to find variables that appeared in the block being `compute_inline` 'd. However, the `UndefinedVars` would visit `Buffer` and counts variables in buffer shape/strides,  and actually, we only care about buffer access indices. 

# Fix
This PR fixes the issue by adding a `visit_buffer` argument in the `UndefinedVars` function so that we can exclude variables that appeared in buffer shape/strides.

cc @Lunderberg @cyx-6 @junrushao @wrongtest-intellif

### DIFF
--- a/src/tir/schedule/primitive/compute_inline.cc
+++ b/src/tir/schedule/primitive/compute_inline.cc
@@ -296,7 +296,7 @@ class BaseInliner : public StmtExprMutator {
    */
   static int GetNumUndefinedNonpointerVars(const Stmt& stmt) {
     // Do not visit buffer shape/strides.
-    auto undefined_vars = UndefinedVars(stmt, {}, false);
+    auto undefined_vars = UndefinedVars(/*stmt=*/stmt, /*defs=*/{}, /*visit_buffer=*/false);
     // Buffer pointers and the inlined indices are allowed, but no
     // other variables may appear in the inlined block.
     int num_nonpointer_vars = 0;
--- a/src/tir/schedule/primitive/compute_inline.cc
+++ b/src/tir/schedule/primitive/compute_inline.cc
@@ -16,6 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
+#include "../../analysis/var_use_def_analysis.h"
 #include "../utils.h"
 
 namespace tvm {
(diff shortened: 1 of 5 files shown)


src/tir/schedule/primitive/compute_inline.cc
```diff
@@ -295,7 +295,8 @@ class BaseInliner : public StmtExprMutator {
    * \param stmt The statement in which to count undefined variables
    */
   static int GetNumUndefinedNonpointerVars(const Stmt& stmt) {
-    auto undefined_vars = UndefinedVars(stmt, {});
+    // Do not visit buffer shape/strides.
+    auto undefined_vars = UndefinedVars(/*stmt=*/stmt, /*defs=*/{}, /*visit_buffer=*/false);
```
Lunderberg, line 299: If I'm reading it correctly, it looks like this would also ignore variables that are used in the shape/stride of buffers that are declared internal to `stmt`, but the suppression of the undefined variable failures should only apply to buffers that are declared externally to `stmt`.

Instead of defining contexts in which variables should not be considered, could we instead define variables should be ignored by `UndefinedVars`?  If a variable's definition occurs outside the LCA of the producer/consumer, then it would be legal to use in the inlined statement.  The dynamic buffer shapes would then need to follow this more general condition, rather than being ignored entirely.

Lunderberg: As I'm thinking on it, this would also allow `compute_inline` to apply in cases that are currently forbidden, but are semantically valid.  In the example below, the block `"B"` couldn't be inlined, because it refers to the variable `value`.  However, `value`'s definition is accessible from both block `"B"` and block `"C"`, so it could be inlined without risk of an undefined variable.

```python
@T.prim_func
def func(A: T.Buffer(64, "int32"), C: T.Buffer(64, "int32")):
    B = T.alloc_buffer(64, "int32")

    value: T.int32 = 42

    for i in range(64):
        with T.block("B"):
            vi = T.axis.remap("S", [i])
            B[vi] = value

    for i in range(64):
        with T.block("C"):
            vi = T.axis.remap("S", [i])
            C[vi] = value
```

yzh119: @Lunderberg Thanks for your suggestion!
I agree with you that collecting variables outside LCA producer/consumer is a better idea. I'll revise the code accordingly and ping you again once it's ready.

yzh119: Hi @Lunderberg sorry for being late, would you mind helping review #14451 instead? The `GetNumUndefinedNonpointerVars ` check doesn't seem to work on nested blocks (in dynamic workloads we always blockize the dynamic loop, which would emit nested blocks). So I prefer moving on with the solution in #14451 which uses a simpler check.

PR #14507, opened 2023-04-05
[Unity][Refactor] Use NameSupply for unique symbol generation and remove NameTable

masahi:
`NameTable` is used in Relax code but we now have `NameSupply` in `main`. They serve exactly the same purpose (gensym), so we should use the more standard one (`NameSupply`) for consistency. 

@tqchen @slyubomirsky

### DIFF
--- a/include/tvm/ir/name_supply.h
+++ b/include/tvm/ir/name_supply.h
@@ -54,7 +54,9 @@ class NameSupplyNode : public Object {
    * \brief Generates a unique name from this NameSupply.
    * \param name The name from which the generated name is derived.
    * \param add_prefix If set to true, then the prefix of this NameSupply will be prepended to the
-   * name. \return A unique name.
+   * name.
+   * \param add_underscore If set to true, add '_' between prefix and a digit.
+   * \return A unique name.
    */
   String FreshName(const String& name, bool add_prefix = true, bool add_underscore = true);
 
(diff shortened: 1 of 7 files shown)


include/tvm/ir/name_supply.h
```diff
@@ -54,9 +54,11 @@ class NameSupplyNode : public Object {
    * \brief Generates a unique name from this NameSupply.
    * \param name The name from which the generated name is derived.
    * \param add_prefix If set to true, then the prefix of this NameSupply will be prepended to the
-   * name. \return A unique name.
+   * name.
+   * \param add_underscore If set to true, add '_' between prefix and a digit.
+   * \return A unique name.
    */
-  String FreshName(const String& name, bool add_prefix = true);
+  String FreshName(const String& name, bool add_prefix = true, bool add_underscore = true);
```
masahi, line 61: This change is only for backward-compat according to the variable naming convention of `BlockBuilder` (`lv1` vs `lv_1`). `NameSupply` adds an underscore before a digit by default. 

If we are ok with introducing such trivial variable naming change, I'm more than happy to remove this diff.

tqchen: I think no suffix is better overall for UX, we can upstream this change

tqchen (approved):

PR #14509, opened 2023-04-05
[Unity][Analysis] Get symbolic TIR vars from struct info

MasterJH5574:
This PR adds an analysis function which helps to fetch all the TIR variables that appear in a given struct info object. This is helpful for passes that need to be aware of symbolic shape dynamism.

PR #14511, opened 2023-04-05
[Unity] Memory planning with TIR var upper bound

MasterJH5574:
This PR enhances the memory planning pass with TIR var upper bound
annotation, so that it supports dynamic shape cases when we have
an estimate for the maximum values of the dynamic vars.

### DIFF
--- a/src/relax/transform/static_plan_block_memory.cc
+++ b/src/relax/transform/static_plan_block_memory.cc
@@ -48,11 +48,26 @@
  * - insert kill_storage at the end of each binding block, for all the storage
  * tokens that are allocated inside the binding block, as the memory planning
  * only works on block level.
+ *
+ * The memory planning pass "supports" dynamic shape in the way of TIR variable
+ * upper bound annotation. To be more specific, we can annotate the attribute
+ * "tir_var_upper_bound" to Relax functions. The attribute value is a dict from
+ * strings to integers, denoting the name of TIR variables to the upper bound
+ * values of the TIR vars. **The annotated upper bound attribute only applies
+ * to TIR vars in the function signature for clarity.**
+ *
+ * For example, we can annotate a Relax function with
+ *   `R.func_attr({"tir_var_upper_bound": {"n": 1024}})`.
+ * It means the maximum value of variable that names "n" in the function
+ * signature will have upper bound 1024. And we will use 1024 as its value
+ * during memory planning.
  */
+#include <tvm/arith/analyzer.h>
 #include <tvm/relax/analysis.h>
 #include <tvm/relax/expr_functor.h>
 #include <tvm/relax/nested_msg.h>
 #include <tvm/relax/transform.h>
+#include <tvm/tir/stmt_functor.h>
 
 #include <map>
 #include <set>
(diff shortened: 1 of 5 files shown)


MasterJH5574:
Depends on #14509.

src/relax/transform/static_plan_block_memory.cc
```diff
@@ -476,11 +521,15 @@ class StorageAllocatorInit : public StorageAllocatorBaseVisitor {
     token2block_.erase(token_to_discard.get());
   }
 
+  /*! \brief The arithmetic analyzer. */
+  arith::Analyzer ana;
```
tqchen, line 525: ana_

src/relax/transform/static_plan_block_memory.cc
```diff
@@ -476,11 +521,15 @@ class StorageAllocatorInit : public StorageAllocatorBaseVisitor {
...
   /*!
    * \brief The context IRModule, used for checking if a callee function is
    * a PrimFunc inside the IRModule.
    */
   const IRModule& ctx_mod_;
+  /*! \brief The mapping from TIR variables to their respective upper bound values. */
+  std::unordered_map<const tir::VarNode*, IntImm> var_upper_bound_;
```
tqchen, line 532: prefer `unordered_map<tir::Var, IntImm, ObjectPtrHash, ObjectPtrEqual>`, mainly to avoid var get de-allocated and re-allocated, does not matter in here

src/relax/transform/static_plan_block_memory.cc
MasterJH5574: Ah thanks my bad. It was added in the last minute so missed this point.

tqchen (commented):
Some minor nits and LGTM

src/relax/transform/static_plan_block_memory.cc
```diff
@@ -401,8 +431,23 @@ class StorageAllocatorInit : public StorageAllocatorBaseVisitor {
     ICHECK(sinfo->dtype == Downcast<DataTypeImm>(call->args[1])->value);
     ICHECK(!token_map_.count(call));
 
-    // No support for symbolic shape at this moment.
+    // Use the upper bounds of TIR vars as their values.
+    Array<PrimExpr> upper_bounded_shape;
+    upper_bounded_shape.reserve(shape->values.size());
     for (const PrimExpr& dim_len : shape->values) {
+      upper_bounded_shape.push_back(
+          ana_.Simplify(tir::Substitute(dim_len, [this](const tir::Var& v) -> tvm::PrimExpr {
```
tqchen, line 439: Do not use substitute, instead use ConstIntBound. see https://github.com/apache/tvm/blob/main/include/tvm/arith/analyzer.h#L131, this helps to avoid cases where expression is not monotonically increasing with the var

MasterJH5574: Thanks for bringing it up! Fixed.

tqchen (approved):

yzh119:
In SparseTIR we have statements like `T.assume_bound`, and we might unify them in the future :)

spectrometerHBH (approved):

PR #14221, opened 2023-03-07
[Unity][Transform] SimplifyNormInference

SiriusNEO:
This PR adds a new pass SimplifyNormInference to unpack the norm operator into a sequence of operators, which is same as the pass SimplifyInference in Relay.

Co-authored-by: Siyuan Feng <hzfengsy@sjtu.edu.cn>

PR #14512, opened 2023-04-05
[Unity][Transform] Some Improvements on pass DecomposeOps

SiriusNEO:
Prior to this PR, we first have a pass named `SimplifyNormInference` (https://github.com/apache/tvm/pull/14221), which is used to decompose `batch_norm` into simple operators for optimization during inference. And there are some follow-up changes:

- https://github.com/apache/tvm/pull/14282 This PR introduces a policy to decompose `tensor_to_shape` op.
- https://github.com/mlc-ai/relax/pull/170 In mlc, we make `SimplifyNormInference` support op simplification during training (Because `batch_norm` behaves differently in inference and training).
- And in the future maybe there will be more operators needed to be decomposed.

So this PR polishes the changes in https://github.com/apache/tvm/pull/14282 mainly in the following aspects:
- Use one mutator (instead of two) to visit and decompose ops.
- Separate inference and training by registering two passes (Adviced by TQ).
- Shorten the name (DecomposeCompositeOps -> DecomposeOps)

Now the code is clear and easy to read. Now we use `if-else if-else if ...` pattern to recognize the op we want to decompose. In the future maybe we can introduce a map or register the decomposition policy in the op attribute, just like what we do in the `LegalizeOps` pass.


SiriusNEO:
cc @tqchen @sunggg @Hzfengsy

src/relax/transform/decompose_ops.cc
```diff
@@ -0,0 +1,267 @@
...
+  Expr one_minus_mom = MakeConstantScalar(1 - attrs->momentum, sinfo->dtype);
+
+  return Tuple({
+      out,
+      add(multiply(one_minus_mom, moving_mean), multiply(momentum, data_mean)),
+      add(multiply(one_minus_mom, moving_var), multiply(momentum, data_var)),
+  });
+}
+
+Expr SimplifyLayerNorm(const Call& call) {
```
vinx13, line 123: We don't need to decompose `LayerNorm` since it can be implemented more efficiently as a single op

sunggg: Is there any complication for fusion if we have a composite ops in general? (not specifically about `LayerNorm`). i.e., Is there a case where we cannot fuse the composite op but we can when we decompose?

sunggg (commented):
LGTM! Thank you for the improvement. 
Just a few comments.

src/relax/transform/decompose_ops.cc
SiriusNEO: Ah here we simplify `LayerNorm` for now just because it is easy to get its gradient. If it is more efficient as a single op, maybe in the future we can remove this from the pass after we finish the gradient function of the whole `LayerNorm`

SiriusNEO: Or maybe we can add some arguments for the pass to let the user specify what operators he want to simplify?

Hzfengsy (approved):

tqchen:
@sunggg please take another look

src/relax/transform/decompose_ops.cc
vinx13: > Is there any complication for fusion if we have a composite ops in general? 
Yes, we need to be aware of the implementation when setting op type (`kOpaque` or `kInjective`). This will cause different fusion results.

PR #14513, opened 2023-04-05
[Unity] Add rewriting for CUDA graph capturing

vinx13:
This PR added the pass for CUDA graph rewriting. It identifies the regions that can be executed with CUDA graph and lifts them into new functions for runtime capturing.

cc @tqchen @junrushao @spectrometerHBH @MasterJH5574 @masahi @csullivan @sunggg


python/tvm/relax/transform/transform.py
```diff
@@ -726,6 +726,17 @@ def SplitCallTIRByPattern(patterns, fcodegen) -> tvm.ir.transform.Pass:
     return _ffi_api.SplitCallTIRByPattern(patterns, fcodegen)  # type: ignore
 
 
+def RewriteCUDAGraph() -> tvm.ir.transform.Pass:
+    """Rewrite a Relax module for executing with CUDA graph. This pass identifies the regions that
+    can be executed with CUDA graph and lifts them into new functions for runtime graph capturing.
+
+    Returns
+    -------
+    ret: tvm.ir.transform.Pass
```
Hzfengsy, line 735:
```suggestion
    ret: tvm.ir.transform.Pass
        The registered pass for rewriting cuda graph
```

python/tvm/relax/vm_build.py
```diff
@@ -297,6 +297,10 @@ def foo(x: Tensor((3, 4), "float32"), y: Tensor((3, 4), "float32")):
     passes.append(relax.transform.ToNonDataflow())
     passes.append(relax.transform.CallTIRRewrite())
     passes.append(relax.transform.StaticPlanBlockMemory())
+
+    if tvm.transform.PassContext.current().config.get("relax.backend.use_cuda_graph", False):
```
Hzfengsy, line 301: As we want to keep the building pipeline minimum, I'm not sure if there is any other way to decouple cuda graph rewriting and the "minimum" building pipeline. I know it may depends on the static memory planning, so I won't block it but would like us to consider it carefully. Also cc @tqchen

src/relax/transform/rewrite_cuda_graph.cc
```diff
@@ -0,0 +1,512 @@
...
+      return;
+    }
+    ExprMutator::VisitBinding_(op);
+  }
+
+  std::unordered_map<const VarNode*, LiftedFunctionRewritePlan> subgraph_launches;
+  std::unordered_map<const VarNode*, Expr> var_redef_;
+  std::unordered_set<const VarNode*> lifted_bindings_;
+  int index_alloc = 0;
+  int index_capture = 0;
```
Hzfengsy, line 490:
```suggestion
  int index_alloc_ = 0;
  int index_capture_ = 0;
```

src/relax/transform/rewrite_cuda_graph.cc
```diff
@@ -0,0 +1,512 @@
...
+      return;
+    }
+    if (lifted_bindings_.count(op->var.get())) {
+      // The binding is lifted to the subgraph and will be removed from the original function.
+      return;
+    }
+    ExprMutator::VisitBinding_(op);
+  }
+
+  std::unordered_map<const VarNode*, LiftedFunctionRewritePlan> subgraph_launches;
```
Hzfengsy, line 486:
```suggestion
  std::unordered_map<const VarNode*, LiftedFunctionRewritePlan> subgraph_launches_;
```

python/tvm/relax/vm_build.py
tqchen: Perhaps one approach is to lift out the static memory planning and TIR rewrite phases into another function. 

Then cuda graph can call these passes from outside and then pass to build, which detects if the function have already pass the runtime optimization stage, and skips the passes

src/relax/transform/rewrite_cuda_graph.cc
```diff
@@ -0,0 +1,512 @@
...
+ * \brief Pass for transforming Relax module to execute with CUDA graph.
+ *
+ * CUDA graph provides a way to capture a sequence of CUDA kernel launches in the runtime and
+ * save them as a graph. The graph can be executed multiple times with less overhead than launching
+ * kernels individually. This pass rewrites the Relax module to execute with CUDA graph.
+ *
+ * The transformation is done in two steps:
+ *
+ * 1. Identify the regions that can be captured by CUDA graph and create the rewriting plan with
+ * `CUDAGraphRewritePlaner`.
```
masahi, line 31: typo `Planer`

src/relax/transform/rewrite_cuda_graph.cc
```diff
@@ -0,0 +1,512 @@
...
+ * A region is a subgraph of the Relax function that are executed statically. A region is executed
+ * statically if 1) it only depends on the memory allocated internally in the Relax function with
+ * constant shapes, 2) it only contains kernel launches (there are no control flow).
+ *
+ * This transformation is expected to run after `StaticPlanBlockMemory`. After
+ * `StaticPlanBlockMemory`, all the tensors that can be statically allocated are allocated with
+ * `R.memory.alloc_storage` and `R.memory.alloc_tensor`, while other tensors will be allocated via
+ * `R.builtin.alloc_tensor`.
+ *
+ * `CUDAGraphRewritePlaner` is executed at the level of BindingBlock. It first identify all the
```
masahi, line 42: same typo

PR #14517, opened 2023-04-06
[Unity][Op] Negative Log Likelihood Loss

SiriusNEO:
This op is widely used when we want to do training in Relax. Since we decide to upstream the Gradient pass from [mlc](https://github.com/mlc-ai/relax), we need to first upstream this operator. Its interface is similar with Relay and TOPI.


SiriusNEO:
cc @MasterJH5574

yongwww (approved):
looks awesome!

MasterJH5574 (approved):

PR #14525, opened 2023-04-07
[Unity] Properly handle tuple-outputting function in `FuseOpsByPattern`

masahi:
Previously, it has been assumed that when a fused function outputs a tuple, `TupleGetItem` needs to be attached to each output element to properly remap variables. But when an op that emits a tuple (e.g. `split`) is wrapped into a new function by this pass, that tuple can become the output of the fused function on its own. In such cases, a user of the output of such functions expects to get a tuple, so `TupleGetItem` must not be attached.

This PR also fixes a bug in the grouping algorithm, where `TupleGetItemPattern` that has been matched is not put into the matching group. 

@cyx-6


cyx-6 (approved):
LGTM!

PR #14531, opened 2023-04-07
[TVMScript] Preserve LetStmt of constants

Lunderberg:
Prior to this commit, the `bind_assign_value` implementation for TIR would treat assignment of constants (e.g. `j = 42`) as meta-variables to be inserted at their point of use.  This commit updates the parsing to treat `j = 42` as a shorthand for the TIR `LetStmt`, similar to the update made in https://github.com/apache/tvm/pull/14320.  This behavior is more consistent with the other uses of `bind_assign_value` as assignment, with `j = T.meta_var(42)` used to represent meta-variables.

### DIFF
--- a/python/tvm/script/parser/tir/parser.py
+++ b/python/tvm/script/parser/tir/parser.py
@@ -142,14 +142,14 @@ def bind_assign_value(self: Parser, node: doc.expr, var_name: str, value: Any) -
     ):
         IRBuilder.name(var_name, value)
         return value
-    elif isinstance(value, PrimExpr):
+    else:
+        value = tvm.runtime.convert(value)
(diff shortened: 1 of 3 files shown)


junrushao:
The change makes sense to me! Originally we had to take care of this only for 100% backward compatibility reasons, but in fact this behavior is much better in terms of consistency. CC @wrongtest-intellif let me know what you think given you guys are heavily using TVMScript

Lunderberg:
@junrushao That makes sense, especially when preparing the new parsing implementation to be a drop-in replacement for the older implementation.  Thank you for pinging @wrongtest-intellif , and I've marked it as ready-for-review.

tqchen (approved):

PR #14532, opened 2023-04-07
[Arith] Implement statistics counters for RewriteSimplifier

Lunderberg:
Previously, so long as `RewriteSimplifier` produces the same output, unit tests of its behavior would pass.  This could have severe performance regressions, such as the one resolved in https://github.com/apache/tvm/pull/14528, which caused the runtime of two test to increase from ~1.5 seconds to ~10 minutes each.

This commit implements statistics counts in RewriteSimplifier, which are exposed through both the C++ and Python APIs, and uses these to guard against the known performance regression from https://github.com/apache/tvm/pull/14528.


src/arith/rewrite_simplify.h
```diff
@@ -39,6 +40,36 @@ namespace arith {
 
 using namespace tir;
 
+struct RewriteSimplifierStatsNode : Object {
+  int nodes_visited{0};
+  int constraints_entered{0};
+  int rewrites_attempted{0};
```
tqchen, line 46: use int64_t

tqchen:
One thing that worth thinking about is the relation of simplification and persistence of analyzer. Since we encourage persistence of analyzer in general within a pass, so we should document in max step to encourage users to keep that in mind

src/arith/rewrite_simplify.h
Lunderberg: Thank you, updated here and in the `SetMaximumRewriteSteps` method.

Lunderberg:
Good point on the re-use of the same `Analyzer` instance.  I've added that in the description of `SetMaximumRewriteSteps`, that re-use of the same object is necessary in order to have accurate usage counters.

src/arith/rewrite_simplify.h
```diff
@@ -87,9 +125,29 @@ class RewriteSimplifier::Impl : public IRMutatorWithAnalyzer {
...
+  void RecordAttemptedRewrite() { stats_.rewrites_attempted++; }
+  void RecordRewrite() {
+    stats_.rewrites_performed++;
+
+    ICHECK(maximum_rewrite_steps_ <= 0 || stats_.rewrites_performed <= maximum_rewrite_steps_)
+        << "RewriteSimplifier exceeded maximum number of rewrites allowed ("
+        << maximum_rewrite_steps_ << ")";
+  }
+
+  bool is_currently_visiting_{false};
```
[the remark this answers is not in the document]
Lunderberg: Thank you, and removed.  This was part of an additional counter that I ended up not implementing, but failed to strip out of the final PR.

src/arith/rewrite_simplify.h
```diff
@@ -39,6 +40,41 @@ namespace arith {
 
 using namespace tir;
 
+/* Record of
```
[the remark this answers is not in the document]
Lunderberg: Thank you, and added.

src/tir/transforms/remove_no_op.cc
```diff
@@ -316,14 +323,19 @@ Pass RemoveNoOp() {
...
-      touch_pattern.emplace(f->body);
+      touch_pattern.emplace(f->body, config->max_simplification_steps);
     }
 
     arith::Analyzer analyzer;
+    analyzer.rewrite_simplify.SetMaximumRewriteSteps(config->max_simplification_steps);
 
-    auto* n = f.CopyOnWrite();
-    n->body = NoOpRemover::Apply(std::move(n->body), &analyzer, std::move(touch_pattern), nullptr);
+    {
```
Aleksei-grovety, line 334: Is there a specific reason for adding curly braces here?

Lunderberg: They aren't necessary for correctness here, but were added present to limit the scope of `write_ptr`.  It doesn't have much impact in a function of this size, especially with the return statement just afterward, but for longer functions it can help to limit the scope of variables that are only needed for a few following lines.

That said, the implementation no longer requires updating the signature of `NoOpRemover::Apply`, so this change has become unrelated to the overall PR.

Aleksei-grovety (approved):

tqchen (approved):

PR #14536, opened 2023-04-08
[QNN] Implement 'qnn.softmax'

maekawatoshiki:
This PR implements `qnn.softmax` which is the quantized operator of `nn.softmax`.
The implementation is based on the algorithm proposed in https://arxiv.org/pdf/2207.01405.pdf.

### DIFF
--- a/src/relay/qnn/op/softmax.cc
+++ b/src/relay/qnn/op/softmax.cc
@@ -1,0 +1,144 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/*!
+ * \file src/relay/qnn/op/softmax.cc
+ * \brief QNN softmax operator.
+ */
+#include <tvm/relay/analysis.h>
+#include <tvm/relay/op_attr_types.h>
+
+#include "op_common.h"
+#include "tvm/ir/expr.h"
+#include "tvm/relay/attrs/nn.h"
+#include "tvm/relay/type.h"
+#include "tvm/runtime/data_type.h"
+#include "tvm/runtime/logging.h"
+#include "tvm/topi/reduction.h"
+
+namespace tvm {
+namespace relay {
+namespace qnn {
+
+bool QnnSoftmaxRel(const Array<Type>& types, int num_inputs, const Attrs& attrs,
+                   const TypeReporter& reporter) {
+  // Expected Types: input, scale, zero_point, output_scale, output_zero_point, output
+  ICHECK_EQ(types.size(), 6);
+  const auto* x = types[0].as<TensorTypeNode>();
+  if (x == nullptr) return false;
+  ICHECK(x->dtype == DataType::Int(8))
+      << "Expected quantized softmax type(int8) for input but was " << x->dtype;
+
+  // Check the types of scale and zero points.
+  for (size_t i = 1; i < 5; ++i) {
+    if (types[i].as<IncompleteTypeNode>()) {
+      return false;
+    }
+  }
+
+  ICHECK(IsScalarType(types[1], DataType::Float(32)));  // scale
+  ICHECK(IsScalarType(types[2], DataType::Int(32)));    // zero_point
+  ICHECK(IsScalarType(types[3], DataType::Float(32)));  // scale
+  ICHECK(IsScalarType(types[4], DataType::Int(32)));    // zero_point
+
+  // Assign types for scale and zero points.
+  reporter->Assign(types[1], TensorType({}, DataType::Float(32)));  // scale
+  reporter->Assign(types[2], TensorType({}, DataType::Int(32)));    // zero_point
+  reporter->Assign(types[3], TensorType({}, DataType::Float(32)));  // scale
+  reporter->Assign(types[4], TensorType({}, DataType::Int(32)));    // zero_point
+
+  // Collect the input tensor and output tensor devoid of scale and zero points to reuse Relay
+  // IdentityRel infer type function.
+  Array<Type> tensor_types = {types[0], types[5]};
+  return IdentityRel(tensor_types, 2, attrs, reporter);
+}
+
+// Positional relay function to create quantized softmax operator used by frontend FFI.
+Expr MakeQuantizedSoftmax(Expr x, int axis, Expr scale, Expr zero_point, Expr output_scale,
+                          Expr output_zero_point) {
+  auto attrs = make_object<SoftmaxAttrs>();
+  attrs->axis = axis;
+  static const Op& op = Op::Get("qnn.softmax");
+  return Call(op, {x, scale, zero_point, output_scale, output_zero_point}, Attrs(attrs), {});
+}
+
+/*
+ * \brief Canonicalizes the QNN softmax op.
+ */
+Expr QnnSoftmaxCanonicalize(const Attrs& attrs, const Array<Expr>& new_args,
+                            const Array<tvm::relay::Type>& arg_types) {
+  // Expected: input, scale, zero_point, output_scale, output_zero_point
+  ICHECK_EQ(new_args.size(), 5);
+
+  const Expr input_scale = new_args[1];
+  const Expr input_zero_point = new_args[2];
+  const Expr output_scale = new_args[3];
+  const Expr output_zero_point = new_args[4];
+  const int axis = attrs.as<SoftmaxAttrs>()->axis;
+
+  // Refer to the Algorithm 1 in https://arxiv.org/pdf/2207.01405.pdf
+
+  const Expr quantized_data =
+      Subtract(Cast(new_args[0], DataType::Int(32)), Cast(input_zero_point, DataType::Int(32)));
+
+  const Expr x_0 = ConvertDtype(
+      Round(Divide(MakeConstantScalar(DataType::Float(32), 1.f), input_scale)), DataType::Int(32));
+  const Expr max = Max(quantized_data, {axis}, true, false);
+  const Expr x = Subtract(quantized_data, max);
+
+  const auto const_i32 = [&](int32_t val) { return MakeConstantScalar(DataType::Int(32), val); };
+  const int n = 8;
+  const int m = 30;
+  const int bits = 8;
+  const Expr x_p = Subtract(Add(x, RightShift(x, const_i32(1))), RightShift(x, const_i32(4)));
+  const Expr q = Divide(x_p, Negative(x_0));
+  const Expr r = Subtract(x_p, Multiply(q, Negative(x_0)));
+  const Expr x_b = Add(RightShift(r, const_i32(1)), x_0);
+  const Expr exps = LeftShift(x_b, Subtract(const_i32(n), q));
+  const Expr sums = Sum(exps, {axis}, true, false);
+  const Expr output =
+      RightShift(Multiply(Divide(const_i32(1 << m), sums), exps), const_i32(m - (bits - 1)));
+  const Expr requantized =
+      Requantize(output, arg_types[0].as<TensorTypeNode>()->shape,
+                 MakeConstantScalar(DataType::Float(32), 1.f / (1 << (bits - 1))), const_i32(0),
+                 output_scale, output_zero_point, DataType::Int(bits), 0);
+
+  return requantized;
+}
+
+RELAY_REGISTER_OP("qnn.softmax")
+    .describe("Softmax for quantized tensors.")
+    .set_attrs_type<SoftmaxAttrs>()
+    .set_num_inputs(5)
+    .add_argument("data", "Quantized Tensor", "The input data.")
+    .add_argument("scale", "Tensor", "The quantization scale of the input tensor.")
+    .add_argument("zero_point", "Tensor", "The quantization zero_point of the input tensor.")
+    .add_argument("output_scale", "Tensor", "The quantization scale of the output tensor.")
+    .add_argument("output_zero_point", "Tensor",
+                  "The quantization zero_point of the output tensor.")
+    .set_support_level(11)
+    .add_type_rel("QSoftmax", QnnSoftmaxRel)
+    .set_attr<TNonComputational>("TNonComputational", true)
+    .set_attr<FTVMLegalize>("FTVMQnnCanonicalize", QnnSoftmaxCanonicalize);
+
+TVM_REGISTER_GLOBAL("relay.qnn.op._make.softmax").set_body_typed(MakeQuantizedSoftmax);
+
+}  // namespace qnn
+}  // namespace relay
+}  // namespace tvm
--- a/src/relay/qnn/op/softmax.cc
+++ b/src/relay/qnn/op/softmax.cc
@@ -80,6 +80,12 @@ Expr MakeQuantizedSoftmax(Expr x, int axis, Expr scale, Expr zero_point, Expr ou
 
 /*
  * \brief Canonicalizes the QNN softmax op.
+ * \param attrs The Softmax attrs.
+ * \param new_args The new mutated args to the call node.
+ * \param arg_types The types of input and output.
+ * \return The sequence of Relay ops for softmax op.
+ * \note This op is highly experimental and sometimes lacks accuracy.
+ *       Be aware that the input scale must be in the range of 0 to 1.
  */
 Expr QnnSoftmaxCanonicalize(const Attrs& attrs, const Array<Expr>& new_args,
                             const Array<tvm::relay::Type>& arg_types) {
--- a/src/relay/qnn/op/softmax.cc
+++ b/src/relay/qnn/op/softmax.cc
@@ -108,14 +108,14 @@ Expr QnnSoftmaxCanonicalize(const Attrs& attrs, const Array<Expr>& new_args,
   const Expr max = Max(quantized_data, {axis}, true, false);
   const Expr x = Subtract(quantized_data, max);
 
-  const int n = 8;
   const int m = 30;
   const int bits = 8;
   const Expr x_p = Subtract(Add(x, RightShift(x, const_i32(1))), RightShift(x, const_i32(4)));
-  const Expr q = Divide(x_p, Negative(x_0));
+  const Expr q = Clip(Divide(x_p, Negative(x_0)), 0, 20);
+  const Expr max_q = Max(q, {axis}, true, false);
   const Expr r = Subtract(x_p, Multiply(q, Negative(x_0)));
   const Expr x_b = Add(RightShift(r, const_i32(1)), x_0);
-  const Expr exps = LeftShift(x_b, Subtract(const_i32(n), q));
+  const Expr exps = LeftShift(x_b, Subtract(max_q, q));
   const Expr sums = Sum(exps, {axis}, true, false);
   const Expr output =
       RightShift(Multiply(Divide(const_i32(1 << m), sums), exps), const_i32(m - (bits - 1)));
--- a/src/relay/qnn/op/softmax.cc
+++ b/src/relay/qnn/op/softmax.cc
@@ -86,7 +86,15 @@ Expr QnnSoftmaxCanonicalize(const Attrs& attrs, const Array<Expr>& new_args,
   // Expected: input, scale, zero_point, output_scale, output_zero_point
   ICHECK_EQ(new_args.size(), 5);
 
-  const Expr input_scale = new_args[1];
+  const auto const_i32 = [&](int32_t val) { return MakeConstantScalar(DataType::Int(32), val); };
+  const auto const_f32 = [&](float val) { return MakeConstantScalar(DataType::Float(32), val); };
+
+  const auto const_input_scale = new_args[1].as<ConstantNode>();
+  ICHECK(const_input_scale) << "Input scale should be constant.";
+  ICHECK(const_input_scale->is_scalar()) << "Input scale should be scalar.";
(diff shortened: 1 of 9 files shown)


src/relay/qnn/op/softmax.cc
```diff
@@ -0,0 +1,148 @@
...
+bool QnnSoftmaxRel(const Array<Type>& types, int num_inputs, const Attrs& attrs,
+                   const TypeReporter& reporter) {
+  // Expected Types: input, scale, zero_point, output_scale, output_zero_point, output
+  ICHECK_EQ(types.size(), 6);
+  const auto* x = types[0].as<TensorTypeNode>();
+  if (x == nullptr) return false;
+  ICHECK(x->dtype == DataType::Int(8))
+      << "Expected quantized softmax type(int8) for input but was " << x->dtype;
+  const auto* param = attrs.as<SoftmaxAttrs>();
+  ICHECK(param != nullptr) << "SoftmaxAttrs cannot be nullptr.";
```
masahi, line 48: `param` can be removed since it's not used

src/relay/qnn/op/softmax.cc
```diff
@@ -0,0 +1,148 @@
...
+  const Expr x_b = Add(RightShift(r, MakeConstantScalar(DataType::Int(32), 1)), x_0);
+  const Expr exps = RightShift(LeftShift(x_b, MakeConstantScalar(DataType::Int(32), n)), q);
+  const Expr sums = Sum(exps, {axis}, true, false);
+  const Expr output =
+      RightShift(Multiply(Divide(MakeConstantScalar(DataType::Int(32), 1 << m), sums), exps),
+                 MakeConstantScalar(DataType::Int(32), m - (bits - 1)));
+  const Expr requantized =
+      Requantize(output, arg_types[0].as<TensorTypeNode>()->shape,
+                 MakeConstantScalar(DataType::Float(32), 1.f / (1 << (bits - 1))),
+                 MakeConstantScalar(DataType::Int(32), 0), output_scale, output_zero_point,
```
masahi, line 123: I suggest creating a helper function to clean up many `MakeConstantScalar(DataType::Int(32), ...)`.

maekawatoshiki:
Thank you for the review, @masahi. I've just modified the code.

src/relay/qnn/op/softmax.cc
```diff
@@ -0,0 +1,144 @@
...
+
+  const auto const_i32 = [&](int32_t val) { return MakeConstantScalar(DataType::Int(32), val); };
+  const int n = 8;
+  const int m = 30;
+  const int bits = 8;
+  const Expr x_p = Subtract(Add(x, RightShift(x, const_i32(1))), RightShift(x, const_i32(4)));
+  const Expr q = Divide(Negative(x_p), x_0);
+  const Expr r = Subtract(x_p, Multiply(q, Negative(x_0)));
+  const Expr x_b = Add(RightShift(r, const_i32(1)), x_0);
+  const Expr exps = RightShift(LeftShift(x_b, const_i32(n)), q);
```
masahi, line 113: Could `q` be bigger than `n`? If so, aren't we ending up doing a left shift by a negative number?

src/relay/qnn/op/softmax.cc
```diff
@@ -0,0 +1,144 @@
...
+      Round(Divide(MakeConstantScalar(DataType::Float(32), 1.f), input_scale)), DataType::Int(32));
+  const Expr max = Max(quantized_data, {axis}, true, false);
+  const Expr x = Subtract(quantized_data, max);
+
+  const auto const_i32 = [&](int32_t val) { return MakeConstantScalar(DataType::Int(32), val); };
+  const int n = 8;
+  const int m = 30;
+  const int bits = 8;
+  const Expr x_p = Subtract(Add(x, RightShift(x, const_i32(1))), RightShift(x, const_i32(4)));
+  const Expr q = Divide(Negative(x_p), x_0);
```
masahi, line 110: floor missing after divide?

tests/python/relay/test_pass_fake_quantization_to_integer.py
```diff
@@ -1114,5 +1114,36 @@ def test_fake_quantize_take():
...
+        .evaluate()(*args)
+        .numpy()
+    )
+    result_int = (
+        relay.create_executor("vm", mod=mod_int, device=tvm.cpu(), target="llvm")
+        .evaluate()(*args)
+        .numpy()
+    )
+
+    assert np.allclose(result_int, result, rtol=1, atol=0.05)
```
masahi, line 1145: I think `rtol = 1` is a bit too large... is this a reasonable choice or could there be accuracy issues in the arithmetic?

tests/python/relay/test_pass_fake_quantization_to_integer.py
```diff
@@ -1114,5 +1114,36 @@ def test_fake_quantize_take():
...
+def test_fake_quantize_softmax():
+    x = relay.var("x", shape=[5, 10], dtype="float32")
+
+    x = relay.qnn.op.quantize(x, relay.const(0.08), relay.const(-48), out_dtype="int8")
+    x = relay.qnn.op.dequantize(x, relay.const(0.08), relay.const(-48))
+    op = relay.op.nn.softmax(x, axis=1)
+    op = relay.qnn.op.quantize(op, relay.const(0.0038), relay.const(-128), out_dtype="int8")
+    op = relay.qnn.op.dequantize(op, relay.const(0.0038), relay.const(-128))
+
+    x_np = np.random.random_sample([5, 10]).astype(np.float32) * 20.0 - 6
```
masahi, line 1126: Please test on more inputs, ideally on sampling from all int8 / uint8 range. This choice looks very contrived.

src/relay/qnn/op/softmax.cc
masahi: From the paper, it's not clear to me why it is always safe to right-shift by q. I don't see an obvious bound on it.

masahi:
cc @AndrewZhaoLuo @ibsidorenko @Icemist please help review.

src/relay/qnn/op/softmax.cc
masahi: Actually I'm not sure their eq. 14 is correct for a positive exponent, since we cannot decompose `S * I` into `(-q)` + fraction where `q` is positive... thought?

src/relay/qnn/op/softmax.cc
maekawatoshiki: >  I don't quite understand why they negate here.

From my understanding, since `x_p` is 0. or less than 0., `Divide(Negative(x_p), x_0)` is to get positive values.

> floor missing after divide?

This is an integer arithmetic so that it would not need floor, I think.

tests/python/relay/test_pass_fake_quantization_to_integer.py
maekawatoshiki: What do you think about this: https://github.com/apache/tvm/pull/14536/commits/35b554875943de8fca899a3b86faae448f53ebf0#diff-3068125dc5426cffbc64c13858f102aa8f7eaee8565a6aaacb4badde71faf7b1

tests/python/relay/test_pass_fake_quantization_to_integer.py
```diff
@@ -1114,5 +1114,36 @@ def test_fake_quantize_take():
     compare_fq_to_int(op, [x_np])
 
 
+def test_fake_quantize_softmax():
```
ibsidorenko, line 1117: Looks like this test does not allow to check accuracy in full.

I have printed out output and found that ~70% of output values is equal to **0.0** in this test. This is because output after `qnn.quantize` operation is equal to **"-128"**. It is not very interesting/representative case for **"int8"** data type.

Can you slightly modify this test in the following way:

1. Remove second `qnn.dequantize`. Let's check output of `qnn.dequantize` + `softmax` + `qnn.quantize` only
2. Play with QNN parameters (zero point, scale) in a such way that output from quantize will be in the range [-100, +100] fo r example. Not only **"-128"** like now

P.S.
I have checked output after `qnn.quantize` and see that some of value have diff by **7**. I think it is too much and the accuracy is unsatisfactory... any thoughts?

ibsidorenko (changes requested):

ibsidorenko:
Hi, @maekawatoshiki ! Thank you for your PR, relay very interesting work! 
But I have my doubts about accuracy. Can you check my comment for unit test.

src/relay/qnn/op/softmax.cc
masahi: > since x_p is 0. or less than 0

Why is that? It's `q * log_2_e` right?

tests/python/relay/test_pass_fake_quantization_to_integer.py
masahi: Yes it looks good. We should also test on uint8. A new feature involving subtle arithmetic like this should be tested very thoroughly. See also the remark by @ibsidorenko

src/relay/qnn/op/softmax.cc
```diff
@@ -0,0 +1,144 @@
...
+  const Expr output_scale = new_args[3];
+  const Expr output_zero_point = new_args[4];
+  const int axis = attrs.as<SoftmaxAttrs>()->axis;
+
+  // Refer to the Algorithm 1 in https://arxiv.org/pdf/2207.01405.pdf
+
+  const Expr quantized_data =
+      Subtract(Cast(new_args[0], DataType::Int(32)), Cast(input_zero_point, DataType::Int(32)));
+
+  const Expr x_0 = ConvertDtype(
```
AndrewZhaoLuo, line 100: image

This is in `ShiftExp` under `algorithm 1`

First off, if S > 1 then we already have issues as `x_0` (`I_0` in paper) maybe 0... 

For attention I would expect output activation range to potentially be very large (see LLM int8 paper) so having a high scale factor is not unreasonable for some schemes.

AndrewZhaoLuo: Maybe I am not understanding something but seems like an obvious flaw...

I think you can get around this by not rounding I_0 in `algorithm1` but keeping it a float and rounding when needed. However this would introduce runtime FLOPS.

tests/python/relay/test_pass_fake_quantization_to_integer.py
```diff
@@ -1114,5 +1114,36 @@ def test_fake_quantize_take():
...
+    shape = [50, 10]
+    x = relay.var("x", shape=shape, dtype="int8")
+
+    x = relay.qnn.op.dequantize(x, relay.const(0.08), relay.const(-48))
```
AndrewZhaoLuo, line 1121: Things we probably want to test:

1. Different dtypes
2. Different scale factors
3. Different distributions along the axis of reduction (e.g. flat distribution should give flat probabilities, muiltiple spikes, etc.)

src/relay/qnn/op/softmax.cc
AndrewZhaoLuo: Another option is when decomposing `2^(S * I_p)` into integer and decimal components, you instead decompose:

`2^[(S * 2^-n * I_p) * 2^n]`. At runtime we can choose an `n` to make `S * 2^-n << 1` to get around this problem. You can then decompose the internal terms in parentheses and the outer `2^n` now merely becomes another shift.

AndrewZhaoLuo (commented):
Very interesting PR. I agree with folks that if this is the default implementation, it better be fairly accurate across a wide range of scales and inputs. While the paper you reference targets ViTs, we cannot assume only ViTs will use this implementation.

The paper itself is kind of poor when it comes to consistent notation but I think I've deciphered the intention.

Unfortunately, I have doubts on the base-line implementation discussed in the paper on the grounds of accuracy, much like many of the others have commented. 

Furthermore, I believe a main issue with the algorithm is it can only deal with small input scale factors. I think this is probably a tenable problem but requires more work and rigorous testing.

src/relay/qnn/op/softmax.cc
masahi: Yeah but the point is to decompose in a way that the integer part is negative, so that we can "divide" by this integer. I don't see how such decomposition is possible if the input is positive.

AndrewZhaoLuo: Hmm, I believe it is due to the fact that the integers are all normalized by subtracting the maximum value. Therefore all values are either <= 0

src/relay/qnn/op/softmax.cc
maekawatoshiki: `x_p` is the same as `I_p` in the Algorithm 1 from the paper.
And here: https://github.com/apache/tvm/blob/35b554875943de8fca899a3b86faae448f53ebf0/src/relay/qnn/op/softmax.cc#L102-L103 we subtract the max value in the axis.

src/relay/qnn/op/softmax.cc
masahi: Isn't S < 0 by definition? Since it should be 1.0 / max(calib_data)

AndrewZhaoLuo: I believe S here is defined as:

image

masahi: We can fallback to the fp32-based impl when S > 1.

AndrewZhaoLuo: Yeah I think that is another reasonable thing to do.

src/relay/qnn/op/softmax.cc
masahi: Ok. But I'm still not sure about my first question https://github.com/apache/tvm/pull/14536#discussion_r1161619325.

tests/python/relay/test_pass_fake_quantization_to_integer.py
ibsidorenko: Yes, sure, it's up to you. My main concern here is to avoid the case when all (or almost all) output values are equal to "-128" (it is not representative case for "int8" data type.)

src/relay/qnn/op/softmax.cc
AndrewZhaoLuo: Ah yes I think it can which is bad. I think the way around this might be exploring decomposing `2^[(S * 2^-k * I_p) * 2^k]` instead of `2^(S * I_p)` for some well chosen k.

src/relay/qnn/op/softmax.cc
```diff
@@ -0,0 +1,144 @@
...
+  const Expr input_scale = new_args[1];
+  const Expr input_zero_point = new_args[2];
+  const Expr output_scale = new_args[3];
+  const Expr output_zero_point = new_args[4];
+  const int axis = attrs.as<SoftmaxAttrs>()->axis;
+
+  // Refer to the Algorithm 1 in https://arxiv.org/pdf/2207.01405.pdf
+
+  const Expr quantized_data =
+      Subtract(Cast(new_args[0], DataType::Int(32)), Cast(input_zero_point, DataType::Int(32)));
```
Aleksei-grovety, line 98: Is it necessary to cast input_zero_point to int32? It is assumed that input_zero_point is of type int32.

maekawatoshiki: This is apparently a redundant cast. I should remove it.

maekawatoshiki:
I took a look at the latest tensorflow implementation and noticed it is, for some reason, using float arithmetic: https://github.com/tensorflow/tensorflow/blob/98a187a8bfcdcf0c55c16f07b4a06b50e06a9a26/tensorflow/lite/kernels/internal/optimized/optimized_ops.h#L3471-L3477.

I also found another paper proposing quantized softmax (https://arxiv.org/pdf/2101.01321.pdf) and its implementation (https://github.com/kssteven418/I-BERT/blob/1b09c759d6aeb71312df9c6ef74fa268a87c934e/fairseq/quantization/utils/quant_modules.py#L578).
However, unlike the proposed algorithm in the paper, its implementation looks like using float arithmetic in several places of code.

I realized that it's difficult to implement integer-only quantized softmax with satisfying quality in a variety of input scales.

Let me abandon this PR for now to investigate further, and hopefully, I'll make another PR.

AndrewZhaoLuo:
Yeah it is challenging problem. I would not necessarily constrain yourself to making a single implementation that can handle all cases however. 

Feel free to plan for several future specialization with fallbacks if needed imo. Just make sure tests reflect this.

Though, will leave up to you.

shinh:
Does this make sense to add this `qnn.softmax` implementation as an optional feature? By default, we wouldn't enable qnn.softmax, but users could activate it when they find its precision satisfactory for their use case. To be more specific, I propose the following:

1. Add `@register_optional_fake_quantization_to_integer` and use it in fake_quantization_to_integer.py for `softmax`:

```
@register_optional_fake_quantization_to_integer("nn.softmax")
def softmax(expr, type_map):
  ...
```

2. Modify `fake_quantization_to_integer.cc` so that optional rewriters will be ignored unless users explicitly state they want to use quantized softmax by something like

```
relay.transform.FakeQuantizationToInteger(optional_qnn_ops={"nn.softmax"})(mod)
```

I guess it's OK to relax checks of unittests if this feature is optional? What are your thoughts?

masahi:
@shinh Sounds good to me, as long as things won't break by default, I don't see any problem.

src/relay/qnn/op/softmax.cc
```diff
@@ -0,0 +1,148 @@
...
+  ICHECK_EQ(new_args.size(), 5);
+
+  const auto const_i32 = [&](int32_t val) { return MakeConstantScalar(DataType::Int(32), val); };
+  const auto const_f32 = [&](float val) { return MakeConstantScalar(DataType::Float(32), val); };
+
+  const auto const_input_scale = new_args[1].as<ConstantNode>();
+  ICHECK(const_input_scale) << "Input scale should be constant.";
+  ICHECK(const_input_scale->is_scalar()) << "Input scale should be scalar.";
+  const float input_scale = static_cast<float*>(const_input_scale->data->data)[0];
+  ICHECK(input_scale <= 1.f) << "Input scale should be less than or equal to 1.";
```
maekawatoshiki, line 96: The assertion fails when the input scale does not meet the condition.

tests/python/relay/test_pass_fake_quantization_to_integer.py
```diff
@@ -1114,5 +1114,53 @@ def test_fake_quantize_take():
...
+
+        x_np = np.random.randint(-128, 127, size=shape, dtype="int8")
+        x_np = np.sort(x_np)
+        args = [x_np]
+
+        mod = tvm.IRModule.from_expr(op)
+        mod = tvm.relay.transform.InferType()(mod)
+        mod_int = tvm.relay.transform.FakeQuantizationToInteger(
+            hard_fail=True, optional_qnn_ops=["nn.softmax"]
+        )(mod)
```
maekawatoshiki, line 1138: The pass does not use the fully-integer implementation for `nn.softmax`, unless it is specified as `optional_qnn_ops`.

tests/python/relay/test_pass_fake_quantization_to_integer.py
```diff
@@ -1114,5 +1114,53 @@ def test_fake_quantize_take():
...
+        # since it is difficult to use allclose due to not-so-good accuracy.
+        for qdq, qop in zip(result, result_int):
+            assert is_sorted(qdq)
+            assert is_sorted(qop)
+
+        try:
+            np.testing.assert_allclose(result_int, result, atol=1)
+        except AssertionError as e:
+            # To see the difference
+            print(e)
```
maekawatoshiki, line 1162: What do you think about this?
I checked the max absolute difference here is, in most cases, 0~6.
Also the overall trend of the softmax output didn't differ much between the QOp and QDQ implementation.

maekawatoshiki (commented):
I re-implemented the QOp softmax as an optional feature.
I'd appreciate your advice.

src/relay/qnn/op/softmax.cc
```diff
@@ -0,0 +1,148 @@
...
+  const Expr x = Subtract(quantized_data, max);
+
+  const int m = 30;
+  const int bits = 8;
+  const Expr x_p = Subtract(Add(x, RightShift(x, const_i32(1))), RightShift(x, const_i32(4)));
+  const Expr q = Clip(Divide(x_p, Negative(x_0)), 0, 20);
+  const Expr max_q = Max(q, {axis}, true, false);
+  const Expr r = Subtract(x_p, Multiply(q, Negative(x_0)));
+  const Expr x_b = Add(RightShift(r, const_i32(1)), x_0);
+  const Expr exps = LeftShift(x_b, Subtract(max_q, q));
```
masahi, line 118: If we are left-shifting by `max_q`, shouldn't we update the output scale to divide by 2^(max_q - 1) as well?

src/relay/qnn/op/softmax.cc
```diff
@@ -0,0 +1,148 @@
...
+Expr MakeQuantizedSoftmax(Expr x, int axis, Expr scale, Expr zero_point, Expr output_scale,
+                          Expr output_zero_point) {
+  auto attrs = make_object<SoftmaxAttrs>();
+  attrs->axis = axis;
+  static const Op& op = Op::Get("qnn.softmax");
+  return Call(op, {x, scale, zero_point, output_scale, output_zero_point}, Attrs(attrs), {});
+}
+
+/*
+ * \brief Canonicalizes the QNN softmax op.
```
masahi, line 82: Please document that this implementation is highly experimental.

tests/python/relay/test_pass_fake_quantization_to_integer.py
```diff
@@ -1114,5 +1114,53 @@ def test_fake_quantize_take():
...
+            relay.create_executor("vm", mod=mod_int, device=tvm.cpu(), target="llvm")
+            .evaluate()(*args)
+            .numpy()
+        )
+
+        # Check at least the softmax output is in ascending order,
+        # since it is difficult to use allclose due to not-so-good accuracy.
+        for qdq, qop in zip(result, result_int):
+            assert is_sorted(qdq)
+            assert is_sorted(qop)
```
masahi, line 1156: Why should they be sorted?

tests/python/relay/test_pass_fake_quantization_to_integer.py
masahi: As long as this implementation is useful for your use case, I'm fine with this. cc @ibsidorenko @AndrewZhaoLuo

src/relay/qnn/op/softmax.cc
maekawatoshiki: Added comment

src/relay/qnn/op/softmax.cc
masahi: I see, I guess we don't have to compute `S_exp` because the scale factors cancel in the denominator / numerator in `I_out`?

maekawatoshiki: Yes, I believe so.

masahi (approved):

PR #8165, opened 2021-05-31
[Relay, TOPI] Refactor strided_slice and add axes argument

masahi:
This PR has two purposes:
1. Add `axes` argument to static param `strided_slice`
2. Refactor `strided_slice` related code so that adding `axes` argument would not increase clutter. Also I did further refactoring to remove duplicated output shape calculation between topi/relay type rel (left as [TODO](https://github.com/apache/tvm/blob/main/include/tvm/topi/transform.h#L641-L642)) and minimize the number of `te::compute` for strided slice (currently there are 4 of them).

For 1, currently slicing along the **second** axis with input shape `(?, 3)`, begin `[0]`, end `[1]` calls the dynamic parameter variant of relay/topi `strided_slice` with `begin = (0, 0)` and `end = (?, ?)`, which results in the output shape `(?, ?)` rather than desired `(?, 1)`. This causes a problem when the output is fed to `gather_nd` which requires a certain dimension to be static. 

The root problem is that our `strided_slice` op does not have a notion of `axes` and always tries to slice along all axes. So when some of the dimensions are dynamic, we call dynamic parameter variant of `strided_slice` which makes all input params and the output shape dynamic, even if we only need to slice along static dimensions with static begin/end. My solution adds `axes` argument to static parameter variant of `strided_slice`so that slicing would not touch dynamic dimensions and preserve static dimensions along provided `axes`. 

For 2, we accumulated many patches for `strided_slice` over the last few years and I found the current implementation extremely messy and hard to maintain. In order not to make the situation worse by introducing `axes` argument, I decided to roll major refactoring of our `strided_slice` related code.

First, the output shape calculation logic is duplicated between topi and relay type rel:
* https://github.com/apache/tvm/blob/main/include/tvm/topi/transform.h#L651-L714
* https://github.com/apache/tvm/blob/main/src/relay/op/tensor/transform.cc#L2451-L2534

This has been fixed by introducing [StridedSliceOutputShape function](https://github.com/apache/tvm/blob/64c1bbf27607de08620c942d41592bae71c0a667/include/tvm/topi/transform.h#L655) in topi and use it from both topi/relay. 

Second, currently there are 4 `te::compute` for `strided_slice`, each with slightly different expectation on its input and input param canonicalization. 
* https://github.com/apache/tvm/blob/main/include/tvm/topi/transform.h#L716-L726
* https://github.com/apache/tvm/blob/main/include/tvm/topi/transform.h#L628-L636
[...]

PR #14548, opened 2023-04-09
[Unity][Op] Dynamic Strided Slice

sunggg:
This PR brings dynamic strided slice, which will be the first stab for data-dependent ops in Unity. 

### Overview 
It consists of three parts (Op, TOPI, Legalization) and their test cases. 

Data-dependent ops, like dynamic strided slice, could be tricky when we cannot automatically deduce their output shape. 
In such cases, we cannot lower them since TE infra requires a concrete output shape, which should be defined with symbolic variables at least. Therefore, manual shape function registration is inevitable for those operators to let the compiler know how to compute their output shapes. 

With this PR, users can register the shape func in TOPI and insert it with match cast mechanism during the legalization. 
It's worth noting that for data-dependent ops, current TOPI creates symbolic variables whenever the shape value of certain dimension is unknown, then later mechanism handles the binding and so on. (see [link](https://github.com/apache/tvm/blob/unity/include/tvm/topi/transform.h#L672))
However, with this PR, the legalizer would be in charge of creating symbolic variables and binding them explicitly, and then pass the output shape to TOPI so that TOPI can simply use it to define its compute. 
```Python
@register_legalize("relax.dynamic_strided_slice")
def _dynamic_strided_slice(bb: BlockBuilder, call: Call) -> Expr:
    # 1. Insert shape function
    output_shape = bb.normalize(
        bb.call_te(
            topi.shape_func_dynamic_strided_slice,
            call.args[0],
            call.args[1],
            call.args[2],
            call.args[3],
        )
    )
    # 2. Convert tensor to shape and match cast with new symbolic vars
    # Get shape length
    ndim = int(output_shape.struct_info.shape[0])
    output_shape = bb.emit(
        Call(
            ExternFunc("vm.builtin.tensor_to_shape"),
            [output_shape],
            sinfo_args=[ShapeStructInfo(ndim=ndim)],
        )
    )
    output_shape_vars = [tir.Var("s", "int64") for i in range(ndim)]
    bb.match_cast(output_shape, ShapeStructInfo(output_shape_vars))

   # 3. Pass the output shape vars to TOPI
    return bb.call_te(
        topi.dynamic_strided_slice,
        call.args[0],
        call.args[1],
        call.args[2],
        call.args[3],
        output_shape=output_shape_vars,
    )
```
Through the internal discussion with @junrushao, we confirmed that this should comply with WIP PR https://github.com/apache/tvm/pull/14278. 
Also, since this requires the change in existing `topi::dynamic_strided_slice` (see [link](https://github.com/apache/tvm/blob/unity/include/tvm/topi/transform.h#L709)),  this PR creates relax namespace and implements its own version. 

### Notes
* Currently, in relax, we already have `relax.strided_slice` op that covers non-data-dependent scenarios. This is still useful since it can cover the current limitation of `relax.dynamic_strided_slice` op: its shape analysis may not be informative enough for its users like `tensor_to_shape`, which requires the known integer shape at compile-time.  
We will revisit their unification when we have a better understanding in the future. 
* This PR adapts the Relay's `topi::dynamic_strided_slice` to relax standards without fixing its current limitation. Therefore, it expects to perform preprocessing for `begin/end/strides` tensors to make them have the equal length with the dimension of `data` tensor.  PR https://github.com/apache/tvm/pull/14493 would be helpful for such pre-processing. 

### Discussion
* Currently, I named it "dynamic" strided slice only for simplicity and following the tradition. But, strictly speaking, data-dependent strided slice is more accurate since `relax.strided_slice` can already cover shape dynamism for `data` tensor. Any suggestion would be very helpful. 
* To maintain the relax version of TOPI, I created the relax namespace for now. Would there be a better way? 

cc. @yongwww @jwfromm @psrivas2 @slyubomirsky @junrushao @tqchen @MasterJH5574 @masahi


include/tvm/topi/transform.h
```diff
@@ -2035,6 +2034,73 @@ inline Tensor adv_index(const Tensor& data, const Array<Tensor>& indices,
...
+          PrimExpr len1 = ceildiv(begin - end, -stride);
+          PrimExpr len2 = ceildiv(end - begin, stride);
+          return if_then_else(stride < 0, len1, len2);
+        };
+        PrimExpr length(-1);
+        int ndim = data.ndim();
+        for (int i = 0; i < ndim; i++) {
+          length = if_then_else(indices[0] == i, data->shape[i], length);
+        }
+        return GetLength(begin(indices), end(indices), strides(indices), length);
```
masahi, line 2097: > Through the internal discussion with @junrushao, we confirmed that this should comply with WIP PR https://github.com/apache/tvm/pull/14278.

Not sure what you meant by above. Is the whole point being able to avoid repeated compute definition like this?

src/relax/op/tensor/index.cc
```diff
@@ -239,5 +239,78 @@ TVM_REGISTER_OP("relax.strided_slice")
...
+          << "values to be all int64. However, " << name << " has dtype " << sinfo->dtype << ".";
+    }
+  };
+  diag_def(begin_sinfo, "begin");
+  diag_def(end_sinfo, "end");
+  diag_def(strides_sinfo, "stride");
+
+  // The output shape will depend on the runtime value in begin/end/stride tensors.
+  // TODO(tvm-team): Extract more compile-time info when those tensors are constants.
+  return TensorStructInfo(data_sinfo->dtype, n_axis);
```
masahi, line 303: I'm a bit concerned about this. It seems we can only express slicing with either "all static" or "all dynamic" axes. But partially static / dynamic slicing is very common in practice (e.g., slicing only along the dynamic "batch" (the number of detected boxes) dimension object detection models). If shape func needs to be implemented via topi, I don't see an way to express such partially-static output shape.

For Relay I added a hacky workaround https://github.com/apache/tvm/pull/8165 for this issue.

python/tvm/relax/transform/legalize_ops/index.py
```diff
@@ -59,3 +60,39 @@ def _strided_slice(bb: BlockBuilder, call: Call) -> Expr:
...
         slice_mode="end",
     )
+
+
+@register_legalize("relax.dynamic_strided_slice")
+def _dynamic_strided_slice(bb: BlockBuilder, call: Call) -> Expr:
+    # 1. Insert shape function
+    output_shape = bb.normalize(
+        bb.call_te(
+            topi.shape_func_dynamic_strided_slice,
```
masahi, line 70: Could there be an alternative way to implement shape funcs? Because te tensor cannot express "constant"-ness.

python/tvm/relax/transform/legalize_ops/index.py
```diff
@@ -59,3 +60,39 @@ def _strided_slice(bb: BlockBuilder, call: Call) -> Expr:
...
+            call.args[0],
+            call.args[1],
+            call.args[2],
+            call.args[3],
```
masahi, line 74: Two questions:

* Why `shape_func_dynamic_strided_slice` needs to be implemented in topi / c++? This seems like a one-off function, why not just create `te.compute` here in python?
* Are we introducing a standard API for shape func (like Relay), or is each op going to pick any API as it sees fit (like this example)? In the latter case I don't see any reason shape funcs need to belong to topi.

python/tvm/relax/transform/legalize_ops/index.py
sunggg: I considered two options: TE and TIR, and chose TE since it is easier to write. 
I did not consider the hybrid script since I did not want users to learn new thing only for the shape function. 
Would hybrid script help in this case? Also, I'm open for other suggestions as well.

python/tvm/relax/transform/legalize_ops/index.py
sunggg: > Are we introducing a standard API for shape func (like Relay), or is each op going to pick any API as it sees fit (like this example)? In the latter case I don't see any reason shape funcs need to belong to topi.

The latter case. 

Great point. I put it in TOPI-side because I viewed TOPI as the registry of TEs and developers can easily check related implementations in the single file. I don't have any strong preference here, so I'm open to hear what folks prefer.

src/relax/op/tensor/index.cc
sunggg: Yes, that is current limitation of this PR so I left it as TODO. 

Actually, I have a question regarding https://github.com/apache/tvm/pull/8165. Can `te::Tensor` contain symbolic variables? 
Please correct me if I'm wrong, based on my current understanding, data within `te::Tensor` will be certain values rather than symbolic variables. So I cannot see how partially static output shape will be represented if we go through F1->F2 route below. 
```C++
// F1
inline te::Tensor dynamic_strided_slice(const te::Tensor& x, const te::Tensor& begin,
                                        const te::Tensor& end, const te::Tensor& strides, ...)
{
     // ...
     Array<PrimExpr> begin_expr, end_expr, strides_expr;
     for (int64_t i = 0; i < num_dynamic_axes; ++i) {
         auto ind = make_const(index_dtype, i);
         begin_expr.push_back(begin(ind));    // <- Can `begin(ind)` be symbolic in practice?
         end_expr.push_back(end(ind));
         strides_expr.push_back(strides(ind));
      } 
      // Call F2
      return dynamic_strided_slice(x, begin_expr, end_expr, strides_expr, name, tag);
}

// F2
inline Tensor dynamic_strided_slice(const Tensor& x, const Array<PrimExpr>& begin,
                                    const Array<PrimExpr>& end, const Array<PrimExpr>& strides, ...)
```

masahi: I think `te::Tensor` can be created from symbolic values, but we cannot extract such values from the tensor.

python/tvm/relax/transform/legalize_ops/index.py
masahi: I was thinking about TVMScript. Why not implement shape func in TVMScript, which should be able to represent partially-static shape as well?

include/tvm/topi/transform.h
masahi: I think the question is whether or not we consider shape funcs a part of "op definition". This is the case for Relay, and one of the reason adding a new op in Relay is complicated. I don't know what the goal of #14278 is, but the current discussion in this PR suggests shape funcs need not implemented under topi. It seems more like "implementation details" of the legalizer.

 So unless #14278 is aiming to generate an op legalizer definition, I think this work is completely decoupled from #14278.

python/tvm/relax/transform/legalize_ops/index.py
masahi: Since there is no standard API for shape func, and there is no "registration" for shape funcs specifically, I think it is better not to introduce a rigid convention, in particular shape funcs be implemented in topi / c++. 

Since they are only used by the legalizer, I view shape funcs in Relax simply as implementation details, one-off function. So developers should be able to express data-dependent output shape computation logic in any way in anywhere, including TVMScript.

include/tvm/topi/transform.h
```diff
@@ -2035,6 +2034,73 @@ inline Tensor adv_index(const Tensor& data, const Array<Tensor>& indices,
       name, tag);
 }
 
+namespace relax {
+// relax dynamic slice
+inline te::Tensor dynamic_strided_slice(const te::Tensor& x, const te::Tensor& begin,
+                                        const te::Tensor& end, const te::Tensor& strides,
+                                        Array<PrimExpr> output_shape,
+                                        std::string name = "T_strided_slice_dynamic",
+                                        std::string tag = kInjective) {
```
masahi, line 2043: Why can't we use the existing definition of `dynamic_strided_slice`? https://github.com/apache/tvm/blob/4e07a8ed6687a08b6b27db21af019a5a179b9ee1/include/tvm/topi/transform.h#L648-L651

masahi: sorry maybe I'm confused about the existing implementation of `dynamic_strided_slice` and its relation with `strided_slice` shape func. If `dynamic_strided_slice` can calculate the right output shape from its dynamic input, why do we need a shape func for it at all...

include/tvm/topi/transform.h
sunggg: The one of the main goals of https://github.com/apache/tvm/pull/14278 is to automate the current repetitive & tedious op registration process including legalization, struct_info, etc. Unlike current flow which requires us to look at multiple different sites and do the manual job, https://github.com/apache/tvm/pull/14278 will allow us to look at the single place to put all those information so that the parser will handle the rest. I believe legalization function should be there, too. 

> I think the question is whether or not we consider shape funcs a part of "op definition". 

You brought up the very good points. I see that shape func is a part of op definition just like legalization function for each operator. When https://github.com/apache/tvm/pull/14278 lands, shape function can be registered together only if necessary so that we don't complicate the op registration unnecessarily. 

>  but the current discussion in this PR suggests shape funcs need not be implemented under topi.

I forgot to mention earlier, but https://github.com/apache/tvm/pull/14278 deduce output shape of operator by using the existing shape computation logic in TOPI to automatically generate the struct_info logics. Not sure if this is the hard requirement tho. @junrushao, would it be okay to put the shape computation in elsewhere, for example legalizer?

src/relax/op/tensor/index.cc
sunggg: Interesting. Is there any example code that I can try? 
Also, if we can create the `te::Tensor` with symbolic variables, may I ask why we cannot extract those values? Can't we access the value by using the index in the above example?

python/tvm/relax/transform/legalize_ops/index.py
sunggg: Correct me if I'm wrong, but don't we need `Array<PrimExpr>` to represent the partially-static shape? If this is true, I don't think this is currently supported in TVMScript. AFAIK, we usually carry those forms around within op attribute. 

Also, if we go with TVMScript, do we require users to write the TIR primfunc by hand?

src/relax/op/tensor/index.cc
masahi: Assuming "symbolic variable" you mentioned is just another `PrimExpr`, I expect we can fill in a tensor by that value, just like any other `PrimExpr`. 

> Can't we access the value by using the index in the above example?
What I meant was, as soon as we put a symbolic expression in a TE tensor, we lost all symbolic-specific information. You can index it, but what you get is an opaque value. So we cannot exploit any symbolic information about it. 

So output shape represented by TE Tensor loses symbolic or constant shape information. That's why I think it is better to directly generate shape func via TIR.

python/tvm/relax/transform/legalize_ops/index.py
sunggg: I think this is very good point. we can continue this discussion here: https://github.com/apache/tvm/pull/14548#discussion_r1162233085

include/tvm/topi/transform.h
sunggg: No worries! :)

python/tvm/relax/transform/legalize_ops/index.py
masahi: >  If this is true, I don't think this is currently supported in TVMScript

hmm interesting. Don't we create TVMScript with shapes like (`n`, 16) all the time? I hope the output of a shape func can also be expressed like that. 

> Also, if we go with TVMScript, do we require users to write the TIR primfunc by hand?

Yes but writing in TVMScript is only required if a user wants to exploit "partially-static" shape. I think this is a relatively rare and advanced use case. `te.compute`-based definition should always be possible.

And other than dynamic slicing, I cannot think of other ops that may want to implement their shape func in TVMScript.  Maybe `reshape` and other shape-changing ops might apply but I haven't met a real-world use case.

include/tvm/topi/transform.h
masahi: so do you have an answer to my original question? Why can't we simply create the output shape by

```
  Array<PrimExpr> out_shape;

  for (size_t i = 0; i < num_slice_axes; ++i) {
    auto d = indexdiv(end[i] - begin[i], strides[i]);
    if (d->IsInstance<tvm::IntImmNode>()) {
      // Preserve static dimension if possible
      out_shape.push_back(d);
    } else {
      out_shape.push_back(tvm::tir::Var("dim"));
    }
  }
```
?

Sorry if I'm missing something obvious.

include/tvm/topi/transform.h
masahi: > The one of the main goals of https://github.com/apache/tvm/pull/14278 is to automate the current repetitive & tedious op registration process including legalization

ok, but you are introducing `shape_func_dynamic_strided_slice` which looks like a repetitive definition of output shape logic to me. Are you saying that https://github.com/apache/tvm/pull/14278 is going to remove that definition? Otherwise I still don't get why you meant when you say this PR "complies" with #14278.

python/tvm/relax/transform/legalize_ops/index.py
yongwww: I agree that "partially-static" cases are common. However, writing TIR manually might make it more challenging for contributors to develop data-dependent ops. Simplifying the shape function implementation could be beneficial. I think several ops like reshape or max_pool2d could be used as partially-dynamic, such as allowing dynamic or partially-dynamic strides in max_pool even I haven't encountered the real-world use cases.

sunggg: @masahi Yeah, I believe that is the shape field in struct info. In this case, we need to pass `Array<PrimExpr>` as an argument of `call_tir` for the shape func.

include/tvm/topi/transform.h
sunggg: My apology. It seems like I misread your comment. 

The problem was how to insert match cast logic and bind those variables. I couldn't see a way to insert them in TOPI side.  So I did it in the legalizer and it became explicit and straightforward. 
https://github.com/apache/tvm/blob/4901e3db16673ee48776cde3bff2024efc2d3a39/python/tvm/relax/transform/legalize_ops/index.py#L77-L88

masahi: This, again, sounds odd to me to talk about legalizer vs topi distinction. Since `dynamic_strided_slice` and the shape func is only meant to be used by the legalizer, why not just define them here? There is no need to be concerned about "TOPI side".

src/relax/op/tensor/index.cc
sunggg: Based on our discussion today, decided to revisit when https://github.com/apache/tvm/pull/14278 lands. Left the TODO comment.

python/tvm/relax/transform/legalize_ops/index.py
sunggg: Based on our discussion today, decided to revisit when https://github.com/apache/tvm/pull/14278 lands. Left the TODO comment.

sunggg:
@masahi @yongwww reflected your feedback. would you take another look?

masahi (approved):

masahi:
@sunggg `unity` CI has been broken after merging this PR:
https://github.com/apache/tvm/commits/unity
https://ci.tlcpack.ai/blue/organizations/jenkins/tvm-unity/detail/PR-14612/1/pipeline 

Can you send a follow-up PR?

 cc @tqchen

tqchen:
I also noticed this and disabled the tests temporarily in my unity merge, @sunggg please followup

Hzfengsy:
Hi @sunggg, I noticed this PR failed the tests of `test_relax_dynamic_strided_slice`, saying
```
AttributeError: module 'tvm.topi' has no attribute 'shape_func_dynamic_strided_slice'
```

you can reproduce it with the latest unity branch. 
```
pytest tests/python/topi/test_topi_transform.py::test_relax_dynamic_strided_slice
``` 

Could you please fix it if you have time? 

PS: The other failure tests in the same file are not related to this PR, and I've fixed it in the [branch]( https://github.com/apache/tvm/tree/test_all_cases_on_unity)

PR #14551, opened 2023-04-10
[TVMScript] Int64 as default

cyx-6:
work in progress


FrozenGene:
does us have one rfc or specific reason why we choose i64 as default?

junrushao:
@FrozenGene I don't think we are going to merge this PR. This is used to test the feasibility.

PR #14552, opened 2023-04-10
[Unity] Fix emit_te with symbolic input

tqchen:
This PR fixes the use of emit_te with symbolic shape input.

Testcases are added


yongwww (approved):
Thanks for the fix!!!

jinhongyii (approved):

PR #14553, opened 2023-04-10
[Unity] Make FuseOps work on a call_tir which has a ShapeExpr arg

jinhongyii:
the pattern of the shapeexpr should be set to `kOpaque` in  GraphCreator. Previously it was not handled and will run into an error


jinhongyii:
cc: @tqchen @Hzfengsy

Hzfengsy:

jinhongyii:
test added

Hzfengsy (approved):

PR #14555, opened 2023-04-10
[Unity] Preserve symbolic var args when applying call_tir

jinhongyii:
Currently fuse_tir will remove the symbolic var args of call_tir. This PR fixes this behavior.


jinhongyii:
cc: @tqchen @Hzfengsy

Hzfengsy (approved):

PR #14556, opened 2023-04-10
[Bugfix] [Relay] fix a bug caused by IncompleteTypeNode in EinsumRel while doing MergeComposite

kfeng123:
Previously, it was reported that when using MergeComposite on a model with broadcast_to OP, an error caused by IncompleteTypeNode occurs during compiling model. See https://discuss.tvm.apache.org/t/incomplete-type-in-broadcasttorel-while-doing-mergecomposite/10439

A similar bug also occurs if the model contains einsum OP. This PR fixes this bug for einsum OP and add a test case.

### DIFF
--- a/src/relay/op/tensor/math.cc
+++ b/src/relay/op/tensor/math.cc
@@ -59,6 +59,10 @@ bool EinsumRel(const Array<Type>& types, int num_inputs, const Attrs& attrs,
   }
 
   // Check the input tuple consists of tensors with consistent dtype.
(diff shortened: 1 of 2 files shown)


kfeng123:
cc @comaniac

comaniac:
LGTM. Can you add a unit test?

kfeng123:
> LGTM. Can you add a unit test?

Hi, I have added a test case of MergeComposite: the old code crashed and the updated code works correctly.

comaniac (approved):
LGTM

PR #14559, opened 2023-04-10
[Unity][Op] Symbolic shape support of take grad

SiriusNEO:
Prior to this PR the legalization of `R.grad.take_backward` doesn't support symbolic shape. This PR enables the support by some small fixes and adds tests.


python/tvm/relax/transform/legalize_ops/grad.py
```diff
@@ -162,7 +162,7 @@ def gen_ir(output_grad_ptr, x_ptr, indices_ptr, out_ptr):
             with ib.for_range(0, fused_shape) as i:
                 out[i] = tir.const(0, dtype=x_ptr.dtype)
 
-            indices_len = indices_ptr.shape[0].value  # must be 1-dim
+            indices_len = indices_ptr.shape[0]  # must be 1-dim
```
yongwww, line 165: consider adding an assert for the "must be..."

tqchen (approved):

PR #14564, opened 2023-04-10
[LLVM] Validate generated LLVM module before optimization

Lunderberg:
Because LLVM's optimizations assume that the generated module is valid, validation should be done before optimization, rather than after.  This has the additional benefit of providing error messages from LLVM that more closely relate to the TVM-generated LLVM IR, rather than the optimized LLVM IR.


kparzysz-quic (approved):

Lunderberg:
@tvm-bot rerun

PR #14567, opened 2023-04-10
[TIR][Hexagon] Use the "target" value in T.func_attr for VTCM limit

Lunderberg:
For the VerifyVTCMLimit, read directly from the function attribute, if the function has already been annotated with the target.


masahi (approved):

masahi:
@Lunderberg Need to resolve the conflict.

PR #14568, opened 2023-04-10
[LLVM] Expand tvm::Type to DWARF conversion

Lunderberg:
Prior to this commit, DWARF debug symbols for `float32`, `int8`, and `int32` could be generated, with other datatypes resulting in an error.  This commit expands the range of types with debug symbols to allow for any `DLDataTypeCode::kDLInt`, `kDLUint`, or `kDLFloat`, regardless of the bitsize.

### DIFF
--- a/src/target/llvm/codegen_cpu.cc
+++ b/src/target/llvm/codegen_cpu.cc
@@ -55,6 +55,7 @@
 
 #include <algorithm>
 #include <memory>
+#include <tuple>
 #include <unordered_map>
 #include <unordered_set>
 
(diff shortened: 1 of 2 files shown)


src/target/llvm/codegen_cpu.cc
```diff
@@ -300,6 +296,36 @@ llvm::DIType* CodeGenCPU::GetDebugType(const Type& ty_tir, llvm::Type* ty_llvm)
...
+
+    std::stringstream ss;
+    ss << name;
+
+    if (!dtype.is_bool()) {
+      ss << dtype.bits();
+    }
+    if (dtype.lanes() > 1) {
+      ss << "x" << dtype.lanes();
+    }
```
kparzysz-quic, line 324: There is already a function that gets the string from `DataType` [here](https://github.com/apache/tvm/blob/51dcafb4abc7dbc3583c1719d3c483ae52d67859/include/tvm/runtime/data_type.h#L316).

Please add `#include <sstream>` for `std::stringstream` if you decide to keep it.

src/target/llvm/codegen_cpu.cc
```diff
@@ -300,6 +296,36 @@ llvm::DIType* CodeGenCPU::GetDebugType(const Type& ty_tir, llvm::Type* ty_llvm)
                                              : nullptr;
     return dbg_info_->di_builder_->createPointerType(pointee_type,
                                                      ty_llvm->getPrimitiveSizeInBits());
+
+  } else if (auto* prim_type = ty_tir.as<PrimTypeNode>()) {
+    DataType dtype = prim_type->dtype;
+    auto [name, dwarf_type] = [&]() -> std::tuple<const char*, llvm::dwarf::TypeKind> {
```
kparzysz-quic, line 302: `llvm::StringRef` is the "LLVM way" rather than `const char*`.

src/target/llvm/codegen_cpu.cc
Lunderberg: Thank you, and updated to use `DLDataType2String`.

Lunderberg:
@echuraev Good point.  Probably not strictly required, as the existing DWARF handling isn't tested, but this does impact what function signatures are allowed when not using the PackedFunc API.  I've added a unit test that fails on main, and passes with this change.

Lunderberg:
@kparzysz-quic Thank you, and thank you for the catch!  Looks like I added the unit test in my local branch, then forgot to push.

PR #14572, opened 2023-04-11
[Relay] Enhance type infer for dynamic shapes

shengxinhu:


shengxinhu:
Enhance to support type_infer to support PrimExpr such as  tir.IndexMod(relay.Any(),5) in unifier

PR #14574, opened 2023-04-11
[IR][SIBuilder]

chunit-quic:
- Add SIBuilder to handle the span propagation between passes
- Add SequentialSpan for multiple source expressions conversion between passes
- Add test cases for SIBuilder and SequentialSpan

### DIFF
--- a/include/tvm/ir/si_builder.h
+++ b/include/tvm/ir/si_builder.h
@@ -1,0 +1,103 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/*!
+ * \file tvm/ir/si_builder.h
+ * \brief build a source info during rewriting expressions.
+ */
+#ifndef TVM_IR_SI_BUILDER_H_
+#define TVM_IR_SI_BUILDER_H_
+
+#include <tvm/relay/expr.h>
+#include <tvm/relay/expr_functor.h>
+#include <tvm/tir/stmt.h>
+
+#include <memory>
+#include <unordered_set>
+
+namespace tvm {
+
+/*!
+ * \brief SIBuilder provides helper APIs for filling spans,
+ *        particularly useful for one-to-many, many-to-one and many-to-many pass transformations.
+ */
+class SIBuilder {
+ public:
+  /*!
+   * \brief Create SIBuilder from a given span
+   */
+  explicit SIBuilder(const Span& span = Span());
+
+  /*!
+   * \brief Create SIBuilder from a given span sequence
+   */
+  explicit SIBuilder(const Array<Span>& spans = Array<Span>());
+  explicit SIBuilder(const std::initializer_list<Span>& init);
+
+  /*!
+   * \brief Create SIBuilder via a subgraph,
+   *        Will construct span based on the exprs in the subgraph. Including the inputs exprs.
+   *
+   * \param entry Entry expr for subgraph
+   * \param inputs End exprs for subgraph
+   */
+  template <typename T, typename = std::enable_if_t<std::is_base_of<BaseExpr, T>::value>>
+  explicit SIBuilder(const T& entry, const tvm::Array<T>& inputs = {});
+  explicit SIBuilder(const tir::Stmt& entry, const tvm::Array<PrimExpr>& inputs = {});
+  explicit SIBuilder(const tir::Stmt& entry, const tvm::Array<tir::Stmt>& inputs = {});
+
+  ~SIBuilder();
+
+  SIBuilder(const SIBuilder&) = delete;
+  SIBuilder& operator=(const SIBuilder&) = delete;
+
+  /*!
+   * \brief create new source info based on the given span or subgraph.
+   *
+   * \return The given span, or reconstructed span from subgraph.
+   */
+  Span CreateSpan() const;
+
+  /*!
+   * \brief Recursively fill all span of exprs in subgraph from entry until inputs.
+   *
+   * \param entry Entry expr for subgraph.
+   * \param inputs End exprs for subgraph, will not be filled with new span.
+   */
+  template <typename T, typename = std::enable_if_t<std::is_base_of<BaseExpr, T>::value>>
+  void RecursivelyFillSpan(
+      const T& entry, const std::unordered_set<T, ObjectPtrHash, ObjectPtrEqual>& inputs) const;
+
+  void RecursivelyFillSpan(
+      const tir::Stmt& entry,
+      const std::unordered_set<PrimExpr, ObjectPtrHash, ObjectPtrEqual>& inputs) const;
+  void RecursivelyFillSpan(
+      const tir::Stmt& entry,
+      const std::unordered_set<tir::Stmt, ObjectPtrHash, ObjectPtrEqual>& inputs) const;
+
+ private:
+  struct Impl;
+  std::unique_ptr<Impl> impl_;
+
+  std::unique_ptr<Impl> CreateImpl(const Span& span);
+};
+
+}  // namespace tvm
+
+#endif  // TVM_IR_SI_BUILDER_H_
--- a/include/tvm/ir/si_builder.h
+++ b/include/tvm/ir/si_builder.h
@@ -34,8 +34,8 @@
 namespace tvm {
 
 /*!
- * \brief SIBuilder provides helper APIs for filling spans,
- *        particularly useful for one-to-many, many-to-one and many-to-many pass transformations.
+ * \brief Source Information Builder, SIBuilder provides helper APIs for filling spans,
+ *        particularly useful for one-to-many, many-to-one and many-to-many IR transformations.
  */
 class SIBuilder {
  public:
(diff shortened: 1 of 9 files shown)


chunit-quic:
Hi community,

This is third part of TVM Explorer infrastructures. Based on this patch, we can tag source
information, the spans of source expressions, to the generated expressions during the pass transformations.

To give you more context, we briefly describe what have been done below in this PR. Here is the
list of main modifications:

1. Create a class SequentialSpan which is derived from Span.
2. Create a class SIBuilder to manage the span propagation during the pass transformations.
3. Use the pass context config, enable_si_builder to active SIBuilder.

You can reference to [preRFC](https://discuss.tvm.apache.org/t/pre-rfc-tvm-explorer-infrastructure/13457#pass-source-information-builder-6) for more details. Thanks! :D

--- 

chunit-quic:
> I wonder if it's relay specific. If so, could you please move it under relay's namespace

Hi @Hzfengsy,
Thank you for reply. :D
Although the span filling we have done are mainly in relay passes,  we also implement the functionalities for TIR in [SIBuilder](https://github.com/apache/tvm/pull/14574/files#diff-a948171162b02fc584b9ff95f8548486786d78ace286c6c58948d4608003c071R88). 
Perhaps it would be fine to tag with IR? 
Thanks!

chunit-quic:
> I wonder if it's relay specific. If so, could you please move it under relay's namespace

Hi @Hzfengsy,
Just a gentle ping, would you mind to review this PR. Or is there any other is recommended for this topic? Thank you :D

Hzfengsy:
Sorry, I'm not an expert on `Span` system. cc @tqchen @junrushao

include/tvm/ir/si_builder.h
```diff
@@ -0,0 +1,103 @@
...
+#include <memory>
+#include <unordered_set>
+
+namespace tvm {
+
+/*!
+ * \brief SIBuilder provides helper APIs for filling spans,
+ *        particularly useful for one-to-many, many-to-one and many-to-many pass transformations.
+ */
+class SIBuilder {
```
masahi, line 40: Please document what "SI" stands for

include/tvm/ir/si_builder.h
```diff
@@ -0,0 +1,103 @@
...
+#include <tvm/tir/stmt.h>
+
+#include <memory>
+#include <unordered_set>
+
+namespace tvm {
+
+/*!
+ * \brief SIBuilder provides helper APIs for filling spans,
+ *        particularly useful for one-to-many, many-to-one and many-to-many pass transformations.
```
masahi, line 38: pass transformations -> IR transformations?

src/ir/si_builder.cc
```diff
@@ -0,0 +1,341 @@
...
+
+using RelayExprSet = std::unordered_set<relay::Expr, ObjectPtrHash, ObjectPtrEqual>;
+using PrimExprSet = std::unordered_set<PrimExpr, ObjectPtrHash, ObjectPtrEqual>;
+using StmtSet = std::unordered_set<tir::Stmt, ObjectPtrHash, ObjectPtrEqual>;
+
+class RelayCollapse : public relay::ExprVisitor {
+ public:
+  explicit RelayCollapse(const RelayExprSet& inputs = {}) : inputs_(inputs) {}
+
+  Span Collapse(const relay::Expr& entry);
```
masahi, line 40: Please document what `Collapse` and `Fill` mean in this file

python/tvm/ir/base.py
```diff
@@ -69,6 +69,20 @@ def __init__(self, source_name, line, end_line, column, end_column):
         )
 
 
+@register_object("SequentialSpan")
+class SequentialSpan(Object):
+    """Specifies a location in a source program.
```
masahi, line 74: Please update this doc, since it refers to only a single span. 

Does this class need to be exposed to Python? If it is used only by c++ code, no need to define a python class.

src/ir/si_builder.cc
masahi: I think `CollectSpans` will be a more descriptive name for `Collapse`?

src/ir/si_builder.cc
```diff
@@ -0,0 +1,341 @@
...
+void SIBuilder::RecursivelyFillSpan(const tir::Stmt& entry, const PrimExprSet& inputs) const {
+  impl_->RecursivelyFillSpan(entry, inputs);
+}
+
+void SIBuilder::RecursivelyFillSpan(const tir::Stmt& entry, const StmtSet& inputs) const {
+  impl_->RecursivelyFillSpan(entry, inputs);
+}
+
+std::unique_ptr<SIBuilder::Impl> SIBuilder::CreateImpl(const Span& span) {
+  struct NullImpl : public SIBuilder::Impl {
```
masahi, line 246: This class and the base class `SIBuilder::Impl ` don't seem to be necessary. If SI collection is not enabled, there is no need to collect spans or create the impl object. We should check `enable_si_builder` condition much earlier.

masahi (changes requested):

chunit-quic (commented):
Hi @masahi,
Sorry for late reply and thank you very much for reviewing our codes!
Based on the comments we have done changes. Yet we would like to check with you some details of them before push a new patch.

Any comment is appreciated. Thank you :D

include/tvm/ir/si_builder.h
```diff
@@ -0,0 +1,103 @@
...
+  explicit SIBuilder(const tir::Stmt& entry, const tvm::Array<PrimExpr>& inputs = {});
+  explicit SIBuilder(const tir::Stmt& entry, const tvm::Array<tir::Stmt>& inputs = {});
+
+  ~SIBuilder();
+
+  SIBuilder(const SIBuilder&) = delete;
+  SIBuilder& operator=(const SIBuilder&) = delete;
+
+  /*!
+   * \brief create new source info based on the given span or subgraph.
```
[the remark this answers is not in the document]
masahi: yes looks good

src/ir/si_builder.cc
masahi: What I meant was, if si collection is not enabled, we don't have to call `CreateImpl` at all. What you described still sounds too complicated for its purpose.

chunit-quic: I see. I agree this implementation seems to be a bit too complicated in this case.
Basically we had two strategies to enable/disable the functionalities of SIBuilder. May you
tell us which one you prefer more or please kindly give us some advices? :D

In the following, I take a real case from one of our upcoming pass transformation PR for example,
I will list pro and con of our two strategies for you.

# The use case of SIBuilder
Here we uses `SimplifyReshape` in `${tvm}/src/relay/transforms/simplify_expr.cc` for example. It is
a many-to-one transformation. The SIBuilder will be invoked like following block:

```C++
class SimplifyReshape : public DFPatternRewrite {
 public:
  SimplifyReshape() {
    x_ = IsWildcard();
    auto reshape1 = IsOp("reshape") || IsOp("contrib_reverse_reshape");
    auto reshape2 = IsOp("reshape") || IsOp("contrib_reverse_reshape");
    pattern_ = reshape1({reshape2({x_})});
  }

  Expr Callback(const Expr& pre, const Expr& post,
                const Map<DFPattern, Array<Expr>>& node_map) const override {
    //...
    if (const_shape) {
      auto x = node_map[x_][0];
      auto ret = MakeReshape(x, newshape);

      SIBuilder si_builder(/* entry */ node_map[pattern_][0], /* inputs */ {x});
      ret->span = si_builder.Build();

      return ret;
    }
    return post;
  }
//..
};
```

In above case the si_builder is initialized by a sub graph, and invokes its member function to
build a new span. To make its functionalities enabled/disabled, we have 2 proposals

# Proposals
1. Null implementation
Just like what we have now. The pro is, we check `enable_si_builder` in a place(CreateImpl) only.
The con is obviously, it seems to be complicated in this case.

2. Embed `enable_si_builder` checking in each place
Another way is we can embed `enable_si_builder` checking at the beginning constructors or functions.
Take RecursivelyFillSpan an example, we can add the checking like this:

```C++
template <>
void SIBuilder::RecursivelyFillSpan(const relay::Expr& entry, const RelayExprSet& inputs) const {
    if (enabled) {
        RecursivelyFillSpan(entry, inputs);
    }
}
```
The pro is, the codes are very straightforward now. The con is the duplicated checking of
`enable_si_builder` will be added to multiple constructors and functions. It might seem to be a bit
redundant.

Currently we use the first proposal because we can reduce the checking to a place only, though it
is a bit complicated. Yet we are good to use the proposal 2. Or perhaps may you kindly point out
some strategies we might miss out?

Thank you for reading this long rely. We will get back to you soon. :D

masahi: Yes, 2. is what I meant by "checking early". If that ends up with too-much repetitive check, I'm ok with your solution of making `NullImpl` a base class.

chunit-quic: Thank you very much for clarifying and understanding. :D
We would like to keep this NullImpl. Because we made many constructors to serve both Relay and TIR, the second proposal will lead us to add quite a lot number of checking (perhaps more than ten). 

We will fix the the codes based your comments and push it once we finished.

masahi (approved):

junrushao:
This PR introduces a potential issue as suggested by clang, which we might want to patch:

```
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.3.sdk/usr/include/c++/v1/__memory/unique_ptr.h:48:5: warning: delete called on non-final 'tvm::SIBuilder::Impl' that has virtual functions but non-virtual destructor [-Wdelete-non-abstract-non-virtual-dtor]
    delete __ptr;
    ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.3.sdk/usr/include/c++/v1/__memory/unique_ptr.h:305:7: note: in instantiation of member function 'std::default_delete<tvm::SIBuilder::Impl>::operator()' requested here
      __ptr_.second()(__tmp);
      ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.3.sdk/usr/include/c++/v1/__memory/unique_ptr.h:259:19: note: in instantiation of member function 'std::unique_ptr<tvm::SIBuilder::Impl>::reset' requested here
  ~unique_ptr() { reset(); }
                  ^
/Users/jshao/Projects/tvm-dev/src/ir/si_builder.cc:229:12: note: in instantiation of member function 'std::unique_ptr<tvm::SIBuilder::Impl>::~unique_ptr' requested here
SIBuilder::~SIBuilder() = default;
           ^
```

PR #14576, opened 2023-04-11
[TOPI] Enhance ConfigSpace to support dynamic shape for schedules

shengxinhu:
If we want to infer dynamic function on llvm, it is likely that we need to add an extra dynamic schedule for some ops, such as batch_matmul.  One of the reasons is that auto-tvm use ConfigSpace to define space which do not support dynamic shapes. This pr try to enhance it to support dynamic shapes, to enable dynamic functions to run successfully without additional dynamic schedules.


shengxinhu:
@merrymercy

tqchen:
Thanks @shengxinhu . The particular approach can be limited as we still need to exploring splitting dynamic dimension as well.

Recently there are a lot of pushes in enabling dynamism in tensorIR and meta-schedule template, you are more than welcome to checkout that line of work.

PR #14581, opened 2023-04-11
[Unity] Update legalizer for variance

Ubospica:
Now the legalization method for relax.variance is:
```
var(x) = mean((x - mean(x)) ** 2)
```
This PR changes the legalization method into
```
var(x) = mean(x * x) - mean(x) ** 2
```
for better data locality (which means better performance)


psrivas2 (approved):

tqchen:
there might be some numerical stability issues here, so would be good to put this on hold a bit and come and visit later.

PR #14587, opened 2023-04-11
[microNPU] Fix skip tests when Vela is not present

lhutton1:
This was failing locally due to an indirect import in `infra.py`. Also ensure both tests are run by using the standard `tvm.testing.main()` function to run the tests.

cc @arina-grovety @sergio-grovety @ekalda


ekalda (approved):
LGTM!

ekalda:
Thanks @lhutton1!

PR #14588, opened 2023-04-11
[Fix][TIR] UnifyThreadBinding creating unit loop with annotation

MasterJH5574:
This PR fixes a behavior of the UnifyThreadBinding pass which (at one place) assumes a return value is always a ForNode, which is not right.

To be more specific, when a thread-binding loop has an annotation, the current behavior is assuming that the post-recursive-mutation value is also a ForNode, and apply the previous annotation directly to the new loop. However, the post-recursive-mutation value is also possibly not a ForNode. In this case, the current behavior is incorrect.

This PR creates a new unit-length loop in this case to preserve the annotation.

Thanks Bohan for catching this issue.

### DIFF
--- a/src/tir/transforms/unify_thread_binding.cc
+++ b/src/tir/transforms/unify_thread_binding.cc
@@ -64,9 +64,20 @@ class ThreadBindingUnifier : public StmtExprMutator {
     if (annotations.empty()) {
       return stmt;
     }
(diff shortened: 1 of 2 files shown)


tqchen (approved):

MasterJH5574:
@tvm-bot rerun

PR #14590, opened 2023-04-11
[Docker] Support rootless docker when using docker/bash.sh

Lunderberg:
The `docker/bash.sh` script should omit the `--pid=host` argument and `docker/with_the_same_user` arguments when the host's docker daemon is running in [rootless mode](https://docs.docker.com/engine/security/rootless/).  The `with_the_same_user` script is unnecessary in this mode, as rootless docker daemons already use the privileges of the user.

The `--pid=host` flag is required when running in CI, and while it may be useful when running tests locally, it is only supported for rootless docker in versions docker versions 22.06 or greater ([requires this commit](https://github.com/moby/moby/pull/41893)).


echuraev (approved):
LGTM

PR #14591, opened 2023-04-11
[MetaSchedule] Handle cases when no features found by FeatureExtractor

Ubospica:
This PR fixes the problem that ops like `topi.full` cannot be correctly tuned.

Currently, metaSchedule system adopts a XGBoost model to predict the running performance of an IRModule based on features extracted from it. 

However, Some operators like `topi.full` has so simple IRModule that no feature can be extracted. `topi.full` contains only one BufferStore node with a FloatImm value, so `PerStoreFeature` extractor cannot find any feature from it. And the XGBModel also cannot handle cases when there is no features.

This PR specifically handles this case in `PerStoreFeatureNode` and `XGBModel`.

Thanks to @junrushao


junrushao (approved):
LGTM! Please fix the lint :-)

junrushao:
Nice! Should be ready to get in :-)

PR #14592, opened 2023-04-11
[TensorIR] Fix ComputeAt with perfect symbolic bound

tqchen:
This PR fixes a case where we have perfect symbolic bound in compute at and can be eliminated.

Testcases are added.


MasterJH5574 (approved):
LGTM

PR #11999, opened 2022-07-03
[TIR] Add sugar method `Schedule.work_on`

junrushao1994:
This PR introduces `Schedule.work_on`, which instructs
`Schedule.get_block` to find the correct PrimFunc to retrieve from
without having to specify `func_name` in every time if the PrimFunc's
name is not `main`.

PR #14594, opened 2023-04-11
[TIR][Schedule] Method returning the function being worked on

MasterJH5574:
Cherry-pick of #14593.

---

PR #11999 introduces the sugar method `work_on` to TIR Schedule, with a field `func_working_on_` newly added to the ScheduleNode. In some cases we may want to know which function a ScheduleNode is working on, which is not supported previously.

Therefore, this PR introduces a method to ScheduleNode that returns the function (more accurately, GlobalVar) currently being worked on. With this we are able to know the function being worked on.


tqchen:
we will do a merge of main into tir, and let us wait until then to put it into unity

PR #14595, opened 2023-04-11
[Unity][MetaSchedule] BlockCollector focusing on current func

MasterJH5574:
PR #14182 introduces the default GPU schedule pass, which makes uses
of the BlockCollector in Meta-Schedule. When requested, the
BlockCollector goes through every function in the input IRModule,
while in the case of applying default GPU schedule where we apply
schedule to one PrimFunc at a time, going through every PrimFunc is
wasteful.

Therefore, this PR enhances the BlockCollector so that it only focuses
on the function being worked on.


junrushao (approved):
LGTM! Perhaps we should send both PRs to mainline.

PR #14599, opened 2023-04-12
[TIR][SSA] Add patch support of thread binding axes in ConvertSSA

wrongtest-intellif:
Currently the attrstmt of `thread_extent` and `virtual_thread` serves as the def of corresponding iter vars, thus it seems to be proper to take them into SSA conversion.


tqchen (approved):

tqchen:
this is useful but a bit stale, closing for now(since ci trigger may not work), would be great to raise a new one

PR #14603, opened 2023-04-12
[Unity] Fix FuseTIR when the same buffer is read multiple times with different access pattern

masahi:
When the same buffer is read multiple times with different access patterns in a single expression, the check below fails
https://github.com/apache/tvm/blob/unity/src/relax/transform/fuse_tir.cc#L266

But this case should be allowed, for example in the following subgraph `inp_0` is used twice in different read regions. See the test case for details. This subgraph arises if we run `ConvertLayout` on SD UNet from web-stable diffusion.

```
lv: R.Tensor((2, 4, 64, 64), dtype="float32") = R.concat((inp_0, inp_0), axis=0)
lv_2: R.Tensor((2, 64, 64, 4), dtype="float32") = R.permute_dims(lv, axes=[0, 2, 3, 1])
```

@Hzfengsy

### DIFF
--- a/tests/python/relax/test_transform_fuse_tir.py
+++ b/tests/python/relax/test_transform_fuse_tir.py
@@ -883,5 +883,125 @@ def test_symbolic_var_in_call_tir_args():
     _check(Before, Expected)
 
 
+def test_same_buffer_multiple_read():
+    @I.ir_module
+    class Module:
+        @T.prim_func
+        def concatenate(
+            rxplaceholder: T.Buffer((T.int64(1), T.int64(4), T.int64(64), T.int64(64)), "float32"),
+            rxplaceholder_1: T.Buffer(
+                (T.int64(1), T.int64(4), T.int64(64), T.int64(64)), "float32"
+            ),
+            T_concat: T.Buffer((T.int64(2), T.int64(4), T.int64(64), T.int64(64)), "float32"),
+        ):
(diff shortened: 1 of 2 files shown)


tests/python/relax/test_transform_fuse_tir.py
```diff
@@ -883,5 +883,125 @@ def main(
...
+            ),
+            T_concat: T.Buffer((T.int64(2), T.int64(4), T.int64(64), T.int64(64)), "float32"),
+        ):
+            T.func_attr({"op_pattern": 2, "tir.noalias": T.bool(True)})
+            for ax0, ax1, ax2, ax3 in T.grid(T.int64(2), T.int64(4), T.int64(64), T.int64(64)):
+                with T.block("T_concat"):
+                    v_ax0, v_ax1, v_ax2, v_ax3 = T.axis.remap("SSSS", [ax0, ax1, ax2, ax3])
+                    T.reads(
+                        rxplaceholder_1[v_ax0 - T.int64(1), v_ax1, v_ax2, v_ax3],
+                        rxplaceholder[v_ax0, v_ax1, v_ax2, v_ax3],
```
masahi, line 903: Note that in this test case, `rxplaceholder_1` and `rxplaceholder` refer to the same buffer `inp_0`. But they are used with different access patterns.

Hzfengsy (approved):

PR #14605, opened 2023-04-12
[Unity] Fix ForceNarrowI32 with pod arguments

tqchen:
This PR fixes ForceNarrowI32 for functions that comes with pod arguments.

Test cases are added


tqchen:
will send a separate PR to main and cover the force narrow code.

junrushao (approved):
LGTM!

junrushao:
We might want to send this PR to mainline as well

tqchen:
agree, forcenarrow function is not yet in main, and we should send it later

PR #14606, opened 2023-04-12
[Unity][UX] Symbolic Variables Used in Multiple Functions

SiriusNEO:
Prior to this PR, there is no constraint to prevent user defining multiple functions which may use the same symbolic TIR var. For example, user may write the following script:
```
batch_size = tir.Var("batch_size", "int64")

@I.ir_module
class Test:
    @R.function
    def main(
        x: R.Tensor((batch_size, 10), "float32"),
    ):
        with R.dataflow():
            lv = R.sum(x, axis=1)
            lv1 = R.mean(x, axis=1)
            out = R.add(lv, lv1)
            R.output(out)
        return out

    @R.function
    def main1(
        x: R.Tensor((batch_size, 10), "float32"),
        y: R.Tensor((batch_size, 10), "float32"),
    ):
        with R.dataflow():
            out = R.subtract(x, y)
            R.output(out)
        return out

lowered_mod = LegalizeOps()(Test)

ex = relax.build(lowered_mod, "llvm")
vm = relax.VirtualMachine(ex, tvm.cpu(0))
```
But this script can not be built because `VMShapeLower` can not handle this case well.
Since it is a illegal behaviour, we need to prevent user from doing this. Specifically, this PR contains two parts of work:
- Let well form checker to check this case.
- Let `CopyWithNewVars` util copies the symbolic vars in the struct info inside the function.

### DIFF
--- a/python/tvm/relax/ir/instrument.py
+++ b/python/tvm/relax/ir/instrument.py
@@ -34,5 +34,4 @@ class WellFormedInstrument:
 
     def run_after_pass(self, mod, pass_info):
         if pass_info.name not in self.skip_pass_name:
-            mod.show()
             assert relax.analysis.well_formed(mod)
--- a/python/tvm/relax/ir/instrument.py
+++ b/python/tvm/relax/ir/instrument.py
@@ -34,4 +34,5 @@ class WellFormedInstrument:
 
     def run_after_pass(self, mod, pass_info):
         if pass_info.name not in self.skip_pass_name:
+            mod.show()
             assert relax.analysis.well_formed(mod)
--- a/src/relax/analysis/well_formed.cc
+++ b/src/relax/analysis/well_formed.cc
@@ -30,16 +30,17 @@
  *    3. When a Function has a corresponding GlobalVar and a `global_symbol`
  *       attribute, the name of the GlobalVar must equal the value of the
  *       `global_symbol` attribute value.
- *    4. Any variable cannot used as different function parameters in the same IRModule
- *    5. Vars are defined before use.
- *    6. Vars are defined exactly once.
- *    7. Symbolic Vars are defined before use.
- *    8. DataflowVars cannot be defined inside BindingBlock.
- *    9. Vars defined in IfNode, except the return Var, are invisible
+ *    4. Any variable cannot used as different function parameters in the same IRModule.
+ *    5. Any symbolic var cannot present across different functions in the same IRModule.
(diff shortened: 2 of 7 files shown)


SiriusNEO:
For the `CopyWithNewVars` part, cc @Ubospica @yelite

yongwww:
Thanks @SiriusNEO for the effort. I am curious about the error in VMShapeLower for this case, could you elaborate it a bit?

SiriusNEO:
@yongwww Sure! For the example above, if you run it you will get error in this line:
```
ex = relax.build(lowered_mod, "llvm")
```
And the error msg is:
```
        at /home/sirius/catalyst/mlc-ai/src/relax/backend/vm/vm_shape_lower.cc:366
  File "/home/sirius/catalyst/mlc-ai/src/relax/backend/vm/vm_shape_lower.cc", line 366
TVMError:
---------------------------------------------------------------
An error occurred during the execution of TVM.
For more information, please see: https://tvm.apache.org/docs/errors.html
---------------------------------------------------------------
  Check failed: (slot->value_computed) is false: PrimExpr T.int64(4) * batch_size * T.int64(10) has not been computed
```
But if you remove a function, either `main` or `main1` from the module `Test`, it goes well.
Another trial is that I clear `slot_vec_` and `slot_map_` in the beginning of `Function Rewrite(GlobalVar gvar, Function func)` in the vm shape lower, and then it also works for this example.
So after digging it a bit, I found the problem lies in the dup tir var in multiple functions (For more details I can't figure it out completely, since I'm not quite familiar with the logic of `VMShapeLower`). And after communicating with TQ, this behaviour (Use the same TIR var in multiple functions) is considered illegal. So I send this PR.

yongwww:
@SiriusNEO Gotcha, thanks!

python/tvm/relax/ir/instrument.py
```diff
@@ -34,4 +34,5 @@ def run_before_pass(self, mod, pass_info):
 
     def run_after_pass(self, mod, pass_info):
         if pass_info.name not in self.skip_pass_name:
+            mod.show()
```
yongwww, line 37: remove this?

yongwww (approved):

Hzfengsy (approved):

Hzfengsy:
cc @LeshengJin

MasterJH5574 (approved):

PR #14608, opened 2023-04-12
[Unity][BYOC] Add fused patterns for stacked attention

cyx-6:
In some models, the input Q, K and V for attention ops are from a stacked tensor initially, and then they are splitted and reshaped to call attention op, like

stacked_qkv -> split -> reshape -> attention.

Actually, we could to skip the split and reshape ops, by manipulating the layout parameters in codegen.

This PR adds the such fused patterns for stacked attention in BYOC. So that we are able to codegen directly from stacked_qkv.

### DIFF
--- a/python/tvm/contrib/cutlass/attention_operation.py
+++ b/python/tvm/contrib/cutlass/attention_operation.py
@@ -86,13 +86,13 @@ def instantiate_attention_template(attrs):
   p.v_strideH = p.head_dim_value; // H'
 
   // stride for S
-  p.q_strideM = p.k_strideM = p.v_strideM = 
-    p.q_strideH * p.num_heads + 
-    p.k_strideH * p.num_heads + 
+  p.q_strideM = p.k_strideM = p.v_strideM =
+    p.q_strideH * p.num_heads +
+    p.k_strideH * p.num_heads +
     p.v_strideH * p.num_heads; // H * N + H * N + H * N'
 
   // stride for B
-  p.q_strideB = p.k_strideB = p.v_strideB = 
+  p.q_strideB = p.k_strideB = p.v_strideB =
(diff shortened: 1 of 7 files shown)


vinx13 (approved):

python/tvm/contrib/cutlass/attention_operation.py
```diff
@@ -19,45 +19,87 @@
...
+        "qkv_stacked": """
+  p.query_ptr = reinterpret_cast<T *>(${qkv}->data);
+  p.key_ptr = reinterpret_cast<T *>(${qkv}->data) + p.head_dim * p.num_heads;
+  p.value_ptr = reinterpret_cast<T *>(${qkv}->data) + p.head_dim * p.num_heads * 2;
+  CHECK(${qkv}->ndim == 3); // B, S, NH + NH + NH'
+
+  // stride for N
+  p.q_strideH = p.head_dim; // H
+  p.k_strideH = p.head_dim; // H
+  p.v_strideH = p.head_dim_value; // H'
```
vinx13, line 86: should these be (H+H+H')?

python/tvm/relax/backend/contrib/cutlass.py
```diff
@@ -244,6 +245,14 @@ def attention_patterns():
             "cutlass.attention_bias",
             *make_attention_pattern(with_bias=True),
         ),
+        (
+            "cutlass.stacked_attention",
```
vinx13, line 249: does the order here matter? If we have a subgraph containing both reshape and attention, will `cutlass.attention` that matches only a single attention operation be selected first?

cyx-6: the order matters here. I tried to change the order with stacked attention first, however, the original attention matches first.

PR #14612, opened 2023-04-13
[Unity][AMP] Fix merging concrete type and "unknown" type

masahi:
In AMP, an abstract type representing "unknown" type has a string representation "". `String2DLDataType` returns `DataType::Void()`  https://github.com/apache/tvm/blob/c452e6966c33047512155f42f63aef4e0586d129/include/tvm/runtime/data_type.h#L351-L352 for such string. 

So when the abstract type is `unknown_`,  `ICHECK_EQ(a.code(), b.code())` in `NTypeMerge` always fails because we try to merge a concrete type such as "float32" with void.

@spectrometerHBH

### DIFF
--- a/src/relax/transform/infer_amp_utils.cc
+++ b/src/relax/transform/infer_amp_utils.cc
@@ -40,6 +40,11 @@ NType NTypeMerge(const NType& a, const NType& b) {
   auto fcombine = [&](const String& a_str, const String& b_str) -> String {
     DataType a = DataType(String2DLDataType(a_str));
     DataType b = DataType(String2DLDataType(b_str));
+    if (a_str == "") {
+      return b_str;
+    } else if (b_str == "") {
+      return a_str;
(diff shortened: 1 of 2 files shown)


src/relax/transform/infer_amp_utils.cc
```diff
@@ -40,6 +40,11 @@ NType NTypeMerge(const NType& a, const NType& b) {
   auto fcombine = [&](const String& a_str, const String& b_str) -> String {
     DataType a = DataType(String2DLDataType(a_str));
     DataType b = DataType(String2DLDataType(b_str));
+    if (a_str == "") {
```
psrivas2, line 43: should we move this if-else-if before lines 41 and 42?

masahi:
@tvm-bot rerun

yongwww (approved):
LGTM

PR #14613, opened 2023-04-13
[Unity][CUTLASS] Support batched matmul + residual fusion

masahi:
Follow-up to https://github.com/apache/tvm/pull/14317 to support batched GEMM + residual block fusion.

@vinx13


masahi:
@tvm-bot rerun

PR #14617, opened 2023-04-13
[Unity] Update specific builtins for LM

tqchen:
This PR updates the specific builtins for LM
and move them to lm_support.cc

The kv_create now takes an initial data and copies it instead of consumes it. This will enable us to create kv within a VM more easily.


yongwww (approved):

junrushao (approved):
LGTM

PR #14624, opened 2023-04-14
[Unity] Introduce FewShotTuning Pass

zxybazh:
Based on feedback for `DefaultGPUSchedule` pass, we noticed that while PrimFuncs can be supported without tuning on GPU devices, the performance is not competitive. Therefore we introduce a new FewShotTuning pass for static shape PrimFuncs that directly samples from the search space and make sure the returned candidate can pass post-processors.

By design, this pass enables most of the performance optimizations according to the target, and avoided tuning by:
- Disable benchmarking by default.
- Only do random sampling.
- Limit valid trials to user specified small number, by default 1.
- Skip tuning apis and directly do sampling and compilation.
- Still guarantee result is compilable.

This PR is a follow-up for @psrivas2's work and based on discussion with @sunggg and @psrivas2.

CC: @jwfromm @junrushao

### DIFF
--- a/include/tvm/relax/transform.h
+++ b/include/tvm/relax/transform.h
@@ -499,10 +499,11 @@ TVM_DLL Pass RewriteCUDAGraph();
  *  on MetaSchedule schedule rules but directly samples from the search space instead of using the
  *  tuning algorithm. User can specify the number of valid counts to try and whether to use runner
  *  for evaluation.
+ * \param valid_count The number of valid counts to try.
+ * \param runner The runner to evaluate the generated schedules.
  * \return The Pass.
  */
-
-TVM_DLL Pass FewShotTuning(int valid_count, ObjectRef runner);
+TVM_DLL Pass FewShotTuning(Integer valid_count, ObjectRef runner);
 
 }  // namespace transform
 }  // namespace relax
--- a/include/tvm/relax/transform.h
+++ b/include/tvm/relax/transform.h
@@ -493,6 +493,17 @@ TVM_DLL Pass ToMixedPrecision(const DataType& out_dtype,
  */
 TVM_DLL Pass RewriteCUDAGraph();
 
+/*!
+ * \brief The pass is designed for few shot tuning for static shape PrimFuncs. It examines all the
+ *  blocks within the PrimFunc and conducts loop fusion, splitting, and other transformations based
+ *  on MetaSchedule schedule rules but directly samples from the search space instead of using the
+ *  tuning algorithm. User can specify the number of valid counts to try and whether to use runner
+ *  for evaluation.
+ * \return The Pass.
+ */
+
+TVM_DLL Pass FewShotTuning(int valid_count, ObjectRef runner);
+
 }  // namespace transform
 }  // namespace relax
 }  // namespace tvm
--- a/include/tvm/relax/transform.h
+++ b/include/tvm/relax/transform.h
@@ -498,12 +498,12 @@ TVM_DLL Pass RewriteCUDAGraph();
  *  blocks within the PrimFunc and conducts loop fusion, splitting, and other transformations based
  *  on MetaSchedule schedule rules but directly samples from the search space instead of using the
  *  tuning algorithm. User can specify the number of valid counts to try and whether to use runner
- *  for evaluation.
+ *  for benchmarking.
  * \param valid_count The number of valid counts to try.
- * \param runner The runner to evaluate the generated schedules.
+ * \param benchmark Whether to use runner for benchmarking.
  * \return The Pass.
  */
-TVM_DLL Pass FewShotTuning(Integer valid_count, ObjectRef runner);
+TVM_DLL Pass FewShotTuning(int valid_count, bool benchmark);
 
 }  // namespace transform
 }  // namespace relax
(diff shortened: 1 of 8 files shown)


tests/python/relax/test_transform_few_shot_tuning.py
```diff
@@ -0,0 +1,56 @@
...
+                    T.writes(C[v_i, v_j])
+                    with T.init():
+                        C[v_i, v_j] = T.float16(0)
+                    C[v_i, v_j] = C[v_i, v_j] + A[v_i, v_k] * B[v_k, v_j]
+    # fmt: on
+    # pylint: enable=no-self-argument,missing-class-docstring,line-too-long
+    target = tvm.target.Target("nvidia/geforce-rtx-3070")
+    with target, tvm.transform.PassContext(opt_level=3):
+        After = FewShotTuning()(Before)
+        After.show()
```
sunggg, line 52: Let's use the structural equality checks e.g. https://github.com/tlc-pack/relax/blob/relax/tests/python/relax/test_transform_legalize_ops_binary.py#L195

src/relax/transform/few_shot_tuning.cc
```diff
@@ -0,0 +1,169 @@
...
+  }
+  int best_idx = 0;
+  if (runner.defined()) {
+    for (int i = 1; i < costs.size(); ++i) {
+      if (costs[i] < costs[best_idx]) {
+        best_idx = i;
+      }
+    }
+  } else {
+    best_idx = results.size() - 1;
```
sunggg, line 125: So, if we don't define runner, we pick the last one among the buildable candidates. 
Is there any benefit over using `valid_count` of 1 in this case?

include/tvm/relax/transform.h
```diff
@@ -481,6 +481,18 @@ TVM_DLL Pass DeadCodeElimination(Array<runtime::String> entry_functions);
...
+ * \brief The pass is designed for few shot tuning for static shape PrimFuncs. It examines all the
+ *  blocks within the PrimFunc and conducts loop fusion, splitting, and other transformations based
+ *  on MetaSchedule schedule rules but directly samples from the search space instead of using the
+ *  tuning algorithm. User can specify the number of valid counts to try and whether to use runner
+ *  for evaluation.
+ * \param valid_count The number of valid counts to try.
+ * \param runner The runner to evaluate the generated schedules.
+ * \return The Pass.
+ */
+TVM_DLL Pass FewShotTuning(Integer valid_count, ObjectRef runner);
```
sunggg, line 494: Passing `runner` might be too much overhead for most of the cases. 
Can we use the default runner and use the boolean argument like `enable_runner`?

sunggg (commented):
Thank you, @zxybazh! I left some questions.

tests/python/relax/test_transform_few_shot_tuning.py
zxybazh: Thanks for the suggestion! It seems we cannot know for sure which search space we will be using and what decision will be used in this case. Therefore, `structural_equal` might be unfit here, what about we check for the trace and see if the trace is one of our search space?

src/relax/transform/few_shot_tuning.cc
zxybazh: Good point, if `valid_count` is `1`, I think there's no gain / benefit to generate more candidate without benchmarking. We can fix the option to default to 1 when benchmarking is not enabled.

include/tvm/relax/transform.h
zxybazh: Yeah sure, will do!

zxybazh:
@sunggg Now all the comments are addressed, and the tests are working. For testing, note that we cannot know which search space and what decision is picked becaused it’s randomly sampling from the search space. Therefore, I only made sure that the result is correct and schedule is applied (so that it can run on cuda). Please take another look.

tests/python/relax/test_transform_few_shot_tuning.py
```diff
@@ -0,0 +1,436 @@
...
+    ]
+    actual_rt_mod(*actual_data)
+    return actual_data[-1].numpy()
+
+
+def _assert_allclose(mod: tvm.ir.IRModule, actual: tvm.ir.IRModule, func: tvm.tir.PrimFunc):
+    inputs, output_shape, output_dtype = _get_input_output_info(func)
+    expected_output = _expected_results(mod, inputs, output_shape, output_dtype)
+    actual_output = _actual_results(actual, inputs, output_shape, output_dtype)
+    tvm.testing.assert_allclose(expected_output, actual_output, rtol=1e-2, atol=1e-2)
```
sunggg, line 103: We usually use `rtol=1e-7, atol=1e-7`. Shall we change to this?

sunggg (commented):
One last comment and I think it is good to go. 
Thank you for addressing my comments!

tests/python/relax/test_transform_few_shot_tuning.py
zxybazh: Thanks Sung, the reason we were using 1e-2 is because the unit test was running on cuda, which has slight difference compare to on our baseline on llvm. Now that the default target is changed to llvm I'll change the threshold accordingly!

zxybazh: The accuracy has been fixed. For softmax test, fp16 doesn't satisfy 1e-7 accuracy but fp32 can work so I changed the PrimFunc to work in fp32. For Fused Variance Test I found that there's an issue with RewriteReduction Postproc so I've reported this issue with a reproducible script in #14791.

Other than that, the tests are all fixed and working.

tqchen (approved):

PR #14627, opened 2023-04-14
[Unity][CODEGEN] Fix metal codegen when with only single working dim

tqchen:
This PR fixes the metal introduces by a previous commit that removes the workdim remapping that caused issues for kernels with only threadIdx.x and blockIdx.x


tqchen:
cc @Lunderberg was caused by recent merging that bought in #14495 . This is a hotfix, will also need send to main

MasterJH5574 (approved):

tqchen:
right i think it should be covered by tests already but was not run due to metal being osx only

PR #14628, opened 2023-04-14
[CODEGEN] Fix metal codegen when with only single working dim

tqchen:
This PR fixes the metal introduces by a previous commit that removes the workdim remapping that caused issues for kernels with only threadIdx.x and blockIdx.x


tqchen:
cc @Lunderberg

tqchen:
@tvm-bot rerun

spectrometerHBH (approved):

PR #14636, opened 2023-04-16
Fix bug about wrong attribute name

jikechao:
The attribute 'errors' is a wrong name. When code was executed at this line, It will throw an error: message `"AttributeError: module 'tvm' has no attribute 'errors'"`. This PR fix it!


jikechao:
@vvchernov @anwang2009 Could you help me review it?  Thanks!

tqchen (approved):

PR #14639, opened 2023-04-16
[Unity][Frontend] Add `no_bind_return_tuple` for PyTorch FX Translator

SiriusNEO:
Previously, if we have a Torch model with multiple return values like:
```
class Mod(nn.Module):
    ...
    def forward(...):
        ...
        return logits, loss
```
The translator will always bind the return tuple to a Relax var and return it:
```
    gv: R.Tuple(...) = (logits, loss)
    return gv
```
This brings inconvenience if we want to use `Gradient` pass to set the loss as the target (i.e. set `target_index=1`). Because if the return value is a single Relax var, the `Gradient` pass will not take it as multiple return values.

This PR brings a flag to let user choose between different return styles. If the user set `no_bind_return_tuple = True`, they will get
```
    return (logits, loss)
```

### DIFF
--- a/tests/python/relax/test_frontend_from_fx.py
+++ b/tests/python/relax/test_frontend_from_fx.py
@@ -2708,6 +2708,40 @@ def test_unwrap_unit_return_tuple():
 
 
 @tvm.testing.requires_gpu
+def test_no_bind_return_tuple():
+    import torch.fx as fx
+    from torch.nn import Module
+    from tvm.relax.frontend.torch import from_fx
+
+    class Identity(Module):
+        def __init__(self):
+            super().__init__()
+
+        def forward(self, x, y):
+            return (x, y)
+
+    @tvm.script.ir_module
+    class Expected:
+        @R.function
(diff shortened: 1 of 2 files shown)


tests/python/relax/test_frontend_from_fx.py
```diff
@@ -2707,6 +2707,40 @@ def main(
...
+                gv: R.Tensor((256, 256), dtype="float32") = inp_0
+                gv1: R.Tensor((256, 256), dtype="float32") = inp_1
+                R.output(gv, gv1)
+            return (gv, gv1)
+
+    graph_model = fx.symbolic_trace(Identity())
+    mod = from_fx(
+        graph_model, [([256, 256], "float32"), ([256, 256], "float32")], no_bind_return_tuple=True
+    )
+    mod.show()
```
Hzfengsy, line 2740: remove printing :)

Hzfengsy (approved):
LGTM except a nit

PR #14640, opened 2023-04-16
[Unity] Improve error message in webgpu request

tqchen:
Sometimes user may get confused with an old version of browser, add more precise error message

### DIFF
--- a/web/src/webgpu.ts
+++ b/web/src/webgpu.ts
@@ -40,12 +40,46 @@ export async function detectGPUDevice(): Promise<GPUDeviceDetectOutput | undefin
     if (adapter == null) {
       throw Error("Cannot find adapter that matches the request");
     }
+    const computeMB = (value: number) => {
+      return Math.ceil(value  / (1 << 20)) + "MB";
+    }
+
+    // more detailed error message
+    const requiedMaxBufferSize = 1 << 30;
+    if (requiedMaxBufferSize > adapter.limits.maxBufferSize) {
+      throw Error(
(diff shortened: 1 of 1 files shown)


junrushao (approved):

DustinBrett:
Are these hard requirements? I am trying to get WebLLM going on Android and the only requirement it can't do is `maxStorageBufferBindingSize`.

tqchen:
the `maxStorageBufferBindingSize` likely is due to the size of the models we are dealing with. I remember trying to halving it and get an error, you can try to reduce it a bit though.

Also i think these constraints will get lifted in the future versions, dawn's matrix channel would be a good place to ask

DustinBrett:
> the `maxStorageBufferBindingSize` likely is due to the size of the models we are dealing with. I remember trying to halving it and get an error, you can try to reduce it a bit though.
> 
> Also i think these constraints will get lifted in the future versions, dawn's matrix channel would be a good place to ask

Ok, thanks for the quick reply! I am using Chrome Canary v115 on my Samsung Galaxy S20, but the limit seems to only be 128MB when it requires 1024MB. I will keep an eye on the progress and am happy to hear this could be lifted in the future. After seeing people running [mlc-llm](https://github.com/mlc-ai/mlc-llm) on mobile I am eager to do the same with WebLLM.

beaufortfrancois:
Any updates on this issue?

CharlieFRuan:
Hi @DustinBrett, the demo now has three models (ones with -1k suffix) that should be able to run with 128MB: https://webllm.mlc.ai/

PR #14641, opened 2023-04-16
[Unity][TARGET] Updates vulkan codegen for DeclBuffer

tqchen:
This PR updates vulkan codegen to support for DeclBuffer. We also disabled subgroup operations.

This is because we find that on Nvidia GPU subgroup were detected to be 8 and the generated code for softmax produces wrong results.

Setting it to be 32(the typical size of CUDA warp) also does not produce the right result.

We directly hard set the value to 1 to be ensure correctness. The team can revisit this behavior later.

### DIFF
--- a/src/runtime/vulkan/vulkan_device.cc
+++ b/src/runtime/vulkan/vulkan_device.cc
@@ -144,7 +144,11 @@ VulkanDeviceProperties::VulkanDeviceProperties(const VulkanInstance& instance,
   max_num_threads = properties.properties.limits.maxComputeWorkGroupInvocations;
 
   // Even if we can't query it, warp size must be at least 1.
-  thread_warp_size = std::max(subgroup.subgroupSize, 1U);
+  // thread_warp_size = std::max(subgroup.subgroupSize, 1U);
+  // vulkan's subgroup may not directly map to warp and atm
+  // can cause issues in softmax allreduce in NVidia GPU
+  // disable warp setting to be safe.
+  thread_warp_size = 1U;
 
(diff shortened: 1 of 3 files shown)


tqchen:
Note: i find this when trying vulkan on stable diffusion models on cuda, so sending to unity first for hotfix, we can cherrypick to main.

yongwww (approved):

tqchen:
Some steps to reproduces for those who are interested, in the `web-stable-diffusion` repo.

Without this patch will error at softmax op

```python
python build.py --target "vulkan -from_device=1"
python build.py --target "cuda"
python tests/debug/compare_lib.py --stage vae --cmp vulkan 
```



```python
```

tqchen:
FYI @Lunderberg @masahi

masahi (approved):

Hzfengsy (approved):

Lunderberg:
Interesting.  What GPU/driver is this on?  (On my local RTX 3070, both `subgroupSize` and `cudaDevAttrWarpSize` return 32.)

Also, is there a specific version of `web-stable-diffusion` that is required to reproduce the issue?  I'm getting an error when calling `relax.transform.RemoveUnusedFunction` ([here in build.py](https://github.com/mlc-ai/web-stable-diffusion/blob/main/build.py#L113)), as this function was removed in https://github.com/apache/tvm/pull/14336.

tqchen:
Mine was on RTX 4080, not sure about the driver. We are updating the script and can report back once it is ready

```
| NVIDIA-SMI 525.89.02    Driver Version: 525.89.02    CUDA Version: 12.0     |

```

PR #14649, opened 2023-04-17
[Unity][BYOC] Fuse attention pattern with `strided_slice`

cyx-6:
This PR expands the support for fused stacked attention patterns strating with `strided_slice`. Initially, we only support fused stacked attention pattern starting with `split` in #14608. But with the help of #14583, we may have similar patterns starting with `strided_slice` as well.

### DIFF
--- a/python/tvm/relax/backend/patterns.py
+++ b/python/tvm/relax/backend/patterns.py
@@ -197,12 +197,15 @@ def make_attention_pattern(with_bias: bool = False):
     return out, annotations
 
 
-def make_stacked_attention_pattern(with_bias: bool = False):
+def make_stacked_attention_pattern(start_op: str, with_bias: bool = False):
     """
     Create pattern for fused multi head attention with stacked input.
 
     Parameters
     ----------
+    start_op: str
+        The starting op for pattern, i.e. `R.split` or `R.strided_slice`.
+
     with_bias: bool
         Whether or not to include bias addition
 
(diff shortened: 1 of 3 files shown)


python/tvm/relax/backend/patterns.py
```diff
@@ -217,13 +220,23 @@ def make_stacked_attention_pattern(with_bias: bool = False):
...
-    qkv_tuple = is_op("relax.split")(stacked_qkv)
+    if start_op == "split":
+        qkv_tuple = is_op("relax.split")(stacked_qkv)
+        query_raw = is_tuple_get_item(qkv_tuple, 0)
+        key_raw = is_tuple_get_item(qkv_tuple, 1)
+        value_raw = is_tuple_get_item(qkv_tuple, 2)
+    elif start_op == "strided_slice":
+        query_raw = is_op("relax.strided_slice")(stacked_qkv)
+        key_raw = is_op("relax.strided_slice")(stacked_qkv)
+        value_raw = is_op("relax.strided_slice")(stacked_qkv)
```
vinx13, line 231: we will also need to check the begin/end in strides_slice if value has a different sequence length

masahi:
I found that `stride_slice` after combined matmul is very expensive. So I'm replacing #branches `strided_slice` with one `split`.

tests/python/relax/test_codegen_cutlass.py
```diff
@@ -676,10 +676,23 @@ def get_relax_stacked_attention_module(qkv, b, s, n, h, h_v, bias=None, qk_scale
...
-                q = R.reshape(qkv_tuple[0], [b, s, n, h])
-                k = R.reshape(qkv_tuple[1], [b, s, n, h])
-                v = R.reshape(qkv_tuple[2], [b, s, n, h_v])
+                if op == "split":
+                    qkv_tuple = R.split(qkv, [n * h, n * h * 2], axis=2)
+                    q = R.reshape(qkv_tuple[0], [b, s, n, h])
+                    k = R.reshape(qkv_tuple[1], [b, s, n, h])
+                    v = R.reshape(qkv_tuple[2], [b, s, n, h_v])
+                elif op == "strided_slice":
+                    qkv_tuple = R.split(qkv, [n * h, n * h * 2], axis=2)
```
masahi, line 685: `qkv_tuple` not used (I was confused)

cyx-6: yes, this is useless from copy and paste.

masahi (approved):

PR #14656, opened 2023-04-18
[Runtime] Fix Can't "query_imports" Bug of VM Executable

Johnson9009:
This "LOG(FATAL)" will error out when the function isn't found in “executable”, so it will prevent the recursive query the function in the imported modules of "executable".

For example, if the module hierarchy is something like below.
executable
|---LLVMModuleNode
|___NPUModuleNode

This bug will cause we can't get the function in NPUModuleNode by the call "executalbe.get_function("npu_func1", query_imports=True)".

From Arm China.


Johnson9009:
@junrushao @Hzfengsy @tqchen Please help to review, this is so little bug fix, thanks.

junrushao:
The PR looks good to me. Retriggering CI

junrushao (approved):
LGTM!

PR #14658, opened 2023-04-18
[Unity][Training] Training Utils: Loss functions, Optimizers, Trainer, SetupTrainer

Ubospica:
This PR introduces several utils to support training models in Relax IR, including:
- Loss functions. A library that generates relax functions representing common loss functions, such as MSELoss and CrossEntropyLoss.
- AppendLoss pass. You can specify one function (the backbone function) in the IRModule, and one loss function to AppendLoss. It will attach the loss function to the backbone function, and add the result function to the IRModule.
- Optimizers. A library that generates relax functions representing common optimizers, such as SGD and Adam.
- SetupTrainer. A pass that transform the backbone module (a IRModule containing only one function named "backbone") into a module that contains all functions used in the training process.
- Trainer. A runtime utility that stores parameters, states, etc, and helps to run the training IRModule.

Co-authored-by: Chaofan Lin <siriusneo@sjtu.edu.cn>


Ubospica:
Separated into many pull requests

PR #14659, opened 2023-04-18
[Unity] Update docs for operators

Ubospica:
This PR:
- updates docs for several operators,
- add an `1` default value for the `axis` parameter of batch_norm,
- rename module `tvm.relax.transform.legalize_ops.creation` to `tvm.relax.transform.legalize_ops.create`, which is align with `tvm.relax.op.create`, and
- fixes a previous upstream error in tests/python/relax/test_transform_legalize_ops_grad.py


yongwww (approved):

python/tvm/relax/op/nn/nn.py
```diff
@@ -659,14 +659,15 @@ def batch_norm(
     beta: Expr,
     moving_mean: Expr,
     moving_var: Expr,
-    axis: int,
+    axis: int = 1,
```
Hzfengsy, line 662: Is there any reason to add such default value?

Ubospica: Common batch norm operators takes the axis indexed 1 as the axis to calculate mean and variance, such as torch.nn.functional.batch_norm.

Ubospica: Our difference from torch is that we may have layouts like NHWC, but torch only accepts NCHW. Considering this maybe we can remove the default value 1.

PR #14665, opened 2023-04-18
[Unity] Add MultiSpan and attach it to CallNode produced by FuseOps

areusch:
I'm sending this PR to solicit feedback on this approach and am open to revisions or upstreaming parts of it into the `main` branch if needed.

This PR makes it possible for Relax pipelines to keep track of which original Relax operators were fused into which PackedFunc. To accomplish this, it introduces a new IR node: MultiSpan, which contains an Array<Span>. , it subclasses Span. Therefore, in a MultiSpan instance, the `Span` fields should be ignored. A more proper way to do this would be to rename Span to e.g. SingleSpan, but this would create so many conflicts that it would need to happen upstream first.

At the moment, this is only implemented in the FuseOps Relax pass. There will be limitations of this approach--for example, a `VarNode` referenced from multiple places in an IRModule will not be able to have distinct Spans. My interest here lies *only* in tracking this information in Relax functions where fusion and other "graph-level" transforms will occur, so I'm not sure I can justify doing something more elaborate such as introducing a new ObjectPath-based IR tracking facility. However, my hope is that the benefits are spelled out enough here to motivate a somewhat inferior approach as a starting point, and we can learn from this and improve as needed.

Please let me know your thoughts.

@junrushao @yelite @tqchen @masahi @jwfromm

### DIFF
--- a/include/tvm/ir/source_map.h
+++ b/include/tvm/ir/source_map.h
@@ -114,7 +114,7 @@ class SpanNode : public Object {
   }
 
   static constexpr const char* _type_key = "Span";
-  TVM_DECLARE_FINAL_OBJECT_INFO(SpanNode, Object);
+  TVM_DECLARE_BASE_OBJECT_INFO(SpanNode, Object);
 };
 
 class Span : public ObjectRef {
@@ -125,6 +125,31 @@ class Span : public ObjectRef {
   TVM_DLL Span Merge(const Span& other) const;
 
   TVM_DEFINE_OBJECT_REF_METHODS(Span, ObjectRef, SpanNode);
+};
+
+class MultiSpanNode : public SpanNode {
+ public:
+  Map<Span, ObjectRef> spans;
+
+  // override attr visitor
+  void VisitAttrs(AttrVisitor* v) { v->Visit("spans", &spans); }
+
+  bool SEqualReduce(const MultiSpanNode* other, SEqualReducer equal) const {
+    return equal(spans, other->spans);
+  }
+
+  static constexpr const char* _type_key = "MultiSpan";
+  TVM_DECLARE_FINAL_OBJECT_INFO(MultiSpanNode, SpanNode);
+};
+
+/*!
+ * \brief Attaches multiple disjoint Span to a given IR node.
+ */
+class MultiSpan : public Span {
+ public:
+  TVM_DLL MultiSpan(Map<Span, ObjectRef> spans);
+
+  TVM_DEFINE_NOTNULLABLE_OBJECT_REF_METHODS(MultiSpan, Span, MultiSpanNode);
 };
 
 /*! \brief A program source in any language.
--- a/include/tvm/ir/source_map.h
+++ b/include/tvm/ir/source_map.h
@@ -129,7 +129,7 @@ class Span : public ObjectRef {
 
 class MultiSpanNode : public SpanNode {
  public:
-  Map<Span, ObjectRef> spans;
+  Map<String, Span> spans;
 
   // override attr visitor
   void VisitAttrs(AttrVisitor* v) { v->Visit("spans", &spans); }
@@ -147,7 +147,7 @@ class MultiSpanNode : public SpanNode {
  */
 class MultiSpan : public Span {
  public:
-  TVM_DLL MultiSpan(Map<Span, ObjectRef> spans);
+  TVM_DLL MultiSpan(Map<String, Span> spans);
 
   TVM_DEFINE_NOTNULLABLE_OBJECT_REF_METHODS(MultiSpan, Span, MultiSpanNode);
 };
--- a/src/relax/transform/fuse_ops.cc
+++ b/src/relax/transform/fuse_ops.cc
@@ -490,7 +490,7 @@ class FunctionCreator : public ExprMutator {
    * \brief Build a MultiSpan that contains all Span from fused Call nodes.
    */
   MultiSpan BuildMultiSpanForCallSite() {
-    Map<Span, ObjectRef> spans;
+    Map<String, Span> spans;
     for (Binding binding : bindings_) {
       const auto* var_binding = binding.as<VarBindingNode>();
       if (!var_binding->value->IsInstance<CallNode>()) {
@@ -498,7 +498,9 @@ class FunctionCreator : public ExprMutator {
       }
       Call call = Downcast<Call>(var_binding->value);
       if (call->span.defined()) {
-        spans.Set(call->span, call);
+        std::stringstream ss;
+        ss << call->op << "@" << call.get();
+        spans.Set(String(ss.str()), call->span);
       }
     }
     return MultiSpan(spans);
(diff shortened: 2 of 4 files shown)


src/relax/transform/fuse_ops.cc
```diff
@@ -730,7 +751,8 @@ class OperatorFusor : public ExprMutator {
       //  - If this binding is an output binding, emit an output variable.
       //  - Otherwise, emit a dataflow variable.
       Var new_var;
-      Call call_to_emit = Call(gv, UpdateArgs(func_info.arguments_));
+      Call call_to_emit = Call(gv, UpdateArgs(func_info.arguments_), {}, {}, new_span);
+      LOG(INFO) << "Created fused call " << call_to_emit << " with span " << new_span;
```
Hzfengsy, line 755: Please remove debug info

Hzfengsy (commented):
Could you please add some test cases for it? Could you please check if this enhancement also applied to `FuseOpsByPattern`?

junrushao:
I'm not sure if we should go through this code path. First of all, the possibility of tracking span during deep optimizations is minimal, and I am not aware of a mature solution that is functioning and maintainable at the same time; Second, I am not sure if works well with existing infra, for example, error reporting in the BlockBuilder. Happy to change my mind though :-)

include/tvm/ir/source_map.h
```diff
@@ -127,6 +127,31 @@ class Span : public ObjectRef {
   TVM_DEFINE_OBJECT_REF_METHODS(Span, ObjectRef, SpanNode);
 };
 
+class MultiSpanNode : public SpanNode {
+ public:
+  Map<Span, ObjectRef> spans;
```
tqchen, line 132: I am not sure if it should be a dictionary, the span should be self-contained ideally and can be detached, so at best it should be an Array of Span to show the original tracking history

tqchen (changes requested):

include/tvm/ir/source_map.h
areusch: the thing is, in addition to the Span I need the name of the Relax op it was attached to. Happy to report less stuff here, but I do need more than just what's in the span. Hence the map.

areusch:
@Hzfengsy yes it should work for FuseOpsByPattern (I've updated CompositeFunctionAnnotator to make this work).

@junrushao I agree there are limitations here when it comes to Spans in deep schedule optimizations. However, it provides considerable introspection into the graph-level optimizations done by the compiler. I'm not attached to the way this information is tracked, but I do think it's reasonable to want to know which ops got fused how. I see this as a way to surface that information. 

What's the thinking behind building a tracking method that works for Spans (or similar provenance) inside scheduled compute before surfacing graph-level fusion information? At present, this method is essentially confined to Call() in the `main` function. Any such system to track in-compute spans should be a superset of this, and it seems like it'd make more sense to wait for a concrete use case before we design it. The concrete use case for this tracking is: I want to know which Relax ops were generated as part of a model import and which ones were fused together.

include/tvm/ir/source_map.h
tqchen: Putting the object itself is more of an anti-pattern. You can however put in more string information into the fused span. Ideally span should not refer back to the IR fragement, this can cause things like cylic deps and other bad issues

areusch: that's reasonable. i'll put the op then, unless someone has a more general suggestion.

tqchen: I think it should be something that is only related to span plus string in this particular case, since the main purpose of span is location tracking and nothing else

areusch: changed to Map<String,Span>

areusch:
@tqchen @Hzfengsy ptal

yelite:
Have we considered keeping this information in a contextual state outside the IR graph? For example, 
```
with TransformTrackingContext() as tracking_context:
    mod = FuseOpsByPattern(...)(mod)
    ...

tracking_context.get_fusion_map()
```

This will reduce the friction when we want to track more specific data during fusion or other passes, since we don't need to touch the IR definition to update what gets tracked.

areusch:
@yelite it seems like we'd still need to make passes aware of that out-of-band information though, right? so the friction would still be there, just it wouldn't be like a compile-time error.

yelite:
@areusch Yes, passes need to call methods of `TransformTrackingContext` to record these mappings. The difference is that it doesn't need to change files under `include/tvm/ir/`, which are depended by virtually everything.

areusch:
@yelite to be honest i don't follow your argument. could you include some more detail? yes--changing the IR requires changing include/tvm/ir. but i do not understand why you are proposing to add a second way to track spans that lives outside the IR. this could lead to conflicting duplication of span information. in the past we've intentionally make the opposite design choice to make it easier to keep such records in sync during IR transformations.

yelite:
> but i do not understand why you are proposing to add a second way to track spans that lives outside the IR. this could lead to conflicting duplication of span information

I think what we are discussing here is about tracking IR graph transformation, that is, many-to-many relationship between IR nodes before&after a pass. The imaginary API of `TransformTrackingContext` operates on IR nodes rather than their span. `get_fusion_map()` could return type `Map<Call, Array<Call>>`, a mapping from calls to the fused functions to original calls. Based on this map, we can also get the `MultiSpan` proposed in this PR if it's needed.

My point is that tracking this information outside IR graph will give us more flexibility in the future, because:
1. By staying out of the IR graph, we can represent the mapping in a more structured way. For example, we can keep track of a `Map<Call, Array<Call>>`. It contains more information than `Map<String, Span>`.
2. The definition of IR nodes is more concise and with less ambiguity. If we want to have good interoperability with other frameworks, it's crucial to keep our IR definition simple to make it easier to reach consensus. Letting each node potentially carry a `Map<String, Span>` in addition to a `Span` is like going to the opposite direction.
3. We can avoid some constraints from the structure of IR graph. In this specific task (track FuseOps), I think it's more natural to attach the mapping to the fused function rather than the call node to the fused function, as it's easier to go from the call node to the function being called, than the opposite. However it will be harder to implement if we store this information in the Function node, because multiple call nodes can call to the same fused function.

It's not clear in this PR how this MultiSpan will be consumed. I assume it's for recording more metrics during the benchmark. If that's the case, I don't think storing this information outside IR will make it harder to consume.

yelite:
@areusch Just want to clarify, I don't think the approach I shared is strictly better than the one implemented in this PR. I just wanted to share some thought on this, because I faced the same problem when benchmarking TVM on some models. I think which solution is better really depends on the use case. If it's for error reporting at some point of the transformation pipeline, being confined in IR graph is a much better solution.

> there's no forcing function to ensure that passes which run after this one also adjust the tracking information placed in the IR

Just for my own understanding, how does the MultiSpan approach ensure the span being passed along during transformation? Does it rely on the pass to copy span when creating a new function out of the old one?

areusch:
@yelite ah gotcha. the purpose here is to be able to map the results of the profiler to the frontend layers that are computed in a given PackedFunc. there are some other reasons you also might want this mapping, such as the TVM Explorer RFC. 

> Just for my own understanding, how does the MultiSpan approach ensure the span being passed along during transformation? Does it rely on the pass to copy span when creating a new function out of the old one?

Right--the main thing is that if a Span isn't propagated, at this level the solution is either to simply copy the span (if mutating `main()` and replacing/updating the CallNode or drop it if doing that doesn't make sense). So there's a clear set of rules that everyone in the repo can follow and those follow the IR. If tracking things out-of-band, then it kind of brings up the question of why a particular pass should have to care about the out-of-band tracking info.

PR #14667, opened 2023-04-19
[TFLite][Frontend] Support for quantized squared difference

ashutosh-arm:
In line with the TFLu implementation, this commit
enables mapping of quantized squared difference
to Relay's float32 squred difference.

cc @leandron


leandron (approved):
LGTM, thanks @ashutosh-arm

NicolaLancellotti (approved):
LGTM!

PR #14671, opened 2023-04-19
[METAL] Fix flaky memory issue due to racing

tqchen:
This PR aims to fix a relatively rare flaky race issuse when we are attempting to purge memory and the buffer is still in use in command buffer


MasterJH5574 (approved):
LGTM.

jinhongyii (approved):

MasterJH5574:
The CI issue seems to be irrelevant. So will go with merge.

PR #14675, opened 2023-04-20
[Unity] Use custom hash in `BlockBuilder` to avoid hashing large constants

masahi:
I found that if I use `BindParams` early in my BYOC workflow, some Relax passes become extremely slow. For example, `partition_for_cutlass` takes more than 40 sec on SD UNet if parameters are bound, while otherwise it only takes 1 sec.

This massive regression is due to the `unordered_map` used by `BlockBuilder`, https://github.com/apache/tvm/blob/unity/src/relax/ir/block_builder.cc#L381-L382

This map is keyed on `BaseFunc` and it uses structural equal / hash. So if large parameters are bound on each func, those params are hashed / compared repeatedly. 

This is solved by introducing a custom struct hash that ignores NDArray content.

@tqchen @Hzfengsy @slyubomirsky

### DIFF
--- a/src/meta_schedule/module_equality.cc
+++ b/src/meta_schedule/module_equality.cc
@@ -53,19 +53,6 @@ class SEqualHandlerIgnoreNDArray : public SEqualHandlerDefault {
   }
 };
 
-class SHashHandlerIgnoreNDArray : public SHashHandlerDefault {
- protected:
-  void DispatchSHash(const ObjectRef& object, bool map_free_vars) override {
-    ICHECK(object.defined());
-    if (auto ndarray = object.as<runtime::NDArray::Container>()) {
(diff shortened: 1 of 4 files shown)


tqchen:
@masahi can we try keep structural equality, and cache the structural hash(can ignore bind values)(and use that to check inequality? before resorting to structural equal). I feel that should handle most of the case

masahi:
> can we try keep structural equality, and cache the structural hash(can ignore bind values)(and use that to check inequality?

I haven't modified the equality part. Only the hash part is modified to ignore NDArray. Is there reason to prefer cache-based solution at the cost of more complexity?

tqchen:
Ah ok, this works, thanks masa.

PR #14651, opened 2023-04-18
[CI] Modify test cases to accommodate the CI upgrades

yongwww:
update with the images built from https://github.com/apache/tvm/pull/14635

Co-authored by @driazati

PR #14680, opened 2023-04-20
[CI] Downgrade ci_cpu llvm version back to 11

neildhickey:
The upgrade to llvm 15 caused failures in topi tests python/topi/python/test_topi_math.py::test_cast[float64-float16-llvm]. It seems to be a miscompilation due to opaque pointers. Downgrading while working to fix the issue.


ekalda:
@tvm-bot rerun

yongwww:
@neildhickey  thanks for the investigation! I saw this issue in my ci upgrade work (https://github.com/apache/tvm/pull/14651), it would better to fix this cast issue, looks the cast from f64 to f16 fails.

neildhickey:
@yongwww Yes I do want to fix the issue properly but put up this patch as a temporary work around until I am able to fix it properly

leandron (approved):
Thanks for looking into it @neildhickey

leandron:
Merging this to make the current `ci_cpu` image sources to work again.

PR #14686, opened 2023-04-20
[CMAKE] Update search pattern of config

tqchen:
This PR updates the search patter of config.cmake to CMAKE_BINARY_DIR. This behavior remains unchanged for TVM build.

The difference only happens when tvm was included as a subdirectory by another project via add_subdirectory.
In such cases, we will search the config.cmake on the root of the parent project rather than the tvm folder.


tqchen:
cc @junrushao

junrushao (approved):
That makes sense! Thanks for the change!

leandron (approved):
LGTM, thanks @tqchen @junrushao

PR #14692, opened 2023-04-21
[Unity][CUTLASS] Fix CUTLASS codegen for occasional variable name conflict

masahi:
I've hit a case where an invalid signature like this is generated:
```
void fused_relax_nn_conv2d_relax_add2_cutlass_(DLTensor* lv8, DLTensor* lv7, DLTensor* lv8, DLTensor* out0){
```

This happens because for some reasons different parameter `Var`s in the same function end up having the same `name_hint`.

I fixed this by generating a unique name for each parameter.

@cyx-6 @vinx13


cyx-6 (approved):
LGTM!

PR #14693, opened 2023-04-21
[TIR] Enhance TVMScript Buffer Slice Access

lightzhan-intellif:
This PR tries to fix a bug, which will case the error:
```
error: unsupported operand type(s) for +: 'BufferRegion' and 'BufferRegion'
 --> /workspace/code_newest/tvm/private_test/test_meta_programming.py:40:12
    |  
 40 |    B[0:5] = A[0:5] + B[0:5]
    |             ^^^^^^^^^^^^^^^
```
The problem will occur when it is a binary operation between two BufferSlice accesses, because the parser do not know how to deal with it.

This PR solves this by adding corresponding ops to `BufferRegion`. Any further suggestion is welcomed!

@junrushao @Hzfengsy @wrongtest-intellif


Hzfengsy:
cc @cyx-6

python/tvm/tir/stmt.py
```diff
@@ -536,6 +536,12 @@ class BufferRegion(Object, Scriptable):
     def __init__(self, buffer: Buffer, region: List[Range]):
         self.__init_handle_by_constructor__(_ffi_api.BufferRegion, buffer, region)  # type: ignore
 
+    def __eq__(self, other):
```
Hzfengsy, line 539: Not sure if it's safe. I tried doing similar things in Relax operators, but it may influence the behavior of `StructualEqual`, see  https://github.com/apache/tvm/blob/8569b9fe632b9488978bede30e169d457aafd8c9/python/tvm/relax/expr.py#L152

junrushao (commented):
I don't have problem with the fix per se, but was thinking if this ad-hoc fix is sufficient. To be more precise, vectorized buffer access being recognized as PrimExpr instead of BufferRegion.

ekalda:
Thanks @lightzhan-intellif for the fix! I think it would be good to also test that the scripted output of `VectorizeLoop` can be parsed as this is something that is currently broken and that this change should fix as well.

lightzhan-intellif:
> Thanks @lightzhan-intellif for the fix! I think it would be good to also test that the scripted output of `VectorizeLoop` can be parsed as this is something that is currently broken and that this change should fix as well.

Thanks for your suggestion, I will have a look.

Hzfengsy (approved):

PR #14694, opened 2023-04-21
unify search path approach to various libs

t-vi:
The FindCUDA looks for the various libraries in a more or less limited set of locations.
For nvrtc we currently have the most complete search, the others are lacking one or more locations.
This makes FindCUDA look in all typical linux lib dirs (e.g. Debian's CUDA install wants that).

Thank you for considering.


t-vi:
@junrushao I think you touched this last, and @tqchen before
If you have other suggestions for reviewers, I would also appriciate a hint.

tqchen (approved):

tqchen:
Thank you @t-vi !

t-vi:
Thank you, @tqchen !

PR #14700, opened 2023-04-22
[TensorIR][Schedule] Fix dtype bugs in Rfactor rule

Ubospica:
Currently test `test_reduction_rfactor_zero_dim` in file `tests/python/unittest/test_tir_schedule_rfactor.py` fails. 

The reason is there exists mixture of int32/int64 dtype in this testcase as well as in TIR RFactor schedule rule. This PR fixes these problems by unifying dtype to int64.

Besides, this PR adds a `tvm::tir::DefaultIndexType()` interface to determine the correct dtype.

`test_reduction_rfactor_zero_dim` is a regression test so we does not need more unit tests.

### DIFF
--- a/include/tvm/tir/buffer.h
+++ b/include/tvm/tir/buffer.h
@@ -34,6 +34,18 @@
 namespace tvm {
 namespace tir {
 
+#ifndef TVM_INDEX_DEFAULT_I64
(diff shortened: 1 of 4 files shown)


Ubospica:
Moved to #14705

PR #14702, opened 2023-04-22
[Unity] MetaScheduleApplyDatabase using workload from records

MasterJH5574:
This PR fixes an issue in MetaScheduleApplyDatabase, which uses the IRModule being queried as the base module of TIR Schedule. This will cause the inconsistency between the IRModule being scheduled and the trace of the record. For example, the trace of the record may have `get-block` instructions which tries to get block from a given name, which exists in the IRModule of the record while does not exist in the IRModule being queried (which is right the IRModule being scheduled).

Therefore, this PR adds a case discussion. When anchor-op equality is not applied, we create the TIR schedule from the IRModule of the record, so that we can prevent the issue above from happening.

This PR adds a unit test that can demonstrate the issue.


tqchen (approved):

junrushao (approved):
LGTM

PR #14705, opened 2023-04-23
[Unity][TensorIR][Schedule] Fix dtype bugs in Rfactor rule

Ubospica:
Currently test `test_reduction_rfactor_zero_dim` in file `tests/python/unittest/test_tir_schedule_rfactor.py` fails on unity branch.

The reason is there exists mixture of int32/int64 dtype in this testcase as well as in TIR RFactor schedule rule. 

This PR fixes these problems by unifying dtype to int64 as current unity branch sets  int64 as default dtype.

Besides, this PR adds a `tvm::tir::DefaultIndexType()` interface to determine the correct dtype.

`test_reduction_rfactor_zero_dim` is a regression test so we does not need more unit tests.

### DIFF
--- a/src/script/ir_builder/tir/ir.cc
+++ b/src/script/ir_builder/tir/ir.cc
@@ -45,7 +45,7 @@ Buffer BufferDecl(Array<PrimExpr> shape, DataType dtype, String buffer_name, Opt
     buffer_data = data.value();
   }
   if (!elem_offset.defined() && offset_factor) {
-    DataType shape_dtype = shape.empty() ? DataType::Int(32) : shape[0]->dtype;
+    DataType shape_dtype = shape.empty() ? tvm::tir::DefaultIndexType() : shape[0]->dtype;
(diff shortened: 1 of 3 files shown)


MasterJH5574:
For TIR primitive fixes we can go to the main branch.

Ubospica:
> For TIR primitive fixes we can go to the main branch.

The problem only occurs in unity branch, but not in main branch. The reason for that should be we require int64 as default dtype (and due to some limitations, in only a portion of places) only in unity, and this resulted in specific bugs in Unity.

tqchen:
@Ubospica would be great to followup on this. please send a PR to main

Ubospica:
The essence of the problem that this PR wants to solve is the inconsistency between the main and unity branches regarding dtype (int32 vs int64). However, current fix in the test cannot be unified in the main branch and unity branch. Since it is not a blocker, this PR will be closed temporarily and reopened after conprehensive consideration.

PR #14707, opened 2023-04-23
[Keras][Bugfix] fix a bug about alpha attribute in LeakyReLU which lead to passes conflict

jikechao:
The `alpha` attribute in LeakyReLU lacks exception checking. If the `alpha=nan` , the keras frontend can convert it to relay ir successful, but in the optimization stage, it will trigger an  unexpected crash and throws `TVMError: Observed 100 rewrite passes, possible conflicting passes?`

The relay IR is the following:
image

This patch fixes the bug!

The StackTrace:
```
    model = relay.build_module.create_executor("graph", mod, tvm.cpu(0), 'llvm', params).evaluate()  # crash
  File "/workplace/software/tvm/tvm/python/tvm/relay/backend/interpreter.py", line 171, in evaluate
    return self._make_executor()
  File "/workplace/software/tvm/tvm/python/tvm/relay/build_module.py", line 519, in _make_executor
    mod = build(self.mod, target=self.target)
  File "/workplace/software/tvm/tvm/python/tvm/relay/build_module.py", line 372, in build
    mod_name=mod_name,
  File "/workplace/software/tvm/tvm/python/tvm/relay/build_module.py", line 169, in build
    mod_name,
  File "/workplace/software/tvm/tvm/python/tvm/_ffi/_ctypes/packed_func.py", line 237, in __call__
    raise get_last_ffi_error()
tvm._ffi.base.TVMError: Traceback (most recent call last):
  14: TVMFuncCall
  13: tvm::relay::backend::RelayBuildModule::GetFunction(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, tvm::runtime::ObjectPtr<tvm::runtime::Object> const&)::{lambda(tvm::runtime::TVMArgs, tvm::runtime::TVMRetValue*)#3}::operator()(tvm::runtime::TVMArgs, tvm::runtime::TVMRetValue*) const
  12: tvm::relay::backend::RelayBuildModule::Build(tvm::IRModule, tvm::runtime::Array<tvm::Target, void> const&, tvm::Target const&, tvm::relay::Executor const&, tvm::relay::Runtime const&, tvm::WorkspaceMemoryPools const&, tvm::ConstantMemoryPools const&, tvm::runtime::String)
  11: tvm::relay::backend::RelayBuildModule::BuildRelay(tvm::IRModule, tvm::runtime::String const&)
  10: tvm::relay::backend::RelayBuildModule::OptimizeImpl(tvm::IRModule)
  9: tvm::transform::Pass::operator()(tvm::IRModule) const
  8: tvm::transform::Pass::operator()(tvm::IRModule, tvm::transform::PassContext const&) const
  7: tvm::transform::SequentialNode::operator()(tvm::IRModule, tvm::transform::PassContext const&) const
  6: tvm::transform::Pass::operator()(tvm::IRModule, tvm::transform::PassContext const&) const
  5: tvm::relay::transform::FunctionPassNode::operator()(tvm::IRModule, tvm::transform::PassContext const&) const
  4: tvm::runtime::PackedFuncObj::Extractor<tvm::runtime::PackedFuncSubObj<tvm::runtime::TypedPackedFunc<tvm::relay::Function (tvm::relay::Function, tvm::IRModule, tvm::transform::PassContext)>::AssignTypedLambda<tvm::relay::transform::SimplifyExpr()::$_0>(tvm::relay::transform::SimplifyExpr()::$_0)::{lambda(tvm::runtime::TVMArgs const&, tvm::runtime::TVMRetValue*)#1}> >::Call(tvm::runtime::PackedFuncObj const*, tvm::runtime::TVMArgs, tvm::runtime::TVMRetValue*)
  3: tvm::relay::SimplifyExpr(tvm::RelayExpr const&, tvm::IRModule const&)
  2: tvm::relay::RewritePatterns(tvm::runtime::Array<tvm::relay::DFPatternCallback, void>, tvm::RelayExpr, tvm::IRModule)
  1: tvm::relay::PatternRewriter::Rewrite(tvm::runtime::Array<tvm::relay::DFPatternCallback, void> const&, tvm::RelayExpr const&)
  0: _ZN3tvm7runtime6detail
  File "/workplace/software/tvm/tvm/src/relay/ir/dataflow_matcher.cc", line 829
TVMError: Observed 100 rewrite passes, possible conflicting passes?

```

### DIFF
--- a/tests/python/frontend/keras/test_forward.py
+++ b/tests/python/frontend/keras/test_forward.py
@@ -214,8 +214,8 @@ class TestKeras:
 
     def test_forward_activations_except(self, keras_mod):
         """
-        test invalid attribute alpha=None for LeakyReLU,
-        after version 2.3.1 in keras, checking was added to reject the invalid api call: 
+        test invalid attribute alpha=None for LeakyReLU.
+        after version 2.3.1 in keras, checking was added to reject the invalid api call:
         LeakyReLU(alpha=None), (issue: https://github.com/tensorflow/tensorflow/pull/47017)
         Thus, it's necessary to check the keras version to avoid crash at LeakyReLU(alpha=None)
         """
--- a/tests/python/frontend/keras/test_forward.py
+++ b/tests/python/frontend/keras/test_forward.py
@@ -215,9 +215,9 @@ class TestKeras:
     def test_forward_activations_except(self, keras_mod):
         """
         test invalid attribute alpha=None for LeakyReLU,
-        after version 2.3.1 in keras, keras add check to reject the valid api call: LeakyReLU(alpha=None),
-        (related issue: https://github.com/tensorflow/tensorflow/pull/47017)
-        Thus, it's necessary to check the keras version for avoiding crash when call LeakyReLU(alpha=None)
+        after version 2.3.1 in keras, checking was added to reject the invalid api call: 
+        LeakyReLU(alpha=None), (issue: https://github.com/tensorflow/tensorflow/pull/47017)
+        Thus, it's necessary to check the keras version to avoid crash at LeakyReLU(alpha=None)
         """
         if package_version.parse(keras_mod.__version__.split("-tf")[0]) <= package_version.parse(
             "2.3.1"
--- a/tests/python/frontend/keras/test_forward.py
+++ b/tests/python/frontend/keras/test_forward.py
@@ -765,6 +765,7 @@ if __name__ == "__main__":
         sut.test_forward_merge_dot(keras_mod=k)
         sut.test_forward_merge(keras_mod=k)
         sut.test_forward_activations(keras_mod=k)
+        sut.test_forward_activations_special(keras_mod=k)
         sut.test_forward_dense(keras_mod=k)
         sut.test_forward_permute(keras_mod=k)
         sut.test_forward_sequential(keras_mod=k)
--- a/tests/python/frontend/keras/test_forward.py
+++ b/tests/python/frontend/keras/test_forward.py
@@ -211,13 +211,18 @@ class TestKeras:
             keras_model = keras_mod.models.Model(data, x)
             verify_keras_frontend(keras_model)
             verify_keras_frontend(keras_model, need_transpose=False, layout="NHWC")
-        # test invalid attribute alpha=None for LeakyReLU
+
+    def test_forward_activations_special(self, keras_mod):
+        # test invalid attribute alpha=None for LeakyReLU,
+        # after version 2.3.1 in keras, keras add check to reject the valid api call:  LeakyReLU(alpha=None),
+        # (related issue: https://github.com/tensorflow/tensorflow/pull/47017)
+        # Thus, it's necessary to check the keras version for avoiding crash when call  LeakyReLU(alpha=None)
         if package_version.parse(keras_mod.__version__.split("-tf")[0]) <= package_version.parse(
             "2.3.1"
         ):
             data = keras_mod.layers.Input(shape=(2, 3, 4))
-            x = keras_mod.layers.LeakyReLU(alpha=None)(data)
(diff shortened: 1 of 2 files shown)


jikechao:
@yongwww Could your willing to review this PR? Many thanks!

Hzfengsy:
Please add a regression test for it.

tests/python/frontend/keras/test_forward.py
```diff
@@ -209,6 +211,15 @@ def test_forward_activations(self, keras_mod):
             keras_model = keras_mod.models.Model(data, x)
             verify_keras_frontend(keras_model)
             verify_keras_frontend(keras_model, need_transpose=False, layout="NHWC")
+        # test invalid attribute alpha=None for LeakyReLU
+        if package_version.parse(keras_mod.__version__.split("-tf")[0]) <= package_version.parse(
+            "2.3.1"
+        ):
```
echuraev, line 217: Why do you need such check on the package version?

jikechao: The following model is generated using keras (version 2.2) with the script. However, this model is not a valid model, which should be rejected in TVM front-end checking.  
```
input_shape = (2, 3, 4)
layer = keras.layers.LeakyReLU(alpha=None)  
y = layer(x)
model = models.Model(x, y)
```

image

However, in the latest keras. Keras found `keras.layers.LeakyReLU(alpha=None)` is not a valid usage and added an exception value checking to avoid generating the invalid model ([see this issue](https://github.com/tensorflow/tensorflow/pull/47017)). Thus, for triggering this crash, we should use keras <=2.3.1.

echuraev: I see your point, thank you. I'm not sure if it is the best option to have such condition in the test. Another variant, we can catch exception from TF if keras version is higher as 2.3.1 and if no exception was generated by TF then we will try to verify our frontend.
Current option is also LGTM. Let's discuss with other folks what is the best option in the test. @AndrewZhaoLuo, @leandron what do you think, should we check Keras version in test?

leandron: I think we can do that, as we are not very restrictive in terms of which Keras version we support. Some version validation is done so that we can update dependencies and keep CI running, but at some point, tests that only pass with ancient versions should be removed. That's the only case I see it is OK to add version checking within a test.

echuraev: And the reasonable question: how we can track such tests, which can be removed when the version starts to be ancient? Should be a github issue created or specific comment in the code can be used?

AndrewZhaoLuo: I personally think a comment will suffice for this. 

Github issues should be more active problems imo. Having one hang out for a long time is bad (though I am guilty of this :P).

tests/python/frontend/keras/test_forward.py
```diff
@@ -209,6 +211,15 @@ def test_forward_activations(self, keras_mod):
             keras_model = keras_mod.models.Model(data, x)
             verify_keras_frontend(keras_model)
             verify_keras_frontend(keras_model, need_transpose=False, layout="NHWC")
+        # test invalid attribute alpha=None for LeakyReLU
```
AndrewZhaoLuo, line 214: I think it's probably cleaner to just spin this out into it's own test.

AndrewZhaoLuo (commented):
I think it looks fine. Thanks for fixing this. 

Please add a comment to indicate to people why things are gated on version revision.

echuraev (approved):
LGTM. Thank you for this PR and fixing this issue.

echuraev:
@tvm-bot rerun

echuraev:
@tvm-bot rerun

echuraev:
@tvm-bot rerun

echuraev:
@tvm-bot rerun

PR #14712, opened 2023-04-24
[release] Bump version numbers to v0.12.0

ysh329:
This bumps all the **version numbers on v0.12.0 branch** as correct **v0.12.0**.

cc @driazati @Johnson9009 @junrushao

### DIFF
--- a/conda/recipe/meta.yaml
+++ b/conda/recipe/meta.yaml
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-{% set version = '0.12.dev0' %}
+{% set version = '0.12.0' %}
 {% set pkg_name = 'tvm' %}
 {% set cuda_tag = cuda_version | replace('.', '') %} # [cuda]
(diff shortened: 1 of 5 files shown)


ysh329:
So strange, log shows conflict as below:
image

leandron:
> So strange, log shows conflict as below:
> image

I think this has to do with the fact that CI will try to merge your current patch with "top of the repository", and as the top of the repository is now fixed to `0.13.dev0`, it diverges from what you propose here.

So I don't think that there is anything wrong with your patch, we probably just need to fix CI somewhere. @driazati can you help?

driazati:
Yeah this is a CI problem, I think we should just disable that behavior for releases, I filed a PR to do it: #14714. Once that is merged into v0.12.0, we can rebase this PR and it should use the updated CI configs

ysh329:
@tvm-bot rerun

ysh329:
I rebase failed and re PR with #14738

PR #14722, opened 2023-04-26
[Unity] Cache ndarray-cache.json

DustinBrett:
Based on a [comment](https://github.com/mlc-ai/web-llm/issues/57#issuecomment-1518639252) from @tqchen I've gone ahead and made a PR for my [suggestion/question](https://github.com/mlc-ai/web-llm/issues/57) of caching `ndarray-cache.json` before caching the shards. From my point of view this change means subsequent loads will pull 100% from CacheStorage, avoiding any external requests. Also I think this might make more sense because if your `ndarrayCacheUrl` is not pinned to a specific commit then it could perhaps become out of sync with the cached shards.


DustinBrett:
Adding CC comment for reviewers @quic-sanirudh & @tqchen as the contributing guidelines suggest I do.

tqchen (approved):

tqchen:
Thank you @DustinBrett

PR #14724, opened 2023-04-26
[bugfix][Relay] fix wrong documentation and add layout checking in keras frontend

jikechao:
This pr fixes 2 questions.
1. The wrong documentation about attribute `layout` in `relay.from_keras()`. 
     By analysis the source code [Here](https://github.com/jikechao/tvm/blob/main_abc/python/tvm/relay/frontend/keras.py#:~:text=assert%20layout%20in,NHWC%20or%20NDHWC%22), we can find that `input_layout` has 4 valid values (e.g, 'NWC', 'NCHW', NHWC and NDHWC"). However, in the [source code comments](https://github.com/jikechao/tvm/blob/main_abc/python/tvm/relay/frontend/keras.py#:~:text=layout%3A%20str,better%20across%20TVM.) and [official documentation](https://tvm.apache.org/docs/reference/api/python/relay/frontend.html?highlight=from_keras#tvm.relay.frontend.from_keras:~:text=layout%20(str)%20%E2%80%93%20One%20of%20%E2%80%98NCHW%E2%80%99%20or%20%E2%80%98NHWC%E2%80%99%2C%20indicates%20how%20data%20should%20be%20arranged%20in%20the%20output%20model.%20Default%20layout%20is%20%E2%80%98NCHW%E2%80%99%20as%20it%20in%20general%20performs%20better%20across%20TVM.) , it says only 'nchw' and 'nhwc' are supported. They are conflicted. 

2. Add exception checking to reject the mismatched `layout` and `input_shape` immediately.
    For the following script: relay can convert the keras model to RelayIR successful, It crashed when executed the last statements **(e.g., `model = model.evaluate()`) in the script and throw:   "Check failed: (!axes.defined() || static_cast<int>(axes.size()) == ndim) is false: Dimension mismatch: axes has 4 elements, but data.ndim = 3".**  This exception checking is too late, and the crash message will increase the difficulty of debugging. Thus, I added an exception checking in keras importer and give a more clear exception message.

```
import tvm
import tvm.relay as relay
import numpy as np
from tensorflow import keras


input_shape = (1, 2, 3)   
# the dimesion=3 and the default layout = 'nchw' which constructs an invalid test case 
# but it passes the frontend check and generates a RealyIR.

layer = keras.layers.Flatten()

y = layer(x)
model = models.Model(x, y)

mod, params = relay.frontend.from_keras(model, shape_dict)    # convert to RelayIR successfully.
print(mod)

    model = relay.build_module.create_executor("vm", mod, tvm.cpu(0), 'llvm', params)
    model = model.evaluate()    # crash here!!!
```

cc @echuraev


jikechao:
This PR led to many test cases crashing. Thus, I'm confused if this fix is valid. Wish for your comments.

jikechao:
@tvm-bot rerun

PR #14730, opened 2023-04-27
[Unity] Add options to MS tuning pass to enable more fine-grained tuning

masahi:
Currently, `MetaScheduleTuneIRMod` pass only has `max_trials_global` param to control the duration of tuning. Now that we have good BYOC support and `DefaultGPUSchedule` pass, we should add more options to allow more fine-grained tuning. 

For example, in SD UNet, I want to tune only layer and group norm, for small number of trials each. And I don't want to spend any seconds "tuning" trivial ops such as `cast`, `concat` etc. 

To realize these goals, I'm adding `max_trials_per_task` and `op_names` params. The latter is used to select which op to tune. For SD UNet I'd use it like `op_names=["layer_norm", "group_norm"]`.

@sunggg @zxybazh

### DIFF
--- a/python/tvm/meta_schedule/relax_integration.py
+++ b/python/tvm/meta_schedule/relax_integration.py
@@ -194,6 +194,8 @@ def tune_relax(
         The maximum number of trials to run
     max_trials_per_task : Optional[int]
         The maximum number of trials to run for each task
+    op_names: Optional[List[str]]
+        A list of operator names to specify which op to tune
     num_trials_per_iter : int
         The number of trials to run per iteration
     builder : BuilderType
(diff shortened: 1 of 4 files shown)


python/tvm/meta_schedule/relax_integration.py
```diff
@@ -193,6 +194,8 @@ def tune_relax(
         The maximum number of trials to run
     max_trials_per_task : Optional[int]
         The maximum number of trials to run for each task
+    op_names: Optional[List[str]]
+        A list of operator names to specify which op to tune
```
zxybazh, line 198: Maybe we can mention that by default is None and means all op would be tuned.

zxybazh (approved):
LGTM, minor comment on docs.

python/tvm/meta_schedule/relax_integration.py
masahi: Done

PR #14731, opened 2023-04-27
[Unity][CUTLASS] Add layer norm support

masahi:
Layer norm in SD UNet is expensive and so far we've been using MS tuning for it. CUTLASS has a layer norm kernel which I found to be more than 2x faster than a MS-tuned one on SD workloads.  

@cyx-6 @yelite


cyx-6 (approved):
LGTM

yelite (approved):
This looks great. Thanks for adding this!

PR #14732, opened 2023-04-27
[MetaSchedule] Skip custom_rule in case of None

apeskov:
Custom rule mechanics allows to have "schedule_rule" attribute equals "None". This value will be ignored by `ApplyCustomRuleNode`. But also this block will be ignore by all other rules. It happens because of improper check in `PostOrderApplyNode`.

This patch allows to use regular rules in case of `schedule_rule=="None"`.


Hzfengsy:
cc @junrushao

apeskov:
@junrushao could you please take a look. In your previous patch [#13195](https://github.com/apache/tvm/pull/13195) with refactoring of custom rules you already had a [discussion](https://github.com/apache/tvm/pull/13195/files#r1018952166) with @masahi about auto tensorization issue. 

But there is an another aspect. In case of `schedule_rule == "None"`, ApplyCustomRule and `sch->Unannotate` will not be called. So block will not be processed by any schedule rule never. I'm not sure if it was in the design.

junrushao:
I vaguely recall that there is a reason that we use `custom_rule: None`, which means we wanted to design space generator to intentionally skip a block. My assumption was that if we have a good usecase if we want the generator not to skip, then perhaps we could unannotate the block first - but I'm happy to change my mind if the situation above has been taken good consideration

PR #14734, opened 2023-04-27
[Unity] Add system lib build option to relax

tqchen:
This PR add system lib build option to relax
also adds an util that creates an archived static
library in compilation


junrushao (approved):
LGTM! BTW, the name "system lib" has been around for a while, but less known to introductory users, so shall we clarify why it is named this way?

tqchen:
Good suggestion, that is a great topic for discussion.

MasterJH5574 (approved):

Issue #14661, opened 2023-04-18
[Bug][ONNX] cann't parse strides for LpPool


jikechao:
For the below onnx model with LpPool, we can see that no `strides` attribute in LpPool. However, tvm frontend try to parse this attribute, and throw an error: `KeyError: 'strides'`. By analysis, the crash is caused by the [node.attribute] https://github.com/apache/tvm/blob/main/python/tvm/relay/frontend/onnx.py#:~:text=attr%20%3D%20self._parse_attr(node.attribute) cannot get the `strides` attribute and lead to a crash when execute `attr["strides"]` in this [line](https://github.com/apache/tvm/blob/main/python/tvm/relay/frontend/onnx.py#:~:text=data%2C-,attr%5B%22strides%22%5D%2C,-attr%5B%22kernel_shape).

### model structure
![image](https://user-images.githubusercontent.com/29506758/234762020-7ebaa8f0-2798-4f65-9b2f-9ca46a96c8d0.png)

### Crash message
```
Traceback (most recent call last):
  File "test.py", line 7, in <module>
    irmod, params = relay.frontend.from_onnx(model, {'x': (1, 3, 32, 32)}, freeze_params=True)
  File "/workplace/software/tvm/tvm/python/tvm/relay/frontend/onnx.py", line 7347, in from_onnx
    mod, params = g.from_onnx(graph, opset)
  File "/workplace/software/tvm/tvm/python/tvm/relay/frontend/onnx.py", line 6964, in from_onnx
    self._construct_nodes(graph)
  File "/workplace/software/tvm/tvm/python/tvm/relay/frontend/onnx.py", line 7079, in _construct_nodes
    op = self._convert_operator(op_name, inputs, attr, self.opset)
  File "/workplace/software/tvm/tvm/python/tvm/relay/frontend/onnx.py", line 7205, in _convert_operator
    sym = convert_map[op_name](inputs, attrs, self._params)
  File "/workplace/software/tvm/tvm/python/tvm/relay/frontend/onnx.py", line 1999, in _impl_v1
    attr["strides"],
KeyError: 'strides'
```

### Steps to reproduce
1. download the [lppool_model.onnx](https://github.com/jikechao/onnx_models/blob/main/lppool_model.onnx)
2. execute the script.
```
import onnx
from tvm import relay

onnx_model_path = "lppool_model.onnx"
model = onnx.load(onnx_model_path)
[...]

PR #14740, opened 2023-04-28
[Relay][Bugfix] Fix stride in LpPool for default

jikechao:
This Pr tries to fix the bug in https://github.com/apache/tvm/issues/14661

### DIFF
--- a/python/tvm/relay/frontend/onnx.py
+++ b/python/tvm/relay/frontend/onnx.py
@@ -1996,7 +1996,7 @@ class LpPool(OnnxOpConverter):
                 # one will need to run dynamic_to_static on this model after import
                 data = autopad(
                     data,
-                    attr.get('strides', [1] * num_spatial_dims),
+                    attr.get("strides", [1] * num_spatial_dims),
                     attr["kernel_shape"],
                     [1] * ndim,
                     mode=attr["auto_pad"],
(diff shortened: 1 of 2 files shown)


jikechao:
@tvm-bot rerun

yongwww (approved):

python/tvm/relay/frontend/onnx.py
```diff
@@ -1988,14 +1988,15 @@ def _impl_v1(cls, inputs, attr, params):
...
+        num_spatial_dims = ndim - 2
         if "auto_pad" in attr:
             attr["auto_pad"] = attr["auto_pad"].decode("utf-8")
             if attr["auto_pad"] in ("SAME_UPPER", "SAME_LOWER"):
                 # Warning: LpPool does not yet support dynamic shapes,
                 # one will need to run dynamic_to_static on this model after import
                 data = autopad(
                     data,
-                    attr["strides"],
+                    attr.get("strides", [1] * num_spatial_dims),
```
AndrewZhaoLuo, line 1999: `_impl_v1` implies a handling of the v1 version of the opset. This is technically 11+. 

I think the pattern here is this is ok, as creating a `_impl_v11` implies significant interface change. 

That being said, please add a comment here that this is meant to handle the field being optional for opsets 11+

AndrewZhaoLuo (approved):
LGTM, one nit.

PR #14741, opened 2023-04-28
[TIR] Keep block annotations from tensorization

apeskov:
TIR TensorIntrin implementation may contains some annotation. Current logic will skip all existing annotations. But in some cases it may be helpful to transfer annotations to resulting schedule.


Hzfengsy (approved):
LGTM. Please fix the CI :)

src/tir/schedule/primitive/blockize_tensorize.cc
```diff
@@ -604,6 +604,15 @@ void Tensorize(ScheduleState self, const StmtSRef& sref, const TensorIntrin& int
     BlockNode* block = block_realize.CopyOnWrite()->block.CopyOnWrite();
     block->body = impl_block->body;
     block->match_buffers = std::move(match_buffer_regions);
+    for (auto [key, val] : impl_block->annotations) {
```
echuraev, line 607: nit: Can we use `auto&` instead of `auto`?

apeskov: updated

echuraev (approved):

PR #14748, opened 2023-05-01
【PaddlePaddle Hackathon 4 No.232】add attribute support for assign/conv2d/conv2d_transpose/scale

siriuslay:
add "output" attributes for assign op.
add "data_format" attribute for conv2d and conv2d_transpose op. 
add "act" attribute for scale op.

### DIFF
--- a/python/tvm/relay/frontend/paddlepaddle.py
+++ b/python/tvm/relay/frontend/paddlepaddle.py
@@ -178,8 +178,12 @@ def convert_argsort(g, op, block):
 def convert_assign(g, op, block):
     """Operator converter for assign."""
 
-    out = g.get_node(op.input("X")[0])
-    g.add_node(op.output("Out")[0], out)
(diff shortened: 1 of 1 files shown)


python/tvm/relay/frontend/paddlepaddle.py
```diff
@@ -1935,6 +1945,7 @@ def convert_scale(g, op, block):
     scale = op.attr("scale")
     bias = op.attr("bias")
     bias_after_scale = op.attr("bias_after_scale")
+    act = op.attr("act")
```
jiangjiajun, line 1948: There's no "act" attribute in operator "scale". While we use `paddle.scale(x, act="relu")`, this will generate 2 op, include "scale" and "relu"

jiangjiajun (changes requested):
Please add test code for this pull request

PR #14751, opened 2023-05-02
[CUDA] Fixed the call of the min function in the schedule for cuda

valmat07:
When using a scatter layer for cuda target, it gives the following error:

`ValueError: Cannot use and / or / not operator to Expr, hint: use tvm.tir.all / tvm.tir.any instead`

This is due to the call to the min function here:
https://github.com/apache/tvm/blob/d32dea800baeda14d144d0524a5da435c9cb160b/python/tvm/topi/cuda/scatter.py#L228-L231

This PR fixes this bug

### DIFF
--- a/tests/python/relay/test_any.py
+++ b/tests/python/relay/test_any.py
@@ -2149,6 +2149,29 @@ def test_scatter_nd():
 
 
 @tvm.testing.uses_gpu
+def test_scatter_nd_any_updates():
+    def verify_scatter_nd_any_updates(data_np, indices_np, updates_np, ref_res):
+        indices_shape = (2, relay.Any())
+        updates_shape = (relay.Any(), relay.Any())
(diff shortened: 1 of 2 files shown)


echuraev (commented):
Could you please add an unit test?

tests/python/relay/test_any.py
```diff
@@ -59,6 +59,7 @@ def check_result(
                 continue
             if kind == "debug" and (only_vm or dev.device_type != tvm.cpu().device_type):
                 continue
+            print(tgt)
```
echuraev, line 62: Do you really need this print?

echuraev (approved):
In general LGTM. Please remove redundant print.

echuraev:
cc: @junrushao, @masahi

echuraev:
@tvm-bot rerun

echuraev:
@tvm-bot rerun

PR #14752, opened 2023-05-02
[Unity] Limit number of characters in logger names

jwfromm:
This little one liner prevents metaschedule from creating logs with names so large they are invalid for the host operating system. Here we limit to 100 characters for the function name, which should be plenty to make it clear to the user which function corresponds to the log.


zxybazh (approved):
Thanks for noticing issue and the quick fix!

PR #14757, opened 2023-05-03
[Unity][Training] Categorical cross entropy loss

ndemashov:
This PR adds support of categorical cross entropy loss

### DIFF
--- a/python/tvm/relax/training/loss.py
+++ b/python/tvm/relax/training/loss.py
@@ -27,7 +27,7 @@ from typing_extensions import Literal
 from ..block_builder import BlockBuilder
 from ..expr import Expr, Var, Function, StructInfo
 
-from ..op import abs, sum, mean, subtract, multiply
+from ..op import abs, sum, mean, subtract, multiply, reshape, argmax
 from ..op.nn import log_softmax, nll_loss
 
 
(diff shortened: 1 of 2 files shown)


python/tvm/relax/training/loss.py
```diff
@@ -290,3 +290,94 @@ def __call__(
...
+
+        predictions = _create_param_var(predictions, "predictions")
+        targets = _create_param_var(targets, "targets")
+
+        arg_list = [predictions, targets]
+        if weights:
+            weights = _create_param_var(weights, "weights")
+            arg_list.append(weights)
+
+        if self.ignore_index >= 0:
```
echuraev, line 357: Could you please add a clarification commentary why you have two implementations which are depends on `ignore_index`. I think it might be useful in the future.

echuraev (commented):
Thank you for the PR. I added a minor comment. @Ubospica, @SiriusNEO could you please review this PR?

python/tvm/relax/training/loss.py
```diff
@@ -290,3 +290,94 @@ def __call__(
...
+    ) -> Function:
+        """Get the relax function of CategoricalCrossEntropyLoss. If the parameters are
+        struct info, it will create corresponding variables.
+
+        Parameters
+        ----------
+        predictions : Union[Var, StructInfo]
+            The predictions of the model in the calculation of loss.
+
+        targets : Union[Var, StructInfo]
```
echuraev, line 336: Should we check the type of the targets? As far as I remember, only `int64` data type is applicable for `targets`.

SiriusNEO (approved):
Overall looks great, just a few comments.

python/tvm/relax/training/loss.py
```diff
@@ -290,3 +290,94 @@ def __call__(
...
+                        reshape(argmax(targets, axis=1), shape=(targets.struct_info.shape[0],))
+                    )
+                    loss = bb.emit_output(
+                        nll_loss(logits, targets, weights, self._reduction, self.ignore_index)
+                    )
+                bb.emit_func_output(loss)
+        else:
+            with bb.function(self._loss_name, arg_list):
+                with bb.dataflow():
+                    logits = bb.emit(log_softmax(predictions))
```
SiriusNEO, line 371: Would it be good if we lift the common part like
```
with bb.function(self._loss_name, arg_list):
    with bb.dataflow():
        logits = bb.emit(log_softmax(predictions))
        if self.ignore_index >= 0:
            ...
        else:
            ...
```

SiriusNEO (approved):
Overall looks great, just a few comments.

python/tvm/relax/training/loss.py
ndemashov: Done, thank you.

python/tvm/relax/training/loss.py
ndemashov: Good idea, thank you. I added a check like it's done in the nllloss function.

python/tvm/relax/training/loss.py
ndemashov: Done, thank you.

python/tvm/relax/training/loss.py
```diff
@@ -290,3 +290,91 @@ def __call__(
...
+        weights : Optional[Union[Var, StructInfo]]
+            a manual rescaling weight given to each class. It has to be a Tensor of size C.
+
+        Returns
+        -------
+        The relax function of CategoricalCrossEntropyLoss with the loss name as its global symbol.
+        """
+
+        if targets.dtype != 'int64':
+            raise TypeError(f'Dtype of targets expected to be int/uint. However, the dtype of targets is {targets.dtype}')
```
echuraev, line 348: It is written that type should be `int/uint` but you check only for `int`

ndemashov: Fixed, thank you.

echuraev (approved):
LGTM

echuraev:
@tvm-bot rerun

echuraev:
@tvm-bot rerun

echuraev:
@tvm-bot rerun

PR #14766, opened 2023-05-04
[TIR][Schedule] Improve blockize to support blockizing multiple blocks

multiverstack-intellif:
This is to improve blockize primitive to support blockizing multiple blocks into a nested block. This helps when we want to schedule a big part of IR as a whole block.
Some possible simplification is performed during the blockize process, to keep TIR not that complex, so some other test case may need adjustment.

cc @Hzfengsy @junrushao @wrongtest-intellif

### DIFF
--- a/python/tvm/tir/schedule/schedule.py
+++ b/python/tvm/tir/schedule/schedule.py
@@ -2694,7 +2694,7 @@ class Schedule(Object):
     def blockize(
         self, target: Union[LoopRV, List[BlockRV]], preserve_unit_iters: bool = True
     ) -> BlockRV:
-        """Convert the subtree rooted at a specific loop into a block.
+        """Convert multiple blocks or the subtree rooted at a specific loop into a block.
 
         Parameters
         ----------
--- a/python/tvm/tir/schedule/schedule.py
+++ b/python/tvm/tir/schedule/schedule.py
@@ -2691,13 +2691,15 @@ class Schedule(Object):
     ########## Schedule: Blockize & Tensorize ##########
 
     @type_checked
-    def blockize(self, loop: LoopRV, preserve_unit_iters: bool = True) -> BlockRV:
+    def blockize(
+        self, target: Union[LoopRV, List[BlockRV]], preserve_unit_iters: bool = True
+    ) -> BlockRV:
         """Convert the subtree rooted at a specific loop into a block.
 
         Parameters
         ----------
-        loop : LoopRV
-            The root of the subtree.
+        target : LoopRV or List[BlockRV]
+            The root of the subtree or the specified blocks.
         preserve_unit_iters : bool
             Whether or not to preserve unit iterators in block bindings
 
@@ -2764,7 +2766,7 @@ class Schedule(Object):
         block are divisible by the subspace represented by the loops starting at the given loop.
         """
 
-        return _ffi_api.ScheduleBlockize(self, loop, preserve_unit_iters)  # type: ignore # pylint: disable=no-member
+        return _ffi_api.ScheduleBlockize(self, target, preserve_unit_iters)  # type: ignore # pylint: disable=no-member
 
     @type_checked
     def tensorize(
--- a/src/tir/schedule/primitive/blockize_tensorize.cc
+++ b/src/tir/schedule/primitive/blockize_tensorize.cc
@@ -245,7 +245,12 @@ Map<Var, PrimExpr> DeriveBlockBinding(const Array<IterVar>& iter_vars,          
     if (is_one(inner_mark->extent)) {
       // Skip inner var when extent is 1
       // substitution
-      sub = is_one(outer_mark->extent) ? outer_mark->extent - 1 : outer_iter;
+      if (is_one(outer_mark->extent) && !preserve_unit_iters) {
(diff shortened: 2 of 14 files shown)


tests/python/unittest/test_meta_schedule_schedule_rule_mlt_tc.py
```diff
@@ -988,15 +988,15 @@ def conv2d_1x1_0(inputs: T.Buffer((1, 16, 16, 64), "float16"), weight: T.Buffer(
                                 v2, v3 = T.axis.remap("SS", [ax2_1, ax3])
                                 v4_o = T.axis.spatial(1, 0)
                                 v5_o = T.axis.spatial(1, 0)
-                                T.reads(conv2d_nhwc_reindex_shared_wmma_accumulator[v0, v1, v2, v3, 0:16, 0:16])
-                                T.writes(conv2d_nhwc_reindex_shared[v0, v1, v2, v3, 0:16, 0:16])
+                                T.reads(conv2d_nhwc_reindex_shared_wmma_accumulator[v0, v1, 0, 0, 0:16, 0:16])
```
Hzfengsy, line 991: I think this kind of chengs may break current tensorcore support IMO. cc @junrushao @vinx13

vinx13: yes these unit iters can be important in some workloads (1x1 conv, or batch size = 1 gemm)

multiverstack-intellif: Thanks for the reminder. Do you mean tensorcore needs read/write ranges to be in a specific form? What kind of requirement does it like?

vinx13: it doesn't require specific forms of reads/writes. however, it needs iter vars with extent = 1 to be kept for dimensions used for padding

multiverstack-intellif: Thanks. Do you mean preserving the for loop with unit extend that `preserve_unit_iters` ensure? The for loop in this case is not changed, only the read/write range is changed here. Does it require all the iter vars be kept anywhere in reads/writes and expressions even if it's always constant 0?

python/tvm/tir/schedule/schedule.py
```diff
@@ -2691,13 +2691,15 @@ def after_set_dtype(
     ########## Schedule: Blockize & Tensorize ##########
 
     @type_checked
-    def blockize(self, loop: LoopRV, preserve_unit_iters: bool = True) -> BlockRV:
+    def blockize(
+        self, target: Union[LoopRV, List[BlockRV]], preserve_unit_iters: bool = True
+    ) -> BlockRV:
         """Convert the subtree rooted at a specific loop into a block.
```
vinx13, line 2697: need to update doc here

src/tir/schedule/primitive/blockize_tensorize.cc
```diff
@@ -152,7 +152,8 @@ Array<Array<arith::IterMark>> SubspaceDivide(const BlockRealize& realize,
                                              const StmtSRef& block_sref,  //
```
vinx13, line 152: update function doc for the new param

tests/python/unittest/test_meta_schedule_schedule_rule_mlt_tc.py
vinx13: I verified these workloads and confirmed this PR doesn't break auto tensorization. Keeping these unit iters preserve more semantics of original IR, but I don't have strong opinion

src/tir/schedule/primitive/blockize_tensorize.cc
multiverstack-intellif: Thanks, updated now.

python/tvm/tir/schedule/schedule.py
multiverstack-intellif: Thanks, updated now.

src/tir/schedule/primitive/blockize_tensorize.cc
```diff
@@ -219,30 +227,43 @@ Map<Var, PrimExpr> DeriveBlockBinding(const Array<IterVar>& iter_vars,
...
-    // substitution
     PrimExpr sub{nullptr};
-    if (is_one(outer_mark->extent)) {
-      sub = inner_iter->var;
+    if (is_one(inner_mark->extent)) {
+      // Skip inner var when extent is 1
+      // substitution
+      if (is_one(outer_mark->extent) && !preserve_unit_iters) {
+        // Simplify outer if not preserve_unit_iters
+        sub = outer_mark->extent - 1;
```
vinx13, line 250: nit: is this equivalent to `make_zero(outer_mark.dtype()`? I feel this is more readable

vinx13 (approved):

src/tir/schedule/primitive/blockize_tensorize.cc
multiverstack-intellif: Yes, it's more direct, thanks.

PR #14767, opened 2023-05-04
[OpenCLML] Transposed convolution support and other fixes

srkreddy1238:
Added support for transposed convolution.
Epsilon support for batchnorm op added - part of v3.0. CLML version query bug fixed.


src/relay/backend/contrib/clml/codegen.cc
```diff
@@ -189,22 +192,27 @@ class CLMLJSONSerializer : public backend::contrib::JSONSerializer {
...
+      } else {
+        name = "conv2d";
+        ICHECK(conv_attr->kernel_layout == "OIHW")
+            << "Kernel layout must be OHWI, has the module been pre-processed correctly?";
+      }
+    } else if (backend::IsOp(nodes.conv, "nn.conv2d_transpose")) {
+      name = "conv2d_transpose";
+      const auto* conv_transpose_attr = nodes.conv->attrs.as<Conv2DTransposeAttrs>();
+      ICHECK(conv_transpose_attr);
+      ICHECK(conv_transpose_attr->kernel_layout == "OIHW")
```
echuraev, line 215: Is it necessary to check `kernel_layout` here? You have already done it in `python/tvm/relay/op/contrib/clml.py`. If it is not necessary, then probably the same checks can be removed for convolution.

python/tvm/relay/op/contrib/clml.py
```diff
@@ -37,7 +37,7 @@
 def clml_sdk_version():
     """Utility function to get clml version version"""
 
-    return tvm.support.libinfo().get("TVM_CLML_VERSION", 2)
+    return int(tvm.support.libinfo().get("CLML_VERSION_MAJOR", 2))
```
echuraev, line 40: Why did you change `TVM_CLML_VERSION` on `CLML_VERSION_MAJOR`?

echuraev (commented):
Thank you for your PR. In general LGTM. Some minor comments.

src/relay/backend/contrib/clml/codegen.cc
srkreddy1238: I will remove this check in python frontend. Codegen is the final component to make sure we don't generate incompatible code.

python/tvm/relay/op/contrib/clml.py
srkreddy1238: The version check from libinfo was broken earlier and not functional. ```TVM_CLML_VERSION``` was compiler flag and cmake option is ```CLML_VERSION_MAJOR```. We can maintain same for both.

echuraev (approved):
LGTM. Thanks

PR #14768, opened 2023-05-04
[Unity][CI] Use the upgraded images

yongwww:
Use the newly upgraded images for unity


yongwww:
The Change of [unity_jenkinsfile.groovy](https://github.com/apache/tvm/pull/14768/files#diff-12632478e80c5aba61acddb58d83cc0dbcdf5ba348251394efa6f6b1caa6d5b7) needs to be merged first to take effect, which is why the CI for this PR is failing (CI is still using old container, and llvm-15 is not there). I have tested the changes with the new containers in my local environment. It should be okay to land this change. I didn't update the link container in this pr, it requires updates on pylintrc and some python files, which could lead to duplication with the main branch. It seems that the upcoming merge from main will need to address the pylint updates accordingly. cc: @tqchen @driazati @slyubomirsky

tqchen (approved):

PR #13307, opened 2022-11-07
[RUNTIME][ALIGNMENT] Configurable kAllocAlignment if needed

srkreddy1238:
Not all plarforms 64bit aligned allocations. Platforms with 32bit alignment fail to support set_input_zero_copy even though the ndarray is allocated by the tvm runtime itself.

This change enabled configurable option for such targets.

PR #11391, opened 2022-05-20
[VM] Memory alignment check for `set_input_zero_copy`

KJlaccHoeUM9l:
[PR](https://github.com/apache/tvm/pull/11003) has added the ability to skip copying data when creating input `NDArray` tensors if the source input is in `DLTensor` format.

However, when adding this functionality, memory alignment was not checked, as it was done for [GraphExecutor](https://github.com/apache/tvm/blob/c216cbec5bb795a8b13bdb1e177b523e4f7e4ca8/src/runtime/graph_executor/graph_executor.cc#L168).
In view of this, runtime errors (`Segmentation fault (core dumped)`) are possible, because TVM uses [aligned memory](https://github.com/apache/tvm/blob/c216cbec5bb795a8b13bdb1e177b523e4f7e4ca8/src/runtime/c_runtime_api.cc#L166).

This PR adds this check.

PR #14771, opened 2023-05-04
[TIR][Runtime] Allow use of external non-compact/non-aligned buffers

Lunderberg:
Prior to this commit, any use of `tvm.nd.from_dlpack` to create a strided `NDArray`, or a `NDArray` whose alignment was less than `tvm::runtime::kAllocAlignment` would raise an error.  As a result, views into larger arrays, which are unlikely to be aligned and compact, could only be shared when copied into an aligned and compact buffer.

This commit moves the compact/aligned check from the `NDArray` class into the generated TIR code as part of DLTensor unpacking.  These checks were initially introduced in https://github.com/apache/tvm/pull/11391, to avoid segfaults caused by use of non-aligned buffers in code intended for aligned buffers.  The new checks will provide the same safeguard as the alignment is checked prior to use, but allows the alignment requirement to be relaxed on a per-buffer basis.

This approach also removes a potential bug resulting from compile-time configuration of `tvm::runtime::kAllocAlignment`, first introduced in https://github.com/apache/tvm/pull/13307.  Since TVM supports cross-compiling, the installation of TVM used to compile a kernel may assume a larger value of `kAllocAlignment` than is provided by the runtime installation of TVM.  By validating the alignment within the generated kernel, rather than as part of the runtime, this potential inconsistency would be caught.

### DIFF
--- a/apps/bundle_deploy/bundle.c
+++ b/apps/bundle_deploy/bundle.c
@@ -32,8 +32,10 @@
 
 #define CRT_MEMORY_NUM_PAGES 16384
 #define CRT_MEMORY_PAGE_SIZE_LOG2 10
+#define CRT_MEMORY_PAGE_SIZE_BYTES (1 << CRT_MEMORY_PAGE_SIZE_LOG2)
 
-static uint8_t g_crt_memory[CRT_MEMORY_NUM_PAGES * (1 << CRT_MEMORY_PAGE_SIZE_LOG2)];
+static uint8_t g_crt_memory[CRT_MEMORY_NUM_PAGES * CRT_MEMORY_PAGE_SIZE_BYTES]
+    __attribute__((aligned(CRT_MEMORY_PAGE_SIZE_BYTES)));
 static MemoryManagerInterface* g_memory_manager;
 
 /*! \brief macro to do C API call */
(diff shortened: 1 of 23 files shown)


Lunderberg:
It looks like this commit breaks most of the Rust tests, as the alignment assumed by the LLVM kernel was never validated when running through Rust.  Unfortunately, I can't find a good way to produce aligned allocations in Rust, as [`Vec::new_in`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.new_in) is unstable.  For now, I've updated the Rust tests to opt out of requiring aligned buffers, but this isn't the best long-term solution.

Lunderberg:
Rebased onto main to resolve conflict with https://github.com/apache/tvm/pull/15132.  The individual commits are now cleaned up and ordered such that, if CI failures continue to occur, they can be segmented out into independent PRs.

Lunderberg:
Rebased this onto main, as I attempt to clear out some of my (very) old PRs.

Lunderberg:
Closing this PR, as it is unlikely that I will have a chance to get back to it.

If anybody else would like to pick this up, the last roadblock was that the Rust allocator does not provide the higher alignment requirement that is later assumed that memory allocation, and doesn't have a clear way to provide the larger-than-element-type alignment.

PR #14774, opened 2023-05-05
[microNPU][ETHOSU] Fix Softmax quantization parameters

Aleksei-grovety:
Fix zero point and scale values for operations according to the values in Vela, the test is updated to check case with different input and output zero point.


Aleksei-grovety:
@tvm-bot rerun

leandron:
Thanks @ekalda @Aleksei-grovety!

PR #14777, opened 2023-05-05
[Unity][Training] More Relax operators gradient supported

SiriusNEO:
This PR introduces gradient functions for the following Relax operators:
- `R.maximum`
- `R.triu`
- `R.broadcast_to`
- `R.silu`
- `R.minimum`


python/tvm/relax/op/_op_gradient.py
```diff
@@ -231,7 +232,33 @@ def power_grad(
     ]
 
 
+@register_gradient("relax.maximum")
```
yongwww, line 235: How about adding relax.minimum as well?

yongwww (approved):

PR #14779, opened 2023-05-05
fix: use arm on demand instead of spot

naut-brian:
Spot instances for Arm is failing right now so we're going to go back to on demand.

### DIFF
--- a/ci/jenkins/templates/utils/macros.j2
+++ b/ci/jenkins/templates/utils/macros.j2
@@ -33,7 +33,7 @@ sh(
 {% set test_dir_name = name.replace(":", "").replace(" ", "-").replace("-", "_")|string %}
 def {{ method_name }}(node_type='{{ node }}-SPOT', on_demand=false) {
   if ({{ condition }}) {
-    if (on_demand==true) {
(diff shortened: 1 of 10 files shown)


ci/jenkins/templates/utils/macros.j2
```diff
@@ -33,7 +33,7 @@ sh(
 {% set test_dir_name = name.replace(":", "").replace(" ", "-").replace("-", "_")|string %}
 def {{ method_name }}(node_type='{{ node }}-SPOT', on_demand=false) {
   if ({{ condition }}) {
-    if (on_demand==true) {
+    if (on_demand==true or 'ARM' in node_type) {
```
driazati, line 36: without being wrapped in `{{ ... }}` this happens inside groovy at runtime, so `in` and `or` don't work
```suggestion
    if (on_demand==true || node_type.contains('ARM')) {
```

driazati (approved):
approach lgtm

yongwww (approved):

yongwww:
@tvm-bot rerun

junrushao (approved):

PR #14783, opened 2023-05-06
[MetaSchedule] Add an API to dump a pruned database

junrushao:
This PR introduces the `Database.dump_pruned` API, which makes possible to keep the only optimal record for the workload on disk for easy re-distribution of a MetaSchedule database.


junrushao:
CC: @tqchen @spectrometerHBH

spectrometerHBH (approved):

PR #14786, opened 2023-05-06
[TVMScript] Add `__name__` attr for parsed PrimFunc and IRModule

Hzfengsy:
This PR adds `__name__` attr to indicate the func/mod name for parsed PrimFunc and IRModule.

cc @tqchen @cyx-6


junrushao (approved):
LGTM! Curious question: is the python-side attr going to affect the AST at all?

Hzfengsy:
It will not affect the AST, just for a hint only in Python env for better UX :)

PR #14792, opened 2023-05-06
[OpenCL] Refactor OpenCL runtime to support SPIRV binary ingestion

masahi:
This is the first step toward supporting Intel XMX (tensorcore equivalent of Intel dGPUs). They expose XMX via spirv, and such spirv modules can be consumed by their OpenCL runtime.  It is going to be a long game, since [their extension](https://github.com/intel/llvm/blob/sycl/sycl/doc/design/spirv-extensions/SPV_INTEL_joint_matrix.asciidoc) is not even in the spirv toolchain yet. I have the simplest test case of XMX tensorization (m8n8k16 matmul) working using [this script](https://github.com/masahi/tensorir-experiment/blob/master/spirv-opencl/joint_matrix.py) but beyond that I've been struggling with strange correctness issues (see https://github.com/intel/intel-graphics-compiler/issues/279).

For now, I'm abusing the existing `-device` target option to decide when to enable spirv code gen for the opencl target (`target = "opencl -device=spirv"`). I don't have a strong opinion on this, suggestions welcome. 

The next steps are going to be
* Update our spirv codegen to support emitting the OpenCL flavor ("Kernel" capability, https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Env.html)
* Support Intel joint matrix extension in our spirv codegen. It's very similar to the NV tensorcore extension https://github.com/apache/tvm/pull/14770

@echuraev @tqchen @csullivan


src/runtime/opencl/opencl_module.cc
```diff
@@ -108,7 +109,27 @@ class OpenCLWrappedFunc {
   LaunchParamConfig launch_param_config_;
 };
 
-OpenCLModuleNode::~OpenCLModuleNode() {
+class OpenCLSPIRVModuleNode : public OpenCLModuleNodeBase {
```
echuraev, line 112: Probably all runtime part related to SPIRV can be moved in a new directory in `src/runtime`? I suppose that SPIRV can be useful no only for Vulcan, so SPIRV can be separated from other runtimes initially. What do you think?

src/target/source/codegen_opencl.cc
```diff
@@ -585,6 +586,14 @@ void CodeGenOpenCL::SetTextureScope(
 }
 
 runtime::Module BuildOpenCL(IRModule mod, Target target) {
+#if TVM_USE_VULKAN
```
echuraev, line 589: Probably this variable can be device-agnostic? Like: `TVM_USE_SPIRV` or something like that.

src/runtime/opencl/opencl_module.cc
masahi: This class is still specific to OpenCL. Other than the input binary format and kernel initialization, most runtime code are shared between the normal OpenCL runtime and the new one that consumes SPIRV. So I want to put it here.

src/target/source/codegen_opencl.cc
masahi: Okay I'll change it to `TVM_HAS_SPIRV` which, for now, is set to true only when `USE_VULKAN=ON`.

masahi: Changed to `TVM_ENABLE_SPIRV`, thanks for the suggestion!

src/runtime/opencl/opencl_module.cc
echuraev: Probably in this case, `SPIRVShader` from Vulcan can be moved to OpenCL directory? And I'm ok if it is in the same files with OpenCL runtime. Just as an idea, probably the SPIRV part can be moved to new files such as `spirv_module.h/cc`?

masahi: Ok, made the following changes:
* Add `runtime/spirv` dir and put `SPIRVShader` there
* Add `runtime/opencl_module_spirv.cc` and put `OpenCLSPIRVModuleNode` and related code there.

Overall I also like this organization. Please take a look.

src/runtime/spirv/spirv_shader.h
```diff
@@ -31,7 +31,7 @@ namespace tvm {
 namespace runtime {
 namespace vulkan {
```
echuraev, line 32: I suppose it should be another namespace.
```suggestion
namespace spirv {
```

src/target/spirv/spirv_utils.cc
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+  return std::make_pair(smap, code_data.str());
+}
+
+#else
+
+std::pair<std::unordered_map<std::string, runtime::SPIRVShader>, std::string> LowerToSPIRV(
+    IRModule mod, Target target) {
+  LOG(FATAL) << "LowerToSPIRV is called but Vulkan is not enabled.";
```
echuraev, line 181: Probably Vulkan should be removed from the error message, then error message will be more generic?

echuraev (commented):
In general LGTM. Several minor comments.

echuraev (approved):
LGTM. Thank you!

PR #14797, opened 2023-05-07
[NDArray] Allow arbitrary stride when the corresponding shape is 1

masahi:
As discussed in https://github.com/pytorch/pytorch/issues/99803, as of PT 1.13, PT started to fill in a stride with 1 whenever the corresponding shape axis has size 1 during PT tensor to DLPack conversion. This breaks `tvm.runtime.ndarray.from_dlpack` when the shape is like `(1, 77)` (clip for SD) or whenever the batch size is 1 in vision models. So this is going to be a common problem that people would encounter in practice (see https://discuss.tvm.apache.org/t/a-pytorch-and-tvm-version-incompatibility-problem-while-using-from-dlpack-and-to-dlpack/14771).

This PR fixes this issue by relaxing the runtime stride check. An alternative is to modify the strides during `from_dlpack`, but since PT introduced this change for a good reason, I felt that trying to calculate the "correct" strides on our side again is not a good idea.   

@tqchen @junrushao


tqchen (approved):

junrushao (approved):
Good catch!

PR #14800, opened 2023-05-08
[Unity] Fix CUDA graph rewrite var used before def

vinx13:
CUDA graph rewriting may result reordering of original bindings, for example when a variable is used as an input of the lifted function. If the variable comes from the output of another function, we need to make sure output unpacking is emitted.


masahi (approved):
Confirmed that this fixes the issue I had. Need to fix lint.

masahi:
@vinx13 Please remove this debug log https://github.com/apache/tvm/blob/unity/src/runtime/relax_vm/cuda/cuda_graph_builtin.cc#L90

masahi:
@tvm-bot rerun

PR #14802, opened 2023-05-09
[Unity] Add pass to allocate big workspace and pass it to all functions that need temp storage  

masahi:
As discussed in https://github.com/apache/tvm/pull/14723, some cutlass kernels, in particular the attention kernel, require temporary storage to hold intermediate results. This results in repeated calls to cudaMalloc/Free, and the bug like #14723. 

This PR adds a pass that allocate a big workspace at the beginning of `main` and threads that through all functions that need temp storage. This lets us eliminate all redundant malloc / free, and improves e2e perf slightly. 

An external function can specify its workspace requirement by the `kWorkspaceSize` attribute. See how it is done for the cutlass attention kernel for example. 

@vinx13 @cyx-6 @yelite

### DIFF
--- a/include/tvm/relax/expr.h
+++ b/include/tvm/relax/expr.h
@@ -983,6 +983,8 @@ constexpr const char* kCodegen = "Codegen";
 constexpr const char* kComposite = "Composite";
 /*! \brief Indicate the function was created by the Pattern Partitioning Pass. */
 constexpr const char* kPartitionedFromPattern = "PartitionedFromPattern";
+/*! \brief The required workspace for an external function. */
+constexpr const char* kWorkspaceSize = "WorkspaceSize";
 }  // namespace attr
 
 /*! \brief The extern function, which can represent packed function. */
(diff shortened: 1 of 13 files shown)


vinx13 (approved):

yelite:
Does this also work for dynamic shape? It looks so but just want to confirm. Although the codegen doesn't support dynamic shape attention yet, it would be great if the workspace allocation works with dynamic shape.

tqchen:
Thanks @masahi ! Some high level notes:

- We don't need to use zeros, can use relax's builtin alloc_tensor


Some ideas on making it even more composable:
- We can have cutlass BYOC to generate TIR functions that contains the allocate and call_extern part into cutlass (the T.allocate here can depend on dyn shape)
- We can then have a pass that lift the workspace allocation into relax
- Finally, the relax memory planning will tie up the reusable memory through memory planning

masahi:
> Does this also work for dynamic shape?

@yelite For now no, since we need to find the maximum of the workspace sizes required by all extern functions. In principle we may be able to specify the size via a symbolic expression and doing max at runtime, but that's for future PR :) 

Attention with dynamic seq len sounds interesting. For that, the kernel needs to check the given workspace size and if that's smaller than the actual runtime size it requires, we need to fallback to additional malloc / free. If there is a known upper bound on the dynamic dim size, we can use that when we annotate workspace requirements.

masahi:
> We don't need to use zeros, can use relax's builtin alloc_tensor

@tqchen It seems `alloc_storage` requires a device index, which we don't know during this pass but `zeros` automatically takes care of the device issue.

tqchen:
device_index=0 should be sufficient and is also the default behavior(the default device). Hetro execution would need to have smarter selection of device index anyway.

PR #14807, opened 2023-05-09
[Unity][BYOC] Fix incorrect bias stride in matmul cutlass offload

yelite:
This PR makes the cutlass codegen use the correct bias stride when bias has more than 2 dimensions. For example, if the input bias has shape (1, n, 4096), the original code will set ldc to 0, which produces incorrect result. This enables offloading matmul in LLaMA.

The same PR is merged in mlc relax https://github.com/mlc-ai/relax/pull/212. 

cc @vinx13 @masahi

### DIFF
--- a/tests/python/relax/test_codegen_cutlass.py
+++ b/tests/python/relax/test_codegen_cutlass.py
@@ -374,6 +374,44 @@ def test_matmul_offload(
     tvm.testing.assert_allclose(out, ref, rtol=1e-2, atol=1e-2)
 
 
+def test_matmul_with_3d_bias_offload():
+    x_shape = (1, 4, 8)
+    y_shape = (1, 8, 16)
+    dtype = "float16"
+
+    x = np.random.randn(*x_shape).astype(dtype)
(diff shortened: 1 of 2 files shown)


masahi (approved):

tests/python/relax/test_codegen_cutlass.py
```diff
@@ -374,6 +374,44 @@ def test_matmul_offload(
...
+    x = np.random.randn(*x_shape).astype(dtype)
+    y = np.random.randn(*y_shape).astype(dtype)
+    bias = np.random.randn(1, x_shape[-2], y_shape[-1]).astype(dtype)
+    args = (x, y, bias)
+
+    mod = get_relax_matmul_module(
+        x_shape,
+        y_shape,
+        dtype,
+        with_bias=True,
```
masahi, line 391: This test is broken, since `get_relax_matmul_module` doesn't have `with_bias` param.

yelite: I created #14881 to fix it. Thanks for the heads up

PR #14809, opened 2023-05-09
[Unity][BYOC] Add shape validation for bias arg in cuBLAS

yelite:
This PR adds more logic on bias shape in cuBLAS partitioning. This prevents matmul with residual input to be offloaded to cuBLAS as matmul with bias. cuBLAS only support 1D bias input.

This PR also fixes the type key for cublas json runtime, so that it can be deserialized correctly.

cc @masahi @vinx13


masahi (approved):
Need to fix lint.

PR #14812, opened 2023-05-09
[TVMScript] Handle AllocatedPoolInfo, ConstantPoolInfo, ConstantInfo

Lunderberg:
These objects may appear as part of a TIR PrimFunc following the `tir.usmp.ConvertPoolAllocationsToOffsets` transform.  Prior to this commit, any such PrimFuncs would result in errors when printing TVMScript, along with a fallback to the legacy repr.

### DIFF
--- a/python/tvm/ir/memory_pools.py
+++ b/python/tvm/ir/memory_pools.py
@@ -245,7 +245,7 @@ class ConstantMemoryPools(Object):
         )
 
 
-@register_object("ir.ConstantMemoryPools")
+@register_object("ir.AllocatedPoolInfo")
 class AllocatedPoolInfo(Object):
     """Allocate memory in a given pool.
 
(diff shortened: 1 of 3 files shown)


python/tvm/ir/memory_pools.py
```diff
@@ -245,7 +245,7 @@ def __init__(
         )
 
 
-@register_object("ir.ConstantMemoryPools")
```
junrushao, line 248: A bit off the topic, but to some extent, I believe it might be less ideal if we expose those feature-dependent component in `tvm/ir` folder, which is supposed to be generic.

Lunderberg: Yeah, I wasn't sure how much refactoring to include in this PR.  There were some clear inconsistencies, such as the wrong object name being in `@register_object`, but this feels like something that should be in the `tir.usmp` namespace.

junrushao: Agreed. This seems less organized. Let's leave this to future change

junrushao (approved):

PR #14814, opened 2023-05-09
[Unity][Relax] gelu-tanh operator

yzh119:
# Motivation
The gelu operator has different implementations, in [PyTorch](https://pytorch.org/docs/stable/generated/torch.nn.GELU.html), they are distinguished by `approximate='none'` or `approximate='tanh'`. Currently, Relax only implements the case where `approximate="none"`  which depends on erf function, this PR implements the gelu operator approximated by tanh. In hugging face transformers, the gelu with tanh operator is called [gelu_new](https://github.com/huggingface/transformers/blob/366a8ca09e8dd92dfb1956c8be3118b5a2b13639/src/transformers/activations.py#L49) and was used in many models.

## Difference with erf gelu
Actually, I only observed a very slight numerical difference between the two implementations, so I'm still not sure whether it necessary to include this variant or not.

Below is a figure showing the numerical difference between the two, and the script:
image

```python
import numpy as np
import torch
import torch.nn.functional as F
import matplotlib.pyplot as plt

x = np.linspace(-6, 6, 200)
x = torch.from_numpy(x).float()
y_none = F.gelu(x, approximate='none')
y_tanh = F.gelu(x, approximate='tanh')
plt.plot(x.numpy(), (y_tanh - y_none).numpy(), label='difference')
plt.legend()
plt.show()
```

The largest difference is 5e-4.

### DIFF
--- a/python/tvm/relax/backend/contrib/cutlass.py
+++ b/python/tvm/relax/backend/contrib/cutlass.py
@@ -158,6 +158,8 @@ def _check_matmul(context: PatternCheckContext) -> bool:
 def _get_activation_from_name(pattern_name):
     if "_relu" in pattern_name:
         return "relax.nn.relu"
+    elif "_gelu_tanh" in pattern_name:
+        return "relax.nn.gelu_tanh"
     elif "_gelu" in pattern_name:
         return "relax.nn.gelu"
     elif "_silu" in pattern_name:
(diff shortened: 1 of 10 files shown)


Hzfengsy (approved):

yongwww (approved):
LGTM

PR #14817, opened 2023-05-10
[Vulkan] Add cooperative matrix support

mei-ye:
[Vulkan] Add SPIR-V code generation for "SPV_NV_cooperative_matrix" extension
Add im2col implementation for direct Conv2D.  Currently only 16x16x16 FP16 wmma fragments with FP32 intermediates are supported.  Add "min_design_space" as a parameter to give minimum design space for meta scheduler tuning. Add "use_int32_const" as a paramter to use int32 type for constants. Allow target query to be called from the schedules so that samplings are constrained to produce legal schedules. Do not allow the reuse of buffers with different dtypes. Add a unit test test_wmma.py.

### DIFF
--- a/src/target/spirv/codegen_spirv.cc
+++ b/src/target/spirv/codegen_spirv.cc
@@ -129,6 +129,7 @@ void CodeGenSPIRV::InitFuncState() {
   builder_.reset(new spirv::IRBuilder(spirv_support_));
   builder_->InitHeader();
   shared_memory_bytes_used_ = 0;
+  fragment_info_.clear();
 }
 
 spirv::Value CodeGenSPIRV::GetThreadIndex(const IterVar& iv, const PrimExpr& extent) {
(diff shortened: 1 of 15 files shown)


junrushao:
CC @masahi seems similar to your PR: https://github.com/apache/tvm/pull/14770

masahi:
Yeah, hi Mei, it's interesting to know that we worked on this extension at the same time! I'll look through this PR and think about how to integrate our work.

mei-ye:
> Yeah, hi Mei, it's interesting to know that we worked on this extension around the same time! I'll look through this PR and think about how to integrate our work.
> 
> BTW, I'm aware that AMD's Windows VK driver supports this extension. Is Linux driver going to get the support for this extension as well?

Yes, Linux driver supports it.  Though let me ask whether it is made into release.

src/target/spirv/codegen_spirv.cc
```diff
@@ -395,6 +396,135 @@ spirv::Value CodeGenSPIRV::VisitExpr_(const CallNode* op) {
...
+        builder_->StructArrayAccess(ptr_type_a, var_map_[buffer_a], MakeValue(index_a));
+    spirv::Value ptr_b =
+        builder_->StructArrayAccess(ptr_type_b, var_map_[buffer_b], MakeValue(index_b));
+    spirv::Value ptr_c =
+        is_equal ? ptr_d
+                 : builder_->StructArrayAccess(ptr_type_c, var_map_[buffer_c], MakeValue(index_c));
+    uint32_t mask = spv::MemoryAccessMaskNone;
+    spirv::Value loaded_a = builder_->MakeValue(spv::OpLoad, fragment_type_a, ptr_a, mask);
+    spirv::Value loaded_b = builder_->MakeValue(spv::OpLoad, fragment_type_b, ptr_b, mask);
+    spirv::Value loaded_c = builder_->MakeValue(spv::OpLoad, fragment_type_c, ptr_c, mask);
```
masahi, line 487: See the discussion in https://github.com/apache/tvm/pull/14770#discussion_r1187486958. I was thinking that we cannot load / store matrices to a buffer, but looking at this code I'm realizing that that was a wrong assumption. 

But even then, I prefer my solution that materializes individual matrices as phi value. Can you share your thought to the discussion in #14770?

mei-ye:
> > Yeah, hi Mei, it's interesting to know that we worked on this extension around the same time! I'll look through this PR and think about how to integrate our work.
> > BTW, I'm aware that AMD's Windows VK driver supports this extension. Is Linux driver going to get the support for this extension as well?
> 
> Yes, Linux driver supports it. Though let me ask whether it is made into release.

Linux driver has been released in April with this extension support.

mei-ye:
> > Linux driver has been released in April with this extension support
> 
> Do you mean AMDVLK or AMDGPU Pro? I looked into AMDVLK source code but didn't find this extension supported.

AMDGPU pro.

masahi:
@mei-ye Can you split topi / relay / ms change from this PR and send them later? And please add a minimum test case for cooperative matrix, such as the matmul test in my PR.

mei-ye:
> @mei-ye Can you split topi / relay / ms change from this PR and send them later? And please add a minimum test case for cooperative matrix, such as the matmul test in my PR.

@masahi : are you going to check in your SPIR-V code gen?  In this patch, I added an unit test: test_wmma.py.

masahi:
> > @mei-ye Can you split topi / relay / ms change from this PR and send them later? And please add a minimum test case for cooperative matrix, such as the matmul test in my PR.
> 
> @masahi : are you going to check in your SPIR-V code gen? In this patch, I added an unit test: test_wmma.py.

That's still open for discussion. If people prefer your approach, I'll close my PR. In which case I want your PR to be easier to review.

src/target/spirv/codegen_spirv.cc
```diff
@@ -395,6 +396,135 @@ spirv::Value CodeGenSPIRV::VisitExpr_(const CallNode* op) {
...
+    int stride = static_cast<int>(Downcast<IntImm>(op->args[6])->value);
+    auto type_int = builder_->GetSType(DataType::Int(32));
+    spirv::Value stride_val = builder_->IntImm(type_int, stride);
+    std::string layout = (op->args[7].as<StringImmNode>())->value;
+    spirv::SType dst_ptr_type =
+        builder_->GetPointerType(fragment_type, fragment_info_[buffer_node].sclass);
+    spirv::Value dst_ptr =
+        builder_->StructArrayAccess(dst_ptr_type, var_map_[buffer_node], MakeValue(dst_index));
+    const CallNode* call_node = src_ptr_expr.as<CallNode>();
+    ICHECK(call_node && call_node->op.same_as(builtin::address_of()));
```
masahi, line 434: Lower `address_of` in its own else if branch (see my PR)

src/target/spirv/codegen_spirv.cc
```diff
@@ -726,5 +885,51 @@ void CodeGenSPIRV::VisitStmt_(const SeqStmtNode* op) {
...
+int CodeGenSPIRV::Stoi(const std::string& str) {
+  try {
+    return std::stoi(str);
+  } catch (std::invalid_argument& e) {
+    LOG(FATAL) << "Cannot convert \"" << str << "\" to int";
+    throw;
+  }
+}
+
+std::pair<int32_t, int32_t> CodeGenSPIRV::GetWmmaFragmentSize(const VarNode* variable) {
```
masahi, line 897: Please don't duplicate this with `codegen_cuda.cc`

src/target/spirv/ir_builder.cc
```diff
@@ -500,7 +519,18 @@ SType IRBuilder::DeclareType(const DataType& dtype) {
...
     t.type = dtype;
     SType base_type = GetSType(dtype.element_of());
-    ib_.Begin(spv::OpTypeVector).AddSeq(t, base_type, dtype.lanes()).Commit(&global_);
+
+    if (row * col == 0) {
+      ICHECK((row == 0) && (col == 0));
+      ib_.Begin(spv::OpTypeVector).AddSeq(t, base_type, dtype.lanes()).Commit(&global_);
+    } else {
+      Value v_row = GetSpecConst(GetSType(DataType::UInt(32)), row);
+      Value v_col = GetSpecConst(GetSType(DataType::UInt(32)), col);
```
masahi, line 528: I don't get why we need specialization constants. Aren't they for configuring values that can change (or values that are only known) at runtime?

mei-ye: Yes, you can give the specialization constant a new value at runtime.

masahi: But why do you want to do that for cooperative matrix?

src/target/spirv/codegen_spirv.cc
mei-ye: I factor out the lowering of "address_of" in the latest commit.   However, calls to "address_of" is an argument of calls to  "tvm_load_matrix_sync", like below:  
@tir.tvm_load_matrix_sync(meta[tir.Var][1], 16, 16, 16, 0, @tir.address_of(X_shared_2: Buffer(X_shared, float16, [1], [], scope="shared")[0], dtype=handle), 16, "row_major", dtype=handle)

The call to "address_of" is not recursively visited in: spirv::Value CodeGenSPIRV::VisitExpr_(const CallNode* op)

src/target/spirv/codegen_spirv.cc
mei-ye: I factor out this code now.

src/target/spirv/codegen_spirv.cc
masahi: You can add additional `VisitExpr(call->args[5])`

mei-ye: Do you prefer VisitExpr or the factoring approach?

masahi: Not sure what you meant by "the factoring approach", but I prefer having 

```
else if (op->op.same_as(builtin::address_of())) {
}
```

since it allows other use of `address_of` other than `load_matrix`.

mei-ye: OK, fix is checked in now.  I recombined the fix of several issues and the original code into one commit.

masahi:
@tvm-bot rerun

src/target/spirv/codegen_spirv.cc
```diff
@@ -394,6 +395,120 @@ spirv::Value CodeGenSPIRV::VisitExpr_(const CallNode* op) {
...
     return spirv::Value();
+  } else if (op->op.same_as(builtin::tvm_fill_fragment())) {
+    ICHECK_EQ(op->args.size(), 6U);
+    const VarNode* buffer_node = op->args[0].as<VarNode>();
+    ICHECK(buffer_node && fragment_info_.count(buffer_node));
+    DataType ele_dtype = GetElementDataType(buffer_node);
+    ICHECK(ele_dtype.is_float()) << "Only floating point fragment accumulator is supported";
+    spirv::SType ele_stype = builder_->GetSType(ele_dtype);
+    spirv::SType& fragment_type = fragment_info_[buffer_node].stype;
+    double init = static_cast<uint64_t>(Downcast<FloatImm>(op->args[5])->value);
```
masahi, line 406: why cast to uint64?

src/target/spirv/spirv_support.h
```diff
@@ -276,6 +276,20 @@ struct SPIRVSupport {
...
+  /*!
+   * \brief  Whether the driver supports operations involving cooperative matrix.
+   *
+   * Vulkan extension: VK_NV_cooperative_matrix
+   * SPV Extension name: SPV_NV_cooperative_matrix
+   * SPV Capability: spv::CapabilityCooperativeMatrixNV
+   *
+   * If support is present, can perform cooperative matrix operations.  If
+   * support is not present, codegen will throw exception on
+   * attempting to perform cooperative matrix.
```
masahi, line 289: `perform cooperative matrix operations`

masahi (approved):

src/target/spirv/codegen_spirv.cc
mei-ye: I can't recall a good reason.  Removing this cast works fine.  Should I reset and re-patch?

masahi:
I'll address the nit issues in my upcoming PR.

tqchen:
awesome, thank you @mei-ye @masahi !

masahi:
I've updated my vk 4K matmul test https://github.com/masahi/tensorir-experiment/blob/master/vk_cooperative_matrix_nv/test_4k.py to use this extension support. It gets ~90 TFLOPs on RTX4080. 

The cool part is that by changing the target from vk to cuda the exact same schedule / script works with the same performance. https://github.com/masahi/tensorir-experiment/blob/master/vk_cooperative_matrix_nv/test_4k.py#L171-L172

masahi:
Auto tensorization on vk seems to work as well, but the result is not correct after tuning. For CUDA the result is correct after auto-tensorization tuning, so this is a VK-specific issue.

PR #14822, opened 2023-05-10
[TOPI] Use f-strings for string formatting, NFC

kparzysz-quic:


kparzysz-quic:
Not sure what people think about this.  I find f-strings clearer than % or format.

kparzysz-quic:
@tvm-bot rerun

masahi (approved):

PR #14824, opened 2023-05-11
[Relay][BugFix] fix a bug about ReLu in the threshold attribute which causes a different results with keras

jikechao:
### Expected Behavior 
TVM rejects the invalid attribute value `threshold=None` immediately rather than accepting it and producing a different result with Keras. 

### Actual Behavior
image

### Steps to reproduce
```
import tvm
import tvm.relay as relay
import numpy as np
from tensorflow import keras


input_shape = (1, 2, 3, 4)

layer = keras.layers.ReLU(threshold=None)  # threshold of a ReLU layer cannot be the None value. Received: None

y = layer(x)
model = models.Model(x, y)
print(model.summary())
# model.save('model.h5')
res_keras = model.predict(input_data)

mod, params = relay.frontend.from_keras(model, shape_dict)

with tvm.transform.PassContext(opt_level=1):
    model = relay.build_module.create_executor("graph", mod, tvm.cpu(0), 'llvm', params).evaluate()  # crash

test_x_tvm = input_data
res_tvm = model(tvm.nd.array(test_x_tvm.astype('float32'))).numpy()

np.testing.assert_allclose(res_keras, res_tvm, atol=1e-3, rtol=1e-3)
```

cc @echuraev @Hzfengsy

### DIFF
--- a/tests/python/frontend/keras/test_forward.py
+++ b/tests/python/frontend/keras/test_forward.py
@@ -227,6 +227,7 @@ class TestKeras:
             act_funcs = [
                 keras_mod.layers.LeakyReLU(alpha=None),
                 keras_mod.layers.LEU(2, 3, 4),
+                keras_mod.layers.ReLU(threshold=None),
             ]
             data = keras_mod.layers.Input(shape=(2, 3, 4))
             for act_func in act_funcs:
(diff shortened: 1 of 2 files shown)


Hzfengsy (approved):

tests/python/frontend/keras/test_forward.py
```diff
@@ -227,6 +227,7 @@ def test_forward_activations_except(self, keras_mod):
             act_funcs = [
                 keras_mod.layers.LeakyReLU(alpha=None),
                 keras_mod.layers.LEU(2, 3, 4),
+                keras_mod.layers.ReLU(threshold=None),
```
echuraev, line 230: Is this behavior actual only for Keras <= 2.3.1?

jikechao: Yes, When Keras > 2.3.1, an assertion have been added to the Keras to crash this usage. Thus, this test is only for Keras <= 2.3.1

echuraev (approved):
Thank you for this PR and clarification about keras version.

PR #14832, opened 2023-05-11
[relay][convert layout]: Infer correct layout for nn.batch_flatten

f2013519:
Currently during the ConvertLayout pass, whenever there is a nn.batch_flatten op, additional layout transforms are inserted before the nn.batch_flatten op. This may not be efficient.

In addition, for BYOC targets (which expects ops to have a certain layout), insertion of these transforms can create additional overhead - especially if they cannot be handled by target which then creates additional compiler regions.

This can be avoided by propagating the updated layouts to the nn.batch_flatten op.


f2013519:
Dropping as this can lead to issues with downstream ops such as dense layer whose weights assume original layout from the flatten layer.

PR #14833, opened 2023-05-12
[Codegen][OpenCL] fix amibiguous selection operator call

spectrometerHBH:
Per OpenCL specification

https://man.opencl.org/select.html

### DIFF
--- a/src/target/source/codegen_opencl.cc
+++ b/src/target/source/codegen_opencl.cc
@@ -437,7 +437,7 @@ void CodeGenOpenCL::VisitExpr_(const CallNode* op, std::ostream& os) {
       LOG(FATAL) << "Unsupported type: " << op->dtype
                  << ", currently only float and half are supported for image2d OpenCL codegen.";
     }
-  this->PrintExpr(op->args[0], ss);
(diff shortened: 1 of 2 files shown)


tqchen (approved):

src/target/source/codegen_opencl.cc
```diff
@@ -567,11 +567,7 @@ void CodeGenOpenCL::VisitExpr_(const SelectNode* op, std::ostream& os) {
   os << ", ";
   PrintExpr(op->condition, oss);
   if (op->dtype.is_float()) {
-    if (op->condition.dtype().is_uint() || op->condition.dtype().is_int()) {
-      os << oss.str();
-    } else {
-      os << CastTo(oss.str(), DataType::Int(op->dtype.bits(), op->dtype.lanes()));
-    }
+    os << CastTo(oss.str(), DataType::Int(op->dtype.bits(), op->dtype.lanes()));
```
echuraev, line 570: Could you please add a unit test for this change? I suppose it might be similar with [test_opencl_type_casting](https://github.com/apache/tvm/blob/main/tests/python/unittest/test_target_codegen_opencl.py#L142-L184).

PR #12534, opened 2022-08-22
[Docker][CI][RISC-V] Build riscv-isa-sim (spike) in ci_riscv Docker image to enable RISC-V unit testing

PhilippvK:
## Context

See Discussion in https://discuss.tvm.apache.org/t/coordination-of-risc-v-integration-in-tvm/13133

## Summary

1. Update `Dockerfile.ci_riscv` to download RISC-V toolchain and compile Spike simulator (including proxy kernel)

   - Using SiFives GCC toolchain for Embedded targets (See below)
   - Two variants of proxy kernel (`pk`, `pk64`) are build to support `rv32gc` as well as `rv64gc` targets

2. Introduce AotTestRunner to use spike in unit tests (currently `rv32gc` only)

3. Enable spike based AoTTestRunner for AoT/CRT related test cases

## Future work

I have a followup ready based on this PR, which enable usage of spike for MicroTVM deployment (ProjectAPI Server).

## Questions

- Why not use existing CSI-NN GCC toolchain?

That toolchain (`Xuantie-900-gcc-linux`) is mainly targeting 64-bit devices capable of running Linux. While it could be used for 32-bit bare metal targets as well (using the `-static` compile flag), it is missing relevant multilib arches, such as `rv32gc+ilp32d` to be  usable. Thus, I use Sifives GCC (which does currenlty not support the RVV1.0 extension)

- Should we fix the version of spike being build or just use the latest development branch?

cc @Mousius @areusch @driazati @gigiblender

PR #14836, opened 2023-05-12
[microTVM][RISCV] Tensorization for conv_2d_nchw_int8 with RVV extension

katebern-grovety:
Dot product tensorization with RISC-V V-extension intrinsic for conv_2d_nchw_int8. Using llvm intrinsic.
Testing with AOT test runner on Spike.


python/tvm/topi/riscv_cpu/conv2d_int8.py
```diff
@@ -0,0 +1,362 @@
...
+    Checks to ensure that we can use int8 on riscv_cpu.
+    1) The datatypes are correct.
+    2) The vector extension "V" is used.
+    """
+    # 1) Check datatypes.
+    is_dtype_support = data_dtype == "uint8" and kernel_dtype == "int8"
+
+    # 2) Check target.
+    current_target = target.Target.current(allow_none=False)
+    has_attr = "+v" in current_target.mattr
```
PhilippvK, line 54: Could you please look up if your schedules would also be supported by the sub-extensions `zve32x`, `zve32f`, `zve64x`, `zve64f`, `zve64d`? If this is the case, please extend this check to be aware of these `mattr` flags.

python/tvm/topi/riscv_cpu/tensor_intrin.py
```diff
@@ -0,0 +1,217 @@
...
+//   output[i] = 0;
+//   for (unsigned j = 0; j < 4; ++j) {{
+//     output[i] += data[k] * kernel[i][j];
+//   }}
+// }}
+
+#ifdef __cplusplus
+extern "C"
+#endif
+int32_t dot_uint8_int8_int32_body(uint8_t* data, int8_t* kernel, int32_t* output) {{
```
PhilippvK, line 123: Do you by any chance have some benchmark results on how this performs compared to the RVV autovectorizer which is provided by RISCV LLVM/GCC?

tests/python/relay/aot/riscv.mk
```diff
@@ -0,0 +1,85 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
PhilippvK, line 1: I tried to upstream a patch like this, before and if was rejected because of the `AOTTestRunner` being removed at some point a favor of the MicroTVM Project API interface. (See https://github.com/apache/tvm/pull/12534#discussion_r961100014)

I actually have MicroTVM Project API template for Spike for a long time now (and use it regulary). I still need to clean it up before upstreaming: https://github.com/PhilippvK/microtvm-spike-template

tests/python/relay/aot/riscv.mk
```diff
@@ -0,0 +1,85 @@
...
+ENABLE_TVM_PLATFORM_ABORT_BACKTRACE = 0
+DMLC_CORE=$(TVM_ROOT)/3rdparty/dmlc-core
+TOOLCHAIN_PATH=$(shell dirname $(shell which riscv64-unknown-linux-gnu-gcc))/..
+
+CC = clang-16
+CXX = clang++-16
+
+TARGET_CFLAGS = --target=riscv64-unknown-linux-gnu -march=rv64gcv -static
+RUNNER = spike
+RUNNER_OPT = --isa=rv64gcv $(shell which pk)
```
PhilippvK, line 29: I would like to have support for `rv32gc` here as well.

tests/python/relay/aot/riscv.mk
```diff
@@ -0,0 +1,85 @@
...
+
+PKG_CFLAGS = ${PKG_COMPILE_OPTS} ${TARGET_CFLAGS} -O2 \
```
PhilippvK, line 31: Some reason why using `-O2` instead of `-O3` or `-Os`?

tests/python/relay/aot/riscv.mk
```diff
@@ -0,0 +1,85 @@
...
+# specific language governing permissions and limitations
+# under the License.
+
+AOT_ROOT ?= $(CRT_ROOT)/aot
+
+ENABLE_TVM_PLATFORM_ABORT_BACKTRACE = 0
+DMLC_CORE=$(TVM_ROOT)/3rdparty/dmlc-core
+TOOLCHAIN_PATH=$(shell dirname $(shell which riscv64-unknown-linux-gnu-gcc))/..
+
+CC = clang-16
```
PhilippvK, line 24: Clang-16 and RISC-V GCC could both be supported here?

python/tvm/topi/riscv_cpu/conv2d_int8.py
```diff
@@ -0,0 +1,362 @@
...
+    is_arch_support = "v" in current_target.arch[2:]
+    if not is_arch_support and "march" in current_target.attrs:
+        is_arch_support = "v" in current_target.attrs["march"]
```
PhilippvK, line 57: I was not aware that `-march=` is actually exposed to the TVM target. This is great because it would also work with non-llvm TVM targets.

python/tvm/topi/riscv_cpu/conv2d_int8.py
```diff
@@ -0,0 +1,362 @@
...
+                batch, ic_chunk, ih, _, ic_block = s[data_vec].op.axis
+                parallel_axis = s[data_vec].fuse(batch, ic_chunk, ih)
+                s[data_vec].parallel(parallel_axis)
+
+                # conv2d_nchwc_int8 has 7D kernel
+                oc_chunk, ic_chunk, oh, ow, ic_block, oc_block, _ = s[kernel_vec].op.axis
+                s[kernel_vec].reorder(oc_chunk, oh, ic_chunk, ow, ic_block, oc_block)
+                oc_bn = cfg["tile_oc"].size[-1]
+                if oc_bn > 1:
+                    s[kernel_vec].vectorize(oc_block)
```
PhilippvK, line 182: You are adding LLVM-exclusive schedule primitives but only testing them on TVMs C-backends (with `tir.disable_vectorize=1`). We eventually need to also test if your vector intrinsics does interfere with the auto-vectorization going on in LLVM. Testing this should be possible without compiling and running it for Spike, since errors would already show up during `tvm.build()`.

tests/python/relay/aot/riscv.mk
katebern-grovety: I added Project API template for RISC-V with Spike based on CRT

tests/python/relay/aot/riscv.mk
katebern-grovety: No, there was no reason

tests/python/relay/aot/riscv.mk
PhilippvK: Thank you for coming up with that change. Would it be possible to split this up into a different PR with separate tests etc.?

That would make reviewing the individual changes much easier.

apps/microtvm/riscv/template_project/CMakeLists.txt.template
```diff
@@ -0,0 +1,64 @@
...
+endforeach(crt_lib_name ${CRT_LIBS})
+
+# Build model files
+add_library(tvm_model)
+file(GLOB_RECURSE tvm_model_srcs model/codegen/host/src/*.c model/codegen/host/lib/*.o)
+target_sources(tvm_model PRIVATE ${tvm_model_srcs})
+target_include_directories(tvm_model PRIVATE ${CMAKE_SOURCE_DIR}/include crt_config crt/include)
+target_compile_options(tvm_model PRIVATE -Wno-error=unused-variable -Wno-error=missing-braces -Wno-error=unused-const-variable -Wno-unused-variable)
+set_target_properties(tvm_model PROPERTIES LINKER_LANGUAGE C)
+target_link_libraries(main PRIVATE tvm_model)
```
PhilippvK, line 59: I think you would need to add `m` (aka. libmath) here to support Softmax layers.

apps/microtvm/riscv/template_project/microtvm_api_server.py
```diff
@@ -0,0 +1,285 @@
...
+        assert (new_flag & os.O_NONBLOCK) != 0, "Cannot set file descriptor {fd} to non-blocking"
+
+    def open_transport(self, options):
+        toolchain_path = options.get("toolchain_path")
+        target = options.get("target")
+        march = options.get("march")
+        spike_args = ["spike", f"--isa={march}", os.path.join(toolchain_path, target, "bin/pk")]
+        self._proc = subprocess.Popen(
+            spike_args + [self.BUILD_TARGET],
+            stdin=subprocess.PIPE,
```
PhilippvK, line 214: I was wondering if you also run into the issue with the broken STDIN support in recent version of spike mentioned here: https://github.com/riscv-software-src/riscv-isa-sim/issues/1400

Are you using an version of spike which does not have that problem or did you find another workaround?

apps/microtvm/riscv/template_project/CMakeLists.txt.template
```diff
@@ -0,0 +1,64 @@
...
+# specific language governing permissions and limitations
+# under the License.
+
+# SPDX-License-Identifier: Apache-2.0
+
+cmake_minimum_required(VERSION 3.18)
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_C_COMPILER "clang")
+set(CMAKE_CXX_COMPILER "clang++")
+set(FLAGS "--target=${TARGET} --sysroot=${TOOLCHAIN_PATH}/sysroot --gcc-toolchain=${TOOLCHAIN_PATH} -march=${MARCH} -static")
```
PhilippvK, line 24: I think we should also expose `-mabi` to the user.

apps/microtvm/riscv/template_project/microtvm_api_server.py
```diff
@@ -0,0 +1,285 @@
...
+        flag = fcntl.fcntl(fd, fcntl.F_GETFL)
+        fcntl.fcntl(fd, fcntl.F_SETFL, flag | os.O_NONBLOCK)
+        new_flag = fcntl.fcntl(fd, fcntl.F_GETFL)
+        assert (new_flag & os.O_NONBLOCK) != 0, "Cannot set file descriptor {fd} to non-blocking"
+
+    def open_transport(self, options):
+        toolchain_path = options.get("toolchain_path")
+        target = options.get("target")
+        march = options.get("march")
+        spike_args = ["spike", f"--isa={march}", os.path.join(toolchain_path, target, "bin/pk")]
```
PhilippvK, line 211: The `--isa` argument of spike will not always be the same as the `-march` used with clang. I would prefer having isa exposed as project option (defaulting to use the same value as march) to overwrite it if required.

apps/microtvm/riscv/template_project/microtvm_api_server.py
```diff
@@ -0,0 +1,285 @@
...
+                    help="Sets the value of target.",
+                ),
+                server.ProjectOption(
+                    "march",
+                    optional=["generate_project"],
+                    type="str",
+                    default="rv64gc",
+                    help="Sets the value of target architecture.",
+                ),
+            ],
```
PhilippvK, line 99: To compare how this performs on different vector length, we should make it configurable either by exposing `vlen` and `elen` as project options or by adding a generic `spike_extra_args` option to add any user defined arguments to the cmdline. See here: https://github.com/PhilippvK/microtvm-spike-template/blob/main/template_project/microtvm_api_server.py#L136-L149

apps/microtvm/riscv/template_project/microtvm_api_server.py
katebern-grovety: On the latest version of spike, when using instruction counting, I got an error
`An illegal instruction was executed!`
now I'm using spike version 1.1.0 and this error does not occur on it.

apps/microtvm/riscv/template_project/CMakeLists.txt.template
katebern-grovety: I'll add it.

tests/python/relay/aot/riscv.mk
katebern-grovety: Yes, sure. I'll create a new PR Project API

apps/microtvm/riscv/template_project/CMakeLists.txt.template
katebern-grovety: I'll add it.

apps/microtvm/riscv/template_project/microtvm_api_server.py
katebern-grovety: I'll add isa as project option.

PR #14838, opened 2023-05-12
[Relay] Use f-strings for string formatting, NFC

kparzysz-quic:

Reformat modified files.


slyubomirsky (approved):
Thanks for tidying up the code base!

PR #14778, opened 2023-05-05
[TIR] Require buffer declaration (DeclBuffer) before use

Lunderberg:
Part of [RFC#70](https://github.com/apache/tvm-rfcs/blob/main/rfcs/0070-introducing-decl-buffer.md#engineering-plan), requiring buffers to be declared prior to their usage.  This is currently a draft PR, to see what errors are found by the CI.

PR #14840, opened 2023-05-12
[TVMScript] Print Executor, Runtime, and FunctionInfo as metadata

Lunderberg:
Instances of `relay::Executor`, `relay::Runtime`, and `relay::backend::FunctionInfo` may be inserted into a PrimFunc's attributes for bookkeeping purposes.  In these flows, the function should still be able to be printed for debugging purposes.

### DIFF
--- a/src/script/printer/ir/relay.cc
+++ b/src/script/printer/ir/relay.cc
@@ -1,0 +1,48 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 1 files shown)


junrushao (approved):
Not sure if we wanted to print anything for Relay under `printer/ir/` folder, because we are not maintaining Relay any more. However, I don't object if we printed it out if it helps with some usecases in production

Lunderberg:
That makes sense, and agreed that it doesn't make sense to pull in portions that aren't maintained.  This mainly arose while debugging some CI failures in https://github.com/apache/tvm/pull/14778, where these types being used as attributes caused some debug print statements to fail.

Maybe it would be better to have a fallback catch-all that prints the type?  That way, we avoid the relay dependency, but can still print the portions that are maintained.

PR #14842, opened 2023-05-12
[Docker] Install PyTorch on cpu image

masahi:
As discussed in https://github.com/apache/tvm/pull/14841, it is desirable to install PT on cpu image as well, to remove many workarounds in tests. Even for the gpu image, we have been using the cpu-build of PT, since we only use it for reference / model import.


yongwww (approved):
LGTM

tqchen (approved):

PR #14843, opened 2023-05-12
[LLVM][RUNTIME] Enable multi systemlib with device code

tqchen:
This PR enables combination of multiple system lib into the same static library with a system_lib_prefix attribute. This can open doors for multiple models to be compiled separately then packaged into the same app via static library.

It resolves a previous issue that prevents multiple system lib to be linked together when they come with extra binary component such as CUDA due to symbol conflict.


junrushao (approved):
LGTM! Please fix the CI failure

PR #14821, opened 2023-05-10
[Bugfix][Relay] Fix softplus about the wrong calculation formula in Relay PyTorch frontend

jikechao:
Due to lacking consider of the attribute `threshold` in `Softplus`, the inference results in TVM  are different from PyTorch.
You can see the definition of Softplus in [Pytorch documentation](https://pytorch.org/docs/1.7.1/generated/torch.nn.Softplus.html?highlight=softplus#torch.nn.Softplus:~:text=threshold%20%E2%80%93%20values%20above%20this%20revert%20to%20a%20linear%20function.%20Default%3A%2020)

### Expected Behavior
The TVM gives the same inference results as PyTorch.
### Actual Behavior
image

### Steps for reproduce
```
import torch
from tvm import relay
import tvm
import numpy as np

m = torch.nn.Softplus(1, 2,)
input_data = torch.tensor([[1.0, 4.0]], dtype=torch.float32)

torch_outputs = m(input_data)

trace = torch.jit.trace(m, input_data)
input_shapes = [('input0', torch.Size([1, 2]))]

mod, params = relay.frontend.from_pytorch(trace, input_shapes)

with tvm.transform.PassContext(opt_level=3):
    exe = relay.create_executor('graph', mod=mod, params=params, device=tvm.device('llvm', 0), target='llvm').evaluate()
input_tvm = {'input0': np.array([[1.,  4.]], dtype='float32')}
tvm_outputs = exe(**input_tvm).asnumpy()

np.testing.assert_allclose(torch_outputs, tvm_outputs, rtol=1e-3, atol=1e-3)
```

cc @Hzfengsy @echuraev

PR #14845, opened 2023-05-13
[bugfix][Relay] Fix softplus in paddlepaddle frontend

jikechao:
This PR fixed a similar bug with #14821  in the PaddlePaddle frontend.

For the [PaddlePaddle documentation](https://www.paddlepaddle.org.cn/documentation/docs/en/api/paddle/nn/Softplus_en.html), when input * beta > threshold, Softplus(x) = x. 
image

TVM miss this special case handling. It would cause wrong inference results.

A patch and a Test case that can trigger this bug are submitted in this PR.

cc @Hzfengsy @echuraev @jiangjiajun


echuraev (approved):
LGTM. Thanks

PR #14847, opened 2023-05-13
[CI] Update CPU image to install PyTorch

masahi:
Follow up to https://github.com/apache/tvm/pull/14842 to actually update the image

### DIFF
--- a/ci/jenkins/docker-images.ini
+++ b/ci/jenkins/docker-images.ini
@@ -19,7 +19,7 @@
 [jenkins]
 ci_arm: tlcpackstaging/ci_arm:20230504-142417-4d37a0a0
 ci_cortexm: tlcpackstaging/ci_cortexm:20230504-142417-4d37a0a0
-ci_cpu: tlcpackstaging/ci_cpu:20230504-142417-4d37a0a0
+ci_cpu: tlcpackstaging/ci_cpu:20230513-200357-e54bbc73
 ci_gpu: tlcpackstaging/ci_gpu:20230504-142417-4d37a0a0
 ci_hexagon: tlcpackstaging/ci_hexagon:20230504-142417-4d37a0a0
 ci_i386: tlcpackstaging/ci_i386:20230504-142417-4d37a0a0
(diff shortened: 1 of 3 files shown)


ci/jenkins/docker-images.ini
```diff
@@ -19,7 +19,7 @@
 [jenkins]
 ci_arm: tlcpackstaging/ci_arm:20230504-142417-4d37a0a0
 ci_cortexm: tlcpackstaging/ci_cortexm:20230504-142417-4d37a0a0
-ci_cpu: tlcpackstaging/ci_cpu:20230504-142417-4d37a0a0
+ci_cpu: tlcpackstaging/ci_cpu:20230513-200357-e54bbc73
```
masahi, line 22: I noticed that all images are pointing to `tlcpackstaging` account rather than `tlcpack` since https://github.com/apache/tvm/pull/14651. Should we fix them?

yongwww: looks the images were uploaded into https://hub.docker.com/u/tlcpack already, we can replace tlcpackstaging with tlcpack for the images with tag 20230504-142417-4d37a0a0.  Probably we can do it in follow-up pr after the cpu image with tag 20230513-200357-e54bbc73 is uploaded into tlcpack

yongwww: cc: @driazati

yongwww (approved):
LGTM

ci/jenkins/docker-images.ini
masahi: According to https://tvm.apache.org/docs/contribute/ci.html#updating-a-docker-image-tag, the current protocol is to reference `tlcpack` in a PR that updates an image, and CI automatically fetches from the corresponding image from `tlcpackstaging`. The images are automatically uploaded to `tlcpack` after successful merge.

masahi:
I don't get why we are still getting the `free(): invalid pointer` error even though I'm using `--link-static` and `-DHIDE_PRIVATE_SYMBOLS=ON`.

yongwww:
Previously what I did to debug the tests related to CI changes is to reproduce it locally first, the ec2 instance type for cpu container is r5.large. 
Steps to reproduce:
1). pull the docker image, and create a container from this image
2). pull the tvm code
3). build tvm (generate the same config.cmake as ci: `tests/scripts/task_config_build_cpu.sh build`), set env variables like TVM_HOME

Then the try to run the failed test, it should be reproduced.

PR #14849, opened 2023-05-14
[Bugfix][TVMScript] Capture fails if var appears only in annotation

junrushao:
Consider the case below:

```python
dtype = "float32"

@T.prim_func:
def f(
  A: T.Buffer((1, ), dtype),
  B: T.Buffer((1, ), dtype),
):
  ...
```

The variable `dtype` only appears in the type annotation of the function being parsed. In this case, the python interpreter will evaluate the annotation first before invoking the decorator, and thus if `dtype` doesn't appear in the function body, it will not be considered as being captured by the function itself. As a result, `inspect` module will be unable to supply the value of `dtype` during parsing, leading to failure.

This PR fixes the bug by maintaining a copy of function annotations that are already parsed. Whenever expression evaluation fails during parsing, it falls back to using the copy that is evaluated by python interpreter.

### DIFF
--- a/python/tvm/script/parser/core/entry.py
+++ b/python/tvm/script/parser/core/entry.py
@@ -15,12 +15,13 @@
 # specific language governing permissions and limitations
 # under the License.
 """The entry point of TVM parser."""
-
+import inspect
 from typing import Any, Dict, Union
 
 from ...ir_builder import IRBuilder
 from . import doc
 from .diagnostics import Source
+from .error import ParserError
 from .parser import Parser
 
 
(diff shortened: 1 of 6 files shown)


junrushao:
CC: @tqchen @cyx-6

cyx-6 (approved):
Great fix!

junrushao:
A side note: we might want to migrate all `report_error` to raising `ParserError` to avoid nested error reporting

Hzfengsy (approved):

PR #14850, opened 2023-05-14
[BugFix][Target][Web] Fix the wrong dtype for the shift expression

yanghaku:
This PR fixes the wrong data type for the shift expression in WGSL.

In [WGSL standard](https://www.w3.org/TR/WGSL/#bit-expr), The number of bits to shift must be u32, but the generated code may has i32 type, such as the following:
```wgsl
pad_temp_shared[((i32(threadIdx.z) * 7) + i32(threadIdx.x))] = p0[(((((rc_outer * 784) + ((i32(threadIdx.z) >> 2i) * 196)) + (i32(blockIdx.y) * 28)) + ((i32(threadIdx.z) & 3) * 7)) + i32(threadIdx.x))];
```
And the error message in tint:
```console
error: no matching overload for operator >> (i32, i32)
```

This PR adds a convert, if the number of bits to shift is i32, it will be converted to u32. E.g.   `2i` -> `u32(2i)`.

### DIFF
--- a/src/target/source/codegen_webgpu.cc
+++ b/src/target/source/codegen_webgpu.cc
@@ -252,6 +252,24 @@ void CodeGenWebGPU::VisitExpr_(const CallNode* op, std::ostream& os) {  // NOLIN
       this->EndScope(else_scope);
     }
     os << result;
+  } else if (op->op.same_as(builtin::shift_left()) || op->op.same_as(builtin::shift_right())) {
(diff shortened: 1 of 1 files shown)


tqchen:
Thanks @yanghaku . There is already a patch in unity branch https://github.com/apache/tvm/blob/unity/src/target/source/codegen_webgpu.cc#L365 you are more than welcome to try it out.

yanghaku:
@tqchen  Thanks, I'll try the unity branch and close this PR.

PR #14848, opened 2023-05-13
[Unity][CI] Update CPU image to install PyTorch

masahi:
Follow up to https://github.com/apache/tvm/pull/14842 to actually update the image.

PR #14852, opened 2023-05-15
[Unity][CI] Update CPU image to install PyTorch (fix)

masahi:
I sent https://github.com/apache/tvm/pull/14848 before but it was a mistake, for unity I need to modify this file. 

cc @junrushao @tqchen


junrushao (approved):
Thanks for the quick update!

PR #14857, opened 2023-05-15
[Bugfix] Fix batch_norm

liquanfeng:
Fix batch_norm on PyTorch frontend. 
Gamma and beta should be considered separately.


vinx13:
could you add a test case?

liquanfeng:
> could you add a test case?

Done. PTAL, thanks! @vinx13

vinx13 (approved):

PR #14858, opened 2023-05-15
[relay][simplify_expr]: Add pass to remove trivial transpose ops

f2013519:
With my BYOC flow, I see that during various transformations, sometimes trivial no-ops like relay.transpose(x, axes=[0, 1, 2, 3]) are created. I would expect these to be optimized out with the SimplifyExpr pass which is not happening currently. This creates issues when I try to perform pattern matching on the graph.

Currently, SimplifyExpr looks for back to back transpose ops to optimize. This PR splits the current pass into two allowing us to optimize out these trivial ops as well.

### DIFF
--- a/src/relay/transforms/simplify_expr.cc
+++ b/src/relay/transforms/simplify_expr.cc
@@ -261,6 +261,38 @@ class SimplifyCastClip : public DFPatternRewrite {
 };
 
 /*!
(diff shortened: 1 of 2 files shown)


f2013519:
cc @shingjan @junrushao @masahi

vinx13 (approved):

masahi (approved):

PR #14861, opened 2023-05-16
[microNPU][ETHOSU] Add offloading to the NPU the nn.avg_pool2d operator with a stride > 3

sergio-grovety:
The nn.avg_pool2d operator with a stride size greater than 3 in any of the spatial dimensions is rewritten as ethosu avg_pool with strides=[1,1] if the case satisfies the additional conditions (no AvgPool2D padding, spatial dimensions of ifm and shape of pooling are equal).

### DIFF
--- a/python/tvm/relay/op/contrib/ethosu.py
+++ b/python/tvm/relay/op/contrib/ethosu.py
@@ -650,12 +650,33 @@ class AvgPool2DParams:
         """
         This function checks whether AvgPool2D has compatible attributes with the NPU
         """
+
+        def check_extra_strides():
+            """
+            This function checks whether AvgPool2D could be legalized as ethosu_pooling
+            supported by the NPU.
+            We consider only specific case: when there is no AvgPool2D padding, the spatial
+            dimensions of ifm and the shape of pooling are equal, but stride size exceed 3
+            by any of dimensions, e.g:
+            ifm: (1, 8, 8, _), strides: (8, 8), pool_shape: (8, 8)
+            ifm: (1, 25, 5, _), strides: (25, 5), pool_shape: (25, 5)
+            """
+            if self.pooling_type != "AVG":
+                return False
+            if list(self.padding) != [0, 0, 0, 0]:
+                return False
+            if [self.ifm.shape[1], self.ifm.shape[2]] != list(self.pool_shape):
+                return False
+            if list(self.strides) != list(self.pool_shape):
+                return False
+            return True
+
         tensor_params = [self.ifm, self.ofm]
         if not check_valid_dtypes(tensor_params, supported_dtypes=[np.uint8, np.int8]):
             return False
         if self.ifm.dtype != self.ofm.dtype:
             return False
-        if not check_strides(self.strides):
+        if not check_strides(self.strides) and not check_extra_strides():
             return False
         if not check_batch_size(self.ifm):
             return False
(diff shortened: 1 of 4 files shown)


python/tvm/relay/op/contrib/ethosu.py
```diff
@@ -650,12 +650,33 @@ def is_valid(self):
...
+            """
+            This function checks whether AvgPool2D could be legalized as ethosu_pooling
+            supported by the NPU.
+            We consider only specific case: when there is no AvgPool2D padding, the spatial
+            dimensions of ifm and the shape of pooling are equal, but stride size exceed 3
+            by any of dimensions, e.g:
+            ifm: (1, 8, 8, _), strides: (8, 8), pool_shape: (8, 8)
+            ifm: (1, 25, 5, _), strides: (25, 5), pool_shape: (25, 5)
+            """
+            if self.pooling_type != "AVG":
```
ekalda, line 664: Why can't we use the same trick for max pool as well?

python/tvm/relay/op/contrib/ethosu.py
```diff
@@ -650,12 +650,33 @@ def is_valid(self):
...
+                return False
+            if list(self.padding) != [0, 0, 0, 0]:
+                return False
+            if [self.ifm.shape[1], self.ifm.shape[2]] != list(self.pool_shape):
+                return False
+            if list(self.strides) != list(self.pool_shape):
```
ekalda, line 670: Why is this condition necessary? To me it sounds like this patch supports a case where the kernel is the same shape as ifm, so the stride value doesn't matter since we can't move the kernel anyway?

python/tvm/relay/op/contrib/ethosu.py
```diff
@@ -650,12 +650,33 @@ def is_valid(self):
         """
         This function checks whether AvgPool2D has compatible attributes with the NPU
         """
+
+        def check_extra_strides():
```
ekalda, line 654: Maybe we can come up with a more self explanatory name for this function? It currently sounds as if some operators have more than one set of strides. Maybe something like `check_same_ifm_and_kernel_shape` or `legalize_large_stride`?

tests/python/contrib/test_ethosu/test_legalize.py
```diff
@@ -875,6 +875,91 @@ def verify(ext_func):
     verify(mod["tvmgen_default_ethos_u_main_0"])
 
 
+def test_tflite_pool2d_extra_strides_legalize():
```
ekalda, line 878: This legalization test is fine, it checks whether the pooling matching the specific conditions gets partitioned for microNPU. I think it would be good to also test that it gets correctly legalized to `ethosu.pooling` op (i.e. the strides will be correctly replaced with [1, 1]).

ekalda (changes requested):
Thanks @sergio-grovety, looks good in general, nice documentation! Left some suggestions and questions

python/tvm/relay/op/contrib/ethosu.py
arina-grovety: Yes, you're right, now I see redundancy as well.

Aleksei-grovety (commented):
Looks good just one suggestion.

python/tvm/relay/op/contrib/ethosu.py
arina-grovety: Hi @ekalda, exactly, this approach can also be applied to MAX_POOL, done

ekalda (approved):
LGTM! Thanks @arina-grovety and @sergio-grovety!

ekalda:
Thanks @arina-grovety and @sergio-grovety, this is now merged!

PR #14862, opened 2023-05-16
[TIR] Support PrimFunc-to-PrimFunc calls with primitive arguments

Lunderberg:
This PR allows TIR PrimFuncs to call subroutines located within the same IRModule.  The immediate goal is to make it easier to hand-write TIR that represents output of possible optimizations, to determine whether those optimizations would be worth implementing.  A longer-term goal is to enable multi-device scheduling, such as tensor-parallel compute, in which a single TIR function may be scheduled to delegate portions of the compute onto multiple devices.

The use cases enabled by this PR are shown in the unit tests in `tests/python/unittest/test_tir_subroutine_call.py`, which is the best place to start for reviewing.  (And both thanks and apologies in advance to reviewers for the larger PR!)  The changes enabling these use cases are summarized below:

* Cherry-pick TVMScript changes from unity branch, to parse `module.func_name` as a GlobalVar

* Update TIR building to only use IRModule to IRModule transforms, without using the intermediate `Map<Target, IRModule>`.  This ensures that each pass can identify the callee within the same IRModule.  External functions that accept `Map<Target, IRModule>` as input are normalized before lowering.

* Refactor `SplitHostDevice` to output PrimFunc-to-PrimFunc calls, with a subsequent `LowerDeviceKernelLaunch` pass to generate compute kernel launches, as well as handling any user-written host-to-device.

* Update several passes to accept input PrimFuncs without a `kGlobalSymbol` attribute

* Update LLVM codegen to handle `GlobalVar` as the `CallNode::op`, producing calls to the subroutine.

### DIFF
--- a/apps/extension/tests/test_ext.py
+++ b/apps/extension/tests/test_ext.py
@@ -39,7 +39,7 @@ def test_ext_dev():
     def check_llvm():
         if not tvm.testing.device_enabled("llvm"):
             return
-        f = tvm.build(s, [A, B], "ext_dev", "llvm")
+        f = tvm.build(s, [A, B], "ext_dev", "ext_dev")
         dev = tvm.ext_dev(0)
         # launch the kernel.
         a = tvm.nd.array(np.random.uniform(size=n).astype(A.dtype), dev)
--- a/include/tvm/driver/driver_api.h
+++ b/include/tvm/driver/driver_api.h
@@ -54,7 +54,8 @@ using tvm::transform::Pass;
  * \param target The device Target.
  * \return The composite Pass for the fused module.
 //  */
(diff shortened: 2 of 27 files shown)


Lunderberg:
The majority of this PR is now split out, and waiting for the components to pass CI.  Of the initial ~2.5k lines of changes, about 1k have landed so far, 1.6k are currently making their way through CI, leaving about 500 lines in the current PR.  Increased line count when splitting out the changes was largely due to increased test coverage, and was well worth the effort.

Commits https://github.com/apache/tvm/pull/14862/commits/721da7348d4e557c80b663525f65c626b0f0aa00 and https://github.com/apache/tvm/pull/14862/commits/b4025fd967cb6acbc93cd683284e20ed3b8e0943 probably deserve to be their own PRs as well, but they depend on the current batch so they can't easily be split out yet.

Issue #14829, opened 2023-05-11
[Bug][Unity] `LazyTransformParams` introduces impure actions inside `DataflowBlock`s


slyubomirsky:
The test case for the pass `LazyTransformParams` reveals a significant bug that should require rethinking how the pass works:

https://github.com/apache/tvm/blob/28567bdc248464c9b1d4f728f3aca3b64966fa90/tests/python/relax/test_transform_lazy_transform_params.py#L77-L98

Here is the intended transformed function:
```python
        @R.function
        def main_transform_params() -> R.Tuple(R.Object, R.Object):
            cls = Expected
            with R.dataflow():
                lv: R.Object = R.call_packed("get_item", R.prim_value(1), sinfo_args=(R.Object,))
                lv1: R.Object = R.call_packed(
                    "set_item", R.prim_value(0), lv, sinfo_args=(R.Object,)
                )
                lv2: R.Tuple = R.vm.kill_object(lv)
                lv1_1: R.Object = R.call_packed("get_item", R.prim_value(0), sinfo_args=(R.Object,))
                lv3 = R.call_tir(
                    cls.transform_layout_IOHW_to_OIHW,
                    (lv1_1,),
                    out_sinfo=R.Tensor((16, 3, 3, 3), dtype="float32"),
                )
                lv4: R.Object = R.call_packed(
                    "set_item", R.prim_value(1), lv3, sinfo_args=(R.Object,)
                )
                lv5: R.Tuple = R.vm.kill_object(lv1_1)
                gv: R.Tuple(R.Object, R.Object) = (lv1, lv4)
                R.output(gv)
            return gv
```
Notice that there are calls to a `PackedFunc` called `set_item` inside the dataflow block. This is an impure action and thus should not be permitted inside the dataflow block. In principle, I would argue this pass should come after `ToNonDataflow` in the phase ordering, but the pass also includes a liveness analysis that _relies_ on dataflow blocks, so that is not an option. Either way, there should be some careful redesign here. I don't think it's valid to include that mutation inside a dataflow block.

cc @quic-sanirudh

tqchen:
cc @jinhongyii  

I think ideally it should be done after ToNonDataflow

slyubomirsky:
I tried running `ToNonDataflow` (and agree that it's what one would expect to work), but the pass fails due to the liveness analysis breaking.

jinhongyii:
[...]

PR #14864, opened 2023-05-16
[Unity]Lazy transform param now only work on non-dataflow block

jinhongyii:
As #14829 mentioned, `set_item` is an impure function, so LazyTransformParam Pass should be done after manually applyiyng ToNonDataflow pass. This PR corrects the behavior.

cc: @slyubomirsky


jinhongyii:
@tqchen

slyubomirsky:
Thank you for making the change :)

PR #14874, opened 2023-05-17
[TVMScript] Use default fallback for un-registered type

Lunderberg:
Previously, even if an object is supported by TVMScript, printing that object could fail if the object contains references to un-registered objects.  In particular, the attributes in the `PrimFunc` and `AttrStmt` may contain any `ObjectRef`, and are frequently used for external bookkeeping.  This commit adds a default printer for unregistered types in TVMScript.

This is a follow-up to changes made in https://github.com/apache/tvm/pull/14840


junrushao (approved):
Thanks a lot for this change!! Instead of a direct crash, this approach provides a nice and readable safety net for all un-registered types!

junrushao (approved):
Thanks a lot for this change!! Instead of a direct crash, this approach provides a nice and readable safety net for all un-registered types!

PR #14875, opened 2023-05-17
[Quantization]: Update simulated_quantize to infer correct layout

f2013519:
In our BYOC flow, we invoke the Convert Layout pass after simulated_quantize annotations are inserted in the IR. 

However, after Convert Layout, we observe layout transforms to the original layout are inserted before the simulated_quantize annotations, as the compiler is currently unable to infer the layouts from the input.

This PR allows propagation of the input layouts to avoid the insertion of additional layout transforms.


f2013519:
cc: @junrushao @masahi @vinx13

masahi (approved):

PR #14876, opened 2023-05-18
[CONTRIB] Enable create_staticlib to take in tar files

tqchen:
This PR enahnces create_staticlib to take tar files which can be handy when combining multiple libs into a single a file.


junrushao (approved):
LGTM!

junrushao:
@echuraev Let's get another round of review and get it in!

echuraev (approved):
LGTM. Thanks!

Issue #14878, opened 2023-05-18
[Bug] [Build Failure] build the tvm source code crash and threw "use of undeclared identifier 'f'"


jikechao:
I update the TVM source code today (git_id: b4475b83) and rebuilt it.  The build crashed and threw a crash message:
![image](https://github.com/apache/tvm/assets/29506758/7bd79fd7-463c-4ca3-903f-96c789ad2611)

This crash was introduced after PR #14570 3 days before. 

### Environment

clang version 8.0.0-3~ubuntu16.04.1
TVM: git_id: b4475b83 (May 18, 2023)

### Steps to reproduce

git pull
make

cc @Lunderberg @kparzysz-quic @tqchen

Lunderberg:
@jikechao Thank you for the bugix!  Since #14879, can we close this issue?

PR #14879, opened 2023-05-18
[bugfix] fix the undeclared identifier 'f'

jikechao:
This pr fixed the source code build problem found in issue #14878 

The variable 'f' is a local variable that is inviable for the outside code. This pr moved the 'f' definition to the outside of the condition and fix the crash.

cc @Lunderberg @kparzysz-quic

### DIFF
--- a/src/target/llvm/codegen_cpu.cc
+++ b/src/target/llvm/codegen_cpu.cc
@@ -466,13 +466,14 @@ llvm::Value* CodeGenCPU::CreateCallExtern(Type ret_type, String global_symbol,
   llvm::FunctionType* ftype = llvm::FunctionType::get(GetLLVMType(ret_type), arg_types, false);
   // Check if it is available in global function table as injected function.
 
+  llvm::Function* f = module_->getFunction(MakeStringRef(global_symbol));
   auto callee = [&]() -> llvm::Value* {
     if (auto it = gv_func_map_.find(global_symbol); it != gv_func_map_.end()) {
       if (it->second == nullptr) {
         it->second = InitContextPtr(ftype->getPointerTo(), "__" + global_symbol);
       }
       return GetContextPtr(it->second);
-    } else if (llvm::Function* f = module_->getFunction(MakeStringRef(global_symbol))) {
+    } else if (f) {
       return f;
     } else {
       return llvm::Function::Create(ftype, llvm::Function::ExternalLinkage,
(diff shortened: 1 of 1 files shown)


jikechao:
@tvm-bot rerun

src/target/llvm/codegen_cpu.cc
```diff
@@ -466,13 +466,14 @@ llvm::Value* CodeGenCPU::CreateCallExtern(Type ret_type, String global_symbol,
   llvm::FunctionType* ftype = llvm::FunctionType::get(GetLLVMType(ret_type), arg_types, false);
   // Check if it is available in global function table as injected function.
 
+  llvm::Function* f = module_->getFunction(MakeStringRef(global_symbol));
```
Lunderberg, line 469: Instead of exposing `llvm::Function* f` outside the scope, could we instead replace `auto ext_callee = f;` on line 490 with `auto ext_callee = callee;`?  The current change would allow compilation, but would skip the optional pointer-cast of the callee.

Lunderberg (changes requested):
Thank you for the catch!  I think we could have a smaller fix within the ifdef, which would also avoid needing to expose the definition of `f` to the larger scope.

src/target/llvm/codegen_cpu.cc
```diff
@@ -466,13 +466,14 @@ llvm::Value* CodeGenCPU::CreateCallExtern(Type ret_type, String global_symbol,
...
   auto callee = [&]() -> llvm::Value* {
```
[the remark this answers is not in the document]
Lunderberg: The `callee` was intended to be used in both cases, with the `#if` normalizing `callee` into either a `llvm::Value*` or a `llvm::FunctionCallee` depending on the LLVM version.  I think we should keep the definition of `callee` outside the conditional, but should replace `auto ext_callee = f;` with `auto ext_callee = callee;`.  That way, `callee` is initialized as appropriate module/context/external function pointer in both branches, and is then converted to the type required by the LLVM API inside the directive.

src/target/llvm/codegen_cpu.cc
Lunderberg: Or, alternatively, I think we could simplify it by returning `builder_->CreateCall(ftype, callee, arg_values);`.  This method is available as far back as 3.7.0 ([commit link](https://github.com/llvm/llvm-project/commit/3a6a69c535066c7cadc434d693eeeb5e5886abb1)), and since TVM requires LLVM 4.0 or higher, we would avoid needing the `#if` directive altogether.

jikechao: @Lunderberg @quic-sanirudh Thanks for your guidance. I have corrected the patch, could you recheck the code change again?

junrushao (approved):
Thanks for fixing the legacy issue!

Lunderberg (approved):
Thank you for the changes, and looks good to me!

quic-sanirudh (approved):
LGTM, Thanks. Guess we can merge this now as Eric has already approved it.

PR #14885, opened 2023-05-18
[bugfix] [Relay] fix broadcast in PyTorch frontend 

jikechao:
`broadcast_shapes` is a newly introduced operator after PyTorch 1.8.0. If you install PyTorch < 1.8.0, the usage `torch.broadcast_shapes` will lead to a crash and throw:   **AttributeError: module 'torch' has no attribute 'broadcast_shapes'**.

This PR replaces `broadcast_shapes` with an old operator `broadcast_tensors`, which was introduced since PyTorch 1.0.0.

### Reference: [PyTorch Documentation](https://pytorch.org/docs/1.8.0/generated/torch.broadcast_shapes.html?highlight=broadcast_shapes#torch.broadcast_shapes)
image

### Reproducible script
```
import torch
from tvm import relay
import tvm

m = torch.nn.L1Loss()
input_shape_ = [10, 3]
input_data=[torch.randn(input_shape_, dtype=torch.float64), torch.randn(input_shape_, dtype=torch.float64)]
trace = torch.jit.trace(m, input_data)

#input_shapes = [('input0', torch.Size(input_shape))]
input_shapes = [('input0', torch.Size(input_shape_)), ('input1', torch.Size(input_shape_))]
mod, params = relay.frontend.from_pytorch(trace, input_shapes)

    exe = relay.create_executor('vm', mod=mod, params=params, device=tvm.device('llvm', 0), target='llvm').evaluate()
```

BTW, this bug can be triggered by existing test cases. Thus, we needn't add new test cases.

cc @echuraev @vvchernov @yongwww

### DIFF
--- a/python/tvm/relay/frontend/pytorch.py
+++ b/python/tvm/relay/frontend/pytorch.py
@@ -2261,7 +2261,11 @@ class PyTorchOpConverter:
         import torch
 
         infer_shape_value = [self.infer_shape(t) for t in tensor_list]
-        res_shape = list(torch.broadcast_tensors(*map(torch.empty, infer_shape_value))[0].shape)
+        # "torch.broadcast_shapes" is available after PyTorch 1.8.0
+        if hasattr(torch, "broadcast_shapes"):
+            res_shape = list(torch.broadcast_shapes(*infer_shape_value))
+        else:
+            res_shape = list(torch.broadcast_tensors(*map(torch.empty, infer_shape_value))[0].shape)
         return [_op.broadcast_to(tensor, res_shape) for tensor in tensor_list]
 
     def Bool(self, inputs, input_types):
--- a/python/tvm/relay/frontend/pytorch.py
+++ b/python/tvm/relay/frontend/pytorch.py
@@ -2260,7 +2260,8 @@ class PyTorchOpConverter:
         tensor_list = inputs[0]
         import torch
 
-        res_shape = list(torch.broadcast_shapes(*[self.infer_shape(t) for t in tensor_list]))
+        infer_shape_value = [self.infer_shape(t) for t in tensor_list]
+        res_shape = list(torch.broadcast_tensors(*map(torch.empty, infer_shape_value))[0].shape)
         return [_op.broadcast_to(tensor, res_shape) for tensor in tensor_list]
 
     def Bool(self, inputs, input_types):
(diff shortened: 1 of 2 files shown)


python/tvm/relay/frontend/pytorch.py
```diff
@@ -2260,7 +2260,8 @@ def broadcast_tensors(self, inputs, input_types):
         tensor_list = inputs[0]
         import torch
 
-        res_shape = list(torch.broadcast_shapes(*[self.infer_shape(t) for t in tensor_list]))
+        infer_shape_value = [self.infer_shape(t) for t in tensor_list]
```
shingjan, line 2263: If we are trying to be backward-compatible here, would that be possible that we can check if the `broadcast_shape` available first, and fall back to `broadcast_tensors.shape` if not so?

shingjan (changes requested):

python/tvm/relay/frontend/pytorch.py
jikechao: @shingjan Thanks for your suggestion. Adding an available check seems a better solution. This pr has been changed!

echuraev (commented):
Could you please add a unit test?

jikechao:
@echuraev Thanks for your review. Test cases have been added. These test cases can cover the bugging line. However only when the PyTorch version < 1.8.0 in your environments, these test cases can trigger this crash.

tests/python/frontend/pytorch/test_forward.py
```diff
@@ -4817,6 +4818,7 @@ def test_forward_mse_loss():
     verify_model(torch.nn.MSELoss().eval(), input_data=[predictions, targets])
     verify_model(torch.nn.MSELoss(reduction="sum").eval(), input_data=[predictions, targets])
     verify_model(torch.nn.MSELoss(reduction="none").eval(), input_data=[predictions, targets])
+    verify_model(torch.nn.MSELoss().eval(), input_data=[predictions, targets])
```
shingjan, line 4821: This test seems redundant to me. If there is no good way of mocking a missing attribute, I would prefer just leaving the comment.

shingjan (changes requested):

tests/python/frontend/pytorch/test_forward.py
```diff
@@ -4798,6 +4798,7 @@ def test_forward_l1_loss():
     verify_model(torch.nn.L1Loss().eval(), input_data=[predictions, targets])
     verify_model(torch.nn.L1Loss(reduction="sum").eval(), input_data=[predictions, targets])
     verify_model(torch.nn.L1Loss(reduction="none").eval(), input_data=[predictions, targets])
+    verify_model(torch.nn.L1Loss().eval(), input_data=[predictions, targets])
```
echuraev, line 4801: It looks like the same test exists two lines above. If there is no any other way to test this change then I agree with @shingjan that the added tests can be removed and just the comment will be enough.

jikechao:
@shingjan @echuraev Thanks for your scrutiny. These redundant test cases have been removed.

echuraev (approved):
LGTM. Thanks

PR #14886, opened 2023-05-18
[Relay][PyTorch] Add tvm_relay_backend as torch_dynamo_backends

shingjan:
This is a move to have our own registration of tvm as one of the backends of `torch.compile()`. Follows the contract of [Custom Backends of PT2.0](https://pytorch.org/docs/stable/dynamo/custom-backends.html). Put it as draft for discussions.

cc: @masahi


python/tvm/relay/frontend/pytorch.py
```diff
@@ -5005,3 +5006,180 @@ def from_pytorch(
         export_c_graph(export_renamed_c_graph_path, graph)
 
     return transform.RemoveUnusedFunctions()(mod), tvm_params
+
+
+def tvm_relay_backend(gm, example_inputs, *, scheduler=None, trials=20000):
```
FrozenGene, line 5011: It is better to add VM backend too.

PR #14889, opened 2023-05-19
[TIR][TVMScript] Added format/parsing of subroutine calls

Lunderberg:
Similar to `module.relax_func(args)` syntax used when parsing Relax functions, this allows `module.tir_func(args)` to be used when parsing TIR PrimFuncs.

### DIFF
--- a/src/script/printer/ir/ir.cc
+++ b/src/script/printer/ir/ir.cc
@@ -97,8 +97,6 @@ TVM_STATIC_IR_FUNCTOR(IRDocsifier, vtable)
         }
       }
       return HeaderWrapper(d, ClassDoc(module_doc, {IR(d, "ir_module")}, (*f)->stmts));
-      // return HeaderWrapper(d, ClassDoc(IdDoc(GetBindingName(d).value_or("Module")),
-      //                                  {IR(d, "ir_module")}, (*f)->stmts));
     });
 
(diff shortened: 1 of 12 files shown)


src/script/printer/ir/ir.cc
```diff
@@ -84,8 +96,9 @@ TVM_STATIC_IR_FUNCTOR(IRDocsifier, vtable)
           (*f)->stmts.push_back(Downcast<FunctionDoc>(doc));
         }
       }
-      return HeaderWrapper(d, ClassDoc(IdDoc(GetBindingName(d).value_or("Module")),
-                                       {IR(d, "ir_module")}, (*f)->stmts));
+      return HeaderWrapper(d, ClassDoc(module_doc, {IR(d, "ir_module")}, (*f)->stmts));
+      // return HeaderWrapper(d, ClassDoc(IdDoc(GetBindingName(d).value_or("Module")),
+      //                                  {IR(d, "ir_module")}, (*f)->stmts));
```
csullivan, line 101: Remove

csullivan (approved):

src/script/printer/ir/ir.cc
Lunderberg: Thank you for the catch, and removed!

PR #14892, opened 2023-05-19
[TIR] Expand unit tests for ConvertSSA

Lunderberg:
Prior to this PR, there was a single test which invoked ConvertSSA and checked that no error was thrown.  This PR adds additional test cases for nested variable definition, and for variables de-duplicated across separate function calls in an `IRModule`.

Of the additional tests, the behavior tested by `TestDedupAutoBroadcastBuffer` and `TestReusedBufferParameter` fails on main, and is resolved by this PR.


Lunderberg:

tests/python/unittest/test_tir_transform_convert_ssa.py
```diff
@@ -0,0 +1,255 @@
...
+class TestReuseInSequentialLetStmt(BaseBeforeAfter):
+    """De-dup sequential variable bindings"""
+
+    def before(self):
+        @T.prim_func
+        def func():
+            i1 = T.handle("int32")
+            with T.LetStmt(i1, var=i1):
+                T.evaluate(0)
+            with T.LetStmt(i1, var=i1):
```
tqchen, line 36: I am not sure if TVMScript support creation of non-SSA form, since  future parsers could choose to do renaming, or not support such case. 

It would be safer to do things with ir builder

tqchen (changes requested):

tests/python/unittest/test_tir_transform_convert_ssa.py
Lunderberg: Good point, especially since all the before cases are malformed in some way.  The IRModule-wide checks should be safe, since the duplication is introduced after parsing, but both `TestReuseInSequentialLetStmt` and `TestReuseInNestedLetStmt` should be updated to prevent future improvements to the parser from fixing them before the test runs.

Lunderberg: Unit tests are now updated so that all SSA violations are introduced outside of TVMScript.

tqchen (approved):

PR #14895, opened 2023-05-19
[Unity] Allow eliminating only call nodes in CSE pass

masahi:
The CSE pass eliminates expressions very eagerly. This can result in an undesirable result, for example `relax.arange(...)` op  which requires all of its inputs to be `PrimValue` complains when its inputs are CSE variables. 

https://github.com/apache/tvm/blob/unity/src/relax/op/tensor/create.cc#L248-L253

```
  File "/Users/masa/projects/dev/tvm/src/relax/ir/block_builder.cc", line 138
TVMError: Arange expects the `start` to be a PrimValue, but got relax.expr.DataflowVar
```

That error can be easily fixed, but I've also hit similar issues with other ops like `strided_slice` as well. Since I only need common CallNodes to be eliminated, I'm adding an option to avoid the default fine-grained elimination.

cc @slyubomirsky @sunggg

### DIFF
--- a/include/tvm/relax/transform.h
+++ b/include/tvm/relax/transform.h
@@ -169,8 +169,9 @@ TVM_DLL Pass CanonicalizeBindings();
  *
  * \note For functions local to dataflow blocks, this pass performs
  * CSE *within* those functions.
+ * \param call_only If true, enable eliminating only call nodes.
  */
-TVM_DLL Pass EliminateCommonSubexpr();
+TVM_DLL Pass EliminateCommonSubexpr(bool call_only = false);
 
 /*!
  * \brief Bind params of function of the module to constant tensors.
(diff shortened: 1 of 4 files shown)


slyubomirsky:
The proposed changes are implemented correctly and I don't object to them per se. But I wonder if this is the right solution. Perhaps we should exclude PrimValues from CSE in general instead? There isn't really a good reason to CSE them in the first place. Maybe there are other types of expressions we shouldn't CSE either.

slyubomirsky (approved):
I've decided to approve because, while we might consider a setting to leave only scalar constants and PrimValues (we could call them atomic values), calls-only is indeed a reasonable simplification that will account for almost all expensive computations. (Other situations besides calls, though, where CSE is useful include large tensor constants, repeated inner functions, and large tuples.)

PR #14898, opened 2023-05-20
[TVMScript] set default global_symbol of T.prim_func

lyuchuny3:
set the default global_symbol of prim_func at the stage of `@T.prim_func`
so that the user no need to add it manually.


lyuchuny3:
@junrushao , please help to review this change, thanks

tqchen:
Thanks @lyuchuny3 , we intentionally avoid global symbol setting so that the name can be changed in the future

tqchen:
https://github.com/apache/tvm/issues/14899

PR #5107, opened 2020-03-19
[REFACTOR][TIR][TARGET] Make target codegen take IRModule and PrimFunc.

tqchen:
As part of the unified IR refactor.
This PR refactors the target codegen to use IRModule containing tir::PrimFuncs.

In order to break the refactor into several steps without breaking the codebase,
we built an conversion pass to convert Array<LoweredFunc> into IRModule.

The follow-up refactors will gradually move the passes covered by IRModule up
until we cover all the passes. Then we can remove the additional redundant
concepts such as LoweredFunc.

PR #14901, opened 2023-05-21
[LLVM] Codegen subroutine call when CallNode::op is GlobalVar

Lunderberg:
Previously, the `CallNode::op` must be a known built-in operation.  This commit allows LLVM codegen to produce a subroutine call to another function within the same IRModule, with `CallNode::op` specifying the `GlobalVar` that represents that function.

### DIFF
--- a/src/target/llvm/codegen_amdgpu.cc
+++ b/src/target/llvm/codegen_amdgpu.cc
@@ -87,9 +87,9 @@ class CodeGenAMDGPU : public CodeGenLLVM {
   CodeGenAMDGPU() = default;
   virtual ~CodeGenAMDGPU() = default;
 
-  void AddFunction(const PrimFunc& f) final {
+  void AddFunction(const GlobalVar& gvar, const PrimFunc& f) final {
     // add function as void return value
-    CodeGenLLVM::AddFunctionInternal(f, true);
+    CodeGenLLVM::AddFunctionInternal(gvar, f, true);
     function_->setCallingConv(llvm::CallingConv::AMDGPU_KERNEL);
     std::ostringstream attr;
     attr << "1," << DetectROCMmaxThreadsPerBlock();
@@ -262,11 +262,7 @@ runtime::Module BuildAMDGPU(IRModule mod, Target target) {
 
   cg->Init("TVMAMDGPUModule", llvm_target.get(), NullOpt, false, false);
 
-  cg->AddFunctionsOrdered(mod->functions.begin(), mod->functions.end(), [](auto& kv) {
-    ICHECK(kv.second->template IsInstance<PrimFuncNode>())
-        << "Can only lower IR Module with PrimFuncs";
-    return Downcast<PrimFunc>(kv.second);
-  });
+  cg->AddFunctionsOrdered(mod->functions.begin(), mod->functions.end());
 
   llvm::TargetMachine* tm = llvm_target->GetOrCreateTargetMachine();
   const auto* find_rocm_bitcodes = tvm::runtime::Registry::Get("tvm_callback_rocm_bitcode_path");
--- a/src/target/llvm/codegen_cpu.cc
+++ b/src/target/llvm/codegen_cpu.cc
@@ -220,18 +220,17 @@ llvm::DISubprogram* CodeGenCPU::CreateDebugFunction(const PrimFunc& f) {
 #endif
 }
 
-void CodeGenCPU::AddFunction(const PrimFunc& f) {
+void CodeGenCPU::AddFunction(const GlobalVar& gvar, const PrimFunc& f) {
 #if TVM_LLVM_VERSION >= 50
   di_subprogram_ = CreateDebugFunction(f);
 #endif
   EmitDebugLocation(f->span);
-  CodeGenLLVM::AddFunction(f);
+  CodeGenLLVM::AddFunction(gvar, f);
   if (f_tvm_register_system_symbol_ != nullptr) {
-    auto global_symbol = f->GetAttr<String>(tvm::attr::kGlobalSymbol);
-    ICHECK(global_symbol.defined())
-        << "CodeGenLLVM: Expect PrimFunc to have the global_symbol attribute";
-    export_system_symbols_.emplace_back(
-        std::make_pair(global_symbol.value().operator std::string(), function_));
+    if (auto global_symbol = f->GetAttr<String>(tvm::attr::kGlobalSymbol)) {
+      export_system_symbols_.emplace_back(
+          std::make_pair(global_symbol.value().operator std::string(), function_));
+    }
   }
   AddDebugInformation(f, function_);
 }
--- a/src/target/llvm/codegen_cpu.h
+++ b/src/target/llvm/codegen_cpu.h
@@ -67,7 +67,7 @@ class CodeGenCPU : public CodeGenLLVM {
   void Init(const std::string& module_name, LLVMTarget* llvm_target,
             Optional<String> system_lib_prefix, bool dynamic_lookup,
             bool target_c_runtime) override;
-  void AddFunction(const PrimFunc& f) override;
+  void AddFunction(const GlobalVar& gvar, const PrimFunc& f) override;
   void AddMainFunction(const std::string& entry_func_name) override;
   std::unique_ptr<llvm::Module> Finish() override;
   void VisitStmt_(const AssertStmtNode* op) override;
--- a/src/target/llvm/codegen_hexagon.cc
+++ b/src/target/llvm/codegen_hexagon.cc
@@ -547,7 +547,6 @@ runtime::Module BuildHexagon(IRModule mod, Target target) {
 
   auto cg = std::make_unique<CodeGenHexagon>();
 
-  std::vector<PrimFunc> funcs;
   std::string entry_func;
 
   for (auto kv : mod->functions) {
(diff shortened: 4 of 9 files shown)


Lunderberg:

masahi (approved):

LeiWang1999:
looks like after this pr was merged,  the ret val is 

code to reproduce:

```python
import tvm
from tvm import te
import numpy as np
import tvm.testing
from tvm.tir import TensorIntrin


M = 64
N = 64

@tvm.script.ir_module
class MyModule:
    @T.prim_func
        T.func_attr({"global_symbol": "main"})
        A = T.match_buffer(a, (M, N), dtype="float32")
        B = T.match_buffer(b, (M, N), dtype="float32")
        for i, j in T.grid(M, N):
            with T.block("B"):
                B[vi, vj] = A[vi, vj] * 2.0


ir_module = MyModule
sch = tvm.tir.Schedule(ir_module, debug_mask="all")

block_b = sch.get_block("B")

i, j = sch.get_loops(block_b)
sch.bind(i, "blockIdx.x")
sch.bind(j, "threadIdx.x")

print(sch.mod["main"].script())

ctx = tvm.rocm(0)
with tvm.transform.PassContext():
    rocm_mod = tvm.build(sch.mod, target="rocm")

'''
  File "../../tvm_rocm/memory_copy.py", line 38, in <module>
    rocm_mod = tvm.build(sch.mod, target="rocm -mcpu=gfx90a")
  File "/home/aiscuser/v-leiwang3/tvm/python/tvm/driver/build_module.py", line 281, in build
    rt_mod_host = _driver_ffi.tir_to_runtime(annotated_mods, target_host)
  File "/home/aiscuser/v-leiwang3/tvm/python/tvm/_ffi/_ctypes/packed_func.py", line 238, in __call__
    raise get_last_ffi_error()
tvm._ffi.base.TVMError: Traceback (most recent call last):
  7: TVMFuncCall
  6: tvm::runtime::PackedFuncObj::Extractor<tvm::runtime::PackedFuncSubObj<tvm::runtime::TypedPackedFunc<tvm::runtime::Module (tvm::runtime::Map<tvm::Target, tvm::IRModule, void, void> const&, tvm::Target)>::AssignTypedLambda<tvm::{lambda(tvm::runtime::Map<tvm::Target, tvm::IRModule, void, void> const&, tvm::Target)#6}>(tvm::{lambda(tvm::runtime::Map<tvm::Target, tvm::IRModule, void, void> const&, tvm::Target)#6}, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >)::{lambda(tvm::runtime::TVMArgs const&, tvm::runtime::TVMRetValue*)#1}> >::Call(tvm::runtime::PackedFuncObj const*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, tvm::runtime::TVMRetValue)
  5: tvm::TIRToRuntime(tvm::runtime::Map<tvm::Target, tvm::IRModule, void, void> const&, tvm::Target const&)
  4: tvm::codegen::Build(tvm::IRModule, tvm::Target)
  3: tvm::runtime::PackedFuncObj::Extractor<tvm::runtime::PackedFuncSubObj<tvm::runtime::TypedPackedFunc<tvm::runtime::Module (tvm::IRModule, tvm::Target)>::AssignTypedLambda<tvm::runtime::Module (*)(tvm::IRModule, tvm::Target)>(tvm::runtime::Module (*)(tvm::IRModule, tvm::Target), std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >)::{lambda(tvm::runtime::TVMArgs const&, tvm::runtime::TVMRetValue*)#1}> >::Call(tvm::runtime::PackedFuncObj const*, tvm::runtime::TVMArgs, tvm::runtime::TVMRetValue*)
  2: tvm::codegen::BuildAMDGPU(tvm::IRModule, tvm::Target)
  1: tvm::codegen::CodeGenLLVM::Finish()
  0: tvm::codegen::CodeGenLLVM::Verify() const
  File "/home/aiscuser/v-leiwang3/tvm/src/target/llvm/codegen_llvm.cc", line 361
TVMError: LLVM module verification failed with the following errors: 
Calling convention requires void return type
ptr @main_kernel
Function return type does not match operand type of return inst!
  ret void
 i32
'''
```

this can be resolved through:

```c++
src/target/llvm/codegen_amdgpu.cc
class CodeGenAMDGPU : public CodeGenLLVM {
 public:
  CodeGenAMDGPU() = default;
  virtual ~CodeGenAMDGPU() = default;

  llvm::Function* DeclareFunction(const GlobalVar& gvar, const PrimFunc& f) {
    // amd gpu kernel function should be void return
    return this->DeclareFunctionInternal(gvar, f, true);
  }
```

because amd gpu kernels should be void return but by default it wll call a llvm codegen implementation which returns int by default, Previously there were no issues, until this code snippet was added.

```c++
// src/target/llvm/codegen_llvm.cc 
  if (auto it = functions_.find(gvar.get()); it != functions_.end()) {
    return it->second;
  }
```

and if we wanna leverage shared memory, there's another issue:

```
TVMError: LLVM module verification failed with the following errors: 
Global is external, but doesn't have external or weak linkage!
ptr addrspace(3) @shmem
Global is external, but doesn't have external or weak linkage!
ptr addrspace(3) @shmem.1
```

Please take a look @Lunderberg

tqchen:
We should not support external global in the device codegen

tqchen:
Following up on this. Device side codegen follows a different convention and likely we cannot simply have a call from the host to device via the external call. Instead, we need to do them through packed func calls (and we need to do them via PackedFunc convention) 

So I wonder if we should revert some of the changes in the device codegen(nvptx, rocm) and only support subroutine rewriting in host side.

Lunderberg:
Agreed, the device-side codegen should not handle host-to-device calls.  If any such calls exist (e.g. resulting from SplitHostDevice), they are lowered to `PackedFunc` API in `LowerDeviceKernelLaunch`.

I'm investigating the issue at the moment, and can reproduce it locally.  (Thank you @LeiWang1999 for the test case!)  It looks like it is a discrepancy between the `PrimFunc::ret_type` and the `CallNode::dtype`, currently checking where that discrepancy gets introduced.

Lunderberg:
@LeiWang1999 It looks like the discrepancy is specifically between the return type of the PrimFunc (`DataType::Void()`) and the return type provided to LLVM-IR (`i32`).  I can confirm that your fix (adding `CodeGenAMDGPU::DeclareFunction`) does avoid the issue in the test script, but I can't reproduce the shared memory issue.  Is that a different test case for that failure mode?

Long-term, rather than providing the `ret_void = true` argument, it may be more robust to use the `PrimFunc::ret_type` directly.  That is, because the `PrimFunc::ret_type` is void, then the `ret_void` argument duplicates information without adding expressibility.  There's a `TODO` in the comments (from [PR#5107](https://github.com/apache/tvm/pull/5107/files#diff-f61b04b100f5145f2681340c81d3f2af221239594ed01e2e24896522329ce92cR125)) stating that this was an original intention of the code.

Lunderberg:
> So I wonder if we should revert some of the changes in the device codegen(nvptx, rocm) and only support subroutine rewriting in host side. cc @Lunderberg

@tqchen Forgot to respond to this part.  That is the intent, that the LLVM subroutines are supported only on the host side.  The primary change to nvptx/rocm were (a) to make sure they could provide the `GlobalVar` when delegating to `CodeGenLLVM::AddFunction` and (b) to call `DeclareFunction` prior to `AddFunction`.  If we want to explicitly prevent TIR-to-TIR calls in nvptx/rocm, we could revert the changes of type (b) without impacting the changes of type(a).  I suppose we could also move the callee resolution into `CodeGenLLVM::Finish`, since that would avoid the need for `DeclareFunction` altogether, but that would be a larger change.

Lunderberg:
@LeiWang1999 Can you try running your test case with PR #15127 ?  This is a draft PR that removes the `ret_void` argument altogether, instead using `prim_func->ret_type`.  This resolved the failure mode in the test case you provided, though I haven't tested it against the shared memory issue you mentioned.

LeiWang1999:
Thanks @Lunderberg , I will take a look later, and to reproduce the shared memory external global issue, the test case can be accessed through https://gist.github.com/LeiWang1999/d0db6e802f358a2a42fca32b5cd42441

Lunderberg:
@LeiWang1999 Thank you for the test case.  I can reproduce the shmem error, though it looks like it predates the LLVM changes in this PR.  Unfortunately, that means there isn't a clear before/after that could be compared and fixed.  Do you know if the shared memory on rocm has worked in the past?

FYI, I also condensed the test case down the the following test case, which reproduces the same error:

```python
import tvm

@T.prim_func
def func():
    A = T.alloc_buffer(1, scope="shared")

    for i in T.thread_binding(1, thread="threadIdx.x"):
        A[0] = 0.0

tvm.build(func, target="rocm")
```

LeiWang1999:
@Lunderberg at least from this commit https://github.com/apache/tvm/commit/1d9863470e0e97413d05b98f2852dc7de60611a0 the rocm shared memory should work.

LeiWang1999:
> @LeiWang1999 Can you try running your test case with PR #15127 ? This is a draft PR that removes the `ret_void` argument altogether, instead using `prim_func->ret_type`. This resolved the failure mode in the test case you provided, though I haven't tested it against the shared memory issue you mentioned.

I think ret void argument issue has been fixed through the related pr, thanks @Lunderberg ! but from my test the external global function issue is still remains.

zhangxiao-stack:
> @LeiWang1999 Thank you for the test case. I can reproduce the shmem error, though it looks like it predates the LLVM changes in this PR. Unfortunately, that means there isn't a clear before/after that could be compared and fixed. Do you know if the shared memory on rocm has worked in the past?
> 
> FYI, I also condensed the test case down the the following test case, which reproduces the same error:
> 
> ```python
> import tvm
> from tvm.script import tir as T
> 
> @T.prim_func
> def func():
>     A = T.alloc_buffer(1, scope="shared")
> 
>     for i in T.thread_binding(1, thread="threadIdx.x"):
>         A[0] = 0.0
> 
> tvm.build(func, target="rocm")
> ```

I am  facing the same issue on the ROCm platform. Is there a solution yet?

LeiWang1999:
@zhangxiao-stack checkout this branch https://github.com/LeiWang1999/tvm/blob/lei/feat-hip/tests/python/unittest/test_tir_schedule_tensorize_hip_mfma.py#L112 this is a alter hip source code generation, it might be one possible impermanent solution.

PR #14902, opened 2023-05-21
[RUNTIME] Update Module and Registry to use String Container

tqchen:
This PR updates the Module and Registry's DLL function to use String container instead of std::string.

While it is impossible(and un-necessary) to obtain a stable ABI due to the nature of c++, it is helpful to keep small set of tvm/runtime functions to work with use a String so it is more stable across compilers.


tqchen:
@tvm-bot rerun

junrushao (approved):

PR #14903, opened 2023-05-22
[Test Cases][BugFix] Add some version check to make test cases run in all PyTorch versions

jikechao:
This PR added some version checks to avoid calling the operator/attribute, which has not been introduced yet！

When the PyTorch <=1.7.1, test cases will fail.

```
FAILED frontend/pytorch/test_forward.py::test_forward_tensor_split - AttributeError: module 'torch' has no attribute 'tensor_split'
FAILED frontend/pytorch/test_forward.py::test_grid_sample - ValueError: nn.functional.grid_sample(): expected mode to be 'bilinear' or 'nearest', but got: 'bicubic'
```

cc @echuraev @Hzfengsy @shingjan

### DIFF
--- a/tests/python/frontend/pytorch/test_forward.py
+++ b/tests/python/frontend/pytorch/test_forward.py
@@ -1012,11 +1012,13 @@ def test_forward_tensor_split():
         def forward(self, *args):
             return torch.tensor_split(args[0], self.split_size_or_sections, self.dim)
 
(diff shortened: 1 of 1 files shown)


jikechao:
@tvm-bot rerun

echuraev (approved):
LGTM!

PR #14907, opened 2023-05-22
[Unity] Support causal mask for `R.nn.attention`

cyx-6:
This PR introduces the causal mask for `R.nn.attention` and its cutlass dispatch and legalization. The `causal_mask` argument accepts 2 types of causal masks - "TopLeft" and "BottomRight". 

For example, with `seq_len = 4`, `seq_len_kv = 2`, 
mask for "TopLeft":
```
[[1, 0],
 [1, 1],
 [1, 1],
 [1, 1]]
```
mask for "BottomRight":
```
[[1, 1],
 [1, 1],
 [1, 1],
 [1, 1]]
```
with `seq_len = 2`, `seq_len_kv = 4`,
mask for "TopLeft":
```
[[1, 0, 0, 0],
 [1, 1, 0, 0]]
```
mask for "BottomRight":
```
[[1, 1, 1, 0],
 [1, 1, 1, 1]]
```

@vinx13 @masahi @yelite

### DIFF
--- a/python/tvm/relax/frontend/torch/fx_translator.py
+++ b/python/tvm/relax/frontend/torch/fx_translator.py
@@ -1022,9 +1022,7 @@ class TorchFXImporter:
         query = transpose_S_H(self.env[node.args[0]])
         key = transpose_S_H(self.env[node.args[1]])
         value = transpose_S_H(self.env[node.args[2]])
-        causal_mask = (
-            "TopLeft" if "is_causal" in node.kwargs and node.kwargs["is_causal"] is True else None
-        )
+        causal_mask = "TopLeft" if node.kwargs.get("is_causal", False) else None
 
         if len(node.args) == 4:
             mask = self.env[node.args[3]]
--- a/python/tvm/relax/frontend/torch/fx_translator.py
+++ b/python/tvm/relax/frontend/torch/fx_translator.py
@@ -1015,19 +1015,24 @@ class TorchFXImporter:
         )
 
     def _scaled_dot_product_attention(self, node: fx.node.Node) -> relax.Var:
-        assert len(node.args) <= 4, "Dropout, and causal masking are not supported."
+        assert (
+            len(node.args) <= 4
+        ), "Dropout is not supported, and is_causal should be called by kwargs."
         transpose_S_H = lambda tensor: relax.op.permute_dims(tensor, [0, 2, 1, 3])
         query = transpose_S_H(self.env[node.args[0]])
         key = transpose_S_H(self.env[node.args[1]])
         value = transpose_S_H(self.env[node.args[2]])
+        causal_mask = (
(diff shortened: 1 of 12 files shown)


masahi (commented):
* Please update `test_transform_legalize_ops.py`
* Would be great to update the FX translator too. https://github.com/apache/tvm/blob/unity/python/tvm/relax/frontend/torch/fx_translator.py#L1018

cyx-6:
@masahi The unit test for `LegalizeOps` with causal attention has been added, and the `fx_translator` has been updated as well.

python/tvm/relax/op/nn/nn.py
```diff
@@ -1035,13 +1036,32 @@ def attention(
...
+        mask for 'TopLeft':
+        [[1, 0],
+         [1, 1],
+         [1, 1],
+         [1, 1]]
+        mask for 'BottomRight':
+        [[1, 1],
+         [1, 1],
+         [1, 1],
+         [1, 1]]
```
masahi, line 1051: All elements are one. Is this correct?

cyx-6: It is wierd. Because refer to the unit test `test_codegen_cutlass.py::test_attention_causal_offload`, the result between cutlass and numpy mismatches when we use
```
[[0, 0],
 [0, 0],
 [1, 0],
 [1, 1]]
```
In other word, in cutlass impl, the "BottomRight" means `np.tril(*, k=abs(seq_len - seq_len_kv))`, rather than `k=seq_len - seq_len_kv` or vice versa. I have no idea if cutlass impl is correct.

python/tvm/relax/frontend/torch/fx_translator.py
```diff
@@ -1015,19 +1015,24 @@ def _cross_entropy(self, node: fx.node.Node) -> relax.Expr:
...
-        assert len(node.args) <= 4, "Dropout, and causal masking are not supported."
+        assert (
+            len(node.args) <= 4
+        ), "Dropout is not supported, and is_causal should be called by kwargs."
         transpose_S_H = lambda tensor: relax.op.permute_dims(tensor, [0, 2, 1, 3])
         query = transpose_S_H(self.env[node.args[0]])
         key = transpose_S_H(self.env[node.args[1]])
         value = transpose_S_H(self.env[node.args[2]])
+        causal_mask = (
+            "TopLeft" if "is_causal" in node.kwargs and node.kwargs["is_causal"] is True else None
```
masahi, line 1026: I think we can use `node.kwargs.get("is_causal", False)` to simplify the condition

python/tvm/relax/op/nn/nn.py
cyx-6: sure, I have added the detailed mask definition in the docs.

masahi (approved):

PR #14909, opened 2023-05-22
[ETHOSU][MicroNPU][Pass] Add a pass to replicate pads

sergio-grovety:
Added a pass to to handle the situation when nn.pad operator has more than one qnn.conv2d consumer.
```
       pad
     /     \
 Conv2D    Conv2D
```
In this case, because of the peculiarities of pattern parsing, conv2d does not get into the composite for the NPU. Therefore, pads are added so that each has only one consumer.


sergio-grovety:
cc @neildhickey, @ekalda, @ilyag-grovety, @Aleksei-grovety @arina-grovety
Please review the PR

Aleksei-grovety:
@tvm-bot rerun

ekalda (changes requested):
Thanks @sergio-grovety, great work! Bar @Aleksei-grovety suggestions, I only had one nit and two more general points:

1. This pass is necessary due to how we do pattern matching for Ethos-U and is therefore coupled to Ethos-U integration, so I don't think it makes much sense to have it among generic Relay transformation passes. Maybe move it to somewhere into `ethosu` namespace, e.g. into [codegen](https://github.com/apache/tvm/blob/main/python/tvm/relay/backend/contrib/ethosu/codegen.py).
2. Testing - while legalization test is the most important test for this kind of change, I think it would also be good to have a simple lightweight codegen test for this kind of graph topology. Since the pass creates a new conv2d, there is some risk of constants getting messed up, but a simple codegen test should eliminate that risk.

sergio-grovety:
@tvm-bot rerun

tests/python/contrib/test_ethosu/test_codegen.py
```diff
@@ -157,6 +157,76 @@ def conv2d_double(x):
...
+@pytest.mark.parametrize(
+    "kernel_shape, activation_function",
+    [((3, 3), "RELU"), ((1, 2), "NONE")],
+)
+@pytest.mark.parametrize("strides, dilation", [((3, 2), (1, 1))])
+@pytest.mark.parametrize("op_padding", ["SAME", "VALID"])
+@pytest.mark.parametrize("sep_padding", [(0, 0, 1, 1), (7, 5, 4, 5)])
+@pytest.mark.parametrize(
+    "op_pairs", [("conv2d", "conv2d"), ("depthwise", "depthwise"), ("conv2d", "depthwise")]
+)
```
ekalda, line 171: This is 128 codegen tests, which would noticeably increase CI time. I think the variations on `kernel_shape`, `activation_function`, `op_padding` and `sep_padding` are exercised by other tests as well, so these values could be kept constant. Also, if we are parametrizing over one value, it doesn't need to be expressed with `parametrize` and can be included into the test code.

ekalda (changes requested):
Almost there, one suggestion regarding to testing!

ekalda (approved):
LGTM! Thanks @sergio-grovety, @Aleksei-grovety and @arina-grovety!

PR #14913, opened 2023-05-22
[TIR] Handle subroutine calls in MakePackedAPI

Lunderberg:
Previously, the `MakePackedAPI` lowering pass treated all functions as externally-visible functions, which require the `"global_symbol"` attribute and are updated to use the `PackedFunc` API.  This commit updates `MakePackedAPI` to be aware of subroutines, distinguishing between externally-visible subroutines, which have the `"global_symbol"` attribute, and internal subroutines, which do not have the `"global_symbol"` attribute.

Externally-visible subroutines may be called directly by the user of a `runtime.Module`, and must use the packed func interface.  Any internal callers of an externally-visible subroutine must be updated to use the `builtin::tvm_call_cpacked()`.

Internal subroutines may not be called directly by an external user, and neither the subroutine's signature nor any internal callers should be updated to the packed func interface.

### DIFF
--- a/src/tir/transforms/make_packed_api.cc
+++ b/src/tir/transforms/make_packed_api.cc
@@ -135,6 +135,47 @@ Stmt RewriteReturn(Stmt body, Var ret_var, Var ret_tcode) {
   return rewriter(body);
 }
 
+class SubroutineCallRewriter : public StmtExprMutator {
+ public:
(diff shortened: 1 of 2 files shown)


Lunderberg:

csullivan (approved):

PR #14916, opened 2023-05-22
[TypoFix] fix some typo problem in keras frontend

jikechao:
This pr fixed a typo problem which could give an ambiguous crash message.

@echuraev @merrymercy @masahi


echuraev (approved):
LGTM. Thanks.

PR #14918, opened 2023-05-22
[TIR] SplitHostDevice, handle subroutines

Lunderberg:
This PR refactors `SplitHostDevice` into three separate transformations.  Previously, `SplitHostDevice` would replace device regions with a `builtin::tvm_call_packed()` node to replace the extracted region.  After this PR, this process is performed in three separate steps.

1. `AnnotateDeviceRegion`: Annotate the regions that should be executed on another target.
2. `SplitHostDevice`: Extract the annotated region into an independent PrimFunc, with a `GlobalVar` to represent the call from into the new subroutine.
3. `LowerDeviceKernelLaunch`: For any subroutine call where the caller and callee are on different devices, replace with a device kernel launch.

### DIFF
--- a/src/tir/transforms/lower_device_kernel_launch.cc
+++ b/src/tir/transforms/lower_device_kernel_launch.cc
@@ -36,13 +36,67 @@ namespace tvm {
 namespace tir {
 
 namespace {
-struct DeviceInfo {
+struct KernelInfo {
+  // The device on which the PrimFunc runs
   Target target;
-  Array<Var> params;
-  Optional<Array<String>> launch_params;
-  Map<String, PrimExpr> thread_extent;
-  Optional<PrimExpr> dyn_shmem_size{NullOpt};
 
+  // The externally visible symbol which may refer to the PrimFunc
+  // when launching a device kernel.
+  String global_symbol;
+
+  // The parameters accepted by the PrimFunc.  Used to rewrite
+  // `launch_args` to be in terms of the calling scope.
+  Array<Var> params;
+
+  // The launch parameters that should annotate the PrimFunc, if the
+  // kernel is ever called from the host.
+  Array<String> launch_params;
+
+  // Additional arguments which must be provided to the host-side
+  // PackedFunc.  These may be in terms of the function's parameters
+  // (e.g. a function that computes the average of `N` elements, and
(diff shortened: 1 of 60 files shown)


Lunderberg:

In addition, this PR depends on https://github.com/apache/tvm/pull/14915 for the new unit tests, and should be rebased onto main after #14915 lands.

src/tir/transforms/lower_device_kernel_launch.cc
```diff
@@ -0,0 +1,314 @@
...
+ * \brief Visitor class to collect device-side program information.
+ */
+class DeviceInfoCollector : public StmtVisitor {
+ public:
+  static KernelInfo Collect(const GlobalVar& gvar, const PrimFuncNode* func) {
+    DeviceInfoCollector collector;
+    collector.info_.target = [&]() -> Target {
+      auto target_attr = func->GetAttr<Target>(tvm::attr::kTarget).value();
+      bool is_host_func =
+          func->GetAttr<Bool>(tvm::tir::attr::kIsHostFunc).value_or(Bool(false))->value;
```
csullivan, line 72: Re. discussion offline, do we want to update to not rely on kIsHostFunc?

Lunderberg: Agreed, and updated.  The `SplitHostDevice` and `LowerDeviceKernelLaunch` no longer use the `kIsHostFunc` attribute at all.

Lunderberg:
(Merged https://github.com/apache/tvm/pull/14950 into this PR to allow CI to run.  No conflicts expected after CI, so github's squash/merge will remove it from the final commit.)

csullivan (approved):
LGTM, thank you for the refactoring! Looking forward to the additional functionality having SplitHostDevice split out into separate steps will enable during lowering.

tqchen:
there seems to be a regression caused by this PR, please look into it

Lunderberg:
Found the issue.  It was a combination of this PR and #14945, which is why it wasn't caught by CI.  A full description of the issue, and a bugfix, are in PR #14982.

tqchen:
A note that we should revisit some of the assumptions https://github.com/apache/tvm/issues/16237

PR #14922, opened 2023-05-23
[OpenCLML] Reactor and introduce on chip memory and memory planner

srkreddy1238:
Introduced thread context with CLMLWorkspace.
Organized the code as runtime, utils and memory planners Introcuded recording queue support and on chip memory support. On chip memory allocation planner to acommodate multiple tensors at a time. DDR memory planner introduced to reuse the underlaying memory across multiple tensor descriptors.

Dense layer support refactored to use GEMM.
CLML binary operators doesn't support broadcasting. Hence introduced an explicite broadcast op as a work around.

clml SDK codegen is enhanced accordingly.


src/runtime/contrib/clml/clml_memory_planner.cc
```diff
@@ -0,0 +1,269 @@
...
+      if ((it->second.first >= size) &&
+          (layer->ddr_storage_ref_map.find(it->first) == layer->ddr_storage_ref_map.end())) {
+        // Found a buffer in global pool. Insert in local pool and then use.
+        if (best_fit > it->second.first) {
+          memptr = it->first;
+          best_fit = it->second.first;
+        }
+      }
+    }
+  }
```
echuraev, line 231:
```suggestion
  }
  // No available buffer in local pool, look for global pool
  for (auto it = cws->ddr_global_pool.begin(); it != cws->ddr_global_pool.end(); it++) {
    if ((it->second.first >= size) &&
        (layer->ddr_storage_ref_map.find(it->first) == layer->ddr_storage_ref_map.end())) {
      // Found a buffer in global pool. Insert in local pool and then use.
      if (best_fit > it->second.first) {
        memptr = it->first;
        best_fit = it->second.first;
      }
    }
  }
```

src/runtime/contrib/clml/clml_memory_planner.cc
```diff
@@ -0,0 +1,269 @@
...
+    layer->ddr_storage_ref_map.insert(
+        {memptr, std::make_pair(cws->ddr_global_pool[memptr].first, true)});
+    return memptr;
+  } else {
+    // Allocate a fresh buffer in global then use in local pool.
+    LOG_MEM << "Allocating fresh buffer in global pool";
+    memptr = AllocateDDRTensorMemory(size);
+    cws->ddr_global_pool.insert({memptr, std::make_pair(size, 1)});
+    layer->ddr_storage_ref_map.insert({memptr, std::make_pair(size, true)});
+  }
```
echuraev, line 245:
```suggestion
  }
  // Allocate a fresh buffer in global then use in local pool.
  LOG_MEM << "Allocating fresh buffer in global pool";
  memptr = AllocateDDRTensorMemory(size);
  cws->ddr_global_pool.insert({memptr, std::make_pair(size, 1)});
  layer->ddr_storage_ref_map.insert({memptr, std::make_pair(size, true)});
```

src/runtime/contrib/clml/clml_utils.cc
```diff
@@ -0,0 +1,258 @@
...
+cl_ml_tensor_qcom DeviceMakeCLMLTensor(cl_context context, tensor_dims_t dims,
+                                       cl_ml_tensor_layout_qcom layout, cl_channel_type dtype) {
+  cl_ml_tensor_qcom tensor;
+  cl_int result = CL_OUT_OF_RESOURCES;
+
+  cl_ml_tensor_desc_qcom desc = {
+      dtype, layout, dims.n, dims.c, dims.h, dims.w, 0, CL_TENSOR_DIMENSIONS_4D_QCOM, {0}};
+  result = CLML_INTF->clCreateMLTensorQCOM(CLML_CTX, nullptr, &desc, &tensor);
+  ICHECK(tensor && result == CL_SUCCESS) << "clCreateMLTensorQCOM:" << result;
+  (void)result;
```
echuraev, line 94: Why do you need this line?

src/runtime/contrib/clml/clml_runtime.cc
```diff
@@ -198,7 +229,7 @@ class CLMLRuntime : public JSONRuntimeBase {
         if (header != kTVMCLMLTuningCacheMagic) break;
         if (!strm->Read(&reserve)) break;
         if (!strm->Read(&tune_symbol)) break;
-        LOG(INFO) << "Tuning Cache Symbol:" << tune_symbol;
+        // LOG(INFO) << "Tuning Cache Symbol:" << tune_symbol;
```
echuraev, line 232: Probably can be removed?

echuraev (commented):
Is it necessary to add new tests for memory planner?

srkreddy1238:
> Is it necessary to add new tests for memory planner?

We definitely need few test cases. Let me find a way of exposing the plan to verify externally.

echuraev:
Probably you can take a look at the OpenCL tests: https://github.com/apache/tvm/blob/main/tests/cpp-runtime/opencl/opencl_texture_pool_test.cc

src/runtime/contrib/clml/clml_utils.cc
srkreddy1238: typo . Removing.

srkreddy1238:
@echuraev have you ever built gtests (opencl-cpptest bin) for Android ?

echuraev:
> @echuraev have you ever built gtests (opencl-cpptest bin) for Android ?

Yes, I did it. As far as I remember it won't print test results on the host machine, you have to take a look at the log file of `tvm_rpc` or it is necessary to extend the current implementation. I suppose that the same tests for hexagon can transfer test output through RPC to host. 

You can build `opencl-cpptest` by the following commands:
```
$ cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_TOOLCHAIN_FILE=${ANDROID_NDK}/build/cmake/android.toolchain.cmake -DANDROID_ABI=arm64-v8a -DANDROID_NATIVE_API_LEVEL=android-23 -DCMAKE_FIND_ROOT_PATH_MODE_PACKAGE=ON -DANDROID_STL=c++_static -DUSE_OPENCL=ON -DUSE_CPP_RPC=ON -DUSE_HEXAGON_SDK=OFF -DUSE_OPENCL_GTEST=/path/to/googletests ..

$ make -j8 opencl-cpptest
```

cmake/modules/OpenCL.cmake
```diff
@@ -59,20 +59,33 @@ if(USE_OPENCL)
...
-    add_executable(opencl-cpptest ${OPENCL_TEST_SRCS})
-    target_link_libraries(opencl-cpptest PRIVATE gtest_main tvm_runtime)
+    if(Build_OpenCL_GTests)
+        message(STATUS "Building OpenCL-Gtests")
+        tvm_file_glob(GLOB_RECURSE OPENCL_TEST_SRCS
+          "${CMAKE_SOURCE_DIR}/tests/cpp-runtime/opencl/*.cc"
+        )
+        add_executable(opencl-cpptest ${OPENCL_TEST_SRCS})
+        target_link_libraries(opencl-cpptest PRIVATE gtest_main tvm_runtime)
+    endif()
```
echuraev, line 88: I suppose it might be useful to write a warning message in case when we cannot build OpenCL GTests.

tests/cpp-runtime/opencl/clml_memory_planner.cc
```diff
@@ -0,0 +1,481 @@
...
+  }
+}
+
+void CompareDDRPlan(tvm::runtime::contrib::CachedLayer& layer, const std::vector<int>& ddr_plan) {
+  for (auto& nid : ddr_plan) {
+    EXPECT_EQ(layer.ddr_alloc_plan.find(nid) == layer.ddr_alloc_plan.end(), false);
+  }
+}
+
+TEST(CLMLMemoryPlanner, sequential_all_on_chip) {
```
echuraev, line 119: You can use `TEST_F` instead of `TEST`. In this case, you can move all common parts to the test class. As an example, you can take a look at: https://github.com/apache/tvm/blob/main/tests/cpp-runtime/opencl/opencl_compile_to_bin.cc

tests/cpp-runtime/opencl/clml_memory_planner.cc
```diff
@@ -0,0 +1,481 @@
...
+            << " On-Chip:" << in_chip_total_alloc << " Reject:" << total_reject;
+
+  for (auto it = cws->ddr_global_pool.begin(); it != cws->ddr_global_pool.end(); it++) {
+    LOG(WARNING) << "DDR Global pool - size:" << it->second.first << " Ref:" << it->second.second;
+  }
+  for (auto it = layer.ddr_storage_ref_map.begin();
+       it != layer.ddr_storage_ref_map.end(); it++) {
+    LOG(WARNING) << "DDR Local pool - size:" << it->second.first << " Ref cnt:" << it->second.second;
+  }
+#endif
```
echuraev, line 103: Do you really need this debug part?

tests/cpp-runtime/opencl/clml_memory_planner.cc
```diff
@@ -0,0 +1,481 @@
...
+#if defined(TVM_GRAPH_EXECUTOR_CLML)
+#include "../src/runtime/contrib/clml/clml_memory_planner.h"
+#include "../src/runtime/contrib/clml/clml_runtime.h"
+#include "../src/runtime/opencl/opencl_common.h"
+#include "../src/runtime/texture.h"
+
+using namespace tvm::runtime;
+using namespace tvm::runtime::cl;
+
+void InitMemoryPlan(tvm::runtime::contrib::CachedLayer& layer) {
```
echuraev, line 32: I saw that in class `CLMLRuntime` you have almost the same methods. Probably it is better to test them? You can change `private` to `protected` and inherit `CLMLRuntime` class in your test class. It should be like a test wrapper above `CLMLRuntime` class.

srkreddy1238: Only ```InitMemoryPlan``` can be reused but ```PlanMemory```  has dependency with ```JSONRuntime``` nodes. initializing CLMLRuntime here requires Json graph and it's dependents. Hence I tried isolating the test environment within "CahcedLayer``` object. Let me see how much I can reuse from CLMLRuntime.

tests/cpp-runtime/opencl/clml_memory_planner.cc
srkreddy1238: Yes, I will leave this. Helps to diagnose quickly what's going on with failed test cases. Let me see if I can associate this with failure status.

echuraev: I believe you can do that in this way: `testing::Test::HasFailure()`

echuraev (approved):
LGTM. Thanks!

PR #14930, opened 2023-05-23
Fix pytorch axis

mikeseven:
relay.frontend.from_pytorch maps torch.squeeze() using 1 axis while pytorch allows multiple axis.
This fixes that. 
Note that this pattern seems to be used for other functions. This frontend may need to be updated elsewhere. I only tested for squeeze for my current needs.

cc @areusch


python/tvm/relay/frontend/pytorch.py
```diff
@@ -399,7 +399,10 @@ def squeeze(self, inputs, input_types):
             axis = None
         else:
             # TODO (t-vi): why is the cast to int needed? similarly elsewhere
-            axis = [int(inputs[1])]
+            # [MBS] squeeze axis can be a list
+            # axis = [int(inputs[1])]
```
echuraev, line 403: Probably the old code can be removed?

echuraev (commented):
Please add a unit test for your changes.

python/tvm/relay/frontend/pytorch.py
mikeseven: yes, removed.

mikeseven: test added with other tests for this frontend.

tests/python/frontend/pytorch/test_forward.py
```diff
@@ -578,9 +578,15 @@ class Squeeze2(Module):
...
             return args[0].squeeze(1)
 
+    class Squeeze3(Module):
+        def forward(self, *args):
+            return args[0].squeeze((1, 3))
+
     input_data = torch.rand(input_shape).float()
     verify_model(Squeeze1().float().eval(), input_data=input_data)
     verify_model(Squeeze2().float().eval(), input_data=input_data)
+    if package_version.parse(torch.__version__) >= package_version.parse("2.0.0"):
```
echuraev, line 588: Why did you add this check on torch version?

tests/python/frontend/pytorch/test_forward.py
```diff
@@ -578,9 +578,15 @@ class Squeeze2(Module):
         def forward(self, *args):
             return args[0].squeeze(1)
 
+    class Squeeze3(Module):
+        def forward(self, *args):
+            return args[0].squeeze((1, 3))
```
echuraev, line 583: Probably this code can be modified and all these classes can be squeezed into one class. This one class will take the arguments for squeeze.

tests/python/frontend/pytorch/test_forward.py
mikeseven: Because support for list of int for axis is only in Pytorch 2. In Pytorch 1.x, its only 1 int.

echuraev: Got it, thank you!

tests/python/frontend/pytorch/test_forward.py
echuraev: I took a look in another tests. And many of them are implemented in the same style. So, let's keep it as you have implemented. Thank you!

PR #14931, opened 2023-05-23
[Unity][Transform] Fix bug for tir expression in shape in fuse_tir

Ubospica:
Currently fuse_tir cannot handle cases when the shape of arguments of call_tir contains tir expressions (e.g. `n-1` other than `n`). 

In this case, 
1. The tir vars in the tir expression will be collected and provided as the third parameter of call_tir
2. fuse_tir has a special logic to handle these vars. However, it wrongly checks such vars cannot already exist in `symbolic_var_remap`. In fact, the var has been stored in `symbolic_var_remap` when previously matching arguments to tir buffers, where the shape of the argument is visited.

This PR fixes this problem by using `symbolic_var_matcher.Match` instead.


Hzfengsy (approved):
Generally LGTM

PR #14932, opened 2023-05-23
[Unity][Training] Enhance op gradient

Ubospica:
This PR updates the logic of the `_fit_shape` helper function to first use BlockBuilder to acquire the shape, instead of providing the shape. 

It will make it easier for expressions that are hard to provide shapes, such as matmul.

The correctness of these op gradients are checked in `tests/python/relax/test_op_gradient_numeric.py`, and does not need additional checks.


Ubospica:
@tvm-bot rerun

Ubospica:
@tvm-bot rerun

PR #14941, opened 2023-05-24
[TVMScript] Prevent bool to int conversion in T.Assert condition

Lunderberg:
Previously, while literal `True` and `False` values were converted to `tvm::Bool` instances, constant-foldable expressions (e.g. `0 == 1`) would be evaluated to `True`, but were then passed directly to the FFI.  Because the FFI uses the same representation for integer and boolean values, the conversion to `PrimExpr` resulted in a `tvm::Integer` instead of `tvm::Bool`.

This commit converts the argument of `T.Assert` to a `tvm::Bool` before calling the FFI, avoiding the ambiguity.  In addition, the `AssertStmt` constructor now validates the datatype of the condition, to prevent it from re-occurring.

This was first caught in the unit test `test_debug_info.py::test_llvm_ir_debug_info`, which failed validation on some versions of LLVM due to the use of `i32` as the condition of an assert.

### DIFF
--- a/python/tvm/script/ir_builder/tir/ir.py
+++ b/python/tvm/script/ir_builder/tir/ir.py
@@ -862,6 +862,8 @@ def Assert(condition: PrimExpr, message: str) -> frame.AssertFrame:  # pylint: d
     res : frame.AssertFrame
         The result AssertFrame.
     """
+    if isinstance(condition, bool):
+        condition = IntImm("bool", condition)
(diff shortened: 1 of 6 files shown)


junrushao (approved):
LGTM!

Lunderberg:
Rebased onto https://github.com/apache/tvm/pull/14950, so CI can use its bugfix.

PR #14942, opened 2023-05-24
[TIR] Preserve existing kTarget function attribute in BindTarget

Lunderberg:
Previously, if a function already has a `tvm::attr::kTarget` attribute, it will be overwritten by the `tir.BindTarget` transform. This commit updates the behavior such that `tir.BindTarget` adds annotations to functions that are missing a target annotation, but preserves any existing target annotations.

### DIFF
--- a/include/tvm/target/target.h
+++ b/include/tvm/target/target.h
@@ -218,6 +218,9 @@ class Target : public ObjectRef {
    */
   static Target WithHost(const Target& target, const Target& host);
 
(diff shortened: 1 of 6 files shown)


junrushao (approved):
LGTM!

Lunderberg:
Rebased onto main, failing tests required #14915.

Lunderberg:
Rebased onto https://github.com/apache/tvm/pull/14950, so CI can run with its bugfix.

PR #14943, opened 2023-05-24
[TIR] Restrict tir.transform.InstallDebugSpans to host functions

Lunderberg:
Previously, the `tir.InstallDebugSpans` pass required the module to contain only a single PrimFunc.  This commit relaxes the requirement, to require a single host-side PrimFunc, and to ignore any other device-side functions.


Lunderberg:
PRs #14943, #14944, and #14945 perform analogous changes to three different host-only lowering passes.   Each of these three contains the commits that add utility functions `tvm::tir::IsHostFunc` and `TargetNode::HasKey`, then apply the new utility to a specific pass.  Whichever of the three lands first can have the others rebased on top to prevent any conflicts.

These PRs in turn are a subset of the functionality in #14862.  As that PR had become unmanageably large, I am carving it out into several smaller independent PRs for ease of review.

junrushao (approved):
tbh i dont know if this pass is actually used but lets get it in

Lunderberg:
> tbh i dont know if this pass is actually used but lets get it in

It looks like it requires the `"tir.enable_debug"` option in the PassConfig.  I'm not sure how often that gets enabled, but it was added relatively recently (~5 months ago by @driazati ), so I'm guessing there's some usage of it.

There's also some potential improvements, since an entire IRModule can be printed out, which could probably remove the host-only and one-function-only limitations, but I didn't want to mix functionality changes with implementation changes at the moment.

Lunderberg:
@tvm-bot rerun

PR #14949, opened 2023-05-25
[CodegenC] Updated unit test for sorted CodegenC output

Lunderberg:
Previously, this unit test generated a `Map<tvm::Target, IRModule>` whose default iteration order was not sorted by function name, built the `Map` of modules, then validated whether the resulting C code was a sorted list of 4 elements.  However, this condition was stricter than necessary, as it depended on the number of items added to the `Map` until it was unsorted.

This commit updates the test to instead validate that `std::is_sorted` returns true.


Lunderberg:
This PR resolves a failure that resulted from changes in development of https://github.com/apache/tvm/pull/14862, but is otherwise independent of that PR.

masahi (approved):

PR #14914, opened 2023-05-22
[TIR] Handle subroutine calls in MakeUnpackedAPI

Lunderberg:
Prior to this commit, MakeUnpackedAPI required all functions to be annotated with `kGlobalSymbol` (`"global_symbol"`).  This commit updates the transformation to apply only to functions that have the `kGlobalSymbol` attribute, and to update any internal callsites of the modified functions.

This is analogous to the changes made in https://github.com/apache/tvm/pull/14913, which updates `MakePackedAPI`.

PR #14950, opened 2023-05-25
[Bugfix] Avoid symbol conflicts in MakePackedAPI/MakeUnpackedAPI

Lunderberg:
PRs https://github.com/apache/tvm/pull/14913 and https://github.com/apache/tvm/pull/14914 made analogous changes to `MakePackedAPI` and `MakeUnpackedAPI` to handle subroutine calls. Both PRs introduced the same symbol, `tvm::tir::SubroutineCallRewriter`, a local utility to update internal calls to a modified function.  While each PR passed CI individually, and was therefore able to merge, having both changes caused a duplicate symbol.

This commit updates `MakePackedAPI` and `MakeUnpackedAPI` to place their local utilities into anonymous namespaces, avoiding the conflict.

### DIFF
--- a/src/tir/transforms/make_packed_api.cc
+++ b/src/tir/transforms/make_packed_api.cc
@@ -42,6 +42,7 @@ namespace tir {
 
 static constexpr const char* kDeviceContextVar = "device_api_context";
 
+namespace {
 class ReturnRewriter : public StmtMutator {
  public:
   explicit ReturnRewriter(Var ret_var, Var ret_tcode) : ret_var_(ret_var), ret_tcode_(ret_tcode) {}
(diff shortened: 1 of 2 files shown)


csullivan (approved):

tqchen:
does this hint perhaps we should have a common utility somehow as internal header

csullivan:
Thanks @tqchen. Yeah, had the same thought when discussing offline with @Lunderberg but perhaps we can remove MakeUnpackedAPI altogether later.

In discussing with @Lunderberg it seems the special case handling MakeUnpackedAPI introduces for AOT should be able to be removed once annotations are used for whether a function requires packing or not. MakePackedAPI could then code generate the packed api when the functions are annotated and AOT functions would remain unpacked as they lack the annotation.

csullivan (approved):

PR #14951, opened 2023-05-25
[TIR] Keep trivial LetStmt in tir.Simplify when used in buffer decl

Lunderberg:
Prior to this commit, any trivial let binding of `var1 = var2` is inlined.  However, buffer definitions are not updated, so this can result in dangling `tir::Var` instances.  This commit updates the `tir.Simplify` pass to keep trivial let bindings if they are used as part of a buffer definition.

Ideally, the trivial `LetStmt` variable would be inlined into the buffer definition as well as other expressions.  However, because a buffer may be implicitly declared, the first usage may be within a constrained context.  If that happens, the simplified shape/strides expression cannot be used to update the buffer definition, as that simplification is not valid at all possible usage points of the buffer.

```python
for i in range(n):
    elem_offset = i
    view = T.Buffer(1, data=buf, elem_offset = elem_offset)
    if i == 0:
        # First occurrence in TIR is here, where elem_offset would
        # simplify to zero.
        view[0] = 1
    else:
        # But the same buffer is used here, where elem_offset doesn't
        # simplify to zero.
        view[0] = 2
```

This will be resolvable after https://github.com/apache/tvm/pull/14778 lands, requiring all buffers to be declared with `DeclBuffer` prior to usage.

```python
for i in range(n):
    elem_offset = i
    # All variables used by the DeclBuffer are valid across the entire
    # body of the DeclBuffer.
    view = T.decl_buffer(1, data=buf, elem_offset = elem_offset)
    if i == 0:
        view[0] = 1
    else:
        view[0] = 2
```

### DIFF
--- a/src/tir/transforms/simplify.cc
+++ b/src/tir/transforms/simplify.cc
@@ -33,6 +33,7 @@
 
 #include "../../arith/ir_mutator_with_analyzer.h"
 #include "../../tir/analysis/control_flow_graph.h"
+#include "../../tir/analysis/var_use_def_analysis.h"
 
 namespace tvm {
 namespace arith {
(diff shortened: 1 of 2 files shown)


Lunderberg:
@tvm-bot rerun

tests/python/unittest/test_tir_transform_split_host_device.py
```diff
@@ -168,5 +168,21 @@ def main_kernel(n: T.int32):
         return mod
 
 
+class TestSplitHostDevice(BaseCompare):
```
masahi, line 171: Is this test relevant? Looks like a dup from https://github.com/apache/tvm/pull/14988

masahi (approved):

tests/python/unittest/test_tir_transform_split_host_device.py
Lunderberg: Hmm, it was a dup from https://github.com/apache/tvm/pull/14982, which I had based this PR on top of to make allow the CI to use that bugfix.  I think the "squash and merge" should remove it from this PR as part of merging, but just to be on the safe side, rebasing onto main.

Lunderberg: And after rebase, no changes remain in `test_tir_transform_split_host_device.py`.

PR #14952, opened 2023-05-25
[TIR] Avoid re-defining `var = arg_var` in ArgBinder

Lunderberg:
Prior to this commit, `ArgBinder` would always introduce a new variable to represent the input argument, even if the argument already a primitive type.  This introduces trivial let bindings that are expected to be simplified out, but which can produce dangling `tir::Var` usage in some cases (see https://github.com/apache/tvm/pull/14951).

This commit updates `ArgBinder` to prefer using the original `tir::Var` when possible.  That is, when a function takes `n: T.int32` as input, the packed function should produce a binding `n: T.int32 = T.tvm_struct_get(...)`, rather than producing a binding `arg_n = T.tvm_struct_get(...)` followed by `n = arg_n`.

### DIFF
--- a/src/tir/transforms/make_packed_api.cc
+++ b/src/tir/transforms/make_packed_api.cc
@@ -259,24 +259,13 @@ PrimFunc MakePackedAPI(PrimFunc func) {
     return res;
   };
 
-  // Need to re-declare vars, in case some arguments also appears in the buffer.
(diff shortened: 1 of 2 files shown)


Lunderberg:
@tvm-bot rerun

masahi (approved):

PR #14958, opened 2023-05-25
[Codegen][LLVM] Allow void return type from PackedFunc 

Lunderberg:
Previously, calling a packed func that returns void would result in an error being raised from `tir::APIType`, as there is no runtime representation of a void type.  This commit updates `CodeGenCPU::MakeCallPackedLowered` to only read the return value and type of a `PackedFunc` when the TIR return type is non-void.


Lunderberg:
To allow CI to run, the commit for the bugfix in https://github.com/apache/tvm/pull/14950 is in this PR, but is the same commit ID and shouldn't result in any merge conflicts.

tqchen (approved):

PR #14962, opened 2023-05-26
[METAL] Fix int8 vectorized cast

Hzfengsy:
Current codegen output `(half4)*(device uint*)A` tries to create a `int32` number and then cast it to `half4`, which is not the expected behavior.

As Metal natively supports `uchar4` and `char4` types ([ref](https://developer.apple.com/documentation/metal/mtldatatype/uchar4)), we can direct use them to solve that problem.

cc @tqchen


echuraev (approved):
LGTM!

tqchen (approved):

PR #14963, opened 2023-05-26
[Bugfix][PyTorch] Support use_input_stats in instance_norm

### LINKED ISSUE #14926: [Bug] [Relay] attribute track_running_stats of InstanceNorm lead to wrong inference results
For the layer `InstanceNorm1d` or `InstanceNorm3d`, if attribute `track_running_stats` was set as `True`, TVM will give different inference results with PyTorch.

### Expected behavior
For the same input data, TVM and PyTorch give the same inference results.

### Actual behavior

image

### Steps to reproduce
```
import torch
from tvm import relay
import tvm
import numpy as np

m = torch.nn.InstanceNorm1d(3, track_running_stats=True).eval()
input_data = [torch.randn([1, 3, 5], dtype=torch.float32)]

torch_outputs = m(*[input_.clone() for input_ in input_data ])
trace = torch.jit.trace(m, input_data)
input_shapes = [('input0', torch.Size([1,3,5]))]

mod, params = relay.frontend.from_pytorch(trace, input_shapes)

    exe = relay.create_executor('aot', mod=mod, params=params, device=tvm.cpu(0), target='llvm').evaluate()
input_tvm = {'input0': np.array(input_data[0], dtype='float32')}
tvm_outputs = exe(**input_tvm).asnumpy()

np.testing.assert_allclose(torch_outputs, tvm_outputs, rtol=1e-3, atol=1e-3)
```

### Triage
* frontend:torch
* needs-triage

cc @shingjan

liquanfeng:
Fix #14926 .
The input should be normalized with the provided running mean and running variance instead of computing the mean and variance from the input data when use_input_stats is set to false.

### DIFF
--- a/python/tvm/relay/frontend/pytorch.py
+++ b/python/tvm/relay/frontend/pytorch.py
@@ -1385,6 +1385,9 @@ class PyTorchOpConverter:
         data = inputs[0]
         data_type = input_types[0]
         channels = self.infer_shape(data)
+        running_mean = inputs[3]
+        running_var = inputs[4]
+        use_input_stats = inputs[5]
 
         if isinstance(inputs[1], _expr.Expr) and isinstance(inputs[2], _expr.Expr):
             scale = center = True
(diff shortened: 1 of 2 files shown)


python/tvm/relay/frontend/pytorch.py
```diff
@@ -1385,6 +1385,9 @@ def instance_norm(self, inputs, input_types):
         data = inputs[0]
         data_type = input_types[0]
         channels = self.infer_shape(data)
+        running_mean = inputs[3]
```
shingjan, line 1388: Is this a most recent api change from pytorch or a long ignored params on our side? The former will need us to think about BC.

liquanfeng: The latter I think. The "track_running_stats" has been ignored in the unit test.
By the way, there will be another PR later about BC, also on a similar issue.

masahi (approved):

PR #11038, opened 2022-04-18
[OpenCL] Fix type casting

echuraev:
The previous PR https://github.com/apache/tvm/pull/11021 was reverted in https://github.com/apache/tvm/pull/11035 due
to it affected performance of generated OpenCL code.

This PR fixed the same issue but doesn't lead to performance
degradation. Tested on Resnet50_v2 network.

PR #11021, opened 2022-04-15
[OpenCL] Fix type casting error

echuraev:
Faced situation when generated OpenCL kernel contained the following if
condition:
```
if (uint4(...) && (int4(...) == int4(...)))
```

In this case, got the following error:
"can't convert between vector values of different size ('uint4' and 'int __attribute__((ext_vector_type(4)))')"

Added casts for binary ops. But it was necessary to modify `CastFromTo`
and add new method `CastTo`. Because with `CastFromTo` the following
code was generated:
```
if (uint4(...) && (convert_uint4(int4(...)) == convert_uint4(int4(...))))
```
But the OpenCL compiler still generated the same error.

This is why added new method `CastTo`. In this method we don't check the
current type of op and just add cast to a new type.

Finally the following code will be generated:
```
if (uint4(...) && convert_uint4(convert_uint4(int4(...)) == convert_uint4(int4(...))))
```

Thanks for contributing to TVM!   Please refer to guideline https://tvm.apache.org/docs/contribute/ for useful information and tips. After the pull request is submitted, please request code reviews from [Reviewers](https://github.com/apache/incubator-tvm/blob/master/CONTRIBUTORS.md#reviewers) by @ them in the pull request thread.

PR #14972, opened 2023-05-29
[OPENCL] Always use convert_T for type conversion

tqchen:
This PR changes the Cast in OpenCL to always relying on convert_T to get closer to the spec and more reliable.

### DIFF
--- a/tests/python/unittest/test_target_codegen_opencl.py
+++ b/tests/python/unittest/test_target_codegen_opencl.py
@@ -32,10 +32,12 @@ def test_opencl_ternary_expression():
         max_lhs = tvm.tir.const(2, dtype=dtype)
         max_rhs = tvm.tir.if_then_else(A[0] > 0, true_value, false_value)
         C = te.compute((n,), lambda i: tvm.te.max(max_lhs, max_rhs), name="C")
-        s = te.create_schedule(C.op)
-        s[C].bind(s[C].op.axis[0], te.thread_axis("threadIdx.x"))
-        fun = tvm.build(s, [A, C], target)
 
+        func = te.create_prim_func([A, C])
+        sch = tvm.tir.Schedule(func)
+        (x,) = sch.get_loops(sch.get_block("C"))
+        sch.bind(x, "threadIdx.x")
+        fun = tvm.build(sch.mod, target=target)
         a = tvm.nd.empty((n,), A.dtype, dev)
         c = tvm.nd.empty((n,), A.dtype, dev)
         # Only need to test compiling here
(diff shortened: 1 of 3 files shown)


tests/python/unittest/test_target_codegen_opencl.py
```diff
@@ -150,50 +158,38 @@ def check_type_casting(ctx, n, dtype):
                 tvm.tir.all(
                     *[
                         i // block_size == tvm.tir.const(3, "int32"),
-                        i % block_size == tvm.tir.const(3, "int32"),
```
echuraev, line 153: This test has covered a bug in TVM which was fixed. Why did you change the compute function? Can we keep it as it was?

tqchen: i also migrated the original test to the latest TensorIR so the test is future proof.

Interestingly TensorIR becomes smarter :) so the original condition get simplified and becomes `i == 15` here. So I modified the test to ensure we cover most behavior we intend to cover around type casting

tqchen: Unfortunately relying on string pattern was too reliable given the set of possible transformations here. So we need to adapt the testcase as we go. If there is a new test cases in TVMScript, i think we can also put it here

echuraev: > so the original condition get simplified and becomes i == 15 here

Do you mean that this condition: 
```
tvm.tir.all(
    *[
        i // block_size == tvm.tir.const(3, "int32"),
        i % block_size == tvm.tir.const(3, "int32"),
    ]
),
```
Was transformed to `i == 15`?

I took a look at the git history and the original problem which should be covered by this test was that left and right parts of condition should have the same data type. You can see detailed description in #11021. And next this test was added in #11038.

By decreasing the number of conditions in `tvm.tir.all`, you prevent compiler to generate code such as `lcond && rcond`. Am I right?  In this case, the original problem won't be tested. Sorry if I missed something.

echuraev: I took a look at the git history and the original problem which should be covered by this test was that left and right parts of condition should have the same data type. You can see detailed description in #11021. And next this test was added in #11038.

By decreasing the number of conditions in `tvm.tir.all`, you prevent compiler to generate code such as `lcond && rcond`. Am I right?  In this case, the original problem won't be tested. Sorry if I missed something.

tqchen: Sorry I wasn't being clear, right 

```
tvm.tir.all(
    *[
        i // block_size == tvm.tir.const(3, "int32"),
        i % block_size == tvm.tir.const(3, "int32"),
    ]
)
```

In TensorIR schedule the ^ condition get transformed to `i == 15`, as indeed that makes sense. 

So we might need a different condition to test the chains.

tqchen: OK updated to a case where simplification won't happen

echuraev (approved):
LGTM! Thank you :)

tqchen:
@tvm-bot rerun

echuraev:
@tvm-bot rerun

MasterJH5574 (approved):

PR #14975, opened 2023-05-30
[CMAKE] Add a dummy target to defer libtvm dep

tqchen:
This PR add a dummy target that can be renamed as libtvm and can be used to create a dependency on libtvm without compiling everything. This can be useful for down stream in platforms such as windows.


tqchen:
@tvm-bot  rerun

junrushao (approved):

PR #14981, opened 2023-05-30
[target] Use native architecture for `llvm` target

FranklandJack:
Set the default `-device=` key for llvm targets based on the native architecture rather than hard coding to `cpu` which is x86 specific. This means that when llvm target triples are not specified  we will test `arm_cpu` schedules on Arm®-based architectures and `cpu` schedules on x86 based architectures.

Fix any schedule test failures that result from this fix.

### DIFF
--- a/python/tvm/relay/op/strategy/arm_cpu.py
+++ b/python/tvm/relay/op/strategy/arm_cpu.py
@@ -152,6 +152,7 @@ def conv2d_strategy_arm_cpu(attrs, inputs, out_type, target):
                 is_winograd_applicable = (
                     "float" in data.dtype
                     and "float" in kernel.dtype
+                    and not data.dtype.count("custom")
                     and kh == 3
                     and kw == 3
                     and stride_h == 1
@@ -284,8 +285,21 @@ def conv2d_strategy_arm_cpu(attrs, inputs, out_type, target):
                     name="depthwise_conv2d_nchw.x86",
                 )
         elif layout == "NHWC":
-            assert kernel_layout == "HWOI"
-            if target.features.has_asimd:
+            # TODO(@FranklandJack)
+            # Handle HWOI in arm_cpu schedules/compute definition.
+            if kernel_layout != "HWOI":
+                logger.warning(
+                    """depthwise_conv2d with layout NHWC and HWOI
+                               kernel layout is not optimized for arm_cpu target.
+                               """
+                )
+                strategy.add_implementation(
+                    wrap_compute_conv2d(topi.nn.depthwise_conv2d_nhwc, need_kernel_layout=True),
+                    wrap_topi_schedule(conv2d_generic.schedule_depthwise_conv2d_nhwc),
+                    name="depthwise_conv2d_nhwc.generic",
+                )
+
+            elif target.features.has_asimd:
                 strategy.add_implementation(
                     wrap_compute_conv2d(topi.arm_cpu.compute_depthwise_conv2d_nhwc),
                     wrap_topi_schedule(topi.arm_cpu.schedule_depthwise_conv2d_nhwc),
@@ -304,8 +318,11 @@ def conv2d_strategy_arm_cpu(attrs, inputs, out_type, target):
                 and kernel.shape[3] == 1  # channel_multiplier == 1
                 and out_type.dtype == "int32"
                 and (
-                    (data.shape[3] % 4 == 0 and data.dtype == "int8" and target.features.has_dsp)
-                    or (data.shape[3] % 2 == 0 and data.dtype == "int16")
+                    (
+                        (data.shape[3] % 4 == 0 and data.dtype == "int8")
+                        or (data.shape[3] % 2 == 0 and data.dtype == "int16")
+                    )
+                    and target.features.has_dsp
                 )
                 and (padding != "SAME" or data.shape[1] % stride_h == data.shape[2] % stride_w == 0)
                 # Ideally we should check that kernel is a Relay constant, but strategy functions
--- a/python/tvm/topi/arm_cpu/conv2d.py
+++ b/python/tvm/topi/arm_cpu/conv2d.py
@@ -23,10 +23,11 @@ from tvm import te
 from tvm import autotvm
 import tvm.contrib.nnpack
 
-from ..utils import traverse_inline, get_const_tuple
+from ..utils import traverse_inline, get_const_tuple, conv2d_infer_layout_helper
 from .. import nn
 from ..nn.utils import get_const_int, get_pad_tuple
 from ..nn.winograd_util import winograd_transform_matrices
+from ..nn.conv2d import conv2d_infer_layout
 from .conv2d_spatial_pack import (
     conv2d_spatial_pack_nchw,
     conv2d_spatial_pack_nhwc,
@@ -509,3 +510,8 @@ def conv2d_nhwc_dsp(cfg, data, kernel, strides, padding, dilation, out_dtype):
 def schedule_conv2d_nhwc_dsp(cfg, outs):
     """Create schedule for conv2d_nhwc_dsp"""
     return conv2d_nhwc_dsp_schedule(cfg, outs)
+
+
+@conv2d_infer_layout.register("arm_cpu")
+def _conv2d_infer_layout(workload, cfg):
+    return conv2d_infer_layout_helper(workload, cfg)
--- a/python/tvm/topi/arm_cpu/injective.py
+++ b/python/tvm/topi/arm_cpu/injective.py
@@ -69,8 +69,10 @@ def schedule_injective(outs):
     if list(s[x].op.axis):
         # do not vectorize for broadcast
         dtype = "uint16" if x.dtype == "bfloat16" else x.dtype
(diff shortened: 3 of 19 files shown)


src/target/target_kind.cc
```diff
@@ -257,6 +257,12 @@ TargetJSON TestTargetParser(TargetJSON target) {
 
 /**********  Register Target kinds and attributes  **********/
 
+#if defined(__arm__) || defined(__aarch64__)
+#define NATIVE_CPU "arm_cpu"
+#else
+#define NATIVE_CPU "cpu"
+#endif
```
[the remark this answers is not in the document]
FranklandJack: A general solution would definitely be preferable, do you have a suggestion of how one would use the target parser to do this?

Since we've gone from being x86 specfic to arm and x86 specfic I'd argue this is still an improvement over what we had before. We can always add support for more targets in future PRs.

It's worth noting that this has exposed a lot of false positive tests on the `arm_cpu` target, if the same thing happens for other targets it might be advantageous to do this iteratively in order to keep the diff contained.

python/tvm/topi/arm_cpu/conv2d.py
```diff
@@ -509,3 +510,24 @@ def conv2d_nhwc_dsp(cfg, data, kernel, strides, padding, dilation, out_dtype):
...
+    dilated_kernel_h = (k_height - 1) * hdilation + 1
+    dilated_kernel_w = (k_width - 1) * wdilation + 1
+    out_height = idxdiv(in_height + pt + pb - dilated_kernel_h, strides[0]) + 1
+    out_width = idxdiv(in_width + pl + pr - dilated_kernel_w, strides[1]) + 1
+    tile_ic, tile_oc = cfg["tile_ic"].size[-1], cfg["tile_oc"].size[-1]
+    in_shape = (batch_size, idxdiv(in_channel, tile_ic), in_height, in_width, tile_ic)
+    in_layout = f"NCHW{tile_ic}c"
+    out_shape = (batch_size, idxdiv(out_channel, tile_oc), out_height, out_width, tile_oc)
+    out_layout = f"NCHW{tile_oc}c"
+    return ((in_shape, in_layout),), ((out_shape, out_layout),)
```
ekalda, line 533: Is this function identical to the ones in `x86` and `intel_graphics`? If yes, then while you are at it, maybe move the body of that function to some common location (e.g. `python/tvm/topi/utils.py`) to reduce the amount of duplicated code.

python/tvm/relay/op/strategy/arm_cpu.py
```diff
@@ -284,8 +288,21 @@ def conv2d_strategy_arm_cpu(attrs, inputs, out_type, target):
                     name="depthwise_conv2d_nchw.x86",
                 )
         elif layout == "NHWC":
-            assert kernel_layout == "HWOI"
-            if target.features.has_asimd:
+            # TODO(@FranklandJack)
+            # Handle HWOI in arm_cpu scheudles/compute definition.
```
ekalda, line 292: Nit:
```suggestion
            # Handle HWOI in arm_cpu schedules/compute definition.
```

tests/python/unittest/test_meta_schedule_relay_integration.py
```diff
@@ -420,6 +420,7 @@ def test_extract_task_arm_conv2d_nchwc():
 
 
 def test_meta_schedule_te2primfunc_argument_order_and_lowering():
+    target = Target("llvm --num-cores=16")
```
ekalda, line 423: Just curious, what does that change do?

ekalda (changes requested):
Thanks @FranklandJack, I think it's a really good improvement to the testing of Arm targets and lowering the amount of duplicated testing we do in the upstream CI. Mostly clarifications and nits...

python/tvm/relay/op/strategy/arm_cpu.py
```diff
@@ -132,13 +132,16 @@ def conv2d_strategy_arm_cpu(attrs, inputs, out_type, target):
                         plevel=15,
                     )
                 else:
+                    # TODO(@FranklandJack)
+                    # Investigate why this producing output tensor of
+                    # incorrect dimensions in
+                    # test_runtime_module_based_interface.py
```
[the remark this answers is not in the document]
FranklandJack: That's a really good point. Perhaps the best solution here is to just fix this schedule as part of this PR?

python/tvm/relay/op/strategy/arm_cpu.py
FranklandJack: Haha whoops! I should turn my spell checker on, will fix :)

tests/python/unittest/test_meta_schedule_relay_integration.py
FranklandJack: Ach, this is a debugging artefact that I forgot to clean up, will remove. Good catch!

python/tvm/topi/arm_cpu/conv2d.py
FranklandJack: Yep it was copied directly from there. I guess this will only work for scheduling definitions that include 'tile_ic' and 'tile_oc' parameters (which is true for x86, intel_graphics and arm_cpu) but maybe not others? Is that okay? I guess we can document that.

python/tvm/relay/op/strategy/arm_cpu.py
```diff
@@ -284,8 +288,21 @@ def conv2d_strategy_arm_cpu(attrs, inputs, out_type, target):
...
+                               kernel layout is not optimized for arm_cpu target.
+                               """
+                )
+                strategy.add_implementation(
+                    wrap_compute_conv2d(topi.nn.depthwise_conv2d_nhwc, need_kernel_layout=True),
+                    wrap_topi_schedule(conv2d_generic.schedule_depthwise_conv2d_nhwc),
+                    name="depthwise_conv2d_nhwc.generic",
+                )
+
+            elif target.features.has_asimd:
```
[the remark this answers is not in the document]
FranklandJack: Ah, so I think this might be a bug. The condition I think should be `if kernel_layout != "HWOI"` so that we fallback on the generic condition in the unsupported case, does that make sense?

python/tvm/relay/op/strategy/arm_cpu.py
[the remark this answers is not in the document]
FranklandJack: For the time being I've just disabled the test that this causes to fail.

python/tvm/topi/arm_cpu/injective.py
```diff
@@ -69,8 +69,10 @@ def schedule_injective(outs):
     if list(s[x].op.axis):
         # do not vectorize for broadcast
         dtype = "uint16" if x.dtype == "bfloat16" else x.dtype
-        (io, ii) = s[x].split(list(s[x].op.axis)[-1], 16 // np.dtype(dtype).itemsize)
-        s[x].vectorize(ii)
+        # do not vectorize for custom data types
+        if 0 == dtype.count("custom"):
```
lhutton1, line 73: nit: this could just be `if dtype.count("custom")`

python/tvm/relay/op/strategy/arm_cpu.py
```diff
@@ -304,8 +321,11 @@ def conv2d_strategy_arm_cpu(attrs, inputs, out_type, target):
                 and kernel.shape[3] == 1  # channel_multiplier == 1
                 and out_type.dtype == "int32"
                 and (
-                    (data.shape[3] % 4 == 0 and data.dtype == "int8" and target.features.has_dsp)
-                    or (data.shape[3] % 2 == 0 and data.dtype == "int16")
+                    (
+                        (data.shape[3] % 4 == 0 and data.dtype == "int8")
+                        or (data.shape[3] % 2 == 0 and data.dtype == "int16")
+                    )
+                    and target.features.has_dsp
```
lhutton1, line 328: The logic seems to have changed here, was this intentional?

lhutton1 (commented):
Thanks @FranklandJack, I think it's very important to be testing the `arm_cpu` schedules in CI. Just a couple of comments

src/target/target_kind.cc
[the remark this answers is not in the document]
lhutton1: To make this more generic I wonder if we can use something like: https://github.com/apache/tvm/pull/14946/files#diff-f61b04b100f5145f2681340c81d3f2af221239594ed01e2e24896522329ce92cR2087. I agree that adding architectures iteratively will likely be simpler due to CI breakages, but this way we can set the precedence.

In addition, I think it would be good to add a warning when the host architecture could not be determined. E.g. something like: `Host architecture could not be determined. Compiling using 'cpu' fallback key which may result in unoptimized code. Please try adding a target.` (probably needs thinking through a bit more) in order to prompt users.

python/tvm/topi/arm_cpu/injective.py
FranklandJack: Sure, fixed.

python/tvm/relay/op/strategy/arm_cpu.py
lhutton1: Aha yep agreed, seems like a bug :)

lhutton1:
Closing as superseded by: https://github.com/apache/tvm/pull/16513

PR #14944, opened 2023-05-24
[TIR] Restrict tir.transform.LowerTVMBuiltin to host functions

Lunderberg:
Previously, the `tir.transform.LowerTVMBuiltin` pass applied to all functions in an `IRModule`, but was only applied to modules that contain only host functions.  This commit updates `tir.transform.LowerTVMBuiltin` to apply only to host functions.

PR #14985, opened 2023-05-30
[Driver] Single-module lowering flow in driver_api.cc

Lunderberg:
Prior to this commit, a build that used multiple targets needed to provide `tvm::build` with a `Map<Target, IRModule>` specifying which target should be used to compile each `IRModule`.  As a result, lowering passes could not introduce new targets based on a PrimFunc's content (e.g. a `with T.target()` frame to delegate out to another device), nor simplify based on cross-device subroutines (e.g. simplify a host-side conditional based on the known output of a device-side internal subroutine).

This commit makes the `tvm::attr::kTarget` attribute (`"target"`) be the single source of truth for where a `PrimFunc` will be executed. Other existing methods for specifying the target (the `target` parameter for `tvm.build`, the keys in a `Map<Target,IRModule>`, the parameter to the pass `tir::transform::BindTarget`) are still accepted as inputs, and may provide a default value for `tvm::attr::kTarget` if the attribute is missing, but may not overwrite the target attribute.

### DIFF
--- a/apps/extension/tests/test_ext.py
+++ b/apps/extension/tests/test_ext.py
@@ -39,7 +39,7 @@ def test_ext_dev():
     def check_llvm():
         if not tvm.testing.device_enabled("llvm"):
             return
-        f = tvm.build(s, [A, B], "ext_dev", "llvm")
+        f = tvm.build(s, [A, B], "ext_dev", "ext_dev")
         dev = tvm.ext_dev(0)
         # launch the kernel.
         a = tvm.nd.array(np.random.uniform(size=n).astype(A.dtype), dev)
--- a/include/tvm/driver/driver_api.h
+++ b/include/tvm/driver/driver_api.h
@@ -54,7 +54,8 @@ using tvm::transform::Pass;
  * \param target The device Target.
  * \return The composite Pass for the fused module.
 //  */
-TVM_DLL transform::Sequential MixedModulePassManager(IRModule mixed_mod, Target target);
+TVM_DLL transform::Sequential MixedModulePassManager(IRModule mixed_mod,
+                                                     Optional<Target> target = NullOpt);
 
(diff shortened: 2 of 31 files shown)


Lunderberg:
This PR is a subset of the changes introduced in https://github.com/apache/tvm/pull/14862, split out for ease of testing and reviewing.  It is currently a draft PR, as it depends on changes made in https://github.com/apache/tvm/pull/14944, and will be rebased onto main after https://github.com/apache/tvm/pull/14944 lands.

Lunderberg:
Rebased onto main now that #14944 has landed, marked as ready for review.

Lunderberg:
Converting to draft, as the fixes for the last few CI failures depend on changes in https://github.com/apache/tvm/pull/14986, and so this PR should wait until https://github.com/apache/tvm/pull/14986 has landed.

Lunderberg:
With https://github.com/apache/tvm/pull/16184 landed, this PR should (hopefully) be able to land without requiring all low-level codegen to directly support function calls.  For codegens that do not yet support function calls, device-side kernels can have subroutines inlined to avoid a recurrence of https://github.com/apache/tvm/pull/16033.

Currently, this PR is rebased on top of main, but does not use of `InlinePrivateFunctions`.  I expect it to fail in CI for Metal codegen, and to have the same ethos-u failure modes as previously.  These should be resolvable by applying `InlinePrivateFunctions`, conditional on the target being used.

Lunderberg:
The failing unit tests here are blocked by https://github.com/apache/tvm/pull/15757.  The `InlinePrivateFunctions` transform requires all buffers to be declared before use.  However, the `DeclBuffer` nodes are currently stripped out during `FlattenBuffer`, and are no longer available when required after `MakePackedAPI`.

PR #14986, opened 2023-05-30
[TIR] Move SplitHostDevice to before MakePackedAPI

Lunderberg:
This simplifies the logic used in MakePackedAPI, as it is now the last user of the host parameter in a function's target.  After MakePackedAPI, every PrimFunc has a "target" attribute without a "host".

### DIFF
--- a/python/tvm/utils/roofline/__init__.py
+++ b/python/tvm/utils/roofline/__init__.py
@@ -51,10 +51,16 @@ def _create_args(mod: IRModule, dev: Device, func_name: str = "main", remote=Non
 
 @pass_instrument
 class SaveLoweredTIR:
-    """Save TIR functions from right before final lowering. Right now this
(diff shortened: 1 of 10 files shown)


Lunderberg:
This PR is a subset of the changes introduced in https://github.com/apache/tvm/pull/14862, split out for ease of testing and reviewing.  It is currently a draft PR, as it interacts with the same functionality that is being fixed in https://github.com/apache/tvm/pull/14982, and should probably re-run CI to avoid a similar breakage being reintroduced.

While I've tested this PR against the specific VTA failure that is https://github.com/apache/tvm/pull/14982 fixes, we may want to re-run CI on top of main to avoid similar breakage.

Lunderberg:
@tvm-bot rerun

csullivan (approved):

PR #14988, opened 2023-05-30
[TIR] Handle callees on same target, different codegen

Lunderberg:
Prior to this commit, any caller that uses a different `Target` than its callee is lowered to a device-kernel launch.  However, if the caller and callee are on the same device, despite using a different target (e.g. `Target("llvm")` and `Target("c")` both use `kDLCPU`), then the kernel launch is unnecessary.

This commit updates `LowerDeviceKernelLaunch` to produce a kernel launch only when the callee is on another device, and to produce `T.call_extern` for callees on the same device.

### DIFF
--- a/src/tir/transforms/lower_device_kernel_launch.cc
+++ b/src/tir/transforms/lower_device_kernel_launch.cc
@@ -170,14 +170,27 @@ class DeviceKernelMutator : public StmtExprMutator {
   }
 
   PrimFunc UpdateKernelAttributes(const GlobalVar& gvar, PrimFunc func) const {
-    if (device_kernel_launch_.count(gvar.get())) {
+    bool is_kernel_launch = device_kernel_launch_.count(gvar.get());
+    bool is_call_extern = extern_method_call_.count(gvar.get());
+    CHECK(!is_kernel_launch || !is_call_extern)
+        << "Function " << gvar << " has multiple callees, "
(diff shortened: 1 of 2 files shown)


Lunderberg:
This is a subset of functionality from https://github.com/apache/tvm/pull/14862, split out into an independent PR for ease of testing and review.

masahi (approved):
Makes sense to me.

src/tir/transforms/lower_device_kernel_launch.cc
```diff
@@ -243,6 +275,7 @@ class DeviceKernelMutator : public StmtExprMutator {
   Optional<Target> current_target_;
   std::unordered_map<const GlobalVarNode*, KernelInfo> device_info_map_;
   std::unordered_set<const GlobalVarNode*> device_kernel_launch_;
+  std::unordered_set<const GlobalVarNode*> extern_method_call_;
```
masahi, line 278: `method` -> `func`? Since we don't use that term.

Lunderberg: Good point, and renamed `extern_method_call_` to `extern_function_call_`.

Lunderberg:
Similar to https://github.com/apache/tvm/pull/14951, rebased this PR on top of main, now that https://github.com/apache/tvm/pull/14982 has landed.

PR #14989, opened 2023-05-30
[TIR] Cleanup of MakePackedAPI

Lunderberg:
Prior to this commit, the `RequiresPackedAPI` function checked whether a function needed the packed func API.  This was used both to generate a list of call-sites to update, and as part of the updates to `PrimFunc` signatures.  However, the function that updates the `PrimFunc` signature could still return the original function unmodified, breaking internal method calls.  This occurred for functions with a `kTarget` attribute without a host.

This commit updates `MakePackedAPI` to first update all `PrimFunc` signatures that require the packed func API, then use the result to determine which call-sites must be updated.  This resolves the discrepancy for host-less target annotations, and removes the possibility of similar discrepancies in the future.


Lunderberg:
CI failures require #14986 to resolve.  Merging that dev branch into this one to run CI, and changing to draft PR as this shouldn't be merged until after #14986 lands.

PR #14990, opened 2023-05-30
[IR,TE,TIR] Use f-strings for string formatting, NFC

kparzysz-quic:

Reformat modified files.

### DIFF
--- a/python/tvm/ir/container.py
+++ b/python/tvm/ir/container.py
@@ -46,7 +46,7 @@ class Array(Object):
             raise AttributeError("handle is not set")
         if name == "type_key":
             return super().__getattr__(name)
-        raise AttributeError("%s has no attribute %s" % (str(type(self)), name))
+        raise AttributeError(f"{type(self)} has no attribute {name}")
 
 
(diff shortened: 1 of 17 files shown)


Hzfengsy (approved):
LGTM. Thanks for the work, but I'm curious about the reason behind these changes. Does it mean string formatting will be deprecated in the future by Python?

kparzysz-quic:
F-strings embed the expressions into the string itself, unlike `%` and `format`, which separate them from the string.  This typically makes f-strings more compact and easier to read.  This can be a judgment call sometimes and I tried not to sacrifice readability if the existing code was clearer than f-strings.

Maybe a small bonus for moving away from `%` is that `%` uses different format specification from `format` and f-strings.  The latter two use the same one, so that helps with consistency a little bit.

I created the first PR not knowing how it will be received, but since the reception was positive, I decided to keep going.

kparzysz-quic:
Forgot to mention---I am not aware of any intentions to deprecate `%` from python.  As far as I know it's here to stay.

PR #14994, opened 2023-05-31
[Bugfix][TVMC] Fix tvmc option for printing which operators are offloaded to the Ethos-U

sergio-grovety:
Lines from the initial Relay, which don't correspond to the `relay.Call` now are printed to the output with the `--dump-offloads` option enabled. Thus we get rid of the incomprehensible gaps in the line numbers of the initial relay, which occurred before.


sergio-grovety:
cc @neildhickey, @ekalda, @ilyag-grovety, @Aleksei-grovety @arina-grovety
Please review the PR

Aleksei-grovety (commented):
Looks good, just some suggestions.

lhutton1 (approved):
Thanks @sergio-grovety @arina-grovety, I had just one small nit

PR #14996, opened 2023-05-31
[OpenCL][Adreno] Fix conv2d when output channels < 4

echuraev:
When the number of output channels is less than 4, then we cannot pack such convolution to textures, although we can repack and extend tensors from 4d to 5d in runtime.

It is happened because function `PropBoundToInputs` is invoked for all stages in when InferBound pass or LowerSchedule function is called.

`PropBoundToInputs` has a logic in it that helps to developer to avoid out of bound access. And based on the output shape, it propagates it to inputs.

Imagine that we want to transform a 4d tensor with 3 channels to 5d, extend its number of channels to 4 and then transform it back to 4d tensor with number of channels equal to 3. Example below:
```
[1, 3, 6, 6] -> [1, 1, 6, 6, 4] -> [1, 3, 6, 6]
```
In this case, we might write a boundary check in the repacking and extending compute function to handle the case when the iterator by the channels is out of bounds for the intermediate tensor.

To avoid such problem, `PropBoundToInputs` has a logic which propagates a bounds from output tensor to input. In case when it is a possible situation that the compute has out of bounds access, then the range is decreased to value when such situation cannot be achieved. That means that for the example above the loop by channels which should filling intermediate tensor will iterate in range [0, 2] instead of [0, 3].

As it was mentioned, to avoid such problem we use buffers and cuda schedules instead of textures for cases when number of output channels is less than 4 and we cannot pack such tensor to texture. I evaluated performance of such approach and it doesn't introduce any performance degradation. For such small convolutions performance with buffers even a bit better than with textures.

### DIFF
--- a/python/tvm/relay/op/strategy/adreno.py
+++ b/python/tvm/relay/op/strategy/adreno.py
@@ -42,9 +42,18 @@ def conv2d_strategy_adreno(attrs, inputs, out_type, target):
             or (data_layout == "NCHW" and kernel_layout == "OIHW4o")
         ):
             if len(kernel.shape) == 4:
-                _, _, kh, kw = get_const_tuple(kernel.shape)
+                oc, _, kh, kw = get_const_tuple(kernel.shape)
             else:
-                _, _, kh, kw, _ = get_const_tuple(kernel.shape)
+                oc, _, kh, kw, _ = get_const_tuple(kernel.shape)
+            # We cannot use textures for case than number of channels is less than 4.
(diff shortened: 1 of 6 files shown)


elvin-n (approved):

echuraev:
@masahi, @csullivan could you please review this PR?

masahi (approved):

srkreddy1238:
Thanks for the PR @echuraev .
We too came across this and used  similar workaround for some recent popular networks.

PR #14997, opened 2023-05-31
[Hexagon] Remove "c" as aot_host_target tvm/contrib/hexagon/pytest_pl…

kparzysz-quic:
…ugin.py

After PR#14918 (changes to SplitHostDevice), LowerIntrin code generates fma, which then becomes the LLVM intrinsic fmuladd. Problem is, the code then goes to the C codegen, which knows nothing about LLVM intrinsics. The result is an abort at compile-time:
```
E     File ".../src/target/source/codegen_c.cc", line 611
E   TVMError: Unresolved call Op(tir.call_llvm_pure_intrin)
```

Why this didn't happen before is unclear at the moment, but using C codegen with Hexagon code doesn't sound like a winning combination. Hexagon-related code assumes LLVM codegen, and the above crash is the result of that.

### DIFF
--- a/python/tvm/contrib/hexagon/pytest_plugin.py
+++ b/python/tvm/contrib/hexagon/pytest_plugin.py
@@ -309,7 +309,7 @@ def terminate_rpc_servers():
         os.system("ps ax | grep tvm_rpc_x86 | awk '{print $1}' | xargs kill")
 
 
-aot_host_target = tvm.testing.parameter("c", HEXAGON_AOT_LLVM_TARGET)
+aot_host_target = tvm.testing.parameter(HEXAGON_AOT_LLVM_TARGET)
 
 
 @tvm.testing.fixture


kparzysz-quic:
cc: @Lunderberg 
Maybe you know why it worked before without spending time on investigating...

Lunderberg:
Hmm.  I think it might be related to [this change](https://github.com/apache/tvm/pull/14918/files#diff-c795c8cdf29dcfda3dcb7bf342247b22deee71e2a9c037046254625e91bdd152L264) in particular.  Previously, the host function had its `kTarget` attribute removed altogether.  Afterwards, the host function had its `kTarget` attribute updated to be the original functions `target->GetHost()`.  Effectively, that would result in any per-function host annotations being replaced with the host passed to `tvm.build`, which defaults to LLVM.

That still wouldn't explain how the LLVM intrinsics were inserted in the first place, as `LowerIntrin` should use whatever the function annotation is at the point of the pass being applied, but it might be close to an answer.

kparzysz-quic:
LowerIntrin inserted `fma`, which is a TIR builtin.  Then, Hexagon code legalized it to an LLVM intrinsic.

junrushao (approved):

PR #15003, opened 2023-06-01
[DOCKER][ADRENO] Docker setup changes for multi user environments

srkreddy1238:
Target setup shouldn't conflict while multiple users on the same target device libc++_shared.so may not be available for some android devices from system


echuraev (approved):
LGTM! Thanks for the PR.

PR #15011, opened 2023-06-02
[Vulkan] Add im2col wmma implementation for CHW direct conv2d

mei-ye:
Currently only 16x16x16 FP16 wmma fragments with FP32 intermediates are supported.  Add "min_design_space" as a parameter to give minimum design space for meta scheduler tuning. Add "use_int32_const" as a paramter to use int32 type for constants. Allow target query to be called from the schedules so that samplings are constrained to produce legal schedules. Add a unit test test_meta_schedule_vulkan_wmma.py

### DIFF
--- a/src/runtime/relax_vm/ndarray_cache_support.cc
+++ b/src/runtime/relax_vm/ndarray_cache_support.cc
@@ -267,7 +267,28 @@ class NDArrayCache {
 };
 
 TVM_REGISTER_GLOBAL("vm.builtin.ndarray_cache.get").set_body_typed(NDArrayCache::Get);
-TVM_REGISTER_GLOBAL("vm.builtin.ndarray_cache.update").set_body_typed(NDArrayCache::Update);
+TVM_REGISTER_GLOBAL("vm.builtin.ndarray_cache.update").set_body([](TVMArgs args, TVMRetValue* rv) {
+  CHECK(args.size() == 2 || args.size() == 3);
+  String name = args[0];
+  bool is_override = args.size() == 2 ? false : args[2];
+
+  NDArray arr;
+  if (args[1].type_code() == kTVMNDArrayHandle) {
+    arr = args[1];
+  } else {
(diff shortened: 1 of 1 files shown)


Hzfengsy:
cc @junrushao @vinx13 @masahi

mei-ye:
@tvm-bot rerun

tests/python/unittest/test_meta_schedule_vulkan_wmma.py
```diff
@@ -0,0 +1,203 @@
...
+
+
+def _create_tmp_database(tmpdir: str, mod_eq: str = "structural") -> ms.database.JSONDatabase:
+    path_workload = osp.join(tmpdir, "database_workloads.json")
+    path_tuning_record = osp.join(tmpdir, "database_tuning_records.json")
+    return ms.database.JSONDatabase(path_workload, path_tuning_record, module_equality=mod_eq)
+
+
+@tvm.testing.requires_vulkan
+def test_wmma(
```
masahi, line 70: This test should go into `test_meta_schedule_relay_integration.py`.

masahi (commented):
It's going to take some time to review this PR. We don't have a topi schedule that does TIR tensorization like this, so there are some design-level questions. cc @vinx13

python/tvm/topi/cuda/conv2d.py
```diff
@@ -148,3 +154,352 @@ def conv2d_backward_weight_cudnn(
...
+            ),
+            name="conv2d_nchw_mma",
+            attrs={
+                "schedule_rule": "conv2d_nchw_mma",
+            },
+        )
+    return C
+
+
+def schedule_rule_conv2d_nchw_mma(sch: Schedule, block: BlockRV):
```
masahi, line 247: Rather than implementing a manual schedule like this, I recommend trying to get auto tensorization for tensor core working for vulkan. I expect the resulting schedule to be very similar.

PR #15001, opened 2023-05-31
[Unity][CUTLASS]  Fix circular import bug in relax cutlass backend

digital-nomad-cheng:
There exists circular import bug in `tvm/python/tvm/relax/backend/contrib/cutlass.py`

Code to reproduce:
```
from tvm.contrib.cutlass import (
    has_cutlass,
    num_cutlass_partitions,
    finalize_modules
)
```

Error:      
```
File "/home/work/tmp/test.py", line 1, in <module>
    from tvm.contrib.cutlass import (
  File "/home/tvm/python/tvm/contrib/cutlass/__init__.py", line 18, in <module>
    from .build import has_cutlass, num_cutlass_partitions, finalize_modules, finalize_modules_vm
  File "/home/tvm/python/tvm/contrib/cutlass/build.py", line 28, in <module>
    from tvm import relax, relay, runtime
  File "/home/tvm/python/tvm/relax/__init__.py", line 90, in <module>
    from . import backend
  File "/home/tvm/python/tvm/relax/backend/__init__.py", line 19, in <module>
    from . import contrib
  File "/home/tvm/python/tvm/relax/backend/contrib/__init__.py", line 20, in <module>
    from .cutlass import partition_for_cutlass
  File "/home/tvm/python/tvm/relax/backend/contrib/cutlass.py", line 24, in <module>
    from tvm.contrib.cutlass.build import is_shape_valid_for_cutlass_matmul
ImportError: cannot import name 'is_shape_valid_for_cutlass_matmul' from partially initialized module 'tvm.contrib.cutlass.build' (most likely due to a circular import) (/home/tvm/python/tvm/contrib/cutlass/build.py)
```

PR #15023, opened 2023-06-05
[Unity][FIX] add init file to `relax.backend.contrib`

Hzfengsy:
This PR adds `__init__.py` to `relax.backend.contrib`, fixing the package issue reported at https://github.com/mlc-ai/mlc-llm/issues/311

cc @yzh119 @tqchen

### DIFF
--- a/python/tvm/relax/backend/contrib/__init__.py
+++ b/python/tvm/relax/backend/contrib/__init__.py
@@ -15,5 +15,3 @@
 # specific language governing permissions and limitations
 # under the License.
 """Relax backends contrib"""
-from . import cublas
-from . import cutlass
--- a/python/tvm/relax/backend/contrib/__init__.py
+++ b/python/tvm/relax/backend/contrib/__init__.py
@@ -1,0 +1,19 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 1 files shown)


masahi:
The file was removed recently to fix a circular dep problem https://github.com/apache/tvm/pull/15001

Hzfengsy:
@masahi Thanks for sharing. However, `__init__` file is needed as setuptools will ignore files if there is no `__init__` file in this folder. Maybe we need to consider an alternate way to avoid circular dep.

python/tvm/relax/backend/contrib/__init__.py
```diff
@@ -0,0 +1,19 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+"""Relax backends contrib"""
+from . import cublas
+from . import cutlass
```
masahi, line 19: Does keeping an empty init file fix the issue? We don't want `from . import cutlass` (probably cublas as well) here.

Hzfengsy: The empty init file is fine.

masahi: Here is the summary of the previous issue for reference:

* It is fine for `backend/contrib/cutlass.py` to depend on `contrib/cutlass/build.py`
* But `contrib/cutlass/build.py` does `import relax`, which transitively imports `backend/contrib/cutlass.py`

PR #15024, opened 2023-06-05
[Unity] Added bounds checking on TupleGetItem index

Lunderberg:
The index provided must be non-negative, and less than the size of the tuple.  When the tuple being accessed has `TupleStructInfo`, the upper bound can also be checked immediately.


slyubomirsky (approved):
Great change, thank you for spotting this!

PR #15029, opened 2023-06-06
[BYOC] Hide internal cutlass symbols

vinx13:
Cutlass internally may use static variables flags to control initialization. When multiple modules are loaded at runtime, duplicating symbols in the dynamic libraries cause subsequent modules failure to initialize cutlass.

cc @masahi @jwfromm @junrushao


junrushao (approved):
LGTM

PR #15031, opened 2023-06-06
[Docker] Update ci_cpu docker image to one containing polly package f…

neildhickey:
…or llvm-16


ashutosh-arm (approved):
Thanks @neildhickey. LGTM!

PR #15032, opened 2023-06-06
[AOT] Remove duplication in tvm.testing.aot.compile_models

Lunderberg:
Previously, the body of an if/else were nearly identical, except for a `with apply_fixed_config` context.  This commit removes the duplication using `contextlib.ExitStack` to conditionally enter the context.


lhutton1 (approved):
Thanks for the simplification @Lunderberg!

PR #15034, opened 2023-06-06
[TIR] Refactor BlockScope outside schedule

quic-sanirudh:
This PR does 2 major changes to TIR.

1. Move `stage_pipeline` member into `BlockInfo` from `BlockScope`
2. Extract `BlockScope` outside schedule

This is the first step of a change to expose block level dependence analysis to TIR passes. This PR moves BlockScope to be outside schedules so that it can be used in a follow-up analysis pass to compute block level dependences if needed for a transformation pass.

There is some extra discussion about this change in [this discussion forum post](https://discuss.tvm.apache.org/t/why-is-block-dependency-info-attached-to-schedules/14975)


quic-sanirudh:
cc @Hzfengsy @junrushao

Hzfengsy (approved):

PR #15050, opened 2023-06-07
[TIR] [Schedule] Fix decompose_padding bug with dtypes

quic-sanirudh:
Ran into a type mismatch error when the primfunc uses int64 dimensions, but the extent introduced in a particular case was just int32.


quic-sanirudh:
cc @wrongtest-intellif

yzh119 (approved):
LGTM

Issue #15039, opened 2023-06-06
[Bug] Installing python on Docker image fails due to require-hashes violation


Liam-Sturge:
### Expected behavior

running `ubuntu_install_python.sh` should install Python and the required dependencies in the Docker container.

### Actual behavior

Fails when attempting the command `pip3 install -U --require-hashes -r <(echo "${pip_spec}") \
     -c /install/python/bootstrap/lockfiles/constraints-${PYTHON_VERSION}.txt`.

Error produced:

```bash

Collecting cachecontrol[filecache]<0.13.0,>=0.12.9
ERROR: In --require-hashes mode, all requirements must have their versions pinned with ==. These do not:
     cachecontrol[filecache]<0.13.0,>=0.12.9 from https://files.pythonhosted.org/packages/a2/ab/fc647c0069493a904be5519dd03a535a2d49a51e98d5e3cb5d93d332293d/CacheControl-0.12.13-py2.py3-none-any.whl#sha256=431fc10c5ab1a1589ce08c05b948abac31c0f76962d5fc9efab9da280c9790aa (from poetry==1.2.0b1->-r /install/python/bootstrap/lockfiles/constraints-3.8.txt (line 202))
```

### Environment

Platform: AArch64
TVM version: latest (v0.12.0)

### Steps to reproduce

Download the latest changes in TVM (v0.12.0), then use the script ./build.sh to build a container.

Or,

Within a recently built Docker container such as tvm.ci_cpu, run `pip3 install -U --require-hashes -r <(echo "pip==22.1.2 --hash=sha256:a3edacb89022ef5258bf61852728bf866632a394da837ca49eb4303635835f17") \
     -c /install/python/bootstrap/lockfiles/constraints-3.8.txt`.

### Triage

* needs-triage
* dev:ci

ashutosh-arm:
I observed something similar while building cortex-m image on x86 machine.

Liam-Sturge:
Some tlcpack Jenkins jobs, such as the one for daily Docker image rebuilding have been failing for the same reason. See:
https://ci.tlcpack.ai/job/docker-images-ci/job/daily-docker-image-rebuild/744/console

Liam-Sturge:
Have been monitoring CI and confirmed this appears to be fixed.

PR #15056, opened 2023-06-07
[Bugfix][CI] Fix CacheControl version constraint violation

Liam-Sturge:
A change introduced in the dependencies of Poetry have introduced a bug when building Docker containers and running `ubuntu_install_python.sh`.

The problem links back to CacheControl where one of the linked dependencies is not pinned to a version.

See bug raised for TVM here: https://github.com/apache/tvm/issues/15039

A fix was released for this bug, documented here:
  https://github.com/psf/cachecontrol/issues/311

The suggested solution to the problem is to upgrade to v0.12.14, so this patch sets the version constraints accordingly.

cc @areusch @neildhickey @ashutosh-arm @tqchen @Mousius


neildhickey (approved):
LGTM

ashutosh-arm (approved):
Thanks @Liam-Sturge 👍

PR #14983, opened 2023-05-30
[CI] Make Graviton3 default AArch64 job runner node

Liam-Sturge:
In order to support SVE testing, migrating the current default AArch64 nodes to Graviton3 based nodes.

cc @areusch @leandron @Mousius @tqchen

PR #15061, opened 2023-06-08
[TOPI][Adreno] Fix problem with ceil_log2

echuraev:
On Adreno devices double type is not supported by OpenCL. Modified `ceil_log2` operation to cast value to float instead of double in case of Adreno.


masahi:
See https://github.com/apache/tvm/pull/14983#issuecomment-1583221170 for the arm build issue.

masahi (approved):

PR #15063, opened 2023-06-08
[microNPU][ETHOSU] Fix CopyComputeReordering pass arguments

Aleksei-grovety:
No arguments were passed to CopyComputeReordering pass and the same parameters were used for all targets. This fix takes arguments for CopyComputeReordering pass from Vela.
In networks tests, the amount of memory used for U65 has increased because now after CopyComputeReordering pass, the number of DMA commands executed in parallel has increased.


ekalda:
Thanks @Aleksei-grovety!

PR #15064, opened 2023-06-08
[Unity] Add an API to create multiple kv caches with single allocation

yelite:
This would be useful when creating multiple kv caches with the same shape. On A10G, compared to creating 64 kv caches separately in LLaMA from mlc-llm, doing a single allocation can save about 35 ms.

@tqchen @junrushao

### DIFF
--- a/src/runtime/relax_vm/lm_support.cc
+++ b/src/runtime/relax_vm/lm_support.cc
@@ -167,11 +167,58 @@ class AttentionKVCache : public ObjectRef {
 
 TVM_REGISTER_OBJECT_TYPE(AttentionKVCacheObj);
 
+/*!
+ * \brief Create multiple kv caches with same shape, from single memory allocation.
+ * \param init_data The initial data to put into the cache. Ignored if init_fill_count is
+ *        less than 0.
+ * \param reserve_shape The shape of cache.
+ * \param init_fill_count The initial row to fill into
+ *        the cache.
+ * \param num_caches Number of caches to create.
+ */
+Array<AttentionKVCache> CreateMultipleKVCaches(NDArray init_data, ShapeTuple reserve_shape,
+                                               int init_fill_count, int num_caches) {
+  DLDataType dtype = init_data->dtype;
+
+  int64_t cache_size = (dtype.bits * dtype.lanes + 7) / 8;
+  for (const auto dim : reserve_shape) {
+    cache_size *= dim;
+  }
+
+  // Add padding to make each cache align to kAllocAlignment
+  using tvm::runtime::kAllocAlignment;
+  int64_t padding = (kAllocAlignment - cache_size % kAllocAlignment) % kAllocAlignment;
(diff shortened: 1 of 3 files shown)


src/runtime/relax_vm/lm_support.cc
```diff
@@ -167,12 +167,59 @@ class AttentionKVCache : public ObjectRef {
...
+  int64_t padding = (kAllocAlignment - cache_size % kAllocAlignment) % kAllocAlignment;
+  int64_t cache_offset = cache_size + padding;
+
+  auto block = NDArray::Empty(ShapeTuple({cache_offset * num_caches}), dtype, init_data->device);
+  auto block_view = block.CreateView(reserve_shape, dtype);
+
+  Array<AttentionKVCache> result;
+  for (int i = 0; i < num_caches; ++i) {
+    // Use DLManagedTensor to prevent underlying memory from being freed
+    DLManagedTensor* data_view = block_view.ToDLPack();
```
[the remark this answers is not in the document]
tqchen: https://github.com/apache/tvm/blob/unity/include/tvm/runtime/relax_vm/memory_manager.h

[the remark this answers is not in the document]
yelite: Thanks! I updated the code to use storage interface and it looks cleaner. But now it could print a warning message if the requested allocator type mismatches from the allocator that is created at VM initialization.

tqchen:
cc @yzh119 @Hzfengsy

src/runtime/relax_vm/lm_support.cc
```diff
@@ -167,12 +168,56 @@ class AttentionKVCache : public ObjectRef {
...
+ * \param reserve_shape The shape of cache.
+ * \param init_fill_count The initial row to fill into
+ *        the cache.
+ * \param num_caches Number of caches to create.
+ */
+Array<AttentionKVCache> CreateMultipleKVCaches(NDArray init_data, ShapeTuple reserve_shape,
+                                               int init_fill_count, int num_caches) {
+  DLDataType dtype = init_data->dtype;
+
+  int64_t cache_size = (dtype.bits * dtype.lanes + 7) / 8;
```
yzh119, line 184: So currently the dtype is smaller than one byte, then we would pad it to one byte, is that correct?
FYI: [Flexgen](https://arxiv.org/abs/2303.06865) uses 4-bit KV cache, we can support it later.

yzh119 (approved):
LGTM and thanks for doing this, I just have a few minor comments.

src/runtime/relax_vm/lm_support.cc
tqchen: I think it is fine for now. Since subbyte are usually packed manually(the dtype is i32)

yzh119: Thanks for the clarification, make sense to me.

PR #15068, opened 2023-06-09
[Relay] Improve the "clip" op optimization in simplify expr pass

kfeng123:
This PR improves the optimization of clip in simplify expr pass.
Previously, there is a SimplifyCastClip class, it deals with the pattern "cast->clip": it checks if the a_min/a_max values of clip equals the min/max values of the data type of cast OP.
This PR improve SimplifyCastClip in 2 aspects:

1. In the original code, it is restricted that the OP before "clip" is a "cast" OP. This restriction is not necessary, and is relaxed.
2. In the original code,  CheckDataTypeMaxMinValue checks if the a_min/a_max values of clip equals the min/max values of dtype. I relax this to check "a_max >= dtype.max and a_min <= dtype.min". Also, I add the case of bfloat16 in CheckDataTypeMaxMinValue.

Since the new optimization of "clip" does not rely on "cast" op, the name is changed to SimplifyClip.


kfeng123:
cc @ibsidorenko

src/relay/transforms/simplify_expr.cc
```diff
@@ -234,30 +232,38 @@ class SimplifyClipAndConsecutiveCast : public DFPatternRewrite {
...
-    DataType cast_dtype = Downcast<TensorType>(cast->checked_type())->dtype;
+    DataType cast_dtype = Downcast<TensorType>(pre->checked_type())->dtype;
 
-    auto clip = Downcast<Call>(post);
-    const CallNode* clip_node = clip.as<CallNode>();
+    const CallNode* clip_node = post.as<CallNode>();
     const ClipAttrs* clip_attrs = clip_node->attrs.as<ClipAttrs>();
 
+    // TODO(kfeng123): For now, the arg of "clip" is forced to not be "qnn.requantize" and
+    // "ann.add". This is to avoid destroying the structure required by LegalizeQnnOpForDnnl
```
ibsidorenko, line 250: "qnn.add"?

kfeng123: Thanks. Typo fixed.

ibsidorenko (approved):
By the way, what is motivation for this change?  New case  came from the real life DNN model... or this is required for some kind of synthetic test?

masahi (approved):

PR #15074, opened 2023-06-12
[Relax] Add bitwise and logical ops (AND, NOT, OR, XOR)

vvchernov:
Support bitwise and logical unary (NOT) and binary (AND, OR, XOR) ops on Relax side.

cc @tqchen


vvchernov:
Mistake with branch to pull (main instead of unity). See PR #15075

Issue #15019, opened 2023-06-04
[BUG] [Relay] PlanDevices Pass Failure when Two Operators with Different Target Devices Share the Same Input


lecoan:
While trying to utilize Apache TVM for heterogeneous execution, I've encountered a problem with the PlanDevices pass. Specifically, when two operators share the same input but are assigned to different target devices, PlanDevices pass fails.

As an example, consider the operation `(a+b)-(b+c)`. If I intend to assign the first `add` operator to the CPU and the second one to the GPU, PlanDevices pass fails as it appears to have difficulty determining the appropriate device for `b`.

Given that it's common for multiple layers in a neural network to process the same input, this seems to be a bug in TVM that warrants attention.

### Expected behavior

There are two behaviors I would expect from the PlanDevices pass in this scenario:

1. Automatic addition of a `device_copy`: If CPU is the default device, the PlanDevices pass should append a `device_copy` between `b` and the last `add` operator.
2. Input replication: The PlanDevices pass could replicate `b` on both CPU and GPU.

### Actual behavior

Assuming the PlanDevices pass initially visits `(a+b)`, it marks `b` for the CPU. However, when it visits `(b+c)`, it throws an error as it attempts to place `b` on GPU.

Here is the error message:

```log
TVMError: Function parameters and result VirtualDevices do not match those of call. Call:
free_var %b: Tensor[(5, 7), float32] ;
free_var %c: Tensor[(5, 7), float32] ;
%0 = add(%b, %c) ;
on_device(%0, virtual_device=VirtualDevice(device_type=2, virtual_device_id=0, target=Target(id=12cebdda0, kind='cuda', keys={'cuda', 'gpu'}, attrs={'max_num_threads': 1024, 'thread_warp_size': 32, 'arch': "sm_50"}, host=Target(id=12ceba810, kind='llvm', keys={'cpu'})))) 
with function virtual devices:
fn(?4828570296?VirtualDevice(device_type=2, virtual_device_id=0, target=Target(id=11fcaac00, kind='cuda', keys={'cuda', 'gpu'}, attrs={'max_num_threads': 1024, 'thread_warp_size': 32, 'arch': "sm_50"}, host=Target(id=11fcc96b0, kind='llvm', keys={'cpu'})))):?4828570408?
and implied call virtual devices:
[...]

PR #15090, opened 2023-06-13
[Bugfix] [Relay] Insertion of "device_copy" CallNode to Resolve Device Conflict on Unconstrained Nodes

lecoan:
This PR addresses an issue #15019 I opened previously, regarding the PlanDevices pass's failure in cases where two operators share the same input but are intended to be assigned to different target devices. This scenario can often occur in the context of a neural network, where multiple layers can process the same input.

In the specific case of an operation `(a+b)*(b+c)`, where the first add operator is assigned to the CPU and the second one to the GPU, PlanDevices pass would fail as it had difficulty determining the correct device for b.

The problematic behavior seemed to be due to PlanDevices pass marking `b` for the CPU (when it first visits `a+b`), and then throwing an error when it attempts to place b on the GPU while visiting `b+c`.

The solution I've implemented in this PR is the automatic addition of a `device_copy`. This means that if `b` is assigned to CPU after visiting `a+b` , the PlanDevices pass will append a `device_copy` to copy `b` to GPU when visiting `b+c`.

### DIFF
--- a/src/relay/transforms/device_planner.cc
+++ b/src/relay/transforms/device_planner.cc
@@ -77,8 +77,8 @@
  *
  * Phase 1
  * -------
- * We iterately process the programs and find nodes with conflicting virtual devices. If the
- * virtual devices (d1 and d2) are joinable, they are replaced with a joined device (d). If
+ * We iteratively process the programs and find nodes with conflicting virtual devices. If the
+ * virtual devices ( \p d1 and \p d2 ) are joinable, they are replaced with a joined device \p d. If
  * they are unjoinable, a "device_copy" CallNode is inserted to copy the node output to the second
  * device.
  *
@@ -412,7 +412,7 @@ class RewriteOnDevices : public ExprMutator {
 /* =============== Phase 1 =============== */
 
 /*!
- * \brief Add "device_copy" calls for these nodes that have conflicted virtual_devices.
+ * \brief Add "device_copy" calls for nodes that have conflicting virtual devices.
  *
  * Eg Suppose an IRModule contains the following expr:
  * \code
@@ -421,7 +421,7 @@ class RewriteOnDevices : public ExprMutator {
  *   %2 = add(%b, %c);
  *   %3 = on_device(%2, virtual_device=d2);
  * \endcode
- * Then \p %b has two possbile virtual_devices \p d1 and \p d2 from the above expr:
+ * In the above example, node %b has two possible virtual devices: \p d1 and \p d2.
  *
  * - If \p d1 and \p d2 are joinable, replace \p d1 and \p d2 with the joined device \p d:
  * \code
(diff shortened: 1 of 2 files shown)


lecoan:
@mbs-octoml @Lunderberg  Hi, all the test cases passed. Can you help me review the patch?

src/relay/transforms/device_planner.cc
```diff
@@ -77,6 +77,12 @@
  *
  * Phase 1
  * -------
+ *  We iterate process the programs to find those nodes with conflicting virtual devices. If the
```
masahi, line 80: iteratively

src/relay/transforms/device_planner.cc
```diff
@@ -404,6 +410,201 @@ class RewriteOnDevices : public ExprMutator {
...
+    auto itr = conflicted_nodes.find(expr);
+    return itr != conflicted_nodes.end();
+  }
+
+  std::unordered_set<const ExprNode*> conflicted_nodes;
+  std::unordered_map<const ExprNode*, VirtualDevice> expr_to_device;
+};
+
+/*!
+ * \brief Flows the device constraints over the module and find all the conflicted nodes. The
```
masahi, line 478: Flow

src/relay/transforms/device_planner.cc
```diff
@@ -404,6 +410,201 @@ class RewriteOnDevices : public ExprMutator {
...
+        dev_ctx_->conflicted_nodes.emplace(arg.get());
+      }
+    }
+    for (auto& expr : call_node->args) {
+      VisitExpr(expr);
+    }
+  }
+
+  IRModule mod_;
+  std::unique_ptr<DeviceContext> dev_ctx_;
```
masahi, line 551: Why need pointer here?

src/relay/transforms/device_planner.cc
lecoan: Thanks for your review! My apologies for the spelling errors. I have double-checked the comment using ChatGPT to ensure accuracy.

src/relay/transforms/device_planner.cc
lecoan: My apologies for the spelling errors. I have double-checked the comment using ChatGPT to ensure accuracy.

src/relay/transforms/device_planner.cc
lecoan: The reason is similar to the `PlanDevicesCore` sub-pass, which uses a pointer for `DeviceDomains` to prevent unnecessary copying. Since the necessary information is contained in `dev_ctx_`, which is created in `ConflictedNodeFinder` and then passed to `ConflictedNodeRewriter`, we also use a pointer here.

src/relay/transforms/device_planner.cc
```diff
@@ -77,6 +77,13 @@
  *
  * Phase 1
  * -------
+ * We iterately process the programs and find nodes with conflicting virtual devices. If the
```
masahi, line 80: Still wrong, "iteratively"

lecoan: fixed

masahi (approved):

PR #15091, opened 2023-06-13
[TIR][TVMScript] Convert tir.op operands to PrimExpr

Lunderberg:
Previously, some operations in `tvm.tir.op` required the operand to already be a `PrimExpr`, and raised an error if the operand was instead a python object.  This could cause failures round-trip through TVMScript, as int32 and float32 types are printed as literals in TVMScript, relying on automatic conversion to convert them back into `PrimExpr` objects.  For example, `T.expr2(T.int32(0))` is printed as `T.expr2(0)`, which then passes the python object `0` into `tvm.tir.op.expr2` when parsing the TVMScript.

This commit updates the operations in `tvm.tir.op` to convert operands prior to accessing `operand.dtype`, avoiding the error and allowing these expressions to round-trip.

This issue was first noticed in https://github.com/apache/tvm/pull/15076 for `T.ret(0)`, but applied to any operator that used the operand's dtype without first normalizing to `PrimExpr`.

### DIFF
--- a/python/tvm/tir/op.py
+++ b/python/tvm/tir/op.py
@@ -1460,6 +1460,8 @@ def ret(val):
     ret : PrimExpr
         The return expression
     """
+
+    val = convert(val)
     return call_intrin(val.dtype, "tir.ret", val)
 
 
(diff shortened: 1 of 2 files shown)


Lunderberg:
@tvm-bot rerun

csullivan (approved):

PR #15102, opened 2023-06-14
[TIR] Update primfunc host attachment to include host

tqchen:
This PR updates the host function attachment to include host attribute so it can be lowered through MakePackedAPI.


tqchen:
cc @Lunderberg this is a temp workaround after #14986 , we might want to revisit the invariance of MakePackedAPI require a host attribute (it makes sense in some way but might also get misuse)

yzh119 (approved):

Lunderberg:
Thank you @tqchen, and this change makes sense, and would prevent any later use of `BindTarget` from adding an erroneous host.  I assume that this is needed for compatibility in the unity branch?  (Trying to think of functionality tests that could be added on the main branch to validate this behavior.)

Regarding `MakePackedAPI`, I think the main issue is that `MakePackedAPI` needs to know both the host, so that the resulting device function can have the same host as is used by other functions in the module, and needs the device so that the correct device type can be checked in `ArgBinder::BindDLTensor`.  I've been thinking about whether `tir::Buffer` could have an optional parameter for the device type, which could be populated during `SplitHostDevice` to provide the device type for `MakePackedAPI` to use.

Lunderberg (approved):

PR #15104, opened 2023-06-15
[microNPU][ETHOSU] Fix minimum buffer size

Aleksei-grovety:
Fix minimum buffer size for DMA operations according to alignment.


ekalda (commented):
Thanks @Aleksei-grovety! I think a test would be good there as well.

Aleksei-grovety:
I added a test for this case.

ekalda (approved):
LGTM, thanks @Aleksei-grovety!

Issue #14899, opened 2023-05-20
[Tracking Issue][UX][Unity] Private annotations in TVMScipt


tqchen:
Global symbol right now needs to be manually attached, for better UX, it is desirable to move to the following model

```python
class IRModule:
    # implies a private function
    @R.function(private=True)
    def func1():
        pass

    # implies a function with global_symbol=func2, and we do not need to fill in the attr field.
    @R.function
    def func2():
         pass
```

cc @quic-sanirudh

tqchen:
cc @yongwww @Hzfengsy @cyx-6

tqchen:
i think below would work

```python
def dec(f=None, label=None):
     def wrap_function(f):
        def wrapper(*args, **kwargs):
            if f is None:
           
    if f is not None:
         return wrap_function(f)
    else:
         return wrap_function
```

slyubomirsky:
Wow, that is a very clever solution and it indeed works. I will implement this solution for purity annotations and we can also use it for privacy too.

slyubomirsky:
Implemented

PR #15109, opened 2023-06-15
[Unity][Relax][UX] Specify function purity in the @R.function decorator

slyubomirsky:
This PR implements the suggestion proposed in #14899 and allows function purity to be annotated in the `@R.function` decorator instead of using a fake method call (`R.is_pure()` or `R.is_impure()`). This required various changes to the parser to support, but can be used as a template for adding further optional parameters to `@R.function`.

For example, this is how an impure function can be declared:
```python
@R.function(pure=False)
def f(...): ...
```

The default value of `pure` is `True`, so there is no need to include an annotation for a function that is already pure (which will usually be the case in Relax).

Note that it will also still be valid to use the decorator without any arguments, as here:
```python
@R.function
def f(...): ...
# f will be treated as pure
```

See `entry.py` for how the `function` decorator has been modified to support this.

Many thanks to @junrushao for guiding me through some of the moving parts in the parser. Further thanks is due to @tqchen, as he showed how the convenient decorator interface could be implemented in the discussion for #14899.

### DIFF
--- a/include/tvm/script/ir_builder/relax/ir.h
+++ b/include/tvm/script/ir_builder/relax/ir.h
@@ -33,9 +33,10 @@ namespace relax {
 
 /*!
  * \brief Start a function frame.
+ * \param is_pure Whether the function is annotated as pure.
  * \return The created ir_builder Function frame.
  */
-TVM_DLL FunctionFrame Function();
+TVM_DLL FunctionFrame Function(const Bool& is_pure);
 
 /*!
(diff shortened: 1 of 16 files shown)


tqchen:
cc @yongwww @cyx-6

Hzfengsy (approved):

tqchen:
Great work! would be great to also followup with private decorator

PR #15112, opened 2023-06-15
[Unity][Relax] Make RewriteDataflowReshape only rewrite volume-preserving ops

kparzysz-quic:
The reshape operator expects that the number of elements in the source is the same as the number of elements in the result. There are operators that could have a reshape pattern that don't meet this requirement (e.g. strided_slice), and they should not be converted to reshape.

### DIFF
--- a/src/relax/transform/rewrite_dataflow_reshape.cc
+++ b/src/relax/transform/rewrite_dataflow_reshape.cc
@@ -27,7 +27,6 @@
 #include <tvm/tir/analysis.h>
 #include <tvm/tir/function.h>
 
-#include <numeric>
 #include <vector>
 
 #include "../op/tensor/manipulate.h"
@@ -73,10 +72,8 @@ class DataflowReshapeRewriter : public ExprMutator {
   }
 
   Expr VisitExpr_(const CallNode* call) final {
-    auto Unchanged = GetRef<Call>(call);
-
-    if (!IsCallingTIRReshape(call)) {
-      return Unchanged;
+    if (call->args.size() < 2) {
+      return GetRef<Call>(call);
     }
 
     // We bring the calls of reshape PrimFunc back to calls of high-level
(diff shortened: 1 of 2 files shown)


src/relax/transform/rewrite_dataflow_reshape.cc
```diff
@@ -89,8 +95,32 @@ class DataflowReshapeRewriter : public ExprMutator {
...
+    auto product = [](Array<PrimExpr> args) -> PrimExpr {
+      ICHECK(!args.empty());
+      return std::reduce(args.begin(), args.end(), PrimExpr(1),
+                         [](auto a, auto b) { return a * b; });
+    };
+    auto arg_count = product(arg_sinfo->GetShape().value());
+    auto res_count = product(res_sinfo->GetShape().value());
+    if (!arith::Analyzer().CanProveEqual(arg_count, res_count)) {
+      return Unchanged;
+    }
```
masahi, line 122: How about moving this logic to `IsCallingTIRReshape`?

src/relax/transform/rewrite_dataflow_reshape.cc
```diff
@@ -69,8 +73,10 @@ class DataflowReshapeRewriter : public ExprMutator {
   }
 
   Expr VisitExpr_(const CallNode* call) final {
+    auto Unchanged = GetRef<Call>(call);
```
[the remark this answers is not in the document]
kparzysz-quic: Ugh, my mistake.  Done.

src/relax/transform/rewrite_dataflow_reshape.cc
kparzysz-quic: This code needs the input arg, so I'll move the call to `IsCallingTIRReshape` until the arg is known.  I hope that's ok.

masahi (approved):

PR #15117, opened 2023-06-16
[FRONTEND][TFLITE] Add support for TFLite's regular NMS operator

ilyag-grovety:
This PR adds support of regular NMS operator.

Open questions:

1. How to properly test added functionality? 
Other NMS implementations, e.g., fast NMS, use a TF frozen graph from TF official website to convert a model to TFLite and keep NMS operations only. In order to create a similar test, we need to find an archive on TF official website that contains a frozen graph of a model compiled with `--use-regular-nms=True` flag. We haven't found it yet, so any help is appreciated.
3. Regular NMS requires two sort operations:
   - Sorting the scores after selecting scores above `nms_score_threshold`. This PR implements this with a simple bubble sort in order to prove the algorithm's semantics. We tried to replace it with `tvm.contrib.sort.argsort`. It works well when testing the regular NMS with `run_tvm_graph` as fast NMS test does or building and running the regular NMS with llvm target. At the same time, it fails to build (error is provided below) when `target=ethos-u,cmsis-nn,c`. It seems that `__tvm_module_ctx` variable is only being initialized when `cpp` runtime is chosen.
   The error:
`error: '__tvm_module_ctx' undeclared (first use in this function)
203 | if (TVMBackendGetFuncFromEnv(__tvm_module_ctx, "tvm.contrib.sort.argsort", &tvm_contrib_sort_argsort_packed) != 0) {`
   - Sorting the scores of previous and current NMS steps. There are two alternatives here:
      - implement some sorting algorithm as part of hybrid script (to replace current bubble sort)
      - save the result of each NMS step and use `argsort` after the hybrid script part. This approach has a drawback as it requires significant amount of memory to store the results of each NMS step.

### DIFF
--- a/python/tvm/relay/op/vision/_vision.py
+++ b/python/tvm/relay/op/vision/_vision.py
@@ -48,6 +48,9 @@ reg.register_pattern("vision.non_max_suppression", OpPattern.OPAQUE)
 reg.register_strategy("vision.all_class_non_max_suppression", strategy.all_class_nms_strategy)
 reg.register_pattern("vision.all_class_non_max_suppression", OpPattern.OPAQUE)
 
+reg.register_strategy("vision.regular_non_max_suppression", strategy.regular_nms_strategy)
+reg.register_pattern("vision.regular_non_max_suppression", OpPattern.OPAQUE)
+
 
 @script
 def _get_valid_counts_shape_func(data_shape):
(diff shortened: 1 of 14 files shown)


ilyag-grovety:
@leandron @Aleksei-grovety

sergio-grovety:
cc @ekalda

ekalda:
Thanks for the patch @ilyag-grovety, looks all very neat and good to me. Hopefully the CI will go green as well once the model URL will be properly enabled.

Regarding to testing, from my research, I didn't find a way to create that operator through current TF or Keras APIs either, so I think using the model consisting of this operator is the best approach here.

I think using simple sorting algorithm is fine for now. I don't know what it would take to enable `argsort` for `cpp` runtime, but any effort of that kind should be a separate patch anyway. If bubble sort becomes a massive performance bottleneck in the future, then it is possible to further enhance this operator with enabling `argsort` or implementing a different sorting algorithm, but for now it should be fine.

Aleksei-grovety:
Hi @mehrdadh! Could you please upload [the model] (https://github.com/Grovety/ModelZoo/raw/52fb82156ae8c8e3f62c7d7caf6867b25261dda4/models/object_detection/ssd_mobilenet_v1/tflite_int8/tflite_graph_with_regular_nms.pb)to S3? It's necessary for testing NMS.

python/tvm/relay/op/vision/_vision.py
```diff
@@ -122,6 +125,33 @@ def all_class_nms_shape_func(attrs, inputs, _):
...
+    out_boxes_shape[1] = int64(attrs.max_detections)
+    out_boxes_shape[2] = int64(4)
+
+    out_classes_shape[0] = boxes_shape[0]
+    out_classes_shape[1] = int64(attrs.max_detections)
+
+    out_scores_shape[0] = boxes_shape[0]
+    out_scores_shape[1] = int64(attrs.max_detections)
+
+    out_num_detections_shape[0] = int64(1)
```
Aleksei-grovety, line 145: Shouldn't there be `out_num_detections_shape[0] = boxes_shape[0]`? As further comments point to `num_detections` of size `(batch_size,)`.

Aleksei-grovety (commented):
Thanks @ilyag-grovety! Looks good just a couple comments.

python/tvm/topi/cuda/ssd/multibox.py
```diff
@@ -341,10 +360,12 @@ def transform_loc(loc, loc_base_idx, anchor, anchor_base_idx, clip, vx, vy, vw,
         i = idxd(tid, num_anchors)
         j = idxm(tid, num_anchors)
 
-        with ib.if_scope(cls_id[tid] > 0):
+        with ib.if_scope(tvm.tir.any(keep_background == 1, cls_id[tid] > 0)):
             with ib.if_scope(j == 0):
                 out_base_idx = i * num_anchors * 6
-                out_loc[out_base_idx] = cls_id[tid] - 1.0
+                out_loc[out_base_idx] = (
+                    cls_id[tid] - 0.0 if keep_background == 1 else cls_id[tid] - 1.0
```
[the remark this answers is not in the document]
ilyag-grovety: No, I'll remove, thanks

python/tvm/relay/op/vision/_vision.py
ilyag-grovety: You're right, fixed

sergio-grovety:
@tvm-bot rerun

PR #15118, opened 2023-06-16
[Unity] Allocate workspace for all functions

yelite:
This PR makes `AllocateWorkspace` allocate workspace in all Relax functions that doesn't have `Codegen` or `Composite` attributes.

cc @masahi


yelite:
@masahi I think what you described is ideal and the workspace param can be made as `R.Object` to workaround the multi-caller limitation. I am just not sure what we lose by stripping out shape information from the workspace param. If all the usages of workspace validates the actual size of workspace, and allocate if workspace is not large enough, like what attention does currently, I think it should be fine.

masahi (approved):
`R.Object` sounds like a good solution. For now this patch should be fine.

PR #15120, opened 2023-06-17
[ARITH] NormalizeToIterSum

tqchen:
This PR adds an arith util function NormalizeToIterSum.

This function normalizes an index to sum of iterators by scale.

The iterator splits are simplified when possible and ordered by their constant scale and number of prod symbols in descenting order.

We also ensures that each scale is in the form of symbol * const (e.g. n * m * 4). This function is useful to detect stride patterns in buffer access.


tqchen:
@tvm-bot  rerun

MasterJH5574 (approved):

PR #15121, opened 2023-06-19
GraphExecutor: Fix wild pointer assign when input and output are reshape

minbo-wang:
If reshape in our relay expression, like reshape + matmul expr, when we use set_input_zero_copy or set_output_zero_copy，the result goes wrong.

### DIFF
--- a/src/runtime/graph_executor/graph_executor.cc
+++ b/src/runtime/graph_executor/graph_executor.cc
@@ -230,6 +230,16 @@ void GraphExecutor::SetOutputZeroCopy(int index, DLTensor* data_ref) {
   // check the consistency of output
   CheckExternalDLTensor(data_ref, output_node_eid);
 
+  if (nodes_[output_node.node_id].op_type == "tvm_op" &&
+      nodes_[output_node.node_id].param.func_name == "__nop") {
+    const NodeEntry& input_node = nodes_[output_node.node_id].inputs[0];
+    output_node_eid = this->entry_id(input_node);
+    ICHECK_NE(node_output_dltensors_[output_node_eid].size(), 0);
+    for (DLTensor* t : node_output_dltensors_[output_node_eid]) {
+      t->data = static_cast<char*>(data_ref->data) + data_ref->byte_offset;
+    }
+  }
+
   // Update the data pointer for output op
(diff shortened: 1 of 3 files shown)


echuraev (changes requested):
Could you please add correspondent unit tests?

minbo-wang:
> Could you please add correspondent unit tests?

minbo-wang:
> Could you please add correspondent unit tests?
ok

src/runtime/graph_executor/graph_executor.cc
```diff
@@ -566,7 +588,7 @@ std::pair<std::function<void()>, std::shared_ptr<GraphExecutor::OpArgs>> GraphEx
   }
 
   if (param.func_name == "__nop") {
-    return {[]() {}, arg_ptr};
+    return {[arg_ptr]() {}, arg_ptr};
```
echuraev, line 591: What is the reason of capturing `arg_ptr` in empty lambda?

tests/python/contrib/test_reshape_0_copy.py
echuraev, line 1: IMHO, I would say that better use `zero` instead of `0` in the name of the file, e.g `test_reshape_zero_copy.py`.
Also, probably you can move this test to the [file](https://github.com/apache/tvm/blob/main/tests/python/unittest/test_runtime_module_based_interface.py#L693) with existing tests on zero_copy.

echuraev (commented):
A couple of minor comments. @vinx13, @yinghai could you please review?

echuraev (approved):
In general LGTM. Just a couple of moments which were described in previous comment.

corehalt:
@echuraev , cc: @masahi 
Hi, can we please merge this? current set_output_zero_copy doesn't work without these changes when we have a (reshape) nop at the end of a graph (it prevents to forward the pointer of the actual output area through the nop) :pray: 

Thank you!

corehalt:
oh also @vinx13 was involved sorry, forgot to mention

masahi:
@corehalt The PR needs to resolve the conflict. Feel free to take over this PR and create a new one.

masahi:
Yes it's fine.

PR #15130, opened 2023-06-20
[Unity][Relax] Check for call_tir early in DataflowReshapeRewriter

kparzysz-quic:
In an earlier commit, the check for `call_tir` was moved until later, which exposed some of the code to any calls, not just `call_tir`. The code was making assumptions about some of the arguments, which now are not necessarily met. Specifically, that there are at least two arguments, and that the first argument was a `GlobalVar`.

Move the check for `call_tir` back to the beginning of the function.


kparzysz-quic:
Sorry, I missed that thing in the last commit.

masahi:
yeah I have https://github.com/apache/tvm/pull/15125 as well

kparzysz-quic:
You have a testcase, so let's go with yours.

PR #15133, opened 2023-06-21
[XGBoost,MetaSchedule] Support xgb set tree method

vincentccc:
support to set specific xgboost tree method

### DIFF
--- a/python/tvm/meta_schedule/cost_model/xgb_model.py
+++ b/python/tvm/meta_schedule/cost_model/xgb_model.py
@@ -344,7 +344,7 @@ class XGBModel(PyCostModel):
         average_peak_n: int = 32,
         adaptive_training: bool = True,
         num_tuning_cores: Optional[int] = None,
-        tree_method: Optional[str] = None,
+        tree_method: Optional[Literal["auto", "exact", "approx", "hist", "gpu_hist"]] = None,
     ):
         super().__init__()
         if not isinstance(extractor, FeatureExtractor):
--- a/python/tvm/meta_schedule/cost_model/xgb_model.py
+++ b/python/tvm/meta_schedule/cost_model/xgb_model.py
@@ -21,6 +21,8 @@ from collections import OrderedDict
 from itertools import chain as itertools_chain
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, NamedTuple, Optional, Tuple
 
+from typing_extensions import Literal
+
 import numpy as np  # type: ignore
 
 from ...contrib.tar import tar, untar
@@ -219,7 +221,7 @@ class XGBConfig(NamedTuple):
     nthread : Optional[int],
         The number of threads to use.
         Default is None, which means to use physical number of cores.
-    tree_method str :
+    tree_method : Literal["auto", "exact", "approx", "hist", "gpu_hist"]
         The tree construction algorithm used in XGBoost.
     """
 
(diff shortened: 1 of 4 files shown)


zxybazh:
@tvm-bot rerun

python/tvm/meta_schedule/tir_integration.py
```diff
@@ -58,6 +58,7 @@ def tune_tir(  # pylint: disable=too-many-locals
     seed: Optional[int] = None,
     module_equality: str = "structural",
     special_space: Optional[Mapping[str, SpaceGenerator.SpaceGeneratorType]] = None,
+    **kwargs,
```
zxybazh, line 61: May I ask if this is specifically added for cost model? If so, IMHO it would be better to avoid confusion and only add one parameter in cost model's constructor function instead.

python/tvm/meta_schedule/tune.py
```diff
@@ -108,7 +109,7 @@ def tune_tasks(
     elif not isinstance(database, Database):
         database = Database.create(database, module_equality=module_equality)
     if not isinstance(cost_model, CostModel):
-        cost_model = CostModel.create(cost_model, num_tuning_cores=num_cores)
+        cost_model = CostModel.create(cost_model, num_tuning_cores=num_cores, **kwargs)
```
zxybazh, line 112: Same as previous comment, IMHO if we only want to add a single parameter to the cost model, it might be more specific to directly create a cost model with `tree_method` parameter or you can modify the default construction parameters.

python/tvm/meta_schedule/cost_model/xgb_model.py
```diff
@@ -217,6 +219,8 @@ class XGBConfig(NamedTuple):
     nthread : Optional[int],
         The number of threads to use.
         Default is None, which means to use physical number of cores.
+    tree_method str :
```
zxybazh, line 222: Typo, might be changed to `tree_method: str`.

zxybazh: Also, can you please specify the options in the doc? For example what methods are supported here? In that case you can use `Literal["auto", "...", "..."]` (replace the dots with other options) as the type annotation.

zxybazh (commented):
Thanks for adding the new tree method option! Generally looks ok to me, would you please fix the typos and some docs as commented?

python/tvm/meta_schedule/tir_integration.py
vincentccc: Hi zhou, thanks for your time to review, there are some other cost models and set_tree_method only work for xgb model，so I think it is not a proper way to specialize the parameter here. And there may be some other parameters need to be set in other cases, use **kwargs may be more flexible.

python/tvm/meta_schedule/cost_model/xgb_model.py
vincentccc: Thanks a lot, I will fix it.

python/tvm/meta_schedule/tir_integration.py
zxybazh: Hi Vincent, thanks for the explanation, it makes sense to me that `set_tree_method` only work for xgb model and we shouldn't specialize it for all the cost models. On the other hand, the `**kwargs` here is used in a `tune_tir` api and it would be a bit hard to figure out if certain argument is given for cost model or database, or even other classes. Another issue is when multiple class use the same argument, it would be hard to handle. IMHO we may try to let the users configure cost model separately and call in the `tune_tir` api as follows:
```python
cost_model = CostModel.create("xgb", num_tuning_cores=num_cores, tree_method = "auto")
tune_tir(mod, ..., cost_model=cost_model)
```
What do you think?

vincentccc: Thanks for your time! I‘m agree with your proposal, I will fix it.

zxybazh (approved):
LGTM, thanks a lot!

PR #8313, opened 2021-06-23
[Metal] Add pass for splitting kernel with huge number of args

echuraev:
The Metal has some limitations on the number of input parameters. More
information can be found here:
https://developer.apple.com/documentation/metal/buffers/about_argument_buffers?language=objc

In this commit a new pass for splitting functions with big number of
arguments to smaller parts was added. In parameter `max_function_args`
we can specify the maximum number of kernel arguments for specific
target and then split kernel when the number of arguments exceeds the
value of `max_function_args`. Currently this pass works only for concat
layer.

Thanks for contributing to TVM!   Please refer to guideline https://tvm.apache.org/docs/contribute/ for useful information and tips. After the pull request is submitted, please request code reviews from [Reviewers](https://github.com/apache/incubator-tvm/blob/master/CONTRIBUTORS.md#reviewers) by @ them in the pull request thread.

PR #15137, opened 2023-06-21
[Relay] Introduce arguments limit to FuseOps pass

echuraev:
In PR #8313 a parameter `max_function_args` was introduced. It leads to
limit the number of function arguments and in the case when this value is
exceeded, then the concatenation layer is split into a several concat
operations.

I faced a problem on Adreno GPU that for kernel with a big number of
arguments the enqueueNDRange was crashed without any errors. The
problem appeared because of the huge number of arguments. But in this
case not only concat layer was a root cause of the problem. Also after
fusing several operations the final functions had a big number of
arguments.

As it was discussed in #8313, adding a limitation on the number of
function arguments to the FuseOps pass might be a good improvement. In
this PR I introduced such a mechanism for limiting the number of function
arguments for FuseOps pass and add an arguments limit to OpenCL devices
at 128 parameters.

The idea of the current approach is to calculate the number of arguments for
each node in the fusing algorithm and in case then the number of function
arguments exceeds the limit, specified by `max_function_args`, then the
fusing should be stopped. In case when a node has several inputs and for
some of the inputs the number of arguments wasn't computed, then we
postpone fusing for this node and will try to fuse this node later when
the number of arguments will be computed for all inputs. This approach
with postponed fusing helps to avoid additional computations during
compilation.

Additionally, the case of dynamic shapes should be handled.  In this case, 
function arguments also included sizes of dynamic dimension and strides. 
The number of strides can be computed by calculating the number of tensor 
dimensions (the number of strides equals the rank of the tensor). The number 
of additional parameters with sizes of dynamic dimensions can be calculated 
by computing the number of dynamic dimensions.

cc: @Hzfengsy, @masahi, @csullivan


echuraev:
@tqchen During implementing a limitation on the number of function arguments, I met a problem. In case of tensor with dynamic shape, it is possible that in addition to the parameters of the kernel and indices by dynamic dimensions, strides will be passed to the function. For example, below an example of the OpenCL code which was generated for one of concats from the test on split_args:
```opencl
__kernel void vm_mod_fused_concatenate_kernel(__global float* restrict T_concat, 
                                              __global float* restrict p0,
                                              __global float* restrict p1,
                                              __global float* restrict p2,
                                              __global float* restrict p3,
                                              __global float* restrict p4,
                                              int any_dim,
                                              int any_dim_1,
                                              int any_dim_2,
                                              int any_dim_3,
                                              int any_dim_4,
                                              int stride,
                                              int stride_1,
                                              int stride_2,
                                              int stride_3,
                                              int stride_4,
                                              int stride_5,
                                              int stride_6,
                                              int stride_7,
                                              int stride_8,
                                              int stride_9,
                                              int stride_10,
                                              int stride_11,
                                              int stride_12,
                                              int stride_13,
                                              int stride_14,
                                              int stride_15,
                                              int stride_16) {
   // ...
}
```
As you can see, OpenCL kernel has three types of arguments: buffers, indices of dynamic dimensions and strides. Unfortunately, I'm not able to get the number of strides in `SplitArgs` and `FuseOps` passes. Strides appear only after the stage of `LowerTE`. 

To handle this problem, I added a new attribute to target `max_function_args_dyn` which is used in case if at least one of the tensors has dynamic shape. This attribute should have less value than `max_function_args` and it helps to run topologies, which previously didn't work in TVM. But I have some doubts about this solution. Probably you know the better way, how can I calculate the number of strides which will be passed to the kernel in case of dynamic shapes?

echuraev:
@csullivan, @masahi, @Hzfengsy now all functional and performance issues are fixed. Could you please review this PR?

Hzfengsy:
@masahi Could you please review this one?

include/tvm/relay/transform.h
```diff
@@ -120,9 +120,12 @@ TVM_DLL Pass FoldConstant(bool fold_qnn = false);
 /*!
  * \brief Split function with huge number of arguments to smaller pieces.
  *
+ * \param max_function_args Maximum number of function arguments. If it is 0 then SplitArgs won't
+ *                          split funciton.
```
masahi, line 124: typo

src/relay/analysis/graph_partitioner.h
```diff
@@ -205,10 +209,14 @@ class GraphPartitioner {
...
   /*! \brief The maximum number of operations in one fused function */
   size_t max_fuse_depth_;
+  /*! \brief The maximum number of arguments in one fused function */
+  size_t max_function_args_;
   /*! \brief The internal groups. */
   std::vector<Group*> groups_;
   /*! \brief internal field used for deduplication */
   std::unordered_set<IndexedForwardGraph::Node*> visited_;
+  /*! \brief internal field used for hashing arguments number for a node */
+  std::unordered_map<const tvm::Object*, size_t> argsMap_;
```
masahi, line 219: args_map_

"arguments number" is a weird phrase

masahi: Also the name `args_map_` doesn't tell what it is for.

src/relay/analysis/graph_partitioner.h
```diff
@@ -247,6 +255,9 @@ class GraphPartitioner {
   void CommitFuse(IndexedForwardGraph::Node* src, IndexedForwardGraph::Node* sink);
 
   size_t CountNodesUptoSink_(IndexedForwardGraph::Node* src, IndexedForwardGraph::Node* sink);
+  size_t CountAdditionalArgs_(const TensorTypeNode* ttype, bool with_strides = true);
```
masahi, line 258: Describe what "Additional" means

src/relay/analysis/graph_partitioner.cc
```diff
@@ -238,6 +320,8 @@ void GraphPartitioner::InitGroups(const IndexedForwardGraph& graph) {
 void GraphPartitioner::RunFuse(const IndexedForwardGraph& graph,    //
                                const DominatorTree& post_dom_tree,  //
                                int phase) {
+  IndexedForwardGraph::Node* prev_node = nullptr;
```
masahi, line 323: Describe what is meant by "previous". It is only used for counting args but the name is too generic.

src/relay/analysis/graph_partitioner.h
```diff
@@ -247,6 +255,9 @@ class GraphPartitioner {
...
+  size_t CountArgs_(const tvm::Object* child, const tvm::Object* till_node);
+  size_t CountArgsLimit_(const IndexedForwardGraph::Node* child);
```
masahi, line 260: Describe what this function is for, even though there is already `max_function_args_`.

echuraev:
Hi @masahi! 

Thank you for your review, and sorry for the delay in the response. I found one accuracy problem in previous implementation and it was necessary to reimplement this algorithm almost from scratch. I have added a few new tests and also checked the accuracy on several networks. Now everything is ok. Also, I have applied some of your comments in the new commit. 

Could you please take a look at this PR once again?

csullivan (commented):
Thanks @echuraev!

echuraev:
@csullivan In the bug #15358 I have described a problem related to the FuseOps pass algorithm. The same problem applicable for the functionality which was introduced in this PR. When we stop fusing because the number of the function arguments was exceeded, then the first PrimFunc will contain the maximum possible number of base functions (if one more base function is added, then the `max_function_args` is exceeded). But other PrimFuncs will contain only one base function in it. 

I suppose it is a generic problem in current fusion algorithm and it should be fixed separately in another PR.

masahi (commented):
There are too many obvious grammar errors in the comments. Please go through them.

src/relay/analysis/graph_partitioner.cc
```diff
@@ -220,7 +225,113 @@ size_t GraphPartitioner::CountFusedNodesWithNewChild(IndexedForwardGraph::Node*
...
+}
+
+size_t GraphPartitioner::CountArgs_(IndexedForwardGraph::Node* src,
+                                    const IndexedForwardGraph& graph, bool update_postpone) {
+  std::unordered_set<Group*> visited_groups;
+  Group* gnode = groups_[src->index];
+  ICHECK(gnode != nullptr);
+  auto sum = gnode->args_num;
+  visited_groups.insert(gnode->FindRoot());
+  auto calcArgs = [this, src, &graph, &visited_groups,
```
masahi, line 246: code style

echuraev:
@masahi  Thank you for your review! I have corrected all mistakes that I have found. Please, let me know if I missed something.

src/relay/analysis/graph_partitioner.h
```diff
@@ -205,10 +213,21 @@ class GraphPartitioner {
...
+  size_t max_function_args_;
   /*! \brief The internal groups. */
   std::vector<Group*> groups_;
   /*! \brief internal field used for deduplication */
   std::unordered_set<IndexedForwardGraph::Node*> visited_;
+  /*! \brief The map with nodes which were postponed for fusing. */
+  std::unordered_multimap<const IndexedForwardGraph::Node*, IndexedForwardGraph::Node*>
+      postponed_fusing_map_;
+  /*!
+   * \brief Fusing of this node should be postponed till all child nodes will be evaluated.
```
masahi, line 226: till all child nodes **are** evaluated.

src/relay/analysis/graph_partitioner.h
```diff
@@ -247,6 +266,23 @@ class GraphPartitioner {
...
+  // Calculate the number of arguments for the node.
+  size_t CountArgs_(IndexedForwardGraph::Node* src, const IndexedForwardGraph& graph,
+                    bool update_postpone = true);
+  // Count the actual limit of arguments for a generated function.
+  // max_function_args_ specifies the number of maximum function arguments. But
+  // usually, output tensors are also passed to the function as arguments.
+  // Additionally, in the case of dynamic shape, it is necessary to take into
+  // account the number of parameters which specifies the sizes of the dynamic
+  // dimensions.
+  // This function computes the limit of arguments by the following formula:
```
masahi, line 283: the limit of arguments -> the maximum number of arguments

masahi (approved):

echuraev:
@masahi thank you for your review and help! Applied all comments.

echuraev:
@tvm-bot rerun

PR #15138, opened 2023-06-21
[Test] Improve check for TVMError exception in test_cast

kparzysz-quic:
The original code has
```
with pytest.raises(tvm.error.TVMError):
    s.astype("int")
    assert "Can't cast a handle to other types" in str(e)
```
The `astype` is expected to throw a TVMError exception, but if it does, the following assertion never gets executed.

Add try/except to the test to verify that the expected exception is thrown.


echuraev (approved):
LGTM!

PR #15139, opened 2023-06-21
[LLVM] Minor refactor to LLVMModuleNode::SaveToFile

Lunderberg:
Previously, the `#if TVM_LLVM_VERSION` checks made it difficult to determine the logic of  LVMModuleNode::SaveToFile while debugging. This commit pulls out the preprocessor directives into wrapper functions that maintain the same compatibility, making it easier to follow the logic of the `SaveToFile` function.


kparzysz-quic (approved):
LGTM, thanks!

PR #15140, opened 2023-06-21
[Unity][IR][UX] Privacy annotation in Relax

slyubomirsky:
This PR implements the proposal in #14899. Namely, the `@R.function` decorator now has an optional `private` attribute. If a function is marked as private, then it will not have a global symbol attached to it and thus will not be externally accessible. By default, functions are not private, so the parser does insert a global symbol for them.

Here is an example of how to use the privacy attribute:
```python
@I.ir_module
class Module:
    # Will not have a global symbol! Often not very useful to make main private...
    @R.function(private=True)
    def main(...): ...

    # private is false by default. This function will have a global symbol ("other")
    @R.function
    def other(...): ...
```

Passes that introduce new functions should decide whether a global symbol should be included with those functions. Most of the time, this is not the case (e.g., the functions extracted by the lambda lifting pass probably should not be publicly accessible).

Note that there are alternative approaches we can consider. For example, we may have "private" as a field on a function node and fill in the global symbol attributes _later_. This may be important for some passes.

### DIFF
--- a/src/relax/transform/lift_transform_params.cc
+++ b/src/relax/transform/lift_transform_params.cc
@@ -251,7 +251,10 @@ class TransformParamsLifter : public ExprMutator {
     lift_plan_ = planner.Plan(func, num_input);
 
     // Step 2: Add the lifted function to the module
-    builder_->AddFunction(lift_plan_.f_transform_params, new_func_name);
+    // (The lifted function should be public so we add a global symbol to it)
+    auto lift_func =
+        WithAttr(lift_plan_.f_transform_params, tvm::attr::kGlobalSymbol, new_func_name);
+    builder_->AddFunction(lift_func, new_func_name);
 
     // Step 3: Update the current function.
 
--- a/src/script/printer/relax/function.cc
+++ b/src/script/printer/relax/function.cc
@@ -110,7 +110,7 @@ TVM_STATIC_IR_FUNCTOR(IRDocsifier, vtable)
       // if the function is global or is not in a module and does not have a global symbol,
       // indicate that it's private
       if (AtTopLevelFunction(d) &&
-          (!n->attrs.defined() || !n->attrs->dict.count("global_symbol"))) {
+          (!n->attrs.defined() || !n->attrs->dict.count(tvm::attr::kGlobalSymbol))) {
         dec_keys.push_back("private");
         dec_values.push_back(LiteralDoc::Boolean(true, Optional<ObjectPath>()));
       }
--- a/src/script/printer/relax/function.cc
+++ b/src/script/printer/relax/function.cc
@@ -43,7 +43,15 @@ TVM_STATIC_IR_FUNCTOR(IRDocsifier, vtable)
     .set_dispatch<relax::Function>("", [](relax::Function n, ObjectPath n_p, IRDocsifier d) -> Doc {
       std::unordered_set<const tir::VarNode*> func_vars;
       With<RelaxFrame> f(d);
-      IdDoc func_name = d->Define(n, f(), FindFunctionName(d, n).value_or("main"));
+
+      IdDoc func_name("");
+      // if we are binding a local definition, then calling d->Define
+      // will result in a repeated definition and an incorrect displayed name
+      if (Optional<String> name = GetBindingName(d)) {
+        func_name = std::move(IdDoc(name.value()));
+      } else {
+        func_name = std::move(d->Define(n, f(), FindFunctionName(d, n).value_or("main")));
+      }
       (*f)->AddDispatchToken(d, "relax");
       (*f)->is_func = true;
       (*f)->func_vars = &func_vars;
--- a/src/script/printer/relax/function.cc
+++ b/src/script/printer/relax/function.cc
@@ -22,6 +22,21 @@ namespace tvm {
 namespace script {
 namespace printer {
 
+bool AtTopLevelFunction(const IRDocsifier& d) {
+  // fewer than 2 frames: not in a function at all
+  if (d->frames.size() < 2) {
+    return false;
+  }
+  // if the first frame is a RelaxFrame, then this is not inside a module.
+  // 2 frames => we are at a function (more than 2 => nested function)
+  if (d->frames[0]->IsInstance<RelaxFrameNode>()) {
+    return d->frames.size() == 2;
+  }
+  // otherwise the first two frames pertain to an IR module,
+  // so 3 frames => we are at a top-level function (more than 3 => nested function)
+  return d->frames.size() == 3;
+}
+
 TVM_REGISTER_NODE_TYPE(RelaxFrameNode);
 
 TVM_STATIC_IR_FUNCTOR(IRDocsifier, vtable)
@@ -54,7 +69,7 @@ TVM_STATIC_IR_FUNCTOR(IRDocsifier, vtable)
       if (n->attrs.defined() && !n->attrs->dict.empty()) {
         // if the function is a global function and has a global symbol,
         // then don't print the global symbol (it will be implicit from not being private)
-        if (d->frames.size() == 3 && n->attrs->dict.count("global_symbol")) {
+        if (AtTopLevelFunction(d) && n->attrs->dict.count("global_symbol")) {
           Map<String, ObjectRef> new_attrs;
           for (auto kv : n->attrs->dict) {
             if (kv.first != "global_symbol") {
@@ -81,8 +96,9 @@ TVM_STATIC_IR_FUNCTOR(IRDocsifier, vtable)
         dec_keys.push_back("pure");
         dec_values.push_back(LiteralDoc::Boolean(false, Optional<ObjectPath>()));
       }
-      // if the function is global and does not have a global symbol, indicate that it's private
-      if (d->frames.size() == 3 &&
+      // if the function is global or is not in a module and does not have a global symbol,
+      // indicate that it's private
+      if (AtTopLevelFunction(d) &&
           (!n->attrs.defined() || !n->attrs->dict.count("global_symbol"))) {
         dec_keys.push_back("private");
         dec_values.push_back(LiteralDoc::Boolean(true, Optional<ObjectPath>()));
--- a/src/script/printer/relax/function.cc
+++ b/src/script/printer/relax/function.cc
@@ -52,17 +52,45 @@ TVM_STATIC_IR_FUNCTOR(IRDocsifier, vtable)
       (*f)->func_vars = nullptr;
       // Step 4. Print attributes
       if (n->attrs.defined() && !n->attrs->dict.empty()) {
-        (*f)->stmts.push_back(
-            ExprStmtDoc(Relax(d, "func_attr")  //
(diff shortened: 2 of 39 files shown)


tests/python/relax/test_transform_merge_composite_functions.py
```diff
@@ -41,7 +41,7 @@ def main(
             R.output(gv)
         return gv
 
-    @R.function
+    @R.function(private=True)
```
slyubomirsky, line 44: @masahi I requested your review because I am not sure why exactly I needed to insert the privacy annotations for these tests. Leaving it off prevents any of the functions from being merged, but I'm not sure why that would happen. Do you know where it might matter that the functions have a `global_symbol` attribute?

src/relax/transform/lift_transform_params.cc
```diff
@@ -251,7 +251,10 @@ class TransformParamsLifter : public ExprMutator {
     lift_plan_ = planner.Plan(func, num_input);
 
     // Step 2: Add the lifted function to the module
-    builder_->AddFunction(lift_plan_.f_transform_params, new_func_name);
+    // (The lifted function should be public so we add a global symbol to it)
```
slyubomirsky, line 254: I am assuming we want them to be public.

tests/python/relax/test_tvmscript_printer_relax.py
```diff
@@ -47,8 +47,10 @@ def main(a: R.Tensor((10, 10))) -> R.Tensor((10, 10)):
 
 
 def test_extern_func():
+    # note: this function will be treated as private unless a global symbol is added
```
slyubomirsky, line 50: This seems annoying but I'm not sure there's anything else we can do for the case of a function outside of an IRModule

tqchen: How about we treat function outside as public always?

tqchen:
cc @yongwww

tests/python/relax/test_tvmscript_printer_relax.py
slyubomirsky: I'll see if I can detect when a function is not inside an IRModule

src/script/printer/relax/function.cc
```diff
@@ -52,17 +52,45 @@ TVM_STATIC_IR_FUNCTOR(IRDocsifier, vtable)
       (*f)->func_vars = nullptr;
       // Step 4. Print attributes
       if (n->attrs.defined() && !n->attrs->dict.empty()) {
-        (*f)->stmts.push_back(
-            ExprStmtDoc(Relax(d, "func_attr")  //
-                            ->Call({d->AsDoc<ExprDoc>(n->attrs, n_p->Attr("attrs"))})));
+        // if the function is a global function and has a global symbol,
+        // then don't print the global symbol (it will be implicit from not being private)
+        if (d->frames.size() == 3 && n->attrs->dict.count("global_symbol")) {
```
slyubomirsky, line 57: Not fond of having this magic number, but it was the only way I could figure out to check that the function is at the top level of a module. Possibly a hack--is there a better way I could check for that?

slyubomirsky: @yongwww might there be a more robust way to check if we're inside a module?

tests/python/relax/test_tvmscript_printer_relax.py
tqchen: likely we don;t need such detection? since @R.function itself implies a public function by default i assume and we would need explicit `private=True` flag

tests/python/relax/test_tvmscript_parser.py
```diff
@@ -254,7 +256,7 @@ def main(x: R.Tensor((10, 20), "float32")) -> R.Tensor((10, 20), dtype="float32"
 
     bb = relax.BlockBuilder()
     x = relax.Var("x", relax.TensorStructInfo([10, 20], "float32"))
-    with bb.function("main", [x]):
+    with bb.function("main", [x], {"global_symbol": "main"}):
```
tqchen, line 257: how about we also have private=False flag here(default to False)

tqchen:
Thanks @slyubomirsky on the default behavior, would be great if we default `private=False` if not specified, and ask for explicit private=True. 

It will involve more handling in printing(like not printing global symbol, and when global symbol do not exist, print as private=True), and modifying the test cases that comes with global_symbol (by removing them). But it would be much cleaner from user pov

tests/python/relax/test_tvmscript_parser.py
slyubomirsky: `private=False` in the parser (which is the default) means that there's a global symbol in the attributes.

slyubomirsky:
Do you mean that private should be another attribute? Or that it should be a field in the function AST node? RIght now, privacy is handled entirely in the parser (if `private` is `False`, which is the default, the parser adds a global symbol).

tests/python/relax/test_tvmscript_parser.py
slyubomirsky: Do you mean the block builder should also have a private flag in the constructor?

slyubomirsky:
The tough part is dealing with nested functions. There is no way to have a public nested function, but the BlockBuilder/printer/etc would have to know whether we're inside another function or not.

tqchen:
sorry i am not talking about changing the AST. I only mean the interface

```
@R.function
def public_func():
    pass
    
    
@R.function(private=True):
def private_func():
    pass
```

And make it happen also for tir

tests/python/relax/test_tvmscript_parser.py
tqchen: Right, this way it is closer to what function provide

slyubomirsky:
Then, yes, that is already how I had implemented it. False is the default value for private.

tests/python/relax/test_tvmscript_printer_relax.py
slyubomirsky: We don't want to put a global symbol on inner functions (I assume), so that's why the current behavior is to make sure that the function is inside an IRModule.

I think I can detect this case in `entry.py`, however. It might be a little hacky, though.

tests/python/relax/test_tvmscript_parser.py
slyubomirsky: I've implemented that, good idea :+1:

slyubomirsky:
Yeah, there are printer tests included. I am making some further tweaks because the implementation right now is rather hacky.

tests/python/relax/test_tvmscript_printer_relax.py
```diff
@@ -73,6 +94,40 @@ def func(a: R.Tensor((10, 10))) -> R.Tensor((10, 10)):
     )
 
 
+def test_nested_function():
```
slyubomirsky, line 97: This actually uncovered a seemingly unrelated bug in the printer, where the nested function would be printed with a suffix after its name (see the changes in `src/script/printer/utils.h`) because the LHS variable in the `VarBinding` was treated as already having been encountered. Also fixed.

src/script/printer/utils.h
```diff
@@ -161,13 +161,25 @@ inline Optional<String> FindFunctionName(const IRDocsifier& d, const BaseFunc& f
...
+  // (this comes up in the case of defining a local function: the local function
+  // is the RHS of a binding. The var name on the LHS will have been looked at first;
+  // without this check, we would print a suffix on the function name even though it
+  // is actually the first definition)
+  for (const auto& bound_name : binding_names) {
+    if (name == bound_name) {
+      return name;
+    }
+  }
+
```
slyubomirsky, line 182: Fix to the bug noted in `test_tvmscript_printer_relax.py`, except I'm not sure if this is the best approach to detecting the case of a function definition. This is a result of the handling of `VarBinding`s in `src/script/printer/relax/binding.cc`

slyubomirsky: I think this could have false positives, like if you use the name of the function currently being bound inside the function's body, so I might have to think of a better way to handle the case of a nested function.

slyubomirsky: Found a better alternative solution

slyubomirsky:
I will implement the same changes for TIR in a separate PR

tests/python/relax/test_transform_merge_composite_functions.py
yelite: With this diff, these tests will pass without making those function private
```diff
diff --git a/src/relax/transform/dead_code_elimination.cc b/src/relax/transform/dead_code_elimination.cc
index fe36eb28ef..596c68ef3e 100644
--- a/src/relax/transform/dead_code_elimination.cc
+++ b/src/relax/transform/dead_code_elimination.cc
@@ -96,7 +96,7 @@ IRModule RemoveUnusedFunctions(IRModule mod_, Array<runtime::String> entry_funcs
   for (auto f : existing_functions) {
     // If a function has an external linkage type, we do not remove it.
     // Otherwise, we check the function and remove it if it is not used anywhere.
-    if (f.second->GetLinkageType() == LinkageType::kInternal && !tracer.check_if_called(f.first)) {
+    if (!tracer.check_if_called(f.first)) {
       mod_->Remove(f.first);
     }
   }
diff --git a/src/relax/transform/merge_composite_functions.cc b/src/relax/transform/merge_composite_functions.cc
index 0bc92ba923..bdae719e31 100644
--- a/src/relax/transform/merge_composite_functions.cc
+++ b/src/relax/transform/merge_composite_functions.cc
@@ -60,6 +60,8 @@
 #include <tvm/tir/function.h>
 
 #include "../../support/arena.h"
+#include "tvm/ir/attrs.h"
+#include "tvm/ir/function.h"
 #include "utils.h"
 
 namespace tvm {
@@ -311,7 +313,8 @@ class CompositeInliner : public ExprMutator {
 
       if (func->GetAttr<String>(attr::kComposite)) {
         if (!inlined_functions_.count(func)) {
-          inlined_functions_.Set(func, CopyWithNewVars(func));
+          inlined_functions_.Set(func,
+                                 WithoutAttr(CopyWithNewVars(func), tvm::attr::kGlobalSymbol));
         }
         return Call(inlined_functions_[func], call->args);
       }
```

At the end of MergeCompositeFunction, a DCE pass will be called to removed the merged functions. However, DCE will not remove public functions, i.e. those with a global symbol. Making those functions public prevents them from being cleaned up by DCE.

I think the logic here is correct. Those fused_xxx functions should be private as they are not meant to be called externally. So we probably don't want to change anything here.

tqchen (approved):

tests/python/relax/test_transform_merge_composite_functions.py
slyubomirsky: Thank you very much for the explanation, @yelite. So, to confirm, you don't think there needs to be a change to the merge composite pass?

yelite: Yes, the merge composite pass shouldn't be changed. We might need to update the fusion pass if it produces fused functions with global symbol.

tqchen: I think by default fusion pass should not produce functions with global symbol. Ideally we should prevent fusing functions that with global symbol --- because user expects to access the original function

PR #15142, opened 2023-06-22
[TIR][Schedule] Enhance `compute-inline` for fusion

junrushao:
This PR enhances `compute-inline` and `reverse-compute-inline` to handle more complicated fusion patterns where the RHS of the equation could be generic expressions rather than strict buffer indexing. This could be used in multiple cases in LLM inference.

### DIFF
--- a/tests/python/unittest/test_meta_schedule_schedule_rule_mlt_tc.py
+++ b/tests/python/unittest/test_meta_schedule_schedule_rule_mlt_tc.py
@@ -26,6 +26,7 @@ from tvm.meta_schedule.testing.space_generation import (
     check_sketches,
     generate_design_space,
     get_rules,
+    print_sketches,
 )
 from tvm.script import tir as T
 from tvm.tir.tensor_intrin.cuda import get_wmma_intrin_group
@@ -211,7 +212,7 @@ def test_matmul_relu(shared_scope):
     actual = generate_design_space(
         kind="cuda",
         mod=mod,
-        target=tvm.target.Target("cuda"),
+        target=tvm.target.Target("cuda --arch=sm_70"),
         types=None,
         sch_rules=[
             multi_level_tiling_tensor_core(
@@ -362,7 +363,7 @@ def test_matmul_relu_with_fallback():
     actual = generate_design_space(
         kind="cuda",
         mod=mod,
-        target=tvm.target.Target("cuda"),
+        target=tvm.target.Target("cuda --arch=sm_70"),
         types=None,
         sch_rules=[
             multi_level_tiling_tensor_core(),
@@ -525,7 +526,7 @@ def test_conv2d(shared_scope):
     actual = generate_design_space(
         kind="cuda",
         mod=mod,
-        target=tvm.target.Target("cuda"),
+        target=tvm.target.Target("cuda --arch=sm_70"),
         types=None,
         sch_rules=[
             multi_level_tiling_tensor_core(
@@ -545,7 +546,7 @@ def test_conv2d(shared_scope):
     actual = generate_design_space(
         kind="cuda",
         mod=mod,
-        target=tvm.target.Target("cuda"),
+        target=tvm.target.Target("cuda --arch=sm_70"),
         types=None,
         sch_rules=[
             multi_level_tiling_tensor_core(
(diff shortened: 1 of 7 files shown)


tests/python/unittest/test_tir_schedule_compute_inline.py
```diff
@@ -1159,5 +1159,180 @@ def test_reverse_compute_inline_producer_predicate_disallowed():
     )
 
 
+def test_compute_inline_softmax():
```
Hzfengsy, line 1162: Could we simplify the test cases for better readability?

junrushao: I made some attempts but it didn't turn out smoothly - looks like it's challenging to further minimize the length of the test

junrushao:
I just realized that `tests/python/unittest/test_tir_schedule_tensorize_ldmatrix_mma.py::test_i8i8i32_m16n16k32` fails on my machine when running the tests. It's likely not tested on CI environment

junrushao:
@tqchen @Hzfengsy This PR should be good to go

tests/python/unittest/test_meta_schedule_schedule_rule_mlt_tc.py
```diff
@@ -211,7 +212,7 @@ def matmul_relu_0(A: T.Buffer((128, 128), "float16"), B: T.Buffer((128, 128), "f
     actual = generate_design_space(
         kind="cuda",
         mod=mod,
-        target=tvm.target.Target("cuda"),
+        target=tvm.target.Target("cuda --arch=sm_50"),
```
Hzfengsy, line 215: It's so confusing that we mark arch `sm_50` for tensor core unit tests. I don't think these changes are related to the `compute_inline` enhancement

tests/python/unittest/test_tir_schedule_compute_inline.py
```diff
@@ -622,8 +621,8 @@ def elementwise_overcomputed_producer_reverse_inlined(
     for i, j in T.grid(128, 128):
         with T.block("B"):
             vi, vj = T.axis.remap("SS", [i, j])
-            T.where(i < 127 and j < 127)
-            C[vi, vj] = A[vi, vj] * 2.0 + 1.0
+            if vi < 127 and vj < 127:
```
Hzfengsy, line 624: What's the reason to change `T.where` to `IF` branches? There are two major concerns:
1. with `If` branch we may lose changes for further schedule opportunities after `compute_inline`
2. It's breaking changes for the existing codebase

Hzfengsy (changes requested):

tests/python/unittest/test_meta_schedule_schedule_rule_mlt_tc.py
junrushao: Let me explain: without an sm version that doesn’t support async pipeline, those unittests will fail due to software pipelining. In fact, all those tests will fail on my machine locally

tests/python/unittest/test_tir_schedule_compute_inline.py
junrushao: To be very clear, T.where limits the iteration domain, meaning it’s going to change the affinity of iteration space, which is not desirable. This PR fixed this behavior and I count it as a bug fix rather than a breaking change.

tqchen: We had some discussions on this topic. T.where is a mechanism to use to handle the iterator predicate outside the block. We had a restriction that the affine map indicated by T.where plus the axis binding must match the loop domain declared.

In the specific case of reverse compute inline with over relaxing part, we are not exactly changing the loop domain, but change the internal operations, since we are iterating over the loops and would like to skip some operations. So having if here is better and matches the design of block isolation

tests/python/unittest/test_meta_schedule_schedule_rule_mlt_tc.py
Hzfengsy: In this case, I recommend changing to `sm_70` or `sm_75`

Hzfengsy (approved):
Please address the `sm` version comment, and we can get it in

tests/python/unittest/test_meta_schedule_schedule_rule_mlt_tc.py
junrushao: Will do

PR #15146, opened 2023-06-23
[TIR] Block dependence analysis without schedules

quic-sanirudh:
This work introduces a new object called `BlockDependenceInfo` that computes and returns block dependences. The idea is to be able to expose block level dependences to TIR passes without having to create an explicit schedules.

The patch introduces 2 main classes:
1. `SRefTreeCreator` - This creates and returns a new SRefTree and returns a map from original statements to corresponding srefs
2. `BlockDependenceInfo` - This object computes the actual dependences between blocks within a block scope and returns it for access in TIR passes

This is a continuation to
[PR #15034](https://github.com/apache/tvm/pull/15034) and completes the work started there to expose block level dependences to TIR passes

Note: One major difference between the SRef Tree created for dependence analysis here versus the one already present in schedules is that this SRef tree only contains block nodes and not loops. This makes it easier to find the parent blocks (by just accessing `parent` member)

### DIFF
--- a/include/tvm/tir/block_dependence_info.h
+++ b/include/tvm/tir/block_dependence_info.h
@@ -1,0 +1,102 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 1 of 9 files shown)


quic-sanirudh:
cc @Hzfengsy @junrushao

Hzfengsy (approved):
LGTM. Thanks @quic-sanirudh for the great effort on the refactoring

quic-sanirudh:
> LGTM. Thanks @quic-sanirudh for the great effort on the refactoring

Thanks for all your help in design and reviews

PR #15147, opened 2023-06-23
[microNPU][ETHOSU] Add option to disable copying constants for case without cascader

Aleksei-grovety:
Added the option to disable copying constants for the case when the user determines in the linker script for section ".rodata.tvm" that the constants are located in SRAM.
To activate this option, pass the `--target-ethos-u-disable_copying_constants=1` argument to tvmc.

### DIFF
--- a/src/relay/backend/contrib/ethosu/compiler_attrs.cc
+++ b/src/relay/backend/contrib/ethosu/compiler_attrs.cc
@@ -41,6 +41,7 @@ struct EthosUCompilerConfigNode : public tvm::AttrsNode<EthosUCompilerConfigNode
   String accelerator_config;
   bool enable_cascader;
   bool enable_striping;
+  bool disable_copying_constants;
   String dev_force_block_config;
   String dev_max_open_plans;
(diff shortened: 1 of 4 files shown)


src/relay/backend/contrib/ethosu/compiler_attrs.cc
```diff
@@ -59,6 +60,9 @@ struct EthosUCompilerConfigNode : public tvm::AttrsNode<EthosUCompilerConfigNode
     TVM_ATTR_FIELD(enable_cascader)
         .describe("Whether the cascader should be enabled")
         .set_default(false);
+    TVM_ATTR_FIELD(disable_copying_constants)
+        .describe("Whether copying constants is disabled for case without cascader")
```
lhutton1, line 64: Perhaps it's worth mentioning here that constants are expected to be in SRAM when this option is enabled?

Aleksei-grovety: Yes, sure.

lhutton1 (approved):

PR #15149, opened 2023-06-23
[Docker] Fix build.sh environment variables

neildhickey:
The build.sh script was not actually passing the environment variable to the docker image.


leandron (approved):
LGTM, thanks @neildhickey

PR #15156, opened 2023-06-26
[Unity][Dlight] Add reduction rules

junrushao:
This PR adds a dlight rule that effectively allows fusing all reduction and spatial blocks together and executed within a single GPU kernel. This is meant to schedule softmax, layer norm, RMS norm, etc.


tqchen:
@tvm-bot rerun

tqchen:
Flaky marked in https://github.com/apache/tvm/issues/15158, maybe we can just fix it along the way

junrushao:
Retriggered

tqchen:
Let us fix the flaky if possible, seems a matter of increase atol and rtol there

PR #15162, opened 2023-06-27
[Runtime] Support void as dtype in FFI

junrushao:
In TVM C++, the dtype `void` is designated to represent "non-existing" data types, but this was not exposed to Python at the moment. This PR brings up such support on Python side for consistency.


echuraev (approved):
LGTM. Thanks!

PR #15163, opened 2023-06-27
[TIR] Fix RenewDef for symbolic input shapes

junrushao:
There are cases where the shapes of input buffers are symbolic, but the first symbol is a composite PrimExpr rather than a TIR Var, which the original implementation does not take this into account.

Example:

```python
@T.prim_func
    m = T.int64()
    A = T.match_buffer(a, (m * 2,))  // `m` first appears as composite
    B = T.match_buffer(b, (m, 2))
```

### DIFF
--- a/tests/python/unittest/test_tir_renew_defs.py
+++ b/tests/python/unittest/test_tir_renew_defs.py
@@ -15,9 +15,9 @@
 # specific language governing permissions and limitations
 # under the License.
 
-import pytest
 import sys
 
+import pytest
 import tvm
 import tvm.testing
 from tvm.script import tir as T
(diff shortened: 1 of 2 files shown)


tests/python/unittest/test_tir_renew_defs.py
```diff
@@ -169,5 +170,21 @@ def symbolic_func(a: T.handle, b: T.handle, n: T.int32):
...
+        A = T.match_buffer(a, (m * 2,))
+        B = T.match_buffer(b, (m, 2))
+        for i, j in T.grid(m, 2):
+            with T.block("B"):
+                vi, vj = T.axis.remap("SS", [i, j])
+                B[vi, vj] = A[vi * 2 + vj]
+
+    f1 = main
+    f2 = tvm.tir.stmt_functor.renew_defs(main)
+    tvm.ir.assert_structural_equal(f1, f2)
```
Hzfengsy, line 186: Please check the `m` in `f1` and `f2` are not the same object

Hzfengsy (approved):

tests/python/unittest/test_tir_renew_defs.py
junrushao: sg

junrushao: added

PR #15164, opened 2023-06-27
[CMake] Support LLVM-16 static linking

junrushao:
LLVM static linking in TVM relies on default flags provided by llvm-config, which however does not depend on modern CMake features, namely, imported targets, leading to a series of issues on builds particularly on macOS and Windows. This PR strengthens CMake configuration process to avoid such behavior.


yzh119 (approved):

WPLF:
thanks

PR #15166, opened 2023-06-27
[Target] Add MetaSchedule-compatible attributes to OpenCL

junrushao:
MetaSchedule prefers `max_threads_per_block` and
`max_shared_memory_per_block` for GPU targets.

### DIFF
--- a/src/target/target_kind.cc
+++ b/src/target/target_kind.cc
@@ -362,6 +362,8 @@ TVM_REGISTER_TARGET_KIND("rocm", kDLROCM)
     .set_target_parser(UpdateROCmAttrs);
 
 TVM_REGISTER_TARGET_KIND("opencl", kDLOpenCL)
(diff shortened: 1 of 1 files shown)


src/target/target_kind.cc
```diff
@@ -362,6 +362,8 @@ TVM_REGISTER_TARGET_KIND("rocm", kDLROCM)
     .set_target_parser(UpdateROCmAttrs);
 
 TVM_REGISTER_TARGET_KIND("opencl", kDLOpenCL)
+    .add_attr_option<Integer>("max_threads_per_block", Integer(256))
```
echuraev, line 365: I suppose that `max_num_threads` has the same meaning as `max_threads_per_block`, am I right?

tqchen: I think so, the main reason is that all other targets switched to the new flags

tqchen (approved):

PR #15168, opened 2023-06-27
[TIR][Schedule] Support padding-by-factor in PadEinsum

junrushao:
This PR introduces a minor behavior change in Pad-Einsum that makes it support padding-by-factor instead of padding-by-constant-number.

### DIFF
--- a/src/tir/schedule/primitive.h
+++ b/src/tir/schedule/primitive.h
@@ -672,7 +672,6 @@ TVM_DLL StmtSRef DecomposePadding(ScheduleState self, const StmtSRef& block_sref
  */
 TVM_DLL void PadEinsum(ScheduleState self, const StmtSRef& block_sref,
                        const Array<Integer>& padding);
-
 /******** Schedule: Buffer transformation ********/
 /*!
  * \brief Compute the target buffer via rolling buffering.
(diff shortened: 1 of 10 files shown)


src/tir/schedule/primitive.h
```diff
@@ -632,8 +632,7 @@ TVM_DLL StmtSRef DecomposePadding(ScheduleState self, const StmtSRef& block_sref
  * \param block_sref The block sref that matches the Einsum pattern.
  * \param padding The padding for each block iter.
  */
-TVM_DLL void PadEinsum(ScheduleState self, const StmtSRef& block_sref,
-                       const Array<Integer>& padding);
+TVM_DLL void PadEinsum(ScheduleState self, StmtSRef block_sref, Array<Integer> padding);
```
vinx13, line 635: use const ref?

vinx13 (commented):
could you also update example here https://github.com/apache/tvm/blob/main/python/tvm/tir/schedule/schedule.py#L3515

src/tir/schedule/primitive.h
junrushao: Fixed! There's no fundamental difference though.

junrushao:
Thanks @vinx13 for the huge help fixing the tensor core case!

junrushao:
Fixed a merge conflict

spectrometerHBH (approved):

PR #15170, opened 2023-06-27
[Unity][UX][Tweak] Make it an error to mark a function private and specify a global symbol

slyubomirsky:
A small follow-up to #15140, which was merged before I could make this addition: This change throws an error if a function has been annotated as private but the user manually specifies the global symbol. Since it is a contradiction to do so, it is best to detect this error rather than have it silently work (with the effect of actually making the function public).

### DIFF
--- a/src/script/ir_builder/relax/ir.cc
+++ b/src/script/ir_builder/relax/ir.cc
@@ -86,6 +86,10 @@ void FuncAttrs(Map<String, ObjectRef> attrs) {
   if (!frame->attrs.empty()) {
     LOG(FATAL) << "ValueError: Duplicate function attrs, previous one is:\n" << frame->attrs;
   }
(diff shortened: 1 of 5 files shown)


junrushao:
This is a really cool addition! A bit off the topic: do you think we should sugar the `global_symbol` attribute to make it part of the decorator? For example,

```
@R.function(global_symbol=True) # implies the global symbol name as "my_relax_func"
def my_relax_func: ...
@I.prim_func(global_symbol=True) # implies the global symbol name as "my_relax_func"
def my_tir_func: ...
@I.prim_func # implies the private symbol
def my_private_func: ...
```

slyubomirsky:
Having syntax sugar for global symbol might make sense. However, as established in the last PR, I think we would prefer for the default not to be private

junrushao:
in fact default not being private sounds better to me :-)

tqchen:
I agree, we would go for default not private

yongwww (approved):

PR #15171, opened 2023-06-27
[Unity][UX][TIR] Implement privacy annotation for the @prim_func decorator

slyubomirsky:
Implements proposal #14899 for TIR, by the same method as PR #15140. Namely, a TIR PrimFunc can be specified to be private (without a global symbol attribute) in TVMScript in the `prim_func` decorator. By default, `PrimFunc`s are not private, so they will have a `global_symbol` attribute that is mapped to their name.

Example usage:
```python
# not private: its global symbol will be "func"
@T.prim_func
def func(...): ...

# no global symbol included
@T.prim_func(private=True)
def func(...): ...
```

This did require changing very, very many tests, unfortunately.


python/tvm/tir/schedule/testing.py
```diff
@@ -16,14 +16,27 @@
...
-from typing import Sequence, Union
+from typing import Any, Sequence, Union
 
 import tvm
 from tvm.ir import IRModule, assert_structural_equal
 from tvm.tir import PrimFunc
 from tvm.tir.schedule import Schedule, Trace
 
 
+def assert_structural_equal_gs(f1: PrimFunc, f2: PrimFunc, *args: Any, **kwargs: Any) -> None:
```
slyubomirsky, line 27: I don't like the name but I didn't want it to be too verbose because this has to be used in a lot of places. Also, is there a better file where this should reside? It's used in tests outside of those related to scheduling.

tests/python/unittest/test_tir_lower_match_buffer.py
```diff
@@ -22,10 +22,10 @@
 
 
 def _check(original, transformed):
-    mod = tvm.IRModule.from_expr(original)
+    mod = tvm.IRModule.from_expr(original.with_attr("global_symbol", "main"))
```
slyubomirsky, line 25: Should `from_expr` include a global symbol by default? The only reason I hesitated on implementing that is because Relay and Relax also use `from_expr` and I didn't want to mess with that.

tests/python/relax/test_transform_fuse_ops.py
```diff
@@ -1001,7 +1001,7 @@ def main(
 
     @I.ir_module
     class Expected:
-        @T.prim_func
+        @T.prim_func(private=True)
```
slyubomirsky, line 1004: Throughout these Relax tests, I assumed that most PrimFuncs we will add through fusion, etc, are likely not intended to be public. Please let me know if this assumption is mistaken.

python/tvm/relax/block_builder.py
```diff
@@ -346,8 +351,12 @@ def emit_te(self, func: Callable, *args: Any, **kwargs: Any) -> Var:
 
         kwargs : Any, optional
             The keyword arguments passed to the function.
-            Note that the key "primfunc_name_hint" is reserved for passing name hint
-            to the PrimFunc that gets generated.
+            Note that the below keys are reserved:
+            * "primfunc_name_hint" is reserved for passing a name hint
+              to the PrimFunc that gets generated.
+            * "primfunc_public" is reserved for indicating whether the
```
slyubomirsky, line 357: For emitting PrimFuncs from Relax, I assume private is the default we want. Happy to do it the other way if that's desirable.

slyubomirsky:
If anyone has any ideas, I don't know what's wrong with the current failing GPU tests in the last CI. Nothing in them seems to have anything to do with the changes in this PR (in both cases, no global symbols were added or removed: all PrimFuncs either already didn't have them or already had them).

tqchen:
Thank you @steven, do you mind sending this PR to main instead? we can then merge it back to unity

slyubomirsky:
Reposted for mainline: https://github.com/apache/tvm/pull/15214

PR #15173, opened 2023-06-28
[FRONTEND][TFLITE][BugFix] Fix int16 transpose conv loading

wooseok-cadence:
Loading int16 conv transpose op in tflite model currently fails because output type is not int64.

This patch adjusts output type to int64 for int16 quantized transpose convolution operation. In addition, one typo in QnnConv2DTransposeRel is fixed.

Test script is also included to evaluate the loading of int16 quantized transpose convolution op.


leandron (approved):
LGTM, thanks @wooseok-cadence

PR #15178, opened 2023-06-28
[Unity] Add memory scope and nd allocation support in allocators

quic-sanirudh:
This patch adds a new Allocator type called NDAllocator as discussed in [this discussion](https://discuss.tvm.apache.org/t/memory-scope-for-vm-alloc-storage-builtins/15172).

The new allocator allows allocating to other memory scopes and allows nd-allocation.

### DIFF
--- a/include/tvm/runtime/relax_vm/memory_manager.h
+++ b/include/tvm/runtime/relax_vm/memory_manager.h
@@ -71,6 +71,13 @@ class Allocator {
    *  \return A sized allocation in the form of a buffer.
    */
   virtual Buffer Alloc(size_t nbytes, size_t alignment, DLDataType type_hint) = 0;
+  /*! \brief Allocate a buffer given a size, alignment and type.
+   *  \param shape The shape of allocated tensor.
+   *  \param dtype A type hint to the allocator.
+   *  \param mem_scope The memory scope of allocated tensor.
+   *  \return A sized allocation in the form of a buffer.
+   */
+  virtual Buffer Alloc(ShapeTuple shape, DLDataType dtype, String mem_scope);
   /*! \brief Free a buffer allocated by the allocator.
    *  \param buffer The buffer to free.
    */
--- a/python/tvm/relax/op/vm/vm.py
+++ b/python/tvm/relax/op/vm/vm.py
@@ -17,23 +17,24 @@
 
 from typing import Union
 from . import _ffi_api
-from ...expr import Expr, Call, PrimValue, DataTypeImm, Tuple
+from ...expr import Expr, Call, PrimValue, DataTypeImm, Tuple, StringImm
 from ...utils import args_converter
 
 
 @args_converter.auto
 def alloc_storage(
-    size: Expr,
+    shape: Expr,
     runtime_device_index: Union[int, Expr],
     dtype: Union[str, Expr],
+    storage_scope: Union[str, StringImm] = "global",
 ) -> Call:
     """Construct a Call to allocate a storage with specific size,
     runtime_device_index, and dtype.
 
     Parameters
     ----------
-    size : Expr
-        The size of the storage to be allocated.
+    shape : Expr
+        The shape of the storage to be allocated.
 
     runtime_device_index : Union[int, Expr]
         The device index indicating on which device the tensor is to
@@ -42,6 +43,9 @@ def alloc_storage(
     dtype : Union[str, Expr]
         The datatype of the storage to be allocated.
 
+    storage_scope : Union[str, StringImm]
+        The storage scope of the storage to allocate. Default is global.
+
     Returns
     -------
     result : Call
@@ -49,9 +53,11 @@ def alloc_storage(
     """
     if isinstance(dtype, str):
         dtype = DataTypeImm(dtype)
+    if isinstance(storage_scope, str):
+        storage_scope = StringImm(storage_scope)
     if isinstance(runtime_device_index, int):
         runtime_device_index = PrimValue(runtime_device_index)
-    return _ffi_api.alloc_storage(size, runtime_device_index, dtype)  # type: ignore
+    return _ffi_api.alloc_storage(shape, runtime_device_index, dtype, storage_scope)  # type: ignore
 
 
 @args_converter.auto
--- a/python/tvm/runtime/relax_vm.py
+++ b/python/tvm/runtime/relax_vm.py
@@ -326,6 +326,7 @@ class VirtualMachine(object):
             If the result is a tuple, it returns a list of the fields.
             The fields are potentially also tuples, so these can be arbitrily nested.
         """
+
         # to deal with potentially nested tuples, we need to query for arity recursively
         def get_output_rec(func_name, *idx):
             arity = self._get_output_arity(func_name, *idx)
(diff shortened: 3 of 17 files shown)


quic-sanirudh:
cc @tqchen

quic-sanirudh:
With this PR, the `alloc_storage`'s `size` parameter seems to have 2 different meanings based on which allocator we use. In the existing allocators, size is always expected to be 1d and it means the actual bytes of storage. With nd-allocator, it represents the physical shape of the buffer, and size is calculated from that.

This mostly came up because of the AllocDataSpace API doing a similar thing, but I think we might need to make it both represent shape, or both represent the size of allocation. In case of 2d, it'll be size of allocation for each physical dimension.

include/tvm/runtime/relax_vm/memory_manager.h
```diff
@@ -71,6 +72,14 @@ class Allocator {
...
    */
   virtual Buffer Alloc(size_t nbytes, size_t alignment, DLDataType type_hint) = 0;
+  /*! \brief Allocate a buffer given a size, alignment and type.
+   * \param ndim The number of dimension of allocated tensor.
+   * \param shape The shape of allocated tensor.
+   *  \param type_hint A type hint to the allocator.
+   * \param mem_scope The memory scope of allocated tensor.
+   *  \return A sized allocation in the form of a buffer.
+   */
+  virtual Buffer Alloc(int ndim, const int64_t* shape, DLDataType type_hint, String mem_scope);
```
tqchen, line 82: Let us change int ndim, const int64_t* shape to ShapeTuple

quic-sanirudh: Thanks, I've updated the API with `ShapeTuple`.

quic-sanirudh: Actually I just realized that we could probably unify the 2 Alloc overloads into one and call the appropriate AllocDataSpace based on the size of the ShapeTuple. Was that intended change mentioned here?

tqchen: let us keep things in two for now as there might be some future thoughts on byte reuse, as the original alloca is dtype agnostic

include/tvm/runtime/relax_vm/memory_manager.h
```diff
@@ -71,6 +72,13 @@ class Allocator {
    *  \return A sized allocation in the form of a buffer.
    */
   virtual Buffer Alloc(size_t nbytes, size_t alignment, DLDataType type_hint) = 0;
+  /*! \brief Allocate a buffer given a size, alignment and type.
+   * \param shape The shape of allocated tensor.
+   *  \param type_hint A type hint to the allocator.
+   * \param mem_scope The memory scope of allocated tensor.
+   *  \return A sized allocation in the form of a buffer.
+   */
+  virtual Buffer Alloc(ShapeTuple shape, DLDataType type_hint, String mem_scope);
```
tqchen, line 81: AllocDataSpaceWithScope to be consistent with the original one

tqchen: In this case type_hint should just be dtype

src/runtime/relax_vm/nd_allocator.h
```diff
@@ -0,0 +1,86 @@
...
+    buf.data = runtime::DeviceAPI::Get(device_)->AllocDataSpace(device_, shape.size(), shape.data(),
+                                                                type_hint, mem_scope);
+    used_memory_.fetch_add(nbytes, std::memory_order_relaxed);
+    DLOG(INFO) << "allocate " << nbytes << " B, used memory " << used_memory_ << " B";
+    return buf;
+  }
+
+  Buffer Alloc(size_t nbytes, size_t alignment, DLDataType type_hint) override {
+    Buffer buf;
+    buf.device = device_;
```
tqchen, line 62: seems we don't need ND Allocator, just can let Naive allocator redirects to device API. We might need some better strategy later for things like hexagon allocator incase caching or pooling logic is nedded

tqchen:
Thanks @quic-sanirudh , left some extra comments. Agree about your point about allocator convention. One idea is always use the AllocWithScope API and in the case of raw bytes, pass in dtype as the element.

include/tvm/runtime/relax_vm/memory_manager.h
quic-sanirudh: right makes sense, I'll fix that, thanks.

src/runtime/relax_vm/nd_allocator.h
tqchen: Sorry, i was meant to comment on the nd version's `Alloc`. Seems that we can have both 1D and ND directly implemented in the NaiveAllocator atm.

quic-sanirudh: ah okay, yeah that makes sense. I'll get rid of the NDAllocator and implement ND version with NaiveAllocator.

quic-sanirudh: One problem I see though is that the difference in how `alloc_storage` with scope (which takes physical shape) differs from existing `alloc_storage` (that takes raw bytes) could cause confusion. Especially when we get an alloc_storage call with 1d shape, and global memory scope, do we consider the 1d shape as raw bytes, or just physical shape (number of elements)?

I think unifying this discrepancy one way or another is needed if we have to get rid of NDAllocator.

tqchen:
I actually mean that we can update `allocate_storage` to always take in shape instead of nbytes. We can always `allocate_storage` with dtype='uint8' that matches the older behavior. Then at least from VM API level the confusion is addressed, and we can continue to propagate that below

quic-sanirudh:
> I actually mean that we can update `allocate_storage` to always take in shape instead of nbytes. We can always `allocate_storage` with dtype='uint8' that matches the older behavior. Then at least from VM API level the confusion is addressed, and we can continue to propagate that below

Thanks for the clarification. That makes sense. I'll update the PR that way.

include/tvm/runtime/relax_vm/memory_manager.h
quic-sanirudh: Done thanks

src/runtime/relax_vm/nd_allocator.h
quic-sanirudh: I've removed `NDAllocator`. Thanks for the feedback.

quic-sanirudh:
I've updated the PR to change alloc_storage to take shape. Right now, the `VMBuiltinLower` is updated so that `vm.alloc_storage` always allocates with size as shape and dtype as `uint8`. One thing to note is that AllocDataSpace takes a type_hint argument, but as far as I can tell, only opengl runtime needed that and so for now, passing in a different dtype to AllocDataSpace isn't going to hurt any backends, but we may want to fix it.

I'm planning to do that in a follow-up PR where I change the `relax.memory.alloc_storage` builtin generated by StaticPlanBlockMemory pass is also updated to take in shape and generates it with proper dtype. This would also allow me to later extend it to perhaps do the memory planning on a different storage_scope and probably NDAllocation.

tqchen (approved):

tqchen:
Thanks @quic-sanirudh , this is now merged

tqchen:
also cc @MasterJH5574 @srkreddy1238

quic-sanirudh:
> Thanks @quic-sanirudh , this is now merged

Thanks a lot for all the help.

PR #15179, opened 2023-06-29
[Testing] Utility method to run TVM on remote device

junrushao:
This PR introduces `tvm.testing.rpc_run`, a utility method that allows a `runtime.Module` to run on a remote device via TVM RPC.

Example:

```python
import numpy as np
import tvm
from tvm.testing import rpc_run

@T.prim_func
def cuda_kernel(
    A: T.Buffer((128,), "float32"),
    B: T.Buffer((128,), "float32"),
):
    for bx in T.thread_binding(4, thread="blockIdx.x"):
        for tx in T.thread_binding(32, thread="threadIdx.x"):
            x = bx * 32 + tx
            B[x] = A[x] + 1.0

def main():
    np_a = np.random.randn(128).astype("float32")
    np_b = np_a + 1.0

    rt_mod = tvm.build(cuda_kernel, target="nvidia/geforce-rtx-3090-ti")
    tvm_a, tvm_b = rpc_run(
        rt_mod,
        "cuda",
        [np_a, np_b],
    )

    assert np.allclose(tvm_b, np_b)
```

Result:

```
Execution time summary:
 mean (ms)   median (ms)    max (ms)     min (ms)     std (ms)
  0.0023       0.0023       0.0023       0.0023       0.0000
```

### DIFF
--- a/python/tvm/testing/rpc_run.py
+++ b/python/tvm/testing/rpc_run.py
@@ -1,0 +1,162 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+# pylint: disable=invalid-name, missing-function-docstring
+"""A utility method to run a TVM module on a remote device."""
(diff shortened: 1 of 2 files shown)


python/tvm/testing/rpc_run.py
```diff
@@ -0,0 +1,160 @@
...
+    elif callable(export_func):
+        if output_format is None:
+            raise ValueError("output_format must be specified if `export_func` is callable")
+    else:
+        raise ValueError(f"Unsupported export_func: {export_func}")
+    return export_func, output_format
+
+
+def rpc_run(  # pylint: disable=too-many-arguments,too-many-locals
+    rt_mod: "Module",
```
MasterJH5574, line 79: Just regarding the naming here: the input module is local instead of remote, so is it a good idea to use `rt_mod`?

junrushao: well i just realized it's ambiguous - "rt" here means runtime actually

junrushao: any ideas?

MasterJH5574: I feel like my understanding has been “remote” for quite a long while... Sorry about that. Perhaps `mod` is good enough? Or do you think `mod` is exclusive for IRModule.

junrushao: OK changed to `mod` and `remote_mod`

PR #15183, opened 2023-06-29
[Runtime] Disable legalization passes when hardware supports bf16

yzh119:
# Motivation
Currently, the legalization passes would be triggered whether the hardware supports it or not.
This PR disables the pass when the pass context sets `target_support_bf16` to True.

cc @tqchen @Hzfengsy @vinx13


yzh119:
Hi @yongwww would you mind helping set up `ml_dtypes` for CI images like tlcpack/ci-arm?

PR #15184, opened 2023-06-29
[Unity] Legalization for LayoutTransform

LeshengJin:
This pr implements layout_transform legalization mentioned in https://discuss.tvm.apache.org/t/codegen-support-for-relax-layout-transform/15187.

### DIFF
--- a/python/tvm/relax/transform/legalize_ops/manipulate.py
+++ b/python/tvm/relax/transform/legalize_ops/manipulate.py
@@ -173,7 +173,7 @@ def _layout_transform(bb: BlockBuilder, call: Call) -> Expr:
     def te_layout_transform(data):
         inverse, padding_predicate = index_map.non_surjective_inverse(data.shape)
         output_shape = index_map.map_shape(data.shape)
-        if isinstance(padding_predicate, tvm.tir.expr.IntImm) and bool(padding_predicate) == False:
+        if isinstance(padding_predicate, tvm.tir.expr.IntImm) and bool(padding_predicate) is False:
             return te.compute(
                 output_shape,
                 lambda *idx: data(*inverse.map_indices(idx)),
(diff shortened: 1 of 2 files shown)


tests/python/relax/test_transform_legalize_ops_manipulate.py
```diff
@@ -1522,5 +1522,45 @@ def main(
...
 
 
+def test_layout_transform():
+    transformation = lambda a, b, c: (a, c, b // 3, b % 3)
+    pad_value = 2
+    # fmt: off
+    @I.ir_module
+    class LayoutTransform:
+        @R.function
+        def main(x: R.Tensor((10, 20, 30), "float32")):
```
psrivas2, line 1532: should we also add a test case with symbolic dimensions?

psrivas2 (approved):
Looks good! Thanks @LeshengJin !

tests/python/relax/test_transform_legalize_ops_manipulate.py
LeshengJin: Good idea! Have added a symbolic shape test case.

python/tvm/relax/transform/legalize_ops/manipulate.py
```diff
@@ -163,3 +163,30 @@ def _scatter_elements(bb: BlockBuilder, call: Call) -> Expr:
...
+def _layout_transform(bb: BlockBuilder, call: Call) -> Expr:
+    index_map: tvm.tir.IndexMap = call.attrs.index_map
+    pad_value = call.attrs.pad_value.value
+
+    def te_layout_transform(data):
+        inverse, padding_predicate = index_map.non_surjective_inverse(data.shape)
+        output_shape = index_map.map_shape(data.shape)
+        return te.compute(
+            output_shape,
+            lambda *idx: tvm.te.if_then_else(
```
vinx13, line 178: shall we simplified if_then_else when the predicate is not needed?

LeshengJin: Yes, that's better. Added an if-else statement to check if the predicated is needed.

Hzfengsy (approved):

PR #15186, opened 2023-06-30
[microNPU][ETHOSU] Add Vela's logic to select configuration block

Aleksei-grovety:
For the case when cascader is enabled, the logic of choosing the optimal configuration block from TVM will be used in other cases, the Vela's logic will be used except the cases when dev_force_block_config option is specified.


Aleksei-grovety:
@tvm-bot retun

ekalda (approved):
Thanks @Aleksei-grovety, looks all good to me 🚀

Issue #15151, opened 2023-06-24
[Bug] autotvm_matmul  Check failed:(code == RPCCode::kReturn)is false: code =kShutdown


MalaJeans:
When using the official tutorial: autotvm_matmul_x86.py, the tuning trials is set to 10,
 and the matrix size is set to N, L, M = 2043*2043*2043. An error occurred during the execution as follows:
 Check failed:(code == RPCCode::kReturn)is false: code =1
 Check failed:(code == RPCCode::kReturn)is false: code =kShutdown
 How can I solve this problem?

 My environment is:
 Ubuntu 20.04.6

Thanks!

Johnson9009:
It seems the RPC server timeout when executing, so try to increase the "session_timeout".

MalaJeans:
> 
Change the session_timeout of rpc.connect(), right？
In which file rpc.connect() needs to be modified?
Thanks!

MalaJeans:
> It seems the RPC server timeout when executing, so try to increase the "session_timeout".
Change the session_timeout of rpc.connect(), right？
In which file rpc.connect() needs to be modified?
Thanks!

MalaJeans:
@Johnson9009 
Change the session_timeout of rpc.connect(), right？
In which file rpc.connect() needs to be modified?
Thanks!

Johnson9009:
> @Johnson9009 Change the session_timeout of rpc.connect(), right？ In which file rpc.connect() needs to be modified? Thanks!

Yes, just try increase the session_timeout of rpc.connect, if you use RPC tracker, it is the session_timeout of "request".

```python3
    tracker = rpc.connect_tracker(tracker_host, tracker_port)
    remote = tracker.request(key, priority=0, session_timeout=60)
```

Johnson9009:
Please change the tile of your issue, so others can know what the problem about and help you.

MalaJeans:
OK,  thanks！I will have a try!

MrGouYou:
> > @Johnson9009 Change the session_timeout of rpc.connect(), right？ In which file rpc.connect() needs to be modified? Thanks!
> 
> Yes, just try increase the session_timeout of rpc.connect, if you use RPC tracker, it is the session_timeout of "request".
> 
> ```python
>     tracker = rpc.connect_tracker(tracker_host, tracker_port)
>     remote = tracker.request(key, priority=0, session_timeout=60)
> ```

Actually, i am running with 'LocalRunner', and  also get similar problem. 
[...]

PR #15187, opened 2023-06-30
[RPC] Report RPC Session Timeout to Client Instead of "kShutdown"

Johnson9009:
By using RPC server in NPU board, at some time a compiled model will hang the NPU, because of the buggy operator libraries of NPU toolchain, so we must to use the session_timeout to ensure the board resource can be released by the hang jobs.

Currently the handling of session timeout error in RPC server is not good, it just kill the server loop sub process, then in the destructor of  class `RPCEndpoint` will send the code of `kShutdown` to the RPC client, but the RPC client expect receive the code of `kReturn` or `kException`, so users will see the error message that like the one reported in  https://github.com/apache/tvm/issues/15151, this error report will make users very confused and don't know what's happened.

When using tuning to search a good schedule for operators, we only want to ignore the RPC session timeout error that indicate the schedule generated is an illegal one, but other error reported by the RPC server may help us find the potential bug of our tool chain built on top of TVM, so the RPC session timeout error should be split to a standalone TVM error class.

This PR implemented these requirements by sending the RPC session timeout error message as a PRC server exception to the RPC client before kill the server loop sub process.

### DIFF
--- a/python/tvm/error.py
+++ b/python/tvm/error.py
@@ -62,6 +62,11 @@ class RPCError(TVMError):
 
 
 @register_error
+class RPCSessionTimeoutError(RPCError, TimeoutError):
+    """Error thrown by the remote server when the RPC session has expired."""
+
+
+@register_error
 class OpError(TVMError):
     """Base class of all operator errors in frontends."""
 
--- a/python/tvm/rpc/server.py
+++ b/python/tvm/rpc/server.py
@@ -25,6 +25,7 @@ Server is TCP based with the following protocol:
    - {server|client}:device-type[:random-key] [-timeout=timeout]
 """
 # pylint: disable=invalid-name
+import os
 import ctypes
 import socket
 import select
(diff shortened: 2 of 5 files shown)


junrushao:
This is a nice addition! I'm curious what its implication is to the existing auto tuning system though - for example, will it affect AutoTVM's time out mechanism? CC @zxybazh

Johnson9009:
> This is a nice addition! I'm curious what its implication is to the existing auto tuning system though - for example, will it affect AutoTVM's time out mechanism? CC @zxybazh

@junrushao In my opinion, there isn't any effect for the existing auto tuning system, because currently they all catch the very general error type, e.g., Exception, TVMError, and the RPCSessionTimeoutError is a subclass of TVMError, so it will be caught too.

https://github.com/apache/tvm/blob/e178375e1bbaaa4255d59a23f66d0595fe445991/python/tvm/autotvm/measure/measure_methods.py#L369-L375

https://github.com/apache/tvm/blob/e178375e1bbaaa4255d59a23f66d0595fe445991/python/tvm/autotvm/measure/measure_methods.py#L683-L693

https://github.com/apache/tvm/blob/e178375e1bbaaa4255d59a23f66d0595fe445991/python/tvm/auto_scheduler/measure.py#L1150-L1159

junrushao:
Yep, and that's why I am curious. Thanks for the explanation!

junrushao (approved):

PR #15188, opened 2023-06-30
[RPC] Disable socket SO_REUSEADDR for Windows

jzm-intel:
This PR fix a RPC issue on Windows by disabling socket SO_REUSEADDR flag. This flag works as expected on POSIX platforms, but will cause indeterminate behaviors on Windows sockets.

Reference: https://learn.microsoft.com/en-us/windows/win32/winsock/using-so-reuseaddr-and-so-exclusiveaddruse#using-so_reuseaddr

### DIFF
--- a/python/tvm/exec/microtvm_debug_shell.py
+++ b/python/tvm/exec/microtvm_debug_shell.py
@@ -32,6 +32,7 @@ import argparse
 import logging
 import socket
 import struct
+import sys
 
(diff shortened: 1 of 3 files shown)


jzm-intel:
@tqchen Please take a look, thanks

junrushao (approved):
Interesting observation! Thanks for the bugfix on Windows!

PR #15192, opened 2023-07-02
[TIR] Support cross-threaad reduction lowering with thread-broadcasting rewrite

MasterJH5574:
This PR enhances the LowerCrossThreadReduction pass with the thread-broadcasting block rewrite.

Specifically, previously whenever a TIR block has thread-broadcast behavior (i.e., there exists some thread var which is free for the block), we never insert a predicate for the block and therefore the generated final code has race condition, which sometimes lead to wrong computation results.

This PR enhances the pass by collecting thread var information along transformation, and rewrite the thread-broadcast TIR block with additional predicate clauses which bound the thread vars and effectively state that "only execute the block when `thread_var == 0`". Therefore, the race condition issue in such blocks is resolved.

### DIFF
--- a/src/tir/transforms/lower_cross_thread_reduction.cc
+++ b/src/tir/transforms/lower_cross_thread_reduction.cc
@@ -25,11 +25,30 @@
 #include <tvm/tir/stmt_functor.h>
 #include <tvm/tir/transform.h>
 
+#include "../../runtime/thread_storage_scope.h"
+#include "../../support/utils.h"
 #include "../schedule/analysis.h"
 #include "./ir_utils.h"
 
 namespace tvm {
 namespace tir {
+
+using runtime::ThreadScope;
+using support::StartsWith;
+
+// Implement a hash and equality function for ThreadScope so that
+// ThreadScope can serve as map key class
+struct ThreadScopeHash {
+  size_t operator()(const ThreadScope& scope) const {
+    return static_cast<size_t>(scope.rank * 30 + scope.dim_index);
+  }
+};
+
+struct ThreadScopeEqual {
+  bool operator()(const ThreadScope& a, const ThreadScope& b) const {
+    return a.rank == b.rank && a.dim_index == b.dim_index;
+  }
+};
 
 /*!
  * \brief Checks if a loop is bound to threadIdx.x/y/z
@@ -478,6 +497,27 @@ class CrossThreadReductionTransformer : public StmtMutator {
     return need ? reduction_loops : std::vector<const ForNode*>{};
   }
 
+  // Check if the input block needs thread broadcast rewrite.
+  // One block needs broadcast rewrite when there exists one or more thread
+  // vars which vars free variables to this block.
+  std::vector<std::pair<ThreadScope, Range>> NeedCrossThreadBroadcast(
+      const BlockRealizeNode* realize) {
+    std::unordered_map<ThreadScope, Range, ThreadScopeHash, ThreadScopeEqual> unbound_thread2range =
+        thread2range_;
+    for (const ForNode* loop : loop_stack_) {
+      if (loop->thread_binding.defined()) {
+        ThreadScope scope = ThreadScope::Create(loop->thread_binding.value()->thread_tag);
+        unbound_thread2range.erase(scope);
+      }
+    }
+
+    std::vector<std::pair<ThreadScope, Range>> unbound_thread2range_list;
+    for (auto [scope, range] : unbound_thread2range) {
+      unbound_thread2range_list.emplace_back(scope, range);
+    }
+    return unbound_thread2range_list;
+  }
+
   /*!
    * \brief Given that the input block needs cross-thread reduction, check if cross-thread reduction
    * can be applied to the block (i.e., the block satisfies all necessary conditions of cross-thread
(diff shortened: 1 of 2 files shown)


junrushao (approved):

tqchen:
@tvm-bot rerun

src/tir/transforms/lower_cross_thread_reduction.cc
```diff
@@ -645,13 +707,66 @@ class CrossThreadReductionTransformer : public StmtMutator {
...
+    // Step 2. Update the BlockRealize with the new predicate.
+    ObjectPtr<BlockRealizeNode> p_realize = make_object<BlockRealizeNode>(*realize);
+    p_realize->predicate = std::move(predicate);
+
+    // Step 3. Wrap the updated BlockRealize with the new loops.
+    Stmt body(p_realize);
+    for (int i = 0; i < static_cast<int>(unbound_thread2range.size()); ++i) {
+      std::string dim_index(1, static_cast<char>(unbound_thread2range[i].first.dim_index + 'x'));
+      body = For(
+          /*loop_var=*/loop_vars[i],                          //
```
yzh119, line 740: What's the purpose of following `//`?

yzh119 (approved):
Overall LGTM, some slight comments :)

src/tir/transforms/lower_cross_thread_reduction.cc
```diff
@@ -578,9 +621,31 @@ class CrossThreadReductionTransformer : public StmtMutator {
...
+
+    // Collect loop-thread information:
+    // - when encountering a threadIdx loop, we keep note of its domain and
+    // the "loop var -> thread scope" relation, in order to collect all existing
+    // threads within a thread block.
+    // - we are careful about thread block boundary for safety.
+    int old_thread_block_depth = thread_block_depth_;
+    if (loop->kind == ForKind::kThreadBinding) {
+      ThreadScope scope = ThreadScope::Create(loop->thread_binding.value()->thread_tag);
+      if (scope.rank == 0 || !thread_block_depth_) {
```
tqchen, line 633: Do clear immediately after VisitLoop, since that helps to remove prior states

src/tir/transforms/lower_cross_thread_reduction.cc
```diff
@@ -645,13 +707,66 @@ class CrossThreadReductionTransformer : public StmtMutator {
...
+  Stmt VisitStmt_(const BlockRealizeNode* realize) final {
+    // Part 1. Check if the block needs cross-thread reduction rewrite.
+    std::vector<const ForNode*> reduction_loops = NeedCrossThreadReduction(realize);
+    if (!reduction_loops.empty()) {
+      // Return an empty statement, because the transformation result will
+      // be inserted when returning to the first reduction-related loop.
+      MakeCrossThreadReduction(realize, reduction_loops);
+      return Stmt{nullptr};
+    }
+
```
tqchen, line 761: Only checks if we already have cross thread reduction, this will reduce the amount of checks needed for other realize.

src/tir/transforms/lower_cross_thread_reduction.cc
```diff
@@ -478,6 +497,30 @@ class CrossThreadReductionTransformer : public StmtMutator {
...
   }
 
+  // Check if the input block needs thread broadcast rewrite.
+  // One block needs broadcast rewrite when there exists one or more thread
+  // vars which vars free variables to this block.
+  std::vector<std::pair<ThreadScope, Range>> NeedCrossThreadBroadcast(
+      const BlockRealizeNode* realize) {
+    std::unordered_map<ThreadScope, Range, ThreadScopeHash, ThreadScopeEqual> unbound_thread2range =
+        thread2range_;
+    for (const PrimExpr& iter_value : realize->iter_values) {
```
tqchen, line 507: Checking iter_values have a few disadvantages, for example, unit loop could be unbound. Instead, check the surrounding `loop_stack_` which is much more robust

tqchen (changes requested):

src/tir/transforms/lower_cross_thread_reduction.cc
MasterJH5574: To force a line break for formatter. So that it doesn't format it into
```c++
For(/*loop_var=*/loop_vars[i], /*min=*/...)
```

src/tir/transforms/lower_cross_thread_reduction.cc
MasterJH5574: Now changed the logic to

> When exiting a loop,
> * if it is a `blockIdx` loop, clear the map,
> * if it is a `threadIdx` loop, clear the map when both `threadIdx` depth and `blockIdx` depth are 0.

src/tir/transforms/lower_cross_thread_reduction.cc
MasterJH5574: Yep updated, as this pattern basically appears only when cross-thread reduction exists.

src/tir/transforms/lower_cross_thread_reduction.cc
MasterJH5574: Thanks for the good catch!! Updated.

tqchen (approved):

PR #15207, opened 2023-07-02
[TIR][Transform] Add LiftThreadBinding Pass

junrushao:
This PR adds a pass LiftThreadBinding to TIR.

Previously, during GPU cross-thread reduction, a temporary local buffer will be created in the RF buffer, as a concrete example:

```python
rf_local = T.alloc_buffer(..., scope="local")

// Step 1. Data parallel RF block
for tx in T.thread_binding(..., thread="threadIdx.x")
  rf_local[tx, ...] =

// Step 2. Cross-thread reduction to accumuate rf_local
for ...:
  for tx' in T.thread_binding(..., thread="threadIdx.x"):
    ... += rf_local[tx', ...]
```

In this case, the buffer `rf_local` will only be accessed by a single point `tx` or `tx'`, but during the pass `CompactBuffeRegion`, the two variables as thread bindings are treated as two separate variables, i.e. the information that `tx` and `tx'` are always identical to each other is discarded, which means the accessed region on `rf_local` are estimated as `Union({tx}, {tx'})` as opposed to `{tx}`, leading over allocation of local registers.

This pass is introduced to address this issue by lifting thread bindings to their LCAs.

### DIFF
--- a/include/tvm/tir/var.h
+++ b/include/tvm/tir/var.h
@@ -104,6 +104,12 @@ class Var : public PrimExpr {
    */
   TVM_DLL explicit Var(String name_hint, Type type_annotation, Span span = Span());
   /*!
+   * \brief Make a new copy of var with same type, but a different nam
+   * \param name The new name to be used.
+   * \return the new Var copy
+   */
+  TVM_DLL Var copy_with_name(const String& name) const;
+  /*!
    * \brief Make a new copy of var with same type, append suffix
    * \param suffix The suffix to be appended.
    * \return the new Var copy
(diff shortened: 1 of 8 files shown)


include/tvm/tir/var.h
```diff
@@ -103,6 +103,12 @@ class Var : public PrimExpr {
    * \param span The location of this object in the source code.
    */
   TVM_DLL explicit Var(String name_hint, Type type_annotation, Span span = Span());
+  /*!
+   * \brief Make a new copy of var with same type, but a different nam
+   * \param name The new name to be used.
+   * \return the new Var copy
+   */
+  TVM_DLL Var copy_with_name(const String& name) const;
```
tqchen, line 111: CamelCase

junrushao: it follows the cnovention existing methods like `Var::copy_with_dtype`

jinhongyii (approved):

PR #15212, opened 2023-07-03
[Unity] [ONNX frontend] Add ability to register custom operator for conversion

Civitasv:
I cannot find a suitable way to register custom operator of ONNX in TVM. So I try to implement it.

Actually, it is really simple, we just need to update the convert_map which is used to convert ONNX to Relax IR.

An example:

```py
import onnx
from onnx import helper, TensorProto
from tvm import relax
from tvm.relax.frontend.onnx import OnnxOpConverter
from tvm.relax.frontend.onnx import register_custom_op

op_name = "my_custom_op"
op_domain = "ai.onnx.contrib"


class MyCustomOpConverter(OnnxOpConverter):
    @classmethod
    def _impl_v13(cls, bb, inputs, attr, params):
        return relax.op.abs(inputs[0])


def create_custom_opset():
    opset = helper.make_opsetid(op_domain, 2)

    return opset


def create_custom_model():
    shape = [32, 32]
    node = helper.make_node(op_name, ["x"], ["y"], domain=op_domain)
    graph = helper.make_graph(
        [node],
        "custom_op_graph",
        inputs=[helper.make_tensor_value_info("x", TensorProto.FLOAT, shape)],
        outputs=[helper.make_tensor_value_info("y", TensorProto.FLOAT, shape)],
    )

    model = helper.make_model(graph, producer_name="custom_op_model")

    model.opset_import.append(create_custom_opset())
    return model


if __name__ == "__main__":
    model = create_custom_model()

    # try to check it using onnx.check, it should success
    try:
        onnx.checker.check_model(model)
        print("Model is valid.")
    except Exception as exception:
        print("Model is not valid: ", str(exception))

    # register new operator
    register_custom_op(op_name, MyCustomOpConverter)

    ir_mod = from_onnx(model, keep_params_in_input=True)
    ir_mod, params = relax.frontend.detach_params(ir_mod)

    ir_mod.show()
```

It will show:

```txt
Model is valid.


@I.ir_module
class Module:
    @R.function
    def main(
        x: R.Tensor((32, 32), dtype="float32")
    ) -> R.Tensor((32, 32), dtype="float32"):
        R.func_attr({"global_symbol": "main", "num_input": 1})
        with R.dataflow():
            gv: R.Tensor((32, 32), dtype="float32") = R.abs(x)
            R.output(gv)
        return gv
```


Civitasv:
@liangfu, please review it when you are free. It's a small yet useful feature.

Civitasv:
Any progress here?

Hzfengsy:
cc @jwfromm Could you please take a look?

Civitasv:
This PR is simply a *plugin*, not a *logic rewrite*, it simply add the ability for users to register non-existing operators or override existing operators. If users don't register any operators, then the code will work as before.

Hzfengsy:
Thanks @Civitasv. I'm not sure what's the "custom operator" in the importer. IMO, if there is an operator that the relax importer does not support, we can directly enhance it. So I wonder about the use case of this PR.

Civitasv:
For example, if the user want to customize the *Add* operator. Previously, they must change the source code of tvm(or update the IRModule after importing it), but with this pr, they can just maintain a customized operators library and update the implementation at frontend stage, which I think is very useful. What do you think?

Hzfengsy:
Of course, I think it's good to have one. But I'm still thinking if it's the best place to customize operators. I guess your usage is to use different low-level implementations of the operator (e.g. call library to compute matmul instead of te compute), rather than change the importing behavior (e.g. converting an ONNX operator add to relax multiply operator)
If so, operator legalizer would be a better place to customize

This does not mean I would block this PR, but want to clarify the usage. cc @tqchen for visibility.

Civitasv:
Thanks. That make context clearer.

tqchen:
If the goal is to add a customization of an existing op like Add, then having a legalizer pass that rewrites is better and achieves the same goal, also works across multiple frontends.

If the goal is to patch up non-existent operators, then having it in the importer makes sense

Civitasv:
> If the goal is to patch up non-existent operators, then having it in the importer makes sense

Yeah. The initial reason I made this PR is to provide a way to patch up non-existent operators. 

> If the goal is to add a customization of an existing op like Add, then having a legalizer pass that rewrites is better and achieves the same goal, also works across multiple frontends.

I don’t know how to do it yet. Can you give me an example?

PR #15214, opened 2023-07-03
[TIR][UX] Implement privacy annotations in TIR

slyubomirsky:
Implements proposal #14899 for TIR. Adapted from Unity PR #15171, by the same method as PR #15140. Namely, a TIR PrimFunc can be specified to be private (without a global symbol attribute) in TVMScript in the `prim_func` decorator. By default, `PrimFunc`s are not private, so they will have a `global_symbol` attribute that is mapped to their name.

Example usage:
```python
# not private: its global symbol will be "func"
@T.prim_func
def func(...): ...

# no global symbol included
@T.prim_func(private=True)
def func(...): ...
```

This did require changing very, very many tests, unfortunately.


slyubomirsky:
@junrushao Any insight on the test failures involving `check_sketches` (e.g., in `tests/python/unittest/test_meta_schedule_schedule_rule_cross_thread_reduction.py`)? Getting rid of the global symbols in those tests didn't affect them, so I'm not sure where my changes could affect these

slyubomirsky:
The test errors turned out to global symbols not matching within the meta scheduler (thanks @vinx13 for the help). Hopefully will be solved shortly.

slyubomirsky:
@tqchen @vinx13 @junrushao Now that the tests pass, I would appreciate close review, as there were very many tests changed and there seemed to be a lot of functionality affected by global symbols.

include/tvm/script/ir_builder/tir/ir.h
```diff
@@ -56,7 +56,7 @@ Buffer BufferDecl(Array<PrimExpr> shape, DataType dtype, String buffer_name, Opt
  * \brief The primitive function statement.
  * \return The PrimFuncFrame.
  */
-PrimFuncFrame PrimFunc();
+PrimFuncFrame PrimFunc(const Bool& is_private);
```
junrushao, line 59: let's use `bool` here given we don't need nullability:

```suggestion
PrimFuncFrame PrimFunc(bool is_private);
```

src/ir/function.cc
```diff
@@ -44,4 +45,16 @@ TVM_REGISTER_GLOBAL("ir.BaseFuncWithAttr")
       LOG(FATAL) << "Do not support function type " << func->GetTypeKey();
     });
 
+TVM_REGISTER_GLOBAL("ir.BaseFuncWithoutAttr")
```
junrushao, line 48: Nice! Looks like we always forgot to upstream this piece from the unity branch

python/tvm/tir/schedule/testing.py
```diff
@@ -16,14 +16,33 @@
...
-from typing import Sequence, Union
+from typing import Any, Sequence, Union
 
 import tvm
 from tvm.ir import IRModule, assert_structural_equal
 from tvm.tir import PrimFunc
 from tvm.tir.schedule import Schedule, Trace
 
 
+def assert_structural_equal_gs(
```
junrushao, line 27: shall we expand `gs` to `global_symbol`? feel free to push back if it's a huge ask (like affecting 80 files)

include/tvm/script/ir_builder/tir/ir.h
slyubomirsky: Thanks, I've done it

python/tvm/tir/schedule/testing.py
slyubomirsky: I'm not a fan of the abbreviation, but it will probably break a lot of formatting haha. I'll see what will happen

slyubomirsky: It was actually a very easy change via VSCode's "rename" shortcut and applying `black` to all files affected (only 29). :+1:

slyubomirsky:
@tvm-bot rerun

junrushao (approved):
This is extremely meaningful contribution to overall clarity and formalization of TIR! LGTM

PR #15224, opened 2023-07-04
[Unity] Fix memory statistics issues in estimate_memory_usage

Ubospica:
Previously `tvm.relax.analysis.estimate_memory_usage` will ignore `R.builtin.alloc_tensor` operators when calculating memory usage of the module **after** memory planning. However, such operators indeed increases the memory usage. 

This PR will take such operators into account so as to get correct memory statistics.


MasterJH5574 (approved):

MasterJH5574:
Need to fix lint.

PR #15231, opened 2023-07-04
[Unity][Training] Registering te gradient

Ubospica:
This PR supports a new intrinsic call_tir_with_grad to register te gradient functions. The Gradient pass will call te gradient functions when meet call_tir_with_grad nodes in forward code.

Current workflow is as follows:
```
# register te gradient function
@register_te_gradient("f_mul_grad")
def f_mul_grad(output_grad: te.Tensor, src1: te.Tensor, src2: te.Tensor, k: int):
    # returns a list of te tensors, representing gradients w.r.t. src1, src2
    # k is a constant parameter
    ...

# irmodule definition
@I.ir_module
class Module:
    @T.prim_func
    def f_mul(A, B, result):
        ...

    @R.function
    def main(a, b):
        cls = Module
        with R.dataflow():
            lv = R.call_tir_with_grad(cls.f_mul, (a, b), te_grad_name="f_mul_grad", te_grad_kwargs={"k": 1})
            gv = R.output(lv)
        return gv
```


tqchen:
Thanks @Ubospica . let us introduce a new intrinsic `call_tir_with_grad`, given this is a temp work around before tir AD.

PR #15236, opened 2023-07-05
[TIR][Schedule] Scoped CacheRead/Write producing compact region

MasterJH5574:
This PR enhances CacheRead/Write so that when a cache operation is performed under an inner block, the generated cache buffer will have the shape as compact as possible, by region consumption analysis.

The motivation of this change comes from the needs of dynamic shape TIR scheduling, in which case we may isolate a "static shape" internal block using blockize, and do further scheduling inside the internal block. For such cases, the current CacheRead/Write inside the static-shape block will still produce dynamic-shape cache buffers, which is not ideal for analysis and subsequent scheduling.

One thing that worths noting is that, to ensure the IR correctness after inserting the cache block, we will only compact the cache buffer when all the consumer blocks of the read buffer (for CacheRead) or the write buffer (for CacheWrite) are children blocks of the cache block insertion location. Otherwise we will insist allocating the full-size cache buffer.

### DIFF
--- a/src/tir/schedule/primitive.h
+++ b/src/tir/schedule/primitive.h
@@ -105,7 +105,7 @@ TVM_DLL std::vector<int64_t> SamplePerfectTile(
  *  The sampled tile size will be partitioned into two parts. The second part has a guarantee
  *  that their extent's product have a factor of `innerpart_factor`. The first part is loops at
  *  [0, partition_pos); the second part is loops at [partition_pos, n) and we will have
- *  `innerpart_factor` | \prod_{l=partition_pos}^{n-1} l.extent
+ *  `innerpart_factor` | prod_{l=partition_pos}^{n-1} l.extent
  *
  * \param rand_state The random state
  * \param extent The loop extent to be tiled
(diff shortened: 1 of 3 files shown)


src/tir/schedule/primitive.h
```diff
@@ -105,7 +105,7 @@ TVM_DLL std::vector<int64_t> SamplePerfectTile(
  *  The sampled tile size will be partitioned into two parts. The second part has a guarantee
  *  that their extent's product have a factor of `innerpart_factor`. The first part is loops at
  *  [0, partition_pos); the second part is loops at [partition_pos, n) and we will have
- *  `innerpart_factor` | \prod_{l=partition_pos}^{n-1} l.extent
+ *  `innerpart_factor` | prod_{l=partition_pos}^{n-1} l.extent
```
MasterJH5574, line 108: Updating this since Clang will warn
```
1 warning generated.
In file included from /home/ruihangl/tvm/src/tir/schedule/primitive/for_kind.cc:19:
In file included from /home/ruihangl/tvm/src/tir/schedule/primitive/../utils.h:49:
/home/ruihangl/tvm/src/tir/schedule/primitive/.././primitive.h:108:26: warning: unknown command tag name [-Wdocumentation-unknown-command]
 *  `innerpart_factor` | \prod_{l=partition_pos}^{n-1} l.extent
                         ^~~~~~
```

src/tir/schedule/primitive/cache_read_write.cc
```diff
@@ -242,22 +246,48 @@ Block MakeCacheStage(const BufferRegion& cache_region, CacheStageInfo* info,
...
+    iter_values.push_back(cache_full_region ? (axis_range->min + loop_var) : loop_var);
   }
   // block variables
   Array<IterVar> block_vars;
   // block access region for read/write buffers
-  Region access_region;
+  Region read_access_region, write_access_region;
   // indices used in block body
-  Array<PrimExpr> access_indices;
+  Array<PrimExpr> read_access_indices, write_access_indices;
```
junrushao, line 256: nit: split them into multiple lines

junrushao (approved):
Overall LGTM! Likely it won't break anything as scopes don't happen a lot in existing TIR/MS pipeline

PR #15237, opened 2023-07-05
[Relay] ExprMutator Return Origin Expr When All Fields Isn't Changed

Johnson9009:
The Python version of Relay ExprMutator class haven't do the copy on write operation, this will cause the pass write using Python version of Relay ExprMutator can't distinguish the nodes that is changed.


Johnson9009:
@junrushao @tqchen @Hzfengsy Please help to review or cc other reviewers, thanks.

tqchen (approved):

PR #15239, opened 2023-07-05
[TVMScript] Handle parsing of PrimFunc calls with non-void return

Lunderberg:
Prior to this commit, the return type of all internal function calls was hard-coded as `"void"`.  After this commit, the `GlobalVar` representing the internal function has type annotation based on the callee's signature, which is then used as the return type of the internal call.


Hzfengsy (approved):

Lunderberg:
@tvm-bot rerun

PR #15240, opened 2023-07-05
[Frontend][pytorch]Add copy_ op to pytorch frontend

shukun-ziqiangxu:
@masahi


shukun-ziqiangxu:
@shingjan @yelite

masahi:
See https://github.com/apache/tvm/pull/9375. This came up often, but so far we are not supporting this op.

PR #15241, opened 2023-07-05
[TIR] Return error code from kernels in SplitHostDevice   

Lunderberg:
Some codegen types delegate to `CodeGenCPU` for their compute kernels, as they may delegate work to packed functions.  Because `CodeGenCPU` assumes that it can return an error code at any point (e.g. when launching a parallel for loop), the compute kernel should return an error code.

This PR resolves a potential bug introduced in https://github.com/apache/tvm/pull/15127, which removed the hard-coded override of return type.

The unit tests in this PR rely on changes made in https://github.com/apache/tvm/pull/15239, to allow TVMScript to represent the call into a compute kernel with non-void return type.


tqchen:
@Lunderberg can you clarify an example that leverages CodegenCPU for kernel. Just want to make sure that our existing path of GPU kernel separated codegen continues to function as error there are propagated by call packed mechanism

Lunderberg:
> can you clarify an example that leverages CodegenCPU for kernel.

@tqchen Certainly.  This mainly comes up in a few edge cases found when debugging a single-module lowering flow (https://github.com/apache/tvm/pull/14985), used for https://github.com/apache/tvm/pull/14862.  The issue arose when a `kDLExtDev` target or a custom `TIRToRuntime` hook was implemented by subclassing `CodeGenCPU` or `CodeGenCHost` (e.g. [`CodeGenCMSISNN`](https://github.com/apache/tvm/blob/main/src/relay/backend/contrib/cmsisnn/tir_to_runtime.cc#L39)).  In those cases, the base class assumes that it is safe to return an error code (e.g. in [`CodeGenCHost::PrintGetFuncFromBackend`](https://github.com/apache/tvm/blob/main/src/target/source/codegen_c_host.cc#L221), even if that occurs within a portion that has been separated into an independent function.

These cases are mostly suppressed by the fix in https://github.com/apache/tvm/pull/15102, but can still happen if there's an explicit `T.target("my_custom_extension", host="llvm")`.  In those cases, the compute kernels occur within a function generated by `"my_custom_extension"`, with the DLTensor-unpacking should still be handled by the usual LLVM codegen.

> Just want to make sure that our existing path of GPU kernel separated codegen continues to function as error there are propagated by call packed mechanism

Definitely agreed.  I updated the original PR to limit the `int32_t` return type to targets that may be executed on the CPU, so that the separated GPU kernels are unaffected.  This is sufficient for the functionality in https://github.com/apache/tvm/pull/14862, while avoiding changes to the GPU path.

tqchen (approved):

PR #15244, opened 2023-07-05
[Runtime] Serialization/Deserialization of runtime module

sunggg:
This PR implements serialization and deserialization of such runtime module by registering `.set_creator` and `.set_repr_bytes`. 
If runtime module is binary serializable (see https://github.com/apache/tvm/pull/14406), its roundtrip is possible. 
When the runtime module is not binary serializable, such as csource_module, it will raise an error. 

cc. @tqchen


tests/python/unittest/test_roundtrip_runtime_module.py
```diff
@@ -0,0 +1,75 @@
...
+""" Test roundtrip of runtime modules """
+# pylint: disable=missing-docstring
+
+import pytest
+import tvm
+import tvm.testing
+from tvm import relay
+from tvm.meta_schedule.testing.custom_builder_runner import (
+    build_relay_with_tensorrt,
+)
```
tqchen, line 26: given this depends on relay, let us move to test/relay

src/node/structural_hash.cc
```diff
@@ -360,6 +360,46 @@ struct ADTObjTrait {
...
+      ICHECK(load_func) << load_func_key << " is not registered.";
+      // deserialize
+      runtime::Module rtmod = (*load_func)(&b64strm);
+      return RefToObjectPtr::Get(rtmod);
+    })
+    .set_repr_bytes([](const Object* n) -> std::string {
+      const auto* rtmod = static_cast<const runtime::ModuleNode*>(n);
+      std::string blob;
+      dmlc::MemoryStringStream mstrm(&blob);
+      support::Base64OutStream b64strm(&mstrm);
```
tqchen, line 390: This is not sufficient, we will need to collect the import tree, as in normal export, so not only module itself get exported, but also the imported modules

tqchen (changes requested):

tests/python/unittest/test_roundtrip_runtime_module.py
tqchen: actually, on testcase, instead of attaching a byoc, let us just compile a cuda module and attach as extra_mods, this would also cover recursive import(as cuda contains host and device modules)

tqchen:
Thanks @sunggg. The main comment is to include import tree information so we can recover modules that imports each other(e.g. a host module that imports cuda).

We can try to refactor and bring out logic in https://github.com/apache/tvm/blob/main/src/target/codegen.cc#L63 to achieve that goal. Also we need to check if the modules are all serializable and report error accordingly if they are not

tests/python/unittest/test_roundtrip_runtime_module.py
sunggg: Thanks for the suggestion, the testcase is added.

src/node/structural_hash.cc
sunggg: Thanks for the catch! This is addressed by using https://github.com/apache/tvm/blob/main/src/runtime/dso_library.cc#L152 which also collects import tree.

sunggg:
Thanks for the valuable feedback, @tqchen! 
I found the existing support for serialization/deserialization of runtime module (including its import tree) that leverages python-side `export_library` mechanism. 

To check if the modules are all exportable, it leverages the check introduced by https://github.com/apache/tvm/pull/14406. 
The example of the error message is as below:
```Bash
AssertionError: Module RelayGraphRuntimeCodegenModule should be either dso exportable or binary serializable.
```

include/tvm/target/codegen.h
```diff
@@ -77,6 +84,16 @@ std::string PackImportsToC(const runtime::Module& m, bool system_lib,
...
+
+/*
+ * Encode TVM runtime module to base64 stream
+ */
+std::string serialize(tvm::runtime::Module module);
+
+/*
+ * Decode TVM runtime module from base64 stream
+ */
+runtime::Module deserialize(std::string state);
```
tqchen, line 89: DeserializeModuleFromBase64

include/tvm/target/codegen.h
```diff
@@ -47,6 +47,13 @@ using runtime::TVMRetValue;
  */
 runtime::Module Build(IRModule mod, Target target);
 
+/*!
+ * \brief Serialize runtime module including
+ *
+ * \param mod The runtime module to serialize including its import tree.
+ */
+std::string SerializeModule(const runtime::Module& mod);
```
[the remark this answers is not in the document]
sunggg: Done

sunggg:
@tvm-bot rerun

src/target/codegen.cc
```diff
@@ -324,6 +333,46 @@ runtime::Module PackImportsToLLVM(const runtime::Module& mod, bool system_lib,
...
+};
+
+std::string SerializeModuleToBase64(tvm::runtime::Module module) {
+  static const runtime::PackedFunc* f_to_str = runtime::Registry::Get("serialize_runtime_module");
+  ICHECK(f_to_str) << "IndexError: Cannot find the packed function "
+                      "`serialize_runtime_module` in the global registry";
+  return (*f_to_str)(module);
+}
+
+tvm::runtime::Module DeserializeModuleFromBase64(std::string state) {
```
tqchen, line 353: state=> base64

- For loading (back to IRModule), actually we do not want to export load it as so, instead, we would like to directly load back, this is something that have not been implemented but likely very related with the process  blob

src/target/codegen.cc
```diff
@@ -83,17 +85,19 @@ class ModuleSerializer {
 
     for (const auto& group : mod_group_vec_) {
       ICHECK_NE(group.size(), 0) << "Every allocated group must have at least one module";
-      if (!group[0]->IsDSOExportable()) {
```
tqchen, line 86: We need two modes here. If the 

- export mode (when we call export_library): Then we save as lib as long as it is DSOExportable.
- serialize mode (when we serialize): Always serialize when it is binary serializable. report an error when it is not, since we cannot recover the model

tqchen (changes requested):

sunggg:
Thank you, @tqchen for the suggestion. By reflecting your feedback, the latest commit cleans-up the APIs as follows:
```C++
// support full roundtrip of `runtime::Module` if the root and its all submodules are binary serializable
// These are used for structural hash of `runtime::Module` and thus also used for `SaveJson` and `LoadJson`
// Please note that `SerializeModuleToBytes` supports `include_dso` to keep the existing usage (e.g., `PackImportsToC`). 
// When `include_dso` is disabled and it encounters `DSOModules`, error will be raised.  
std::string SerializeModuleToBytes(const runtime::Module& mod, bool include_dso = true);
runtime::Module DeserializeModuleFromBytes(std::string blob);

// supports export/import of `runtime::Module` if the root and all submodules are either binary serializable or dso exportable
// this is currently used for pytorch integration: https://github.com/apache/tvm/pull/12232
std::string ExportModuleToBase64(tvm::runtime::Module module);
runtime::Module ImportModuleFromBase64(std::string base64str);
```
Whenever each API faces module/submodule that is not expected, the error will be triggered. 

Also, `SHashReduce` and `SEqualReduce` are defined to enable structural equality check. 
Thus, new test-cases verify roundtrippability via structural equality check.

src/target/codegen.cc
```diff
@@ -63,13 +66,16 @@ class ModuleSerializer {
  public:
   explicit ModuleSerializer(runtime::Module mod) : mod_(mod) { Init(); }
 
-  void SerializeModule(dmlc::Stream* stream) {
+  void SerializeModuleToBytes(dmlc::Stream* stream, bool include_dso) {
```
tqchen, line 69: include_dso => export_dso

src/target/codegen.cc
```diff
@@ -83,17 +89,20 @@ class ModuleSerializer {
 
     for (const auto& group : mod_group_vec_) {
       ICHECK_NE(group.size(), 0) << "Every allocated group must have at least one module";
-      if (!group[0]->IsDSOExportable()) {
+      if (group[0]->IsBinarySerializable()) {
```
tqchen, line 92: When we are in export_dso mode, we should first check if it is IsDSOExportable, 

```c++
if (export_dso) {
    if (group[0]->IsDSOExportable()) {
     if (group[0]->IsBinarySerializable()) {
    }
} else {
    CHECK(group[0]->IsBinarySerializable());
}
```

```

File: src/node/structural_hash.cc [hunk 9]
```diff
@@ -360,6 +361,39 @@ struct ADTObjTrait {
 
 TVM_REGISTER_REFLECTION_VTABLE(runtime::ADTObj, ADTObjTrait);
 
+struct ModuleNodeTrait {
+  static constexpr const std::nullptr_t VisitAttrs = nullptr;
+  static void SHashReduce(const runtime::ModuleNode* key, SHashReducer hash_reduce) {
```
tqchen, line 366: avoid structural equality check of runtime module as it is not guaranteed to be valid for all. In this case let us avoid use structural equality for check and instead do some special handling

include/tvm/target/codegen.h
```diff
@@ -47,6 +47,37 @@ using runtime::TVMRetValue;
  */
 runtime::Module Build(IRModule mod, Target target);
 
+/*!
+ * \brief Serialize runtime module including its submodules
+ * \param mod The runtime module to serialize including its import tree.
+ * \param include_dso By default, include the info of DSOExportable modules. If disabled, an error
+ * will be raised when encountering DSO modules.
+ */
+std::string SerializeModuleToBytes(const runtime::Module& mod, bool include_dso = true);
```
tqchen, line 56: These functions should remain internal and avoid exposing for now

python/tvm/contrib/torch/optimize_torch.py
```diff
@@ -51,14 +51,6 @@ def forward(self, *torch_inputs: Tuple[torch.Tensor]):
         return ret
 
 
-@register_func("script_torch.save_to_base64")
```
tqchen, line 54: let us keep torch related functionalities untouched for now, since it would be good to first focus on runtime module serialization mechanism that **do not** handle shared libraries, but mainly binary exportables.

tqchen (changes requested):
Thanks @sunggg some followup comments

include/tvm/target/codegen.h
sunggg: `SerializeModuleToBytes` and `DeserailizeModuleFromBytes` are defined here because they are used in `src/node/structural_hash.cc`. Should I use these functions indirectly via packed function instead? 
```
TVM_REGISTER_REFLECTION_VTABLE(runtime::ModuleNode, ModuleNodeTrait)
    .set_creator([](const std::string& blob) {
      runtime::Module rtmod = codegen::DeserializeModuleFromBytes(blob);
      return RefToObjectPtr::Get(rtmod);
    })
    .set_repr_bytes([](const Object* n) -> std::string {
      const auto* rtmod = static_cast<const runtime::ModuleNode*>(n);
      return codegen::SerializeModuleToBytes(GetRef<runtime::Module>(rtmod), /*include_dso*/ false);
    });
```

python/tvm/contrib/torch/optimize_torch.py
sunggg: Thanks for the feedback. They are moved back to torch related functionalities while keeping the current changes (e.g., remove the unrecommended use of `tmpnam`, keep the better naming).

src/node/structural_hash.cc
sunggg: > avoid structural equality check of runtime module as it is not guaranteed to be valid for all.

Is it because runtime module is not always binary serializable? If so, current design would raise an error within the structural equality. I assumed the serialization is always deterministic.

src/target/codegen.cc
sunggg: Thank you for the catch, it is reflected.

include/tvm/target/codegen.h
tqchen: we should avoid serialization and implementing structural hash/equal in module for now, mainly because they are not useful for all instead rely on special checking

src/node/structural_hash.cc
sunggg: Based on offline discussion and https://github.com/apache/tvm/pull/15244#discussion_r1294883567, we agreed to disable it for now and revisit when needed in the future.

include/tvm/target/codegen.h
sunggg: Based on offline discussion, we agreed to keep `SerializeModuleToBytes` and `DeserailizeModuleFromBytes` as they are now.

src/support/base64.h
```diff
@@ -293,6 +293,40 @@ class Base64OutStream : public dmlc::Stream {
     }
   }
 };
+
+inline size_t b64strlen(const std::string b64str) {
+  ICHECK(b64str.size() % 4 == 0) << "invalid base64 encoding";
```
tqchen, line 298: Move these two functions to callers, since they are not necessarily the optimal way of decoding b64. The b64Stream is the preferred way. So avoid making them in support(and gradually we move the contrib impl over, or keep them self-contained to a sub module scope) is better

tqchen (changes requested):
Some final nits and we are good to go

src/support/base64.h
sunggg: Ah, I see. I moved them to contrib side (`src/contrib/torch/tvm_module_wrapper/RuntimeModuleWrapperTVM.cc`) since it is the only user currently.

junrushao:
Hey just wanted to chime in a little bit - are we on the right track to get it in? Seems an important PR

PR #15246, opened 2023-07-05
[Unity][Pass] `BindSymVar` pass

sunggg:
This PR introduces an utility pass that binds symbolic variables to user-provided integer values.
For example, say we have a following IRModule. 
```Python
@tvm.script.ir_module
class Before:
  @R.function
  def main(
    x: R.Tensor(("m", "n")),
    y: R.Tensor(("m", "n"))
   ) -> R.Tensor(("m", "n")):
      m = T.Var("m", "int64")
      n = T.Var("m", "int64")
      with R.dataflow():
        out = R.matmul(x, y)
        R.output(out)
      return out
```
We can conveniently bind the symbolic variable by applying `After = relax.transform.BindSymVars("main", {"m": 10, "n": 10})(Before)`.
```Python
@tvm.script.ir_module
class After:
  @R.function
  def main(
    x: R.Tensor((10, 10)),
    y: R.Tensor((10, 10))
   ) -> R.Tensor((10, 10)):
      with R.dataflow():
        out = R.matmul(x, y)
        R.output(out)
      return out
```

This would be useful when specializing shape and providing compile-time shape info (e.g., model params or batch sizes) by eliminating the need to rewrite the model. 
cc. @tqchen @psrivas2

### DIFF
--- a/include/tvm/relax/attrs/nn.h
+++ b/include/tvm/relax/attrs/nn.h
@@ -296,6 +296,17 @@ struct GroupNormAttrs : public tvm::AttrsNode<GroupNormAttrs> {
   }
 };  // struct GroupNormAttrs
 
+/*! \brief Attributes used in rms_norm operator */
+struct RMSNormAttrs : public tvm::AttrsNode<RMSNormAttrs> {
+  Array<Integer> axes;
+  double epsilon;
+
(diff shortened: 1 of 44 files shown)


sunggg:
@tvm-bot rerun

sunggg:
Closed as it is merged with another PR https://github.com/apache/tvm/pull/15509

PR #15251, opened 2023-07-06
[TIR] Blockize keeping T.init in inner block when outer block does no reduction

MasterJH5574:
This PR enhances the behavior of tir.Schedule.blockize, so that when the outer block after blockization does no effective reduction, the `T.init` part will be kept in the inner block.

No new unit test is added as the update of the existing tests in this PR has already covered the correctness and ensured the expected behavior.

### DIFF
--- a/tests/python/unittest/test_tir_schedule_blockize.py
+++ b/tests/python/unittest/test_tir_schedule_blockize.py
@@ -42,13 +42,14 @@ def test_blockize_outer():
         A: T.Buffer((128, 128), "float32"),
         B: T.Buffer((128, 128), "float32"),
     ) -> None:
-        with T.block("blockized_B"):
-            vio = T.axis.spatial(1, 0)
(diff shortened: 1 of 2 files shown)


MasterJH5574:
Depending on #15251.

tests/python/unittest/test_tir_schedule_blockize.py
```diff
@@ -355,4 +353,5 @@ def after_blocks_blockize(
 
 
 if __name__ == "__main__":
-    tvm.testing.main()
+    test_blockize_outer()
```
yzh119, line 356: Keep it `tvm.testing.main()`.

yzh119 (approved):
LGTM

MasterJH5574:
Closed since the intention of this PR seems to break MS tuning which is more complicated for fix.

PR #15252, opened 2023-07-06
[Unity][Dlight] Handle Epilogue Broadcasting

junrushao:
This PR improves the Decode-GEMV scheduling by further analyzing its epilogue pattern.

The existing behavior assumes that the outcome of cross-thread reduction stays in register files local to each thread, which is further used to calculate the epilogue in the same thread.

This strategy means the cross-thread reduction outcome is stored only on thread 0, while the other threads cannot participate in subsequent computation (i.e. epilogue). Related: https://github.com/apache/tvm/pull/15192.

When the epilogue is relatively lightweight, i.e. elementwise add, casting on scalars, this strategy is optimal. However, once the outcome needs to be broadcasted to compute over a non-trivial region, for example, act as a normalizer of `np.mean`, it would become much slower because only one thread in a thread block is effectively used.

In this case, we will need to broadcast the cross-thread reduction outcome in shared memory, making it visible to other threads, and then bind the compute region to all threads in the threadblock.

### DIFF
--- a/python/tvm/dlight/base/analysis.py
+++ b/python/tvm/dlight/base/analysis.py
@@ -17,12 +17,13 @@
 """Analysis on TIR blocks, loops and functions."""
 from typing import List, Optional, Union
 
-from tvm import tir
+from typing_extensions import Literal
+
+from tvm import ir, tir
 from tvm._ffi import get_global_func
 from tvm.target.target import Target
 from tvm.tir import Schedule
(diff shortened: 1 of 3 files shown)


junrushao:
CC @MasterJH5574 on cross-thread reduction, @Hzfengsy and @spectrometerHBH for e2e performance

Hzfengsy (approved):

PR #15259, opened 2023-07-07
[FRONTEND][TFLITE][BugFix] Fix variable typo in batchmatmul converting func

rebel-shshin:
Fix a couple of variable names for tflite frontend.

### DIFF
--- a/python/tvm/relay/frontend/tflite.py
+++ b/python/tvm/relay/frontend/tflite.py
@@ -3041,9 +3041,9 @@ class OperatorConverter(object):
                 _op.concatenate([out_batch, _op.strided_slice(shape_b, [rank_b - 2], [rank_b])], 0)
             )
             if not tvm.ir.structural_equal(shape_a, a_broadcasted_shape):
-                input_a = _op.transform.broadcast_to(a, a_broadcasted_shape)
+                input_a = _op.transform.broadcast_to(input_a, a_broadcasted_shape)
             if not tvm.ir.structural_equal(shape_b, b_broadcasted_shape):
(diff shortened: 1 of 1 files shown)


rebel-shshin:
Please take a look :-). @leandron @masahi @xiaosibai @neildhickey 
cheers~

leandron (changes requested):
Interesting that we don't see any test failing because of that. Any chance you could add a small test case to make sure it doesn't regress in future?

rebel-shshin:
> Interesting that we don't see any test failing because of that. Any chance you could add a small test case to make sure it doesn't regress in future?
> 
> Thanks for the PR @rebel-shshin

Yeap I will do that :)

rebel-shshin:
cc @leandron 
Hi @leandron, I added two test cases as you requested. Please take a look :-)

rebel-shshin:
Hi @leandron, any idea why `gpu/pr-head` fail?
I think my patch should not be affected the gpu test.

leandron:
@tvm-bot rerun

leandron:
> Hi @leandron, any idea why `gpu/pr-head` fails?
> I think my patch should not be affected the gpu test.

Tips probably a transient - I triggered a new set of CI for this patch now.

rebel-shshin:
> > Hi @leandron, any idea why `gpu/pr-head` fails?
> > I think my patch should not be affected the gpu test.
> 
> It's probably a transient - I triggered a new set of CI for this patch now.

Thanks

leandron:
This is merged now, thanks all.

Issue #15176, opened 2023-06-28
[Bug] [Relay][Keras] attribute go_backwards in LSTM converter function not be considered lead to wrong inference results


jikechao:
The following test cases produced different inference results with Keras. 
The TVM frontend didn't consider the go_backwards=True in function `_convert_lstm` (Line 985-1030) in file [relay/frontend/keras.py](https://github.com/apache/tvm/blob/main/python/tvm/relay/frontend/keras.py)
```
keras.layers.LSTM(units=32, go_backwards=True)
input_shape=[1, 2, 64]
```

### Actual behavior
![image](https://github.com/apache/tvm/assets/29506758/24c68248-af6b-49a3-bd56-1e958d5ab397)

### Triage
* frontend:keras
* needs-triage

cc @shingjan

jikechao:
This bug has been solved by PR #15261.

PR #15261, opened 2023-07-07
[bugfix][frontend][keras] Fix go_backwards attribute of LSTM in keras frontend

jikechao:
This PR fixed the bug proposed in #15176 and added bug-triggering test cases.

If `go_backwards=True`, the LSTM layer processes the input sequence backward and returns the reversed sequence.
Missing the support for `go_backwards` will lead to wrong inference results in TVM.

cc @Hzfengsy @echuraev


Hzfengsy (approved):

jikechao:
@Hzfengsy, Thanks for your review!

PR #15264, opened 2023-07-08
[TIR] Allow symbolic bounds in IndexMap analysis

junrushao:
This PR adds the bounds of shape variables to the arithmetic analyzer so that it is possible to simplify certain expressions.


src/arith/iter_affine_map.cc
```diff
@@ -2061,6 +2061,9 @@ class IterMapToExprNormalizer : public ExprMutator {
     if (analyzer_->CanProve(expr->extent == expr->source->extent) && is_one(expr->lower_factor)) {
       return source * expr->scale;
     } else if (analyzer_->CanProve(expr->source->extent == expr->lower_factor * expr->extent)) {
+      if (is_one(expr->extent) && !is_one(expr->lower_factor)) {
```
tqchen, line 2064: would be nice ot have a rationale here

tqchen: I think it is better to use

```c++
`is_one(expr->extent) && !is_one(expr->source->extent)`
```

This will cover all the non-trivial iterator case

tqchen (approved):

tqchen:
@LeiWang1999 what you met seems is related to the index bound(i32 i64 related), which is a bit unfortunate but as we transition to enable i64 it is necessary.

```
index_map = IndexMap.from_func(ldmatrix_trans_permutation_16x16_32x8_16x16, index_dtype="int32")
```

if you add index_dtype="int32", given your loops are in i32, the inverse map seems to be OK. if we don't do that, there will be cast as a result we cannot make the inverse

LeiWang1999:
Thanks tq, I found that under `index_map = IndexMap.from_func(ldmatrix_trans_permutation_16x16_32x8_16x16, index_dtype="int32")` though, the inverse map is not ok, the output is:

```bash
========================inject transform=============================

@T.prim_func
def main(A: T.Buffer((16, 16), "float16"), B: T.Buffer((16, 16), "float16")):
    # with T.block("root"):
    for i, j in T.grid(16, 16):
        with T.block("B"):
            T.reads(B[(vi * 2 + vj // 8) // 16 * 8 + (vi * 2 + vj // 8) % 8, (vi * 2 + vj // 8) % 16 // 8 * 8 + vj % 8])
            T.writes(A[vi, vj])
            A[vi, vj] = B[(vi * 2 + vj // 8) // 16 * 8 + (vi * 2 + vj // 8) % 8, (vi * 2 + vj // 8) % 16 // 8 * 8 + vj % 8]
========================inverse inject transform=============================

@T.prim_func
def main(A: T.Buffer((16, 16), "float16"), B: T.Buffer((16, 16), "float16")):
    # with T.block("root"):
    for i, j in T.grid(16, 16):yu
        with T.block("B"):
            T.reads(B[(vi * 2 + vj // 8) // 2, vj % 16])
            T.writes(A[vi, vj])
            A[vi, vj] = B[(vi * 2 + vj // 8) // 2, vj % 16]
```

After applying the inverse_map to the map, the layout should remain consistent with its state prior to the transformation. 

Before this pull request, the code was functioning as expected.

```bash
========================inject transform=============================
@T.prim_func
def func(A: T.Buffer[(16, 16), "float16"], B: T.Buffer[(16, 16), "float16"]):
    # function attr dict
    T.func_attr({"tir.noalias": True, "global_symbol": "main"})
    # body
    # with T.block("root")
    for i, j in T.grid(16, 16):
        with T.block("B"):
            T.reads(B[vi // 8 * 8 + vi % 4 * 2 + vj // 8, vi % 8 // 4 * 8 + vj % 8])
            T.writes(A[vi, vj])
            A[vi, vj] = B[vi // 8 * 8 + vi % 4 * 2 + vj // 8, vi % 8 // 4 * 8 + vj % 8]

========================inverse inject transform=============================
@T.prim_func
def func(A: T.Buffer[(16, 16), "float16"], B: T.Buffer[(16, 16), "float16"]):
    # function attr dict
    T.func_attr({"tir.noalias": True, "global_symbol": "main"})
    # body
    # with T.block("root")
    for i, j in T.grid(16, 16):
        with T.block("B"):
            T.reads(B[vi, vj])
            T.writes(A[vi, vj])
            A[vi, vj] = B[vi, vj]
 ```

I'll take a look tomorrow.

LeiWang1999:
insert `(*indices).MutateByApply(
        [&](const PrimExpr& e) { return SimplifyNonTrivialExpr(e, analyzer_); });` 
before  https://github.com/apache/tvm/blob/f36a093c20963a3827be0cf4fd5e15a0cac69f31/src/tir/schedule/primitive/layout_transformation.cc#L806-L807

to simplify the indice works for me. looks like it's removed by this commit, wonder if there’s a specific rationale behind it.

tqchen:
Previously, we would like to have option that preserves trivial simplify (simplifies iterator of range 1 to zero).

See implementation here https://github.com/apache/tvm/blob/main/src/arith/ir_mutator_with_analyzer.cc#L52

Running `SimplifyNonTrivialExpr`  before means we cannot have the non-trivial option. I think it might be fine to update  IterMapSimplifyWithContext  to include`SimplifyNonTrivialExpr` but still preserve the non-trivial option. 

I thin the best approach would be to enhance IterMapSimplify to handle this case, since this involves affine map transformation and the best case we don;t rely on other simplifier.

We can can see `(vi * 2 + vj // 8) // 16` should be readily transformed to `(vi // 8 + vj // 128)`, my suspect is that if you apply IterMap simplify again on the result it should also work (maybe some intermediate values did not further simplifes). Good to dig a bit what happens in this case without relying on other simplfier(so it have potential to work for future symbolic case)

PR #15279, opened 2023-07-10
[Bugfix][Dlight] Fix the schedule rule for decode-gemv

yzh119:
Currently, mlc-llm produces meaningless results if we compile models with dlight and set the quantization mode to `q4f16_0` (which transposes the weight matrix).
This is due to a typo in `decode-gemv`, thank @Hzfengsy for finding this bug.

cc @junrushao @Hzfengsy


Hzfengsy:
Would be great to add a regression test

junrushao (approved):

yzh119:
Regression test for speed?

yzh119:
It turns out this PR itself cannot solve the issue.
Currently failed cases include:
1. `q3f16_0`
2. `q4f16_0`
3. `q4f32_0`

tqchen:
still looks like a right direction. let us add a test and merge this in first, then iterate

MasterJH5574:
Closed since the regression is covered by #15330.

PR #15290, opened 2023-07-11
[Unity][Relax][Transform] Do not remove MatchCast for RemoveAllUnused

L0phTg:
Match_cast can be used to capture symbolic shapes, remove MatchCast binding will cause `Prim Expr xxx has not been computed` in build stage.

### DIFF
--- a/src/relax/ir/binding_rewrite.cc
+++ b/src/relax/ir/binding_rewrite.cc
@@ -249,7 +249,7 @@ class RemoveUnusedVars : public ExprMutator {
     auto prev_dfb = GetRef<DataflowBlock>(block);
     builder_->BeginDataflowBlock();
     for (Binding binding : block->bindings) {
-      if (!unused_vars.count(binding->var)) {
+      if (!unused_vars.count(binding->var) || binding.as<MatchCastNode>()) {
         VisitBinding(binding);
       }
     }
--- a/tests/python/relax/test_analysis.py
+++ b/tests/python/relax/test_analysis.py
@@ -156,6 +156,29 @@ def test_edge_binding_block_fake_unused_remove_all_unused():
     tvm.ir.assert_structural_equal(optimized, IdentityUnused["main"])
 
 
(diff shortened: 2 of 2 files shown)


L0phTg:
if we call `relax.analysis.remove_all_unused` to update this function, will cause bugs in build stage
```python
@R.function
def main(
    shape_tensor: R.Tensor((3,), dtype="int64")
) -> R.Tensor(dtype="int32", ndim=3):
    m = T.int64()
    n = T.int64()
    k = T.int64()
    with R.dataflow():
        lv: R.Shape(ndim=3) = R.call_pure_packed(
            "vm.builtin.tensor_to_shape", shape_tensor, sinfo_args=(R.Shape(ndim=3),)
        )
        lv1: R.Shape([m, n, k]) = R.match_cast(lv, R.Shape([m, n, k]))
        lv2: R.Tensor((m, 1, k), dtype="int32") = R.full(
            R.shape([m, 1, k]), R.const(1, "int32"), dtype="int32"
        )
        gv: R.Tensor((m, 1, k), dtype="int32") = lv2
        R.output(gv)
    return gv
```

Hzfengsy:
Thanks for the fixing. Could you please add a regression test, i.e., a test case that fails before this fix but passes after this patch?

L0phTg:
> Thanks for the fixing. Could you please add a regression test, i.e., a test case that fails before this fix but passes after this patch?

sure

L0phTg:
add regression test

Hzfengsy (approved):

Hzfengsy:
Please fix the CI and we can get it in

Hzfengsy:
@tvm-bot rerun

tqchen:
cc @slyubomirsky @ganler as well

tqchen:
A likely better approach is to also analyze the use of symbolic vars and remove match-cast if the generated symbolic vars are not referenced in followup struct info and Exprs. But this is a good first step to ensure correctness

ganler (approved):
LGTM. Great catch!

PR #15300, opened 2023-07-12
[Ethos-U] Replace ethos-u.constants with AllocateConst

Lunderberg:
Previously, constants for ethos-u were tracked using a function attribute `"ethos-u.constants"`.  This predates the introduction of `AllocateConst`, and had comments indicating that it should be replaced with `AllocateConst` when possible.

To minimize impact to existing passes, this commit preserves the `"ethos-u.constants"` attribute during ethosu-specific lowering passes.  The attribute is converted to `AllocateConst` at the end of the `lower_ethosu` pass, just prior to lowering with the usual TIR passes.

### DIFF
--- a/python/tvm/relay/backend/contrib/ethosu/codegen.py
+++ b/python/tvm/relay/backend/contrib/ethosu/codegen.py
@@ -16,7 +16,7 @@
 # under the License.
 """Codegen for Arm(R) Ethos(TM)-U NPU"""
 from collections import defaultdict
-from typing import List, Callable
+from typing import List, Callable, Dict
 
 from ethosu.vela import api as vapi
 import tvm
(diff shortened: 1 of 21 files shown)


Lunderberg:
This change originated while debugging https://github.com/apache/tvm/pull/14985, but ended up being unnecessary for it, and is therefore kept as a separate PR.

python/tvm/relay/backend/contrib/ethosu/codegen.py
```diff
@@ -532,6 +532,34 @@ def relay_to_tir(mod: tvm.ir.IRModule) -> tvm.ir.IRModule:
     return tir_mod
 
 
+def collect_consts(mod: tvm.IRModule) -> Dict[tvm.tir.Var, tvm.nd.NDArray]:
+    """Collect any AllocateCont
```
ekalda, line 536: Nit:
```suggestion
    """Collect any AllocateConst
```

ekalda (commented):
Thanks for these improvements @Lunderberg! Looks like it needs a rebase and I'm not sure the changes to `codegen_c_host.cc` and `split_host_device.cc` were intentional for this this patch, but otherwise looks all good to me!

Lunderberg:
Thank you for catching the extra changes in this PR.  The `codegen_c_host.cc` changes definitely don't belong here, and are removed.  The `split_host_device.cc` can probably be removed, as they're more closely related to https://github.com/apache/tvm/pull/14985.

PR #15303, opened 2023-07-12
[ETHOSU][microTVM] MicroTVM tutorial for ALIF EthosU NPU

sergio-grovety:
This PR is intended to add EthosU demo run on the ALIF AppKit board alongside with existing FVP runs.
So now you can run this tutorial on
- FVP baremetal
- FVP FreeRTOS
- [Alif AppKit board](https://alifsemi.com/support/kits/ensemble-devkit/)

To run the demo on Alif AppKit run following command:
```./run_demo.sh --alif_target_board BOARD_AppKit_Alpha1```
This will download and compile the model and than
```./flash --alif_toolkit_path path_to_SETools --alif_console_port tty_device```
to upload firmware to the MCU and get the results.

cc @ekalda, @Aleksei-grovety

### CODE REVIEW

**Review by leandron — CHANGES_REQUESTED** [review]
The existing tutorial running on the Fixed Virtual Platform is still useful for many users. Therefore, I suggest either:
* putting another demo application, or
* add toggles/command line options so that users can reuse the same tutorial in both platforms.

Finally, I think it would be great to have a written tutorial alongside https://github.com/apache/tvm/tree/main/docs/topic/microtvm to cover embedded accelerator hardware. What do you think?

**Comment by sergio-grovety** [comment]
Hi @leandron 
This is the exact behavior we implemented in this PR. We kept FVP run on both baremetal and FreeRTOS and added baremetal run on the Alif hardware. It wasn't described clearly initially but now we've updated README and PR's description.

> Finally, I think it would be great to have a written tutorial alongside https://github.com/apache/tvm/tree/main/docs/topic/microtvm to cover embedded accelerator hardware. What do you think?

Yes, that's a great idea, but we planned to write a complete tutorial in the following PR  with the ProjectAPI template for Alif's EthosU

File: apps/microtvm/ethosu/Makefile_alif.mk [hunk 1]
```diff
@@ -0,0 +1,132 @@
...
+.PHONY: clean
+
+CC := arm-none-eabi-gcc
+CXX := arm-none-eabi-g++
+AR := arm-none-eabi-ar
+OBJCOPY := arm-none-eabi-objcopy
+
+PWD := $(abspath $(shell pwd))
+BUILD_DIR := $(PWD)/build
+PLATFORM ?= RTSS_HP
```
Aleksei-grovety, line 28: Maybe add information in the README that we can choose between RTSS_HP and RTSS_HE

apps/microtvm/ethosu/run_demo.sh
```diff
@@ -148,6 +190,7 @@ python3 -m tvm.driver.tvmc compile --target=ethos-u,cmsis-nn,c \
     --target-ethos-u-accelerator_config=ethos-u55-256 \
     --target-cmsis-nn-mcpu=cortex-m55 \
     --target-c-mcpu=cortex-m55 \
+    --target-ethos-u-disable_copying_constants=1 \
```
Aleksei-grovety, line 193: Maybe add information in the README that we use mode when weights should be placed in the SRAM.

Aleksei-grovety (commented):
Thanks @sergio-grovety! Looks good just several suggestions.

apps/microtvm/ethosu/run_demo.sh
sergio-grovety: This is a demo and I'm not sure we need to go deep in describing specific parameters, because we already have pre-defined memory layout in *.ld files. Changing of the `--target-ethos-u-disable_copying_constants` parameter requires clear understanding Alif's memory and modification of LD scripts. So IMHO this topic is out of scope of the demo, but it could be a part of the another big document describing TVM on EthosU.

apps/microtvm/ethosu/README.md
```diff
@@ -16,26 +16,24 @@
 <!--- under the License. -->
 
 
-Running TVM on bare metal Arm(R) Cortex(R)-M55 CPU, Ethos(TM)-U55 NPU and CMSIS-NN
```
Aleksei-grovety, line 19: CMSIS-NN is removed from the description, but it remains in run_demo.sh as target.

sergio-grovety: Hi @Aleksei-grovety I'd like to keep CMSIS-NN target for the case if someone want to use that tvmc compilation command for their model. I've added additional comment about possible CMSIS-NN dependency inside `run_demo.sh`

Aleksei-grovety (approved):

PR #15305, opened 2023-07-12
[Runtime] Flush L2 cache in time eval

spectrometerHBH:
This PR introduces an optional cache flush functionality to `time_evaluator`. It is implemented by allocating two large empty NDArrays on the device so that the L2 cache are flushed. This gives us more accurate evaluation on the performance of a runtime function.

### DIFF
--- a/src/runtime/profiling.cc
+++ b/src/runtime/profiling.cc
@@ -861,7 +861,7 @@ TVM_REGISTER_GLOBAL("runtime.profiling.ProfileFunction")
 
 PackedFunc WrapTimeEvaluator(PackedFunc pf, Device dev, int number, int repeat, int min_repeat_ms,
                              int limit_zero_time_iterations, int cooldown_interval_ms,
-                             int repeats_to_cooldown, PackedFunc f_preproc) {
+                             int repeats_to_cooldown, int cache_flush_bytes, PackedFunc f_preproc) {
   ICHECK(pf != nullptr);
 
   if (static_cast<int>(dev.device_type) == static_cast<int>(kDLMicroDev)) {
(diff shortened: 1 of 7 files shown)


src/runtime/profiling.cc
```diff
@@ -878,6 +878,10 @@ PackedFunc WrapTimeEvaluator(PackedFunc pf, Device dev, int number, int repeat,
     // skip first time call, to activate lazy compilation components.
     pf.CallPacked(args, &temp);
 
+    // allocate two large arrays to flush L2 cache
+    NDArray arr1 = NDArray::Empty({256 * 1000000 / 4}, {kDLInt, 32, 1}, dev);
```
tqchen, line 882: How about we add a parameter, L2 flush_size, which enables us to choose to do so or not, and when it is 0 we skip

junrushao:
Also CC: @yzh119

yzh119:
I suppose we already have a `l2_cache_flush_cuda` flag in time_evalutor function: https://github.com/apache/tvm/pull/13726

and you can reuse the `l2_cache_flush_cuda` function defined in `l2_cache_flush.cc`.

spectrometerHBH:
> I suppose we already have a `l2_cache_flush_cuda` flag in time_evalutor function: #13726
> 
> and you can reuse the `l2_cache_flush_cuda` function defined in `l2_cache_flush.cc`.

I don't want it to be cuda only

yzh119:
Yes the major concern is that L2 cache size is device specific, and later architectures may have L2 cache greater than 256mb

tqchen:
To make it generalized. how about we instead introduce a l2_cache_flush_bytes, which default to 0, and use that as a parameter to indicate what array to allocate. This way it would generalize across GPUs as long as we set this argument right

junrushao:
The implementation per se is not specific to L2 either. We could say it’s cache_flush_bytes

yzh119:
cache_flush_bytes sounds good to me

yzh119 (approved):
LGTM

junrushao (approved):

tqchen (approved):

tqchen:
@tvm-bot  rerun

PR #15309, opened 2023-07-13
[Unity] Update CUTLASS Attention to incorprate upstream change

masahi:
https://github.com/NVIDIA/cutlass/pull/992

On RTX 4080, I got modest speed up on SD (49 -> 50 it / sec), and tiny speed up on Vicuna 7b (+ 0.2 - 0.3 tok / s). 

They recommend "setting blockSize to 64x128 rather than 32x128 after Sm75". On RTX 4080 + SD, there is no perf difference. But 64x128 is slightly slower than 32x128 on the attention workload in the vicuna decoder. So I didn't change the block size.   

@vinx13 @cyx-6 @sunggg @jwfromm


sunggg (approved):
LGTM. Thank you for the update!

MasterJH5574 (approved):

PR #15314, opened 2023-07-14
[Unity][OP] Add `rms_norm`

cyx-6:
This PR introduces the operator root mean square, `rms_norm`, into TOPI and relax, and its legalize transform.


MasterJH5574 (approved):
Thanks! I feel that the TOPI part can also go to the main branch which will be helpful, right?

tqchen:
agree we can send topi to main

junrushao (approved):
Very nice!

PR #15323, opened 2023-07-14
[Fix][TIR] LowerThreadAllreduce with correct thread mask

MasterJH5574:
This PR fixes a bug in the LowerThreadAllreduce pass.

Prior to this PR, in multi-group settings, the thread mask is not correctly set: when the reduction extent is 32, the thread mask will always be 0. This bug was not spotted because even when the mask is 0, the CUDA program still gives correct result. But in any way, having the zero mask is dangerous and should be fixed.


MasterJH5574:
cc @yzh119

yzh119 (approved):
Thank @MasterJH5574 for spotting the bug!

junrushao (approved):

PR #15332, opened 2023-07-16
[Runtime] Device API to query L2 cache size

yzh119:
Followup of #15305 , this PR creates API to query device L2 cache size in bytes.
Currently, the API-supported devices includes CUDA, OpenCL, and ROCM.

Note that OpenCL's API does not return the accurate device L2 cache size.
I cannot find a Vulkan API that returns L2 texture cache size, but the `vkCmdPipelineBarrier` call will flush the L2 texture cache automatically(https://zeux.io/2020/02/27/writing-an-efficient-vulkan-renderer/), thus we return 0 by default.

### DIFF
--- a/include/tvm/runtime/device_api.h
+++ b/include/tvm/runtime/device_api.h
@@ -48,7 +48,8 @@ enum DeviceAttrKind : int {
   kMaxRegistersPerBlock = 9,
   kGcnArch = 10,
   kApiVersion = 11,
-  kDriverVersion = 12
+  kDriverVersion = 12,
+  kGlobalMemCacheSizeBytes = 13,
 };
 
 #ifdef TVM_KALLOC_ALIGNMENT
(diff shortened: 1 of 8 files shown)


MasterJH5574:
Just curious what will happen if we try to get the `l2_cache_size_bytes` for other not-yet-supported targets like Metal. As long as the behavior is not undefined I think it is fine.

yzh119:
@MasterJH5574 it will return 0 by default. However metal has its own API to obtain l2 cache size: https://developer.apple.com/documentation/kernel/ml_processor_info_t/3609284-l2_cache_size/, I'll add support for metal.

MasterJH5574:
Sounds good, thank you!

MasterJH5574 (approved):

junrushao (approved):

PR #15333, opened 2023-07-16
[Unity][Dlight] Rule matmul avoiding blockIdx.z

MasterJH5574:
Prior to this PR, the matmul rule of dlight binds loops to `blockIdx.z`. However, not every device supports this blockIdx dimension (for example, WebGPU does not support `blockIdx.z`), which makes dlight fails to apply and build.

Therefore, this PR fuses the `blockIdx.z` loop with other `blockIdx` loop.


MasterJH5574:
cc @Hzfengsy

python/tvm/dlight/gpu/matmul.py
```diff
@@ -328,7 +328,7 @@ def is_spatial(block: BlockRV) -> bool:
         by, vy, ty, yi = sch.split(y, [None, vthread_y, block_size_y, micro_size_y])
         ko, ki = sch.split(k, factors=[None, micro_size_k])
         sch.reorder(bx, by, vy, vx, ty, tx, ko, ki, yi, xi)
-        sch.bind(batch, "blockIdx.z")
+        bx = sch.fuse(batch, bx)
```
yzh119, line 331: I suppose fusing `batch` with `by` will more faithfully inherit the semantics of `sch.bind(batch, "blockIdx.z")` because the physical id is `blockIdx.z * (gridDim.x * gridDim.y) + blockIdx.y * (gridDim.x) + blockIdx.x`.

yzh119 (approved):
LGTM, thank you for the patch.

python/tvm/dlight/gpu/matmul.py
MasterJH5574: Yes you're right. Just updated. Does it look good to you?

yzh119: It looks great, thank you!

PR #15334, opened 2023-07-17
[Bugfix][Relay][Frontend][Keras] Add a assertion to reject a invalid value for attribute units in RNN layers

jikechao:
This PR fixes two types of bugs:
Bug 1. fix a typo about the wrong API name in the test case.
Bug 2-5. add an assertion to reject the invalid model, which can be constructed in the old version of Keras. The attribute `units` in LSTM, RNN, GRU, and Dense must have a positive value. Such invalid values can be used to construct a DL model successfully and can be converted to RelayIR successfully. In this pr, we reject these invalid models containing such layers.

## Steps to reproduce to bug-2 about LSTM: (Keras <= 2.6)
```
import tvm
import tvm.relay as relay
import numpy as np
from tensorflow import keras

input_shape = (1, 2, 4)
input_data = np.random.random(size=input_shape)
dtype = 'float32'

x = layers.Input(shape=input_shape[1:], dtype=dtype)

layer = keras.layers.LSTM(units=0)        # invalid input but can not reject in keras <= 2.4

y = layer(x)
model = models.Model(x, y)
res_keras = model.predict(input_data)

print(model.summary())
res_keras2 = model(input_data)

mod, params = relay.frontend.from_keras(model, shape_dict)
print(mod)
```

In the latest version of Keras, the invalid value for units will be rejected. Thus, the test case can be triggered using the Keras <=2.4.
Thus, we cannot new test cases into the unit tests.

cc @echuraev @leandron

### DIFF
--- a/python/tvm/relay/frontend/keras.py
+++ b/python/tvm/relay/frontend/keras.py
@@ -1008,6 +1008,7 @@ def _convert_lstm(
     if keras_layer.go_backwards:
         in_data = _op.reverse(in_data, axis=1)
     units = list(weightList[0].shape)[1]
+    assert units > 0, "The value of units must be a positive integer"
     time_steps = in_shape[1]
     in_data = _op.squeeze(in_data, axis=[0])
     in_data = _op.split(in_data, indices_or_sections=time_steps, axis=0)
@@ -1051,6 +1052,7 @@ def _convert_simple_rnn(
     if keras_layer.use_bias:
         in_bias = etab.new_const(weightList[2])
     units = list(weightList[0].shape)[1]
+    assert units > 0, "The value of units must be a positive integer"
     in_data = _op.nn.batch_flatten(in_data)
     ixh = _op.nn.dense(in_data, kernel_weight, units=units)
     if keras_layer.use_bias:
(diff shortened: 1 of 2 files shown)


echuraev (commented):
In general LGTM. I have some doubts. How did these models with incorrect units number work in old versions of Keras? I'm afraid that we can break a backward compatibility by this patch. Probably we could process this units number in the way how it was implemented in Keras? Or we can check Keras version and generate assert for newer versions of Keras. What do you think?

jikechao:
@echuraev  Thanks for your comments!

If `unit =0`, the output shape will contain a zero dimension. Such models will be meaningless. It looks like a bug in the old version of Keras, and this bug was fixed in the later version. Thus, I would like to reject these invalid models.
image

In the new version of Keras, the called statement `layer = keras.layers.LSTM(units=0)` will crash immediately by checking in Keras. Thus, the newly added assertion statement only works for that old version of Keras .

jikechao:
BTW, in addition to LSTM, RNN, and GRU, the Dense has the same bug. Thus we added an extra patch to fix ` _convert_dense` simultaneously!

jikechao:
@tvm-bot rerun

echuraev (approved):
Thank you for clarification. LGTM.

jikechao:
@Hzfengsy @echuraev @leandron 
CI passed. Could you help me merge this PR?

Hzfengsy:
Thanks for the fixing!

PR #15340, opened 2023-07-17
[Unity][Dlight] Fix DecodeGeMV rule for spatial-inner with grouping

MasterJH5574:
This PR fixes a bug of DecodeGeMV dlight rule when the innermost tensor dimension is spatial with `unroll_factor` (for example, the grouping used in group quantization).

Prior to this PR, a reduction loop that is bound to threadIdx was reordered to reside outside a split spatial loop, which prevents the TIR LowerCrossThreadReduction pass to successfully apply due to some safety-guard requirement.

This PR fixes this issue by not reordering the split spatial loop after the reduction loop, so that the pass can be applied. Note that we can do this as the order of thread-binding loops does not matter.

### DIFF
--- a/python/tvm/dlight/gpu/decode_gemv.py
+++ b/python/tvm/dlight/gpu/decode_gemv.py
@@ -220,8 +220,7 @@ class DecodeGEMV(ScheduleRule):
         s = sch.fuse(*s)
         sch.reorder(s, r)
         if unroll_spatial_factor:
-            s, inner = sch.split(s, factors=[None, unroll_spatial_factor])
-            sch.reorder(s, r, inner)
+            s, _ = sch.split(s, factors=[None, unroll_spatial_factor])
         sch.bind(s, "threadIdx.x")
         sch.bind(r, "threadIdx.y")
         # Schedule epilogue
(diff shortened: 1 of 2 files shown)


python/tvm/dlight/gpu/decode_gemv.py
```diff
@@ -220,8 +220,7 @@ def _sch_inner_spatial(
         s = sch.fuse(*s)
         sch.reorder(s, r)
         if unroll_spatial_factor:
-            s, inner = sch.split(s, factors=[None, unroll_spatial_factor])
-            sch.reorder(s, r, inner)
+            s, _ = sch.split(s, factors=[None, unroll_spatial_factor])
```
MasterJH5574, line 223: We can do this as the order of thread-binding loops does not matter.

yzh119: and this implies the new loop order is `(s, inner, r)` which meets the requirement of CrossThreadReduction pass :)

yzh119 (approved):
LGTM, thank you for the patch.

junrushao (approved):

Issue #15245, opened 2023-07-05
[Unity][Tracking Issue] CoreML BYOC


sunggg:
### This issue is to track progress for Unity CoreML
- [x] P1. `PrimValue` support in `FuseOpsByPattern`: https://github.com/apache/tvm/pull/15217
- [x] P2. Handle `R.output` with `CanonicalizeBindings`: https://github.com/apache/tvm/pull/15341
- [x] P3. Match behavior between `FuseOpsByPattern(..., annotate_codegen=True)` and `MergeCompositeFunctions()`: https://github.com/apache/tvm/pull/15491
- [ ] P4. CoreML codegen and runtime: https://github.com/apache/tvm/pull/15556
- [ ] P5. Performance debugging

cc. @tqchen 

cc @quic-sanirudh

PR #15341, opened 2023-07-18
[Unity][Transform] Elide redundant bindings of dataflow vars

slyubomirsky:
This PR addresses the second issue in #15245. In particular, it handles the case that can arise when using `CanonicalizeBindings` of having a dataflow var whose only use is being bound to the output var, like so:

```python
@R.function
def main() -> R.Tensor((), "int32"):
        with R.dataflow():
            y = R.const(1)
            n = y
            R.output(n)
        return n
```
The only use for `y` is to be bound to `n`, which is the `DataflowBlock`'s output.

This PR introduces adds a pass called `FoldDataflowBlockOutput` that detects this case and eliminates the intermediate binding, leaving in this example simply this:
```python
@R.function
def main() -> R.Tensor((), "int32"):
    with R.dataflow():
        n = R.const(1)
        R.output(n)
    return n
```


slyubomirsky:
I'm a little surprised by how verbose this check turned out to be, so I wonder if there's a more elegant way to implement it that I've managed to miss. Perhaps the condition is too complex and there are more circumstances where we might want to eliminate these intermediate bindings.

I'm also not sure if the dead code elimination pass is the best place to make this happen, though I don't know what other pass would be better for this check. It could potentially be its own mini-pass.

sunggg (commented):
Thank you, @slyubomirsky for the quick action!

> I'm a little surprised by how verbose this check turned out to be, so I wonder if there's a more elegant way to implement it that I've managed to miss. Perhaps the condition is too complex and there are more circumstances where we might want to eliminate these intermediate bindings.

If we specifically target this case, one potential approach is to use dataflow pattern matcher. [Here](https://gist.github.com/sunggg/293b8bc644f6a4458eed3be424c6a995) is the example.

> I'm also not sure if the dead code elimination pass is the best place to make this happen, though I don't know what other pass would be better for this check. It could potentially be its own mini-pass.
I hear your concern and now that I think about it, this seems more like a propagation/folding than DCE. One potential approach might be to create a propagation/folding pass and implement opts like constant foldings.

slyubomirsky:
I've separated the logic out to another pass. I will investigate using the pattern language to make the matching more concise, though I think it may not actually suffice here because I would still need to search to find _which_ var to eliminate.

slyubomirsky:
Unfortunately, it seems the pattern given in @sunggg's example will not handle the case of multiple variables to fold, like this one:
```python
@R.function
def func() -> R.Tensor((), "int32"):
    with R.dataflow():
        y = R.const(1)
        z = R.const(2)
        m = y
        n = z
        R.output(m, n)
    return n
```
The pattern given will identify `y` and `m`, but not `n` and `z`.

sunggg (approved):
Turned out pattern-based approach has more issues than expected. 
Although the extension to the multiple match is possible, there are the correctness issues in the match result. This happens even for the single match case. Also, pattern definition works bit oddly as well. 
For example,
```
@I.ir_module
class Mod:
    @R.function
    def should_find(x: R.Tensor):
        with R.dataflow():
            y = R.const(1)
            n = y
            R.output(n)
        return n
        
def detect(expr):
    with PatternContext() as ctx:
        (~is_dfv()).only_used_by(is_dfv())
        # This should be (is_dfv()).only_used_by(~is_dfv())
        return ctx.match_dfb(expr)
        
print(detect(Mod["should_find"].body.blocks[0]))
# This will  return {DataflowVarPattern(): m, !(DataflowVarPattern()): y}.
# But it should be opposite. 
```
Since it is not main usecase scenario that is important now, I think we can revisit this when we have free cycle in the future. Current implementation looks good to me. 
Thank you for implementing this, @slyubomirsky!

PR #15343, opened 2023-07-18
[Unity][Runtime] Flush L2 cache in time eval

MasterJH5574:
This PR introduces an optional cache flush functionality to `time_evaluator`. It is implemented by allocating two large empty NDArrays on the device so that the L2 cache are flushed. This gives us more accurate evaluation on the performance of a runtime function.


MasterJH5574:
#15305 has been keeping failing GPU CI for more than one day long uncontrollable reason (spot got shutdown), which blocks our developing progress to some extent. Thus now sending to unity first.

Hzfengsy (approved):

PR #15344, opened 2023-07-18
[ACL] Update Compute Library to v23.05

leandron:
Update Compute Library to follow latest developments in 23.05.

Changelog: https://arm-software.github.io/ComputeLibrary/v23.05

cc @neildhickey @lhutton1 @ashutosh-arm for reviews


ashutosh-arm (approved):
LGTM - thanks @leandron 
Also, noticed there is a new patch available https://github.com/ARM-software/ComputeLibrary/releases. But I guess if our integration tests don't break, it should be find to go ahead with this major release.

PR #15346, opened 2023-07-18
Fix Cython compile for v3.0.0 release

lhutton1:
Cython `v3.0.0` was recently released (https://github.com/cython/cython/releases/tag/3.0.0) and is used in newly built docker images. This causes a compilation issue since 3.0.0 expects function definitions to be explicitly declared with the `noexcept` annotation. This change should be backwards compatible to `v0.29.31`. For more details see the discussion here: https://github.com/scipy/scipy/issues/17234#issuecomment-1493423202.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -16,7 +16,7 @@ runs:
       environment-file: conda/build-environment.yaml
       auto-activate-base: false
       use-only-tar-bz2: true
-      python-version: 3.7
+      python-version: 3.8
       condarc-file: conda/condarc
   - uses: conda-incubator/setup-miniconda@v2
     if: steps.conda1.outcome == 'failure'
@@ -26,7 +26,7 @@ runs:
       environment-file: conda/build-environment.yaml
       auto-activate-base: false
       use-only-tar-bz2: true
-      python-version: 3.7
+      python-version: 3.8
       condarc-file: conda/condarc
   - name: Conda info
     shell: pwsh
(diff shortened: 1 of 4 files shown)


lhutton1:
Windows and MacOS builds are failing as they use cython==0.29.28 which is not compatible with the new `noexcept` keyword. This functionality was added in 0.29.31 (https://github.com/cython/cython/blob/master/CHANGES.rst#02931-2022-07-27). We could either upgrade the Windows and MacOS builds to use a newer version of cython (I'm not too sure yet how to get later versions in conda), or we can constrain the version in main CI to <= 0.29.36.

Mousius:
My guess would be that because the Conda build uses `3.7` of Python, it isn't receiving newer builds of Cython.

https://github.com/apache/tvm/blob/e2d65111616dfa95797c0dd7e082e4050b71701d/conda/build-environment.yaml#L28

That may be quite concerning given `3.7` just passed end of life (https://devguide.python.org/versions/).

tqchen:
We can remove the 3.7 here from tehe build env

tqchen (approved):

junrushao:
Let's get this PR merged instead!

lhutton1:
After updating the python version to 3.8, the same issues mentioned in https://github.com/apache/tvm/pull/15346#issuecomment-1640409277 seem to persist

tqchen:
https://github.com/apache/tvm/pull/15469

PR #15347, opened 2023-07-18
[PyTorch] Add aten::view_as

huanmei9:
support aten::view_as op for the frontend of pytorch. The info of this op is available at this [link](https://pytorch.org/docs/stable/generated/torch.Tensor.view_as.html).


junrushao (approved):
Relay doesn't support tensor aliasing, meaning it could potentially lead to certain underlying issues if not taken good care of. However, there's unfortunately no alternative approach forward with Relay due to its design limitation, and therefore, I'd prefer merging this in

huanmei9:
@junrushao 
Thanks for your reply. BTW, what is tensor aliasing, and could you provide an example?

PR #15349, opened 2023-07-18
[TVMC] Add tvmc flag to print compilation time per pass

Anndrey24:
Added a new flag `--print-pass-times` for tvmc compile to provide debugging information for tvmc users using `PassTimingInstrument`.   
Also added a test to check the printing of timing results.  
Example output:  
```
Compilation time breakdown by pass:
sequential: 431498us [48us] (88.48%; 88.48%)
	RemoveUnusedFunctions: 14us [14us] (0.00%; 0.00%)
	ToBasicBlockNormalForm: 66us [66us] (0.01%; 0.02%)
	qnn.Legalize: 2167us [24us] (0.44%; 0.50%)
		InferType: 617us [617us] (0.13%; 28.48%)
```

### DIFF
--- a/python/tvm/driver/tvmc/compiler.py
+++ b/python/tvm/driver/tvmc/compiler.py
@@ -31,7 +31,7 @@ import tvm
 from tvm import autotvm, auto_scheduler
 from tvm import relay
 from tvm.driver.tvmc.registry import generate_registry_args, reconstruct_registry_entity
-from tvm.ir.instrument import PassInstrument
+from tvm.ir.instrument import PassInstrument, PassTimingInstrument
 from tvm.ir.memory_pools import WorkspaceMemoryPools
 from tvm.target import Target
 from tvm.relay.backend import Executor, Runtime
@@ -157,6 +157,11 @@ def add_compile_parser(subparsers, _, json_params):
         default="default",
         help="The output module name. Defaults to 'default'.",
     )
+    parser.add_argument(
+        "--print-pass-times",
+        action="store_true",
+        help="print compilation time per pass",
+    )
     for one_entry in json_params:
         parser.set_defaults(**one_entry)
 
@@ -191,6 +196,7 @@ def drive_compile(args):
 
     additional_targets = reconstruct_target_args(args)
     workspace_pools_target, extra_targets = target_from_cli(args.target, additional_targets)
+    instruments = [PassTimingInstrument()] if args.print_pass_times else None
     transform_args = parse_graph_transform_args(args)
 
     compile_model(
@@ -214,6 +220,7 @@ def drive_compile(args):
         workspace_pools=(
             workspace_pools_recombobulate(args, [workspace_pools_target], extra_targets)
         ),
+        instruments=instruments,
         **transform_args,
     )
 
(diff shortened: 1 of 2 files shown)


tests/python/driver/tvmc/test_command_line.py
```diff
@@ -272,3 +272,20 @@ def test_tvmc_logger_set_basicConfig(monkeypatch, tmpdir_factory, keras_simple):
...
+    # Compile model
+    module_file = os.path.join(tmpdir, "keras-tvm.tar")
+    compile_cmd = f"tvmc compile --target 'llvm' {keras_simple} --output {module_file} {print_cmd}"
+    compile_args = compile_cmd.split(" ")[1:]
+    _main(compile_args)
+
+    # Check for timing results output
+    expected_out = "Printing results of timing profile..."
+    captured_out = capsys.readouterr().out
+    assert expected_out in captured_out
```
lhutton1, line 291: We should also try to check the pass breakdown exists and that it makes some form of sense. Could we check for some known parts of the outputs like pass names and the timing units?

python/tvm/driver/tvmc/compiler.py
```diff
@@ -356,6 +363,13 @@ def compile_model(
         if codegen["config_key"] is not None:
             config[codegen["config_key"]] = codegen_from_cli["opts"]
 
+    timing_inst = None
```
lhutton1, line 366: `compile_model` is also used by the TVMC Python API, so if a user adds `PassTimingInstrument` to their list of instruments and calls `compile_model` providing the list of instruments, they may get an unexpected output printed to the console. 

Since it is relatively easy for a python API user to add the instrument and print it themselves, I think it would be better to move the printing functionality up to `drive_compile` so that the feature is only exposed to the command line API. It would also simplify checking if the option was enabled.

Happy to hear more thoughts from others though.

python/tvm/driver/tvmc/compiler.py
```diff
@@ -442,6 +456,11 @@ def compile_model(
         if dumps:
             save_dumps(package_path, dumps)
 
+        # Print compilation times per pass
+        if timing_inst:
+            print("Printing results of timing profile...")
```
lhutton1, line 461: Perhaps we could rename to "Compilation time breakdown by pass:" - I feel like it's a bit more informative

lhutton1 (commented):
Thanks @Anndrey24, I think this will be a really useful feature for debugging compilation performance issues! The patch looks great, I just had some small questions/comments below. If possible could you add a small example of the output to the PR description?

python/tvm/driver/tvmc/compiler.py
Anndrey24: We discussed this offline. Because the `PassTimingInstrument` object clears its timing profile (retrieved using `.render()`) upon exiting the `PassContext` used in `compile_model()`, thus making the results unavailable outside of its scope, the decision was to leave the printing functionality inside `compile_model()`.  
However, to facilitate using this functionality through the TVMC Python API and avoid any unexpected outputs, a `print_pass_times` bool parameter would be added to `compile_model()`, meaning the python API user does not need to add the timing instrument themselves.

lhutton1 (approved):
Thanks @Anndrey24, LGTM!

PR #15352, opened 2023-07-18
[CI] Make Graviton3 default AArch64 job runner node

Mousius:
In order to support SVE testing, migrating the current default AArch64 nodes to Graviton3 based nodes. Using r7g.large instances which have the memory requirements to support the TVM workloads.

### DIFF
--- a/ci/jenkins/templates/arm_jenkinsfile.groovy.j2
+++ b/ci/jenkins/templates/arm_jenkinsfile.groovy.j2
@@ -19,7 +19,7 @@
 
 {% call m.invoke_build(
   name='BUILD: arm',
-  node='ARM-SMALL',
+  node='ARM-GRAVITON3',
   condition='!skip_ci && is_docs_only_build != 1',
   ws='tvm/build-arm',
   docker_image='ci_arm',
@@ -40,7 +40,7 @@
 {% call(shard_index, num_shards) m.sharded_test_step(
   name="integration: aarch64",
   num_shards=4,
-  node="ARM-SMALL",
+  node="ARM-GRAVITON3",
   ws="tvm/ut-python-arm",
   platform="arm",
   docker_image="ci_arm",
@@ -57,7 +57,7 @@
 
 {% call(shard_index, num_shards) m.sharded_test_step(
   name="topi: aarch64",
-  node="ARM-SMALL",
+  node="ARM-GRAVITON3",
   ws="tvm/ut-python-arm",
   platform="arm",
   docker_image="ci_arm",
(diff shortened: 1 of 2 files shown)


ci/jenkins/templates/arm_jenkinsfile.groovy.j2
```diff
@@ -19,7 +19,7 @@
 
 {% call m.invoke_build(
   name='BUILD: arm',
-  node='ARM-SMALL',
+  node='ARM-GRAVITON3',
```
tqchen, line 22: Thanks, it would be useful to have a analysis of cost and the way we structure the tests. 

As of now running the UTs directly through e2e compilation can take up a lot of CI time. A lot of that comes from tests that are as a matter of fact integration tests

Would be great for us to isolate out a limited set of integration tests (in cases with `tests/arm_sve/`) and only run limited set of testcases over these would be useful. Like our `require_cuda` tag, while majority of tests do not have to go through the specific nodes

tqchen (changes requested):

tqchen:
Thanks, it would be useful to have a analysis of cost and the way we structure the tests.

As of now running the UTs directly through e2e compilation can take up a lot of CI time. A lot of that comes from tests that are as a matter of fact integration tests that do not have to do with SVE.

Ideally we should isolate out a limited set of integration tests(e.g. via `tests/arm_sve/`) and only run those. There is also a preliminary solution via `require_cuda` tag, that optionally disables the related tests when they are not available. As such majority of tests do not have to go through the specific hw.

Mousius:
@tqchen the new instance type is slightly more expensive on paper, as detailed below:

| CI | Instance Type | On-Demand  | Reserved | Minimum Spot |
| ---- | ---- | ------------- | ------------- | ------ |
| `ARM-GRAVITON3` | r7g.large | $87.5270 monthly | $57.8890 monthly | $49.1290 monthly |
| `ARM-SMALL` | r6g.large |  $82.3440 monthly | $51.9030 monthly | $48.0340 monthly |

However, the new generation of instance has been proven to improve performance (see: [Re:invent presentation](https://d1.awsstatic.com/events/Summits/reinvent2022/CMP327_AWS-Graviton-deep-dive-The-best-price-performance-for-your-AWS-workloads.pdf)). Which indicates this is an improvement for CI costs.

If you look at the diff, this replaces the `r6g.large` instances with `r7g.large` instances and does not add any additional nodes, replacing the existing AArch64 CI with a new instance type that can also run the SVE tests. That means the SVE tests will already be targeted appropriately to AArch64, and this isn't an entirely new set of nodes specifically catering to just SVE.

tqchen:
OK get it, seems to be good on this

tqchen (approved):

PR #15353, opened 2023-07-18
[CI] Pin cython version to fix cython compilation

lhutton1:
Cython 3.0.0 was recently released, but it is incompatible with the current .pxi definitions in python/tvm/_ffi/_python. Pinning the cython version until a working fix is created.

Posting this PR as an alternative to https://github.com/apache/tvm/pull/15346.

### DIFF
--- a/docker/install/ubuntu_install_python_package.sh
+++ b/docker/install/ubuntu_install_python_package.sh
@@ -25,7 +25,7 @@ pip3 install --upgrade \
     "Pygments>=2.4.0" \
     attrs \
     cloudpickle \
-    cython \
+    cython==0.29.34 \
     decorator \
     mypy \
     numpy==1.21.* \


Mousius (approved):
This seems sensible given updating in Conda likely requires a bit of effort

tqchen:
I think it is better to see if we can get https://github.com/apache/tvm/pull/15346 through, we could remove 3.7 requirement from the conda there

junrushao (changes requested):
Instead of pinning package to a lower version, we might want to fix this particular issue instead to get consistent CI vs user/developer-end behavior

Mousius:
@tqchen / @junrushao in an ideal world, we would fix everything. As you can see from https://github.com/apache/tvm/pull/15346, it's not straightforward due to how neglected Conda is. I will merge this to resolve the issue and get things back into a stable state; if you can correct Conda and update Cython, that'd be great, we lack the expertise in this area 😸

PR #15359, opened 2023-07-19
[CMAKE] Conditionally link "clog" in NNPack install

lhutton1:
"cpuinfo" is a direct dependency of NNPack. It is installed via the NNPack's cmake script, but the version is not pinned. The "clog" dependency was recently removed from "cpuinfo"
(https://github.com/pytorch/cpuinfo/commit/3dc310302210c1891ffcfb12ae67b11a3ad3a150), which leads to issues when compiling TVM with support for NNPack.

When building TVM with newly generated docker images and NNPack enabled, we run into the following cmake configuration error:
```
The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files: NNPACK_CLOG_CONTRIB_LIB
```

This commit intends to fix this failure while also keeping compatibility with older versions of cpuinfo. It conditionally adds "clog" to the linker libs variable if it was found.


cchung100m (approved):
LGTM, thanks to @lhutton1

ashutosh-arm (approved):
Thanks @lhutton1 for deep follow through on series of issues with CI. LGTM!

PR #15362, opened 2023-07-19
[Relay] improve SimplifyClipAndConsecutiveCast pass

kfeng123:
This PR improves SimplifyClipAndConsecutiveCast pass in several aspects.
First, it allows that the dtype of `clip` and the dtype of the last `cast` differ.
Second, it allows that there are arbitrary number of `cast` after `clip`. This is achieved via the recursion of patterns.
Both of the above two cases araise from the compilation of a prequantized model.

See the [pre-RFC](https://discuss.tvm.apache.org/t/on-the-applications-of-the-composition-of-dataflow-patterns/15344?u=kfeng123)

### DIFF
--- a/src/relay/transforms/simplify_expr.cc
+++ b/src/relay/transforms/simplify_expr.cc
@@ -160,6 +160,9 @@ class SimplifyConsecutiveCast : public DFPatternRewrite {
   DFPattern cast1_;
 };
 
+/*! If mode == 0, return true if the interval [min_value, max_value] contains the range of dtype,
+ * and return false otherwise. If mode == 1, return true if the interval [min_value, max_value] is
+ * contained by the range of dtype, and return false otherwise.*/
 bool CheckDataTypeMaxMinValue(DataType dtype, double min_value, double max_value, int mode = 0) {
   double lbound{}, ubound{};
   if (dtype.is_int() || dtype.is_uint()) {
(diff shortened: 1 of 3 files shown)


kfeng123:
Perhaps the failure of test is due to the optimization of this PR: this PR may reduce the memory the model requires.
I file a [issue](https://github.com/apache/tvm/issues/15365).
Hopefully, some experts of usmp may help to verify this issue.

kfeng123:
> Perhaps the failure of test is due to the optimization of this PR: this PR may reduce the memory the model requires. I file a [issue](https://github.com/apache/tvm/issues/15365). Hopefully, some experts of usmp may help to verify this issue.

For now I just modify the ground truth of the test case to pass tests. Not sure if this is correct. Correct me if the original ground truth is indeed correct.

src/relay/transforms/simplify_expr.cc
```diff
@@ -160,7 +160,7 @@ class SimplifyConsecutiveCast : public DFPatternRewrite {
   DFPattern cast1_;
 };
 
-bool CheckDataTypeMaxMinValue(DataType dtype, double min_value, double max_value) {
+bool CheckDataTypeMaxMinValue(DataType dtype, double min_value, double max_value, int mode = 0) {
```
Hzfengsy, line 163: Please comment the meaning of different `mode`

kfeng123: Done

Hzfengsy (approved):

Issue #15319, opened 2023-07-14
[Unity][Tracking Issue] In-place operations


slyubomirsky:
Per the [discussion on in-place updates](https://discuss.tvm.apache.org/t/discuss-inplace-update-in-dataflow-block/14669), this is a tracking issue to discuss the steps and implementation details.

- [x] Implement a basic version of the `call_tir_inplace` operator. This will handle the "simple case" described in the discussion thread, where the input tensor must be at least large enough to hold the desired output. At this stage, we will not handle memory planning for the cases where the input tensor is too small to hold the output (thus requiring the memory planner to ensure that the underlying storage is large enough).
- [x] Implement a conservative alias analysis to ensure that candidates for in-place operations are not aliased. We may also need to implement special operators to assert that values are not aliased or assert that a given PackedCall does not create aliases.
- [x] Implement a liveness analysis to determine that candidates for an in-place operation are not live later in the program.
- [x] Implement a pass that uses the alias and liveness analyses to identify candidates for in-place operations and replaces them with `call_tir_inplace` invocations. At this stage, the memory planner should also be modified to handle cases where the input tensor needs a larger underlying storage.

cc @quic-sanirudh

slyubomirsky:
@tqchen A question about implementing one of the relatively simple cases, an in-place operation where the result is smaller than the input. I discussed with @MasterJH5574 and we weren't entirely sure about how this should work.

Let's suppose we have a call `out = call_tir_inplace(some_func, (t1,), return_shape)`, where `return_shape` is smaller (in at least one dimension) than `t1`. `out` will be treated as a tensor of the smaller shape (`return_shape`), but will in reality be stored in the same place as `t1`. @MasterJH5574 points out that slicing is an example where this could arise.

The question is, do we need to have any special handling in the memory planner for this case? Would Relax's runtime treat `out` as being of `t1`'s shape even though it's supposed to be smaller? Where might we need to make changes to handle this case? I could imagine some difficulties potentially arising with stride, layout, etc.

(For now, I will implement the scenario where the output shape matches the input shape exactly.)

slyubomirsky:
[...]

PR #15372, opened 2023-07-20
[Unity][Op] Implement basic `call_tir_inplace` operator

slyubomirsky:
As the first step in proposal #15319, this PR adds a `call_tir_inplace` operator that is responsible for invoking `PrimFunc`s that perform in-place changes. Namely, this means that the `PrimFunc`s directly mutate an input and do not expect a freshly allocated output tensor like in a typical DPS style `call_tir` invocation.

`call_tir_inplace` has the same signature as `call_tir`, except with one additional argument: `inplace_indices`, which lists an argument index (or -1) corresponding to each output of the `PrimFunc` call. Suppose there are `N` tensor arguments to `call_tir_inplace`. Then `inplace_indices` has the following property:
1. If `inplace_indices[i] = j`, where `0 <= j < N`, then the `i`th output of the call will alias the `j`th argument. I.e., the `j`th argument is expected to be mutated in-place.
2. If `inplace_indices[i] = -1`, then the `i`th output of the call will be a freshly allocated tensor. The `PrimFunc` will thus expect a newly allocated tensor corresponding to that output in the usual DPS style.

Note that `call_tir_inplace` is treated as pure by the type system even though it is not, in fact, pure. This means that it is dangerous to call this operator directly! It is provided right now for testing purposes; in the future (per proposal #15319), these calls will be inserted only by passes that first verify that it is _safe_ to perform ops in-place. (We could discuss the phase ordering here; perhaps it will make more sense to simply perform the in-place lowering late in compilation, after getting rid of dataflow blocks.)


slyubomirsky:
I think the test failure is spurious

slyubomirsky:
@tvm-bot rerun

tqchen (approved):

PR #15380, opened 2023-07-21
[Target] Add Jetson Orin Nano tag

mshr-h:
cc @junrushao @zxybazh @masahi 

I have the dev board and the change works fine with it.


Hzfengsy (approved):

zxybazh (approved):
Thanks for adding this new target, LGTM!

PR #15382, opened 2023-07-21
[Unity] Avoid overloaded-virtual warnings

quic-sanirudh:
This patch fixes a couple of build warnings `-Woverloaded-virtual` that come up due to the `VisitExpr_` definitions


junrushao (approved):
Thanks for the fix!

PR #15383, opened 2023-07-21
[Unity][Dlight] Fix reduction rule, aligning last block's iters

MasterJH5574:
Prior to this PR, the reduction dlight rule sometimes may over-normalizes the last block and removes some spatial loops, making the last block have fewer block iters than previous reduction blocks. In this case, the dlight rule is not able to apply due to assertion failure while it is supposed to apply.

This PR thus adds unit block iter to the last block when it has fewer iters than other reduction blocks in the front, so that the numbers of block iters can align.

One softmax unit test is added to ensure that dlight reduction rule can work.


MasterJH5574:
Waiting for #15381 to get in first to reduce conflicts.

Hzfengsy:
Please rebase it as #15381 has been merged

Hzfengsy (approved):

PR #15385, opened 2023-07-22
[Bugfix] Work around "Internal Compiler Error" in MSVC

junrushao:
This PR found a workaround that prevents the nigthly windows pypi package of TVM Unity to be released. Basically it splits out a deeply nested method into several smaller ones so that MSVC's internal handling is supposed to be slightly simpler and less error prone.


junrushao:
Cherry-picked from https://github.com/mlc-ai/relax/pull/271

tqchen (approved):

Issue #15004, opened 2023-06-01
[Bug][Relay] func ` _expr.const(c)` get an unexpected crash: 'NoneType' object has no attribute 'dtype'


jikechao:
The PyTorch model with the `adaptive_max_pool1d` layer will lead to a crash.

### Actual behavior
```
Traceback (most recent call last):
  File "test.py", line 22, in <module>
    mod, params = relay.frontend.from_pytorch(trace, input_shapes)
  File "/workplace/software/tvm/tvm/python/tvm/relay/frontend/pytorch.py", line 4970, in from_pytorch
    outputs = converter.convert_operators(operator_nodes, outputs, ret_name)
  File "/workplace/software/tvm/tvm/python/tvm/relay/frontend/pytorch.py", line 4256, in convert_operators
    return [_wrap_const(outputs[ret_name]) for ret_name in ret_names]
  File "/workplace/software/tvm/tvm/python/tvm/relay/frontend/pytorch.py", line 4256, in <listcomp>
    return [_wrap_const(outputs[ret_name]) for ret_name in ret_names]
  File "/workplace/software/tvm/tvm/python/tvm/relay/frontend/pytorch.py", line 4403, in _wrap_const
    return _expr.const(c)
  File "/workplace/software/tvm/tvm/python/tvm/relay/expr.py", line 677, in const
    value.dtype, None
AttributeError: 'NoneType' object has no attribute 'dtype'
```

### Steps to reproduce

```
import torch
from tvm import relay
import tvm
import numpy as np
from torch.nn import Module

para_0 = torch.randn([1, 2, 2], dtype=torch.float32)

class adaptive_max_pool1d(Module):
        def forward(self, *args):
            return torch.nn.functional.adaptive_max_pool1d(args[0], 2, True)
m = adaptive_max_pool1d().float().eval()
print(m)

input_data=para_0

torch_outputs = m(input_data)
print(torch_outputs)
trace = torch.jit.trace(m, input_data)
input_shapes = [('input0', torch.Size([1, 2, 2]))]

mod, params = relay.frontend.from_pytorch(trace, input_shapes)
```

### Triage

* frontend::pytorch

cc @echuraev @shingjan

haoyang9804:
If I get it correctly, here is how the error occurs.
[...]

PR #15386, opened 2023-07-23
[Relay] Fix an adaptive_max_pool1d operator conversion bug

haoyang9804:
This is a code patch for fixing this issue: https://github.com/apache/tvm/issues/15004
Please refer to my comments in this link for the root cause of this issue.


haoyang9804:
cc @jikechao

jikechao (changes requested):
Thanks for your fixing, LGTM. Please add a test!

jikechao (approved):

jikechao:
@echuraev @Hzfengsy @vvchernov 
Please recheck it.

Hzfengsy:
Please add a regression test for it

haoyang9804:
> From one hand LGTM, from another one it confuses me due to fix for specific op was done on very high level (and CI test was constructed for specific op not for all cases). It looks like it has been assumed that outputs exist at any `ret_name in ret_names`. I need time to analyze it by my-self or explanation what is wrong with the initial assumption.

I agree that "fix for specific op was done on very high level". So I make the fix only specific to the buggy pytorch op. This is only a workaround currently. I believe there are some deeper issues make this bug happen.

jikechao:
cc @vvchernov

python/tvm/relay/frontend/pytorch.py
```diff
@@ -4291,7 +4291,11 @@ def _handel_nested_input(inputs):
 
             self.current_op.pop()
 
-        return [_wrap_const(outputs[ret_name]) for ret_name in ret_names]
+        return [
+            _wrap_const(outputs[ret_name])
+            for ret_name in ret_names
+            if ret_name != "aten::adaptive_max_pool1d_0_1"
```
vvchernov, line 4297: LGTM, but please set TODO somewhere near. Due to it is still workaround for specific op on high-level API and we do not miss it for more correct fix in the future.

vvchernov (changes requested):

python/tvm/relay/frontend/pytorch.py
haoyang9804: I found that all TODO in python requires @someone. In this issue, who should I @?

vvchernov: Of course your-self) In general It does not matter due to it does not mean that the man who wrote TODO should fix it, it is a tag for anybody to understand that there is some issue

vvchernov (approved):

echuraev (approved):
LGTM. Thanks!

PR #15392, opened 2023-07-24
[Relay] add redirecting operation to dataflow pattern graph

kfeng123:
This PR adds the redirecting operation for dataflow pattern, which can redirect a WildcardPattern to any other DFPattern. With the redirecting operation, one can construct recursive dataflow pattern. Some test cases are added.

This PR is part of the [pre-RFC](https://discuss.tvm.apache.org/t/on-the-applications-of-the-composition-of-dataflow-patterns/15344?u=kfeng123)


kfeng123:
cc @masahi

masahi (approved):

PR #15398, opened 2023-07-25
[Unity][Module] Add Core Data Structure

junrushao:
This PR introduces the core data structure for the new `nn.Module` API, including:

- Tensor, a wrapper on top of relax.Expr whose struct_info is a TensorStructInfo, providing more convenient access shape and dtype information. Tensor is always symbolc and not bound to any concrete values.
- Parameter, a special tensor which could be bound or not bound to concrete values.
- Module, a container of nn.Parameters and sub nn.Modules.
- Effect, a non-user-facing class that encloses potential side effects, for example, IO, impure external function callings, inplace mutation, etc.

### DIFF
--- a/python/tvm/relax/frontend/nn/__init__.py
+++ b/python/tvm/relax/frontend/nn/__init__.py
@@ -14,6 +14,5 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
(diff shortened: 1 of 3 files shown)


junrushao:
CC: @LeshengJin @cyx-6

cyx-6 (approved):
LGTM!

PR #15401, opened 2023-07-25
[Codegen][Metal] Support metal warp-level primitive

MasterJH5574:
This PR introduces the warp-level shuffle primitives used in Metal Shading Language, and uses them in the implementation of allreduce lowering.

The introduced primitives are:
* `simd_shuffle`,
* `simd_shuffle_up`,
* `simd_shuffle_down`.

See section 6.9.2 of https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf for details.

The correctness are validated by `test_allreduce_cuda` with the backend changed to Metal. Given we do not have Metal CI tests, the correctness is checked only locally.

Given the Metal shuffle primitives do not support (or need) masking, the pass LowerThreadAllreduce is updated to support such backend which does not have masks. One unit test for metal is added to ensure that no mask is used.

### DIFF
--- a/src/target/source/intrin_rule_metal.cc
+++ b/src/target/source/intrin_rule_metal.cc
@@ -30,6 +30,28 @@ namespace codegen {
 namespace intrin {
 using tir::FLowerIntrinsic;
 
+struct MetalWarpIntrinsic {
+  const Op operator()(DataType t, const Op& orig_op) const {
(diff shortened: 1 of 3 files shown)


MasterJH5574:
Depending on #15399 to fix the allreduce correctness issues first.

tqchen (approved):

echuraev (approved):

PR #15403, opened 2023-07-25
[CI][Docker] Add riscv-gnu-toolchain to ci_riscv

katebern-grovety:
Currently the ci_riscv docker image uses GNU toolchain for Xuantie RISC-V CPU, which does not contain V-extension instructions. This PR adds GNU toolchain for RISC-V from RISC-V Collaboration. This update is required for the [PR](https://github.com/apache/tvm/pull/14836)

### DIFF
--- a/docker/install/ubuntu_install_riscv_gcc_toolchain.sh
+++ b/docker/install/ubuntu_install_riscv_gcc_toolchain.sh
@@ -32,33 +32,46 @@ if [ "$#" -lt 1 -o "$1" == "--help" -o "$1" == "-h" ]; then
     exit 1
 fi
 
-RISCV_TOOLCHAIN_VERSION="2023.07.07"
+RISCV_TOOLCHAIN_VERSION="2023.01.04"
 SPIKE_VERSION="v1.1.0"
 TMPDIR=$(mktemp -d)
 INSTALLATION_PATH=$1
-export PATH=$INSTALLATION_PATH/bin:$PATH
+INSTALLATION_PATH_32=${INSTALLATION_PATH}/32
+INSTALLATION_PATH_64=${INSTALLATION_PATH}/64
 shift
 
 # Install the packages required to build the toolchain
-sudo apt-install-and-clear -y --no-install-recommends device-tree-compiler autoconf automake autotools-dev curl libmpc-dev libmpfr-dev libgmp-dev gawk build-essential bison flex texinfo gperf libtool patchutils bc zlib1g-dev libexpat-dev ninja-build git libglib2.0-dev
+sudo apt-install-and-clear -y --no-install-recommends device-tree-compiler
 
-# Create installation path directory
-mkdir -p "${INSTALLATION_PATH}"
+# Create installation path directories
+mkdir -p "${INSTALLATION_PATH_32}"
+mkdir -p "${INSTALLATION_PATH_64}"
+cd $TMPDIR
 
 # Install RISC-V GNU toolchain
-cd $TMPDIR
-git clone https://github.com/riscv/riscv-gnu-toolchain -b $RISCV_TOOLCHAIN_VERSION
-pushd riscv-gnu-toolchain
-./configure --prefix=$INSTALLATION_PATH --enable-multilib
-make linux
-popd
+# 32
+DOWNLOAD_PATH="$TMPDIR/${RISCV_TOOLCHAIN_VERSION}_32.tar.gz"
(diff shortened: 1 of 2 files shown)


katebern-grovety:
cc @Mousius, @areusch, @leandron, @PhilippvK, @alter-xp

docker/install/ubuntu_install_riscv_gcc_toolchain.sh
```diff
@@ -0,0 +1,81 @@
...
+shift
+
+# Create installation path directory
+mkdir -p "${INSTALLATION_PATH}"
+
+# Download and extract RISC-V gcc
+TMPDIR=$(mktemp -d)
+RISCV_GCC_EXT="tar.gz"
+RLS_TAG="2023.07.07"
+RISCV_GCC_URL="https://github.com/riscv-collab/riscv-gnu-toolchain/releases/download/${RLS_TAG}/riscv64-glibc-ubuntu-22.04-gcc-nightly-${RLS_TAG}-nightly.${RISCV_GCC_EXT}"
```
PhilippvK, line 45: In contrast to the Xuantie GCC, this is a non-multilib build of the RISC-V GNU Toolchain. Support is probably limited to `rv64gc/lp64d`.

In the long term, we should find a way to support `riscv32` and `riscv64` (eventually, with different levels of floating-point support).

docker/install/ubuntu_install_riscv_gcc_toolchain.sh
```diff
@@ -0,0 +1,81 @@
...
+
+# Install pk
+git clone https://github.com/riscv/riscv-pk.git
+pushd riscv-pk
+mkdir build
+pushd build
+../configure --prefix=`pwd`/install --host=riscv64-unknown-linux-gnu
+make -j`nproc`
+make install
+cp ./pk $INSTALLATION_PATH/bin/pk
```
PhilippvK, line 76: Same issue here. RISC-V proxy kernel needs to be built separately for rv32gc and rv64gc.

PhilippvK:
The RISC-V GCC is only supposed to be used for linking and providing the `libc` headers to Clang right?

docker/install/ubuntu_install_riscv_gcc_toolchain.sh
katebern-grovety: Thanks @PhilippvK! I'll replace it with a multilib build.

docker/install/ubuntu_install_riscv_gcc_toolchain.sh
katebern-grovety: I'll add build for rv32gc and rv64gc.

katebern-grovety:
> The RISC-V GCC is only supposed to be used for linking and providing the `libc` headers to Clang right?

Yes, RISC-V GCC is used for linking and providing libs.

docker/install/ubuntu_install_riscv_gcc_toolchain.sh
PhilippvK: Do you have an guess on how long the from source build will take? I am not sure If this would be acceptable… Maybe there is a multilib build available to use here: https://github.com/sifive/freedom-tools/releases

alter-xp:
The current version of Xuantie GCC supports V-extension. However, as I mentioned in the [PR,](https://github.com/apache/tvm/pull/14836) is it syntactically correct to directly use `vwredsum `?

docker/install/ubuntu_install_riscv_gcc_toolchain.sh
katebern-grovety: I guess it will take a few hours, by the link you specified there are builds from two years ago. As a solution, I can publish a release in forked riscv-gnu-toolchain repository.

katebern-grovety:
> The current version of Xuantie GCC supports V-extension. However, as I mentioned in the [PR,](https://github.com/apache/tvm/pull/14836) is it syntactically correct to directly use `vwredsum `?

But on the current version of the toolchain, the error 
`/opt/riscv/riscv64-unknown-linux-gnu/bin/riscv64-unknown-linux-gnu-ld: -march=rv64i2p0_m2p0_a2p0_f2p0_d2p0_c2p0_v1p0_zve32f1p0_zve32x1p0_zve64d1p0_zve64f1p0_zve64x1p0_zvl128b1p0_zvl32b1p0_zvl64b1p0: Invalid or unknown z ISA extension: 'zve'` occurs, and there are no errors with the updated toolchain.

Instead of `vwredsum` I'll use `vwredsum_vs_i16m1_i32m1`.

PR #15404, opened 2023-07-25
[TIR] Allow starred expressions in TIR script

kparzysz-quic:
Small change in the evaluator to allow it to handle starred expressions (i.e. list/tuple splicing).


tqchen (approved):

Lunderberg:
I really like this, and it reduces some of the tedious parts of writing generic TIR functions.  Playing around with it, I found a few more cases where it could be useful, and opened https://github.com/apache/tvm/pull/15442 to extend the functionality to for-loops (e.g. `for *spatial,reduction in T.grid(*buf.shape)`).

PR #15412, opened 2023-07-26
[Unity][DLight] Fix Reduction Rule

Hzfengsy:
This PR fixes a correctness issue when the epilogue length is different with the `len_tx`.

cc @junrushao @MasterJH5574


MasterJH5574 (approved):
I see. Thanks for the fix. I was assuming the spatial loop is divided into bx and tx only, and was not aware that there could be another loop which is not bound to any thread (e.g., the loop with length `8` in your added test case).

PR #15413, opened 2023-07-26
[TIR] Fix Primitive Rfactor DType

Hzfengsy:
The rfactor primitive will create/rewrite two blocks, together with the block read/write regions. However, the generated read/write region extents are invalid when it's an int64 index. This commit fixes the issue.

cc @MasterJH5574


MasterJH5574 (approved):
Thanks! dtype is indeed something we didn't pay attention to when wrote these schedule primitives.

junrushao (approved):
Thanks for the fix!

PR #15414, opened 2023-07-26
[Hexagon] Add default vtcm capacity for targets

quic-sanirudh:
This patch adds VTCM default capacity values for different target architectures so that it can be queried at compile time from targets


quic-sanirudh:
cc @kparzysz-quic

PR #15419, opened 2023-07-27
[VM][Textures] Enable OpenCL textures for VM 

echuraev:
This commit introduces memory scope to VM and enables using textures.

The following changes have been made:
  - AnnotateMemoryScope pass is used in VM compilation pipeline
  - VM allows to use more than one device with the same device type.
    Also, virtual devices in VM contains information about memory
    scope.
  - Instructions LoadConst and AllocStorage were extended to support
    textures.
  - VM bytecode was updated to support memory scope.
  - Annotate texture storage pass was updated to support dynamic shape.
  - Some other minor changes have been made.
  
cc: @masahi, @csullivan, @elvin-n, @srkreddy1238


echuraev:
@masahi, @csullivan, @elvin-n, @srkreddy1238 CI is green. Could you please review this PR?

include/tvm/runtime/vm/bytecode.h
```diff
@@ -201,6 +203,10 @@ struct Instruction {
       Index alignment;
       /*! \brief The hint of the dtype. */
       DLDataType dtype_hint;
+      /*! \brief The number of dimensions. */
+      uint32_t ndim;
+      /*! \brief The shape of tensor. */
+      int64_t* shape;
```
masahi, line 209: With these information, does the existing field, `allocation_size`, become redundant? How about putting them into a union to save one int64?

echuraev: Done, put `shape` and `allocation_size` into a union.

src/runtime/vm/bytecode.cc
```diff
@@ -338,14 +340,22 @@ Instruction Instruction::AllocTensorReg(RegName storage, RegName offset, RegName
...
   Instruction instr;
   instr.op = Opcode::AllocStorage;
   instr.dst = dst;
   instr.alloc_storage.allocation_size = size;
   instr.alloc_storage.alignment = alignment;
   instr.alloc_storage.dtype_hint = dtype_hint;
   instr.alloc_storage.device_index = device_index;
+  instr.alloc_storage.ndim = static_cast<uint32_t>(shape.size());
+  if (instr.alloc_storage.ndim > 0) {
+    instr.alloc_storage.shape = new int64_t[shape.size()];
```
masahi, line 354: This alloc is leaking?

src/runtime/vm/vm.cc
```diff
@@ -899,8 +920,9 @@ void VirtualMachine::RunLoop(const std::vector<Index>& output_tensor_reg_indices
         ICHECK_EQ(actual_src_dev.device_type, inst_src_dev.device_type);
         ICHECK_EQ(actual_src_dev.device_id, inst_src_dev.device_id);
         Device dst_dev = GetDevice(instr.device_copy.dst_device_index);
+        auto mem_scope = exec_->virtual_devices[instr.device_copy.dst_device_index].second;
 
-        NDArray dst_data = src_data.CopyTo(dst_dev);
+        NDArray dst_data = src_data.CopyTo(dst_dev, String(mem_scope));
```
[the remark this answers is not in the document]
echuraev: I tried to do that, but it looks like we cannot avoid it. We cannot store `mem_scope` in `virtual_devices` as a `String` because of serialization problems. We could use `std::string` in `CopyTo` method, but it looks like in `NDArray` it is better to use `String` because of keeping uniform interface.

src/runtime/vm/bytecode.cc
echuraev: Thank you. Fixed.

src/runtime/vm/bytecode.cc
```diff
@@ -250,13 +252,15 @@ Instruction::~Instruction() {
...
     case Opcode::LoadConsti:
-    case Opcode::AllocStorage:
     case Opcode::ShapeOf:
     case Opcode::ReshapeTensor:
     case Opcode::DeviceCopy:
     case Opcode::Fatal:
     case Opcode::KillRegister:
       return;
+    case Opcode::AllocStorage:
+      delete[] this->alloc_storage.shape;
```
masahi, line 262: Need a null ptr check here.

echuraev: You are right. Thank you!

src/runtime/vm/bytecode.cc
```diff
@@ -252,13 +252,17 @@ Instruction::~Instruction() {
...
     case Opcode::LoadConsti:
-    case Opcode::AllocStorage:
     case Opcode::ShapeOf:
     case Opcode::ReshapeTensor:
     case Opcode::DeviceCopy:
     case Opcode::Fatal:
     case Opcode::KillRegister:
       return;
+    case Opcode::AllocStorage:
+      if (this->alloc_storage.ndim > 0) {
```
masahi, line 262: I'd prefer more direct

`this->alloc_storage.shape != nullptr`

since `this->alloc_storage.ndim > 0` is assuming the logic implemented elsewhere that the shape is non-null only when `this->alloc_storage.ndim > 0`.

masahi (approved):

src/runtime/vm/bytecode.cc
echuraev: I'm afraid that check on `nullptr` might be not enough, because `shape` and `allocation_size` are in the same `union`. When we change the value of `allocation_size` then `shape` is not equal to `nullptr` and when we try to release "allocated" memory in `shape` then the program might be crashed with segfault.

With current implementation of `alloc_storage` instruction, we allocate memory for shape only in case when `ndim > 0` so this is why I think that it is a valid condition for releasing memory. 

If you have any ideas how to implement this logic more clearly, then let me know.

src/runtime/vm/vm.cc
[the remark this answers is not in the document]
csullivan: Does the `instr.device_copy.dst` NDArray not have the correct memory scope already? 

1. If so, then can we use it instead of the copy here?

2. If not, then I would not expect `instr.device_copy.src` to have the correct mem_scope either. In that case you would need to construct a new ndarray from a container using the src dataptr and the correct memory scope.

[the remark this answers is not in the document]
echuraev: Sorry, @csullivan, probably I didn't get your question. So, I'll try to reply in details and please, correct me if I won't answer on your questions.

`device_copy` instruction contains three fields: `src register`, `src device index` and `dst device index`. So I extract `mem_scope` for dst memory from `dst device index` and it gets me the correct memory scope for destination array.

Answering on the first question: If you speak about copy data to dst NDArray, then no, we have to do such copy because dst NDArray will be created after this copy. If you speak about creating a string copy, then I have answered on this question in the previous message. 

Answering on the second question: I extract destination memory scope from `inst.device_copy.dst_device_index`, so extracted memory scope is correct. And I create a new NDArray from a src dataptr with correct memory scope.

Issue #15420, opened 2023-07-27
[Bug] [CI] Error installing Jax for Ubuntu ci_cpu Docker build


Liam-Sturge:
### Expected behavior

When building an Ubuntu cpu container the script `docker/install/ubuntu_install_jax.sh` should install Jaxlib correctly without issues.

### Actual behavior

Jaxlib fails to install with the following error reported:

```bash
Collecting orbax
  Using cached orbax-0.1.8.tar.gz (1.6 kB)
    ERROR: Command errored out with exit status 1:
     command: /usr/bin/python3 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-sy4y9jni/orbax/setup.py'"'"'; __file__='"'"'/tmp/pip-install-sy4y9jni/orbax/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-install-sy4y9jni/orbax/pip-egg-info
         cwd: /tmp/pip-install-sy4y9jni/orbax/
    Complete output (3 lines):
    
    *** Orbax is a namespace, and not a standalone package. For model checkpointing and exporting utilities, please install `orbax-checkpoint` and `orbax-export` respectively (instead of `orbax`). ***
    
    ----------------------------------------
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
```

This is due to the latest release of Orbax (v0.1.8) which enforces the deprecation of installing orbax as a standalone package. The version of Flax that `docker/install/ubuntu_install_jax.sh` attempts to install (v0.6.8) has `orbax` rather than `orbax-checkpoint` specified as an installation dependency.

### Environment

An x86 platform running Ubuntu 22.04.

### Steps to reproduce

Occurs when setting up an Ubuntu 22.04 cpu container using `Dockerfile.ci_cpu`.

Can also be reproduced by running the following commands in an Ubuntu environment:

```bash
pip3 install jaxlib==0.4.7 "jax[cpu]==0.4.7"
pip3 install flax==0.6.8
```

### Triage

* dev:ci

cc @Mousius @areusch @leandron

Liam-Sturge:
No longer seeing the Installation error after bumping the version of Jaxlib and Flax.

PR #15421, opened 2023-07-27
[CI] Bump Flax and Jaxlib versions to fix Jaxlib install error

Liam-Sturge:
The Flax dependency `Orbax v0.1.8` has enforced the deprecation of installing Orbax as a standalone package. Flax v0.6.8 (which is currently pinned in the script `docker/install/ubuntu_install_jax.sh`) attempts to install Orbax as a standalone package and raises an error about doing so.

Going forward, the package `orbax-checkpoint` should be installed instead. Flax v0.6.8 does not recognise this and attempts to install Orbax instead of orbax-checkpoint and the installation fails.

In order to resolve Jax installation issues, bumping the version of Flax to be at least 0.6.9, which resolves the problem. Flax >= 0.6.9 does not pin the version of orbax-checkpoint that it installs and the latest version requires Jax >= 0.4.9 to be installed so the two must be updated together.

Issue: [15420](https://github.com/apache/tvm/issues/15420) raised that details the error which occurs and steps to reproduce.

cc @leandron @Mousius @lhutton1 @ekalda @areusch @konturn @tqchen

### DIFF
--- a/docker/install/ubuntu_install_jax.sh
+++ b/docker/install/ubuntu_install_jax.sh
@@ -23,13 +23,13 @@ set -o pipefail
 # Install jax and jaxlib
 if [ "$1" == "cuda" ]; then
     pip3 install --upgrade \
-        jaxlib==0.4.7 \
-        "jax[cuda11_pip]==0.4.7" -f https://storage.googleapis.com/jax-releases/jax_cuda_releases.html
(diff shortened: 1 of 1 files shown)


lhutton1 (approved):

lhutton1:
Thanks @Liam-Sturge @yongwww

PR #15423, opened 2023-07-28
[Submodule] Add Flash attention v2

masahi:
[Flash attention v2](https://github.com/Dao-AILab/flash-attention) was released last week and it is claimed to give good speed up compared to existing implementations. I integrated flash v2 into our Relax CUTLASS BYOC and got 2-3 iter / sec speed up on SD v1.5 UNet compared to our existing flow that uses the xformer kernel.

The original code depends on libtorch so it is hard to integrate. A version without torch dep was created in https://github.com/tlc-pack/libflash_attn and I'm adding it as a submodule.

@tqchen @vinx13 @yzh119 @sunggg


tqchen:
let us also update https://github.com/apache/tvm/blob/main/LICENSE#L248 and https://github.com/apache/tvm/tree/main/licenses

junrushao:
Very cool work! This will substantially help with TVM performance on stable diffusion and LLM training workloads

tqchen (approved):

PR #15424, opened 2023-07-28
[Unity] nn.Module Torch Integration

junrushao:
This PR introduces a `nn.Module.jit` method that effectively converts an `nn.Module` to a collection of torch-in-torch-out callables for users to debug with.


cyx-6 (approved):
LGTM!

PR #15434, opened 2023-07-28
[Unity][Frontends][Onnx] Improve ConstantOfShape behavior

jwfromm:
This small PR improves how the onnx importer deals with ConstantOfShape. Previously we were creating constants that could be massive if the provided shape was large. This caused an explosion in the resulting model parameter size. Using `broadcast_to` instead provides behavior closer to the intention of the source graph.

I also noticed relax now has `relax.op.image.resize2d` so i replaced our topi call with it which should help produce more readable output graphs.

### DIFF
--- a/python/tvm/relax/frontend/onnx/onnx_frontend.py
+++ b/python/tvm/relax/frontend/onnx/onnx_frontend.py
@@ -249,7 +249,6 @@ class Unsqueeze(OnnxOpConverter):
     def _impl_v13(cls, bb, inputs, attr, params):
         data = inputs[0]
         axes = get_constant(inputs[1], params)
-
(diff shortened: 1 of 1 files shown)


junrushao (approved):

junrushao:
Thanks for this fix! It was a pain dealing with those issues

masahi:
We expect test cases for a PR like this before merging.

junrushao:
I think I agree. @jwfromm do you like to add a test case?

jwfromm:
Its unclear what the right way to handle tests for this is since there is no functional difference in the outputs from before and after. We would have to add graph analysis to the onnx frontend tests which I'm not sure we want.

PR #15435, opened 2023-07-28
[Unity] cuda graph support for cublas

sunggg:
This PR enables cuda graph with cublas.

### DIFF
--- a/src/runtime/contrib/cublas/cublas.cc
+++ b/src/runtime/contrib/cublas/cublas.cc
@@ -135,8 +135,9 @@ int roundoff(int v, int d) { return (v + d - 1) / d * d; }
 
 #if CUDART_VERSION >= 10010
 
-void CallCublasLt(cublasLtHandle_t hdl, const DLTensor* A, const DLTensor* B, const DLTensor* bias,
-                  const DLTensor* C, bool transa, bool transb, cublasLtEpilogue_t epilogue) {
+void CallCublasLt(cublasLtHandle_t hdl, cudaStream_t stream, const DLTensor* A, const DLTensor* B,
+                  const DLTensor* bias, const DLTensor* C, bool transa, bool transb,
+                  cublasLtEpilogue_t epilogue) {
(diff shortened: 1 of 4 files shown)


src/runtime/contrib/cublas/cublas.cc
```diff
@@ -239,8 +239,12 @@ void CallCublasLt(cublasLtHandle_t hdl, const DLTensor* A, const DLTensor* B, co
   auto B_data = static_cast<char*>(B->data) + B->byte_offset;
   auto C_data = static_cast<char*>(C->data) + C->byte_offset;
 
+  auto func = tvm::runtime::Registry::Get("runtime.get_cuda_stream");
+  ICHECK(func != nullptr);
+  cudaStream_t stream = static_cast<cudaStream_t>((*func)().operator void*());
```
masahi, line 244: It's better to get the stream at https://github.com/apache/tvm/blob/unity/src/runtime/contrib/cublas/cublas_json_runtime.cc#L54 once rather than repeatedly here.

sunggg: Thanks for the suggestion! Reflected.

masahi (approved):

LeiWang1999:
hi @masahi , @sunggg , would you mind providing some detailed information about the CUDA graph? I'm currently attempting to integrate cuDNN into Relax via BYOC, and any insights would be greatly appreciated.

masahi:
What information do you need?

LeiWang1999:
> What information do you need?

@masahi  thanks, 
Trustfully, I'm not sure about the purpose of the cuda graph, is this a new concept that introduced by unity?

masahi:
No, it is a standard CUDA concept to reduce host-side overhead. See
https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#cuda-graphs
https://www.nvidia.com/ja-jp/on-demand/session/gtcspring23-s51211/

LeiWang1999:
@masahi thanks! seems sth like multi stream parallel.

PR #15439, opened 2023-07-30
[Vulkan] Add spirv shuffle instruction support

masahi:
This allows using `builtin::tvm_warp_shuffle()`, `builtin::tvm_warp_shuffle_up/down()` for VK. We need to specify the warp size explicitly as `target = "vulkan -from_device=0 -thread_warp_size=32` to enable shuffle instructions.

On NV it works correctly. TODO: test on AMD and add unit tests.

On Intel the result is incorrect for some reason. In general subgroup on Intel GPU seems tricky, since unlike NV or AMD there is no well defined "subgroup (warp) size". Each instruction can execute with different subgroup sizes.  To disable shuffle instructions, we can add `-supported_subgroup_operations=0` to the target string. 

Reference:
https://www.khronos.org/blog/vulkan-subgroup-tutorial
https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_non_uniform_instructions
https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpGroupNonUniformShuffleDown


masahi:
See https://en.wikipedia.org/wiki/RDNA_(microarchitecture)#Architecture. The HW-native wavefront size of recent consumer GPUs is actually 32. But they can also work with wave64 by breaking up a wave64 instruction into two.

src/tir/transforms/lower_thread_allreduce.cc
```diff
@@ -719,12 +719,16 @@ class ThreadAllreduceBuilder final : public StmtExprMutator {
   // Also, the warp/wavefront size differs (64 on rocm, 32 on cuda and metal).
   bool IsWarpReduction(const std::vector<DataType>& types, int group_extent, int reduce_extent,
                        int contiguous_reduce_extent) {
-    if ((target_->kind->name != "cuda") && (target_->kind->name != "rocm") &&
-        (target_->kind->name != "metal")) {
+    if (target_->kind->name == "vulkan") {
+      if (target_->GetAttr<Integer>("supported_subgroup_operations") == 0) {
```
masahi, line 723: Here, ideally we should check the availability of each subgroup feature by bit mask such as `VK_SUBGROUP_FEATURE_SHUFFLE_BIT`, and `VK_SUBGROUP_FEATURE_SHUFFLE_RELATIVE_BIT`. But we cannot include a vulkan header in this file, so for simplicity I'm assuming that non-zero `supported_subgroup_operations` implies most common shuffle operations are supported.

tqchen:
@mei-ye do you have some insights into this?

PR #15441, opened 2023-07-31
[Unity] dynamo with dynamic shape

liquanfeng:
support dynamo with dynamic shape.


Hzfengsy (approved):
Lgtm. Thanks @liquanfeng

PR #15445, opened 2023-07-31
[Unity][Ops] Support for erf in relax

jwfromm:
This PR adds the `relax.op.erf` function. I added erf as a unary operator since it is arithmetic and elementwise, however, unlike other unary operators it requires some special handling. Specifically, we cast to fp32 from fp16 since it is not implemented on many devices for half precision. Let me know if you think it should go somewhere else due to this special behavior.


Hzfengsy (approved):
LGTM

masahi (approved):

Issue #15101, opened 2023-06-14
[Unity] [Tracking Issue] Heterogeneous execution for Relax


yongwww:
### This issue is to track progress for Relax Heterogenous support proposed in [here](https://discuss.tvm.apache.org/t/rfc-unity-relax-heterogeneous-execution-for-relax/14670).
- [x] P1.  Add the VDevice, update TVMScript parser and printer to support it.
- [x] P2.  Implement hint_on_device, to_vdevice, and builtin to_device.
- [x] P3. Update StructInfo Deduction and VM Codegen.
- [x] P4. Add UpdateVDevice pass.
- [x] P5. Develop RealizeVDevice pass.

cc @quic-sanirudh

csullivan:
Thank you for bringing support for heterogeneous graphs @yongwww! Shall we close this as completed?

qelk123:
Thank you for your contirbution! I am wondering are there any e2e test case I could follow?  It seems I can't specify more than one target in the `relax.build` API. How should I compile when nodes in my graph are heterogeneous to each other? (CPU node and CUDA node for example)
In `relay.vm.compile` I could offer a target dict like: `target={"cpu": tvm.target.Target("llvm"), "cuda": tvm.target.Target("cuda")}` to support multi-device compilation. What should I do when I using relax vm?

yongwww:
@csullivan will close this tracking issue in the coming days, I have a pr to be up to add e2e test cases.

yongwww:
@qelk123 great question!!! currently relax.build doesn't support multiple targets. I have a pr to enable this e2e test case, the pr is supposed to be up in the coming days (hopefully by end of this week). What we plan to do in Relax is:
- use the target information defined in vdevice list of global_infos of ir_module, then we don't need to feed a list/dict of targets to relax.build.
- modify tvm.build to lower the TIR based on the target attribute of the prim_func.
- update relax virtual machine to accept multiple devices.

qelk123:
@yongwww Thank you very much! I am currently working on adding end-to-end multi-device support for my Relax heterogeneous graph based on this issue. At the moment, it works on some simple graphs. As you mentioned, these three  aspects are also the main aspects I am focusing on during development. However, I have encountered some constraints during my experiments. One such constraint is that when I am not using the R.dataflow scope, the vdevice information does not appear to be properly propagated. Is the current design intended to support heterogeneous execution only within the dataflow scope?

yongwww:
[...]

PR #15447, opened 2023-08-01
[Unity] Multi-device support for Relax

yongwww:
This PR is specifically for the P1 & P2 of #15101.


yongwww:
cc: @tqchen @psrivas2 @junrushao @csullivan @slyubomirsky

src/relay/backend/contrib/cmsisnn/target.cc
```diff
@@ -31,13 +31,14 @@ namespace contrib {
...
 tvm::transform::Pass RelayToTIR();
+using FTVMTIRToRuntime = runtime::TypedPackedFunc<runtime::Module(IRModule, Target)>;
 runtime::Module TIRToRuntime(IRModule mod, Target target);
 
 TVM_REGISTER_TARGET_KIND("cmsis-nn", kDLCPU)
     .add_attr_option<Array<String>>("mattr")
     .add_attr_option<String>("mcpu")
     .add_attr_option<Bool>("debug_last_error")
-    .set_attr<FTVMRelayToTIR>(tvm::attr::kRelayToTIR, RelayToTIR())
+    .set_attr<tvm::transform::Pass>(tvm::attr::kRelayToTIR, RelayToTIR())
```
tqchen, line 41: these changes maybe should goto main?

tqchen: Maybe we don't need these changes

include/tvm/relax/struct_info.h
```diff
@@ -180,17 +184,20 @@ class TensorStructInfoNode : public StructInfoNode {
...
     v->Visit("dtype", &dtype);
+    v->Visit("vdevice", &vdevice);
     v->Visit("ndim", &ndim);
     v->Visit("span", &span);
   }
 
   bool SEqualReduce(const TensorStructInfoNode* other, SEqualReducer equal) const {
-    return equal(shape, other->shape) && equal(ndim, other->ndim) && equal(dtype, other->dtype);
+    return equal(shape, other->shape) && equal(ndim, other->ndim) &&
+           equal(vdevice, other->vdevice) && equal(dtype, other->dtype);
```
tqchen, line 194: This changes the json format, would be good to add a unity updater here https://github.com/apache/tvm/blob/main/python/tvm/ir/json_compact.py#L60 to upgrade the json so when we load older versions they still work

src/relay/backend/contrib/cmsisnn/target.cc
yongwww: I am planning to submit my update for the Target to the main branch. In the current Target implementation, there is a dependency on IRModule : `target.h -> target_kind.h - > transform.h - > module.h`. In this pr, we introduced Target as a member of VDevice, and VDevice could be part of the global_infos of IRModule, leading to the dependency `module.h->target.h`", resulting in circular dependencies. My changes are trying to resolve this issue by removing the dependency of IRModule on Target.

include/tvm/relax/struct_info.h
yongwww: [the updater](https://github.com/apache/tvm/pull/15447/files#diff-1a7f4665eeaa8ac3b32bb983b901e52eee4a2e38661c507e1902bd6c46ce7785R60-R72) was added in the new commit. Currently we don't have version management for Unity branch, I used `0.13` as the from_version, `0.14` as the to_version after the upgrade in the updater.

tqchen (approved):

tqchen:
Thank you @yongwww ! this is merged

PR #15451, opened 2023-08-01
[Unity]: fix error on enum

Lantianyou:


Lantianyou:
I will submit a PR fix the eslint and typescript config and fixes suggested by them to improve the DX and reliability of the current code. 

Would you mind reviewing it then? Thanks in advance.

tqchen:
Thank you @Lantianyou for the PRs , yes they are really appreciated

PR #15453, opened 2023-08-01
[Unity][NN] Allow nn.Module to generate subroutines

Lunderberg:
Prior to this commit, `nn.Module` generated the relax expression within the body of `relax.BlockBuilder.current()`.  For large models, this can result in extremely large function bodies, making it difficult to identify regions within the module.

This commit adds an option `nn.Module.define_subroutine`, which defaults to `False` (current behavior).  If set to `True`; either within a subclass, within an instance, or globally in `nn.Module`; function calls into the module will produce a subroutine representing the module's execution, and a call into that subroutine.  For example, calling a `class Linear(nn.Module)` would produce a `def linear(arg, weights)` function definition and a `module.linear(arg, weights)` function call.  To ensure correct shape propagation, a subroutine is generated for each unique set of argument shapes passed to the `nn.Module` subclass.

The PR branch contains three commits for ease of review.  The first commit extends `relax.BlockBuilder` to allow definitions to be generated while already making a function.  The second commit extends `nn.Module` to optionally generate and call subroutine functions.  The third commit adds backwards compatibility type checks to retain compatibility with [`mlc-ai`'s implementation of llama](https://github.com/mlc-ai/mlc-llm/blob/main/mlc_llm/models/llama.py) when generating subroutines.

### DIFF
--- a/python/tvm/relax/frontend/nn/core.py
+++ b/python/tvm/relax/frontend/nn/core.py
@@ -282,7 +282,7 @@ class Module(SubroutineMixin):
         ------
         Parameter - The module's parameter
         """
-        for name, param in self.named_parameters():
+        for _, param in self.named_parameters():
             yield param
 
     def state_dict(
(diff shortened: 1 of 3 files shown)


Lunderberg:
From discussion with @junrushao , extended implementation to also support the new `relax.frontend.nn.Module` API.

junrushao:
This is super cool work and really improved the UX overall! Thanks @Lunderberg and @slyubomirsky!

PR #15455, opened 2023-08-01
[Unity] Implement R.macro for Relax macros

kparzysz-quic:
In the same way as T.macro, R.macro support hygiene (via "hygienic" keyword), but unlike TIR macros, Relax macros produce expressions, and have to return a value: the last statement in R.macro must be "return".

### DIFF
--- a/python/tvm/script/ir_builder/relax/ir.py
+++ b/python/tvm/script/ir_builder/relax/ir.py
@@ -441,6 +441,19 @@ def emit_var_binding(value: VarBinding) -> Var:
     return _ffi_api.EmitVarBinding(value)  # type: ignore
 
 
+############################### SeqExpr ###############################
+
+
+def SeqExpr() -> frame.SeqExprFrame:  # pylint: disable=invalid-name
+    """Create a SeqExpr frame.
+    Returns
+    -------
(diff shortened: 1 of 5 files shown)


kparzysz-quic:
cc: @cyx-6 @yzh119 @slyubomirsky @masahi

junrushao:
CC @slyubomirsky if you’d love to review

slyubomirsky:
I'll review, sorry I didn't see the pings.

slyubomirsky:
Can macros take in Relax vars as arguments and define new Relax vars? That would be an important test case for hygienic macros. Otherwise, everything in this PR looks good to me.

junrushao:
Any updates?

kparzysz-quic:
Sorry, I was on vacation in the last two weeks.

Relax macros are technically `SeqExpr`s, so they cannot define variables that are visible outside.  The value of the macro is the last expression, which syntactically must be wrapped in a `return` expression.

kparzysz-quic:
Rebased, and added a test for using (outside of a macro) a variable defined in a macro.

junrushao (approved):
Looking good to me! @slyubomirsky would you like to take a second look?

PR #15457, opened 2023-08-01
[Unity][Fix][Op] Add groups to conv1d

LeshengJin:
- Enable groups as parameters of conv1d
- Update the output shape of conv2d.


yongwww (approved):
LGTM

PR #15322, opened 2023-07-14
[Dlight] Benchmarking Tools for Dynamic Shape PrimFuncs & Relax Function

zxybazh:
This PR introduces benchmarking tools for dynamic shape PrimFuncs and Relax. It facilitates the following functionalities:
1. Automatically benchmark PrimFunc performance with user-speicified dynamic shape sampling function and input information. E.g., `n=random.randint(50, 100)` will benchmark performance of PrimFunc with Dynamic dimension `n` between 50-100.
2. Extract self-contained PrimFunc benchmarking files from Relax Module. E.g., it can produce multiple python files for each function in a Relax Module and automatically extract dynamic shape input information from bindings.
3. Conduct Relax Function level benchmarking using the same dynamic shape sample value across the Relax Function. E.g., the same value `n` is used consistently in a Relax Function for any PrimFunction call so we can figure out what is the performance bottleneck when we pin down the value of all dynamic dimensions like `n` or `m` (Yes, there could be multiple dynamic dimensions).
4. It can automatically generate valid input even when the input is not the dynamic shape but the value of the dynamic dimension, which is `n`. This is specific to `rotatry_embedding` for now.

Co-authored by @cyx-6
Thanks to @junrushao for advices.

Example Usage
```python
def benchmark_prim_func_full_rpc():
    with LocalRPC() as rpc:
        rpc_config = ms.runner.RPCConfig(
            tracker_host=rpc.tracker_host,
            tracker_port=rpc.tracker_port,
            tracker_key=rpc.tracker_key,
            session_priority=1,
            session_timeout_sec=100,
        )
        benchmark_prim_func(
            cuda_workload,
            args=[
                ((1, "m", 4096), "float32"),
                ((4096, 4096), "float32"),
                ((1, "m", 4096), "float32"),
            ],
            dym_var_dict={"m": "int32"},
            target="nvidia/geforce-rtx-3070",
            dev=tvm.cuda(),
            rpc_config=rpc_config,
            evaluator_config=ms.runner.EvaluatorConfig(
                number=10,
                repeat=10,
                min_repeat_ms=0,
                enable_cpu_cache_flush=False,
            ),
        )
```

Expected Results for the tested PrimFunc:
```
  InputInfo   Time(us)    Std(us)  Weight  WxTime(ms)
0   m = 126  752.48000  19.417429       1    0.752480
1    m = 56  430.68955   0.244274       1    0.430690
2    m = 13  340.11350   0.241286       1    0.340114
[...]

PR #15466, opened 2023-08-03
[Dlight] Fix Customized Dynamic Shape Sampling

zxybazh:
This PR is a follow up for #15322 to better support model level workload benchmarking as in https://github.com/mlc-ai/dlight-bench/pull/6

introduces some minor fixes to enable easier dynamic shape sampling for model level workload benchmarking. For example, a dynamic shape sampling function can take `sample_idx` and `sample_num` as input to generate better selection of shapes. It also makes life easier by converting all the dynamic variable to str when parsing from PrimFuncs.

Other than that, this PR includes a few minor changes, including type annotation and supporting customizing the columns in the benchmarking output.

### DIFF
--- a/python/tvm/dlight/benchmark/bench.py
+++ b/python/tvm/dlight/benchmark/bench.py
@@ -85,11 +85,13 @@ def benchmark(
     # produce IRModule and function name
     if isinstance(mod_or_func, PrimFunc):
         func_name = "main" if func_name is None else func_name
-        mod = IRModule.from_expr(mod_or_func.with_attr("global_symbol", func_name))
     else:
-        mod = mod_or_func
         # assume only one global function
-        (func_name,) = mod.get_global_vars()
+        (func_gv,) = mod_or_func.get_global_vars()
+        func_name = func_gv.name_hint
+        mod_or_func = mod_or_func[func_gv]
+    mod = IRModule.from_expr(mod_or_func.with_attr("global_symbol", func_name))
+
     # produce input shapes
     if args is None:
         args, _ = extract_func_info_from_prim_func(mod[func_name])
@@ -130,7 +132,7 @@ def benchmark(
     # run benchmark
     if rpc_config is None:
         profile_result = rt_mod.time_evaluator(
-            func_name,
+            func_name if isinstance(func_name, str) else func_name.name_hint,
             dev=dev,
             number=evaluator_config.number,
             repeat=evaluator_config.repeat,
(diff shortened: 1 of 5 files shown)


python/tvm/dlight/benchmark/utils.py
```diff
@@ -110,6 +119,11 @@ def default_dym_var_sample_func(dym_var_dict: Dict[str, str]) -> Dict[str, int]:
     ----------
     dym_var_dict : Dict[str, str]
         Dynamic shape variable dictionary, e.g., {"n": "int32", "m": "int32"}
```
sunggg, line 121: Is there any case where we express shape values other than "int64"?

python/tvm/dlight/benchmark/utils.py
```diff
@@ -101,7 +106,11 @@ def populuate_input_shape(
     return results
 
 
-def default_dym_var_sample_func(dym_var_dict: Dict[str, str]) -> Dict[str, int]:
+def default_dym_var_sample_func(
+    dym_var_dict: Dict[str, str],
+    sample_idx: int,  # pylint: disable=unused-argument
```
sunggg, line 111: Seems like `sample_idx` is not used?

python/tvm/dlight/benchmark/utils.py
zxybazh: I don't think so, the data types I've seen are only `int32` and `int64`. Given this is part of the input shapes, it's expected to be integer so I think it also makes sense if we just remove the data type annotation and only use a set to store the variables.

python/tvm/dlight/benchmark/utils.py
zxybazh: Yes, this default sample function doesn't use `sample_idx` but it could be very useful for other use cases. For example, we may want to use `n=2^sample_idx` for the samplings.

python/tvm/dlight/benchmark/utils.py
sunggg: interesting. do you know where we use `int32`? It would be nice to standardize it if there is no specific reason behind. Then, I believe we can remove this dict and simplify.

python/tvm/dlight/benchmark/utils.py
sunggg: Ah, I see. then it would be nice to give the example in the comments for new users.

Hzfengsy:
any update?

zxybazh:
Thanks for checking in! The issues are all fixed, and demo is updated in https://github.com/mlc-ai/dlight-bench/pull/6 . Ready for review.

After this PR it should be really easy to conduct PrimFunc level benchmarking & customize dynamic shape. I'm creating a notebook to demonstrate all the use cases. Will post in the thread.

zxybazh:
The mentioned colab notebook can be found here: https://colab.research.google.com/drive/1wof8KvUnAaqoXI7HAcItlOquB4cPfiny?usp=sharing

CC: @sunggg

sunggg (approved):
LGTM, thank you @zxybazh!

python/tvm/dlight/benchmark/bench.py
```diff
@@ -199,28 +212,41 @@ def benchmark_prim_func(
...
         )
+        _, total_input_bytes = inputs_infos
         row = {
-            "InputInfo": ", ".join([f"{k} = {v}" for k, v in dym_var_sample.items()]),
+            "InputInfo": ", ".join([f"{k} = {v}" for k, v in dym_var_sample.items()])
+            if len(dym_vars) > 0
+            else "static",
             "Time(us)": median * 1e6,
             "Std(us)": std * 1e6,
+            "Memory(GB/s)": total_input_bytes / median / 1024**3,
```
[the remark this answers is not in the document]
zxybazh: Thanks for the tip! I think it makes sense to have a maximum bandwidth if the workload is memory bond. On the other hand, this is a rough calculation of throughput so it may not reflect the computation flops, we can show peak FLOPS of the hardware for comparison.

PR #15467, opened 2023-08-03
[Unity][BYOC] Integrate Flash attention v2 kernel into CUTLASS BYOC

masahi:
This PR enables using the flash v2 kernel for offloading the attention op. Flash v2 is implemented in CUTLASS, so it is reasonable to integrate it into our CUTLASS BYOC.  

The flash v2 kernel is used whenever applicable, because it is expected to be faster than the xformer kernel. The xformer repo itself is now [admitting that flash is always faster](https://github.com/facebookresearch/xformers/blob/main/xformers/ops/fmha/dispatch.py#L15-L16). 

For now, flash is disabled for causal attention case, since I haven't validated its correctness thoroughly. 

UPDATE: It seems for a single-token decoding in LLM, the flash v2 kernel with `is_causal = False` is much slower than the xformer kernel. So for LLM inference, it is recommended to always use the causal mask which will disable the flash attn kernel. The flash v2 kernel is non-trivially faster on workloads like SD. 

@cyx-6 @vinx13 @yzh119 @sunggg

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -843,4 +843,8 @@ if(USE_CUDA AND USE_CUTLASS)
   install(TARGETS fpA_intB_gemm EXPORT ${PROJECT_NAME}Targets DESTINATION lib${LIB_SUFFIX})
   target_link_libraries(tvm PRIVATE fpA_intB_gemm)
   target_link_libraries(tvm_runtime PRIVATE fpA_intB_gemm)
+
+  install(TARGETS flash_attn EXPORT ${PROJECT_NAME}Targets DESTINATION lib${LIB_SUFFIX})
+  target_link_libraries(tvm PRIVATE -Wl,--no-as-needed flash_attn)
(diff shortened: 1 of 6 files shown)


Hzfengsy:
Is it possible to push cutlass parts to the `main` branch and remain only relax part to unity branch? it would be easier for us to sync branches later

masahi:
`main` doesn't have an `attention` op, so this is unity specific.

sunggg (approved):
LGTM. Two questions tho. 
1. Is v2 always better than v1?  
2. Are we deprecating v1?

masahi:
> Are we deprecating v1?

We didn't have support for flash v1 before, so we are not deprecating it. The one we have used is so-called "xformer kernel" available in the cutlass example https://github.com/NVIDIA/cutlass/tree/main/examples/41_fused_multi_head_attention

masahi:
Hi can somebody merge this?

PR #12455, opened 2022-08-16
[Target] Replace utility functions with target.features

Mousius:
Following on from https://github.com/apache/tvm/pull/12454 this patch removes the utility functions in favour of the centralised `target.features` property.

PR #15468, opened 2023-08-03
[Relay][Strategy] Fix `arm_cpu` int8 conv2d schedule selection for 32-bit targets

lhutton1:
https://github.com/apache/tvm/pull/12455 slightly altered the behaviour when selecting an int8 conv2d schedule. Previously conditions that decide which schedule to select used `is_aarch64` which checks for the existance of `aarch64` in the target triple. However, the conditions now use `has_asimd` which is true if `aarch64` exists in the target triple OR `+neon` is used in the mattr.

Both `conv2d_NHWC_quantized_interleaved.arm_cpu` and `depthwise_conv2d_nhwc.arm_cpu` makes calls to LLVM intrinsics that require both `aarch64` and `+neon`. But in the case of the target `rasp4b`, the updated conditions result in compilation failure since the target has `+neon` but doesn't have `aarch64` in the target triple. The conditions have been updated to fix the compilation failure.

Likewise, the previous behaviour of the condition for `conv2d_nhwc_spatial_pack.arm_cpu` has been restored ensure a program with a 32-bit target can still be compiled.

Finally, we should only select the `depthwise_conv2d_nhwc_dsp.arm_cpu` schedule when a backend that understands `pragma_import_c` has been selected, i.e. "c".

For a more detailed discussion of the issue please see: https://discuss.tvm.apache.org/t/tflite-llvm-llvm-error-when-compiling-tflite-model/15411

cc @Mousius @ashutosh-arm @ekalda @neildhickey

### DIFF
--- a/python/tvm/relay/op/strategy/arm_cpu.py
+++ b/python/tvm/relay/op/strategy/arm_cpu.py
@@ -211,21 +211,23 @@ def conv2d_strategy_arm_cpu(attrs, inputs, out_type, target):
                     name="conv2d_nhwc_dsp.arm_cpu",
                 )
             elif kernel_layout == "HWIO":
+                is_aarch64 = target.features.is_aarch64
                 has_asimd = target.features.has_asimd
                 has_dot_prod = target.features.has_dotprod
+
                 if has_dot_prod and data.dtype in ["int8", "uint8"]:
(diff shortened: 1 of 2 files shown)


ekalda:
Thanks @lhutton1!

PR #15469, opened 2023-08-03
[CYTHON] Make cython compatible with 3.0

tqchen:
This PR makes the cython code compatible with 3.0 also in a way that is backward compatible with older cython versions


tqchen:
Thanks to @caojy1998 , @frozenbugs

PR #15470, opened 2023-08-03
[Relay][Strategy] Use x86 dense schedules for arm_cpu

lhutton1:
Currently the fallback used when compiling a dense operation with targets such as `llvm -device=arm_cpu` is `dense.generic`. This results in very poor performance. Although https://github.com/apache/tvm/pull/13775 meant that x86 schedules are used in cases where no strategy is provided by arm_cpu, the dense strategy is registered due to the existence of specialized schedules for arm_cpu e.g. a schedule for embedded devices. This commit ensures x86 schedules are used inplace of a generic schedule which yields much better performance.

The commit also follows the same approach for the `dense.generic` schedule as the x86 strategy. This will only be used when auto-scheduler is enabled.

A test has been added to check the intended schedules are picked when compiling with `arm_cpu`.

cc @ekalda @neildhickey

### DIFF
--- a/python/tvm/relay/op/strategy/arm_cpu.py
+++ b/python/tvm/relay/op/strategy/arm_cpu.py
@@ -559,33 +559,53 @@ def schedule_dense_arm_cpu(attrs, inputs, out_type, target):
             wrap_topi_schedule(topi.arm_cpu.schedule_dense_dsp),
             name="dense_dsp.arm_cpu",
         )
-    else:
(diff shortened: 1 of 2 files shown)


ekalda (approved):
Thanks @lhutton1, LGTM!

leandron (approved):
LGTM, thanks! Merging this now, thanks @ekalda @lhutton1!

PR #15471, opened 2023-08-03
[ARITH] Enhance Canonical Simplify for LE

Hzfengsy:
This PR enhances the canonical simplifier to support the following patterns:

`x0 * s0 + x1 * s1 + ... + xn + c < 0, let d = gcd(s0, s1, ..., s{n-1}, c)`

1. if can prove -d < xn < d, then we can simplify the expression to `x0 * (s0/d) + x1 * (s1/d) + ... + x{n-1} * (s{n-1}/d) < c/d`, e.g. `x * 8 + y < 16` where `y` \in [0, 8), we can simplify it to `x < 2`
2. if xn is in pattern of yn % m, where m % d == 0, convert it to yn // d % (m/d) e.g. `x1 * 64 + (x2 * 8 + x3) % 64 < 120`, `x3` \in [0, 8), we can simplify it to `x1 * 8 + (x2 * 8 + x3) // 8 % 8 < 15` ==> `x1 * 8 + x2 % 8 < 15`

cc @tqchen


tests/python/unittest/test_target_codegen_c_host.py
```diff
@@ -64,7 +64,8 @@ def test_add_pipeline():
     s[C].pragma(xo1, "parallel_launch_point")
     s[C].pragma(xo2, "parallel_stride_pattern")
     s[C].pragma(xo2, "parallel_barrier_when_finish")
-    s[C].vectorize(xi)
+    # FIXME(tvm-team): vector operators are not supported for codegen to C yet
```
Hzfengsy, line 67: This case is because of the poor support of vectorization for c_codegen.
Before this PR, TVM failed to simplify the branch predicate and skip the vectorization. This PR enhances the arith, and makes it TRUE vectoring, but failed on codegen stage.

In short:
1. It's not related to this PR, as it's codegen issue
2. It's not a regression, vectorized step is skipped before this PR.

tqchen (approved):

PR #15473, opened 2023-08-03
[Relay] [Bugfix] Fix some bugs of dominator pattern

kfeng123:
This PR solves Issue 1 and Issue 2 of dominator pattern as described in the [pre-RFC](https://discuss.tvm.apache.org/t/on-the-applications-of-the-composition-of-dataflow-patterns/15344)

### DIFF
--- a/src/relay/ir/dataflow_matcher.cc
+++ b/src/relay/ir/dataflow_matcher.cc
@@ -302,12 +302,12 @@ bool DFPatternMatcher::VisitDFPattern_(const CallPatternNode* op, const Expr& ex
 bool DFPatternMatcher::MatchesPath(const DominatorPatternNode* op, const Expr& expr) {
   auto call_node = expr.as<CallNode>();
   auto index_node = expr_to_node(expr);
+  size_t arg_counter{0};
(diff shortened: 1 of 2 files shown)


tqchen:
cc @ganler @masahi @yelite  would be good if you can help review. @kfeng123 can you also send a PR to unity and update dominator pattern in https://github.com/apache/tvm/blob/unity/include/tvm/relax/dataflow_pattern.h Thank you!

masahi (approved):

ganler:
> cc @ganler @masahi @yelite would be good if you can help review. @kfeng123 can you also send a PR to unity and update dominator pattern in https://github.com/apache/tvm/blob/unity/include/tvm/relax/dataflow_pattern.h Thank you!

Note that Relax pattern language does not syntactically model topological patterns such as the dominator pattern, but instead dependency constraints need to be specified to model such patterns. 

I need to take a while to follow up the pre-RFC for better understanding the intent of this PR but feel free to go ahead merging it without me. :)

kfeng123:
I am not familiar with the unity branch. I would like to learn the pattern matching design there and contribute to it. But it may take some time.

PR #15474, opened 2023-08-03
[Unity][Pass] Fixpoint simplification

slyubomirsky:
This PR implements a simple simplification pass that applies various other simplification passes until fixpoint (i.e., apply them until the module stops changing). This probably should not be part of the standard compilation, since it has the potential to repeat passes many times, but I doubt that we would be in danger of having it happen that much. I am certain that it terminates, since all of the passes included either make the module "more simplified" or leave it unchanged if they cannot find any opportunities to make changes (i.e., no pass has the potential to undo the work of other passes). Nevertheless, the pass includes an iteration limit in case passes are later added that in some situations do not converge (not recommended).

~~Incidentally, it also fixes a bug that I happened to discover in the binding canonicalization pass: It did not handle the case of a non-dataflow var being bound to a dataflow var and gave a segfault if it was encountered.~~

A few points of discussion:
1. I wasn't sure if constant folding should be included in this. I am not sure new opportunities for constant folding could arise with these passes, though maybe it's possible (if we could come up with an example).
2. There are many normalizations happening between each round but I encountered a bug when I tried to go function-by-function (StructInfo would not update correctly), so this manner of implementation ensures fewer headaches. Passes all assume that the input has been normalized so not doing it every time has the potential to cause unexpected errors.
3. Are there more simplifications we would like to throw in here? Lots of little passes like the dataflow block output folding behavior would now be a lot more feasible to just throw in here and see some use.


slyubomirsky:
Looks like a spurious/flaky test failure

slyubomirsky:
@tvm-bot rerun

kparzysz-quic:
All of this should be doable in a single execution of a sequence of optimizations.  If we don't get it. we should examine our passes for deficiencies.  Doing fixed point transformations is (IMO) a warning sign to look at what we're missing that we're trying to make up for with repeated iterations.

Even in the pattern matching pass (where repetition is justifiable) we should have a max number of iterations to perform.

slyubomirsky:
It's a tradeoff between increasing the complexity of the individual passes to handle more cases versus rerunning existing passes. I agree that running to fixpoint is a bit of a lazy "shotgun approach" but it could be useful in that it would allow for easily throwing in new passes to handle small transformations.

kparzysz-quic:
I disagree with this approach.  This is not the right way to compose transformations.

kparzysz-quic:
Sure.  I'll be out on vacation in the next 2 weeks, but I'd be happy to propose/work on improvements after I'm back.

slyubomirsky:
I'll split the bugfix into its own PR then and we can leave this open for further discussion.

slyubomirsky:
The specific motivation for this one has been superseded by later PRs like #15791 so there is no reason to keep this around.

PR #15479, opened 2023-08-04
[CMake] Use statically linked zstd

junrushao:
This PR makes it possible to actually use statically linked zstd in TVM build.


junrushao:
It's unclear to me why the statically linked zstd crashes windows.

junrushao:
Superseded by #16032

PR #15483, opened 2023-08-04
[Script] Be more careful when generating ast.ExtSlice for Subscript

kparzysz-quic:
The ast.ExtSlice expects a non-empty list, otherwise evaluation fails with `error: empty dims on ExtSlice`. Also, each element in "dims" list of ExtSlice must be either Slice or Index.

In python3.8 an expression `A[()]` is parsed (by ast) as Subscript with slice being `Index(value=Tuple(elts=[]))`. When we translate a subscript from doc.AST to ast, we unconditionally convert every tuple to ast.ExtSlice, which in this case is incorrect.

The fix is to map empty tuple back to the `Index(Tuple[]))` instead of ExtSlice. In other cases, ensure that members of ExtSlice are of correct types.

### DIFF
--- a/python/tvm/script/parser/core/doc.py
+++ b/python/tvm/script/parser/core/doc.py
@@ -414,6 +414,7 @@ def _register_subscription_handling():
                 ctx=from_doc(x.ctx),
             )
         elif isinstance(x.slice, doc.Tuple):
+
             def remap_dim(doc_item: doc.Expr) -> ast.Expr:
                 ast_item = from_doc(doc_item)
                 if isinstance(ast_item, (ast.Index, ast.Slice)):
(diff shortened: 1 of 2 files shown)


kparzysz-quic:
cc: @Hzfengsy @junrushao @yzh119

Lunderberg (approved):
Thank you for the fix, and LGTM!   It looks like this is specifically for python 3.8 and earlier, as the `ast.Index` and `ast.ExtSlice` classes were deprecated in python 3.9.

I'm rather surprised on how [this test](https://github.com/apache/tvm/blob/main/tests/python/unittest/test_tvmscript_roundtrip.py#L2843) was able to pass, as the CI runs on python3.8.16, and this test case also uses length-zero tuples as an index.  Does the bug only trigger when the object is passed to an external macro?

kparzysz-quic:
This happens when we evaluate and expression containing `()`, specifically when it get translated back into ast and then passed to `eval`.  I encountered this while using macros, and I also thought it should happen without macros too, but I wasn't able to reproduce it with a simple testcase.  I can dig into this some more to see why it worked there, but failed in a macro.

junrushao:
Interesting findings! When initiating TVMScript, we tested bunch of cases in certain python versions, but not all - which might be the source of such issues

PR #14888, opened 2023-05-19
[TEST] Enable arm_cpu qnn.conv2d i8mm test

lhutton1:
Enabling the test now we have LLVM > 11 in the docker images.

cc @ekalda @neildhickey

PR #15484, opened 2023-08-04
[Bugfix][TOPI] Fix a bug in arm_cpu int8 conv2d i8mm schedule

Anndrey24:
`topi.arm_cpu.schedule_conv2d_NHWC_quantized_interleaved` was failing compilation with the `+i8mm` extension enabled (as done in apache/tvm#14888) whenever the output height and output width were both equal to 1, such that OH x OW = 1.

Padding was being removed during the `tir.BufferShapeLegalize` pass, causing an error in the `tir.BufferBindUnwrapper` pass. Some of the removed padding was necessary for tensorize (using the `gemm_acc_2x2_int8_int8_int32` intrinsic), which expects 2x2 output tiles. However, because of the optimisations mentioned above, the output tensor `C_interleaved` was reduced to having 1x2 tiles instead.

e.g. for A = [1x1x1x8], W = [1x1x8x24], C = [1x1x1x24]:
- Before fix: `C_interleaved = T.Buffer((1, 1, 2, 1, 6, 1, 2), "int32”)`
- After fix: `C_interleaved = T.Buffer((1, 1, 2, 1, 6, 2, 2), "int32”)`

To make sure the required padding is left untouched, while the rest of it is still removed, a dummy reference to the needed axis is declared.

In the end, the leftover padding is still disregarded when computing the final output tensor `C`.

### DIFF
--- a/python/tvm/topi/arm_cpu/conv2d_gemm.py
+++ b/python/tvm/topi/arm_cpu/conv2d_gemm.py
@@ -211,18 +211,45 @@ def compute_conv2d_gemm_without_weight_transform(
                 ),
                 name="C_interleaved",
             )
+            # Ensure the padding needed for tensorize does not get removed during tir passes
+            # by adding a dummy reference to the specific padded area of the result
+            zero = (
+                tvm.tir.const(1, C_interleaved.dtype)
+                * C_interleaved[
+                    batches - 1,
+                    M // tile_rows_A,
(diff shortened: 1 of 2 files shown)


Anndrey24:
cc @ekalda @neildhickey @lhutton1 @leandron

ekalda:
Thanks @Anndrey24!

PR #15487, opened 2023-08-04
[Unity] nn.Module external modules

cyx-6:
This PR introduces the feature of importing external `*.o` modules into our `nn.Module` frontend.

### DIFF
--- a/python/tvm/relax/frontend/nn/spec.py
+++ b/python/tvm/relax/frontend/nn/spec.py
@@ -15,6 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 """Compilation specifications, for example, dynamic shape inputs."""
+from collections import defaultdict
 import inspect
 import threading
 from typing import Any, Callable, Dict, List, Sequence, Tuple, Union
@@ -307,17 +308,15 @@ class SpecBuilder:
             return result
 
         def _extern_modules() -> List[Tuple[str, List[str]]]:
-            mod2func = {}
+            mod2func = defaultdict(set)
             for _, extern_module in core._attribute_finder(
                 spec.module, "", condition_yield=lambda x: isinstance(x, core.ExternModule)
             ):
                 module_spec = extern_module.module_spec
-                if not module_spec.filename in mod2func:
-                    mod2func[module_spec.filename] = set()
                 mod2func[module_spec.filename].update(
                     [function_spec.symbol for function_spec in module_spec.functions]
                 )
-            return [(mod, list(mod2func[mod])) for mod in mod2func]
+            return [(mod, list(funcs)) for mod, funcs in mod2func.items()]
 
         # pylint: enable=protected-access
 
@@ -341,7 +340,7 @@ class SpecBuilder:
         if extern_modules:
             original_external_mods = mod.get_attr("external_mods")
             if original_external_mods is not None:
-                external_mods.extend(original_external_mods)
+                external_mods = original_external_mods + extern_modules
             mod = mod.with_attr("external_mods", external_mods)
         return mod, params
 
--- a/python/tvm/relax/frontend/nn/spec.py
+++ b/python/tvm/relax/frontend/nn/spec.py
@@ -21,6 +21,7 @@ from typing import Any, Callable, Dict, List, Sequence, Tuple, Union
 
 from tvm import tir
 from tvm.ir import IRModule
+from tvm.runtime import load_static_library
 
 from ... import expr as rx
 from ...block_builder import BlockBuilder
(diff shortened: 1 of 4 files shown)


junrushao:
I'd love to review later this week. Was meant to review Monday but got distracted for quite a while

python/tvm/relax/frontend/nn/torch.py
```diff
@@ -42,6 +42,7 @@ def __init__(  # pylint: disable=invalid-name
         spec: _spec.ModuleSpec,
         vm: VirtualMachine,
         params: List[NDArray],
+        effects: List[Any] = None,
```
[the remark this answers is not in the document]
cyx-6: yes

python/tvm/relax/frontend/nn/spec.py
```diff
@@ -268,10 +306,24 @@ def _effects() -> List[Tuple[str, core.Effect]]:
...
             return result
 
+        def _extern_modules() -> List[Tuple[str, List[str]]]:
+            mod2func = {}
+            for _, extern_module in core._attribute_finder(
+                spec.module, "", condition_yield=lambda x: isinstance(x, core.ExternModule)
+            ):
+                module_spec = extern_module.module_spec
+                if not module_spec.filename in mod2func:
+                    mod2func[module_spec.filename] = set()
```
junrushao, line 316: using

```
from collections import defaultdict

mod2func = defaultdict(set)
```

to avoid checking if `mod2func` contains `module_spec.filename`.

python/tvm/relax/frontend/nn/spec.py
```diff
@@ -268,10 +306,24 @@ def _effects() -> List[Tuple[str, core.Effect]]:
...
+                mod2func[module_spec.filename].update(
+                    [function_spec.symbol for function_spec in module_spec.functions]
+                )
+            return [(mod, list(mod2func[mod])) for mod in mod2func]
```
junrushao, line 320: always prefer `[f(k, v) for k, v in dict.items()]` to `[f(k, dict[k]) for k in dict]`

python/tvm/relax/frontend/nn/spec.py
```diff
@@ -282,7 +334,16 @@ def _effects() -> List[Tuple[str, core.Effect]]:
...
                     self.builder.emit_func_output(outputs, inputs)
-        return self.builder.get(), params
+        external_mods = []
+        for lib_path, func_names in extern_modules:
+            external_mods.append(load_static_library(path=lib_path, func_names=func_names))
+        mod = self.builder.get()
+        if extern_modules:
+            original_external_mods = mod.get_attr("external_mods")
+            if original_external_mods is not None:
+                external_mods.extend(original_external_mods)
```
junrushao, line 344: let's follow the ordering and put the original ones in the beginning.

```
external_mods = list(original_external_mods) + external_mods
```

tests/python/relax/test_frontend_nn_extern_module.py
```diff
@@ -0,0 +1,122 @@
...
+
+    Module = get_relax_matmul_module(
+        shape_converter(shape_a, var_table),
+        shape_converter(shape_b, var_table),
+        dtype,
+    )
+
+    mod = partition_for_cutlass(Module, True)
+    codegen_pass = relax.transform.RunCodegen({"cutlass": {"sm": 80, "find_first_valid": True}})
+    mod = codegen_pass(mod)
```
junrushao, line 71: iiuc this part is only used for generating `./tmp/cutlass.o` - is that correct?

if so, is it possible to:
1) use a much simpler and more unit-testable source code? for example:

```C++
void Add(NDArray a, NDArray b, NDArray c) {
  ICHECK(a->ndim == 0 && a->dtype == DataType::Float(32));
  ICHECK(b->ndim == 0 && b->dtype == DataType::Float(32));
  ICHECK(c->ndim == 0 && c->dtype == DataType::Float(32));
  *(reinterpret_cast<double*>(c->data)) = *(reinterpret_cast<double*>(a->data)) + *(reinterpret_cast<double*>(b->data));
}
```

2) isolate this logic to a separate method, say, "def generate_test_module(path: str)"

junrushao:
Otherwise LGTM

junrushao (approved):

Cydia2018:
Thank you very much for your work. 

I would like to ask if the function of external modules has not been fully developed yet? I tried to embed a handwritten attention operator and it compiled normally, but an error was reported when running: Cannot find PackedFunc attention in either Relax VM kernel library.

Looking forward to your reply!

cyx-6:
@Cydia2018 Thanks for reaching out! Have you ever wrapped your operator with `TVM_DLL_EXPORT_TYPED_FUNC` or other TVM registration functor? Feel free to create a new issue with more context. And it would be great if you could provide a reproducible script for us. :)

Issue #15233, opened 2023-07-05
[Unity][MSC][Tracking Issue] Introduction to Multi-System Compiler


Archermmt:
- [ ] [M0] Build MSCGraph core parts. Enable translation between Relay, Relax and MSCGraph without lossing information.
- [ ] [M1] Finish RuntimeManager for relax, and torch, so that a compiling process can be test based on MSCGraph.
- [ ] [M2] Use MSCProcessor to manage the compiling pipeline. Add pruner, prune the graph by given density.
- [ ] [M3] Add MSCGym, enable auto pruning process. Add distiller, enable knowledge distilliation for pruning.
- [ ] [M4] Add quantizer and collecter for quantization and debugging.
- [ ] [M5] [Optional] Add MSCWrapper as compression toolchain. MSCWrappers are wrapper format of MSCProcessor, which can be used to wrap and optimize the models in training system.
- [ ] [M6] [Optional] Add MSC_PIPE for Saas. Build a local service and client for test the Saas

Archermmt:
Intorduction @ https://discuss.tvm.apache.org/t/rfc-unity-msc-introduction-to-multi-system-compiler/15251

Archermmt:
TODO: add tests for M0.2 after M0.3

Archermmt:
Discussion on translate relay to relax without loss info: https://discuss.tvm.apache.org/t/msc-translate-relay-to-relax-without-loss-info/15650

Lunderberg:
I'm somewhat concerned about the `relay -> python codegen -> relax` code path used in `tvm.contrib.msc.framework.torch.frontend.translate.from_torch` when `via_relax=False`.  This is a duplication of the serialization/parsing used in TVMScript (`tvm.script`), and can cause CI failures (e.g https://github.com/apache/tvm/pull/15783) due to this duplication.

While I agree with the need for a operator-level conversion from relay to relax, I think it should be done through extending the existing `relax.testing.relay_translator.from_relay` converter rather than having an additional python code-generator.

Archermmt:
@Lunderberg  sorry for the late reply.... I've checkout the failures, which is cause by from_fx function of relax. The failures of test_translate_relax.py and test_translate_torch.py have the same reason. I think MSC is not the source of the failures, it just build new modules based on relax and relay, without touching the basic codes of them.

And the reasons way build a duplicate "relay -> relax" method:
[...]

PR #15489, opened 2023-08-05
[RFC][Unity][MSC] MileStone 0

Archermmt:
RFC @ https://discuss.tvm.apache.org/t/rfc-unity-msc-introduction-to-multi-system-compiler/15251
Tacking issue @ https://github.com/apache/tvm/issues/15233

This is the M0 of MSC(Multi-System Compiler): Build MSCGraph core parts. Enable translation between Relay, Relax and MSCGraph without loss information.

Core parts:
1. MSCGraph: A DAG structure for store info from relax and relay. A MSCGraph can be build from relax or relay function
2. Codegen: A printer to generate target codes. A MSCGraph can be translated to relax

This milestone add examples on translating relay to relax without loss information. This solution can be used to use relay-based features in relax, like quantization, pruning and some optimization passes.

Both python and cpp codes are managed in a new module, so there are no change to the current code base. To use MSC, just set the USE_MSC in config.cmake to ON.

### DIFF
--- a/src/tir/transforms/thread_storage_sync.cc
+++ b/src/tir/transforms/thread_storage_sync.cc
@@ -50,11 +50,27 @@ class ThreadSyncPlanner : public StorageAccessVisitor {
   }
   // Plan the sync
   std::vector<AccessEntry> Summarize(std::vector<StmtEntry> seq, const ForNode* loop) final {
+    // Redirect all "shared.dyn" buffer access to the same buffer var
+    // so that the accesses can be planned together.
+    Var shared_dyn_buf;
+    for (StmtEntry& entry : seq) {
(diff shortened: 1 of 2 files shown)


junrushao:
Well I mean this is awesome to have MSC, but is there any way we could potentially split the PR to certain reviewable size?

Archermmt:
> Well I mean this is awesome to have MSC, but is there any way we could potentially split the PR to certain reviewable size?

let me think of that. Maybe split it into some small parts as following:
M0.5: Test (tests/python/contrib/test_msc && related helper modules in python)

Each step is functional isolated and play as pre-request of next step. I can try add some test after each part.

PR #15491, opened 2023-08-06
[Unity][BYOC] Unify the interface between `FuseOpsByPattern(.., annotate_codegen=True)`and `MergeCompositeFunctions()`

sunggg:
Currently, we have two paths for BYOC.

Path1: `seq = [FuseOpsByPattern(patterns, annotate_codegen=False), MergeCompositeFunctions(), RunCodegen()]`
Path2: `seq = [FuseOpsByPattern(patterns, annotate_codegen=True), RunCodegen()]`

Path1 is used for inference engines such as TensorRT, DNNL, CoreML, etc. which are more efficient with larger subgraph and thus better off merge adjacent composite functions. In constrast, Path1 is used for kernel libraries, such as Cutlass, Cublas, and not need to merge. 

For better UX, this PR aims to make both paths have same interface with various BYOC codegens. 
In addition, it removes unnecessary `kPrimitive` attribute in the function that will be offloaded to external codegen.
One remaining caveat is that two paths would have different function names of each composite function and thus structural equality check would fail between two paths.  Instead, this PR compares each path to their expected forms which have the same interface except function names.  

This PR is P3 of CoreML tracker: https://github.com/apache/tvm/issues/15245

cc. @masahi

### DIFF
--- a/include/tvm/relax/utils.h
+++ b/include/tvm/relax/utils.h
@@ -105,6 +105,11 @@ TVM_DLL bool IsImpureCall(const Call& call);
  */
 TVM_DLL Function CopyWithNewVars(Function func);
 
+/*!
+ * \brief Transform all dataflow structure to non-dataflow version.
+ */
+Expr ToNonDataflow(const Expr& e);
+
 }  // namespace relax
 }  // namespace tvm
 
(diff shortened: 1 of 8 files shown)


masahi:
By "interface difference", do you mean the unnecessary, "outer" `with R.dataflow()` generated by `MergeCompositeFunctions`?

sunggg:
@masahi yes. and the behavior for redundant bindings of dataflow vars was different and it had addressed by PR https://github.com/apache/tvm/pull/15341

masahi (approved):

PR #15492, opened 2023-08-06
[TVMScript] Allow use of Python builtins in script

kparzysz-quic:
The builtins are already supported by `eval` (they are automatically injected in the global scope), but they are not recognized by the evaluator's checks. When the evaluator sees doc.Name, it looks it up in the current `var_table`, and flags an error if it's not there.

Make the evaluator also consult the current builtins before erroring out.


kparzysz-quic:
cc: @junrushao @Hzfengsy @yzh119

Hzfengsy (approved):

junrushao (approved):
LGTM! Nice addition

PR #15493, opened 2023-08-06
[TIR] Output DeclBuffer in SplitHostDevice

Lunderberg:
If the generated device function uses a buffer, generate a DeclBuffer for the buffer at the top of the device function.

This is a subset of the changes made in https://github.com/apache/tvm/pull/14778, broken out for ease of testing and review.


junrushao (approved):
The PR looks great to me!

PR #15495, opened 2023-08-06
[Unity] Recursive pattern match in rewrite_call

Lunderberg:
Prior to this commit, pattern matches were evaluated a single time, in the order of the block bindings.  If a pattern rewrite produced a node which would also match the pattern (e.g. recursive reordering), the pattern would still only be evaluated once.

This commit updates `PatternMatcher` to iterate until convergence when provided with a pattern, analogous to its current behavior when called through `rewrite_bindings`.


kparzysz-quic (approved):
We should probably add some guards against looping forever eventually in the fixed point rewriting.

PR #15498, opened 2023-08-07
[Unity][ONNX] Support ONNX Symbolic Shape Deduction

Hzfengsy:
This PR improves ONNX importer to support symbolic shape importing. With this PR, we can import BERT with symbolic shapes into Relax.

cc @jwfromm


Hzfengsy:
ping @tqchen

jwfromm:
Awesome work, thank you @Hzfengsy!

PR #15500, opened 2023-08-07
[Unity][MSC][M0.1] Enable set name and layout for exprs

Archermmt:

This is the first part of Milestone 0: Build MSCGraph core parts. Enable translation between Relay, Relax and MSCGraph without loss information.

To limit each PR in reviewable size, the Milestone 0 will be split into 5 steps:
M0.5: Translation test (relax/relay test && related helper modules in python)

This is the M0.1 step. Passes are added to set name and layouts via Span. 
The names are essential for developing optimization algorithms, especially when developer want to control the optimization tensor by tensor. This always happens in developing of advanced quantization.
Layout is essential for pruning process, where channel relation between weights are built. And building channel relationship need layout of tensors to be set. Some quantization algorithm also need layout for quantize weights per channel or per token.

These attributes need to be set to Call, Constant as well as Var. So the best way is to find a common attributes holder for them all. In the end I chose span as the attributes holder, and special xml format like <key>value</kay> is used to mark key:value pair in span. 

cc @Hzfengsy


python/tvm/contrib/msc/core/transform/pattern.py
```diff
@@ -0,0 +1,490 @@
...
+import tvm
+from tvm.relax.dpl import pattern as relax_pattern
+from tvm.relay import dataflow_pattern as relay_pattern
+
+from tvm.relax.transform import PatternCheckContext
+from tvm.relax.backend.pattern_registry import register_patterns
+from tvm.relay.op.contrib.register import register_pattern_table
+
+
+def make_relax_conv_bias_pattern(op_name):
```
Hzfengsy, line 29: Please add type annotation if possible.
```suggestion
def make_relax_conv_bias_pattern(op_name: str) -> Tuple[bool, Dict[str, Pattern]:
```

python/tvm/contrib/msc/core/transform/pattern.py
```diff
@@ -0,0 +1,490 @@
...
+    """A simple utility to create patterns for an operation fused with bias.
+
+    Parameters
+    ----------
+    op_name: str
+        The name of a Relax op, such as "relax.nn.conv2d"
+
+    Returns
+    -------
+    pattern: DFPattern
```
Hzfengsy, line 39: Looks like it's not correct

python/tvm/contrib/msc/core/transform/pattern.py
```diff
@@ -0,0 +1,490 @@
...
+
+def _check_relax_conv_bias(context: PatternCheckContext) -> bool:
+    """Check if conv_bias fuse pattern is correct."""
+    bias = context.annotated_expr["bias"]
+    reshape = context.annotated_expr["reshape"]
+    non_one_dims = len([i for i in reshape.struct_info.shape.values if i > 1])
+    return non_one_dims <= 1 and bias.struct_info.ndim == 1
+
+
+def make_relax_linear_pattern():
```
Hzfengsy, line 62: Type annotation

python/tvm/contrib/msc/core/transform/pattern.py
```diff
@@ -0,0 +1,490 @@
...
+
+
+def _check_relax_linear(context: PatternCheckContext) -> bool:
+    """Check if linear pattern is correct."""
+    weight = context.annotated_expr["weight"]
+    permute = context.annotated_expr["permute"]
+    return weight.struct_info.ndim == 2 and not permute.attrs["axes"]
+
+
+def make_relax_linear_bias_pattern():
```
Hzfengsy, line 86: +1

Hzfengsy (commented):
Generally LGTM. One key problem is that how do we test MSC on CI?

Archermmt:
> Generally LGTM. One key problem is that how do we test MSC on CI?

I add some tests under tests/python/contrib/test_msc. Tests in this folder will cover most of the modules and can be used as test in CI. However, I plan to add these tests to CI after Milestone4, when end to end demos are added and most features are stable.
After the e2e demo, I will discuss with the unity on whether MSC should be used as infrastructure for model optimization and how it be further developed. If the design of MSC helps in model optimization and will be developed by other developers, CI tests are necessary. Before that, design and modules may be modified and reconstructed, tests may also be changed accordingly. So I prefer run the tests manually before PR, until the most features are stable.

Hzfengsy:
Thanks @Archermmt. To be more specific, this PR does not change the CI script, and the CI will not turn on MSC. I wonder if we can turn it on (not sure if there are dependency issues). Also, please confirm that the folder `tests/python/contrib/test_msc` is running in the `unity` CI

Archermmt:
Seems ok. Should I add the test in tests/scripts/ci.py(together with task_config_build_msc.sh && task_python_msc.sh)? I've tested with "python tests/scripts/ci.py cpu" as suggested in https://tvm.apache.org/docs/contribute/pull_request.html#guidelines

Hzfengsy:
`unity` branch does not follow that way. Please add it into file `tests/scripts/unity/task_python_relax.sh`

Hzfengsy:
If you know how to add contrib module in CI under `unity` branch? @tqchen.

yongwww:
@Archermmt looks you might need to update these two files (`echo set\(USE_MSC ON\) >> config.cmake` in `tests/scripts/task_config_build_cpu.sh` and `task_config_build_gpu.sh`) to enable the MSC build on the CI machines for Unity

Archermmt:
Thanks, I'll try open USE_MSC in build_gpu and add the tests to scripts/unity. After Milestone2, tests on TensorRT will be added, so gpu is needed.

ci/jenkins/unity_jenkinsfile.groovy
```diff
@@ -318,6 +318,7 @@ stage('Build and Test') {
           sh "${docker_run}  ${ci_gpu} ./tests/scripts/task_config_build_gpu.sh build"
           make("${ci_gpu}", 'build', '-j2')
           sh "${docker_run} ${ci_gpu} ./tests/scripts/unity/task_python_relax_gpuonly.sh"
+          sh "${docker_run} ${ci_gpu} ./tests/scripts/unity/task_python_msc.sh"
```
yongwww, line 321: It appears that the test cases for your MSC were not executed as indicated by the following link: https://ci.tlcpack.ai/blue/organizations/jenkins/tvm-unity/detail/PR-15500/7/pipeline/60/. The changes to the Jenkins file will only take effect once they are merged. To run your test cases on CI machines for this pull request, you might consider adding them to the test_python_relax.py or test_python_relax_gpu.py files.

Archermmt: I add test_msc to task_python_relax.sh. Seems ok: https://ci.tlcpack.ai/blue/organizations/jenkins/tvm-unity/detail/PR-15500/8/pipeline/60

yongwww: yep, the tests work in CI.

Archermmt:
All tests passed. Please check if there are anything else I need to change before the PR get approved. Thanks!
cc @Hzfengsy

Hzfengsy (approved):

Hzfengsy:
Thanks @Archermmt for the awesome work and reviews from @yongwww

PR #15503, opened 2023-08-08
[Unity][Dlight] Matmul tensorization for SM70

MasterJH5574:
Prior to this PR, TensorCore tensorization is not enabled for SM70 while SM70 devices have TensorCore support. This PR enables the tensorization.


MasterJH5574:
cc @Hzfengsy @adfwer233

Hzfengsy (approved):

PR #15506, opened 2023-08-08
[Relay][Strategy] Use x86 pool schedules for arm_cpu

lhutton1:
Similar to https://github.com/apache/tvm/pull/15470, x86 schedules are used in place of generic schedules to improve performance.

Since the pooling strategy does not use `OpStrategy`, mocking is used to ensure the relevant `schedule_pool` function is called when lowing a Relay pooling operation with respect to a given target.

cc @ekalda @ashutosh-arm @neildhickey

### DIFF
--- a/python/tvm/relay/op/strategy/arm_cpu.py
+++ b/python/tvm/relay/op/strategy/arm_cpu.py
@@ -74,8 +74,7 @@ def schedule_pool_arm_cpu(attrs, outs, target):
             and layout in ("NWC", "NHWC")
         ):
             return topi.arm_cpu.schedule_pool(outs, layout)
-        logger.warning("pool is not optimized for arm cpu.")
-        return topi.generic.schedule_pool(outs, layout)
+        return topi.x86.schedule_pool(outs, layout)
 
 
 def _get_padding_width(padding):
--- a/tests/python/relay/strategy/test_select_implementation.py
+++ b/tests/python/relay/strategy/test_select_implementation.py
@@ -19,12 +19,14 @@
 
 import pytest
 import numpy as np
+from unittest.mock import MagicMock
 
 import tvm
 from tvm import relay
 from tvm import te
 from tvm.relay.testing import run_infer_type
 import tvm.testing
+from tvm import topi
 
 
 @pytest.mark.parametrize(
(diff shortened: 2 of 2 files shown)


python/tvm/relay/op/strategy/arm_cpu.py
```diff
@@ -74,8 +74,7 @@ def schedule_pool_arm_cpu(attrs, outs, target):
             and layout in ("NWC", "NHWC")
         ):
             return topi.arm_cpu.schedule_pool(outs, layout)
-        logger.warning("pool is not optimized for arm cpu.")
-        return topi.generic.schedule_pool(outs, layout)
+        return topi.x86.schedule_pool(outs, layout)
```
ashutosh-arm, line 77: question: should we still wrap this under an `arm_cpu` schedule inside topi? In the future, if needed, the API can then further be optimized for the specific device.

tests/python/relay/strategy/test_select_implementation.py
```diff
@@ -187,5 +189,38 @@ def test_dense(target, expected_valid_impl, expected_impl):
...
+    out = tvm.IRModule.from_expr(out)
+
+    # Since pool does not use OpStrategy to determine the relevant schedule,
+    # we cannot simply check the schedule name that was selected with
+    # `select_implementation`. With this implementation of schedule selection,
+    # "pool.arm_cpu" will always be the schedule name, regardless of what schedule
+    # was selected. Instead, this test checks that the relevant schedule function
+    # is called when building the module.
+    mock_schedule = MagicMock()
+    mock_schedule.side_effect = lambda outs, layout: topi.generic.schedule_pool(outs, layout)
```
ashutosh-arm, line 216: question: `side_effect` is new to me. Does this work in the same way as described in the following places?
https://docs.pytest.org/en/7.1.x/how-to/monkeypatch.html#monkeypatching-functions 
https://docs.python.org/3/library/unittest.mock-examples.html#side-effect-functions-and-iterables 

Would it better to have a mock `schedule_pool()` within the test and check the value returned by it / prints done by it? This will make sure that the replacement actually happened only when x86 schedule was invoked.

ashutosh-arm (commented):
This looks like a step in the right direction. A few minor comments/questions below.

tests/python/relay/strategy/test_select_implementation.py
lhutton1: Yes essentially it was setting a return value for the mocked function.

The return type of `schedule_pool` is a schedule. I took a look at the object but it doesn't really offer anything indicative of an x86 schedule having been used. Also, I don't believe `schedule` can be equality checked very easily. Therefore, I think the simplest way is to check whether the expected function was called.

You did give me an idea that the test could be simplified a little. Rather than check from `relay.build` we can simply check from `schedule_pool_arm_cpu`. This will remove the side-effect nastiness.

ashutosh-arm (approved):
LGTM - thanks @lhutton1.

lhutton1:
Thanks @ashutosh-arm!

PR #15507, opened 2023-08-08
[Unity] Remove unused local function definitions

Lunderberg:
Prior to this commit, the `relax.analysis.remove_all_unused` utility only removed variable bindings that occur within a dataflow block. This satisfies that the condition that a variable can be removed if is unused and if its computation has no side effects.  However, it doesn't include any values from a `BindingBlock`.

This commit updates the `RemoveUnusedVars` pass to also inspect `BindingBlock` instances for removeable variables.  Currently, this is limited to unused local function definitions.

### DIFF
--- a/tests/python/relax/test_analysis.py
+++ b/tests/python/relax/test_analysis.py
@@ -117,6 +117,53 @@ def test_binding_block_remove_all_unused():
     tvm.ir.assert_structural_equal(optimized, GroundTruth["main"])
 
 
+def test_binding_block_remove_all_unused_without_dataflow():
+    @tvm.script.ir_module
+    class IdentityUnused:
+        @R.function
+        def main(x: R.Tensor((32, 32), "float32")) -> R.Tensor:
+            lv0 = x
+            unused0 = R.call_dps_packed("my_sigmoid", (x,), R.Tensor((32, 32), dtype="float32"))
+            unused1 = R.call_dps_packed(
+                "my_dps_func", (unused0,), R.Tensor((32, 32), dtype="float32")
+            )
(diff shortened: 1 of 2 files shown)


tests/python/relax/test_analysis.py
```diff
@@ -117,6 +117,53 @@ def main(x: R.Tensor((32, 32), "float32")) -> R.Tensor:
...
+            unused0 = R.call_dps_packed("my_sigmoid", (x,), R.Tensor((32, 32), dtype="float32"))
+            unused1 = R.call_dps_packed(
+                "my_dps_func", (unused0,), R.Tensor((32, 32), dtype="float32")
+            )
+            z = R.call_packed("vm.builtin.copy", lv0, sinfo_args=(R.Tensor((32, 32), "float32")))
+            return z
+
+    optimized = remove_all_unused(IdentityUnused["main"])
+
+    GroundTruth = IdentityUnused
```
kparzysz-quic, line 135: Shouldn't the two `unused`s be removed?  I think this whole algorithm (not your change specifically) is somewhat deficient if it doesn't keep track of newly-unused objects.

Lunderberg: For this specific change, because (to my understanding), the `call_dps_packed` doesn't guarantee that it has no side effects.  Within a `R.dataflow()` environment, the expression may not have side effects, so it is safe to remove.

For newly-unused objects, I believe that is handled by the `GetUnusedVars` function, called during construction.  It iteratively walks from the function outputs to the predecessors required to generate those outputs, finding the set of all variables required to generate the output.  Anything not within that set is considered unused and can be removed, even it occurs within another expression (because that expression is also going to be removed).

kparzysz-quic (approved):

PR #15509, opened 2023-08-08
[Unity] Implement relax.Function.bind_symbolic_vars

Lunderberg:
If a function has dynamic shape parameters, it can be useful to replace them with static parameters (e.g. when producing several models within the same family).  This commit introduces a utility function `relax.Function.bind_symbolic_vars`, which allows symbolic variables to be replaced with static values.

This is a related to the parameter binding done in `relax.transform.BindParam`, but does not require the bound parameter to be fully static data array.

### DIFF
--- a/src/relax/analysis/struct_info_analysis.cc
+++ b/src/relax/analysis/struct_info_analysis.cc
@@ -963,26 +963,34 @@ class SymbolicVarCollector : public relax::ExprVisitor,
   using tir::ExprVisitor::VisitExpr;
   using tir::ExprVisitor::VisitExpr_;
 
-  // Possible mode of visitor
-  enum class VisitMode {
-    /*! \brief Check all vars are well-defined. */
-    kDefault = 0,
-    /*! \brief Match define the vars on first occurrence. */
-    kMatchVarDef = 1,
-    /*! \brief Define any variables, do not visit subexpressions */
-    kProvideDefinitions = 2,
+  // Possible mode of visitor, used as bit-flags
+  enum VisitMode {
+    /*! \brief Do nothing on encountering a symbolic variable */
+    kNone = 0,
+
+    /*! \brief Provide a variable definition on first occurrence.
+     *
+     * If a symbolic variable occurs at a site where a definition can
+     * be provided, mark the variable as having a definition.
+     */
+    kProvideDefinition = 1,
+
+    /*! \brief Require a variable definition on occurrence.
+     *
+     * If a symbolic variable occurs, and has not previously been
+     * defined, mark the variable as being free/undefined.
+     */
+    kRequireDefinition = 2,
   };
 
   void VisitExpr_(const FunctionNode* op) final {
-    WithMode(VisitMode::kProvideDefinitions, [&]() {
-      ICHECK(mode_ == VisitMode::kProvideDefinitions);
+    WithMode(VisitMode::kProvideDefinition, [&]() {
       for (Var param : op->params) {
         relax::StructInfoVisitor::VisitStructInfo(GetStructInfo(param));
       }
     });
 
-    WithMode(VisitMode::kMatchVarDef, [&]() {
-      ICHECK(mode_ == VisitMode::kMatchVarDef);
+    WithMode(VisitMode::kRequireDefinition, [&]() {
       for (Var param : op->params) {
         relax::StructInfoVisitor::VisitStructInfo(GetStructInfo(param));
       }
@@ -992,7 +1000,8 @@ class SymbolicVarCollector : public relax::ExprVisitor,
   }
 
   void VisitBinding_(const MatchCastNode* binding) final {
-    WithMode(VisitMode::kMatchVarDef, [&]() { this->VisitStructInfo(binding->struct_info); });
+    WithMode(VisitMode(VisitMode::kProvideDefinition | VisitMode::kRequireDefinition),
+             [&]() { this->VisitStructInfo(binding->struct_info); });
 
     relax::ExprVisitor::VisitBinding_(binding);
   }
@@ -1007,15 +1016,13 @@ class SymbolicVarCollector : public relax::ExprVisitor,
       // time to collect usages.  Otherwise, a symbolic variable
       // defined by a later parameter may be treated as undefined when
       // used by an earlier parameter.
-      WithMode(VisitMode::kProvideDefinitions, [&]() {
-        ICHECK(mode_ == VisitMode::kProvideDefinitions);
+      WithMode(VisitMode::kProvideDefinition, [&]() {
         for (StructInfo param : op->params.value()) {
           this->VisitStructInfo(param);
         }
       });
 
-      WithMode(VisitMode::kMatchVarDef, [&]() {
-        ICHECK(mode_ == VisitMode::kMatchVarDef);
+      WithMode(VisitMode::kRequireDefinition, [&]() {
         for (StructInfo param : op->params.value()) {
           this->VisitStructInfo(param);
         }
@@ -1034,12 +1041,12 @@ class SymbolicVarCollector : public relax::ExprVisitor,
   }
 
   void VisitStructInfoExprField(const PrimExpr& expr) final {
-    if (mode_ >= VisitMode::kMatchVarDef) {
+    if (mode_ & VisitMode::kProvideDefinition) {
       if (auto var = expr.as<tir::Var>()) {
         defined_symbolic_var_.insert(var.value());
       }
     }
-    if (mode_ <= VisitMode::kMatchVarDef) {
+    if (mode_ & VisitMode::kRequireDefinition) {
       tir::ExprVisitor::VisitExpr(expr);
     }
   }
@@ -1061,7 +1068,7 @@ class SymbolicVarCollector : public relax::ExprVisitor,
   }
 
   /*! \brief The current visit mode. */
-  VisitMode mode_ = VisitMode::kDefault;
+  VisitMode mode_ = VisitMode::kRequireDefinition;
   /*! \brief The set of defined symbolic vars. */
   std::unordered_set<tir::Var, ObjectPtrHash, ObjectPtrEqual> defined_symbolic_var_;
   /*! \brief The set of free/undefined symbolic vars. */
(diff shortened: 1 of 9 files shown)


Lunderberg:
Ooh, thank you for pointing it out!  That would explain why I didn't see the functionality anywhere in the `unity` branch itself.

Looking through it, I think the main differences are:

* This PR accepts either `tir.Var` (less ambiguous, easier to generate programmatically), or strings (more ambiguous, much easier to hand-write), where #15246 only accepts strings.

* This PR operates on a relax function, where #15246 operates on an IRModule.

* This PR would ignore functions that are defined within a local function definition, where #15246 would update their symbolic variables.

* This PR would replace any TIR usage supported by `tir::ExprMutator`, where #15246 checks for specific TIR nodes.

* This PR raises an error if no symbolic variable is known by that name, where #15246 would silently ignore it.

Thoughts on making a merged PR with the best of both implementations:

* Providing methods to update either a IRModule transform or a relax Function would provide the most flexibility.
* Implementing a IRModule transform on top of a relax Function update would be require constructing fewer intermediates than implementing a relax Function update on top of an IRModule transform.

Any preference on which of the two PRs should be updated overall? Since both provide similar functionality, I don't have a huge preference.

sunggg:
Thank you for the great summary! 
In fact, I like your PR in the sense that it offers broader supports for TIR and it accepts both strings and `tir.Var`.
Since I think a pass would be convenient when we are dealing with IRModule with multiple functions inside, how about creating a function pass that wraps your utility function? I agree that this would be easier path to merge our PRs.

Lunderberg:
@sunggg How does the IRModule pass in the latest commit look?  

* Swapped order of `binding_map` and `func_name`, to allow a default value of `func_name = None`.  If no function name is specified, all functions within the module with be updated.
* Pulled in the error checking from the `relax.Function.bind_symbolic_vars` for the `IRModule` pass.  If a single function is updated, each value in the `binding_map` must be used in that function.  If multiple functions are updated, each values in the `binding_map` must occur as a replacement in at least one function.
* Allows replacement either by TIR variable or by name.  If multiple functions are updated, a replacement by name may result in updates to a different TIR variable in each function, so long as it uniquely maps to a TIR variable within a given function.  (e.g. Replacing `"batch_size"` in all functions in a module.)

The error checking was why I ended up going with a module pass instead of just wrapping the utility with a function pass.  Since the string names would be very convenient to hand-write, they're also the greatest risk of a user typo, which I wanted to guard against.  Having a function pass would require either disabling the "all replacements are used", or restricting the updates to a single function.

tests/python/relax/test_bind_symbolic_vars.py
```diff
@@ -0,0 +1,205 @@
...
+        symbolic_var_map = {M: 128, K: 64, N: 32}
+    else:
+        symbolic_var_map = {"M": 128, "K": 64, "N": 32}
+
+    after = before.bind_symbolic_vars(symbolic_var_map)
+    tvm.ir.assert_structural_equal(expected, after)
+
+
+def test_error_with_duplicate_var_names():
+    """Duplicate variable names may not be replaced by string
```
sunggg, line 55: I'm surprised to see this case passing the well_formed check. I think this should be filtered before any pass is applied, because this seems incorrect to me.

sunggg (approved):
Overall, LGTM. Thank you for merging with my PR and making it better!
One comment that is not directly related to this PR.

tests/python/relax/test_bind_symbolic_vars.py
Lunderberg: On the C++ side, both TIR variables and Relax variables have reference identity, and duplicate names can exist within the same function.  (Technically, the variables only have a `name_hint`, not a name, which I believe is to avoid the assumption of unique names.  (Though, double parenthetically, some of the python APIs expose `name_hint` as `name`.))  TVMScript has a tighter restriction, where variables tracked by the parser may not repeat within the same scope.  In this test, I define the variables with the same TIR variable name outside of TVMScript, to produce the ambiguous test case anyways.

As a bit of fun, the function can be inspected, and even though TVMScript modifies the variable names to avoid printing duplicate variables, the underlying objects still have duplicate names.

```python
M = tir.Var("M", "int64")
N1 = tir.Var("N", "int64")
N2 = tir.Var("N", "int64")

dtype = "float32"


@R.function
def func(
    A: R.Tensor([M, N2], dtype), B: R.Tensor([N1, N2], dtype)
) -> R.Tensor([M, N1], dtype):
    A: R.Tensor([M, N2], dtype) = A
    B: R.Tensor([N1, N2], dtype) = B
    out = R.linear(A, B)
    return out


# Prints  B: R.Tensor(("N", "N_1"), dtype="float32")
func.show()
# But the true name can still be extracted
assert func.params[1].struct_info.shape[0].name == "N"
assert func.params[1].struct_info.shape[1].name == "N"
# And the function is well-formed, despite the duplicate variable
# names.
assert tvm.relax.analysis.well_formed(tvm.IRModule.from_expr(func))
```

sunggg: Thanks for the explanation, and yeah I see why it currently passes the well-formed checker. But I'm wondering if this is what we want or what it is supposed to be. I feel like this duplicated name would be very confusing if we are debugging things, especially when we print out IRs. Is there any necessity for this feature?

Lunderberg: Hmm.  It is really useful if you're merging parts of TIR/Relax that were generated independently.  Those are guaranteed to have different `tir.Var` and `relax.Var` objects, but can very easily have duplicate names.  I think the only place we de-duplicate the names themselves is in an `IRModule`, and even that gets kind of hairy with both `String` and `GlobalVar` acting as unrelated unique keys.

I think that re-use of the same name should be well-formed, but there should be a utility pass (potentially called several times during lowering) that de-duplicates the name within a function.  That way, it provides the better debugging experience, but doesn't require renaming variables when constructing TIR.  It also avoids promising that variable names will remain stable, or that they should be relied upon, since that isn't a stable assumption at the moment.

Lunderberg:
@sunggg Also, I've added you as a co-author for this PR.

tests/python/relax/test_bind_symbolic_vars.py
sunggg: Ah, my apology. I missed the part that it already shows renamed varname when printing functions in your example. Just realized by trying by my own. My concern was printing the same name for different variables, which is turned out to be a false alarm. So, I agree with you, this seems okay. Thank you for clarifying!

src/relax/analysis/struct_info_analysis.cc
```diff
@@ -994,6 +1003,17 @@ class SymbolicVarCollector : public relax::ExprVisitor,
 
   void VisitStructInfo_(const FuncStructInfoNode* op) final {
     if (op->params.defined()) {
+      // Visit the parameters once to collect bindings, and another
+      // time to collect usages.  Otherwise, a symbolic variable
+      // defined by a later parameter may be treated as undefined when
+      // used by an earlier parameter.
```
slyubomirsky, line 1009: When does this happen? I think by definition, the symbolic variables are defined at their first appearance in the signature (going left to right)

src/relax/analysis/struct_info_analysis.cc
```diff
@@ -1014,14 +1034,14 @@ class SymbolicVarCollector : public relax::ExprVisitor,
   }
 
   void VisitStructInfoExprField(const PrimExpr& expr) final {
-    if (mode_ == VisitMode::kMatchVarDef && expr->IsInstance<tir::VarNode>()) {
-      // populate symbolic var in first occurrence
-      const auto& var = Downcast<tir::Var>(expr);
-      if (defined_symbolic_var_.count(var) == 0) {
-        defined_symbolic_var_.insert(var);
+    if (mode_ >= VisitMode::kMatchVarDef) {
```
slyubomirsky, line 1037: Do we use this pattern elsewhere? I'm not personally very fond of it because it could create problems for adding more modes later, but I'm not against adding it if we do this already.

slyubomirsky (approved):
These seem like solid changes and the implementation is straightforward. Thank you very much! Please have a look at my comments if you have the chance.

src/relax/utils.cc
```diff
@@ -61,6 +65,49 @@ class ExprBinder : public ExprMutator {
...
+    //
+    // The strided_slice operator currently stores the begins/ends in
+    // the CallNode::attrs.  Because the CallNode::attrs is only
+    // intended to store static information, any PrimExpr members in
+    // the attributes are not visited by `ExprMutator::VisitPrimExpr`.
+    // Therefore, these must be explicitly visited.
+    //
+    // When the strided_slice operator is updated to store begins/ends
+    // as a tuple of `relax::PrimValue` in the arguments, this special
+    // case can be removed.
```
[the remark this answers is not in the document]
Lunderberg: That's true, and I'd like to update `strided_slice` at some point as well.  The comments in the `strided_slice` implementation state that it should be updated when `relax.PrimValue` is more widely supported, so that is the long-term goal.  However, I think that's a larger change, as it would be a change to the Relax IR, and not just a transformation between Relax functions, and should probably be left for a later PR.

src/relax/analysis/struct_info_analysis.cc
Lunderberg: The symbolic variables are defined in their first appearance in the signature, but only if that appearance is sufficient to provide a definition.  If the symbolic variable can be matched to a specific expression in the caller (e.g. ), then it is bound.  If the symbolic variable cannot be matched to a specific expression in the caller, then it is a usage-site instead.  So, `R.Tensor(shape=[16])` binding to `R.Tensor(shape=[sym_var])` implies a binding of `{sym_var: 16}`, but `R.Tensor(shape=[16])` binding to `R.Tensor(shape=[sym_var1 * sym_var2])` cannot provide any definitions.

So, this edge case occurs in cases like the one shown below, where the first occurrence is a usage, and the definition doesn't occur until a later parameter.

```python
def func(A: R.Tensor(shape=[sym_var1*sym_var2]), B: R.Tensor(shape=[sym_var1, sym_var2])):
    #                       ^^^^^^^^                                ^^^^^^^^
    #                       Usage of sym_var1                       Definition of sym_var1
    ...
```

src/relax/analysis/struct_info_analysis.cc
Lunderberg: The pattern of GT/LT enums is used a couple of places, most commonly in the `OpPatternKind` when checking if an operation is as permissible or more permissible than required for a transformation.  That said, I agree that this isn't the most readable, and it probably would be better to have separate visitors to collect defined variables and to collect free variables.

src/relax/analysis/struct_info_analysis.cc
slyubomirsky: Aha, I hadn't thought of that case. Very good.

PR #15513, opened 2023-08-09
[TOPI] check empty array of x86 injective's iters

wrongtest-intellif:
Just add a missing check, for following simple case:
```
#[version = "0.0.5"]
def @main(%v: float32 /* ty=float32 */) -> float32 {
  %0 = fn (%X: float32 /* ty=float32 */, Primitive=1) -> float32 {
    add(%X, 1f /* ty=float32 */) /* ty=float32 */
  } /* ty=fn (float32) -> float32 */;
  %0(%v) /* ty=float32 */
}
```


vinx13 (approved):

lhutton1:
Is it possible to add a test case to this fix to prevent regression in the future?

Issue #15463, opened 2023-08-02
[TensorIR] Enhance PointerTypeRewrite


tqchen:
Our current GPU backends sometimes can generate the following code:

```c++
void kernel(float* global_data, float* res) {
    float data[4];
    ((float4*)data)[0] = ((float4*)global_data)[0];
     res[0] = data[0] + data[1] + data[2] + data[3];
}
```

Here we are doing vectorized store into a local memory then do a scalarized access from the local memory. While such pattern can be supported by compilers like cuda, it is not always the most robust path. Additionally, backends like vulkan, webgpu comes with the restriction that we can only have one kind of access to the data, resulting the code to be scalarized.

A better version of the code is as follows, when the following conditions are met:
- We detect we have aligned vector store into data
- For all read access to the data, they are scalar access, and can be represented by extract_element(Shuffle), this can be done by running modular analysis and check that all access have coeff that is multiple of the vector lane 

When such condition are met, we can still rewrite data into the vector type, and rewrite all read access of the data to extract_element(represented via shuffle).

```c++
void new_kernel(float4* global_data, float* res) {
    float4 vec_data[1];
    vec_data[0] = global_data[0];
    // access the elements of the vector.
    res[0] = vec_data[0].x + vec_data[0].y + vec_data[0].z + vec_data[0].w;
}
```

This is a pretty common case in GPU programs so it would be desirable to upgrade  `PointerTypeRewrite` to support such cases. This would likely help to robustify our overall support as the new_kernel is also what most programmer will write by handle 

## Action Items

- [ ] Enhance the var analysis to decouple buffer read and write/opaque access, 
     - The read part needs to contain a coeff entry via Modular analysis that summarizes whether rewrite to extract_element is possible
     - For all read access, do Modular analysis to get coeff, so we know that index = x * coeff + offset
     - If coeff % vector_lane == 0, then it means we can rewrite all scalar access to vector access `vec_data[(index - offset % veclane) // vec_lane].extract_element(offset % veclane)`. We can keep GCD of all coeff access to trigger the condition
     - To keep backward compatibility, we can put it behind a flag `rewrite_scalar_read_to_vector_shuffle` 
- [ ] Rewrite scalar read access into extract_element when we find such cases.
- [ ] Support extract_element shuffle pattern (more general shuffle pattern is OK but extract_element is the main one) in codegen , then turn `rewrite_scalar_read_to_vector_shuffle` on for the cases

[...]

PR #15517, opened 2023-08-09
[TIR] Shuffle in PointerValueTypeRewrite for scalar reads 

vinx13:
Added an option `rewrite_scalar_read_to_vector_shuffle` in `PointerValueTypeRewrite` (currently
only enabled for Vulkan). When enabled, when a buffer has both scalar
and vector reads, the buffer will be vectorized if possible and scalar
reads will be achieved via T.Shuffle.

Close https://github.com/apache/tvm/issues/15463.

cc @tqchen @Lunderberg @sunggg

### DIFF
--- a/src/tir/transforms/storage_rewrite.cc
+++ b/src/tir/transforms/storage_rewrite.cc
@@ -33,6 +33,7 @@
 #include <tvm/tir/transform.h>
 
 #include <map>
+#include <numeric>
 #include <unordered_map>
 #include <unordered_set>
 
@@ -1083,17 +1084,14 @@ struct BufferVarInfo {
 
     DataType preferred_base_type = *base_access_dtype.begin();
 
-    // If there is only one vectorizable size used to access the
-    // buffer, and if that access size is compatible with the array
-    // size, then the buffer is vectorizable.  In the future, this
-    // could be improved to allow vectorized buffer access of size
-    // GCD(*lanes_used), if necessary.
     int preferred_lanes = element_dtype.lanes();
-    if ((element_dtype.lanes() == 1) && (access_dtype.size() == 1)) {
+    if (element_dtype.lanes() == 1) {
+      int lanes = access_dtype.begin()->lanes();
+      for (auto dtype : access_dtype) {
+        lanes = std::gcd(lanes, dtype.lanes());
+      }
       arith::Analyzer analyzer_;
       arith::ModularSet me = analyzer_.modular_set(extent);
-
-      int lanes = access_dtype.begin()->lanes();
       if ((me->coeff % lanes == 0) && (me->base % lanes == 0)) {
         preferred_lanes = lanes;
       }
@@ -1120,8 +1118,10 @@ class VectorTypeAccessChecker : public StmtExprVisitor {
    * type as it is later accessed, with scalar element types.
    */
   VectorTypeAccessChecker(const Array<tir::Var>& params, const Map<Var, Buffer>& buffer_map,
-                          bool allow_untyped_pointers = false)
-      : allow_untyped_pointers_(allow_untyped_pointers) {
+                          bool allow_untyped_pointers = false,
+                          bool detect_scalar_read_patterns = true)
+      : allow_untyped_pointers_(allow_untyped_pointers),
+        detect_scalar_read_patterns_(detect_scalar_read_patterns) {
     // If a parameter is in the buffer map, we want to track the
     // version in the map.
     for (auto it : buffer_map) {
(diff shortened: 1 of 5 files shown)


tqchen:
@Lunderberg do you mind help reviewing this

Lunderberg:
I can, though I probably won't have time available to do so until later next week.

tqchen:
ah ok, i can help to take a look  then

src/tir/transforms/storage_rewrite.cc
```diff
@@ -1304,6 +1307,18 @@ class VectorTypeAccessChecker : public StmtExprVisitor {
...
     }
 
+    if (detect_scalar_read_patterns_ && is_buffer_load && indices.size()) {
+      const PrimExpr last_dim_index = indices[indices.size() - 1];
+      if (last_dim_index.dtype().lanes() == 1) {
+        arith::ModularSet me = analyzer_.modular_set(last_dim_index);
+        if (me->coeff > 0) {
+          // When coeff == 0, the index is constant and doesn't need to be recorded since it can
+          // always be rewritten to shuffle.
+          var_info.access_dtype.insert(access_dtype.with_lanes(me->coeff));
```
tqchen, line 1317: i think we need a data structure that captures two things: 

```c++
struct VarReadInfo {
   DataType access_dtype;
   // maintained as GCD of all coef of access index
   int64_t simd_coeff;
};
```
This is mainly to ensure that we don't end up use a very large vector here

tqchen (commented):
Thanks wuwei, some comment

src/tir/transforms/storage_rewrite.cc
vinx13: access_dtype also contains dtype for write, so the result dtype will be eventually bound by the vectorization lanes of writes

src/tir/transforms/storage_rewrite.cc
```diff
@@ -1083,17 +1084,14 @@ struct BufferVarInfo {
...
-    // buffer, and if that access size is compatible with the array
-    // size, then the buffer is vectorizable.  In the future, this
-    // could be improved to allow vectorized buffer access of size
-    // GCD(*lanes_used), if necessary.
     int preferred_lanes = element_dtype.lanes();
-    if ((element_dtype.lanes() == 1) && (access_dtype.size() == 1)) {
+    if (element_dtype.lanes() == 1) {
+      int lanes = access_dtype.begin()->lanes();
+      for (auto dtype : access_dtype) {
+        lanes = std::gcd(lanes, dtype.lanes());
```
tqchen, line 1091: use https://github.com/apache/tvm/blob/main/src/arith/int_operator.h#L171

src/tir/transforms/storage_rewrite.cc
tqchen: the main corner case would be when we do not have writes and only reads.

vinx13: updated

tqchen (approved):

src/tir/transforms/storage_rewrite.cc
```diff
@@ -1081,19 +1088,24 @@ struct BufferVarInfo {
...
+    if (access_dtype.empty()) {
+      return element_dtype;
+    }
     DataType preferred_base_type = *base_access_dtype.begin();
 
-    // If there is only one vectorizable size used to access the
-    // buffer, and if that access size is compatible with the array
-    // size, then the buffer is vectorizable.  In the future, this
-    // could be improved to allow vectorized buffer access of size
-    // GCD(*lanes_used), if necessary.
```
vinx13, line 1090: it seems taking gcd can cause regressions in some source codegen cases (e.g cuda). For example, when a buffer is accessed via `ramp(lanes=2)` and `ramp(lanes=4`), rewriting the buffer to float2 will disable the vectorization for float4.

tqchen: How about we only trigger rewriting if:
- there is vector write
- original dtype was float4
- there is only scalar read access

In a sense that comparing to original case, the only difference is that we allow scalar read from vector write

vinx13: That makes sense, I've updated rewriting. Now the rewriting only trigger if there is one vector access dtype with lanes same as the coeff scalar access.

MasterJH5574:
Hi @vinx13, we noticed that this PR breaks the WebGPU codegen as the WebGPU codegen right now does not support tir::ShuffleNode. Therefore, exceptions are thrown in pass `PointerValueTypeRewrite` at the beginning of WebGPU codegen.

Here is a reproducible example https://gist.github.com/MasterJH5574/3e6141a1d6dcafd383ed6f7ac27e3318. Need to run it under this commit https://github.com/mlc-ai/relax/commit/631f37b6bf8b101d16ecc55de7e6a749a3588570 since the WebGPU codegen in mlc-ai/relax is slightly different from the codegen here due to some conflicts.

cc @tqchen

vinx13:
Is it possible to support it in codegen? Usually this can be supported via element extraction  e.g `vex.x/y/z`. Alternatively we can set `rewrite_scalar_access_to_vector_shuffle` to false

tqchen:
I think we should support via codegen

CharlieFRuan:
Will look into adding Shuffle support for WebGPU!

guoyaol:
Hi @vinx13 
I was building TVM mod for stable diffusion, I found some kernels failed to pass the ICHECK in this PR. Wonder if you have any insights about how to tackle this? Thanks.
```
  File "/Users/guoyaoli/tvm_work/tvm/src/tir/transforms/storage_rewrite.cc", line 1466
InternalError: Check failed: (me->coeff == 0 || info.factor() % me->coeff == 0) is false: 
```
here's a minimum reproducible script, should reproduce the same error running on up-to-date unity branch: https://gist.github.com/guoyaol/af7d1161124987b69b8eb2744b1c399e

cc @MasterJH5574 @tqchen

vinx13:
@guoyaol it may be unaligned vectorizations or a false alarm when the arithmetic analyzer can't handle the index pattern

PR #15519, opened 2023-08-10
[TEST] Run tests/python/relay/aot tests in ci-cortexm

ekalda:
The tests in `tests/python/relay/aot` were run as part of `task_python_integration.sh` which is run for `ci-cpu` and `ci-arm`, but some of tests in that folder require CMSIS-NN or FVP, so in practice they were not running anywhere in the upstream CI.

Since all the tests in that folder could run on `ci-cortexm`, change the scripts to run them there.

Also fix some tests that started failing as a result of #15103 (but didn't manifest in upstream CI) and `test_device_api_hooks_unpacked_api` that clearly hasn't been running in any CI for a while.

### DIFF
--- a/tests/scripts/task_python_integration.sh
+++ b/tests/scripts/task_python_integration.sh
@@ -64,7 +64,7 @@ run_pytest ctypes ${TVM_INTEGRATION_TESTSUITE_NAME}-integration tests/python/int
 run_pytest ctypes ${TVM_INTEGRATION_TESTSUITE_NAME}-contrib tests/python/contrib --ignore=tests/python/contrib/test_ethosu --ignore=tests/python/contrib/test_cmsisnn
 # forked is needed because the global registry gets contaminated
 TVM_TEST_TARGETS="${TVM_RELAY_TEST_TARGETS:-llvm;cuda}" \
-    run_pytest ctypes ${TVM_INTEGRATION_TESTSUITE_NAME}-relay tests/python/relay
+    run_pytest ctypes ${TVM_INTEGRATION_TESTSUITE_NAME}-relay tests/python/relay --ignore=tests/python/relay/aot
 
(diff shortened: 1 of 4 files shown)


tests/scripts/task_python_integration.sh
```diff
@@ -64,7 +64,7 @@ run_pytest ctypes ${TVM_INTEGRATION_TESTSUITE_NAME}-integration tests/python/int
 run_pytest ctypes ${TVM_INTEGRATION_TESTSUITE_NAME}-contrib tests/python/contrib --ignore=tests/python/contrib/test_ethosu --ignore=tests/python/contrib/test_cmsisnn
 # forked is needed because the global registry gets contaminated
 TVM_TEST_TARGETS="${TVM_RELAY_TEST_TARGETS:-llvm;cuda}" \
-    run_pytest ctypes ${TVM_INTEGRATION_TESTSUITE_NAME}-relay tests/python/relay
+    run_pytest ctypes ${TVM_INTEGRATION_TESTSUITE_NAME}-relay tests/python/relay --ignore=tests/python/relay/aot
```
lhutton1, line 67: nit: we probably don't need to ignore these on ci_cpu as the irrelevant tests will be skipped

lhutton1 (approved):
LGTM!

tests/scripts/task_python_integration.sh
ekalda: Yes I was thinking about it, but there are also few hundred tests in that folder that won't be skipped on neither `ci_cpu` nor `ci_cortexm` and are not really dependent on anything in these environments, so I don't see a reason to duplicate these tests.

lhutton1: good point, let's leave it :)

lhutton1:
Thanks @ekalda!

PR #15523, opened 2023-08-10
[Unity][Bugfix] Handle assignment of non-dataflow var to a dataflow var in CanonicalizeBindings

slyubomirsky:
Split off from #15474: This is a simple bugfix that corrects a segfault in the `CanonicalizeBindings` pass. Namely, it did not handle the case of a non-dataflow var being bound to a dataflow var (which is permitted) and gave a segfault if it was encountered.


yongwww (approved):
LGTM

PR #15537, opened 2023-08-12
[CPP_RPC] export listdir for RPC

Ever-Kid:
This PR register `tvm.rpc.server.listdir` using `ListDir` and export it in python frontend for PRC use.

### DIFF
--- a/apps/cpp_rpc/rpc_env.cc
+++ b/apps/cpp_rpc/rpc_env.cc
@@ -73,6 +73,13 @@ namespace runtime {
 void CleanDir(const std::string& dirname);
 
 /*!
+ * \brief ListDir get the list of files in a directory
+ * \param dirname The root directory name
+ * \return vector Files in directory.
+ */
+std::vector<std::string> ListDir(const std::string& dirname);
+
+/*!
  * \brief buld a shared library if necessary
  *
  *        This function will automatically call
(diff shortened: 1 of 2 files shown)


apps/cpp_rpc/rpc_env.cc
```diff
@@ -72,6 +72,13 @@ namespace runtime {
  */
 void CleanDir(const std::string& dirname);
 
+/*!
+ * \brief ListDir get the list of files in a directory
+ * \param dirname The root directory name
+ * \return vector Files in directory.
+ */
+std::vector<std::string> ListDir(const std::string& dirname);
```
wrongtest-intellif, line 80: Is the implementation missing?

apps/cpp_rpc/rpc_env.cc
```diff
@@ -123,6 +130,15 @@ RPCEnv::RPCEnv(const std::string& wd) {
     *rv = this->GetPath(args[0]);
   });
 
+  TVM_REGISTER_GLOBAL("tvm.rpc.server.listdir").set_body([this](TVMArgs args, TVMRetValue* rv) {
+    std::string dir = this->GetPath(args[0]);
+    std::string ret;
+    for (auto d : ListDir(dir)) {
+      ret.append(d.append(","));
```
wrongtest-intellif, line 137: Could we use ostringstream here?

python/tvm/rpc/client.py
```diff
@@ -145,6 +145,18 @@ def remove(self, path):
             self._remote_funcs["remove"] = self.get_function("tvm.rpc.server.remove")
         self._remote_funcs["remove"](path)
 
+    def listdir(self, path):
+        """ls files from remote temp folder.
+
+        Parameters
+        ----------
+        path: str
+            The relative location to remote temp folder.
```
wrongtest-intellif, line 154: It would be great to describe the return type and split token.

wrongtest-intellif (changes requested):

apps/cpp_rpc/rpc_env.cc
Ever-Kid: already coded downside the same file

wrongtest-intellif (approved):

PR #15547, opened 2023-08-14
[TVMScript] Create loop var with min_val dtype in for frame

Lucien0:
Currently, int is always used when creating loop var in a for frame. So if the iteration variable is uint when creating the for frame, it will fail.

The explicit type brings great convenience to subsequent analysis. So I think it's better to create explicit variables rather than convert to int.


Hzfengsy (approved):
LGTM. There are two questions beyond this PR:
1. What's the purpose of using `uint` in loops? For arithmetic analysis or it's related to the hardware backend.
2. What's the Script syntax when we use `uint`, i.e., you've added a test case for the ir_builder, which is great. But I wonder how it can be used in the parser

Lucien0:
Thanks for reply. I think these two problems can be shown in one example:

```
uint_var = T.call_extern("uint32", "some_function")
divided_var = T.truncdiv(128, uint_var)
for i in range(divided_var):
    xxx

if divided_var % 4 == 0:
    xxx
```

1. Our hardware has an architecture like OpenCL, here this function may like get_local_id. I used the divided_var after the for loop in mod, which will analysis its min_val in pass lower intrin. If set to int, its min_val will be -128 rather than 0.
Actually, I can indeed cast like `for i in range(T.int32(divided_var))`, but it is horrible to add this in every for statement and is unnecessary. Just let the variable adapt to the later type.

3. I just create uint var through call_extern. 

BTW, I come from Arm China. And we discovered this issue when utilizing TVM Script to create a DSL for our custom operators during the process of writing the actual operators. cc @Johnson9009

PR #15551, opened 2023-08-14
[VTA] Use std c++ allocation functions rather than memalign/free

grg-haas:
The VTA runtime uses `memalign` in two places to allocate aligned memory. While technically deprecated, this function does usually work in typical software environments. However, when working with mixed C/C++ runtime environments, this approach can error catastrophically as the two languages allocate the same memory pages on their respective heaps. These two instances of `memalign`/`free` are the only two remaining ones which use C-like allocation APIs. Therefore, this PR moves these over to `std::aligned_alloc`/`std::free` in order to fully standardize on C++ allocation APIs and avoid this issue.

### DIFF
--- a/vta/runtime/runtime.cc
+++ b/vta/runtime/runtime.cc
@@ -76,9 +76,9 @@ class AlignmentAllocator : public std::allocator<T> {
 
   inline const_pointer address(const_reference r) const { return &r; }
 
-  inline pointer allocate(size_type n) { return (pointer)memalign(N, n * sizeof(value_type)); }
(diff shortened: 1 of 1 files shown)


grg-haas:
Unsure who to tag here as a reviewer, but I believe some of this code is by @tmoreau89 and seems to have been semi-recently worked on by @Lunderberg

Lunderberg:
Thank you for the update, and I like the standardization on `std::aligned_alloc`.  During the recent updates to ensure aligned memory allocations, I had mostly retained whatever style of allocation was done at each point, to minimize the extent of the changes, and hadn't considered the potential of incompatible C/C++ runtimes.

PR #15566, opened 2023-08-15
[Unity][FX] Add fx support for functional linear and conv2d

jwfromm:
This PR adds support for the functional variants of `linear` and `conv2d` in the relax fx importer. I was trying out using the fx path to import SDXL and found that it generated functional nodes. After this change, the full import goes smoothly.


MasterJH5574 (approved):

MasterJH5574:
Thank you @jwfromm! Also CC @guoyaol who is looking into SDXL.

MasterJH5574:
Hmmm looks like `fx_translator.py` is conflict with the latest commit https://github.com/apache/tvm/commit/fd08c1c6cd7bbcf7e7a47f6b421d82d29adac986

PR #15568, opened 2023-08-15
[Testing] Allow Capitalized name in CompareBeforeAfter

vinx13:
This PR allows to use capitalized variable name `Before` and `Expected` in `CompareBeforeAfter`. When `before / after` are classes or ir module, using capitalized variable names is more consistent with python naming convention.

cc @Lunderberg

### DIFF
--- a/python/tvm/testing/utils.py
+++ b/python/tvm/testing/utils.py
@@ -1953,6 +1953,10 @@ class CompareBeforeAfter:
     """
 
     def __init_subclass__(cls):
+        assert (hasattr(cls, "before") and hasattr(cls, "Expected")) or (
+            hasattr(cls, "Before") and hasattr(cls, "Expected")
+        ), "The subclass of CompareBeforeAfter should have either 'before' and 'expected', or "
+        "'Before' and 'Expected' defined."
         for name in ["before", "Before"]:
             if hasattr(cls, name):
                 cls.before = cls._normalize_before(getattr(cls, name))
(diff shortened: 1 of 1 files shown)


Hzfengsy (approved):

python/tvm/testing/utils.py
```diff
@@ -1953,10 +1953,13 @@ def expected(A: T.Buffer(1, "int32")):
     """
 
     def __init_subclass__(cls):
-        if hasattr(cls, "before"):
-            cls.before = cls._normalize_before(cls.before)
-        if hasattr(cls, "expected"):
-            cls.expected = cls._normalize_expected(cls.expected)
+        for name in ["before", "Before"]:
```
Lunderberg, line 1956: Can we add an `assert` that at most one of the two options are present?

Lunderberg (commented):
Looks good, just one change to avoid accidental misuse.  I like the change, and I'm always unsure how to name a `@I.ir_module`-annotated class.  On the one hand, it looks like a class definition, but on the other hand, it produces something that acts more like a variable.  Supporting both sounds reasonable to me.

python/tvm/testing/utils.py
Lunderberg: I'm not sure if should be that strong of an assert, as that would prevent defining an intermediate test that defines `transform` and `expected`, but doesn't define `before`.  This could be useful for defining several tests that should all normalize to produce the same result.  This would also break existing cases where the `transform` is defined, then used across several tests in a file.

What if instead, we were to assert that no duplicates names exist?

```python
assert len([getattr(cls,name) for name in ['before','Before'] if hasattr(cls,name)]) <= 1
assert len([getattr(cls,name) for name in ['expected','Expected'] if hasattr(cls,name)]) <= 1
```

PR #15569, opened 2023-08-15
[Unity][MSC][M0.2] MSCGraph core

Archermmt:

This is the first part of Milestone 0: Build MSCGraph core parts. Enable translation between Relay, Relax and MSCGraph without loss information.

To limit each PR in reviewable size, the Milestone 0 will be split into 5 steps:
M0.5: Translation test (relax/relay test && related helper modules in python)

This PR add the very basic data structure for MSC, thus MSCGraph and WeightGraph.

MSCGraph is a DAG structure built from relax or relay function. It saves the info of every Expr and will be used in all the MSC submodules. Structure of MSCGraph is similar with the tf.Graph in tensorflow, thus a graph has nodes (MSCJoint) and tensors (MSCTensor). Each node has parent and children, and each tensor has producer and consumers. To manage the weights in computational graph, the MSCGraph define some Constants as weights, for example the 2nd input of Call(nn.conv2d) will be defined as a weight("weight") of the node. That is essential when we process the weights in computational graph, especially for training stage.

WeightGraph is a DAG structure built from MSCGraph. It describes the topology of weights in MSCGraph. WeightGraph is needed in weight pruning and mask generation for sparse training, when relationship between weights are necessary to make prune strategy. WeightGraph also contains nodes (WeightJoint) and tensors (MSCTensor).

Tests of the MSCGraph will be added in next step(M0.3 MSCGraph Builder). Because the MSCGraph can only be built from relax or relay, no tests can be execute before we build a MSCGraph.

cc @Hzfengsy

### DIFF
--- a/python/tvm/contrib/msc/core/ir/__init__.py
+++ b/python/tvm/contrib/msc/core/ir/__init__.py
@@ -17,4 +17,3 @@
 """tvm.contrib.msc.core.ir"""
 
 from .graph import *
-from .translate import *
--- a/python/tvm/contrib/msc/core/ir/__init__.py
+++ b/python/tvm/contrib/msc/core/ir/__init__.py
@@ -1,0 +1,20 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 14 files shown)


Hzfengsy:
One general question: Is it possible to add testcase in this PR? If not, please add a todo in the tracking issue, and add them in the following PR.

Archermmt:
//Todo: Add test for M0.2 in M0.3, reason: Graph can tested only by build

Archermmt:
Seems like I have no access to adding labels in tracking issue.... I add a comment to mark that. Test will for MSCGraph will be added in M0.3:MSCGraph Builder

Hzfengsy (approved):
LGTM. Let's leave it open for another 2 days to see if anyone else would take a look

Hzfengsy:
Thanks @Archermmt

PR #15572, opened 2023-08-16
[Unity] don't do actual profile when input is dynamic shape

leiwen83:
Since current relax not support value range binding for dynamic shape. It is meaningless to do the real profile, as the result would be always inf.


Hzfengsy:
cc @masahi

masahi:
please send this to main.

leiwen83:
I also want to know whether we could support pass like value range for the dynamic shape?

So that BYOC like tensorRT take range to do further opt? tensorRT opt the model use [min, best, max] range.

masahi (changes requested):
Marking as "Request changes" to prevent this from merging into unity.

PR #15576, opened 2023-08-16
don't do actual profile when input is dynamic shape

leiwen83:
Since current relax not support value range binding for dynamic shape. It is meaningless to do the real profile, as the result would be always inf.


leiwen83:
cc @masahi

PR #10724, opened 2022-03-23
[ARM] Support NCHWc alter layout in the fallback mode

masahi:
I want to extract tuning tasks for ARM int8 tensorization. The current `alter_layout` code in `topi/arm_cpu` doesn't fire in the fallback mode, which has been fixed in the PR.

@tkonolige @comaniac

PR #15584, opened 2023-08-17
[Bugfix][Relay][Strategy] Enable compile time transformation of weights matrix for arm_cpu NHWC quantized conv2d

Anndrey24:
Fixed arm_cpu strategy bug which was causing tensorization errors when using the `AlterOpLayout` pass for the quantized NHWC conv2d schedules, as discovered in https://github.com/apache/tvm/pull/10724. Therefore, we can now also enable the usage of `AlterOpLayout` for these schedules in order to transform the weight matrix at compile time, instead of runtime as before.  

I also modified the padding in `Conv2DGemmWeightTransformRel` and `interleave_transpose_weights` to reflect the changes made in https://github.com/apache/tvm/pull/13669 and updated the AlterOpLayout tests accordingly.  

cc @ekalda @lhutton1 @leandron

### DIFF
--- a/python/tvm/relay/op/strategy/arm_cpu.py
+++ b/python/tvm/relay/op/strategy/arm_cpu.py
@@ -468,18 +468,23 @@ def conv2d_gemm_without_weight_transform_strategy_arm_cpu(attrs, inputs, out_typ
     layout = attrs.data_layout
     data = inputs[0]
     strategy = _op.OpStrategy()
+    is_aarch64 = target.features.is_aarch64
+    has_asimd = target.features.has_asimd
+    has_dot_prod = target.features.has_dotprod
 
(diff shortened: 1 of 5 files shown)


lhutton1 (commented):
Thanks @Anndrey24, this looks like a great improvement! It seems we have test coverage for the alter op layout pass which is great, although, it would be good to also have a test to check    no weight transform happens at runtime. Is it possible to extend https://github.com/apache/tvm/blob/8b37d4d11ac94da3c9cfcced50804784f38d3f91/tests/python/relay/strategy/test_select_implementation.py#L60 to check the schedule that gets selected is "conv2d_NHWC_quantized_*_without_transform.arm_cpu"?

ekalda (approved):
Thanks @Anndrey24, very well spotted!

PR #15587, opened 2023-08-18
[Unity][Frontend][NN] Conv1D/LayerNorm NN module

LeshengJin:
Add support for Conv1D and LayerNorm in our torch-like nn api.

cc: @junrushao @cyx-6 @jwfromm

### DIFF
--- a/python/tvm/relax/frontend/nn/op.py
+++ b/python/tvm/relax/frontend/nn/op.py
@@ -179,6 +179,89 @@ def matmul(a: Tensor, b: Tensor, out_dtype: Optional[str] = None, name: str = "m
     return _wrap_nested(_op.matmul(a._expr, b._expr, out_dtype=out_dtype), name)
 
 
+def conv1d(
+    input: Tensor,
+    weight: Tensor,
+    bias: Optional[Tensor] = None,
+    stride: Optional[Union[int, Tuple]] = 1,
+    padding: Optional[Union[int, Tuple, str]] = 0,
+    dilation: Optional[Union[int, Tuple]] = 1,
+    groups: Optional[int] = 1,
+    name: str = "conv1d",
+) -> Tensor:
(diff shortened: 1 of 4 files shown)


python/tvm/relax/frontend/nn/op.py
```diff
@@ -179,6 +179,89 @@ def matmul(a: Tensor, b: Tensor, out_dtype: Optional[str] = None, name: str = "m
     return _wrap_nested(_op.matmul(a._expr, b._expr, out_dtype=out_dtype), name)
 
 
+def conv1d(
+    input: Tensor,
```
jwfromm, line 183: Although i originally used this name for conv2d as well, I decided to switch to `x` for consistency since thats what every other op does. We should probably do the same here.

python/tvm/relax/frontend/nn/op.py
```diff
@@ -14,7 +14,7 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-# pylint: disable=too-many-lines,invalid-name,protected-access
+# pylint: disable=too-many-lines,invalid-name,protected-access,redefined-builtin
```
jwfromm, line 17: This can be reverted once `input` is renamed to `x`

jwfromm (approved):
LGTM, just minor nitpicks on variable names in `op.py`

python/tvm/relax/frontend/nn/op.py
LeshengJin: Great idea. Changed to `x`. Thanks for your review!

jwfromm:
@LeshengJin seems like there's some minor conflict, once resolved we can merge this :)

LeshengJin:
@jwfromm Resolved! Let's merge it after ci finished. Thanks!

PR #15594, opened 2023-08-20
[Runtime] Ensure a DSOLibrary is loaded only once

junrushao:
This PR introduces a DSOLibrary cache that ensures a DSO library is only dlopen-ed once in a single process. This guard is less useful per se as Linux already provides such guanratee backed by reference counter:

> If the same library is loaded again with dlopen(), the same file handle
> is returned. The dl library maintains reference counts for library
> handles, so a dynamic library is not deallocated until dlclose() has
> been called on it as many times as dlopen() has succeeded on it. The
> _init() routine, if present, is only called once. But a subsequent call
> with RTLD_NOW may force symbol resolution for a library earlier loaded
> with RTLD_LAZY.

Link: https://linux.die.net/man/3/dlopen.

However, it will help TVM metadata to be more properly handled at loading time in a multi-threading environment.

### DIFF
--- a/src/runtime/dso_library.cc
+++ b/src/runtime/dso_library.cc
@@ -26,6 +26,9 @@
 #include <tvm/runtime/packed_func.h>
 #include <tvm/runtime/registry.h>
 
+#include <mutex>
+#include <unordered_map>
+
 #include "library_module.h"
 
 #if defined(_WIN32)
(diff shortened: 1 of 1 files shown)


junrushao:
@tqchen please review. Also, I was a bit uncertain if we should also guard anything in [`CreateModuleFromLibrary`](https://github.com/apache/tvm/blob/main/src/runtime/library_module.cc#L207-L228) in a multi-threading environment?

junrushao:
I do not believe this is a right fix.

PR #15595, opened 2023-08-20
[Runtime] Enhance PackedFunc Metaprogramming with `PackArgs`

junrushao:
This PR introduces a convenient method `PackArgs` that allows packing variadic C++ parameters into TVM's PackedFunc calling convention without recursion using C++ parameter pack tricks.

As an example,

```C++
template <typename... Args>
void MyVariadicMethod(Args&&... args) {
  constexpr int kNumArgs = sizeof...(Args);
  TVMValue values[kNumArgs];
  int type_codes[kNumArgs];
  tvm::runtime::PackArgs(values, type_codes, std::forward<Args>(args)...);
  ...
}
```

### DIFF
--- a/include/tvm/runtime/packed_func.h
+++ b/include/tvm/runtime/packed_func.h
@@ -1284,13 +1284,27 @@ namespace parameter_pack {
 
 template <typename... EnumArgs>
 struct EnumeratedParamPack {
-  struct Invoke {
-    template <template <size_t i, typename TArgument> class Functor, typename... ExtraParams>
-    static void F(ExtraParams&&... extra_params) {
+  struct InvokeWithoutArg {
+    template <template <size_t i, typename TArgument> class Functor, typename ExtraParams>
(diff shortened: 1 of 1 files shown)


junrushao:
CC: @cyx-6 @tqchen

cyx-6 (approved):
LGTM!

include/tvm/runtime/packed_func.h
```diff
@@ -1622,6 +1641,20 @@ inline TVMRetValue PackedFunc::operator()(Args&&... args) const {
   return rv;
 }
 
+template <int i, typename T>
```
kparzysz-quic, line 1644: This should be `size_t` not `int`.  Otherwise it fails to compile with clang.

junrushao: Yeah I found the same thing: https://github.com/apache/tvm/pull/15637...It works with my GCC but not clang

PR #15597, opened 2023-08-21
add tir_var_best_tuned function attr

leiwen83:
For cutlass lowlevel profiler, it need concrete shape for profiler usage. For the matmul/conv2d ops, we may pass dynamic shape, so that it need to how to interpret that variable when profiler meet it.


leiwen83:
cc @masahi

PR #15601, opened 2023-08-21
[Unity] Add instruments to relay translator

quic-sanirudh:
Sometimes it is useful to instrument relay passes that are run during relay to relax translation, and this patch adds a new argument to relay translator to accept instruments list.

The instruments can be passed on as

```python
relay_translator.from_relay(..., instruments=[MyInstrumentsObject])
```

### DIFF
--- a/python/tvm/relax/testing/relay_translator.py
+++ b/python/tvm/relax/testing/relay_translator.py
@@ -18,11 +18,12 @@
 # pylint: disable=too-many-nested-blocks, unused-variable
 """Relay to Relax translator."""
 
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, Sequence
 
 import tvm
 from tvm import relax, relay
 from tvm.ir.module import IRModule
+from tvm.ir.instrument import PassInstrument
 from tvm.relax.testing import nn
 from tvm.relay.backend.te_compiler import select_implementation
 from tvm.runtime import NDArray
@@ -37,6 +38,7 @@ def from_relay(
     *,
     opt_level: int = 3,
     pass_config: Optional[Dict[str, Any]] = None,
+    instruments: Optional[Sequence[PassInstrument]] = None,
     disabled_pass: Optional[List[str]] = None,
     translate_op_with_tir: Optional[Dict[str, tvm.tir.PrimFunc]] = None,
     append_op_attrs: bool = False,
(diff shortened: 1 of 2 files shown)


quic-sanirudh:
@tvm-bot rerun

quic-sanirudh:
@tvm-bot rerun

quic-sanirudh:
@tvm-bot rerun

quic-sanirudh:
cc @junrushao @yongwww

quic-sanirudh:
cc @YuchenJin @Lunderberg

python/tvm/relax/testing/relay_translator.py
```diff
@@ -255,6 +261,7 @@ def visit_func(node):
         opt_level=opt_level,
         config=pass_config,
         disabled_pass=disabled_pass,
+        instruments=instruments,
```
Lunderberg, line 264: I like it, though I'm wondering if we should have it forward the instrumentation of the existing `PassContext`, rather than explicitly accept it as an argument.  That way, debug instrumentation set by a much higher scope would also be preserved, without requiring an update to the immediately-calling scope.

```python
instruments = tvm.transform.PassCurrent().instruments
```

Lunderberg (changes requested):

python/tvm/relax/testing/relay_translator.py
quic-sanirudh: Thanks for your review @Lunderberg.

I actually did it that way initially, but since the other arguments to `PassContext` such as `opt_level` and `pass_config` were taken as arguments to the relay translator, I modified it take instruments as an argument as well.

I can change it back to be forwarded from current PassContext as you've suggested, either way is fine with me, but in that case, should we also remove `opt_level` and `pass_config` from being arguments and forward them from current context as well? Since that would be a breaking change, I'm a bit hesitant to do that.

Lunderberg: Hmm, good point.  Since the other arguments are currently specified explicitly, probably better to keep it that way for consistency.

In the long-term, I think that entering a PassContext should probably distinguish between options that override the previous (`opt_level`), options that append to the previous (`instruments`, `disabled_passes`).

Lunderberg (approved):
Approved after discussion.  Long-term, the `PassContext` constructed here should probably take all options other than `pass_config` from the current pass context.

quic-sanirudh:
> Approved after discussion. Long-term, the `PassContext` constructed here should probably take all options other than `pass_config` from the current pass context.

Sure, I agree that in the long term, it might be better to pick it up from PassContext. Perhaps to avoid a breaking change initially, we can keep the arguments and prioritize them when provided, but when they're not provided, they can be picked up from the current PassContext.

PR #15415, opened 2023-07-26
[Frontend][ONNX] add onnx Mish operator

Aarsh2001:
[Mish](https://onnx.ai/onnx/operators/onnx__Mish.html)  - tests pass locally for `llvm` target on a `cpu`

PR #15602, opened 2023-08-21
[ONNX][BugFix] Support If body with free variable from graph input 

gangmul12:
When graph inputs are used in an inner body of If node, the original TVM ONNX
frontend did not set the span properly. Because of the wrong or partial span,
new relay.Var is introduced and failed to match identical Var. Firstly, there
was an issue where the free variable of the inner body was updated in _node but
not applied to _input. Secondly, although the free variable of the then body
successfully updated to relay.Var in _node, but this was obscured by the update
of _node in the else body.

This commit fixes the ONNX importer and adds an ONNX import testcase for the
revised code.

### DIFF
--- a/python/tvm/relay/frontend/onnx.py
+++ b/python/tvm/relay/frontend/onnx.py
@@ -6959,6 +6959,7 @@ class GraphProto:
                 op = fold_constant(op)
             else:
                 op = _expr.TupleWrapper(fold_constant(op.astuple()), len(op))
+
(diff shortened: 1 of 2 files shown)


gangmul12:
cc @Hzfengsy . Can you review this when you are free? I found you reviewed the most recent ONNX PR https://github.com/apache/tvm/pull/15415 . I'm not sure who to cc..

Hzfengsy (approved):

PR #15603, opened 2023-08-21
[Unity][Frontend][NN] Add Timesteps layer to NN Module API

jwfromm:
This PR adds support for the huggingface diffusers `Timesteps` layer. This is useful when constructing stable diffusion like models. The definition of the layer required padding, so I also implemented `relax.op.nn.pad` using semantics as close to pytorch as possible.


cyx-6 (approved):
It looks good to me!

PR #15604, opened 2023-08-22
[Unity][Frontend][NN] Op print_ 

LeshengJin:
This pr introduces `op.print_`, which prints runtime tensor value to the screen in one-line code.

Example:

```python
class Model(Module):
    def test(self, x: Tensor):
        z = op.add(x, x)
        op.print_(z) # In the runtime, the value of z would be printed to screen.
        return x
```

cc: @junrushao @cyx-6 @jwfromm


junrushao (approved):
Cool stuff!

junrushao:
Waiting for the CI to pass

PR #15605, opened 2023-08-22
[CMake] Add NCCL to TVM and TVM Runtime

junrushao:
This PR introduces NCCL in the cmake system.
NCCL is NVIDIA's library for distributed communication.


Hzfengsy:
Not sure about NCCL. Here maybe two noob questions:
1. Is NCCL included in the Cuda package or do we need to install it separately? 
2. If it's part of Cuda, what are the benefits of setting a separate switch? i.e. why not turn it on by default if we turn on cuda

junrushao:
@Hzfengsy Thanks for asking!

> Is NCCL included in the Cuda package or do we need to install it separately?

You will need to install it separately.

csullivan:
Verified on my side, this is merged. Thanks @junrushao

PR #15609, opened 2023-08-23
[Unity][Frontend][NN] Add diffusers style Attention layer

jwfromm:
This PR adds support for the `Attention` layer in the nn module API. This layer mimics the behavior of the [Attention layer used in huggingface Diffusers](https://github.com/huggingface/diffusers/blob/80871ac5971fe7e708befa3b553463c4e61b22ab/src/diffusers/models/attention_processor.py#L36). Under the hood it uses scaled dot product attention. Notably, there are still some missing features. For example I didnt add support for attention masks yet. I also am assuming 3 dimensional inputs so this likely wouldnt be useful for the VAE of stable diffusion. However, it does allow us to represent the all attention layers of the UNET. I'm hoping we can expand on the functionality more if needed in future PRs.

### DIFF
--- a/python/tvm/relax/frontend/nn/core.py
+++ b/python/tvm/relax/frontend/nn/core.py
@@ -486,6 +486,9 @@ class ModuleList(Module):
     def __len__(self):
         return len(self.modules)
 
+    def append(self, module):
+        self.modules.append(module)
+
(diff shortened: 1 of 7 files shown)


Hzfengsy:
cc @junrushao @cyx-6

LeshengJin (approved):
Attention is all you need! Thanks, @jwfromm

PR #15612, opened 2023-08-23
[Unity] Prevent Relax VM from being corrupted after getting incorrect shape

yelite:
cc @jwfromm @junrushao


junrushao (approved):
This is a great idea to guard the frames in case of any exceptions thrown!

jwfromm (approved):
Super excited to have this fixed. Thank you Lite!

PR #15614, opened 2023-08-23
[Unity] Replacing unary ops with LookUpTable and Take op to improve performance.

sdalvi-quic:
**Motivation** : Unary ops calculate the value during runtime which is costly. Since we are handling quantized data hence, it is better to pre compute the value during compile time (for 0 -256 value) and store in a Look Up Table which is accessed during the runtime. 

**Implementation** : This pass generates the Look Up Table for each unary op during compile time. It iterates over the original graph and replaces the unary ops call node, (example tanh, sqrt) to a call node that accesses a Look Up Table corresponding to the unary op, and return the value. 

**Testing** : The test_take checks for both value correctness and structural correctness of the output generated after the pass.

### DIFF
--- a/python/tvm/contrib/hexagon/generate_take_op.py
+++ b/python/tvm/contrib/hexagon/generate_take_op.py
@@ -1,0 +1,86 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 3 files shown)


sdalvi-quic:
@jverma-quic @quic-sanirudh @kparzysz-quic

quic-sanirudh:
@sdalvi-quic Can you fix the lint errors, and maybe also update the title to add `[Unity]` tag so that its obvious that this patch goes into unity branch?

PR #15617, opened 2023-08-25
[Unity] fix masked_fill operation

vvchernov:
Fix is needed for correct work of masked_fill op which particularly used by MPT-like models

@masahi


junrushao:
I suppose it should be good to merge :)

PR #15622, opened 2023-08-25
[Unity] Disco: A Framework-Agnostic SPMD Runtime for Distributed Inference/Training

junrushao:
Disco is a distributed runtime that consists of a controler and a cluster of workers. The
controler is responsible for managing the workers by broadcasting commands to all the workers
together, and the workers are responsible for executing the commands and. The controler and
workers communicate with each other through a bi-directional channel.

Different from a generic system, Disco is designed to as "single-program-multiple-data" (SPMD)
runtime, which means that all the workers execute the same instruction at the same time, but the
data they are working on may be different. For example, in data parallelism, each worker may
work on a different batches of the data, but they all execute the same set of instructions.
Therefore, imagine there is a virtual machine that executes the program, the structures of
workers' register files could be considered as "identical" (single program) although the values
may differ (multiple data).

**DRef.** Following the design above, consider the program in SPMD in a virtual ISA, then each
worker is a virtual machine instance to execute the ISA maintaining its own register file.
The controler denotes each of their register files with a unique integer "register id",
and the workers use this id to refer to the register file that resides on itself.
DRef is a control-side object backed by such a register id. The data it contains is not assumed
to be directly accessible by the controler, with an exception for worker-0, which is a special
worker that is always co-located with the controler.

**Worker-0.** Worker-0 is a special worker that is always co-located with the controler.
It is assumed that the controler can synchronize with and access the registers of worker-0.
The Disco session provides multiple APIs to interact specifically with the worker-0.
To shared data with other workers, a common paradigm in Disco is to copy data from the
controler-side NDArray to the worker-0, and then copy it to other workers using primitives on
the data plane, for example, `broadcast` and `send`.

**Control plane.** The controler broadcasts commands to all the workers as control signals.
For example, the control may ask all workers to load a library or call a function respectively.
Common control signals include: shutdown, retrievel a global PackedFunc, call packed function,
etc. The controler is assumed to keep a message channel to each worker to implement the broadcast
behavior, and the message channel may vary depends on usecases.

**Data plane.** The data channel is usually used to exchange data between workers, especially for
tensor data which is usually large. For example, performing an allreduce operator for sharded
matrix multiplication, or broadcasting for an input tensor. For efficiency, the data channel is
usually backed by NCCL on NVIDIA GPUs, RCCL on AMD GPUs, or MPI on CPUs.

**Session.** A Disco session is a primary interface to interact with the Disco runtime, serving
as a global context that manages the control and workers. It could be implemented as a
multi-threaded with a pool of workers for single-node multi-gpu scenarios, or TCP sockets for
workloads that span over a cluster of nodes.

**Channel.** Disco channel is a bi-directional communication channel between the controler and
workers for exchanging control signals. It is no different from a generic RPC channel, but
adopts TVM's PackedFunc calling convention to support polymorphic and variadic arguments.

Joined work with @LeshengJin


src/runtime/disco/session.cc
```diff
@@ -0,0 +1,165 @@
...
+}
+
+DRef SessionObj::CallPackedWithPacked(const TVMArgs& args) {
+  constexpr uint64_t one = 1;
+  constexpr int offset = 4;
+  TVMValue* values = const_cast<TVMValue*>(args.values);
+  int* type_codes = const_cast<int*>(args.type_codes);
+  int num_args = args.num_args;
+  int reg_id = AllocateReg();
+  uint64_t is_dref = 0;
```
tqchen, line 77: introduce a special type code for dref

src/runtime/disco/worker.h
```diff
@@ -0,0 +1,184 @@
...
+  /*! \brief The id of the worker.*/
+  int worker_id;
+  /*! \brief Total number of workers */
+  int num_workers;
+  /*! \brief The default device to allocate data if not specified */
+  Device default_device;
+  /*! \brief The name of the underlying collective communication library. */
+  String ccl;
+  /*!
+   * \brief The data shared between worker-0 and the controler. It's a nullptr if
```
tqchen, line 169: \note this data structure is owned by controller

python/tvm/runtime/disco/session.py
```diff
@@ -0,0 +1,294 @@
...
+from ..._ffi.runtime_ctypes import Device
+from ..ndarray import NDArray
+from ..object import Object
+from . import _ffi_api
+
+
+@register_object("runtime.disco.DRef")
+class DRef(Object):
+    """An object that exists on all workers. The controller process assigns a unique "register id"
+    to each object, and the worker process uses this id to refer to the object residing on itself.
```
jinhongyii, line 32:
```suggestion
    """An object that exists on all workers. The controller assigns a unique "register id"
    to each object, and the worker uses this id to refer to the object residing on itself.
```
remove the word "process" to avoid confusion

python/tvm/runtime/disco/session.py
```diff
@@ -0,0 +1,294 @@
...
+    """
+
+    @property
+    def session(self) -> "Session":
+        """Get the session that this DRef belongs to."""
+        return _ffi_api.DRefSession(self)  # type: ignore # pylint: disable=no-member
+
+    def debug_get_from_remote(self, worker_id: int) -> Any:
+        """Get the value of a DRef from a remote worker.
```
jinhongyii, line 41: add document to emphasize this is only used for debug and its limitations.

python/tvm/runtime/disco/session.py
```diff
@@ -0,0 +1,294 @@
...
+        del dref.handle
+
+    def __getitem__(self, name: str) -> DPackedFunc:
+        func = self.session._get_cached_method("runtime.ModuleGetFunction")
+        return DPackedFunc(func(self, name, False))
+
+
+@register_object("runtime.disco.Session")
+class Session(Object):
+    """A Disco interactive session. It allows users to interact with the Disco command queue with
```
jinhongyii, line 81: emphasize this is controller-side

python/tvm/runtime/disco/session.py
```diff
@@ -0,0 +1,294 @@
...
+    various PackedFunc calling convention."""
+
+    @staticmethod
+    def threaded_session(num_workers: int) -> "Session":
+        """Create a threaded session."""
```
jinhongyii, line 86:
```suggestion
        """Create a thread-scope session."""
```
I don't understand what "threaded" means at first glance.
Maybe use a better name for threaded_session as well.

python/tvm/runtime/disco/session.py
```diff
@@ -0,0 +1,294 @@
...
+            func = cache[name]
+        return func
+
+    def empty(
+        self,
+        shape: Sequence[int],
+        dtype: str,
+        device: Optional[Device] = None,
+    ) -> DRef:
+        """Create an empty NDArray on all workers.
```
jinhongyii, line 106: The implementation of runtime.disco.empty does not match the document. It only creates an empty array on controller instead of workers.

src/runtime/disco/nccl/nccl.cc
```diff
@@ -0,0 +1,141 @@
...
+#include "../session.h"
+#include "./utils.h"
+
+namespace tvm {
+namespace runtime {
+namespace nccl {
+
+struct NCCLGlobalContext {
+  std::vector<int> device_ids;
+  std::vector<ncclComm_t> communicators;
```
jinhongyii, line 37: add a todo to support explicit broadcast if involving socket.

src/runtime/disco/worker.h
```diff
@@ -0,0 +1,184 @@
...
+      return "kMin";
+    case ReduceKind::kMax:
+      return "kMax";
+    case ReduceKind::kAvg:
+      return "kAvg";
+  }
+  LOG(FATAL) << "ValueError: Unknown ReduceKind: " << static_cast<int>(kind);
+}
+
+/*! \brief A bi-directional channel for controler-worker communication. */
```
jinhongyii, line 104:
```suggestion
/*! \brief A bi-directional command channel for controler-worker communication. */
```
To emphasize this is not for transferring ndarray data.

src/runtime/disco/threaded_session.cc
```diff
@@ -0,0 +1,184 @@
...
+};
+
+class ThreadedCommandQueueObj final : public CommandQueueObj {
+ public:
+  explicit ThreadedCommandQueueObj(int n_workers) {
+    for (int i = 0; i < n_workers; ++i) {
+      std::shared_ptr<DiscoThreadChannel> channel = std::make_shared<DiscoThreadChannel>();
+      WorkerZeroData* data = (i == 0) ? &worker_zero_data_ : nullptr;
+      workers_.emplace_back(std::make_unique<DiscoWorker>(i, n_workers, data, channel));
+      channels_.emplace_back(std::move(channel));
```
jinhongyii, line 133: why are we opening channels for each worker? Isn't the controller supposed to communicate with worker 0 only?

jinhongyii (changes requested):

python/tvm/runtime/disco/session.py
tqchen: This is indeed creating an array on the worker side, and attach a DRef that references these array. 

Perhaps we can update the docs to

```
Create an empty NDArray on all workers and attach them to a  DRef
```

src/runtime/disco/threaded_session.cc
tqchen: This is an implementation detail. It is possible to have an impl that have control messages work with all workers. Especially in threaded case. But it does not have to be the only implementation. So this is fine as a first step for threaded.

src/runtime/disco/worker.h
tqchen: Maybe stating 

```
A bi-directional channel for controler-worker communication. This channel is primarily used to transfer control messages and not data.
```

python/tvm/runtime/disco/session.py
tqchen: to expand a bit

```
Create a disco session backed by multiple threads in the same process.
```

python/tvm/runtime/disco/session.py
jinhongyii: ok I was wrong.

src/runtime/disco/session.cc
junrushao: done. we will split it into a separate PR to main

LeshengJin (approved):
Great work! Would bring more followup prs after the merge.

jinhongyii (approved):
LGTM!

include/tvm/runtime/disco/session.h
```diff
@@ -0,0 +1,286 @@
...
+   * \param worker_id The id of the worker to be fetched from.
+   * \return The value of the register.
+   */
+  inline TVMRetValue DebugGetFromRemote(int worker_id);
+  /*!
+   * \brief Copy from the NDArray provided to a remote worker.
+   * \param worker_id The id of the worker to be copied to.
+   * \param source The NDArray to be copied.
+   */
+  void DebugCopyFrom(int worker_id, NDArray source);
```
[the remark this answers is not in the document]
junrushao: I think it copies from a given source, so `CopyFrom` might make more sense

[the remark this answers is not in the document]
jinhongyii: CopyFrom is too similar to GetFrom. Might need to differ from GetFromRemote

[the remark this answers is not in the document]
jinhongyii: It copies an ndarray to a worker, so CopyFrom also makes sense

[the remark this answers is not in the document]
junrushao: I'd love to clarify with a common usecase below:

```python
disco_array = sess.empty(shape, dtype)
disco_array.debug_copy_from(numpy_array)
```

In this case, the subject that takes the "copy" action is `disco_array`, and the action is to copy `numpy_array` to the subject itself. And therefore, from the perspective of the subject who takes action, the action is indeed better to be characterized as copying from.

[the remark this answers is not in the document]
jinhongyii: ok this makes sense

dtrealm:
Great work, do you have any plans to implement MPI communication for multiple nodes?

PR #15625, opened 2023-08-27
[Runtime] Add Stringification to DLDeviceType and DLDevice

junrushao:
This PR adds `DLDeviceType2String` and overloading of ostream's `operator<<` on `DLDevice` so that they could be printed as strings more with better developer ergonomics.


junrushao:
Moved to https://github.com/apache/tvm/pull/15630

PR #15626, opened 2023-08-27
[Unity] Implement relax.Function.bind_params

Lunderberg:
Similar to `relax.Function.bind_symbolic_vars`, implemented in https://github.com/apache/tvm/pull/15509, this commit introduces `relax.Function.bind_params` to allow Relax parameters to be manipulated on a per-function basis.  This utility function and the existing `BindParams` transform both use the same underlying implementation.

### DIFF
--- a/include/tvm/relax/transform.h
+++ b/include/tvm/relax/transform.h
@@ -180,7 +180,7 @@ TVM_DLL Pass EliminateCommonSubexpr(bool call_only = false);
  *
  * \return The Pass.
  */
-TVM_DLL Pass BindParams(String func_name, Map<String, runtime::NDArray> params);
+TVM_DLL Pass BindParams(String func_name, Map<ObjectRef, ObjectRef> params);
 
 /*!
  * \brief Bind symbolic vars to constant shape values.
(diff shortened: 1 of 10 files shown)


Lunderberg:
@sunggg This is a follow-up to the `BindSymbolicVars` commit we made a few weeks ago.  Since `BindSymbolicVars` and `BindParams` are so similar, I wanted to have the same functionality for both.

sunggg (approved):
LGTM! Thank you @Lunderberg. 
Just one basic question, what would be the real use-case for this pass? Is this for multi-gpu?

Lunderberg:
At the moment, this would primarily be to write a single Relax function, then manually manipulate it into a specialized form.  For example, writing an attention function that accepts a boolean filter tensor, then using BindParams to make a sibling function that replaces it with a constant tensor which o Lu contains `True`.

In the past, this would have been used to statically bind model parameters into an end-to-end model.  However, since that requires the compiling machine to have sufficient RAM, it isn’t as flexible as the `LiftTransformParams`.

csullivan (approved):

csullivan:
Thanks @Lunderberg and @sunggg! This is merged.

PR #15628, opened 2023-08-27
[Arith] Added simplification rule for multiple equality compares

Lunderberg:
The expression `(x==y) && (x==z)` requires that `y==z`.  When `y` and `z` are constants, this can allow better constant folding by rewriting `(x==c1) && (x==c2)` into `(x==c1) && (c1==c2)`.

This commit adds the above rewrite, and the corresponding rewrite of the negative expression.  This was initially part of https://github.com/apache/tvm/pull/15627, but does not depend on Relax features, and is split out in order to upstream into TVM main.


junrushao (approved):
LGTM!

PR #15629, opened 2023-08-27
[BugFix][VTA] tvm.tir.Call has no name attribute

youxiudeshouyeren:
tvm.tir.Call has no name attribute, this attribute belongs to op.


youxiudeshouyeren:
Code Review: @junrushao @Lunderberg   Thanks!

Lunderberg (approved):
Good catch, and thank you for the fix!

PR #15631, opened 2023-08-27
[RPC] Enhance RPC Protocol to support TVM Object

junrushao:
This PR introduces object support in TVM RPC protocol by introducing three new interfaces in `rpc_reference.h`:
- `uint64_t GetObjectBytes(Object* obj)`, which is a required implementation that returns the length of the object during serialization;
- `void WriteObject(Object* obj)` used to serialize an object to a writable channel;
- `void ReadObject(int* type_code, TVMValue* value)`, which deserializes a TVM Object from a channel.

To serialize an object, a recommended paradigm is to write its `type_index` first, and then its content. For example, `ShapeTuple` can be serialized as:

```C++
// pseudocode
void WriteObject(Object* obj) {
  if (obj is ShapeTuple) {
    this->Write<uint32_t>(type_index of ShapeTuple);
    this->Write<int32_t>(obj->ndim);
    this->WriteArray<int64_t>(obj->shape);
  } else {
    throw Unsupported;
  }
}

uint64_t GetObjectBytes(Object* obj) {
  uint64_t result = 0;
  if (obj is ShapeTuple) {
    result += sizeof(uint32_t); # for `type_index`
    result += sizeof(int32_t);  # for `ndim`
    result += sizeof(int64_t) * obj->ndim; # for content of the shape
  } else {
    throw Unsupported;
  }
  return result;
}
```

To deserialize an object, similar to serialization, the recommended approach paradigm is to read `type_index` and disptch based on it.

Caveat on deserialization: RPC Reference itself does not own or allocate any memory to store objects, meaning extra logic is usually required in `ReadObject` to keep their liveness.

### DIFF
--- a/include/tvm/runtime/object.h
+++ b/include/tvm/runtime/object.h
@@ -72,6 +72,8 @@ struct TypeIndex {
     kRuntimeShapeTuple = 6,
     /*! \brief runtime::PackedFunc. */
     kRuntimePackedFunc = 7,
+    /*! \brief runtime::DRef */
+    kRuntimeDiscoDRef = 8,
     // static assignments that may subject to change.
     kRuntimeClosure,
     kRuntimeADT,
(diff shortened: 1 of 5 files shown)


junrushao:
CC: @tqchen

tqchen (approved):

PR #15634, opened 2023-08-28
[Disco] Support ShapeTuple in Disco Protocol

junrushao:
ShapeTuple is an essential item used in Relax stack as the runtime representation of shapes. It is also used as a current workaround to represent integers (1-d shape) given standalone non-constant integers are currently absent in Relax.

This PR introduces formal support for ShapeTuple in Disco's communication protocol based on the recent enhancement of TVM RPC system to support TVM Objects: https://github.com/apache/tvm/pull/15631.

### DIFF
--- a/src/runtime/disco/threaded_session.cc
+++ b/src/runtime/disco/threaded_session.cc
@@ -110,6 +110,9 @@ class DiscoThreadedMessageQueue : public dmlc::Stream {
   uint64_t GetObjectBytes(Object* obj) {
     if (obj->IsInstance<DRefObj>()) {
       return sizeof(uint32_t) + sizeof(int64_t);
+    } else if (obj->IsInstance<StringObj>()) {
+      uint64_t size = static_cast<StringObj*>(obj)->size;
+      return sizeof(uint32_t) + sizeof(uint64_t) + size * sizeof(char);
     } else if (obj->IsInstance<ShapeTupleObj>()) {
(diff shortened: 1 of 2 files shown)


junrushao:
CC @jinhongyii @LeshengJin @tqchen

LeshengJin (approved):
LGTM! Thanks @junrushao

src/runtime/disco/threaded_session.cc
```diff
@@ -146,13 +152,17 @@ class DiscoThreadedMessageQueue : public dmlc::Stream {
       this->Read<int64_t>(&dref->reg_id);
       dref->session = Session{nullptr};
       result = ObjectRef(std::move(dref));
+    } else if (type_index == TypeIndex::kRuntimeString) {
+      uint64_t size = 0;
```
tqchen, line 156: To make things consistent, it is helpful to pass in string length, instead of len+1(that includes termination string)

junrushao: I'm not sure if my understanding is correct here. In `WriteObj`, we write the string's actual length to dmlc::Stream, and in `ReadObj` here, we create an `std::string` of length `len + 1`, and update its first `len` characters. I suppose in either case, the terminal character is not included.

jinhongyii (approved):

src/runtime/disco/threaded_session.cc
tqchen: i misread the code. no questions for now

PR #15635, opened 2023-08-28
[Runtime] Fix ICE from Clang

junrushao:
This PR fixes an error reporting from Clang on the line below:

```
/.../include/tvm/runtime/packed_func.h:1706:3: error: no matching function for call to 'F'
  detail::parameter_pack::EnumerateWithArg<Args...>::template F<TVMArgsSetterApply>(
  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

PR #15637, opened 2023-08-28
[Backport][Runtime] Fix ICE from Clang

junrushao:
Backported from #15635. This template is never instantiated on `apache:main` yet, but in case of future ICE, I backported the related fix from Unity branch.


junrushao:
CC: @kparzysz-quic

Hzfengsy (approved):

PR #14690, opened 2023-04-21
[Arith] Add IntegerSetNode to work with Presburger Set

multiverstack-intellif:
Add basic structure for IntegerSet to work with Presburger Set in MLIR as the 1st step for https://github.com/apache/tvm-rfcs/pull/99 , also together with some other necessary modifications:

- Add USE_MLIR (default OFF) as the switch to enable the feature or not.
- Move some reflection related class definition to header file for type checking convenience in other places.

PR #15638, opened 2023-08-28
[Arith] MLIR PresburgerSet compile fix mlir >= 160

cbalint13:
Hi folks,

Some fixes for MLIR based analyzer module introduced by #14690 .

---

* Make CMake at par with LLVM info:
```
{...}
-- Use llvm-config=llvm-config-64
-- LLVM libdir: /usr/lib64
-- Found MLIR
-- Build with MLIR
-- Set TVM_MLIR_VERSION=160
-- Found LLVM_INCLUDE_DIRS=/usr/include
{...}
--    USE_MKL                            : OFF
--    USE_MLIR                           : ON
--    USE_MSVC_MT                        : OFF
{...}
```

* Fix several compilation errors:
```
error: cannot convert 'llvm::SmallVector<long int>' to 'llvm::ArrayRef<mlir::presburger::MPInt>'
error: no matching function for call to 'tvm::IntImm::IntImm(tvm::runtime::DataType, mlir::presburger::MPInt&)'
note:   no known conversion for argument 2 from 'mlir::presburger::MPInt' to 'int64_t' {aka 'long int'}
```

Tested using: ```llvm/mlir 16.0.6```, ```llvm/mlir 15.0.7```,  ```llvm/mlir 17.0.0rc3```

Thanks,
~Cristian.

Cc @multiverstack-intellif , @wrongtest-intellif , @vinx13, @Hzfengsy , @junrushao , @tqchen


junrushao (approved):
Thanks for the bugfixes! Definitely not an expert here, and do you think we need to add a test for changes introduced in presurger_set.cc?

cbalint13:
> Thanks for the bugfixes! Definitely not an expert here, and do you think we need to add a test for changes introduced in presurger_set.cc?

@junrushao ,

Me either, not expert on this topic, unsure how #14690 passed it's own test (can't see related CI logs).
Since here we have compilation fixes (no changes), the  #14690 test case included should be sufficient.

---

Let's wait the tests to pass in green, reported issue here might be related to:
 **(a)** a different ```mlir = 16``` version on my side
 **(b)** a newer ```gcc 13``` not willing to infer casts  (unplausible)

If **(a)**  will adjust this PR with some little ```#if version``` checks (LLVM/MLIR  comes with frequent API changes).

Let's also ask authors: @multiverstack-intellif , @wrongtest-intellif

cbalint13:
> If **(a)** will adjust this PR with some little `#if version` to keep pace with MLIR API changes (mlir >= 15, as expected).
> Let's also ask authors: @multiverstack-intellif , @wrongtest-intellif

I've updated the PR to be consistent with all ```mlir >= 15```  releases.
Elaborated for: ```llvm/mlir 15.0.7```, ```llvm/mlir 16.0.6```, ```llvm/mlir 17.0.0rc3```.

multiverstack-intellif:
> > Thanks for the bugfixes! Definitely not an expert here, and do you think we need to add a test for changes introduced in presurger_set.cc?
> 
> @junrushao ,
> 
> Me either, not expert on this topic, unsure how #14690 passed it's own test (can't see related CI logs). Since here we have compilation fixes (no changes), the #14690 test case included should be sufficient.
> 
> Let's wait the tests to pass in green, reported issue here might be related to: **(a)** a different `mlir = 16` version on my side **(b)** a newer `gcc 13` not willing to infer casts (unplausible)
> 
> If **(a)** will adjust this PR with some little `#if version` to keep pace with MLIR API changes (mlir >= 15, as expected).
> 
> Let's also ask authors: @multiverstack-intellif , @wrongtest-intellif

I believe it's because of a), USE_MLIR is not turned ON in CI test, AFAIK, and only LLVM 14 & 15 is tested locally when  #14690 is submitted.

junrushao:
Sounds great, and it seems that we all agree with the changes being made! Will merge it in as soon as the CI is green.

cbalint13:
> 

@multiverstack-intellif , @junrushao 

Trying to tackle similar issue as described at [interblock analisys](https://github.com/multiverstack-intellif/tvm-rfcs/blob/improve_intset/rfcs/0099-introduce-PresburgerSet.md#interblock-analysis) I noticed it can't compile on newer mlir.
I think just like LLVM, MLIR will have frequent api changes across releases, so maintenance like this PR will be needed.

---

To refresh, this PR now covers (tested on ) llvm/mlir: ```15.0.7```, ```16.0.6``` and ```17.0.0rc3```.

junrushao:
Thanks @cbalint13! This is merged!

PR #15639, opened 2023-08-28
Do not link LLVM libraries into cpptest binary

kparzysz-quic:
They are already linked into libtvm.so, and cpptests links against libtvm.so, so they are not needed again. On the other hand, linking them twice can lead to the following error:
```
CommandLine Error: Option 'print-summary-global-ids' registered more than once!
LLVM ERROR: inconsistency in registered CommandLine options
```


junrushao (approved):
Looks great! Thanks for fixing this issue!

PR #15640, opened 2023-08-29
[Disco] Support Sharding via Scatter

junrushao:
WIP

### DIFF
--- a/src/runtime/disco/builtin.h
+++ b/src/runtime/disco/builtin.h
@@ -1,0 +1,80 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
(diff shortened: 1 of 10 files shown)


src/runtime/disco/builtin.h
```diff
@@ -0,0 +1,63 @@
...
+namespace tvm {
+namespace runtime {
+
+/*!
+ * \brief Load a runtime Module, then create and initialize a RelaxVM
+ * \param path The path to the runtime Module (a DSO file) to be loaded
+ * \param device The default device used to initialize the RelaxVM
+ * \return The RelaxVM as a runtime Module
+ */
+Module LoadVMModule(std::string path, Device device);
```
tqchen, line 37: These are all remote functions and likely should not be part of .h file. instead, only register them locally at worker and expose controller/session APIs

junrushao: I exposed them to be used by other worker-side methods. How about keeping them in a `builtin/` or `worker_builtin/` folder?

tqchen: worker_builtin might be ok, although ideally worker side builtin should be minimized, which means they can be contained as a single cc file. The main thing is to avoid being confused from controller side code

junrushao:
Continued in: #15652, #15653, #15654, #15655

PR #15642, opened 2023-08-29
[Unity][Frontend][NN] Collection of new NN operators

jwfromm:
This PR adds a bunch of new NN operators that conform to torch semantics. I also updated a couple of existing operators like LayerNorm to be more inline with the torch equivalent. Specifically this PR adds
* unsqueeze
* concat
* subtract
* Identity
* chunk
* pad
I also fixed an issue in ReshapeDataflowRewrite when operating on scalars.


junrushao (approved):
Awesome!

PR #15647, opened 2023-08-30
[Unity][Dlight] GeMV rule max_num_threads awareness

MasterJH5574:
Prior to this PR, the GeMV dlight rule is not aware of the maximum number of threads of the backend target. This PR brings the awareness. This is mainly for the purpose of WebGPU where each thread block can have 256 threads at most.


junrushao (approved):
LGTM!

PR #15648, opened 2023-08-31
[Relay][TOPI] Remove input padding for arm_cpu conv2d int8 native schedule in Legalize pass

Anndrey24:
The Legalize pass was unnecessarily padding the input channels for conv2d int8 native implementations. Since the conv2d schedule itself can add padding more efficiently, I skipped padding during the pass and further optimised the schedule to deal with it.  
I also added a test to check whether or not the Legalize pass pads the conv2d input data.  

The benchmark results of a single int8 conv2d operation where input padding along the K axis is necessary, with input = (1 x 224 x 224 x 3), filter = (16 x 3 x 3 x 3), output = (1 x 112 x 112 x 16), tested for the following target `"llvm --device=arm_cpu --mtriple=aarch64-linux-gnu -mattr=+v8.2a,+dotprod"` , before and after the changes, is given below:

|  Before (ms) | After (ms) | Speedup (%) |
|-------------|------------|-------------|
| 0.5186      | 0.0663     | 87.22       |

### DIFF
--- a/python/tvm/topi/arm_cpu/conv2d_gemm.py
+++ b/python/tvm/topi/arm_cpu/conv2d_gemm.py
@@ -328,12 +328,7 @@ def schedule_conv2d_gemm_interleaved(cfg, s, out, final_out):
 
     b, m, n = data_im2col.op.axis
     if data_im2col.op.name == "data_im2col":
-        n_size = data_im2col.shape[2]
-        if n_size % 16 == 0:
-            split_factor = 16
-        else:
-            split_factor = 8
-        n_outer, n_inner = s[data_im2col].split(n, split_factor)
+        n_outer, n_inner = s[data_im2col].split(n, 16)
         s[data_im2col].unroll(n_outer)
         s[data_im2col].vectorize(n_inner)
         b_m_fused = s[data_im2col].fuse(b, m)
@@ -418,7 +413,8 @@ def schedule_conv2d_gemm_native(cfg, s, out, final_out):
     b, x, y = C.op.axis
     (k,) = C.op.reduce_axis
     k_outer, k_inner = s[C].split(k, 16)
-    x_outer, y_outer, x_inner, y_inner = s[C].tile(x, y, x_factor=4, y_factor=16)
+    y_tile_size = 16
+    x_outer, y_outer, x_inner, y_inner = s[C].tile(x, y, x_factor=4, y_factor=y_tile_size)
     s[C].reorder(b, x_outer, y_outer, k_outer, x_inner, y_inner, k_inner)
     gemm_acc = gemm_acc_nx16_int8_int8_int32(in_type, rows=1)
     s[C].unroll(x_inner)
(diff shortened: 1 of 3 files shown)


python/tvm/topi/arm_cpu/conv2d_gemm.py
```diff
@@ -428,12 +435,29 @@ def schedule_conv2d_gemm_native(cfg, s, out, final_out):
 
     b, m, n = data_im2col.op.axis
     if data_im2col.op.name == "data_im2col":
-        n_outer, n_inner = s[data_im2col].split(n, 16)
+        if A.op.name == "A_padded_K":
```
lhutton1, line 438: would be good to add a comment here stating that if there are both "padding_K" and "padding_M" ops, this path will be selected

python/tvm/topi/arm_cpu/conv2d_gemm.py
```diff
@@ -428,12 +435,29 @@ def schedule_conv2d_gemm_native(cfg, s, out, final_out):
...
+            ic = data_im2col.op.input_tensors[0].shape[3]
+            split_factor = n_size // ic
+
+        n_outer, n_inner = s[data_im2col].split(n, split_factor)
         s[data_im2col].unroll(n_outer)
         s[data_im2col].vectorize(n_inner)
-        s[data_im2col].parallel(m)
     elif padding_A:
         s[data_im2col].compute_inline()
+        _, n_inner = s[A].split(A.op.axis[2], 16)
```
lhutton1, line 459: For the 16 here, let's make it a variable to maintain the link to https://github.com/apache/tvm/pull/15648/files#diff-42b1313a1be464c7f2c94f75d656be725f1ccb54b9391cd5b27c33009ac0e2d5R421

lhutton1 (commented):
Overall LGTM! I just had a couple of comments to improve readability of the schedules. It would also be good to check the schedule tests cover the cases that have been added

python/tvm/topi/arm_cpu/conv2d_gemm.py
```diff
@@ -326,7 +328,12 @@ def schedule_conv2d_gemm_interleaved(cfg, s, out, final_out):
 
     b, m, n = data_im2col.op.axis
     if data_im2col.op.name == "data_im2col":
-        n_outer, n_inner = s[data_im2col].split(n, 16)
+        n_size = data_im2col.shape[2]
+        if n_size % 16 == 0:
+            split_factor = 16
+        else:
+            split_factor = 8
+        n_outer, n_inner = s[data_im2col].split(n, split_factor)
```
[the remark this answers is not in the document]
Anndrey24: That does make more sense. I'll split it out into a separate PR

Anndrey24:
Regarding the tests, it seems that all of the padding cases are already visited at least once in: https://github.com/apache/tvm/blob/476d9d1d41980ad35cd6a4d044c17fea2210f352/tests/python/topi/python/test_topi_conv2d_int8.py#L110

lhutton1:
Awesome, thanks for checking!

lhutton1 (approved):
Thanks @Anndrey24! Will leave open for a bit incase @ekalda wants to take a look

lhutton1:
Thanks @Anndrey24!

Issue #15643, opened 2023-08-30
[Bug] [Ethos] apps/microtvm/ethosu/run_demo.sh gets wrong inference result


toyowata:
A run_demo.sh script in the apps/microtvm/ethosu is able to build microtvm executable binary for Ethos-U and kick to FVP to run the demo, but the inference result is wrong.

i.e. A cat image is classified as goldfinch.

### Expected behavior

The inference result is 'tabby'.

### Actual behavior

The inference result is 'goldfinch'.

FVP output:
```
telnetterminal0: Listening for serial connection on port 5000
telnetterminal1: Listening for serial connection on port 5001
telnetterminal2: Listening for serial connection on port 5002
telnetterminal5: Listening for serial connection on port 5003

    Ethos-U rev fa6e5f88 --- Aug  9 2021 09:54:41
    (C) COPYRIGHT 2019-2021 Arm Limited
    ALL RIGHTS RESERVED

Starting Demo
I: Initializing NPU: base_address=0x48102000, fast_memory=0x0, fast_memory_size=0, secure=1, privileged=1
I: Soft reset NPU
I: New NPU driver registered (handle: 0x0x20000de0, NPU: 0x0x48102000)
Running inference
I: Acquiring NPU driver handle
D: ethosu_reserve_driver(): NPU driver handle 0x20000de0 reserved
D: ethosu_invoke_async(): OPTIMIZER_CONFIG
I: Optimizer release nbr: 0 patch: 1
I: Optimizer config. product=0, cmd_stream_version=0, macs_per_cc=8, shram_size=48, custom_dma=0
I: Optimizer config. arch version: 1.0.6
I: Ethos-U config. product=0, cmd_stream_version=0, macs_per_cc=8, shram_size=48, custom_dma=1
I: Ethos-U. arch version=1.1.0
D: ethosu_invoke_async(): NOP
D: ethosu_invoke_async(): NOP
D: ethosu_invoke_async(): NOP
D: ethosu_invoke_async(): COMMAND_STREAM
I: handle_command_stream: cmd_stream=0x60025be0, cms_length 2663
I: Soft reset NPU
D: ethosu_dev_set_clock_and_power(): CMD=0x00000000
D: ethosu_dev_run_command_stream(): QBASE=0x0000000060025be0, QSIZE=10652, cmd_stream_ptr=0x60025be0
D: ethosu_dev_run_command_stream(): BASEP0=0x0000000060028580
[...]

PR #15649, opened 2023-08-31
Add output_data_sec section in corstone300.ld

toyowata:
Fixed output_data_sec section was missing in the corstone300.ld for Ethos-U.

See more detail: https://github.com/apache/tvm/issues/15643

### DIFF
--- a/apps/microtvm/ethosu/corstone300.ld
+++ b/apps/microtvm/ethosu/corstone300.ld
@@ -138,6 +138,7 @@ SECTIONS
   {
     . = ALIGN(16);
     *(ethosu_scratch)
+    *(output_data_sec)
     . = ALIGN (16);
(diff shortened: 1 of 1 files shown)


lhutton1:
Thanks for creating the PR @toyowata. cc @ekalda @ashutosh-arm could you take a look? Perhaps we can do it in a follow-up PR, but I think the `task_demo_microtvm.sh` script could check the correct prediction was made to prevent regression in the future

ashutosh-arm:
I don't have much idea of the new flag. It will be good to get it reviewed offline from someone who knows more about linker scripts.

lhutton1 (approved):
Thanks @toyowata LGTM!

lhutton1:
Thanks @toyowata @ashutosh-arm!

PR #15650, opened 2023-08-31
[Unity][Frontend][NN] Make effects optional in nn module.

jwfromm:
This PR adds the `debug` argument to `export_tvm`. When `debug` is `False`, effects are not included in the output graph. This can make deploying models less cumbersome since its not often theyll use effects. I also added automatic annotation of the `num_inputs` function attribute since it is useful for a few passes.

### DIFF
--- a/python/tvm/relax/frontend/nn/core.py
+++ b/python/tvm/relax/frontend/nn/core.py
@@ -394,6 +394,7 @@ class Module(SubroutineMixin):
         device: str = "cpu",
         pipeline: str = "zero",
         out_format: str = "torch",
+        debug: bool = False,
     ) -> Callable:
         """Just-in-time compilation of a nn.model to an executable"""
         from tvm import relax  # pylint: disable=import-outside-toplevel
@@ -402,7 +403,7 @@ class Module(SubroutineMixin):
 
         # Convert nn.Module to IRModule
         spec = _spec.ModuleSpec.from_raw(spec, self)
-        mod, params = _spec.SpecBuilder().build(spec)
+        mod, params = _spec.SpecBuilder().build(spec, debug=debug)
 
         # Convert parameters
         device = _str_to_device(device)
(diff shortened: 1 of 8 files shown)


python/tvm/relax/frontend/nn/spec.py
```diff
@@ -322,15 +324,18 @@ def _extern_modules() -> List[Tuple[str, List[str]]]:
         # pylint: enable=protected-access
 
         params = _params()
-        effects = _effects()
+        effects = _effects() if debug else None
```
junrushao, line 327: Let's only set `io_effect` to None if `debug` flag is off, because there could be other effects, for example, kv cache

junrushao: by setting `self.io_effect` to None in `SpecBuilder.__init__`

jwfromm: This is great feedback and makes perfect sense. I've updated the code so that debug only turns off io effects based on your suggestion. Can you give it another look and see if this seems reasonable to you?

python/tvm/relax/frontend/nn/core.py
```diff
@@ -360,12 +360,31 @@ def to(self, dtype: Optional[str] = None) -> None:  # pylint: disable=invalid-na
     def export_tvm(
         self,
         spec: "_spec.ModuleSpecType",
+        debug: bool = False,
```
junrushao, line 363: Shall we also add this `debug` flag to the `jit` method?

junrushao (approved):
LGTM. A minor comment. Feel free to skip as a follow-up PR

python/tvm/relax/frontend/nn/core.py
junrushao: Oh actually you will need the `debug` flag in line 405

jwfromm: Good point, now added.

PR #15652, opened 2023-09-01
[Runtime] ShapeTuple.Product and ShapeTuple Printing

junrushao:
This PR adds two convenient methods for `ShapeTuple`.

```C++
// Returns the number of elements in the shape,
// i.e. the product of all dimensions
ShapeTupleObj::index_type ShapeTupleObj::Product();

// Printing method for shape
std::ostream& operator<<(std::ostream& os, const ShapeTuple& shape);
```

### DIFF
--- a/include/tvm/runtime/container/shape_tuple.h
+++ b/include/tvm/runtime/container/shape_tuple.h
@@ -24,6 +24,7 @@
 #ifndef TVM_RUNTIME_CONTAINER_SHAPE_TUPLE_H_
 #define TVM_RUNTIME_CONTAINER_SHAPE_TUPLE_H_
 
+#include <ostream>
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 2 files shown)


junrushao:
CC: @tqchen

include/tvm/runtime/container/shape_tuple.h
```diff
@@ -42,6 +43,9 @@ class ShapeTupleObj : public Object {
   /*! \brief The size of the shape tuple object. */
   uint64_t size;
 
+  /*! \brief Get number of elements in the shape */
+  index_type Numel() const;
```
tqchen, line 47: trying to dig a bit about related APIs. How about Product? the product of the shape values

tqchen: numel makes sense for an array api. e.g. array.numel. Just not sure if it is right to use it for shape.

junrushao: How about NumElements()?

junrushao: OK. I don't have strong opinion to use `Product`. It's just `numel` is a well-known convention, but `Product` is a newly-invented word with ambiguity, but I don't have strong opinion.

tqchen (approved):

PR #15653, opened 2023-09-01
[Disco] Add `Scatter-From-Worker0`

junrushao:
This PR introduces an API:

```C++
void ScatterFromWorker0(Optional<NDArray> send, NDArray recv);
```

This mimics the MPI-style scattering primitive that divides NDArray
`send` into chunks, and send them to each worker's NDArray `recv`.

### DIFF
--- a/src/runtime/disco/builtin.cc
+++ b/src/runtime/disco/builtin.cc
@@ -16,15 +16,13 @@
  * specific language governing permissions and limitations
  * under the License.
  */
+#include "./builtin.h"
+
+#include <dlpack/dlpack.h>
 #include <tvm/runtime/container/shape_tuple.h>
-#include <tvm/runtime/data_type.h>
 #include <tvm/runtime/disco/session.h>
-#include <tvm/runtime/module.h>
-#include <tvm/runtime/ndarray.h>
-#include <tvm/runtime/object.h>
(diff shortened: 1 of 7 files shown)


junrushao:
CC: @tqchen

src/runtime/disco/builtin.cc
```diff
@@ -72,38 +70,48 @@ Module LoadVMModule(std::string path, Device device) {
...
-      pf->CallPacked(args, rv);
+NDArray AllReduce(NDArray send, ReduceKind reduce_kind) {
+  return GetCCLFunc("allreduce")(send, static_cast<int>(reduce_kind));
+}
+
+NDArray BroadcastFromWorker0(NDArray buffer) {
+  return GetCCLFunc("broadcast_from_worker0")(buffer);
+}
+
+void ScatterFromWorker0(const Array<NDArray>& buffers) {
```
tqchen, line 94: API choice, let us avoid using scatter. Since MPI scatter have a specific expectation (scatter from a single buffer).

Instead, let us use SendToWorkers

junrushao: I will replace it with a more standard MPI-style scatter

tqchen (approved):

PR #15661, opened 2023-09-03
[Unity][DLight] Improve the performance of matmul on Adreno GPU

Hzfengsy:
The current matmul implementation is targeted for NVIDIA GPU, which is not optimized for Adreno GPU (also other edge devices). This patch improves the performance, improving performance from 7170.14ms to 554.57ms for a very large kernel


Hzfengsy:
cc @spectrometerHBH

spectrometerHBH (approved):

PR #15662, opened 2023-09-03
[Unity][Frontend][NN] Add nn.MultiLinear

junrushao:
This PR introduces `nn.MultiLinear` as an extension to `nn.Linear` which is arithmetically equivalent to applying multiple `nn.Linear` to the same input.

For example, QKV projection in attention, represented in `nn.Linear` being:

```python
q_proj = nn.Linear(hidden_dim, ...)
k_proj = nn.Linear(hidden_dim, ...)
v_proj = nn.Linear(hidden_dim, ...)

q = q_proj(x)
k = k_proj(x)
v = v_proj(x)
```

could be simplified using `nn.MultiLinear` as:

```python
qkv_proj = nn.MultiLinear((hidden_dim, hidden_dim, hidden_dim), ...)
q, k, v = qkv_proj(x)
```

The underlying implementation, by default, will fuse all the projection matrices into a single one, to allow potentially larger room for optimization.

### DIFF
--- a/python/tvm/relax/frontend/nn/modules.py
+++ b/python/tvm/relax/frontend/nn/modules.py
@@ -18,13 +18,15 @@
 """Builtin Modules."""
 from typing import List, Optional, Sequence, Union
 
+import numpy as np
+
 from tvm import relax as rx
 from tvm import tir
 from tvm._ffi import register_func
 from tvm.runtime import NDArray
 
 from . import op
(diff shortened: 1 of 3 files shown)


junrushao:
CC: @MasterJH5574 @jwfromm

Hzfengsy (approved):

PR #15665, opened 2023-09-04
[Arith] Fix detect non-divisible iteration form like (x % 255) // 16

wrongtest-intellif:
Previously, the optimization from `floordiv(floormod(..))` to `floormod(floordiv(..))` do not check the divisibility. Which may create wrong  iteration form. The change add a default handling and let it just fail in followup validity checks.

For example, ` (x % 255) // 16` should not get transformed to `(x // 16) % 15`. The change make it as `(x % 255) // 16 % ceildiv(255, 16)`.

It currently could not pass the surjective check though the mapping actually is surjective. We could improve it in the future.

### DIFF
--- a/src/arith/iter_affine_map.cc
+++ b/src/arith/iter_affine_map.cc
@@ -1897,10 +1897,18 @@ PrimExpr IterMapRewriter::SplitFloorDivConst(IterSplitExpr lhs, PrimExpr base, P
   // = floormod(sc2+t, c2)
   // = floormod(floordiv(y, c1), c2)
   // = floormod(floordiv(iter, lower_factor*c1), c2), where c1=rhs, c2=extent/rhs
-  IterSplitExpr new_split(padded->source,
-                          /* lower_factor = */ padded->lower_factor * rhs,
-                          /* extent = */ analyzer_->Simplify(floordiv(padded->extent, rhs)),
-                          /* scale = */ padded->scale);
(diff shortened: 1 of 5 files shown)


Hzfengsy (approved):

Hzfengsy:
Thanks for the fix!

vinx13:
do we need another `IterMark`? I think in this case we can rewrite to `(x // 16) % ceildiv(255, 16)`

Lunderberg:
@vinx13 I don't think that simplification would hold.  If I use `x==-16`, `(x % 255) // 16` evaluates to 14, while `(x // 16) % ceildiv(255, 16)` evaluates to 15.

That said, if we know that `x` is on the range `0 <= x < 255`, such that `x % 255 == x`, then the simplification would be valid.

vinx13:
The simplification @Lunderberg mentioned should work. Actually we can get the extent from the source `IterMark` of the split.
Say we want to compute `floordiv(lhs, rhs)` where `lhs == floormod(floordiv(x, c1), ext)`, `x == IterMark(extent=ext_x)`, and `ext % rhs != 0`.

if we can prove `ext_x // c1 \in [0, ext)`, it can be simplified as `floordiv(lhs, rhs) == floordiv(x, c1 * rhs) == IterSplit(x, lower_factor=c1*rhs, extent=ceildiv(ext, rhs)`

PR #15668, opened 2023-09-04
[CI] Allow Limit CPUs in Docker

junrushao:
This PR adds a new flag `--cpus` in `./docker/bash.sh`, which is passed to docker command that allows limiting the number of CPU cores of a docker container.

Related materials: https://docs.docker.com/config/containers/resource_constraints/#configure-the-default-cfs-scheduler


junrushao:
CC: @tqchen

tqchen (approved):

PR #15673, opened 2023-09-05
[Disco] Add LoadAll method to Disco Shard Loader

jinhongyii:
This PR adds a `LoadAll` method to the Disco shard loader to load
parameters all at once.

### DIFF
--- a/src/runtime/disco/loader.cc
+++ b/src/runtime/disco/loader.cc
@@ -46,6 +46,8 @@ class ShardLoaderObj : public Object {
                           TypedPackedFunc<void(DLTensor*, int, DLTensor*)> f_shard);
   /*! \brief Load the i-th parameter */
   NDArray Load(int weight_index) const;
+  /*! \brief Load all the parameters */
+  Array<NDArray> LoadAll() const;
   /*! \brief Slice the given tensor at a specific dimension */
   NDArray Shard(NDArray source, int dim, int num_slices) const;
 
@@ -63,6 +65,8 @@ class ShardLoaderObj : public Object {
   NDArrayCacheMetadata metadata_;
   /*! \brief Sharding information for each weight */
   std::vector<ShardInfo> shard_info_;
+  /*! \brief Maps the name of a shard to its index */
+  std::unordered_map<std::string, int> param_name_to_index_;
   /*! \brief A method to slice a 3-D tensor */
   TypedPackedFunc<void(DLTensor*, int, DLTensor*)> f_shard_;
   /*! \brief The current file opened to load weights in it */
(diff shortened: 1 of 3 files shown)


jinhongyii:
cc: @junrushao @tqchen

src/runtime/disco/loader.cc
```diff
@@ -93,9 +93,18 @@ inline std::vector<ShapeTuple::index_type> ShardShape(const ShapeTuple& shape, i
   return result;
 }
 
+class ShardInfoCompare {
+  public:
+    bool operator()(const ShardLoaderObj::ShardInfo& a, const ShardLoaderObj::ShardInfo& b) const {
+      int a_id = std::stoi(a.param->name.substr(6));
+      int b_id = std::stoi(b.param->name.substr(6));
+      return a_id < b_id;
```
junrushao, line 101: you don't need an actual "sort", instead you may retrieve the parameter names in ndarray-cache-metadata and calculate their names correspondingly by "params_0", "params_1", etc

src/runtime/disco/loader.cc
```diff
@@ -43,7 +43,7 @@ class ShardLoaderObj : public Object {
   /*! \brief Create a shard loader. */
   static ObjectRef Create(const std::string& path_to_metadata, const std::string& metadata,
                           const std::string& shard_info,
-                          TypedPackedFunc<void(DLTensor*, int, DLTensor*)> f_shard);
+                          PackedFunc f_shard);
```
junrushao, line 46: Why this change?

jinhongyii: TypedPackedFunc cannot be null

src/runtime/disco/loader.cc
```diff
@@ -187,5 +231,16 @@ TVM_REGISTER_GLOBAL("runtime.disco.ShardLoaderLoad")
       return loader->Load(IntegerFromShapeTuple(weight_index));
     });
 
+TVM_REGISTER_GLOBAL("runtime.disco.ShardLoaderLoadAll").set_body_typed([](ObjectRef loader_obj) {
+  const auto* loader = loader_obj.as<ShardLoaderObj>();
+  CHECK(loader != nullptr) << "TypeError: Expected ShardLoaderObj, but gets: "
+                           << loader_obj->GetTypeKey();
+  Array<NDArray> shards;
+  for (int i = 0; i < static_cast<int>(loader->shard_info_.size()); i++) {
```
junrushao, line 239: does `shard_info_.size()` imply all parameters?

junrushao:
To facilitate PR merge, I made the related changes accordingly which will save @jinhongyii's time

PR #15674, opened 2023-09-05
[Unity] Add `has_function` to RelaxVM

jinhongyii:
add a "has_function" function in executable

### DIFF
--- a/include/tvm/runtime/disco/session.h
+++ b/include/tvm/runtime/disco/session.h
@@ -155,7 +155,7 @@ class DRefObj : public Object {
  */
 class DRef : public ObjectRef {
  public:
-  TVM_DEFINE_MUTABLE_OBJECT_REF_METHODS(DRef, ObjectRef, DRefObj);
+  TVM_DEFINE_MUTABLE_NOTNULLABLE_OBJECT_REF_METHODS(DRef, ObjectRef, DRefObj);
 };
 
 /*!
@@ -195,18 +195,6 @@ class SessionObj : public Object {
    * \param host_array The array to be copied to worker-0
    * \param remote_array The NDArray on worker-0
    */
-
-  DRef CopyToWorker0Wrapper(const NDArray& host_array) { 
-    
-    Device dev {DLDeviceType(0), 0};
-    auto func = this->GetGlobalFunc("runtime.disco.empty");
-    ShapeTuple shape = host_array.Shape();
-    DataType dtype = host_array.DataType();
-    DRef dref = this->CallPacked(func, shape, dtype ,dev);
-    this->CopyToWorker0(host_array, dref);
-    return dref;
-  }
-
   virtual void CopyToWorker0(const NDArray& host_array, const DRef& remote_array) = 0;
   /*!
    * \brief Synchrnoize the controler with a worker, and it will wait until worker finishes
@@ -248,7 +236,7 @@ class Session : public ObjectRef {
  public:
   /*! \brief Create a session backed by a thread pool of workers */
   static Session ThreadedSession(int num_workers);
-  TVM_DEFINE_MUTABLE_OBJECT_REF_METHODS(Session, ObjectRef, SessionObj);
+  TVM_DEFINE_MUTABLE_NOTNULLABLE_OBJECT_REF_METHODS(Session, ObjectRef, SessionObj);
 };
 
 /*!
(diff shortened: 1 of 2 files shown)


jinhongyii:
cc: @junrushao @tqchen

include/tvm/runtime/disco/session.h
```diff
@@ -195,6 +195,18 @@ class SessionObj : public Object {
    * \param host_array The array to be copied to worker-0
    * \param remote_array The NDArray on worker-0
    */
+
+  DRef CopyToWorker0Wrapper(const NDArray& host_array) {
```
tqchen, line 199: To minimize the interface, let us consider only expose remote.Empty.

Then move the wrapper to the upper level applications

src/runtime/relax_vm/executable.cc
```diff
@@ -75,6 +75,10 @@ PackedFunc Executable::GetFunction(const String& name, const ObjectPtr<Object>&
       vm->LoadExecutable(GetObjectPtr<Executable>(this));
       *rv = Module(vm);
     });
+  } else if (name == "func_exists") {
```
tqchen, line 78: LGTM

include/tvm/runtime/disco/session.h
```diff
@@ -236,7 +248,7 @@ class Session : public ObjectRef {
  public:
   /*! \brief Create a session backed by a thread pool of workers */
   static Session ThreadedSession(int num_workers);
-  TVM_DEFINE_MUTABLE_NOTNULLABLE_OBJECT_REF_METHODS(Session, ObjectRef, SessionObj);
+  TVM_DEFINE_MUTABLE_OBJECT_REF_METHODS(Session, ObjectRef, SessionObj);
```
junrushao, line 251: Why this change?

include/tvm/runtime/disco/session.h
junrushao: `Empty` is exposed on python: https://github.com/apache/tvm/blob/unity/python/tvm/runtime/disco/session.py#L117-L142.

Shall we migrate some of the python pieces to C++?

include/tvm/runtime/disco/session.h
```diff
@@ -155,7 +155,7 @@ class DRefObj : public Object {
  */
 class DRef : public ObjectRef {
  public:
-  TVM_DEFINE_MUTABLE_NOTNULLABLE_OBJECT_REF_METHODS(DRef, ObjectRef, DRefObj);
+  TVM_DEFINE_MUTABLE_OBJECT_REF_METHODS(DRef, ObjectRef, DRefObj);
```
junrushao, line 158: If there isn't a particular reason, please keep objects as "not-nullable" as possible

include/tvm/runtime/disco/session.h
jinhongyii: In MLC we want to make Session a field. To initialize it, we are using ThreadSession(worker_num) instead of its constructor to create a session, so it has to be nullable to compile. Another reason is that in definition of DRef, we are already having  
``` /*! \brief Back-pointer to the host controler session */
  ObjectRef session{nullptr};
```
So there's no reason session can't be nullable.

junrushao: To be clear, it's not-nullable since it's properly created. If you wanted to cover the case where Session could be a field of another TVM Object, the common paradigm we are using across TVM codebase is to mark it as "initially null":

```C++
Session sess{nullptr};
```

Note that this pattern is widely used in MetaSchedule already.

There are multiple folds of reasons for using not-nullable, as TVM object system will enforce more checking, and also developers of other modules could feel very safe to use an object without worrying it's null.

jinhongyii: sounds good

include/tvm/runtime/disco/session.h
jinhongyii: If we want to expose in c++, that's only 2 lines of code. So I don't think it's necessary.

jinhongyii:
please review @junrushao @tqchen

tqchen (approved):

src/runtime/relax_vm/executable.cc
tqchen, line 78: thinking a bit more, how about rename to has_function

tqchen (commented):
one minor nit

src/runtime/relax_vm/executable.cc
junrushao: sounds good. I'm updating the name on my end to accelerate PR merge

Issue #14070, opened 2023-02-21
[Tracking Issue][Unity] Relax Layout Planning


psrivas2:
### This issue is to track progress for Layout Planning in Relax as per the proposal [here](https://github.com/tlc-pack/relax/issues/277)
- [x] P1. Add layout_transform operator in Relax (https://github.com/tlc-pack/relax/pull/403)
- [x] P2. Add PrimFunc level analysis pass to suggest layout transform to blocks and buffers based on layout transforms on outputs (https://github.com/apache/tvm/pull/14066)
- [x] P3. Add AlterOpLayout pass that allows users to modify the layout of certain operators. (https://github.com/apache/tvm/pull/14215)
- [ ] P4. Add OptimizeLayout pass that uses P1 & P2 to flow layouts across ops and cancel/fuse adjacent layout transforms.
- [ ] P5. Extend P2 to handle implicit padding in output buffer transformations.

#### Future Possibilities
- Using P2 to get operator level layout planning pass as discussed in comments in [proposal](https://github.com/tlc-pack/relax/issues/277)
- Extend P2 to handle multiple write buffers in a block.
- Extend P2 to handle operators such as `upsampling` whose buffer access are difficult to analyze.

cc @Hzfengsy @junrushao @quic-sanirudh

PR #15678, opened 2023-09-06
[UNITY][Pass] Optimize redundant layout transform ops

abhikran-quic:
Relax `AlterOpImpl` pass introduces `layout_transform` operations. If the layouts match for consecutive `layout_transform` operations, they can be cancelled out. This pass tries to optimize redundant `transform_layout` operations.

This is also addressing P4 in https://github.com/apache/tvm/issues/14070

### DIFF
--- a/python/tvm/relax/transform/optimize_layout_transform.py
+++ b/python/tvm/relax/transform/optimize_layout_transform.py
@@ -14,13 +14,12 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-# pylint: disable=invalid-name, unused-argument
+# pylint: disable=invalid-name, unused-argument, redefined-argument-from-local
 """Relax Optimize Layout Transform pass."""
-import tvm
 from tvm.ir.module import IRModule
 from tvm.ir.transform import PassContext
 from tvm.relax import Expr, Function
-from tvm.relax.dpl import *
+from tvm.relax.dpl import is_op, rewrite_call, wildcard
 from . import function_pass
 
 
@@ -29,18 +28,6 @@ class OptimizeLayoutTransform:
     """
     Pass to remove redundant transform layout operators
     introduced by AlterOpImpl pass.
-
-    Parameters
-    ----------
-    func: Expr
-        The relax function to be optimized
-
-    mod: IRModule
-        The ir module
-
-    ctx: PassContext
-        Relax pass context
-
     """
 
     def __init__(self):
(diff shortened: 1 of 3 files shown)


abhikran-quic:
cc: @psrivas2 @masahi @junrushao

python/tvm/relax/transform/optimize_layout_transform.py
```diff
@@ -0,0 +1,147 @@
...
+from typing import Union
+import tvm
+from tvm import IRModule, relax
+from tvm.ir.transform import module_pass
+from tvm.relax.analysis import remove_all_unused
+from tvm.relax.expr_functor import mutator, PyExprMutator
+
+
+@mutator
+class OptimizeLayoutTranformMutator(PyExprMutator):
```
masahi, line 28: Can we use the built-in pattern rewriter instead?

https://github.com/apache/tvm/blob/unity/python/tvm/relax/dpl/rewrite.py#L26-L28

abhikran-quic: Thanks @masahi for sharing this. 
I had gone through this earlier but I wasn't able to find a rewriter that can work on a complete IRModule. Hence, I implemented my own class. 
Do you know of a function that can take IRModule and rewrite all the patterns of interest ? 
If there is no such function, I can work on it. Basically it'll be a wrapper on top of `rewrite_call` to iterate over the whole graph.

masahi: Why do you need a module-based pass for this rewriting? I think our pattern rewriter works at the function level partly because that's where the actual rewriting happens.

abhikran-quic: Alright. I understand the concern here.  I will rewrite the pass as a function pass.

python/tvm/relax/transform/optimize_layout_transform.py
```diff
@@ -0,0 +1,75 @@
...
+"""Relax Optimize Layout Transform pass."""
+from tvm.ir.module import IRModule
+from tvm.ir.transform import PassContext
+from tvm.relax import Expr, Function
+from tvm.relax.dpl import is_op, rewrite_call, wildcard
+from . import function_pass
+
+
+@function_pass(opt_level=0)
+class OptimizeLayoutTransform:
```
masahi, line 27: The name "OptimizeLayoutTransform" sounds too generic for the simple thing it does. I'd suggest introducing "RemoveNop" pass and add this pattern as a first step.

abhikran-quic: The reason for keeping the name as `OptimizeLayoutTransform` was to optimize two `layout_transform` operations into a single one when the index maps don't match. 
Right now the implementation is optimizing our redundant `layout_transform` operations when index_maps are matching. 
Eventually, we plan to extend this pass to converge two `transform_layout` operations into a single one.

masahi: ok makes sense

masahi (approved):

PR #15679, opened 2023-09-06
[Unity] Support Padding Reversal in Alter-Op pass

rutkoor:
This patch adds support for non-bijective transformations in alter-op-impl pass. The idea is to introduce remove_pad operator, which copies non-padded buffer entries from padded_buffer. This additional copy operation will suffice as a placeholder for supporting non-bijective transformations. Further we can eliminate this copy operation in later stages of compilation pipeline.

The intermediate remove_pad operators that are created can be cancelled out by using OptimizeLayoutTransform pass(https://github.com/apache/tvm/pull/15678).


rutkoor:
@tvm-bot rerun

rutkoor:
@tvm-bot re-run

rutkoor:
@tvm-bot re-build

rutkoor:
@tvm-bot rerun

rutkoor:
cc: @psrivas2 @masahi @junrushao

rutkoor:
Gentle reminder to please review this PR.

python/tvm/relax/transform/legalize_ops/manipulate.py
```diff
@@ -205,3 +213,16 @@ def te_layout_transform(data, name):
...
+
+@register_legalize("relax.remove_pad")
+def _remove_pad(bb: BlockBuilder, call: Call) -> Expr:
+    orig_shape = call.attrs.orig_shape
+
+    def te_remove_pad(data):
+        """
+        Returns a new compute that restrict the original expression to the shape of orig_shape
+        """
+        return te.compute(orig_shape, data, name="te_remove_pad")
```
masahi, line 226: How do you know that this removes the correct padded values?

rutkoor: Hi @masahi, currently the elements are padded in the end of the original buffer. So, when we iterate over orig_shape(dimensions of non-padded buffer), we can be sure that we will be iterating over non-padded entries.
The one assumption that I have taken here is that padding will always be happening in the end of the buffer.
In future, if AlterOp introduces padding in-between original buffer then we should modify remove_pad operator to reflect the same.

masahi: > currently the elements are padded in the end of the original buffer

Isn't that a specific behavior of the layout transform pass? Here you are defining a general function that undoes padding, and padding can be anything along the begin / end of each axis.

rutkoor: Technically the behavior of remove_pad can be extended to remove padding along the given axis and given offset. In this PR, the support is added for AlterOp pass explicitly, to remove the padding which is being introduced by layout transform pass. In the next PR, I plan to work on generalizing remove_pad functionality to be generic.

Will it be good idea to separate out this PR from generalizing remove_pad functionality?

masahi: We shouldn't introduce a half-baked op definition that is meant to be used only by a certain pass. The proper order is that a generic op definition should come first.

masahi: Better yet, is this actually needs to be an "op"? Since the op definition is so trivial, how about moving this compute definition to `alter_op_layout.cc` and use it there? You can remove all boilerplate and test cases required for a Relax op definition.

masahi: For example, https://github.com/mlc-ai/mlc-llm uses weight quantize / dequantize functions (called `encode / decode`), RMS norm etc implemented in TE without introducing Relax ops for them. They are represented by `call_tir`.

rutkoor: Thanks @masahi for your helpful suggestions. In the latest changes I have defined remove_pad operator inside `alter_op_impl.cc`.
As of now, the test case in this PR fails because the `te::compute` introduces min Range value to `int32(0)`, whereas when reading `Expected` relax testcase the min Range value is being read as `int64(0)`.
Please let me know what I should do with min Range problem.

masahi: Can you show the error?

rutkoor: Hi @masahi, the error looks something like this,

`ValueError: StructuralEqual check failed, caused by lhs at <root>.functions[I.GlobalVar("remove_pad")].body.block.body.body.block.iter_vars[0].dom.min.dtype and rhs at <root>.functions[I.GlobalVar("remove_pad")].body.block.body.body.block.iter_vars[0].dom.min.dtype`

where the lhs dtype is coming out to be `int32` when running the `alter_op_impl `pass, and rhs dtype is `int64` which is coming from `Expected module`.

I have commited a change in `src/te/operation/compute_op.cc` which is resolving the issue.
My question is, does the change in `src/te/operation/compute_op.cc` should go as a separate PR or can I have it in this PR only?

masahi: If all tests pass, I think it is fine to include such change in this PR.

python/tvm/relax/transform/legalize_ops/manipulate.py
```diff
@@ -182,7 +182,15 @@ def te_layout_transform(data, name):
         )
 
     index_map: tvm.tir.IndexMap = call.attrs.index_map
-    pad_value = call.attrs.pad_value.value
+    pad_value = call.attrs.pad_value
+    if pad_value is not None:
+        pad_value = pad_value.value
+    else:
+        if "int" in call.args[0].struct_info.dtype:
+            pad_value = int(pad_value)
```
masahi, line 190: pad_value is None here

src/relax/transform/alter_op_impl.cc
```diff
@@ -176,16 +190,41 @@ class AlterOpImplMutator : public ExprMutator {
...
+          },
+          "output", topi::kElementWise);
+
+      String op_name = "remove_pad";
+      PrimFunc remove_pad_with_frozen_layout =
+          WithAttr(CreatePrimFunc({placeholder_tensor, output_tensor}), kOperatorName, op_name);
+
+      GlobalVar gv_remove_pad = builder_->AddFunction(remove_pad_with_frozen_layout, op_name);
+      builder_->UpdateFunction(gv_remove_pad,
+                               WithoutAttr(remove_pad_with_frozen_layout, "global_symbol"));
```
masahi, line 224: Code between L210 - L224 is hard to read and it seems it doesn't need to be defined here. I think you can define `gv_remove_pad` with dynamic input shape once at the constructor and use that repeatedly here.

python/tvm/relax/transform/legalize_ops/manipulate.py
rutkoor: Corrected.

src/relax/transform/alter_op_impl.cc
rutkoor: I have introduced `GetOrCreateRemovePadOp` function which returns the `GlobalVar` of the `remove_pad PrimFunc` if it is present in `remove_pad_map_` map variable.

src/relax/transform/alter_op_impl.cc
```diff
@@ -314,6 +377,8 @@ class AlterOpImplMutator : public ExprMutator {
   Map<PrimFunc, GlobalVar> cache_;
   /*! \brief Input IRModule */
   const IRModule& mod_;
+  /*! \brief Map from shape_dim attribute to the remove_pad GlobalVar */
+  Map<String, GlobalVar> remove_pad_map_;
```
masahi, line 381: Why not `std::unordered_map<int, GlobalVar>'?

rutkoor: Using `std::unordered_map<int, GlobalVar>` now.

masahi (approved):

PR #15681, opened 2023-09-06
[TFLite][Frontend] Support for quantized SQUARE

tlopex:
The commit enables convert_square function for the quantized tensor. The op SQUARE can be tested by previous testing.

cc @leandron

### DIFF
--- a/include/tvm/runtime/container/variant.h
+++ b/include/tvm/runtime/container/variant.h
@@ -1,0 +1,123 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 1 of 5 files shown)


leandron (changes requested):
@tlopex thanks for the PR! I see it is failing on a formatting error. Please have a look at https://tvm.apache.org/docs/contribute/pull_request.html on the linting steps.

Apart from that, can you add tests to this, so that the quantised version of `SQUARE` is tested?

jikechao (changes requested):
Thanks for the PR! @tlopex 
You should correct the code and add a corresponding test case.
Specifically, You need to remove the OPNotImplementaion warning in Lines 1313-1316.
In addition, the patch can be simplified to avoid redundant statements. (e.g., `input_tensors = self.get_input_tensors(op)` in Line 1295 and 1304)

PR #15682, opened 2023-09-06
[Relay] Fuse clip operators in yolov5

youxiudeshouyeren:
When I specify the ReLU6 activation function in yolov5， This redundancy occurs.
For example:
image
It generates this IR：
```
  %233 = nn.conv2d(%231, meta[relay.Constant][54] /* ty=Tensor[(256, 256, 1, 1), int8] */, padding=[0, 0, 0, 0], channels=256, kernel_size=[1, 1], out_dtype="int32") /* ty=Tensor[(1, 256, 16, 16), int32] */;
  %234 = add(%233, meta[relay.Constant][55] /* ty=Tensor[(256, 1, 1), int32] */) /* ty=Tensor[(1, 256, 16, 16), int32] */;
  %235 = fixed_point_multiply(%234, multiplier=2097356939, shift=-6) /* ty=Tensor[(1, 256, 16, 16), int32] */;
  %236 = clip(%235, a_min=-128f, a_max=127f) /* ty=Tensor[(1, 256, 16, 16), int32] */;
  %237 = cast(%236, dtype="int8") /* ty=Tensor[(1, 256, 16, 16), int8] */;
  %238 = clip(%237, a_min=0f, a_max=48f) /* ty=Tensor[(1, 256, 16, 16), int8] */;
```

The redundancy pattern is：
```
 Example:
  %1 = clip(%0, a_min=0f, a_max=255f) [type=int32]
  %2 = cast(%1, dtype="uint8") [type=uint8]
  %3 = clip(%2, a_min=20f, a_max=66f) [type=uint8]
```

It can be optimized to:
```
 %1 = clip(%0, a_min=20f, a_max=66f) [type=int32]
 %2 = cast(%1, dtype="uint8") [type=uint8]
```

This PR is to simplify `clip->cast->clip` to `clip->cast`

### DIFF
--- a/src/relay/transforms/simplify_expr.cc
+++ b/src/relay/transforms/simplify_expr.cc
@@ -327,6 +327,56 @@ static std::vector<int> GetTransposeAxisOrder(const Call& call, int ndim) {
 }
 
 /*!
+ * \brief SimplifyClipCastClip matches the pattern clip->cast->clip and fuse clips based on Clip
+ *    min/max values.
+ *
+ * Example:
+ *   %1 = clip(%0, a_min=0f, a_max=255f) [type=int32]
+ *   %2 = cast(%1, dtype="uint8") [type=uint8]
+ *   %3 = clip(%2, a_min=20f, a_max=66f) [type=uint8]
+ *
+ * Optimized to (fuse Clips):
+ *   %1 = clip(%0, a_min=20f, a_max=66f) [type=uint8]
+ *   %2 = cast(%1, dtype="uint8") [type=uint8]
+ */
+class SimplifyClipCastClip : public DFPatternRewrite {
+ public:
+  SimplifyClipCastClip() {
+    data_ = IsWildcard();
+    clip1_ = IsOp("clip")({data_});
+    cast_ = IsOp("cast")({clip1_});
+    pattern_ = IsOp("clip")({cast_});
+  }
+
+  Expr Callback(const Expr& pre, const Expr& post,
+                const Map<DFPattern, Array<Expr>>& node_map) const override {
+    auto clip1 = Downcast<Call>(node_map[clip1_][0]);
+    const CallNode* clip1_node = clip1.as<CallNode>();
+    const ClipAttrs* clip1_attrs = clip1_node->attrs.as<ClipAttrs>();
+
+    auto cast = Downcast<Call>(node_map[cast_][0]);
+    const CallNode* cast_node = cast.as<CallNode>();
+    DataType cast_type = Downcast<TensorType>(cast_node->checked_type())->dtype;
+
+    auto clip2 = Downcast<Call>(post);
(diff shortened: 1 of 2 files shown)


youxiudeshouyeren:
@tvm-bot re-run

src/relay/transforms/simplify_expr.cc
```diff
@@ -326,6 +326,56 @@ static std::vector<int> GetTransposeAxisOrder(const Call& call, int ndim) {
...
+ *
+ * Example:
+ *   %1 = clip(%0, a_min=0f, a_max=255f) [type=int32]
+ *   %2 = cast(%1, dtype="uint8") [type=uint8]
+ *   %3 = clip(%2, a_min=20f, a_max=66f) [type=uint8]
+ *
+ * Optimized to (fuse Clips):
+ *   %1 = clip(%0, a_min=20f, a_max=66f) [type=uint8]
+ *   %2 = cast(%1, dtype="uint8") [type=uint8]
+ */
```
jikechao, line 341: Thanks for the PR! @youxiudeshouyeren. This optimization behavior seems not thorough enough; It can be further optimized to  `%1 = clip(%0, a_min=20f, a_max=66f) [type=uint8]`.

jikechao (commented):
It seems that this example should have been optimized by `SimplifyClip`/`SimplifyClipAndConsecutiveCast` passes, but it was not executed. Perhaps you need to prioritize checking why existing passes have not been optimized to this pattern.

src/relay/transforms/simplify_expr.cc
youxiudeshouyeren: I realize it was my example that was written incorrectly, the first data type is int32. It should be:
```
 * Optimized to (fuse Clips):
 *   %1 = clip(%0, a_min=20f, a_max=66f) [type=int32]
 *   %2 = cast(%1, dtype="uint8") [type=uint8]
 */
```
I'll correct the example in the comment.

youxiudeshouyeren:
> Thanks for the PR! @youxiudeshouyeren. It seems that this example should have been optimized by `SimplifyClip`/`SimplifyClipAndConsecutiveCast` passes, but it was not executed. Perhaps you need to prioritize checking why existing passes have not been optimized to this pattern.

They are different optimizations and both have different patterns. SimplifyClipAndConsecutiveCast targets clip(int32)-cast(uint8)->cast(int32). This simplification targets clip(int32)->cast(int8)->clip(int8). I'm sure the pass SimplifyClipAndConsecutiveCast is enabled, but it doesn't eliminate this redundancy in yolov5.

jikechao:
@youxiudeshouyeren Thanks for your explanation! 
I mean that this redundancy may be caused by a bug in the implementation of `SimplifyClip`. 
Definitely, adding a brand new pass in this PR can solve this problem. However, It will be better to fix this bug rather than add the new pass.

Issue #15282, opened 2023-07-10
[Bug] tensor type Tensor[(1), float64] has 1 dimensions, while float64 has 0 dimension


jikechao:
torch.nn.functional.instance_norm(args[0], use_input_stats=True) and input_dtype=float64  wil lead to a crash: 
**Error: tensor type `Tensor[(1), float64]` has 1 dimensions, while `float64` has 0 dimension**

### Actual behavior
```
Traceback (most recent call last):
  File "test.py", line 20, in <module>
    mod, params = relay.frontend.from_pytorch(trace, input_shapes)
  File "/workplace/software/tvm/tvm_/python/tvm/relay/frontend/pytorch.py", line 5002, in from_pytorch
    outputs = converter.convert_operators(operator_nodes, outputs, ret_name)
  File "/workplace/software/tvm/tvm_/python/tvm/relay/frontend/pytorch.py", line 4263, in convert_operators
    self.record_output_type(relay_out)
  File "/workplace/software/tvm/tvm_/python/tvm/relay/frontend/pytorch.py", line 238, in record_output_type
    self.infer_type_with_prelude(output)
  File "/workplace/software/tvm/tvm_/python/tvm/relay/frontend/pytorch.py", line 174, in infer_type_with_prelude
    body = self.infer_type(val, self.prelude.mod)
  File "/workplace/software/tvm/tvm_/python/tvm/relay/frontend/pytorch.py", line 167, in infer_type
    new_mod = transform.InferType()(new_mod)
  File "/workplace/software/tvm/tvm_/python/tvm/ir/transform.py", line 160, in __call__
    return _ffi_transform_api.RunPass(self, mod)
  File "/workplace/software/tvm/tvm_/python/tvm/_ffi/_ctypes/packed_func.py", line 237, in __call__
    raise get_last_ffi_error()
tvm._ffi.base.TVMError: Traceback (most recent call last):
  10: TVMFuncCall
[...]

PR #15683, opened 2023-09-06
Fix a bug caused by PyTorch instance_norm when the input shape is [1,1,1,2]

haoyang9804:
Fix [Issue 15282](https://github.com/apache/tvm/issues/15282)


haoyang9804:
cc @vvchernov @echuraev

jikechao (approved):
LGTM, thanks for your fixing.

python/tvm/relay/frontend/pytorch.py
```diff
@@ -4424,7 +4424,10 @@ def _create_typed_const(data, dtype):
     dtype should be a TVM dtype"""
 
     if dtype == "float64":
-        typed_data = _expr.const(np.float64(data), dtype=dtype)
+        if len(data) == 1:
+            typed_data = _expr.const(np.array([np.float64(data)]), dtype=dtype)
+        else:
+            typed_data = _expr.const(np.float64(data), dtype=dtype)
```
echuraev, line 4430: I read your updates in the issue, thank you for detailed description. Here it looks like a workaround above numpy bug. I think that the best way to fix this issue is to create an issue or a PR in the numpy repo with fix.

Currently, this implementation won't work after fixing the bug in numpy. My suggestion is to create an issue or PR with fix to numpy repository and at the same time change the current implementation in the following way:
```suggestion
        typed_data = np.float64(data)
        if np.isscalar(typed_data):
            typed_data = _expr.const(np.array([typed_data]), dtype=dtype)
        else:
            typed_data = _expr.const(typed_data, dtype=dtype)
```
This code should work also after fixing the bug in Numpy.

echuraev (commented):
Thank you for your PR!

python/tvm/relay/frontend/pytorch.py
haoyang9804: Thanks! I have submitted a issue to Numpy community [here](https://github.com/numpy/numpy/issues/24661).

echuraev: Thank you! In this case, something like `typed_data = _expr.const(np.asarray(data, dtype="float64"), dtype=dtype)` might be the best solution for this problem.

python/tvm/relay/frontend/pytorch.py
```diff
@@ -4424,7 +4424,11 @@ def _create_typed_const(data, dtype):
     dtype should be a TVM dtype"""
 
     if dtype == "float64":
-        typed_data = _expr.const(np.float64(data), dtype=dtype)
+        typed_data = np.float64(data)
+        if np.isscalar(typed_data):
+            typed_data = _expr.const(np.asarray(data, dtype="float64"), dtype=dtype)
+        else:
+            typed_data = _expr.const(typed_data, dtype=dtype)
```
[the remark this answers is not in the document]
haoyang9804: True. Thanks for your reminder. Have changed the patch

echuraev (approved):
LGTM. Thanks!

PR #15686, opened 2023-09-06
[Unity] Add new Relax annotation ops: smooth and absmax

ibsidorenko:
This commit adds 2 new Relax ops:
1. "relax.op.smooth" (R.smooth): Multiply elements from a tensor by a scale (if it operates like "multiply") or pass input as is (if it operates like "identity").
2. "relax.annotate.absmax" (R.absmax): Helper op, that does the following transform: abs()->max()->squeeze() for the input tensor.

Both of them are used in SmoothQuant TVM passes used in MLC-llm.

This PR does not include unit tests for these new ops, since they are single-purpose ops and not widely used. This way we can save time on regression testing.


python/tvm/relax/transform/legalize_ops/annotate.py
```diff
@@ -0,0 +1,47 @@
...
+from ...expr import Call, Expr
+from .common import register_legalize
+
+
+@register_legalize("relax.annotate.smooth")
+def _smooth(bb_: BlockBuilder, call: Call) -> Expr:
+    if call.attrs.mode == "identity":
+        return bb_.call_te(topi.identity, call.args[0])
+
+    assert call.attrs.mode == "multiply"
```
masahi, line 30: What if `mode == "quantize"`?

masahi:
I'm taking a look at https://github.com/mlc-ai/mlc-llm/pull/855 and it makes me wonder if we need those two ops at all. The `identity` mode in smooth seems odd and unnecessary, and `R.absmax` is legalized immediately before stat collection.

ibsidorenko:
> I'm taking a look at [mlc-ai/mlc-llm#855](https://github.com/mlc-ai/mlc-llm/pull/855) and it makes me wonder if we need those two ops at all. The `identity` mode in smooth seems odd and unnecessary, and `R.absmax` is legalized immediately before stat collection.

Thank you for review! I saw your comments there. Rather interesting idea and looks like it is possible to avoid from using these ops. I have added there in comments about motivation of adding `R.absmax`

python/tvm/relax/transform/legalize_ops/annotate.py
ibsidorenko: This is helper op for SmoothQuant. In case of `mode == "quantize"` it should be lowered by SmoothQuantLegalizer.
mode == "quantize" was introduced only due to absence of **QNN** dialect.

python/tvm/relax/op/annotate.py
```diff
@@ -0,0 +1,63 @@
...
+        The input data
+
+    scale : relax.Expr
+        Scale multiplier
+
+    kind : int
+        Kind of argument to be annotated. Can be one of: activation or weight.
+
+    mode : str
+        Execution mode for the op. Can be one of: "identity", "multiply" or "quantize".
```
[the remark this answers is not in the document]
ibsidorenko: Yes, this is intentional. Legalization for mode == "quantize" was not implemented. Ideally, I need `qnn.quantize` here

yongwww:
@ibsidorenko thanks for the effort! looks we might not need use these two ops, shall we close this pr?

ibsidorenko:
> @ibsidorenko thanks for the effort! looks we might not need use these two ops, shall we close this pr?

Yes, I think we can close this PR

PR #15688, opened 2023-09-06
[Unity] Delegate DataflowVar visitor to Var by default

Lunderberg:
Prior to this commit, writing a subclass of `relax::ExprVisitor` or `relax::ExprMutator` required separate overrides for visiting a `relax::DataflowVar` and a `relax::Var`.  In the majority of cases, these two types should be treated identically, and failure to handle a `DataflowVar` equivalently would be a bug.

This commit updates the `relax::ExprVisitor` and `relax::ExprMutator` base classes to visit `DataflowVar` by delegate to the visitor of `relax::Var`.  As a result, any derived class that overrides the `relax::Var` visitor will also update the behavior for `relax::DataflowVar`.  A pass that requires different behavior for `relax::Var` and `relax::DataflowVar` can still explicitly override both methods in order to provide different behavior.

### DIFF
--- a/src/relax/analysis/analysis.cc
+++ b/src/relax/analysis/analysis.cc
@@ -94,8 +94,6 @@ class VarVisitor : protected ExprVisitor {
 
   void VisitExpr_(const VarNode* var) final { vars_.Insert(GetRef<Var>(var)); }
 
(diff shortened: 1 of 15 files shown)


slyubomirsky (approved):
Good idea. I don't see a way that this can cause problems and it's a nice simplification, so I'm for it.

PR #15692, opened 2023-09-07
fixup resize2d dynamic input support

leiwen83:
If Resize take shape like input's shape is dynamic one, and we use scales sampling mode, then it would report error like below, for currently relax onnx frontend still deem all shape in input a const value. Then the dynamic tir.Var would has no value to extract out.

AttributeError: <class 'tvm.tir.expr.SizeVar'> has no attribute value

Use tir expr to get calculation result for scaled size.


python/tvm/relax/frontend/onnx/onnx_frontend.py
```diff
@@ -1302,8 +1303,10 @@ def _impl_v18(cls, bb, inputs, attr, params):
         if scales is not None:
             assert isinstance(scales, relax.Constant), "Only constant scales currently supported."
             scales = scales.data.numpy()
-            sizes_shape = [dim.value for dim in x.struct_info.shape]
-            sizes = (sizes_shape * scales)[2:].astype("int64").tolist()
+            sizes = []
+            for i, dim in enumerate(x.struct_info.shape):
+                sizes.append(T.cast(scales[i] * dim, "int64"))
```
Hzfengsy, line 1308: It would be great to use `tir.cast` instead of `T.cast` as the later one is designed for TVMScript for most cases

Hzfengsy (approved):
Overall LGTM

python/tvm/relax/frontend/onnx/onnx_frontend.py
leiwen83: Change from T.cast to directly use cast which is imported from tvm.tir.generic

Hzfengsy:
Please make the CI green :)

leiwen83:
> Please make the CI green :)

All CI green now ~

PR #15694, opened 2023-09-07
[Unity] Implemented SameShapeConstraint for dataflow pattern matches

Lunderberg:
Prior to this commit, a shape could be explicitly specified using the `ShapePattern`, but could not be specified relative to the shape of another expression.  As a result, patterns with restricted shapes became very difficult to

This commit implements `SameShapeConstraint`, which can be applied between any patterns that participate in the match.  For example, matching against `R.add(lhs,rhs)` where `lhs` and `rhs` have the same non-broadcasted shape.  Because these constraints operate between patterns that do not necessarily share a consumer/producer relationship, they could not previously be expressed using the existing `PairCons` functionality, and were instead implemented in terms of a new `DFConstraint` base class.

### DIFF
--- a/src/relax/ir/dataflow_matcher.cc
+++ b/src/relax/ir/dataflow_matcher.cc
@@ -22,6 +22,7 @@
  * \brief The dataflow pattern matcher for Relax.
  */
 
+#include <tvm/arith/analyzer.h>
 #include <tvm/node/structural_equal.h>
 #include <tvm/relax/analysis.h>
 #include <tvm/relax/dataflow_matcher.h>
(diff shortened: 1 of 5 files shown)


masahi:
@ganler

src/relax/ir/dataflow_matcher.cc
```diff
@@ -663,11 +758,51 @@ static std::optional<MatchState> TryMatch(const PNode& p, const RNode& r,
   return new_match;
 }
 
+static std::optional<MatchState> TryValidate(
+    const MatchState& current_match,
+    const std::unordered_map<const DFPatternNode*, PNode>& pattern2node,
+    const std::vector<DFConstraint>& validation_constraints,
+
```
masahi, line 765: Remove?

masahi (approved):
Very interesting!

src/relax/ir/dataflow_matcher.cc
```diff
@@ -443,6 +444,92 @@ bool DFPatternMatcher::VisitDFPattern_(const ShapePatternNode* op, const Expr& e
   return false;
 }
 
+Optional<Bool> SameShapeConstraintNode::IsConstraintSatisfied(
```
ganler, line 447: Just curious if the return value will always be concrete. Is there any cases that the returned `Bool` is a predicate expression?

src/relax/ir/dataflow_matcher.cc
```diff
@@ -443,6 +444,92 @@ bool DFPatternMatcher::VisitDFPattern_(const ShapePatternNode* op, const Expr& e
...
+  // even if some shapes are still unknown
+  // (e.g. SameShapeConstraint(A,B,C) can return false if A and B have
+  // different shapes, even if C is unknown).  The passing constraint
+  // is only valid if all of the shapes are known.
+  if (all_shapes_defined) {
+    // If all shapes are known and have the same dimentionality, then
+    // we just need to prove that the sizes of each dimension match.
+    // If we cannot prove it at this point, we won't get more
+    // information later.
+    return Bool(analyzer->CanProve(all_dimensions_equal));
```
ganler, line 518: If it cannot be proved, shall we return `NullOpt` instead since "unproved" does not mean "must be unequal"?

ganler (approved):
This is a very cool and useful feature! Besides syntactical and topological constraints, we now also support some more complicated semantic constraints!

More importantly in the future, I think we may even customize such semantic constraints given user-provided predicate functions.

src/relax/ir/dataflow_matcher.cc
Lunderberg: Thank you, and the empty line is now removed.

src/relax/ir/dataflow_matcher.cc
Lunderberg: In the current API, the return `Optional<Bool>` can only contain true/false/nullopt.  The potential use case for having a full predicate expression be returned would be to allow constraints to interact with each other.

Currently, the API uses the three possible values of `Optional<Bool>` to decide whether to backtrack due to a failed condition (`Bool(false)`), mark a condition as satisfied (`Bool(true)`), or to mark a condition as uncheckable (`NullOpt`).  If we instead returned a `std::tuple<PrimExpr, bool>`, where the `PrimExpr` gives a necessary condition for the constraint to be satisfied, and the boolean indicates whether the condition is also sufficient for the constraint to be satisfied.  That would allow constraints to respond that the constraint is satisfied, so long as the returned expression evaluates to true.  The pattern matcher could then take the AND of all the necessary conditions, checking if the entire expression can be validated.

For example, suppose we have tensors `A: R.Tensor(shape=[16, 16])` and `B: R.Tensor(shape=[16, n])`.  The current pattern matcher is conservative, and cannot verify that `SameShapeConstraint([pat_A, pat_B])` should match, because it cannot prove that `n==16`.  Returning a `PrimExpr` would allow the calling scope to determine whether this should impose additional constraints by binding the value of `n`, or to treat it as a mismatch.

src/relax/ir/dataflow_matcher.cc
Lunderberg: The pattern matcher should be conservative in the case of ambiguity, as this may be used to enable/disable rewrite rules whose validity depends on the pattern being matched.  Previously, I had been thinking of the return value as "yes", "no", and "ask me again later".  The "ask me again later" option is no longer valid at this point, as no further shape information will be received.  As a result, the shape cannot be proven equal, and the pattern match should fail.

This would be another argument for returning `(PrimExpr condition, bool condition_is_sufficient)` from this function, to avoid the potential confusion here.  If this choice of backtracking/attempting to combine with other expressions/etc is made at the calling scope, this function wouldn't need to distinguish between these uses of `Bool(false)` and `NullOpt`, and could instead return `{condition, true}`.

ganler: Thanks for the clarification. The rationale makes sense to me.

src/relax/ir/dataflow_matcher.cc
ganler: Oh I see. Computing a complement constraint can be very useful to debug the existance of optmizations. In the exemplary case you demonstrated, we may indicate/warn the users that if additional constraints/invariants can be satisfied, larger set of optimizations could be applied.

Lunderberg: And updated to use a `AsPrimExpr` method to convert a constraint to an expression that must evaluate to true, rather than returning an `Optional<Bool>`.

csullivan:
Thanks @Lunderberg @ganler @masahi, this is merged.

PR #15696, opened 2023-09-07
[Unity] Unwrap tuples during CanonicalizeBindings

Lunderberg:
Prior to this commit, `CanonicalizeBindings` would identify trivial bindings of `var_y = var_x` and replace subsequent usage of `var_y` with `var_x`, but bindings of the form `tuple_var = (var_x,); var_y = var_x[0]` would not be canonicalized.

This commit updates the `CanonicalizeBindings` pass to identify trivial bindings that occur across a `TupleGetItem` expression, handling them as if they were direct assignments.


slyubomirsky (approved):
Great change, thank you!

PR #15697, opened 2023-09-07
[Unity] Mark result of LazyTransformParams as impure function

Lunderberg:
The function produced by `LazyTransformParams` has only side effects, through the `"get_item"` and `"set_item"` PackedFunc calls, and should be marked as impure.

### DIFF
--- a/tests/python/relax/test_transform_lazy_transform_params.py
+++ b/tests/python/relax/test_transform_lazy_transform_params.py
@@ -74,9 +74,8 @@ def test_lazy_transform_params():
                     T.writes(out[o, i, h, w])
                     out[o, i, h, w] = w1[i, o, h, w]
 
-        @R.function
+        @R.function(pure=False)
         def main_transform_params() -> R.Tuple:
-            R.func_attr({"relax.force_pure": True})
             cls = Expected
             lv: R.Object = R.call_packed("get_item", R.prim_value(1), sinfo_args=(R.Object,))
             _: R.Object = R.call_packed("set_item", R.prim_value(0), lv, sinfo_args=(R.Object,))
@@ -137,10 +136,8 @@ def test_lazy_transform_params_with_symbolic_vars():
 
     @I.ir_module
     class Expected:
-        @R.function
+        @R.function(pure=False)
         def main_transform_params(slice_shape_expr: R.Shape(["slice_index"])):
-            # we expect ToNonDataflow and RemovePurityTracking to be invoked first
-            R.func_attr({"relax.force_pure": True})
             cls = Expected
 
             slice_index = T.int64()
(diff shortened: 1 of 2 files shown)


tests/python/relax/test_transform_lazy_transform_params.py
```diff
@@ -137,10 +136,8 @@ def slice_buffer(
 
     @I.ir_module
     class Expected:
-        @R.function
+        @R.function(pure=False)
         def main_transform_params(slice_shape_expr: R.Shape(["slice_index"])):
-            # we expect ToNonDataflow and RemovePurityTracking to be invoked first
```
slyubomirsky, line 142: Where does `LazyTransformParams` fall in the phase ordering? Do we expect it to be used early or late in compilation?

Lunderberg: Currently, it isn't present in the lowering flow, and is explicitly called by an end-user.  (Based on grep for `"LazyTransformParams"` across the TVM repo.)  While the comment suggests that it was intended to be used in a later stage, so far this seems like just an extra requirement for an end-user to perform, before calling a function intended for an end-user's convenience (e.g. [here](https://github.com/mlc-ai/mlc-llm/blob/main/mlc_llm/utils.py#L219)).  By outputting the correct purity annotations, we avoid this requirement.

That said, looking at it again, for correctness, any `DataflowBlock` in which the mutation occurs should be broken apart into a regular `BindingBlock`.

slyubomirsky: Okay, if it's not specifically meant for later stages in compilation, then this change is definitely appropriate.

slyubomirsky: What side effects would `get_item` have? I can't find where it's defined. Seems a little unusual for a read to be impure.

slyubomirsky (approved):

tests/python/relax/test_transform_lazy_transform_params.py
Lunderberg: It wouldn't be impure due to side effects, but impure due to failure to return the same result for each execution.  Because the model weights are no longer expressed as input arguments, they rely on external state.

slyubomirsky: Ah, that's interesting. I would still call it pure since it's not modifying any state (as distinct from, say, a call to a random number generator that modifies the internal state of the generator), but the dependency on external state is a potential bit of trickiness.

Lunderberg: Hmm.  The analogy I'd see would be calls to a `get_current_time()` function.  Calling the function doesn't modify any of the internal state of the system, but still wouldn't be a pure function due to the dependency on external state.

slyubomirsky: Ah, that's a good point. Most definitions of purity consider a function referring to external state not to be pure, so I think you're right. I should emphasize this in the specification, since I had neglected it myself.

PR #15698, opened 2023-09-07
[Unity][Analysis] Check for usage of DataflowVar in all_vars()

Lunderberg:
Prior to this commit, the `VarVisitor` used in the implementation of `all_vars` and `free_vars` only collected variable usage sites where the variable was a `const VarNode*`, and ignored usage sites of a `const DataflowVarNode*`.  When analyzing an entire function, these variables were found in the `const VisitVarBinding*`, and could pass the existing tests.  However, when analyzing a single expression, these variables would be erroneously excluded.

This commit adds a `VisitExpr_(const DataflowVarNode*)` implementation in `VarVisitor`, to collect variable usage regardless of the type of variable.

### DIFF
--- a/src/relax/analysis/analysis.cc
+++ b/src/relax/analysis/analysis.cc
@@ -94,6 +94,8 @@ class VarVisitor : protected ExprVisitor {
 
   void VisitExpr_(const VarNode* var) final { vars_.Insert(GetRef<Var>(var)); }
 
+  void VisitExpr_(const DataflowVarNode* var) final { vars_.Insert(GetRef<Var>(var)); }
+
   void VisitExpr_(const FunctionNode* op) final {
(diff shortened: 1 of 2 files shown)


Lunderberg:
This bug was the motivation for the more general fix in https://github.com/apache/tvm/pull/15688, which delegates from `VisitExpr_(const DataflowVarNode*)` to `VisitExpr_(const VarNode*)` by default.

Hzfengsy (approved):

Hzfengsy:
Should we close #15688 as this got merged?

Lunderberg:
That said, #15688 should be rebased on top of this PR, so that it can also remove the more specific fix as it will no longer be necessary.

PR #15699, opened 2023-09-07
[Unity] Propagate extra symbolic vars through LiftTransformParams

Lunderberg:
Prior to this PR, a symbolic variable could be provided to the `LiftTransformParams`, which could then be used to produce the transformed parameters.  For example, providing the `rank` and `world_size` of a parameter set in a multi-GPU setup.  However, these symbolic variables may be left undefined in the inference function.

This PR updates `LiftTransformParams` to track the symbolic variables that are definable by the inference function, and to propagate any symbolic variables that would otherwise be undefined.

### DIFF
--- a/src/relax/transform/lift_transform_params.cc
+++ b/src/relax/transform/lift_transform_params.cc
@@ -31,6 +31,8 @@
 #include <iostream>
 #include <vector>
 
+#include "../../support/ordered_set.h"
+
 namespace tvm {
 namespace relax {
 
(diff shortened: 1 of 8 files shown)


src/relax/transform/lift_transform_params.cc
```diff
@@ -43,26 +43,92 @@ struct LiftTransformParamsInfoPlan {
       lifted_bindings;  // the bindings of the original function that are lifted
 };
 
+namespace {
+template <typename T>
+struct InsertionOrderSet {
```
vinx13, line 48: I implemented something similar here https://github.com/apache/tvm/blob/unity/src/support/ordered_set.h

Lunderberg: Thank you, and I'll switch this PR over to it.

Though, related to it, `std::list` is usually implemented as a linked list, and has very poor performance as a result.  Is it currently being used in a context where the less-frequent iterator invalidation of `std::list` is required?

vinx13: That’s a good point. I agree vector is better when deletion is infrequent.

Lunderberg: And the PR is now updated to use `support::OrderedSet`.  I did need to make a few adjustments to it, so that it could support both hashable types (using `std::unordered_map<T, ValueType>`) and TVM objects (using `std::unordered_map<T, ValueType, ObjectPtrHash, ObjectPtrEqual>`), but that seemed like good functionality to have overall.

PR #15703, opened 2023-09-07
[Unity][TVMScript] Produce var = R.ExternFunc("") statements

Lunderberg:
Prior to this commit, any `ExternFunc` usage in a relax function would print the string name of the function on its own line, omitting any variable definition, and later use of the variable would occur without a definition.  This commit updates the printing of `R.ExternFunc` to appear as a normal relax variable.


junrushao (approved):
This is great change! Previously we don't have a proper format for ExternFunc, and this PR has addressed this issue perfectly. Thanks for your contribution @Lunderberg!

PR #15705, opened 2023-09-08
[TVMScript][FIX] Disable `show_object_address` in printing by default

Hzfengsy:
The new feature of printing object addresses is introduced in #15579. It's an excellent feature for debugging, but it should not be turned on by default.

cc @Lunderberg @junrushao


junrushao (approved):
LGTM. Looks like a small typo

Lunderberg:
Thank you for the fix.  This was intended to be disabled by default.

PR #15710, opened 2023-09-08
[TOPI] Ensure vectorization of input padding in `arm_cpu` int8 conv2d interleaved schedule

Anndrey24:
When padding the input data, the int8 conv2d interleaved schedule tries to split the `data_im2col` cols axis by a factor of 16 in order to then vectorize over those splits. However, the size of the axis is `n_size = KH x KW x IC` and the `Legalize` pass only pads the number of input channels up to a multiple of 8. Therefore, `n_size` is only guaranteed to be a multiple of 8, not 16.   

I modified the schedule to check whether a split factor of 16 is appropriate, otherwise use 8 instead, in order to ensure vectorization is performed in all cases.

cc @ekalda @lhutton1 @leandron @neildhickey

### DIFF
--- a/python/tvm/topi/arm_cpu/conv2d_gemm.py
+++ b/python/tvm/topi/arm_cpu/conv2d_gemm.py
@@ -326,7 +326,12 @@ def schedule_conv2d_gemm_interleaved(cfg, s, out, final_out):
 
     b, m, n = data_im2col.op.axis
     if data_im2col.op.name == "data_im2col":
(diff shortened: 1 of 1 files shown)


lhutton1 (approved):
LGTM! I don't see a way this change can easily be tested. However, functionally both paths should already be covered under https://github.com/apache/tvm/blob/93356cdcbe2160b90827fdcb4c4a88f765d6dd03/tests/python/topi/python/test_topi_conv2d_int8.py#L74

PR #15711, opened 2023-09-08
[Bugfix][Strategy] Fix `arm_cpu` int8 conv2d strategy for dotprod and i8mm targets

Anndrey24:
Whenever both dotprod and i8mm were available together on a target (e.g. `"llvm --device=arm_cpu --mtriple=aarch64-linux-gnu -mattr=+v8.2a,+dotprod,+i8mm"`), the native int8 conv2d implementation corresponding to the `+dotprod` attribute would be selected, but the compute definition of the conv2d operation would be constructed for the `+i8mm` attribute and its related interleaved schedule instead.  
The reason for this was a different order of conditional statements being used in 2 separate files:
 - `arm_cpu.py`: When selecting the conv2d implementation, the program first checked for `dotprod` support. If present, it chose the native schedule
 - `conv2d_gemm.py`: when constructing the compute definition, `i8mm` support is checked first, then `dotprod` 
 
To fix this, I modified the int8 conv2d strategy to prioritize `i8mm` over `dotprod` when both are available too. I also added new test cases to cover this situation.  

cc @ekalda @lhutton1 @leandron @neildhickey

### DIFF
--- a/python/tvm/relay/op/strategy/arm_cpu.py
+++ b/python/tvm/relay/op/strategy/arm_cpu.py
@@ -213,19 +213,35 @@ def conv2d_strategy_arm_cpu(attrs, inputs, out_type, target):
                 is_aarch64 = target.features.is_aarch64
                 has_asimd = target.features.has_asimd
                 has_dot_prod = target.features.has_dotprod
+                has_matmul_i8 = target.features.has_matmul_i8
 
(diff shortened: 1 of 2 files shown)


lhutton1 (approved):
Thanks @Anndrey24, LGTM!

PR #15712, opened 2023-09-08
[Unity] Clear slots before each shape lowering

vinx13:
cc @sunggg @cyx-6


sunggg (approved):

cyx-6 (approved):
Thanks @vinx13! It really helps!

PR #15713, opened 2023-09-08
[Unity] Dynamic-shape param support in LazyTransformParams

MasterJH5574:
This PR brings the support for dynamic-shape parameters to pass LazyTransformParams.

Prior to this PR, the symbolic variables in the dynamic-shape parameters are not properly popped out. This PR uses MatchCast to make sure the symbolic variables are always popped out and thereby support the dynamic-shape parameters.

This PR also fixes a previvously failed test.


junrushao (approved):
LGTM!

PR #15720, opened 2023-09-10
[Disco] Make Session.CallWithPacked public

junrushao:
This PR contains necessary tweaks to run Llama2 in Tensor Parallelism.


src/relax/transform/fuse_tir.cc
```diff
@@ -848,6 +848,14 @@ std::vector<size_t> GetTupleAccessedIndices(const FunctionNode* func, const Var&
 class TIRFuseMutator : public ExprMutator {
  public:
   static IRModule Transform(const IRModule& mod) {
+    Map<String, BaseFunc> tir_funcs;
```
junrushao, line 851: This changes allows TIRs with global symbols to be preserved if they are not called by relax - this is useful for the shard loader.

jinhongyii (approved):

PR #15725, opened 2023-09-12
Revert "[CodeGenC] Handle GlobalVar callee as internal function call"

junrushao:
This PR temporarily reverts a recent change that breaks the Metal backend. Happy to get this commit back once the codegen is fixed.

Reverts apache/tvm#15103

### DIFF
--- a/python/tvm/topi/arm_cpu/mprofile/dsp/micro_kernel/avg_pool.py
+++ b/python/tvm/topi/arm_cpu/mprofile/dsp/micro_kernel/avg_pool.py
@@ -55,7 +55,7 @@ def intrin_sum(shape, in_dtype, out_dtype, reset=False):
             ib = tvm.tir.ir_builder.create()
             ib.emit(
                 tvm.tir.call_extern(
-                    "int32",
+                    cc.dtype,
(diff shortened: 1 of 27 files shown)


junrushao:
Related discussion: https://github.com/apache/tvm/pull/15103#issuecomment-1714790711.

CC: @Lunderberg @MasterJH5574

MasterJH5574 (approved):

MasterJH5574:
Thank you @junrushao! @Lunderberg I’m sorry for this. Please help take a look at the issue!

csullivan:
Prior to merging I would request that we please include a test or otherwise link to a test that fails so that there is something semi-minimal for which can be used to support the desired use case and prevent the breakage in the future.

junrushao:
Agreed @csullivan. Given it's a codegen test, it's a bit challenging though to have a minimal test if we don't have unittest infra that could potentially access Metal-capable devices. As suggested by @MasterJH5574 [in the prior thread](https://github.com/apache/tvm/pull/15103#issuecomment-1714790711), how about using tests in `tests/python/unittest/test_allreduce_cuda.py` for Metal? They are around 10-15 lines of code in core.

PR #15735, opened 2023-09-12
[Unity][disco] Change collective communication to destination-passing style 

jinhongyii:
This eliminates extra malloc in collective communication op


jinhongyii:
cc: @junrushao  @tqchen

PR #15736, opened 2023-09-13
[Unity] Skip shape checking on transformed params

vinx13:
cc @masahi @sunggg


vinx13:
seems we can't simply skip the weights. we need to check whether weights are dynamic, and only skip checking weights with static shape as for dynamic shapes we need to update shape heap

jwfromm (approved):
LGTM, thanks Wuwei!

PR #15742, opened 2023-09-13
[Unity][Disco] separate computation and communication into 2 stream

jinhongyii:
put computation on default stream and put communication on a new stream.

### DIFF
--- a/src/runtime/disco/nccl/nccl.cc
+++ b/src/runtime/disco/nccl/nccl.cc
@@ -137,7 +137,8 @@ void ScatterFromWorker0(Optional<NDArray> send, NDArray recv) {
     NCCL_CALL(ncclGroupStart());
     uint8_t* data = static_cast<uint8_t*>(buffer->data);
     for (int i = 0; i < num_workers; ++i) {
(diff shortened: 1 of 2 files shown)


jinhongyii:
cc: @junrushao

junrushao:
i'm trying to figure out - which stream is used for compute and which is for communication? are we using null stream for compute?

jinhongyii:
null stream (or default stream equivalently) is used for compute. The stream created in init_ccl is for communication

src/runtime/disco/nccl/nccl.cc
```diff
@@ -91,26 +90,34 @@ void AllReduce(NDArray send, ReduceKind reduce_kind, NDArray recv) {
   NCCLThreadLocalContext* ctx = NCCLThreadLocalContext::Get();
   ShapeTuple shape = send.Shape();
   int64_t numel = shape->Product();
+  Device device = ctx->worker->default_device;
+  DeviceAPI::Get(device)->SyncStreamFromTo(device, nullptr, ctx->stream);
```
junrushao, line 94: one minor comment: please add a field `ctx->compute_stream` and rename `ctx->stream` to `ctx->comm_stream`

jinhongyii: makes sense

junrushao (approved):

PR #15745, opened 2023-09-14
[OpenCL] Don't initialize OpenCL runtime on host

echuraev:
After adding OpenCL wrapper, it is possible to build TVM with OpenCL support also on the host which doesn't have OpenCL libraries. But if you want to compile OpenCL module for a remote device on such host machine then you will see an error that OpenCL lib cannot be open.

To avoid such problem, we need to call OpenCL functions only in runtime. So function for initializing OpenCL workspace was removed from OpenCLModuleNode. And a new function `IsProgramCreated` was added. The last function is necessary to prepare vectors with OpenCL programs, associated with OpenCL devices. Previously it was done during OpenCLModule initialization. So, now we create such vectors only in runtime after getting list of available OpenCL devices.

### DIFF
--- a/src/runtime/opencl/opencl_module.cc
+++ b/src/runtime/opencl/opencl_module.cc
@@ -215,6 +215,7 @@ bool OpenCLModuleNode::IsProgramCreated(const std::string& func_name, int device
   auto dev_size = GetGlobalWorkspace()->devices.size();
   ICHECK(device_id < static_cast<int>(dev_size))
       << "Device id " << device_id << " is bigger than number of available devices";
+  // zero initialize cl_program pointers for each device kernel
   if (size == 0) programs_[func_name].resize(dev_size, nullptr);
   return false;
 }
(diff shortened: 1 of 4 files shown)


echuraev:
cc: @elvin-n

src/runtime/opencl/opencl_module.cc
```diff
@@ -353,9 +357,10 @@ PackedFunc OpenCLModuleNode::GetFunction(const String& name,
 }
 
 Module OpenCLModuleCreate(std::string data, std::string fmt,
-                          std::unordered_map<std::string, FunctionInfo> fmap, std::string source) {
+                          std::unordered_map<std::string, FunctionInfo> fmap, std::string source,
+                          bool dry_run) {
   auto n = make_object<OpenCLModuleNode>(data, fmt, fmap, source);
-  n->Init();
+  n->Init(dry_run);
```
elvin-n, line 363: `dry_run` name would assume that we need to have such object and call some API, but results of these calls are not important.

On the other hand, I would assume that no API is called and creation and initialization of OpenCL is just a resuilt of bad design. So, we need either not to create instance of this class or it we have several purposes inside this class and it dedicate for both runtime and some service capabilities, we need to rename this parameter

elvin-n (changes requested):

src/runtime/opencl/opencl_module.cc
echuraev: Thank you for your comment! Previous implementation was refactored and now we don't need this parameter.

echuraev:
@masahi, @elvin-n Could you please review?

masahi (approved):

PR #15748, opened 2023-09-14
[BugFix] Move symbols that are relevant to the runtime from libtvm to…

neucer:
[BugFix] Move symbols that are relevant to the runtime from libtvm to libtvm-runtime


junrushao:
CC: @yelite

yelite (approved):
Thanks for fixing this!

MasterJH5574 (approved):

PR #15750, opened 2023-09-14
[Unity] Fix BYOC codegen for dynamic shapes

vinx13:
The backend function may support dynamic shapes. In this case, the ret sinfo of the function shouldn't be reused.

cc @sunggg @masahi


masahi (approved):
I was just getting an error `InternalError: Check failed: (ptr) is false: The struct_info is not populated, check if you have normalized the expr` while working on a GEMM benchmark script with dynamic `M`. This change fixes it, great!

PR #15755, opened 2023-09-15
[Disco] Use Non-Null Stream for Compute

junrushao:
This PR changes the compute stream to a non-null one to avoid potential implicit synchronization.


junrushao:
Moved to #15766

PR #15756, opened 2023-09-15
[Unittest][Metal] Add minimal metal functionality test to CI

Lunderberg:
Prior to this commit, the CI compiled TVM with `USE_METAL=ON` on OSX, as defined in `conda/recipe/build.sh`, but did not validate the execution of any generated metal kernels.  As a result, breakage could occur without being caught by the CI, such as found following https://github.com/apache/tvm/pull/15103.

This commit adds the execution of a single metal kernel as a minimal functionality test of the metal backend.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -69,7 +69,11 @@ jobs:
         shell: bash -l {0}
         run: >-
           python -m pytest -v tests/python/all-platform-minimal-test
-      - name: Minimal Metal Functionality
+      - name: Minimal Metal Compile-Only
+        shell: bash -l {0}
+        run: >-
+          python -m pytest -v -s 'tests/python/unittest/test_allreduce.py::test_allreduce_sum_compile'
+      - name: Minimal Metal Compile-and-Run
         shell: bash -l {0}
         run: >-
           python -m pytest -v -s 'tests/python/unittest/test_allreduce.py::test_allreduce_sum[dims0-metal]'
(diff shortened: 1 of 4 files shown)


Lunderberg:
The initial version of this PR has a deliberate failure by asserting that `target != "metal"`.  If the CI updates are implemented correctly, the "CI / MacOS" stage should fail due to this assert, and all other CI tests should pass.  This step is to verify that (1) the updated `main.yaml` is enabling additional tests, (2) that the CI build for OSX has the metal backend enabled, and (3) that the CI execution for OSX can find a metal-capable device.

After these three items are verified, the assert statement will be removed, and all CI steps including the "CI / MacOS" stage should pass.

echuraev (commented):
Thank you! Awesome improvement to add minimal Metal testing to CI. Excited about it!

Lunderberg:
Hmm, first time that I've been disappointed to have the CI show green.  The "CI / MacOS" section skipped the Metal test, rather than failing on the assert.

Lunderberg:
And getting closer.  The "compile-only" test has two versions, one of which defines the PackedFunc `"tvm_callback_metal_compile"` and one of which does not.  The version that does not define the PackedFunc finishes without error, as it would provide the generated code to the driver at runtime.  The version that does define the PackedFunc has the compilation function called at compile-time.

Now, the final validation of the CI: Generating code in `CodeGenMetal` that should result in a compile-time error.

Lunderberg:
And the results look good, with the intentionally-incorrect codegen resulting in a failure found at compile-time of the `*.metallib`, during execution of the `xcrun` command.

The intentionally-incorrect codegen is now removed, and the compile-only test should now pass.  Assuming this latest update passes CI, this PR will be marked as ready for review.

Lunderberg:
And ready for review!

echuraev (approved):
LGTM! Thank you! Glad to see such tests in CI.

PR #15757, opened 2023-09-15
[TIR] Output DeclBuffer in FlattenBuffer

Lunderberg:
If a flattened buffer is produced for use in `BufferLoad` and `BufferStore` statements, generate a `DeclBuffer`.  This PR updates the behavior for both TE-derived schedules (`tir.StorageFlatten` transform) and TIR-derived schedules (`tir.FlattenBuffer` transform).

This is a subset of the changes made in https://github.com/apache/tvm/pull/14778, broken out for ease of testing and review.


Lunderberg:
This is blocked on the ethos-u and ethos-m codegen, which does not currently support `DeclBuffer` nodes.

PR #15762, opened 2023-09-15
[TVMScript] Use environment variable TVM_BLACK_FORMAT for .show()

Lunderberg:
Prior to this commit, the default behavior of the `black_format` argument in TVMScript printing has changed back and forth, based on conflicting user preferences.  This commit allows the default to be specified by each using using the `TVM_BLACK_FORMAT` environment variable.  If unspecified in a `obj.show()` method call, this environment variable is used to determine the default.

### DIFF
--- a/python/tvm/runtime/script_printer.py
+++ b/python/tvm/runtime/script_printer.py
@@ -15,6 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 """Configuration of TVMScript printer"""
+import os
 from typing import Dict, List, Optional, Sequence
 
 from tvm._ffi import get_global_func, register_object
@@ -199,7 +200,7 @@ class Scriptable:
     def show(
         self,
         style: Optional[str] = None,
-        black_format: bool = False,
+        black_format: Optional[bool] = None,
         *,
         name: Optional[str] = None,
         show_meta: bool = False,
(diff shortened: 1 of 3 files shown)


junrushao:
Love this PR. Moving my comments from the [previous one](https://github.com/apache/tvm/pull/15706#issuecomment-1721978241).

> Thanks @Lunderberg! 100% agreed with your comments, and particularly, feeling the same as you that the previous generation of TVMScript printer comes with many subtle issues making it not good looking or non-pythonic at all, as you mentioned, extra parentheses, non-fused loops, and ramp/broadcast syntax. The environment variable based approach is even better to me. Happy to get it in!
>
> I have been thinking about this issue as well in the recent days, as it's something we all care about a lot. I think all of us agree with the formatting perspective, and my personal source of pain is mainly about the line width. How about @Hzfengsy? If line-width is the only thing we have different opinions, we can make it configurable using an environment variable.
>
> Let's also move the discussion to @Lunderberg's new PR for more visbility!
>
> Thanks again Eric for being always super thoughtful!

junrushao:
BTW, `line_length` is something we could configure in black: https://github.com/psf/black/blob/19.3b0/black.py#L168

Hzfengsy:
Great PR! It's a great solution to make people with different tastes happy! Thanks @Lunderberg

Hzfengsy (approved):

PR #15763, opened 2023-09-15
[TIR] Do not drop 4th argument to tir.max

Lunderberg:
The `tir.op.comm_reducer` utility provides two distinct APIs, either reducing along a tensor axis or reducing along a list of arguments.

Prior to this commit, when reducing along a list of arguments, the 4th argument was silently dropped.  For example, `tvm.tir.max(1,2,3,4,3,2,1)` would return `3`.


csullivan (approved):

Hzfengsy (approved):

junrushao (approved):
This is great addition! Thanks for the work!

PR #15766, opened 2023-09-17
[Disco] Integrate NCCL with CUDAGraph

junrushao:
This PR integrates NCCL with CUDAGraph by using the default stream that CUDAGraph uses for NCCL communication.


junrushao:
CC: @vinx13 @jinhongyii

PR #15772, opened 2023-09-18
[Unity][QNN] Add QNN quantize/dequantize ops

ibsidorenko:
This PR is a first step to introduce QNN dialect to Unity branch and includes the following changes:
1. Add R.qnn.quantize and R.qnn.dequantize ops.
2. Add Relax QnnCanonicalize pass that is responsible for QNN lowering.


ibsidorenko:
cc @masahi

python/tvm/relax/transform/transform.py
```diff
@@ -1466,3 +1466,16 @@ def create_dataflowblock_pass(pass_arg):
     if pass_func:
         return create_dataflowblock_pass(pass_func)
     return create_dataflowblock_pass
+
+
+def QnnCanonicalize() -> tvm.ir.transform.Pass:
+    """Group one or multiple composite functions created by FuseOpsByPattern into a new function.
+    The new function will be annotated with "Codegen" and "global_symbol" attributes, and it
+    is intented to be offloaded to an external backend.
```
masahi, line 1474: Needs doc update

masahi (commented):
I think it's worth discussing if we need a full-blown QNN "dialect" in Relax, like Relay.

If we want to have other qnn ops like conv2d, dense etc in the future, having a dialect, with a separate "canonicalization" step, is probably necessary. But the industry is moving toward "QDQ representation" for quantization model representation, which only requires quantize and dequantize ops.

If Relax also adopts QDQ for quantized representation, we don't need a large dialect like Relay does. We can define quantize / dequantize as a normal relax ops, and the logic in the canonicalize pass can go directly to the legalize pass.

cc @tqchen @MasterJH5574 @Hzfengsy this PR is motivated by activation quantization (smooth quant, in particular).

python/tvm/relax/transform/transform.py
ibsidorenko: Fixed

quic-sanirudh:
> I think it's worth discussing if we need a full-blown QNN "dialect" in Relax, like Relay.
> 
> If we want to have other qnn ops like conv2d, dense etc in the future, having a dialect, with a separate "canonicalization" step, is probably necessary. But the industry is moving toward "QDQ representation" for quantization model representation, which only requires quantize and dequantize ops.
> 
> If Relax also adopts QDQ for quantized representation, we don't need a large dialect like Relay does. We can define quantize / dequantize as a normal relax ops, and the logic in the canonicalize pass can go directly to the legalize pass.
> 
> cc @tqchen @MasterJH5574 @Hzfengsy this PR is motivated by activation quantization (smooth quant, in particular).

I agree that it might be good to discuss the design of how to support quantization in relax before introducing this.

One approach we've been thinking about is to perhaps introduced a new `QuantTensor` type that might represent quantized tensor and includes the `zero_point` and `scale` params as attributes. This way, we would be able to maintain quantization params as part of the tensor it applies to, and maybe avoid introducing new quantized ops (just reuse the existing ops with `QuantTensor` arguments to represent quantized op). This could eventually be lowered (canonicalized) to a regular op later in the flow if needed.

masahi:
> One approach we've been thinking about is to perhaps introduce a new `QuantTensor` type that might represent quantized tensor and includes the `zero_point` and `scale` as attributes.

To create a variable of such type, don't we end up introducing something like `qnn.quantize` op?

quic-sanirudh:
> To create a variable of such type, don't we end up introducing something like `qnn.quantize` op?

Yes, I think we would need a quantize/dequantize op anyway, but as you suggested above, we could just add it to regular relax ops, and only the type of input/output tensor would be either regular Tensor or `QuantTensor`. This way, the operators are just the same definition, but legalization would change based on type of tensor instead of a canonicalization step, and we get the advantage of a full quantized operator implementation.

I actually did not understand how we might be able to work with just "QDQ representation" with legalization. Would we be doing the work of both `FQ2I` and `Canonicalize` pass together in one shot to get an actual quantized implementation of the operator and schedule that primfunc?

masahi:
> I actually did not understand how we might be able to work with just "QDQ representation" with legalization. Would we be doing the work of both FQ2I and Canonicalize pass together in one shot to get an actual quantized implementation of the operator and schedule that primfunc?

I think it depends on how a backend wants to process a quantized model. For CPU or GPU backends where "an actual quantized implementation" is not necessary (as long as they can see (int8, int8) -> int32 GEMM in the lowered model, for example), indeed the composition of FQ2I and Canonicalize might be sufficient. For more advanced cases where one wants to minimize int8 <-> fp32 conversions, a backend would need to do something like "QDQ propagation" done by TRT https://docs.nvidia.com/deeplearning/tensorrt/developer-guide/index.html#tensorrt-process-qdq. Although TRT says they only support the QDQ representation for quantized models, they clearly have an internal int8 representation to represent the result of such propagation step.

ibsidorenko:
Thank you all, @masahi and @quic-sanirudh  for opinion and discussion!
Since there is still no consensus on the need of QNN dialect, I suggest to add only R.quantize/R.dequantize ops and support them in LegalizeOps pass. This will be enough for my experiments. At the same time overall design will not be changed in this case (like QNN dialect introduction).

yongwww:
cc: @slyubomirsky @psrivas2

ibsidorenko:
I have submitted another [PR#15849](https://github.com/apache/tvm/pull/15849) that adds `R.quantize` and `R.dequantize`. Based on our discussion these ops were supported in `LegalizeOps` pass only (without QNN canonicalization).

masahi:
Closing since https://github.com/apache/tvm/pull/15849 was merged

PR #15773, opened 2023-09-18
[BugFix][CPP] Fix cpp deploy bug

PeterH0323:
## Motivation

Fix bug when using `apps/howto_deploy/run_example.sh`, it will cause `core dumped`

image

After fixed, all test pass.

image

## Modification

Add runtime for `tvm.build`


PeterH0323:
cc @masahi

junrushao (approved):

Issue #15148, opened 2023-06-23
[Tracking Issue] TFLite operator support


leandron:
In https://github.com/apache/tvm/issues/9187 we implemented quantised version of operators in TFLite frontend. Recently, I just noticed a few more operators (with varying priorities) that can be taken as beginner friendly tasks, for someone who's starting in TVM.

I'm creating this as a tracking issue to mark those operators that need to have quantization support implemented and tested:

- [x] `FLOOR_DIV` (#15724)
- [x] `FLOOR_MOD` (#15733)
- [x] `DIV` (#15768)
- [x] `POW` (#15798)
- [ ] `REVERSE_SEQUENCE` (#15915)
- [x] `SQUARE` (#15914)
- [x] `ELU` (#15821)
- [ ] `LOCAL_RESPONSE_NORMALIZATION` (LRN)
- [ ] `L2_NORMALIZATION`
- [x] `MIRROR_PAD` (#16243)
- [ ] `ARG_MIN` (#15922)
- [x] `NOT_EQUAL` (#15769)
- [x] `LESS` (#15746)
- [x] `LESS_EQUAL` (#15790)
- [x] `GREATER_EQUAL` (#15775)

For each operator above, there are actions to be taken:
- Remove the conditional block `if self.is_quantized(op): ...`
- Implement unit tests
- ask any [committer](https://github.com/apache/tvm/blob/main/CONTRIBUTORS.md) in the project to update this issue, reflecting the PR to support the operators on the list above

https://github.com/apache/tvm/pull/9165 can be used as an example.

In case you're interested, a good starting point is the TFLite frontend:
https://github.com/apache/tvm/blob/bee073b0c8e8625216184a2dbb0204c0a376fc26/python/tvm/relay/frontend/tflite.py#L78-L185

p3achyjr:
hi--if no one has taken this yet, can I give this a shot?

leandron:
> hi--if no one has taken this yet, can I give this a shot?

Absolutely @p3achyjr. Each operator can be submitted individually, so we have self contained, easier to review PRs. 

Let us know if you need any help. Thanks!

tlopex:
@leandron  I see that the code has changed a lot from the example#9165. All tests are all combined to one function, so if I just need to remove the conditional block and test it? Besides, because I am totally new to it, I find it a little bit tough to have a test, can you offer me some simple ways?

p3achyjr:
[...]

PR #15775, opened 2023-09-18
[TFLite][Frontend] Support quantized GREATER_EQUAL

tlopex:
Support GREATER_EQUAL quantization operation as part of #15148


lhutton1 (approved):
LGTM!

PR #15778, opened 2023-09-19
[Hotfix] Fix Windows Pipe

junrushao:
Our cross-platform pipe support is reported to be broken on Windows. See also: https://github.com/mlc-ai/mlc-llm/issues/930.


MasterJH5574 (approved):
Loos good, thank you @junrushao!

PR #15781, opened 2023-09-19
[Unity][Dlight] Fix inline consumer in matmul tensorize rule

jinhongyii:
Some blocks have failed to be inlined to the producer cache-write stage. This could be due to another producer block that has not been scheduled. This PR fixes this by using the same function in matmul rule.


jinhongyii:
cc: @Hzfengsy

Hzfengsy:
Can we skip black format for the function in testcase file? it's a bit too long

Hzfengsy (approved):

PR #15783, opened 2023-09-19
[Unity] Use PrimValue as offset in R.tril and R.triu

Lunderberg:
This mirrors the support in `topi`, which supports a `PrimExpr` as the offset of the diagonal.


Lunderberg:
@Archermmt Can you take a look at the MSC-specific unit tests?  I've parametrized the tests and marked the failing cases (`via_relax=False`) with `pytest.mark.xfail`, but it would be good to resolve them in the long-term.

vinx13 (approved):

PR #15791, opened 2023-09-20
[Unity][Pass] Include `FoldDataflowBlockOutput` in `CanonicalizeBindings`

slyubomirsky:
As discussed in certain community meetings, `FoldDataflowBlockOutput` is a bit of a clumsy outlier of a pass, resulting in a cludge like #15474 to try to make it easier to use. In this PR, the functionality of that pass is included in `CanonicalizeBindings` (it can be done just as a further transformation on `DataflowBlock`s) and the separate pass is removed.


slyubomirsky:
Please review @sunggg @kparzysz-quic

kparzysz-quic (commented):
For
```
@tvm.script.ir_module
class Input:
    @R.function
    def main() -> R.Tensor((), "int32"):
        with R.dataflow():
            a = R.const(1)
            b = a
            c = b
            d = c
            n = d
            R.output(n)
        return n
```
the result is
```
@I.ir_module
class Module:
    @R.function
    def main() -> R.Tensor((), dtype="int32"):
        with R.dataflow():
            a: R.Tensor((), dtype="int32") = R.const(1, "int32")
            b: R.Tensor((), dtype="int32") = a
            c: R.Tensor((), dtype="int32") = a
            d: R.Tensor((), dtype="int32") = a
            n: R.Tensor((), dtype="int32") = a
            R.output(n)
        return n
```
The result should be
```
class Module:
    @R.function
    def main() -> R.Tensor((), dtype="int32"):
        with R.dataflow():
            n: R.Tensor((), dtype="int32") = R.const(1, "int32")
            R.output(n)
        return n
```

src/relax/transform/canonicalize_bindings.cc
```diff
@@ -102,6 +103,108 @@ class BindingCanonicalizer : public ExprMutator {
...
+  // Special case: for dataflow blocks, we will check for dataflow vars that solely exist
+  // to be bound to the output. In this case, we will get rid of those bindings and
+  // use the dataflow var's definition directly
+  BindingBlock VisitBindingBlock_(const DataflowBlockNode* block) override {
+    auto new_block = Downcast<DataflowBlock>(ExprMutator::VisitBindingBlock_(block));
+    std::unordered_map<DataflowVar, Expr, ObjectPtrHash, ObjectPtrEqual> binding_map;
+    std::unordered_set<DataflowVar, ObjectPtrHash, ObjectPtrEqual> disqualified_set;
+    std::unordered_set<DataflowVar, ObjectPtrHash, ObjectPtrEqual> output_vars;
+
+    for (auto binding : new_block->bindings) {
```
kparzysz-quic, line 115: Whether a variable is a potential "output" variable, and whether it's disqualified from that depends on how it's used.  It's easier to collect that information going backwards in the dataflow order.

kparzysz-quic (commented):
A more general question is whether we also want to eliminate `a = b`, where both are solely dataflow vars.  All occurrences of `a` could be replaced with `b`.

src/relax/transform/canonicalize_bindings.cc
slyubomirsky: Good point

slyubomirsky:
I hadn't thrown in the functionality from dead code elimination, but it would probably be easy to add it, so I'll try that.

slyubomirsky:
It was indeed easy to throw in the extra functionality.

slyubomirsky:
As it happens, #15840 handles the elision of unnecessary bindings, so I'll remove it from this PR.

slyubomirsky:
@tvm-bot rerun

kparzysz-quic (approved):
LGTM!

PR #15793, opened 2023-09-20
[Unity] Fix dlight bench when func_name is global var

vinx13:
This PR fixed error when func name is global var. `time_evaluator` expects `func_name` to be a string instead of global var.

cc @zxybazh


zxybazh (approved):
Thanks for the fix!

jikechao (approved):
LGTM, thanks for your PR. @vinx13

MasterJH5574 (approved):

PR #15794, opened 2023-09-21
[IR] Use CoW in StmtFunctor to prevent mutating buffer

vinx13:
cc @tqchen @junrushao


junrushao (approved):
Thanks!

tqchen:
@vinx13 we might need to think about possible needs to remap the use site of the buffer, either in the pass, or in followups

Lunderberg:
How large is the performance benefit of `StmtMutator::CopyOnWrite` overall?  Since we only construct new objects if there's a change, and all internal state is held by pointer, my initial guess would be that the performance impact would be minimal.  If so, then we may want to avoid this entire category of bugs by disabling it altogether.

This would also minimize surprises for developers, as currently the `StmtMutator::VisitStmt_` functions may mutate their argument, even though the arguments were provided through `const Node*` and contain no member variables that were defined as `mutable`.

vinx13:
superseded by #15816

PR #15796, opened 2023-09-21
[Disco] Improve RCCL Integration Logic

junrushao:
This PR makes the following changes:
- Move `runtime/disco/ccl` to `runtime/disco/nccl` as RCCL itself uses NCCL as a prefix to its APIs;
- Consolidate `utils.h` to `nccl.cc` to avoid potential misuse of the private header file;
- Reduce the number of macros used in `nccl.cc`.


LeshengJin (approved):
Neat! Learned a lot. Thanks @junrushao

PR #15803, opened 2023-09-21
[Unity] Update test cases of LazyTransformParams as impure

vinx13:
Mark the rest of test cases as impure. This is a follow up of #15697 

cc @Lunderberg


junrushao (approved):
Thanks for taking quick actions!

PR #15806, opened 2023-09-22
[UNITY][Pass] Remove redundant reshape

sjain58:
In Mobilenet float32, we see this sequence of ops:

%179 = squeeze(%178, axis=[2, 3]) /* ty=Tensor[(1, 1001), float16] span=MobilenetEdgeTPU/Logits/Squeeze:0:0 /;
%180 = reshape(%179, newshape=[-1, 1001]) / ty=Tensor[(1, 1001), float16] span=MobilenetEdgeTPU/Predictions/
Reshape:0:0 */;

which later gets transformed to

lv171 = R.call_tir(cls.primfunc_hmx_conv2d26_add20, (lv169, metadata["relax.expr.Constant"][116], metadata["relax.expr.Constant"][117]), out_sinfo=R.Tensor((1, 1001, 1, 1), dtype="float16"))
lv172: R.Tensor((1, 1001), dtype="float16") = R.reshape(lv171, R.shape([1, 1001]))
lv173: R.Tensor((1, 1001), dtype="float16") = R.reshape(lv172, R.shape([1, 1001]))

We can eliminate the redundant reshape.

lv171 = R.call_tir(cls.primfunc_hmx_conv2d26_add20, (lv169, metadata["relax.expr.Constant"][116], metadata["relax.expr.Constant"][117]), out_sinfo=R.Tensor((1, 1001, 1, 1), dtype="float16"))
lv173: R.Tensor((1, 1001), dtype="float16") = R.reshape(lv171, R.shape([1, 1001]))

### DIFF
--- a/python/tvm/relax/transform/remove_redundant_reshape.py
+++ b/python/tvm/relax/transform/remove_redundant_reshape.py
@@ -16,7 +16,6 @@
 # under the License.
 # pylint: disable=invalid-name, unused-argument, missing-function-docstring, abstract-method
 """Relax Remove Redundant Reshape ops"""
-import tvm
 from tvm import IRModule, relax
 from tvm.ir.transform import PassContext
 from tvm.ir import structural_equal
@@ -35,10 +34,10 @@ class RemoveRedundantReshape:
         self.input1 = wildcard()
         shape1 = wildcard()
         pattern_redundant_reshape = is_op("relax.reshape")(self.input1, shape1)
-        self.pattern1 = pattern_redundant_reshape
+        self.no_op_reshape = pattern_redundant_reshape
         shape2 = wildcard()
-        self.pattern2 = is_op("relax.reshape")(pattern_redundant_reshape, shape2)
-        self.pattern = self.pattern2 | self.pattern1
+        self.repeated_reshape = is_op("relax.reshape")(pattern_redundant_reshape, shape2)
+        self.pattern = self.repeated_reshape | self.no_op_reshape
 
     def transform_function(self, func: Expr, mod: IRModule, ctx: PassContext) -> IRModule:
         """
@@ -68,14 +67,12 @@ class RemoveRedundantReshape:
 
             def rewriter(expr, matches):
                 args = matches[self.pattern]
-                if self.pattern2 in matches and args == matches[self.pattern2]:
+                if self.repeated_reshape in matches:
                     return relax.op.reshape(matches[self.input1], args.args[1])
-                elif self.pattern1 in matches and args == matches[self.pattern1]:
+                elif self.no_op_reshape in matches:
                     if args.args[0].struct_info.shape:
                         if structural_equal(args.args[0].struct_info.shape, args.args[1]):
                             return args.args[0]
-                    else:
-                        raise Exception("Tensor of unknown dimension or full shape is not known")
                 return expr
 
             updated_func = rewrite_call(self.pattern, rewriter, funct)
--- a/python/tvm/relax/transform/remove_redundant_reshape.py
+++ b/python/tvm/relax/transform/remove_redundant_reshape.py
@@ -15,9 +15,11 @@
 # specific language governing permissions and limitations
 # under the License.
 # pylint: disable=invalid-name, unused-argument, missing-function-docstring, abstract-method
-
-from tvm.ir.module import IRModule
+"""Relax Remove Redundant Reshape ops"""
+import tvm
+from tvm import IRModule, relax
 from tvm.ir.transform import PassContext
+from tvm.ir import structural_equal
 from tvm.relax import Expr, Function
 from tvm.relax.dpl import is_op, rewrite_call, wildcard
 from . import function_pass
@@ -30,10 +32,13 @@ class RemoveRedundantReshape:
     """
 
     def __init__(self):
-        input = wildcard()
+        self.input1 = wildcard()
         shape1 = wildcard()
-        pattern_redundant_reshape = is_op("relax.reshape")(input, shape1)
-        self.pattern = pattern_redundant_reshape
+        pattern_redundant_reshape = is_op("relax.reshape")(self.input1, shape1)
+        self.pattern1 = pattern_redundant_reshape
+        shape2 = wildcard()
+        self.pattern2 = is_op("relax.reshape")(pattern_redundant_reshape, shape2)
+        self.pattern = self.pattern2 | self.pattern1
 
     def transform_function(self, func: Expr, mod: IRModule, ctx: PassContext) -> IRModule:
         """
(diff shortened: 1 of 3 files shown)


python/tvm/relax/transform/remove_redundant_reshape.py
```diff
@@ -0,0 +1,73 @@
...
+            # Skip non-relax functions
+            if not isinstance(func, Function):
+                continue
+            # Skip primitive functions
+            if "Primitive" in func.attrs.keys() and func.attrs["Primitive"] != 0:
+                continue
+
+            def rewriter(expr, matches):
+                args = matches[self.pattern]
+                if list(args.args[0].struct_info.shape) == list(args.args[1]):
```
Lunderberg, line 66: Instead of using `list(a) == list(b)`, can we use `tvm.ir.structural_equal(a,b)`?  This would ensure that symbolic shapes with `PrimExpr` expressions are correctly handled as well as static shapes.

python/tvm/relax/transform/remove_redundant_reshape.py
```diff
@@ -0,0 +1,73 @@
...
+class RemoveRedundantReshape:
+    """
+    Transformation pass to remove redundant reshape operator
+    """
+
+    def __init__(self):
+        input = wildcard()
+        shape1 = wildcard()
+        pattern_redundant_reshape = is_op("relax.reshape")(input, shape1)
+        self.pattern = pattern_redundant_reshape
```
Lunderberg, line 36: This pattern only checks for reshapes whose argument already matches the new shape.  Should we also check for reshapes whose argument is itself a reshape?  That is, `reshape(reshape(arg, shape1), shape2)` can always be written as `reshape(arg, shape2)`.

tests/python/relax/test_remove_redundant_reshape.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def _run_pass_compare_output(Before, Expected):
+    fused_mod = RemoveRedundantReshape()(Before)
+    if not relax.analysis.well_formed(fused_mod):
+        print("IRModule is not well-formed")
+
+    fused_mod = DeadCodeElimination()(fused_mod)
+    if not relax.analysis.well_formed(fused_mod):
+        print("IRModule is not well-formed")
```
Lunderberg, line 35: Should these cases throw an exception instead of just printing to terminal?

python/tvm/relax/transform/remove_redundant_reshape.py
```diff
@@ -0,0 +1,73 @@
...
+        for _, func in mod.functions.items():
+            # Skip non-relax functions
+            if not isinstance(func, Function):
+                continue
+            # Skip primitive functions
+            if "Primitive" in func.attrs.keys() and func.attrs["Primitive"] != 0:
+                continue
+
+            def rewriter(expr, matches):
+                args = matches[self.pattern]
```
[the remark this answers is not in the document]
sjain58: Thanks @Lunderberg, this makes sense. I have added a condition to check if tensor dimension is not known or full shape of the tensor is unknown as you suggested.

python/tvm/relax/transform/remove_redundant_reshape.py
sjain58: Thanks @Lunderberg, now using `tvm.ir.structural_equal(a,b)` instead of `list(a) == list(b)`.

python/tvm/relax/transform/remove_redundant_reshape.py
sjain58: Thanks @Lunderberg, this is something even we were having in mind as a second step. Now we have support for `reshape(reshape(arg, shape1), shape2)`, and able to remove multiple consecutive redundant reshapes. Thanks a lot!

tests/python/relax/test_remove_redundant_reshape.py
sjain58: Raising exception instead of print statement completely make sense. Thanks for pointing this out!

python/tvm/relax/transform/remove_redundant_reshape.py
```diff
@@ -0,0 +1,83 @@
...
+class RemoveRedundantReshape:
+    """
+    Transformation pass to remove redundant reshape operator
+    """
+
+    def __init__(self):
+        self.input1 = wildcard()
+        shape1 = wildcard()
+        pattern_redundant_reshape = is_op("relax.reshape")(self.input1, shape1)
+        self.pattern1 = pattern_redundant_reshape
```
Lunderberg, line 38: Nitpick: Can `self.pattern1` and `self.pattern2` be renamed to `self.no_op_reshape` and `self.repeated_reshape`?  That would make it easier for future readers to understand why each case is considered redundant.

python/tvm/relax/transform/remove_redundant_reshape.py
```diff
@@ -0,0 +1,83 @@
...
+            # Skip non-relax functions
+            if not isinstance(funct, Function):
+                continue
+            # Skip primitive functions
+            if "Primitive" in funct.attrs.keys() and funct.attrs["Primitive"] != 0:
+                continue
+
+            def rewriter(expr, matches):
+                args = matches[self.pattern]
+                if self.pattern2 in matches and args == matches[self.pattern2]:
```
Lunderberg, line 71: Nit: The `args == matches[self.pattern2]` is unnecessary, as a pattern is only contained in the `matches` lookup if that subpattern was part of the match.

python/tvm/relax/transform/remove_redundant_reshape.py
```diff
@@ -0,0 +1,83 @@
...
+                    return relax.op.reshape(matches[self.input1], args.args[1])
+                elif self.pattern1 in matches and args == matches[self.pattern1]:
+                    if args.args[0].struct_info.shape:
+                        if structural_equal(args.args[0].struct_info.shape, args.args[1]):
+                            return args.args[0]
+                    else:
+                        raise Exception("Tensor of unknown dimension or full shape is not known")
```
Lunderberg, line 78: Should this fall through to the `return expr` instead of raising an exception?  Since tensors are not required to have a fully-defined shape at compile-time, this would raise an exception for some well-defined Relax functions.

tests/python/relax/test_remove_redundant_reshape.py
```diff
@@ -0,0 +1,94 @@
...
+import tvm.testing
+from tvm import relax
+from tvm.relax.transform import DeadCodeElimination
+from tvm.relax.transform import RemoveRedundantReshape
+from tvm.script import ir as I, relax as R
+
+
+def _run_pass_compare_output(Before, Expected):
+    fused_mod = RemoveRedundantReshape()(Before)
+    assert relax.analysis.well_formed(fused_mod), "IRModule is not well-formed"
```
Lunderberg, line 30: Nitpick: I like the checks for well-formedness in general, but they are redundant here.  The test environment for relax unit tests already has a well-formed check before and after each transform.  ([location in `conftest.py`](https://github.com/apache/tvm/blob/unity/tests/python/relax/conftest.py#L23))

Lunderberg (commented):
Thank you for the changes, and I like the implementation!  I just have a couple of nitpicks remaining on it.

(FYI, it looks like the CI is currently stalled on lint checks, with an unused `import tvm`.)

python/tvm/relax/transform/remove_redundant_reshape.py
sjain58: Thanks for the suggestion @Lunderberg! The suggested names are better for future reading. I have made the suggested changes.

python/tvm/relax/transform/remove_redundant_reshape.py
sjain58: Yes, this condition is not needed, and I have removed it. Thanks!

python/tvm/relax/transform/remove_redundant_reshape.py
sjain58: Thank you for pointing this out! I have removed this exception and in that case, it would follow the `return expression` .

tests/python/relax/test_remove_redundant_reshape.py
sjain58: Thanks for confirming that @Lunderberg! In that case we don't need well-formed checks, hence I have removed them.

Lunderberg (approved):
Thank you for making the changes, and looks good to me!

PR #15807, opened 2023-09-22
[Unity] Extend RemoveAllUnused to support relax::Expr

Lunderberg:
Prior to this commit, the `RemoveAllUnused` utility could only apply to `relax::Function` instances.  This commit extends the allowed usage to apply to any `relax::Expr` that contains variable bindings, such as `relax::SeqExpr`.

### DIFF
--- a/include/tvm/relax/analysis.h
+++ b/include/tvm/relax/analysis.h
@@ -399,18 +399,25 @@ TVM_DLL Map<Var, Array<Var>> DataflowBlockUseDef(const DataflowBlock& dfb);
 /*!
  * \brief Get the use-def chain of variables inside a function.
  *
- * \param fn The function to be analyzed.
- * \return A map from variable definitions to a set of uses and variables needed by return value.
+ * \param expr The expression to be analyzed.
+ *
+ * \return A tuple of variable usage and variable outputs.  The first
+ * element is a map from variable definitions to the set of downstream
+ * users of that definition.  The second element is a list of
+ * variables whose usage occurs outside of any variable binding,
+ * typically the output body of a relax::Function or a relax::SeqExpr.
  */
-std::pair<Map<Var, Array<Var>>, Array<Var>> FunctionUseDef(const Function& fn);
+std::pair<Map<Var, Array<Var>>, Array<Var>> FunctionUseDef(const Expr& expr);
 
 /*!
  * \brief Remove unused statements inside DataflowBlocks.
  *
- * \param fn The function to remove unused statements.
- * \return The function that contains no unused statements in DataflowBlock.
+ * \param expr The expression (typically a relax::Function) from which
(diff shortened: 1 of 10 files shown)


csullivan (approved):

jinhongyii:
This PR has a bug when handling a use-def chain which ends at an impure call.
Suppose the code is like 
```
b = 2*a
c = impure_call(b)
return None
```
Then b is an unused var (by your definition), but not an impure call, so it will be removed, which breaks the program.

jinhongyii:
btw, since we have an assumption that binding block is impure, why do we try to remove unused in binding block?

Lunderberg:
> Then b is an unused var (by your definition), but not an impure call, so it will be removed, which breaks the program.

Thank you, and I can reproduce the error with the test case below.

```python
def test_binding_block_keep_pure_func_used_only_for_impure():
    """Remove unused dataflow bindings

    Removal of unused bindings may not remove side effects.  Unused
    bindings whose value is an impure operation (e.g. `R.call_packed`)
    may not be removed, nor may any of their inputs.
    """

    @R.function(private=True)
    def before(x: R.Tensor((32, 32), "int32")):
        y = x * R.const(2)
        z = R.call_packed(
            "function_maybe_with_side_effects", y, sinfo_args=(R.Tensor((32, 32), "int32"))
        )
        return R.tuple()

    expected = before

    after = remove_all_unused(before)
```

For the purpose of removing unused bindings, an argument to an impure function is equivalent to an output, and should be retained.  Thankfully, this should be a pretty straightforward change to make.

> btw, since we have an assumption that binding block is impure, why do we try to remove unused in binding block?

I'd phrase it slightly differently: A `BindingBlock` has no assumptions about the purity of its bound values, where a `DataflowBlock` may assume that all bound values are pure.  That is, we don't need to assume that every step of a `BindingBlock` is impure, but we also cannot assume that any step is pure.  Simplifications that require bound values to be pure are still legal within a `BindingBlock`, so long as the purity is validated before making a change.  The advantage of the `DataflowBlock` is that it lets us skip that validation step, because the contents are required to be pure.

In context of this PR, this check is encoded [here](https://github.com/apache/tvm/pull/15807/files#diff-1f74c84f9a522574f3529c92e5eb683811e6c62c83412cf8575ff8e5b66167ecR249-R250).  A binding is removable if it is unused, and if it's computation is pure, where purity can be established either from being within a `DataflowBlock`, or from checking the purity explicitly.  The bug is that the set of `unused_vars_` is determined by flowing backwards from function outputs, where it should be determined by flowing backwards from function outputs *and* arguments to impure functions.

Lunderberg:
@jinhongyii I've opened https://github.com/apache/tvm/pull/16036, which updates the variable usage tracking to check for arguments to impure functions.

PR #15809, opened 2023-09-22
[Unity] Implement LowerAllocTensor to remove R.builtin.alloc_tensor

Lunderberg:
The `StaticPlanBlockMemory` transform is provided a module that expresses all allocations with `R.builtin.alloc_tensor`, and produces a module that uses `R.memory.alloc_storage` and
`R.memory.alloc_tensor` to express static allocations, while dynamic allocations continue to use `R.builtin.alloc_tensor`.

Prior to this commit, this mixed output was handled as part of `VMBuiltinLower`. This commit extracts the lowering of `R.builtin.alloc_tensor` to a new pass, `LowerAllocTensor`.  This pass runs after `StaticPlanBlockMemory`, and replaces any remaining `R.builtin.alloc_tensor` with calls to `R.memory.alloc_storage` and `R.memory.alloc_tensor`.

### DIFF
--- a/python/tvm/relax/vm_build.py
+++ b/python/tvm/relax/vm_build.py
@@ -314,8 +314,8 @@ def build(
     if tvm.transform.PassContext.current().config.get("relax.backend.use_cuda_graph", False):
         passes.append(relax.transform.RewriteCUDAGraph())
 
-    passes.append(relax.transform.KillAfterLastUse())
     passes.append(relax.transform.LowerAllocTensor())
+    passes.append(relax.transform.KillAfterLastUse())
 
     passes.append(relax.transform.VMBuiltinLower())
     passes.append(relax.transform.VMShapeLower())
--- a/python/tvm/relax/vm_build.py
+++ b/python/tvm/relax/vm_build.py
@@ -310,10 +310,12 @@ def build(
     passes.append(relax.transform.RemovePurityChecking())
     passes.append(relax.transform.CallTIRRewrite())
     passes.append(relax.transform.StaticPlanBlockMemory())
-    passes.append(relax.transform.KillAfterLastUse())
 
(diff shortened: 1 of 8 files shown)


python/tvm/relax/vm_build.py
```diff
@@ -310,6 +310,7 @@ def foo(x: Tensor((3, 4), "float32"), y: Tensor((3, 4), "float32")):
     passes.append(relax.transform.RemovePurityChecking())
     passes.append(relax.transform.CallTIRRewrite())
     passes.append(relax.transform.StaticPlanBlockMemory())
+    passes.append(relax.transform.LowerAllocTensor())
 
     if tvm.transform.PassContext.current().config.get("relax.backend.use_cuda_graph", False):
         passes.append(relax.transform.RewriteCUDAGraph())
```
csullivan, line 316: Did you look in to what assumptions `relax.transform.RewriteCUDAGraph` makes on static vs dynamic allocations to ensure the expectations match before and after this change, given that there is some dependence on alloc_tensor in RewriteCudaGraph?

Lunderberg: It appears to be a bug in `RewriteCUDAGraph`, which occurs when there is a `R.memory.alloc_storage` that is then used in a trivial var-to-var rebinding.

Lunderberg: This bug ended up being trickier to track down than I had expected.  It will be much simpler to solve after https://github.com/apache/tvm/pull/15810 lands, since the `.kill_*` methods won't be inserted yet.  For now, I've re-ordered the passes so the `LowerAllocTensor` occurs after `RewriteCUDAGraph`.

csullivan: Awesome, many thanks for looking into this @Lunderberg, changing the ordering until the fix is in makes sense. Let's land this an #15810.

csullivan (approved):

PR #15815, opened 2023-09-25
[Unity] Extend EliminateCommonSubexpr to operate on relax::Expr

Lunderberg:
Prior to this commit, the `EliminateCommonSubexpr` utility could only apply to `relax::Function` instances.  This commit extends the allowed usage to apply to any `relax::Expr` that contains variable bindings. This is only included as an internal utility within the C++ implementation, and is not currently exposed for external use.


slyubomirsky (approved):
Great change, thank you very much.

PR #15153, opened 2023-06-24
[TensorIR][Visitor] Visit buffer members in `match_buffer`'s in block visitor functions

yzh119:
# Motivation
Currently, the `elem_offset` member in `match_buffer`'s will not be visited in default block visitor functions, resulting in some passes cannot correctly rewrite the program because some fields have not been entered.

This PR fixes the issue by visiting this field in default StmtVisitor and StmtExpr.

PR #15816, opened 2023-09-26
Revert "[TensorIR][Visitor] Visit buffer members in `match_buffer`'s in block visitor functions (#15153)

vinx13:
This reverts commit 34637d7ee38f2636b1948548a39c15838d7a8db6.

It is better not visiting fields of buffer when visiting `MatchBufferRegion` since it's more common that the buffer is not mutated in `StmtMutator` (it can be left to the derived passes to handle case by case). It is also consistent with visitor/mutator for `BufferLoad` and `BufferStore` that do not visit buffer fields.
This fixes some broken tests in software pipelining where strides of some buffers in `MatchBufferRegion` are unintentionally mutated in-place.

cc @yzh119 @tqchen @junrushao

### DIFF
--- a/src/tir/ir/stmt_functor.cc
+++ b/src/tir/ir/stmt_functor.cc
@@ -135,15 +135,8 @@ void StmtVisitor::VisitStmt_(const BlockNode* op) {
   VisitArray(op->reads, fvisit_buffer_region);
   VisitArray(op->writes, fvisit_buffer_region);
   VisitArray(op->match_buffers,
-             [this, fvisit_buffer_region](const MatchBufferRegion& match_buffer_region) {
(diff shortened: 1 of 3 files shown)


yzh119 (approved):
I discussed with @vinx13 and I think it's reasonable not to visit buffers in the base mutator but explicitly do that in derived classes if necessary.

Lunderberg:
As a potential third option, I think that we could have the buffers be visited at their point of definition.  That is, any `DeclBuffer` statement would have the buffer internals visited once, and the mutated buffer would then be substituted in at all points of use.  This would require the `DeclBuffer` to be present for all buffer objects, but that's part of the long-term plan anyways.

PR #15817, opened 2023-09-26
[TE] Support using tir::Var as CreatePrimFunc args

LeshengJin:
- Allow `CreatePrimFunc` args to be a mixture of `te::Tensor` and `tir::Var`.
- Integrate `CreatePrimFunc` and `CreateRelaxPrimFunc` into one function.

```python
idx = te.var("idx", dtype="int64")
m = te.var("m", dtype="int64")
n = te.var("n", dtype="int64")
tensor = te.placeholder((m, n), name="tensor")
slice0 = te.compute((idx, n), lambda i, j: tensor[i, j], name="slice")
# use idx as an arg
te.create_prim_func([tensor, idx, slice])
```

cc: @junrushao @Hzfengsy @yongwww

### DIFF
--- a/python/tvm/te/operation.py
+++ b/python/tvm/te/operation.py
@@ -19,7 +19,7 @@ import inspect
 
 # pylint: disable=invalid-name
 from numbers import Integral as _Integral
-from typing import List, Optional
+from typing import List, Optional, Union
 
 import tvm._ffi
 import tvm.arith._ffi_api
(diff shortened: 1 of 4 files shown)


junrushao:
I'm a big fan of this change! CC: @yongwww @masahi @Hzfengsy

Hzfengsy (approved):

MasterJH5574:
Looks like the PR is mainly about TE and TIR, so I’m wondering if there is any blocker for this PR to be sent to main branch?

junrushao:
The motivation behind is to unify [`CreatePrimFunc`](https://github.com/apache/tvm/blob/b7471f8315ac50ed42da33db733dfbe3a94eff06/src/te/operation/create_primfunc.cc#L646) and [`CreateRelaxPrimFunc`](https://github.com/apache/tvm/blob/b7471f8315ac50ed42da33db733dfbe3a94eff06/src/te/operation/create_primfunc.cc#L652) that have been used semi-interchangeably but neither are consistent with the existing `tvm.build` API that accepts both TIR Var and TE Tensor in `args`. Let's get this in first, and then isolate the non-Relax part and send to mainline as a follow-up step

yongwww:
thanks for the change, looks good to me!

PR #15818, opened 2023-09-26
[CLI TOOLS][RTVM] Improve rtvm tool with new options to measure native performance

srkreddy1238:
Few fixes and enhancements that affects model loading times New options to measure performance.

### DIFF
--- a/apps/cpp_rtvm/main.cc
+++ b/apps/cpp_rtvm/main.cc
@@ -29,6 +29,7 @@
 #endif
 #include <dmlc/logging.h>
 
+#include <chrono>
 #include <cstring>
 #include <iostream>
 #include <sstream>
@@ -54,7 +55,11 @@ static const string kUsage =
     "--input        - Numpy file for the model input (optional and we use random of not given)\n"
     "--output       - Numpy file name to dump the model output as numpy\n"
     "--dump-meta    - Dump model meta information\n"
-    "--pre-compiled - The file name of a file where pre-compiled programs should be stored"
+    "--pre-compiled - The file name of a file where pre-compiled programs should be stored\n"
+    "--profile      - Profile over all execution\n"
+    "--dry-run      - Profile after given dry runs, default 10\n"
+    "--run-count    - Profile for given runs, default 50\n"
+    "--zero-copy    - Profile with zero copy api\n"
     "\n"
     "  Example\n"
     "  ./rtvm --model=keras-resnet50 --device=\"opencl\" --dump-meta\n"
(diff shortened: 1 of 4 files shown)


apps/cpp_rtvm/main.cc
```diff
@@ -192,59 +226,171 @@ int ExecuteModel(ToolArgs& args) {
 #endif
 
   // Initialize TVM Runner
-  TVMRunner runner = TVMRunner(args.model, args.device);
+  auto runner = new TVMRunner(args.model, args.device);
```
echuraev, line 229: Why do you need this change? Probably I missed something, but I didn't see any place where pointer to `runner` is necessary. In case if you need the pointer to `runner`, I would recommend to use smart pointers instead of raw pointer.

apps/cpp_rtvm/tvm_runner.cc
```diff
@@ -156,8 +193,7 @@ inline size_t GetMemSize(NDArray& narr) {
  * \return The memory size.
  */
 size_t TVMRunner::GetInputMemSize(std::string input_id) {
-  LOG(INFO) << "TVMRunner::GetInputMemSize:" << input_id;
-
+  // LOG(INFO) << "TVMRunner::GetInputMemSize:" << input_id;
```
echuraev, line 196: nit: Such comments can be removed.

echuraev (commented):
Thank you for your PR! Just a couple of small comments

apps/cpp_rtvm/main.cc
srkreddy1238: Handled. Don't need a pointer for now.

apps/cpp_rtvm/main.cc
```diff
@@ -350,43 +350,46 @@ int ExecuteModel(ToolArgs& args) {
...
     }
   }
+
+  if (args.profile) {
+    // Print Stats
+    runner.PrintStats();
+  }
   auto tstart = std::chrono::high_resolution_clock::now();
-  delete runner;
+  runner.~TVMRunner();
```
echuraev, line 389: Why do you need to call dtor here? Probably, I missed something in the logic.
I suppose that the object will be automatically destroyed in the end of the program.

srkreddy1238: We are timing it to measure the unload time.

srkreddy1238:
@echuraev There are some requirements and concerns on high load and unload timings with TVM in general. With PreCompiled bin we could reduce the majority and there exist some more scope with params loading. Unload time is also a bit significant. I am working on few improvements around these areas ...

echuraev (approved):
LGTM! @srkreddy1238 just up to you if you want to return dynamic allocation or not. Feel free to merge this PR when the CI is green.

Issue #15754, opened 2023-09-15
[CI Problem] ci_gpu fails to build due to missing oneflow version


ashutosh-arm:
ERROR: No matching distribution found for oneflow==0.7.0+cpu

The error might be new as the [last gpu docker build](https://hub.docker.com/r/tlcpack/ci-gpu/tags) is 4 months old. When I checked the [command line](https://github.com/apache/tvm/blob/67df20faeeffaf3c145698783f6cd157b8cdbd9c/docker/install/ubuntu_install_oneflow.sh#L25) that is used to install oneflow dependency, the source https://release.oneflow.info/ seems to be unreachable. What should be the way out?

### Branch/PR Failing

https://github.com/apache/tvm/pull/15747

### Jenkins Link

https://ci.tlcpack.ai/blue/rest/organizations/jenkins/pipelines/tvm-docker/branches/PR-15747/runs/2/nodes/74/steps/240/log/?start=0

* needs-triage

leandron:
cc @hhhfccz @BBuf for visibility.

lhutton1:
fixed by https://github.com/apache/tvm/pull/15836

PR #15819, opened 2023-09-26
[Docker] Install oneflow from PyPi

lhutton1:
Installing oneflow from the current link (https://release.oneflow.info) seems to be broken as reported in #15754, which is impacting other unrelated changes in CI. This commit attempts to fix the install by using a version from PyPi.


lhutton1:
cc @ashutosh-arm @leandron @neildhickey

PR #15820, opened 2023-09-26
[Unity][Analysis] Change warning to info for non-affine transform

Lunderberg:
A `PrimFunc` may contain transformations that are not affine, or that are not recognizable as being affine.  Prior to this commit, this produced a `LOG(WARNING)` stating that this was an error in the index analysis.  Because this is the expected behavior for these functions, this commit updates the message to instead write to `DLOG(INFO)`.

The other uses of `LOG(WARNING)` inside `layout_transformation.cc` are kept as-is, as they relate to limitations of the analysis, rather than being a statement about the analyzed `PrimFunc`.


quic-sanirudh (approved):
LGTM, thanks.

PR #15821, opened 2023-09-26
[TFLite][Frontend] Support quantized ELU

tlopex:
Support ELU quantization operation as part of #15148


tlopex:
cc @leandron @lhutton1

leandron (approved):
LGTM, thanks @tlopex! This is merged.

PR #15823, opened 2023-09-26
[Unity] Add Relax multi-device e2e cases 

yongwww:
This Pull Request enables Relax multi-device end-to-end test cases, including multi-CPU, multi-GPU, and heterogeneous CPU-GPU cases.
- Add the `kTarget` attribute to `prim_func` during `LegalizeOps` if the related vdevice is annotated.
- Change CallTIRRewrite pass to `ModulePass`,  find and fill the device_index in the emitted `alloc_tensor` by looking up the global `vdevice` list of context mod.
- Update DefaultGPUSchedule to only works for prim_funcs that are supposed to be on GPU, allowing prim_funcs to work for multiple targets.
- Make the `target` argument of `relax.build` optional, and use the annotated `target` in the `vdevice` list for compilation if it is not specified.

cc: @tqchen @psrivas2 @junrushao @csullivan @Lunderberg @zxybazh @qelk123


yongwww:
@tvm-bot  rerun

src/tir/transforms/default_gpu_schedule.cc
```diff
@@ -98,24 +98,48 @@ IRModule MarkScheduled(const IRModule& mod) {
...
                   mod->import_set_,       // import_set
                   mod->source_map,        // map
-                  mod->attrs);            // attrs);
+                  mod->attrs,             // attrs
+                  mod->global_infos);     // global_infos
+}
+
+bool IsScheduledOnGPU(const BaseFunc& func) {
+  Optional<tvm::Target> target = func->attrs.GetAttr<tvm::Target>(tvm::attr::kTarget);
+  if (target.defined()) {
```
zxybazh, line 107: Do we need to also check target from context here, i.e., GPU target defined in context but not in the function attribute?

tqchen:
cc @zxybazh @Hzfengsy

zxybazh (approved):
LGTM, thanks @yongwww!

PR #15824, opened 2023-09-26
[TVMScript] Preserve traceback across TVMScript parsing

Lunderberg:
Prior to this commit, exceptions raised during the parsing of TVMScript would be caught and replaced with a new exception.  While this does allow the TVMScript location of the error to be included in the exception, it also removes the stack trace of the original error. This commit updates the `Parser.report_error` function to provide the original stack trace alongside the updated exception object.


junrushao (approved):
Awesome update!

PR #15825, opened 2023-09-26
[Unity] Dtype check in legalization of R.matmul

Lunderberg:
Prior to this commit, if `R.matmul` has arguments with an unknown datatype, it would produce an error from within the `R.call_te` implementation of `BlockBuilder.call_te`, stating `TVMError: cannot make const for type handle`.  While true, this is not the most easily interpreted error message.

This error message occurs due to the use of an unknown data type in Relax as a concrete data type in TE/TIR.  While Relax supports unknown datatypes, TIR does not.  When the Relax representation of an unknown datatype, `DataType::Void()`, is passed into TIR, this is interpreted as an opaque handle, which cannot have a scalar constant produced for it.

This commit validates the types being used in Relax, prior to generating the legalized implementation of `R.matmul`.  By catching the error earlier, the error message can be in terms of the Relax objects being used.

### DIFF
--- a/python/tvm/relax/transform/legalize_ops/linear_algebra.py
+++ b/python/tvm/relax/transform/legalize_ops/linear_algebra.py
@@ -89,6 +89,14 @@ def _matmul(bb: BlockBuilder, call: Call) -> Expr:
             name="matmul",
         )
 
+    lhs, rhs = call.args
+    lhs_sinfo = call.args[0].struct_info
+    rhs_sinfo = call.args[1].struct_info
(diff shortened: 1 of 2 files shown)


slyubomirsky:
Are you certain that `Void` is treated as false in that check? There's no test case included.

Lunderberg:
I had tested it locally, but initially didn't add a test case as it only changes where an error is thrown from, and no additional functionality.  That said, your point on validating the assertion is a good one, and I've added a test to ensure that the unknown dtype triggers the correct error.

slyubomirsky (approved):
Thank you for adding the test case!

PR #15826, opened 2023-09-26
[Disco] Loading-time sharding support

LeshengJin:
In our previous implementation, parameter sharding relies on pre-quantization weight processing, meaning each set of quantized weights corresponds strictly to a hardcoded constant `num_shards`, and re-quantization is strictly required upon each change of #GPUs, e.g. from 4-GPU to 8-GPU setting. This PR makes it possible to move parameter sharding to post-quantization loading-time. During loading, we iterate over all parameters and apply the sharding operation based on the provided sharding information.

To make this happen, this PR makes an enhancement to the existing `shard_info.json` to include the sharding function being used at loading time. Each parameter is attached to a list of loading-time preprocessing methods that are serially applied to it to transform this parameter to the desired shape, as shown in the example below:

```python
shard_info = {
  "x_0": [ # name of the parameter
    [ # a list of preprocessing functions to be applied
      "tests.disco.shard_dim_1",  # name of the sharding function
      [(num_shards, 64, 64), "float16"],  # output shape/dtype of `tests.disco.shard_dim_1`
      num_shards,  # extra inputs to `tests.disco.shard_dim_1`
    ],
  ],
  "x_1": [...],
}
```

To parameter `x_0`, it means we will call method `tests.disco.shard_dim_1` which has the signature:

```python
def shard_dim_1(
  input: NDArray,
  num_shards, # extra inputs
  output: NDArray, # and its shape is (num_shards, 64, 64), and dtype is "float16"
) -> None: ...
```

This approach simplifies parameter sharding for users and ensures correctness.

### DIFF
--- a/src/runtime/disco/loader.cc
+++ b/src/runtime/disco/loader.cc
@@ -23,6 +23,7 @@
 #include <functional>
 #include <numeric>
 #include <string>
+#include <unordered_map>
 #include <vector>
 
 #include "../file_utils.h"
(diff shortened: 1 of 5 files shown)


junrushao:
CC: @jinhongyii

src/runtime/disco/loader.cc
```diff
@@ -60,6 +61,10 @@ class ShardLoaderObj : public Object {
     const FileRecord* file;
     const ParamRecord* param;
     int shard_dim;
+    PackedFunc f_reorder_fp16_;
```
jinhongyii, line 64: Why do we need 3 different functions for different dtypes? Why can't we pass different packed func in shard info when dtypes are differnt?

jinhongyii: Can we rename to f_preprocess, because it's more general?

jinhongyii: I'm thinking of future cases where we might need more kinds of preprocessing other than reorder. If we let users pass a preprocess function's base name in shard info, like r`eorder_qkv`, and look for corresponding dtype variant `reorder_qkv_fp16` in IRModule automatically, it might be more flexible. If we want to do that, we can look up preprocess func when we call `Shard` instead of `Create`

junrushao (changes requested):
Let's unify sharding and reordering into a single function - no need for shard3d any more :)

jinhongyii:
Can we separate reorganizing function and sharding function again? I'm thinking of some situation where there can be 2 reorganizing function ( doing nothing and combine qkv), and n different sharding functions( different tensor shape and different sharding dimension). In this case, we will need at most 2n functions in total in IRModule.

junrushao (approved):

jinhongyii (approved):

PR #15827, opened 2023-09-26
[Unity][Disco] Use default stream in disco

jinhongyii:
running on default stream and synchronizing with it is significantly faster on amd gpu than running on other stream and synchronizing that stream, so we switch to using default stream in init_ccl.

I also tested on nvidia gpu and confirmed that there's no perf regression.

### DIFF
--- a/src/runtime/disco/nccl/nccl.cc
+++ b/src/runtime/disco/nccl/nccl.cc
@@ -137,20 +137,12 @@ inline ncclRedOp_t AsNCCLRedOp(ReduceKind kind) {
 struct CCLThreadLocalContext {
   DiscoWorker* worker;
   int device_id;
-  deviceStream_t default_stream;
+  deviceStream_t default_stream = nullptr;
   ncclComm_t comm;
 
-  void Clear() {
-    NCCL_CALL(ncclCommDestroy(comm));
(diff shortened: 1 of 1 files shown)


jinhongyii:
cc: @junrushao @tqchen

junrushao (approved):

src/runtime/disco/nccl/nccl.cc
```diff
@@ -137,20 +137,12 @@ inline ncclRedOp_t AsNCCLRedOp(ReduceKind kind) {
...
-    StreamDestroy(default_stream);
-  }
+  void Clear() { NCCL_CALL(ncclCommDestroy(comm)); }
 
-  deviceStream_t GetDefaultStream() {
-    const auto* func = tvm::runtime::Registry::Get("runtime.get_" TVM_DISCO_DEVICE_NAME "_stream");
-    ICHECK(func != nullptr);
-    deviceStream_t stream = static_cast<deviceStream_t>((*func)().operator void*());
-    return stream == nullptr ? default_stream : stream;
-  }
```
junrushao, line 153: it seems to break cudagraph integration because cudagraph provides a default stream

PR #15829, opened 2023-09-27
[Relay][Keras][Bugfix] fix the converters of GRU and SimpleRNN about the go_backwards attribute

jikechao:
Two similar bugs with the issue: https://github.com/apache/tvm/issues/15176

For the convert function of GRU and SimpleRNN in the Keras frontend, they overlook the parsing for `go_backwards` attribute.
Thus, they can lead to wrong inference results. 

This PR fixed them and added two corresponding regression tests.

cc @echuraev @Hzfengsy @merrymercy


vvchernov (approved):
LGTM

echuraev (approved):
LGTM. Thanks!

PR #15830, opened 2023-09-27
[ADRENO] Minor changes for Adreno docs and help scripts

srkreddy1238:
NCHW is mandatory layout for CLML offload. Updated the docs. CI scripts will keep OpenCL enbaled as fallback always. Enable configurable device bind ports. Helps in multi user environments.


echuraev (approved):
LGTM. Thanks!

PR #15058, opened 2023-06-07
[OpenCL][Texture] Improved texture memory planning

srkreddy1238:
Motivated form the fact that textures can be allocated over a clBuffer object and the size of backing clBuffer can be computed based on hardware image pitch alignment.

This optimizes the overall memory allocation on device and helps greately the models with large memory requirements.

Improvised the graph memory planner to not differentiate buffer and texture storage tokens and reuse them across. The texture pool in OpenCL runtime is rebranded as memory pool that handles allocation for both buffer and image objects.

NDArray to DeviceAPI interface is extended with AllocDataSpaceView and FreeDataSpaceView. These new API's acommodates accessing same physical memory as clBuffer / clImage objects.

PR #15833, opened 2023-09-27
[VM] Memory Manager moved up to runtime

srkreddy1238:
Now graph runtime also uses the same memory manager This acommodates a common memory manager with pooled and naive support.

As a follow up we can move the WorkspacePool to use this common memory manager.

This is a prerequisite to accommodate a common two stage memory allocation as described in #15058

### DIFF
--- a/include/tvm/runtime/memory/memory_manager.h
+++ b/include/tvm/runtime/memory/memory_manager.h
@@ -66,8 +66,7 @@ class Allocator {
    *  \param mem_scope The device memory scope hint.
    *  \return The empty NDArray.
    */
-  NDArray Empty(std::vector<int64_t> shape, DLDataType dtype, Device dev,
-                Optional<String> mem_scope);
+  NDArray Empty(ShapeTuple shape, DLDataType dtype, Device dev, Optional<String> mem_scope);
   /*! \brief Return the allocator type. */
   inline AllocatorType type() const { return type_; }
   /*! \brief Allocate a buffer given a size, alignment and type.
@@ -135,7 +134,7 @@ class StorageObj : public Object {
   Buffer buffer;
 
   /*! \brief Allocate an NDArray from a given piece of storage. */
-  NDArray AllocNDArray(size_t offset, std::vector<int64_t> shape, DLDataType dtype);
+  NDArray AllocNDArray(size_t offset, ShapeTuple shape, DLDataType dtype);
 
   /*! \brief The deleter for an NDArray when allocated from underlying storage. */
   static void Deleter(Object* ptr);
--- a/include/tvm/runtime/memory/memory_manager.h
+++ b/include/tvm/runtime/memory/memory_manager.h
@@ -39,6 +39,11 @@ namespace tvm {
 namespace runtime {
 namespace memory {
 
+enum AllocatorType {
+  kNaive = 1,
+  kPooled,
+};
+
 struct Buffer {
   /*! \brief The pointer to the allocated block of memory. */
   void* data{nullptr};
@@ -46,11 +51,8 @@ struct Buffer {
   size_t size{0};
   /*! \brief The context of the allocated buffers. */
   Device device;
-};
-
-enum AllocatorType {
-  kNaive = 1,
-  kPooled,
+  /*! \brief The allocator that created this buffer. */
+  AllocatorType alloc_type;
 };
 
 class Allocator {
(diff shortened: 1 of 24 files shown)


include/tvm/runtime/memory_manager.h
```diff
@@ -37,9 +37,8 @@
 
 namespace tvm {
 namespace runtime {
-namespace vm {
 
-struct Buffer {
+struct MBuffer {
```
tqchen, line 41: It would be useful if we can move to a new namespace runtime::memory and keep the same buffer name.

tqchen (changes requested):

include/tvm/runtime/memory_manager.h
```diff
@@ -130,7 +131,7 @@ class MemoryManager {
 class StorageObj : public Object {
  public:
   /*! \brief The index into the VM function table. */
-  Buffer buffer;
+  MBuffer buffer;
 
   /*! \brief Allocate an NDArray from a given piece of storage. */
   NDArray AllocNDArray(size_t offset, std::vector<int64_t> shape, DLDataType dtype);
```
yongwww, line 137: It would be good to update vector<int_64_t> shape to `ShapeTuple`. I can work on this once this pr lands

srkreddy1238: Handled ShapeTuple across memory_manager and Allocators

include/tvm/runtime/memory/memory_manager.h
```diff
@@ -37,15 +37,15 @@
...
+namespace memory {
 
 struct Buffer {
   /*! \brief The pointer to the allocated block of memory. */
   void* data{nullptr};
   /*! \brief The size of the block. */
   size_t size{0};
   /*! \brief The shape of the tensor. */
-  std::vector<int64_t> shape;
+  ShapeTuple shape;
```
yongwww, line 48: we can remove this member `ShapeTuple shape;`, and related `buf.shape.push_back(shape[i]);` in naive_allocator.h. Looks it is not used.

src/runtime/memory/memory_manager.cc
```diff
@@ -154,29 +154,30 @@ Allocator* MemoryManager::GetAllocator(Device dev) {
...
-  *buffer = this->Alloc(size, alignment, dtype);
+  if (!mem_scope.defined() || mem_scope == "global") {
+    *buffer = this->Alloc(size, alignment, dtype);
+  } else {
+    *buffer = this->Alloc(shape, dtype, mem_scope.value());
+  }
   container->manager_ctx = reinterpret_cast<void*>(buffer);
   container->dl_tensor.data = buffer->data;
   return NDArray(GetObjectPtr<Object>(container));
 }
```
yongwww, line 174: it would be helpful to add the definition of Storage as below in in memory_manager.cc, it was not defined yet.

```
Storage::Storage(Buffer buffer) {
  auto n = make_object<StorageObj>();
  n->buffer = std::move(buffer);
  data_ = std::move(n);
}
```

yongwww (commented):
Thanks for the work, overall it looks good to me, have left some comments.

include/tvm/runtime/memory/memory_manager.h
srkreddy1238: Agreed. Even the two stage memory allocation for OpenCL texture doesn't need it.

yongwww (approved):
LGTM

tqchen (approved):

src/runtime/memory/memory_manager.cc
```diff
@@ -154,29 +160,30 @@ Allocator* MemoryManager::GetAllocator(Device dev) {
...
+                         Optional<String> mem_scope) {
   VerifyDataType(dtype);
   NDArray::Container* container = new NDArray::Container(nullptr, shape, dtype, dev);
   container->SetDeleter(BufferDeleter);
-  size_t size = GetDataSize(container->dl_tensor);
+  size_t size = DeviceAPI::Get(dev)->GetDataSize(container->dl_tensor);
   size_t alignment = GetDataAlignment(container->dl_tensor);
   Buffer* buffer = new Buffer;
-  *buffer = this->Alloc(size, alignment, dtype);
+  if (!mem_scope.defined() || mem_scope == "global") {
```
echuraev, line 171: Also `mem_scope` can be empty:
```suggestion
  if (!mem_scope.defined() || mem_scope.empty() || mem_scope == "global") {
```

echuraev (changes requested):
In general LGTM. Thank you for your PR! One comment which should be applied.

echuraev (approved):

yongwww:
This was merged, thanks for the effort!

PR #15835, opened 2023-09-27
[CodeGenC][Redo] Handle GlobalVar callee as internal function call

Lunderberg:
This reverts commit [`e88d0d`](https://github.com/apache/tvm/pull/15725), which itself reverted [`9ff71f`](https://github.com/apache/tvm/pull/15103) for breakages on the metal backend.  Now that the CI contains compile-time testing of the metal codegen, the original breakage should be identifiable.

### DIFF
--- a/python/tvm/contrib/xcode.py
+++ b/python/tvm/contrib/xcode.py
@@ -131,6 +131,11 @@ def compile_metal(code, path_target=None, sdk="macosx", min_os_version=None):
     temp_ir = temp.relpath("my_lib.air")
     temp_target = temp.relpath("my_lib.metallib")
 
+    # Debug printout for https://github.com/apache/tvm/pull/15835
+    print('/////////////// BEGIN METAL CODEGEN ///////////////')
+    print(code)
+    print('///////////////   END METAL CODEGEN ///////////////', flush=True)
+
     with open(temp_code, "w") as out_file:
         out_file.write(code)
     file_target = path_target if path_target else temp_target
--- a/python/tvm/contrib/xcode.py
+++ b/python/tvm/contrib/xcode.py
@@ -131,11 +131,6 @@ def compile_metal(code, path_target=None, sdk="macosx", min_os_version=None):
     temp_ir = temp.relpath("my_lib.air")
     temp_target = temp.relpath("my_lib.metallib")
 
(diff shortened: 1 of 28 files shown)


Lunderberg:
This PR is currently listed as a draft, to ensure that the CI can catch the failure mode that was reported [here](https://github.com/apache/tvm/pull/15103#issuecomment-1714790711).

Lunderberg:
And the fix is now in.  It ended up being a <10 line fix once the CI was able to catch the problem, but getting the CI to that point was the harder part.

@MasterJH5574  Can you verify that PR doesn't re-introduce the issue that you reported [here](https://github.com/apache/tvm/pull/15103#issuecomment-1714790711).  The CI tests should be sufficient, but it would be good to confirm it as well.

Lunderberg:
@MasterJH5574 Have you had a chance to verify that this PR does not cause a regression on OSX?  I'd like to avoid leaving it idle for too long, as that allows conflicts with unrelated changes to creep in.

csullivan (approved):
With #15756 the original cause for reversion (#15725) has been resolved.

tqchen:
I think one issue revealed by the latest set of regression is that we are trying to build a generic set of functionalities that are not necessarily used in some of the subclass settings.

For example, in the case of metal codegen the easiest is to actually decouple from the normal host c codegen, reduce possible re-use while ensure local code changes are correct. In the GPU shader generations, we do not need to handle GlobalVar callee anyway, so we don't have to bear the extra level of indirections

tqchen:
My read is that for kernel generators(where we know there won't be cross function call, or we know we do not yet support kernel function call), we can safely remove 

https://github.com/Lunderberg/tvm/blob/e8ef2f25f45f44aabfa1652027a05b3aae4c67f8/src/target/source/codegen_metal.cc#L352C2-L354C6 and 

add a comment that we skip declaration as the cross-kernel calls are not supported

tqchen:
actually it turns out to be more complicated, although the above analysis indeed is right. I think given shader generator won't have a need for inter function call, we can keep the original logic which will ensure printing the signature and auxiliary data structure once

MasterJH5574:
Filed a PR https://github.com/apache/tvm/pull/16033 that restores the Metal codegen logic. I verified that it has no problem with both iPhone and Metal codegen.

PR #15837, opened 2023-09-28
[Unity][BYOC] Support attention over batched sequences of variable sequence lengths 

masahi:
CUTLASS fMHA and Flash Attention can compute attention over batched sequences of variable sequence lengths in one kernel call. This is very useful for batched "prefill" attention in LLM workloads. vllm and TGI rely on such kernels to do prefill attention efficiently. 

To support this, I introduced a new variant of `nn.attention` op that takes concatenated queries, values, and keys as input. In addition, sequence length information is encoded in a somewhat odd argument (`seqstart_q`). This arg is required by both CUTLASS fMHA and Flash Attention for the variable length mode, so it is a standard one.  

The test case demonstrates how to use and offload this op to CUTLASS for a typical batched prefill workload in LLM inference.

@cyx-6 @vinx13 @MasterJH5574 @yelite @sunggg

### DIFF
--- a/python/tvm/contrib/cutlass/build.py
+++ b/python/tvm/contrib/cutlass/build.py
@@ -862,6 +862,8 @@ class CutlassRelaxFunctionAnnotator(relax.PyExprMutator):
     def handle_attention(self, f, op_type):
         """Annotate an attention op."""
         signature = _extract_relax_function_signature(f)
+        arg_idx = _extract_arg_idx(op_type, f)
+
         if _get_call_node(f.body, "relax.nn.attention") is not None:
             op_attrs = _get_call_node(f.body, "relax.nn.attention").attrs
         elif _get_call_node(f.body, "relax.nn.attention_bias") is not None:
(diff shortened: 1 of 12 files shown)


python/tvm/contrib/cutlass/build.py
```diff
@@ -862,10 +862,13 @@ def handle_matmul(self, f, op_type):
     def handle_attention(self, f, op_type):
         """Annotate an attention op."""
         signature = _extract_relax_function_signature(f)
+
         if _get_call_node(f.body, "relax.nn.attention") is not None:
             op_attrs = _get_call_node(f.body, "relax.nn.attention").attrs
         elif _get_call_node(f.body, "relax.nn.attention_bias") is not None:
             op_attrs = _get_call_node(f.body, "relax.nn.attention_bias").attrs
+        elif _get_call_node(f.body, "relax.nn.attention_var_len") is not None:
```
yelite, line 870: I am wondering if there is flexibility in Relax ops system to unify all these attention variations. The current approach of having a different op name for each variation is okay, but doesn't seem to be scalable if we want to introduce more variations (like var_len + bias, customized mask)

yelite (approved):

python/tvm/contrib/cutlass/build.py
masahi: yeah, it would be nice if we can define an "attention input" object to pass all inputs, rather than individually.

vinx13 (approved):

python/tvm/contrib/cutlass/build.py
sunggg: Or another alternative could be having a single attention with a superset of attributes for different variants of attention. Based on how the attributes are set, we may be able to differentiate different attentions.

tests/python/relax/test_codegen_cutlass.py
```diff
@@ -1992,5 +1992,107 @@ def main(
...
+            np.array(seq_lens, dtype="int32"),
+        ],
+        "cuda",
+    )
+
+    tvm.testing.assert_allclose(out, ref, rtol=1e-2, atol=1e-2)
+
+    ############# xformer reference for verification #############
+
+    # attn_bias = BlockDiagonalCausalMask.from_seqlens(seq_lens)
```
sunggg, line 2082: Are these commented out due to the extra dependency?

masahi: yes, it depends on torch and xformer.

python/tvm/contrib/cutlass/build.py
masahi: That doesn't work because different kinds of attention op require different inputs, not attributes. And Relax doesn't support variable-length inputs.

PR #15839, opened 2023-09-28
[Unity] Include last kernel launch in captured CudaGraph

Lunderberg:
Prior to this commit, the last kernel launch would not be included in
a captured CUDA graph.  This commit updates `RewriteCUDAGraph` to
include the last kernel launch.

The previous implementation assumed that any calls to
`R.builtin.alloc_tensor` that remain after `StaticPlanBlockMemory` are
dynamic allocations.  This is not the case, as the allocation of a
static-shaped output tensor may still use `R.builtin.alloc_tensor`.
The primary change of this commit was to update `RewriteCUDAGraph` to
check for static allocations directly, rather than inferring a static
allocation based on the operation being used.

This change triggered an additional bug, in which the previous
implementation only checked for output variables if they occurred as
part of a `VarBinding`, and not if they occurred as the body of a
`SeqExpr`.  As a result, a captured CUDA graph whose output was
immediately used as the output of the containing Relax function would
contain an undefined variable.  This commit updates `RewriteCUDAGraph`
to operate on a `SeqExpr` rather than a `BindingBlock`, so that the
`SeqExprNode::body` may be inspected for output variables.

### DIFF
--- a/python/tvm/relax/vm_build.py
+++ b/python/tvm/relax/vm_build.py
@@ -310,10 +310,11 @@ def build(
     passes.append(relax.transform.RemovePurityChecking())
     passes.append(relax.transform.CallTIRRewrite())
     passes.append(relax.transform.StaticPlanBlockMemory())
-    passes.append(relax.transform.KillAfterLastUse())
 
     if tvm.transform.PassContext.current().config.get("relax.backend.use_cuda_graph", False):
         passes.append(relax.transform.RewriteCUDAGraph())
+
+    passes.append(relax.transform.KillAfterLastUse())
 
     passes.append(relax.transform.VMBuiltinLower())
     passes.append(relax.transform.VMShapeLower())
--- a/src/relax/transform/rewrite_cuda_graph.cc
+++ b/src/relax/transform/rewrite_cuda_graph.cc
@@ -222,14 +222,13 @@ class CUDAGraphRewritePlanner : public ExprVisitor {
 
   void VisitBinding_(const VarBindingNode* binding, const CallNode* call) final {
     static const auto& mem_alloc_storage_op = Op::Get("relax.memory.alloc_storage");
-    static const auto& mem_kill_storage_op = Op::Get("relax.memory.kill_storage");
     static const auto& builtin_alloc_tensor_op = Op::Get("relax.builtin.alloc_tensor");
     static const auto& call_builtin_with_ctx_op = Op::Get("relax.call_builtin_with_ctx");
 
     if (call->op.same_as(mem_alloc_storage_op) && IsStaticAllocStorage(binding)) {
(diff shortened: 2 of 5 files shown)


Lunderberg:
As part of this implementation, moving the `R.memory.kill_tensor` and `R.memory.kill_storage` to be generated in a later `KillAfterLastUse` pass meant that the `RemoveCUDAGraph` pass did not need to track these calls.

masahi:
cc @vinx13

vinx13:
I didn't consider `R.builtin.alloc_tensor` previously because it can be either dynamic allocation or memory visible to external (tensors for input/output). The implication here is running the model with cuda graph enabled will overwrite the output tensors of previous runs.

Lunderberg:
I'm afraid I don't quite understand.  Either `R.memory.alloc_storage` or `R.builtin.alloc_tensor` could be independent allocations that could be returned as output.  (e.g. A static memory plan allocations using `R.memory.alloc_storage`, reuses that buffer several times internally, before returning it with the final result.)

Oh, I think I see.  That would mean that its relying on the current behavior of `StaticBlockPlanMemory`, and assumes that the user-provided parameters are identical across each call.  Effectively, the last kernel launch would be also serving as a memcpy from the memory used by the captured cudagraph to other memory space.  Is that understanding correct?

vinx13:
That's correct. It's relying on the current behavior of `StaticBlockPlanMemory` that doesn't rewrite the input/output tensors. So only tensors from `R.builtin.alloc_tensor` can be returned (similar to the current implementation of this pass, `StaticBlockPlanMemory` doesn't rewrite the body of `SeqExpr`).

Lunderberg:
> It doesn't assume that the user-provided parameters are identical across each call since kernels taking user-provided parameters (usually the first kernel) are not captured.

Isn't that assumption present in `test_capture_fixed_inputs` at [this point](https://github.com/apache/tvm/blob/unity/tests/python/relax/test_transform_rewrite_cuda_graph.py#L620-L638), since the first provided parameters will be used for later calls?  Though, perhaps we're using "user-provided" to mean two separate things.  I was thinking of it as "the user of the relax function", but it could also mean "the user of the framework in which the relax function exists".  The `params` input would be a "user-provided parameter" by the first definition, but not by the second definition.

Lunderberg:
I think I'll close this PR, and perhaps later submit a modified version that explicitly encodes the logic of preventing user-owned buffers from being overwritten.  That way, the `RewriteCUDAGraph` pass could be agnostic to its location within the lowering flow, and the implementation details of `StaticPlanBlockMemory`.

vinx13:
`params` is an exception of the user-inputs. Since it’s usually fixed we use `num_input` func attribute to identify such input and slow capturing.

PR #15840, opened 2023-09-28
[Unity] Avoid trivial var-to-var bindings in CanonicalizeBindings

Lunderberg:
Prior to this commit, the `relax.transform.CanonicalizeBindings` transform would detect trivial bindings `var_y = var_x`, and replace later usage of `var_y` with `var_x`.  However, the trivial binding `var_y = var_x` would be left in the canonicalized function.

This commit updates the `CanonicalizeBindings` transform to remove trivial bindings.  This is not intended as a full dead-code elimination, as that is better handled as a separate pass, but is instead intended to avoid introduction of dead code during canonicalization.

### DIFF
--- a/src/relax/transform/canonicalize_bindings.cc
+++ b/src/relax/transform/canonicalize_bindings.cc
@@ -38,16 +38,6 @@ class BindingCanonicalizer : public ExprMutator {
 
   using ExprMutator::VisitExpr_;
 
-  Expr VisitExpr_(const VarNode* op) override {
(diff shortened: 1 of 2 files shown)


slyubomirsky:
Oh lol I just folded this functionality into #15791 without seeing this. This implementation is slightly more compact so we can use it and I can clean up #15791.

slyubomirsky (approved):
This all seems sound, thank you for the change!

PR #15841, opened 2023-09-29
[Unity] Remove end-of-life handling from StaticPlanBlockMemory

Lunderberg:
The catch-all in `relax.transform.KillAfterLastUse` can be used for releasing tensor/storage, not just those that were not previously handled.  This simplifies the `StaticPlanBlockMemory` pass, as it does not need to track the location at which `R.memory.kill_storage` and `R.memory.kill_tensor` should be inserted.

### DIFF
--- a/src/relax/transform/rewrite_cuda_graph.cc
+++ b/src/relax/transform/rewrite_cuda_graph.cc
@@ -222,14 +222,13 @@ class CUDAGraphRewritePlanner : public ExprVisitor {
 
   void VisitBinding_(const VarBindingNode* binding, const CallNode* call) final {
     static const auto& mem_alloc_storage_op = Op::Get("relax.memory.alloc_storage");
-    static const auto& mem_kill_storage_op = Op::Get("relax.memory.kill_storage");
     static const auto& builtin_alloc_tensor_op = Op::Get("relax.builtin.alloc_tensor");
     static const auto& call_builtin_with_ctx_op = Op::Get("relax.call_builtin_with_ctx");
 
     if (call->op.same_as(mem_alloc_storage_op) && IsStaticAllocStorage(binding)) {
       AddStaticBinding(binding, /*is_alloc_storage=*/true);
       return;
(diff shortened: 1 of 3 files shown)


Lunderberg:
@vinx13 This contains some of the refactoring that was present in https://github.com/apache/tvm/pull/15839, but without the breaking changes to the `RewriteCUDAGraph` pass.

MasterJH5574 (commented):
Hi Eric, thanks for the update for the memory planning pass! I just want to make sure the memory planning still function properly. So for each of the existing test cases, could we have a corresponding test which showcases the effect of “StaticPlanBlockMemory + KillAfterLastUse” with this PR can have the same effect of “killing object” as only using StaticPlanBlockMemory before this PR? I think having this explicitness here can help prove the correctness.

Lunderberg:
MasterJH5574 I'm not sure if that would be the best test cases to add, for a couple of reasons.

1. I fully expect there to be differences between the previous single-pass of `StaticPlanBlockMemory` and the new 2-pass sequence of `SPBM -> KillAfterLastUse`, due to cases that aren't handled by the static memory planning.  (e.g. Failure to de-allocate buffers that are aliased using `R.reshape`, or dynamic buffers.)

2. Writing a test on the application of both passes in sequence assumes that the passes are used adjacently in the lowering flow.  This is already not the case, as `RewriteCUDAGraph` and `LowerAllocTensor` occur between the two.

That said, I think it would be good to add tests verifying that the memory planning does still have the intended reduction of memory use.  I think the best way to do this would be to monitor the allocations/deallocations and verifying that they occur in the correct order.  The allocations could be captured by overriding `"vm.builtin.alloc_storage"`, but deallocations occur based on the reference-counter in `NDArray`.  Do you know of a good way to capture/track those deallocations?

Lunderberg:
Rebased onto current `unity`, as the CI results from three weeks ago are rather stale.  Would appreciate reviews, to avoid the results going stale again.

junrushao (approved):

PR #15844, opened 2023-09-30
[TFLite][Frontend] Support quantized floor_mod

p3achyjr:
Tests were failing b/c TVM and TFLite disagree on div-by-zero cases.

floor_mod still fails, this is probably a behaviorial difference.

### DIFF
--- a/tests/python/frontend/tflite/test_forward.py
+++ b/tests/python/frontend/tflite/test_forward.py
@@ -2452,6 +2452,7 @@ def _test_elemwise(
     qnn_op=None,
     same_qnn_params=False,
     comparison_op=False,
+    exclude_zero_point=False,
 ):
     """One iteration of elemwise"""
 
@@ -2479,6 +2480,16 @@ def _test_elemwise(
             if same_qnn_params:
                 inq0_min, inq0_max = (out_min, out_max)
                 inq1_min, inq1_max = (out_min, out_max)
+
+            if exclude_zero_point:
+                if inq1_max == inq1_min:
+                    raise ZeroDivisionError('Input range is 0.')
+
+                # only compute for rhs.
+                quant_scale = 255 / (inq1_max - inq1_min)
+                zero_point = int(round(-inq1_min * quant_scale))
+                data[1][data[1] == zero_point] += 1
+                data[1][data[1] == 0] += 1
 
             # fake_quant will keep the tensors in float32 until the conversion in the session
             inq_data = [
@@ -2619,6 +2630,7 @@ def _test_div(data, fused_activation_function=None, quantized=False, qnn_op=None
         quantized,
         qnn_op,
         same_qnn_params=True,
+        exclude_zero_point=True,
     )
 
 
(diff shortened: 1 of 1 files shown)


tests/python/frontend/tflite/test_forward.py
```diff
@@ -2480,6 +2481,16 @@ def __test_elemwise(in_data):
...
 
+            if exclude_zero_point:
+                if inq1_max == inq1_min:
+                    raise ZeroDivisionError("Input range is 0.")
+
+                # only compute for rhs.
+                quant_scale = 255 / (inq1_max - inq1_min)
+                zero_point = int(round(-inq1_min * quant_scale))
+                data[1][data[1] == zero_point] += 1
+                data[1][data[1] == 0] += 1
```
Lunderberg, line 2492: Is this second adjustment required?  Looking at the two next two each other, its seems like the first (when `data[1] == zero_point`) would only be required when these are pre-quantized values, and the second (when `data[1] == 0`) would only be required when these are post-quantization values, and that only one of those can be true at a time.

p3achyjr: I agree--but when I was testing, I needed to keep the second adjustment for all tests to pass.

Lunderberg: Interesting, and thank you for checking.  Possibly due to tests that quantize one or the other argument, leading `data` to be used in both contexts.

tests/python/frontend/tflite/test_forward.py
```diff
@@ -2916,11 +2929,6 @@ def test_all_elemwise():
         _test_forward_elemwise(_test_floor_divide)
         _test_forward_elemwise_quantized(_test_floor_divide)
         _test_forward_elemwise(_test_floor_mod)
-        # This test of quantized floor mod is currently disabled due
```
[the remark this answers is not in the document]
p3achyjr: added back--sorry about that

p3achyjr:
@Lunderberg hi! sorry to bug. is this PR still needed?

Lunderberg:
@p3achyjr Thank you for the reminder, and it would still be good to get this PR in, even if re-enabling the flaky `floor_mod` test is saved for a follow-up PR.

Lunderberg (approved):

p3achyjr:
@Lunderberg thanks for getting back!!

tests/python/frontend/tflite/test_forward.py
```diff
@@ -2480,6 +2481,16 @@ def __test_elemwise(in_data):
                 inq0_min, inq0_max = (out_min, out_max)
                 inq1_min, inq1_max = (out_min, out_max)
 
+            if exclude_zero_point:
+                if inq1_max == inq1_min:
+                    raise ZeroDivisionError("Input range is 0.")
+
+                # only compute for rhs.
+                quant_scale = 255 / (inq1_max - inq1_min)
```
Aleksei-grovety, line 2489: Shouldn't there be a dependency on the data type for the value 255? Since the data type can be int8, int16.

PR #15845, opened 2023-09-30
[Unity][Relax]fix behaviors for importing torch.cat/concat

guoyaol:
Fix behaviors for importing torch.cat/concat
Now we can import Stable Diffusion XL and Real-ESRGAN more smoothly

Before fixing, we only support **torch.cat((x, y), dim=n_dim)**

Add support for importing:
1. **torch.cat((x, y))**        with defualt dim=0
2. **torch.cat((x, y), n_dim)**      when user doesn't explicitly state dim=
3. **torch.concat**       an alias of torch.cat


Hzfengsy (approved):

MasterJH5574 (approved):
Thank you @guoyaol for the first contribution!

PR #15849, opened 2023-10-01
[Unity] Add R.quantize and R.dequantize ops

ibsidorenko:
This commit adds 2 new operations (`R.quantize` and `R.dequantize`) and supports them in `LegalizeOps` pass.


ibsidorenko:
cc @masahi @quic-sanirudh

python/tvm/relax/op/qdq.py
```diff
@@ -0,0 +1,88 @@
...
+    Parameters
+    ----------
+    data : tvm.relax.Expr
+        The input tensor to be dequantized.
+
+    scale : tvm.relax.Expr
+        The output scale.
+
+    zero_point : tvm.relay.Expr
+        The output zero_point.
```
masahi, line 74: I think "input" scale / zp would be more correct?

src/relax/op/tensor/qdq.cc
```diff
@@ -0,0 +1,201 @@
...
+
+  // Check datatype of scale param:
+  if (scale_sinfo->dtype != DataType::Float(32) && scale_sinfo->dtype != DataType::Float(16)) {
+    ctx->ReportFatal(Diagnostic::Error(call)
+                     << "scale param datatype should be one of [float16, float32], but got "
+                     << scale_sinfo->dtype);
+  }
+
+  // Check datatype of zero_point param:
+  if (zp_sinfo->dtype != DataType::Int(32)) {
```
masahi, line 76: I remember that zp needs to be 8 bit in practice. How about enforcing int8 here?

src/relax/op/tensor/qdq.h
```diff
@@ -0,0 +1,61 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  Sex The License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/*!
+ * \file src/relax/op/tensor/qdq.h
+ * \brief The functions to make Relax quantize/dequantize operator calls.
```
masahi, line 22: Is this header needed?

masahi (commented):
Also needs to fix the conflict.

src/relax/op/tensor/qdq.cc
ibsidorenko: I think it is possible. I have never met the cases when `zero_point` is out of [-128, 127] range.

src/relax/op/tensor/qdq.h
ibsidorenko: There's no need for that in this PR. But it was added for the future when we want to construct `quantize`/`dequantize` op in one of TVM passes.

python/tvm/relax/op/qdq.py
ibsidorenko: Done

masahi (approved):

PR #15850, opened 2023-10-01
[Unity] Avoid Using Wildcard Imports

Hzfengsy:
Wildcard importing is not recommended in Python projects: https://docs.python.org/3/tutorial/modules.html#importing-from-a-package

I acknowledge that disabling it will bring extra effort to adding new operators, analysis and transforms. But it will make the code clear, also benefit the API reference doc.

cc @junrushao @tqchen


junrushao:
This is great and is something I've been thinking for a while! Thanks for taking actions getting it fixed!

junrushao (approved):

PR #15810, opened 2023-09-22
[Unity] Implement relax.transform.KillAfterLastUse

Lunderberg:
Prior to this commit, intermediate objects produced while executing a Relax function would persist until the end of the Relax function. While re-use of static allocations is handled by the `StaticPlanBlockMemory` transform, re-use of dynamic allocations is handled by the `relax_vm::PooledAllocator`.  For large Relax functions representing end-to-end model execution, releasing memory from the VM registers to the `relax_vm::PooledAllocator` at the end of the function call may be insufficient.

This commit introduces a new pass, `relax.transform.KillAfterLastUse`, which identifies the last usage of each Relax variable and inserts a `relax.memory.kill_tensor`, `relax.memory.kill_storage`, or `relax.vm.kill_object` call depending on the object type.  This insertion is suppressed if a Relax variables is already killed, such as static allocations and tensors tracked by `StaticPlanBlockMemory`.

PR #15852, opened 2023-10-01
Revert "[Unity] Implement relax.transform.KillAfterLastUse (#15810)"

junrushao:
This reverts commit aa4587feb5103927d95e5e931149debd0a0aeafc.

Unfortunately, this PR broke MLC LLM's build pipeline, more specifically, the command below

```python
python3 -m mlc_llm.build --model dist/models/llama-2-13b-chat-hf/ --quantization q4f16_1
```

leads to the the following error message:

<details>

```
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/opt/scratch/junrushao/mlc-llm/mlc_llm/build.py", line 13, in <module>
    main()
  File "/opt/scratch/junrushao/mlc-llm/mlc_llm/build.py", line 10, in main
    core.build_model_from_args(parsed_args)
  File "/opt/scratch/junrushao/mlc-llm/mlc_llm/core.py", line 616, in build_model_from_args
    new_params = utils.convert_weights(param_manager, params, args)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/scratch/junrushao/mlc-llm/mlc_llm/utils.py", line 258, in convert_weights
    vm["transform_params"]()
  File "/opt/scratch/junrushao/tvm-dev/python/tvm/_ffi/_ctypes/packed_func.py", line 239, in __call__
    raise_last_ffi_error()
  File "/opt/scratch/junrushao/tvm-dev/python/tvm/_ffi/base.py", line 476, in raise_last_ffi_error
    raise py_err
  File "/opt/scratch/junrushao/tvm-dev/src/runtime/relax_vm/vm.cc", line 634, in tvm::runtime::relax_vm::VirtualMachineImpl::InvokeClosurePacked(tvm::runtime::ObjectRef const&, tvm::runtime::TVMArgs, tvm::runtime::TVMRetValue*)
    clo->impl.CallPacked(TVMArgs(values.data(), tcodes.data(), args.size() + 1), rv);
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/scratch/junrushao/tvm-dev/src/runtime/relax_vm/vm.cc", line 708, in operator()
    *rv = static_cast<VirtualMachineImpl*>(ctx_ptr)->InvokeBytecode(gf_idx, inputs);
[...]

PR #15854, opened 2023-10-02
[Unity][Bugfix] Check for trivial re-bindings in MatchCast nodes

Lunderberg:
The `KillAfterLastUse` pass that was implemented in https://github.com/apache/tvm/pull/15810 checked for trivial re-bindings in `VarBinding` nodes, but not in `MatchCast` nodes.  As a result, `CodeGenVM`'s de-duplication of registers resulted in the object being killed prematurely.

```python
y = R.match_cast(x, R.Tensor(...)) # Trivial rebinding here.
                                   # CodeGenVM has these share a register.

R.memory.kill_tensor(x)  # Kill x after last usage.
                         # Register is set to None.

_ = R.ExternFunc("set_item")(y) # Use of the cleared register through y.
```

### DIFF
--- a/src/relax/transform/kill_after_last_use.cc
+++ b/src/relax/transform/kill_after_last_use.cc
@@ -49,6 +49,13 @@ class UnusedTrivialBindingRemover : public ExprMutator {
         has_trivial_binding.insert(binding->var.get());
         ExprVisitor::VisitBinding_(binding, val);
       }
+      void VisitBinding_(const MatchCastNode* binding) override {
+        if (binding->value.as<VarNode>() &&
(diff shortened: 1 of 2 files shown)


Lunderberg:
@junrushao This change should resolve the MLC-LLM error you found in https://github.com/apache/tvm/pull/15852.  It works in my local setup, and it would be good to validate that it works on your side as well.

junrushao (approved):

masahi:
Just hit the issue of `ValueError: Do not know how to handle type object`, fixed by this PR. Thanks.

PR #15856, opened 2023-10-03
[Unity][CUTLASS] Fix hardcoded epsilon parameter in RMS norm

masahi:
The CUTLASS RMS kernel has been using a hard-coded epsilon value of 1e-6. I learned that recent popular language models such as llama 2 have been trained with eps = 1e-5, and using a slightly different eps can cause accuracy problems for such models on some inputs (see https://github.com/mlc-ai/mlc-llm/issues/923#issuecomment-1735066112). This has been fixed by this PR.

@zxybazh @sunggg @vinx13


zxybazh (approved):
Thanks for the fix, LGTM.

PR #15864, opened 2023-10-03
[Unity] Include LegalizeOps in the default relax.build lowering flow

Lunderberg:
Prior to this commit, `relax.transform.LegalizeOps` needed to be called prior to `relax.build`.  This commit adds `LegalizeOps` to the lowering flow, to simplify the calling steps for an end-user.  If the `IRModule` contains no legalizable functions, a second legalization pass has no effect.

Some test cases relied on this behavior as an implicit assertion that operator fusion patterns applied.  That is, by omitting `LegalizeOps`, a successful compilation `relax.build` would only occur if all legalizable operators have already been removed, and so an incorrect fusion pattern would result in a failure to build the module.  While these tests would be better expressed by comparing against an expected fused pattern, updating the tests is outside the scope of this PR.  To allow these tests to keep their implicit assertions, a `"relax.transform.apply_legalize_ops"` config can be used to disable the `LegalizeOps` pass.

### DIFF
--- a/python/tvm/relax/vm_build.py
+++ b/python/tvm/relax/vm_build.py
@@ -307,6 +307,7 @@ def build(
 
     lowering_passes = tvm.transform.Sequential(
         [
+            relax.transform.LegalizeOps(),
             relax.transform.RewriteDataflowReshape(),
             relax.transform.ToNonDataflow(),
             relax.transform.RemovePurityChecking(),
--- a/src/relax/transform/legalize_ops.cc
+++ b/src/relax/transform/legalize_ops.cc
@@ -31,6 +31,8 @@
 namespace tvm {
 namespace relax {
 
+TVM_REGISTER_PASS_CONFIG_OPTION("relax.transform.apply_legalize_ops", Bool);
+
 /*!
  * \brief Check if a given Tensor/Shape/TupleStructInfo contains shapes whose
  * values are all known.
@@ -206,7 +208,12 @@ namespace transform {
 Pass LegalizeOps(Optional<Map<String, PackedFunc>> cmap, bool enable_warning) {
   runtime::TypedPackedFunc<IRModule(IRModule, PassContext)> pass_func = [=](IRModule mod,
                                                                             PassContext pc) {
-    return LegalizeMutator(mod, cmap, enable_warning).Transform();
+    bool apply_legalize_ops =
+        pc->GetConfig<Bool>("relax.transform.apply_legalize_ops").value_or(Bool(true))->value;
+    if (apply_legalize_ops) {
+      mod = LegalizeMutator(mod, cmap, enable_warning).Transform();
+    }
+    return mod;
   };
   return CreateModulePass(/*pass_function=*/pass_func,
                           /*opt_level=*/0,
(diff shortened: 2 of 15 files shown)


junrushao:
I'm not super sure if it's desirable. Usually after operator lowering, we will apply a series of transformations to reach the end state of the desirable TIR, which we could abstract with `relax.pipeline`, while the default `relax.build` is kept as minimum as possible to be compatible with all potential relax pipelines. One direction I would imagine is to introduce more pipelines so that the build process could be more streamlined

quic-sanirudh:
> I'm not super sure if it's desirable. Usually after operator lowering, we will apply a series of transformations to reach the end state of the desirable TIR, which we could abstract with `relax.pipeline`, while the default `relax.build` is kept as minimum as possible to be compatible with all potential relax pipelines. One direction I would imagine is to introduce more pipelines so that the build process could be more streamlined

I think having LegalizeOps as part of `relax.build` in some format is useful as we've had [multiple queries in our discuss forum](https://discuss.tvm.apache.org/t/relax-permute-dims-not-supported/15635/4) about unsupported operators, which just turned out to be because users were not aware of having to call LegalizeOps explicitly before `relax.build`.

I also agree with the idea of keeping relax.build as minimal as possible and using `relax.pipeline` for the rest, so would it be better if update `relax.build` to take a relax pipeline as an argument and run the passes in the pipeline as part of the build process. That way, we can add a small pipeline with just `LegalizeOps` and make that the default pipeline taken by `relax.build`, and others who want customization qcan pass their own (probably more complex) pipelines.

Lunderberg:
@junrushao  Absolutely agreed, there's a wide variety of hand-selected transformations that optimized models require, many of which must be performed after `LegalizeOps`.  That is essential functionality, and this PR isn't intended to replace those at all.

Workflow 1: A user writes a model, applies writes a sequence of transformations that have been hand-selected to optimize that specific model into the desired form, then passes the result into `relax.build`.  All operations have already been legalized, so applying `LegalizeOps` inside `relax.build` has no effect.

Workflow 2: A user writes a model, then passes the model directly into `relax.build`.  The model still contains operators that require legalization, so `LegalizeOps` has an effect.

The goal of this PR is to enable Workflow 2 as a simpler way to get started, either for a new user or when running a new model, without impacting the optimizations performed in Workflow 1.

Lunderberg:
Philosophically, I like the distinction @quic-sanirudh is drawing between the minimal `relax.build` and the more expansive options in `relax.pipeline`.  I'd see `relax.build` as a tool which converts from any well-formed relax module to an executable, performing any legalizations that are required, but which avoids performing any optimizations as those are outside of its scope.

Lunderberg:
I rebased onto current `unity`, to make sure that no CI breakages have emerged since then.  Any objections to merging if the CI passes?

quic-sanirudh:
I agree with the workflow 2 you've mentioned here, and since `LegalizeOps` probably has no effect on workflow 1, I'm fine with merging this.

quic-sanirudh (approved):

junrushao:
I don’t object either. Happy to merge it in

Lunderberg:
Sounds good, and thank you!

I'm rebasing the commit on top of unity head, as the current CI failures are due to a breakage on `unity` head, and are resolved with https://github.com/apache/tvm/pull/15941. (See [this comment](https://github.com/apache/tvm/pull/15941#issuecomment-1779377445) for details.)

PR #15870, opened 2023-10-04
[Bugfix] Fix Disco-CUDAGraph Integration

junrushao:
This PR fixed a bug introduced in #15827 since which the cudagraph's stream is discarded.

### DIFF
--- a/src/runtime/disco/nccl/nccl.cc
+++ b/src/runtime/disco/nccl/nccl.cc
@@ -140,9 +140,19 @@ struct CCLThreadLocalContext {
   deviceStream_t default_stream = nullptr;
   ncclComm_t comm;
 
-  void Clear() { NCCL_CALL(ncclCommDestroy(comm)); }
+  void Clear() {
+    NCCL_CALL(ncclCommDestroy(comm));
+    if (default_stream != nullptr) {
+      StreamDestroy(default_stream);
+    }
+  }
 
-  deviceStream_t GetDefaultStream() { return nullptr; }
+  deviceStream_t GetDefaultStream() {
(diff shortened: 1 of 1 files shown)


junrushao:
CC: @jinhongyii

Hzfengsy (approved):

src/runtime/disco/nccl/nccl.cc
```diff
@@ -140,9 +140,19 @@ struct CCLThreadLocalContext {
   deviceStream_t default_stream = nullptr;
   ncclComm_t comm;
 
-  void Clear() { NCCL_CALL(ncclCommDestroy(comm)); }
+  void Clear() {
+    NCCL_CALL(ncclCommDestroy(comm));
+    if (default_stream != nullptr) {
```
jinhongyii, line 145: IIUC, default_stream is always nullptr

tqchen: given in this case `default_stream` is a member that can be changed later, i think having such code is better. Note it is not the same as the default stream in the CUDA terminology, but in the nccl runtime

src/runtime/disco/nccl/nccl.cc
```diff
@@ -140,9 +140,19 @@ struct CCLThreadLocalContext {
...
+      StreamDestroy(default_stream);
+    }
+  }
 
-  deviceStream_t GetDefaultStream() { return nullptr; }
+  deviceStream_t GetDefaultStream() {
+    const auto* func = tvm::runtime::Registry::Get("runtime.get_" TVM_DISCO_DEVICE_NAME "_stream");
```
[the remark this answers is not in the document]
junrushao: it's initialized in cudagraph

jinhongyii (approved):

PR #15872, opened 2023-10-04
[TIR] Autocorrect Range min and extent types when possible

kparzysz-quic:
The values of `min` and `extent` in Range should have the same type, but the usage of convenience may often produce type mismatches, for example `Range(0, dim)` will assume int32 for the `min`. These mismatches can then cause hard-to-track errors, in particular assertions in the arithmetic simplifier.
This is already done for `For` statement, but not in other places.

This patch introduces `DataType::WidestOf` function, that returns the widest type from the given list. It is then used to promote the integer bounds if such widest type exists.

### DIFF
--- a/src/tir/ir/stmt.cc
+++ b/src/tir/ir/stmt.cc
@@ -125,14 +125,11 @@ For::For(Var loop_var, PrimExpr min, PrimExpr extent, ForKind kind, Stmt body,
     ICHECK(loop_var.dtype() == widest) << "ValueError: Loop var type (" << loop_var.dtype()
                                        << ") disagree with range types (min:" << min.dtype()
                                        << ", extent:" << extent.dtype() << ')';
-    if (min.dtype() != widest) {
-      if (tir::is_const_number(min)) {
-        min = tvm::cast(widest, min);
-      }
-    } else if (extent.dtype() != widest) {
-      if (tir::is_const_number(extent)) {
-        extent = tvm::cast(widest, extent);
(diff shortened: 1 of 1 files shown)


tqchen:
In most cases the intention was to construct IRs that have consistent types. So insert casting like `cast(i32_val, i64)` is still not desirable for cases like arith simplifier.

How about we do the following things:
- Auto promote trivial constants, in which case no cast will be introduced
- Error out(instead of silently promote) in constructor for cases that are in-consistent, so we can have callers to fix the type tracking when possible.

include/tvm/runtime/data_type.h
```diff
@@ -237,6 +241,49 @@ class DataType {
...
   }
 
+  /*!
+   * \brief Return the widest type from a given array.
+   *
+   * Given an array of types, return the one of those types that can
+   * hold all values representable in the other types, or "Void" if
+   * such type is not present in the array.
+   */
+  static DataType WidestOf(const std::vector<DataType>& types) {
```
tqchen, line 251: is it possible to use template trick or overloading to handle two/three case, mainly avoid the vector passing

kparzysz-quic: Would `std::initializer_list` be ok?

kparzysz-quic:
I'm not sure what you think.  This is adding a lot of code for a simple thing, but I'm hoping that the extra functions may be useful in other cases as well.

PR #15875, opened 2023-10-04
[Unity][Fix] Remove duplicated words from comments, NFC

kparzysz-quic:
Removed instances of accidentally repeated words from comments. There are cases where duplicated words appear legitimately, those cases remain unmodified.


slyubomirsky (approved):
These are very welcome changes.

PR #15878, opened 2023-10-05
[Unity][Op] Introduce `call_inplace_packed` as a counterpart to `call_tir_inplace`

slyubomirsky:
As discussed in the October 3, 2023, community meeting, for cases like the `KVCache` in LLaMa in mlc-llm, it might be desirable to have a way to have packed calls inside dataflow blocks where the packed calls simply "consume" their output and mutate it in-place (performing no other side effects). While this is not pure, if the other conditions for doing an in-place call are met (no other live aliases of the consumed value), this is safe, though users should use caution if applying it. (The in-place update analyses could also be used to make sure the conditions are met.) This PR introduces a `call_inplace_packed` operator to signal this intent, though unlike `call_tir_inplace`, this operator really only conveys the intent—it is up to the implementation of the packed call to behave as it advertises.

@psrivas2 raised the idea of having a different name, which may be worth discussing. We could consider using the word "consume" in it or referring to linear typing if we do not feel "inplace" conveys the intent well enough.


src/relax/transform/remove_purity_checking.cc
```diff
@@ -52,6 +52,12 @@ class PurityRemover : public ExprMutator {
                       call->attrs, call->sinfo_args);
       return VisitExpr(ret);
     }
+    if (call->op == call_inplace_packed_op_) {
+      // call_inplace_packed has its own attrs so we don't pass those down
+      auto ret = Call(call->args[0], Array<Expr>(call->args.begin() + 1, call->args.end()),
+                      tvm::Attrs(), call->sinfo_args);
```
slyubomirsky, line 58: As far as I know, we never use the attrs for `call_pure_packed`, so we could simply drop those and combine those cases

tests/python/relax/test_relax_operators.py
```diff
@@ -248,6 +250,89 @@ def pure_copy(x: R.Tensor((3, 4), "float32")):
...
+    arr_a = np.random.rand(3, 4).astype("float32")
+    arr_b = np.random.rand(3, 4).astype("float32")
+    sum = arr_a + arr_b
+    tvm_arr_a = tvm.nd.array(arr_a)
+    tvm_arr_b = tvm.nd.array(arr_b)
+    result = run_cpu(CallInplaceTuple, "inplace_tuple", tvm_arr_a, tvm_arr_b)
+    assert result[0] == tvm_arr_a
+    assert (result[0].numpy() == sum).all()
+    assert result[1] != tvm_arr_a and result[1] != tvm_arr_b
+    assert (result[1].numpy() == sum).all()
```
slyubomirsky, line 333: If we want, we can check reference equality/disequality at run time as well. I'm not sure it would be worth it, though.

masahi:
> it might be desirable to have a way to have packed calls inside dataflow blocks

Very strong +1 for this need. This week I integrated a KV cache update kernel, and faced a problem of how to call this op when the module being defined is already under a dataflow scope. To workaround this, I ended up [returning in-place updated arguments from the kernel ](https://github.com/masahi/tvm/blob/contrib-vllm/src/runtime/contrib/vllm/cache_kernels.cu#L88) to make it look like a pure operation from outside, and [call it](https://github.com/masahi/mlc-llm/blob/llama-batched-vllm/mlc_llm/relax_model/llama_batched.py#L163-L174) via `call_pure_packed`. Although it gets the job done, it feels very awkward to return in-place updated arguments. I just wished that I could simply call `_ = call_packed(...)` inside a dataflow block and have transform passes not remove this dummy binding.

slyubomirsky:
Were you getting a well-formed error?

masahi:
No, assuming that such check occurs automatically during `relax.build()`.

slyubomirsky:
Good to know. I'm glad it gets caught, though I'm curious how LLaMa is able to build and run despite also having packed calls in it.

masahi:
I mean, the KV cache update operation is disguised as a pure operation in the model. So it does build and run, that's no problem. But semantically it is not quite right and I also need to return the updated cache from the model (even thought it is in-place updated) so that passes that remove unused bindings in a df block don't accidentally remove the KV cache update op.

yongwww (approved):
lgtm

tqchen (approved):

PR #15883, opened 2023-10-06
[Unity][Transform] Allow static Relax arguments to dynamic PrimFunc

Lunderberg:
Prior to this commit, the `relax.transform.FuseTIR` transform required that the shapes arguments passed into a `PrimFunc` be structurally equivalent to the shapes of the parameters, and that any replacement of symbolic `tir.Var` be with a symbolic `tir.Var` in the fused function.

This commit updates the `SymbolicMatcher` to instead extract a `Map<tir::Var, PrimExpr>`.  As a result, a Relax tensor with statically-known shape can be passed into a TIR PrimFunc with dynamic shape.  The resulting fused TIR function is in terms of the statically-known shape, and no longer contains the symbolic variable.

### DIFF
--- a/src/relax/transform/fuse_tir.cc
+++ b/src/relax/transform/fuse_tir.cc
@@ -62,10 +62,6 @@ class SymbolicMatcher : ExprFunctor<bool(const PrimExpr& n, const PrimExpr& othe
     } else {
       return ExprFunctor::VisitExpr(node, other);
     }
-    // bool matched = node.same_as(other) || ((node->type_index() == other->type_index()) &&
-    //                                     node.dtype().code() == other.dtype().code());
-
-    // return matched && ExprFunctor::VisitExpr(node, other);
   }
 
 #define TVM_DECLARE_SYMBOLIC_MATCHER_BINOP(OpName)               \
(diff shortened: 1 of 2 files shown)


masahi:
cc @Hzfengsy

src/relax/transform/fuse_tir.cc
```diff
@@ -39,31 +39,41 @@ namespace tir {
...
-    return matched && ExprFunctor::VisitExpr(n, other);
+  bool VisitExpr(const PrimExpr& node, const PrimExpr& other) {
+    if (node.same_as(other)) {
+      return true;
+    } else if (node.dtype().code() != other.dtype().code()) {
+      return false;
+    } else {
+      return ExprFunctor::VisitExpr(node, other);
+    }
+    // bool matched = node.same_as(other) || ((node->type_index() == other->type_index()) &&
```
Hzfengsy, line 65: Please remove useless lines

Hzfengsy (approved):
General LGTM.

src/relax/transform/fuse_tir.cc
Lunderberg: Thank you for the catch, and removed.

PR #15884, opened 2023-10-07
[Unity][Relax] Support Dynamic Tensor as Index, torch frontend

guoyaol:
Add support: allow one dynamic tensor as torch.tensor's index

In Stable Diffusion XL:
e.g.  `output = input1 [input2.argmax(dim = -1), ]`

**Before:** we can only support int, slice, elipsis, None
**Now:** we can have one dynamic torch.tensor in index, which allows us to represent SD XL correctly. 

Behaviors **aligns with torch.tensor**:
as shown in test cases

**Future TODO:** Support other "fancy indexing" methods in torch.tensor, (e.g. multiple tensors as index) which requires some new relax ops


Hzfengsy (approved):

Hzfengsy:
Thanks for the fix!

PR #15888, opened 2023-10-07
[Unity][Disco] Add runtime.disco.scatter_from_local

jinhongyii:
This is an initial step for supporting shard loading without communication


jinhongyii:
cc: @junrushao

PR #15898, opened 2023-10-09
[Unity][Testing] Show failing module in WellFormedInstrument

Lunderberg:
Prior to this commit, the `WellFormedInstrument` showed any error messages produced by the call to `relax.analysis.well_formed`, but did not show the ill-formed module itself.  Since this instrumentation is enabled for all unit tests in the `tests/python/relax` directory, CI logs do not include the failing module, making the logs less useful for debugging.

This commit updates the `WellFormedInstrument` to display the ill-formed module prior to raising an exception.

### DIFF
--- a/python/tvm/relax/ir/instrument.py
+++ b/python/tvm/relax/ir/instrument.py
@@ -25,13 +25,19 @@ class WellFormedInstrument:
     is well formed. It will skip specific passes, like Normalize.
     """
 
-    def __init__(self):
+    def __init__(self, check_struct_info=True):
         self.skip_pass_name = ["Normalize", "ResolveGlobals"]
+        self.check_struct_info = check_struct_info
 
     def run_before_pass(self, mod, pass_info):
-        if pass_info.name not in self.skip_pass_name:
-            assert relax.analysis.well_formed(mod)
+        self._check(mod, pass_info.name, "Before")
 
     def run_after_pass(self, mod, pass_info):
-        if pass_info.name not in self.skip_pass_name:
-            assert relax.analysis.well_formed(mod)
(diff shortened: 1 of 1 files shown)


python/tvm/relax/ir/instrument.py
```diff
@@ -25,13 +25,19 @@ class WellFormedInstrument:
...
 
     def run_before_pass(self, mod, pass_info):
-        if pass_info.name not in self.skip_pass_name:
-            assert relax.analysis.well_formed(mod)
+        self._check(mod, pass_info.name, "Before")
 
     def run_after_pass(self, mod, pass_info):
-        if pass_info.name not in self.skip_pass_name:
-            assert relax.analysis.well_formed(mod)
+        self._check(mod, pass_info.name, "After")
```
sunggg, line 36: Nit: Do we need in both before and after? If we have consecutive passes, we would have redundant checks running.

sunggg (commented):
Thank you! One nit.

python/tvm/relax/ir/instrument.py
Lunderberg: When doing an end-to-end lowering, it's definitely overkill, but this utility is primarily used for the relax unit tests.  There, I think it's helpful in establishing whether a transformation caused a module to be ill-formed, or whether the usually-hand-written input module was ill-formed from the start.

sunggg: got it. yeah, for testing purpose, I think it is okay. we can revisit in the future if necessary.

sunggg (approved):

vinx13 (approved):

PR #15900, opened 2023-10-09
[Unity] Ignore R.ExternFunc in EliminateCommonSubexpr

Lunderberg:
Prior to this commit, `relax::ExternFunc` nodes would be de-duplicated as part of the `EliminateCommonSubexpr` pass.  This commit instead ignores the `relax::ExternFunc` nodes, retaining the in-line definitions.

### DIFF
--- a/tests/python/relax/test_transform_cse.py
+++ b/tests/python/relax/test_transform_cse.py
@@ -276,5 +276,19 @@ def test_do_not_eliminate_shape_expr():
     verify(Before, Expected)
 
 
+def test_do_not_eliminate_extern_func():
+    @I.ir_module
+    class Before:
+        @R.function(pure=False)
+        def foo(x: R.Tensor((2, 3), dtype="float32")):
+            y = R.call_packed("extern_func_name", x, sinfo_args=R.Tensor([2, 3]))
+            z = R.call_packed("extern_func_name", y, sinfo_args=R.Tensor([2, 3]))
+            return z
+
+    Expected = Before
+
+    verify(Before, Expected)
+
+
 if __name__ == "__main__":
     tvm.testing.main()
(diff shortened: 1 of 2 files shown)


tests/python/relax/test_transform_cse.py
```diff
@@ -276,5 +276,19 @@ def foo(x: R.Tensor((2, 3), dtype="float32"), y: R.Tensor((2, 3), dtype="float32
     verify(Before, Expected)
 
 
+def test_do_not_eliminate_extern_func():
+    @I.ir_module
+    class Before:
+        @R.function(pure=False)
+        def foo(x: R.Tensor((2, 3), dtype="float32")):
+            y = R.call_packed("extern_func_name", x, sinfo_args=R.Tensor([2, 3]))
+            z = R.call_packed("extern_func_name", y, sinfo_args=R.Tensor([2, 3]))
```
kparzysz-quic, line 285: If this was commoned before, then something else is probably wrong. This is equivalent to `extern_func(extern_func(x))`, so there is no opportunity to common anything even if the external function was known to be pure.

Lunderberg: Previously, it wasn't a computational step that was being extracted as the common subexpression, but the `R.ExternFunc` object itself.

```python
@I.ir_module
class Module:
    @R.function(pure=False)
    def foo(x: R.Tensor((2, 3), dtype="float32")):
        gv = R.ExternFunc("extern_func_name")
        y = gv(x, sinfo_args=(R.Tensor((2, 3)),))
        z = gv(y, sinfo_args=(R.Tensor((2, 3)),))
        return z
```

kparzysz-quic: I see.  Yeah, that seems unnecessary.

kparzysz-quic (approved):

PR #15901, opened 2023-10-09
[Unity][Schedule] Prototype of peeling tail iterations

kparzysz-quic:
If iv is a loop induction variable, then `m, t = sch.peel_tail(iv, 1)` will convert
```
  for iv in T.serial(n):
    blah
```
to
```
  for m in T.serial(n-1):
    blah
  for t in T.serial(1):
    blah
```

This is still wip.

### DIFF
--- a/src/tir/schedule/primitive/loop_transformation.cc
+++ b/src/tir/schedule/primitive/loop_transformation.cc
@@ -454,6 +454,56 @@ Array<StmtSRef> Split(ScheduleState self, const StmtSRef& loop_sref, const Array
   return result_srefs;
 }
 
+Array<StmtSRef> Peel(ScheduleState self, const StmtSRef& loop_sref, PrimExpr iter_count,
+                     bool preserve_unit_iters) {
+  const ForNode* loop = TVM_SREF_TO_FOR(loop_sref);
+  if (!loop->annotations.empty() || loop->thread_binding.defined()) {
+    throw HasAnnotationOrThreadBindingError(self->mod, GetRef<For>(loop));
+  }
+
+  Var main_var = loop->loop_var.copy_with_suffix("_main");
+  Var tail_var = loop->loop_var.copy_with_suffix("_tail");
+  Map<Block, Block> opaque_block_reuse;
+
+  auto make_loop_body = [&opaque_block_reuse](Var old_var, PrimExpr new_expr, Stmt body) -> Stmt {
+    SubstituteVarAndCollectOpaqueBlock rewriter(
+        [old_var, new_expr](const Var& v) -> Optional<PrimExpr> {
+          return v.same_as(old_var) ? Optional<PrimExpr>(new_expr) : NullOpt;
+        },
+        &opaque_block_reuse);
+    return rewriter(body);
+  };
+
+  PrimExpr main_extent = loop->extent - iter_count;
+  PrimExpr tail_extent = iter_count;
+
+  // Loop before peeling:     iv:[min, min + extent)
+  // Main loop after peeling: iv0:[0, main_extent)  iv -> iv0 + min
+  // Tail loop after peeling: iv1:[0, tail_extent)  iv -> iv1 + min + main_extent
+  Stmt main_body = make_loop_body(loop->loop_var, main_var + loop->min, loop->body);
+  Stmt tail_body = make_loop_body(loop->loop_var, tail_var + loop->min + main_extent, loop->body);
+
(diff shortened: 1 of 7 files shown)


kparzysz-quic:
Hey @junrushao, @jinhongyii, @Lunderberg, @masahi.  This is still WIP, but I wanted to see if you have any comments or concerns about it so far.

Lunderberg:
I like it overall.  My general view is that anything that can be done automatically while lowering should also be doable manually while scheduling, and this would enable schedule-time loop partitioning.

Instead of having a specific utility for peeling loop iterations off of the end, and we have it apply a general loop partition?  Something like `sch.partition_loop(partition_extents)`, where `partition_extents` is a list of extents for each extracted iterator.  Similar to `.reshape()`, up to one of the extents could be `-1`, indicating that it should be inferred from the extent of the partitioned iterator and the other extents in the partition.  That way, we could express "peel off the last iteration" as `sch.partition_loop([-1, 1])`, but could also express "peel off the first iteration" as `sch.partition_loop([1, -1])`, and "peel off 4 elements from both sides" as `sch.partition_loop([4, -1, 4])`.

src/tir/schedule/primitive/loop_transformation.cc
```diff
@@ -454,6 +454,56 @@ Array<StmtSRef> Split(ScheduleState self, const StmtSRef& loop_sref, const Array
...
+  Stmt tail_body = make_loop_body(loop->loop_var, tail_var + loop->min + main_extent, loop->body);
+
+  Stmt main_loop = IterMapSimplifyBlockBinding::SimplifyBindings(
+      For(main_var, 0, main_extent, ForKind::kSerial, main_body),  // initial loop
+      GetLoops(loop_sref), opaque_block_reuse.CopyOnWrite(), preserve_unit_iters);
+  Stmt tail_loop = IterMapSimplifyBlockBinding::SimplifyBindings(
+      For(tail_var, 0, tail_extent, ForKind::kSerial, tail_body),  // initial_loop
+      GetLoops(loop_sref), opaque_block_reuse.CopyOnWrite(), preserve_unit_iters);
+
+  BlockRealize main_block({}, main_extent > 0, Block({}, {}, {}, "peel_main", main_loop));
```
Lunderberg, line 493: Should we also check whether the block contains subblocks as part of the body, in case those need to be renamed?

kparzysz-quic: Probably.  I'm learning this as I go, so if you think so, then probably yes.  :)

rutkoor:
Hi @Lunderberg,
@kparzysz-quic may not continue to work on this PR, I will try to complete the changes suggested in this PR.  

If I understood the above suggestion, you are proposing generic loop-partitioning scheduling primitive rather than peeling just tail iterations. Can you please confirm, if you are looking into something like below example?

```
for iv in range(n):
    a[i] = b[i] + c[i]

```
Given the above loop, if we want to partition it into three parts, with head iterations as 4, body iterations as n - 4 - 5 and tail iterations as 5, we can use a scheduling primitive like, `sch.partition_loop(iv, 4, 5)`, where `iv` is a loop iteration variable.

```
for iv in range(4):
    a[i] = b[i] + c[i]
for iv in range(4, n-5):
    a[i] = b[i] + c[i]
for iv in range(n-5, n):
    a[i] = b[i] + c[i]
```

The design(`sch.partition_loop(iv, head_iter, tail_iter)`) that I am thinking is, to have head_iterations and tail_iterations as arguments to the sch primitive, which will give fine control to users to peel from start/end of loops.

Please confirm the design, so that I will go ahead and start working on the remaining changes.


Thanks

Lunderberg:
@rutkoor That design seems reasonable as well.  I was thinking that the full partitioning would be more general and may be easier to implement, but being able to peel off the head/tail also works.

rutkoor:
> @rutkoor That design seems reasonable as well. I was thinking that the full partitioning would be more general and may be easier to implement, but being able to peel off the head/tail also works.

Hi @Lunderberg, as per your suggestions I have made necessary changes and finished development of generic loop-partitioning scheduling primitive (https://github.com/apache/tvm/pull/16431). Please take a look at it and let me know if it is aligned with your expectations. 

Thanks.

rutkoor:
The PR (https://github.com/apache/tvm/pull/16431) which introduces loop-partition scheduling primitive got merged. This branch can be safely closed now.

PR #15904, opened 2023-10-09
[Unity][Transform] Canonicalize and use CSE between pattern matches

Lunderberg:
The `PatternRewriter` is intended to iterate until no matching patterns remain, as implemented in #14446 and #15495.  Prior to this commit, this only involved repeating the pattern match rewrite rules.  However, intermediate results produced by pattern replacement could cause the iterative pattern matching to terminate early.

* If two rewrite rules each introduce the same intermediate, there will exist two copies of that intermediate, which can prevent `only_used_by` patterns from matching.  Applying `EliminateCommonSubexpr` allows the pattern matching to continue.

* Applying a rewrite rule may result in dangling intermediates that are no longer used.  These dangling intermediates may prevent the next application of a rewrite rule that uses the `only_used_by` constraint.  Applying `RemoveAllUnused` allows the pattern matching to continue.

* A rewrite rule that returns a `relax::Var` or `relax::TupleGetItem` as the replacement introduces trivial var-to-var rebinding, which are not tracked by `PatternRewriter`.  Applying `CanonicalizeBindings` allows the pattern matching to continue.

While this could be fixed externally by repeatedly applying `rewrite_call`, this would require re-inspecting the entire function, and not just the dataflow block in which the replacement occurred.

### DIFF
--- a/python/tvm/relax/transform/remove_redundant_reshape.py
+++ b/python/tvm/relax/transform/remove_redundant_reshape.py
@@ -66,13 +66,18 @@ class RemoveRedundantReshape:
                 continue
 
             def rewriter(expr, matches):
-                args = matches[self.pattern]
+                arg = matches[self.input1]
+
                 if self.repeated_reshape in matches:
-                    return relax.op.reshape(matches[self.input1], args.args[1])
+                    output_shape = matches[self.repeated_reshape].args[1]
+                    return relax.op.reshape(arg, output_shape)
+
                 elif self.no_op_reshape in matches:
-                    if args.args[0].struct_info.shape:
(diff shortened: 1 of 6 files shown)


masahi (approved):
Makes sense, please fix the tests.

Lunderberg:
Current CI failure is a segfault while generating TVMScript output.  The segfault is caused by a dangling reference, and is resolved by #15923.

masahi:
@tvm-bot rerun

Lunderberg:
Thank you for re-launching the CI, and that's a weird HTTP error from it.  I've started it up again, and hopefully it won't have the same issue.

PR #15909, opened 2023-10-10
[IR] Make sure Range types match when binding to Var

kparzysz-quic:
The values of `min` and `extent` in Range should have the same type, but the usage of convenience may often produce type mismatches, for example `Range(0, dim)` will assume int32 for the min. These mismatches can then cause hard-to-track errors, in particular assertions in the arithmetic simplifier. Such problems are likely to occur when a mismatched range is bound to a variable.

This patch identifies a few occurrences where a Range is bound to a Var, and checks that all types match. In case of a mismatch, types of compile- time immediates can be adjusted, otherwise an error is reported.


junrushao:
Ah I just merged a PR [#16074](https://github.com/apache/tvm/pull/16074), but not sure if it's related.

PR #15910, opened 2023-10-10
[Unity] Paged KV Cache as LM Support

MasterJH5574:
This PR introduces the PagedKVCache object to `lm_support.cc` for the KV cache value management in batching settings.

One test file is included. Note that this file does not contain the test of attention function/kernel. That part will be uploaded and tested separately.

### DIFF
--- a/src/runtime/relax_vm/paged_kv_cache.cc
+++ b/src/runtime/relax_vm/paged_kv_cache.cc
@@ -1,0 +1,661 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+/*!
+ * \file src/runtime/relax_vm/paged_kv_cache.cc
+ * \brief Runtime paged KV cache object for language models.
+ */
+#include <tvm/runtime/device_api.h>
+#include <tvm/runtime/logging.h>
+#include <tvm/runtime/ndarray.h>
+#include <tvm/runtime/registry.h>
+
+namespace tvm {
+namespace runtime {
+namespace relax_vm {
+
+//-------------------------------------------
+// We keep the implementation private as
+// they may subject to future changes.
+//
+// Users can interact with it through the
+// runtime API function calls
+//-------------------------------------------
+
+/*!
+ * \brief The paged KV cache for attention.
+ * - It supports managing the K/V data of **multiple sequences**.
+ * - It manages K/V values by doing paging along the sequence-length
+ * dimension with a configured page size.
+ * - To add a sequence to the cache, use AddSequence.
+ * - The basic example use of the paged KV cache after initialization
+ * in each round of model forwarding is the following:
+ *   - step 1. use `ResetAppendLengths` to reset the appending information
+ *     for preparation,
+ *   - step 2. use `ReserveExtraLengthForAppend` to specify the length
+ *     of K/V data to be appended for each sequence,
+ *   - step 3. use `SyncAuxArrayToDevice` to synchronize auxiliary arrays
+ *     to device for append/attention computation,
+ *   - step 4. for each layer, use `Append` to append the K/V data to the
+ *     cache, and then use `Attention` to compute attention results with
+ *     Q data.
+ */
+class PagedAttentionKVCacheObj : public Object {
+ private:
+  /*! \brief The total number of sequences managed in the KV cache. */
+  int64_t num_total_seqs_ = 0;
+  /*! \brief The number of pages that are in use by the sequences. */
+  int64_t num_pages_in_use_ = 0;
+  /*!
+   * \brief The number of allocated pages, including the in-use pages
+   * and the pages released due to sequence removal.
+   */
+  int64_t num_pages_allocated_ = 0;
+
+  /********************* Configuration *********************/
+
+  /*! \brief The page size (the sequence length each page manages) of the cache. */
+  const int64_t page_size_;
+  /*! \brief The number of layers in the model. */
+  const int64_t num_layers_;
+  /*! \brief The number of heads in the model. */
+  const int64_t num_heads_;
+  /*! \brief The number of features each head has. */
+  const int64_t head_dim_;
+
+  /*! \brief We fix int32 to be the index dtype of auxiliary data. */
+  const DLDataType dtype_aux_ = DLDataType(DataType::Int(32, 1));
+
+  /********************* Page Structures *********************/
+
+  /*!
+   * \brief The KV data managed by the KV cache.
+   * It has layout (num_pages, num_layers, 2, num_heads, page_size, head_dim).
+   * Along on the "2" dimension, index 0 stands for K and 1 stands for V.
+   */
+  NDArray pages_;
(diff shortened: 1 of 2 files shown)


src/runtime/relax_vm/lm_support.cc
```diff
@@ -472,6 +472,430 @@ void ApplySoftmaxWithTemperature(NDArray logits, double temperature) {
...
+  const DLDataType dtype_aux = DataType::Int(32, 1).operator DLDataType();
+
+  /********************* Page Structures *********************/
+
+  const int64_t page_chunk_size = 1;
+
+  NDArray pages;
+  std::vector<int32_t> free_page_ids;
+
+  std::vector<std::vector<int32_t>> page_table;
```
tqchen, line 497: add comment, what does each element in page table mean

MasterJH5574:
Mark as draft for an update of docstring and APIs. Sorry that it is not mature enough.

MasterJH5574:
Documented the paged KV cache and updated some of the interfaces. It's ready for review now.

src/runtime/relax_vm/paged_kv_cache.cc
```diff
@@ -0,0 +1,633 @@
...
+  const int64_t page_size_;
+  /*! \brief The number of layers in the model. */
+  const int64_t num_layers_;
+  /*! \brief The number of heads in the model. */
+  const int64_t num_heads_;
+  /*! \brief The number of features each head has. */
+  const int64_t head_dim_;
+
+  /*! \brief We fix int32 to be the index dtype of auxiliary data. */
+  const DLDataType dtype_aux_ = DataType::Int(32, 1).operator DLDataType();
```
tqchen, line 81: We can store as DataType and auto conversts to DLDataType later

src/runtime/relax_vm/paged_kv_cache.cc
```diff
@@ -0,0 +1,633 @@
...
+    cur_pos2seqid_device_ = NDArray::Empty({reserved_num_pages * page_size}, dtype_aux_, device);
+  }
+
+  /*!
+   * \brief Given a sequence id and a required extra length, allocate new
+   * new pages for the sequence until the total capacity can cover the
+   * current sequence length plus the required extra length.
+   * \note By reserving extra length for the sequence, the subsequent appended
+   * K/V data for the sequence must have the same length as the input length here.
+   * \param seq_id The id of the sequence to process.
```
tqchen, line 169: to allocate a new sequence, call ReserveExtraLengthForAppend with seq_id equals num_total_sequences

src/runtime/relax_vm/paged_kv_cache.cc
```diff
@@ -0,0 +1,633 @@
...
+   * \param extra_length The extra length to reserve for the sequence.
+   */
+  void ReserveExtraLengthForAppend(int64_t seq_id, int64_t extra_length) {
+    CHECK_GE(seq_id, 0) << "Input sequence id should be positive";
+    CHECK_LE(seq_id, num_total_seqs_)
+        << "Invalid input sequence id " << seq_id
+        << ", which is larger than the current total number of sequences.";
+    CHECK_GT(extra_length, 0) << "The input length should be positive.";
+
+    if (seq_id == num_total_seqs_) {
```
tqchen, line 179: I think it is useful to have a separate function to grow the sequence to keep the concern of concurrent sequence growing and seq len growing separately

src/runtime/relax_vm/paged_kv_cache.cc
```diff
@@ -0,0 +1,633 @@
...
+      ++num_total_seqs_;
+    }
+
+    // The reservation is based on the current sequence length.
+    // If "current sequence + input extra length" does not exceed the
+    // current capacity (number of pages * page size), no action is taken.
+    int64_t cur_npage = page_table_[seq_id].size();
+    int64_t tgt_npage = (seq_lengths_[seq_id] + extra_length + page_size_ - 1) / page_size_;
+    for (int64_t page_idx = cur_npage; page_idx < tgt_npage; ++page_idx) {
+      AllocatePageForSequence(seq_id);
```
tqchen, line 193: nit, add a comment stating that this call mutates the page table

src/runtime/relax_vm/paged_kv_cache.cc
```diff
@@ -0,0 +1,633 @@
...
+   * \param f_attention The input attention compute function.
+   * \param q_data The query data. We support the following two layout settings:
+   * - in batch decode settings, q_data has layout (num_total_seqs, 1, num_heads, head_dim)
+   *   where num_total_seqs should exactly equal to the number of sequences in the cache.
+   * - in other settings (single-sequence prefill, batch prefill, or speculation
+   *   verification), q_data has the **flattened layout** to handle raggedness:
+   *   (1, total query length, num_heads, head_dim).
+   * \param layer_id The model layer index of the current attention.
+   * \param output The attention output array.
+   * \param apply_rotary A boolean flag indicating if to apply RoPE.
```
tqchen, line 213: Apply RoPE to the kv? or qkv, would be good to state the purpose

tqchen: It might be useful to add a TODO and indicate that we could move the rope related parameters to the f_attention closure in the future.

src/runtime/relax_vm/paged_kv_cache.cc
```diff
@@ -0,0 +1,633 @@
...
+    }
+    CHECK_EQ(ntoken, q_data->shape[0] * q_data->shape[1]);
+
+    // The auxiliary data structure on device must have been synchronized.
+    CHECK(!dirty_aux_data_device_)
+        << "The auxiliary arrays are not synchronized to device. Please call "
+           "`SyncAuxArrayToDevice` to synchronize before calling `Attention`.";
+
+    // Provide a 8MB size float temp buffer for the attention kernel.
+    static NDArray tmp_buffer =
```
tqchen, line 246: this can be dangerous in multi-threaded setting, can we make this a member of the class?

src/runtime/relax_vm/paged_kv_cache.cc
```diff
@@ -0,0 +1,633 @@
...
+        NDArray::Empty({8 * 1024 * 1024}, DLDataType(DataType::Float(32)), pages_->device);
+
+    f_attention(q_data, pages_,                                                                  //
+                page_table_indptr_device_.CreateView({num_total_seqs_ + 1}, dtype_aux_),         //
```
tqchen, line 250: would be great these views are also created once during sync device and kept there without having to recreate on the fly

src/runtime/relax_vm/paged_kv_cache.cc
```diff
@@ -0,0 +1,633 @@
...
+
+  /*!
+   * \brief Returning the cached K/V values in the form of "an array of NDArray".
+   * Each returned NDArray has layout (num_layers, 2, seqlen, num_heads, head_dim),
+   * where along on the "2" dimension, index 0 stands for K and 1 stands for V.
+   * \param f_view The function used for copying data out from cache.
+   * \return The cached K/V values, one NDArray per sequence.
+   * \note This method is majorly for testing purpose.
+   */
+  Array<NDArray> View(PackedFunc f_view) {
```
tqchen, line 356: This is no longer a view since it copies data out. Let us do DebugGetKV

tqchen (approved):

tests/python/relax/test_runtime_builtin_paged_attention_kv_cache.py
```diff
@@ -0,0 +1,420 @@
...
+    verify_cached_values(cache, [], f_copy_cache)
+
+
+if __name__ == "__main__":
+    test_paged_attention_kv_cache_append_prefill()
+    test_paged_attention_kv_cache_append_decode()
+    test_paged_attention_kv_cache_remove()
+    test_paged_attention_kv_cache_popn()
+    test_paged_attention_kv_cache_clear()
+    # tvm.testing.main()
```
yongwww, line 420: how about uncomment this test.

BTW, looks the test case for "vm.builtin.paged_attention_kv_cache_attention" is missing

MasterJH5574: We have the attention part tested with manual CUDA code developed by @yzh119 (great thanks!). We don't have TIR attention functions yet so it cannot be tested in a tidy way right now in this file. As a result, the attention tests are skipped for now. I just added a comment on that. We will follow up with the tests once we have the attention written in TIR.

PR #15912, opened 2023-10-10
[VM] modify the date pointer of tensor to point to the buffer

yongwww:
This change is from the unity, the kv cache will need it.


tqchen:
Would be great to cross check kv to see if we can work with the offset parameter in DLTensor. The main reason is that the data ptr may not points to an addressable location (it is not an issue for CUDA but can be an issue say for metal).

yongwww:
Thanks for the review!  I am going to close this pr due to the concern.

PR #15914, opened 2023-10-11
[TFLite][Frontend] Support quantized SQUARE

tlopex:
Support SQUARE quantization operation as part of #15148


leandron (changes requested):
Thanks for the PR. @tlopex can you add some testing to make sure this feature don't regress in future? Thanks.

lhutton1 (approved):
LGTM!

leandron (approved):
Thanks!

PR #15915, opened 2023-10-11
[TFLite][Frontend] Support quantized Reverse sequence

tlopex:
Support Reverse sequence quantization operation as part of #15148


leandron:
Hi @tlopex this is failing on a linting error, can you fix that so perhaps we can progress this PR?

PR #15916, opened 2023-10-11
[Unity][Transform] Handle relax.Var as call_tir args when lowering

Lunderberg:
Prior to this commit, several transforms assumed that the arguments passed to a `call_tir` builtin were provided as in-line `relax::Tuple` objects.  Because it would be equally valid for the arguments to instead be a `relax::Var` instance that had previously been bound to a `relax::Tuple` object, or had been passed as an input parameter with `relax::TupleStructInfo`, this assumption shouldn't be made.  This PR updates the `CallTIRRewrite`, `FoldConstant`, `FuseOps`, and `RewriteDataflowReshape` passes to handle variables providing the arguments.

### DIFF
--- a/include/tvm/relax/expr_functor.h
+++ b/include/tvm/relax/expr_functor.h
@@ -278,6 +278,20 @@ class ExprVisitor : public ExprFunctor<void(const Expr&)> {
   virtual void VisitSpan(const Span& span);
   virtual void VisitPrimExpr(const PrimExpr& expr);
 
+  /*!
+   * \brief Look up the value bound to a variable.
+   * \param var The var to be looked up.
+   * \return The value bound to the input \p var.
+   * \note For function parameters, this function returns NullOpt.
+   */
+  inline Optional<Expr> LookupBinding(const Var& var) {
+    if (auto it = binding_table_.find(var->vid); it != binding_table_.end()) {
+      return it->second;
+    } else {
+      return NullOpt;
+    }
+  }
+
  private:
   using TSelf = ExprVisitor;
   using VisitBindingVTable =
@@ -308,6 +322,14 @@ class ExprVisitor : public ExprFunctor<void(const Expr&)> {
   // This visitor is not visible to child classes and only
   // used to supported default visiting behavior.
   DefaultStructInfoFieldVisitor default_struct_info_field_visitor_{this};
+
+  /*! \brief A binding table that maps var to value.
+   *
+   * Unlike ExprMutator, which can rely on the binding table of the
+   * internal BlockBuilder, ExprVisitor needs to track the bindings on
+   * its own.
+   */
+  std::unordered_map<Id, Expr, ObjectPtrHash, ObjectPtrEqual> binding_table_;
 };
 
 void PostOrderVisit(const Expr& node, std::function<void(const Expr&)> fvisit);
--- a/include/tvm/relax/expr_functor.h
+++ b/include/tvm/relax/expr_functor.h
@@ -292,6 +292,23 @@ class ExprVisitor : public ExprFunctor<void(const Expr&)> {
     }
   }
 
+  /*!
+   * \brief Unwrap any known binding
+   * \param expr The expression to be unwrapped.
+   * \return The expression after following any known Var bindings
+   */
+  inline Expr UnwrapBindings(Expr expr) {
+    while (true) {
+      auto as_var = expr.as<Var>();
+      if (!as_var) return expr;
+
+      auto bound_expr = LookupBinding(as_var.value());
+      if (!bound_expr) return expr;
+
+      expr = bound_expr.value();
+    }
+  }
+
  private:
   using TSelf = ExprVisitor;
   using VisitBindingVTable =
@@ -533,6 +550,23 @@ class ExprMutator : public ExprMutatorBase {
    * \note For function parameters, this function returns NullOpt.
    */
   Optional<Expr> LookupBinding(const Var& var);
+
+  /*!
+   * \brief Unwrap any known binding
+   * \param expr The expression to be unwrapped.
+   * \return The expression after following any known Var bindings
+   */
+  inline Expr UnwrapBindings(Expr expr) {
+    while (true) {
+      auto as_var = expr.as<Var>();
+      if (!as_var) return expr;
+
+      auto bound_expr = LookupBinding(as_var.value());
+      if (!bound_expr) return expr;
+
+      expr = bound_expr.value();
+    }
+  }
 
   /*!
    * \brief Post-order rewrite a node and normalize.
--- a/python/tvm/relax/op/base.py
+++ b/python/tvm/relax/op/base.py
@@ -66,6 +66,18 @@ def null_value() -> Call:
     return _ffi_api.null_value()  # type: ignore
 
 
+def _normalize_arg_tuple(args: Expr) -> Expr:
+    if isinstance(args, RxTuple) or isinstance(args.struct_info_, TupleStructInfo):
+        # A tuple, or a Var bound to a tuple, are kept as-is
+        return args
+    elif isinstance(args, Expr):
+        # A single argument is wrapped into a tuple
+        return RxTuple((args,))
+    else:
+        # Anything else is left for the FFI to handle
+        return args
+
+
 @args_converter.auto
 def call_tir(
     gvar: GlobalVar,
@@ -97,12 +109,7 @@ def call_tir(
     ret: Call
         A call node for the call_tir operator.
     """
-    if (
-        isinstance(args, Expr)
-        and not isinstance(args, RxTuple)
-        and not isinstance(args.struct_info_, TupleStructInfo)
-    ):
-        args = RxTuple((args,))
+    args = _normalize_arg_tuple(args)
 
     if not isinstance(out_sinfo, list):
         out_sinfo = [out_sinfo]
@@ -156,8 +163,7 @@ def call_tir_with_grad(
     ret: Call
         A call node for the call_tir_with_grad operator.
     """
-    if isinstance(args, Expr) and not isinstance(args, RxTuple):  # type: ignore
-        args = RxTuple((args,))
+    args = _normalize_arg_tuple(args)
 
     if not isinstance(out_sinfo, list):
         out_sinfo = [out_sinfo]
@@ -224,8 +230,7 @@ def call_tir_inplace(
     ret: Call
         A call node for the call_tir operator.
     """
-    if isinstance(args, Expr) and not isinstance(args, RxTuple):  # type: ignore
-        args = RxTuple((args,))
+    args = _normalize_arg_tuple(args)
 
     if not isinstance(inplace_indices, list):
         inplace_indices = [inplace_indices]
@@ -279,8 +284,7 @@ def call_dps_packed(
     if isinstance(func, str):
         func = ExternFunc(func)
 
-    if isinstance(args, Expr) and not isinstance(args, RxTuple):  # type: ignore
-        args = RxTuple((args,))
+    args = _normalize_arg_tuple(args)
 
     if not isinstance(out_sinfo, list):
         out_sinfo = [out_sinfo]
--- a/python/tvm/relax/op/base.py
+++ b/python/tvm/relax/op/base.py
@@ -97,7 +97,11 @@ def call_tir(
     ret: Call
         A call node for the call_tir operator.
     """
-    if isinstance(args, Expr) and not isinstance(args, RxTuple) and not isinstance(args.struct_info_, TupleStructInfo):  # type: ignore
+    if (
+        isinstance(args, Expr)
+        and not isinstance(args, RxTuple)
+        and not isinstance(args.struct_info_, TupleStructInfo)
+    ):
         args = RxTuple((args,))
 
     if not isinstance(out_sinfo, list):
--- a/python/tvm/relax/op/base.py
+++ b/python/tvm/relax/op/base.py
@@ -25,7 +25,7 @@ from tvm.runtime.object import Object
 from . import _ffi_api
 from ..expr import Expr, StringImm, ShapeExpr, Call, ExternFunc, GlobalVar, Var
 from ..expr import Tuple as RxTuple
-from ..struct_info import StructInfo, TensorStructInfo
+from ..struct_info import StructInfo, TensorStructInfo, TupleStructInfo
 from ...ir import PrimExpr
 from ..utils import args_converter
 
@@ -97,7 +97,7 @@ def call_tir(
     ret: Call
         A call node for the call_tir operator.
     """
-    if isinstance(args, Expr) and not isinstance(args, RxTuple):  # type: ignore
+    if isinstance(args, Expr) and not isinstance(args, RxTuple) and not isinstance(args.struct_info_, TupleStructInfo):  # type: ignore
         args = RxTuple((args,))
 
     if not isinstance(out_sinfo, list):
--- a/src/relax/transform/call_tir_rewrite.cc
+++ b/src/relax/transform/call_tir_rewrite.cc
@@ -77,14 +77,14 @@ class CallTIRMutator : public ExprMutator {
           ICHECK(inplace_attrs->inplace_indices[0].IntValue() != -1)
               << "If calling call_tir_inplace and there is one output, its in-place index must not"
                  " be -1.";
-          outs.push_back(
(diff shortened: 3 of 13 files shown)


Lunderberg:
Rebased onto `unity` head to ensure the CI tests haven't broken in the meantime.

slyubomirsky:
I think you may need to update the StructInfo inference for `call_tir_inplace` like in #15971, since (without modification) that assumes the argument is a tuple literal. The test cases here don't try that case, hence why that doesn't result in an error (I'd recommend adding a test case).

src/relax/transform/call_tir_rewrite.cc
```diff
@@ -111,41 +111,69 @@ class CallTIRMutator : public ExprMutator {
...
+          return args;
+        } else {
+          LOG(FATAL) << "Lowering of " << call
+                     << " requires knowing how many arguments are passed to the function.  "
+                     << "However, the tuple of arguments " << arg_tuple
+                     << " is not itself a tuple, "
+                     << "nor does its struct info " << GetStructInfo(arg_tuple)
+                     << " define the number of arguments.";
         }
+      }();
```
slyubomirsky, line 149: Not that I particularly mind it, but is it preferable to use this construction with a lambda as opposed to just assigning `args` in different branches?
```c++
Array<Expr> args;
if (case1) {
    args = ...
} else if (case2) {
    args = ...
}
// etc.
```
Does it avoid an allocation or something?

python/tvm/relax/op/base.py
```diff
@@ -97,7 +97,11 @@ def call_tir(
     ret: Call
         A call node for the call_tir operator.
     """
-    if isinstance(args, Expr) and not isinstance(args, RxTuple):  # type: ignore
+    if (
+        isinstance(args, Expr)
+        and not isinstance(args, RxTuple)
+        and not isinstance(args.struct_info_, TupleStructInfo)
```
slyubomirsky, line 103: You should do this for the other call_tir variants in this file too.

src/relax/transform/call_tir_rewrite.cc
Lunderberg: I tend to use this construction for a couple of reasons.

* Avoid ever having a partially-initialized variable.
* Limit the scope of temporary variables that should only be used in the initialization.
* Simplify nested if/else cases with early returns.

Effectively, the immediately-invoked lambda expression acts as a block scope with a return type, similar to Rust's braces (e.g. The value of `{let i = 5; i+1}` is `6`) or relax's SeqExpr.

include/tvm/relax/expr_functor.h
```diff
@@ -278,6 +278,37 @@ class ExprVisitor : public ExprFunctor<void(const Expr&)> {
   virtual void VisitSpan(const Span& span);
   virtual void VisitPrimExpr(const PrimExpr& expr);
 
+  /*!
+   * \brief Look up the value bound to a variable.
+   * \param var The var to be looked up.
+   * \return The value bound to the input \p var.
+   * \note For function parameters, this function returns NullOpt.
+   */
+  inline Optional<Expr> LookupBinding(const Var& var) {
```
tqchen, line 287: I think we should instead enforce call_tir to be a more restricted form

tqchen (changes requested):

tqchen:
Thanks for the PR, I know this is indeed a generalization and there are some tradeoffs to be considered here. Specifically, we should consider the following alternative:

- C0: We enforce `call_tir` to only take explicit Tuple and enforce in well-form check

This indeed limit what we can do in terms of language, but would greatly simplify the logic that leverages CallTIR. And such simplicity helps us in a lot of cases since passes are simpler and more passes that depends on `call_tir` pattern matching would enjoy that simplicity.

slyubomirsky:
See also [this issue](https://discuss.tvm.apache.org/t/validity-of-common-subexpression-elimination-pass-that-simplifies-call-tir-args/15832). Allowing `call_tir` to be more general would allow us to avoid special handling for `call_tir` (and its variants) in that case. It's a question of which we'd prefer.

python/tvm/relax/op/base.py
Lunderberg: Good call, and updated.

Lunderberg:
This does increase the complexity of passes that interact with `call_tir`, but it also decreases the complexity of all other passes.  If we require an in-line tuple for `call_tir`, then every other pass must be aware of this restriction and preserve it.  Any `ExprMutator` class that implements `Expr VisitExpr_(const TupleNode*) override` must also check the context of the `Tuple` in order to know what mutations are allowed, and similar checks would be required in `VisitBinding`.  This is complexity that is inherent to interactions with a `call_tir` node, and a IR restriction not exposed through the C++ type system moves that complexity to passes that shouldn't be aware of `call_tir` at all.

I think there are ways we could ensure that all `call_tir` has immediate access to its arguments, working within the C++ type system instead of running a well-formed check

* Provide an explicit IR node for `CallTIR` and variants.  Instead of `R.call_tir(tir_gvar, [arg1, ..., argN])` expanding into `R.Call(tvm.ir.Op("relax.call_tir"), [tir_gvar, R.tuple(arg1, ..., argN)])`, it would expand into `R.CallTIR(tir_gvar, [arg1, ..., argN])`.

* Move the GlobalVar target into a member variable of the `call_tir` operator.  Instead of `R.call_tir(tir_gvar, [arg1, ..., argN])` expanding into `R.Call(tvm.ir.Op("relax.call_tir"), [tir_gvar, R.tuple(arg1, ..., argN)])`, it would expand into `R.Call(R.TIRFunc(tir_gvar), [arg1, ..., argN])`.  This would be analogous to how `R.ExternFunc` is currently handled.longer require a tuple, so it couldn't be indirect.

With either option, the arguments would be immediately accessible by passes that interact with `call_tir`, but this would be enforced at the type system, so other passes wouldn't be required to explicitly check for it at runtime.

Lunderberg:
@slyubomirsky I pulled in the infer struct info improvements and the unit test updates from PR#15971, so this should now include all fixes from both sets of changes.

slyubomirsky (approved):
My concerns were addressed, but we should have agreement on the design question @tqchen posed before we merge (if we will merge).

slyubomirsky:
Making a new IR node for `call_tir` would entail a lot of revisions, but it might be worth it since it is very important to the language. If that's something we would consider, we should give thought to other changes to `Call` node potentially (several have come up before), since making changes to the AST would require careful thought.

Since we have the StructInfo system as a type system, I don't see a reason why `call_tir` needs to be a special case--a tuple is a tuple and that's the beauty of the type system.

tqchen:
The principle here is we make common cases(and their optimizations easy), while placing burdens on less common cases. 

As for the pass writing patterns, most of our current relax passes starts with mutating leaf-node e.g. Call where we can explicit unpack the patterns. They are structured as follows:

- Override Call, pattern match on the op, which allows us to enable structural intrinsics.
- Directly unpack the args since most op likely will entail the structure there, return a new value.

Mutating TupleExpr directly have the issue of needing to consider possible recursions, given our normal form allows nested tuple, exactly for the ease of representing structural information in some special intrinsics. Likely there is really a limited set of passes that involves this directly mutation, especially when they are tied to key optimization needs, and they can be restructured  to the above pattern.

Taking that rationale and given a lot of instrinsics also follow that pattern(e.g. CallPacked and other possible low level variant (e.g. an intrinsic that explicitly marks read and writes). Supporting intrinsics to allow special structures in the signature helps to simplify passes 

Given the CallTIR and other structured intrinsics are central part of a lot of our analysis. I think the tradeoff is worth spending.

slyubomirsky:
If we intend to have special cases like `call_tir` where one argument _must_ be a tuple literal (i.e., not following the normal rule of the type system that any member of the type should be permitted to stand in for any other), we should probably have some centralized place or an API where these rules are defined so that passes know to handle them. I'm not sure what form this should take, but such special cases make it trickier to implement passes like common subexpression elimination: the passes have to know about which operators are special and have special handling for them (as @Lunderberg noted).

I would argue that it doesn't bring that many advantages to require the argument to `call_tir` be a tuple literal, since the only other possibility (in normal form) is for the argument to a variable, and it is easy to look up the definitions of variables (if they're from the same function) or to use `TupleGetItem` if necessary (like the `GetTupleIndex` method in this PR).

Lunderberg:
> My concerns were addressed, but we should have agreement on the design question @tqchen posed before we merge (if we will merge).

Oh, absolutely.  The comment was to ensure that we had a clean implementation for the design comparison.

>  I'm not sure what form this should take, but such special cases make it trickier to implement passes like common subexpression elimination: the passes have to know about which operators are special and have special handling for them

Thank you for this phrasing, as I was trying to put this sentiment into words.  The existence of special requirements that must be preserved makes generic implementations of other passes be much more difficult.

tqchen:
What we are discussing is closely related to our original consideration of what forms a normal form and the general needs. Let me elaborate a bit more

The purpose of normal form is to restrict the set of possibilities in representing the same program thus helping reduce the assumption that one might take when writing pass. Of course it imposes extra demand for passes to ensure things go back to the normal form(the wellform-ness check). Currently, the relax normal form requires that we normalize all non-tuple nesting, but expand all tuple constructions.

That means we encourage forms like

```python
def func(x, y):
    lv0 = call_tir(mm, (x, y))
```
But do not encourage 
```python
def func(x, y):
    lv0 = (x, y)
    lv1 = call_tir(mm, lv0)
```

The reason we do that is to observe a general category of need: 
N0: Structural intrinsic argument
When constructing intrinsics, we would like intrinsics to present certain structures in them, for example `call_tir`, `call_packed`. If we want to analyze read write relations low-level, maybe we would like to have intrinsics of 
 lv0 = call_func_with_read_write(func, read_tuple, write_tuple)

Pattern match and rewriting of these structural intrinsic is a first-class need in our case because of following reasons:
R0:  they are common needs in machine learning passes, i can easily write a customized pass that matches and replaces the call_tir, such needs are also unbounded.
R1: We would like to make that customization happen at a third party, aka the developers who do not necessarily contribute to upstream can do it as simply as possible without worrying about unpacking tuples.

Such a normal form might bring a bit of restrictions, but won’t be too much to other possible passes that are “generic”. This also depends on the past writing patterns we do. Let us consider the following patterns
PT0: Only overload visit call and unpack that during visits
PT1: Possibly overload visit tuple

I would claim that PT0 is actually more desirable and handles almost all the pass needs, and it won’t cause any issues in the structural intrinsics, since all arguments are transformed and pattern matched together.

I think most of the possible ideas of difficulty mentioned are related to passes that follow PT1. Under the current normal form however, we naturally need to be very careful about PT1, because we do not want to lift common values in tuple construction and instead encourage tuples to be always unpacked, it is harder to think about. A CSE pass that tries to detect sub-tuples and lift common values via PT1 actually would violate the normal form requirement. 


A rule of thumb under the current normal form is that we almost always won’t do PT1, unless one passes that populate tuple annotation(aka deduction). If there is another pass that does so, we should visit the assumption and check carefully if the result can violate the normal form assumption. And if there are passes that do tuple replacement, we can still insert an extra helper pass that turns some of the indirect tuple reference into the unpacked form by checking structural info and unpacking them.

Finally, one should note that the passes might touch PT1 are usually developed in repo, as a result, the extra consideration of renormalization won’t be hurt, as our experience is sufficient to tackle these considerations. Burdens of R0/R1 are unbounded, a lot of our designs(e.g. Dataflow block) goes into simplifying them for developers who may not have a deep background. 

Of course all of the above roots back to our rationale of the normal form. One can argue we should take other normal forms(e.g. Tuple value should always be bound to a variable), in which case the consideration would be different and there are other tradeoffs. We considered these tradeoffs before arriving at the current one.

slyubomirsky:
Great work in making the full comparison list, Eric. Are we presently using inline tuples to simplify things significantly? I really don't think it's that big a deal to check the type of a variable and determine if you need to use TupleGetItem, etc.

Lunderberg:
> Are we presently using inline tuples to simplify things significantly?

I haven't found any locations that would require anything more than a binding lookup.  Of the 19 instances of `Downcast<Tuple>` that I found using grep, none of them 

* 8 cases which are removed by this PR
* 8 cases that use `Downcast<Tuple>` inside as part of an `IRModule` transform, and which could use the `UnwrapBindings` utility in this PR.
* 2 cases that occur in VM codegen, to handle `"relax.call_builtin_with_ctx"`.
* 1 case that occurs in a helper function, which is then used in `IRModule` transforms.

The most complicated case involving `relax::Tuple` that I've found is in the dataflow pattern matching, but that already handles the unwrapping of tuple nodes [here](https://github.com/apache/tvm/blob/unity/src/relax/ir/dataflow_matcher.cc#L367) with the `TryGetVarOfVal` line.

tqchen:
The simpler approach, is actually to just enable well-form check to detect the related issues and require tuple to be inlined. 

Normalizer serves as a way to create new bound variables when composite expression occur, which is the common case, but do not necessarily have to handle all the inlining cases.

Lunderberg:
> The simpler approach, is actually to just enable well-form check to detect the related issues and require tuple to be inlined.

Oh, certainly agreed that the well-formed check is the simpler one to introduce in the short-term.  I think it adds complexity in the long-term, though.

* Not evident to readers of the code.  A well-formed check can only detect ill-formed IR once it has been generated.  Encoding it in the C++ types used in the IR informs a reader what structures are allowed.
  
* Not evident to pass writers.  A well-formed check can only detect ill-formed IR after a pass runs.  Encoding it in the C++ types of the IR provides feedback from the C++ compiler, because ill-formed structures would cause a compilation error.
  
* Not comprehensive.  A well-formed check can only detect whether the currently tested input to a pass produces ill-formed IR.  Encoding a requirement in the C++ types of the IR ensures that a pass cannot produce ill-formed IR.

> Where pass writer can simply do the items if needed once detected by wellform check.

This is the case I want to avoid, because this is too late of a warning.  It occurs after a pass has been designed and written, often only when more comprehensive inputs are used in CI.  Adding a new design constraint that late in development often requires re-design of the implementation in order to follow the new constraint, resulting in significant repetition of effort.

> Aka we would like to eliminate the extra binding lookup for both syntax reason, as well as reduction of metral overhead for having to think about optional binding lookup or polymorphism.

I agree with the reasoning of reducing the mental overhead, but think that comes from having as many constraints exposed to the C++ type system as possible.  (See [earlier comment](https://github.com/apache/tvm/pull/15916#issuecomment-1781157045) for ways this could be done.)  By making illegal states be impossible to represent, the mental overhead of avoiding illegal states is removed. If illegal states can be represented, we increase the mental overhead of avoiding those illegal states.

tqchen:
> as many constraints exposed to the C++ type system as possible

This is indeed also a tradeoff here in terms of IR design. We cannot encode all constraints, for example the ANF itself is not encoded directly in the c++ data structure, but still enforced through well-form checkers.

For this particular case, because we can possibly allow unbounded usecases to support **N0: Structural intrinsic argument**.  The possible usecases go beyond CallTIR, that also include call packed, possible nested structures etc. So enabling it in the normal form is a good tradeoff

> It occurs after a pass has been designed and written

A pass that follows the common pattern(rewrites Call and not rewrites Tuple) will not face such issues. A pass that needs to do tuple replacement already have other complications to consider. And if the pass writer is not willing to do so, having a followup util function(like ConvertSSA) will simply unpack the tuple values and recovert things back to the normal form, without need to redesign the original pass. 

Indeed, this extra burden is put on developers that develops pass that involves tuple remapping, such pass in nature would require extra set of considerations. And the extra overhead of insertin a re-normalization is not high. They are less common than the passes that involves pattern match and rewrite structural intrinsic argument.  That is also why the design prioritizes these cases.

Lunderberg:
@slyubomirsky @tqchen Can you take a look at PRs https://github.com/apache/tvm/pull/16067 and https://github.com/apache/tvm/pull/16068?  The former introduces new functionality through a `FNormalize` attribute, and the latter applies that functionality to `relax.op.call_tir`.

1. Downstream passes can make assumptions about the specific form of the AST.  For `call_tir`, the current use of `Downcast<Tuple>` be guaranteed to have an in-line tuple.
2. If the operator can be automatically normalized, then upstream passes do not need to be aware of the operator-specific requirements.  The normalization is applied as part of every ExprMutator usage, the same as any other normalization step.
3. If automatic normalization is impossible, an error is raised during the pass that introduced the ill-formed IR, rather than at the downstream pass that tried to use it.

Properties (2) and (3) fulfill my goals of avoiding an increased mental burden when writing an upstream pass, and of making all IR requirements be explicit within the code.

Lunderberg:
Closing this PR, as it is superseded by #16067 and #16068.

PR #15919, opened 2023-10-12
[Unity] Fix bug in onnx_frontend when multi inputs have same symbolic shape

HongHongHongL:
When multi inputs have same symbolic shape, the function **from_onnx** will cause  

> AssertionError:` For Expr having ShapeStructInfo, emit_te now only supports ShapeExpr

This PR uses a dictionary mapping from the name of ValueInfoProto to SizeVar when parse_graph_input. ValueInfoProto with same name will map to same SizeVar.

### DIFF
--- a/python/tvm/relax/frontend/onnx/onnx_frontend.py
+++ b/python/tvm/relax/frontend/onnx/onnx_frontend.py
@@ -114,7 +114,8 @@ def get_info(info_proto: onnx.onnx_ml_pb2.ValueInfoProto, value_dict={}) -> Tupl
     Returns
     -------
     Tuple[str, List, str, List]
-        The name, shape, type, and shape name of the ValueInfoProto.
+        The name, shape, type, and shape name of the ValueInfoProto, and the
+        value_dict.
     """
     shape = []
     shape_name = []
@@ -135,7 +136,7 @@ def get_info(info_proto: onnx.onnx_ml_pb2.ValueInfoProto, value_dict={}) -> Tupl
         dtype = get_type(info_proto.type.tensor_type.elem_type)
     else:
         dtype = None
-    return name, shape, dtype, shape_name
+    return name, shape, dtype, shape_name, value_dict
 
 
 def get_numpy(tensor_proto: onnx.onnx_ml_pb2.TensorProto) -> _np.ndarray:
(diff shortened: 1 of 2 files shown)


python/tvm/relax/frontend/onnx/onnx_frontend.py
```diff
@@ -100,14 +100,17 @@ def get_constant(
         return var
 
 
-def get_info(info_proto: onnx.onnx_ml_pb2.ValueInfoProto) -> Tuple[str, List, str, List]:
+def get_info(info_proto: onnx.onnx_ml_pb2.ValueInfoProto, value_dict={}) -> Tuple[str, List, str, List]:
```
Hzfengsy, line 103: if `value_dict` is the output of the function, please move it to the return values, instead of reference param

Hzfengsy (changes requested):

python/tvm/relax/frontend/onnx/onnx_frontend.py
HongHongHongL: `value_dict` should be the input of the function. When a symbol appears for the second time, the value in the dict should be used.

python/tvm/relax/frontend/onnx/onnx_frontend.py
```diff
@@ -100,26 +100,35 @@ def get_constant(
...
     ----------
     info_proto: onnx.onnx_ml_pb2.ValueInfoProto
         The ValueInfoProto to extract the info from.
 
+    value_dict: Dict
+        The Dictionary mapping from the name of ValueInfoProto to SizeVar
+
     Returns
     -------
     Tuple[str, List, str, List]
```
Hzfengsy, line 119:
```suggestion
    Tuple[str, List, str, List, Dict]
```

python/tvm/relax/frontend/onnx/onnx_frontend.py
```diff
@@ -100,26 +100,35 @@ def get_constant(
...
+    value_dict: Dict[str, tvm.tir.SizeVar]
+) -> Tuple[str, List, str, List, Dict]:
     """Extract the shape from a ValueInfoProto.
 
     Parameters
     ----------
     info_proto: onnx.onnx_ml_pb2.ValueInfoProto
         The ValueInfoProto to extract the info from.
 
+    value_dict: Dict
```
Hzfengsy, line 114:
```suggestion
    value_dict: Dict[str, tvm.tir.SizeVar]
```

Aleksei-grovety (commented):
Could you add a test to check these changes?

Hzfengsy (approved):

Aleksei-grovety (approved):

Issue #15729, opened 2023-09-12
[Bug][Unity] TVMError when using relax to load model with Trilu operator


Thrsu:
The error message suggests that there is an issue with converting the second argument for the trill operation. The error message is as follows: `TVMError: In function relax.op.tril(0: RelayExpr, 1: int) -> RelayExpr: error while converting argument 1: [06:33:20] /workplace/software/tvm/tvm/include/tvm/runtime/packed_func.h:562: InternalError: Check failed: type_code_ == kDLInt (8 vs. 0) : expected int but got Object`

It is worth noting that when running the model using the relay module, no error occurs. The error only occurs when attempting to load the model using the relax module.

The model structure is as below:
<img width="611" alt="Trilu" src="https://github.com/apache/tvm/assets/89128704/04012b47-ab0e-40bd-ba7e-326eab1afa7c">

### Actual behavior
```
Traceback (most recent call last):
  File "test.py", line 8, in <module>
    tvm_model = from_onnx(model, opset=18, keep_params_in_input=True)
  File "/workplace/software/tvm/tvm/python/tvm/relax/frontend/onnx/onnx_frontend.py", line 2287, in from_onnx
    return g.from_onnx(graph, opset)
  File "/workplace/software/tvm/tvm/python/tvm/relax/frontend/onnx/onnx_frontend.py", line 1932, in from_onnx
    self._construct_nodes(graph)
  File "/workplace/software/tvm/tvm/python/tvm/relax/frontend/onnx/onnx_frontend.py", line 2095, in _construct_nodes
    op = self._convert_operator(op_name, inputs, attr, self.opset)
  File "/workplace/software/tvm/tvm/python/tvm/relax/frontend/onnx/onnx_frontend.py", line 2191, in _convert_operator
    sym = op_function(self.bb, inputs, attrs, [self._nodes, self._params])
  File "/workplace/software/tvm/tvm/python/tvm/relax/frontend/onnx/onnx_frontend.py", line 621, in _impl_v14
    return relax.op.tril(x, k)
  File "/workplace/software/tvm/tvm/python/tvm/relax/op/create.py", line 238, in tril
    return _ffi_api.tril(x, k)  # type: ignore
  File "/workplace/software/tvm/tvm/python/tvm/_ffi/_ctypes/packed_func.py", line 238, in __call__
    raise get_last_ffi_error()
tvm._ffi.base.TVMError: Traceback (most recent call last):
  4: TVMFuncCall
[...]

PR #15924, opened 2023-10-13
[Fix][Unity] Fix TVMError when using relax to load model with Trilu operator

dmilosevic252:
Sets the default value 0 for the k variable, if the value has not been previously set.
This is a code patch for fixing this issue: https://github.com/apache/tvm/issues/15729

### DIFF
--- a/python/tvm/relax/frontend/onnx/onnx_frontend.py
+++ b/python/tvm/relax/frontend/onnx/onnx_frontend.py
@@ -637,6 +637,11 @@ class Trilu(OnnxOpConverter):
         x = inputs[0]
         k = inputs[1] if len(inputs) > 1 else 0
 
+        if isinstance(k, relax.Var) and k.name_hint in params:
+            k = get_constant(k,params)
+        else:
+            k = 0
+
         if upper:
             return relax.op.triu(x, k)
         else:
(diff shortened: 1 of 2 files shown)


python/tvm/relax/frontend/onnx/onnx_frontend.py
```diff
@@ -637,6 +637,11 @@ def _impl_v14(cls, bb, inputs, attr, params):
         x = inputs[0]
         k = inputs[1] if len(inputs) > 1 else 0
 
+        if isinstance(k, relax.Var) and k.name_hint in params:
+            k = get_constant(k, params)
+        else:
+            k = 0
```
Hzfengsy, line 643: Why do we need to set `k` to 0? It should already be done in line 638

Hzfengsy: Also, what will happen if `k` is a constant instead of an input? Please also add a test case for it.

dmilosevic252: > Why do we need to set `k` to 0? It should already be done in line 638

In line 638 'k' is set to 0 if its value hasn't been passed. In line 643 we set it to 0 if 'k' is passed but it's value (constant) cannot be obtained.

> Also, what will happen if `k` is a constant instead of an input? Please also add a test case for it.

Thank you for the feedback! I will add a test case.

Hzfengsy (approved):

PR #15928, opened 2023-10-13
[Relay] Expose qnn ops directly from relay.qnn module

quic-sanirudh:
When we access `nn` ops in relay, we can directly access them as `relay.nn.<op>` where `<op>` can be any nn op like `conv2d`, `avg_pool2d`, etc.

When we access `qnn` ops, we need to access them as `relay.qnn.op.<op>` as they're not exposed to the `relay.qnn` module. This change tries to add that.

### DIFF
--- a/python/tvm/relay/qnn/__init__.py
+++ b/python/tvm/relay/qnn/__init__.py
@@ -14,8 +14,9 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-# pylint: disable=wildcard-import
+# pylint: disable=wildcard-import,redefined-builtin
 """QNN dialect operators and IR passes."""
 from __future__ import absolute_import as _abs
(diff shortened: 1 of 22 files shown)


quic-sanirudh:
cc @ibsidorenko

ibsidorenko (approved):

ibsidorenko:
LGTM. Do you plan to change access names in unit tests?

quic-sanirudh:
> LGTM. Do you plan to change access names in unit tests?

Thanks for the review, and yes, I'll update the tests and push it today. Initially I thought it might introduce a lot of changes that hurts git history un-necessarily, but now I think there could be some advantage to doing it (especially since tests are seen as an initial source of example usage of an API).

kparzysz-quic (approved):
LGTM.  Thanks @quic-sanirudh!

PR #15929, opened 2023-10-13
[microNPU][ETHOSU] Fix rounding mode in requantize operation

Aleksei-grovety:
Before the fix, TFL rounding was used in the requantize operation, and this led to a discrepancy in the results in some cases.


lhutton1 (approved):
LGTM, thanks @Aleksei-grovety!

PR #15932, opened 2023-10-14
[Unity] fixup resize2d dynamic input support

starrkk:
when roi is numpy.array([], )，  if roi is not None statement undetectable.


Hzfengsy (commented):
Please add a regression test

starrkk:
@Hzfengsy  Because roi parameters are generated by the op of relax and are not defined directly via onnx, it is not easy to write regression tests

Hzfengsy:
For a concrete ONNX frontend issue, there must be an ONNX graph that hits the bug. Simplify the graph and that would be a regression test. 

If it's generated by relax, which means it's might be the output of another Node, then you can create an ONNX subgraph with more than one node in the testcase

python/tvm/relax/frontend/onnx/onnx_frontend.py
```diff
@@ -1290,7 +1290,7 @@ def _impl_v18(cls, bb, inputs, attr, params):
         ), "Only one of scales and sizes can be provided in Resize."
 
         # Define relax implementation.
-        if roi is not None:
+        if roi is not None and not _np.array_equal(roi.data.numpy(),  _np.array([], dtype=_np.float32)):
```
Hzfengsy, line 1293: It's not robust to use `equal` check, as the dtype can be different. Please check if it's a 0-ranked tensor

starrkk: Okay, I'll try

PR #15935, opened 2023-10-16
[Unity] [Bugfix] Fix TypeError in interpolate caused by scale_factor as tuple

Thrsu:
This pull request addresses a TypeError that occurs in the interpolate function when the scale_factor parameter is of type tuple. The current implementation fails to handle cases where scale_factor is a tuple, resulting in the error message: "TypeError: int() argument must be a string, a bytes-like object or a number, not 'list'."

This patch allows the interpolate function to handle both scalar and tuple values for the scale_factor parameter without raising a TypeError.

And this bug can be reproduced by the following script:
```python
import torch
from torch import fx
from torch.nn import Module
import tvm
from tvm import relax
from tvm.relax.frontend.torch import from_fx

input_data = torch.randn([1, 2, 4, 4], dtype=torch.float32)
class interpolate(Module):
    def forward(self, input):
        return torch.nn.functional.interpolate(input, size=None, scale_factor=(2.0, 1.0), mode='bilinear', align_corners=False,)

model = interpolate().float()
input_data = [input_data]
input_info = list(zip([list(inp.shape) for inp in input_data], [str(inp.dtype) for inp in input_data]))
fx_model : torch.fx.GraphModule = fx.symbolic_trace(model)
with torch.no_grad():
    mod = from_fx(fx_model, input_info)
```

### DIFF
--- a/python/tvm/relax/frontend/torch/fx_translator.py
+++ b/python/tvm/relax/frontend/torch/fx_translator.py
@@ -1118,7 +1118,9 @@ class TorchFXImporter:
             assert isinstance(shape, relax.ShapeExpr)
             if isinstance(scale_factor, tuple):
                 assert len(scale_factor) == len(shape) - 2
-                size = tuple(int(shape[i].value * scale_factor[i - 2]) for i in range(2, len(shape)))
+                size = tuple(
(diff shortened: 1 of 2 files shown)


jikechao (approved):
LGTM, thanks for your PR. @Thrsu

Hzfengsy (approved):

PR #15938, opened 2023-10-17
[Unity][MSC][Bugfix] Trilu bugfix && special ops support

Archermmt:
The trilu test from test_translate_relay.py results in failures(e.g https://github.com/apache/tvm/pull/15783). This PR fix the bugs.
Besides, I use a customized ops map to convert torch.max(x, dim=(1,2)) to relay and test the torch codegen. Neither relay nor relax support this op (when call max with dims, the outputs behavior is different, sa https://pytorch.org/docs/stable/generated/torch.max.html). The customized ops converter is not included in this PR. I may contribute the parser part after MSC, while currently I focus on MSC. To support the special ops like torch.max with dim, I reconstruct codegen a little.

cc @Lunderberg

### DIFF
--- a/python/tvm/contrib/msc/core/codegen/codegen.py
+++ b/python/tvm/contrib/msc/core/codegen/codegen.py
@@ -142,7 +142,7 @@ def relay_to_relax(
     params: dict of <string:tvm.ndarray>
         The parameters of the IRModule.
     trans_config: dict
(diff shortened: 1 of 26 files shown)


Hzfengsy (approved):

Lunderberg:
Thank you for the fix!

PR #15941, opened 2023-10-17
[Unity][Transform] Improved canonicalization of non-dataflow Var

Lunderberg:
Prior to this commit, `relax.transform.CanonicalizeBindings` removed trivial bindings `var_y = var_x` where a `var_y: relax.DataflowVar` and `var_x: relax.Var`, but did not remove trivial bindings when `var_y: relax.Var` and `var_x: relax.DataflowVar`.  This was to avoid invalid use of a `relax.DataflowVar` outside of a dataflow block.

This commit updates `CanonicalizeBindings` to handle this type of binding as well.  To ensure that no `relax.DataflowVar` instances are used outside of a dataflow block, this is done by replacing `var_y: relax.DataflowVar` at its point of definition, instead of replacing `var_x: relax.Var` at its point of use.

This commit also canonicalizes `relax.Var` definitions to `relax.DataflowVar`, if the binding occurs within a dataflow block, and the variable is never used outside of a dataflow block.

### DIFF
--- a/src/relax/transform/canonicalize_bindings.cc
+++ b/src/relax/transform/canonicalize_bindings.cc
@@ -36,8 +36,8 @@ namespace relax {
 namespace {
 
 struct CanonicalizationPlan {
-  std::unordered_map<Id, Var, ObjectPtrHash, ObjectPtrEqual> replace_usage;
-  std::unordered_map<Id, Var, ObjectPtrHash, ObjectPtrEqual> replace_binding;
+  Map<Id, Var> replace_usage;
+  Map<Id, Var> replace_binding;
   std::unordered_set<Id, ObjectPtrHash, ObjectPtrEqual> bindings_to_remove;
 };
 
@@ -63,27 +63,18 @@ class CanonicalizePlanner : public ExprVisitor {
       Var bound_var = binding_iter.first;
       Var bound_to = binding_iter.second;
 
-      while (true) {
-        if (auto it = visitor.trivial_bindings_.find(bound_to);
-            it != visitor.trivial_bindings_.end()) {
-          // This may be a trivial binding into a trivial binding.  In
-          // that case, unwrap the bindings until we find the earliest
-          // non-trivial binding.
-          bound_to = it->second;
-        } else {
-          break;
-        }
+      while (auto opt = visitor.trivial_bindings_.Get(bound_to)) {
+        // This may be a trivial binding into a trivial binding.  In
+        // that case, unwrap the bindings until we find the earliest
+        // non-trivial binding.
+        bound_to = opt.value();
       }
 
-      while (true) {
-        if (auto it = plan.replace_binding.find(bound_to->vid); it != plan.replace_binding.end()) {
-          // The variable we are binding to may have already been
-          // replaced, if it fell into Case 4 (Var = DataflowVar).  In
-          // that case, we check against its replacement instead.
-          bound_to = it->second;
-        } else {
-          break;
-        }
+      while (auto opt = plan.replace_binding.Get(bound_to->vid)) {
+        // The variable we are binding to may have already been
+        // replaced, if it fell into Case 4 (Var = DataflowVar).  In
+        // that case, we check against its replacement instead.
+        bound_to = opt.value();
       }
 
       if (bound_var.as<DataflowVarNode>() || !bound_to.as<DataflowVarNode>()) {
@@ -94,7 +85,7 @@ class CanonicalizePlanner : public ExprVisitor {
         // For these three cases, the trivial binding can be
         // unwrapped, using the bound variable directly at the point
         // of use.
-        plan.replace_usage[bound_var->vid] = bound_to;
+        plan.replace_usage.Set(bound_var->vid, bound_to);
         plan.bindings_to_remove.insert(bound_var->vid);
         handled.insert(bound_to);
       } else {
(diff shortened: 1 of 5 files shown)


Lunderberg:
Rebased onto unity head and resolved conflicts with the https://github.com/apache/tvm/pull/15791.

src/relax/transform/canonicalize_bindings.cc
```diff
@@ -33,68 +33,198 @@
...
-    } else if (new_var.same_as(binding->var) && new_value.same_as(binding->value)) {
-      this->builder_->EmitNormalized(GetRef<VarBinding>(binding));
+  void VisitBinding(const Binding& binding) override {
+    bool has_same_struct_info = true;
+    Expr value;
+    if (auto ptr = binding.as<VarBindingNode>()) {
+      value = ptr->value;
+    } else if (auto ptr = binding.as<MatchCastNode>()) {
+      has_same_struct_info =
+          StructuralEqual()(GetStructInfo(binding->var), GetStructInfo(ptr->value));
```
slyubomirsky, line 144: Might we want to consider using the arithmetic solver to check if shapes are statically provable to be equal? The struct info can be equivalent without being structurally equal. (We should factor out that checking into a utility function if it isn't already, as this comes up a lot.)

src/relax/transform/canonicalize_bindings.cc
```diff
@@ -33,68 +33,198 @@
...
-        var_remap_[binding->var->vid] = parent.value();
-      } else {
-        // LHS and RHS have the same struct info, but the RHS is not a
-        // drop-in replacement for the LHS.
-        builder_->EmitNormalized(VarBinding(binding->var, new_value));
+    // Unwrap TupleGetItem, if the Tuple being accessed is known.
+    if (auto tuple_get_item = value.as<TupleGetItemNode>()) {
+      Expr tuple = tuple_get_item->tuple;
+      while (true) {
+        if (auto tuple_var = tuple.as<Var>()) {
```
slyubomirsky, line 154: Might it be cleaner to have `while (auto* tuple_var = tuple.as<VarNode>()) {` as the condition? I believe this should work and it would be correct.

src/relax/transform/canonicalize_bindings.cc
```diff
@@ -33,68 +33,198 @@
...
+        if (auto it = visitor.trivial_bindings_.find(bound_to);
+            it != visitor.trivial_bindings_.end()) {
+          // This may be a trivial binding into a trivial binding.  In
+          // that case, unwrap the bindings until we find the earliest
+          // non-trivial binding.
+          bound_to = it->second;
+        } else {
+          break;
+        }
+      }
```
slyubomirsky, line 76: I think this might work instead and might be clearer:
```c++
for (auto it = visitor.trivial_bindings_.find(bound_to);
    it != visitor.trivial_bindings_.end();
    it = visitor.trivial_bindings_.find(bound_to)) {
  bound_to = it->second;
} 
```
(Or the first clause can be moved to before the loop.)

Matter of preference, perhaps, if you consider it simpler than the above.

slyubomirsky (changes requested):
This is an excellent refactor and is very well documented. I made a couple of comments that I think would be worth having responses to before merging, but I think this leaves the transformation in great shape.

src/relax/transform/canonicalize_bindings.cc
slyubomirsky: The same applies to the loop below as well.

src/relax/transform/canonicalize_bindings.cc
Lunderberg: I'd like to do so, yes, though I think that would be best isolated in an independent PR.  Here, the check for having identical struct info is the same check as was done previously in the `CanCanonicalizeVar` method, and was kept the same in order to minimize the extent of changes.

src/relax/transform/canonicalize_bindings.cc
Lunderberg: Agreed, and updated.

src/relax/transform/canonicalize_bindings.cc
Lunderberg: I've gone back and forth on this one a bit.  On the one hand, the current version has an extra level of nesting, but on the other hand, using `for` would require repeating the `.find` clause.  In a perfect world, we would have `while(init_statement; condition)`, but unfortunately (according to [this stackoverflow post](https://stackoverflow.com/q/59985550)), that variation was left out for sake of simplicity.

Lunderberg: I made another update to use `Map<Id, Expr>` instead of `std::unordered_map<Id, Expr>`, so that the loops could be rewritten as `while(auto opt = lookup.Get(key))`, avoiding the `while(true)` altogether.

slyubomirsky: Yeah, I checked to see if `while (stmt1; condition)` would work and it doesn't. I'm not against having any `while(true)`, but avoiding a level of nesting is usually a good change.

src/relax/transform/canonicalize_bindings.cc
slyubomirsky: Okay. We can revisit this issue separately (I bet there are other passes where we're checking structural equality of struct info instead of static equivalence).

slyubomirsky (approved):
Thank you for addressing the comments.

Lunderberg:
No problem!

I'm going to merge this in, as it resolves a few test failures in unity head that resulted from a conflict between [PR#15791](https://github.com/apache/tvm/pull/15791) and [PR#15904](https://github.com/apache/tvm/pull/15904).  Each PR passed CI on its own, but the combination of the two changes caused a few minor test failures.  Since the test failures were resolved in this PR, that avoids needing to make a hotfix PR.

PR #15942, opened 2023-10-17
[Unity] Handle duplicate outputs in LazyTransformParams

Lunderberg:
A parameter transformation may output the same tensor more than once. When this occurs, the `set_item` function should be called for each output tensor.


Lunderberg:
Rebased onto main to resolve conflicts with #15855

masahi (approved):

PR #15951, opened 2023-10-19
[Unity][BYOC] Add support for sliding window in attention op

masahi:
This is necessary for Mistral support. A new attribute `window_size` has been added to `nn.attention(...)` op, and such sliding-window attention runs efficiently via flash attention. I updated our flash attention submodule to pick up the support for sliding window in https://github.com/tlc-pack/libflash_attn/pull/4.

@sunggg @yelite @cyx-6 @vinx13


sunggg (approved):
LGTM. Thank you!

vinx13 (approved):

PR #15954, opened 2023-10-19
[Unity] Split DecomposeOpsForTraining into two steps

Lunderberg:
Prior to this commit, the `DecomposeOpsForTraining` transform directly replaced `relax.nn.batch_norm` into more primitive relax operations. This required the decomposed form of `relax.nn.batch_norm` to be duplicated with `DecomposeOpsForInference`.  This commit refactors the pass to occur in two steps, first to apply training-specific mutations, and then to decompose.

Having a clear `DecomposeOps` pass also has a clear single location for operator decomposition, which may be migrated into the operator definition in the future, similar to `FLegalize`.

### DIFF
--- a/src/ir/transform.cc
+++ b/src/ir/transform.cc
@@ -531,6 +531,23 @@ Pass CreateModulePass(const runtime::TypedPackedFunc<IRModule(IRModule, PassCont
   return ModulePass(pass_func, pass_info);
 }
 
+Pass ApplyPassToFunction(Pass pass, String func_name, bool error_if_function_missing) {
+  auto pass_name =
+      static_cast<const std::stringstream&>(std::stringstream() << "ApplyPassTo" << func_name)
+          .str();
+
+  auto pass_func = [pass, func_name](IRModule mod, PassContext) -> IRModule {
+    auto gvar = mod->GetGlobalVar(func_name);
+    auto func = mod->Lookup(gvar);
+    auto mutated = pass(IRModule({{gvar, func}}));
+
+    mod.CopyOnWrite()->Update(mutated);
+    return mod;
+  };
+
+  return CreateModulePass(pass_func, 0, pass_name, {});
+}
+
 TVM_REGISTER_NODE_TYPE(PassInfoNode);
 
 TVM_REGISTER_GLOBAL("transform.PassInfo")
--- a/src/ir/transform.cc
+++ b/src/ir/transform.cc
@@ -31,6 +31,7 @@
 
 #include <chrono>
 #include <iomanip>
+#include <regex>
 #include <stack>
 #include <unordered_set>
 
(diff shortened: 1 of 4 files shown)


Lunderberg:
@sunggg This is what I was referring to by a different decomposition being applied for training and for inference in https://github.com/apache/tvm/pull/15842.  This PR extracts that difference out into a separate pass, such that both training and inference can then share the same underlying definition of `batch_norm`.

Lunderberg:
Rebased onto `unity` to re-run CI, as it has been long enough that discrepancies could have arisen since the previous run.

src/ir/transform.cc
```diff
@@ -531,6 +531,23 @@ Pass CreateModulePass(const runtime::TypedPackedFunc<IRModule(IRModule, PassCont
   return ModulePass(pass_func, pass_info);
 }
 
+Pass ApplyPassToFunction(Pass pass, String func_name, bool error_if_function_missing) {
```
sunggg, line 534: This function seems useful. Can we get the `List[String]` instead of single function name so that we can generalize a little further?

sunggg (commented):
Sorry, for the late review, Eric! Overall, LGTM. One question.

Lunderberg:
> Sorry, for the late review

No problem, as I've been slow on responding as well.

src/ir/transform.cc
Lunderberg: I think that would work, but only in the case where `error_if_function_missing = false`.  If there's a `List[String]`, then it wouldn't be clear to a reader whether an error is raised when all of the listed functions are absent, or when any of the listed functions are absent.

I've been picturing this utility as being useful for hand-crafted optimization pipelines, where some optimizations should only be applied to specific functions.  In those cases, raising an error when an expected function is missing would allow for earlier detection of an invalid optimization pipeline.

Lunderberg: As a third option, what if the `func_name` parameter is replaced with a `func_name_regex` parameter?  That way, it could handle the cases of a single function and of a list of functions, and could clearly state when an error occurs by renaming the parameter to `error_if_no_function_matches_regex`.

Lunderberg: And the more I think on it, the more I like the regex option.  Implemented, and going through CI now.

csullivan (approved):
This LGTM, thanks @Lunderberg. @sunggg please let @Lunderberg know if you have any follow ups or if we can merge. Thanks!

sunggg (approved):
Regex sounds cool! Thanks @Lunderberg for the update and thanks @csullivan for the reminder!

tqchen:
Quick note. this PR depends on std::regex and we know that there can be a symbol issue with pytorch, see https://github.com/apache/tvm/pull/16249 we may need to revert the usage or change to prefix matching for now so we can unblock unity main transition.

For future usage of regex, let us consider simple pattern and prefix for now.

Issue #15880, opened 2023-10-05
[Bug] [Unity] Fail to print out call trace under relax::Normalizer


jinhongyii:
### Expected behavior

printing out call trace when encountering `ICHECK` failure

### Actual behavior
```
python/tvm/relax/block_builder.py:645: in normalize
    return _ffi_api.BlockBuilderNormalize(self, expr)  # type: ignore
python/tvm/_ffi/_ctypes/packed_func.py:239: in __call__
    raise_last_ffi_error()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

    def raise_last_ffi_error():
        """Raise the previous error from FFI
    
        This should be used instead of `raise get_last_ffi_error()`, as it
        handle propagation of errors across an FFI boundary.  For example,
        if Python passes a callback to a C++ function, and the callback
        raises an exception, the re-thrown exception should contain the
        full stack trace, not just the stack frames that are above the
        outermost FFI call.
        """
    
        _LIB.TVMGetLastPythonError.restype = ctypes.c_void_p
        _LIB.TVMGetLastBacktrace.restype = ctypes.c_char_p
        py_err = _LIB.TVMGetLastPythonError()
        if py_err is None:
            c_err_msg = py_str(_LIB.TVMGetLastError())
            py_err_msg, err_type = c2pyerror(c_err_msg)
            if err_type is not None and err_type.startswith("tvm.error."):
                err_type = err_type[10:]
            py_err = ERROR_TYPE.get(err_type, TVMError)(py_err_msg)
    
        else:
            # TVMGetLastPythonError returns a PyObject*, with NULL when
            # there is no such value.  If we annotated the restype as
            # ctypes.py_object, we would need to return Py_None from the
            # C++ implementation.  This would require introducing a
            # dependency on libpython that we want to avoid when not in a
            # Python environment.  Therefore, casting the resulting void*
            # pointer to PyObject* using ctypes.
            py_err = ctypes.cast(ctypes.c_void_p(py_err), ctypes.py_object).value
    
        tb = py_err.__traceback__
    
[...]

PR #15955, opened 2023-10-20
[FFI][Python] Handle error propagation when line number is missing

Lunderberg:
Prior to this commit, the FFI error propagation in Python assumed that any stack frame with a known file name also had a known line number. This commit updates the error propagation to instead check if a known line number is present.

Closes https://github.com/apache/tvm/issues/15880.

### DIFF
--- a/python/tvm/_ffi/base.py
+++ b/python/tvm/_ffi/base.py
@@ -349,7 +349,7 @@ def _append_traceback_frame(tb, func_name, filepath, lineno: Optional[int]):
     # filepath to show code snippets from that FFI file.
     header = "" if lineno is None else "\n" * (lineno - 1)
     code = compile(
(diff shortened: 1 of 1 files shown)


Lunderberg:
@jinhongyii I have this PR marked as a draft, to prevent merging prior to your confirmation of its effect.

junrushao:
Hey do you think it's ready? I met the same issue and this one seems a good fix to me

Lunderberg:
@junrushao I think it is.  Since you ran into the same issue and can verify that the fix resolves it, I'm marking it as ready for review.  (I wasn't able to reproduce the original error, so I was waiting for confirmation from somebody who could.)

junrushao (approved):
Cannot wait to get it merged!

PR #15957, opened 2023-10-20
[Disco] Add loader for presharded params.

Lunderberg:
Prior to this commit, sharding of model weights was always performed when initializing the model.  This could cause slow initialization, especially for larger numbers of GPUs, as all model weights are initially transferred to GPU-0, before being scattered to all workers.

This commit updates the `tvm::runtime::ShardLoaderObj` to also allow loading of pre-sharded model weights.  With pre-sharded model weights, the tensors are sharded while the model is being built, and each worker independently loads the specific model weights that it requires.

### DIFF
--- a/src/runtime/disco/loader.cc
+++ b/src/runtime/disco/loader.cc
@@ -19,6 +19,7 @@
 #include <tvm/runtime/data_type.h>
 #include <tvm/runtime/packed_func.h>
 #include <tvm/runtime/registry.h>
+#include <unistd.h>
 
 #include <functional>
 #include <numeric>
@@ -47,10 +48,20 @@ class ShardLoaderObj : public Object {
                           std::string shard_info, Module mod);
   /*! \brief Load the i-th parameter */
   NDArray Load(int weight_index) const;
+
   /*! \brief Load all the parameters */
   Array<NDArray> LoadAll() const;
 
   NDArray ApplyShardFunc(const ShardInfo::ShardFunc& shard_func, const NDArray& param) const;
+
+  /*! \brief Load all the pre-sharded parameters */
+  Array<NDArray> LoadAllPresharded() const;
+
+  /*! \brief Load the i-th parameter from presharded binaries */
+  NDArray LoadPresharded(int weight_index) const;
+
+  /*! \brief Slice the given tensor at a specific dimension */
+  NDArray Shard(NDArray source, int dim, int num_slices) const;
 
   static constexpr const char* _type_key = "runtime.disco.ShardLoader";
   TVM_DECLARE_FINAL_OBJECT_INFO(ShardLoaderObj, Object);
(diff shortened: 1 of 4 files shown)


Lunderberg:
This PR is based on and supersedes https://github.com/apache/tvm/pull/15676.

Lunderberg:
Rebased onto main to re-run CI, as 2-week-old CI results are a bit stale for my preferences.

@junrushao Could I get a review on this PR?

masahi (approved):
I confirmed that it works. @junrushao Any concerns not to merge this?

src/runtime/disco/loader.cc
```diff
@@ -19,6 +19,7 @@
 #include <tvm/runtime/data_type.h>
 #include <tvm/runtime/packed_func.h>
 #include <tvm/runtime/registry.h>
+#include <unistd.h>
```
junrushao, line 22: Is it added by accident? Let’s try not to depend on Unix headers

src/runtime/disco/nccl/nccl.cc
```diff
@@ -162,7 +162,7 @@ struct CCLThreadLocalContext {
 
 void InitCCL(Session sess, IntTuple device_ids) {
   DRef func = sess->GetGlobalFunc("runtime.disco." TVM_DISCO_CCL_NAME ".init_ccl_per_worker");
-  LOG(INFO) << "Initializing " TVM_DISCO_CCL_NAME " with devices: " << device_ids;
+  // LOG(INFO) << "Initializing " TVM_DISCO_CCL_NAME " with devices: " << device_ids;
```
junrushao, line 165: Let’s recover this line

junrushao: Oh i got what you mean. Do you want to remove verbose logging message?

src/runtime/disco/loader.cc
Lunderberg: Thank you, and removed.  It was included while debugging, in order to show `getpid()` in debug print statements.

src/runtime/disco/nccl/nccl.cc
Lunderberg: Yup, that was my intent, as correct behavior should generally be silent.  That said, this is useful for debugging, so I've replaced it with `DLOG(INFO)`.

src/runtime/disco/worker.cc
```diff
@@ -62,8 +62,9 @@ void DiscoWorker::SetRegister(int reg_id, TVMArgValue value) {
 struct DiscoWorker::Impl {
   static void MainLoop(DiscoWorker* self) {
     ThreadLocalDiscoWorker::Get()->worker = self;
-    LOG(INFO) << "[Worker #" << self->worker_id << "] " << support::GetProcessIdAndThreadIdHeader()
-              << " started";
+    // LOG(INFO) << "[Worker #" << self->worker_id << "] " <<
```
junrushao, line 65: Let's remove the commented lines

Lunderberg: Thank you for the catch, and removed.

PR #15918, opened 2023-10-11
[TOPI][TIR][TE][x86] Extend x86 SIMD (u)int8 coverage for dense & conv2d

cbalint13:
This PR enhances x86 SIMD (u)int8 coverage for dense and conv2d operators.
It extends current SIMD support with avx2 & ssse3, and adds a new set of non-overflowing SIMD method.

---
#### Tracker:

- [ ] PR for TIR ```call_{pure}_llvm_intrin``` pretty print.
- [ ] PR for new TIR intrinsics
- [ ] PR for ```ArrayIntImm``` node

This PR will hold only the TOPI part.

---

#### Changes:

[x86][TOPI]
  * Extends [current tensorizer set](https://github.com/apache/tvm/blob/dc1b62916f68bf7b7f417eb4ea79bc75e83ec1a5/python/tvm/topi/x86/tensor_intrin.py#L130-L150), now called ```fast-math``` overflowing one, with ```avx2``` and ```ssse3```.
  * Adds a [new tensorizer set](https://github.com/apache/tvm/blob/dc1b62916f68bf7b7f417eb4ea79bc75e83ec1a5/python/tvm/topi/x86/tensor_intrin.py#L151-L198), a "precision" non-overflowing one, supporting: ```avx512```,  ```avx2``` and ```ssse3```.
  * Enable proper operator data alignment to gurantee any data (even smallest) will fit the SIMD vector width.

[TIR][LLVM]
 * Adds new TIR ops mapped to LLVM instrinsics: ```zextend```, ```sextend```, ```truncate``` for type conversions.
 * Adds new TIR ops mapped to LLVM instrinsics: ```vectorpermute```, ```vectorshuffle``` for vector manipulation.
 * The ```call_llvm_pure_intrin``` & ```call_llvm_intrin``` now holds instruction ```StringImm``` instead of ```IntImm``` abstract.
 * Enables TIR op ```atomic_add``` mapped to proper LLVM intrinsic guarnteed (best-effort) to lower to single instruction.

[TE]
 * Introduce new ```ArrayIntImm``` expression for small immediate list of integer constants.

[Target]
* Introduce a flag ```-key=cpu,fast-math``` to switch from the precise SIMD (default) to the overflowing SIMD set.

#### Performance

For the new avx2 & ssse3 the ```fast``` vs. ```precise``` SIMD sets:

```
$ python3 tests/python/contrib/test_gemm_acc32_simd.py
[...]

PR #15960, opened 2023-10-20
[TIR][LLVM] Expose TIR api for llvm sext/zext and trunc native type converter intrinsics

cbalint13:
This PR expose new tir api operators binded to their llvm native intrinsic counterparts.
Adds the ability to emit native cpu intrinsics for atomic type conversions of vectors for tensorizers.

---

#### Changes

 * Adds new TIR ops mapped to LLVM instrinsics: ```zextend```, ```sextend```, ```truncate``` for type conversions.
 * Enables TIR op ```atomic_add``` mapping to proper LLVM intrinsic guaranteed (best-effort) to lower to single instruction.

#### Rationale

Some highly efficient CPU intrinsics related to data type manipulations of whole vectors **are not exposed** by LLVM.
As substitute LLVM offers "higher level functions" with guarantees that will emit the exact & right instruction on CPU.

#### Example

* On x86 we want to expand a vector from ```uint8x16``` ->  ```uint16x16``` or perhaps sign expand to ```int16x16```. 
    In order to do this the [pmovzxwd](https://www.officedaytime.com/simd512e/simdimg/pmovsx.php?f=pmovzxwd) and [pmovsxbw](https://www.officedaytime.com/simd512e/simdimg/pmovsx.php?f=pmovsxbw)   are needed which are **not** exposed by LLVM directly. 
     The new ```zextend``` (non-sign, zero aware) and ```sextend``` (sign aware) functions can now do this:

   * Here is the tir invocation:
    ```
      vec_a_words = tvm.tir.zextend("uint16x8", vec_a_uint8x8)
      vec_b_words = tvm.tir.sextend("int16x8", vec_b_int8x8)
    ```
   * In the lowered x86 assembley there are now *exactly* the mentioned two instructions:
   ```
    1ea3:	c4 e2 79 30 e2       	vpmovzxbw %xmm2,%xmm4
    1ea8:	c4 e2 79 20 12       	vpmovsxbw (%rdx),%xmm2
   ```

#### Notes

A more complete example with real usage in a tensorization process with these new tir operators [can be seen here](https://github.com/apache/tvm/blob/0d338828eebaa3ff705e8521f2a1b3530f73dc7d/python/tvm/topi/x86/tensor_intrin.py#L152-L198) .
This also allows more TOPI/MS data type conversions leveraging precise control on involved atomic CPU instructions.

This PR is indispensable part of #15918 , an effort towards int8 tensorization coverage on x86.

---

Cc: @Lunderberg , @junrushao , @masahi , @vinx13, @ekalda , @lhutton1 , @quic-sanirudh , @kparzysz-quic

### DIFF
--- a/include/tvm/tir/builtin.h
+++ b/include/tvm/tir/builtin.h
@@ -51,6 +51,21 @@ TVM_DLL const Op& ret();
 TVM_DLL const Op& reinterpret();
 
 /*!
+ * \brief Zero extend the value using the target type.
+ */
+TVM_DLL const Op& zextend();
+
+/*!
+ * \brief Sign extend the value using the target type.
+ */
+TVM_DLL const Op& sextend();
+
+/*!
+ * \brief Truncate the value using the target type.
+ */
+TVM_DLL const Op& truncate();
+
+/*!
  * \brief Marks a condition is likely going to happen.
  */
 TVM_DLL const Op& likely();
@@ -769,9 +784,10 @@ TVM_DLL const Op& vectorlow();
 TVM_DLL const Op& vectorcombine();
 
 /*!
- * \brief atomic add instruction, corresponding e.g. to atomicAdd in CUDA
+ * \brief Atomic add instruction.
  */
 TVM_DLL const Op& atomic_add();
+
 /*!
  * \brief Create an Nd memory allocation with storage scope
  */
--- a/python/tvm/script/ir_builder/tir/ir.py
+++ b/python/tvm/script/ir_builder/tir/ir.py
@@ -1869,6 +1869,9 @@ def _dtype_forward(func):
 
 
 reinterpret = _dtype_forward(_tir_op.reinterpret)
+sextend = _dtype_forward(_tir_op.sextend)
+zextend = _dtype_forward(_tir_op.zextend)
+truncate = _dtype_forward(_tir_op.truncate)
 call_extern = _dtype_forward(_tir_op.call_extern)
 call_intrin = _dtype_forward(_tir_op.call_intrin)
 call_llvm_intrin = _dtype_forward(_tir_op.call_llvm_intrin)
@@ -2072,6 +2075,9 @@ __all__ = [
     "q_multiply_shift_per_axis",
     "ret",
     "reinterpret",
+    "sextend",
+    "zextend",
+    "truncate",
     "round",
     "rsqrt",
     "shift_left",
--- a/python/tvm/tir/__init__.py
+++ b/python/tvm/tir/__init__.py
@@ -74,7 +74,7 @@ from .op import (
     create_barriers,
 )
 from .op import vectorlow, vectorhigh, vectorcombine
-from .op import infinity, reinterpret
+from .op import infinity, reinterpret, zextend, sextend, truncate
 from .op import exp, exp2, exp10, log, log2, log10, log1p, ldexp, clz
 from .op import sin, sinh, asin, asinh
 from .op import cos, cosh, acos, acosh
@@ -88,6 +88,7 @@ from .op import comm_reducer, min, max, sum
 from .op import q_multiply_shift, q_multiply_shift_per_axis, shift_left, shift_right
 from .op import TVMBackendAllocWorkspace, TVMBackendFreeWorkspace
 from .op import start_profile_intrinsic, end_profile_intrinsic
+from .op import atomic_add
 from .generic import add, subtract, multiply
 
 from .schedule import StmtSRef, BlockScope, ScheduleState, Schedule, ScheduleError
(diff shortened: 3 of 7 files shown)


cbalint13:
@Lunderberg ,

> Other than lack of hardware support, are there any cases where we wouldn't want to apply these intrinsics?

* Not aware of, in worst case LLVM's sext/zext will map to some "multiple" instructions that mimics that.

> If there aren't, I'm wondering if `llvm::Instruction::ZExt` and `llvm::Instruction::SExt` should be the default codegen for `T.cast(value_uint8x8, 'uint8x16')` and `T.cast(value_uint8x8, 'int8x8')`, respectively.

* Not sure where you want to point, but yes, Zext/Sext could replace those casts as you mentioned.


---

To sum up the needs here:

* The very need here is in this PR at [this proposed tensorization](https://github.com/apache/tvm/blob/0d338828eebaa3ff705e8521f2a1b3530f73dc7d/python/tvm/topi/x86/tensor_intrin.py#L152-L198) of PR#15918 (now splitted to multiple PRs).
* This would bring a ```precise``` (not overflowing !) set of int8 SIMD tensorizer on x86, backward down to ssse3 only.
* Current int8 SIMD set on x86 are overflowing and imprecise (except vnni), also TOPI lacks proper lane-size awareness.

A bit longer:

The ultimate goal, if 15918 is done, would be to have a solid int8 support (again not overflowing on x86 !), working even backward with older hardware (not everyone have vnni or avx512) to make LLM and other quantized models happier.

Lunderberg:
> Not sure where you want to point, but yes, Zext/Sext could replace those casts as you mentioned.

The main thing I'm wondering is whether the support for these conversions should be done through an LLVM-specific intrinsic, or through a change in `CodegenLLVM` to recognize the `T.cast` and call `builder_->CreateZExt` instead of `builder_->CreateIntCast`.  If this is always beneficial to do, then we wouldn't need to introduce an explicit intrinsic for it, as we would already have a way to express the same effect within TIR.

Though, looking at LLVM's implementation of `IRBuilderBase::CreateIntCast`, it looks like it already does produce a `Instruction::ZExt` when applicable.

> The need here for exposing tir sext/zext/truncate would be for fine control of intrinsics in tensorization.

Ah, that makes sense.  So if I understand correctly, we are currently relying on LLVM's choice of intrinsic for performing integer to integer casts, but we want to be able to override that choice of intrinsic by explicitly specifying it.

cbalint13:
@Lunderberg ,

> Though, looking at LLVM's implementation of `IRBuilderBase::CreateIntCast`, it looks like it already does produce a `Instruction::ZExt` when applicable.

Yes it is (a bit higher level abstract), then goes low to ```Zext/Sext``` and maps properly to cpu specific instruction.
In x86 case is [pmovzxwd](https://www.officedaytime.com/simd512e/simdimg/pmovsx.php?f=pmovzxwd) and [pmovsxbw](https://www.officedaytime.com/simd512e/simdimg/pmovsx.php?f=pmovsxbw) but **not** exposed, unavailable via ```llvm_pure_intrinsic("pmovzxwd",{...})```.

> 
> > The need here for exposing tir sext/zext/truncate would be for fine control of intrinsics in tensorization.
> 
> Ah, that makes sense. So if I understand correctly, we are currently relying on LLVM's choice of intrinsic for performing integer to integer casts, but we want to be able to override that choice of intrinsic by explicitly specifying it.

Yes exactly, all need here is in the tensorization process to control and manipulate **exactly** what [instruction is needed](https://github.com/apache/tvm/blob/0d338828eebaa3ff705e8521f2a1b3530f73dc7d/python/tvm/topi/x86/tensor_intrin.py#L152-L198).
 It is sufficient to expose zext, sext, truncate in tir and make it easly available if one need "exact" manipulation on.

---

More notes [maybe bit offtopic]:

To make even a stronger argument towards LLVM's  more stranger world, there is the RISCV RVV (V vector instruction):

 LLVM expose only *latest* RVV v1.0 specs, and tomorrow probably what will be the next RVV 2.x . But there are "real-world" CPU with only with RVV v0.7 spec version. Now LLVM refuse to maintain "versioned cpu extensions" so either accept their high level things emitting RVV 1.0 (latest) or simply go and emit instructions by yourself piece by piece (from the old set or whatever). See rationale [here](https://lists.llvm.org/pipermail/llvm-dev/2021-July/151679.html)
 What the thread does not even mention is that there are also "forks" of RVV like [T-Head's extensions](https://github.com/T-head-Semi/thead-extension-spec) , curiously some of their extended ISA got accepted into LLVM, but only "some".

  There will be more and more cases like this as NPU and all kind of wierd CPU extension landscape will grow.
 
Anyway, our case here is the proposed tensorization effort to put the "performance" on table.

Issue #10141, opened 2022-02-02
[Tracking Issue] Write schedules for RISC-V targets


areusch:
We added initial support for RISC-V in microTVM by proving that we can run regression tests against the Zephyr RISC-V target (backed by QEMU). However, we have not yet added support for any V- or P- extension SIMD instructions to provide more interesting performance numbers out of the box. This issue tracks adding those schedules, which involves:
- [ ] selecting a simulator which can correctly execute those instructions
- [ ] adding the schedules
- [ ] adding tests to verify functional correctness

PhilippvK:
I've recently worked with RISC-V vector instructions using these simulators:

- **Spike (`riscv-isa-sim`)** - https://github.com/riscv-software-src/riscv-isa-sim
  - Supports V-Ext. (v2.0) and P-Ext. (v0.9.2)
  - No pre-build binaries available - Need to compile simulator (`spike`) and proxy kernel (`pk`) from source
  - Debugging support with GDB very limited
- **riscvOVPSimPlus** - https://www.ovpworld.org/library/wikka.php?wakka=riscvOVPsimPlus
  - Version with V-Ext. support: Sign-up required for download, internet connection for usage
  - However, perfomance and usability is great (Much faster than Spike)

Both of them are only ISS, e.g. not cycle accurate, but that should not be an issue.

---

Another thing I would like to point out is that the vectorization support in the RISC-V GCC toolchain seems to be (References: [1](https://github.com/riscv-collab/riscv-gnu-toolchain/issues/975) [2](https://github.com/riscv-collab/riscv-gnu-toolchain/issues/990) [3](https://github.com/riscv-collab/riscv-gnu-toolchain/issues/968)). Thus, one should switch to using LLVM/Clang instead (LLVM also supports all RVV intrinsics).

PR #15964, opened 2023-10-21
[LLVM][RUNTIME] Add optional LLVM ORCJIT runtime executor

cbalint13:
This PR introduce a new LLVM ORC JIT executor for the runtime.
The new ORCJIT may obsolete MCJIT due to better upstream maintenance.

---

#### Changes

  * Old MCJIT is **default** , thus the previous behaviour is unchanged.
  * The new ORCJIT is added as a alternative and **optional** llvm runtime engine.

#### Rationale

  * [MCJIT](https://llvm.org/docs/MCJITDesignAndImplementation.html) is the first implementation in LLVM and looks to be unmaintained by upstream.
  * The new [ORCJIT](https://llvm.org/docs/ORCv2.html) is added and it shares the same flow configuration as the current MCJIT.
  * Targets like RISCV, and possible others too, will not work on MCJIT see some discussions [here](https://reviews.llvm.org/D127842) .
* The new ORCJIT offers flexibile custom compilation and linkage flow that can be useful for future challenges.

#### Usage

* The default is the old MCJIT so no changes in any behaviour of TVM, but being limited it fails on e.g. RISCV:
```
[23:35:43] src/target/llvm/codegen_llvm.cc:185: Warning: Set native vector bits to be 128 for riscv64
LLVM ERROR: Relocation type not implemented yet!
```
* The new ORCJIT can be invoked by ```llvm -jit=orcjit {...}```, this works even on RISCV targets.

This implementation here is backward compatibile and was tested on LLVM={10,11,12,13,14,15,16,17,18}

---

Halide project also [leverage orcjit runtime](https://github.com/halide/Halide/blob/eb66c06539d623459ea3c3abeea5b7cce111ad5f/src/JITModule.cpp#L380).
Also may relate to PR https://github.com/apache/tvm/pull/14836 and TK #10141 ongoing effort toward RISCV tensorization.

### DIFF
--- a/src/target/llvm/llvm_instance.cc
+++ b/src/target/llvm/llvm_instance.cc
@@ -256,8 +256,23 @@ LLVMTargetInfo::LLVMTargetInfo(LLVMInstance& instance, const Target& target) {
     }
   }
 
-  // Target options
+  // LLVM JIT engine options
+  if (const Optional<String>& v = target->GetAttr<String>("jit")) {
+    String value = v.value();
+    if ((value == "mcjit") || (value == "orcjit")) {
+      jit_engine_ = value;
+    } else {
+      LOG(FATAL) << "invalid jit option " << value << " (can be `mcjit` or `orcjit`).";
+    }
+  }
 
+  // RISCV code model
+  auto arch = llvm::Triple(triple_).getArch();
+  if (arch == llvm::Triple::riscv32 || arch == llvm::Triple::riscv64) {
+    code_model_ = llvm::CodeModel::Medium;
+  }
+
+  // Target options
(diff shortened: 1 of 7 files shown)


cbalint13:
This is ready for review.
Cc: @Lunderberg , @junrushao , @masahi , @vinx13, @tqchen , @ekalda , @lhutton1 , @quic-sanirudh , @kparzysz-quic
Cc: @katebern-grovety  , @PhilippvK

tqchen:
@cbalint13 do you mind to rebase on latest main? love to get it in

cbalint13:
@tqchen 

It is rebased now, glad to see it as useful.
Thank you !

PR #15966, opened 2023-10-23
[CMake] Fix order of GNUInstallDirs module

syalavarthi:
In CMakeLists.txt, the variables CMAKE_INSTALL_<dir> are referenced before the execution of GNUInstallDirs module. This order is incorrect as the variables are defined by the GNUInstallDirs module.

Since the variables are undefined, INTERFACE_INCLUDE_DIRECTORIES property is not set in the tvmTargets.cmake. This causes external applications based on TVM unable to find TVM runtime headers.


cbalint13 (approved):
LGTM thanks @syalavarthi !

echuraev (approved):

PR #15969, opened 2023-10-23
Implementation of built-in function log_softmax

vvchernov:
Built-in function "vm.builtin.logprobs_from_logits" calculates log_sofmax from logits. It is needed to log probes calculation on mlc-llm side

cc @tqchen

### DIFF
--- a/src/runtime/relax_vm/lm_support.cc
+++ b/src/runtime/relax_vm/lm_support.cc
@@ -43,7 +43,6 @@
 #include <tvm/runtime/relax_vm/memory_manager.h>
 #include <tvm/runtime/relax_vm/vm.h>
 
-#include <algorithm>
 #include <cmath>
 
 namespace tvm {
--- a/src/runtime/relax_vm/lm_support.cc
+++ b/src/runtime/relax_vm/lm_support.cc
@@ -536,18 +536,16 @@ void LogSoftmax(NDArray logits, const NDArray& output) {
 
   size_t seq_length = logits->shape[logits->ndim - 2];
   size_t vocab_length = logits->shape[logits->ndim - 1];
-  std::vector<float> data(vocab_length);
   float* res_ptr = static_cast<float*>(output->data);
 
   for (size_t seq_ind = 0; seq_ind < seq_length; ++seq_ind) {
-    data.clear();
     const float* plogits = static_cast<float*>(logits->data) + vocab_length * seq_ind;
     float* res_log_probs = res_ptr + vocab_length * seq_ind;
 
-    // Fill data and find max value
+    // Fill intermediate data and find max value
     float max_value = plogits[0];
     for (size_t i = 0; i < vocab_length; ++i) {
-      data[i] = plogits[i];
+      res_log_probs[i] = plogits[i];
       if (max_value < plogits[i]) {
         max_value = plogits[i];
       }
@@ -556,16 +554,16 @@ void LogSoftmax(NDArray logits, const NDArray& output) {
     // Compute denominator
     float sum = 0.0f;
     for (size_t i = 0; i < vocab_length; ++i) {
-      float value = data[i] - max_value;
+      float value = res_log_probs[i] - max_value;
       sum += expf(value);
-      data[i] = value;
+      res_log_probs[i] = value;
     }
 
     float log_sum = logf(sum);
 
-    // Compute log probes
+    // Compute final log probes
     for (size_t i = 0; i < vocab_length; ++i) {
-      res_log_probs[i] = data[i] - log_sum;
+      res_log_probs[i] = res_log_probs[i] - log_sum;
     }
   }
 }
--- a/src/runtime/relax_vm/lm_support.cc
+++ b/src/runtime/relax_vm/lm_support.cc
@@ -511,6 +511,7 @@ int SampleTopPFromProb(NDArray prob, double top_p, double uniform_sample) {
 
 TVM_REGISTER_GLOBAL("vm.builtin.sample_top_p_from_prob").set_body_typed(SampleTopPFromProb);
 
+// NOLINTNEXTLINE(runtime/references)
 void LogSoftmax(NDArray logits, NDArray& output) {
   /* log_softmax from logits is calculated.
    * Both operations are joined to more quick and stable calculations:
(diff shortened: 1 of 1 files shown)


tqchen:
Ideally we should decouple logsoftmax from sampling. So we have a logsoftmax first. Do exp to get prob, then apply sample from probs, this will also allow us to leverage optimizations in sample from prob

vvchernov:
Hello @tqchen! Thank you for your comment! I did it similar with "sample_top_p_from_prob" and "sample_top_p_from_logits" and it was wrong, there is no sampling. I've renamed funcs to avoid misunderstanding. Could you check it again?

src/runtime/relax_vm/lm_support.cc
```diff
@@ -510,6 +511,72 @@ int SampleTopPFromProb(NDArray prob, double top_p, double uniform_sample) {
 
 TVM_REGISTER_GLOBAL("vm.builtin.sample_top_p_from_prob").set_body_typed(SampleTopPFromProb);
 
+NDArray LogProbsFromLogits(NDArray logits) {
```
tqchen, line 514: Just call it logsoftmax

tqchen: Consider pass in the output NDArray, so they can be pre-allocated

tqchen (changes requested):

src/runtime/relax_vm/lm_support.cc
vvchernov: Do you mean I need to pre-allocate output NDArray and pass it as function arg?

src/runtime/relax_vm/lm_support.cc
```diff
@@ -510,6 +511,72 @@ int SampleTopPFromProb(NDArray prob, double top_p, double uniform_sample) {
...
+  auto final_result = NDArray::Empty(logits.Shape(), logits.DataType(), DLDevice{kDLCPU, 0});
+
+  size_t seq_length = logits->shape[logits->ndim - 2];
+  size_t vocab_length = logits->shape[logits->ndim - 1];
+  std::vector<std::pair<float, int>> data;
+  data.resize(logits->shape[logits->ndim - 1]);
+  float* res_ptr = static_cast<float*>(final_result->data);
+
+  for (size_t seq_ind = 0; seq_ind < seq_length; ++seq_ind) {
+    data.clear();
```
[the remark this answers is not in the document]
vvchernov: I sort data to get maximum value for robust calculation of exponents. The calculation is based on torch.log_softmax approach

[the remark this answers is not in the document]
tqchen: we can take max but not sorting, because sorting takes more time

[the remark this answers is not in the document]
vvchernov: A little bit more about performance of the calculation. As I saw in torch implementation, logits are splitted on chunks and exponential sum is calculated for the chunk in parallel threads and after the results are reduced to one sum. It is potentially can be done here and for other similar built-in functions (but in other PR). @tqchen what do you think about it?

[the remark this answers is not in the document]
tqchen: let us first benchmark and see if this becomes a bottleneck, if so we can then check to optimize it. We can in theory do it in GPU if needed

vvchernov:
Hello @tqchen! I've successfully checked the built-in function together with mlc-llm pipeline. Could you finally review the PR?

src/runtime/relax_vm/lm_support.cc
```diff
@@ -510,6 +510,68 @@ int SampleTopPFromProb(NDArray prob, double top_p, double uniform_sample) {
...
+    ICHECK_EQ(logits->shape[i], 1) << "The leading dimensions of logits must be 1";
+  }
+
+  size_t seq_length = logits->shape[logits->ndim - 2];
+  size_t vocab_length = logits->shape[logits->ndim - 1];
+  std::vector<float> data(vocab_length);
+  float* res_ptr = static_cast<float*>(output->data);
+
+  for (size_t seq_ind = 0; seq_ind < seq_length; ++seq_ind) {
+    data.clear();
```
tqchen, line 543: do you needs to clear it? i feel data is not needed. you can just operate on inputs 

and result_log_probs inplace,

First get the max  value, then do the subtraction

tqchen (changes requested):

src/runtime/relax_vm/lm_support.cc
tqchen: aka you can use the output array to store the intermediate result of expf and then run the final subtractions

vvchernov: Hello @tqchen! I've fixed it

vvchernov:
Hello @tqchen! Could you finish review and approve PR?

tqchen:
After visiting some of the use-cases. I think we should not move logsoftmax here, mainly due to efficiency reasons. The cost of calculating related data can become a bottleneck.

Instead, let us simply use `log(prob) if prob > 1e-6 else log(1e-6)` where prob is the result of softmax to get the logprob, this would be stable for most cases and saves the extra set of computations.

vvchernov:
Hello @tqchen! I think there is some misunderstanding.
As I understood you want to transfer high-weight calculation (like expf) onto GPU side.
I based on SampleTopPFromLogits method ("vm.builtin.sample_top_p_from_logits") which is implemented on TVM side, there is [expf](https://github.com/apache/tvm/blob/e7d12be07aba6e6dedade041d47146df33531fa4/src/runtime/relax_vm/lm_support.cc#L355) calculation and it is used by mlc-llm on [CPU side](https://github.com/mlc-ai/mlc-llm/blob/65478c88ab989a09b6c05b21aad7d24a5e547a80/cpp/llm_chat.cc#L1323C12-L1323C37).
Possibly you think about softmax_with_temperature-like method compiled on relax for GPU calculation. I can do so, but I want to do it as model independent method and did not see any examples on mlc-llm side. But I saw SampleTopPFromLogits and followed this way.

Some notes:
1. In my case there are no top_p or temperature (in other words they are both zero).
2. Torch specially uses log_softmax method though it has log and softmax separately. It is explained as more stable calculation (and I repeated it excluding chunk calculation of exp sum). log_softmax(x) = x-max(x) - log(sum(exp(x-max(x))))
3. I'm not aware about performance due to this method is needed for loglikelihood approach of accuracy benchmark. But if it is used for speculative decoding or other logit mechanics like beam search the performance issue become more actual and critical.

PR #15971, opened 2023-10-24
[Unity][Op] Allow the argument to `call_tir` to be a var bound to a tuple, not a tuple literal

slyubomirsky:
In response to the issue in common subexpression elimination noted in this [forum post](https://discuss.tvm.apache.org/t/validity-of-common-subexpression-elimination-pass-that-simplifies-call-tir-args/15832), this PR fixes the case noted by allowing `call_tir` and its variants to correctly handle the case where they are passed a _variable_ bound to a tuple of arguments, rather than a tuple literal. This was an easy change to make and corresponds to user expectations.

In principle, we could address the issue highlighted in other ways, such as by changing common subexpression elimination to avoid doing elimination inside `call_tir` calls or treating `call_tir` as a special case during normalization. I went with this approach because I believe it to be simpler; however, we may consider other approaches if it is absolutely essential that the argument be a tuple _literal_. (I would argue that that would not be a particularly good design.)

### DIFF
--- a/python/tvm/relax/op/base.py
+++ b/python/tvm/relax/op/base.py
@@ -25,7 +25,7 @@ from tvm.runtime.object import Object
 from . import _ffi_api
 from ..expr import Expr, StringImm, ShapeExpr, Call, ExternFunc, GlobalVar, Var
 from ..expr import Tuple as RxTuple
-from ..struct_info import StructInfo, TensorStructInfo
+from ..struct_info import StructInfo, TensorStructInfo, TupleStructInfo
 from ...ir import PrimExpr
 from ..utils import args_converter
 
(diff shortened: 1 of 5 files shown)


quic-sanirudh (approved):
LGTM, thanks for the quickfix

quic-sanirudh:
> This may be a duplicate of #15916, which also resolves the analogous problem for `FuseOps`, `RewriteDataflowReshape`, and `FoldConstant`.

Oh nice, this is great, thanks for this. I was running to issues in the some of the same passes when testing locally, and so I was starting to work on a fix to follow up this PR, but looks like they're all fixed in #15916, that saves me a bunch of time 😄, so thanks again for that.

slyubomirsky:
Didn't see it, yep, they're duplicates. There is one case that the other PR misses so hopefully that can be updated.

PR #15976, opened 2023-10-24
[Unity] Forbid iteration over arbitrary relax.Expr

Lunderberg:
Prior to this commit, iteration over a `relax.Expr` relied on an exception being thrown from `TupleGetItem`, which was then re-raised as an `IndexError`.  This allowed Python's default implementation of iteration over an object that defines `__getitem__`, but doesn't define `__iter__` or `__len__` to terminate.  However, if the object being iterated over did not have `TupleStructInfo`, it would result in an infinite loop.

This commit adds an explicit implementation of `__iter__`, which raises a `TypeError` when called, preventing the default implementation provided by Python.  This is an intentional breakage of backwards compatibility, as arbitrary `relax.Expr` instances were not intended to be iterable.

### DIFF
--- a/python/tvm/relax/expr.py
+++ b/python/tvm/relax/expr.py
@@ -19,7 +19,7 @@
 """The expression nodes of Relax."""
 import typing
 from numbers import Number
-from typing import Any, Callable, Dict, List, Optional, Union, Mapping
+from typing import Any, Callable, Dict, List, Optional, Union, Mapping, Iterator
 
 import numpy as _np  # type: ignore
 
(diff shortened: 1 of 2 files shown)


Lunderberg:
https://github.com/apache/tvm/pull/15975 and https://github.com/apache/tvm/pull/15976 are two possible implementations to resolve issues with iteration over relax objects.  Both are currently in draft form, to allow discussion and to check what issues arise from each implementation.

tqchen:
Thanks @Lunderberg . love the proposed change to help alleviate the potential bugs, can you fix the testcases around this one and we can merge it in

PR #15981, opened 2023-10-25
[Unity][UnitTest] Cleanup test_vm_build.py

Lunderberg:
- Removed unused `import os`

- Used `tvm.testing.main()` inside `if __name__=="__main__"`

- Added parametrized fixture `exec_mode` instead of marking all tests.

- Replace `@pytest.mark.xfail` with `with pytest.raises`, based on intended test behavior in comments.  The `@pytest.mark.xfail` allows a test to fail, and is intended for marking known bugs or not-yet-implemented functionality.  `with pytest.raises` requires that the block throw an instead of `@pytest.mark.xfail`, and is intended for validating that error-checking paths function correctly.

- Renamed `test_save_function_time_evaluator` to `test_save_function_time_evaluator_rpc`, to avoid having two distinct definitions of `test_save_function_time_evaluator`.

### DIFF
--- a/tests/python/relax/test_vm_build.py
+++ b/tests/python/relax/test_vm_build.py
@@ -14,13 +14,13 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-import os
+
 import ctypes
 from typing import Tuple, Callable
 
-
 import numpy as np
 import pytest
+
 import tvm
 import tvm.script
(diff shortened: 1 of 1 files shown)


tqchen (approved):

Lunderberg:
Thank you, @tqchen!  I've rebased to resolve the conflicts.

yongwww (approved):

PR #15983, opened 2023-10-25
[FFI] Allow IntImm arguments to PackedFunc with int parameter

Lunderberg:
TVM containers, such as tvm::runtime::Array, require the contained objects to inherit from `ObjectRef`.  As a result, the wrapper types `IntImm`, `FloatImm`, and `StringImm` are often used to allow native types in the TVM containers.  Conversions into these wrapper type may be required when using a container, and may be performed automatically when passing an object across the FFI.  By also providing conversion to an unwrapped type, these automatic conversions are transparent become transparent to users.

### DIFF
--- a/include/tvm/runtime/packed_func.h
+++ b/include/tvm/runtime/packed_func.h
@@ -589,7 +589,7 @@ class TVMPODValue_ {
     } else if (auto opt = ThroughObjectRef<double>()) {
       return opt.value();
     } else if (auto opt = ThroughObjectRef<int64_t>()) {
-      return opt.value();
+      return static_cast<double>(opt.value());
     }
     TVM_CHECK_TYPE_CODE(type_code_, kDLFloat);
     return value_.v_float64;
--- a/include/tvm/runtime/packed_func.h
+++ b/include/tvm/runtime/packed_func.h
@@ -37,6 +37,7 @@
 #include <functional>
 #include <limits>
 #include <memory>
+#include <optional>
 #include <string>
 #include <tuple>
 #include <type_traits>
@@ -537,6 +538,42 @@ struct ObjectTypeChecker<Map<K, V>> {
   }
 };
 
+class TVMPODValue_;
+
+/*!
+ * \brief Type trait to specify special value conversion rules from
+ *     ObjectRef to primitive types.
+ *
+ *  TVM containers, such as tvm::runtime::Array, require the contained
+ *  objects to inherit from ObjectRef.  As a result, the wrapper types
+ *  IntImm, FloatImm, and StringImm are often used to hold primitive
+ *  types inside a TVM container.  Conversions into this type may be
+ *  required when using a container, and may be performed
+ *  automatically when passing an object across the FFI.  By also
+ *  handling conversions from wrapped to unwrapped types, these
+ *  conversions can be transparent to users.
+ *
+ *  The trait can be specialized to add type specific conversion logic
+ *  from the TVMArgvalue and TVMRetValue.
+ *
+ * \tparam T The type (e.g. int64_t) which may be contained within the
+ *     ObjectRef.
+ *
+ * \tparam (anonymous) An anonymous and unused type parameter, which
+ *     may be used for SFINAE.
+ */
+template <typename T, typename = void>
+struct PackedFuncObjectRefConverter {
+  /*!
+   * \brief Attempt to convert an ObjectRef from an argument value.
+   *
+   * \param obj The ObjectRef which may be convertible to T
+   *
+   * \return The converted result, or std::nullopt if not convertible.
+   */
+  static std::optional<T> TryFrom(const ObjectRef& obj) { return std::nullopt; }
+};
+
 /*!
  * \brief Internal base class to
  *  handle conversion to POD values.
@@ -549,25 +586,41 @@ class TVMPODValue_ {
     // the frontend while the API expects a float.
     if (type_code_ == kDLInt) {
       return static_cast<double>(value_.v_int64);
+    } else if (auto opt = ThroughObjectRef<double>()) {
+      return opt.value();
+    } else if (auto opt = ThroughObjectRef<int64_t>()) {
+      return opt.value();
     }
     TVM_CHECK_TYPE_CODE(type_code_, kDLFloat);
     return value_.v_float64;
   }
   operator int64_t() const {
+    if (auto opt = ThroughObjectRef<int64_t>()) {
+      return opt.value();
+    }
     TVM_CHECK_TYPE_CODE(type_code_, kDLInt);
     return value_.v_int64;
   }
   operator uint64_t() const {
+    if (auto opt = ThroughObjectRef<uint64_t>()) {
+      return opt.value();
+    }
     TVM_CHECK_TYPE_CODE(type_code_, kDLInt);
     return value_.v_int64;
   }
   operator int() const {
+    if (auto opt = ThroughObjectRef<int>()) {
+      return opt.value();
+    }
     TVM_CHECK_TYPE_CODE(type_code_, kDLInt);
     ICHECK_LE(value_.v_int64, std::numeric_limits<int>::max());
     ICHECK_GE(value_.v_int64, std::numeric_limits<int>::min());
     return static_cast<int>(value_.v_int64);
   }
   operator bool() const {
+    if (auto opt = ThroughObjectRef<bool>()) {
+      return opt.value();
+    }
     TVM_CHECK_TYPE_CODE(type_code_, kDLInt);
     return value_.v_int64 != 0;
   }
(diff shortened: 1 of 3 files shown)


include/tvm/runtime/packed_func.h
```diff
@@ -549,25 +586,41 @@ class TVMPODValue_ {
     // the frontend while the API expects a float.
     if (type_code_ == kDLInt) {
       return static_cast<double>(value_.v_int64);
+    } else if (auto opt = ThroughObjectRef<double>()) {
+      return opt.value();
+    } else if (auto opt = ThroughObjectRef<int64_t>()) {
+      return opt.value();
```
Aleksei-grovety, line 592: Shouldn't there be `static_cast<double>`, as in line 588?

include/tvm/runtime/packed_func.h
```diff
@@ -537,6 +538,42 @@ struct ObjectTypeChecker<Map<K, V>> {
...
 };
 
+class TVMPODValue_;
+
+/*!
+ * \brief Type trait to specify special value conversion rules from
+ *     ObjectRef to primitive types.
+ *
+ *  TVM containers, such as tvm::runtime::Array, require the contained
+ *  objects to inherit from ObjectRef.  As a result, the wrapper types
```
Aleksei-grovety, line 548: Trailing whitespace, also in lines 550, 552.

tests/cpp/packed_func_test.cc
```diff
@@ -319,3 +319,62 @@ TEST(TypedPackedFunc, RValue) {
...
+  func(tvm::IntImm(DataType::Int(32), 10));
+}
+
+TEST(TypedPackedFunc, FloatImmWrapper) {
+  using namespace tvm::runtime;
+
+  TypedPackedFunc<void(double)> typed_func = [](double x) {};
+  PackedFunc func = typed_func;
+
+  // Argument may be provided as a floating point.  If provided as an
```
Aleksei-grovety, line 346: Trailing whitespace.

Aleksei-grovety (commented):
Thanks @Lunderberg! Looks good, just some comments.

include/tvm/runtime/packed_func.h
Lunderberg: Good catch.  It isn't strictly necessary in either case, due to C++'s implicit conversion of numeric types, but would be good to call attention to it.  Updated to use a `static_cast<double>` here.

include/tvm/runtime/packed_func.h
Lunderberg: Hmm, I'm not seeing the trailing whitespace in the commit, either using `git show --word-diff-regex="[ ]+|[^ ]+"`, setting my editor to show whitespace, or in the github diff.

Aleksei-grovety: Sorry, I used the wrong definition, there is a double space before "As"

Lunderberg: Ah, I see.  This is intentional, but is more of a stylistic difference than anything else.  Traditionally, (accidental diversion into typography) a space after a sentence would use an em space (the same with as a capital M) after a sentence, wider than the spacing between words.  This isn't possible in monospaced fonts, so typewriters would emulate the em space by using two spaces.  Exactly which is better has since become the topic of a great many flame wars.

I was curious, and it looks like double-spacing after a sentence is twice as common as single-spacing in the TVM repo, with the Apache copyright header being the most prominent example.

```bash
# Count the number of periods sentences followed by one space
$ find . \
  \( -path ./3rdparty -o -path "./build*" \) -prune -o \
  \( -name "*.cc" -o -name "*.h" \) \
  -exec grep '[A-Za-z]\{2\}\. [A-Za-z]' {} /dev/null \; \
    | wc --lines
3958

# Count the number of periods sentences followed by two spaces
$ find . \
  \( -path ./3rdparty -o -path "./build*" \) -prune -o \
  \( -name "*.cc" -o -name "*.h" \) \
  -exec grep '[A-Za-z]\{2\}\.  [A-Za-z]' {} /dev/null \; \
    | wc --lines
7666
```

tqchen:
Given the implementation complexity i feel it would be helpful to not support this implicit behavior for now. This is because IntImm are used in other purposes(e.g. that comes with dtype).

This may start to make sense once we start to introduce a BoxInt that do not contain the dtype field and simply serve as a dtype function.

I assume downstream packedfunc will still be able to handle the function correctly

Lunderberg:
This arose when writing unit tests for access of a relax tuple.  I expected the following unit test to pass on `unity` HEAD.

```python
import tvm.testing

exec_mode = tvm.testing.parameter("bytecode", "compiled")

def test_vm_tuple_get_item(exec_mode):
    @R.function(private=True)
    def func(arg: R.Tuple([R.Prim("int64"), R.Prim("float32")])):
        return arg[0]

    mod = tvm.IRModule({"main": func})

    target = tvm.target.Target("llvm", host="llvm")
    ex = tvm.relax.build(mod, target, exec_mode=exec_mode)
    vm = tvm.relax.VirtualMachine(ex, tvm.cpu())

    res = vm["main"]((17, 42.5))
    assert res == 17
```

The error occurs with the following steps:

1. When passed an argument, python tuples are converted to `tvm.ir.Array`.
2. In order to use `tvm.ir.Array`, the arguments must be converted to `ObjectRef` subclasses.  This converts `(17, 42.5)` into `tvm.ir.Array([ T.int64(17), T.float32(42.5) ])`.
3. The tuple is accessed with a relax VM builtin, and `T.int64(17)` is stored in a relax VM register.
4. The `T.int64(17)` is type-checked as a `R.prim('int64')`, which uses `TVMPODValue_::operator int64_t`.
5. The call to `operator int64_t` fails, as it is an `IntImm`, and uses `TVMArgTypeCode::kTVMObjectHandle` and not `TVMArgTypeCode::kTVMArgInt`.

Any of these steps could be broken in order to avoid the error.  Fundamentally, the error occurs because we apply an automatic conversion (`int` to `IntImm`), but do not automatically apply the reverse conversion (`IntImm` to `int`).  I believe that this PR, which changes step (5), is the best resolution.  Avoiding the error by updating any of the other steps has drawbacks as described below.

1. Change the FFI conversion of python tuples to a runtime-only type, rather than `tvm.ir.Array`.  This would cause breakage for any FFI call that uses `Array` at compile-time.
2. Change the internal type of `tvm.ir.Array` from `ObjectRef` to `TVMArgValue`.  This would allow FFI conversions from python tuples to `Array`, but would break the usage of those `Array` objects.
3. Change the return of the `tuple_getitem` builtin to convert from `IntImm` to `int` when providing a `TVMRetValue`.  Possible, but would drop the dtype information that may be required for later type-checks.
4. Change the type-checking of `R.Prim` in the VM to check for `IntImm`.  This would avoid the error in this case, but this function is used for type-checking before calling external `PackedFunc` instances.  Those calls would throw the same error at a later point.
5. Change the `operator int64_t` implementation to check for an `IntImm`.  That is the change implemented in this PR, and handles any such conversion required.

> Given the implementation complexity

Can you explain what you mean here?  This doesn't seem any more complex than the argument handling that we already do for FFI compatibility purposes.

tqchen:
I understand the error scenario you described. The main issue here is that IntImm and FloatImm are compile time construct that are not defined as partof the runtime (because they are coupled as Expr).

As a result, the implementation won't allow us to support the case where only tvm_runtime is available, which is a common case to support. A better approach would be to enable boxed `BoxInt` and `Float` that are decoupled from IntImm and FloatImm, and migrate the existing usage of IntImm to those runtime boxed cases.

Given tuple of two prim values are less commonly used and we only recently supported the prim values. I think it is better for us to not support the case for now, so we can wait for longer term solution that separates out dedicated runtime boxed values, in which case automatic conversion would makes sense and there won't be limitations that it won't work in a runtime onlys etting.

Lunderberg:
> A better approach would be to enable boxed BoxInt and Float that are decoupled from IntImm and FloatImm, which can exist in the runtime, and migrate the existing usage of IntImm to those runtime boxed cases.

I like the idea of having the separate runtime-only types, as distinct from the compile-time representations.  I was concerned that that would be a larger impact change, but I agree with that as the long-term approach.

For now, closing this PR to avoid any premature merging.

> Given tuple of two prim values are less commonly used and we only recently supported the prim values.

For my understanding, I thought that `PrimValue` had always been present, back to the initial AST definition in #13901.  Use of a `PrimValue` to define a symbolic variable is relatively recent, but the failure mode doesn't depend on that functionality.

Lunderberg:
This PR has been superseded by https://github.com/apache/tvm/pull/16183, which implements boxed primitives that can be used as part of `libtvm_runtime.so`, along with the type conversions required for backwards compatibility.  This includes the conversion of a boxed primitive to an unboxed primitive, which is the functionality required for relax tuples containing primitives.

PR #15986, opened 2023-10-26
[Fix][TIR]fix symbolic strides lower

JackWeiw:
`compact_buffer_region` PASS modify shared buffer stride[0] to

`T.int64(72) * T.min((n + T.int64(63)) // T.int64(64) * T.int64(64), T.int64(96))` and stride[1] is `T.int64(72)`
but in LowerOpaqueBlock PASS it report error:
InternalError: Check failed: (is_zero(floormod(buffer->strides[i - 1], buffer->strides[i]))) is false:

For more detaied discuss, see [here](https://discuss.tvm.apache.org/t/bug-tir-symbolic-floormod/15826)

Another bug occurs in  PASS InjectPTXAsyncCopy .
that is dst_offset.dtype could be int64, the dtype of PrimExpr(index_factor) would be set to default to int32.
cause dtype inconsistent when calling tir::Mul. 

To reproduce the problem in InjectPTXAsyncCopy, see script [here](https://gist.github.com/JackWeiw/5b80956ab44c0f63d4f434f18f42cc89)

### DIFF
--- a/src/tir/transforms/inject_ptx_async_copy.cc
+++ b/src/tir/transforms/inject_ptx_async_copy.cc
@@ -24,6 +24,7 @@
 #include <tvm/tir/analysis.h>
 #include <tvm/tir/builtin.h>
 #include <tvm/tir/expr.h>
+#include <tvm/tir/op.h>
 #include <tvm/tir/stmt_functor.h>
 #include <tvm/tir/transform.h>
 
@@ -79,7 +80,7 @@ class PTXAsyncCopyInjector : public StmtMutator {
         if (indices_lanes == 1) {
           auto src_offset = load->indices[0];
           auto dst_offset = store->indices[0];
-          Array<PrimExpr> args = {store->buffer->data, tir::Mul(dst_offset, PrimExpr(index_factor)),
+          Array<PrimExpr> args = {store->buffer->data, mul(dst_offset, PrimExpr(index_factor)),
                                   load->buffer->data, src_offset, PrimExpr(bytes)};
           // use arguments size to indicate whether or not to use predicated cp.async
           if (predicated) {
@@ -114,7 +115,7 @@ class PTXAsyncCopyInjector : public StmtMutator {
           }();
           if (src_offset.defined() && dst_offset.defined()) {
             return Evaluate(Call(store->buffer->dtype, tvm::tir::builtin::ptx_cp_async(),
-                                 {store->buffer->data, tir::Mul(dst_offset, PrimExpr(index_factor)),
+                                 {store->buffer->data, mul(dst_offset, PrimExpr(index_factor)),
                                   load->buffer->data, src_offset, PrimExpr(bytes)}));
           }
         } else {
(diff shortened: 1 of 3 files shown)


JackWeiw:
CC @Lunderberg @wrongtest-intellif

tests/python/unittest/test_tir_transform_lower_opaque_block.py
```diff
@@ -250,6 +250,34 @@ def transformed_strided_buffer_func(
             C[i0 * 4 + i1, j] = B[i1, j] * T.float32(2)
 
 
+@T.prim_func
+def compacted_symbolic_strided_buffer_func(a: T.handle) -> None:
+    n = T.int64()
+    A = T.match_buffer(a, (1, n, 10240), "float32")
+    for i, j, k in T.grid(((n + 63) // 64 * 4 + 7) // 8, 2, 160):
```
Lunderberg, line 257: Unrelated, the presence of this expression is kind of odd to me.  Assuming this example came from a TIR printout, I would have expected `((n + 63) // 64 * 4 + 7) // 8` to be simplified to the equivalent `(n + 127) // 128`.  The fact that it didn't simplify may indicate that I should take a look at the `CanonicalSimplifier`.

src/tir/transforms/inject_ptx_async_copy.cc
```diff
@@ -79,7 +80,7 @@ class PTXAsyncCopyInjector : public StmtMutator {
         if (indices_lanes == 1) {
           auto src_offset = load->indices[0];
           auto dst_offset = store->indices[0];
-          Array<PrimExpr> args = {store->buffer->data, tir::Mul(dst_offset, PrimExpr(index_factor)),
+          Array<PrimExpr> args = {store->buffer->data, mul(dst_offset, PrimExpr(index_factor)),
```
Lunderberg, line 83: Since your description mentions this as a separate bug, can it either be split out into a separate PR, or (since it is a relatively small change), have a test case added for it?

tests/python/unittest/test_tir_transform_lower_opaque_block.py
```diff
@@ -250,6 +250,34 @@ def transformed_strided_buffer_func(
...
+                    T.reads(A[0, i * 128 + j * 32 + ax0, k * 64 + ax1])
+                    T.writes(A_pad_shared_dyn[0, ax0, ax1])
+                    A_pad_shared_dyn[0, ax0, ax1] = T.if_then_else(i * 128 + j * 32 + ax0 < n, A[0, i * 128 + j * 32 + ax0, k * 64 + ax1], T.float16(0))
+
+
+@T.prim_func
+def transformed_symbolic_strided_buffer_func(a: T.handle):
+        n = T.int64()
+        A = T.match_buffer(a, (1, n, 10240))
+        for i, j, k in T.grid(((n + T.int64(63)) // T.int64(64) * T.int64(4) + T.int64(7)) // T.int64(8), 2, 160):
```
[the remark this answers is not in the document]
JackWeiw: Thank u for the adivice, i've modified it using` T.int32` and pulled out into `padded_size = T.meta_var(T.min((n + T.int64(63)) // T.int64(64) * T.int64(64))` in the test case.

src/tir/transforms/inject_ptx_async_copy.cc
JackWeiw: I open a new PR [here](https://github.com/apache/tvm/pull/16000) . Please have a check. 
I will open a new PR to fix dtype mismatch bug in PASS  `InjectPTXAsyncCopy` after symbolic strides PR is merged

PR #15990, opened 2023-10-26
[ACL] Update Compute Library to v23.08

leandron:
Update Compute Library to follow latest developments in 23.08
Changelog: https://arm-software.github.io/ComputeLibrary/v23.08

cc @lhutton1 @neildhickey @ashutosh-arm for reviews

### DIFF
--- a/docker/install/ubuntu_download_arm_compute_lib_binaries.sh
+++ b/docker/install/ubuntu_download_arm_compute_lib_binaries.sh
@@ -28,7 +28,7 @@ if [ "$architecture_type" != "aarch64" ]; then
     gcc-aarch64-linux-gnu
 fi
 
-compute_lib_version="v23.05.1"
+compute_lib_version="v23.08"
 compute_lib_variant="arm64-v8a-neon"
 compute_lib_full_name="arm_compute-${compute_lib_version}-bin-linux-${compute_lib_variant}"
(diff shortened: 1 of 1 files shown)


lhutton1 (changes requested):
Thanks @leandron! The [change log](https://arm-software.github.io/ComputeLibrary/v23.08/versions_changelogs.xhtml#S2_2_changelog) states `libarm_compute_core` was deprecated and that we should no longer link against it. References such as the following should be removed: https://github.com/apache/tvm/blob/d486e21ef30990273e789af3ac1dee22f5bcb2d4/cmake/modules/contrib/ArmComputeLib.cmake#L62

leandron:
> Thanks @leandron! The [change log](https://arm-software.github.io/ComputeLibrary/v23.08/versions_changelogs.xhtml#S2_2_changelog) states `libarm_compute_core` was deprecated and that we should no longer link against it. References such as the following should be removed:
> 
> https://github.com/apache/tvm/blob/d486e21ef30990273e789af3ac1dee22f5bcb2d4/cmake/modules/contrib/ArmComputeLib.cmake#L62

I missed this change, but thinking about it, shouldn't we introduce the CMake change by the time we actually update CI to use the new image? I tested and TVM still will build fine in the meantime.

lhutton1:
Good point, let’s do that!

lhutton1 (approved):

PR #15991, opened 2023-10-26
[CI][ADRENO] Enhancements to Adreno specific CI utils

srkreddy1238:
RPC seem to be hanging while running bunch of tests with in single session. This enables running them individually.
Also we don't need rerun with in same session that leads to hang some times. Adding PLATFORM which is needed for run_pytest to summarize the failures.

### DIFF
--- a/tests/scripts/task_python_adreno.sh
+++ b/tests/scripts/task_python_adreno.sh
@@ -54,18 +54,33 @@ adb forward tcp:5002 tcp:5002
 env adb shell "cd ${TARGET_FOLDER}; killall -9 tvm_rpc-${USER}; sleep 2; LD_LIBRARY_PATH=${TARGET_FOLDER}/ ./tvm_rpc-${USER} server --host=0.0.0.0 --port=5000 --port-end=5010 --tracker=127.0.0.1:${TVM_TRACKER_PORT} --key=${RPC_DEVICE_KEY}" &
 DEVICE_PID=$!
 sleep 5 # Wait for the device connections
-trap "{ kill ${TRACKER_PID}; kill ${DEVICE_PID}; }" 0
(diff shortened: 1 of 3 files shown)


tests/scripts/task_python_adreno.sh
```diff
@@ -54,18 +54,33 @@ adb forward tcp:5002 tcp:5002
...
 sleep 5 # Wait for the device connections
-trap "{ kill ${TRACKER_PID}; kill ${DEVICE_PID}; }" 0
+trap "{ kill ${TRACKER_PID}; kill ${DEVICE_PID}; cleanup; }" 0
 
 # cleanup pycache
 find . -type f -path "*.pyc" | xargs rm -f
 # Test TVM
 make cython3
 
+# The RPC to remote Android device has issue of hang after few tests with in CI environments.
```
yongwww, line 64: Thanks for looking into this! This workaround should be good if it does not significantly increase the overall CI time. It would be great to identify the root cause of the hang issue as well.

yongwww:
cc: @driazati @leandron @lhutton1

tests/scripts/task_python_adreno.sh
srkreddy1238: True, individual runs increase the times significantly that too running on remote device.

I did a bit of a debug on this and it waits infinitely on socket read on host where as device has nothing to write.
Didn't perused further at the moment. I will definitely revisit the RPC issue a bit later.

srkreddy1238:
@yongwww Can u review and merge this ?  We have few follow up PR's on this.

The RPC issue I will investigate a little later.

yongwww (approved):

yongwww:
@srkreddy1238 it is landed now.

PR #15992, opened 2023-10-26
[Unity] Move VarBinding's and MatchCast's value to base class

Lunderberg:
Prior to this commit, the `VarBinding` and `MatchCast` had equivalent `Expr value` fields.  For use cases that need the bound value (e.g. collecting known values), this required downcasting to each subclass.  This commit moves the `Expr value` to the base `Binding` class, removing the need for the downcasting.

No impact to either C++ or Python usage is expected.  On the C++ side, all use of `VarBindingNode::value` or `MatchCastNode::value` will now access `BindingNode::value`.  On the Python side, all dynamic access of `binding_obj.value` will access the new field.

### DIFF
--- a/include/tvm/relax/expr.h
+++ b/include/tvm/relax/expr.h
@@ -722,6 +722,10 @@ class BindingNode : public Object {
  public:
   /*! \brief The return variable to bound to. */
   Var var;
+
+  /*! \brief The binding value. */
+  Expr value;
+
   mutable Span span;
 
   static constexpr const char* _type_key = "relax.expr.Binding";
@@ -751,8 +755,6 @@ class Binding : public ObjectRef {
  */
 class MatchCastNode : public BindingNode {
  public:
-  /*! \brief The input value to match cast. */
-  Expr value;
   /*! \brief The struct info pattern to match to. */
   StructInfo struct_info;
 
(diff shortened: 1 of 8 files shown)


Lunderberg:
@slyubomirsky I was thinking about the `GetBoundValue` utility function, and realized that there's enough locations that require the bound value that it would make sense to move the field from the `VarBindingNode` and `MatchCastNode` classes into the parent `BindingNode` class.

include/tvm/relax/expr.h
```diff
@@ -796,9 +798,6 @@ class MatchCast : public Binding {
 
 class VarBindingNode : public BindingNode {
  public:
-  /*! \brief The binding value. */
-  Expr value;
```
tqchen, line 800: We intentionally keep the value in the subclass, because in match clas there is extra set of struct info to be considered(that can populates new values), and it is important to keep reminding the users of this. 

The extra dispatch is worth overall for this explicitness

tqchen (changes requested):

include/tvm/relax/expr.h
Lunderberg: Good point, and closing this PR.

PR #15995, opened 2023-10-26
[Unity][Contrib] Add vLLM paged attention kernel

masahi:
This PR adds vLLM paged attention and update kernels as contrib. This can be used by MLC to enable batched inference, see https://github.com/mlc-ai/mlc-llm/pull/1134

The vLLM community has recently added the V2 kernel in https://github.com/vllm-project/vllm/pull/1348. This PR doesn't include the V2 change. Once the V2 kernel becomes mature, we can integrate it.

@tqchen @MasterJH5574 @yzh119 @yelite @sunggg


cmake/modules/CUDA.cmake
```diff
@@ -64,6 +64,7 @@ if(USE_CUDA)
     message(STATUS "Build with Thrust support")
     cmake_minimum_required(VERSION 3.13) # to compile CUDA code
     enable_language(CUDA)
+    set(CMAKE_CUDA_ARCHITECTURES "80;75")
```
yelite, line 67: Why do we want to explicitly set the cuda arch here?

cmake/modules/contrib/vllm.cmake
```diff
@@ -0,0 +1,25 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+if(USE_VLLM)
+  message(STATUS "Build with vllm paged attention kernel.")
+  include_directories(src/runtime/contrib/vllm)
+  enable_language(CUDA)
+  set(CMAKE_CUDA_ARCHITECTURES "80;75")
```
yelite, line 22: Should it let users set the cuda arch they want in `config.cmake` instead?

cmake/modules/CUDA.cmake
masahi: The is due to an odd behavior of cmake that I don't understand. If I remove this and set `USE_THRUST`, I get
```
CMake Error in CMakeLists.txt:
  CUDA_ARCHITECTURES is empty for target "tvm_runtime_objs".
```
. Previously I didn't encounter this problem when I was using `USE_THRUST` (a couple of years ago).

This doesn't apply to `vllm.cmake` for some reason. But there, if I don't have `set(CMAKE_CUDA_ARCHITECTURES "80;75")`, I get 
```
error   : Feature 'f16 arithemetic and compare instructions' requires .target sm_53 or higher
```
during build.

cmake/modules/contrib/vllm.cmake
masahi: Not all users (including me) use `config.cmake`. I think building for sm 80 and 75 should be fine in practice (newer arch can use sm80 code). We can add 70 if we still want to support v100.

cmake/modules/CUDA.cmake
cbalint13: @masahi 

CMAKE_CUDA_ARCHITECTURES is designed to be eprovided by user at cmake config time.
* There are also values like 'auto', 'all' see [CUDA_ARCHITECTURES](https://cmake.org/cmake/help/latest/prop_tgt/CUDA_ARCHITECTURES.html)
* Better throw a warning like: 'Using 'all' for CUDAARCHS, use CMAKE_CUDA_ARCHITECTURES to override'.
* Also it is initialized by env var [CUDAARCHS](https://cmake.org/cmake/help/latest/variable/CMAKE_CUDA_ARCHITECTURES.html) and picked up from there.

cmake/modules/contrib/vllm.cmake
cbalint13: masahi 

CMAKE_CUDA_ARCHITECTURES is designed to be eprovided by user at cmake config time.
* There are also values like 'auto', 'all' see [CUDA_ARCHITECTURES](https://cmake.org/cmake/help/latest/prop_tgt/CUDA_ARCHITECTURES.html)
* Better throw a warning like: 'Using 'all' for CUDAARCHS, use CMAKE_CUDA_ARCHITECTURES to override'.
* Also it is initialized by env var [CUDAARCHS](https://cmake.org/cmake/help/latest/variable/CMAKE_CUDA_ARCHITECTURES.html) and picked up from there.

starrkk:
@masahi  Thank you very much for sharing the open source cause, your pr is very meaningful to me, can you continue to integrate your code

PR #15999, opened 2023-10-27
[FFI] Use the correct function in DecRef 

yelite:
Found this problem when investigating for #15996 

cc @Lunderberg

### DIFF
--- a/src/runtime/registry.cc
+++ b/src/runtime/registry.cc
@@ -183,9 +183,9 @@ class EnvCAPIRegistry {
   }
 
   void DecRef(void* python_obj) {
-    ICHECK(py_inc_ref) << "Attempted to call Py_IncRef through EnvCAPIRegistry, "
-                       << "but Py_IncRef wasn't registered";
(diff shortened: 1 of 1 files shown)


Lunderberg (approved):
Oh my, that's a good catch, thank you!  This would effectively mean that any python exception thrown across the FFI border would never be garbage-collected, and that any local variables in its stack trace would be in memory until the program exits.  This is primarily used to improve debugging in cases where the exception is causing the program to exit anyways, but there's no guarantee that that's the case.

yelite:
The fix in this PR exposes an actual problem of making C++ wrapper of a Python object. Running `pytest tests/python/unittest/test_crt.py` will result in error "the function must be called with the GIL held, but the GIL is released (the current Python thread state is NULL)".

In addition, I think we should also consider the order of shutdown here. After the Python runtime shuts down, DecRef can no longer be called.

@Lunderberg Since this goes far beyond my initial intention of fixing a typo, can you take this over and make proper fix?

PR #16001, opened 2023-10-27
[ONNX] Fix interpreting auto_pad parameters in ConvTranspose operator

padreofthegame:
Fix in interpreting auto_pad parameters SAME_UPPER and SAME_LOWER in ConvTranspose version 1 and 11 operator according to documentation on [link](https://onnx.ai/onnx/operators/onnx__ConvTranspose.html).

Tests that were failing are also added in test_forward.py.

Also, enable unsupported test for ONNX: test_convtranspose_autopad_same.

### DIFF
--- a/python/tvm/relay/frontend/onnx.py
+++ b/python/tvm/relay/frontend/onnx.py
@@ -826,6 +826,15 @@ class Conv(OnnxOpConverter):
         return out
 
 
+def is_ort_version_greater_than(ver):
+    import onnxruntime as ort
+
+    v11, v12, v13 = tuple(int(v) for v in ort.__version__.split("."))
+    v21, v22, v23 = tuple(int(v) for v in ver.split("."))
+
+    return (v11 > v21) or (v11 == v21 and v12 > v22) or ((v11, v12) == (v21, v22) and v13 > v23)
+
+
 class ConvTranspose(OnnxOpConverter):
     """Operator converter for ConvTranspose."""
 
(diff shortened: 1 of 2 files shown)


padreofthegame:
@tvm-bot rerun

yongwww (approved):

padreofthegame:
Hello, 

looks like that the problem with this PRs failure is due to the different implementations of convTranspose operator in different versions of onnxruntime. Precissely, onnxruntime 1.12 and 1.15 gives different results when the parameters SAME_UPPER and SAME_LOWER are used. On my local machine, this tests passes when onnxruntime 1.15 is used, and fail with 1.12 (probably due to wrong implementation), although the operator is implemented according to ONNX documentation.

What could be the next step for this PR? What is the common practice here:
	- to skip this test, as it was done with test_convtranspose_autopad_same?
	- to mark this test as flaky?
	- to add part of the code to check onnxruntime version before test running?

Sorry if I tagged someone not related to the topic.

@masahi @Hzfengsy @jikechao

python/tvm/relay/frontend/onnx.py
```diff
@@ -826,6 +826,15 @@ def _impl_v1(cls, inputs, attr, params):
         return out
 
 
+def is_ort_version_greater_than(ver):
+    import onnxruntime as ort
+
+    v11, v12, v13 = tuple(int(v) for v in ort.__version__.split("."))
```
yongwww, line 832: probably you can try to use get_converter here

padreofthegame: What do you mean exactly?

Frankly speaking, at the end, probably there should be no need for this function at all, due to the fact that for opset 11 of convTranspose operator there is only one valid implementation.
On the other hand, that implementation should be independent from the version of ONNXRuntime version.

But, due to the change in the interpretation of autopad parameters in ONNXRuntime versions 1.12.1 and 1.13.1,the same models produce different results in different ONNXRuntime versions, and since there are tests in TVM that compares TVM results with runtime results, I added this additional check in order to make that TVM results be equal to ONNXRuntime in both cases. 

There is even a warning when trying to run onnx model with convTranspoose operator with version 1.12.1
`[W:onnxruntime:Default, conv_transpose.h:32 ConvTranspose] The existing bug in the padding distribution for auto_pad type SAME_UPPER/SAME_LOWER will be fixed in next ORT 1.13 release and hence the results of ConvTranspose operator using the above auto_pad type(s) will be different.`

padreofthegame:
Hello,

looks like this PR is staying opened this way for quite some time now. Any feedback related to this PR?

padreofthegame:
Hello,

Any updates on this one? Can we get this merged as it has been approved for quite some time now, and is passing all the tests?

yongwww:
@padreofthegame  it was merged! Thanks for the fix!

PR #16007, opened 2023-10-29
[Unity][nn.Module] Support Parameter Packing

junrushao:
This PR introduces support for parameter/effect packing.

**What is parameter packing?** There are two ways to represent weights
as input(s) of Relax functions.

- Unpacked: each weight is passed in as a separate parameter:

```python
@R.function
def f(
  ...,
  weight_0: R.Tensor(...),
  weight_1: R.Tensor(...),
  ...,
)
```
- Packed: all weights are packed together as a Relax Tuple and passed in
  as a single parameter:

```python
@R.function
def f(
  ...,
  weights: R.Tuple((R.Tensor(...), R.Tensor(...))),
  ...,
)
```

**Why support parameter packing?** First of all, currently MLC LLM uses
packed parameter as the default format in its runtime. To align the new
`nn.Module` frontend with the existing behavior, we allow packing as a
non-default option to be turned on.

Second, having a calling convention where the last argument being
packed parameters and the second last being packed effects could help
simplify many runtime behavior, for example, instructing the generated
code to skip type checking, etc.

**How to enable parameter/effect packing?** We allow an extra argument
`$` when creating `nn.spec.MethodSpec`, which is effectively
fine-grained configuration of exported Relax function:

```python
model.export_tvm(
    spec={
        "forward": {
            "x": nn.spec.Tensor([1, model.in_features], "float32"),
            "$": {
                "param_mode": "packed", // Configure parameter packing
                "effect_mode": "none", // Configure effect packing
            },
        }
    }
)
```

### DIFF
--- a/python/tvm/relax/frontend/nn/core.py
+++ b/python/tvm/relax/frontend/nn/core.py
@@ -52,6 +52,8 @@ from ._tensor_op import _TensorOp
 from .subroutine import SubroutineMixin
 
 if TYPE_CHECKING:
+    import torch
+
     from . import spec as _spec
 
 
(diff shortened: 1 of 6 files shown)


cyx-6 (approved):
Looks perfect!

junrushao:
Should be useful to you guys as well @jwfromm @masahi

PR #16008, opened 2023-10-29
[Unity] Allow Pipeline Registration

junrushao:
This commit adds `tvm.relax.register_pipeline` which allows external registration of compilation pipelines.

### DIFF
--- a/python/tvm/relax/pipeline.py
+++ b/python/tvm/relax/pipeline.py
@@ -23,6 +23,7 @@ as it is or serves as a basis to do further composition.
 # pylint: disable=unused-argument
 import tvm
 from tvm import meta_schedule as ms
+
 from . import transform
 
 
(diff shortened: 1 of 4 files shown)


python/tvm/relax/pipeline.py
```diff
@@ -93,9 +94,26 @@ def get_pipeline(name: str = "zero", **kwargs) -> tvm.transform.Pass:
...
-        return PIPELINE_MAP[name](**kwargs)
-    else:
+    if name not in PIPELINE_MAP:
         raise ValueError(
             f"Unknown pre-built pipeline {name}," f"candidates are {list(PIPELINE_MAP.keys())}"
         )
+    return PIPELINE_MAP[name](**kwargs)
+
+
+def register_pipeline(name: str):
```
Hzfengsy, line 104: return annotation plz

Hzfengsy (approved):

python/tvm/relax/pipeline.py
junrushao: it's currently used as a [decorator](https://github.com/mlc-ai/mlc-llm/blob/main/python/mlc_chat/compiler/compiler_pass/pipeline.py#L14), so I suppose we don't need a type annotation by default? I'm happy to add one in a follow-up PR if you think this is the case

PR #16011, opened 2023-10-30
[Unity][Training] Support intermediate vars as require_grads for Gradient pass

Ubospica:
This PR supports providing intermediate vars in require_grads argument of the gradient pass.

This is useful for debugging, since users can find the gradient value of intermediate vars.

This PR also fixes a bug when var that requires gradient is wrapped in the checkpointed range (wrapped in start_checkpoint and end_checkpoint pairs). 

cc @yzh119 @MasterJH5574 @tqchen

### DIFF
--- a/tests/python/relax/test_transform_gradient_checkpoint.py
+++ b/tests/python/relax/test_transform_gradient_checkpoint.py
@@ -16,12 +16,13 @@
 # under the License.
 """Unit tests for gradient with checkpointing."""
 import tvm
-from tvm.relax.block_builder import BlockBuilder
-from tvm.relax.testing.nn import checkpoint, emit_checkpoint, emit_checkpoint_sequential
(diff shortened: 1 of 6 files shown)


tests/python/relax/test_transform_gradient_checkpoint.py
```diff
@@ -16,12 +16,13 @@
 # under the License.
 """Unit tests for gradient with checkpointing."""
 import tvm
-from tvm.relax.block_builder import BlockBuilder
```
Ubospica, line 19: These changes are only style enhancement (reordering imports, deleting redundant "dtype=" annotation). I only add two tests for it (test_checkpoint_dag and test_checkpoint_with_intermediate_require_grads). See the end of this file.

MasterJH5574 (approved):

PR #16014, opened 2023-10-30
[Unity] [Bugfix] Fix argument: 'None' for adaptive_avg_pool2d in torch frontend 

yifan660:
Pytorch AdaptiveAvgPool2d can take None as argument for output size. It will output tensor with same size as input tensor
https://pytorch.org/docs/stable/generated/torch.nn.AdaptiveAvgPool2d.html
This PR adds check for None argument. When detected it converts tensor output size to same as input size.


python/tvm/relax/frontend/torch/fx_translator.py
```diff
@@ -878,7 +879,7 @@ def _avg_pool2d(self, node: fx.node.Node) -> relax.Var:
                 ceil_mode = node.kwargs["ceil_mode"]
             else:
                 ceil_mode = False
-
+
```
Hzfengsy, line 882: style

Hzfengsy (approved):

PR #16017, opened 2023-10-30
[Tests] Fix str vs. int comparison in test_num_threads

PhilippvK:
If either `TVM_NUM_THREADS` or `OMP_NUM_THREADS` the `test_num_threads` would fail because strings can not be compared with integers.

This tiny PR fixes the test by introducing type casts.


quic-sanirudh (approved):
LGTM, Thanks

PR #16018, opened 2023-10-30
[Tests] Fix work_dir location used by test_micro_tuning_with_meta_schedule

PhilippvK:
Converting `work_dir = tvm.contrib.utils.tempdir()` to `str` yields `"<tvm.contrib.utils.TempDirectory object at 0x7fc5322c3790>/"` which creates a so-called file in the current working directory.

This change uses `str(work_dir.path)` instead, resulting in the expected behavior.


quic-sanirudh (approved):
LGTM, Thanks.

PR #16021, opened 2023-10-31
[Bugfix][FFI] Typo fix of IncRef to DecRef

Lunderberg:
Propagation of Python exceptions across C++ stack frames was introduced in https://github.com/apache/tvm/pull/15596.  This commit primarily fixes a typo in the initial implementation, where `Py_IncRef` was used instead of `Py_DecRef`.

In addition, this PR resolves errors that were exposed by this typo fix, which caused test failures in `tests/python/unittest/test_crt.py::test_compile_runtime`.  These were due to use of the `Py_IncRef` and `Py_DecRef` functions on threads that hadn't acquired the GIL.  This usage error has been corrected for both the ctypes and cython FFI handling.


junrushao (approved):
LGTM!

Issue #11414, opened 2022-05-23
[Tracking Issue] pylint the tests!


areusch:
This is a tracking issue for our effort to enable pylint on all Python testing code underneath `tests/python`. Currently we don't lint scripts in that directory. We're going to take an incremental approach here and anyone is welcomed to help with this. Here's how:

1. Pick a directory where to start. I suggest choosing the directory based on the number of Python files.
2. Add a line to `tests/lint/pylint.sh` containing that directory similar to https://github.com/apache/tvm/pull/11625. 
3. Run `docker/lint.sh pylint` locally and fix the lint issues.
4. Submit the PR and reference this issue in the PR so we know to update it.

Here are the tests that need linting 😸 (we will try to tag the PR which fixes them when we merge them):

- [ ] tests/python/all-platform-minimal-test/test_minimal_target_codegen_llvm.py - https://github.com/apache/tvm/pull/11939
- [ ] tests/python/all-platform-minimal-test/test_runtime_ndarray.py - https://github.com/apache/tvm/pull/11939
- [ ] tests/python/all-platform-minimal-test/test_runtime_packed_func.py - https://github.com/apache/tvm/pull/11939
- [x] tests/python/ci/test_ci.py - https://github.com/apache/tvm/pull/11666
- [x] tests/python/ci/test_mergebot.py - https://github.com/apache/tvm/pull/11666
- [x] tests/python/ci/test_script_converter.py - https://github.com/apache/tvm/pull/11666
- [x] tests/python/ci/test_utils.py - https://github.com/apache/tvm/pull/11666
- [ ] tests/python/conftest.py 
- [ ] tests/python/contrib/test_arm_compute_lib/infrastructure.py
- [ ] tests/python/contrib/test_arm_compute_lib/test_add.py
- [ ] tests/python/contrib/test_arm_compute_lib/test_concatenate.py
- [ ] tests/python/contrib/test_arm_compute_lib/test_conv2d.py
- [ ] tests/python/contrib/test_arm_compute_lib/test_dense.py
- [ ] tests/python/contrib/test_arm_compute_lib/test_maximum.py
- [ ] tests/python/contrib/test_arm_compute_lib/test_network.py
- [ ] tests/python/contrib/test_arm_compute_lib/test_pooling.py
- [ ] tests/python/contrib/test_arm_compute_lib/test_reshape.py
- [ ] tests/python/contrib/test_arm_compute_lib/test_runtime.py
[...]

PR #16023, opened 2023-10-31
[Pylint] fix pylint issues for thrust&tflite_runtime&util

tlopex:
See #11414

### DIFF
--- a/tests/python/contrib/test_tflite_runtime.py
+++ b/tests/python/contrib/test_tflite_runtime.py
@@ -14,47 +14,49 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+"""Configure pytest"""
+import importlib
 import pytest
-
-import tvm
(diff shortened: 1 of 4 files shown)


tests/python/contrib/test_tflite_runtime.py
```diff
@@ -14,47 +14,49 @@
...
-        return
+        return None
     if not tvm.get_global_func("tvm.tflite_runtime.create", True):
         print("skip because tflite runtime is not enabled...")
-        return
+        return None
 
     try:
-        import tensorflow as tf
+        tf_lib = importlib.import_module("tensorflow")
```
Hzfengsy, line 36: It's better to use `# pylint: disable=import-outside-toplevel` in this file, rather than importlib

tlopex: Got it. I will alter them.

Hzfengsy (approved):

PR #16025, opened 2023-10-31
[TIR] Use known dtype for integer arguments of Range(0, expr)

Lunderberg:
Prior to this commit, `Range(0, expr)` would automatically convert `0` to `IntImm(DataType::Int(32), 0)` using the automatic conversion of `int` to `PrimExpr`.  This can result in ranges where the bounds are of different dtypes.  Since this is a common idiom (25 uses of `Range::FromMinExtent(0, expr)` and 49 uses of `Range(0, expr)`), it should be supported rather than forbidden.

This commit adds overloads that take `(int, PrimExpr)` arguments for `Range` and `Range::FromMinExtent`.  These overloads convert the first argument to an `IntImm` of the same dtype as the second, and then delegate to the `(PrimExpr, PrimExpr)` overload.


Lunderberg:
This PR is currently a draft, in order to see if any test cases rely on the mixed dtype behavior.

PR #16026, opened 2023-10-31
[Unity] Avoid Emitting Redandunt Bindings in TensorExpr Op

junrushao:
See the testcase for more details :)

### DIFF
--- a/python/tvm/relax/frontend/nn/op.py
+++ b/python/tvm/relax/frontend/nn/op.py
@@ -33,7 +33,7 @@ from .spec import SpecBuilder
 IntExpr = Union[int, _tir.PrimExpr]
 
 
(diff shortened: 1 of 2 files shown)


python/tvm/relax/frontend/nn/op.py
```diff
@@ -50,7 +50,8 @@ def _wrap_nested(expr: rx.Expr, name: str) -> Union[Tensor, Tuple[Tensor, ...]]:
     result : Union[Tensor, Tuple[Tensor]]
         The computed result.
     """
-    expr = BlockBuilder.current().emit(expr, name)
+    if not isinstance(expr, rx.DataflowVar):
```
tqchen, line 53: i believe you can just do https://github.com/apache/tvm/blob/unity/python/tvm/relax/block_builder.py#L632

tqchen (approved):

PR #16033, opened 2023-11-01
[Codegen][Metal] Disable cross-function call in Metal codegen

MasterJH5574:
This PR restores the Metal codegen to the one before #15835. Due to there will likely be no internal function call in Metal, we think it is safe to do so.

Verified that with this PR, the metal codegen and iPhone codegen will not fail and will work properly.

The reason of the iPhone codegen failure is because the multiple declarations of a same function will lead to multiple emissions of a same structs, which is not recognizable by the metal compiler.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -73,6 +73,7 @@ jobs:
         shell: bash -l {0}
         run: >-
           python -m pytest -v -s 'tests/python/unittest/test_allreduce.py::test_allreduce_sum_compile'
+          python -m pytest -v -s 'tests/python/unittest/test_target_codegen_metal.py::test_func_with_trailing_pod_params'
       - name: Minimal Metal Compile-and-Run
         shell: bash -l {0}
         run: >-
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -71,7 +71,7 @@ jobs:
           python -m pytest -v tests/python/all-platform-minimal-test
       - name: Minimal Metal Compile-Only
         shell: bash -l {0}
-        run: >-
+        run: |
           python -m pytest -v -s 'tests/python/unittest/test_allreduce.py::test_allreduce_sum_compile'
           python -m pytest -v -s 'tests/python/unittest/test_target_codegen_metal.py::test_func_with_trailing_pod_params'
       - name: Minimal Metal Compile-and-Run
--- a/src/target/source/codegen_metal.cc
+++ b/src/target/source/codegen_metal.cc
@@ -39,8 +39,6 @@ namespace codegen {
 
 void CodeGenMetal::InitFuncState(const PrimFunc& f) {
   CodeGenC::InitFuncState(f);
-  // skip the first underscore, so SSA variable starts from _1
-  name_supply_->FreshName("v_");
   // analyze the data;
   for (Var arg : f->params) {
     if (arg.dtype().is_handle()) {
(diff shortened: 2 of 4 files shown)


MasterJH5574:
cc @Lunderberg @tqchen

junrushao:
CC @Lunderberg would you mind taking a look? BTW, is there a way to unit-test it as you enabled a while ago?

tqchen:
Thanks @MasterJH5574 , can you also document the rationale of that we specialize this part. aka we do not have multiple cross function calls, so we can keep the delcaration once to keep things simple

MasterJH5574:
> BTW, is there a way to unit-test it as you enabled a while ago?

Reading the failure reason elaborated here https://github.com/apache/tvm/pull/15835#issuecomment-1789308138, I feel the way to reproduce or test the issue is to
* construct an IRModule where one PrimFunc _containing trailing int/float parameters_ is called at least twice,
* build the IRModule to `iphone` target.

Lunderberg:
If possible, I'd like to fix the functionality and not just revert it.  https://github.com/apache/tvm/pull/14985 and https://github.com/apache/tvm/pull/14862 have been on the back burner for quite a while, mainly due to breakage they cause in `tvm.contrib.ethosu`, but are still changes I'd like to get around to landing at some point.

@MasterJH5574 We should be able to test it in CI, since I added compile-only testing of the metal backend in #15756.  Functionality that isn't tested in CI is inherently unstable, and so CI testing is preferable in addition to local testing.

Lunderberg:
> BTW, is there a way to unit-test it as you enabled a while ago?

The two steps to add a Metal specific unit test are to (1) decorate the test with `@tvm.testing.requires_metal(support_required="compile-only")` and (2) add the test in the `Minimal Metal Compile-Only` section of `.github/workflows/main.yml`.  This only validates the compiler support, as we don't have a CI environment that has Metal GPUs, but that should catch the majority of codegen failure modes.

MasterJH5574:
@Lunderberg I agree that it would be better if the issue can be fixed! For this PR, the overall codegen still bases on the updated codegen flow (we did not revert the previous PR). There is no inter function calls between kernels in Metal, so it is okay to keep the former flow in Metal codegen for now as a swift fix to the codegen issue.

And thanks for the pointer on testing. I will add a test soon.

Lunderberg:
There aren't currently any inter-function calls in Metal, but it is part of my planned implementation for https://github.com/apache/tvm/pull/14985.  Breaking a change apart into the underlying infrastructure changes and the usage of that infrastructure allows the changes to be tested and integrated separately.

Lunderberg:
That said, I have an idea for how to work around the lack of inter-function calls in codegen.  Since we can represent inter-function calls in TIR, we could add a `InlineLocalFunctions` pass that removes any internal private functions that result from the single-module lowering flow.

tqchen:
I like the idea of keeping things simple on specific shader backend like metal, inlining sounds like a good approach

MasterJH5574:
Finished adding a test and included it in “minimal Metal compile-only.” Also added a note in `AddFunction` in codegen_metal.cc

Lunderberg (changes requested):
Given the special packing used for the scalar buffers, and that there's a possibility of a workaround, I think these sections makes sense to revert.

In the `void CodeGenMetal::VisitExpr_(const CallNode* op) override`, can we add an explicit check for the inter-function call?  Otherwise, this falls back to the `CodegenC` default that support it.

```c++
CHECK(!op->op.as<GlobalVarNode>())
    << "CodegenMetal does not support inter-function calls, "
    << "but expression " << GetRef<Expr>(op)
    << " calls PrimFunc " << op->op;
```

MasterJH5574:
Thanks @Lunderberg for the suggestions! Added the check in CallNode visit function.

Lunderberg (approved):
Thank you, and looks good to me!  The unit test looks really good; I especially like the explanation of the known portions of the failure mode.

Lunderberg:
Circling back to this, I have PR #16184 submitted, which adds a TIR transform that can inline private function calls.  That way, the user-facing functionality of TIR subroutines can be used, without requiring every codegen to natively support subroutines.

PR #16034, opened 2023-11-01
[Packaging] Include BYOC dynamic libraries into wheel

sunggg:
Current FasterTransformer and FlashAttention BYOCs produce their own shared libraries.
This PR adds them in the list of shared libraries that is going to be packaged into the python wheel.


sunggg:
@tvm-bot rerun

tqchen (approved):

PR #16039, opened 2023-11-02
[Target] Support obtain l2 cache size from target

Ubospica:
This PR supports obtain l2 cache size from target. The API is
```
target.l2_cache_size_bytes
```

The l2 cache size is hardcoded in src/target/tag.cc.

This is useful for specializing schedule rules for different targets.

This PR is followup of #15332

cc @yzh119 @zxybazh

### DIFF
--- a/include/tvm/target/tag.h
+++ b/include/tvm/target/tag.h
@@ -109,7 +109,7 @@ class TargetTagRegEntry {
    * \param key The attribute name
    * \param value The attribute value
    */
-  inline TargetTagRegEntry& add_config(String key, ObjectRef value);
+  inline TargetTagRegEntry& with_config(String key, ObjectRef value);
   /*! \brief Set name of the TargetTag to be the same as registry if it is empty */
   inline TargetTagRegEntry& set_name();
   /*!
@@ -137,7 +137,7 @@ inline TargetTagRegEntry& TargetTagRegEntry::set_config(Map<String, ObjectRef> c
   return *this;
 }
 
-inline TargetTagRegEntry& TargetTagRegEntry::add_config(String key, ObjectRef value) {
+inline TargetTagRegEntry& TargetTagRegEntry::with_config(String key, ObjectRef value) {
   tag_->config.Set(key, value);
   return *this;
 }
--- a/include/tvm/target/tag.h
+++ b/include/tvm/target/tag.h
@@ -104,6 +104,12 @@ class TargetTagRegEntry {
    * \param config The config dict for target creation
    */
   inline TargetTagRegEntry& set_config(Map<String, ObjectRef> config);
+  /*!
+   * \brief Add a key-value pair to the config dict
+   * \param key The attribute name
+   * \param value The attribute value
+   */
+  inline TargetTagRegEntry& add_config(String key, ObjectRef value);
   /*! \brief Set name of the TargetTag to be the same as registry if it is empty */
   inline TargetTagRegEntry& set_name();
   /*!
@@ -128,6 +134,11 @@ class TargetTagRegEntry {
 
 inline TargetTagRegEntry& TargetTagRegEntry::set_config(Map<String, ObjectRef> config) {
   tag_->config = std::move(config);
+  return *this;
+}
+
+inline TargetTagRegEntry& TargetTagRegEntry::add_config(String key, ObjectRef value) {
+  tag_->config.Set(key, value);
   return *this;
 }
 
--- a/src/target/target_kind.cc
+++ b/src/target/target_kind.cc
@@ -334,6 +334,7 @@ TVM_REGISTER_TARGET_KIND("cuda", kDLCUDA)
     .add_attr_option<Integer>("max_threads_per_block")
     .add_attr_option<Integer>("thread_warp_size", Integer(32))
     .add_attr_option<Integer>("registers_per_block")
+    .add_attr_option<Integer>("l2_cache_size_bytes")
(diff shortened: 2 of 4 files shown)


include/tvm/target/tag.h
```diff
@@ -131,6 +137,11 @@ inline TargetTagRegEntry& TargetTagRegEntry::set_config(Map<String, ObjectRef> c
   return *this;
 }
 
+inline TargetTagRegEntry& TargetTagRegEntry::add_config(String key, ObjectRef value) {
```
zxybazh, line 140: I think this is pretty much the same as `set_config` where you can just do
```c++
target.set_config({key, value});
```
Another example can be found  here: https://github.com/apache/tvm/blob/c3ce474dc172c521d7057061fac3b21207e83f0d/src/target/tag.cc#L73-L88

zxybazh (commented):
Thanks a lot for sending this PR! Adding new target attribute makes sense to me as it could be beneficial to all users of this target tag. IMHO we don't need to introduce new function to support `add_config` as it's pretty much the same as `set_config`. Other than this LGTM.

CC: @junrushao

include/tvm/target/tag.h
Ubospica: I see. But the problem here is using set_config will **substitute** the whole config map instread of adding new items. And for macro TVM_REGISTER_CUDA_TAG, a config is already added. 

So maybe we need to find a method to add an additional config?

Ubospica: > I think this is pretty much the same as `set_config` where you can just do
> 
> ```c++
> target.set_config({key, value});
> ```
> 
> Another example can be found here:
> 
> https://github.com/apache/tvm/blob/c3ce474dc172c521d7057061fac3b21207e83f0d/src/target/tag.cc#L73-L88

Thanks for the comment. But the problem here is using set_config will **substitute** the whole config map instread of adding new items. And for macro TVM_REGISTER_CUDA_TAG, a config is already added. 

So maybe we need to find a method to add an additional config?

zxybazh: Thanks for the feedback, it makes sense to me.

include/tvm/target/tag.h
```diff
@@ -104,6 +104,12 @@ class TargetTagRegEntry {
    * \param config The config dict for target creation
    */
   inline TargetTagRegEntry& set_config(Map<String, ObjectRef> config);
+  /*!
+   * \brief Add a key-value pair to the config dict
+   * \param key The attribute name
+   * \param value The attribute value
+   */
+  inline TargetTagRegEntry& add_config(String key, ObjectRef value);
```
zxybazh, line 112: `Add_config` seems to assume the key was not originally in the configuration, what about`with_config` which isn't related to whether the original config has the given key.

src/target/target_kind.cc
```diff
@@ -334,6 +334,7 @@ TVM_REGISTER_TARGET_KIND("cuda", kDLCUDA)
     .add_attr_option<Integer>("max_threads_per_block")
     .add_attr_option<Integer>("thread_warp_size", Integer(32))
     .add_attr_option<Integer>("registers_per_block")
+    .add_attr_option<Integer>("l2_cache_size_bytes")
```
zxybazh, line 337: Nit, naming convention. Previous attributes like `max_shared_memory_per_block` is also using bytes, what about we stick to the same style and use `l2_cache_size`? Please feel free to add a comment to clarify it's in bytes though.

include/tvm/target/tag.h
```diff
@@ -131,6 +137,11 @@ inline TargetTagRegEntry& TargetTagRegEntry::set_config(Map<String, ObjectRef> c
   return *this;
 }
 
+inline TargetTagRegEntry& TargetTagRegEntry::add_config(String key, ObjectRef value) {
+  tag_->config.Set(key, value);
```
zxybazh, line 141: Would you please also add the case when the key is already set?

zxybazh (commented):
Thanks again for the feedback, I did anther round of review and please feel free to reply in the thread!

include/tvm/target/tag.h
Ubospica: Make sense. Will change the name into with_config.

src/target/target_kind.cc
Ubospica: Yeah I agree with that. It's only I am following the previously defined(but not implemented attribute) attribute in target, see https://github.com/apache/tvm/blob/main/python/tvm/target/target.py#L272

And the same attribute is defined and implemented for devices. See https://github.com/apache/tvm/blob/main/python/tvm/_ffi/runtime_ctypes.py#L492

The difference is that for targets, we will hardcoded this attibute (b/c we may build a model on devices different from the provided target), and for devices, we will automatically detect it from real devices.

Ubospica: See #15332

zxybazh: Hi, as I mentioend in the beginning of this commen, I was just nit picking and I don't have strong opinion if you prefer to keep it. On the other hand, I don't think any naming for this would impact your cross compilation usage, just make sure you have consistent naming with Zihao's PR should be good.

include/tvm/target/tag.h
Ubospica: I think it is ok given the name is with_config. `Set` will work whether the key is already set.

src/target/target_kind.cc
Ubospica: I see, okay

zxybazh (approved):
LGTM.

Ubospica:
@zxybazh Thanks for the review

PR #16040, opened 2023-11-02
[Arith] Support Analyzer.int_set(expr) in python and add RewriteSimplifier rules

Ubospica:
This PR:
- Supports `Analyzer.int_set(expr)` from python side. That is, exposing this API from C++ side. Previously we only support `Analyzer.int_set(expr, dom_map)`. 
- Adds some RewriteSimplifier rules for simplifying certain expressions:
    - `min/max(x +- y, x)` where `y>=0` or `y<0`
    - Supporting symbolic analysis in comparision expressions. That enables simplifying `x // 2 < y where x < y * 2` to `True`

cc @spectrometerHBH @tqchen

### DIFF
--- a/python/tvm/arith/analyzer.py
+++ b/python/tvm/arith/analyzer.py
@@ -195,7 +195,7 @@ class Analyzer:
         """
         return self._canonical_simplify(expr)
 
-    def int_set(self, expr, dom_map):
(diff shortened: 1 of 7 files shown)


Ubospica:
Needs more consideration on symbolic analysis

PR #16046, opened 2023-11-02
[Fix] Update mutator name rule

LeshengJin:
When the input `name` is an empty string which means the visitor is at the top level of the module, we should not add a dot to the name of the leaf nodes. 
Before:
```
LlamaForCausalLM
    - .model
    - .lm_head
```
After this update:
```
LlamaForCausalLM
    - model
    - lm_head
```

### DIFF
--- a/python/tvm/relax/frontend/nn/visitor.py
+++ b/python/tvm/relax/frontend/nn/visitor.py
@@ -15,7 +15,8 @@
 # specific language governing permissions and limitations
 # under the License.
 """The visitor and mutator infra for nn.Module."""
-from typing import Any
+from typing import Any, Union
+
 from . import core as nn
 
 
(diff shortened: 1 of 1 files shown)


python/tvm/relax/frontend/nn/visitor.py
```diff
@@ -112,6 +113,14 @@ def visit(self, name: str, node: Any) -> Any:
...
+        def _leaf_name(parent: str, child: str) -> str:
+            if parent == "":
+                # in the top level of the module
+                return child
+            else:
+                return f"{parent}.{child}"
+
         if isinstance(node, nn.ModuleList):
             for i in range(len(node)):
                 if isinstance(node[i], nn.ModuleList):
```
zxybazh, line 126: Shall we also change the following parts?

LeshengJin: It's a good question. I assume the top level is nn.Module, not nn.ModuleList. Hence, when visiting nn.ModuleList, the name should not be left as an empty string. @cyx-6 @junrushao

junrushao: I think it's a reasonable design

PR #16049, opened 2023-11-02
[Unity] [LiftTransformParams] Treat symbolic var in weight shape as constant

jinhongyii:
In some cases the shape of model weight contains a symbolic var, like vocab size, but it should be treated as if it's constant in LiftTransformParams because the symbolic var does not change in each call of decode/prefill functions.

This PR implements this behavior.

### DIFF
--- a/tests/python/relax/test_transform_lift_transform_params.py
+++ b/tests/python/relax/test_transform_lift_transform_params.py
@@ -642,5 +642,95 @@ def test_symbolic_var_from_shape():
     tvm.ir.assert_structural_equal(Expected, after)
 
 
+def test_symbolic_var_in_param_shape():
+    @tvm.script.ir_module
+    class Before:
+        @R.function
+        def main(
+            x: R.Tensor((1, 16, 224, "n"), "float32"),
+            w1: R.Tensor((16, "m", 3, 3), "float32"),
+            w2: R.Tensor((16, "m", 3, 3), "float32"),
+        ) -> R.Tensor((1, 16, 224, 224), "float32"):
+            m = T.int64()
+            n = T.int64()
+            R.func_attr({"num_input": 1})
+            with R.dataflow():
+                zeros = R.zeros((n, n), "float32")
+                w1 = R.add(w1, R.const(1, "float32"))
+                conv1 = R.nn.conv2d(x, w1, padding=(1, 1), data_layout="NCHW", kernel_layout="OIHW")
+                conv2 = R.nn.conv2d(
+                    conv1, w2, padding=(1, 1), data_layout="NCHW", kernel_layout="OIHW"
+                )
+                R.output(conv2)
+            return conv2
+
+    @I.ir_module
+    class Expected:
+        @R.function
+        def main_transform_params(
+            params: R.Tuple(
+                R.Tensor((16, "m", 3, 3), dtype="float32"),
+                R.Tensor((16, "m", 3, 3), dtype="float32"),
+            )
+        ) -> R.Tuple(
+            R.Tensor((16, "m", 3, 3), dtype="float32"), R.Tensor((16, "m", 3, 3), dtype="float32")
+        ):
+            m = T.int64()
+            with R.dataflow():
+                lv: R.Tensor((16, m, 3, 3), dtype="float32") = params[1]
+                lv1: R.Tensor((16, m, 3, 3), dtype="float32") = params[0]
+                lv2: R.Tensor((16, m, 3, 3), dtype="float32") = R.add(lv1, R.const(1, "float32"))
+                gv: R.Tuple(
+                    R.Tensor((16, m, 3, 3), dtype="float32"),
+                    R.Tensor((16, m, 3, 3), dtype="float32"),
+                ) = (lv, lv2)
+                R.output(gv)
+            return gv
+
+        @R.function
+        def main(
+            x: R.Tensor((1, 16, 224, "n"), dtype="float32"),
+            transformed_param_0: R.Tensor((16, "m", 3, 3), dtype="float32"),
+            transformed_param_1: R.Tensor((16, "m", 3, 3), dtype="float32"),
+        ) -> R.Tensor((1, 16, 224, 224), dtype="float32"):
+            n = T.int64()
+            m = T.int64()
+            R.func_attr({"num_input": 1})
+            with R.dataflow():
+                zeros: R.Tensor((n, n), dtype="float32") = R.zeros(R.shape([n, n]), dtype="float32")
+                lv: R.Tensor((16, m, 3, 3), dtype="float32") = transformed_param_1
+                conv1: R.Tensor((1, 16, 224, n), dtype="float32") = R.nn.conv2d(
+                    x,
+                    lv,
+                    strides=[1, 1],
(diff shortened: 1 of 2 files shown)


jinhongyii:
cc: @Lunderberg @vinx13

tests/python/relax/test_transform_lift_transform_params.py
```diff
@@ -642,5 +642,95 @@ def slice(
...
+                    out_layout="NCHW",
+                    out_dtype="void",
+                )
+                R.output(conv2)
+            return conv2
+
+    mod = Before
+    after = relax.transform.LiftTransformParams()(mod)
+    tvm.ir.assert_structural_equal(after, Expected)
+
```
Lunderberg, line 733: Can we add a unit test for the case where the weights require symbolic variables that are used, but not definable, from the shapes of the weights?  This would require the `DefinableTIRVarsInStructInfo` usage described earlier.  The following input would be able to trigger this case.

```python
def test_symbolic_var_defined_in_params_but_used_in_weights():
    """A symbolic variable's occurrence in the weights may not define it

    In order to be a source of definition, a symbolic variable in the
    parameters must occur as a distinct parameter, as a tensor shape
    `R.Tensor(["var"])`, an explicit `R.Shape(["var"])`, or as a
    `R.Prim(value="var")`.  A variable that is part of a larger
    expression, such as `R.Tensor(["m * n"])`, are variable usages,
    not variable definitions.
    """

    @tvm.script.ir_module
    class Before:
        @R.function
        def main(
            x: R.Tensor(["m", "n"], "float32"),
            weight: R.Tensor(["m * n"], "float32"),
        ) -> R.Tensor(["m", "n"], "float32"):
            m = T.int64()
            n = T.int64()
            R.func_attr({"num_input": 1})
            with R.dataflow():
                weight = R.add(weight, R.const(1, "float32"))
                weight = R.reshape(weight, [m, n])
                output = R.multiply(x, weight)
                R.output(output)
            return output

    @tvm.script.ir_module
    class Expected:
        pass  # TODO

    After = relax.transform.LiftTransformParams()(Before)
    tvm.ir.assert_structural_equal(Expected, After)
```

Lunderberg (changes requested):
I like the change, and this is a good improvement for `LiftTransformParams`!  It wasn't needed in my use case, because the symbolic variables ended up canceling out, but this is a better and more robust solution.

I have one requested change to handle an edge case that would produce ill-formed output, and a unit test that would trigger that edge case.

tests/python/relax/test_transform_lift_transform_params.py
jinhongyii: The unittest should be introduced in another PR because it involves another problem. Below is the program after applying the pass and you can see that main_transform_params is not well defined because m and n are not defined.
```@I.ir_module
class Module:
    @R.function
    def main_transform_params(params: R.Tuple(R.Tensor(("m * n",), dtype="float32"))) -> R.Tuple(R.Tensor(dtype="float32", ndim=1)):
        m = T.int64()
        n = T.int64()
        with R.dataflow():
            lv: R.Tensor((m * n,), dtype="float32") = params[0]
            gv: R.Tuple(R.Tensor((m * n,), dtype="float32")) = (lv,)
            R.output(gv)
        return gv

    @R.function
    def main(x: R.Tensor(("m", "n"), dtype="float32"), transformed_param_0: R.Tensor(dtype="float32", ndim=1)) -> R.Tensor(("m", "n"), dtype="float32"):
        m = T.int64()
        n = T.int64()
        R.func_attr({"num_input": 1})
        with R.dataflow():
            lv: R.Tensor(dtype="float32", ndim=1) = transformed_param_0
            weight: R.Tensor(dtype="float32", ndim=1) = R.add(lv, R.const(1, "float32"))
            weight_1: R.Tensor((m, n), dtype="float32") = R.reshape(weight, R.shape([m, n]))
            output: R.Tensor((m, n), dtype="float32") = R.multiply(x, weight_1)
            R.output(output)
        return output
```
An ideal output should be 
```@I.ir_module
class Module:
    @R.function
    def main_transform_params(params: R.Tuple(R.Tensor(("k",), dtype="float32"))) -> R.Tuple(R.Tensor(dtype="float32", ndim=1)):
        k = T.int64()
        with R.dataflow():
            lv: R.Tensor((k,), dtype="float32") = params[0]
            gv: R.Tuple(R.Tensor((k,), dtype="float32")) = (lv,)
            R.output(gv)
        return gv

    @R.function
    def main(x: R.Tensor(("m", "n"), dtype="float32"), transformed_param_0: R.Tensor(dtype="float32", ndim=1)) -> R.Tensor(("m", "n"), dtype="float32"):
        m = T.int64()
        n = T.int64()
        R.func_attr({"num_input": 1})
        with R.dataflow():
            lv: R.Tensor(dtype="float32", ndim=1) = transformed_param_0
            weight: R.Tensor(dtype="float32", ndim=1) = R.add(lv, R.const(1, "float32"))
            weight_1: R.Tensor((m, n), dtype="float32") = R.reshape(weight, R.shape([m, n]))
            output: R.Tensor((m, n), dtype="float32") = R.multiply(x, weight_1)
            R.output(output)
        return output
```

jinhongyii:
@Lunderberg please review again

tests/python/relax/test_transform_lift_transform_params.py
Lunderberg: Hmm, you're right.  I had thought that the proposed unit test would be normalized to `R.Tensor(ndim=1)`, and that the lifting of the parameter would result in an undefined usage of the symbolic variable.  Instead, this test cases causes undefined variables  in the output of `LiftTransformParams` prior to this PR as well.

Can we include the unit test, but mark it with `@pytest.mark.xfail`?  That way, it wouldn't cause the CI to fail, but the known failure mode would be documented.

jinhongyii: ok I just added the test

Lunderberg: Oof, looks like it's a pretty tricky one to find a long-term solution for.

1. `FuncStructInfo` is used both for top-level functions (must define all symbolic vars) and for local bindings of functions (may use symbolic vars from parent scope).
2. `FuncStructInfo` does not distinguish between usage of symbolic variables and definitions of symbolic variables.
3. Because it cannot distinguish between valid usage of symbolic vars from the parent scope and invalid usage of undefined symbolic vars, `EraseToWellDefined` doesn't modify `FuncStructInfo`.
4. For the same reason, the `relax::Function` constructor does not throw an error if a parameter contains context-defined symbolic variables.
5. `LiftTransformParams` generates a `relax::Function` that would be valid in a local binding, but isn't valid as a top-level function.  Therefore, the undefined variables are retained.

I think the long-term resolution will be to either extend `FuncStructInfo` to distinguish between usage/definition of symbolic variables, or to preferentially use local bindings instead of new functions, in order to centralize the special-handling into `relax.transform.LambdaLift`.

For now, I think adding the known failure mode with `@pytest.mark.xfail` is sufficient.

jinhongyii:
@Lunderberg Can we merge this now?

Lunderberg (approved):
Thank you for making the changes, and it looks good to me!

PR #16051, opened 2023-11-03
[Fix][TIR] Fix dtype issues for match_buffer and ramp node

Ubospica:
This PR fixes dtype issues for 
- the strides parameter of match_buffer. 
  - When stride is a tir var, it would become a string like `"A_s0"` in the printed script. 
  - If the strides are int64 vars, when we parse the printed script, the dtype of the strides will be set to int32 by default. That causes dtype misalign and effects the structural_equal check. 
  - We will cast the string stride to the dtype of shape (which is always a primexpr)
- ramp node. If the dtypes of base and stride of a ramp node do not match, the dtype of stride will be casted to the dtype of base.


tqchen:
@tvm-bot rerun

tqchen (approved):

PR #16052, opened 2023-11-03
[TOPI] Enhance `topi.nn.matmul`

Ubospica:
This PR enhances `topi.nn.matmul` to support batch matmuls (i.e. inputs with dims larger than 2). 

This is useful in certain cases where we need to generate a batch matmul kernel.

### DIFF
--- a/python/tvm/topi/nn/dense.py
+++ b/python/tvm/topi/nn/dense.py
@@ -19,7 +19,7 @@
 import tvm
 from tvm import auto_scheduler, te
 
-from .. import tag
+from .. import tag, add
 
 
 def matmul(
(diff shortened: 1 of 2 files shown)


Hzfengsy:
We have a separate [`topi.nn.batch_matmul`](https://github.com/Hzfengsy/tvm/blob/a13eadbddaa94246f2d572aed6e04b70737616c5/python/tvm/topi/nn/batch_matmul.py#L29), I doubt if it's necessary to add it to `matmul`

tqchen:
I agree, in this case, moving to use batch_matmul might be easier

Ubospica:
I see. That makes sense

Ubospica:
I see. That makes sense

Ubospica:
> I agree, in this case, moving to use batch_matmul might be easier

What I need is a matmul where the shape of A is [b, m, k] and B is [n, k], and A or/and B could be transposed. But batch_matmul only supports 3-dim A and B. 

And I guess having a matmul operator supporting all input dimensions may be helpful. As described in [data apis](https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__matmul__.html), the additional dimensions will be broadcasted.

tqchen:
OK, in this case i think we can allow the extended version, provided that it does not break original usage and we have testcases to cover the extended usecase in form of TIR

Ubospica:
@tvm-bot rerun

Hzfengsy (approved):

python/tvm/topi/nn/dense.py
```diff
@@ -65,86 +65,103 @@ def matmul(
...
+        reduce_dim_a, in_dim = tensor_a.shape[-2:]
     else:
-        batch, in_dim = tensor_a.shape
+        in_dim, reduce_dim_a = tensor_a.shape[-2:]
+    batch_dims_a = tensor_a.shape[:-2]
 
     if auto_scheduler_rewritten_layout:
         # Infer shape for the rewritten layout
-        out_dim, red_dim = auto_scheduler.get_shape_from_rewritten_layout(
+        assert len(tensor_b).shape == 2, "only support 2-dim matmul when using auto-scheduler"
```
happyme531, line 85: This failed...

tqchen: cc @Ubospica please followup

PR #16059, opened 2023-11-03
[Unity][Relax Frontend] optimize squeeze op converter.

brainternet:
in class Squeeze's _impl_v13()， the attr arg has't been used, if it has explict axis info, try to use it directly.


Hzfengsy:
Please add a testcase for it

PR #16063, opened 2023-11-03
[TIR] Fix pass RenewDefs error in gather/take case

Hzfengsy:
Pervious implementation of RenewDefs pass will fail in the case of the gather/take function. This is because the pass visit and renew the read/write regions twice. This PR fixes it and adds a regression test.

cc @cyx-6


vinx13 (approved):

cyx-6 (approved):
Thanks for fixing!

junrushao:
Could you take a brief look into the hexagon failure? seems irrelevant...?

PR #16066, opened 2023-11-03
[Runtime] Support clear global memory allocators

MasterJH5574:
This PR supports clearing up all the allocated memory.

Prior to this PR, all the allocated memory are managed in the pool of memory manager. The allocated memory in the pool is on hold and never freed. Consequently, the pool size always goes up monotonically within a single run in a process.

While good to save time of memory allocation, in some cases (e.g., on mobile phones which may have running memory limit) we need to clear the pool and free all the memory in order to prevent the pool from endlessly growing up and some of allocated memory not being effectively utilized (fragmentation).

Therefore, this PR introduces a PackedFunc that helps clean up the memory manager, releasing all the allocated memory. Runtime apps can decide when to invoke this PackedFunc and clean up the pool. Usually, this will happen at some app "reset" or "reload" stage.

### DIFF
--- a/src/runtime/memory/memory_manager.cc
+++ b/src/runtime/memory/memory_manager.cc
@@ -22,6 +22,7 @@
  * \brief Allocate and manage memory for the runtime.
  */
 #include <tvm/runtime/memory/memory_manager.h>
+#include <tvm/runtime/registry.h>
 
 #include <memory>
 #include <utility>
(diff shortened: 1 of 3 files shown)


src/runtime/memory/memory_manager.cc
```diff
@@ -166,6 +167,12 @@ Allocator* MemoryManager::GetAllocator(Device dev, AllocatorType type) {
   return it->second.at(type).get();
 }
 
+void MemoryManager::Clear() {
+  MemoryManager* m = MemoryManager::Global();
+  std::lock_guard<std::mutex> lock(m->mu_);
+  m->allocators_.clear();
```
tqchen, line 173: let us iterate over each allocator and clear each of the freed entries, as oppose to clear the vector, which can cause issues when some of the allocated storage from the allocators get retained and deleted after clear

tqchen (changes requested):

src/runtime/memory/memory_manager.cc
MasterJH5574: Thanks for pointing it out. Just pushed a new commit, not sure if it addresses in the exactly same way as you mentioned.

junrushao:
Any updates? The latest rebase/merge will depend on this PR

src/runtime/memory/memory_manager.cc
```diff
@@ -166,6 +167,18 @@ Allocator* MemoryManager::GetAllocator(Device dev, AllocatorType type) {
   return it->second.at(type).get();
 }
 
+void MemoryManager::Clear() {
+  MemoryManager* m = MemoryManager::Global();
```
tqchen, line 171: We need to add a Clear interface to base Allocator, and iterate over the allocators, to call Allocator->Clear() without removing any allocators

tqchen (approved):

PR #16069, opened 2023-11-04
[Unity][Support] Sample from top-p supports offset

MasterJH5574:
This PR enhances the `SampleTopPFromProb` function, adding a new `unit_offset` parameter. With this PR, the input prob array can contain multiple prob distributions. The offset is used to specify which probability to sample from.

Prior to this PR, this function assumes the input prob array only contains a single probability distribution.
This assumption, though effective, prevents efficient sampling for batching scenarios, since it forces us to split the batched prob array before sampling.

This PR eliminates this issue by introducing the offset.


junrushao (approved):
LGTM!

PR #16070, opened 2023-11-04
[TIR][Transform] Introduce new `InjectPermutedLayout` pass

Ubospica:
This PR introduces the new `InjectPermutedLayout` pass, partially adapted from the old version.

This pass will permute the visiting indices of the shared memory with xor operations to avoid bank conflicts. For permutation details see p39 of [these slides](https://developer.download.nvidia.com/video/gputechconf/gtc/2020/presentations/s21745-developing-cuda-kernels-to-push-tensor-cores-to-the-absolute-limit-on-nvidia-a100.pdf).

The current interface is:
```
with T.block("..."):
    T.block_attr({"permuted_layout": 1})
    # load or store to shared memory
```

The load/store operation could be:
- BufferLoad node
- BufferStore node
- ptx_ldmatrix call
- mma_store call

And the InjectPermutedLayout pass is injected in the lowering process in `driver_api.cc`. So after this pass in the building stage, the visiting indices will be permuted.

Notably this pass works after the `CompactBufferAllocation` pass because it relies on the shape of the shared memory in every threadblock.

cc @spectrometerHBH @vinx13 @Hzfengsy

### DIFF
--- a/src/tir/transforms/inject_permuted_layout.cc
+++ b/src/tir/transforms/inject_permuted_layout.cc
@@ -279,6 +279,7 @@ class PermutedLayoutInjector : private IRMutatorWithAnalyzer {
   static constexpr size_t VECTORIZE_FACTOR = 8;
   static constexpr size_t BANK_SIZE_BYTES = 128;
 
+  // Mapping from data Var of a Buffer to Buffer, for lookup
   std::unordered_map<Var, Buffer, ObjectPtrHash, ObjectPtrEqual> buffer_map_;
   bool permute_ = false;
 };
--- a/src/tir/transforms/inject_permuted_layout.cc
+++ b/src/tir/transforms/inject_permuted_layout.cc
@@ -19,44 +19,55 @@
 
 /*!
  * \file inject_permuted_layout.cc
- * \brief The pass for inject permuted layout.
+ * \brief The pass injects permuted layout for shared memory buffers to avoid bank conflicts.
  */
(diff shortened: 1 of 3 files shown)


src/tir/transforms/inject_permuted_layout.cc
```diff
@@ -77,183 +89,196 @@ class PermutedLayoutInjector : public StmtExprMutator {
...
+    if (!call->op.same_as(builtin::ptx_ldmatrix()) && !call->op.same_as(builtin::mma_store())) {
+      return call;
+    }
+
+    if (call->op.same_as(builtin::ptx_ldmatrix())) {
+      // form: T.ptx_ldmatrix(..., smem_ptr, smem_offset)
+      // smem_ptr: T.tvm_access_ptr(ptype, data, offset, extent, rw_mask)
+      auto access_ptr = call->args[5];
+      PrimExpr smem_offset = call->args[6];
+      auto new_access_ptr = HandleAccessPtrAndOffset(access_ptr, smem_offset);
```
vinx13, line 252: it would be better to record var to buffer mapping during visiting to avoid visit the whole prim func to find the buffer

vinx13 (approved):

Ubospica:
@tvm-bot rerun

Ubospica:
@tvm-bot rerun

yzh119 (approved):

PR #16077, opened 2023-11-05
Revert "[Unity][Support] Sample from top-p supports offset"

MasterJH5574:
This PR reverts apache/tvm#16069.

We found that there are more improvements for the sampling function. We would like to implement them and verify the functionality locally first, and upstream when the stability is confirmed.


junrushao (approved):
We should merge the original commit a bit later

PR #16032, opened 2023-11-01
[CMake] Use llvm-config to locate Findzstd.cmake

lhutton1:
LLVM17 now depends on `-lzstd` when the `--link-static` option is used. I.e:
```
$ llvm-config-16 --link-static --system-libs
-lrt -ldl -lm -lz -ltinfo -lxml2

$ llvm-config-17 --link-static --system-libs
-lrt -ldl -lm -lz -lzstd -ltinfo -lxml2
```

The current cmake rules aren't able to find the "Findzstd.cmake", resulting in a cmake error. This commit uses `llvm-config --cmakedir` to locate "Findzstd.cmake" which defines a set of rules to find zstd in the system. The zstd library is already installed as part of: https://github.com/apache/tvm/pull/15799/files#diff-c2c0605a8fdd4f5600690a8c7b1ec769882426af1b0ed01e8aaa0814e3a8f5dbR48

PR #16081, opened 2023-11-06
[Fix] Broken Windows Build with LLVM

junrushao:
PR #16032 introduces a more stable way of zstd path finding by relying on LLVM's existing Findzstd module, but it turns out that the cmake dir from `llvm-config` is not very compatible with CMake's internal convention, and thus a conversion is required to get rid of the error message below:

```
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD

    D:/a/package/package/tvm/build/CMakeFiles/CMakeScratch/TryCompile-tnlfb1/CMakeLists.txt:2

-- Configuring incomplete, errors occurred!
  when parsing string

    C:\Miniconda\envs\tlcpack-build\Library\lib\cmake\llvm

  Invalid character escape '\M'.
```

CC @lhutton1


tqchen (approved):

lhutton1:
Thanks for checking the windows build @junrushao!

PR #16084, opened 2023-11-07
[Unity][Dlight] Enhance matmul tensorizer with Int8 support

ibsidorenko:
Main goal of this commit was not to provide highly optimized schedule, but to enable tensor core tensorization for i8i8i32 matmul. Now it supports fp16 matmul only.

Anyway, speedup is a case compared to fp16 (fp16fp16->fp16 vs i8i8->i32):

N |        M, N, K*      |   Shape type   | fp16, us | int8, us | speedup |
--|----------------------|----------------|----------|----------|---------|
1 |     256, 256, 256    |  static shapes |   15.03  |   12.55  |  1.198x |
2 |     512, 512, 4096   |  static shapes |  173.25  |  129.06  |  1.342x |
3 | N (=260), 4096, 4096 | dynamic shapes |  350.44  |  298.29  |  1.175x |

where K* - is a reduction axis.

Implementation of Matmul int8 tensorizer was derived from Matmul tensorizer with minimal changes. It is possible to join these implementations (fp16 and int8), but I am not sure this is a good idea. Since it can slow down development of fp16 part.

### DIFF
--- a/python/tvm/dlight/gpu/matmul.py
+++ b/python/tvm/dlight/gpu/matmul.py
@@ -296,6 +296,16 @@ def get_reduction_blocks(sch, blocks) -> bool:
     return reduction_blocks
 
 
+def get_in_out_dtypes(block: tir.Block) -> Tuple[str]:
+    """
+    Detect In/Out data types for the given block based on the analysis if read/write buffers.
+    """
+    assert len(block.reads) > 0 and len(block.writes) > 0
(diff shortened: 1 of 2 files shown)


Hzfengsy:
Awesome work! Thanks very much, @ibsidorenko! cc @vinx13 @masahi for review

vinx13 (approved):

PR #16085, opened 2023-11-07
[Unity] Support Regular expression matching in globalvar dataflow pattern

jinhongyii:
When matching global var of TIR functions, sometimes we want to use regular expression to match because the TIR functions are often named like matmul_123. This PR implements this behavior.


jinhongyii:
cc: @masahi

masahi (approved):

PR #16086, opened 2023-11-07
[Unity] support symbolic var in RewriteDataflowReshape

jinhongyii:
Current RewriteDataflowReshape pass cannot detect reshape from [n, 16, 128] to [1, n, 16, 128] because of simplifier is not symbolic-var-aware. This PR adds an additional IterMapSimplify to solve the problem.


jinhongyii:
cc: @masahi

masahi (approved):

PR #16087, opened 2023-11-07
Add _ffi_api.py under script folder

jinhongyii:
Add _ffi_api.py under script folder so that user can call script.Complete in python.


jinhongyii:
cc: @Hzfengsy

Hzfengsy (approved):

PR #16088, opened 2023-11-07
[Unity] make LazyTransformParam more general

jinhongyii:
1. user can now customize the name of packed func get_item and set_item
2. user can now only rewrite get_item but not set_item. This is useful when we want a tuple as output.

### DIFF
--- a/python/tvm/relax/transform/lazy_transform_params.py
+++ b/python/tvm/relax/transform/lazy_transform_params.py
@@ -16,6 +16,7 @@
 # under the License.
 # pylint: disable=invalid-name, unused-argument, missing-function-docstring, abstract-method
 """Relax LazyTransformParams pass."""
+from typing import Optional
 import tvm
 import itertools
 from tvm import IRModule, relax
@@ -108,8 +109,7 @@ class LivenessAnalysis(PyExprVisitor):
         self.var_liveness_end[binding.var] = self.last_appear_in_var_binding
 
 
-@mutator
-class LazyTransformParamsMutator(PyExprMutator):
+class LazyTransformParamsFuncCreator:
     """
     Transform transform_params functions into a lazy version.
 
@@ -120,21 +120,24 @@ class LazyTransformParamsMutator(PyExprMutator):
     """
 
     def __init__(
-        self, fget_item, fset_item, get_item_param, set_item_param, mod: IRModule = None
+        self,
+        fget_item,
+        fset_item,
+        extra_get_item_params,
+        extra_set_item_params,
+        mod: IRModule = None,
     ) -> None:
-        super().__init__(mod)
         self.mod = mod
         self.fget_item = fget_item
-        self.get_item_param = get_item_param
+        self.extra_get_item_params = extra_get_item_params
         self.fset_item = fset_item
-        self.set_item_param = set_item_param
+        self.extra_set_item_params = extra_set_item_params
         # the only input param, which should be a Tuple
         self.input_tuple_param = None
         self.input_params_set = None
         self.out_tuple_map = None
         self.out_tuple_var = None
         self.memory_free_insertion = None
-        self.killed_vars = set()
 
     def transform(self, func: relax.Function) -> relax.Function:
         self.input_tuple_param = func.params[0]
@@ -154,12 +157,15 @@ class LazyTransformParamsMutator(PyExprMutator):
         self.memory_free_insertion = liveness.var_liveness_end
 
         # Step 3. rewrite get item and set item
-        new_body = self.visit_expr(func.body)
+        new_body = func.body
+        if self.fget_item is not None:
+            new_body = LazyInputMutator(self, self.mod).visit_expr(new_body)
+
+        if self.fset_item is not None:
+            new_body = LazyOutputMutator(self, self.mod).visit_expr(new_body)
 
         # Step 4. Add parameters of get_item and set_item (except index) to the function.
-        params = []
-        for param in itertools.chain(self.get_item_param, self.set_item_param):
-            params.append(param)
+        params = [*self.extra_get_item_params, *self.extra_set_item_params]
 
         # Step 5. Find all shape parameters that should be retained as
         # parameters.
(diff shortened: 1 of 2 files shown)


jinhongyii:
cc: @Lunderberg @MasterJH5574

python/tvm/relax/transform/lazy_transform_params.py
```diff
@@ -225,10 +236,38 @@ class LazyTransformParams:
...
+        call_packed(fget_item, [param1, param2, index])
+    set_item_param: list of relax.Var
+        The parameters of the set_item function except index and value.
+        The given parameters will be placed before index and value.
+        For example, if set_item_param is [param1, param2], then the pass will generate
+        call_packed(fset_item, [param1, param2, index, value])
     """
 
+    def __init__(
+        self, fget_item="get_item", fset_item="set_item", get_item_param=[], set_item_param=[]
```
Lunderberg, line 259: Using a mutable object as a default value can have unexpected consequences.  (e.g. Accidental mutation of an object, which is reused in the next call.)  Instead of having a list as the default value, it is better to have `get_item_param=None`, then to construct the list within the body of the function.

python/tvm/relax/transform/lazy_transform_params.py
```diff
@@ -118,9 +119,15 @@ class LazyTransformParamsMutator(PyExprMutator):
...
     """
 
-    def __init__(self, mod: IRModule = None) -> None:
+    def __init__(
+        self, fget_item, fset_item, get_item_param, set_item_param, mod: IRModule = None
+    ) -> None:
         super().__init__(mod)
         self.mod = mod
+        self.fget_item = fget_item
+        self.get_item_param = get_item_param
```
Lunderberg, line 128: What is the purpose of the `get_item_param` and `set_item_param`?  I don't see any unit test with their intended use.  Since the same parameters are forwarded to all `get_item` and `set_item` calls, it seems the same effect could be achieved by having a stateful callback, rather than passing an argument which is then passed right back to you.

python/tvm/relax/transform/lazy_transform_params.py
```diff
@@ -188,24 +199,24 @@ def visit_var_(self, var: relax.Var) -> None:
         return super().visit_var_(var)
 
     def visit_var_binding_(self, binding: relax.VarBinding) -> None:
-        if binding.var == self.out_tuple_var:
+        if self.fset_item is not None and binding.var == self.out_tuple_var:
```
Lunderberg, line 202: It looks like the handling of `set_item` and of `get_item` is largely independent.  Rather than having them as two features of a single mutator, which requires the `self.fset_item is not None` check to be repeated throughout, can we split it into two distinct mutators?  One mutator replaces access of parameters with `get_item`.  The other mutator replaces binding of output with `set_item`.  If both features are desired, then both mutators are applied.

python/tvm/relax/transform/lazy_transform_params.py
```diff
@@ -149,9 +156,13 @@ def transform(self, func: relax.Function) -> relax.Function:
         # Step 3. rewrite get item and set item
         new_body = self.visit_expr(func.body)
 
-        # Step 4. Find all shape parameters that should be retained as
-        # parameters.
+        # Step 4. Add parameters of get_item and set_item (except index) to the function.
         params = []
+        for param in itertools.chain(self.get_item_param, self.set_item_param):
```
Lunderberg, line 161: Python's `*` expansion avoids the need for a loop.

```python
params = [*self.get_item_param, *self.set_item_param]
```

It's a bit more "magic", but avoids mutation of a variable after assignment.

python/tvm/relax/transform/lazy_transform_params.py
```diff
@@ -118,9 +119,15 @@ class LazyTransformParamsMutator(PyExprMutator):
         The module to be transformed
     """
 
-    def __init__(self, mod: IRModule = None) -> None:
+    def __init__(
+        self, fget_item, fset_item, get_item_param, set_item_param, mod: IRModule = None
```
Lunderberg, line 123: From the usage, the `get_item_param` and `set_item_param` are lists of elements, so they should be plural `get_item_params` and `set_item_params`.  (If they are required overall.  See other comment asking if they are.)

This would also be an advantage of breaking the functionality apart into two separate mutators: A `LazyInput` mutator could have an `extra_params` argument, and a `LazyOutput` mutator could have an `extra_params` argument, and both would be clear from the context.  With both mutations implemented in the same class, if we wanted the name to indicate that these are additional arguments, we'd need to make it be `extra_get_item_params` and `extra_set_item_params`, which is a bit long for ease of use.

python/tvm/relax/transform/lazy_transform_params.py
jinhongyii: My bad. I forget to add a test on this. The case is that I need to call `get_item(loader, index)`. Since shard loader is created at runtime, it cannot be a stateful callback and I have to modify the interface of `get_item`. Does this make sense to you?

python/tvm/relax/transform/lazy_transform_params.py
jinhongyii: sounds good. let me try

python/tvm/relax/transform/lazy_transform_params.py
jinhongyii: I cannot fully separate these mutators because they need to share `input_tuple_param` and func.params[0] will change after one mutator.

jinhongyii: so what I can do is create a class that does all the analysis, calls into the two mutators, and create function. In this case, we have to use `extra_get_item_params` and `extra_set_item_params`

jinhongyii:
@Lunderberg I've tried my best to address your comment, please review again

python/tvm/relax/transform/lazy_transform_params.py
Lunderberg: Hmm, looks like the analysis would need to be split apart as well.  In general, I think having multiple smaller changes is better than a single mutator, but this seems like a refactor than would make sense to ask in a feature addition PR.  The current revision looks good, and any additional changes can be in a follow-up PR if desired.

Lunderberg (approved):

Lunderberg:
Looks good, and approved.  Current CI failures are due to use of `IRModule | None` as a type annotation, which requires [PEP-0604](https://peps.python.org/pep-0604/) support added in python 3.10.  As TVM supports 3.7+, these annotations should be replaced with `typing.Optional[IRModule]`.

PR #16090, opened 2023-11-08
[Unity][BlockBuilder] Depracate `BlockBuilder.get()` and change it to `BlockBuilder.finalize()`

Ubospica:
Previously BlockBuilder would have a problem about naming of public function if we 
1) emit a private function, and then 
2) emit a public function with the same name. 

Previously, BlockBuilder would rename the latter, i.e. the public function. That led the name of the public function did not match the name of its global_symbol attribute.

To tackle this problem, this PR introduces a new pass NormalizeGlobalVar. This pass will possibly rename the GlobalVar in an IRModule to ensure these properties:
1. (Invariant) First ensure every public function has the same name as its "global_symbol" attribute ;
2. To ensure 1., we may need to rename private functions with conflicting names; 
3. Finally, the name of every GlobalVar is unique in the IRModule.

This PR adds a member function `bb.finalize()` to be used at the end of the building process. It will call NormalizeGlobalVar to ensure the invariant, and returns the final IRModule. It essentially substitutes the functionality of `bb.get()`, so we mark the latter deprecated.

About naming: we can still use the interface `bb.get()` and execute a NormalizeGlobalVar pass in it. But such `bb.get()` can only be called once at the end of the building process. So we rename it to `bb.finalize()` for clearity and alignment with C++ side.

cc @junrushao @tqchen

### DIFF
--- a/python/tvm/relax/block_builder.py
+++ b/python/tvm/relax/block_builder.py
@@ -14,7 +14,7 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-# pylint: disable=no-else-return, invalid-name, unused-argument
+# pylint: disable=no-else-return, invalid-name, unused-argument, import-outside-toplevel
 """Developer API of constructing Relax AST."""
 
 from typing import Dict, List, Optional, Union, Any, Callable, Sequence
--- a/python/tvm/relax/block_builder.py
+++ b/python/tvm/relax/block_builder.py
@@ -17,23 +17,19 @@
 # pylint: disable=no-else-return, invalid-name, unused-argument, import-outside-toplevel
 """Developer API of constructing Relax AST."""
 
-from typing import Dict, List, Optional, Union, Any, Callable, Sequence
+from typing import Any, Callable, Dict, List, Optional, Sequence, Union
+
+import tvm
+from tvm import relax as rx
+from tvm import tir
+from tvm.ir.base import deprecated
 from tvm.ir.module import IRModule
 from tvm.runtime import Object
-from tvm import relax as rx, tir
-import tvm
-from .expr import (
-    Expr,
-    Var,
-    GlobalVar,
-    BindingBlock,
-    Tuple,
-    BaseFunc,
-    Binding,
-)
-from .struct_info import StructInfo
-from .op.base import call_tir, call_tir_with_grad
+
 from . import _ffi_api
+from .expr import BaseFunc, Binding, BindingBlock, Expr, GlobalVar, Tuple, Var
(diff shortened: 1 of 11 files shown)


python/tvm/relax/block_builder.py
```diff
@@ -661,12 +661,20 @@ def normalize(self, expr: Expr) -> Expr:
...
         Returns
         -------
         ret : tvm.IRModule
             An IRModule with Relax and TIR functions being built.
         """
-        return _ffi_api.BlockBuilderGetContextIRModule(self)  # type: ignore
+        # avoid circular import
+        from .transform import NormalizeGlobalVar
+
+        mod = _ffi_api.BlockBuilderGetContextIRModule(self)  # type: ignore
```
Hzfengsy, line 676: Is it possible to embed the pass inside the `BlockBuilderGetContextIRModule`? aka we can have the same behavior on both python and cpp side

Ubospica: We have considered that. Here the problem is NormalizeGlobalVar will invalidate all global vars previously acquired. But I cannot ensure all pass will follow the assumption that no global var will be used again after BlockBuilderGetContextIRModule. Actually, some passes like AllocateWorkspace might already break the assumption.

On the other hand, on python side BlockBuilder is mainly to build a new IRModule from scratch, and not involve too complex logic. So apply NormalizeGlobalVar to bb.get() could be a good choice.

tqchen: I discussed it with @Ubospica and think that we should not turn it on by default. Because there are not too many cases we create new public functions, and renormlaization can be costly. We should however add wellform check to ensure normal passes do not generate code that violate this assumption


Maybe we do not make it part of `BlockBuilder.get()` but instead add a new function `BlockBuilder.finalize()` and add remark that it will also normalizes the calls.

Ubospica: But I think this will make it confused to use `bb.finalize()` or `bb.get()`, because their only difference is whether apply NormalizeGlobalVar or not. If we require users only use `bb.finalize()` at the end of the IRModule building process, that means a major behaviour change and all related tests might need to be changed.

tqchen: OK, we can keep this for now and keep the documentaiton.

Ubospica:
cc @Hzfengsy @tqchen @junrushao Plz take a look

Ubospica:
@tvm-bot rerun

Ubospica:
@tvm-bot rerun

Ubospica:
@tvm-bot rerun

tqchen:
@Ubospica you need to push a new commit

Hzfengsy (approved):

PR #16092, opened 2023-11-08
[Unity] Add `axis` field to scatter_from_worker0

jinhongyii:
This PR adds an `axis` field to scatter_from_worker0, which means the tensor axis along which it is scattered. legalize_ops will automatically generate reshape and transpose to preserve the constraint of ccl that collective communication ops must be performed on contiguous memory. For example, if the tensor shape of x is [10, 20], and we have `scatter_from_worker0(x, num_workers=2, axis=1)`, then after legalization it will expand to 

```
x = reshape(x, [10, 2, 10]) # shape: [10, 2, 10]
x = permute_dims(x, [1, 0, 2]) # shape: [2, 10, 10]
x = call_dps_packed("scatter_from_worker0", x) # shape: [10, 10]
```
When axis=0, the behavior is the same as before.

Also, this PR renames ScatterFromWorker0Attrs to ScatterAttrs to enable reuse by other ops like worker-id-aware slicing (scatter_from_worker0 = broadcast_from_worker0 +  worker-id-aware slicing).

### DIFF
--- a/include/tvm/relax/attrs/ccl.h
+++ b/include/tvm/relax/attrs/ccl.h
@@ -41,11 +41,11 @@ struct AllReduceAttrs : public tvm::AttrsNode<AllReduceAttrs> {
 };  // struct AllReduceAttrs
 
 /*! \brief Attributes used in scatter operators */
-struct ScatterAttrs : public tvm::AttrsNode<ScatterAttrs> {
+struct ScatterCollectiveAttrs : public tvm::AttrsNode<ScatterCollectiveAttrs> {
   int num_workers;
   int axis;
 
-  TVM_DECLARE_ATTRS(ScatterAttrs, "relax.attrs.ScatterAttrs") {
+  TVM_DECLARE_ATTRS(ScatterCollectiveAttrs, "relax.attrs.ScatterCollectiveAttrs") {
     TVM_ATTR_FIELD(num_workers)
         .describe(
             "The number of workers, also the number of parts the given buffer should be chunked "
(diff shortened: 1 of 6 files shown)


jinhongyii:
cc: @junrushao

include/tvm/relax/attrs/ccl.h
```diff
@@ -40,17 +40,21 @@ struct AllReduceAttrs : public tvm::AttrsNode<AllReduceAttrs> {
   }
 };  // struct AllReduceAttrs
 
-/*! \brief Attributes used in scatter_from_worker0 operators */
-struct ScatterFromWorker0Attrs : public tvm::AttrsNode<ScatterFromWorker0Attrs> {
+/*! \brief Attributes used in scatter operators */
+struct ScatterAttrs : public tvm::AttrsNode<ScatterAttrs> {
```
junrushao, line 44: Scatter is a bit ambiguous. It could mean an op or a communication collective. What do you have a better naming specifically for this usecase?

jinhongyii: Good point. I have no idea on this naming. @tqchen Do you have any suggestions? This attr is expected to be reused by R.dist.redistribute_replica_to_shard we discussed before.

tqchen: How about ScatterCollectiveAttr?

jinhongyii: sounds good

junrushao:
OK that’s a lot better. Thanks!

PR #16094, opened 2023-11-09
[Unity][DistIR] Enhance PropagateSharding pass

jinhongyii:
1. PropagateSharding can now handle tuples.
2. A better sharding propagation algorithm in TIR which can cover all the operators in llama. There's no need to register how to propagate sharding through a Relax high-level operator. An easier way is to run LegalizeOps first, and then this pass will automatically analyze TIR and figure out a way to propagate.


jinhongyii:
cc: @tqchen

tqchen:
cc @csullivan @Lunderberg  for awareness

tests/python/relax/distributed/test_distributed_transform_propagate_sharding.py
```diff
@@ -68,13 +68,119 @@ def foo(
...
         lv1: R.DTensor((128, 128), "float32", "mesh[0]", "S[1]") = R.nn.gelu(lv0)
-        lv2: R.DTensor((128, 128), "float32", "mesh[0]", "S[1]") = lv1
         lv3: R.DTensor((128, 128), "float32", "mesh[0]", "R") = R.matmul(
-            lv2, weight2, out_dtype="void"
+            lv1, weight2, out_dtype="void"
         )
         return lv3
 
 
+@I.ir_module
```
Lunderberg, line 77: Nitpick: Parsing of TVMScript should be done within the test body (if used for a single test case), or within a helper function called by the body (if used by multiple test cases).  If the TVMScript is at module-scope, parsing occurs when the module is imported.  This causes the entire pytest suite to halt at collection-time.  Keeping TVMScript parsing inside the test body causes the failure to occur within the execution of that unit test.

This is a large usability issue when debugging the TVMScript parser, as it becomes impossible to tell how many test failures exist, let alone if there are any commonalities between them.

tests/python/relax/distributed/test_distributed_transform_propagate_sharding.py
```diff
@@ -1060,16 +1738,19 @@ def test_mlp_pipeline_parallelism():
...
-    after = relax.distributed.transform.PropagateSharding()(mod)
+    after = relax.distributed.transform.PropagateSharding()(LlamaAttentionLayer)
     assert_structural_equal(after, ShardedLlamaAttentionLayer)
 
 
-def test_decoder_layer_dynamic_shape():
-    # mod = relax.transform.LegalizeOps({"relax.reshape": lambda bb, call: bb.normalize(call)})(LlamaAttentionLayer)
-    mod = LlamaAttentionLayerDynamicShape
+def test_decoder_layer_tir():
+    mod = relax.transform.LegalizeOps()(LlamaAttentionLayer)
```
Lunderberg, line 1746: With the length of the expected output in `ShardedLlamaAttentionLayerTIR`, I can't tell what behavior this test is intended to check.  Can we add a docstring describing how TIR calls should be handled when propagating the sharding?

My initial guess would be that they are treated as opaque, because the shapes used by `PrimFunc`s in`ShardedLlamaAttentionLayerTIR` are unsharded, but that isn't readily apparent to a reader.

Lunderberg (commented):
Thank you, and I'll take a more detailed read when I'm able.  For now, just a couple of minor questions.

tests/python/relax/distributed/test_distributed_transform_propagate_sharding.py
jinhongyii: This test checks whether this pass can propagate sharding annotations through TIR, and TIR are not treated as opaque funcs. The shapes are not changed because we are changing Tensor to DTensor which contains an extra field `placement`. The reason that I use such a long test is to test whether the pass can cover all of the common TIRs that appear in LLM.

jinhongyii: I've added a few lines of comments for this test in the latest commit

Lunderberg:
@jinhongyii Comments are updated.  There appears to be a shape discrepancy between the Relax caller and the TIR callee, though the length of the unit test makes it difficult to tell.

tests/python/relax/distributed/test_distributed_transform_propagate_sharding.py
jinhongyii: My apology. This is an intended discrepancy. To be short, the TIR in the output program is called global view, which represent the global computation of the whole mesh, and it will be lowered to local view, representing local computation executed by each worker. A more detailed introduction is here: https://thirsty-airplane-926.notion.site/DistIR-presentation-76d557b0c9344510b37bda790d5db3ec?pvs=4

Lunderberg: Hmm, thank you for the link, and I was unaware of the distinction.  I'm a bit worried about that later lowering pass, and whether it would require the PrimFunc to be dynamic along the split axis for correctness, but having the global vs local distinction makes sense.

I don't think this PR should be delayed by my late arrival on the design discussion, and shouldn't be blocking for this PR.  I think improving the readability of the test itself is the main thing that would still be useful to do.  The part that took the most time to recognize was that all the PrimFunc definitions were unchanged.  Can we explicitly build the `ShardedLlamaAttentionLayerTIR` by copying the `PrimFunc` objects from the original module, and only providing an updated version of the `@R.function \n def foo()` ?

jinhongyii: It's an interesting topic to allow the split axis to be dynamic, but currently we just enforce that the axis has to be static for simplicity. 

I don't quite understand your suggestion. How do I construct `ShardedLlamaAttentionLayerTIR` by not writing all the TIR primfuncs?

Lunderberg: > It's an interesting topic to allow the split axis to be dynamic, but currently we just enforce that the axis has to be static for simplicity.

I was thinking that it would be the other way around.  In order to correctly split a `PrimFunc`, the function must contain enough information to produce a smaller version of itself.  That would either require the split axis to be dynamic, so size changes are handled automatically, or to undo any constant-folding that has been done using the split size.

> How do I construct ShardedLlamaAttentionLayerTIR by not writing all the TIR primfuncs?

The current way to do so is by replacing the `cls.FUNC_NAME` with `LlamaAttentionLayer.get_global_var("FUNC_NAME")` within the TVMScript, then copying the PrimFunc to it afterwards (example below).  Most of the time, the extra steps make it not worth the effort, but here I think it is worthwhile to show that every `PrimFunc` is expected to be identical.

```python
def test_function():
    @tvm.script.ir_module
    class DefinedInTVMScript:
        @R.function
        def main(
            x: R.Tensor((2, 128, 28), dtype="float32"), w: R.Tensor((64, 16, 3), dtype="float32")
        ) -> R.Tensor((2, 64, 13), dtype="float32"):
            gv = R.call_tir(Before.conv1d, (x, w), out_sinfo=R.Tensor((2, 64, 13), dtype="float32"))
            return gv

        @T.prim_func(private=True)
        def conv1d(
            A: T.Buffer((T.int64(2), T.int64(128), T.int64(28)), "float32"),
            B: T.Buffer((T.int64(64), T.int64(16), T.int64(3)), "float32"),
            group_conv1d_ncw: T.Buffer((T.int64(2), T.int64(64), T.int64(13)), "float32"),
        ):
            pass

    @tvm.script.ir_module
    class AddedAfterward:
        @R.function
        def main(
            x: R.Tensor((2, 128, 28), dtype="float32"), w: R.Tensor((64, 16, 3), dtype="float32")
        ) -> R.Tensor((2, 64, 13), dtype="float32"):
            gv = R.call_tir(
                DefinedInTVMScript.get_global_var("conv1d"),
                (x, w),
                out_sinfo=R.Tensor((2, 64, 13), dtype="float32"),
            )
            return gv

    for gvar, func in DefinedInTVMScript.functions.items():
        if gvar.name_hint != "main":
            AddedAfterward[gvar] = func

    tvm.ir.assert_structural_equal(DefinedInTVMScript, AddedAfterward)
```

(There's a slightly easier way to copy over top-level functions, after https://github.com/apache/tvm/pull/15703, but it doesn't work for the `cls.name_of_prim_func` notation.)

jinhongyii: cool. I've updated the test based on your method.

Lunderberg (approved):
Thank you for the changes, and looks good to me!

PR #16095, opened 2023-11-09
[Unity][DistIR] LowerGlobalViewToLocalView

jinhongyii:
This PR adds a new pass called LowerGlobalViewToLocalView. 
It shards loop-level computation by rewriting TIR according to the sharding annotation on computational graph.

Introduction on what global view and local view are: https://thirsty-airplane-926.notion.site/DistIR-presentation-76d557b0c9344510b37bda790d5db3ec?pvs=4


jinhongyii:
ready for review

src/relax/distributed/transform/lower_global_view_to_local_view.cc
```diff
@@ -0,0 +1,442 @@
...
+      Expr new_func_body = this->VisitExpr(func_->body);
+      ObjectPtr<FunctionNode> new_func = make_object<FunctionNode>(*func_);
+      new_func->body = new_func_body;
+      builder_->UpdateFunction(gv, Function(new_func));
+    }
+    return builder_->GetContextIRModule();
+  }
+
+ private:
+  inline Array<DTensorStructInfo> ExtractDTensorStructInfo(Var var) {
```
yongwww, line 376: could it be a nested tuple (eg: `tup(dtensor, tup(dtensor, dtensor))`)? not sure if this case is disallowed

yongwww (commented):
Overall it looks good to me! Thanks for sharing the do; it helped me a lot!

src/relax/distributed/transform/lower_global_view_to_local_view.cc
jinhongyii: good catch. My current assumption is to disallow nested tuple. And we can surely support it if we find it's necessary to do so.

yongwww (approved):
LGTM

PR #16100, opened 2023-11-10
[Unity][nn.module] Rolling KVCache override method

davidpissarra:
Part of the effort in bringing Mistral to the compilation flow with nn.Module on [mlc-llm](https://github.com/mlc-ai/mlc-llm). Adding support for Rolling KVCache override method in the nn.module frontend.

cc @CharlieFRuan

### DIFF
--- a/python/tvm/relax/frontend/nn/modules.py
+++ b/python/tvm/relax/frontend/nn/modules.py
@@ -670,6 +670,28 @@ class KVCache(Effect):
             )
         )
 
+    def override(self, new_element: Tensor, max_cache_size: int) -> None:
+        """
+        Override a new element in KVCache.
+
+        Parameters
+        ----------
+        new_element : Tensor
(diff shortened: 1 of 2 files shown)


tqchen:
cc @junrushao @cyx-6

junrushao:
I think we should move all KVCache logic to MLC LLM instead, given we expect plenty of variants of KVCache implementation. @tqchen @davidpissarra what do you guys think?

tqchen:
Yes, i think moving to mlc is OK, we can move back if there is a point of stablization.

davidpissarra:
> I think we should move all KVCache logic to MLC LLM instead, given we expect plenty of variants of KVCache implementation. @tqchen @davidpissarra what do you guys think?

Thanks @junrushao! Makes sense to move KVCache to MLC LLM

PR #16101, opened 2023-11-10
[Unity][Dlight] Choose perfect spatial factor in reduction rule

vinx13:
fixed https://github.com/mlc-ai/mlc-llm/issues/1105

cc @tqchen @junrushao @Hzfengsy

### DIFF
--- a/python/tvm/dlight/gpu/reduction.py
+++ b/python/tvm/dlight/gpu/reduction.py
@@ -99,7 +99,7 @@ class Reduction(ScheduleRule):
         if is_inner_reduction:
             self._sch_inner_reduction(sch, target, block, c_factor, epilogue)
         else:
-            self._sch_inner_spatial(sch, target, block, c_factor, epilogue)
+            self._sch_inner_spatial(sch, target, block, block_info, c_factor, epilogue)
         return sch
 
     def _normalize(  # pylint: disable=too-many-branches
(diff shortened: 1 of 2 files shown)


python/tvm/dlight/gpu/reduction.py
```diff
@@ -198,12 +198,20 @@ def _sch_inner_spatial(
...
         # pylint: disable=invalid-name
         s, r, _ = sch.get_loops(block)
         len_tx, len_ty = 16, 16
+        s_factor = [i.dom for i in block_info.iters if i.kind == "S"][-1]
+        # get perfect spatial factor, spatial factor should be divide the innermost spatial loop so
+        # that the block after r_factor and be reversed compute at the original scope
+        while len_tx > 1:
+            if s_factor % len_tx == 0:
+                break
+            len_tx -= 1
```
junrushao, line 214: Would be nice if we have a util method `_get_perfect_tile_factor` instead of adding an extra `block_info` parameter

junrushao (approved):
Great work! Thanks for addressing this issue!

Hzfengsy:
A general question: Does "perfect tiling" work for all cases? Specifically, is it the best way for those kernels with prime extents?

vinx13:
This is like a fallback solution. If the extent is prime number, thread size will be either the extent or 1. 
This is a case where the access pattern is no longer pointwise for compute_at because of imperfect tiling. We may change the behavior of compute_at or rfactor to support this

PR #16108, opened 2023-11-11
[Unity][WebGPU] Allow lower max storage buffer binding size

CharlieFRuan:
This PR lowers the required `maxStorageBufferBindingSize` from 1GB to 128MB when necessary.

Previously, we required 1GB, which led to errors when running on some devices (e.g. Android's Chrome, since Android has 128MB as its limit).

Now, if the device does not allow 1GB, we lower it to 128MB. We only throw an error when 128MB still exceeds the limit.

cc @tqchen

### DIFF
--- a/web/src/webgpu.ts
+++ b/web/src/webgpu.ts
@@ -55,13 +55,30 @@ export async function detectGPUDevice(): Promise<GPUDeviceDetectOutput | undefin
       );
     }
 
-    const requiredMaxStorageBufferBindingSize = 1 << 30;
+    let requiredMaxStorageBufferBindingSize = 1 << 30;  // 1GB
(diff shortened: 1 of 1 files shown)


CharlieFRuan:
Example log when we have something between 1GB and 128MB:

Screen Shot 2023-11-10 at 10 20 03 PM

junrushao (approved):
LGTM!

Hzfengsy:
Would be great to send to main and cherry-pick back to `unity` if possible

tqchen:
given most of the web features now functioning on unity, let us keep it here. We can bring it back to main later when we merge unity and main

CharlieFRuan:
Offloaded web-llm specific warning to web-llm in latest commit; ready to be merged!
Screen Shot 2023-11-11 at 3 34 49 PM

PR #16110, opened 2023-11-11
[CI] Refactor unittest folder

Hzfengsy:
The current unittest folder is too large and contains too many files and too many components. This PR refactors the unittest folder by moving the files to the corresponding folders.

cc @junrushao @tqchen


junrushao:
I have a [commit](https://github.com/mlc-ai/relax/commit/03e40d2fb2ddda0375aa444d15d8810036f0561a) here that does more fine-grained directory partitioning if you wanted to directly use. It's based on Unity branch, but as long as you peal off the unity part, it should be usable out-of-the-box.

junrushao (approved):

PR #16115, opened 2023-11-13
[Unity][Transform] Implement ExpandTupleArguments

Lunderberg:
Relax functions used as subroutines may receive a tuple of arguments, but only use a subset of the arguments provided.  For externally-exposed functions, this is unavoidable, as the function signature may not be updated without breaking compatibility.  For internal functions, all callsites are contained within the `IRModule`, and can be updated at the same time as the function signature.

This PR implements `ExpandTupleArguments`, which replaces `R.Tuple` parameters with separate parameters, when they occur in a private function.

### DIFF
--- a/include/tvm/relax/transform.h
+++ b/include/tvm/relax/transform.h
@@ -275,6 +275,12 @@ TVM_DLL Pass LiftTransformParams();
  */
 TVM_DLL Pass UpdateVDevice(VDevice new_vdevice, int64_t index);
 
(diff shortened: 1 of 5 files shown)


Lunderberg:
Rebased onto `unity` to keep CI results fresh.

slyubomirsky (approved):
Looks good to me. The approach is clear and it's a nice simplification to be able to make.

PR #16116, opened 2023-11-13
[Unity][Transform] Implement RemoveUnusedParameters

Lunderberg:
Relax functions used as subroutines may receive a tuple of arguments, but only use a subset of the arguments provided.  For externally-exposed functions, this is unavoidable, as the function signature may not be updated without breaking compatibility.  For internal functions, all callsites are contained within the `IRModule`, and can be updated at the same time as the function signature.

This PR implements `RemoveUnusedParameters`, which removes unused parameters from the function signature, if they occur within a private function.  Together with `ExpandTupleArguments` implemented in https://github.com/apache/tvm/pull/16115, this handles partially-exposed tuple arguments for internal subroutines.

### DIFF
--- a/include/tvm/relax/transform.h
+++ b/include/tvm/relax/transform.h
@@ -275,6 +275,12 @@ TVM_DLL Pass LiftTransformParams();
  */
 TVM_DLL Pass UpdateVDevice(VDevice new_vdevice, int64_t index);
 
+/*! \brief Expand tuple arguments to internal functions
+ *
+ * \return The Pass
+ */
+TVM_DLL Pass ExpandTupleArguments();
+
 /*! \brief Remove unused parameters to internal functions
  *
  * \return The Pass
--- a/python/tvm/relax/frontend/nn/op.py
+++ b/python/tvm/relax/frontend/nn/op.py
@@ -827,17 +827,8 @@ def gelu(x: Tensor, approximate: Optional[str] = None, name: str = "gelu") -> Te
     ----
     The input tensor is required to have float dtype
     """
-    dtype = x._expr.struct_info.dtype
     if approximate == "tanh":
-        tanh_const = rx.const(1 + np.tanh(np.sqrt(2 / np.pi)), dtype=dtype)
-        gelu_out = (
-            rx.const(0.5, dtype)
-            * x._expr
-            * (
-                tanh_const
-                * (x._expr + (rx.const(0.044715, dtype) * _op.power(x._expr, rx.const(3, "int32"))))
-            )
-        )
+        gelu_out = _op.nn.gelu_tanh(x._expr)
     else:
         gelu_out = _op.nn.gelu(x._expr)
     return _wrap_nested(gelu_out, name)
--- a/python/tvm/relax/transform/__init__.py
+++ b/python/tvm/relax/transform/__init__.py
@@ -33,6 +33,7 @@ from .transform import (
     DecomposeOpsForInference,
     DecomposeOpsForTraining,
     EliminateCommonSubexpr,
+    ExpandTupleArguments,
     FewShotTuning,
     FoldConstant,
     FunctionPass,
--- a/python/tvm/relax/transform/legalize_ops/nn.py
+++ b/python/tvm/relax/transform/legalize_ops/nn.py
@@ -302,11 +302,15 @@ def _nn_gelu(bb: BlockBuilder, call: Call) -> Expr:
 def _nn_gelu_tanh(bb: BlockBuilder, call: Call) -> Expr:
     def te_gelu_tanh(x: te.Tensor):
         dtype = x.dtype
-        return tir.const(0.5, dtype) * (
-            tir.const(1.0, dtype)
-            + topi.tanh(
-                tir.const(math.sqrt(2.0 / math.pi), dtype)
-                * (x + tir.const(0.044715, dtype) * topi.power(x, 3))
+        return (
+            tir.const(0.5, dtype)
+            * x
+            * (
+                tir.const(1.0, dtype)
+                + topi.tanh(
+                    tir.const(math.sqrt(2.0 / math.pi), dtype)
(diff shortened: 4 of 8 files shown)


src/relax/transform/remove_unused_parameters.cc
```diff
@@ -0,0 +1,239 @@
...
+#include <optional>
+#include <tuple>
+
+namespace tvm {
+namespace relax {
+
+namespace {
+
+template <typename T>
+using PSet = std::unordered_set<T, ObjectPtrHash, ObjectPtrEqual>;
```
slyubomirsky, line 35: What does the P stand for?

src/relax/transform/remove_unused_parameters.cc
```diff
@@ -0,0 +1,239 @@
...
+
+template <typename T, typename U>
+using PMap = std::unordered_map<T, U, ObjectPtrHash, ObjectPtrEqual>;
+
+struct CalleeAnalysis {
+  // The updated function
+  Function func;
+
+  // A mutator that updates calls at the call site.
+  std::function<Array<Expr>(Array<Expr>)> arg_updater;
```
slyubomirsky, line 45: The comment should probably explain how the function is meant to be used. The use of the word "mutator" can be confused for ExprMutator, so it might be better to be a little more specific.

src/relax/transform/remove_unused_parameters.cc
```diff
@@ -0,0 +1,239 @@
...
+  for (const auto& tir_var : free_tir_vars) {
+    Var relax_var("param_" + tir_var->name_hint, PrimStructInfo(tir_var));
+    params.push_back(relax_var);
+  }
+
+  FuncStructInfo new_sinfo(params.Map([](const auto& var) { return GetStructInfo(var); }),
+                           func->ret_struct_info,
+                           Downcast<FuncStructInfo>(func->struct_info_)->purity);
+
+  auto arg_updater = [parameter_mask, old_relax_params = func->params,
```
slyubomirsky, line 103: >old_relax_params = func->params

I did not know about this syntax, very cool

src/relax/transform/remove_unused_parameters.cc
```diff
@@ -0,0 +1,239 @@
...
+      }
+
+      if (callsite_updaters.empty()) {
+        return mod;
+      }
+      auto write_ptr = mod.CopyOnWrite();
+      for (const auto& it : callsite_updaters) {
+        write_ptr->Remove(it.first);
+      }
+      write_ptr->Update(new_callees);
```
slyubomirsky, line 202: Might be worth putting in a comment that this is to transfer over the functions that will be remaining unchanged. I had to think for a second to realize what this was doing.

tests/python/relax/test_transform_remove_unused_parameters.py
```diff
@@ -0,0 +1,101 @@
...
+        @R.function(private=True)
+        def func(A: R.Tensor(["m", "n"], "float32")) -> R.Tensor(["m", "n"], "float32"):
+            m = T.int64()
+            n = T.int64()
+            zeros = R.zeros(R.shape([m, n]), dtype="float32")
+            out = R.add(A, zeros)
+            return out
+
+    Expected = Before
+
```
slyubomirsky, line 98: One test case that might have interesting results would be with an inner function. I think (haven't thought it through all the way) that you might have to separately invoke the transformation on inner functions. Granted, the issue is moot if lambda lifting is used earlier.

slyubomirsky (changes requested):
Overall, I approve of the changes. I've left a couple of questions or suggestions that we might want to address before merging.

src/relax/transform/remove_unused_parameters.cc
Lunderberg: Thank you.  This one was introduced in C++14, specifically to allow a lambda function to hold a move-only type.  I like using it to reduce the amount of state held by the lambda.

src/relax/transform/remove_unused_parameters.cc
Lunderberg: Good call, and I've I added a comment.  This section is to remove any private functions that have unused parameters, then to replace them with updated versions of the private functions.

src/relax/transform/remove_unused_parameters.cc
```diff
@@ -0,0 +1,239 @@
...
+      mod.CopyOnWrite()->Update(caller_updates);
+    }
+    return mod;
+  };
+  auto inner_pass = CreateModulePass(pass_func, 0, "RemoveUnusedParametersInner", {});
+  return tvm::transform::Sequential(
+      {
+          inner_pass,
+          CanonicalizeBindings(),
+          DeadCodeElimination({}),
```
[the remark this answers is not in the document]
Lunderberg: Primarily to clean things up.  The `RemoveUnusedParameters` doesn't leave nan floating around the way `RemoveUnusedOutputs` does, so these are primarily to clean things up afterward.

That said, as I'm looking at it, these really should only be applied in cases where the functions that call into the modified private functions.  Therefore, moving these into the `CallSiteMutator`, along with a comment explaining why they are being used.

tests/python/relax/test_transform_remove_unused_parameters.py
Lunderberg: Hmm.  Currently, it only looks for callees that are represented by a `GlobalVar`, so it would ignore inner functions as they are represented by a `relax::Var`.  When an inner function is being bound, it would be recursively visited, and that would include making updates to any `Call` nodes encountered inside.

I'll need to add a test for it, but I think it would produce a well-formed output when a private function is called from within an inner function, but wouldn't remove unused parameters that are being used to call an inner function.

slyubomirsky: Ah, good point, I forgot that you were looking only for global functions, so this is okay.

slyubomirsky (approved):
Thank you for addressing my feedback. I found the additional comments to be very helpful.

PR #16117, opened 2023-11-13
[Unity][Transform] Implement RemoveUnusedOutputs

Lunderberg:
This commit implements `relax.transform.RemoveUnusedOutputs`, an extension of the current `DeadCodeElimination` pass.  If an internal callee produces multiple outputs, but only some of those outputs are used, the callee can be rewritten to only produce the outputs that are used.  This is intended to allow more aggressive DCE in the callee, as reducing the set of outputs can result in intermediate computations becoming unused.


src/relax/transform/remove_unused_outputs.cc
```diff
@@ -0,0 +1,326 @@
...
+              CHECK_EQ(usage_mask.size(), old_call_sinfo->fields.size())
+                  << "Function " << call->op << " produces " << usage_mask.size() << " outputs, "
+                  << "but " << call << " was used in a context expecting "
+                  << old_call_sinfo->fields.size() << " outputs.";
+
+              Call new_call(new_gvar, call->args);
+
+              int num_outputs_used = 0;
+              for (bool used : usage_mask) {
+                num_outputs_used += used;
```
slyubomirsky, line 248: I'm surprised you don't get a warning for this cast.

slyubomirsky (approved):
These changes seem to be sound. There are quite a few components but it's clear how everything fits together.

src/relax/transform/remove_unused_outputs.cc
```diff
@@ -0,0 +1,326 @@
...
+              << "Indices for TupleGetItem must be less than the size of the tuple, "
+              << "but expression " << GetRef<Expr>(op) << " uses a tuple index of " << op->index
+              << " for a tuple of size " << used_indices.size();
+          used_indices[index] = true;
+        }
+      }
+    }
+  }
+
+  Expr UnwrapBindings(Expr expr) const {
```
slyubomirsky, line 120: Maybe we should try to generalize some version of this logic? Similar code appears in the `FNormalize` PR.

src/relax/transform/remove_unused_outputs.cc
Lunderberg: IIRC, the usual automatic integer-conversion warnings occur from `-Wnarrowing`, which checks if there's a conversion between integer types that throws away information.  Since every value that is allowed in `bool` can be uniquely written in `int`, there is no loss of information.

Lunderberg:
> There are quite a few components but it's clear how everything fits together.

Thank you.  My goal was to make it easier to test the parts in isolation, and I'm glad that worked out.  I think it's also nice how each of the utility transforms can be beneficial in a general lowering pipeline, while still building together to handle the use case of 'partial tuple usage'.

Lunderberg:
I let this PR sit for a bit longer than intended, and am re-running the CI to make sure there haven't been conflicts introduced in the meantime.

PR #16118, opened 2023-11-13
[Unity][Transform] Keep R.ExternFunc in dead-code elimination

Lunderberg:
Prior to this commit, the `DeadCodeElimination` and `FuseTIR` passes treated `R.ExternFunc` differently with respect to removal of internal functions: `DeadCodeElimination` treated these functions as internal, to be removed if there are no internal callers, while `FuseTIR` treated these functions as external, to be retained in all cases.

Because the visibility of a function is determined by the `tvm::attr::kGlobalSymbol` attribute; which is present for `PrimFunc` and `relax::Function`, but missing for `relax::ExternFunc`; there is no information in the `IRModule` that can determine whether a `relax::ExternFunc` should be visible externally.

For consistency, and to allow the use of `DeadCodeElimination` in future refactors of `FuseTIR`, this commit updates `DeadCodeElimination` to retain instances of `relax::ExternFunc`.

### DIFF
--- a/src/relax/transform/dead_code_elimination.cc
+++ b/src/relax/transform/dead_code_elimination.cc
@@ -124,7 +124,7 @@ IRModule DeadCodeElimination(const IRModule& arg_mod, Array<runtime::String> ent
     entry_functions.insert(mod->GetGlobalVar(name));
   }
   for (const auto& [gv, func] : mod->functions) {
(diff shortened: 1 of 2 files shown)


quic-sanirudh (approved):
LGTM, thanks.

PR #16120, opened 2023-11-13
[Unity][Transform] Extract partial-tuple-usage from FuseTIR

Lunderberg:
Prior to this commit, the `FuseTIR` pass explicitly tracked usage of
tuple arguments, to minimize the set of arguments provided to each
kernel.  The additional tgracking and handling of partially-used
tuples makes it difficult to follow the primary changes being made by
`FuseTIR`.

This commit implements the same functionality in terms of the
`ExpandTupleArguments` and `RemoveUnusedParameters` transforms,
introduced in https://github.com/apache/tvm/pull/16115 and
https://github.com/apache/tvm/pull/16116 respectively.  By using these
passes before the main `FuseOps` changes, partial tuple usage is
already handled at that point.

This commit is intended to minimize any changes to user-facing
behavior, and so these pre-process passes are currently used
internally by `FuseTIR`.  This may be avoided in the future by pulling
this internal delegation out into a lowering pipeline.

### DIFF
--- a/src/relax/transform/fuse_tir.cc
+++ b/src/relax/transform/fuse_tir.cc
@@ -403,7 +403,9 @@ class FusedTIRConstructor : public ExprVisitor {
       func_info_.expr2buffers.Set(relax_param, param_buffers);
     }
 
-    // Move all scalar params after buffer params.
+    // Move all scalar params after buffer params.  To ensure that the
+    // order is deterministic and predictable for testing purposes,
+    // std::stable_sort is used instead of std::sort.
     std::stable_sort(prim_func_params.begin(), prim_func_params.end(),
                      [](const auto& a, const auto& b) {
                        bool a_is_var = a.template as<tir::VarNode>();
@@ -733,9 +735,10 @@ class FusedTIRConstructor : public ExprVisitor {
         out->push_back(Downcast<tir::Var>(var));
       }
     } else {
-      ICHECK(false) << "TypeError: The param type of PrimFunc is expected to be Tensor, Tuple or "
-                       "ShapeExpr, but got "
-                    << struct_info->GetTypeKey();
+      LOG(FATAL) << "TypeError: "
+                 << "The param type of PrimFunc is expected to be "
+                 << "Tensor, PrimValue, or ShapeExpr, "
+                 << "but got " << struct_info->GetTypeKey();
     }
   }
 
--- a/src/relax/transform/fuse_tir.cc
+++ b/src/relax/transform/fuse_tir.cc
@@ -385,58 +385,45 @@ class FusedTIRConstructor : public ExprVisitor {
       : mod_(mod), func_name_(func_name) {}
 
   void VisitExpr_(const FunctionNode* func) final {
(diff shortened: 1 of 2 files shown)


Lunderberg:
This PR is marked as a draft, until the pre-requisite PRs #16115, #16116, #16117, and #16118 land.

Lunderberg:
All pre-requisite PRs have landed, and this is now ready for review.

Lunderberg:
Rebased onto unity to avoid stale CI results.

src/relax/transform/fuse_tir.cc
```diff
@@ -719,64 +697,46 @@ class FusedTIRConstructor : public ExprVisitor {
...
-      symbolic_shape_param_started = true;
-      ICHECK(index == -1) << "TypeError: The ShapeExprNode should not be in a Tuple field.";
+      // Case 3. The relax param is a tuple of scalars, each represented as a tir var
       for (const auto& var : shape_expr->values.value()) {
         ICHECK(var->IsInstance<tir::VarNode>());
-        params.push_back(Downcast<tir::Var>(var));
+        out->push_back(Downcast<tir::Var>(var));
       }
     } else {
       ICHECK(false) << "TypeError: The param type of PrimFunc is expected to be Tensor, Tuple or "
```
slyubomirsky, line 736: Nitpick: tuples are no longer accepted, so the message is inaccurate (though a tuple would have been caught above).

src/relax/transform/fuse_tir.cc
```diff
@@ -385,58 +385,45 @@ class FusedTIRConstructor : public ExprVisitor {
...
 
-      ICHECK_EQ(params.size(), buffers.size());
-      for (size_t i = 0; i < params.size(); ++i) {
-        func_info_.buffer_map.Set(params[i], buffers[i]);
-        func_info_.params.push_back(params[i]);
+      func_info_.expr2buffers.Set(relax_param, param_buffers);
+    }
+
+    // Move all scalar params after buffer params.
+    std::stable_sort(prim_func_params.begin(), prim_func_params.end(),
```
slyubomirsky, line 407: I assume it's important for the relative ordering to be preserved (hence the stable sort), might be good to call that out.

slyubomirsky (approved):
The changes seem to make sense. I'm not fully versed on `FuseTIR` so I'm not aware of all of the broader context, but the expansion of tuples makes this code a lot a simpler and the tests all pass. I had a couple of commenting nitpicks to make, but they shouldn't be blockers. Great work.

src/relax/transform/fuse_tir.cc
Lunderberg: Good catch, and updated to remove `Tuple` from the list, and add `PrimValue` as it is now handled.

src/relax/transform/fuse_tir.cc
Lunderberg: Good call, and updated the comment.  It's mainly to make sure that the order is consistent and predictable for unit tests, not for any correctness of the output.

PR #16121, opened 2023-11-14
[Unity] Allow Customized Pipeline in `relax.build`

junrushao:
The existing `relax.build` method assumes the compilation follows a fixed set of passes to be used. With the introduction of Relax pipeline system, one could effectively manage which passes to use during lowering. This PR generalizes this approach by further allowing what to use during compilation.

### DIFF
--- a/python/tvm/relax/vm_build.py
+++ b/python/tvm/relax/vm_build.py
@@ -16,13 +16,11 @@
 # under the License.
 # pylint: disable=invalid-name, no-member
 """VM build logics"""
-from typing import List, Optional, Union, Dict, Any
+from typing import Any, Dict, List, Optional, Union
 
 import tvm
 from tvm import relax
-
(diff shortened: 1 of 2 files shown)


jinhongyii (approved):

python/tvm/relax/vm_build.py
```diff
@@ -249,6 +248,7 @@ def build(
     mod: tvm.IRModule,
     target: Union[str, tvm.target.Target],
     params: Optional[Dict[str, list]] = None,
+    pipeline: str = "default",
```
Hzfengsy, line 251: I appreciate this PR's work, but I am still concerned about setting the "default" pipeline to the `build` API. 
As we said we'd like to keep it a "minimal" build. It would be great to set the default value to None, while give more time for the community to discuss about the default pipeline in `build` API

Hzfengsy (approved):

PR #16123, opened 2023-11-14
[Relay][Pytorch] Add support for `aten::linalg_vector_norm`

### LINKED ISSUE #16096: [FRONTEND][PyTorch] NotImplementedError: The following operators are not implemented: ['aten::linalg_vector_norm']
### Steps to reproduce

```python
from tvm import relay
import torch

def func(x):
  return torch.linalg.vector_norm(x)

x = torch.rand(3, 3)
scripted_func = torch.jit.trace(func, x)
print(scripted_func(x))

input_infos = [('x', (x.shape, 'float32'))]
mod, params = relay.frontend.from_pytorch(scripted_func, input_infos)
```

### Triage

Please refer to the list of label tags [here](https://github.com/apache/tvm/wiki/Issue-Triage-Labels) to find the relevant tags and add them below in a bullet format (example below).

* frontend:pytorch

cc @shingjan @yelite

mshr-h:
Fix #16096 
Support [torch.linalg.vector_norm](https://pytorch.org/docs/stable/generated/torch.linalg.vector_norm.html).

cc @jikechao @vvchernov @Hzfengsy @junrushao

### DIFF
--- a/python/tvm/relay/frontend/pytorch.py
+++ b/python/tvm/relay/frontend/pytorch.py
@@ -3851,6 +3851,8 @@ class PyTorchOpConverter:
         dim = inputs[2]
         keepdim = inputs[3]
 
+        assert dtype == "float32" or dtype == "float64"
+
         if ord == 0:
             return _op.reduce.sum(
                 _op.cast(_op.not_equal(data, _expr.const(0, dtype=dtype)), dtype=dtype),
--- a/python/tvm/relay/frontend/pytorch.py
+++ b/python/tvm/relay/frontend/pytorch.py
@@ -3844,6 +3844,30 @@ class PyTorchOpConverter:
         # Return
         return _op.scatter_nd(source, indices, values, mode)
 
+    def linalg_vector_norm(self, inputs, input_types):
+        data = inputs[0]
+        dtype = input_types[0]
+        ord = inputs[1]
+        dim = inputs[2]
+        keepdim = inputs[3]
+
+        if ord == 0:
+            return _op.reduce.sum(
+                _op.cast(_op.not_equal(data, _expr.const(0, dtype=dtype)), dtype=dtype),
+                axis=dim,
+                keepdims=keepdim,
+            )
+        elif ord == np.inf:
+            return _op.reduce.max(_op.abs(data), axis=dim, keepdims=keepdim)
+        elif ord == np.NINF:
+            return _op.reduce.min(_op.abs(data), axis=dim, keepdims=keepdim)
+        reci_ord = _expr.const(1.0 / ord, dtype=dtype)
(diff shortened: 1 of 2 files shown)


python/tvm/relay/frontend/pytorch.py
```diff
@@ -3844,6 +3844,30 @@ def inplace_copy(self, inputs, input_types):
         # Return
         return _op.scatter_nd(source, indices, values, mode)
 
+    def linalg_vector_norm(self, inputs, input_types):
+        data = inputs[0]
```
vvchernov, line 3848: It looks like this method is based on torch.linalg.vector_norm. The latter assums that input data type is float, double or complex one; dtype also should be real or complex. Would you check it?

vvchernov (approved):
Thank you for your work! LGTM, see my small comment

python/tvm/relay/frontend/pytorch.py
mshr-h: Thanks for your review. That's true. It's based on torch.linalg.vector_norm and it supports complex input data. Seems like the PyTorch frontend doesn't support complex data type. [_convert_data_type](https://github.com/mshr-h/tvm/blob/f2e6107b56bb921306f4969ff03251e13686e4c1/python/tvm/relay/frontend/pytorch.py#L4512)
Should I add complex data type support?

vvchernov: Hello @mshr-h! My idea was to add something like `assert data.dtype == float or data.dtype == double`. And may be add TODO for further support complex values, but I do not think it is needed just now

mshr-h: Hi @vvchernov ! I've added assertion and testcases for double-precision input data.

tests/python/frontend/pytorch/test_forward.py
```diff
@@ -5432,6 +5430,46 @@ def forward(self, *args):
...
+        def forward(self, x):
+            return torch.linalg.vector_norm(x, ord=np.inf)
+
+    class VectorNorm4(torch.nn.Module):
+        def forward(self, x):
+            return torch.linalg.vector_norm(x, ord=-np.NINF)
+
+    class VectorNorm5(torch.nn.Module):
+        def forward(self, x):
+            return torch.linalg.vector_norm(x, ord=0)
```
masahi, line 5456: if you only need different `ord` for each test case, you don't need to create a different class for each. Please clean them up.

mshr-h: @masahi Revised as suggested. Please review. Thanks!

masahi (approved):

PR #16128, opened 2023-11-14
[Unity][MSC] Enable add attributes while fuse ops

Archermmt:

To enable set custom attributes while fusing ops, attrs_getter is added to set extra attributes for functions.

cc @Hzfengsy


Hzfengsy (approved):

Hzfengsy:
cc @masahi @junrushao to see if you have objections

include/tvm/relax/transform.h
```diff
@@ -322,11 +322,20 @@ class FusionPatternNode : public Object {
    */
   Optional<PackedFunc> check;
 
+  /*!
+   * \brief The function to get attributes for fused function
+   *
+   * It should have signature
+   * Map<String, String>(const PatternCheckContext& context)
```
masahi, line 329: At first glance I don't understand why it needs to take `PatternCheckContext` as argument.

python/tvm/relax/transform/transform.py
```diff
@@ -647,19 +647,21 @@ class FusionPattern(Object):
...
     check: Callable[[PatternCheckContext], bool]
+    attrs_getter: Callable[[PatternCheckContext], Dict[str, str]]
 
     def __init__(
         self,
         name: str,
         pattern: DFPattern,
         annotation_patterns: Optional[Mapping[str, DFPattern]] = None,
         check: Optional[Callable[[Mapping[str, Expr]], bool]] = None,
+        attrs_getter: Optional[Callable[[Mapping[str, Expr]], Dict[str, str]]] = None,
```
masahi, line 658: Please use the same type annotation.

src/relax/transform/fuse_ops.cc
```diff
@@ -1102,6 +1106,13 @@ class PatternBasedPartitioner : ExprVisitor {
       auto parent_group = GetGroupForBoundVar(binding->var);
       ICHECK(parent_group);
       parent_group->attrs.Set(attr::kComposite, pat_name_);
+      if (attrs_getter_ != nullptr) {
+        const auto& custom_attrs =
+            attrs_getter_(CreatePatternCheckContext(call, matches_opt.value()));
```
masahi, line 1111: A dup call to `CreatePatternCheckContext`

https://github.com/apache/tvm/blob/c3f6c07021d0b857d4340e7699f14dceba17b228/src/relax/transform/fuse_ops.cc#L1077

masahi (changes requested):
Please add a test case, I don't understand its intended usage.

include/tvm/relax/transform.h
Archermmt: A context is needed to set extra attributes. The context should have access to all expressions in the pattern, I find out that PatternCheckContext is a good one. It contains all the expressions marked during pattern matching. Some attributes need to be defined based on the expressions

python/tvm/relax/transform/transform.py
Archermmt: Do you mean like: attrs_getter: Optional[Callable[[Mapping[str, Expr]], bool]] = None ? But the attrs_getter return a map of <str, str>, each element contains a attribute for the function. The attrs_getter and check has same argument, thus PatternCheckContext, but different return type.

src/relax/transform/fuse_ops.cc
Archermmt: Emm..... that can be simplified. It can be changed to: call the CreatePatternCheckContext once and use in both check && attrs_getter. That depends on comment 1, if PatternCheckContext can be used as the context for attrs_getter, then I'll change this part.

python/tvm/relax/transform/transform.py
masahi: I mean, I was confused about the type annot difference between L650 and L658.

src/relax/transform/fuse_ops.cc
masahi: Without seeing what `attrs_getter` does, it's hard to say if `CreatePatternCheckContext` is a reasonable source to query attributes from. That's why I'm asking for a test case.

Archermmt: And I'll update the test case when the discussion about "context" is done. Currently adding the attributes is mainly for TensorRT translation. An "id_name" should be used to mark same expression in relax and TensorRT, so that each "layer" in the model can share their optimization strategy. The optimization here mainly refers to pruning, sparse and quantization.

masahi: For example, I think you can recover the same information in the loop 
```
      for (const auto& [pat, match] : matches_opt.value()) {
```
using the `binding` argument to `VisitBinding_`.

include/tvm/relax/transform.h
masahi: What `CreatePatternCheckContext` does is very simple. I suggest extracting what you need from there into a function, and use it in both here and inside `CreatePatternCheckContext`.

python/tvm/relax/transform/transform.py
Archermmt: I just copied from L649 and L657 and change return type ... from the check. I also don't know why different annotation is used for check 🤣...

src/relax/transform/fuse_ops.cc
Archermmt: Em.... Let me add the test case to show why these attributes are needed

python/tvm/relax/transform/transform.py
masahi: I see, let's fix them both.

python/tvm/contrib/msc/core/transform/pattern.py
```diff
@@ -26,6 +27,27 @@
...
+
+    Returns
+    -------
+    attrs: dict<str,str>
+        The extra attributes for msc.
+    """
+
+    fused_cnt = MSCMap.get(MSCKey.FUSED_CNT, 0)
+    unique_name = "msc_fused_" + str(fused_cnt)
+    if anchor in context.annotated_expr:
```
masahi, line 45: So what you really need from `PatternCheckContext` is just `annotated_expr`. Please follow my comment in https://github.com/apache/tvm/pull/16128#discussion_r1396423174. The getter function can just ingest `Map<String, Expr>` rather than `PatternCheckContext

Archermmt: yes, the thing I need is the annotated_expr. I'll check your comment

include/tvm/relax/transform.h
masahi: @Archermmt Please update the function signature above.

masahi (approved):

Archermmt:
@masahi tks for the comments. I've changed the signature for attrs_getter.

PR #16129, opened 2023-11-14
[Unity][Transform] Replace eligible operators with in-place versions in dataflow blocks

slyubomirsky:
Pursuant to issue #15319, this pass implements an in-place transformation for dataflow blocks. For each operator call in the dataflow block, the pass if at least one argument is not live past the call and has no aliases that are live past the call. If this condition is met for at least one argument, the operator can be replaced with an in-place version called via `call_tir_inplace`. The in-place version is a TIR `PrimFunc` that is produced by taking the legalizer for that operator and replacing its output with one of the inputs.

The liveness analysis is very simple since we are handling only single dataflow blocks. Handling general Relax control flow would require a solution more akin to #15689. The alias analysis, by contrast, is more complex since alias analysis is generally a tricky issue. It focuses on being able to handle common cases (especially tuples).


tests/python/relax/test_dataflow_inplace.py
```diff
@@ -0,0 +1,462 @@
...
+    y = tvm.nd.array(np.random.rand(1, 3).astype("float32"))
+    expected = np.zeros((2, 3), dtype="float32")
+
+    target = tvm.target.Target("llvm")
+    ex = relax.build(new_mod, target)
+    vm = relax.VirtualMachine(ex, tvm.cpu())
+    res = vm["main"](x, y)
+    # due to reuse of buffers, the result is actually reference equal to argument x
+    # (we can disable this by setting the arguments to "unknown value" in the alias analysis)
+    assert res == x
```
slyubomirsky, line 457: Mutating arguments might not be something we would want to permit, so this is something we may want to prohibit. I think setting arguments to `-1` in the alias analysis might lead to too many rejections, so it might be better to just treat them as a special case. I welcome other thoughts on this.

slyubomirsky: Changed to prevent this behavior. I've left in the code for specifying inputs that are safe to assume are not aliased and therefore can be freely overwritten since we may need it in the future (it was suggested at today's Relax Open Development meeting that ownership semantics might let us determine that, but we'd need to track that).

src/relax/transform/dataflow_inplace.cc
```diff
@@ -0,0 +1,1017 @@
...
+    for (auto var : remove) {
+      currently_live.erase(var);
+    }
+  }
+
+  return {size_match_list, exact_match_list};
+}
+
+// Replace buffers in a PrimFunc according to the mapping.
+tir::Stmt RemapBuffers(const tir::Stmt& stmt, const Map<tir::Buffer, tir::Buffer>& buffer_map) {
```
Lunderberg, line 663: I think this could be handled by calling [`tir::Specialize`](https://github.com/apache/tvm/blob/main/include/tvm/tir/function.h#L267).

src/relax/transform/dataflow_inplace.cc
```diff
@@ -0,0 +1,1017 @@
...
+      current_match_index++;
+    }
+    return DataflowBlock(new_bindings, block->span);
+  }
+
+  // Given the call and indices of arguments that could be done in-place,
+  // replace the call with a call to an in-place PrimFunc.
+  // (Made public for testing.)
+  Call CreateInplaceCall(const Call& call, const Array<Integer>& inplace_indices) {
+    static const auto& legalize_map = Op::GetAttrMap<FLegalize>("FLegalize");
```
Lunderberg, line 832: Should this use a separate legalization map?  Even if they map to the same function, the per-operator support could then be checked by seeing if `"FLegalizeInPlace"` is defined for an operator.

tests/python/relax/test_dataflow_inplace.py
```diff
@@ -0,0 +1,464 @@
...
+                q = R.multiply(z, y)
+                p = R.add(z, q)
+                n = R.multiply(p, p)
+                R.output(n)
+            return n
+
+    block = BasicLiveness["main"].body.blocks[0]
+    live_ranges = dataflow_liveness_analysis(block)
+    expected_ranges = {
+        "x": (-1, 1),
```
Lunderberg, line 50: I think `"x"` should have a liveness range of `(-1,5)`, because local analysis cannot determine whether `x` is needed again within the calling scope of the function.  Rather than tracking the function parameters separate from the liveness, this would track the function parameters as being live for the entire duration.  That would avoid potential pitfalls in the future, as the liveness results would be the only thing required to track.

tests/python/relax/test_dataflow_inplace.py
```diff
@@ -0,0 +1,464 @@
...
+                a = R.add(z, y)  # this one can be done in-place
+                q = R.multiply(a, y)  # broadcast again, a is eligible
+                r = R.subtract(y, y)  # cannot be done in-place because y is an argument
+                s = R.subtract(r, r)  # No broadcast. Can be done in-place
+                m = R.multiply(q, s)  # should give us all zeros
+                R.output(m)
+            return m
+
+    transform_pass = DataflowUseInplaceCalls()
+    new_mod = transform_pass(EndToEndTest)
```
Lunderberg, line 439: Can we add an expected output as well, with `assert_structural_equal`?

tests/python/relax/test_dataflow_inplace.py
```diff
@@ -0,0 +1,464 @@
...
+    x = tvm.nd.array(np.random.rand(2, 3).astype("float32"))
+    y = tvm.nd.array(np.random.rand(1, 3).astype("float32"))
+    expected = np.zeros((2, 3), dtype="float32")
+
+    target = tvm.target.Target("llvm")
+    ex = relax.build(new_mod, target)
+    vm = relax.VirtualMachine(ex, tvm.cpu())
+    res = vm["main"](x, y)
+    assert (expected == res.numpy()).all()
+
```
Lunderberg, line 461: Do we want a `pytest.xfail` case for internal functions, to indicate that they are not yet supported?  I'm thinking of cases like the following, where the current local analysis would conclude that `Module.func` cannot do in-place operations, but non-local analysis could perform `Module.func` in-place for some cases, depending on the liveness analysis in the calling scope.

```python
@I.ir_module
class Module:
    @R.function
    def main(x: R.Tensor((), "int32")) -> R.Tensor((), "int32"):
        cls = Module
        with R.dataflow():
            # Cannot be in-place, uses a function argument
            y = cls.subroutine(x)
            
            # Can be in-place, uses a local binding
            z = cls.subroutine(y)
            R.output(z)
        return z
    
    @R.function(private=True)
    def subroutine(x: R.Tensor((), "int32"), y: R.Tensor((), "int32")) -> R.Tensor((), "int32"):
        with R.dataflow():
            z = R.add(x,R.const(1, "int32"))
            R.output(z)
        return z
```

python/tvm/relax/testing/transform.py
```diff
@@ -128,3 +129,62 @@ def transform(self):
 def ApplyEmptyCppMutator() -> tvm.ir.transform.Pass:
     packed_func = tvm.get_global_func("relax.testing.transform.ApplyEmptyCppMutator")
     return packed_func()
+
+
+def dataflow_liveness_analysis(block: DataflowBlock) -> Dict[Var, Tuple[int, int]]:
+    """
+    Inner function for the dataflow inplace transformation exposed for testing.
+    """
+    live_ranges = tvm.get_global_func("relax.testing.transform.DataflowLivenessAnalysis")(
```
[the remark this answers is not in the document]
slyubomirsky: Good idea, that's a handy feature.

tests/python/relax/test_dataflow_inplace.py
```diff
@@ -0,0 +1,464 @@
...
+    live_ranges = dataflow_liveness_analysis(block)
+    expected_ranges = {
+        "x": (-1, 1),
+        "y": (0, 2),
+        "z": (1, 3),
+        "q": (2, 3),
+        "p": (3, 4),
+        "n": (4, 5),
+    }
+    for var, live_range in live_ranges.items():
```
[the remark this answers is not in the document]
slyubomirsky: Good observation.

src/relax/transform/dataflow_inplace.cc
```diff
@@ -0,0 +1,1017 @@
...
+  }
+
+  // Made public for testing.
+  IRModule CurrentMod() { return builder_->GetContextIRModule(); }
+
+ private:
+  const IRModule& mod_;
+  Array<GlobalVar>
+      legalizers_added;    // Keep track of legalizers we add so we can clean up at the end.
+  Array<Var> func_params;  // The current function's params will be treated as non-aliased
```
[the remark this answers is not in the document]
slyubomirsky: Yeah, I might have to update the comment. Right now, we treat arguments as sacrosanct. In the future, if we consider any kind of ownership semantics, we might consider situations where it is okay to overwrite an argument, but we're not doing it now.

tests/python/relax/test_dataflow_inplace.py
```diff
@@ -0,0 +1,464 @@
...
+
+import numpy as np
+
+
+def test_liveness_analysis():
+    @I.ir_module
+    class BasicLiveness:
+        @R.function
+        def main(x: R.Tensor((), "int32")) -> R.Tensor((), "int32"):
+            with R.dataflow():
```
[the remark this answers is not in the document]
slyubomirsky: Good question. Yes, the analysis depends on dataflow blocks and makes use of the lack of side effects in particular. The transformation won't do anything unless the analysis finds eligible operators to rewrite.

src/relax/transform/dataflow_inplace.cc
```diff
@@ -0,0 +1,1017 @@
...
+                                                        "relax.nn.silu",  "relax.nn.relu"};
+bool OpSupportsInplace(const Op& op) { return SUPPORTED_OPS.count(op->name); }
+
+// Check for in-place eligibility:
+//  1. see if there's an arg big enough to hold the result
+//  2. see if the arg is live past the call
+//  3. see if the arg has an alias that's live past the call
+// If the conditions are met, record the index of that binding.
+// Returns two lists of lists:
+// 1. A list of bindings where at least one argument meets the in-place conditions and the *size*
```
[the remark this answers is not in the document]
slyubomirsky: It would probably be neater to have a data structure to encode that info, yes. I didn't use one to make things easier for the FFI but that's not a very good reason. I'll use this suggestion.

src/relax/transform/dataflow_inplace.cc
```diff
@@ -0,0 +1,1017 @@
...
+
+    return ret;
+  }
+
+  std::unordered_map<Var, std::unordered_set<int>, ObjectPtrHash, ObjectPtrEqual> alias_map_;
+  std::unordered_map<int, std::vector<std::unordered_set<int>>> tuple_map_;
+  int mem_idx_;
+};
+
+// given a shape, return the allocation size corresponding to it (product of elements)
```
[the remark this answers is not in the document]
slyubomirsky: I hadn't thought about that, good point. That's a good distinction to make.

src/relax/transform/dataflow_inplace.cc
```diff
@@ -0,0 +1,1017 @@
...
+int ShapeSize(const ShapeExpr& shape) {
```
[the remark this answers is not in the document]
slyubomirsky: Yeah, we could do this and later check if the final result is a concrete value. Good idea.

src/relax/transform/dataflow_inplace.cc
```diff
@@ -0,0 +1,1017 @@
...
+    if (target_tensor->shape.defined() && target_tensor->shape.as<ShapeExprNode>() &&
+        arg_tensor->shape.defined() && arg_tensor->shape.as<ShapeExprNode>()) {
+      if (target_tensor->dtype != arg_tensor->dtype) {
+        return {false, false};
+      }
+      auto target_shape = Downcast<ShapeExpr>(target_tensor->shape);
+      auto arg_shape = Downcast<ShapeExpr>(arg_tensor->shape);
+      int target_size = ShapeSize(target_shape);
+      int arg_size = ShapeSize(arg_shape);
+      if (target_size == -1 || arg_size == -1 || target_size < arg_size) {
```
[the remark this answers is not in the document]
slyubomirsky: I don't think the operator implementations I'm using here would work for dynamically shaped tensors anyway, though I can check that. In the dynamic case, I guess we can have the operator implementation check whether the (dynamic) sizes match.

src/relax/transform/dataflow_inplace.cc
```diff
@@ -0,0 +1,1017 @@
...
+    if (target_tup->fields.size() != arg_tup->fields.size()) {
+      return {false, false};
+    }
+    bool all_exact = true;
+    for (size_t i = 0; i < target_tup->fields.size(); i++) {
+      auto element_match = SizeMatches(target_tup->fields[i], arg_tup->fields[i]);
+      if (!element_match.first) {
+        return {false, false};
+      }
+      if (!element_match.second) {
```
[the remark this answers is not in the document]
slyubomirsky: Good catch, I always prefer that style.

src/relax/transform/dataflow_inplace.cc
```diff
@@ -0,0 +1,1017 @@
...
+        all_exact = false;
+      }
+    }
+    return {true, all_exact};
+  } else if (target_info.as<PrimStructInfoNode>() && arg_info.as<PrimStructInfoNode>()) {
```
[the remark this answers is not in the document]
slyubomirsky: Good point.

src/relax/transform/dataflow_inplace.cc
slyubomirsky: Ooh! I didn't know about it

src/relax/transform/dataflow_inplace.cc
slyubomirsky: That's worth thinking about. I didn't use it here because the operators that can be done in-place can also be implemented by applying the transformation given here to their legalized version. @tqchen, @MasterJH5574 (and others), do you think it would be worthwhile to add another map for in-place legalizations?

tests/python/relax/test_dataflow_inplace.py
slyubomirsky: That's a good point, I hadn't thought about it. Values that come from outside the DF block should be considered to be live for the whole thing anyway since we're not checking if they're used again.

tests/python/relax/test_dataflow_inplace.py
slyubomirsky: Sure, will do.

tests/python/relax/test_dataflow_inplace.py
slyubomirsky: I'm always in favor of testing error conditions, so I'll add the test case too.

src/relax/transform/dataflow_inplace.cc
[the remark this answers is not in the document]
Lunderberg: Yeah, good point on the FFI.  I think it's probably worth it for the readability, but the amount of bookkeeping necessary to expose an object through the FFI is more than I'd like.  (And duplicated between the C++ and Python sides.)  At some point, I'd like to take a look at how pybind11 exposes the C++ bindings and see if something similar could be adapted for TVM, as I like the way it can expose C++ types without requiring bindings on the python side.

src/relax/transform/dataflow_inplace.cc
Lunderberg: It's a nice one for a lot of signature manipulations.  There may need to be some post-processing, but it should work in this case.  If the selected in-place input is parameter `i`, and the output is parameter `j`, then it would be called as `auto new_func = Specialize(func, {{func->params[j], func->buffer_map[func->params[i]]}})`.

src/relax/transform/dataflow_inplace.cc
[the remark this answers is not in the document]
Lunderberg: If it's based on `FLegalize`, then I think the dynamic kernels would be generated as needed.  They get translated to TE expressions, then into TIR, with any symbolic variables propagated along.

tests/python/relax/test_dataflow_inplace.py
Lunderberg: Sounds good!  Though, I'd probably see this one as a not-yet-supported test case (`pytest.mark.xfail`), rather than a test of an error case (`with pytest.raises(...)`).  Rather than verifying that a specific error is raised, it would document what behavior isn't covered by the current implementation.

src/relax/transform/dataflow_inplace.cc
slyubomirsky: I would like for a pass like this to come later in compilation, but the issue is the fact it relies on working within single dataflow blocks, so it has to come before we get rid of dataflow blocks. That is also causing me a few issues with thinking about how to handle in-place split and concat.

src/relax/transform/dataflow_inplace.cc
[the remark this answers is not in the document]
slyubomirsky: Actually I think it's neater just to handle that in the tuple case. We can skip any other members that are not tuples or tensors themselves

src/relax/transform/dataflow_inplace.cc
slyubomirsky: Hm, using `tir::Specialize` still left some uses of the old output buffer around, which still seems like it would require the manual rewrite I had. It was odd. Namely, the old buffer was being used in the `writes` field of the block.

src/relax/transform/dataflow_inplace.cc
```diff
@@ -0,0 +1,1017 @@
...
+            ret.insert(member_set.begin(), member_set.end());
+          }
+        }
+      }
+    } else if (auto* call_node = value.as<CallNode>()) {
+      if (auto* op_node = call_node->op.as<OpNode>()) {
+        // call_pure_packed: treat as non-op call
+        if (op_node->name == "relax.call_pure_packed") {
+          return HandleMysteryCall(call_node, bound_var, true);
+        } else if (op_node->name == "relax.split") {
```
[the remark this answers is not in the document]
slyubomirsky: Good idea, I don't think there's a reason to single out split at this point (I am working on supporting in-place split but that's a separate matter anyway)

src/relax/transform/dataflow_inplace.cc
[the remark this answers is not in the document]
slyubomirsky: I will look into supporting the dynamic case, then

tests/python/relax/test_dataflow_inplace.py
slyubomirsky: On second thought, I'm not sure we should add an error-condition test case for something we don't even claim to support (the description of the pass does not reference trying to do function calls in place).

tests/python/relax/test_dataflow_inplace.py
slyubomirsky: Lol, in implementing this I found a pretty-printer roundtripping bug related to `call_tir_inplace` so I'll fix that too.

src/relax/transform/dataflow_inplace.cc
```diff
@@ -0,0 +1,1017 @@
...
+    // function/packed call: chaos reigns, alias with any other argument
+    //   (if tuple is passed, assume also aliased with all members of the tuple)
+    // tuple index: -1 if tuple is not in tuple map, otherwise look up corresponding entry
+    // function constant: give them a fresh index (TODO: we can handle in more detail if this is a
+    // case we need to support) prim value: fresh index if node: should not happen inside dataflow
+    // block
+    if (value.as<ConstantNode>() || value.as<PrimValueNode>() || value.as<FunctionNode>()) {
+      // TODO(@slyubomirsky): We will probably want special handling for closures
+      ret.insert(get_fresh_idx());
+    } else if (auto* target_var_node = value.as<VarNode>()) {
```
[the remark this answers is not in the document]
slyubomirsky: I think this is doing a lot more than what `CanonicalizeBindings` is doing; I don't see an obvious correspondence between them. I'm all for deduplicating if it makes sense, though.

tests/python/relax/test_dataflow_inplace.py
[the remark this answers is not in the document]
slyubomirsky: To clarify, in particular the reason this relies on DF blocks is to be sure that there is no control flow to deal with. Dealing with control flow would require more sophisticated machinery like in my earlier PR https://github.com/apache/tvm/pull/15689 (we can revisit it if we do want to have this be independent of DF blocks)

src/relax/transform/dataflow_inplace.cc
```diff
@@ -0,0 +1,1038 @@
...
+tvm::transform::Pass DataflowUseInplaceCalls() {
+  auto inplace_pass = tvm::transform::CreateModulePass(
+      [](const IRModule& mod, const PassContext& ctx) -> IRModule {
+        ModuleInplaceTransformer transformer(mod);
+        return transformer.Transform();
+      },
+      0, "DataflowInsertInPlaceCalls", {}, false);
+  // odd quirk: if Normalize is not explicitly called, then the function
+  // StructInfo will not be properly updated
+  return tvm::transform::Sequential({inplace_pass, Normalize()});
```
slyubomirsky, line 1006: Not sure why this happened. I assumed it was because the `BlockBuilder` was not renormalizing when doing an update, but called `BlockBuilder::Normalize()` before updating the function also did not fix this issue.

slyubomirsky: Maybe this can be avoided if I use `BlockBuilder` methods throughout the transformation. I will see if I can write it that way.

src/relax/transform/dataflow_inplace.cc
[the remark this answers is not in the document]
Lunderberg: Good point on the different scope of analysis.  I think there is a correspondence between aliased tensors and trivial bindings, since the tensor alias would be created by a trivial binding, but it might be more effort than its worth to de-dup at the moment.

src/relax/transform/dataflow_inplace.cc
Lunderberg: This is probably a larger change, but would it make sense to either (A) implement it in terms of bindings without requiring a dataflow block, or (B) implement the pass as `Sequential(ConvertToDataFlow(), current_pass(), ToNonDataflow())`?

(B) would be easier to implement, but the use of `ToNonDataflow` as a post-processor would require that it be after any passes that require dataflows to be present.

tests/python/relax/test_dataflow_inplace.py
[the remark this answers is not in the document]
Lunderberg: Sounds good.  I don't think the application to non-dataflow blocks is strictly necessary, but wanted to check.

tests/python/relax/test_dataflow_inplace.py
Lunderberg: I could go either way on it.  The high-level name `DataflowUseInplaceCalls` doesn't specify that the `InplaceCalls` only refer to native relax operators, and excludes potentially-in-place calls to other relax functions.  The docstring does specify operators, so you're right that we don't claim support for it there.

I'd see a `@pytest.mark.xfail(reason="Not currently supported")` as a way to explicitly disclaim responsibility for that use case.  It would indicate what that support would look like, and that it could be supported at some point, but that it isn't currently supported.

src/relax/transform/dataflow_inplace.cc
Lunderberg: As another option, it may be easier to rely on the `BlockBuilder` usage in the base `ExprMutator` class. Looks like there are two parts that would need to be updated.

1. Normalization of bindings within `DataflowBlock`
    a. The `VisitBindingBlock_(const DataflowBlockNode* op)` would need perform the analysis, save the result in a member variable, then delegate to `ExprMutator::VisitBindingBlock_`.
    b. The updates would need to occur in `Expr VisitExpr_(const CallNode* op)`.  It would check whether there is a replacement planned in the member variable from (a), returning either the updated object or the original.  Normalization would occur within the base class `ExprMutator::VisitBinding_(const VarBinding*, const CallNode*)`.
    
2. Normalization of `relax::Function`.

   ```c++
    // Instead of
    auto function = GetRef<Function>(func_node);                                       
    auto* function_cow = function.CopyOnWrite();                                       
    auto new_body = VisitExpr(function->body);                                         
                       
    // Replace with                                                               
    auto function = Downcast<Function>(VisitExpr(GetRef<Function>(func_node)));
```
     The copy-on-write can be handled internally by `ExprMutator`, if any mutations occur.

slyubomirsky:
Thank you for drawing my attention to the dynamic case, @Lunderberg, I would not have guessed that it would "just work" like that :smile:

src/relax/transform/dataflow_inplace.cc
slyubomirsky: I tried using `BlockBuilder` methods to build up the new dataflow block and still had an error if I didn't have the explicit normalization in here. I'm not sure what exactly is going wrong. Specifically, the error had to do with the `StructInfo` of the `SeqExpr`'s body, but calling normalize on the function's body only did not solve the problem. 

@yongwww, you mentioned that you've encountered an issue like this before. Do you know what could be going wrong here?

src/relax/transform/dataflow_inplace.cc
slyubomirsky: If we don't want to have the dataflow requirement, we would also need to handle control flow. That's the main issue and there's no obvious way to deal with control flow without going with a solution more like abstract interpretation (see my comment about the general-purpose liveness analysis PR).

src/relax/transform/dataflow_inplace.cc
slyubomirsky: There was no DeclBuffer, there was just a loose variable hanging around.

src/relax/transform/dataflow_inplace.cc
slyubomirsky: Thank you, @Lunderberg, I had tried step 1 but not also step 2. Using the function node visitor fixed everything. There are lots of invariants in `ExprMutator` to know about!

src/relax/transform/dataflow_inplace.cc
slyubomirsky: Using the sequential transformation would be a less ad hoc version of the above.

tests/python/relax/test_dataflow_inplace.py
slyubomirsky: I can appreciate your point of view here, but I would point out that while tests are a form of documentation, documentation (such as doc comments) is also a form of documentation :slightly_smiling_face: 

If there is a strong need to reinforce the point that we are only looking for a small number of operators, we can do this, but I don't think there is right now, especially since handling Relax functions in-place is not something we have even planned on doing. (Possible in principle, but it's not on the horizon.)

src/relax/transform/dataflow_inplace.cc
Lunderberg: Got it.  In that case, it would fall into the same bug that #14565 was intended to resolve.  I think for now, it's probably better to use the separate mutator.  If either RFC#70 gets fully implemented, or if the relax-to-TIR conversion provides a `DeclBuffer` annotation, then it could be updated.

src/relax/transform/dataflow_inplace.cc
Lunderberg: Good call, and agreed that the sequential transform would be the better way to handle it, and a more consistent pattern that could be applied to other passes.

src/relax/transform/dataflow_inplace.cc
Lunderberg: No problem!  I think it's a bit tricky with how the `relax::ExprMutator` exposes a similar interface to the `tir::ExprMutator`, but can have extra requirements in some cases.  I think it's helpful to have the automatic normalization and struct info inference, but it does seem a bit more fragile.

My personal rule of thumb is to implement everything in terms of `VisitVarDef` and `VisitExpr_` overrides whenever possible, as those don't include any interactions with the `block_builder_` in the base `ExprMutator` implementation.

tests/python/relax/test_dataflow_inplace.py
Lunderberg: Good point, and mostly just nitpicks on my side.  While I think documenting potential future expansions is useful, I don't think it's worth delaying this PR.  (Approved!)

PR #16130, opened 2023-11-14
[Unity] Support constant args in `nn.ExternModule`

cyx-6:
This PR introduces `spec.ConstInt`, `spec.ConstFloat` and `spec.ConstString` for the `nn.ExternModule` args. So we will enable the constant input for extern module functions.

### DIFF
--- a/tests/python/relax/test_frontend_nn_extern_module.py
+++ b/tests/python/relax/test_frontend_nn_extern_module.py
@@ -253,6 +253,4 @@ def test_extern_spec():
 
 
 if __name__ == "__main__":
-    test_extern_module()
-    test_extern_spec()
-    # tvm.testing.main()
+    tvm.testing.main()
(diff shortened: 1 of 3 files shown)


tests/python/relax/test_frontend_nn_extern_module.py
```diff
@@ -191,6 +191,68 @@ def test_sym(
...
+            },
+        }
+    )
+    tvm.ir.assert_structural_equal(ir_module, ExpectedModule)
+
+
 if __name__ == "__main__":
     test_extern_module()
+    test_extern_spec()
     # tvm.testing.main()
```
junrushao, line 258: let's recover this line

cyx-6: fixed

junrushao (approved):

PR #16131, opened 2023-11-15
[Relay][Pytorch] Add support for `aten::unflatten`

### LINKED ISSUE #15663: [FRONTEND][PYTORCH] aten:unflatten not supported
Got the following error while trying to convert a CLIP model:

    NotImplementedError: The following operators are not implemented: [‘aten::unflatten’, ‘aten::scaled_dot_product_attention’]

There is already an open issue for `aten::scaled_dot_product_attention` (#15361), so I'm creating this one for `aten::unflatten`.

### Triage

* frontend:pytorch

cc @shingjan @yelite

mshr-h:
Fix #15663
Support [torch.unflatten](https://pytorch.org/docs/stable/generated/torch.unflatten.html).

cc @jikechao @vvchernov @Hzfengsy @junrushao

### DIFF
--- a/python/tvm/relay/frontend/pytorch.py
+++ b/python/tvm/relay/frontend/pytorch.py
@@ -1553,11 +1553,13 @@ class PyTorchOpConverter:
         dshape = get_const_tuple(self.infer_shape_with_prelude(data))
         assert len(dshape) > dim
 
-        mult = 1
-        for s in unflattened_size:
-            if s is not -1:
-                mult *= s
-        assert dshape[dim] % mult == 0
+        assert unflattened_size.count(-1) <= 1
+
+        mult = np.multiply.reduce(unflattened_size)
+        if mult < 0:
+            assert dshape[dim] % mult == 0
+        else:
+            assert dshape[dim] == mult
 
         new_shape = dshape[:dim] + unflattened_size + dshape[dim + 1 :]
         out = _op.reshape(data, new_shape)
--- a/python/tvm/relay/frontend/pytorch.py
+++ b/python/tvm/relay/frontend/pytorch.py
@@ -1546,6 +1546,16 @@ class PyTorchOpConverter:
             out = _op.squeeze(out, axis=squeeze_axes)
         return out
 
+    def unflatten(self, inputs, input_types):
+        data = inputs[0]
+        dim = int(inputs[1])
+        unflattened_size = tuple(inputs[2])
+        dshape = get_const_tuple(self.infer_shape_with_prelude(data))
+        assert len(dshape) > dim
+        new_shape = dshape[:dim] + unflattened_size + dshape[dim + 1 :]
+        out = _op.reshape(data, new_shape)
+        return out
+
     def addmm(self, inputs, input_types):
         input_mat = inputs[0]
         mat1 = inputs[1]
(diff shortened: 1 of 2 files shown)


python/tvm/relay/frontend/pytorch.py
```diff
@@ -1546,6 +1546,16 @@ def flatten(self, inputs, input_types):
             out = _op.squeeze(out, axis=squeeze_axes)
         return out
 
+    def unflatten(self, inputs, input_types):
+        data = inputs[0]
+        dim = int(inputs[1])
+        unflattened_size = tuple(inputs[2])
+        dshape = get_const_tuple(self.infer_shape_with_prelude(data))
+        assert len(dshape) > dim
+        new_shape = dshape[:dim] + unflattened_size + dshape[dim + 1 :]
```
vvchernov, line 1555: Add check that dshape[dim] == multiplication of dimensions in unflattened_size

vvchernov (changes requested):
Hello @mshr-h! LGTM, but it is needed one check

python/tvm/relay/frontend/pytorch.py
mshr-h: @vvchernov Thank you for your comment!
Multiplication of dimensions in `unflattened_size` doesn't always match `dshape[dim]` because `unflattened_size` can contain -1.
How do we check it? Do you have any idea?

mshr-h: @vvchernov Thanks!
I don't think we have to check it because torch.jit.trace does it.
They provide something like the below RuntimeError.
```
RuntimeError: unflatten: Provided sizes [3, 5, 3, -1] don't multiply up to the size of dim 0 (60) in the input tensor
```
Should we add the check in TVM's PyTorch frontend?

vvchernov: Hello @mshr-h! Ok torch.jit.trace do it, but in this case we do not need `assert len(dshape) > dim`.
It looks like TVM usually rechecks all corner cases.
About -1 in unflattened_size. In this case we can multiply together other dimension and check dshape[dim] % mult == 0

mshr-h: @vvchernov Thanks. Added the assertion.

jikechao (approved):
LGTM! @mshr-h Thanks for your PR.

python/tvm/relay/frontend/pytorch.py
vvchernov: Hello @mshr-h! There are two cases: unflattened_size has -1 and does not have. You check only the first one. Example:  dshape[dim] = 8, unflattened_size = [2, 2, 1, 1] pass your assert, but it is failure case

vvchernov: I've checked that _op.reshape does not check size correctness by it-self. Nevertheless I found that dim can be not only -1, but from list {0, -1, -2, -3, -4} See https://github.com/apache/tvm/blob/748882aae7be1435f042e22b0fc67cb236705b6c/python/tvm/relay/op/transform.py#L243
I suggest to check -1 case only. It looks like torch does not have other options

mshr-h: @vvchernov 
Thanks! Updated the shape check.

mshr-h: Ah, dim can be negative but I didn't check that. I'll add the check.

vvchernov (approved):
Hello @mshr-h! Thank you for careful work! Sorry for many small boring corrections. LGTM

python/tvm/relay/frontend/pytorch.py
mshr-h: > Ah, dim can be negative but I didn't check that. I'll add the check.

Done.

masahi (approved):

PR #16133, opened 2023-11-15
[Runtime] Parallel-for with threading backend

MasterJH5574:
This PR introduces the runtime parallel-for helper function in C++ with the threading backend in TVM.

Right now the existing [parallel-for](https://github.com/apache/tvm/blob/bd67d2e5ebde1aec18bcfa74c087516579bda1ae/include/tvm/support/parallel_for.h#L48-L68) in TVM is not thread persistent,
in which case we cannot get persistent TLS for each thread.

The introduced parallel-for-with-threading-backend function leverages the threading backend in TVM and persists threads.


junrushao (approved):
LGTM!

PR #16134, opened 2023-11-15
[Unity][BYOC] Support IGEMM in cuBLASLt

ibsidorenko:
This commit adds support of i8i8i32 matmul and its offloading to cuBLAS.


ibsidorenko:
cc @masahi

masahi (approved):

PR #16135, opened 2023-11-15
[Relay][Pass]Implementation of a custom pass to extend 8bit Conv2D with Checksum Calculation in Relay

jokap11:
Hello Everyone,

I assume that this automated Checksum calculation might be helpful for somebody to detect Soft Faults during the calculation of a Conv2D operation.

So I combined the knowledge from Hari. et. Al's Making Convolutions Resilient Via Algorithm-Based Error Detection Techniques and Uzair's  Efficient Software-Implemented HW Fault Tolerance for TinyML Inference in Safety-critical Applications to support both depthwise and standard Conv2D. 
This pass adds multiple relay nodes and generates one output Tuple with the original result and additional boolean output for each "secured" Conv2D to recognize how many soft faults happened during the whole NN.

I  am grateful for any tips/improvements in the code,
Johannes


jokap11:
@tqchen @mbrookhart   @zhiics 
Could you give some feedback?

Issue #12567, opened 2022-08-24
[Bug] python.contrib.test_onnx.test_resize numerical accuracy issue


leandron:
While working to enable Integration tests in AArch64 with ONNX and Torch installed (which is currently a blind spot in our CI, see https://github.com/apache/tvm/issues/12529), I'm seeing `python.contrib.test_onnx.test_resize` failing with an accuracy error. According to some tests locally, the issue reproduces consistently in x86 and AArch64.

I'm creating this issue so that we can investigate the numerical accuracy issue separately, but for now, I'm marking this test as an `xfail`.

Steps to reproduce:
1. Starting from a `ci_cpu` image, run the ONNX installation script (https://github.com/apache/tvm/blob/989e5a11285503716c2033f4e56f1bba6b6d00c7/docker/install/ubuntu_install_onnx.sh)
2. Run the test `python.contrib.test_onnx.test_resize`

This is the error message I see:
```
def test_resize():
        """Resize unit test."""
    
        def verify_resize(dshape, outsize, method, coord_trans, rounding_method, dtype="float32"):
            x = relay.var("x", relay.ty.TensorType(dshape, dtype))
            y = relay.image.resize2d(
                x,
                outsize,
                None,
                layout="NCHW",
                method=method,
                coordinate_transformation_mode=coord_trans,
                rounding_method=rounding_method,
            )
            func = relay.Function([x], y)
            x_data = np.random.uniform(size=dshape).astype(dtype)
            verify_results(func, [x_data], "test_resize", rtol=1e-4, atol=1e-4)
    
        method = ["nearest_neighbor", "linear", "cubic"]
        coord_trans = ["half_pixel", "align_corners", "asymmetric"]
        rounding_method = ["round", "floor", "ceil"]
    
        isize = (1, 3, 480, 640)
    
        # Downsample
        osize = (240, 320)
        for i in method:
            for j in coord_trans:
                for k in rounding_method:
                    if (i == "nearest_neighbor" and j == "align_corners") or (
                        i == "cubic" and j in ["half_pixel", "align_corners"]
                    ):
                        continue
                    verify_resize(isize, osize, method=i, coord_trans=j, rounding_method=k)
    
        # Upsample
        osize = (960, 1280)
        for i in method:
            for j in coord_trans:
                for k in rounding_method:
                    if (i == "nearest_neighbor" and j == "align_corners") or (i == "cubic"):
                        continue
[...]

PR #16137, opened 2023-11-16
[BugFix][TOPI] fix resize accuracy issue.

JR4er:
"resize" result can be wrong if method="nearest_neighbor" and coordinate_transformation_mode="asymmetric" and rounding_method != "floor" and target_height, target_width = n * image_height, m * image_width (n,m are integers).

#12567 

in buggy implementation:
`can_convert_multiply_to_intdiv` returns `True` if `scaled_size`  is n times `origin_size`, which will result in that `get_closest_index` converts the mapped coordinate by `astype('int32')` without checking `rounding_method`.

in this implementation:
`can_convert_multiply_to_intdiv` returns `True` if `origin_size` is n times `scaled_size`, which means that any coordinate in target_image will be mapped into an integer coordinate in the origin_image. Therefore `get_closest_index` do the `astype(`int32`)` in correct scenario.


Hzfengsy:
Thanks for the fix. Could you please also add a topi test?

JR4er:
> Thanks for the fix. Could you please also add a topi test?

a simple test for this case is added.

tqchen (approved):

tqchen:
@tvm-bot  rerun

tqchen:
@tvm-bot rerun

PR #16141, opened 2023-11-17
[Unity][LLM] Add NaN checks during sampling for better error reporting

Hzfengsy:
The current error message would be confusing:
```
src/runtime/relax_vm/lm_support.cc:421: InternalError: Check failed: sampled_index >= 0 (-1 vs. 0)
```
But most of the case is caused by NaN error. This PR improves the error message


Hzfengsy:
cc @junrushao @tqchen

junrushao (approved):

PR #16146, opened 2023-11-18
[Unity] Flash infer integration

jinhongyii:
Running MLC serving requires calling into flash infer kernels, and it should be moved from dependency of mlc-llm to tvm in order to run models on disco runtime successfully.


jinhongyii:
cc: @junrushao @MasterJH5574 @yzh119 @tqchen

tqchen (approved):

PR #16151, opened 2023-11-20
[Relay] conv3d depthwise bug fix

jonghewk:
# Issue
Relay for depthwise `nn.conv3d` causes error.

# Cause
`depthwise` is not fully implemented for `conv3d`.
The pytorch frontend takes care of the depthwise case and reshapes the weights.
(https://github.com/apache/tvm/blob/main/python/tvm/relay/frontend/pytorch.py#L1247).
However, this is not taken care in `Conv3DRel`.

Note that `Conv2DRel` takes care of this case. 
(https://github.com/apache/tvm/blob/main/src/relay/op/nn/convolution.cc#L273)

# How to reproduce
```
class Conv3d(nn.Module):
    def __init__(self):
        super(Conv3d, self).__init__()
        self.conv = nn.Conv3d(16, 32, kernel_size=1, groups=16)
    
    def forward(self, x):
        x = self.conv(x)
        return x

from tvm import relay
s = [1, 16, 10, 10, 10]
net = Conv3d()
input_data = torch.randn(s)
scripted_model = torch.jit.trace(net, input_data).eval()
mod = relay.frontend.from_pytorch(scripted_model, [("x", s)]) 
```

### DIFF
--- a/src/relay/op/nn/convolution.cc
+++ b/src/relay/op/nn/convolution.cc
@@ -480,10 +480,11 @@ bool Conv3DRel(const Array<Type>& types, int num_inputs, const Attrs& attrs,
     Array<IndexExpr> wshape;
     if (is_depthwise) {
       auto channel_multiplier = indexdiv(param->channels, dshape_ncdhw[1]);
-      wshape = {dshape_ncdhw[1], channel_multiplier, param->kernel_size[0], param->kernel_size[1], param->kernel_size[2]};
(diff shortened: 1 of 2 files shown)


jonghewk:
@masahi Any chance of reviewing this?

masahi:
Can you add your PT test as a test case?

jonghewk:
> Can you add your PT test as a test case?

Done.

masahi (approved):

PR #16155, opened 2023-11-21
[Unity][3rdparty] Remove TVM in 3rdparty of FlashInfer

MasterJH5574:
This PR updates the FlashInfer version in 3rdparty, which has removed TVM from its 3rdparty to avoid recursive TVM dependency.


MasterJH5574:
cc @yzh119 @tqchen

junrushao (approved):

PR #16161, opened 2023-11-24
[Unity][DLight] Fix `general_reduction` for GroupNorm

Hzfengsy:
The pass `lower_thread_allreduce` failed to codegen if the spatial loops are unrolled during the schedule. This PR works around the issue by changing the schedule rules.


tqchen:
is the update on cutlass submodules intentional? we should revert these then we are good to merge

Hzfengsy:
Thanks for pointing it out. Fixed :)

tqchen (approved):

PR #16168, opened 2023-11-27
Mimic the effect of offset by adjusting data ptr (similar to how unit…

arangasa:
…y had handled this)

Currently byte_offset field seems to be ignored (eg. in llvm codegen); this change is based on how unity's relax_vm memory_manager had handled offsets.


tqchen:
We cannot do it for all backends but can afford to do it for subset of backends, e.g. cuda, rocm where the data ptr corresponds to the VRAM pointer. For backends like opencl and metal, the data pointer arith won't work because the pointer do not corresponds to an address but an opaque buffer object in host.

As of now our memory allocator can try to allocate without offset and only enable this behavior for subset of the backends. We are also doing best effort allocating without offset for backends like metal and opencl by creating multiple buffers

arangasa:
Thanks a lot @tqchen for your helpful inputs! Closing this PR.

PR #16169, opened 2023-11-27
[Unity][DistIR] LowerDistIR

jinhongyii:
A pass that lowers DistIR to Relax function that runs on each worker.


jinhongyii:
cc: @yongwww @tqchen

src/relax/distributed/transform/lower_distir.cc
```diff
@@ -0,0 +1,271 @@
...
+  }
+
+  void InputPreprocessing() {
+    for (int i = 0; i < static_cast<int>(func_->params.size()); i++) {
+      Var param = func_->params[i];
+      if (const auto* dtensor_sinfo = GetStructInfoAs<DTensorStructInfoNode>(param)) {
+        EmitBroadcastOrScatter(param, new_params_[i], GetRef<DTensorStructInfo>(dtensor_sinfo));
+      } else if (const auto* tuple_sinfo = GetStructInfoAs<TupleStructInfoNode>(param)) {
+        for (int j = 0; j < static_cast<int>(tuple_sinfo->fields.size()); j++) {
+          if (const auto* dtensor_sinfo = tuple_sinfo->fields[j].as<DTensorStructInfoNode>()) {
```
yongwww, line 162: perhaps we can add a Warning or throw an error for non-DTensorStructInfoNode elements in tuple

yongwww (approved):
LGTM, left one minor comment

src/relax/distributed/transform/lower_distir.cc
jinhongyii: It's possible that the input contains a non-DTensorStructInfoNode element. For example, kv cache is object.

PR #16173, opened 2023-11-28
[Unity] [Transform] Remove iteration over functions in function pass

quic-sanirudh:
There was a small redundancy in a couple of function passes where they iterate over all passes in the mod within the transform function, while function pass itself would also do that iteration.

So we would be un-necessarily iterating through all functions and modifying them, but only one of them would be updated each time transform_function is called.

### DIFF
--- a/python/tvm/relax/transform/optimize_layout_transform.py
+++ b/python/tvm/relax/transform/optimize_layout_transform.py
@@ -19,7 +19,7 @@
 from tvm.ir import structural_equal
 from tvm.ir.module import IRModule
 from tvm.ir.transform import PassContext
-from tvm.relax import Expr, Function
+from tvm.relax import Expr
 from tvm.relax.dpl import TuplePattern, is_op, rewrite_call, wildcard
 
 from . import function_pass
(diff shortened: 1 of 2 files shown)


quic-sanirudh:
cc @masahi @Lunderberg

python/tvm/relax/transform/optimize_layout_transform.py
```diff
@@ -63,27 +63,24 @@ def transform_function(self, func: Expr, mod: IRModule, ctx: PassContext) -> IRM
...
-                if hasattr(arg1.struct_info, "shape") and hasattr(arg2.struct_info, "shape"):
-                    if structural_equal(arg1.struct_info.shape, arg2.struct_info.shape):
-                        return arg2
-                return expr
+            if hasattr(arg1.struct_info, "shape") and hasattr(arg2.struct_info, "shape"):
+                if structural_equal(arg1.struct_info.shape, arg2.struct_info.shape):
+                    return arg2
+            return expr
 
-            updated_func = rewrite_call(self.pattern, rewriter, func)
```
Lunderberg, line 87: Beyond just repetition, it looks like the previous implementation would produce incorrect results for an `IRModule` that contains more than one non-primitive Relax function.  Rather than returning the updated version of the argument `func`, this would return the updated version of whichever non-primitive relax function occurs last in `mod.function_items()`, regardless of the `func` argument.

quic-sanirudh: Thanks for the review, don't know how I missed that.

PR #16175, opened 2023-11-28
[Unity, BYOC] Add check for leaking intemediate variables for cublas and cudnn

vinx13:
Added checks to prevent cyclic dependency between groups during fusion

cc @masahi @Hzfengsy

### DIFF
--- a/tests/python/relax/test_transform_fuse_ops_by_pattern.py
+++ b/tests/python/relax/test_transform_fuse_ops_by_pattern.py
@@ -27,6 +27,7 @@ from tvm.relax.dpl.pattern import (
 )
 from tvm.relax.transform import PatternCheckContext
 from tvm.relax.backend.contrib.cutlass import partition_for_cutlass
+from tvm.relax.backend.contrib.cublas import partition_for_cublas
 from tvm.script import ir as I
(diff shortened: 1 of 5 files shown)


masahi:
I think such bInding can be removed by CanonicalizeBindIng pass

tests/python/relax/test_transform_fuse_ops_by_pattern.py
```diff
@@ -1023,5 +1024,27 @@ def main(
...
+    @I.ir_module
+    class Module:
+        @R.function
+        def main(
+            x: R.Tensor((1, 16), dtype="float16"), w: R.Tensor((16, 16), dtype="float16")
+        ) -> R.Tensor((1, 16), dtype="float16"):
+            with R.dataflow():
+                w1: R.Tensor((16, 16), dtype="float16") = R.permute_dims(w, axes=None)
+                y: R.Tensor((1, 16), dtype="float16") = R.matmul(x, w1)
+                y1: R.Tensor((1, 16), dtype="float16") = y
```
masahi, line 1039: I think such binding can be removed by CanonicalizeBinding pass

masahi (approved):

PR #16183, opened 2023-11-29
[FFI][RUNTIME] Introduce runtime boxed types for int/float/bool

Lunderberg:
Prior to this commit, `int`, `float`, and `bool` arguments from Python were converted to `IntImm`, `FloatImm`, and `Bool`.  These are subtypes of `PrimExpr`, and should only be used at compile-time.  By automatically applying this conversion as part of the FFI, these types are required to be present whenever a primitive is converted to a `tvm::ObjectRef`.

This can become especially fragile for an end-user when storing objects into a TVM container.  Because TVM containers require all contents to be `ObjectRef` subclasses, an automatic conversion may be applied on storing into a container, resulting in an unexpected type being retrieved from the container.  For example, this currently occurs in Relax when extracting a `R.Prim` from a `R.Tuple`.

This commit introduces a `Box<T>` type for storage of boxed primitives at runtime, distinct from the IR types.  This was based on discussion on https://github.com/apache/tvm/pull/15983, which would have further cemented the `IntImm` type in `libtvm_runtime.so`, rather than starting the process of separating it.

* Primitive arguments provided to a PackedFunc that requires an `ObjectRef` will be converted to the corresponding boxed type.  (e.g. Passing a Python `int` to a C++ function accepting `ObjectRef` produces a `Box<int64_t>`.

* Boxed primitives provided to a PackedFunc that requires an unboxed primitive will be converted to the corresponding primitive.

* PackedFunc return values of `ObjectRef` are converted to the corresponding primitive, if present.  (e.g. If a `tuple_getitem` with static return type `ObjectRef` returns a `Box<int64_t>`, it will be unwrapped to a python `int`.)

Together, these three rules provide backwards compatibility for existing PackedFunc definitions, while avoiding exposing the user to any container-induced type conversions betweeen primitive types and `ObjectRef`.


include/tvm/runtime/container/boxed_primitive.h
```diff
@@ -0,0 +1,121 @@
...
+ * Furthermore, the name must be specified in the Python FFI using
+ * `tvm._ffi.register_object`.  This prevents use of
+ * `typeid(T)::name()` to build a unique name, as the name is not
+ * required to be human-readable or consistent across compilers.
+ *
+ * This utility struct exists to bridge that gap, providing a unique
+ * name where required.
+ */
+template <typename Prim>
+struct BoxNodeTypeKey;
```
tqchen, line 52: BoxNodeTrait (so it also can help future traits if needed)

include/tvm/runtime/container/boxed_primitive.h
```diff
@@ -0,0 +1,121 @@
...
+
+template <>
+struct BoxNodeTypeKey<int64_t> {
+  static constexpr const char* _type_key = "runtime.BoxInt";
+};
+
```
tqchen, line 58: bring specialization to later part of the code

include/tvm/runtime/container/boxed_primitive.h
```diff
@@ -0,0 +1,121 @@
...
+   * \param value The value to be boxed
+   */
+  Box(Prim value) : ObjectRef(make_object<BoxNode<Prim>>(value)) {}
+
+  operator Prim() const { return (*this)->value; }
+
+  TVM_DEFINE_NOTNULLABLE_OBJECT_REF_METHODS(Box, ObjectRef, BoxNode<Prim>);
+};
+
+/*! \brief Runtime equivalent of IntImm */
```
tqchen, line 101: runtime should not be dependent on IntImm.

Instead, state "Object boxed version of int, can be used to store POD int values as object"

include/tvm/runtime/container/boxed_primitive.h
```diff
@@ -0,0 +1,121 @@
...
+using BoxInt = Box<int64_t>;
```
tqchen, line 102: I now think perhaps it is easier to do `runtime::Int`, `runtime::Bool`, `runtime::Float`. We should gradually switch the current usecases of 

tvm::Bool and tvm::Integer to runtime::Int, runtime::Bool and deprecate previous alias, so the boxed data type are clear

include/tvm/runtime/container/boxed_primitive.h
```diff
@@ -0,0 +1,121 @@
...
+
+/*! \brief Runtime equivalent of FloatImm */
+using BoxFloat = Box<double>;
```
tqchen, line 105: Instead, state "Object boxed version of floating point in FFI, can be used to store POD int values as object"

include/tvm/runtime/container/boxed_primitive.h
```diff
@@ -0,0 +1,121 @@
...
+/*! \brief Runtime equivalent of IntImm with DataType::Bool()
+ *
+ * When passing from Python to C++, TVM PackedFunc conversion follow
+ * C++ conversion rules, and allow bool->int and int->bool
+ * conversions.  When passing from C++ to Python, the types are
+ * returned as bool or int.  If the C++ function uses ObjectRef to
+ * hold the object, a Python to C++ to Python round trip will preserve
+ * the distinction between bool and int.
+ */
+using BoxBool = Box<bool>;
```
tqchen, line 116: Instead, state "Object boxed version of boolean values in FFI and containers, can be used to store POD int values as object"

tqchen (changes requested):
Thanks for the PR, i made some quick comment. The changes wrt to PackedFunc should have benefit from more pairs of eyes to review as they are centeral to most FFI routines

include/tvm/runtime/packed_func.h
```diff
@@ -2129,6 +2315,42 @@ struct PackedFuncValueConverter<::tvm::runtime::String> {
   }
 };
 
+template <typename T>
+struct PackedFuncValueConverter<Array<T>> {
```
tqchen, line 2319: I feel automatic conversion here is a bit overkill.  Recursive automatic conversion would be too much overhead whe passing through FFI. The developer should ensure the correct typings of the internal contained values

I know it may bring some convenience, but in this case I think we should not do it. 

Suggestion: Remove the conversion logic from this PR

include/tvm/runtime/packed_func.h
```diff
@@ -2035,46 +2167,100 @@ inline TObjectRef TVMPODValue_::AsObjectRef() const {
...
   using ContainerType = typename TObjectRef::ContainerType;
   const Object* ptr = other.get();
-  if (ptr != nullptr) {
-    if (std::is_base_of<NDArray::ContainerType, ContainerType>::value ||
-        (std::is_base_of<ContainerType, NDArray::ContainerType>::value &&
-         ptr->IsInstance<NDArray::ContainerType>())) {
+
+  if constexpr (std::is_base_of_v<ContainerType, NDArray::ContainerType> ||
+                std::is_base_of_v<NDArray::ContainerType, ContainerType>) {
+    if (ptr && (std::is_base_of_v<NDArray::ContainerType, ContainerType> ||
```
tqchen, line 2224: nit: would be nice to keep the originals tructure of `if (ptr != nullptr)`, they  helps to keep the structure of the code, lift out a CSE and also makes the diff more readable

tqchen (changes requested):
I spent sometime to go through packed_func.h logic. Given there are some restructuring, this would need some more careful checks (as they impact everything).

tqchen:
cc @junrushao

include/tvm/runtime/packed_func.h
```diff
@@ -547,29 +551,43 @@ class TVMPODValue_ {
     // Allow automatic conversion from int to float
     // This avoids errors when user pass in int from
     // the frontend while the API expects a float.
-    if (type_code_ == kDLInt) {
-      return static_cast<double>(value_.v_int64);
+    if (auto opt = TryAsBool()) {
+      return opt.value();
+    } else if (auto opt = TryAsInt()) {
```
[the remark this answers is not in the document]
slyubomirsky: Would any bool value also work as an int value? Maybe that's the reason for the ordering given.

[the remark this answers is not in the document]
tqchen: in this case order does not matter for correctness, mainly as an optimization as the original intended type as the most freq path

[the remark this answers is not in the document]
slyubomirsky: Ah, I just looked at how they work. Yeah, order doesn't matter then.

tqchen:
Just highlight some of the main comments that are high level
- naming: we can probably go with `runtime::Int`,  `runtime::Float`,  `runtime::Bool` as they are more understandable terms
   - we should remove tvm::Integer and tvm::bool after migrating their usage to boxed values(in later PR)
- convention: It is cleaner to have a convention where runtime::Int and runtime::Float always get unboxed when passing into TVMArg and TVMRet. 
    - The codegen layer that expects int/float then do not need to deal with object. 
    - That does mean the FFI layer needs to explicitly unbox when setting boxed values  into them (seems the current impl is already doing so for TVMRet, need to check argument setting as well)

include/tvm/runtime/packed_func.h
```diff
@@ -1996,8 +2123,10 @@ inline TObjectRef TVMPODValue_::AsObjectRef() const {
         << "Expect a not null value of " << ContainerType::_type_key;
     return TObjectRef(ObjectPtr<Object>(nullptr));
   }
-  // NOTE: the following code can be optimized by constant folding.
-  if (std::is_base_of<NDArray::ContainerType, ContainerType>::value) {
+
+  // NOTE: The following code uses "if constexpr" wherever possible to
+  // minimize the number of runtime checks.
```
slyubomirsky, line 2128: Sometimes I love modern C++ :slightly_smiling_face:

Lunderberg:
> Thanks for the PR, love the direction we are going with dedicated boxed types.

Thank you!

> The changes wrt to PackedFunc should have benefit from more pairs of eyes to review as they are centeral to most FFI routines

Definitely agreed, and the more eyes the better.  Given that this PR has the potential to break any use of the PackedFunc interface, the surface area is very broad.

> naming: we can probably go with runtime::Int, runtime::Float, runtime::Bool

I like this readability improvement, and will update accordingly

> convention: It is cleaner to have a convention where runtime::Int and runtime::Float always get unboxed when passing into TVMArg and TVMRet. 

Agreed.  Assuming I have it implemented correctly, this is handled in the `TVMArgsSetter::SetObject` and `TVMRetValue::operator=(TObjectRef)` implementations, so a boxed primitive is always unboxed when storing in the FFI type.

The conversions that occur when reading from a `TVMArgsSetter` or `TVMRetValue` are to convert into some form of `ObjectRef` that is constructible from the primitive, based on the signature of the type receiving the argument.

include/tvm/runtime/packed_func.h
```diff
@@ -627,13 +645,54 @@ class TVMPODValue_ {
   template <typename TObjectRef>
   inline TObjectRef AsObjectRef() const;
 
+  std::optional<int64_t> TryAsInt() const {
+    // Helper function to reduce duplication in the variable integer
+    // conversions.  This is publicly exposed, as it can be useful in
+    // specializations of PackedFuncValueConverter.
+    if (auto opt = FromBoxed<int64_t>()) {
```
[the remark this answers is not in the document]
tqchen: That means we do not need to have FromBoxed, since Boxed value do not occur in the ArgValue

include/tvm/runtime/container/boxed_primitive.h
Lunderberg: I like the naming for future extensions, though naming it `BoxNodeTrait` might cause confusion with the `BoxNodeTrait` defined [here](https://github.com/apache/tvm/pull/16183/files#diff-22a23c986d23dd8dc15c9233c738505aa04ad2626c0cb24828b7e672d1f56702R48).  I've renamed this instance to `BoxNodeRuntimeTraits` and the other to `BoxNodeCompileTimeTraits` to make a clear distinction between the two cases, and documented the intended usage for each.

include/tvm/runtime/container/boxed_primitive.h
Lunderberg: Good call, and that makes sense to avoid the documentation dependencies as well as the C++ compiler dependencies.  Docstring updated here, as well as for float/bool.

include/tvm/runtime/container/boxed_primitive.h
Lunderberg: Sounds good.  I had explicitly named them as `BoxFoo` to distinguish them from `tvm::Integer` and `tvm::Bool`, but that's no longer a need if those usages are to be deprecated.

include/tvm/runtime/container/boxed_primitive.h
Lunderberg: Updated similar to suggestion, removing all reference to `FloatImm`, with slight wording adjustments.

include/tvm/runtime/container/boxed_primitive.h
Lunderberg: Updated similar to suggestion, removing all reference to `IntImm` and `DataType::Bool`, with slight wording adjustments.

include/tvm/runtime/packed_func.h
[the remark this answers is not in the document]
Lunderberg: Good call, and reordered here and following conversions.  For each conversion operator, it now checks the expected type first, and only falls back to allowed conversions if the expected type isn't present.

> Would any bool value also work as an int value? Maybe that's the reason for the ordering given.

Nope, no specific reason for the ordering.  These specific `Try*` functions don't apply any conversions, so the order doesn't matter for correctness.

include/tvm/runtime/packed_func.h
```diff
@@ -547,29 +551,43 @@ class TVMPODValue_ {
...
-    return value_.v_float64;
   }
   operator int64_t() const {
-    TVM_CHECK_TYPE_CODE(type_code_, kDLInt);
-    return value_.v_int64;
-  }
-  operator uint64_t() const {
-    TVM_CHECK_TYPE_CODE(type_code_, kDLInt);
-    return value_.v_int64;
+    if (auto opt = TryAsBool()) {
```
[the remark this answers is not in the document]
Lunderberg: Updated here as well.

include/tvm/runtime/packed_func.h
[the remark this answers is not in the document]
Lunderberg: Ah, I had missed the context of the more general comment.  I agree, these extra conversions shouldn't be necessary, and will verify to make sure that is the case.

include/tvm/runtime/packed_func.h
Lunderberg: I think we need these conversions in order to preserve backwards compatibility with existing user-written code.  Otherwise, a user-written function that accepts `Array<PrimExpr>` could no longer be called with a python list `[1,2,3]`.

Lunderberg: Regarding performance, I think we should be good there as well.  The conversion is applied instead of the existing recursive check in `ObjectTypeChecker<Array<T>>`, so no additional recursive walks.  The mutation is implemented in terms of `Array::Map`, which is copy-on-write, so no allocations occur if the user has provided the expected types (the common case).

include/tvm/runtime/packed_func.h
Lunderberg: I mainly restructured it in order to make sure the `if constexpr` checks were on the outermost conditional, so that they would be guaranteed to be applied to the entire contents.  I agree on the readability and repetition though, and will take a look to see if there's a way to make it more readable overall.

Lunderberg: How does it look in the latest commit?  I moved the repeated `if (ptr)` structure out to an early return for `if (ptr == nullptr)`, which preserves the `if constexpr` structure.

include/tvm/runtime/packed_func.h
Lunderberg: Holy cow, yeah.  Each step from  C++03 partial template specialization, to C++11 `std::enable_if`, to C++17 `if constexpr` brings such an improvement in readability.  It probably results in the same optimized binary at the end, since the optimizer would (probably) DCE the constants, but this way the compiler will yell at me if I accidentally insert a runtime value into the condition.

include/tvm/runtime/packed_func.h
tqchen: early return of nullptr sounds good

include/tvm/runtime/packed_func.h
```diff
@@ -2129,6 +2315,42 @@ struct PackedFuncValueConverter<::tvm::runtime::String> {
...
+template <typename T>
+struct PackedFuncValueConverter<Array<T>> {
+  static Array<T> From(const TVMArgValue& val) {
+    auto untyped_array = val.AsObjectRef<Array<ObjectRef>>();
+
+    // Attempt to convert each item of the array into the desired
+    // type.  If the items do not require a conversion, no copies are
+    // made.
+    return untyped_array.Map([](ObjectRef item) {
+      // The TVMArgValue is intentionally defined through
```
tqchen, line 2327: add quick common case short cut, and avoid arg setter conversion logic
```c++
if (auto* ptr = item->as<T::ContainerType>()) {
   return GetRef<T>(ptr);
} 
```

include/tvm/runtime/packed_func.h
```diff
@@ -2129,6 +2315,42 @@ struct PackedFuncValueConverter<::tvm::runtime::String> {
...
+      // `TVMArgsSetter`, rather than defining it with
+      // `value.data_ = item.get();` and type code
+      // `kTVMObjectHandle`.  `TVMArgsSetter::operator()` includes
+      // special handling for unwrapping boxed primitives,
+      // PackedFunc, runtime::Module, etc, which should be checked
+      // before delegating to the array element's
+      // PackedFuncValueConverter implementation.
```
tqchen, line 2334: do not quite get what we are intending to handle, maybe state we would like to allow conversion to classes like PrimExpr, so would like to trigger  that

include/tvm/runtime/packed_func.h
```diff
@@ -2129,6 +2315,42 @@ struct PackedFuncValueConverter<::tvm::runtime::String> {
...
+      TVMArgValue arg(value, type_code);
+      return PackedFuncValueConverter<T>::From(arg);
+    });
+  }
+  static Array<T> From(const TVMRetValue& val) {
+    auto untyped_array = val.AsObjectRef<Array<ObjectRef>>();
+
+    return untyped_array.Map([](ObjectRef item) {
+      TVMRetValue item_val;
+      item_val = std::move(item);
```
tqchen, line 2348: have same quick short cut 
```c++
if (auto* ptr = item->as<T::ContainerType>()) {
   return GetRef<T>(ptr);
} 
```

the fallback conversion can remain the same?

include/tvm/runtime/packed_func.h
tqchen: We should likely check some of those cases, and update quite a few of them to `Array<runtime::Int>`. But agree some might be useful especially for the mixed symbolic shape  and static shape case

include/tvm/runtime/packed_func.h
tqchen: the main reason is that round trip through to TVMArg and TVMRet variant have cost of switching etc, and we should optimize for most freq path

tqchen (changes requested):

include/tvm/runtime/packed_func.h
tqchen: added some comments below on optimizing for freq case

include/tvm/runtime/packed_func.h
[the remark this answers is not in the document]
Lunderberg: Verified, these `FromBoxed` conversions are only required for booleans, and has been removed from the int/double handling.

include/tvm/runtime/packed_func.h
Lunderberg: Sounds good, and going through the comments now.  I like the idea of updating the call sites over time, so that eventually the backwards compatibility conversion would only be triggered for externally-defined PackedFunc instances.

include/tvm/runtime/packed_func.h
Lunderberg: Brainstorming, I think we could avoid the round trip through `TVMArgValue` and `TVMRetValue` by instead changing the signature in `PackedFuncValueConverter` so it could make recursive calls on the `ObjectRef` directly.

```c++
template<typename T>
struct PackedFuncValueConverter<Array<T>> {
  // Current API, value converter converts from TVMArgValue or
  // TVMRetValue.  The From() function is called whenever converting
  // to Array<T>.
  static Array<T> From(const TVMArgValue& val) { ... }
};

template<typename T>
struct PackedFuncValueConverter<Array<T>> {
  // Proposed API, value converter converts from the value contained
  // within the TVMArgValue or TVMRetValue.  The From() function is
  // called whenever converting to T, if the contained value is of an
  // appropriate type.
  //
  // Exposing the individual overloads would allow them to be called
  // without first wrapping in `TVMArgValue` or `TVMRetValue`.
  static runtime::String From(ObjectRef obj) { ... }
};



// For types that are convertible from primitives, an overload
// accepting a primitive argument can be defined.
template<>
struct PackedFuncValueConverter<String> {
  static String From(const char* prim) { ... }

  static String From(const TVMByteArray* arr) { ... }
}

// When delegating to PackedFuncValueConverter, the existence of an
// overload can be checked at C++ compile-time, so the performance
// should be equivalent to the current checks.
inline TVMArgValue::operator T() const {
  switch(type_code_) {
    case kTVMArgInt:
      // If the TVMArgValue holds an integer, and
      // PackedFuncValueConverter defines a conversion from integer to
      // T, then run the converter.
      if constexpr(std::is_invocable_v<PackedFuncValueConverter<T>::From, int64_t>) {
        return PackedFuncValueConverter<T>::From(value_.v_int64);
      }
      break;

    ...
  }

  LOG(FATAL) << "Error message here";
}
```

I think such a change to `PackedFuncValueConverter` would be best done in a follow-up PR, as this one is already on the larger side.

include/tvm/runtime/packed_func.h
Lunderberg: Sounds good.  I changed the comment to describe the use case that it enables.

include/tvm/runtime/packed_func.h
tqchen: let us first go with this one for now

Lunderberg:
After slowly working through all the unit tests and transforms that needed to be updated for this change, this PR is now ready for review!  I'm going to squash all the incremental commits as the (very) long history of this dev branch isn't useful.

Lunderberg:
@tvm-bot rerun

Lunderberg:
@tvm-bot rerun

tqchen (approved):

tqchen:
This is merged, thanks @Lunderberg !

include/tvm/ir/expr.h
```diff
@@ -770,53 +770,121 @@ inline const TTypeNode* RelayExprNode::type_as() const {
...
+ * In previous versions of TVM, IntImm was the default FFI type for
+ * integer arguments, instead of runtime::Int.  For backwards
+ * compatibility where the callee has been updated to expected a
+ * runtime::Int, the caller has not been updated to provide a
+ * runtime::Int (e.g. relay script parsing), and the auto-unboxing of
+ * runtime::Int does not apply (e.g. making an `Array<runtime::Int>`),
+ * allow the IntImm to be generated.
+ */
+template <>
+struct PackedFuncValueConverter<runtime::Int> {
```
tqchen, line 880: Let us consider update the caller to Int as well so we don't have this path, as converting from IntImm -> Int can be less relevant

include/tvm/runtime/c_runtime_api.h
```diff
@@ -207,6 +209,7 @@ typedef DLTensor* TVMArrayHandle;
  */
 typedef union {
   int64_t v_int64;
+  bool v_bool;
```
tqchen, line 212: likely v_bool is not needed and we can simply store value in `v_int64 `, this can help to reduce the switch cases, and also make all value types as 64 bit, might help kernel passing handling to reduce cases as well.

include/tvm/ir/expr.h
Lunderberg: Agreed.  I've made the draft PR https://github.com/apache/tvm/pull/17241 to remove the backwards-compatibility handler, in order to see which callers relied on it.

tqchen:
We just find out some a perf regression introduced by this PR, specifically, during LLM decode function the PackedFunc calling overhead goes up to 1.4 ms. This can impact the ability of downstream system. Likely this is due to some of the automatic conversion logic introduced. 

As a temp measure, I am going to first revert this.

The changes of the PR(boxed types and bool) are valuable. Given the smart auto conversion might introduce extra overhead and the regression we see. I think it is helpful to isolate the PR into two pieces, with one that introduces the boxed types/bool, and another one that introduce a possiblity more conservative version of automatic conversion.

Thanks @Lunderberg for great effort and sorry for the extra trouble. As PackedFunc call gets into the center of our runtime execution, being able to reduce the execution time is now become an important topic, so we need to be more careful balancing the runtime logics

Lunderberg:
Looking into the origin of the performance difference, since it may just result from an increased number of FFI calls, and (maybe) could be resolved without removing the automatic conversions.

tqchen:
maybe it is not too bad to do manual conversion to the array of ir types in python wrapper then leave ffi part lightweight

Lunderberg:
First results, looks like the implicit conversions aren't necessarily causing the overhead.  I benchmarked constructing a TIR buffer, with the shape/strides either provided as a 

On `main` (after the FFI changes), the runtime is about the same, regardless of how the shape/strides are provided.
image

On `5a67a00bcb` (last commit before the FFI changes), it's about 80% slower 's about a 40% overhead when providing python integers (`[1,2,3]`) as compared to providing TIR IntImm (`[T.int32(1), T.int32(2), T.int32(3)]`).

image

This is rather surprising, because I expected the overhead to be present after the FFI changes, but this benchmark suggests that there's less overhead.

Next up, seeing how the FFI changes impact Relax `R.call_tir` operators.

Lunderberg:
Hmm, the same thing is occurring for this test case.  I made a Relax module that computes `x = subtract_one(add_one(x))`, for a large number of iterations.  (10/100/1000 iterations in the benchmarks shown below.)

(Current `main`, after the FFI changes)
image

(Before the FFI changes)
image

Since the effect isn't showing up for any smaller benchmarks, next for this investigation will probably need to be a full model execution.

vinx13:
Here is an example, the number of weights are large enough to see the difference. https://gist.github.com/vinx13/ea7a8c785d8d0d5ae5318e8ace085db2

Lunderberg:
Thank you for the example, @vinx13, and I can reproduce the difference with and without the FFI changes.  My numbers roughly agree with yours (1.1 ms vs 5.2 ms), so I can start narrowing down on the cause.

Lunderberg:
FOUND IT!

The root cause was in the type validation, but ended up being pretty easily solveable.  When converting from the untyped `ArrayNode*` to the typed `Array<T>`, we validate that each `ObjectRef` in the array has the correct type `T`.  However, because each element in the untyped `ArrayNode*` must still be some type of `ObjectRef`, this check is unnecessary.

In the test case, this performance overhead largely occurred during the calls to `"vm.builtin.tuple_getitem"`.  For each call, the validation function iterated over the entire `Array<ObjectRef>` argument.  This overhead can be removed by adding a compile-time check on the type parameter's type.  I've updated the conversions for both `Array<T>` and `Map<T,U>` to skip type checks that will always succeed.

With this addition, the FFI update went from being 5x slower than baseline (increase from 1 ms to 5 ms) to being 30% faster than baseline (decrease from 1ms to 700 us).  I'll make a PR shortly that re-applied the FFI changes, with the additional fix.

PR #16184, opened 2023-11-29
[TIR][Transform] Implement InlinePrivateFunctions

Lunderberg:
The functionality to express a call from one `PrimFunc` to another was introduced in https://github.com/apache/tvm/pull/14889.  While this was initially planned to be supported at codegen for all targets (see https://github.com/apache/tvm/pull/15835), this resulted in breakage on some backends (see https://github.com/apache/tvm/pull/16033). After discussion, the plan was changed to support TIR inlining, which would enable the same high-level functionality in TIR without requiring immediate low-level support across all codegens.

This commit implements and tests a new IRModule transform `InlinePrivateFunctions`, which can be used as part of lowering in a follow-up commit.

Because this is initially implemented for use quite late in the lowering flow, many constructs are not currently supported.  The current implementation has the following restrictions.

* `tir::Block` nodes may not occur in the inlined function.  Because a subroutine may be called multiple times, inlining of a subroutine that contains `tir::Block` would result in non-unique names. Support of subroutines with `tir::Block` instances will require de-duplication of block names.

* The subroutine's callsite must occur within a `tir::Evaluate` block. Because inlining a subroutine inserts the `tir::Stmt` body at the point of use, replacement must occur in a context where a `tir::Stmt` can be returned.  Support of subroutines that are called within an expression (e.g. Replacing `func` in `Buf[0] = func(1) + func(2)`) would require hoisting preprocessing done in the subroutine to the parent `tir::Stmt`.

* The subroutine may only accept primitive arguments, and must have an empty `buffer_map`.  Support of subroutines that are called with `tir::Buffer` or `tir::BufferRegion` arguments would require a way to represent these arguments at the callsite, and substitution of the buffer into the callee.

If these unsupported constructs are used, then the inlining of those functions is skipped.  This commit includes unit tests for these unsupported constructs, to validate that `InlinePrivateFunctions` produces well-formed output even when they are present.

### DIFF
--- a/src/tir/ir/specialize.cc
+++ b/src/tir/ir/specialize.cc
@@ -29,6 +29,7 @@
 
 #include <functional>
 
+#include "../transforms/ir_utils.h"
 #include "functor_common.h"
 
 namespace tvm {
@@ -115,18 +116,18 @@ class PrimFuncSpecializer : public StmtExprMutator {
  private:
   Stmt VisitStmt_(const BlockNode* op) final {
     // Step.0. Define buffer mappings which is allocated inside the block
-    Array<Buffer> alloc_buffers = op->alloc_buffers.Map(
-        std::bind(&PrimFuncSpecializer::MutateAllocBuffer, this, std::placeholders::_1));
+    Array<Buffer> alloc_buffers =
+        op->alloc_buffers.Map([this](const auto& buf) { return MutateAllocBuffer(buf); });
 
     // Step.1. Recursively visit block body
     Stmt stmt = StmtExprMutator::VisitStmt_(op);
     op = stmt.as<BlockNode>();
     ICHECK(op != nullptr);
 
-    Array<BufferRegion> reads = op->reads.Map(
-        std::bind(&PrimFuncSpecializer::MutateBufferRegion, this, std::placeholders::_1));
-    Array<BufferRegion> writes = op->writes.Map(
-        std::bind(&PrimFuncSpecializer::MutateBufferRegion, this, std::placeholders::_1));
+    Array<BufferRegion> reads =
+        op->reads.Map([this](const auto& region) { return MutateBufferRegion(region); });
+    Array<BufferRegion> writes =
+        op->writes.Map([this](const auto& region) { return MutateBufferRegion(region); });
 
     if (alloc_buffers.same_as(op->alloc_buffers) && reads.same_as(op->reads) &&
         writes.same_as(op->writes)) {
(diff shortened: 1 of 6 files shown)


Lunderberg:
Rebased onto main as the CI results were a bit stale.

src/tir/ir/specialize.cc
```diff
@@ -115,18 +116,18 @@ class PrimFuncSpecializer : public StmtExprMutator {
  private:
   Stmt VisitStmt_(const BlockNode* op) final {
     // Step.0. Define buffer mappings which is allocated inside the block
-    Array<Buffer> alloc_buffers = op->alloc_buffers.Map(
-        std::bind(&PrimFuncSpecializer::MutateAllocBuffer, this, std::placeholders::_1));
+    Array<Buffer> alloc_buffers =
+        op->alloc_buffers.Map([this](const auto& buf) { return MutateAllocBuffer(buf); });
```
slyubomirsky, line 120: Much cleaner this way :)

src/tir/ir/specialize.cc
```diff
@@ -140,16 +141,54 @@ class PrimFuncSpecializer : public StmtExprMutator {
...
+    Var old_buffer_var = op->buffer->data;
+    Buffer new_buf = MutateAllocBuffer(op->buffer);
+
+    auto node = Downcast<DeclBuffer>(StmtExprMutator::VisitStmt_(op));
+
+    if (!new_buf.same_as(node->buffer)) {
+      node.CopyOnWrite()->buffer = new_buf;
+    }
+
+    // If the buffer variable is begin remapped to an expression, we
```
slyubomirsky, line 156:
```suggestion
    // If the buffer variable is being remapped to an expression, we
```

src/tir/transforms/inline_private_functions.cc
```diff
@@ -0,0 +1,273 @@
...
+  }
+  return recursive_funcs;
+}
+
+Map<GlobalVar, PrimFunc> CollectInlinablePrimFuncs(const IRModule& mod) {
+  auto recursive_functions = CollectRecursiveFunctions(mod);
+
+  Map<GlobalVar, PrimFunc> output;
+  for (const auto& [gvar, base_func] : mod->functions) {
+    if (auto opt = base_func.as<PrimFunc>()) {
```
slyubomirsky, line 105: Not sure it's entirely necessary, but you can reduce nesting with a construction like
```c++
if (!base_func.as<PrimFunc>()) { continue; }
auto prim_func = Downcast<PrimFunc>(base_func);
// ...
```
I'm a fan of reducing nesting when possible, but that's up to you.

src/tir/transforms/inline_private_functions.cc
```diff
@@ -0,0 +1,273 @@
...
+    return func;
+  }
+
+  PSet<GlobalVar> GetRemovableFunctions() const { return removable_funcs_; }
+
+ private:
+  Stmt VisitStmt_(const EvaluateNode* eval) override {
+    if (auto call = eval->value.as<CallNode>()) {
+      if (auto gvar = call->op.as<GlobalVar>()) {
+        if (auto opt_callee = inlinable_funcs_.Get(gvar.value())) {
```
slyubomirsky, line 165: This is perhaps a place where reducing nesting might improve readability.

src/tir/transforms/inline_private_functions.cc
```diff
@@ -0,0 +1,273 @@
...
+      func.CopyOnWrite()->body = new_body;
+    }
+
+    return func;
+  }
+
+  PSet<GlobalVar> GetRemovableFunctions() const { return removable_funcs_; }
+
+ private:
+  Stmt VisitStmt_(const EvaluateNode* eval) override {
```
slyubomirsky, line 162: You should probably mention the details from the PR description as for why `EvaluateNode` is the only case handled

slyubomirsky (commented):
I am not especially familiar with the TIR code base, but the logic here seems reasonable. Is there a reason the issue of unique names for blocks is a blocker?

src/tir/transforms/inline_private_functions.cc
slyubomirsky, line 162: It might be good to mention the details from the PR description as for why cases other than `EvaluateNode` are not handled.

Lunderberg:
> I am not especially familiar with the TIR code base, but the logic here seems reasonable. Is there a reason the issue of unique names for blocks is a blocker?

Not a strong blocker, just an unsupported case at the moment.  Something that can definitely be extended in the future, but not something required for use late in the lowering pipeline.

src/tir/ir/specialize.cc
Lunderberg: Agreed.  I always need to pause when encountering `std::placeholders`, and try to replace it when reasonable to do so.

src/tir/ir/specialize.cc
Lunderberg: Thank you, and fixed.

src/tir/transforms/inline_private_functions.cc
```diff
@@ -0,0 +1,273 @@
...
+  }
+
+  return visitor.caller_lookup;
+}
+
+PSet<GlobalVar> CollectRecursiveFunctions(const IRModule& mod) {
+  // Collect all direct callers
+  auto call_map = CollectCallMap(mod);
+
+  // Propagate to find all indirect callers
```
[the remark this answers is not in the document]
Lunderberg: That's correct.  After propagation, the map would contain all functions that could exist earlier in a call tree.

src/tir/transforms/inline_private_functions.cc
Lunderberg: I try to avoid using `continue`, as the lack of nesting makes it harder to track when the flow control changes.  If the nesting gets to be too much, I tend to switch to a subroutine with early return.  The early return can mimic any flow control that `continue` could have, but the restricted context available in the subroutine keeps it manageable.  Looking at this case again, I think it would be better to pull these out into an `bool IsInlinablePrimFunc(const PrimFunc& func, PSet<GlobalVar>& recursive_functions)` subroutine, and will update to do so.

src/tir/transforms/inline_private_functions.cc
Lunderberg: Yeah, the nesting is a bit deep here.  Reordered to instead call a `GetInlinedSubroutine` method, and let me know what you think on it.

src/tir/transforms/inline_private_functions.cc
Lunderberg: Good call, and I've added the details in a comment, along with pointing to the `xfail` test case.

slyubomirsky (approved):
Thanks for addressing my suggestions.

Lunderberg:
Doing one last CI re-run before merging.  I don't expect there to be breaking changes introduced over Christmas, but I try to avoid stale CI results either way.

Lunderberg:
CI is passing, except for a flaky unit test.  I've submitted https://github.com/apache/tvm/pull/16337 to disable the flaky unit test.  Re-running the CI to see if I can hit the 2/3 chance of passing the flaky test while I wait on it.

PR #16192, opened 2023-12-01
[Device][Metal] Fix metal warp size

jinhongyii:
metal warp size should be 1 on x86 device, but 32 on M1/M2 device

### DIFF
--- a/src/runtime/metal/metal_device_api.mm
+++ b/src/runtime/metal/metal_device_api.mm
@@ -55,7 +55,6 @@ void MetalWorkspace::GetAttr(Device dev, DeviceAttrKind kind, TVMRetValue* rv) {
         break;
       }
       case kWarpSize: {
-// Set warp size to be 1 for safty reason.
 #if defined(__x86_64__)
         *rv = 1;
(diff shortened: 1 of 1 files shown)


jinhongyii:
cc: @junrushao @tqchen

junrushao (approved):

src/runtime/metal/metal_device_api.mm
```diff
@@ -55,8 +55,13 @@
         break;
       }
       case kWarpSize: {
-        // Set warp size to be 1 for safty reason.
+#if defined(__x86_64__)
         *rv = 1;
+#elif defined(__aarch64__)
+        *rv = 32;
+#else
+        LOG(FATAL) << "Unknown architecture";
```
junrushao, line 63: as a fall back, perhaps we should use `LOG(WARNING)` saying it's not x86 or aarch64, and then return 1 as a fallback

tqchen: i agree @jinhongyii do you might do that?

PR #16193, opened 2023-12-01
[Unity] Check usage location when canonicalizing trivial bindings

Lunderberg:
Prior to this commit, the `CanonicalizeBindings` transform handled trivial bindings, including binding of a `DataflowVar` to a `Var`, and usage of `Var` where a `DataflowVar` should be used instead.  However, it did not handle the case where both of these occur at the same time.  That is, a binding of `var_x = dataflow_y`, where `var_x` is only used within the dataflow block.  For these cases, `CanonicalizeBindings` would need to be run repeatedly.

This commit updates `CanonicalizeBindings` to check if a trivial binding is used outside the dataflow block.  This avoids producing output that requires further canonicalization.

### DIFF
--- a/src/relax/transform/canonicalize_bindings.cc
+++ b/src/relax/transform/canonicalize_bindings.cc
@@ -57,7 +57,17 @@ class CanonicalizePlanner : public ExprVisitor {
 
     CanonicalizationPlan plan;
 
-    std::unordered_set<Var, ObjectPtrHash, ObjectPtrEqual> handled;
+    // If a Var has been defined inside a DataflowBlock, is only used
+    // within a DataflowBlock, and is not already handled by removal
+    // of trivial bindings, then we can replace it with a DataflowVar.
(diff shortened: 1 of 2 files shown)


slyubomirsky (approved):
Great catch, what a fun case.

slyubomirsky:
Possible other case to try: The LHS and RHS are both non-dataflow vars that appear only inside a dataflow block. They should both be turned into dataflow vars and may need to be eliminated. I encountered this case when messing with another pass.

Example:
```
with R.dataflow()
    z = R.add(x, y)
    q = z
    r = q
    w = R.multiply(r, r)
    R.output(z, q, r, w)
...
# let's say only w is used
```

Should be canonicalized into
```
with R.dataflow():
    z = R.add(x, y)
    w = R.multiply(z, z)
    R.output(w)
...
```

I believe your changes should handle it but it would be a good one to check. This did not end up working when I tried it without your fix.

PR #16194, opened 2023-12-01
[Unity][Transform] Implement Relax function inlining

Lunderberg:
This commit adds the ability to inline one Relax function into another.  This is provided with two APIs, one which allows explicit specification of the functions to be inlined, and one which inlines
every private Relax function in an `IRModule`.  Function inlining with explicit replacements is intended for use by an end-user when building a model function (e.g. inlining of a rotary embedding), and the second is intended for use as part of a generic optimization pipeline.

### DIFF
--- a/src/relax/transform/inline_functions.cc
+++ b/src/relax/transform/inline_functions.cc
@@ -17,16 +17,11 @@
  * under the License.
  */
 
-#include <tvm/driver/driver_api.h>
-#include <tvm/ir/function.h>
 #include <tvm/relax/analysis.h>
 #include <tvm/relax/expr.h>
 #include <tvm/relax/expr_functor.h>
 #include <tvm/relax/transform.h>
-#include <tvm/relax/type.h>
-#include <tvm/tir/op.h>
 
-#include <tuple>
 #include <utility>
 
 #include "../../support/ordered_set.h"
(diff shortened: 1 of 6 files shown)


Lunderberg:
This PR is currently marked as a draft, as it relies on improvements implemented in https://github.com/apache/tvm/pull/16193.

src/relax/transform/inline_functions.cc
```diff
@@ -0,0 +1,306 @@
...
+    if (auto callee = op->op.as<GlobalVar>()) {
+      callee_to_caller_[callee.value()].insert(current_caller_.value());
+    }
+  }
+
+  Optional<GlobalVar> current_caller_ = NullOpt;
+  GlobalVarMap callee_to_caller_;
+};
+
+std::unordered_set<GlobalVar, ObjectPtrHash, ObjectPtrEqual> FindRecursiveFunctions(
```
slyubomirsky, line 76: https://github.com/apache/tvm/blob/unity/src/relax/analysis/detect_recursion.cc Not sure if my analysis might be suitable for this purpose.

Lunderberg: Ooh, thank you for pointing that out, and that's exactly what I was looking for.  Updated the PR to use `DetectRecursion`.

src/relax/transform/inline_functions.cc
```diff
@@ -0,0 +1,228 @@
...
+      // from the MatchCast inlined portion.
+      //
+      // Option 4: Define a VarBinding `param[i] = args[i]`, with
+      // CanonicalizeBindings to handle conversion of Var to
+      // DataflowVar, and tvm::relax::Bind to handle substitution of
+      // symbolic variables.  This would result in a well-formed Relax
+      // function, with no duplicate definitions of symbolic
+      // variables.
+      //
+      // This implementation uses Option 4.
```
slyubomirsky, line 127: Very clever solution to avoid dealing with dataflow blocks. It's good that `CanonicalizeBindings` has become a single source of truth for dealing with these issues.

tests/python/relax/test_inline_functions.py
```diff
@@ -0,0 +1,363 @@
...
+    tvm.ir.assert_structural_equal(expected, after)
+
+
+def test_subroutine_with_symbolic_vars():
+    """Inlined subroutines should use the caller's symbolic variables
+
+    Before inlining, the subroutine and the caller have distinct
+    `tir::Var` for each symbolic variables.  After inlining, only the
+    caller's `tir::Var` symbolic variables should remain.
+    """
```
slyubomirsky, line 173: What logic in the pass handles this replacement? The use of CanonicalizeBindings per "option 4"?

tests/python/relax/test_inline_functions.py
```diff
@@ -0,0 +1,363 @@
...
+            if cond:
+                Out = Before.subroutine()
+            else:
+                Out = R.const(0, "int64")
+
+            return Out
+
+    with pytest.raises(Exception):
+        Before["main"].inline_functions({"subroutine": Before["subroutine"]})
+
```
slyubomirsky, line 360: A couple of cases that are also worth testing:
1. Inlining within an inlined function
2. Making sure that mutually recursive functions can't be inlined either (since if you can inline within an inlined function, mutual recursion would also cause infinite inlining)

slyubomirsky (changes requested):
Overall, this change looks great, and it is a long-awaited one. My only request is a couple of test cases to make sure this is as general as it sounds like it should be. Excellent work.

src/relax/transform/inline_functions.cc
Lunderberg: Thank you.  I may have gone a bit overboard with this comment, but I wanted to make sure that it was "clever robust" and not "clever unmaintainable".  I'm also really liking the use of CanonicalizeBindings as a general utility, since the rules about `DataflowVar` have too many edge cases to duplicate the logic everywhere.

tests/python/relax/test_inline_functions.py
Lunderberg: Good catch.  I have test cases for theses two cases for the `InlinePrivateFunctions` transform, but not for the explicit user-driven calls.

Lunderberg: Good catch, thank you!  The inlining within an inlined function should already be tested in `test_inline_nested_subroutine_calls`.  The second case I had missed, and have now added `test_error_when_inlining_mutually_recursive_functions`.

src/relax/transform/inline_functions.cc
slyubomirsky: Yes, exactly.

tests/python/relax/test_inline_functions.py
slyubomirsky: Ah, very nice.

slyubomirsky (approved):
Thank you for responding to my feedback. This is a very good change to have.

PR #16195, opened 2023-12-01
[Unity] Improved error checking for DataflowBlock in nested SeqExpr

Lunderberg:
Prior to this commit, the normalizer printed a warning whenever a lowering pass produced a `SeqExpr` containing a non-dataflow `BindingBlock` within a `DataflowBlock`.  This intermediate is only ill-formed if the non-dataflow `BindingBlock` makes use of any `DataflowVar` instances, as the `BindingBlock` can otherwise be hoisted out.

This commit explicitly checks for use of `DataflowVar` in these cases, as this is most likely due to erroneous use of `BindingBlock` where `DataflowBlock` should be used, and changes the severity from `WARNING` to `FATAL`.  If there are no `DataflowVar` instances used in the `BindingBlock`, no warning is required.

### DIFF
--- a/src/relax/ir/block_builder.cc
+++ b/src/relax/ir/block_builder.cc
@@ -892,8 +892,7 @@ class Normalizer : public BlockBuilderImpl, private ExprFunctor<Expr(const Expr&
                     << "The variable " << binding->var << " is defined within a DataflowBlock, "
                     << "but is bound to a SeqExpr that contains non-dataflow BindingBlocks.  "
                     << "These non-dataflow BindingBlocks use the DataflowVars "
-                    << free_dataflow_vars << ", which is invalid.\n"
-                    << binding;
+                    << free_dataflow_vars << ", which is invalid.";
               }
             }
             ret.push_back(block);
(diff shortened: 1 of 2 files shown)


slyubomirsky:
I think this is a good improvement. Could you also please add a test case?

Lunderberg:
Can do, and I've added a few unit tests.  The tests show which errors can be caught during normalization, which ones cannot, and the well-formed output that would erroneously be warned about if the error-checking were more stringent.

slyubomirsky (approved):
Thank you for adding the test cases. I am a big fan of having tests for error cases.

Lunderberg:
Agreed on testing for error cases.  I tend to add the tests when writing new checks, as I'm already in the process of writing the associated tests, and therefore have a way to drive the function.  When updating/tweaking existing errors I frequently omit them, because it may not be easy/possible to write a test case using only the public APIs that exactly hits the specific error case.

(Granted, difficulty in writing specific tests is an issue in itself, and refactoring to allow better tests is a net positive.  But I tend to go down those rabbit holes with disturbing regularity, and try to cut myself off where possible.)

PR #16199, opened 2023-12-04
[Relay][Frontend] Add support for aten::concat

sweetcocoa:
I think it is a quite simple problem, 

`aten::concat` is just an alias of `aten::cat`, but it is not supported.

https://github.com/pytorch/pytorch/blob/3fbfa8cd0a5cefadb3f116c5cd0d60e96ab8c99e/aten/src/ATen/native/TensorShape.cpp#L667

If needed, I will add a minimal example to reproduce.

```
torch==2.1.1+cpu
numpy==1.23.5
```

### DIFF
--- a/tests/python/frontend/pytorch/test_forward.py
+++ b/tests/python/frontend/pytorch/test_forward.py
@@ -719,7 +719,7 @@ def test_forward_concatenate():
             b = (args[0][:, :, 1] + 3) * 11
             c = (args[0][:, :, 2] + 5) * 13
             return torch.cat([t.unsqueeze(2) for t in [a, b, c]], 2)
-    
+
     class Concatenate3(Module):
         def __init__(self):
             super().__init__()
--- a/tests/python/frontend/pytorch/test_forward.py
+++ b/tests/python/frontend/pytorch/test_forward.py
@@ -721,7 +721,11 @@ def test_forward_concatenate():
             return torch.cat([t.unsqueeze(2) for t in [a, b, c]], 2)
 
     class Concatenate3(Module):
-        # pylint: disable=missing-class-docstring
+        """
+        torch.concat is preserved as aten::concat only when in a nested module.
+        (In the most cases, It is converted to aten::cat instead of aten::concat.)
+        """
+
         def __init__(self):
             super().__init__()
 
--- a/tests/python/frontend/pytorch/test_forward.py
+++ b/tests/python/frontend/pytorch/test_forward.py
@@ -719,10 +719,27 @@ def test_forward_concatenate():
             b = (args[0][:, :, 1] + 3) * 11
             c = (args[0][:, :, 2] + 5) * 13
             return torch.cat([t.unsqueeze(2) for t in [a, b, c]], 2)
+    
+    class Concatenate3(Module):
+        def __init__(self):
+            super().__init__()
+
+            class _Concatenate(Module):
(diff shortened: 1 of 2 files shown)


mshr-h:
@sweetcocoa Thank you for your PR!
It would be nice to include a test to https://github.com/apache/tvm/blob/main/tests/python/frontend/pytorch/test_forward.py
Maybe just adding a test case to `def test_forward_concatenate():` is fine?
What do you think?  @jikechao @vvchernov

vvchernov:
Thank you @sweetcocoa for your PR! I agree the test is needed

jikechao (approved):
LGTM, Thanks!

tests/python/frontend/pytorch/test_forward.py
```diff
@@ -719,10 +719,27 @@ def forward(self, *args):
             b = (args[0][:, :, 1] + 3) * 11
             c = (args[0][:, :, 2] + 5) * 13
             return torch.cat([t.unsqueeze(2) for t in [a, b, c]], 2)
+
```
mshr-h, line 722: Please remove trailing whitespaces.

tests/python/frontend/pytorch/test_forward.py
```diff
@@ -2893,6 +2911,7 @@ def forward(self, inp):
 @tvm.testing.uses_gpu
 def test_simple_rnn():
     """test_simple_rnn"""
+
```
echuraev, line 2914: nit: I suppose that all such white spaces are redundant. Could you please remove them in this test and in the tests below?

sweetcocoa: Sorry I didn't see it, I've reverted it now.

sweetcocoa:
I don't think this CI error stems from this PR, can I try restarting it?

jikechao:
> I don't think this CI error stems from this PR, can I try restarting it?

Yes, you can comment with `@tvm-bot rerun` in the chatbox to rerun the CI test.

sweetcocoa:
@tvm-bot rerun

tests/python/frontend/pytorch/test_forward.py
```diff
@@ -720,9 +720,27 @@ def forward(self, *args):
...
+                    a = (args[0][:, :, 0] + 2) * 7
+                    b = (args[0][:, :, 1] + 3) * 11
+                    c = (args[0][:, :, 2] + 5) * 13
+                    return torch.concat([t.unsqueeze(2) for t in [a, b, c]], 2)
+
+            self.mod = _Concatenate()
+
+        def forward(self, *args):
+            return self.mod(*args)
+
```
echuraev, line 739: Why do you create a class in this way?
Will the following code work in the same way?
```suggestion
    class Concatenate3(Module):
        def forward(self, *args):
            a = (args[0][:, :, 0] + 2) * 7
            b = (args[0][:, :, 1] + 3) * 11
            c = (args[0][:, :, 2] + 5) * 13
            return torch.concat([t.unsqueeze(2) for t in [a, b, c]], 2)
```

echuraev: Thank you for your reply! Could you please in this case specify it in the class description, instead of using `# pylint: disable=missing-class-docstring`?

echuraev (approved):
LGTM! Thank you for your PR!

PR #16202, opened 2023-12-05
Fix IRModule initialization with attrs

jinhongyii:
There's no test on IRModule initialization with attr, and there's an error with it. This PR fixes the behavior.


junrushao (approved):
Agreed that ast shouldn’t be part of IRModule

PR #16203, opened 2023-12-05
[Dlight] Change max_threads on CUDA

jinhongyii:
change max number of threads per block from 256 to 1024, which makes rms_norm faster

### DIFF
--- a/python/tvm/dlight/gpu/utils.py
+++ b/python/tvm/dlight/gpu/utils.py
@@ -50,7 +50,7 @@ def suggest_threads_per_block(
     max_threads_for_dynamic_loop: int = 32,
 ) -> List[int]:
     if target.kind.name == "cuda":
-        threads = 256
+        threads = 1024
     elif target.kind.name == "rocm":
         threads = 256
     elif target.kind.name == "metal":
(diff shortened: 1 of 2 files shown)


jinhongyii:
cc: @junrushao

junrushao:
Given that it’s a global change, is it going to positively or negatively affect other operators on CUDA GPU?

jinhongyii:
It only affects reduction ops. Specifically, in llm, it only means rms_norm and layer_norm. So I believe it's safe to merge

junrushao:
Could you fix the unittests so that I could get it in?

Hzfengsy (changes requested):
It's UNSFAE. As:
1. It's global changes. Although it is only used for specific ops, but it might be used in other cases.
2. For reduction rules, we should also consider the operators that are smaller than 1024

My suggestion is adding analysis in Reduction rules, if the size is large enough, than use 1024, else pick 256

jinhongyii:
When operator is smaller than 1024, suggest_threads_per_block will return the largest thread num that is smaller than loop extent. @Hzfengsy

Hzfengsy (approved):

PR #16204, opened 2023-12-05
[Unity][Transform] Pass for automatically extracting DataflowBlocks

slyubomirsky:
This PR implements a pass that automatically extracts DataflowBlocks out of binding blocks (i.e., groups consecutive pure operations in binding blocks into a dataflow block).

Note that this pass makes use of `CanonicalizeBindings` and in the process of testing it, I encountered bugs in `CanonicalizeBindings` that I deal with in here:
1. `CanonicalizeBindings` dealt with the case of a var that appeared only inside a DataflowBlock (demoted to a dataflow var), but it made a mistake in cases where a var was used only inside DataflowBlocks, plural, meaning outside the one where it was defined (thus it should be exposed as an output and thus not demoted to a dataflow var).
2. `CanonicalizeBindings` did not handle inner functions correctly. It will correctly recurse into inner functions and also treat any captured vars as appearing outside their original dataflow block.


src/relax/transform/extract_dataflow_blocks.cc
```diff
@@ -0,0 +1,140 @@
...
+  explicit DataflowBlockExtractor(size_t min_size) : ExprMutator(), min_size_(min_size) {}
+
+  Expr VisitExpr_(const SeqExprNode* seq) override {
+    Array<BindingBlock> new_blocks;
+    Expr new_body = VisitExpr(seq->body);
+    bool changed = !new_body.same_as(seq->body);
+    for (auto block : seq->blocks) {
+      BindingBlock new_block = this->VisitBindingBlock(block);
+      changed = changed || !new_block.same_as(block);
+      if (new_block.as<DataflowBlock>()) {
```
Lunderberg, line 45: Would be good to comment that we don't need to explicitly merge a new dataflow block with a preceding dataflow block, as that is handled during `IRNormalizer`.

src/relax/transform/extract_dataflow_blocks.cc
```diff
@@ -0,0 +1,140 @@
...
+            // extend the streak
+            dataflow_bindings.push_back(binding);
+          } else {
+            // this is the end of the streak
+            dataflow_streak = false;
+
+            // if the df block is below the minimum length, combine the blocks
+            // and reset the dataflow collection
+            if (dataflow_bindings.size() < min_size_) {
+              non_dataflow_bindings.insert(non_dataflow_bindings.end(), dataflow_bindings.begin(),
```
Lunderberg, line 80: What happens if `dataflow_bindings.size() < min_size_`, but the preceding block was also a dataflow block?  As a user, I'd want it to be merged into the preceding block anyways, but it looks like it would be left as-is.

src/relax/transform/extract_dataflow_blocks.cc
```diff
@@ -0,0 +1,140 @@
...
+      // for a binding block, attempt to extract dataflow blocks inside
+      auto binding_block = Downcast<BindingBlock>(new_block);
+      bool dataflow_streak = false;
+      Array<Binding> dataflow_bindings;
+      Array<Binding> non_dataflow_bindings;
+      for (size_t i = 0; i < binding_block->bindings.size(); i++) {
+        auto binding = binding_block->bindings[i];
+        Expr value = GetBoundValue(binding);
+        // dataflow values: not an if node and not an impure call
+        bool is_dataflow =
```
Lunderberg, line 59: This might be a bit cheeky of a way to implement the functionality, but what if we just made a `DataflowBlock` for every bound value that satisfies the `is_dataflow` check?  The `IRNormalizer` combines adjacent blocks together, so the adjacent dataflow blocks would all get merged together at that point.

tests/python/relax/test_transform_extract_dataflow_blocks.py
```diff
@@ -0,0 +1,268 @@
...
+                v = R.divide(w, z)
+                R.output(inner_func, v, w)
+            q = inner_func(w, v)
+            with R.dataflow():
+                a = R.multiply(q, q)
+                b = R.add(a, a)
+                c = R.multiply(b, a)
+                R.output(c)
+            return c
+
```
Lunderberg, line 265: Can we add two test cases for making dataflow blocks when the individual streak doesn't satisfy the `min_size_` requirement, but the presence of a dataflow block next to the potentially-dataflow binding still avoids making a tiny dataflow block.

```python
class TestMergeWithPrecedingDataflowBlock(ExtractCompare):
    @I.ir_module
    class Before:
        @R.function
        def main(x: R.Tensor, y: R.Tensor) -> R.Tensor:
            with R.dataflow():
                z = R.add(x, y)
                w = R.multiply(z, y)
                R.output(w)

            # The single binding of `v = R.add` would normally not be
            # enough to make a dataflow block, as `1 < min_size == 3`.
            v = R.add(w, x)
            return v

    @I.ir_module
    class Expected:
        @R.function
        def main(x: R.Tensor, y: R.Tensor) -> R.Tensor:
            with R.dataflow():
                z = R.add(x, y)
                w = R.multiply(z, y)
                # However, it occurs just after an existing dataflow
                # block, and can be merged into it.
                v = R.add(w, x)
                R.output(v)
            return v


class TestMergeWithNextDataflowBlock(ExtractCompare):
    @I.ir_module
    class Before:
        @R.function
        def main(x: R.Tensor, y: R.Tensor) -> R.Tensor:
            # The single binding of `z = R.add` would normally not be
            # enough to make a dataflow block, as `1 < min_size == 3`.
            z = R.add(x, y)

            # However, it occurs just before an existing dataflow
            # block, and can be merged into it.

            with R.dataflow():
                w = R.multiply(z, y)
                v = R.add(w, x)
                R.output(v)
            return v

    @I.ir_module
    class Expected:
        @R.function
        def main(x: R.Tensor, y: R.Tensor) -> R.Tensor:
            with R.dataflow():
                z = R.add(x, y)
                w = R.multiply(z, y)
                v = R.add(w, x)
                R.output(v)
            return v
```

Lunderberg (commented):
Overall, looks good.  As a general comment, can we rename `ExtractDataflowBlock` to `ConvertToDataflowBlock`?  My first impression from the name `ExtractDataflowBlock` was that the input would contain a dataflow block, which would then be extracted out into some other structure.  It took going to the test cases (which are excellent, BTW) to see that it is instead generating dataflow blocks.

tests/python/relax/test_transform_extract_dataflow_blocks.py
```diff
@@ -0,0 +1,268 @@
...
+                R.output(f)
+            return f
+
+
+class TestTreatNonCallAsPure(ExtractCompare):
+    @I.ir_module
+    class Before:
+        @R.function
+        def main(t: R.Tuple(R.Tensor, R.Tensor)) -> R.Tensor:
+            x = t[0]
```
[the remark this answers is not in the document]
slyubomirsky: Yeah, good idea.

tests/python/relax/test_transform_extract_dataflow_blocks.py
```diff
@@ -0,0 +1,268 @@
...
+                z = R.add(x, y)
+                w = R.multiply(z, z)
+                R.output(w)
+            return w
+
+
+class TestInnerFunction(ExtractCompare):
+    @I.ir_module
+    class Before:
+        @R.function(pure=False)
```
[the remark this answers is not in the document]
slyubomirsky: You're right, the pass does care about purity, so I'll add those cases.

src/relax/transform/canonicalize_bindings.cc
```diff
@@ -118,9 +119,28 @@ class CanonicalizePlanner : public ExprVisitor {
   }
 
  private:
+  void VisitExpr_(const FunctionNode* func) override {
+    // for functions, treat any free vars as used outside their home DF block
```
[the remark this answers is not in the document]
slyubomirsky: In the specification, I actually wrote that we should not allow DF vars to be captured at all. Unless there's a compelling need to deal with that extra complexity, I'm happy to stick with that.

src/relax/transform/canonicalize_bindings.cc
```diff
@@ -170,17 +191,26 @@ class CanonicalizePlanner : public ExprVisitor {
...
+    // where it was defined, it also needs to be exposed, so also we treat that as
+    // used outside of a dataflow block
+    if (!inside_dataflow_ ||
+        (def_blocks_.count(var_ref) && !current_block_.same_as(def_blocks_.at(var_ref)))) {
+      used_outside_home_dataflow_.insert(GetRef<Var>(var));
     }
   }
 
   bool inside_dataflow_{false};
+  BindingBlock current_block_;
```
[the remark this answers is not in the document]
slyubomirsky: Yeah, I was going to do that for a second reason too: There is no "current block" when you're looking at the `body` field of a SeqExpr. This could lead to weird results.

python/tvm/relax/transform/transform.py
```diff
@@ -262,6 +262,23 @@ def LambdaLift() -> tvm.ir.transform.Pass:
     return _ffi_api.LambdaLift()
 
 
+def ExtractDataflowBlocks(min_size: int = 3) -> tvm.ir.transform.Pass:
```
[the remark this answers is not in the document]
slyubomirsky: It was an arbitrary call on my part, so sure I can change it to 2.

src/relax/transform/extract_dataflow_blocks.cc
slyubomirsky: Good point. I rely on normalization throughout this pass.

src/relax/transform/extract_dataflow_blocks.cc
slyubomirsky: That's a good point. Currently it's merged into the following (non-dataflow) block. This is a case worth tracking, I'll think about how to get it.

tests/python/relax/test_transform_extract_dataflow_blocks.py
slyubomirsky: Great idea. I'll have to handle that case though.

src/relax/transform/extract_dataflow_blocks.cc
slyubomirsky: Cheeky indeed. I'll keep it as-is for now but if we run into trouble, that's a good option haha.

src/relax/transform/canonicalize_bindings.cc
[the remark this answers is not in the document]
Lunderberg: That sounds like a plan, and I agree that it's not currently worth the extra complexity.

src/relax/transform/extract_dataflow_blocks.cc
Lunderberg: Sounds good.  I was thinking that this might be an easy way to implement the folding into adjacent DataflowBlocks, since the merging is already handled there.  The downside is that it wouldn't allow customization of `min_size` as part of the transform.

tests/python/relax/test_transform_extract_dataflow_blocks.py
```diff
@@ -0,0 +1,268 @@
...
+                v = R.add(w, x)
+                R.output(v)
+
+            if R.const(True):
+                with R.dataflow():
+                    q = R.multiply(v, v)
+                    a = R.add(q, q)
+                    b = R.multiply(a, a)
+                    R.output(b)
+                # weird but the parser requires this construct
```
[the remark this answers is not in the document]
slyubomirsky: Yeah, I was just hoping having `b = R.multiply(a, a)` be the last assignment would work. It's not the worst thing in the world that the parser expects a specific assignment like that, just a little weird (per the comment)

tests/python/relax/test_transform_extract_dataflow_blocks.py
slyubomirsky: Done, thank you

slyubomirsky:
Thanks for the excellent feedback, @Lunderberg. I will also add a few more test cases per your suggestions.

tests/python/relax/test_transform_convert_dataflow.py
```diff
@@ -0,0 +1,493 @@
...
+
+        @R.function
+        def shapes() -> R.Shape:
+            s1 = R.shape((1, 2, 3))
+            s2 = R.shape((4, 5, 6))
+            s3 = R.shape((7, 8, 9))
+            return s3
+
+        @R.function
+        def prim_values():
```
slyubomirsky, line 201: Here's a bug I encountered, probably with the parser: Putting `-> R.Prim` as the return would segfault. Maybe it expects `-> R.Prim()`? Not sure why it would segfault though.

slyubomirsky: It appears that it is not permitted to leave both optional fields empty. I guess it's not such a serious bug if it's not permitted anyway but it would be better to have a nicer error than a segfault.

Lunderberg: Whoa, that's an impressive segfault.  It looks like the sequence of events is as follows:

1. We hit [this TypeError](https://github.com/apache/tvm/blob/unity/python/tvm/script/parser/relax/entry.py#L452), which gives a diagnostic that the `R.Prim` needs to have one of the two argument types.
2. The TypeError is caught [here](https://github.com/apache/tvm/blob/unity/python/tvm/script/parser/relax/parser.py#L102), where it is passed to the diagnostic context to pretty-print and re-throw.
3. After showing the error message, [this conditional](https://github.com/apache/tvm/blob/unity/src/ir/diagnostic.cc#L129) resets the diagnostic renderer and raises an exception.
4. The DiagnosticError is caught [here](https://github.com/apache/tvm/blob/unity/python/tvm/script/parser/relax/parser.py#L111), where it is passed to the diagnostic context to pretty-print and re-throw.
5. When reaching [here](https://github.com/apache/tvm/blob/unity/src/ir/diagnostic.cc#L118), `(*this)->renderer` has been replaced with a default constructed instance.  The default-constructed instance from step (3) has a nullptr for the renderer, which causes the segfault.

It looks like there's a lot of `except DiagnosticError: raise` scattered throughout the parser to prevent this from occurring (by preventing step (4) from occurring), but these `eval_struct_info` and `eval_struct_info_proxy` don't have it.  That seems like a rather fragile way to avoid some pretty opaque errors, so I'm wondering if we want to change that in the long term.

Lunderberg: I think a more stable way to have the pretty-printing from TVMScript would be to remove all of the try/catch that only exist to re-throw as DiagnosticError, and instead allow the exception to bubble up to the entry point.  Catching the exception at the entry point, the `exc.__traceback__` can then be inspected.  The TVMScript location of the error is then the lowest stack frame that contains a variable of type `doc.AST`, and that object can be provided to `Parser.report_error` as the location of the error in the TVMScript.

This would avoid requiring every node to have its own try/catch/re-throw, and avoid the segfault that we ran into here.

@junrushao Since you've spent more time on the TVMScript parser than I have, what are your thoughts on this idea?

Lunderberg (approved):
Thank you for making the updates, and looks good to me!

PR #16209, opened 2023-12-06
[Unity] De-duplicate calls to TensorStructInfo constructor

Lunderberg:
Prior to this commit, the `TensorStructInfo` constructor took as input a `VDevice`, with a default value of `VDevice()`, and then assigned it to a member of type `Optional<VDevice>`.  This commit changes the constructor signature to match the member type, which will remove unnecessary type conversions when copying a `TensorStructInfo`.

### DIFF
--- a/src/relax/op/tensor/manipulate.cc
+++ b/src/relax/op/tensor/manipulate.cc
@@ -71,19 +71,11 @@ StructInfo InferStructInfoBroadcastTo(const Call& call, const BlockBuilder& ctx)
 
   // Trust the input target shape when there is no possibility to do any compile-time check.
   if (!data_sinfo->shape.defined()) {
-    if (data_sinfo->vdevice.defined()) {
-      return TensorStructInfo(/*shape=*/call->args[1], data_sinfo->dtype,
-                              data_sinfo->vdevice.value());
-    }
-    return TensorStructInfo(/*shape=*/call->args[1], data_sinfo->dtype);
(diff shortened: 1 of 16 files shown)


src/relax/op/tensor/manipulate.cc
```diff
@@ -967,36 +894,22 @@ StructInfo InferStructInfoSplit(const Call& call, const BlockBuilder& ctx) {
...
     Array<PrimExpr> shape = data_shape->values;
     shape.Set(axis, split_len);
-    if (data_sinfo->vdevice.defined()) {
-      std::vector<StructInfo> output_sinfo(
-          n_section - 1,
-          TensorStructInfo(ShapeExpr(shape), data_sinfo->dtype, data_sinfo->vdevice.value()));
-    }
-    std::vector<StructInfo> output_sinfo(n_section - 1,
-                                         TensorStructInfo(ShapeExpr(shape), data_sinfo->dtype));
+    std::vector<StructInfo> output_sinfo(
```
Lunderberg, line 906: This is the only location in the PR that should make a change on the observed behavior.  Prior to this commit, whenever `data_sinfo->vdevice` is defined, the body of the conditional constructed an instance of `std::vector<StructInfo>`, which is then destructed without being used at the end of the conditional.

With this commit, the `data_sinfo->device` is propagated to the output, which I believe to be the intent of the original.

slyubomirsky (approved):
Seems to be a reasonable change. This suggests that I should discuss VDevice in the Relax spec.

yongwww (approved):

PR #16214, opened 2023-12-08
[Python] Fix setup.py for inplace build

MasterJH5574:
Previously, `python/setup.py` missed to take the inplace build case into account when doing file operations (such as finding paths or copying files), which makes command `make cython` fail.

This PR fixes the issue by checking inplace builds.


MasterJH5574:
@tvm-bot rerun

tqchen (approved):

PR #16219, opened 2023-12-10
[Unity] Speed up NormalizeGlobalVar

Hzfengsy:
The current implementation of NormalizeGlobalVar couples the `AddPublicFunction` and checks if the module needs renaming, which causes the modules that don't need renaming to run unnecessary code, such as `AddFunction` to the new module.

This patch separates the two steps so that the modules that don't need renaming can skip the `AddPublicFunction` step.

cc @Ubospica


yzh119 (approved):
LGTM, thank you for the improvement!

Ubospica (approved):
LGTM

PR #16220, opened 2023-12-10
[Unity] Support out dtype for nn.Linear and nn.MultiLinear

CharlieFRuan:
This is motivated by supporting models like GPTNeoX that have the field `ffn_out_dtype` in config, making us set it as `out_dtype`.

After the `matmul`, `x` would have type `out_dtype` (`x = op.matmul(x, w, out_dtype=self.out_dtype`).

When `bias=True`, we want `bias` to be of type `out_dtype`, even if we call `to()`; otherwise, we may have a type mismatch when computing `x + self.bias`.

Same idea applies for `nn.MultiLinear`.

### DIFF
--- a/python/tvm/relax/frontend/nn/modules.py
+++ b/python/tvm/relax/frontend/nn/modules.py
@@ -192,7 +192,9 @@ class MultiLinear(Module):
         self.out_dtype = out_dtype
         self.weight = Parameter((total_out_features, in_features), dtype)
         if bias:
-            self.bias = Parameter((total_out_features,), dtype)
+            self.bias = Parameter(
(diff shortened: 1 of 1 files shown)


CharlieFRuan:
Tested with downstream mlc-llm for:
- `q0f16` and `q0f32` RedPajama (have `out_dtype` as `float32` for both)
- `q0f16` and `q0f32` GPT2 (no out_dtype)
- Though verified that llama-2 still works (which uses MultiLinear), there is no current downstream use case where we have a bias for MultiLinear, but it should be safe to support that as well.

cyx-6 (approved):
LGTM!

PR #16227, opened 2023-12-12
[ARITH] Simplify nested if_then_else when constant is appearing in then_expr

rutkoor:
When constant is appearing in the else_expr, Simplifier pass is able to simplify the nested if_then_else expression and eliminate the redundant if_then_else. But, when constant is appearing in the then_expr, it is unable to simplify the nested if_then_else statements.

The changes in the PR will address the above problem.

### DIFF
--- a/tests/python/tir-transform/test_tir_transform_simplify.py
+++ b/tests/python/tir-transform/test_tir_transform_simplify.py
@@ -1757,5 +1757,17 @@ class TestBufferShapeConstraintWithOffset(BaseBeforeAfter):
         A[T.int64(1)] = T.float32(0)
 
 
+class TestNestedIfElimination(BaseBeforeAfter):
+    def before(a: T.Buffer((2, 8), "int32"), b: T.Buffer((2, 8), "int32")):
+        for i0, j0 in T.grid(2, 8):
+            b[i0, j0] = T.if_then_else(
+                i0 == 1 and 6 <= j0, 0, T.max(0, T.if_then_else(i0 == 1 and 6 <= j0, 0, a[i0, j0]))
+            )
+
+    def expected(a: T.Buffer((2, 8), "int32"), b: T.Buffer((2, 8), "int32")):
+        for i0, j0 in T.grid(2, 8):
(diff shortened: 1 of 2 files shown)


rutkoor:
@tqchen , @junrushao , @Lunderberg

tests/python/tir-transform/test_tir_transform_simplify.py
```diff
@@ -1757,5 +1757,27 @@ def expected(a: T.handle):
         A[T.int64(1)] = T.float32(0)
 
 
+class TestNestedIfElimination(BaseBeforeAfter):
+    def before(a: T.Buffer((2, 8), "int32"), b: T.Buffer((2, 8), "int32")):
+        for i0, j0 in T.grid(2, 8):
+            with T.block("P"):
```
Lunderberg, line 1763: Does the test case depend on the `BlockNode` being present?  If not, we should remove it.  (Removing the `with T.block`, `T.axis.remap`, `T.reads`, and `T.writes` lines, then using `i0` and `j0` directly.)

Lunderberg (changes requested):
This change looks reasonable, to have parallel optimizations between the true and false branches.  I have one request for the unit test, to remove unnecessary parts of the unit test, and can approve it after that.

I'm a bit confused as to the pre-existing implementation, though.  Since this is a simplification of an expression, I would have expected this to be part of the overall `Analyzer` class, with any context provided by `Analyzer::Bind` or `With<ConstraintContext>`.  The `IRMutatorWithAnalyzer` shouldn't be required for any simplifications.

For the long-term, we should probably move the `IterMapSimplifyWithContext` out of `IRMutatorWithAnalyzer`, and into an object held by `arith::Analyzer` (i.e. an `IterMapSimplifier`).  That way, it would receive context from the usual `Bind` and `With<ConstraintContext>` mechanisms, wouldn't need to re-parse the predicate for each simplification, and would be compatible with other sub-analyzers.

tests/python/tir-transform/test_tir_transform_simplify.py
rutkoor: Removed.

Lunderberg (approved):

quic-sanirudh (approved):

PR #16231, opened 2023-12-13
[Target] Add Jetson AGX Orin tags

rym-h:
cc @junrushao @zxybazh @masahi
Looking at the [NVIDIA documentation](https://www.nvidia.com/content/dam/en-zz/Solutions/gtcf21/jetson-orin/[nvidia-jetson-agx-orin-technical-brief.pdf](https://www.nvidia.com/content/dam/en-zz/Solutions/gtcf21/jetson-orin/nvidia-jetson-agx-orin-technical-brief.pdf)), there are 32GB and 64GB models. The former has 8-core and the latter has 12-core. So, I defined them with different tags.


zxybazh (approved):
LGTM, thanks for introducing the new tags.

PR #16232, opened 2023-12-13
[Unity] Fix legalizing strided slice

vinx13:
strided slice with symbolic bounds is supported in topi, this PR remove the check for dynamic shape in legalization. It also fixed calculation of output shape when stride != 1

cc @Lunderberg @Hzfengsy

### DIFF
--- a/tests/python/relax/test_transform_legalize_ops_index_linear_algebra.py
+++ b/tests/python/relax/test_transform_legalize_ops_index_linear_algebra.py
@@ -168,10 +168,34 @@ def test_strided_slice_symbolic_sliced_axis():
             n = T.int64()
             gv: R.Tensor((2, n), "float32") = R.strided_slice(x, axes=[0], begin=[1], end=[8], strides=[3])
             return gv
+
+    @I.ir_module
+    class Expected:
+        @T.prim_func(private=True)
+        def strided_slice(var_A: T.handle, var_T_dynamic_strided_slice_with_axes: T.handle):
+            T.func_attr({"tir.noalias": T.bool(True)})
+            m, n = T.int64(), T.int64()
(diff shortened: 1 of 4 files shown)


tests/python/relax/test_transform_legalize_ops_index_linear_algebra.py
```diff
@@ -210,6 +210,38 @@ def strided_slice(var_rxplaceholder: T.handle, var_T_strided_slice_with_axes: T.
...
 
 
+def test_strided_slice_symbolic_bound():
+    # fmt: off
+    @tvm.script.ir_module
+    class StridedSlice:
+        @R.function
+        def main(x: R.Tensor((10, "n"), "float32")) -> R.Tensor((3, "n"), "float32"):
+            n = T.int64(is_size_var=True)
+            gv: R.Tensor((3, n), "float32") = R.strided_slice(x, axes=[0, 1], begin=[1, 0], end=[8, n], strides=[3, 1])
```
Lunderberg, line 220: It took me a moment to catch that this one test was validating both the symbolic bound, as the second dimension of the output shape is `ceildiv(n-0, 1) = n`, and the bugfix, as the first dimension of the output shape is `ceildiv(8-1, 3) = 3`.  Could we have these two behaviors be explicitly tested in separate unit tests?

Lunderberg (approved):
Looks good to me!  One possible change to improve the readability of the unit tests.

vinx13:
@Lunderberg I'm combining the test because I have to use symbolic bound to trigger the `topi::dynamic_strided_slice` path

Lunderberg:
@vinx13 That makes sense, and thank you for the explanation.  All looks good, and I think we're just waiting on CI.

PR #16234, opened 2023-12-13
[Support] Add Interrupt Handling in Pipe

sunggg:
cc. @tqchen @LeshengJin


src/support/error_handling.h
```diff
@@ -0,0 +1,77 @@
...
+#ifndef TVM_SUPPORT_ERR_HANDLING_H_
+#define TVM_SUPPORT_ERR_HANDLING_H_
+#include <errno.h>
+
+namespace tvm {
+namespace support {
+   /*!
+   * \return last error of socket operation
+   */
+  static int GetLastErrorCode() {
```
tqchen, line 33: This needs to be passed as a second template function parameter to RetryCallOnEINTR

tqchen: We need specialized function in socket and pipe

For socket it should be `WSAGetLastError()` in windows, for pipe it should be `GetLastError()`

src/support/pipe.h
```diff
@@ -67,8 +69,8 @@ class Pipe : public dmlc::Stream {
     ICHECK(ReadFile(handle_, static_cast<TCHAR*>(ptr), size, &nread, nullptr))
```
tqchen, line 69: on windows, create a lambda function that follows the same convention

```c++
auto fread = [&]() {
    if (!ReadFile(handle_, static_cast<TCHAR*>(ptr), size, &nread, nullptr)) return -1;
    if (nread != nwrite) return -1;
    return nread; 
};

```

tqchen (changes requested):

src/support/error_handling.h
```diff
@@ -0,0 +1,77 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/*!
+ * \file err_handling.h
+ * \brief Common error handling functions for socket.h and pipe.h
+ */
+#ifndef TVM_SUPPORT_ERR_HANDLING_H_
```
tqchen, line 24: errno_handling (specific to handle error number)

PR #16235, opened 2023-12-13
[Unity] Fix IndexDataTypeNormalizer so that it correctly handles corner case

jinhongyii:
There is a corner case: A[i] = i for IndexDataTypeNormalizer. In current implementation, the var in buffer index will be rewritten, but the var in BufferStore value will not, which cases a conflict and produces free var. I change it to two iterations: the first collects var_remap_, and the second rewrites vars, so that in A[i] = i, both var will be rewritten.


jinhongyii:
cc: @tqchen

tqchen (approved):

PR #16236, opened 2023-12-14
[TIR] ConvertSSA process entry func first

jinhongyii:
ConvertSSA is called in SplitHostDevice. The vars which only exists in T.launch_thread extent in device function must be the same as host function, so we need to let ConvertSSA process host function first to avoid var redefine.


jinhongyii:
cc: @tqchen

tqchen (approved):

PR #16242, opened 2023-12-14
[Codegen] Fix if_then_else codegen

spectrometerHBH:
- Evaluating `op.if_then_else` with `a?b:c` is potentially dangerous as it might evaluate expressions in branches without a guard of condition.

- While loop condition expression might require several var bindings generated, so we correct the while loop codegen from `while (cond)` to `while (True)`

### DIFF
--- a/src/runtime/ndarray.cc
+++ b/src/runtime/ndarray.cc
@@ -99,7 +99,7 @@ void ArrayCopyToBytes(const DLTensor* handle, void* data, size_t nbytes) {
 TVM_REGISTER_GLOBAL("mlc.show_dltensor").set_body_typed([](NDArray arr) {
   auto handle = arr.operator->();
   LOG(INFO) << "data: " << handle->data;
-  LOG(INFO) << "device:" <<  handle->device.device_type << " " << handle->device.device_id;
+  LOG(INFO) << "device:" << handle->device.device_type << " " << handle->device.device_id;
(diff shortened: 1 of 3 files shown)


src/runtime/ndarray.cc
```diff
@@ -96,6 +96,27 @@ void ArrayCopyToBytes(const DLTensor* handle, void* data, size_t nbytes) {
   DeviceAPI::Get(handle->device)->StreamSync(handle->device, nullptr);
 }
 
+TVM_REGISTER_GLOBAL("mlc.show_dltensor").set_body_typed([](NDArray arr) {
```
junrushao, line 99: remove debug code

vinx13:
isn't short circuiting the defined behavior for ternary expression?

spectrometerHBH:
> isn't short circuiting the defined behavior for ternary expression?

When generating codes for `a?b:c`, there might be bindings for b or c generated ahead. These binding codes are not protected by condition a.

jinhongyii (approved):

ysh329:
> > isn't short circuiting the defined behavior for ternary expression?
> 
> When generating codes for `a?b:c`, there might be bindings for b or c generated ahead. These binding codes are not protected by condition a.

Hi, bohan, I recently am supporting Ternary Operator(like cond?true_branch:false_branch) in our device. I don't understand this modification. Can u explain this with a case? Thanks a lot!

PR #16243, opened 2023-12-14
[TFLite] Add support for quantized mirror pad

lhutton1:
Removes the restriction on converting quantized mirror pad in the front-end and adds tests to support this change.

cc @leandron @neildhickey


leandron (approved):
Thanks @lhutton1!

PR #16245, opened 2023-12-14
[Unity] Fix IndexDataTypeNormalizer so that it correctly handles corner case

jinhongyii:
Fixes https://github.com/apache/tvm/pull/16235


jinhongyii:
cc: @tqchen

PR #16246, opened 2023-12-14
[Unity][VM] Allow `pipeline=None` in `relax.build`

junrushao:
In sophisticated usecases, developers may want full control of the relax compilation pipeline. A typical workflow looks like:
- Step 1. Apply a custom relax.pipeline to IRModule
- Step 2. Invoke `relax.build` to export relax Executable but without any additional passes;
- Step 3. Manipulate the relax Executable.

In the three steps above, both Step 1 and 3 are already supported, and this PR aims to enable Step 2, which allows `relax.build` to apply a `None` pipeline which contains no passes.

Note: as an advanced behavior, this is enabled only when explicitly setting `pipeline=None` and thus backward compatibility is fully preserved.

### DIFF
--- a/python/tvm/relax/pipeline.py
+++ b/python/tvm/relax/pipeline.py
@@ -95,7 +95,7 @@ def default_build_pipeline():
                 transform.AttachGlobalSymbol(),
             ],
         )
-        mod = seq(mod._move())  # pylint: disable=protected-access
+        mod = seq(mod)
         return mod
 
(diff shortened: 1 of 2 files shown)


junrushao:
CC: @MasterJH5574 @tqchen

slyubomirsky:
Definitely sounds like a good idea for testing

tqchen (approved):

PR #16248, opened 2023-12-15
[Relay][TOPI] Add support for group_conv1d_transpose_ncw for generic

jonghewk:
This PR adds support for `group_conv1d_transpose_ncw` generic.


quic-sanirudh (approved):
There seems to be quite a bit of code overlap between this function and the `conv1d_transpose_ncw` (groups=1) version. Could we try and avoid the duplication by extracting relevant parts into a common function.

This is just a suggestion as I feel in general it's always good to avoid code duplication.

quic-sanirudh (approved):
LGTM, Thanks for the changes

Issue #13156, opened 2022-10-20
[Bug] Cannot use std::regex with TVM when PyTorch is loaded


Mousius:
### Expected behavior

Usage of a standard C++ library (`std::regex`) causes no side effects whatsoever.

### Actual behavior

PyTorch tests to suddenly fail with invalid pointer errors if we use `std::regex`:
```
[2022-10-19T14:38:46.286Z] tests/python/frontend/pytorch/qnn_test.py::test_serialized_modules free(): invalid pointer
[2022-10-19T14:38:46.286Z] Fatal Python error: Aborted
```

### Environment

TVM CI / https://github.com/apache/tvm/pull/12454 / https://ci.tlcpack.ai/blue/rest/organizations/jenkins/pipelines/tvm/branches/PR-12454/runs/10/nodes/484/steps/1515/log/?start=0

### Steps to reproduce

Use `std::regex` in your code, as demonstrated by https://github.com/apache/tvm/pull/12454. May be related to https://github.com/apache/tvm/issues/9362 and https://github.com/pytorch/pytorch/issues/49460.

### Triage

Please refer to the list of label tags [here](https://github.com/apache/tvm/wiki/Issue-Triage-Labels) to find the relevant tags and add them below in a bullet format (example below).

* needs-triage
* frontend:pytorch

PR #16249, opened 2023-12-15
[Unity] Avoid to use `std::regex`

Hzfengsy:
`std::regex` in TVM codebase may cause a symbol conflict with PyTorch, we temporarily disable it before we find a better solution, meanwhile the current usage of `std::regex` is not necessary.

cc @tqchen

### DIFF
--- a/src/node/script_printer.cc
+++ b/src/node/script_printer.cc
@@ -21,7 +21,7 @@
 #include <tvm/node/script_printer.h>
 #include <tvm/runtime/registry.h>
 
-#include <regex>
+#include <algorithm>
 
 namespace tvm {
 
@@ -38,8 +38,17 @@ std::string TVMScriptPrinter::Script(const ObjectRef& node, const Optional<Print
 }
 
 bool IsIdentifier(const std::string& name) {
-  static const std::regex kValidIdentifier("^[a-zA-Z_][a-zA-Z0-9_]*$");
-  return std::regex_match(name, kValidIdentifier);
+  // Python identifiers follow the regex: "^[a-zA-Z_][a-zA-Z0-9_]*$"
+  // `std::regex` would cause a symbol conflict with PyTorch, we avoids to use it in the codebase.
+  //
+  // We convert the regex into following conditions:
+  // 1. The name is not empty.
+  // 2. The first character is either an alphabet or an underscore.
+  // 3. The rest of the characters are either an alphabet, a digit or an underscore.
+  return name.size() > 0 &&                            //
+         (std::isalpha(name[0]) || name[0] == '_') &&  //
+         std::all_of(name.begin() + 1, name.end(),
+                     [](char c) { return std::isalnum(c) || c == '_'; });
 }
 
 PrinterConfig::PrinterConfig(Map<String, ObjectRef> config_dict) {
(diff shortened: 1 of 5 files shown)


tqchen:
let us update the testcase for now given the regex feature is not freq used yet here

junrushao (approved):

Lunderberg:
Is there a reproducible test case that can trigger the symbol conflict?  I don't see any in this PR, and haven't run into this issue in the past.

I have multiple upcoming PRs that use `std::regex`, and while I would like to avoid causing breakage from doing so, there aren't enough details here to try to reproduce the issue.

tqchen:
@Lunderberg this only happens when tvm runs together with PyTorch and in the case of the main examples. Unfortunately it is hard enough to trigger a UT. A rule of thumb is to avoid regex and use exact or prefix match for now

Lunderberg:
Thank you.  I haven't been able to reproduce the issue locally, and I want to avoid leaving this open to potential regressions in the future.

For the environments where this issue occurs, what is the value of `torch.compiled_with_cxx11_abi()`?  I suspect that we'll need the `-D_GLIBCXX_USE_CXX11_ABI=0` whenever `torch.compiled_with_cxx11_abi()` returns, but I can't find why that wouldn't have caused an issue with other libraries.

tqchen:
The previous repro happens when we use ci docker and run a particular test(pytorch frontend qnn ) AFAIK. We will run another unity main test before transition, will keep folks posted.

tqchen:
This happens when `HIDE_PRIVATE_SYMBOLS OFF`

Lunderberg:
Strange that the CI didn't run into it as reported in https://github.com/apache/tvm/issues/13156.

Lunderberg:
Ooh boy, tracking down the CXXABI rabbit hole

* gcc only provides Dual ABI support for pre-C++11 classes (e.g. `std::string`), and not for C++11 onward (e.g. `std::regex`).  The Dual ABI is only intended for compatibility with C++ code that predates C++11.  ([link](https://stackoverflow.com/a/51385122))
* Even though it should only be used for backwards compatibility, Pytorch defaults to being compiled with `USE_CXX11_ABI=0`.  This is to make sure that the wheels can be compiled under the `manylinux` docker image, required by PyPI. ([link](https://github.com/pytorch/pytorch/issues/51039))
* The `manylinux` docker image is based on CentOS 7 ([link](https://github.com/pytorch/pytorch/issues/51039#issuecomment-767012101))
* The compiler recommended by PEP-513 for use in the `manylinux` container does not support the `USE_CXX11_ABI=1` flag.  This is a CentOS 7 limitation, and is closed as cantfix ([link](https://bugzilla.redhat.com/show_bug.cgi?id=1546704)).

Lunderberg:
It looks like we can't match TVM's ABI flag to whichever one pytorch uses, because `-D_USE_CXX11_ABI=0` causes breakage with default LLVM installations.

tqchen:
do we have a strong demand for regex? I feel most of them can be replaced by a callback check, or some more restricted version of prefix matching

Lunderberg:
I think regexes are a reasonable way to extend user-facing utilities that select a name.  Prefix matching assumes that the user has control over the naming scheme in order to provide a common prefix, and callbacks may be more general than desired.  I think regex matching is a good middle ground between those two.

Lunderberg:
I've submitted https://github.com/apache/tvm/pull/16412, a linting rule to raise an error on use `#include <regex>`.  I expect this check to fail due to the current use of `std::regex`.  After the `std::regex` use is replaced, the lint can then be enabled to avoid additional regression.

PR #16251, opened 2023-12-15
[Unity][Dlight] Support dlight gemv rule on nested inner block

jinhongyii:
1. When user wants to partly specialize schedule, they can create an inner block and apply dlight on it. In the meanwhile, the user is free to schedule outer block. This PR supports such usecase on gemv rule. 
2. Fix loop order bugs in reduction rule.


jinhongyii:
cc: @tqchen @spectrometerHBH

spectrometerHBH (approved):

PR #16252, opened 2023-12-16
[Unity] Fix Cutlass Codegen for Dense

Hzfengsy:
As dense op is not used in `relax`, the cutlass codegen is not tested. However, the codegen is broken for dense op. This PR fixes the codegen.

cc @yelite @junrushao


junrushao:
CC @yelite @masahi

masahi (approved):

PR #16257, opened 2023-12-17
[Unity] Fix ccache env for `nn.SourceModule`

cyx-6:
This PR refactors the compilation of `nn.SourceModule` to enable ccache by using the relative path, instead of using absolute path. Also it adds the ccache env to not hash the directory.


junrushao:
Thanks for this patch!

junrushao:
Please rebase against the latest unity branch and let's get it merged in!

PR #16266, opened 2023-12-20
[microNPU][ETHOSU] Add fixed point for tanh

Aleksei-grovety:
Add support for calculation tanh with 16 bits fixed point (legalization of non-quantized tanh operation with quantization by fixed point multiplication).

### DIFF
--- a/python/tvm/relay/backend/contrib/ethosu/legalize.py
+++ b/python/tvm/relay/backend/contrib/ethosu/legalize.py
@@ -139,7 +139,9 @@ def get_lut_from_func(
 ) -> List[int]:
     """Calculates the values of the lookup table based on the calculation function"""
 
-    if dtype == np.int8:
+    assert dtype in ["int8", "int16"]
+
+    if dtype == "int8":
         lut_values = list()
         qmin, qmax = np.iinfo(dtype).min, np.iinfo(dtype).max
         for x in range(qmin, qmax + 1):
@@ -150,7 +152,8 @@ def get_lut_from_func(
             lut_values.append(lut_result)
 
         return lut_values
-    elif dtype == np.int16:
+    else:
+        # dtype == "int16"
         input_min = ifm_scale * (np.iinfo(np.int16).min - ifm_zp)
         input_max = ifm_scale * (np.iinfo(np.int16).max - ifm_zp)
 
@@ -194,8 +197,6 @@ def get_lut_from_func(
             lut[i] = slope + base
 
         return lut
-    else:
-        assert f"Unsupported 'dtype = {dtype}' !"
 
 
 class LutActivationRewriter(DFPatternCallback):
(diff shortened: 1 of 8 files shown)


python/tvm/relay/backend/contrib/ethosu/tir_to_cs_translator.py
```diff
@@ -877,7 +877,7 @@ def _create_npu_activation(serial_activation: spec.SerialActivation) -> vapi.Npu
         return None
     op_map = {
         "CLIP": vapi.NpuActivationOp.NONE_OR_RELU,
-        "TANH": vapi.NpuActivationOp.TABLE_LOOKUP,
+        "TANH": vapi.NpuActivationOp.TANH,
```
ekalda, line 880: This change would make it always use the NPU's builtin tanh function instead of the calculated lookup table values, making it to not match TFLite reference kernels and turning all tanh LUT value calculation into dead code.

python/tvm/relay/backend/contrib/ethosu/legalize.py
```diff
@@ -176,27 +222,39 @@ def callback(self, pre: tvm.relay.Expr, post: tvm.relay.Expr, node_map: tvm.ir.c
...
+                ifm_zero_point=input_zp,
+                ofm_scale=input_scale,
+                ofm_zero_point=input_zp,
+                activation=self.activation_type,
+            )
 
-        return identity
+            return identity
+        elif params.ifm.dtype == "int16":
+            lut_tanh = relay.const([], "int16")
```
ekalda, line 244: Seems like this is adding an empty lookup table to the identity operator?

ekalda (changes requested):
Thanks @Aleksei-grovety! I think in the current form of the patch the calculated `int16` values are not actually used...

python/tvm/relay/backend/contrib/ethosu/legalize.py
Aleksei-grovety: After the update, the calculated lookup table values for Int16 are used.

python/tvm/relay/backend/contrib/ethosu/tir_to_cs_translator.py
Aleksei-grovety: [PR](https://github.com/apache/tvm/pull/16680) was merged and the code has been updated.

python/tvm/relay/backend/contrib/ethosu/legalize.py
```diff
@@ -135,21 +135,68 @@ def get_lut_from_func(
...
+        output_max = ofm_scale * (np.iinfo(np.int16).max - ofm_zp)
+        # Create 16 bit lut following the reference
+        nbr_steps = 512
+        step = (input_max - input_min) / nbr_steps
+        half_step = step / 2
+        output_scaling_inv = (np.iinfo(np.int16).max - np.iinfo(np.int16).min + 1) / (
+            output_max - output_min
+        )
+        table_min = np.iinfo(np.int16).min
+        table_max = np.iinfo(np.int16).max
```
lhutton1, line 170: nit: these results can be used to calculate `input_min`, `input_max` etc above

lhutton1 (approved):
Looks good to me! I'll defer to @ekalda to make sure original concerns were addressed

lhutton1:
Thanks @Aleksei-grovety @ekalda!

Issue #16268, opened 2023-12-21
[Bug] [Pillow] AttributeError: 'ImageDraw' object has no attribute 'textsize' in draw_detections()


y-fjmt:
### Expected behavior
Follow the [Compile YOLO-V2 and YOLO-V3 in DarkNet Models](https://tvm.apache.org/docs/how_to/compile_models/from_darknet.html) guide to run YOLO and draw bounding boxes.

### Actual behavior
[Deprecations and removals in Pillow](https://pillow.readthedocs.io/en/stable/deprecations.html#font-size-and-offset-methodshttps://pillow.readthedocs.io/en/stable/deprecations.html#font-size-and-offset-methods)
According to the above documentation, the `ImageDraw.textsize()` method has been deprecated since pillow 9.2.0 and removed in pillow 10.0, resulting in an attribute error.

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[6], line 80
     77 names = [x.strip() for x in content]
     79 tvm.relay.testing.yolo_detection.show_detections(img, dets, thresh, names, last_layer.classes)
---> 80 tvm.relay.testing.yolo_detection.draw_detections(font_path, img, dets, thresh, names, last_layer.classes)
     82 plt.imshow(img.transpose(1, 2, 0))
     83 plt.show()

File /opt/tvm/python/tvm/relay/testing/yolo_detection.py:255, in draw_detections(font_path, im, dets, thresh, names, classes)
    253 if valid:
    254     rgb = detection["rgb"]
--> 255     label = _get_label(font_path, "".join(detection["labelstr"]), rgb)
    256     _draw_box_width(
    257         im,
    258         detection["left"],
   (...)
    265         rgb[2],
    266     )
    267     _draw_label(im, detection["top"] + detection["width"], detection["left"], label, rgb)

File /opt/tvm/python/tvm/relay/testing/yolo_detection.py:321, in _get_label(font_path, labelstr, rgb)
    319 testDraw = ImageDraw.Draw(Image.new("RGB", (1, 1)))
    320 font = ImageFont.truetype(font_path, 25)
--> 321 width, height = testDraw.textsize(labelstr, font=font)
    322 img = Image.new(
    323     "RGB", (width, height), color=(int(rgb[0] * 255), int(rgb[1] * 255), int(rgb[2] * 255))
    324 )
    325 d = ImageDraw.Draw(img)

AttributeError: 'ImageDraw' object has no attribute 'textsize'
```

[...]

PR #16269, opened 2023-12-21
[BugFix] Update pillow usage

Apricot256:
This is a pull request for Yolo detection boxes helper functions.
Tracking issue: [#16268](https://github.com/apache/tvm/issues/16268)

Update to use a replacement function for the removed function so that it can be executed in Pillow version 10 or later.

### DIFF
--- a/python/tvm/relay/testing/yolo_detection.py
+++ b/python/tvm/relay/testing/yolo_detection.py
@@ -315,10 +315,12 @@ def _get_label(font_path, labelstr, rgb):
     from PIL import ImageFont
 
     text = labelstr
+    textSize = 25
     colorText = "black"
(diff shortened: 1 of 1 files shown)


quic-sanirudh (commented):
LGTM, but could this affect backward compatibility for folks who don't use the latest Pillow version?

Should we perhaps add a version specific check to verify whether the `textlength` function exists before using it?

mshr-h:
The `textlength` function was added in version 8.0.0, released on Oct 15, 2020. It's been 3yrs since its release, so I believe we no longer need to prioritize it.

quic-sanirudh (approved):
LGTM, Thanks. As discussed in the comments, since the `textlength` function was added in 8.0 version almost 3 years ago and since this is just a testing function, I'm fine with not adding any version specific checks for this change.

PR #16279, opened 2023-12-26
replace deprecated np.int with int to avoid crash

jikechao:
Replace the deprecated API `np.int` with `int` to avoid a crash.

In NumPy v1.20.0, the `np.int` has been deprecated as shown in the [documentation](https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated)


junrushao (approved):
LGTM!

PR #16290, opened 2023-12-28
[Unity][Frontend] Introduce StructuredState for State Space modules

Hzfengsy:
We introduce the StateCache to support State Space modules. Unlike KVCache, StateCache does not need the operation `append` but requires the init value.

cc @junrushao @tqchen

### DIFF
--- a/tests/python/relax/test_frontend_nn_modules.py
+++ b/tests/python/relax/test_frontend_nn_modules.py
@@ -498,6 +498,61 @@ def test_kv_cache():
     assert_structural_equal(tvm_mod, Module, True)
 
 
+def test_state_cache():
+    @I.ir_module
+    class Module:
+        @R.function
+        def _initialize_effect() -> R.Tuple(R.Object, R.Object):
+            with R.dataflow():
+                _io: R.Object = R.null_value()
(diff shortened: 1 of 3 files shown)


tests/python/relax/test_frontend_nn_modules.py
```diff
@@ -461,6 +461,61 @@ def forward(self, x: core.Tensor) -> core.Tensor:
...
+                )
+                gv1: R.Tuple(
+                    R.Tensor((1, 8), dtype="float32"), R.Tuple(R.Object, R.Object)
+                ) = lv3, (_io, lv2)
+                R.output(gv1)
+            return gv1
+
+    class StateCacheTest(modules.Module):
+        def __init__(self) -> None:
+            self.cache = modules.StateCache(relax.op.zeros((1, 8), "float32"))
```
Hzfengsy, line 507: I use `relax.op.zeros` here instead of `nn.op.zeros` here, since there is no blockbuilder when we initialize the Module. it indeed a little bit tricky to involve raw relax expr here, love to see if there is a better solution

Hzfengsy:
Migrate to https://github.com/apache/tvm/pull/16568

PR #16293, opened 2023-12-29
[TIR] Extend DP4A tensor intrin

vincentccc:
* Update u8s8s32 s8u8s32 u8u8u32

### DIFF
--- a/src/target/source/literal/cuda_int8_t.h
+++ b/src/target/source/literal/cuda_int8_t.h
@@ -1,0 +1,64 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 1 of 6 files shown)


src/target/source/literal/cuda_int8_t.h
```diff
@@ -0,0 +1,64 @@
...
+    asm volatile ("dp4a.u32.s32 %0, %1, %2, %3;" : "=r"(ret) : "r"(srcA), "r"(srcB), "r"(c));
+    return ret;
+}
+
+__SM_61_INTRINSICS_DECL__ int __dp4a(int srcA, unsigned int srcB, int c) {
+    int ret;
+    asm volatile ("dp4a.s32.u32 %0, %1, %2, %3;" : "=r"(ret) : "r"(srcA), "r"(srcB), "r"(c));
+    return ret;
+}
+#endif /* !__CUDACC_RTC__ && defined(__CUDA_ARCH__) */
```
vinx13, line 58: these definitions are similar to `sm_61_intrinsics.hpp`, is there a reason to include them here?

vincentccc: This is because the `sm_61_intrinsics.hpp`  exclude s8xu8 and u8xs8 case

vinx13 (approved):

PR #16294, opened 2023-12-29
[Unity][Web][Fix] Fix fetchNDArray for f32-to-bf16

CharlieFRuan:
Currently when loading the params, we try to decode from bf16 to f32 regardless of the dtype of the param, since all the params are stored with "format=f32-to-bf16" in the record regardless of the dtype.

We solve it by checking the dtype as well, just like the C++ counter part:
https://github.com/apache/tvm/blob/4e66690a4d033af912f5051c0e5a16c9c10691d9/src/runtime/relax_vm/ndarray_cache_support.cc#L168-L172


junrushao (approved):
Thanks for the fix!

PR #16296, opened 2023-12-29
[TIR] Skip non-pure condition when simplify

wrongtest-intellif:
Hi~ The change wants to fix one simplifier bug on non-pure conditions.  If the condition expression do not hold the same value in its dominating scope, it is not safe to bind it in `arith::Analyzer` and rely on it in simplify.

As an example, the condition `A[i] > 1` is not reliable since it is updated with buffer write.
```python
for i in range(128):
    if A[i] > 1:
        A[i] = A[i] + 1
        # some code use A[i] to be simplified like
        B[i] = T.if_then_else(A[i] - 1 > 0, 0, 1)
```


wrongtest-intellif:
cc @Hzfengsy @MasterJH5574 @Lunderberg  
Hi~ Could you help to take a review of the fix? thank you!

src/arith/ir_mutator_with_analyzer.cc
```diff
@@ -20,13 +20,41 @@
...
+  CallEffectKind side_effect = SideEffect(condition);
+  if (side_effect > CallEffectKind::kReadState) {
+    return false;
+  } else if (side_effect == CallEffectKind::kReadState) {
+    // when the condition maybe stateful, enter into the slow path
+    // to visit the subtree one more time and ensure the state is not mutated.
+    std::vector<const VarNode*> read_vars;
+    UsesVar(condition, [&read_vars](const VarNode* v) {
+      read_vars.push_back(v);
+      return true;
```
Lunderberg, line 44: By returning `true`, the `VarTouchVisitor` will return early and won't visit the rest of the TIR tree.  As a result, any read variables after the first would be missing from `read_vars`.

src/tir/analysis/var_touch.cc
```diff
@@ -45,12 +47,16 @@ class VarTouchVisitor : public StmtExprVisitor {
   void VisitExpr_(const VarNode* op) final { Handle(op); }
 
   void VisitStmt_(const BufferStoreNode* op) final {
-    Handle(op->buffer->data.get());
+    if (touch_write_) {
```
Lunderberg, line 50: This looks like it's a repetition of the logic in `VarUseDefAnalyzer`, which already distinguishes between variable access, reads of a buffer, and writes to a buffer.  I'd recommend using it rather than duplicating similar logic here.

src/arith/ir_mutator_with_analyzer.cc
```diff
@@ -20,13 +20,41 @@
...
 #include "ir_mutator_with_analyzer.h"
 
 #include <tvm/arith/iter_affine_map.h>
 #include <tvm/tir/analysis.h>
 #include <tvm/tir/op.h>
 
 namespace tvm {
+namespace tir {
+
+bool IsPureCondition(const PrimExpr& condition, const Stmt& scope) {
```
Lunderberg, line 34: For any impure conditional, it looks like we'd end up visiting the body three times (`UsesVar`, `WritesVar`, and then the actual mutation).  For a nested conditional, the nested body would be visited several times.

Lunderberg (changes requested):
I like the find, though I'm confused as to why the fix is required.  The existing constraint tracking (e.g. [here](https://github.com/apache/tvm/blob/main/src/arith/rewrite_simplify.cc#L353) for the rewrite simplifier) already filters out constraints with side effects, so this case should already be checked for.

For the default simplifier, I think `IsPureCondition` should only check `return SideEffect(condition) <= CallEffectKind::kPure`, in order to avoid visiting the body of the conditional multiple times.  For cases where we do want data-dependent conditionals to be analyzed, we can use the [`propagate_knowns_to_prove_conditional`](https://github.com/apache/tvm/blob/main/src/tir/transforms/simplify.cc#L56) option to enable full dataflow analysis and constraint propagation.

Lunderberg:
I did some digging, and it looks like this is able to occur because `ConstIntBound` doesn't check for `SideEffect(expr) <= CallEffectKind(kPure)` [here](https://github.com/apache/tvm/blob/main/src/arith/const_int_bound.cc#L664), when entering the a constrained context.  

I think the general approach here is correct, that the checking for side effects should happen at the scope providing the constraint, as the `ConstIntBound` and `RewriteSimplifier` do not have sufficient context to determine whether a constraint holds for the duration of the scope.  At some point, we probably should remove the `SideEffect` check in `RewriteSimplifier` to match that of `ConstIntBound`.

PR #16303, opened 2023-12-29
[Unity][NN] Use Linear name for nn.op.permute_dims

Lunderberg:
The `relax::op::linear` is implemented as `permute_dims`, followed by `matmul`.  In this case, readability can be improved by naming the weights.  For example, if the weights provided to `nn.Linear` are named `linear_layer_1_weight`, then the weights with permuted dimensions can be named `matmul_layer_1_weight` instead of `permute_dims`.


junrushao (approved):
LGTM!

PR #16304, opened 2023-12-29
[Unity][Analysis] Handle PrimStructInfo in EraseToWellDefined

Lunderberg:
Prior to this commit, the `EraseToWellDefined` pass would update symbolic variable definitions in `ShapeStructInfo` and `TensorStructInfo`, but did not in `PrimStructInfo`.  This commit updates the `WellDefinedEraser` to include symbolic variables defined in `PrimStructInfo`.


slyubomirsky (approved):
Good addition, and the test cases are also nice to have. Thank you!

PR #16307, opened 2023-12-29
[Unity] Improved error message in ExprMutator::ReEmitBinding

Lunderberg:
Prior to this commit, an error message would occur in `ExprMutator::ReEmitBinding` if the struct info is missing from the generated value.  However, because this error was generated from inside `GetStructInfo`, it didn't include sufficient context for debugging.  This commit checks the struct info explicitly, and includes the context of the updated variable in the error message.


slyubomirsky (approved):
Good tweak.

PR #16308, opened 2023-12-29
[Unity] Improved error message for matmul shape mismatch

Lunderberg:
If a matrix multiplication cannot be performed due to incompatible shapes, the error message now specifies the arguments, the shape of each argument, and which dimension of the shape has a mismatch. Previously, this error message only provided the dimension of the mismatch.


slyubomirsky (approved):
I'm always happy to see improvements to error messages :+1:

PR #16309, opened 2023-12-29
[Unity][Transform] Use parameter name in BundleModelParams

Lunderberg:
Prior to this commit, the `BundleModelParams` would replace model parameters with `param_tuple[index]` within expressions.  These nested expressions would then be normalized, resulting in `gv = param_tuple[index]` or `lv = param_tuple[index]` variable definitions.  These auto-generated `gv` and `lv` names make it quite difficult to determine which model parameter is being used.

This commit updates the `BundleModelParams` transform to explicitly produce the bound variable, `orig_param_name = param_tuple[index]`, preserving human-readable names from the parameters.


slyubomirsky (approved):
Excellent quality-of-life improvement, thanks a ton.

PR #16310, opened 2023-12-29
[Unity][Analysis] Show objects instead of names in WellFormedChecker

Lunderberg:
Prior to this commit, several diagnostics in the `WellFormedChecker` would explicitly extract the name from `relax::Var`, `tir::Var`, and `GlobalVar` instances.  This is unnecessary, as these classes can be printed directly, and skips any changes to the default printing behavior (e.g. printing of variable addresses) that may be useful while debugging.


slyubomirsky (approved):
Good tweak.

PR #16311, opened 2023-12-29
[Unity] Validate struct info in relax::Call constructor

Lunderberg:
All operations called by a `relax::Call` node must have a `FuncStructInfo`.  Prior to this commit, an invalid struct info would be caught by the `BlockBuilder` during normalization.  This delay between the invalid `relax::Call` being constructed and the invalid `relax::Call` being detected makes debugging difficult. This commit adds an additional check during the `relax::Call` constructor, to provide earlier error detection.


slyubomirsky (approved):
Good improvement.

PR #16313, opened 2023-12-29
[Unity][Transform] Implement relax.transform.ExpandMatmulOfSum

Lunderberg:
An optimization pass that rewrites `x*(A+B)` into `x*A + x*B`, where `x`, `A`, and `B` are relax tensors.

### DIFF
--- a/python/tvm/relax/transform/__init__.py
+++ b/python/tvm/relax/transform/__init__.py
@@ -36,6 +36,7 @@ from .transform import (
     DecomposeOpsForInference,
     DecomposeOpsForTraining,
     EliminateCommonSubexpr,
+    ExpandMatmulOfSum,
     ExpandTupleArguments,
     FewShotTuning,
     FoldConstant,
--- a/python/tvm/relax/transform/transform.py
+++ b/python/tvm/relax/transform/transform.py
@@ -1282,6 +1282,26 @@ def AdjustMatmulOrder():
     return _ffi_api.AdjustMatmulOrder()  # type: ignore
 
 
+def ExpandMatmulOfSum():
+    """Expand `matmul(x, A+B)` to `matmul(x,A) + matmul(x,B)`
+
+    If either operand can be fully computed at compile-time (only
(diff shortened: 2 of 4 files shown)


Lunderberg:
This PR is currently a draft, as it depends on functionality introduced in https://github.com/apache/tvm/pull/16312.

slyubomirsky:
The implementation seems sound, but is this an advantageous rewrite most of the time?

Lunderberg:
Whether or not there's an advantage depends on the specific shapes of the arguments.  The benefit occurs if the sizes of `x*A` and `x*B` are small compared to the size of `A*B`.  In principle, this condition can't be proven in the general case, because it depends on a dynamic `batch_size` in `x`.  In practice, the batch size tends to be small compared to the dimensions of the weight matrices, so it is generally beneficial.

This transform is not intended to be enabled in a default pass, but as a tool to be used in building model-specific optimization pipelines.  Specifically, for batched LoRA evaluation, a computation similar to [S-LoRA](https://arxiv.org/abs/2311.03285) can be implemented with the following sequence:

1. Initial state: `x*(W + A*B)`
2. `ExpandMatmulOfSum`: `x*W + x*(A*B)`
3. `AdjustMatmulOrder`: `x*W + (x*A)*B`
4. `CombineParallelMatmul`: Compuite `x*W` and `x*A` simultaneously as `x*concat(W,A)`

slyubomirsky:
I assume it's similar to #16315 in that regard, then. Perhaps @tqchen and @yzh119 should weigh in as well. Probably wouldn't hurt to have the transformation available though.

Lunderberg:
Yup.  I'd see this transformation as a tool in a toolbox, to be used in optimization pipelines as needed, but not something to be part of a general flow without further testing.

slyubomirsky (approved):
I guess there's no harm in merging even if it's not being used as part of an optimization right now.

Lunderberg:
Rebased onto `main`, now that the main branch includes `unity`.  No other changes made, just wanting to avoid stale CI before merge.

PR #16314, opened 2023-12-29
[Unity][Transform] Implement relax.transform.AdjustMatmulOrder

Lunderberg:
Reorder `x*(A*B)` to `(x*A)*B`.  Intended for optimization of LoRA models, for which `(x*A)*B` has a much smaller memory footprint.


Lunderberg:
This PR is currently a draft, as it depends on functionality introduced in https://github.com/apache/tvm/pull/16312.

Lunderberg:
The prerequisite #16312 has now landed, and this PR is ready for review.

src/relax/transform/adjust_matmul_order.cc
```diff
@@ -0,0 +1,93 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/*!
+ * \file tvm/relax/transform/expand_matmul_of_sum.cc
+ * \brief Expand `matmul(x, A+B)` to `matmul(x, A) + matmul(x,B)`
+ */
```
Hzfengsy, line 23: typo :)

Hzfengsy (commented):
Thanks for the PR, which would help on the specific case. However, I think it's necessary to analyze the memory footprint before rewriting it. 

it would be good to rewrite `x * (A * B) -> (x * A) * B`, but not reasonable to rewrite `A * (B * x) -> (A * B) * x

src/relax/transform/adjust_matmul_order.cc
Lunderberg: Thank you, and fixed.

Lunderberg:
Good point.  I've updated the pattern and the rewrite rule with two changes:

1. Check if `(A*B)*C` should be re-ordered into `A*(B*C)`.  Previously, only the reverse was checked.
2. Only reorder the matmuls in either case when there's a provable benefit to doing so, based on the shapes of the three matrices.

I also added several additional unit tests to validate the behavior, for cases where the initial matmul is left-to-right, and where the initial matmul is right-to-left.

Lunderberg:
@Hzfengsy All changes based on your earlier comments have been made, and the PR is ready for re-review.

Hzfengsy (approved):
LGTM. Thanks for the great work!

PR #16316, opened 2023-12-30
[Unity][Frontend] Introducing Object

MasterJH5574:
This PR supports `Object` as a new spec/frontend type in nn.Module, so that non-tensor opaque objects (such as PagedKVCache) can be effectively represented.

One unit test for PagedKVCache is added.

### DIFF
--- a/python/tvm/relax/frontend/nn/core.py
+++ b/python/tvm/relax/frontend/nn/core.py
@@ -50,7 +50,12 @@ from tvm.target import Target
 
 from ... import expr as rx
 from ...block_builder import BlockBuilder
-from ...struct_info import ShapeStructInfo, TensorStructInfo, TupleStructInfo
+from ...struct_info import (
+    ObjectStructInfo,
+    ShapeStructInfo,
+    TensorStructInfo,
+    TupleStructInfo,
+)
 from ._tensor_op import _TensorOp
 from .subroutine import SubroutineMixin
 
@@ -272,6 +277,22 @@ class Parameter(Tensor):
             self._expr = Tensor.placeholder(  # pylint: disable=protected-access
                 self.shape, dtype=dtype, name="param"
             )._expr
+
+
+class Object:
+    """A wrapper on top of relax.Expr whose struct_info is the base
+    ObjectStructInfo (rather than any its subclass). Object effectively
+    represents non-tensor frontend components such as KV caches.
+    """
+
+    _expr: rx.Var
+
+    def __init__(self, *, _expr: rx.Expr, _name: str) -> None:
+        """Private constructor. Object is never supposed to be constructed directly by users."""
+        if not isinstance(_expr, rx.Var):
+            _expr = BlockBuilder.current().emit(_expr, _name)
+        self._expr = _expr
+        assert isinstance(self._expr.struct_info, ObjectStructInfo)
 
 
 class Effect:
--- a/python/tvm/relax/frontend/nn/__init__.py
+++ b/python/tvm/relax/frontend/nn/__init__.py
@@ -17,7 +17,7 @@
 """A PyTorch-like API to build IRModules."""
 # pylint: disable=redefined-builtin
 from . import op, spec
-from .core import Effect, Module, ModuleList, Parameter, Tensor
+from .core import Effect, Module, ModuleList, Object, Parameter, Tensor
 from .exporter import add_extern
 from .extern import ExternModule, ObjectModule, SourceModule
 from .modules import (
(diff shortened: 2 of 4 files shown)


python/tvm/relax/frontend/nn/core.py
```diff
@@ -274,6 +279,22 @@ def to(self, dtype: Optional[str] = None) -> None:  # pylint: disable=invalid-na
...
+    ObjectStructInfo (rather than any its subclass). Object effectively
+    represents non-tensor frontend components such as KV caches.
+    """
+
+    _expr: rx.Var
+
+    def __init__(self, *, _expr: rx.Expr, _name: str) -> None:
+        """Private constructor. Object is never supposed to be constructed directly by users."""
+        if not isinstance(_expr, rx.Var):
+            _expr = BlockBuilder.current().emit(_expr, _name)
```
junrushao, line 293: Curious - is there a case where we allow `_expr` to be non `relax.Var`?

python/tvm/relax/frontend/nn/core.py
```diff
@@ -274,6 +279,22 @@ def to(self, dtype: Optional[str] = None) -> None:  # pylint: disable=invalid-na
...
+        self._expr = _expr
+        assert type(self._expr.struct_info) is ObjectStructInfo
```
junrushao, line 295:
```suggestion
        assert isinstance(self._expr.struct_info, ObjectStructInfo)
```

python/tvm/relax/frontend/nn/modules.py
```diff
@@ -612,6 +621,154 @@ def append(self, new_element: Tensor) -> None:
         )
 
 
+class PagedKVCache(Object):
```
junrushao, line 624: We might want to think of temporarily leaving this KVCache object to mlc for fast development if its design is still in flux

python/tvm/relax/frontend/nn/core.py
MasterJH5574: Ah I allowed it to be non-Var so that it helps emit non-Var Exprs https://github.com/apache/tvm/blob/f671711e6c7e603a0ecc3621f7ea831beb3ade75/python/tvm/relax/frontend/nn/modules.py#L718-L723. After a second thought, I think it would be semantics-wise clearer to only accept Var here, and emit the non-Var Expr on caller side. How does this sound?

python/tvm/relax/frontend/nn/core.py
MasterJH5574: This is on purpose if we want to make sure the frontend `Object` can only represent “opaque objects” and never express other types (like a tensor).
https://github.com/apache/tvm/blob/f671711e6c7e603a0ecc3621f7ea831beb3ade75/python/tvm/relax/frontend/nn/core.py#L282-L285

Say if we pass in a Var with TensorStructInfo here, `isinstance(sinfo, ObjectStructInfo)` returns true, and `type(self._expr.struct_info) is ObjectStructInfo` returns false.

python/tvm/relax/frontend/nn/modules.py
MasterJH5574: Yah agree with you!

MasterJH5574: Do you have a good idea on where it should be placed under MLC? I do think it will be tweaked for a few times.

python/tvm/relax/frontend/nn/core.py
junrushao: got it! that's actually fine as it is, no need to change anything!

python/tvm/relax/frontend/nn/core.py
junrushao: I was not sure if I fully capture what you mean, but `TensorStructInfo` is not a super class of `ObjectStructInfo`, so I suppose `isinstance(sinfo, ObjectStructInfo)` returns `False` in this case?

python/tvm/relax/frontend/nn/modules.py
junrushao: How about `python/mlc_chat/model/utils/kv_cache.py`?

python/tvm/relax/frontend/nn/core.py
MasterJH5574: Ah yes my bad. I remember TensorStructInfo inherited ObjectStructInfo, but I was wrong obviously. Will update it.

python/tvm/relax/frontend/nn/modules.py
MasterJH5574: Sounds good, just updated. The unit test for PagedKVCache will be moved to MLC accordingly.

python/tvm/relax/frontend/nn/core.py
MasterJH5574: Updated :-)

junrushao (approved):
LGTM! Let's get it in!

PR #16324, opened 2024-01-01
[Relay] make "ToScalar" support directly obtaining "int64_t"

mawnja:
Because on Windows, "long double" is 64 bits instead of 128 bits like on Linux, to avoid overflow from "long double" to "int64_t"


mawnja (commented):
All checks passed

mawnja:
@masahi All checks passed, can you help me merge the changes in?

masahi (approved):

PR #16327, opened 2024-01-01
[Unity][nn.Module] Introduce operator `empty`

junrushao:
This PR introduces an operator `op.empty` in the `nn.Module` frontend. It helps us to create an uninitialized memory from the memory pool, which could be used as temporary scratchpad memory to handcrafted operators.


junrushao:
CC: @MasterJH5574 @yzh119

MasterJH5574 (approved):

PR #16331, opened 2024-01-02
[Unity] Alter op impl handling empty transform for output

rasagna-quic:
This PR handles a bug in Alter op impl.

When output does not need to be transform for an op in alter op impl, segmentation fault occurs due to nullptr exception.

This exception has been handled and the corresponding test case for the same has been updated.


quic-sanirudh (approved):
LGTM, Thanks

PR #16334, opened 2024-01-02
[Unity] allow thread sync under if branch

jinhongyii:
thread sync should be allowed to insert under `if` stmt as long as all threads go to the same branch


src/tir/transforms/thread_storage_sync.cc
```diff
@@ -113,7 +113,6 @@ class ThreadSyncPlanner : public StorageAccessVisitor {
         }
       }
       if (sync_before_stmt) {
-        ICHECK_EQ(condition_counter(), 0) << "Cannot insert syncs inside condition";
```
vinx13, line 116: we need to be careful this may result in a condition that threads can never reach the sync point if they diverge

PR #16335, opened 2024-01-02
[Unity] Set CMAKE_CUDA_ARCHITECTURES default to native

vinx13:
CI images should also be updated to install cmake 3.24

cc @tqchen @junrushao @Lunderberg

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.18)
+cmake_minimum_required(VERSION 3.24)
 project(tvm C CXX)
 
 # Utility functions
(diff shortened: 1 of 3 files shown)


Lunderberg:
I like the change, and this makes sense to me.

Lunderberg:
Since https://github.com/apache/tvm/pull/16302 passed CI, and is now merged in as a temporary fix, can this PR also revert the changes from it?

yongwww (approved):

Lunderberg (approved):
Thank you for the changes, and for handling the changes between TVM and the cutlass kernels!

jwfromm:
Late to the party but just wanted to say i love this change. Thanks Wuwei!

Lunderberg:
It looks like cmake's GPU detection isn't perfect (tested with cmake `3.26.4` and `3.27.7`), and this can cause the following error to occur when trying to determine the cuda architectures to use.

```
CMake Error in CMakeLists.txt:                                    
  CUDA_ARCHITECTURES is set to "native", but no GPU was detected. 
```

I don't have a good long-term solution for this issue, but a current workaround is to explicitly call `set(CMAKE_CUDA_ARCHITECTURES 75)` in your `config.cmake`.

PR #15599, opened 2023-08-21
[Hexagon] F2qi avgpool bug fix

rasagna-quic:
We see an accuracy issue when FQ2I is enabled for the avg_pool2d op in the deeplab_v3 QAT model. The error seems to be happening because the underlying graph has the output of the avg_pool mapped directly to the conv2d without a quantize-dequantize layer in between.

Graph of interest
```mermaid
graph TD;
W(dequantize: A_scale, A_zero_point) --> Conv2d;
A(dequantize: B_scale, B_zero_point) --> avg_pool2d;
avg_pool2d --> Conv2d;
Conv2d --> Output(quantize: C_scale, C_zero_point);
```

Quantization parameters after Fq2i pass:
qnn.avg_pool2d	(Bs, Bz, Cs, Cz)
qnn.conv2d	        (Cs, Cz, As, Az)
qnn.requantize	(Cs*As, 0, Cs, Cz)

Proposed fix for quantization parameters:
qnn.avg_pool2d	(Bs, Bz, Bs, Bz)
qnn.conv2d	        (Bs, Bz, As, Az)
qnn.requantize	(Cs*As, 0, Cs, Cz)

where quantized parameters As, Bs, Cs are scales and Az, Bz, Cz are zeropoints.

PR #16337, opened 2024-01-02
[Hexagon][UnitTest] Disable flaky quantization test

Lunderberg:
The `test_pass_fq2i_avg_pool2d.py::test_avgpool_conv2d` test is sensitive to rounding errors, and failed about a third of the time (42 / 100 tests).  This was first noticed as CI failures in unrelated PRs (e.g. https://ci.tlcpack.ai/blue/organizations/jenkins/tvm-hexagon/detail/PR-16184/6/tests). This commit marks the flaky portions of the test with `pytest.mark.xfail`, to avoid causing breaking CI for other PRs.

To minimize the extent of the disabled test cases, this commit breaks up each of the unit tests.  Where previously a single test performed both hardware/simulation tests and relay graph comparisons, these are now done in separate test functions.  The hardware/simulation tests use `tvm.testing.assert_allclose` and have a tolerance of `1e-02`, while the graph-comparison tests use `tvm.ir.structural_equal`, and require identical floating-point values.  Only the graph-comparison test is disabled here.

The other two test cases in `test_pass_fq2i_avg_pool2d.py` do not show this same sensitivity, with no failures seen in 100 executions.

### DIFF
--- a/tests/python/contrib/test_hexagon/test_pass_fq2i_avg_pool2d.py
+++ b/tests/python/contrib/test_hexagon/test_pass_fq2i_avg_pool2d.py
@@ -18,50 +18,19 @@
 """ Tests for avg_pool2d fake quantization to integer """
 
 import numpy as np
+import pytest
+
 import tvm
 import tvm.testing
 import tvm.topi.testing
 from tvm import relay
 from tvm.contrib.hexagon.session import Session
 from tvm.contrib.hexagon.pytest_plugin import HEXAGON_AOT_LLVM_TARGET
+
 from .infrastructure import quantize_np, build_module, run_module
 
 
-def compare_graphs(expr, ref_expr):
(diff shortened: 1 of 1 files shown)


Lunderberg:
@rasagna-quic Can you take a look at this test case?  It was introduced in https://github.com/apache/tvm/pull/15599, and is failing about 1/3 of the time.  This PR is a stopgap to avoid impacting other work, but it should have a better fix.

rasagna-quic:
@Lunderberg Thank you for these changes, these look good to me. I will work create a new PR to fix this issue.

> @rasagna-quic Can you take a look at this test case? It was introduced in #15599, and is failing about 1/3 of the time. This PR is a stopgap to avoid impacting other work, but it should have a better fix for the long-term.

junrushao (approved):
LGTM!

PR #16341, opened 2024-01-03
[CMake] Use ccache as CMAKE_CUDA_COMPILER_LAUNCHER

Lunderberg:
For older versions of `ccache` that do not support `nvcc`, this has no effect.  For newer versions of `ccache`, this enables caching the compiled `nvcc` outputs.  This was tested using `ccache` versions 3.7.7 (does not cache `nvcc` output) and 4.8.3 (caches `nvcc` output).

This is primarily intended to reduce compilation time of cutlass kernels, which can take several minutes after switching branches.

### DIFF
--- a/cmake/utils/CCache.cmake
+++ b/cmake/utils/CCache.cmake
@@ -16,12 +16,28 @@
 # under the License.
 
 if(USE_CCACHE) # True for AUTO, ON, /path/to/ccache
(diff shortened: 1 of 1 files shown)


csullivan (approved):
Many many thanks! Very happy to see the unnecessary lengthy nvcc spawned ptxas recompiles of cutlass kernels for different archs go away

PR #16343, opened 2024-01-03
[CMake][MSVC] Disable permissive mode for MSVC builds

Lunderberg:
The C++ standard requires two-phase name resolution for templates.  By default, MSVC uses a non-standard name resolution, in which all names are looked up when a template is instantiated.  This has caused MSVC-specific compilation errors ([example](https://github.com/apache/tvm/actions/runs/7400684492/job/20134841480?pr=16183)), which are quite difficult to debug.

This commit updates adds the `/permissive-` flag when building TVM with MSVC, disabling the non-standard name resolution.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -166,6 +166,11 @@ if(MSVC)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP")
   add_compile_options(/bigobj)
 
+  # Use standard-conforming two-phase name resolution for templates.
+  # This minimizes the differences between g++/clang builds on Linux,
+  # and MSVC builds on Windows.
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /permissive-")
+
   # MSVC already errors on undefined symbols, no additional flag needed.
   set(TVM_NO_UNDEFINED_SYMBOLS "")
 


junrushao:
I don't fully understand this patch, but seems that it solves a debuggability issue with MSVC. Is there any downsides for this change?

Lunderberg:
The only potential downside would be if we have any code that depends on the non-standard name resolution used by MSVC.  Most of the differences would result in compile-time failures, but there are some very weird edge cases that can occur.  (e.g. Suppose `void func(int)` is declared before the template, and `void func(bool)` is declared after the template but before the template instantiation.  Inside the template, a call to `func(true)` should resolve to `void func(int)`, because name resolution happens when the template is parsed.  However, MSVC would silently and erroneously resolve it to `void func(bool)`, because it does the name lookup at instantiation time.)

Even though two-phase name resolution has been the C++ standard as long as C++ has had a standard, MSVC still defaults to the non-standard name resolution for backwards compatibility with pre-C++98 code.  Their worry is that switching to standard name resolution would silently switch behavior that was developed and compiled only on MSVC.  Given that we primarily compile, test, and run on gcc/clang builds, I think this is very unlikely that we rely on MSVC's behavior, and any differences are most likely bugs in the MSVC build.

Lunderberg:
TL;DR: Very unlikely to introduce a bug, and any reliance on this behavior would have caused bugs in the gcc builds.

junrushao:
Got it. Yeah I do think it's very unlikely that people rely on MSVC-specific name resolution behavior. Thanks for explaining this to me!

junrushao (approved):
LGTM!

PR #16346, opened 2024-01-04
[Doc] Fix minor error in "Expressions in Relay"

gty111:
This PR fixes minor error in [Expressions in Relay](https://tvm.apache.org/docs/reference/langref/relay_expr.html#let-bindings)


quic-sanirudh (approved):
Thanks for the fix.

PR #16348, opened 2024-01-04
[Docker] Upgrade pip in i386 container

vinx13:
This is to fix installation error of ml_dtypes

cc @tqchen @junrushao


junrushao (approved):

leandron (approved):
Thanks for the fix.

PR #16355, opened 2024-01-05
[Unity] Infer struct info for relax.op.split on dynamic-sized index

Lunderberg:
Prior to this commit, the `relax.op.split` did not provide a known shape if the split was performed over a dynamic-size axis.  This commit updates the shape inference to provide correct shapes in this case.  A test case is also added for `CombineParallelMatmul` to show the intended usage of this feature, to ensure that the split outputs from a combined matmul still have correct shape information.

### DIFF
--- a/src/relax/ir/block_builder.cc
+++ b/src/relax/ir/block_builder.cc
@@ -170,6 +170,10 @@ class BlockBuilderImpl : public BlockBuilderNode {
         auto it = shape_var_map.find(shape_var);
         if (it == shape_var_map.end()) {
           shape_var_map.Set(shape_var, shape_expr);
+          // Expose the shape variable as non-negative, for purposes
+          // of shape inference.  In many cases, knowning that the
+          // shape variable is non-negative allows for simpler
+          // expressions for dynamic shapes.
           analyzer_.MarkGlobalNonNegValue(shape_var);
         } else {
           const PrimExpr& old_shape_expr = (*it).second;
--- a/src/relax/ir/block_builder.cc
+++ b/src/relax/ir/block_builder.cc
@@ -170,13 +170,13 @@ class BlockBuilderImpl : public BlockBuilderNode {
         auto it = shape_var_map.find(shape_var);
         if (it == shape_var_map.end()) {
           shape_var_map.Set(shape_var, shape_expr);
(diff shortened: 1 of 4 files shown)


tests/python/relax/test_op_manipulate.py
```diff
@@ -2176,6 +2176,110 @@ def test_split_indices_or_sections_int64():
...
+            R.Tensor([n - (n + 2) // 3 * 2, 16]),
+        ),
+    )
+
+    # Spliting a dynamic axis at specific indices is allowed.  The
+    # algebraic form here isn't the cleanest, primarily because the
+    # test case doesn't know that `n` is a shape variable.  When
+    # occurring in a relax function, `n` would be marked with
+    # `analyzer_.MarkGlobalNonNegValue`, which would make the shapes
+    # simplify to `[(2,16), (3,16), (n,16)]`.
```
slyubomirsky, line 2271: Is there any way to test that directly? Maybe the BlockBuilder can expose its analyzer for testing.

src/relax/ir/block_builder.cc
```diff
@@ -170,13 +170,13 @@ class BlockBuilderImpl : public BlockBuilderNode {
         auto it = shape_var_map.find(shape_var);
         if (it == shape_var_map.end()) {
           shape_var_map.Set(shape_var, shape_expr);
+          analyzer_.MarkGlobalNonNegValue(shape_var);
```
slyubomirsky, line 173: Might be worth commenting that indicating this results in much simpler symbolic shapes in many cases.

slyubomirsky (approved):
This is a clear improvement and the changes to the code are straightforward.

tests/python/relax/test_op_manipulate.py
slyubomirsky: Also note the typo in the first word :)

Lunderberg: Thank you, and I've updated to fix the typo, and to test the simplified expressions directly.  Rather than exposing the block builder's analyzer, this test case now calls `bb.begin_scope`.  That way, failures resulting from changes to the `BlockBuilderImpl::BeginScope` implementation would also be caught.

Issue #15771, opened 2023-09-17
[Bug] Tune fails on macOS


gmeeker:
tvmc tune appears to have broken between 0.12.0 and 0.13.0.

### Expected behavior

Produce an autotuner json file (which worked in 0.12.0).

### Actual behavior

```
[Task  1/25]  Current/Best:    0.00/   0.00 GFLOPS | Progress: (0/40) | 0.00 sException in thread Thread-1:
Traceback (most recent call last):
  File "/usr/local/Cellar/python@3.8/3.8.18/Frameworks/Python.framework/Versions/3.8/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/usr/local/Cellar/python@3.8/3.8.18/Frameworks/Python.framework/Versions/3.8/lib/python3.8/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/gmeeker/src/tvm_0.13.0/python/tvm/rpc/server.py", line 272, in _listen_loop
    _serving(conn, addr, opts, load_library)
  File "/Users/gmeeker/src/tvm_0.13.0/python/tvm/rpc/server.py", line 143, in _serving
    server_proc.start()
  File "/usr/local/Cellar/python@3.8/3.8.18/Frameworks/Python.framework/Versions/3.8/lib/python3.8/multiprocessing/process.py", line 121, in start
    self._popen = self._Popen(self)
  File "/usr/local/Cellar/python@3.8/3.8.18/Frameworks/Python.framework/Versions/3.8/lib/python3.8/multiprocessing/context.py", line 224, in _Popen
    return _default_context.get_context().Process._Popen(process_obj)
  File "/usr/local/Cellar/python@3.8/3.8.18/Frameworks/Python.framework/Versions/3.8/lib/python3.8/multiprocessing/context.py", line 284, in _Popen
    return Popen(process_obj)
  File "/usr/local/Cellar/python@3.8/3.8.18/Frameworks/Python.framework/Versions/3.8/lib/python3.8/multiprocessing/popen_spawn_posix.py", line 32, in __init__
    super().__init__(process_obj)
  File "/usr/local/Cellar/python@3.8/3.8.18/Frameworks/Python.framework/Versions/3.8/lib/python3.8/multiprocessing/popen_fork.py", line 19, in __init__
    self._launch(process_obj)
[...]

PR #16357, opened 2024-01-06
[RPC] Fix tuning on macOS and Windows (#15771)

gmeeker:
Fix regression in (#15187) when multiprocessing start method is not 'fork', which prevented tuning from working. This affects macOS and Windows. Also in python 3.14 the default start method will be 'spawn'.

### DIFF
--- a/python/tvm/rpc/server.py
+++ b/python/tvm/rpc/server.py
@@ -119,6 +119,15 @@ def _server_env(load_library, work_path=None):
     return temp
 
 
+def _serve_loop(sock, load_library, work_path=None):
+    """Server loop"""
+    sockfd = sock.fileno()
+    temp = _server_env(load_library, work_path)
+    _ffi_api.ServerLoop(sockfd)
+    if not work_path:
+        temp.remove()
+
+
 def _parse_server_opt(opts):
     # parse client options
     ret = {}
@@ -128,25 +137,21 @@ def _parse_server_opt(opts):
     return ret
 
 
-def _serving(sock, addr, opts, load_library):
+def _serving(conn, addr, opts, load_library):
     logger.info(f"connected from {addr}")
     work_path = utils.tempdir()
     old_cwd = os.getcwd()
     os.chdir(work_path.path)  # Avoiding file name conflict between sessions.
     logger.info(f"start serving at {work_path.path}")
 
-    def _serve_loop():
-        _server_env(load_library, work_path)
-        _ffi_api.ServerLoop(sock.fileno())
-
-    server_proc = multiprocessing.Process(target=_serve_loop)
+    server_proc = multiprocessing.Process(target=_serve_loop, args=(conn, load_library, work_path))
     server_proc.start()
     server_proc.join(opts.get("timeout", None))  # Wait until finish or timeout.
 
     if server_proc.is_alive():
         logger.info("timeout in RPC session, kill..")
         _ffi_api.ReturnException(
-            sock.fileno(),
+            conn.fileno(),
             f'RPCSessionTimeoutError: Your {opts["timeout"]}s session has expired, '
             f'try to increase the "session_timeout" value.',
         )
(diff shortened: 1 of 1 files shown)


junrushao:
CC @Johnson9009

python/tvm/rpc/server.py
```diff
@@ -119,6 +119,15 @@ def download_linked_module(file_name):
     return temp
 
 
+def _serve_loop(sock, load_library, work_path=None):
+    """Server loop"""
+    sockfd = sock.fileno()
+    temp = _server_env(load_library, work_path)
+    _ffi_api.ServerLoop(sockfd)
+    if not work_path:
+        temp.remove()
```
Johnson9009, line 128: curious, why need the temp? it seems this function only have one caller, and it will give the work_path.

python/tvm/rpc/server.py
```diff
@@ -119,6 +119,15 @@ def download_linked_module(file_name):
     return temp
 
 
+def _serve_loop(sock, load_library, work_path=None):
+    """Server loop"""
+    sockfd = sock.fileno()
+    temp = _server_env(load_library, work_path)
+    _ffi_api.ServerLoop(sockfd)
```
Johnson9009, line 126: can the sockfd be inline?

python/tvm/rpc/server.py
```diff
@@ -119,6 +119,15 @@ def download_linked_module(file_name):
     return temp
 
 
+def _serve_loop(sock, load_library, work_path=None):
```
Johnson9009, line 122: can we keep the parameter name consistent, like use " conn" all or use " sock" all? 
Is there requirements let work_path be a optional parameter?

python/tvm/rpc/server.py
```diff
@@ -119,6 +119,15 @@ def download_linked_module(file_name):
...
+    """Server loop"""
```
Johnson9009, line 123: del the meaningless comment because it  give no more information beyond the function name, or write a meaning one.

Johnson9009 (commented):
sorry for late review, threre are some questions.

python/tvm/rpc/server.py
gmeeker: The answer to most of your comments is that I reverted part of (https://github.com/apache/tvm/pull/15187) so that _server_loop can be pickled. The code here comes from the code before that commit, and I didn't simplify it. I assumed there was a reason for its style, but I'll take another pass at this.

python/tvm/rpc/server.py
gmeeker: The earlier code that I copied from used "sock" to refer to the socket and "conn" to refer to the resulting object from accept(). The current code uses "sock" everywhere, as in "socket object" but not the same object throughout. It sounds like you prefer that?
No, I'm not sure why work_path was optional.

python/tvm/rpc/server.py
Johnson9009: Thanks for your explanation, I think just move the closure out of the function _serving will fix the issue, so from my side, it is better to keep the code simple, so please keep it as same as possible with the current implementation if there isn't reasonable requirement, thanks.

python/tvm/rpc/server.py
Johnson9009: even though the earlier code, there is only 1 function _serve_loop, the type of its 1st parameter is socket, we can only use 1 name "sock" or "conn", there is 2 caller of the function _serve_loop, just as you said, 1 caller's argument is a original socket, and another caller's argument is a result object from accept(), actually the type of the result object from accept still is socket, so I use the name "sock", because it closer with the type.

Johnson9009 (changes requested):

gmeeker:
The changes from the original code are now minimal, just removing the closure.

Johnson9009 (approved):
Thanks, look good to me.

PR #16362, opened 2024-01-07
[Unity][Fix] Memory planning check value type of 'tir_var_upper_bound'

MasterJH5574:
This PR adds a sanity check to ensure that all `tir_var_upper_bound` attrs used by static memory planning has integer as value type. This check helps avoid mistakes of using wrong value types.

The check is needed since `func->GetAttr<Map<String, IntImm>>` does not apply type check.

### DIFF
--- a/src/relax/transform/static_plan_block_memory.cc
+++ b/src/relax/transform/static_plan_block_memory.cc
@@ -334,8 +334,25 @@ class StorageAllocatorInit : public StorageAllocatorBaseVisitor {
     // memory planning.
     // NOTE: we only apply the annotated upper bounds to the TIR variables that
     // appear in the **function signature**.
-    Map<String, IntImm> var_upper_bound_attr =
-        func->GetAttr<Map<String, IntImm>>("tir_var_upper_bound").value_or(Map<String, IntImm>());
+    Map<ObjectRef, ObjectRef> var_upper_bound_attr_raw =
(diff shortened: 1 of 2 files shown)


src/relax/transform/static_plan_block_memory.cc
```diff
@@ -336,6 +336,13 @@ class StorageAllocatorInit : public StorageAllocatorBaseVisitor {
     // appear in the **function signature**.
     Map<String, IntImm> var_upper_bound_attr =
         func->GetAttr<Map<String, IntImm>>("tir_var_upper_bound").value_or(Map<String, IntImm>());
+    // `GetAttr` does not perform type check.
```
junrushao, line 339: “IsInstance”-style type checking, if can be determined in compile time, will be folded to a constant without actually doing the checking in runtime. Therefore, to ensure runtime type checking happens, we will need to use a step-phase downcasting, where the first phase casts an attribute to Map<ObjectRef, ObjectRef>, and then check the types of each item

Hzfengsy:
Thanks for the fix. Beyond this commits, should we also check the `value >= 0`?

MasterJH5574:
Thanks @junrushao and @Hzfengsy for the suggestions! Just updated the PR. Please take another look at your convenience.

junrushao (approved):
LGTM!

PR #16370, opened 2024-01-09
Add OpenCL Support

BeiningWu:
i enhance  some lines to support opencl in detect_target


mshr-h:
I think we need to include tests in https://github.com/apache/tvm/blob/main/tests/python/target/test_target_target.py

junrushao:
This looks great! Please address the comments from Egor and fix the CI. Happy to get it in ASAP

PR #16376, opened 2024-01-09
[Unity][Contrib] Fix a bug due to typo in vllm `reconstruct_from_cache` kernel and add test

masahi:
This kernel is not originally coming from vllm, I wrote this for my need. Then I introduced this bug.
This is exactly an inverse of the `reshape_and_cache` kernel that vllm does have.

@vinx13


vinx13 (approved):

vinx13:
unity ci is broken after cuda upgrade, need to merge #16377

PR #16379, opened 2024-01-10
[Unity][TVMScript] Update `call_packed` semantics to support empty sinfo_args

Hzfengsy:
In low-level Relax (after the pass `CallTIRewrite`), the `call_packed` nodes do not always have explicit `sinfo_args`. This PR extends the parser to support this case.

### DIFF
--- a/python/tvm/script/ir_builder/relax/ir.py
+++ b/python/tvm/script/ir_builder/relax/ir.py
@@ -330,7 +330,7 @@ def output(*vars: Tuple[Var]) -> None:
 def call_packed(
     func: py_str,
     *args: Expr,
-    sinfo_args: Union[StructInfo, List[StructInfo]],
+    sinfo_args: Optional[Union[StructInfo, List[StructInfo]]] = None,
     **kwargs: Any,
 ) -> Call:
     """Create a relax Call, which calls a packed function.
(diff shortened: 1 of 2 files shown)


Hzfengsy:
cc @junrushao @cyx-6

slyubomirsky:
What is the resulting StructInfo if `sinfo_args` is empty?

Hzfengsy:
It should be the `R.Object`, which only happens in the low-level stage I think

Lunderberg:
> It should be the R.Object, which only happens in the low-level stage I think

At the low-level relax, does this correspond to a `void` return type?

Hzfengsy:
> At the low-level relax, does this correspond to a void return type?

You are right, it's void. But for now it is printed as an Object. Here is a real-world example:

```python
_5_1: R.Object = R.call_packed("vm.builtin.paged_attention_kv_cache_attention", paged_kv_cache, R.prim_value(0), alloc15, alloc16, alloc17, alloc18)
```

Lunderberg:
Ah, thank you.  I had interpreted not having "explicit `sinfo_args`" as having a null pointer for the `Array<StructInfo> sinfo_args`, and wanted to check if that would instead indicate a bug in `CallTIRRewrite`.  Instead, `CallTIRRewrite` produces a well-defined, but empty, `sinfo_args`.  The mismatch is in the TVMScript printer, which suppresses printing of empty `sinfo_args`, and the TVMScript parser, which currently requires `sinfo_args` even if empty.

Thank you for clearing up my confusion.

Lunderberg (approved):

PR #16383, opened 2024-01-10
[TOPI][Target] Add fp16 SIMD support for conv2d on `arm_cpu` targets

Anndrey24:
Optimised fp16 conv2d matrix tiling for Arm(R) Neon(TM) instructions and exposed `+fullfp16` as a target feature for Arm(R) Cortex(R) A-Profile CPUs.

Also, a target test was added to `cpptest` for Arm(R) Cortex(R) A-Profile CPUs which checks that the `has_fp16_simd` flag is set exclusively when the user explicitly passes the `+fullfp16` or `+sve` attributes and a supporting architecture version at target creation.

cc @ekalda @lhutton1 @neildhickey


jwfromm (approved):
LGTM, thanks for the change!

ekalda (approved):
Looks good, thanks @Anndrey24!

PR #16385, opened 2024-01-10
[Unity][Relax][Op] Add Conv3D Operator

jwfromm:
This PR adds full support for `relax.op.nn.conv3d`, the last missing convolution dimension in the relax operator set. I've added end to end correctness tests via the onnx frontend and structural testing but let me know if there are other tests that you'd like to see!


tqchen:
Thanks @jwfromm , can you also add to the PT frontend as well

jwfromm:
Yes will do :)

junrushao (approved):

PR #16387, opened 2024-01-11
[RUNTIME][RPC] Enable RPCObjectRef return in RPC

tqchen:
This PR enables RPCObjectRef return object similar to the disco transporation. This allows us to do advanced remote debugging when remote vm requires advanced object input like kv cache and shape.

To keep the implementation with minRPC(used in some of the limited protocols) forn now, we only support RPCObjectRef for now and do not enable unpacking Shape and String (they can still be constructed with two step constructor, per testcase).


tqchen:
cc @junrushao

Hzfengsy (approved):

Issue #16399, opened 2024-01-15
[Bug] relay pytorch frontend not support logical_or op


taomiao:
Thanks for participating in the TVM community! We use https://discuss.tvm.ai for any general usage questions and discussions. The issue tracker is used for actionable items such as feature proposals discussion, roadmaps, and bug tracking.  You are always welcomed to post on the forum first :smile_cat:

Issues that are inactive for a period of time may get closed. We adopt this policy so that we won't lose track of actionable issues that may fall at the bottom of the pile. Feel free to reopen a new one if you feel there is an additional problem that needs attention when an old one gets closed.

### Expected behavior

work well

### Actual behavior

```
self = <tvm.relay.frontend.pytorch.PyTorchOpConverter object at 0x7f090d88aa40>
op_names = {'aten::logical_or'}

    def report_missing_conversion(self, op_names):
        """Check if all ops in an input graph are supported by TVM"""
        known_ops = [
            "prim::Constant",
            "prim::GetAttr",
            "prim::ListConstruct",
            "prim::ListUnpack",
            "prim::TupleConstruct",
            "prim::TupleUnpack",
            "prim::RaiseException",
            "prim::If",
            "prim::Loop",
        ]
        known_ops += list(self.convert_map.keys())
        known_ops += list(qnn_torch.convert_map.keys())
    
        missing = []
    
        for op_name in op_names:
            # Also take care of in-place variant ops like aten::relu_
            if op_name not in known_ops and not (
                op_name.endswith("_") and op_name[:-1] in known_ops
            ):
                missing.append(op_name)
    
        if missing:
            msg = f"The following operators are not implemented: {missing}"
>           raise NotImplementedError(msg)
E           NotImplementedError: The following operators are not implemented: ['aten::logical_or']

../../../../python/tvm/relay/frontend/pytorch.py:4330: NotImplementedError
```

### Environment

os: ubuntu
python: 3.9
pytorch: 2.0
tvm: main branch

### Steps to reproduce

```
def test_logical_or():
    """test_logical_or"""

    def test_fn(x, y):
        return torch.logical_or(x, y)

    a = torch.tensor([0, 1, 10, 0], dtype=torch.int8)
    b = torch.tensor([4, 0, 1, 0], dtype=torch.int8)
    verify_model(test_fn, [a, b])

    a = torch.tensor([True, False, True])
    b = torch.tensor([True, False, False])
    verify_model(test_fn, [a, b])
```

### Triage

[...]

PR #16400, opened 2024-01-15
[Relay][Frontend][Torch] fix pytorch frontend not support logical or

taomiao:
add logical_or op to relay pytorch frontend. #16399

### DIFF
--- a/python/tvm/relay/frontend/pytorch.py
+++ b/python/tvm/relay/frontend/pytorch.py
@@ -2672,6 +2672,12 @@ class PyTorchOpConverter:
 
         return _op.logical_and(lhs, rhs)
 
+    def logical_or(self, inputs, input_types):
+        lhs = _op.cast(inputs[0], "bool")
+        rhs = _op.cast(inputs[1], "bool")
+
+        return _op.logical_or(lhs, rhs)
+
     def nonzero(self, inputs, input_types, is_numpy_style=False):
         data = inputs[0]
         ret = _op.transform.argwhere(data)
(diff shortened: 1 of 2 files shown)


taomiao:
@kevinthesun would you help check this logical_or is ok?

python/tvm/relay/frontend/pytorch.py
```diff
@@ -4237,7 +4243,9 @@ def create_convert_map(self):
             "torchvision::deform_conv2d": self.deform_conv2d,
             "aten::unbind": self.unbind,
             "aten::__and__": self.logical_and,
+            "aten::__or__": self.logical_or,
```
t-vi, line 4246: I'm afraid this is not tested and also not quite right.
While `torch.logical_or(torch.tensor(2), torch.tensor(1))` gives `tensor(True)`,
`torch.ops.aten.__or__(torch.tensor(2), torch.tensor(1))` gives `tensor(3)`
(and `torch.tensor(2) or torch.tensor(1))` gives `tensor(2)`, so it is different yet)

t-vi (changes requested):
Thank you for submitting the PR. The `logical_or` looks good, I have doubts about the `__or__` .

python/tvm/relay/frontend/pytorch.py
taomiao: Removed aten::__or__ from the convert_map.

t-vi: Thank you. Seems good to me now.

masahi (approved):

PR #16401, opened 2024-01-15
[microNPU][ETHOSU] Add fixed point for matmul

Aleksei-grovety:
Added support for calculation matmul with 16 bits fixed point. Added legalization of non-quantized dense operation with quantization by fixed point multiplication.

### DIFF
--- a/python/tvm/relay/op/contrib/ethosu.py
+++ b/python/tvm/relay/op/contrib/ethosu.py
@@ -24,6 +24,7 @@ import numpy as np  # type: ignore
 import tvm  # type: ignore
 from tvm import relay
 from tvm.ir import Op
+from tvm.relay.backend.contrib.ethosu import util
 from tvm.relay.build_module import bind_params_by_name  # type: ignore
 from tvm.relay.dataflow_pattern import (  # type: ignore
     is_constant,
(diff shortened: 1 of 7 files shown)


src/relay/backend/contrib/ethosu/compiler_attrs.cc
```diff
@@ -72,6 +73,13 @@ struct EthosUCompilerConfigNode : public tvm::AttrsNode<EthosUCompilerConfigNode
             "the linker script for section \".rodata.tvm\" that the constants are located in SRAM)")
         .set_default(Bool(false));
     String dev_warning = "Option is intended for development and debugging purposes only. ";
+    TVM_ATTR_FIELD(enable_fixed_point)
```
lhutton1, line 76: Curious why this needs to be passed as an option, could you imply fixed point when an int16 nn.dense operator is used?

python/tvm/relay/op/contrib/ethosu.py
```diff
@@ -1810,27 +1811,37 @@ def __init__(self, func_body):
         from tvm.relay.backend.contrib.ethosu.util import QDenseArgs  # type: ignore
         from tvm.relay.backend.contrib.ethosu.util import BiasAddArgs, RequantArgs
 
+        fract_scale = tvm.relay.Constant(tvm.nd.array(np.array(1 / 2**15)))
```
lhutton1, line 1814: Out of curiosity, where does `1/2**15` come from?

lhutton1 (commented):
Thanks for the PR @Aleksei-grovety, I had a couple of questions about the implementation

src/relay/backend/contrib/ethosu/compiler_attrs.cc
Aleksei-grovety: It could have been done without the option, but with the latest changes we need to specify the number of bits used to represent the fractional part of a fixed-point number.

python/tvm/relay/op/contrib/ethosu.py
Aleksei-grovety: It was a certain case where 15 bits were used to the fractional part, now this value is parameterized.

src/relay/backend/contrib/ethosu/compiler_attrs.cc
lhutton1: I see thanks! If we do need to expose an option, is it possible to combine these two options into a single one? I think it would be possible to imply fixed point calculation is disabled when `fixed_point_fraction_size` is set to 0 or -1

lhutton1 (commented):
Thanks for the responses @Aleksei-grovety. My main concern is that we need to expose an option to the user, rather than encapsulating this information as part of a relay operator. Are you able to share any more context about where this type of operation comes from?

src/relay/backend/contrib/ethosu/compiler_attrs.cc
Aleksei-grovety: Yes, it's possible. I suppose it's worth separating the FullyConnected and Matmul fixed-point implementations, and then only the fixed_point_fraction_size parameter will be needed.

Aleksei-grovety:
> Are you able to share any more context about where this type of operation comes from?

One of the earliest uses of fixed-point arithmetic was in digital signal processing (DSP) applications, where it was used to perform high-speed and high-precision calculations on audio and video signals. Fixed-point arithmetic was well-suited for these applications because it could provide the necessary precision and speed while minimizing the resources required to perform the calculations. 
Here is article [Introduction to Fixed Point Number Representation](https://inst.eecs.berkeley.edu/~cs61c/sp06/handout/fixedpt.html).

lhutton1 (commented):
Thanks for the cleanup @Aleksei-grovety. I think overall I'm still a bit unsure why `fixed_point_fraction_size` needs to be exposed from the microNPU compiler attrs, rather than the data being encoded in the Relay graph. For example, is it possible to insert `relay.fixed_point_multiply` operators on the inputs/outputs of the graph in between the matmul? Apologies if I'm still missing something here

Aleksei-grovety:
> For example, is it possible to insert `relay.fixed_point_multiply` operators on the inputs/outputs of the graph in between the matmul

Yes, it's possible, I've updated the code.

lhutton1 (approved):

lhutton1 (approved):

lhutton1:
Thanks @Aleksei-grovety!

PR #16403, opened 2024-01-15
[Bugfix][ROCm] Handle tvm_thread_invariant in ROCm backend

junrushao:
PR #16345 introduced a TIR intrinsic `tvm_thread_invariant` as a no-op intrinsic that indicates that a condition is an invariant across threads. It is a necessary change to support MoE group GEMM operations.

However, we noticed that the LLVM-ROCm backend has been broken since then because this no-op intrinsic hasn't been properly handled in this particular backend. This PR fixes this behavior by checking this no-op in the LLVM backend.

CC: @spectrometerHBH @yzhliu @tqchen


junrushao:
Superseded by #16404

PR #16406, opened 2024-01-16
[TIR] Fix of inter thread reduction with shared memory prefetch

LeiWang1999:
This is a fix of `LowerCrossThreadReduction`: The pass will remove all the loops with thread bind under the inter thread reduction block, which will introduce some issues when we meet the case where there could be other non-reduction blocks under the reduction thread.

Before removing a thread-bound loop, check if the block(s) under this loop has reduction block var. If the block(s) under have reduction do not have any reduction block var, it means that block is not reduction, and therefore this thread-bound loop should be kept. Otherwise, we remove the thread-bound loop as usual.

related discussion: https://discuss.tvm.apache.org/t/missing-thread-bind-loops-under-block-reduction-when-transformed-with-tir/16232/6

Please CC @MasterJH5574

### DIFF
--- a/src/tir/transforms/lower_cross_thread_reduction.cc
+++ b/src/tir/transforms/lower_cross_thread_reduction.cc
@@ -252,7 +252,7 @@ class InThreadReducerMaker : private StmtMutator {
       For res = opt_res.value();
       if (res->thread_binding.defined()) {
         UnderLoopReductionBlockVarCollector collector;
-        if (collector.CheckHasReductionBlocks(res)) {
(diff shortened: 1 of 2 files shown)


LeiWang1999:
@tvm-bot rerun

LeiWang1999:
> Thank you @LeiWang1999 for the fix! Could you also add a regression test case here in this file? https://github.com/apache/tvm/blob/main/tests/python/tir-transform/test_tir_transform_lower_cross_thread_reduction.py

for sure, please check out the update.

MasterJH5574 (approved):

LeiWang1999:
@tvm-bot rerun

PR #16411, opened 2024-01-16
[Unity][Transform] Handle symbolic variables in LambdaLift

Lunderberg:
Prior to this commit, symbolic variables used by a lambda function would be duplicated between the caller and the lifted-out function. In addition, shape inference within the lifted-out function was performed without access to the symbolic variables, resulting in unnecessary fallback from `R.Tensor([m, n])` to `R.Tensor(ndim=2)`.

This commit updates the `LambdaLift` transform to handle symbolic variables.  All symbolic variables have unique definitions across the resulting `IRModule`, and shape inference in the lifted-out function is aware of symbolic variables that have been exposed to it.

### DIFF
--- a/src/relax/transform/lambda_lift.cc
+++ b/src/relax/transform/lambda_lift.cc
@@ -240,7 +240,6 @@ class LambdaLifter : public ExprMutator {
     auto cache = current_lambda_var_;
     current_lambda_var_ = binding->var;
 
-    // ExprMutator::VisitBinding_(binding, func_node);
     auto new_value = VisitExpr(binding->value);
     if (!rebind_map_.count(binding->var)) {
       ReEmitBinding(binding, new_value);
@@ -304,9 +303,9 @@ class LambdaLifter : public ExprMutator {
 
     Expr body = func_node->body;
 
-    // recursive call
+    // Defining the rewrite rule prior to visiting the body, so that
+    // recursive closures can be updated.
     if (is_recursive && is_closure) {
-      // it is required by block_blocker, will be updated later
       nested_closure_map_.emplace(
           current_lambda_var_.value(),
           Call(gvar_lifted_func, captured_vars.Map([](Var var) -> Expr { return var; })));
@@ -371,8 +370,7 @@ class LambdaLifter : public ExprMutator {
 
       // Call "relax.invoke_closure" to invoke closure
 
-      if (bool is_closure = IsClosure(var);
-          is_closure && builder_->LookupBinding(var).as<CallNode>()) {
+      if (IsClosure(var) && builder_->LookupBinding(var).as<CallNode>()) {
         // if the original op was pure, we should use invoke_pure_closure
         Call orig_call = Downcast<Call>(builder_->LookupBinding(var));
         bool is_pure = [&]() -> bool {
@@ -424,13 +422,6 @@ class LambdaLifter : public ExprMutator {
   }
 
   bool IsClosure(Expr val) {
-    static int depth = -1;
-    struct Context {
-      explicit Context(int* ptr) : ptr(ptr) { (*ptr)++; }
-      ~Context() { (*ptr)--; }
-      int* ptr;
-    } context(&depth);
-
     if (auto opt_var = val.as<Var>()) {
       if (closures_.count(opt_var.value())) {
         return true;
--- a/src/relax/transform/lambda_lift.cc
+++ b/src/relax/transform/lambda_lift.cc
@@ -236,95 +236,25 @@ class LambdaLifter : public ExprMutator {
 
   using ExprMutator::VisitExpr_;
 
-  void VisitBinding_(const VarBindingNode* binding) final {
-    bool is_lambda = binding->value->IsInstance<FunctionNode>();
-    if (is_lambda) {
-      recur_vars_.push_back(binding->var);
+  void VisitBinding_(const VarBindingNode* binding, const FunctionNode* func_node) final {
+    auto cache = current_lambda_var_;
+    current_lambda_var_ = binding->var;
+
+    // ExprMutator::VisitBinding_(binding, func_node);
+    auto new_value = VisitExpr(binding->value);
+    if (!rebind_map_.count(binding->var)) {
+      ReEmitBinding(binding, new_value);
     }
 
-    Expr new_value = this->VisitExpr(binding->value);
-
-    if (new_value->struct_info_.defined() &&
-        !new_value->struct_info_.same_as(binding->var->struct_info_)) {
-      binding->var->struct_info_ = GetStructInfo(new_value);
-      binding->var->checked_type_ = new_value->checked_type_;
-    }
-    if (new_value.same_as(binding->value)) {
-      builder_->EmitNormalized(GetRef<VarBinding>(binding));
-    } else {
-      builder_->EmitNormalized(VarBinding(binding->var, new_value));
-    }
-    if (is_lambda) {
-      recur_vars_.pop_back();
-    }
-  }
-
-  Expr VisitExpr_(const CallNode* call_node) final {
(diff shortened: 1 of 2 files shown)


src/relax/transform/lambda_lift.cc
```diff
@@ -336,176 +266,235 @@ class LambdaLifter : public ExprMutator {
...
-      visited_func = GetRef<Expr>(func_node);
-    } else if (const auto& body_sinfo = MatchStructInfo<ObjectStructInfo>(body)) {
-      visited_func =
-          Function(params, body, body_sinfo.value(), func_node->is_pure, func_node->attrs);
-    } else {
-      visited_func =
-          Function(params, body, func_node->ret_struct_info, func_node->is_pure, func_node->attrs);
+    // recursive call
+    if (is_recursive && is_closure) {
+      // it is required by block_blocker, will be updated later
```
slyubomirsky, line 309: I assume you mean the BlockBuilder?

src/relax/transform/lambda_lift.cc
```diff
@@ -336,176 +266,235 @@ class LambdaLifter : public ExprMutator {
...
+    return ExprMutator::VisitExpr_(op);
+  }
+
+  bool IsClosure(Expr val) {
+    static int depth = -1;
+    struct Context {
+      explicit Context(int* ptr) : ptr(ptr) { (*ptr)++; }
+      ~Context() { (*ptr)--; }
+      int* ptr;
+    } context(&depth);
```
slyubomirsky, line 432: This doesn't appear to be used anywhere.

src/relax/transform/lambda_lift.cc
```diff
@@ -336,176 +266,235 @@ class LambdaLifter : public ExprMutator {
...
+
+    auto orig_sinfo = Downcast<StructInfo>(call->struct_info_);
+
+    if (auto opt_var = call->op.as<Var>()) {
+      auto var = opt_var.value();
+
+      // Call "relax.invoke_closure" to invoke closure
+
+      if (bool is_closure = IsClosure(var);
+          is_closure && builder_->LookupBinding(var).as<CallNode>()) {
```
slyubomirsky, line 375: Is there a reason to write it this way instead of `IsClosure(var) && ...`?

src/relax/transform/lambda_lift.cc
```diff
@@ -336,176 +266,235 @@ class LambdaLifter : public ExprMutator {
...
-          Function(params, body, body_sinfo.value(), func_node->is_pure, func_node->attrs);
-    } else {
-      visited_func =
-          Function(params, body, func_node->ret_struct_info, func_node->is_pure, func_node->attrs);
+    // recursive call
+    if (is_recursive && is_closure) {
+      // it is required by block_blocker, will be updated later
+      nested_closure_map_.emplace(
+          current_lambda_var_.value(),
+          Call(gvar_lifted_func, captured_vars.Map([](Var var) -> Expr { return var; })));
```
slyubomirsky, line 312: I assume the map is so that an `Array<Var>` is treated as `Array<Expr>`, correct?

src/relax/transform/lambda_lift.cc
```diff
@@ -336,176 +266,235 @@ class LambdaLifter : public ExprMutator {
...
-        func = Function(func->params, VisitExpr(func->body), func->ret_struct_info, func->is_pure,
-                        func->attrs);
-        builder_->UpdateFunction(pair.first, func);
+    for (auto [gvar, base_func] : glob_funcs) {
+      if (auto opt = base_func.as<Function>()) {
+        // Must visit the function itself, and not just the function
+        // body, to ensure that EraseToWellDefined recognized symbolic
+        // variables that are exposed by the function signature.
+        auto func = Downcast<Function>(VisitExpr(opt.value()));
+        builder_->UpdateFunction(gvar, func);
```
slyubomirsky, line 487: Is this the only change needed to handle symbolic vars, fundamentally? I ask because I didn't see any logic specific to symbolic vars above.

slyubomirsky (commented):
These changes definitely seem to be an improvement in terms of code organization, but I have a request for clarification (see comments) as to how this change achieves the goal of handling symbolic vars.

src/relax/transform/lambda_lift.cc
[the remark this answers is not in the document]
Lunderberg: Thank you for the catch.  The typo was present in the original, and the comment is also out of date after this change.  I've removed it, and added an appropriate comment.

src/relax/transform/lambda_lift.cc
```diff
@@ -236,95 +236,25 @@ class LambdaLifter : public ExprMutator {
...
 
-  void VisitBinding_(const VarBindingNode* binding) final {
-    bool is_lambda = binding->value->IsInstance<FunctionNode>();
-    if (is_lambda) {
-      recur_vars_.push_back(binding->var);
+  void VisitBinding_(const VarBindingNode* binding, const FunctionNode* func_node) final {
+    auto cache = current_lambda_var_;
+    current_lambda_var_ = binding->var;
+
+    // ExprMutator::VisitBinding_(binding, func_node);
```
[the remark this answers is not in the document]
Lunderberg: Thank you for the catch, and the comment was some test code, and is now removed.

src/relax/transform/lambda_lift.cc
```diff
@@ -336,176 +266,235 @@ class LambdaLifter : public ExprMutator {
...
   IRModule Lift() {
     auto glob_funcs = mod_->functions;
-    for (auto pair : glob_funcs) {
-      if (auto* n = pair.second.as<FunctionNode>()) {
-        auto func = GetRef<Function>(n);
-        func = Function(func->params, VisitExpr(func->body), func->ret_struct_info, func->is_pure,
-                        func->attrs);
-        builder_->UpdateFunction(pair.first, func);
+    for (auto [gvar, base_func] : glob_funcs) {
+      if (auto opt = base_func.as<Function>()) {
```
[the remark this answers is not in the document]
Lunderberg: Thank you!  This was was added back in https://github.com/apache/tvm/pull/14522, and I really like the way it avoids the repetition in `if(obj->IsInstance<TNode>())` checks and `Downcast<T>(obj)` casts inside the conditional.

src/relax/transform/lambda_lift.cc
Lunderberg: Whoops, debug code for indenting prints during nested calls.  Removed.

src/relax/transform/lambda_lift.cc
Lunderberg: Nope.  A previous implementation of `IsClosure` returned an `Optional<_>`, and this stuck around.  Updated.

src/relax/transform/lambda_lift.cc
Lunderberg: Yup, that's correct.  I'm wondering if there should be an implicit conversion from `Array<DerivedClass>` to `Array<BaseClass>`, to avoid needing this type of conversion.

src/relax/transform/lambda_lift.cc
Lunderberg: Primarily, yes.  Most of the other changes are to allow this specific change to work.  Previously, the mutator could assume that every `FunctionNode` encountered would be a lambda to be lifted out, but now it could be a top-level function as well.

src/relax/transform/lambda_lift.cc
slyubomirsky: Might have to be careful to ensure type safety can't get broken that way, that has tended to lead to soundness issues in languages. (I believe that type of thing is a source of unsoundness in TypeScript, IIRC.)

slyubomirsky (approved):
My comments were addressed.

PR #16412, opened 2024-01-16
[Lint] Add check to prevent usage of #include <regex>

Lunderberg:
Currently, the pytorch wheels available through `pip install` use the pre-C++11 ABI by setting `-DUSE_CXX11_ABI=0` [0].  If TVM were to user the pre-C++11 ABI, this would cause breakages with dynamically-linked LLVM environments.

This commit adds a lint check to search for use of `#include <regex>` in any C++ files.  Use of this header should be avoided, as its implementation is not supported by gcc's dual ABI.  This ABI incompatibility results in runtime errors either when `std::regex` is called from TVM, or when `std::regex` is called from pytorch, depending on which library was loaded first.

This restriction can be removed when a version of pytorch compiled using `-DUSE_CXX11_ABI=1` is available from PyPI.

[0] https://github.com/pytorch/pytorch/issues/51039

### DIFF
--- a/python/tvm/runtime/__init__.py
+++ b/python/tvm/runtime/__init__.py
@@ -41,3 +41,5 @@ from .params import (
 
 from . import executor
 from . import disco
+
+from .support import _regex_match
--- a/python/tvm/runtime/support.py
+++ b/python/tvm/runtime/support.py
@@ -1,0 +1,69 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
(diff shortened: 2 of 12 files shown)


Lunderberg:
The PR https://github.com/apache/tvm/pull/16465 exposes a PackedFunc wrapper for Python's regex matching.  The limited functionality it has, checking for a match without returning any extracted results, is sufficient for all existing uses of `#include <regex>` found by the new lint rule.  After it lands, this PR can be updated to use the new wrapper function instead of `std::regex`.

Lunderberg:

Lunderberg:
Rebased this PR onto main, in order to update the stale CI results.

tqchen (approved):

PR #16413, opened 2024-01-17
[Unity][Contrib] Workspace for cuBLAS backend

MasterJH5574:
This PR adds a 32MB workspace for cuBLAS backend, so that functions like `cublasLtMatmul` can take the workspace as input.

The workspace is managed under CuBlasThreadEntry so that it will be allocated only once in each thread.


MasterJH5574:
cc @vinx13 @masahi @junrushao @tqchen

src/runtime/contrib/cublas/cublas_utils.h
```diff
@@ -80,7 +80,13 @@ struct CuBlasThreadEntry {
 struct CuBlasLtThreadEntry {
   CuBlasLtThreadEntry();
   ~CuBlasLtThreadEntry();
+
   cublasLtHandle_t handle{nullptr};
+  cublasLtMatmulPreference_t matmul_pref_desc{nullptr};
+  void* workspace_ptr{nullptr};
+  // 32MB workspace.
+  constexpr const static size_t workspace_size = 33554432;
```
masahi, line 88: What's special about this particular size?

vinx13: it's recommended value for Hopper. for previous archs, it should be 4MB

masahi: Where is that documented?

vinx13: https://docs.nvidia.com/cuda/cublas/index.html#cublassetworkspace

MasterJH5574: Thanks! Just updated with the source.

MasterJH5574:
> There is a pass to allocate such workspace and append it to arguments of a BYOC function: https://github.com/apache/tvm/blob/unity/python/tvm/relax/transform/transform.py#L1305-L1317
> 
> If this can be used, I think that would be preferred.

@masahi Thank you for the great suggestion! Yes I think this one can be a next step, so that the workspace size can be adjustable. For now, given I am not particular familiar with the BYOC flow, I may not be able to quickly enable the workspace to work with the AllocateWorksapce pass. Adding a fixed size workspace in thread entry is the fastest way to enable workspace. I agree that we can leverage the pass later on.

junrushao:
@masahi would you mind sharing more guidance to @MasterJH5574?

src/runtime/contrib/cublas/cublas_utils.h
```diff
@@ -80,7 +80,14 @@ struct CuBlasThreadEntry {
 struct CuBlasLtThreadEntry {
   CuBlasLtThreadEntry();
   ~CuBlasLtThreadEntry();
+
   cublasLtHandle_t handle{nullptr};
+  cublasLtMatmulPreference_t matmul_pref_desc{nullptr};
+  void* workspace_ptr{nullptr};
+  // 32MB workspace as suggested by NVIDIA
+  // https://docs.nvidia.com/cuda/cublas/index.html#cublassetworkspace.
+  static constexpr const size_t workspace_size = 33554432;
```
masahi, line 89: I'm assuming that 32MB is also good for pre-Hopper since it is bigger than the recommended size, 4MB. @vinx13

masahi (approved):
Ok, given the very specific nature of this workspace, I think a fast path like this is reasonable. The `AllocateWorkspace`-based approach can be used in more general settings but it is overkill when we only need one workspace of a fixed size.

PR #16415, opened 2024-01-17
[RUNTIME][Memory] Implement support for non-zero offset within a storage object in AllocNDArr…

arangasa:
…ay for Hexagon

This is needed for supporting a new static memory planner which allocates a single storage object, and uses non-overlapping offset ranges within the storage object to allocate tensors.

Allocating a storage object is an expensive operation involving OS, whereas tensor allocation (using AllocNDArray) is much less expensive.


arangasa:
@tqchen @srkreddy1238 @yelite @yongwww

tqchen (approved):

PR #16417, opened 2024-01-17
fix pytorch frontend linspace op

taomiao:
torch.linspace(0, 8, 10) output 10 elements array, but tvm output 11 elements array.
```
inline Tensor arange(const PrimExpr& start, const PrimExpr& stop, const PrimExpr& step,
                     DataType dtype, std::string name = "T_arange", std::string tag = kInjective) {
  PrimExpr num_elem = tvm::cast(
      tvm::DataType::Int(32), tvm::ceil(tvm::cast(tvm::DataType::Float(32), stop - start) / step));
  Array<PrimExpr> shape;
  return compute(
      {num_elem},
      [&](const Array<Var>& indices) { return tvm::cast(dtype, start + step * indices[0]); }, name,
      tag);
}
```
float number precision deviation leads to this error.


masahi:
can you fix the lint issue?

taomiao:
> can you fix the lint issue?

ok

masahi (approved):

PR #16421, opened 2024-01-17
[Unity][Transform] Raise error in FuseOpsByPattern for SSA violation

Lunderberg:
Internally, `FuseOpsByPattern` makes a mapping from relax variables to the fused group containing that variable.  If the input module violates SSA, this map may be ill-formed.  While not strictly necessary for FuseOps to handle ill-formed inputs, checking it at this level provides better error handling than propagating it to downstream passes.

This commit checks for ill-formed inputs that would produce invalid fused outputs and raises an error.


Lunderberg:
Rebased onto main to resolve conflicts

slyubomirsky (approved):
It's always good to have better error-checking, especially when the messages can be more meaningful. Thanks!

PR #15761, opened 2023-09-15
[Target] LLVM helper functions for any target info

cbalint13:
Hi folks,

This PR adds helper functions to query any LLVM information about: targets, arches and it's features.
It can enhance TVM target handling with information from LLVM module, just like partially was done in: #15685

---
#### Changes:

 * Introduces ```llvm_get_targets()```, ```llvm_get_cpu_archlist()```, ```llvm_get_cpu_features()```& ```llvm_cpu_has_features()```
 * Extends  current ```target_has_feature()``` to work with *any* target, now also having ```-mtriple```/```-mattr``` awareness
 * Exposes booth C/FFI & PY interfaces for all the functions mentioned above.
 * Adds new unit tests for the function logic and API consistency stress.
 * Ensures proper API compatibility with older LLVM, tested with LLVM={10,14,15,16,17}.

---

#### Usage example

* List all LLVM **targets**:
````
print( tvm.target.codegen.llvm_get_targets() )

["arm", "armeb", "aarch64", "aarch64_be", "aarch64_32", "avr", "bpfel", "bpfeb", "hexagon", 
"loongarch32", "loongarch64", "mips", "mipsel", "mips64", "mips64el", "msp430", "powerpc", 
"powerpcle", "powerpc64", "powerpc64le", "r600", "amdgcn", "riscv32", "riscv64", "sparc", 
"sparcv9", "sparcel", "s390x", "thumb", "thumbeb", "i386", "x86_64", "xcore", "nvptx", 
"nvptx64", "lanai", "wasm32", "wasm64"]
````

* List all **cpu architectures** for a specific **target**:
```
print( tvm.target.codegen.llvm_get_cpu_archlist(tvm.target.Target("llvm -mtriple=hexagon--")))

["generic", "hexagonv5", "hexagonv55", "hexagonv60", "hexagonv62", "hexagonv65", 
"hexagonv66", "hexagonv67", "hexagonv67t", "hexagonv68", "hexagonv69", "hexagonv71", 
"hexagonv71t", "hexagonv73"]

print( tvm.target.codegen.llvm_get_cpu_archlist(tvm.target.Target("llvm -mtriple=riscv64-linux-gnu")))

["generic", "generic-rv32", "generic-rv64", "rocket", "rocket-rv32", "rocket-rv64", 
"sifive-7-series",  "sifive-e20", "sifive-e21", "sifive-e24", "sifive-e31", "sifive-e34", 
[...]

PR #1571, opened 2018-08-08
Vulkan TVM Android Support

dayanandasiet:
1. Vulkan Tvm Context added
2. Vulkan test case for rpc added

Thanks for contributing to TVM!   Please refer to guideline https://docs.tvm.ai/contribute/ for useful information and tips. After the pull request is submitted, please request code reviews from others in the community.

PR #16425, opened 2024-01-18
[Target] Use LLVM target parser for determining Arm(R) A-Profile Architecture features

lhutton1:
Currently, target features are determined by a set of fixed checks on the target string. This works well for checking support of a small number of simple features, but it doesn't scale. Some problems include:
- There are many non-trivial conditions for which a feature may(not) be available. It is easy to miss these with the current implementation.
- The inclusion of some features in a target string can imply other features. For example, "+sve2" implies "+sve". This currently isn't taken into account.
- The tests in tests/cpp/target/parsers/aprofile_test.c suggest that targets such as "llvm -mcpu=cortex-a+neon" and "llvm -mattr=+noneon" are supported target strings. The features will be correctly parsed in TVM, however, they are not valid in LLVM. Therefore, it's possible that TVM and LLVM have different understanding of the features available.

This commit uses the more robust LLVM target parser to determine support for the features in TVM. It leverages previous infrastructure added to TVM for obtaining a list of all supported features given an input target, and uses this to check the existance of certain features we're interested in. It should be trivial to grow this list over time. As a result of this change, the problems mentioned above are solved.

In the current form, this commit drops support for target strings such as "llvm -mcpu=cortex-a+neon" and "llvm -mattr=+noneon". A scan of the codebase suggests this functionality is not in use (only in test cases). Should we feel the need to support them, or have a smoother migration for downstream users of TVM we can add a translator to the parser to convert these into LLVM compatible targets.

cc @Mousius @cbalint13 @ekalda @neildhickey

### DIFF
--- a/src/target/llvm/llvm_instance.cc
+++ b/src/target/llvm/llvm_instance.cc
@@ -262,7 +262,7 @@ LLVMTargetInfo::LLVMTargetInfo(LLVMInstance& instance, const TargetJSON& target)
   }
 
   // LLVM JIT engine options
-  if (const Optional<String>& v = target->GetAttr<String>("jit")) {
+  if (const auto& v = Downcast<Optional<String>>(target.Get("jit"))) {
     String value = v.value();
     if ((value == "mcjit") || (value == "orcjit")) {
       jit_engine_ = value;
--- a/src/target/llvm/llvm_instance.cc
+++ b/src/target/llvm/llvm_instance.cc
@@ -199,21 +199,23 @@ std::ostream& operator<<(std::ostream& os, const LLVMTargetInfo::Option& opt) {
   return os;
 }
 
-LLVMTargetInfo::LLVMTargetInfo(LLVMInstance& instance, const Target& target) {
-  triple_ = target->GetAttr<String>("mtriple").value_or("default");
+LLVMTargetInfo::LLVMTargetInfo(LLVMInstance& instance, const Target& target)
+    : LLVMTargetInfo(instance, target->Export()) {}
 
+LLVMTargetInfo::LLVMTargetInfo(LLVMInstance& instance, const TargetJSON& target) {
+  triple_ = Downcast<String>(target.Get("mtriple").value_or(String("default")));
   if (triple_.empty() || triple_ == "default") {
     triple_ = llvm::sys::getDefaultTargetTriple();
   }
-  cpu_ = target->GetAttr<String>("mcpu").value_or(defaults::cpu);
+  cpu_ = Downcast<String>(target.Get("mcpu").value_or(String(defaults::cpu)));
 
-  if (const Optional<Array<String>>& v = target->GetAttr<Array<String>>("mattr")) {
+  if (const auto& v = Downcast<Optional<Array<String>>>(target.Get("mattr"))) {
     for (const String& s : v.value()) {
       attrs_.push_back(s);
     }
   }
   // llvm module target
-  if (target->kind->name == "llvm") {
+  if (Downcast<String>(target.Get("kind")) == "llvm") {
     // legalize -mcpu with the target -mtriple
     auto arches = GetAllLLVMTargetArches();
     bool has_arch =
@@ -224,7 +226,7 @@ LLVMTargetInfo::LLVMTargetInfo(LLVMInstance& instance, const Target& target) {
     }
   }
 
-  if (const Optional<Array<String>>& v = target->GetAttr<Array<String>>("cl-opt")) {
+  if (const auto& v = Downcast<Optional<Array<String>>>(target.Get("cl-opt"))) {
     llvm::StringMap<llvm::cl::Option*>& options = llvm::cl::getRegisteredOptions();
     bool parse_error = false;
     for (const String& s : v.value()) {
@@ -245,7 +247,7 @@ LLVMTargetInfo::LLVMTargetInfo(LLVMInstance& instance, const Target& target) {
   }
 
   llvm::FloatABI::ABIType float_abi = llvm::FloatABI::Default;
-  if (const Optional<String>& v = target->GetAttr<String>("mfloat-abi")) {
+  if (const auto& v = Downcast<Optional<String>>(target.Get("mfloat-abi"))) {
     String value = v.value();
     if (value == "hard") {
       float_abi = llvm::FloatABI::Hard;
@@ -283,14 +285,14 @@ LLVMTargetInfo::LLVMTargetInfo(LLVMInstance& instance, const Target& target) {
   target_options_.NoInfsFPMath = false;
   target_options_.NoNaNsFPMath = true;
   target_options_.FloatABIType = float_abi;
-  if (const Optional<String>& v = target->GetAttr<String>("mabi")) {
-    target_options_.MCOptions.ABIName = v.value();
+  if (target.find("mabi") != target.end()) {
+    target_options_.MCOptions.ABIName = Downcast<String>(target.Get("mabi"));
   }
 
-  auto maybe_level = target->GetAttr<Integer>("opt-level");
+  auto maybe_level = Downcast<Integer>(target.Get("opt-level"));
 #if TVM_LLVM_VERSION <= 170
   if (maybe_level.defined()) {
-    int level = maybe_level.value()->value;
+    int level = maybe_level->value;
     if (level <= 0) {
       opt_level_ = llvm::CodeGenOpt::None;
     } else if (level == 1) {
@@ -327,7 +329,7 @@ LLVMTargetInfo::LLVMTargetInfo(LLVMInstance& instance, const Target& target) {
   // Fast math options
 
   auto GetBoolFlag = [&target](llvm::StringRef flag) -> bool {
-    return target->GetAttr<Bool>(flag.str()).value_or(Bool(false));
+    return Downcast<Bool>(target.Get(flag.str()).value_or(Bool(false)));
   };
   if (GetBoolFlag("fast-math")) {
 #if TVM_LLVM_VERSION >= 60
@@ -850,7 +852,7 @@ const Array<String> LLVMTargetInfo::GetAllLLVMTargetArches() const {
   return cpu_arches;
 }
 
-const Array<String> LLVMTargetInfo::GetAllLLVMCpuFeatures() const {
+const Map<String, String> LLVMTargetInfo::GetAllLLVMCpuFeatures() const {
   std::string feats = "";
   for (const auto& attr : attrs_) {
     feats += feats.empty() ? attr : ("," + attr);
@@ -864,10 +866,11 @@ const Array<String> LLVMTargetInfo::GetAllLLVMCpuFeatures() const {
 #else
       MCInfo->getAllProcessorFeatures();
 #endif
-  Array<String> cpu_features;
+  // TVM doesn't have an FFI friendly Set, so use a Map instead for now
+  Map<String, String> cpu_features;
   for (const auto& feat : llvm_features) {
     if (MCInfo->checkFeatures("+" + std::string(feat.Key))) {
-      cpu_features.push_back(feat.Key);
+      cpu_features.Set(feat.Key, "");
     }
   }
 
@@ -877,9 +880,7 @@ const Array<String> LLVMTargetInfo::GetAllLLVMCpuFeatures() const {
 const bool LLVMTargetInfo::TargetHasCPUFeature(const std::string& feature) const {
   // lookup features for `-mcpu`
   auto feats = GetAllLLVMCpuFeatures();
-  bool has_feature =
-      std::any_of(feats.begin(), feats.end(), [&](const auto& var) { return var == feature; });
-
+  bool has_feature = feats.find(feature) != feats.end();
   return has_feature;
 }
 
--- a/src/target/llvm/llvm_instance.cc
+++ b/src/target/llvm/llvm_instance.cc
@@ -225,7 +225,7 @@ LLVMTargetInfo::LLVMTargetInfo(LLVMInstance& instance, const TargetJSON& target)
       // give the code a chance to run with a less-specific target.
       LOG(ERROR) << "LLVM cpu architecture `-mcpu=" << cpu_
                  << "` is not valid in `-mtriple=" << triple_ << "`"
-                 << ", cpu architecture ignored";
+                 << ", using default `-mcpu=" << String(defaults::cpu) << "`";
     }
   }
 
--- a/src/target/llvm/llvm_instance.cc
+++ b/src/target/llvm/llvm_instance.cc
@@ -221,8 +221,11 @@ LLVMTargetInfo::LLVMTargetInfo(LLVMInstance& instance, const TargetJSON& target)
     bool has_arch =
         std::any_of(arches.begin(), arches.end(), [&](const auto& var) { return var == cpu_; });
     if (!has_arch) {
-      LOG(FATAL) << "LLVM cpu architecture `-mcpu=" << cpu_
-                 << "` is not valid in `-mtriple=" << triple_ << "`";
+      // Flag an error, but don't abort. This mimicks the behaviour of 'llc' to
+      // give the code a chance to run with a less-specific target.
+      LOG(ERROR) << "LLVM cpu architecture `-mcpu=" << cpu_
+                 << "` is not valid in `-mtriple=" << triple_ << "`"
+                 << ", cpu architecture ignored";
     }
   }
 
--- a/src/target/llvm/llvm_instance.cc
+++ b/src/target/llvm/llvm_instance.cc
@@ -386,41 +386,21 @@ static const llvm::Target* CreateLLVMTargetInstance(const std::string triple,
   return llvm_instance;
 }
 
-static llvm::TargetMachine* CreateLLVMTargetMachine(
+static std::unique_ptr<llvm::TargetMachine> CreateLLVMTargetMachine(
     const llvm::Target* llvm_instance, const std::string& triple, const std::string& cpu,
-    const std::string& features, const llvm::TargetOptions& target_options,
-    const llvm::Reloc::Model& reloc_model, const llvm::CodeModel::Model& code_model,
+    const std::string& features, const llvm::TargetOptions& target_options = {},
+    const llvm::Reloc::Model& reloc_model = llvm::Reloc::Static,
+    const llvm::CodeModel::Model& code_model = llvm::CodeModel::Small,
 #if TVM_LLVM_VERSION <= 170
-    const llvm::CodeGenOpt::Level& opt_level) {
+    const llvm::CodeGenOpt::Level& opt_level = llvm::CodeGenOpt::Level(0)) {
 #else
-    const llvm::CodeGenOptLevel& opt_level) {
+    const llvm::CodeGenOptLevel& opt_level = llvm::CodeGenOptLevel(0)) {
 #endif
   llvm::TargetMachine* tm = llvm_instance->createTargetMachine(
       triple, cpu, features, target_options, reloc_model, code_model, opt_level);
   ICHECK(tm != nullptr);
 
-  return tm;
-}
-
-static const llvm::MCSubtargetInfo* GetLLVMSubtargetInfo(const std::string& triple,
-                                                         const std::string& cpu_name,
-                                                         const std::string& feats) {
-  // create a LLVM instance
-  auto llvm_instance = CreateLLVMTargetInstance(triple, true);
-  // create a target machine
-  // required minimum: llvm::InitializeAllTargetMCs()
-  llvm::TargetOptions target_options;
-  auto tm = CreateLLVMTargetMachine(llvm_instance, triple, cpu_name, feats, target_options,
-                                    llvm::Reloc::Static, llvm::CodeModel::Small,
-#if TVM_LLVM_VERSION <= 170
-                                    llvm::CodeGenOpt::Level(0));
-#else
-                                    llvm::CodeGenOptLevel(0));
-#endif
-  // create subtarget info module
-  const llvm::MCSubtargetInfo* MCInfo = tm->getMCSubtargetInfo();
-
-  return MCInfo;
+  return std::unique_ptr<llvm::TargetMachine>(tm);
 }
 
 llvm::TargetMachine* LLVMTargetInfo::GetOrCreateTargetMachine(bool allow_missing) {
@@ -428,10 +408,9 @@ llvm::TargetMachine* LLVMTargetInfo::GetOrCreateTargetMachine(bool allow_missing
 
   std::string error;
   if (const llvm::Target* llvm_instance = CreateLLVMTargetInstance(triple_, allow_missing)) {
-    llvm::TargetMachine* tm =
+    target_machine_ =
         CreateLLVMTargetMachine(llvm_instance, triple_, cpu_, GetTargetFeatureString(),
                                 target_options_, reloc_model_, code_model_, opt_level_);
-    target_machine_ = std::unique_ptr<llvm::TargetMachine>(tm);
   }
   ICHECK(target_machine_ != nullptr);
   return target_machine_.get();
@@ -837,7 +816,11 @@ const Array<String> LLVMTargetInfo::GetAllLLVMTargets() const {
 const Array<String> LLVMTargetInfo::GetAllLLVMTargetArches() const {
   Array<String> cpu_arches;
   // get the subtarget info module
-  const auto MCInfo = GetLLVMSubtargetInfo(triple_, "", "");
+  auto llvm_instance = CreateLLVMTargetInstance(triple_, true);
+  std::unique_ptr<llvm::TargetMachine> target_machine =
+      CreateLLVMTargetMachine(llvm_instance, triple_, "", "");
+  const auto MCInfo = target_machine->getMCSubtargetInfo();
+
   if (!MCInfo) {
     return cpu_arches;
   }
@@ -861,7 +844,11 @@ const Map<String, String> LLVMTargetInfo::GetAllLLVMCpuFeatures() const {
     feats += feats.empty() ? attr : ("," + attr);
   }
   // get the subtarget info module
-  const auto MCInfo = GetLLVMSubtargetInfo(triple_, cpu_.c_str(), feats);
+  auto llvm_instance = CreateLLVMTargetInstance(triple_, true);
+  std::unique_ptr<llvm::TargetMachine> target_machine =
+      CreateLLVMTargetMachine(llvm_instance, triple_, cpu_.c_str(), feats);
+  const auto MCInfo = target_machine->getMCSubtargetInfo();
+
   // get all features for CPU
   llvm::ArrayRef<llvm::SubtargetFeatureKV> llvm_features =
 #if TVM_LLVM_VERSION < 180
--- a/python/tvm/target/codegen.py
+++ b/python/tvm/target/codegen.py
@@ -183,7 +183,8 @@ def llvm_get_cpu_features(target=None):
         List of available CPU features.
     """
     assert isinstance(target, Target) or target is None
-    return _ffi_api.llvm_get_cpu_features(target)
+    feature_map = _ffi_api.llvm_get_cpu_features(target)
+    return set(feature_map.keys())
 
 
 def llvm_cpu_has_features(cpu_features, target=None):
--- a/src/target/llvm/llvm_instance.h
+++ b/src/target/llvm/llvm_instance.h
@@ -157,6 +157,14 @@ class LLVMTargetInfo {
   // NOLINTNEXTLINE(runtime/references)
   LLVMTargetInfo(LLVMInstance& scope, const std::string& target_str);
   /*!
+   * \brief Constructs LLVMTargetInfo from `Target`
+   * \param scope LLVMInstance object
+   * \param target TVM JSON Target object for target "llvm"
+   */
+  // NOLINTNEXTLINE(runtime/references)
+  LLVMTargetInfo(LLVMInstance& instance, const TargetJSON& target);
+
+  /*!
    * \brief Destroys LLVMTargetInfo object
    */
   ~LLVMTargetInfo();
@@ -290,11 +298,12 @@ class LLVMTargetInfo {
 
   /*!
    * \brief Get all CPU features from target
-   * \return list with all valid cpu features
+   * \return Map with all valid cpu features as keys and empty string as value. The Map
+   *         is intended to be used as a Set, which TVM does not currently support.
    * \note The features are fetched from the LLVM backend using the target `-mtriple`
    *       and the `-mcpu` architecture, but also consider the `-mattr` attributes.
    */
-  const Array<String> GetAllLLVMCpuFeatures() const;
+  const Map<String, String> GetAllLLVMCpuFeatures() const;
 
   /*!
    * \brief Check the target if has a specific cpu feature
--- a/src/target/llvm/llvm_module.cc
+++ b/src/target/llvm/llvm_module.cc
@@ -697,12 +697,12 @@ TVM_REGISTER_GLOBAL("target.llvm_get_cpu_archlist")
     });
 
 TVM_REGISTER_GLOBAL("target.llvm_get_cpu_features")
-    .set_body_typed([](const Target& target) -> Array<String> {
+    .set_body_typed([](const Target& target) -> Map<String, String> {
       auto use_target = target.defined() ? target : Target::Current(false);
       // ignore non "llvm" target
       if (target.defined()) {
         if (target->kind->name != "llvm") {
-          return Array<String>{};
+          return {};
         }
       }
       auto llvm_instance = std::make_unique<LLVMInstance>();
@@ -722,8 +722,7 @@ TVM_REGISTER_GLOBAL("target.llvm_cpu_has_feature")
       auto llvm_instance = std::make_unique<LLVMInstance>();
       LLVMTargetInfo llvm_backend(*llvm_instance, use_target);
       auto cpu_features = llvm_backend.GetAllLLVMCpuFeatures();
-      bool has_feature = std::any_of(cpu_features.begin(), cpu_features.end(),
-                                     [&](auto& var) { return var == feature; });
+      bool has_feature = cpu_features.find(feature) != cpu_features.end();
       return has_feature;
     });
 
--- a/src/target/parsers/aprofile.cc
+++ b/src/target/parsers/aprofile.cc
@@ -24,9 +24,11 @@
 
 #include "aprofile.h"
 
+#include <memory>
 #include <string>
 
 #include "../../support/utils.h"
+#include "../llvm/llvm_instance.h"
 
 namespace tvm {
 namespace target {
@@ -52,33 +54,6 @@ double GetArchVersion(Optional<Array<String>> attr) {
   return GetArchVersion(attr.value());
 }
 
-static inline bool HasFlag(String attr, std::string flag) {
-  std::string attr_str = attr;
-  return attr_str.find(flag) != std::string::npos;
-}
-
-static inline bool HasFlag(Optional<String> attr, std::string flag) {
-  if (!attr) {
-    return false;
-  }
-  return HasFlag(attr.value(), flag);
-}
-
-static inline bool HasFlag(Optional<Array<String>> attr, std::string flag) {
-  if (!attr) {
-    return false;
-  }
-  Array<String> attr_array = attr.value();
-
-  auto matching_attr = std::find_if(attr_array.begin(), attr_array.end(),
-                                    [flag](String attr_str) { return HasFlag(attr_str, flag); });
-  return matching_attr != attr_array.end();
-}
-
-static bool HasFlag(Optional<String> mcpu, Optional<Array<String>> mattr, std::string flag) {
-  return HasFlag(mcpu, flag) || HasFlag(mattr, flag);
-}
-
 bool IsAArch32(Optional<String> mtriple, Optional<String> mcpu) {
   if (mtriple) {
     bool is_mprofile = mcpu && support::StartsWith(mcpu.value(), "cortex-m");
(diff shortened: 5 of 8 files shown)


cbalint13:
Hi @lhutton1 !

Thanks a lot for picking up the recently introduce llvm reflection (well, kind of) for ARM targets ! 
I also believe it's a far better to query  llvm submodule own inventory than "guessing" what llvm might want.


> Currently, target features are determined by a set of fixed checks on the target string. This works well for checking support of a small number of simple features, but it doesn't scale. Some problems include:
> 
> * There are many non-trivial conditions for which a feature may(not) be available. It is easy to miss these with the current implementation.
> * The inclusion of some features in a target string can imply other features. For example, "+sve2" implies "+sve". This currently isn't taken into account.
> * The tests in tests/cpp/target/parsers/aprofile_test.c suggest that targets such as "llvm -mcpu=cortex-a+neon" and "llvm -mattr=+noneon" are supported target strings. The features will be correctly parsed in TVM, however, they are not valid in LLVM. Therefore, it's possible that TVM and LLVM have different understanding of the features available.

Yes I agree, I was also concerned about, all the mentioned target strings are "non-legit" from llvm point of view.

To reiterate, if one want to use llvm targets properly than the flags concerning llvm submodule should:
  * Target **must have** a  ```-mtriple=<triple>``` and the ```triple``` should be **only** a arch triple.
  * Target **may have** a  ```-mcpu=<cpumodel>``` and the ```cpumodel``` should be **valid** with the ```mtriple``` arch.
  * Target **may have** a ```-mattr=<cpuflags>```, comma separated list where ```'+feat'``` add,  ```'-feat'``` subtracts features.

> This commit uses the more robust LLVM target parser to determine support for the features in TVM. It leverages previous infrastructure added to TVM for obtaining a list of all supported features given an input target, and uses this to check the existance of certain features we're interested in. It should be trivial to grow this list over time. As a result of this change, the problems mentioned above are solved.
> 
> In the current form, this commit drops support for target strings such as "llvm -mcpu=cortex-a+neon" and "llvm -mattr=+noneon". A scan of the codebase suggests this functionality is not in use (only in test cases). Should we feel the need to support them, or have a smoother migration for downstream users of TVM we can add a translator to the parser to convert these into LLVM compatible targets.

As you also underlined subtracting a feature may disable whole set of futures, and llvm do it being cpu+arch aware.
Addition or subtraction can be done with ```+feat``` or ```-feat```, a.f.a.i.k. there is no such thing as ```+no<whatever>``` in llvm.

Here is an example subtracting feature (avx512f aka foundation) that leads to complete disablement of **all** avx512:
```
target = tvm.target.Target("llvm -mcpu=sapphirerapids")
print("BEFORE", tvm.target.codegen.llvm_get_cpu_features(target) )
print()
target = tvm.target.Target("llvm -mcpu=sapphirerapids -mattr=-avx512f")
print( "AFTER", tvm.target.codegen.llvm_get_cpu_features(target) )

["64bit", "64bit-mode", "adx", "aes", "allow-light-256-bit", "amx-bf16", 
"amx-int8", "amx-tile", "avx", "avx2", "avx512bf16", "avx512bitalg", 
"avx512bw", "avx512cd", "avx512dq", "avx512f", "avx512fp16", 
"avx512ifma", "avx512vbmi", "avx512vbmi2", "avx512vl", "avx512vnni", 
"avx512vpopcntdq", "avxvnni", ....]

["64bit", "64bit-mode", "adx", "aes", "allow-light-256-bit", "amx-bf16", 
"amx-int8", "amx-tile", "avx", "avx2", "avxvnni", ... ]
```


I salute this PR, thanks again @lhutton1  !

src/target/llvm/llvm_instance.cc
```diff
@@ -845,10 +847,11 @@ const Array<String> LLVMTargetInfo::GetAllLLVMCpuFeatures() const {
 #else
       MCInfo->getAllProcessorFeatures();
 #endif
-  Array<String> cpu_features;
+  // TVM doesn't have an FFI friendly Set, so use a Map instead for now
+  Map<String, String> cpu_features;
```
cbalint13, line 851: This looks ugly in python backend, would be std::set better ?

See:
```
$ cat ./tvm-target.py 
#!/usr/bin/python3

target = tvm.target.Target("llvm -mcpu=x86-64")
print( tvm.target.codegen.llvm_get_cpu_features(target) )
```
* BEFORE
```
$ ./tvm-target.py 
["64bit", "64bit-mode", "cmov", "cx8", "fxsr", "idivq-to-divl", "macrofusion", 
"mmx", "nopl", "slow-3ops-lea", "slow-incdec", "sse", "sse2", "vzeroupper", "x87"]
```
* AFTER
```
$ ./tvm-target.py 
{"mmx": "", "cmov": "", "idivq-to-divl": "", "slow-incdec": "", "macrofusion": "", 
"fxsr": "", "slow-3ops-lea": "", "x87": "", "nopl": "", "sse2": "", "vzeroupper": "", 
"64bit": "", "sse": "", "cx8": "", "64bit-mode": ""}

```

lhutton1: Good point thanks, I'd forgotten to check this. Due to the restriction on the FFI, perhaps we can fix this on the python side? For example:
```
feature_map = _ffi_api.llvm_get_cpu_features(target)
return set(feature_map.keys())
```
We could also use std::set for the c++ API if desired, we'd just have to make the conversion to/from Map before/after crossing the FFI boundary.

cbalint13: > Good point thanks, 

However I think my request concerns the "pedantic" side and ignores the "performance" gains.

>I'd forgotten to check this. Due to the restriction on the FFI, perhaps we can fix this on the python side? For example:
> 
> ```
> feature_map = _ffi_api.llvm_get_cpu_features(target)
> return set(feature_map.keys())
> ```

Super ! 

That would be enough for python side.

> 
> We could also use std::set for the c++ API if desired, we'd just have to make the conversion to/from Map before/after crossing the FFI boundary.

See, well I think ```Map``` is fine and yes is may be the fastest option concerning lookup speeds in hot-paths.

Thanks @lhutton1 !

src/target/llvm/llvm_instance.cc
```diff
@@ -199,32 +199,37 @@ std::ostream& operator<<(std::ostream& os, const LLVMTargetInfo::Option& opt) {
...
     bool has_arch =
         std::any_of(arches.begin(), arches.end(), [&](const auto& var) { return var == cpu_; });
     if (!has_arch) {
-      LOG(FATAL) << "LLVM cpu architecture `-mcpu=" << cpu_
-                 << "` is not valid in `-mtriple=" << triple_ << "`";
+      // Flag an error, but don't abort. This mimicks the behaviour of 'llc' to
+      // give the code a chance to run with a less-specific target.
+      LOG(ERROR) << "LLVM cpu architecture `-mcpu=" << cpu_
+                 << "` is not valid in `-mtriple=" << triple_ << "`"
+                 << ", cpu architecture ignored";
```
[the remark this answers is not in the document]
lhutton1: Thanks, yes, I agree, that's a good point! It's looks like I've still got some tests to fix in i386, so I'll add it when I update that :)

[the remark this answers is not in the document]
lhutton1: The result of `GetLLVMSubtargetInfo(triple_, "", "")->getCPU().data()` always seems to be generic, so I stuck to hardcoding that for now

lhutton1:
also cc @kparzysz-quic

lhutton1:
friendly ping on this

lhutton1:
@tvm-bot rerun

lhutton1:
the duplication of the ci-arm CI jobs still appears to be occurring - trying a force push to see if that helps

lhutton1:
It seems CI was failing due to a memory leak observed when calling `GetAllLLVMCpuFeatures()` and `GetAllLLVMTargetArches()`. The following is a valgrind report for an executable that creates a new target `tvm::Target("llvm -mtriple=aarch64-linux-gnu")` with this PR's changes:
```
...
==356347== 8,040 (1,560 direct, 6,480 indirect) bytes in 1 blocks are definitely lost in loss record 42,621 of 42,667
==356347==    at 0x4849013: operator new(unsigned long) (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==356347==    by 0x1136B1B9: ??? (in /usr/lib/x86_64-linux-gnu/libLLVM-17.so.1)
==356347==    by 0xBC44347: llvm::Target::createTargetMachine(llvm::StringRef, llvm::StringRef, llvm::StringRef, llvm::TargetOptions const&, std::optional<llvm::Reloc::Model>, std::optional<llvm::CodeModel::Model>, llvm::CodeGenOpt::Level, bool) const (TargetRegistry.h:488)
==356347==    by 0xBC3EF2D: tvm::codegen::CreateLLVMTargetMachine(llvm::Target const*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, llvm::TargetOptions const&, llvm::Reloc::Model const&, llvm::CodeModel::Model const&, llvm::CodeGenOpt::Level const&) (llvm_instance.cc:398)
==356347==    by 0xBC3F0B2: tvm::codegen::GetLLVMSubtargetInfo(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) (llvm_instance.cc:413)
==356347==    by 0xBC41E09: tvm::codegen::LLVMTargetInfo::GetAllLLVMTargetArches() (llvm_instance.cc:843)
==356347==    by 0xBC3D5CA: tvm::codegen::LLVMTargetInfo::LLVMTargetInfo(tvm::codegen::LLVMInstance&, tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void> const&) (llvm_instance.cc:220)
==356347==    by 0xAD6FCE8: tvm::target::parsers::aprofile::GetFeatures(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>) (aprofile.cc:101)
==356347==    by 0xAD70A5E: tvm::target::parsers::aprofile::ParseTarget(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>) (aprofile.cc:137)
==356347==    by 0xAD71C25: tvm::target::parsers::cpu::ParseTarget(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>) (cpu.cc:55)
==356347==    by 0xAEAB127: tvm::runtime::TypedPackedFunc<tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void> (tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>)>::AssignTypedLambda<tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void> (*)(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>)>(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void> (*)(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>))::{lambda(tvm::runtime::TVMArgs const&, tvm::runtime::TVMRetValue*)#1}::operator()(tvm::runtime::TVMArgs const&, tvm::runtime::TVMRetValue*) const (packed_func.h:1826)
==356347==    by 0xAEB3382: tvm::runtime::PackedFuncObj::Extractor<tvm::runtime::PackedFuncSubObj<tvm::runtime::TypedPackedFunc<tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void> (tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>)>::AssignTypedLambda<tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void> (*)(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>)>(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void> (*)(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>))::{lambda(tvm::runtime::TVMArgs const&, tvm::runtime::TVMRetValue*)#1}> >::Call(tvm::runtime::PackedFuncObj const*, tvm::runtime::TVMArgs, tvm::runtime::TVMRetValue*) (packed_func.h:1252)
==356347== 
==356347== 8,040 (1,560 direct, 6,480 indirect) bytes in 1 blocks are definitely lost in loss record 42,622 of 42,667
==356347==    at 0x4849013: operator new(unsigned long) (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==356347==    by 0x1136B1B9: ??? (in /usr/lib/x86_64-linux-gnu/libLLVM-17.so.1)
==356347==    by 0xBC44347: llvm::Target::createTargetMachine(llvm::StringRef, llvm::StringRef, llvm::StringRef, llvm::TargetOptions const&, std::optional<llvm::Reloc::Model>, std::optional<llvm::CodeModel::Model>, llvm::CodeGenOpt::Level, bool) const (TargetRegistry.h:488)
==356347==    by 0xBC3EF2D: tvm::codegen::CreateLLVMTargetMachine(llvm::Target const*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, llvm::TargetOptions const&, llvm::Reloc::Model const&, llvm::CodeModel::Model const&, llvm::CodeGenOpt::Level const&) (llvm_instance.cc:398)
==356347==    by 0xBC3F0B2: tvm::codegen::GetLLVMSubtargetInfo(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) (llvm_instance.cc:413)
==356347==    by 0xBC4217B: tvm::codegen::LLVMTargetInfo::GetAllLLVMCpuFeatures() (llvm_instance.cc:868)
==356347==    by 0xAD6FD01: tvm::target::parsers::aprofile::GetFeatures(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>) (aprofile.cc:102)
==356347==    by 0xAD70A5E: tvm::target::parsers::aprofile::ParseTarget(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>) (aprofile.cc:137)
==356347==    by 0xAD71C25: tvm::target::parsers::cpu::ParseTarget(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>) (cpu.cc:55)
==356347==    by 0xAEAB127: tvm::runtime::TypedPackedFunc<tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void> (tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>)>::AssignTypedLambda<tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void> (*)(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>)>(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void> (*)(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>))::{lambda(tvm::runtime::TVMArgs const&, tvm::runtime::TVMRetValue*)#1}::operator()(tvm::runtime::TVMArgs const&, tvm::runtime::TVMRetValue*) const (packed_func.h:1826)
==356347==    by 0xAEB3382: tvm::runtime::PackedFuncObj::Extractor<tvm::runtime::PackedFuncSubObj<tvm::runtime::TypedPackedFunc<tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void> (tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>)>::AssignTypedLambda<tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void> (*)(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>)>(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void> (*)(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>))::{lambda(tvm::runtime::TVMArgs const&, tvm::runtime::TVMRetValue*)#1}> >::Call(tvm::runtime::PackedFuncObj const*, tvm::runtime::TVMArgs, tvm::runtime::TVMRetValue*) (packed_func.h:1252)
==356347==    by 0xAE8D453: CallPacked (packed_func.h:1256)
==356347==    by 0xAE8D453: tvm::runtime::TVMRetValue tvm::runtime::PackedFunc::operator()<tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void> >(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>&&) const (packed_func.h:1784)
...
```

The problem seems to come from `GetLLVMSubtargetInfo(...)` which creates a target machine instance to get a pointer to an `MCSubtargetInfo` object. The reference to the created target machine is lost and the memory is not freed. I've tried to naively rework the code to remove the GetLLVMSubtargetInfo(...) function and therefore avoid the leak, but happy to hear any better ideas (cc @cbalint13, @kparzysz)

cbalint13:
> It seems CI was failing due to a memory leak observed when calling `GetAllLLVMCpuFeatures()` and `GetAllLLVMTargetArches()`. The following is a valgrind report for an executable that creates a new target `tvm::Target("llvm -mtriple=aarch64-linux-gnu")` with this PR's changes:
> 
> ```
> ...
> ==356347== 8,040 (1,560 direct, 6,480 indirect) bytes in 1 blocks are definitely lost in loss record 42,621 of 42,667
> ==356347==    at 0x4849013: operator new(unsigned long) (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
> ==356347==    by 0x1136B1B9: ??? (in /usr/lib/x86_64-linux-gnu/libLLVM-17.so.1)
> ==356347==    by 0xBC44347: llvm::Target::createTargetMachine(llvm::StringRef, llvm::StringRef, llvm::StringRef, llvm::TargetOptions const&, std::optional<llvm::Reloc::Model>, std::optional<llvm::CodeModel::Model>, llvm::CodeGenOpt::Level, bool) const (TargetRegistry.h:488)
> ==356347==    by 0xBC3EF2D: tvm::codegen::CreateLLVMTargetMachine(llvm::Target const*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, llvm::TargetOptions const&, llvm::Reloc::Model const&, llvm::CodeModel::Model const&, llvm::CodeGenOpt::Level const&) (llvm_instance.cc:398)
> ==356347==    by 0xBC3F0B2: tvm::codegen::GetLLVMSubtargetInfo(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) (llvm_instance.cc:413)
> ==356347==    by 0xBC41E09: tvm::codegen::LLVMTargetInfo::GetAllLLVMTargetArches() (llvm_instance.cc:843)
> ==356347==    by 0xBC3D5CA: tvm::codegen::LLVMTargetInfo::LLVMTargetInfo(tvm::codegen::LLVMInstance&, tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void> const&) (llvm_instance.cc:220)
> ==356347==    by 0xAD6FCE8: tvm::target::parsers::aprofile::GetFeatures(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>) (aprofile.cc:101)
> ==356347==    by 0xAD70A5E: tvm::target::parsers::aprofile::ParseTarget(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>) (aprofile.cc:137)
> ==356347==    by 0xAD71C25: tvm::target::parsers::cpu::ParseTarget(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>) (cpu.cc:55)
> ==356347==    by 0xAEAB127: tvm::runtime::TypedPackedFunc<tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void> (tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>)>::AssignTypedLambda<tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void> (*)(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>)>(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void> (*)(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>))::{lambda(tvm::runtime::TVMArgs const&, tvm::runtime::TVMRetValue*)#1}::operator()(tvm::runtime::TVMArgs const&, tvm::runtime::TVMRetValue*) const (packed_func.h:1826)
> ==356347==    by 0xAEB3382: tvm::runtime::PackedFuncObj::Extractor<tvm::runtime::PackedFuncSubObj<tvm::runtime::TypedPackedFunc<tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void> (tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>)>::AssignTypedLambda<tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void> (*)(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>)>(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void> (*)(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>))::{lambda(tvm::runtime::TVMArgs const&, tvm::runtime::TVMRetValue*)#1}> >::Call(tvm::runtime::PackedFuncObj const*, tvm::runtime::TVMArgs, tvm::runtime::TVMRetValue*) (packed_func.h:1252)
> ==356347== 
> ==356347== 8,040 (1,560 direct, 6,480 indirect) bytes in 1 blocks are definitely lost in loss record 42,622 of 42,667
> ==356347==    at 0x4849013: operator new(unsigned long) (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
> ==356347==    by 0x1136B1B9: ??? (in /usr/lib/x86_64-linux-gnu/libLLVM-17.so.1)
> ==356347==    by 0xBC44347: llvm::Target::createTargetMachine(llvm::StringRef, llvm::StringRef, llvm::StringRef, llvm::TargetOptions const&, std::optional<llvm::Reloc::Model>, std::optional<llvm::CodeModel::Model>, llvm::CodeGenOpt::Level, bool) const (TargetRegistry.h:488)
> ==356347==    by 0xBC3EF2D: tvm::codegen::CreateLLVMTargetMachine(llvm::Target const*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, llvm::TargetOptions const&, llvm::Reloc::Model const&, llvm::CodeModel::Model const&, llvm::CodeGenOpt::Level const&) (llvm_instance.cc:398)
> ==356347==    by 0xBC3F0B2: tvm::codegen::GetLLVMSubtargetInfo(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) (llvm_instance.cc:413)
> ==356347==    by 0xBC4217B: tvm::codegen::LLVMTargetInfo::GetAllLLVMCpuFeatures() (llvm_instance.cc:868)
> ==356347==    by 0xAD6FD01: tvm::target::parsers::aprofile::GetFeatures(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>) (aprofile.cc:102)
> ==356347==    by 0xAD70A5E: tvm::target::parsers::aprofile::ParseTarget(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>) (aprofile.cc:137)
> ==356347==    by 0xAD71C25: tvm::target::parsers::cpu::ParseTarget(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>) (cpu.cc:55)
> ==356347==    by 0xAEAB127: tvm::runtime::TypedPackedFunc<tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void> (tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>)>::AssignTypedLambda<tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void> (*)(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>)>(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void> (*)(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>))::{lambda(tvm::runtime::TVMArgs const&, tvm::runtime::TVMRetValue*)#1}::operator()(tvm::runtime::TVMArgs const&, tvm::runtime::TVMRetValue*) const (packed_func.h:1826)
> ==356347==    by 0xAEB3382: tvm::runtime::PackedFuncObj::Extractor<tvm::runtime::PackedFuncSubObj<tvm::runtime::TypedPackedFunc<tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void> (tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>)>::AssignTypedLambda<tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void> (*)(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>)>(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void> (*)(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>))::{lambda(tvm::runtime::TVMArgs const&, tvm::runtime::TVMRetValue*)#1}> >::Call(tvm::runtime::PackedFuncObj const*, tvm::runtime::TVMArgs, tvm::runtime::TVMRetValue*) (packed_func.h:1252)
> ==356347==    by 0xAE8D453: CallPacked (packed_func.h:1256)
> ==356347==    by 0xAE8D453: tvm::runtime::TVMRetValue tvm::runtime::PackedFunc::operator()<tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void> >(tvm::runtime::Map<tvm::runtime::String, tvm::runtime::ObjectRef, void, void>&&) const (packed_func.h:1784)
> ...
> ```
> 
> The problem seems to come from `GetLLVMSubtargetInfo(...)` which creates a target machine instance to get a pointer to an `MCSubtargetInfo` object. The reference to the created target machine is lost and the memory is not freed. I've tried to naively rework the code to remove the GetLLVMSubtargetInfo(...) function and therefore avoid the leak, but happy to hear any better ideas (cc @cbalint13, @kparzysz)

* Is this happening with llvm18 too ?
* I look into this (across multiple llvm versions), could help me with a simple testcase ?

lhutton1:
Thanks for the quick response @cbalint13! I didn't try with llvm18 yet, only llvm17. Calling GetAllLLVMCpuFeatures() and GetAllLLVMTargetArches() should reproduce it, but I'll be able to come up with a more concrete example tomorrow

cbalint13:
> Thanks for the quick response @cbalint13! I didn't try with llvm18 yet, only llvm17. Calling GetAllLLVMCpuFeatures() and GetAllLLVMTargetArches() should reproduce it, but I'll be able to come up with a more concrete example tomorrow

I'l test it for llvm18 too, just ping me if you have a concrete sample, until than I try invoking as you said (hope to catch it).

lhutton1:
Here is a reproducer:
mem_leak.cpp
```c++
#include "tvm/runtime/registry.h"
#include "tvm/target/target.h"

int main() {
  auto pf = tvm::runtime::Registry::Get("target.llvm_get_cpu_archlist");
  (*pf)(tvm::Target("llvm"));
}
```
Compile:
```bash
g++ -std=c++17 -O2 -fPIC -I{TVM_DIR}/include -I{TVM_DIR}/3rdparty/dmlc-core/include -I{TVM_DIR}/tvm/3rdparty/dlpack/include -DDMLC_USE_LOGGING_LIBRARY=\<tvm/runtime/logging.h\> -o mem_leak_exec mem_leak.cpp -L{TVM_BUILD_DIR} -ldl -ltvm -pthread
```
Run with valgrind:
```bash
LD_PRELOAD="{TVM_BUILD_DIR}/libtvm.so" valgrind --leak-check=full -v --track-origins=yes ./mem_leak_exec
```

Output:
```
...
==475237== 12,369 (1,560 direct, 10,809 indirect) bytes in 1 blocks are definitely lost in loss record 42,596 of 42,630
==475237==    at 0x4849013: operator new(unsigned long) (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==475237==    by 0x12244479: ??? (in /usr/lib/x86_64-linux-gnu/libLLVM-17.so.1)
==475237==    by 0xBC0131B: llvm::Target::createTargetMachine(llvm::StringRef, llvm::StringRef, llvm::StringRef, llvm::TargetOptions const&, std::optional<llvm::Reloc::Model>, std::optional<llvm::CodeModel::Model>, llvm::CodeGenOpt::Level, bool) const (TargetRegistry.h:488)
==475237==    by 0xBBFBC05: tvm::codegen::CreateLLVMTargetMachine(llvm::Target const*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, llvm::TargetOptions const&, llvm::Reloc::Model const&, llvm::CodeModel::Model const&, llvm::CodeGenOpt::Level const&) (llvm_instance.cc:393)
==475237==    by 0xBBFBD8A: tvm::codegen::GetLLVMSubtargetInfo(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) (llvm_instance.cc:408)
==475237==    by 0xBBFEAE1: tvm::codegen::LLVMTargetInfo::GetAllLLVMTargetArches() const (llvm_instance.cc:835)
==475237==    by 0xBBFA2BB: tvm::codegen::LLVMTargetInfo::LLVMTargetInfo(tvm::codegen::LLVMInstance&, tvm::Target const&) (llvm_instance.cc:218)
==475237==    by 0xBC0EE16: tvm::codegen::__mk_TVM8::{lambda(tvm::Target const&)#1}::operator()(tvm::Target const) const (llvm_module.cc:695)
==475237==    by 0xBC188B0: tvm::runtime::TypedPackedFunc<tvm::runtime::Array<tvm::runtime::String, void> (tvm::Target const&)>::AssignTypedLambda<tvm::codegen::__mk_TVM8::{lambda(tvm::Target const&)#1}>(tvm::codegen::__mk_TVM8::{lambda(tvm::Target const&)#1}, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >)::{lambda(tvm::runtime::TVMArgs const&, tvm::runtime::TVMRetValue*)#1}::operator()(tvm::runtime::TVMArgs const, tvm::runtime::TVMRetValue) const (packed_func.h:1826)
==475237==    by 0xBC233EE: tvm::runtime::PackedFuncObj::Extractor<tvm::runtime::PackedFuncSubObj<tvm::runtime::TypedPackedFunc<tvm::runtime::Array<tvm::runtime::String, void> (tvm::Target const&)>::AssignTypedLambda<tvm::codegen::__mk_TVM8::{lambda(tvm::Target const&)#1}>(tvm::codegen::__mk_TVM8::{lambda(tvm::Target const&)#1}, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >)::{lambda(tvm::runtime::TVMArgs const&, tvm::runtime::TVMRetValue*)#1}> >::Call(tvm::runtime::PackedFuncObj const*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, tvm::runtime::TVMRetValue) (packed_func.h:1252)
==475237==    by 0x1092F8: main (in /workspaces/tvm-ethosn/src/tvm/test_mem_leak/cpp_deploy)
...
```

cbalint13:
> > Here is a reproducer:
> > mem_leak.cpp
> 
> Thanks a lot for this, I start to look at it now.

@lhutton1 ,

Here is a patch:  [tvm-llvm-memleak.diff.gz](https://github.com/apache/tvm/files/14762596/tvm-llvm-memleak.diff.gz)
Can confirm that on your side is fine ?

cbalint13:
> > > Here is a reproducer:
> > > mem_leak.cpp
> > 
> > 
> > Thanks a lot for this, I start to look at it now.
> 
> @lhutton1 ,
> 
> Here is a patch: [tvm-llvm-memleak.diff.gz](https://github.com/apache/tvm/files/14762596/tvm-llvm-memleak.diff.gz) Can confirm that on your side is fine ?
> 
> Later EDIT: the patch is against vanilla tvm latest.

@lhutton1 ,

Few comments on the fix here:

* First these LLVM module interrogation functions creates it's **own sessions** independent from the main instance.
* It seems that within same context (the LLVMTargetInfo class) there will be multiple LLVM instances going memleak

I wanted these functions as separate class apart from ```TVMTargetInfo``` but was a requirement at review time of the PR

1. Your fix **is not naive** at all: https://github.com/lhutton1/tvm/commit/31927904b378c142be3044419e38501425805cd4
      - You force creation within a smart pointers until the needed SubtargetInfo interface, this is a effective way IMO
2. [My fix](https://github.com/apache/tvm/files/14762596/tvm-llvm-memleak.diff.gz) only **make sure** to re-use the llvm_instance + llvm_target_machine using existing ```GetOrCreate*``` interface.
      - But we **loose now the constness** of the querying functions here
      - All query parameters (tripe_, cpu_, attrs_) **now must be passed** at ```LLVMTargetInfo``` class creation

---

I honestly would go with your #1 , because it **keeps the constness** and **the independent parameters** of querying.

cbalint13 (approved):

lhutton1:
Thanks @cbalint13 I can confirm your fix also worked at my end. Happy to go with my changes, your reasoning makes sense to me :)

ekalda (approved):
Thanks for all your hard work on that @lhutton1 and @cbalint13 for reviewing and helping to troubleshoot!

Lunderberg:
This PR causes spurious error messages to be printed when loading `libtvm.so`, when the local version of LLVM available does not support one of the CPU architectures defined in [`tag.cc`](https://github.com/apache/tvm/blob/main/src/target/tag.cc#L85).  For example, the following error messages are printed when LLVM 10.0 is used.

image

cbalint13:
> This PR causes spurious error messages to be printed when loading `libtvm.so`, when the local version of LLVM available does not support one of the CPU architectures defined in [`tag.cc`](https://github.com/apache/tvm/blob/main/src/target/tag.cc#L85). For example, the following error messages are printed when LLVM 10.0 is used.
> 
> image

Hi @Lunderberg ,

The messages looks legit, LLVM10 not support those CPU invocations.
* The message maybe could be more explicit by adding **"using LLVM version X.Y"** like e.g.:
   ```LLVM cpu architecture -mcpu=cortex-a78 is not valid in -mtriple=aarch64{...} using LLVM version X.Y```
* Maybe also we could issue "WARN" instead of "ERROR" level here, but the user should be still notified somehow.

---

This behaviour was intentionate and introduced [here @ PR#15761](https://github.com/apache/tvm/pull/15761/files#diff-f13c79368edd27b0c05d044df267746187448906d4310d5617bce89003b8813bR201-R210)
With this ```#1571``` TVM has the ability to lookup into LLVM's *internal catalog* of arches, cpus and their exact features.

Lunderberg:
The messages look reasonable, based on the available support, but I think they shouldn't be emitted unless the user is attempting to use the invalid target, or making an explicit query of the target parameters.  We should not producing an error message when importing TVM.

Regarding changing from a warning to an error, even commenting-out the `LOG(WARNING)` entirely does not fully remove the error messages, as some are produced from LLVM internally.

Lunderberg:
I agree on the use of internal architectures being preferable to hard-coded lists.  The tags were primarily introduced (IIRC) to handle cases where there is no internal architecture that could be queried, such as GPUs with no readily available table to look up, and with limited availability where the compilation shouldn't require queries to a local GPU.

cbalint13:
@Lunderberg ,

> We should not producing an error message when importing TVM.

Can give me a script line how to reproduce (beside LLVM10 presence) ? 
Now that's sounds odd, if true I address a fix to this, will test all iterations llvm=range(10,19)

> Thanks for the discussion @Lunderberg @cbalint13. I agree that we shouldn't remove the error message completely. Just thinking out loud - the problem here seems to be that the targets registered in tag.cc are parsed when loading tvm, is it possible to defer parsing of these registered targets to when they are actually used by the user?

If turns true, we should take this one check out from ```LLVTargetInfo``` constructor and see a better place for it.

I take care of this, if you don't mind this it will be a new PR.

Lunderberg:
> Can give me a script line how to reproduce (beside LLVM10 presence) ?  Now that's sounds odd, if true I address a fix to this, will test all iterations llvm=range(10,19)

@cbalint13 If you add a new tag to `tag.cc` with the `"mcpu"`  host attribute set to a non-existent type (or edit an existing tag), you can reproduce the error with a newer LLVM version.

> I take care of this, if you don't mind this it will be a new PR.

Thank you, and making a new PR would be perfect!

tqchen:
I am also getting additional errors like

```
python test.py 
[15:23:28] /home/tqchen/github/tvm/src/target/parsers/aprofile.cc:97: Warning: Cannot parse target features. LLVM was not compiled with support for Arm(R)-based targets.
[15:23:28] /home/tqchen/github/tvm/src/target/parsers/aprofile.cc:97: Warning: Cannot parse target features. LLVM was not compiled with support for Arm(R)-based targets.
```

this is on a LLVM that was built for rocm. We should not send out an error message during static loading time if ARM target is not used, and only have such error message when we attempt to  use tags in aprofile

LeiWang1999:
It seems that this pull request may lead to a `segmentation fault` issue when build with `llvm+16.0.0`

```bash
root@e01d939002c0:~/pr_workspace/main_tvm# python
Python 3.10.14 (main, Apr  6 2024, 18:45:05) [GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import tvm
Segmentation fault (core dumped)
```

but works with `llvm+10.0.0`

cbalint13:
> It seems that this pull request may lead to a `segmentation fault` issue when build with `llvm+16.0.0`
> 
> ```shell
> root@e01d939002c0:~/pr_workspace/main_tvm# python
> Python 3.10.14 (main, Apr  6 2024, 18:45:05) [GCC 9.4.0] on linux
> Type "help", "copyright", "credits" or "license" for more information.
> >>> import tvm
> Segmentation fault (core dumped)
> ```
> 
> but works with `llvm+10.0.0`

@LeiWang1999 ,

I'll look at this.
You are on llvm-16.0.0 / https://github.com/llvm/llvm-project/releases/tag/llvmorg-16.0.0 / 08d094a ?

LeiWang1999:
Thanks @cbalint13  , I was using prebuilt binary at:

10.0.1: [clang+llvm-10.0.1-x86_64-linux-gnu-ubuntu-16.04.tar.xz](https://github.com/llvm/llvm-project/releases/download/llvmorg-10.0.1/clang+llvm-10.0.1-x86_64-linux-gnu-ubuntu-16.04.tar.xz)
16.0.0: [clang+llvm-16.0.0-x86_64-linux-gnu-ubuntu-18.04.tar.xz](https://github.com/llvm/llvm-project/releases/download/llvmorg-16.0.0/clang+llvm-16.0.0-x86_64-linux-gnu-ubuntu-18.04.tar.xz)

LeiWang1999:
build with llvm+16.0.0 works on commit `31927904b378c142be3044419e38501425805cd4`

LeiWang1999:
hi @cbalint13 , with bisect debugging, I found that commit `726a1416497eeca7bfb7dcdbd799d00b33c39f79` is the first bad commit.

cbalint13:
@LeiWang1999 

* Can give a gdb backtrace (```bt full```, would be very useful) ?
* Also what is your tvm's cmake options and build hash (to see what features might be absent/present) ?

I cannot reproduce the crash on my side.

LeiWang1999:
Thanks @cbalint13 , BT FULL Trace: [gdb.txt](https://github.com/user-attachments/files/17421504/gdb.txt)

cbalint13:
> Thanks @cbalint13 , BT FULL Trace: [gdb.txt](https://github.com/user-attachments/files/17421504/gdb.txt)

Still not able to reproduce :-(
```
$ rpm -q clang llvm16-devel;  git rev-parse HEAD
clang-19.1.0-1.fc42.x86_64
llvm16-devel-16.0.6-9.fc41.x86_64
726a1416497eeca7bfb7dcdbd799d00b33c39f79

$ python -c 'import tvm; print(tvm.target.codegen.llvm_version_major())'
16

$ readelf -a /usr/lib64/libtvm.so | grep NEED
  [ 4] .gnu.version_r    VERNEED          00000000001ae73c  001ae73c
 0x0000000000000001 (NEEDED)             Shared library: [libstdc++.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libLLVM-16.so]
 0x0000000000000001 (NEEDED)             Shared library: [libopenblas.so.0]
 0x0000000000000001 (NEEDED)             Shared library: [libm.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libgcc_s.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [ld-linux-x86-64.so.2]
 0x000000006ffffffe (VERNEED)            0x1ae73c
 0x000000006fffffff (VERNEEDNUM)         6
```


* Can re-upload ```bt full``` with ```set print frame-arguments all``` to see why is```llvm::RegisterTargetMachine``` offended ?
* If I can't still infer the cause, an ultimate request would be to provide a Docker receipt for me to reproduce the bug.

LeiWang1999:
@cbalint13 , after I transfered llvm into 16.0.6, the bug disappeared, it's interesting, thanks.

cbalint13:
@LeiWang1999 

I had only 16.0.6 (precompiled) at hand for current tests, thinking that issue might be in tvm side.
I recommend to use more recent or latest 19.x if this is possible, llvm upstream changes at steep pace.

Thank you for you patience and help !

PR #16429, opened 2024-01-19
[Unity][Dlight][Fix] Reduction rule support dyn-shape epilogue

MasterJH5574:
This PR fixes a bug in the dlight reduction rule, which fails to recognize the pattern of a dynamic-shape epilogue block and therefore errors out when transforming such blocks.

One regression test is provided.


junrushao (approved):
LGTM!

PR #16366, opened 2024-01-08
[CI] Upgrade sccache version to 0.7.*

lhutton1:
Sccache was previously pinned at version 0.3.3, however, there has recently been an issue when installing some of its dependencies:

```
error[E0277]: expected a `Fn(reqwest::Request, &'a mut task_local_extensions::extensions::Extensions, reqwest_middleware::Next<'a>)` closure, found `RetryAfterMiddleware`
```

Upgrading the version of sccache to the latest seems to resolve this error.

In addition, the JFrog artifactory link for boost has been flaky recently, so changing to an alternative
mirror suggested in https://github.com/boostorg/boost/issues/842.

PR #16435, opened 2024-01-19
[CI] Update image tag to 20240126-070121-8ade9c30e

yongwww:
update the images to align with the latest upgrades of emsdk and nodejs.

### DIFF
--- a/ci/jenkins/docker-images.ini
+++ b/ci/jenkins/docker-images.ini
@@ -17,13 +17,13 @@
 
 # This data file is read during when Jenkins runs job to determine docker images.
 [jenkins]
-ci_arm: tlcpack/ci-arm:20240105-165030-51bdaec6
-ci_cortexm: tlcpack/ci-cortexm:20240105-165030-51bdaec6
-ci_cpu: tlcpack/ci-cpu:20240105-165030-51bdaec6
-ci_gpu: tlcpack/ci-gpu:20240105-165030-51bdaec6
-ci_hexagon: tlcpack/ci-hexagon:20240105-165030-51bdaec6
-ci_i386: tlcpack/ci-i386:20240105-165030-51bdaec6
-ci_lint: tlcpack/ci-lint:20240105-165030-51bdaec6
-ci_minimal: tlcpack/ci-minimal:20240105-165030-51bdaec6
-ci_riscv: tlcpack/ci-riscv:20240105-165030-51bdaec6
-ci_wasm: tlcpack/ci-wasm:20240105-165030-51bdaec6
+ci_arm: tlcpack/ci-arm:20240119-070113-614a7a9e3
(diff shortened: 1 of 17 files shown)


tqchen (approved):

tqchen:
cc @Hzfengsy

yongwww:
The CI failed during `sccache --start-server` with s3 bucket, it was introduced by the sccache upgrade before #16366. Seems we can set env. variables `SCCACHE_BUCKET`, `SCCACHE_REGION`, and `AWS_PROFILE`(or `AWS_ACCESS_KEY_ID` plus `AWS_SECRET_ACCESS_KEY` instead of `AWS_PROFILE`+.aws/credentials) as mentioned in [sccache s3](https://github.com/mozilla/sccache/blob/main/docs/S3.md)

yongwww:
updated to the newer image (tag:20240126-070121-8ade9c30e), it was built on Jan 26. CI is green now. 

Actually the current CI doesn’t really use the s3 bucket tvm-sccache-prod for sccache. Instead, the local sccahe cache is used. I have a [testing pr ](https://github.com/apache/tvm/pull/16480) to hardcode an invalid bucket and CI still passed. The sccache 0.3.3 doesn’t verify the aws credential during the —start-server, whereas the newer sccache 0.7.x does. In this PR, I changed it to use local cache if no aws credentials is available.

I conducted a benchmark comparing sccache with and without an S3 bucket on an AWS g5.4xlarge instance. The results showed that the build time without the S3 bucket (using a local cache, with an average build time of 219.48 seconds) is faster compared to using an S3 bucket (232.01 seconds average build time). Given these findings, I recommend not enabling the S3 bucket for sccache at this time. 


Regarding AWS credentials, they are currently not passed to the worker container. One method is to pass AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY through the container as shown [here](https://github.com/apache/tvm/blob/main/ci/jenkins/generated/cpu_jenkinsfile.groovy#L115). However, this approach raises security concerns. We should explore alternatives like using IAM roles, task IAMs, or similar solutions for future needs, especially if we decide to enable S3 for sccache. Useful references include discussions on the best practices for passing AWS credentials to Docker containers ([Stack Overflow](https://stackoverflow.com/questions/36354423/what-is-the-best-way-to-pass-aws-credentials-to-a-docker-container)) and information about IAM roles for Amazon EC2 ([AWS EC2 User Guide](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html)) and task IAM roles ([AWS ECS Developer Guide](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html)).

cc: @tqchen @vinx13 @ajtulloch @naut-brian

Issue #16433, opened 2024-01-19
[Bug] TIR issued buffer name when Applying FuseTIR with duplicated block.


LeiWang1999:
The code to reproduce:
```python
import tvm
import tvm.testing
from tvm import relax
from tvm.script import ir as I, relax as R, tir as T
from tvm import tir
from tvm.ir import IRModule
from tvm.ir.transform import PassContext, module_pass


@I.ir_module
class Before:
    @T.prim_func(private=True)
    def add(
        A: T.Buffer((T.int64(4096), T.int64(4096)), "float16"),
        Out: T.Buffer((T.int64(4096), T.int64(4096)), "float16"),
    ):
        for i, j in T.grid(T.int64(4096), T.int64(4096)):
            with T.block("add"):
                vi, vj = T.axis.remap("SS", [i, j])
                Out[vi, vj] = A[vi, vj] + T.float16(1.0)

    @T.prim_func(private=True)
    def take(
        A: T.Buffer((T.int64(4096), T.int64(4096)), "float16"),
        B: T.Buffer((T.int64(1),), "int32"),
        T_take: T.Buffer((T.int64(1), T.int64(4096)), "float16"),
    ):
        for ax0, ax1 in T.grid(T.int64(1), T.int64(4096)):
            with T.block("T_take"):
                v_ax0, v_ax1 = T.axis.remap("SS", [ax0, ax1])
                T_take[v_ax0, v_ax1] = A[B[v_ax0], v_ax1]

    @T.prim_func(private=True)
    def add1(
        A: T.Buffer((T.int64(1), T.int64(4096)), "float16"),
        Out: T.Buffer((T.int64(1), T.int64(4096)), "float16"),
    ):
        for i, j in T.grid(T.int64(1), T.int64(4096)):
            with T.block("add"):
                vi, vj = T.axis.remap("SS", [i, j])
                Out[vi, vj] = A[vi, vj] + T.float16(2.0)

    @R.function
    def main(
        input_ids: R.Tensor((1,), dtype="int32"),
        input_embeds: R.Tensor((4096, 4096), dtype="float16"),
    ) -> R.Tensor((1, 4096), dtype="float16"):
        cls = Before
        with R.dataflow():
            gv: R.Tensor((1, 4096), dtype="float16") = cls.fused_func(input_ids, input_embeds)
            R.output(gv)
        return gv

    @R.function(private=True)
    def fused_func(
        input_ids: R.Tensor((1,), dtype="int32"),
[...]

PR #16437, opened 2024-01-19
[Unity] Improve buffer allocation for handling duplicated buffer names.

LeiWang1999:
Refer to this issue 
https://github.com/apache/tvm/issues/16433 , the relax Pass FuseTIR creates intermediate name by 

```c++
n->name = param->name_hint + "_intermediate";
```

which may lead to some bugs when we encounter two fused blocks with the same param name, for example what I encountered in resnet-18, the conv+add+multiply+add, the fused two add operators' block may point to the same block.

So we should provide a method to get an unique name for each in duplicate buffers.

```cpp
auto unify_name_hints = [this, &buffer, &param]() {
        String base_name = buffer->name;
        String unique_name = base_name + "_intermediate";
        size_t unique_id = 0;
        std::unordered_set<std::string> names;

        for (auto& _buffer : func_info_.alloc_buffers) {
          names.insert(_buffer->name);
        }

        while (names.find(unique_name) != names.end()) {
          unique_name = unique_name + "_" + std::to_string(++unique_id);
        }
 return unique_name;
 };
```

### DIFF
--- a/src/relax/transform/fuse_tir.cc
+++ b/src/relax/transform/fuse_tir.cc
@@ -682,10 +682,25 @@ class FusedTIRConstructor : public ExprVisitor {
       const tir::Var& param = output_params[i];
       const tir::Buffer& buffer = func->buffer_map.at(param);
 
+      auto unify_name_hints = [this, &buffer, &param]() {
+        String base_name = buffer->name;
+        String unique_name = base_name + "_intermediate";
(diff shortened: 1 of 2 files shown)


LeiWang1999:
Please CC @Hzfengsy

junrushao (approved):
LGTM!

PR #16438, opened 2024-01-19
[Metal] Dispatch numerically stable tanh for metal

CharlieFRuan:
Prior to this PR, `tanh(x)`returns `NaN` on metal when `x > 45.0`.

Metal's built-in tanh is implemented as `(t - 1.0) / (t + 1.0)`, where `t = exp(2.0 * x)`. Hence for large `x`, `t` becomes `inf`, causing `tanh(x)` to be `NaN`.

A numerically stable `tanh` is implemented for `llvm`, this PR lifts it to `src/target/intrin_rule.cc` and apply the same rule for metal as well.

### DIFF
--- a/src/target/intrin_rule.cc
+++ b/src/target/intrin_rule.cc
@@ -134,6 +134,24 @@ PrimExpr DispatchFastErf(const PrimExpr& e) {
   return res;
 }
 
+PrimExpr DispatchNumericalStableTanh(const PrimExpr& e) {
+  using tir::make_const;
+  using tir::make_zero;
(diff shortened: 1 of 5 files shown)


CharlieFRuan:
cc @MasterJH5574 @tqchen

MasterJH5574 (approved):

CharlieFRuan:
Observed the same issue on webgpu as well; updated to fix.

junrushao (approved):
Thanks! This looks great!

PR #16439, opened 2024-01-20
[Unity]  Support cumsum with pure int32

jinhongyii:
This PR fixes a bug on attr handling in data type rewriter and enforces i32 buffer in cumsum function definition, which ensures that cumsum can be run on a machine with int32 but not int64.


jinhongyii:
cc: @MasterJH5574

junrushao (approved):
LGTM!

PR #16454, opened 2024-01-23
[BugTIR] fix thread_sync occurs in letstmt

JackWeiw:
See original discuss
[LayerNorm Error in thread_storage_sync when read x into shared memory](https://discuss.tvm.apache.org/t/layernorm-error-in-thread-storage-sync-when-read-x-into-shared-memory/16269)

I try to read x into shared memory to accelerate layernorm, [script here 1](https://gist.github.com/JackWeiw/f873daaff32212b0b19cf91fda463007)

but error occurs in pass thread_sorage_sync pass.I found it is because the error in lower LetStmt ,[lowered script](https://gist.github.com/JackWeiw/6737f4faa1b486b389721cf7f3f4ad4d)

### DIFF
--- a/src/tir/transforms/storage_access.cc
+++ b/src/tir/transforms/storage_access.cc
@@ -51,6 +51,10 @@ void StorageAccessVisitor::VisitExpr_(const BufferLoadNode* op) {
   }
   // traverse child
   StmtExprVisitor::VisitExpr_(op);
+  // push to the scope
+  scope_.back().push_back(curr_stmt_);
+  // clear access entry.
+  curr_stmt_.access.clear();
 }
 
 void StorageAccessVisitor::VisitStmt_(const BufferStoreNode* op) {
@@ -83,6 +87,20 @@ void StorageAccessVisitor::VisitStmt_(const BufferStoreNode* op) {
 
 void StorageAccessVisitor::VisitStmt_(const EvaluateNode* op) {
   allow_append_ = true;
+  ICHECK_EQ(curr_stmt_.access.size(), 0U);
+  curr_stmt_.stmt = op;
+  StmtExprVisitor::VisitStmt_(op);
+  // push to the scope
+  if (curr_stmt_.access.size() != 0) {
+    scope_.back().push_back(curr_stmt_);
+    curr_stmt_.access.clear();
+  }
+  allow_append_ = false;
+}
+
+void StorageAccessVisitor::VisitStmt_(const LetStmtNode* op) {
+  allow_append_ = true;
+  curr_stmt_.access.clear();
   ICHECK_EQ(curr_stmt_.access.size(), 0U);
   curr_stmt_.stmt = op;
   StmtExprVisitor::VisitStmt_(op);
--- a/src/tir/transforms/storage_access.cc
+++ b/src/tir/transforms/storage_access.cc
@@ -51,10 +51,6 @@ void StorageAccessVisitor::VisitExpr_(const BufferLoadNode* op) {
   }
   // traverse child
   StmtExprVisitor::VisitExpr_(op);
-  // push to the scope
-  scope_.back().push_back(curr_stmt_);
-  // clear access entry.
-  curr_stmt_.access.clear();
 }
 
 void StorageAccessVisitor::VisitStmt_(const BufferStoreNode* op) {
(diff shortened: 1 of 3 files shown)


Hzfengsy:
cc @vinx13 @spectrometerHBH

JackWeiw:
cc @vinx13 @spectrometerHBH please spend a little time do a review if you see this

src/tir/transforms/storage_access.cc
```diff
@@ -94,6 +98,20 @@ void StorageAccessVisitor::VisitStmt_(const EvaluateNode* op) {
   allow_append_ = false;
 }
 
+void StorageAccessVisitor::VisitStmt_(const LetStmtNode* op) {
+  allow_append_ = true;
+  curr_stmt_.access.clear();
```
vinx13, line 103: shouldn't call clear here, it can only be used after finishing handling a statement

tests/python/tir-transform/test_tir_transform_thread_sync.py
```diff
@@ -160,8 +160,49 @@ def expected(A: T.Buffer((4, 4), "float32"), E: T.Buffer((4, 4), "float32")):
...
+            T.tvm_thread_allreduce(
+                T.uint32(1),
+                in_thread_A_temp_1[0],
+                T.bool(True),
+                cross_thread_A_temp_1[0],
+                threadIdx_x,
+            )
+
+    mod = run_passes(func)
+    assert "T.tvm_storage_sync" in str(mod)
```
vinx13, line 200: it's better to use `tvm.ir.assert_structural_equal` if possible since it's not obvious where `T.tvm_storage_sync` should be inserted

vinx13 (changes requested):

tests/python/tir-transform/test_tir_transform_thread_sync.py
```diff
@@ -160,8 +160,49 @@ def expected(A: T.Buffer((4, 4), "float32"), E: T.Buffer((4, 4), "float32")):
...
+        A_shared = T.allocate([512], "float32", "shared")
+        in_thread_A_temp = T.allocate([1], "float32", "local")
+        cross_thread_A_temp = T.allocate([1], "float32", "local")
+        threadIdx_x = T.launch_thread("threadIdx.x", 128)
+        A_shared_1 = T.Buffer((512,), data=A_shared, scope="shared")
+        for ax0 in range(512):
+            A_shared_1[ax0] = A[blockIdx_x * 512 + ax0]
+        in_thread_A_temp_1 = T.Buffer((1,), data=in_thread_A_temp, scope="local")
+        in_thread_A_temp_1[0] = T.float32(0)
+        with T.LetStmt(in_thread_A_temp_1[0] + A_shared_1[threadIdx_x]) as A_temp:
```
JackWeiw, line 177: in case here, before take change, `StorageAccessVisitor` will call  `VisitExpr_(const BufferLoadNode* op)` directly (not called by father call like `VisitStmt_(const BufferStoreNode* op)` traverse child), so `ICHECK(allow_append_) << op << " " << scope.to_string();` will return error.

if we take `VisitStmt_(const LetStmtNode* op)` and traverse child of `LetStmtNode` 
(case here it will traverse `BufferLoadNode` A_shared which will increase curr_stmt_.access by 1, and then visit `BufferStoreNode` `in_thread_A_temp` , `ICHECK_EQ(curr_stmt_.access.size(), 0U);` will return error)
Do you have any insights on how to solve this problem? @vinx13

src/tir/transforms/storage_access.cc
JackWeiw: can you check my new solution?

src/tir/transforms/storage_access.cc
```diff
@@ -94,6 +94,21 @@ void StorageAccessVisitor::VisitStmt_(const EvaluateNode* op) {
...
```
vinx13, line 99:
```suggestion
```
This should work. You can remove this line since the assertion `ICHECK_EQ(curr_stmt_.access.size(), 0U);` holds

JackWeiw: thank you for your review, the unecessary clear has been removed

vinx13 (approved):

PR #16462, opened 2024-01-24
[Unity][Relax] Avoid removing `PrimFunc`s that are used for in-place calls in `FuseTIR`

slyubomirsky:
This is a temporary fix that prevents `FuseTIR` from removing `PrimFunc`s that are used in in-place calls. Previously, `FuseTIR` removed all `PrimFunc`s that are not marked public other than those that can be handled in fusion. The permanent solution to this issue would be to handle these `PrimFunc`s in `FuseTIR` and `FuseOps`, which I am planning to attempt, but that should not be a blocker on using in-place `PrimFunc`s with those passes (which are part of the compilation pipeline).

### DIFF
--- a/src/relax/transform/dataflow_inplace.cc
+++ b/src/relax/transform/dataflow_inplace.cc
@@ -928,12 +928,7 @@ class ModuleInplaceTransformer : public ExprMutator {
 
     // note: this might be a good time to get rid of the old legalized function, but we don't do it
     // now because later ops might need the same one. Instead, we will clean up at the end
-    
-    // TODO(@tvm-team): Making these public so that FuseTIR does not get rid of them.
-    // The fix for this would be to properly handle inplace calls in FuseOps and FuseTIR
-    auto new_gv = builder_->AddFunction(
-        WithAttr(legal_primfunc, tvm::attr::kGlobalSymbol, inline_legal_op_name),
-        inline_legal_op_name);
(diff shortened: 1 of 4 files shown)


Lunderberg:
Would it be easier to fix the underlying issue in `FuseTIR`?  It looks like it should be doable by having [this loop](https://github.com/apache/tvm/blob/main/src/relax/transform/fuse_tir.cc#L883) maintain a copy of every function in `funcs_to_keep`, rather than a subset, then adding `DeadCodeElimination()` to [this list of passes](https://github.com/apache/tvm/blob/main/src/relax/transform/fuse_tir.cc#L1046).  That way, `FuseTIR` keeps around all the original unfused functions, and it gets cleaned up by the dedicated pass afterwards.

slyubomirsky:
I'm working on getting fusion for in-place cases working in PrimFuncs. These PrimFuncs will not be dead code either, since they will actually be called in the final program. I'll see if it will be just as easy as putting them in the `funcs_to_keep` set, though.

slyubomirsky:
Now this checks for the in-place calls in FuseTIR (only temporary, I hope, as I plan to handle fusion for in-place PrimFuncs correctly soon). We can decide which way we prefer.

Lunderberg (approved):

slyubomirsky:
No longer necessary thanks to [16487](https://github.com/apache/tvm/pull/16487)

PR #16463, opened 2024-01-24
[Unity][Cutlass] Fix C source generation of dense operation

creaiter:
It seems there are two issues while generating c source of dense operation using cutlass.

1. Even though the dense operation contains bias parameters, the generated c code doesn't reflect that bias correctly.
2. The weight shape of dense operation looks flipped for the 1st and 2nd dimensions.

I tried to fix above bugs. Please check that and leave feedback.


masahi:
You can now send a PR to the `main`. Can you switch a branch or reopen?

creaiter:
Okay! I'm going to reopen this PR to the `main`.

I noticed that the second issue already resolved, thus the reopened PR will only contain the change of the first issue.

PR #16465, opened 2024-01-24
[Redo][Unity] Split DecomposeOpsForTraining into two steps

Lunderberg:
This is a reapplication of https://github.com/apache/tvm/pull/15954, after resolving the breakages that required reverting in https://github.com/apache/tvm/pull/16442.  The regex matching is now implemented without the `#include <regex>` from the C++ stdlib, to avoid ABI incompatibility with pytorch.

Prior to this commit, the `DecomposeOpsForTraining` transform directly replaced `relax.nn.batch_norm` into more primitive relax operations.  This required the decomposed form of `relax.nn.batch_norm` to be duplicated with `DecomposeOpsForInference`.  This commit refactors the pass to occur in two steps, first to apply training-specific mutations, and then to decompose.

Having a clear `DecomposeOps` pass also has a clear single location for operator decomposition, which may be migrated into the operator definition in the future, similar to `FLegalize`.


slyubomirsky:
Is having the two separate passes necessary for reducing code duplication for batch norm? It does come at the cost of an extra traversal.

Lunderberg:
It isn't strictly necessary, but I'd like to move in that direction as a first step in removing the `LegalizeOpsForInference` step altogether.  This came up in a conversation [here](https://github.com/apache/tvm/pull/15842#discussion_r1358449742) regarding potential use cases of allowing `FLegalize` to be implemented in terms of other relax operations.

Currently, there are two distinct transforms, `DecomposeOpsForTraining` and `DecomposeOpsForInference`.  If a model uses either `R.nn.batch_norm` or `R.nn.layer_norm`, a user must apply one of these two transforms prior to calling `tvm.relax.build`.  This long-term goal of this change is to remove that requirement.  By splitting the transform into two steps, the training flow has an optional pass `MutateOpsForTraining` and a mandatory pass `DecomposeOps`, while the inference flow has a single mandatory pass `DecomposeOps`.

Since the `DecomposeOps` pass is required for both use cases, and the change it makes is a special case of `FLegalize`, a follow-up PR can then define `FLegalize` for `R.nn.batch_norm` and `R.nn.layer_norm`, and remove `DecomposeOps` altogether.

slyubomirsky:
Ah, that seems like a good reason then. :+1: Some bigger simplifications in the works.

slyubomirsky (approved):
These changes seem reasonable and, per your comment, set us up for further simplifications down the line.

Lunderberg:
Sounds good.  Re-running CI as I let the results get more stale than I'd like, then (assuming no new failures arise) merging in.

PR #16466, opened 2024-01-24
[Relax] Memory planning for "partially dynamic" shapes

MasterJH5574:
This PR improves the memory planning to plan for tensors with shape such as `(m, n)` where only `m` has an integer upper bound (say `m <= 20`).

Prior to this PR, the tensor is not planned even though `m` has an upper bound. This is not as ideal since it does not fully leverage the upper bound info.

This PR enhances the planning, so that a memory chunk of `20 * n` will be allocated first, and the tensor will then be allocated out from this chunk. This enables us to fully leverage the upper bound info.

A unit test is provided.


junrushao:
This idea is pretty much similar to the "unified memory pool" in S-LoRA. It's not as feature-complete as S-LoRA, so please take a look whenever you have time

junrushao (approved):

PR #16468, opened 2024-01-25
[LINT] Fix pylint issues in test_dma_builtin.py

ekalda:
We see `no-self-argument` and `wrong-import-order` pylint errors in downstream CI. Eyeballing the file, both of these errors are legit, it's not clear why the upstream CI didn't catch them, potentially because of the inherent instability of pylint running on multiple threads.


lhutton1 (approved):
Thanks @ekalda!

PR #16471, opened 2024-01-25
[Transform] Preserve all reachable functions in FuseTIR

Lunderberg:
Prior to this commit, the `FuseTIR` pass had custom logic to determine whether a `PrimFunc` should be kept in the output `IRModule`.  This commit replaces this check in `FuseTIR` with a post-processing by `DeadCodeElimination`.


slyubomirsky:
Could you add the test case from that PR? It probably will solve that problem.

src/relax/transform/fuse_tir.cc
```diff
@@ -950,63 +940,66 @@ class TIRFuseMutator : public ExprMutator {
...
-            arg_list.push_back(arg);
+      GlobalVar old_gvar = Downcast<GlobalVar>(call->op);
+      auto relax_func = Downcast<Function>(orig_mod_->Lookup(old_gvar));
+
+      if (!relax_func->HasNonzeroAttr(attr::kPrimitive)) {
+        // The callee is not a primitive function, no need to fuse.
+        return call;
+      }
+
+      tir::PrimFunc fused_tir = FusedTIRConstructor::GetFusedTIR(orig_mod_, old_gvar);
```
slyubomirsky, line 951: Are you sure we aren't likely to call this more than once per global var? As long as it's not likely or costly, this approach is probably fine.

slyubomirsky:
See [16487](https://github.com/apache/tvm/pull/16487), the in-place case will be handled.

src/relax/transform/fuse_tir.cc
Lunderberg: It probably would be called several times.  The fused Relax function is generated once per pattern match in `FuseOpsByPattern`, and is then de-duplicated, leaving many calls to a single `kPrimitive` function.  This would generate the fused TIR function once per callsite, after which the resulting functions would be de-duplicated.  Whether that is good or bad would probably depend on the usefulness for `call_tir_inplace`, so I'll take a look at #16487.

slyubomirsky: Oh so there's probably no issue with calling it more often than in the current way the pass is written. That seems good to me.

Lunderberg:
> See https://github.com/apache/tvm/pull/16487, the in-place case will be handled.

Sounds good.  I think I still like having this change overall, as it avoids having duplicate logic for DCE.  When 16487, I'll rebase this PR on top of it.

PR #16472, opened 2024-01-25
[Unity] Improved error message in tvm::relax::UpdateStructInfo

Lunderberg:
If the struct info cannot be updated, provide the expression, the previous struct info, and the new struct info in the error message.


slyubomirsky (approved):
Thanks! It's a handy improvement.

Lunderberg:
Thank you!  Any time that I need more information for debugging, I try to add it to the error message so that it’s easier the next time around.

PR #16473, opened 2024-01-25
[Unity] Improve error message in tensor_to_shape struct inference

Lunderberg:
If the tensor argument does not have a valid shape, then the error message will include the expression, the argument, and the struct info of the argument.


slyubomirsky (approved):
Thanks!

PR #16482, opened 2024-01-28
[Doc] Fix Docstring in `extern.py` for Sphinx

Hzfengsy:
Fix a missing line in the docstring of `extern.py`, which would cause Sphinx to fail to build the documentation.


yongwww (approved):
LGTM

PR #16484, opened 2024-01-29
[SVE] Add vscale builtin

ekalda:
Add a `vscale` builtin and lowering to `llvm.vscale`. This will be used in subsequent patches for expressing scalable vectors in TIR.

### DIFF
--- a/include/tvm/tir/builtin.h
+++ b/include/tvm/tir/builtin.h
@@ -909,6 +909,11 @@ TVM_DLL const Op& anylist_setitem_call_packed();
  */
 TVM_DLL const Op& anylist_setitem_call_cpacked();
 
+/*!
+ * \brief Get the target's vscale value
+ */
+TVM_DLL const Op& vscale();
+
 /*! \brief The kind of structure field info used in intrinsic */
 enum TVMStructFieldKind : int {
   // array head address
--- a/include/tvm/tir/builtin.h
+++ b/include/tvm/tir/builtin.h
@@ -910,7 +910,8 @@ TVM_DLL const Op& anylist_setitem_call_packed();
 TVM_DLL const Op& anylist_setitem_call_cpacked();
 
 /*!
- * \brief Get the target's vscale value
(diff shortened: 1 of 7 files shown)


zhangwm-pt (approved):

include/tvm/tir/builtin.h
```diff
@@ -909,6 +909,11 @@ TVM_DLL const Op& anylist_setitem_call_packed();
  */
 TVM_DLL const Op& anylist_setitem_call_cpacked();
 
+/*!
+ * \brief Get the target's vscale value
```
lhutton1, line 913: Is it worth adding a link to the LLVM lang ref here? https://llvm.org/docs/LangRef.html#llvm-vscale-intrinsic

lhutton1 (approved):
I had one small suggestion, but otherwise looks good to me!

also cc @tqchen @cbalint13 @Lunderberg @kparzysz-quic

tests/python/codegen/test_target_codegen_aarch64.py
```diff
@@ -476,5 +474,23 @@ def check_correct_assembly(type):
     check_correct_assembly(type=dtype)
 
 
+@pytest.mark.skipif(
+    llvm_version_major() < 10, reason="Vscale is not supported in earlier versions of LLVM"
```
cbalint13, line 478: Just small nit, 
In ```codegen_llvm.cc``` we have here ```#if TVM_LLVM_VERSION >= 110```, so it would be  ```major < 11``` instead of ```major < 10``` ?
I might be wrong, but being a test case only it is waivable.

cbalint13 (approved):
This looks good to me.

include/tvm/tir/builtin.h
ekalda: Good idea, added it to the docstring :)

tests/python/codegen/test_target_codegen_aarch64.py
ekalda: Thanks, good catch, fixed now!

PR #16487, opened 2024-01-30
[Unity][Transform] Handle `call_tir_inplace` in `FuseTIR` and `FuseOps`

slyubomirsky:
This PR handles in-place `PrimFunc`s in `FuseTIR`, namely by propagating the in-place indices. This required a few updates in the two passes but nothing very heavy-duty.

### DIFF
--- a/src/relax/transform/fuse_tir.cc
+++ b/src/relax/transform/fuse_tir.cc
@@ -17,6 +17,7 @@
  * under the License.
  */
 #include <tvm/relax/analysis.h>
+#include <tvm/relax/attrs/op.h>
 #include <tvm/relax/expr_functor.h>
 #include <tvm/relax/struct_info.h>
 #include <tvm/relax/transform.h>
@@ -476,8 +477,11 @@ class FusedTIRConstructor : public ExprVisitor {
   void VisitExpr_(const CallNode* call) final {
     ExprVisitor::VisitExpr_(call);
     static const Op& call_tir_op_ = Op::Get("relax.call_tir");
-    ICHECK(call->op == call_tir_op_)
-        << "Only call_tir is supported in primitive function, but got: " << GetRef<Expr>(call);
+    static const Op& call_tir_inplace_op_ = Op::Get("relax.call_tir_inplace");
+
+    ICHECK(call->op == call_tir_op_ || call->op == call_tir_inplace_op_)
+        << "Only call_tir and call_tir_inplace are supported in primitive function, but got: "
+        << GetRef<Expr>(call);
 
     // Step 1. Get Global var and PrimFunc
     GlobalVar gv = Downcast<GlobalVar>(call->args[0]);
@@ -503,7 +507,9 @@ class FusedTIRConstructor : public ExprVisitor {
     MapInputBuffer(prim_func, call->args[1]);
     const Array<Array<PrimExpr>>& output_buffer_shapes = GetCallTIROutputShapes(call);
 
-    AllocateIntermediateBuffer(GetRef<Expr>(call), prim_func, output_buffer_shapes);
+    // TODO(@tvm-team): We should be able to avoid intermediate allocations for in-place calls.
+    // Currently the same logic is done for in-place calls to avoid crashes or garbage output.
+    AllocateIntermediateBuffer(call, prim_func, output_buffer_shapes);
 
     // Step 6. Update tir_vars
     if (call->args.size() > 2) {
@@ -566,7 +572,8 @@ class FusedTIRConstructor : public ExprVisitor {
    */
   static Array<Array<PrimExpr>> GetCallTIROutputShapes(const CallNode* call) {
     static const Op& call_tir_op_ = Op::Get("relax.call_tir");
-    ICHECK(call->op.same_as(call_tir_op_));
+    static const Op& call_tir_inplace_op_ = Op::Get("relax.call_tir_inplace");
+    ICHECK(call->op.same_as(call_tir_op_) || call->op.same_as(call_tir_inplace_op_));
     ICHECK_EQ(call->sinfo_args.size(), 1);
     auto get_tensor_shape = [](const TensorStructInfoNode* sinfo) {
       const auto* shape_expr = sinfo->shape.as<ShapeExprNode>();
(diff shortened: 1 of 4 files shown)


tests/python/relax/test_transform_fuse_tir.py
```diff
@@ -1930,5 +1930,251 @@ def main(
...
+                    A[v_ax0, v_ax1] = A[v_ax0, v_ax1]
+
+        @R.function(private=True)
+        def fused_add_exp_squeeze(
+            x: R.Tensor((10, 20), dtype="float32"), p0: R.Tensor((), dtype="float32")
+        ) -> R.Tensor((10, 20), dtype="float32"):
+            R.func_attr({"Primitive": 1})
+            cls = Module
+            with R.dataflow():
+                # this overwrites x and is actually evil but we are doing it just to test the pass
```
Lunderberg, line 1977: I like the comment calling it out.  Since this is a private function, I could see it being valid (and useful!) in cases where the calling scope never uses `x` again.  Seems like we keep circling around linear types, and whether transfer of ownership should be representable when calling relax functions.

Lunderberg: Can we add to the comment that this use of `call_tir_inplace` is deliberately not inserted by `relax.transform.DataflowUseInplaceCalls`?

src/relax/transform/fuse_tir.cc
```diff
@@ -438,9 +440,35 @@ class FusedTIRConstructor : public ExprVisitor {
...
+
     const Array<tir::Buffer>& buffers = (*it).second;
+
+    // map of input buffers to indices (helpful for detecting in-place inputs)
+    std::unordered_map<tir::Buffer, Integer, ObjectPtrHash, ObjectPtrEqual> buffer_to_idx;
+    std::unordered_map<tir::Var, Integer, ObjectPtrHash, ObjectPtrEqual> input_to_idx;
+    for (size_t i = 0; i < func_info_.params.size(); i++) {
+      input_to_idx[func_info_.params[i]] = Integer(i);
+    }
+    for (auto kv : func_info_.buffer_map) {
```
Lunderberg, line 452: Nitpick: Using structured bindings (`for(auto [var, buffer]: func_info_.buffer_map)`) would make this easier to read.

src/relax/transform/fuse_tir.cc
```diff
@@ -438,9 +440,35 @@ class FusedTIRConstructor : public ExprVisitor {
...
+
+    // numbered separately because the number of output *vars* might differ from the
+    // number of outputs if there are in-place inputs
+    int out_idx = 0;
     for (size_t i = 0; i < buffers.size(); ++i) {
-      tir::Var param = tir::Var("p_output" + std::to_string(i), PrimType(DataType::Handle()));
+      // Do not add output vars for in-place inputs
+      // (i.e., already listed in the buffer map. This would result
+      // in duplicates in the buffer map otherwise)
+      if (buffer_to_idx.count(buffers[i])) {
```
Lunderberg, line 465: Nitpick: Using `.count` followed by `operator[]` requires two lookups into the map.  This should be avoided, by using `.find` instead.

```c++
if (auto it = buffer_to_idx.count(buffers[i]); it != buffer_to_idx.end()) {
    inplace_indices_.push_back(it.second);
}
```

src/relax/transform/fuse_tir.cc
```diff
@@ -367,17 +368,18 @@ class FusedTIRConstructor : public ExprVisitor {
...
+                                                              const GlobalVar& gv) {
     FusedTIRConstructor visitor(mod, gv->name_hint);
     BaseFunc f = mod->Lookup(gv);
     CHECK(f->IsInstance<relax::FunctionNode>())
         << "Expected relax functions, but got: " << f->GetTypeKey();
     CHECK(f->HasNonzeroAttr(relax::attr::kPrimitive))
         << "Expected a function with attr `kPrimitive`";
     visitor(Downcast<relax::Function>(f));
-    return visitor.fused_tir_;
+    return {visitor.fused_tir_, visitor.inplace_indices_};
```
Lunderberg, line 382: Can we add a safety check that all values in `visitor.inplace_indices_` are unique?  Non-unique values would mean that the fused function returns two tensors, one of which is an in-place operation that overwrites the other.

This should only be a safety check, as any failures would mean that an upstream pass had made an error by inserting `call_tir_inplace` in the first place.  However, since accidental aliasing would be rather tricky to hunt down at runtime, and since this would be an easy place to catch it, I think it would be worth adding a check.

src/relax/transform/fuse_tir.cc
```diff
@@ -897,8 +973,11 @@ class TIRFuseMutator : public ExprMutator {
     for (const auto& [gv, func] : mod->functions) {
       // Only fuse primitive relax functions
       if (func->IsInstance<relax::FunctionNode>() && func->HasNonzeroAttr(attr::kPrimitive)) {
-        tir::PrimFunc fused_tir = FusedTIRConstructor::GetFusedTIR(mod, gv);
-        mutator.fused_tir_funcs_.Set(gv, fused_tir);
+        const auto& [prim_func, indices] = FusedTIRConstructor::GetFusedTIR(mod, gv);
+        mutator.fused_tir_funcs_.Set(gv, prim_func);
+        if (!indices.empty()) {
```
Lunderberg, line 978: Nitpick: Using `if (indices.size())` instead of `if (!indices.empty())` would avoiding double-negatives and make the condition easier to read.  (Though, personal preference as `if (indices.size())` relies on conversion of non-zero `size_t` to `true`.)

src/relax/transform/fuse_tir.cc
```diff
@@ -985,26 +1065,34 @@ class TIRFuseMutator : public ExprMutator {
...
-        // Step b. Create call_tir
+        // Step b. Create call_tir or call_tir_inplace
         Array<Expr> call_args = {fused_tir_gv, Tuple(arg_list)};
         if (!tir_vars.empty()) {
           call_args.push_back(ShapeExpr(tir_vars));
         }
-        return Call(call_tir_op_, call_args, call->attrs, {GetStructInfo(call)});
+        Op call_op = call_tir_op_;
+        Attrs call_attrs = call->attrs;
+        if (inplace_indices_.count(old_gv)) {
```
Lunderberg, line 1083: Same request here for using `.count` followed by `.at`.

```c++
if(auto it = inplace_indices_.find(old_gv); it != inplace_indices_.end()) {
    ...
}
```

tests/python/relax/test_transform_fuse_tir.py
```diff
@@ -1930,5 +1930,251 @@ def main(
...
+                    cls.fused_add_exp_squeeze,
+                    (x, p0),
+                    out_sinfo=R.Tensor((10, 20), dtype="float32"),
+                )
+                R.output(gv1)
+            return gv1
+
+    _check(Module, Expected)
+
+
```
Lunderberg, line 2178: Can we add a test case in which `main` calls two different relax `kPrimitive` functions, both of which use the same underlying TIR functions, but only one of which uses  `T.call_tir_inplace`?

I think it should already be handled correctly in the current implementation, but a lot of potential optimization (e.g. caching the result of `MapArgsToBuffer` by the argument StructInfo) would break it, and so it would good to include.

tests/python/relax/test_transform_fuse_tir.py
```diff
@@ -1930,5 +1930,251 @@ def main(
...
+        @T.prim_func(private=True)
+        def add_inplace(
+            A: T.Buffer((T.int64(10), T.int64(20)), "float32"), B: T.Buffer((), "float32")
+        ):
+            T.func_attr({"tir.noalias": T.bool(True)})
+            for ax0, ax1 in T.grid(T.int64(10), T.int64(20)):
+                with T.block("T_add"):
+                    v_ax0, v_ax1 = T.axis.remap("SS", [ax0, ax1])
+                    T.reads(A[v_ax0, v_ax1], B[()])
```
Lunderberg, line 1946: Nitpick: Unless they're relevant to the specific test case, removing the `T.reads` and `T.writes` annotations can make the test case easier to read.

Lunderberg (changes requested):
Overall, looks good!  The main requests for changing are to make it more robust against upstream errors.

src/relax/transform/fuse_tir.cc
slyubomirsky: Good advice

src/relax/transform/fuse_tir.cc
slyubomirsky: The entries in the buffer map are supposed to be unique anyway, but we can use a set instead to enforce that.

slyubomirsky: Better to catch the error earlier than later, I agree.

src/relax/transform/fuse_tir.cc
slyubomirsky: Kind of subjective, I think "not empty" is readable (sounds more like how it would be phrased verbally). I'll change it if you prefer it.

tests/python/relax/test_transform_fuse_tir.py
slyubomirsky: Good idea

tests/python/relax/test_transform_fuse_tir.py
slyubomirsky: This speaks to my lack of familiarity with TIR :sweat_smile: I'll see how I can compress it

tests/python/relax/test_transform_fuse_tir.py
slyubomirsky: How exactly should this test case work? PrimFuncs used in-place will differ from those used in destination-passing style due to not having a destination parameter.

tests/python/relax/test_transform_fuse_tir.py
slyubomirsky: Yes, ownership is exactly the situation where you would want to do this. This is doubly evil because the function is still marked and treated as pure but technically isn't. I will expand the comment.

tests/python/relax/test_transform_fuse_tir.py
slyubomirsky: I added one, is it what you had in mind?

src/relax/transform/fuse_tir.cc
Lunderberg: Probably more personal preference, so either works.

tests/python/relax/test_transform_fuse_tir.py
Lunderberg: Thank you, and yes, that is what I had in mind.

tests/python/relax/test_transform_fuse_tir.py
Lunderberg: No worries, and thank you for the updates!  My preference is to make the test cases have as little extraneous information as possible, even when it requires a few TIR/TVMScript tricks to do so.

I'm wondering if it would be useful to update the TVMScript printing, so the `T.reads` and `T.writes` annotations are only printed if they are different from what would automatically be inferred.  That way, when developers print a function for use in a test case, the duplicate information isn't printed.

Lunderberg (approved):
Thank you for the changes, and looks good!

PR #16493, opened 2024-01-30
[Relax][Pass] Skip data type node for CSE pass

abhikran-quic:
- The problem is seen when an arg of relax op is dtype


slyubomirsky (approved):
Thanks, this is probably another case that doesn't make sense to fold. Could you please also update the comment in that function?

abhikran-quic:
> Thanks, this is probably another case that doesn't make sense to fold. Could you please also update the comment in that function?

Thanks @slyubomirsky! I have updated the comment.

PR #16497, opened 2024-01-31
[Instrument] Add default instrument to print all passes

quic-sanirudh:
A small patch to add default instruments that prints all passes before or after each pass. This is used very frequently when debugging and we don't know which pass could cause a particular change


yongwww (approved):
it would be good to add a test case for it

quic-sanirudh:
> it would be good to add a test case for it

Thanks for the review. I wasn't sure exactly what to check for in the tests and I've tried something simple, let me know if it can be improved.

yongwww:
@quic-sanirudh it was landed. Thanks for the work!

PR #16498, opened 2024-01-31
[Relax][Frontent] "tensor_ir_inplace" op

MasterJH5574:
This PR introduces the `tensor_ir_inplace_op` for frontend so that we can leverage our `call_tir_inplace` in SLM model definition flow.

One unit test is added. This PR also fixed a few typos in type annotations.

### DIFF
--- a/tests/python/relax/test_frontend_nn_op.py
+++ b/tests/python/relax/test_frontend_nn_op.py
@@ -619,6 +619,111 @@ def test_tensor_ir_op():
     tvm.ir.assert_structural_equal(irmodule, Expected)
 
 
+def test_tensor_ir_inplace_op():
+    hidden_size = 4096
+    dtype = "float16"
+
+    @T.prim_func
+    def inplace_take(
+        var_weight: T.handle, var_pos: T.handle, var_embeddings: T.handle, offset: T.int64
+    ):
+        T.func_attr({"tir.noalias": T.bool(True)})
+        vocab_size = T.int64()
+        weight = T.match_buffer(var_weight, (vocab_size, hidden_size), dtype)
+        seq_len = T.int64()
+        total_seq_len = T.int64()
+        pos = T.match_buffer(var_pos, (seq_len,), "int32")
+        embeddings = T.match_buffer(var_embeddings, (total_seq_len, hidden_size), dtype)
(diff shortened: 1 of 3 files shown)


slyubomirsky:
Thanks for fixing the typos :slightly_smiling_face: Can't believe I had them there in the first place. Is the main reason to add this op so you could use an inline `PrimFunc` with it?

MasterJH5574:
@slyubomirsky You're welcome! The main reason introducing this op is to use `call_tir_inplace` under the principle of our new nn.Module interface: so that we can hide the definition of the TIR in a file (e.g., https://github.com/mlc-ai/mlc-llm/blob/0c5f88155/python/mlc_chat/model/llama/llama_model.py#L295-L303), and expose a clean interface without Relax concept to frontend model definition side (e.g., https://github.com/mlc-ai/mlc-llm/blob/0c5f88155/python/mlc_chat/model/llama/llama_model.py#L295-L303).

jinhongyii (approved):

tests/python/relax/test_frontend_nn_op.py
```diff
@@ -619,6 +619,111 @@ def test(qkv: R.Tensor((1, 1, 24, 16), dtype="float16"), offset: R.Shape(["offse
...
+                v0, v1 = T.axis.remap("SS", [ax0, ax1])
+                T.reads(weight[pos[v0], v1], pos[v0])
+                T.writes(embeddings[v0, v1])
+                embeddings[v0 + offset, v1] = weight[pos[v0], v1]
+
+    class Model(Module):
+        def test(
+            self, embedding_table: Tensor, input_ids: Tensor, embedding_dst: Tensor, offset: int
+        ):
+            tensor_expr_op_out = op.tensor_ir_op(
```
csullivan, line 648: Did you mean to call `op.tensor_ir_inplace_op` here? It doesn't look like you are testing the new `nn.op.tensor_ir_inplace_op` added in this PR.

MasterJH5574: Ooooooops sorry my bad. Is that updated? Or I can find a chance to update the test next time.

MasterJH5574: Thank you so much for letting me know!

PR #16500, opened 2024-01-31
[Unity] Support storage reuse for dynamic shapes

vinx13:
Before this PR, dynamic shapes require upper bound of variables to be provided in order to use storage planning. We can relax this requirement, for shapes with unknown bound, we can look up other tensors with the same symbolic shapes. This can be helpful for deep learning models where the layers with the same configurations are usually repeated since there are many objects with the same shapes.

This PR changed the `StorageToken` to use `PrimExpr` bytes which can be integer or symbolic. For symbolic shapes, we put the tokens into a special buckets for looking up.

cc @MasterJH5574 @tqchen


MasterJH5574 (approved):
LGTM. Thank you @vinx13!

PR #16507, opened 2024-02-02
[TIR] Fix typo in StorgeRewrite

jinhongyii:
This PR correctly disables smem reusing when target is vulkan


jinhongyii:
cc: @MasterJH5574

MasterJH5574:
A bit more context here: when building to Vulkan, we don't want to enable shared memory reuse, in order to avoid dtype mismatch issues when reusing smem from buffers with different dtypes.

Previously the flag for Vulkan is incorrectly set, which didn't effectively disable the shared memory reuse for Vulkan.

And this PR corrects the issue.

PR #16512, opened 2024-02-02
Cut pytest-lazy-fixture

Liam-Sturge:
Pytest-lazy-fixture doesn't work with `pytest==8.0.0`

The following error occurs when pytest-lazy-fixture is called:

    `AttributeError: 'CallSpec2' object has no attribute 'funcargs'`

This has been raised as an issue on the pytest-lazy-fixture project, but the repository is not in active development. See https://github.com/TvoroG/pytest-lazy-fixture/issues/65 for details.

TVM project code-base makes very little reference to pytest-lazy-fixture, so this patch removes the library from the project altogether, in order to resolve the problem.

CC @neildhickey @lhutton1 @tqchen

### DIFF
--- a/docker/install/ubuntu2004_install_python_package.sh
+++ b/docker/install/ubuntu2004_install_python_package.sh
@@ -43,5 +43,4 @@ pip3 install --upgrade \
     junitparser==2.4.2 \
     six \
     tornado \
-    pytest-lazy-fixture \
     git+https://github.com/jax-ml/ml_dtypes.git@v0.2.0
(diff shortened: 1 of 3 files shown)


neildhickey (approved):
@Liam-Sturge thanks for this contribution.

We'll have to regenerate the correct docker images and update once this change has gone in.

Perhaps we can put up a draft review pointing to the new docker images once they are built? Just to be doubly sure CI works still

lhutton1 (approved):
LGTM! The checks on the docker image have passed: https://github.com/apache/tvm/pull/16522

lhutton1:
Thanks @Liam-Sturge @neildhickey!

PR #16513, opened 2024-02-02
[Target] Automatically detect system triple when not specified by the user

lhutton1:
Currently, when a default compile target such as `llvm` is specified, it implies `llvm -keys=cpu` which tends to imply x86 related components being used during compilation e.g. the schedules registered in TOPI. This can be confusing for a user when compiling on other architectures, especially when other tools such as `llc` infer the default target based on the host.

When the target kind is `llvm`, this commit uses the "target.llvm_get_system_triple" functionality to automatically detect
`mtriple` when one has not been provided in the target string. The target will be updated to one that uses the `mtriple` of the host: `llvm -> llvm -mtriple=<system-triple>`. When compiling on Arm(R)-based targets, this has the added benefit of automatically introducing `-keys=arm_cpu` to the target improving the schedule selection.

Lots of tests are currently using targets such as `llvm` or similar which has resulted in a lack of coverage of other targets such as `arm_cpu`. As part of this commit, failing test cases which have simple / obvious issues have been fixed. Others that likely need more thought have been skipped. In doing so, it reduces the number of modifications and simplifies the review for this change.

Note: ~This PR is marked as draft while checking and fixing other failures in CI. Tests marked as skipped containing "issue link" in the reason will have issues added and the related reason will be updated when CI is green.~

This commit is a follow up of the changes made in: #14981

Co-authored-by: Jack Frankland <jack.frankland@arm.com>

### DIFF
--- a/src/target/parsers/cpu.cc
+++ b/src/target/parsers/cpu.cc
@@ -44,8 +44,6 @@ TargetJSON ParseTarget(TargetJSON target) {
   // Try to fill in the blanks by detecting target information from the system
   if (kind == "llvm" && !mtriple.defined() && !mcpu.defined()) {
     String system_triple = DetectSystemTriple().value_or("");
-    LOG(WARNING) << "Explicit mtriple or mcpu was not provided. Using system detected mtriple: "
-                 << system_triple << ".";
     target.Set("mtriple", system_triple);
   }
 
--- a/src/target/parsers/cpu.cc
+++ b/src/target/parsers/cpu.cc
@@ -43,7 +43,10 @@ TargetJSON ParseTarget(TargetJSON target) {
 
   // Try to fill in the blanks by detecting target information from the system
   if (kind == "llvm" && !mtriple.defined() && !mcpu.defined()) {
-    target.Set("mtriple", DetectSystemTriple().value_or(""));
+    String system_triple = DetectSystemTriple().value_or("");
+    LOG(WARNING) << "Explicit mtriple or mcpu was not provided. Using system detected mtriple: "
+                 << system_triple << ".";
+    target.Set("mtriple", system_triple);
   }
 
   if (mprofile::IsArch(target)) {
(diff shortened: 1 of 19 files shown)


lhutton1:
cc @ekalda @neildhickey @cbalint13 @kparzysz-quic would be good to hear your thoughts before I proceed with creating the issues for the failing tests (due to test coverage in CI increasing).

cbalint13 (approved):
It is awesome👌 to see this llvm backend reflecting itself !

Implementation here looks good to me.
Thanks a lot for picking up such effort to go through the (quite many) testcases with it.

src/target/parsers/cpu.cc
```diff
@@ -28,7 +28,24 @@ namespace target {
...
+
 TargetJSON ParseTarget(TargetJSON target) {
+  String kind = Downcast<String>(target.Get("kind"));
+  Optional<String> mtriple = Downcast<Optional<String>>(target.Get("mtriple"));
+  Optional<String> mcpu = Downcast<Optional<String>>(target.Get("mcpu"));
+
+  // Try to fill in the blanks by detecting target information from the system
+  if (kind == "llvm" && !mtriple.defined() && !mcpu.defined()) {
+    target.Set("mtriple", DetectSystemTriple().value_or(""));
+  }
```
cbalint13, line 47: One small nit, in this case would be nice to leave a message for the user about facts:

 * explicit mtriple and/or explict mcpu wasnt provided
 * backend here detected an X mtriple and/or and Y mcpu

lhutton1: Just added, thanks!

lhutton1: After looking through the CI logs, I'm wondering if emitting a warning message is not the best idea here - there are way too many warnings. Perhaps it can be implemented in some driver of tvm instead, e.g. TVMC, and emitted once?

ekalda (approved):
Thanks @lhutton1 and @FranklandJack for looking into this! The patch has been open for a bit, so let's see if CI is still happy...

ekalda:
Thanks @lhutton1 @FranklandJack @cbalint13!

PR #16523, opened 2024-02-05
[SVE] Change the dtype of Ramp and Broadcast lanes to PrimExpr

ekalda:
This change will allow us to express scalable vectors through Ramp and Broadcast nodes, e.g.
```
vec = tvm.tir.expr.Ramp(0, 1, 4 * tvm.tir.vscale())
```
We will use negative values for `runtime::DataType` the encode the scalable lane values, e.g. the above example would result in `lanes` = -4. That's because the lanes in `runtime::DataType` are tied to DLPack standard which uses `uint16_t` for the lanes. The conversion happens in the node definitions and `runtime::DataType`, so the `int` and `uint16_t` values should never be exposed to the API user, especially after the string support has been added.

Also include the TVMScript support for scalable Ramp and Broadcasts.

Note that this patch doesn't include lowering to the appropriate LLVM vectors, support for data type string representation or `LoopVectorizer` support. All of these will be part of future patches.


ekalda:
cc @neildhickey @lhutton1 @tqchen @cbalint13 @Lunderberg @Anndrey24

ekalda:
@tvm-bot rerun

lhutton1 (commented):
Thanks for the great work @ekalda! I had a look over and noticed a few nitpicks :)

src/arith/pattern_match.h
```diff
@@ -628,10 +628,11 @@ inline PRampExpr<TBase, TStride, TLanes> ramp(const Pattern<TBase>& base,
...
 
 template <typename TBase>
-inline PRampExpr<TBase, PConstWithTypeLike<TBase>, PConst<int>> ramp(const Pattern<TBase>& base,
-                                                                     int stride, int lanes) {
-  return PRampExpr<TBase, PConstWithTypeLike<TBase>, PConst<int>>(
-      base.derived(), PConstWithTypeLike<TBase>(base.derived(), stride), PConst<int>(lanes));
+inline PRampExpr<TBase, PConstWithTypeLike<TBase>, PConstWithTypeLike<TBase>> ramp(
+    const Pattern<TBase>& base, int stride, int lanes) {
+  return PRampExpr<TBase, PConstWithTypeLike<TBase>, PConstWithTypeLike<TBase>>(
+      base.derived(), PConstWithTypeLike<TBase>(base.derived(), stride),
```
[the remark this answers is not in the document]
Lunderberg: Thank you for drawing attention to it, and I think this overload looks reasonable with the `int lanes` argument.  The purpose of this overload is to define patterns that have a specific value, but whose dtype doesn't matter for the purpose of matching.  So, `ramp(pvar_base, 1, 4)` would match vectorized groups of 4 values, regardless of whether the `4` was stored as `IntImm(DataType::Int(32), 4)` or `IntImm(DataType::Int(64), 4)`.

src/arith/rewrite_simplify.h
```diff
@@ -221,6 +221,8 @@ class RewriteSimplifier::Impl : public IRMutatorWithAnalyzer {
   bool CanProveGreaterEqual(const PrimExpr& x, int64_t val) {
     return analyzer_->CanProveGreaterEqual(x, val);
   }
+  // Whether the lanes are scalable
+  bool ScalableLanes(const PrimExpr& lanes) { return !lanes.as<IntImmNode>(); }
```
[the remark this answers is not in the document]
Lunderberg: Seconding the stricter checks here.  As it is, this function would state that `Mul(2, 3)` is a scalable expression, because it is something other than an integer.

src/arith/scalable_expression.cc
```diff
@@ -0,0 +1,62 @@
...
+PrimExpr CanonicalizeScalableLanes(const PrimExpr& lanes) {
+  PVar<IntImm> multiplier;
+  PCallExpr<PVscaleOp> vscale;
+
+  PrimExpr new_lanes;
+
+  if ((multiplier * vscale).Match(lanes)) {
+    new_lanes = lanes;
+  } else if ((vscale * multiplier).Match(lanes)) {
+    new_lanes =
```
Lunderberg, line 52: Is this canonicalization necessary?  It seems like it would occur by default when applying the simplification passes.  If it is necessary, we should match the behavior of the simplifier, with constants collected to the RHS of expressions.

I'd recommend instead having a `std::optional<int> ExtractScalableLanes(const PrimExpr& lanes)` method.  That way, the calling scope could use `if(auto scale = ExtractScalableLanes(lanes)) { ... }`, to immediately have access to the extracted value.  As it is, the two uses of this function need to first check if the expression is scalable, then manually extract the scale factor.

```c++
if (PMatchesOneOf{
      multiplier * vscale,
      vscale * multiplier,
    }.Match(lanes)) {
  return vscale.Eval()->value;
} else  {
  return std::nullopt;
}
```

include/tvm/runtime/data_type.h
```diff
@@ -114,17 +118,28 @@ class DataType {
...
+  bool is_vector() const {
+    int encoded_lanes = static_cast<int16_t>(data_.lanes);
+    return encoded_lanes != 0 && encoded_lanes != 1;
+  }
   /*! \return whether type is a bool vector type. */
   bool is_vector_bool() const { return is_vector() && bits() == 1; }
   /*! \return whether type is a Void type. */
   bool is_void() const { return code() == DataType::kHandle && bits() == 0 && lanes() == 0; }
+  /*! \return Whether the type is scalable. */
+  bool is_scalable() const { return static_cast<int16_t>(data_.lanes) < 0; }
```
tqchen, line 130: is_scalable_vector

include/tvm/runtime/data_type.h
```diff
@@ -114,17 +118,28 @@ class DataType {
...
   /*!
    * \brief Create a new data type by change lanes to a specified value.
    * \param lanes The target number of lanes.
    * \return the result type.
    */
   DataType with_lanes(int lanes) const { return DataType(data_.code, data_.bits, lanes); }
+  /*!
+   * \brief Create a new scalable data type by changing the lanes to a specified value.
+   * \param lanes The target number of lanes.
```
tqchen, line 139: I feel the lanes concept was different from the real lanes. This indicate a scaling factor of vscale, maybe vscale_factor? and with_scalable_vscale_factor?

include/tvm/runtime/data_type.h
```diff
@@ -247,6 +262,9 @@ class DataType {
  * \return Number of bytes needed.
  */
 inline int GetVectorBytes(DataType dtype) {
+  if (dtype.is_scalable()) {
+    LOG(FATAL) << "Cannot get vector bytes of scalable vector";
```
tqchen, line 266: runtime Data type cannot be scalable vector

tqchen (changes requested):

include/tvm/runtime/data_type.h
```diff
@@ -90,7 +91,10 @@ class DataType {
   /*! \return number of bytes to store each scalar. */
   int bytes() const { return (bits() + 7) / 8; }
   /*! \return number of lanes in the data. */
-  int lanes() const { return static_cast<int>(data_.lanes); }
+  int lanes() const {
```
tqchen, line 94: it might be useful to distinguish between lanes and vscale_factor, by having two functions, only allow lanes to corresponds to static vector lanes

tqchen: We can use a different function for scalable vector vscale multiplier

include/tvm/runtime/data_type.h
ekalda: Thanks @tqchen, thinking about it, you're right, the "lanes" of the scalable vectors in the current implementation is a bit of a misnomer and in general causes a lot of issues where the scalability is silently ignored or dropped in the passes. So I'm in favour of separating the APIs for fixed length and scalable vectors. Here's a proposal for cleaning it up (it should address your other comments in this file as well):

* Rename `is_scalable()` -> `is_scalable_vector()` - return `True` if it is scalable vector, `False` otherwise
* Add `is_fixed_length_vector()` method to check if it is a fixed length vector
* `is_vector()` should return `True` if it is a vector (scalable or fixed length) and `False` otherwise
* Reserve `lanes()` for fixed length vectors, i.e. if this function is called on a scalable vector, return an error
* Add `vscale_factor()` that returns the integer multiplier (so it would be a scalable vector equivalent of `lanes()`)
* Rename `with_scalable_lanes(lanes)` -> `with_scalable_vscale_factor(vscale_factor)`


Throughout the codebase `dtype.lanes() != 1` is used as a shorthand to test for "vectorness", it would be good to replace these instances with `is_vector()`/`is_fixed_length_vector()`/`is_scalable_vector()` as appropriate.

src/arith/int_set.cc
```diff
@@ -466,14 +466,21 @@ class IntervalSetEvaluator : public ExprFunctor<IntervalSet(const PrimExpr&)> {
...
+                              op->dtype);
+        } else {
+          return Combine<Add>(analyzer_, base,
+                              IntervalSet(make_const(t, vstride * (lanes - 1)), make_zero(t)),
+                              op->dtype);
+        }
+      } else { /* Scalable vector */
+        if (vstride > 0) {
+          return Combine<Add>(analyzer_, base, IntervalSet(make_zero(t), pos_inf()), op->dtype);
+        }
```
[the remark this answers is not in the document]
ekalda: Yeah we probably should :)

src/arith/rewrite_simplify.h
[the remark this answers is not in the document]
ekalda: Agreed, I'll change it to explicitly check for `vscale`.

src/arith/scalable_expression.cc
ekalda: > Is this canonicalization necessary? It seems like it would occur by default when applying the simplification passes. If it is necessary, we should match the behavior of the simplifier, with constants collected to the RHS of expressions.

I suppose this canonicalization is not strictly necessary, but it has proven useful in things like testing the scalable Ramps in isolation. It shouldn't be a lot of computational overhead, so I think it wouldn't hurt to force the order. But yes, let's go for `vscale * 4` to match the behaviour of the rest of the stack.

> I'd recommend instead having a std::optional<int> ExtractScalableLanes(const PrimExpr& lanes) method. That way, the calling scope could use if(auto scale = ExtractScalableLanes(lanes)) { ... }, to immediately have access to the extracted value. As it is, the two uses of this function need to first check if the expression is scalable, then manually extract the scale factor.

```
if (PMatchesOneOf{
      multiplier * vscale,
      vscale * multiplier,
    }.Match(lanes)) {
  return vscale.Eval()->value;
} else  {
  return std::nullopt;
}
```

Thanks, yes that sounds much better :)

include/tvm/runtime/data_type.h
tqchen: in this case, perhaps we can rename `is_vector` to `is_scalable_or_fixed_length_vector()` to be more explicit.

Otherwise LGTM

src/arith/scalable_expression.cc
ekalda: This has been addressed in the latest version of the PR.

include/tvm/runtime/data_type.h
ekalda: The fixed length and scalable APIs have been separated as outlined here in the latest patch.

include/tvm/runtime/data_type.h
ekalda: Done

include/tvm/runtime/data_type.h
ekalda: Done

tqchen (approved):

lhutton1:
Thanks @ekalda @tqchen @Lunderberg

PR #16525, opened 2024-02-05
[WEB] Move ArtifactCache to Interface, Support Cache delete and Batch Delete, Remove typo

DiegoCao:
- Move old artifact cache as a interface template to support future IndexDB cache fetch
- Add deleteInCache api in cache template
- Fix one test path error
- Parallelize Downloading Process

Co-authored-by: DavidGOrtega <g.ortega.david@gmail.com> for parallel download

### DIFF
--- a/web/src/runtime.ts
+++ b/web/src/runtime.ts
@@ -26,6 +26,7 @@ import { Memory, CachedCallStack } from "./memory";
 import { assert, StringToUint8Array } from "./support";
 import { Environment } from "./environment";
 import { FunctionInfo, WebGPUContext } from "./webgpu";
+import { ArtifactCacheTemplate } from "./artifact_cache";
 
 import * as compact from "./compact";
 import * as ctypes from "./ctypes";
@@ -985,7 +986,7 @@ export type InitProgressCallback = (report: InitProgressReport) => void;
 /**
  * Cache to store model related data.
  */
-export class ArtifactCache {
+export class ArtifactCache implements ArtifactCacheTemplate {
   private scope: string;
   private cache?: Cache;
 
@@ -1017,6 +1018,14 @@ export class ArtifactCache {
       .then(requests => requests.map(request => request.url))
       .then(cacheKeys => keys.every(key => cacheKeys.indexOf(key) !== -1))
       .catch(err => false);
+  }
+
+  async deleteInCache(url: string) {
+    if (this.cache === undefined) {
+      this.cache = await caches.open(this.scope);
+    }
+    const result = await this.cache.delete(url);
+    return result;
   }
 }
 
@@ -1451,7 +1460,7 @@ export class Instance implements Disposable {
   }
 
   /**
-   * Fetch NDArray cache from url.
+   * Given cacheUrl, search up items to fetch based on cacheUrl/ndarray-cache.json
    *
    * @param ndarrayCacheUrl The cache url.
    * @param device The device to be fetched to.
@@ -1477,6 +1486,7 @@ export class Instance implements Disposable {
     this.cacheMetadata = { ...this.cacheMetadata, ...(list["metadata"] as Record<string, any>) };
   }
 
+
   /**
    * Fetch list of NDArray into the NDArrayCache.
    *
@@ -1489,7 +1499,7 @@ export class Instance implements Disposable {
     ndarrayCacheUrl: string,
     list: Array<NDArrayShardEntry>,
     device: DLDevice,
-    artifactCache: ArtifactCache
+    artifactCache: ArtifactCacheTemplate
   ) {
     const perf = compact.getPerformance();
     const tstart = perf.now();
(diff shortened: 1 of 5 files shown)


tqchen:
please fix the ci errors

DiegoCao:
Will re-open after supporting batch delete weights

CharlieFRuan:
Will take a look tomorrow; thank you @DiegoCao!

web/src/runtime.ts
```diff
@@ -1477,6 +1489,41 @@ export class Instance implements Disposable {
     this.cacheMetadata = { ...this.cacheMetadata, ...(list["metadata"] as Record<string, any>) };
   }
 
+  /**
+   * Delete NDArray from cache url.
```
CharlieFRuan, line 1493: Would be great if we could have more documentation on this function's behavior. e.g. "given `cacheUrl`, we search up items to delete based on `cacheUrl/ndarray-cache.json`", etc. 

One thing worth pointing out is that, the `ndarray-cache.json` and tokenizer files would still linger in the cache after this function. We only delete the shard files.

web/src/runtime.ts
```diff
@@ -1477,6 +1489,41 @@ export class Instance implements Disposable {
...
+   *
+   * @param ndarrayCacheUrl
+   * @param device
+   * @param cacheScope
+   */
+  async deleteNDArrayCache(
+    ndarrayCacheUrl: string,
+    device: DLDevice,
```
CharlieFRuan, line 1501: device is not used here, perhaps we can remove

web/src/runtime.ts
```diff
@@ -1477,6 +1489,41 @@ export class Instance implements Disposable {
...
+      list = await result.json();
+    }
+    const arrayentry = list["records"] as Array<NDArrayShardEntry>;
+    for (let i = 0; i < arrayentry.length; ++i){
+      const dataUrl = new URL(arrayentry[i].dataPath, ndarrayCacheUrl).href;
+      try {
+        const result = await(await artifactCache.deleteInCache(dataUrl));
+        if (result === false){
+          return false;
+        }
```
CharlieFRuan, line 1518: Should we end the deletion if one of the entry is not found? Or should we ensure everything inside `ndarry-cache` is deleted? I prefer the latter -- let me know what you think! If we go with the latter, this function does not need to return a `boolean` but can be `void`. Whichever we go with, we need to make sure the function's behavior is well defined, matching its docstring.

web/src/runtime.ts
```diff
@@ -1477,6 +1489,41 @@ export class Instance implements Disposable {
...
   }
 
+  /**
+   * Delete NDArray from cache url.
+   *
+   * @param ndarrayCacheUrl
+   * @param device
+   * @param cacheScope
+   */
+  async deleteNDArrayCache(
```
CharlieFRuan, line 1499: I feel like this function should be outside of `Instance`, just like `hasNDArrayInCache()`, since it is more like a helper function. We can then expose this function in https://github.com/apache/tvm/blob/main/web/src/index.ts#L25, so that tvmjs users can call this function.

web/src/runtime.ts
```diff
@@ -1477,6 +1489,41 @@ export class Instance implements Disposable {
...
+    const arrayentry = list["records"] as Array<NDArrayShardEntry>;
+    for (let i = 0; i < arrayentry.length; ++i){
+      const dataUrl = new URL(arrayentry[i].dataPath, ndarrayCacheUrl).href;
+      try {
+        const result = await(await artifactCache.deleteInCache(dataUrl));
+        if (result === false){
+          return false;
+        }
+      } catch (err){
+        this.env.logger("Error: Cannot fetch " + dataUrl + " err= " + err);
```
CharlieFRuan, line 1520: Should be `Cannot delete` rather than `Cannot fetch`

CharlieFRuan (commented):
The new interface and `deleteInCache()` looks good to me! Thanks for the changes.

I left some comments on `deleteNDArrayCache()`, let me know what you think!

web/src/runtime.ts
DiegoCao: I also prefer the latter one. Will change the function to void

web/src/runtime.ts
DiegoCao: Received, also plan to move the fetchNDArrayInCache to the export

web/src/runtime.ts
```diff
@@ -2432,3 +2444,28 @@ export async function hasNDArrayInCache(
...
 }
+
+/**
+ * Given cacheUrl, search up items to delete based on cacheUrl/ndarray-cache.json
+ *
+ * @param ndarrayCacheUrl
+ * @param cacheScope
+ */
+export async function deleteNDArrayCache(
+  ndarrayCacheUrl: string,
```
CharlieFRuan, line 2455: We should replace `ndarrayCacheUrl` to `cacheUrl` to match the docstring and the semantics.

CharlieFRuan (commented):
Everything else looks good, thank you!

DiegoCao:
All issues are resolved

CharlieFRuan:
Thank you! I think this is ready to be merged cc @tqchen

tqchen (approved):

PR #16542, opened 2024-02-08
[Relax] Support callback as argument

Lunderberg:
Prior to this commit, calls from Relax to external PackedFuncs could only be done through the TVM global registry.  While Relax functions accepting a callback could be written as `callback_arg: R.Callable(arg_struct_info, ret_struct_info)`, attempting to compile these functions would raise an error during the `CodeGenVM` step of `relax.build`.  In addition, the global registry is only queried when initializing the `relax.VirtualMachine`, and so later changes requires restarting the VM.

This commit updates both the `CodeGenVM` lowering pass and the relax VM to support callbacks.  The motivating use case is with the `LazyTransformParams` pass, to improve flexibility by avoiding use of the global registry.

### DIFF
--- a/include/tvm/runtime/relax_vm/bytecode.h
+++ b/include/tvm/runtime/relax_vm/bytecode.h
@@ -58,6 +58,7 @@ enum class Opcode {
   Ret = 2U,
   Goto = 3U,
   If = 4U,
+  CallFromRegister = 5U,
 };
 
 /*! \brief A single virtual machine instruction.
@@ -183,10 +184,15 @@ struct Instruction {
   /*! \brief The instruction opcode. */
   Opcode op;
   union {
-    struct /* Call */ {
+    struct /* Call, CallFromRegister */ {
       /*! \brief The destination register. */
       RegName dst;
-      /*! \brief The index into the packed function table. */
+      /*! \brief The index of the function.
+       *
+       * For `OpCode::Call`, this is an index into the table of static
+       * functions.  For `OpCode::CallFromRegister`, this is an index
+       * of a register.
+       */
       Index func_idx;
       /*! \brief The number of arguments to the packed function. */
       Index num_args;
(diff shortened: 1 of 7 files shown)


Lunderberg:

slyubomirsky (approved):
The change seems straightforward and it's good that there are some test cases too. I'm glad that this case will work now.

include/tvm/runtime/relax_vm/bytecode.h
```diff
@@ -183,10 +184,15 @@ struct Instruction {
...
   Opcode op;
   union {
-    struct /* Call */ {
+    struct /* Call, CallFromRegister */ {
       /*! \brief The destination register. */
       RegName dst;
-      /*! \brief The index into the packed function table. */
+      /*! \brief The index of the function.
+       *
+       * For `OpCode::Call`, this is an index into the table of static
```
tqchen, line 192: we should not use CallFromRegister, instead we should lower to https://github.com/apache/tvm/blob/main/src/runtime/relax_vm/builtin.cc#L373

tqchen:
Sorry for being late on this one. In this particular case, it is better to use  `vm.builtin.invoke_closure` to lower those possible variables that contains the Callable. There are two reasons for this

- One goal of relax VM is to ensure minimal set of instructions, so far we have been keeping the miminum variant via Call and all the rest call variant using builtins like `vm.builtin.invoke_closure`
- We need to handle the case where callback is either a PackedFunc or Closure,  `invoke_closure` handles the case. This enables reuse for compiled cases

@Lunderberg  would be great to followup on this soon given this would change some binary format of the VM https://github.com/apache/tvm/blob/main/src/runtime/relax_vm/builtin.cc#L373

tqchen:
cc @yongwww  as well

Lunderberg:
@tqchen Thank you, and I'll make a follow-up PR.  I had thought that `vm.builtin.invoke_closure` was only for use with internally-generated closure objects.

Lunderberg:
I've submitted https://github.com/apache/tvm/pull/16573, which updates the handling of callback functions, reverts the VM changes, and passes the unit tests that were added in this PR.

slyubomirsky:
Good to know re: invoke_closure

PR #16543, opened 2024-02-08
[TIR] Fix segfaults from ordering of Let/Assert in MakePackedAPI

Lunderberg:
Prior to this commit, the `MakePackedAPI` pass would output steps in the following order:

1. Check the number of arguments.
2. All `LetStmt` produced by the `ArgBinder`
3. `AssertStmt` for the Type code checks for each argument.
4. Additional `AssertStmt` produced by the `ArgBinder`.

This order can cause segfaults if a function was provided incorrect arguments.  For example, an integer argument passed to a function expecting a `DLTensor*` would be dereferenced to find the tensor's data pointer (step (2)) before checking if it is valid to perform that dereference (step (3)).  The same would occur when reading the size of a tensor's axes (step (2)) before checking whether the tensor is the correct dimensionality (step (4)).

This commit updates the steps to the following order.

1. Check the number of arguments.
2. Check the type code of each argument.
3. All `LetStmt` and `AssertStmt` produced by the `ArgBinder`, in the order in which they are generated.

### DIFF
--- a/rust/tvm-graph-rt/tests/test_tvm_basic/build.rs
+++ b/rust/tvm-graph-rt/tests/test_tvm_basic/build.rs
@@ -48,10 +48,6 @@ fn main() -> Result<()> {
         obj_file.exists(),
         "Could not build tvm lib: {}",
         String::from_utf8(output.stderr)?
-            .trim()
-            .split("\n")
-            .last()
-            .unwrap_or("")
     );
 
     let mut builder = Builder::new(File::create(&lib_file)?);
--- a/rust/tvm-graph-rt/tests/test_tvm_basic/build.rs
+++ b/rust/tvm-graph-rt/tests/test_tvm_basic/build.rs
@@ -48,6 +48,10 @@ fn main() -> Result<()> {
         obj_file.exists(),
         "Could not build tvm lib: {}",
         String::from_utf8(output.stderr)?
+            .trim()
+            .split("\n")
+            .last()
+            .unwrap_or("")
     );
 
     let mut builder = Builder::new(File::create(&lib_file)?);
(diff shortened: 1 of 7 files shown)


Lunderberg:
This came about while debugging the implementation of https://github.com/apache/tvm/pull/16542, but is otherwise unrelated.

Lunderberg:

rust/tvm-graph-rt/tests/test_tvm_basic/build.rs
```diff
@@ -48,10 +48,6 @@ fn main() -> Result<()> {
         obj_file.exists(),
         "Could not build tvm lib: {}",
         String::from_utf8(output.stderr)?
-            .trim()
-            .split("\n")
-            .last()
-            .unwrap_or("")
```
slyubomirsky, line 54: Wonder how this came up. Just for readability?

src/tir/transforms/arg_binder.cc
```diff
@@ -186,18 +191,8 @@ void ArgBinder::BindDLTensor(const Buffer& buffer, const PrimExpr& device_type,
   if (!(buffer->dtype == DataType::Int(1) || buffer->dtype == DataType::Int(4) ||
         buffer->dtype == DataType::UInt(4))) {
     auto type_msg = tvm::tir::StringImm(type_err_msg.str());
-    asserts_.emplace_back(AssertStmt(a_ndim == v_ndim, msg, nop));
```
slyubomirsky, line 189: Was this just a duplicate?

slyubomirsky (approved):
Definitely good to have clean errors instead of segfaults and more validation. I'm also glad to see tests for these error cases.

rust/tvm-graph-rt/tests/test_tvm_basic/build.rs
Lunderberg: Oh, that's really weird.  I'm guessing it was from bouncing over to the PR branch of https://github.com/apache/tvm/pull/16183, which touched a number of the FFI bindings.  I've removed this delta from the PR.

src/tir/transforms/arg_binder.cc
Lunderberg: Yup.  The buffer's dimensionality is checked earlier, so this is entirely a duplicate check on the dimensionality.

Lunderberg:
Last CI failure is due to a flaky hexagon test.  I've added a `@pytest.mark.skip` to this PR, now just waiting on CI.

PR #16544, opened 2024-02-08
[TIR] Expand debug symbol output for CodeGenLLVM

Lunderberg:
Prior to this commit, the `CodeGenLLVM` would include DWARF symbols specifying the function signature of each TIR function.  This commit expands the information exposed in the debug symbols.

* Name functions based on the name of the corresponding TIR function. This is taken either from the `attr::kGlobalSymbol` if present, or the PrimFunc's `GlobalVar` otherwise.

* Name function parameter based on their TIR variable name.

* Annotate the name, the type signature, and parameter names for the private function produced by the `tir::attr::compute_scope` attribute.

* Name local variables based on the name and type of the corresponding TIR variable.


Lunderberg:
This came about while debugging the implementation of https://github.com/apache/tvm/pull/16542, but is otherwise unrelated.

Lunderberg:

slyubomirsky (approved):
I'm all for propagating more debug information and I don't see anything to take issue with here.

PR #16545, opened 2024-02-08
[Runtime] Fix dtype conversion for bf16 and fp8

vinx13:
cc @tqchen


tqchen (approved):

tqchen:
@vinx13 we should disable the rust e2e test that depends on mxnet by simply removing it, as mxnet is getting phased out

tqchen:
https://github.com/apache/tvm/issues/16547

PR #16546, opened 2024-02-08
[CI] Disable testing that downloads from mxnet

Lunderberg:
The mxnet project was archived in 2023-09 ([link](https://attic.apache.org/projects/mxnet.html)). Downloads from the mxnet AWS
links (e.g. `https://apache-mxnet.s3-accelerate.dualstack.amazonaws.com/gluon/models/resnet18_v1-a0666292.zip`) no longer can be performed.  This commit disables TVM CI steps that depend on downloading models from these locations.

### DIFF
--- a/gallery/how_to/compile_models/from_mxnet.py
+++ b/gallery/how_to/compile_models/from_mxnet.py
@@ -37,6 +37,7 @@ https://mxnet.apache.org/versions/master/install/index.html
 # sphinx_gallery_start_ignore
 # sphinx_gallery_requires_cuda = True
 # sphinx_gallery_end_ignore
+import sys
 import mxnet as mx
 import tvm
 import tvm.relay as relay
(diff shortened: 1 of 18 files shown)


tqchen:
mxnet is getting deprecated, so we should remove the testcases/example that depends on it

tqchen:
opened a related tracking issue https://github.com/apache/tvm/issues/16547

Lunderberg:
That was my guess, and agreed that it doesn't make sense to maintain integration with a deprecated framework.

Marking this PR as ready for review, so that it can unblock CI.  It still attempts to download files, but should bail out on error rather than failing CI altogether.  A follow-up PR can then remove the test cases altogether.

yongwww (approved):
good to have it to unblock the CI.

Lunderberg:
Looks like CI is passing, except for the ARM tests, which are stalled waiting for `ARM-GRAVITON3` instances.  1/8 has run since I checked this morning, so hopefully they'll get queued up sooner rather than later.

yongwww:
arm tests are stuck https://github.com/apache/tvm/issues/16549

Lunderberg:
Thank you for mentioning that.  Looks like the ARM CI was related to another of my PRs, which I recently rebased onto main.  I've unstuck the ARM tests, and CI has now passed here.

PR #16554, opened 2024-02-11
[Web] Compatibility with PagedKVCache in WebGPU

CharlieFRuan:
This PR introduces various WebGPU changes to accommodate the new `PagedKVCache` interface. All changes below are essential for making models that use PagedKVCache runnable under WebGPU:

- Require exactly same-dtype matching for WebGPU smem reuse in `storage_rewrite.cc`
- Rename `AttentionKVCache` to `AttentionKVCacheLegacy` for the old KVcache interface in `lm_support.cc`; include `paged_kv_cache.cc` when making `wasm_runtime` subsequently
- In WebGPU codegen:
  - Declare local variables within the function scope rather than the module scope
  - Generate `while (true)` rather than `while (1)`
- Require 10 `maxStorageBuffersPerShaderStage` rather than the default 8 from the WebGPU device when initializing runtime; this is required for new kernels introduced in PagedKVCache
- In `deviceCopyToCPU()`, when raw bytes to write are not multiples of 4, we pad them, as required by WebGPU's `writeBuffer()`.

Co-authored-by: Rick Zhou <rickzhoucmu@gmail.com>

### DIFF
--- a/src/runtime/relax_vm/lm_support.cc
+++ b/src/runtime/relax_vm/lm_support.cc
@@ -226,18 +226,19 @@ class AttentionKVCacheObj : public Object {
   }
 
   static constexpr const uint32_t _type_index = TypeIndex::kDynamic;
-  static constexpr const char* _type_key = "relax.vm.AttentionKVCache";
+  static constexpr const char* _type_key = "relax.vm.AttentionKVCacheLegacy";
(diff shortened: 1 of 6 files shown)


MasterJH5574 (approved):
LGTM. Thank you @CharlieFRuan and @rickzx for carrying through!

PR #16562, opened 2024-02-13
[Bugfix][TVMScript] Handle R.match_cast as last binding in if/else

Lunderberg:
Prior to this commit, using `R.match_cast` as the last binding would produce a segfault, as `var_binding->value` was used instead of `match_cast->value`.  In addition, because the last binding of each branch was removed, any changes to the struct info resulting from the match cast were silently discarded.

This commit updates the TVMScript parsing of if/else statements to remove the segfault and maintain the struct info changes produced by the `R.match_cast`.

### DIFF
--- a/src/script/ir_builder/relax/frame.cc
+++ b/src/script/ir_builder/relax/frame.cc
@@ -263,7 +263,9 @@ void ElseFrameNode::ExitWithScope() {
   IfFrame frame = FindIfFrame("R.Else");
   frame->else_expr = output;
   CHECK(frame->var_name == var_name)
-      << "This last binding of both branches must have the same variable.";
+      << "This last binding of both branches must provide the same variable.  "
+      << "However, the R.Then branch provides variable " << frame->var_name
+      << ", while the R.Else branch provides variable " << var_name;
 }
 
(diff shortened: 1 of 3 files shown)


Lunderberg:
> I'm surprised that the previous logic failed.

Yeah, I was pretty surprised at it, too.  I think the core issue is the mismatch between the internal representation of a `relax::If` node and the python-ish TVMScript that represents it.  In order to bridge that mismatch, something needs to either be duplicated, or hoisted outside the `relax::If`, and that hoisting is where the error crept in.

(Because it introduces a new scope, I think the body of a `relax::If` is closer to a new python function than to a python if/else block.  Hypothetically, we could represent it with a new function in TVMScript, but that representation would be ambiguous with a local `relax::Function` definition.)

I suspect there's some similar edge cases that could be hunted down within a similar space.  For example, a `relax::MatchCast` whose bound value is `relax::If`, or a `relax::If` whose output value is not the last binding of the branch.  I suspect that those cases would most the `relax::MatchCast` inside the `relax::If`, or introduce a trivial binding, but I haven't tested it.  At least, those differences should be semantically equivalent, even if they have slightly different IR representations.

PR #16563, opened 2024-02-13
[Relax] Implement operators to read runtime DLTensor* information

Lunderberg:
Relax is capable of expressing tensors whose element type is unknown.  However, these must typically be replaced with a known dtype prior to compilation, as most operators require known data types prior to legalization.  This can be done by using a `relax::MatchCast` node, such as accepting a parameter `arg: R.Tensor([16,16])`, then defining the dtype using `R.match_cast(arg, R.Tensor([16,16],'float16'))`.

However, using a `R.match_cast` node requires knowing which data type should be used in the new `R.Tensor`, and raises an error for an incorrect data type.  If an argument may be one of two distinct data types, `R.match_cast` cannot be used to check which data type is in use.

This commit adds Relax operators to read the runtime values of a `DLTensor*` argument.  These can be be used to normalize arguments prior to a compute step.  For example, pre-processing a model weight that may be provided in either `float16` or `bfloat16` format.

### DIFF
--- a/python/tvm/relax/expr.py
+++ b/python/tvm/relax/expr.py
@@ -244,6 +244,192 @@ class ExprWithOp(Expr, Scriptable):
                 raise IndexError from err
             raise
 
+    def _check_for_tensor_struct_info(self):
+        """Raise an error if this is something other than a Tensor
+
+        Used for early checks in `expr.dtype` and `expr.shape`
+        accessors.  While invalid usage would cause errors to be
+        raised durin shape inference, an earlier check makes it easier
+        to find the invalid usage.
+        """
+        if self.struct_info_ is None:
+            return
+
+        if not isinstance(self.struct_info_, tvm.relax.TensorStructInfo):
+            raise TypeError(
+                f"Runtime unpacking of DLDataType is only implemented for tensors, "
+                f"but was applied to object {self} of type {type(self)}."
+            )
+
+    @property
+    def dtype(self) -> "_DLTensorDTypeProxy":
+        """Returns a proxy object for accessing DLTensor::dtype"""
+        self._check_for_tensor_struct_info()
+        return _DLTensorDTypeProxy(self)
+
+    @property
+    def ndim(self) -> "Expr":
+        """Returns the runtime value of DLTensor::ndim"""
+        self._check_for_tensor_struct_info()
+        op = tvm.ir.Op.get("relax.tensor_ndim")
+        return tvm.relax.Call(op, [self])
+
+    @property
+    def shape(self) -> "_DLTensorShapeProxy":
+        """Returns a proxy object for accessing DLTensor::shape"""
+        self._check_for_tensor_struct_info()
+        return _DLTensorShapeProxy(self)
+
+
+class _DLTensorDTypeProxy(tvm.runtime.ObjectGeneric):
+    """A proxy object for unpacking DLDatatype from DLTensor
+
+    Exposes accessors for `DLDataType` fields `type_code`, `lanes`,
+    and `bits` within a `DLTensor::dtype`.  Accessing these fields
+    will produce `relax.Call` expressions, representing the field's
+    runtime value.  If the datatype of the tensor is known at
+    compile-time, the `relax.Call` will be normalized into a
+    `relax.PrimValue`, with no runtime cost.
+
+    Parameters
+    ----------
+    tensor: relax.Expr
+
+        The relax tensor (or a variable referring to a relax tensor),
+        whose runtime shape is being inspected.
+
+    """
+
+    def __init__(self, tensor):
+        self.tensor = tensor
+
+    def asobject(self):
+        """Provide expected in error message
+
+        This method is called when `_DLTensorDTypeProxy` is used in a
+        context that requires a `relax.Expr`.  This usage is not
+        supported, and raising an error here can provide suggested
+        fixes that are not present in the default error message from
+        `tvm.runtime.convert_to_object`.
+        """
+
+        fields = [f"{self.tensor}.dtype.{field}" for field in ["type_code", "bits", "lanes"]]
+        raise TypeError(
+            f"{self.tensor}.dtype cannot be converted to a relax expression, "
+            f"and should be used as a proxy object to access "
+            f"fields {fields}"
+        )
+
+    @property
+    def type_code(self) -> Expr:
+        """Accessor for the DLDataType::bits field
+
+        Returns
+        -------
+        type_code: Expr
+
+            The type code of the DLTensor.  See the `DLDeviceType`
+            enum in `dlpack.h` for more information.
+        """
+        op = tvm.ir.Op.get("relax.tensor_dtype_code")
+        return tvm.relax.Call(op, [self.tensor])
+
+    @property
+    def lanes(self) -> Expr:
+        """Accessor for the DLDataType::bits field
+
+        Returns
+        -------
+        lanes: Expr
+
+            The number of lanes in the DLDataType
+        """
+        op = tvm.ir.Op.get("relax.tensor_dtype_lanes")
+        return tvm.relax.Call(op, [self.tensor])
+
+    @property
+    def bits(self) -> Expr:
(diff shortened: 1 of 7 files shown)


python/tvm/relax/expr.py
```diff
@@ -244,6 +244,192 @@ def __getitem__(self, index: int) -> "ExprWithOp":
                 raise IndexError from err
             raise
 
+    def _check_for_tensor_struct_info(self):
+        """Raise an error if this is something other than a Tensor
+
+        Used for early checks in `expr.dtype` and `expr.shape`
+        accessors.  While invalid usage would cause errors to be
+        raised durin shape inference, an earlier check makes it easier
```
slyubomirsky, line 252:
```suggestion
        raised during shape inference, an earlier check makes it easier
```
typo

python/tvm/relax/expr.py
```diff
@@ -244,6 +244,192 @@ def __getitem__(self, index: int) -> "ExprWithOp":
...
+        """Returns a proxy object for accessing DLTensor::shape"""
+        self._check_for_tensor_struct_info()
+        return _DLTensorShapeProxy(self)
+
+
+class _DLTensorDTypeProxy(tvm.runtime.ObjectGeneric):
+    """A proxy object for unpacking DLDatatype from DLTensor
+
+    Exposes accessors for `DLDataType` fields `type_code`, `lanes`,
+    and `bits` within a `DLTensor::dtype`.  Accessing these fields
```
slyubomirsky, line 288: These are good to have. Offset might also be useful to add, as it might help for memory reuse.

src/relax/op/tensor/unpack.cc
```diff
@@ -0,0 +1,349 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/*!
+ * \file unpack.cc
```
slyubomirsky, line 21: I wonder if "unpack" is the best name. Perhaps "accessors" or "runtime_accessors" could be a little more descriptive? I'm not sure.

src/relax/transform/legalize_ops.cc
```diff
@@ -157,17 +167,72 @@ class LegalizeMutator : public ExprMutator {
...
+        // For data-dependent ops, the output shape will be identified
+        // at runtime.  The Legalizer will insert their shape
+        // functions, which are manually registered for each
+        // data-dependent op, and match cast to define symbolic output
+        // shapes.  These symbolic output shapes at compile time can
+        // be by later operations to refer to the runtime shape.
+        //
+        // TODO(Lunderberg): Make a new operator attribute
+        // `.set_attr<Bool>("DataDependent")`, rather than relying on
+        // the name of the operator.
```
slyubomirsky, line 227: I agree with this, probably should be a separate PR.

src/relax/transform/legalize_ops.cc
```diff
@@ -157,17 +167,72 @@ class LegalizeMutator : public ExprMutator {
...
+        //     for `R.Tensor(shape=[m, n])`, where `m` and `n` are new
+        //     shape variables.  This would allow legalization into
+        //     dynamic TIR PrimFuncs.
+        //
+        //     This fallback would only be applicable for cases where
+        //     both the dtype and the dimensionality are known.  While
+        //     Relax can express a tensor with unknown dtype and
+        //     dimensionality as `TensorStructInfo(DataType::Void(),
+        //     kUnknownNDim)`, TIR cannot express unknown dtype or
+        //     unknown dimensionality.
```
slyubomirsky, line 204: Interesting idea. This could be done by inserting a `MatchCast` that introduces the new vars. Perhaps this should be filed as an issue rather than made a long comment.

python/tvm/relax/expr.py
```diff
@@ -244,6 +244,192 @@ def __getitem__(self, index: int) -> "ExprWithOp":
...
+            return
+
+        if not isinstance(self.struct_info_, tvm.relax.TensorStructInfo):
+            raise TypeError(
+                f"Runtime unpacking of DLDataType is only implemented for tensors, "
+                f"but was applied to object {self} of type {type(self)}."
+            )
+
+    @property
+    def dtype(self) -> "_DLTensorDTypeProxy":
```
slyubomirsky, line 265: Why does the return type have to be in quotes? I assume it has to do with the property decorator.

slyubomirsky (approved):
The implementation seems solid, a very good change. I like the parameterized test cases too. The use of the _DLTensorShapeProxy resulted in an elegant UI in tvmscript.

Idle musing: I wonder if there's any way the PrimFuncs can be unrolled in cases where the parameters are known at compile time.

src/relax/op/tensor/unpack.cc
tqchen: how about relax.inspect(as a sub namespace for the ops)

Lunderberg:
> The implementation seems solid, a very good change. I like the parameterized test cases too. The use of the _DLTensorShapeProxy resulted in an elegant UI in tvmscript.

Thank you!  I really like polishing up the interface to be as clean as possible.  (For my own sake if nothing else, as I am liable to forget a builtin-function name, but am much less likely to forget `obj.dtype`.)

> Idle musing: I wonder if there's any way the PrimFuncs can be unrolled in cases where the parameters are known at compile time.

Prior to `LegalizeOps`, this is implicitly done by the `FNormalize` implementation for each operator.  After `LegalizeOps`, not so much.  While the `FoldConstant` pass can compile/run TIR functions if their arguments are known, there isn't a good way to indicate that a TIR function only requires the DLTensor struct, and not the data itself.

python/tvm/relax/expr.py
Lunderberg: More the order of definitions in the file.   Function annotations are evaluated when the class is being defined.  Since the `_DLTensorDTypeProxy` class is defined lower in the file, the type annotations are provided as a string, rather than as a class object.

src/relax/transform/legalize_ops.cc
Lunderberg: Good call, though I haven't had time to flesh out the idea yet.  This would be part of a general cleanup I'd propose for the `StructInfo` interactions:

* Remove the `Optional<Expr> shape` and `int ndim` in `TensorStructInfo`.  Instead, have `Optional<ShapeStructInfo>`.
* Remove the `ndim` field from `ShapeStructInfo`.  Instead, the `ndim` is unknown when  `Optional<Array<PrimExpr>>` is `NullOpt`.
* If the dimensionality of a `ShapeStructInfo` is known, every dimension must have an associated `PrimExpr`.  The constructor that accepts `ndim` initializes fresh TIR variables to represent the unknown size.

src/relax/transform/legalize_ops.cc
Lunderberg: Agreed.  The note is more so that it has a record somewhere, but it would be enough of a change that it should be part of a separate PR.  As it was, I wanted to make as few changes to `LegalizeOps` as possible in this PR.

python/tvm/relax/expr.py
Lunderberg: Good point.  At the moment, I stuck with the values that have a direct presence elsewhere in Relax, but I agree that it would be good to be able to extract any of the `DLTensor*` fields.

src/relax/op/tensor/unpack.cc
Lunderberg: I like `relax.inspect` quite a bit, and will update the tensors to that namespace.

python/tvm/relax/expr.py
Lunderberg: Thank you, and fixed

Lunderberg:
All CI tests passing, and thank you for the review @slyubomirsky !  I'll follow up with another PR to add inspection of the remainder of the `DLTensor*` fields.

PR #16565, opened 2024-02-13
[Bugfix][Transform] Keep private non-primitive functions in FuseTIR

Lunderberg:
Prior to this commit, private non-primitive relax functions would be discarded by `FuseTIR`.  If any calls to these functions exist, the resulting `IRModule` would be ill-formed.  This commit updates `FuseTIR` so that it only applies updates to functions with `attr::kPrimitive`, and calls into those functions.

To retain backwards compatibility, `DeadCodeElimination` is applied as a post-processing step.


tqchen:
cc @Hzfengsy

Hzfengsy (approved):

PR #16568, opened 2024-02-14
[Relax][Runtime] RNNState for Space State Models

Hzfengsy:
This commit adds the RNNState class to the Relax VM, similar to the PagedKVCache, for space state models like RWKV and mamba

cc @tqchen @MasterJH5574


Hzfengsy:
Refactored. Now AttentionKVCache and RNNState have the same base class. would be great if @MasterJH5574 @tqchen Can take a look

tqchen (approved):

PR #9341, opened 2021-10-21
[CUDA] Support memory reuse for dynamic shared memory

jinhongyii:
This PR applies part of the algorithm of storage rewrite to detect memory reuse possibility for dynamic shared memory, and does rewrite to the body. This functionality is important in matmul because we may want to allocate the shared memory for data fetch and the shared memory for write back in the same location.

PR #16569, opened 2024-02-14
[Unity][Parser] Check well-formedness in the parser

slyubomirsky:
As discussed in several TVM Open Development meetings, we were permitting invalid programs to be parsed by not checking well-formedness during parsing. This PR makes a small change to the parser to check well-formedness in both Relax and TIR proactively, though this required correcting bugs in tests. Those bugs should not have been there in the first place!

I had held off on making this PR until [this PR on MLC-LLM](https://github.com/mlc-ai/mlc-llm/pull/1696), which ensured that there would not be invalid constructs in models constructed by MLC-LLM.

One issue is that the well-formed checker does not create the most readable error messages. If the normalizer's error reporting could also be used in the well-formed checker, that would be a useful change to make as well.

### DIFF
--- a/python/tvm/script/parser/core/entry.py
+++ b/python/tvm/script/parser/core/entry.py
@@ -89,13 +89,13 @@ def parse(
             parser.report_error(err.node, err.args[0])
     ret = builder.get()
     # check well-formedness in both Relax and TIR
-    if (
-        check_well_formed
-        and isinstance(ret, IRModule)
-        and not (relax_well_formed(ret) and tir_well_formed(ret, assert_mode=False))
-    ):
-        parser.report_error(
-            source.as_ast(),
-            err="Program containing Relax functions is not well-formed",
-        )
+    if check_well_formed:
+        check_ret = ret
+        if not isinstance(check_ret, IRModule):
+            check_ret = IRModule.from_expr(ret)
+        if not relax_well_formed(check_ret) or not tir_well_formed(check_ret, assert_mode=False):
+            parser.report_error(
+                source.as_ast(),
+                err="Program is not well-formed",
+            )
     return ret
--- a/python/tvm/script/parser/core/entry.py
+++ b/python/tvm/script/parser/core/entry.py
@@ -95,10 +95,10 @@ def parse(
     ret = builder.get()
     # check well-formedness in both Relax and TIR
     if check_well_formed:
-        # do the imports here to avoid a circular import at the start
-        # (since importing Relax will import a dependenc on the parser)
-        from ....relax.analysis import well_formed as relax_well_formed
-        from ....tir.analysis import verify_well_formed as tir_well_formed
+        # (C0415 = import-outside-toplevel. It is necessary here to avoid a circular dependency,
+        # since importing Relax imports a dependency on the parser)
+        from ....relax.analysis import well_formed as relax_well_formed # pylint: disable=C0415
+        from ....tir.analysis import verify_well_formed as tir_well_formed # pylint: disable=C0415
 
         check_ret = ret
         if not isinstance(check_ret, IRModule):
@@ -108,7 +108,7 @@ def parse(
             parser.report_error(source_ast, err=WELL_FORMED_ERROR_MESSAGE)
         try:
             tir_well_formed(check_ret)
-        except Exception as err:
+        except Exception as err: # pylint: disable=broad-exception-caught
             parser.report_error(
                 source_ast,
                 err=f"{WELL_FORMED_ERROR_MESSAGE}\n\nTraceback: {str(err)}",
--- a/python/tvm/script/parser/core/entry.py
+++ b/python/tvm/script/parser/core/entry.py
@@ -18,7 +18,8 @@
 import inspect
 from typing import Any, Dict, Union
 
-from ....relax.analysis import well_formed
+from ....relax.analysis import well_formed as relax_well_formed
+from ....tir.analysis import verify_well_formed as tir_well_formed
 from ....ir.module import IRModule
 from ...ir_builder import IRBuilder
 from . import doc
@@ -87,8 +88,12 @@ def parse(
         except ParserError as err:
             parser.report_error(err.node, err.args[0])
     ret = builder.get()
-    # well-formedness check will ignore any non-Relax functions
-    if check_well_formed and isinstance(ret, IRModule) and not well_formed(ret):
+    # check well-formedness in both Relax and TIR
+    if (
+        check_well_formed
+        and isinstance(ret, IRModule)
+        and not (relax_well_formed(ret) and tir_well_formed(ret, assert_mode=False))
+    ):
         parser.report_error(
             source.as_ast(),
             err="Program containing Relax functions is not well-formed",
--- a/python/tvm/script/parser/core/entry.py
+++ b/python/tvm/script/parser/core/entry.py
@@ -45,23 +45,11 @@ def scan_macro(program: Union[Any, str], extra_vars: Dict[str, Any] = None) -> A
     return source, closure_vars
 
 
-def find_decorator_annotation(node: doc.Module, annotation: str, default: bool = True) -> bool:
-    """
-    Check the value of given annotation (argument name) in the function decorator.
-    Returns the value of the annotation if present, otherwise giving the default value.
-    """
-    # Note: A Module body is always a list containing a single ClassDef.
-    # The ClassDef has a decorator list
-    for dec in node.body[0].decorator_list:
-        if not isinstance(dec, doc.Call) or dec.func.attr != "ir_module":
-            continue
-        for keyword in dec.keywords:
-            if keyword.arg == annotation:
-                return keyword.value.value
-    return default
-
-
-def parse(program: Union[doc.AST, Any, str], extra_vars: Dict[str, Any] = None) -> Any:
+def parse(
+    program: Union[doc.AST, Any, str],
+    extra_vars: Dict[str, Any] = None,
+    check_well_formed: bool = True,
+) -> Any:
     """Register a method for a operand type, AST operator node and operand index.
 
     Parameters
@@ -71,6 +59,9 @@ def parse(program: Union[doc.AST, Any, str], extra_vars: Dict[str, Any] = None) 
 
     extra_vars : Dict[str, Any]
         The extra variable table for parsing.
+
+    check_well_formed : bool
+        Whether to check well-formedness after parsing.
 
     Returns
     -------
@@ -97,13 +88,9 @@ def parse(program: Union[doc.AST, Any, str], extra_vars: Dict[str, Any] = None) 
             parser.report_error(err.node, err.args[0])
     ret = builder.get()
     # well-formedness check will ignore any non-Relax functions
-    if isinstance(ret, IRModule):
-        # note: use the walrus operator (:=) once the project Python version
-        # supports it, would be more concise
-        source_ast = source.as_ast()
-        if find_decorator_annotation(source_ast, "check_well_formed") and not well_formed(ret):
-            parser.report_error(
-                source_ast,
-                err="Program containing Relax functions is not well-formed",
-            )
+    if check_well_formed and isinstance(ret, IRModule) and not well_formed(ret):
+        parser.report_error(
+            source.as_ast(),
+            err="Program containing Relax functions is not well-formed",
+        )
     return ret
--- a/python/tvm/script/parser/core/entry.py
+++ b/python/tvm/script/parser/core/entry.py
@@ -18,8 +18,6 @@
 import inspect
 from typing import Any, Dict, Union
 
-from ....relax.analysis import well_formed as relax_well_formed
-from ....tir.analysis import verify_well_formed as tir_well_formed
 from ....ir.module import IRModule
 from ...ir_builder import IRBuilder
 from . import doc
@@ -97,6 +95,11 @@ def parse(
     ret = builder.get()
     # check well-formedness in both Relax and TIR
     if check_well_formed:
+        # do the imports here to avoid a circular import at the start
+        # (since importing Relax will import a dependenc on the parser)
+        from ....relax.analysis import well_formed as relax_well_formed
+        from ....tir.analysis import verify_well_formed as tir_well_formed
+
         check_ret = ret
         if not isinstance(check_ret, IRModule):
             check_ret = IRModule.from_expr(ret)
--- a/python/tvm/script/parser/core/entry.py
+++ b/python/tvm/script/parser/core/entry.py
@@ -93,9 +93,11 @@ def parse(
         check_ret = ret
         if not isinstance(check_ret, IRModule):
             check_ret = IRModule.from_expr(ret)
-        if not relax_well_formed(check_ret) or not tir_well_formed(check_ret, assert_mode=False):
-            parser.report_error(
-                source.as_ast(),
-                err="Program is not well-formed",
-            )
+        source_ast = source.as_ast()
+        if not relax_well_formed(check_ret):
+            parser.report_error(source_ast, err="Program is not well-formed")
+        try:
+            tir_well_formed(check_ret)
+        except Exception as err:
+            parser.report_error(source_ast, err=err)
     return ret
--- a/python/tvm/script/parser/core/entry.py
+++ b/python/tvm/script/parser/core/entry.py
@@ -27,6 +27,13 @@ from .diagnostics import Source
 from .error import ParserError
 from .parser import Parser
 
+WELL_FORMED_ERROR_MESSAGE = (
+    "Program is not well-formed. If this is deliberate, consider "
+    "setting check_well_formed in the top-level decorator to False "
+    "(e.g., @I.ir_module(check_well_formed=False) or "
+    "@R.function(check_well_formed=False))."
+)
+
 
 def _default_globals() -> Dict[str, Any]:
     import tvm  # pylint: disable=import-outside-toplevel
@@ -95,9 +102,12 @@ def parse(
             check_ret = IRModule.from_expr(ret)
         source_ast = source.as_ast()
         if not relax_well_formed(check_ret):
-            parser.report_error(source_ast, err="Program is not well-formed")
+            parser.report_error(source_ast, err=WELL_FORMED_ERROR_MESSAGE)
         try:
             tir_well_formed(check_ret)
         except Exception as err:
-            parser.report_error(source_ast, err=err)
+            parser.report_error(
+                source_ast,
+                err=f"{WELL_FORMED_ERROR_MESSAGE}\n\nTraceback: {str(err)}",
+            )
     return ret
--- a/python/tvm/script/parser/core/entry.py
+++ b/python/tvm/script/parser/core/entry.py
@@ -97,8 +97,8 @@ def parse(
     if check_well_formed:
         # (C0415 = import-outside-toplevel. It is necessary here to avoid a circular dependency,
         # since importing Relax imports a dependency on the parser)
-        from ....relax.analysis import well_formed as relax_well_formed # pylint: disable=C0415
-        from ....tir.analysis import verify_well_formed as tir_well_formed # pylint: disable=C0415
+        from ....relax.analysis import well_formed as relax_well_formed  # pylint: disable=C0415
+        from ....tir.analysis import verify_well_formed as tir_well_formed  # pylint: disable=C0415
 
         check_ret = ret
         if not isinstance(check_ret, IRModule):
@@ -108,7 +108,7 @@ def parse(
             parser.report_error(source_ast, err=WELL_FORMED_ERROR_MESSAGE)
         try:
             tir_well_formed(check_ret)
-        except Exception as err: # pylint: disable=broad-exception-caught
+        except Exception as err:  # pylint: disable=broad-exception-caught
             parser.report_error(
                 source_ast,
                 err=f"{WELL_FORMED_ERROR_MESSAGE}\n\nTraceback: {str(err)}",
--- a/python/tvm/script/parser/core/entry.py
+++ b/python/tvm/script/parser/core/entry.py
@@ -18,6 +18,8 @@
 import inspect
 from typing import Any, Dict, Union
 
+from ....relax.analysis import well_formed
+from ....ir.module import IRModule
 from ...ir_builder import IRBuilder
 from . import doc
 from .diagnostics import Source
@@ -41,6 +43,22 @@ def scan_macro(program: Union[Any, str], extra_vars: Dict[str, Any] = None) -> A
     source = Source(program)
     closure_vars = extra_vars or _default_globals()
     return source, closure_vars
+
+
+def find_decorator_annotation(node: doc.Module, annotation: str, default: bool = True) -> bool:
+    """
+    Check the value of given annotation (argument name) in the function decorator.
+    Returns the value of the annotation if present, otherwise giving the default value.
+    """
+    # Note: A Module body is always a list containing a single ClassDef.
+    # The ClassDef has a decorator list
+    for dec in node.body[0].decorator_list:
+        if not isinstance(dec, doc.Call) or dec.func.attr != "ir_module":
+            continue
+        for keyword in dec.keywords:
+            if keyword.arg == annotation:
+                return keyword.value.value
+    return default
 
 
 def parse(program: Union[doc.AST, Any, str], extra_vars: Dict[str, Any] = None) -> Any:
@@ -77,4 +95,15 @@ def parse(program: Union[doc.AST, Any, str], extra_vars: Dict[str, Any] = None) 
             parser.parse(extra_vars=extra_vars)
         except ParserError as err:
             parser.report_error(err.node, err.args[0])
-    return builder.get()
+    ret = builder.get()
+    # well-formedness check will ignore any non-Relax functions
+    if isinstance(ret, IRModule):
+        # note: use the walrus operator (:=) once the project Python version
+        # supports it, would be more concise
+        source_ast = source.as_ast()
+        if find_decorator_annotation(source_ast, "check_well_formed") and not well_formed(ret):
+            parser.report_error(
+                source_ast,
+                err="Program containing Relax functions is not well-formed",
+            )
+    return ret
--- a/src/driver/driver_api.cc
+++ b/src/driver/driver_api.cc
@@ -590,6 +590,7 @@ transform::Sequential MixedModulePassManager(IRModule mixed_mod, Target target) 
 
   mixed_pass_list.push_back(tir::transform::ThreadSync("shared"));
   mixed_pass_list.push_back(tir::transform::ThreadSync("shared.dyn"));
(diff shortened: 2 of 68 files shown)


tests/python/relax/test_transform_operator_specific_normalization.py
```diff
@@ -261,9 +303,9 @@ def multiply_by_two(A: T.Buffer(16, "float32"), B: T.Buffer(16, "float32")):
 def test_normalize_to_inline_tuple_for_call_tir_inplace(custom_op):
     """FNormalize in-lines the argument tuple for R.call_tir_inplace"""
 
-    # The CallTIRInplaceAttrs cannot be constructed from the Python
-    # API.  Therefore, declaring the Expected output first, so that
-    # the attributes can be used for the non-normalized Before.
```
slyubomirsky, line 266: I am pretty sure it is possible to construct attrs in Python (I've done it for Relay) but it's very tedious, so I corrected the comment. It is definitely easier to rely on the one that will be constructed on the C++ side.

python/tvm/script/parser/core/entry.py
```diff
@@ -77,4 +95,15 @@ def parse(program: Union[doc.AST, Any, str], extra_vars: Dict[str, Any] = None)
...
         except ParserError as err:
             parser.report_error(err.node, err.args[0])
-    return builder.get()
+    ret = builder.get()
+    # well-formedness check will ignore any non-Relax functions
+    if isinstance(ret, IRModule):
+        # note: use the walrus operator (:=) once the project Python version
+        # supports it, would be more concise
+        source_ast = source.as_ast()
+        if find_decorator_annotation(source_ast, "check_well_formed") and not well_formed(ret):
```
Lunderberg, line 104: Can we add a unit test validated that `@I.ir_module(check_well_formed=False)` disables the well-formed check?  This is the only location where I see `check_well_formed` in this PR.

python/tvm/script/parser/core/entry.py
```diff
@@ -77,4 +95,15 @@ def parse(program: Union[doc.AST, Any, str], extra_vars: Dict[str, Any] = None)
             parser.parse(extra_vars=extra_vars)
         except ParserError as err:
             parser.report_error(err.node, err.args[0])
-    return builder.get()
+    ret = builder.get()
+    # well-formedness check will ignore any non-Relax functions
+    if isinstance(ret, IRModule):
```
Lunderberg, line 100: Can this be applied to a single relax function as well?  If the result is something other than an `IRModule`, we could construct `IRModule.from_expr(ret)`, and apply the well-formed check there.

python/tvm/script/parser/core/entry.py
```diff
@@ -77,4 +95,15 @@ def parse(program: Union[doc.AST, Any, str], extra_vars: Dict[str, Any] = None)
             parser.parse(extra_vars=extra_vars)
         except ParserError as err:
             parser.report_error(err.node, err.args[0])
-    return builder.get()
+    ret = builder.get()
+    # well-formedness check will ignore any non-Relax functions
```
Lunderberg, line 99: Instead of ignoring the non-Relax functions, can we run the TIR equivalent `tvm.tir.analysis.verify_well_formed` function?

tests/python/relax/test_transform_normalize_global_var.py
```diff
@@ -15,31 +15,31 @@
...
-        def f():
-            return R.const(1, "int32")
-
-        @R.function
-        def f1():
-            R.func_attr({"global_symbol": "f"})
-            cls = Before
-            gv: R.Tensor((), dtype="int32") = cls.f()
-            return gv
+    # parser will check well-formedness so we can't use it to construct this example
```
Lunderberg, line 30: Instead of changing to the `relax.BlockBuilder`, can we use the `check_well_formed=False` argument?

tests/python/relax/test_transform_operator_specific_normalization.py
Lunderberg: Good point.  It's definitely more tedious than I'd like, and (I think) requires going through the `make_node` API rather than having a python constructor.

Lunderberg (changes requested):
I like the PR, and definitely good to have the early validation.  I think the main changes

1. Verifying that the TIR is well-formed as well.
2. Unit test for the `check_well_formed=False` attribute
3. Using the `check_well_formed=False` attribute instead of `relax.BlockBuilder` for unit tests that require ill-formed inputs.  This would be both for the improved readability of TVMScript, and because we may want to enable well-formed checks for some uses of the `relax.BlockBuilder` as well.

python/tvm/script/parser/core/entry.py
Lunderberg: Long-term, I think it would be good to have a single validation pass, rather than the IR-specific passes, because an `IRModule` may contain both TIR and Relax functions.

python/tvm/script/parser/core/entry.py
slyubomirsky: Oh I thought I had gotten rid of it. It doesn't actually work because the parser, for whatever reason, does not save a decorator list for the IRModule and I don't know how to change it so that it does. I would prefer having it, but the present parser can't save that information.

python/tvm/script/parser/core/entry.py
slyubomirsky: Sure, we could do it that way.

tests/python/relax/test_transform_normalize_global_var.py
slyubomirsky: I wish (see above comments) :cry:

python/tvm/script/parser/core/entry.py
slyubomirsky: Will do, I was not aware that there was a similar analysis in TIR. We should definitely apply both.

python/tvm/script/parser/core/entry.py
slyubomirsky: Mmm, handling the failing test cases will be a lot harder without this capability so I will see how I can improve the parser.

slyubomirsky: Ah, I think I know why: Unlike the decorators for TIR and Relax functions (which are inside the module), the decorator for the module itself will not be saved when the module is parsed (i.e., the raw AST sent to the parser will not include the decorator because the decorator will have been "applied") so I'll have to find another means of passing that information along.

slyubomirsky:
@Lunderberg thank you for the review. I've pushed a change to enable a well-formed check for individual functions and PrimFuncs, which can also be disabled with the `check_well_formed` argument, but it led to some TIR well-formedness failures. I am not sure what could be wrong with most of them so I'd appreciate a look if you have time once the CI finishes. Oddly, setting `check_well_formed` to false on those did not correct the errors, so I'm not sure what the issue was.

tests/python/arith/test_arith_domain_touched.py
```diff
@@ -71,7 +71,8 @@ def test_domain_touched():
 def test_domain_touched_vector():
     m = tvm.runtime.convert(128)
 
-    @T.prim_func
+    # n is undefined
```
slyubomirsky, line 74: Is this intended?

tests/python/tir-base/test_tir_renew_defs.py
```diff
@@ -82,7 +82,8 @@ def _get_block(f):
 
 
 def test_match_buffer():
-    @T.prim_func
+    # well-formed checker complains about multiple definitions for a variable A0_s1>?
```
slyubomirsky, line 85: Not sure what was going on here, but the well-formed checker complains. Bug or intended?

tests/python/tir-base/test_tir_specialize.py
```diff
@@ -65,7 +65,7 @@ def matmul_m_128(a: T.handle, b: T.handle, c: T.handle) -> None:
             C[vi, vj] = C[vi, vj] + A[vi, vk] * B[vj, vk]
 
 
-@T.prim_func
+@T.prim_func(check_well_formed=False)
```
slyubomirsky, line 68: Not clear to me what's wrong here

tests/python/tvmscript/test_tvmscript_parser_tir.py
```diff
@@ -272,7 +272,7 @@ def test_tir_starred_for_loop():
     @T.prim_func(private=True)
     def starred(a: T.handle, b: T.handle):
         A = T.match_buffer(a, [*dims, 128], "int32")
-        B = T.match_buffer(a, dims, "int32")
+        B = T.match_buffer(b, dims, "int32")
```
slyubomirsky, line 275: This caused a well-formed error but I'm pretty sure it was a bug in the first place.

slyubomirsky:
I've determined that the segfault in the tests comes from `tests/python/tir-transform/test_transform_inject_rolling_buffer.py`. According to the backtrace, this comes from checking the StructInfo for a global variable, but oddly, it does not appear to come from checking well-formedness. Any idea what could be the issue? @Lunderberg

tests/python/arith/test_arith_domain_touched.py
Lunderberg: Probably not.  We could make it well formed by adding `n: T.int32` to the arguments.  It looks like this test is validating that a fixed integer extent can be inferred, even when `n` is dynamic.

tests/python/tir-base/test_tir_renew_defs.py
Lunderberg: I'd say bug in the unit test.  The `A0_s1` is a variable generated to represent `A.strides[1]`.  The `strides = [s, s]` should probably be `strides = [s, 1]`.

tests/python/tir-base/test_tir_specialize.py
Lunderberg: I believe this is from `x` being undefined in the TIR.  It appears as part of a shape in `T.match_buffer`, but as `x * 8*` rather than on its own.

tests/python/tvmscript/test_tvmscript_parser_tir.py
Lunderberg: Agreed, I think this was a bug in the unit test.

Lunderberg (approved):

yongwww (approved):
looks good!

slyubomirsky:
No clue why `tests/python/tir-transform/test_tir_transform_force_narrow_index_to_i32.py::test_thread_axis2` is failing. There is no well-formed error there, but I get a complaint about dtypes not matching (for the loop iterator `i0_i1_i2_i3_fused_2`). Not sure why it wouldn't have failed before.

slyubomirsky:
Note that `tests/python/tir-transform/test_tir_transform_hoist_if.py::test_hoisting_block_scope_4` and `test_tir_transform_force_narrow_index_to_i32.py::test_thread_axis2` also fail on mainline, so we might have to fix real (unrelated) bugs there or disable the tests.

Lunderberg:
I did a bisect on `test_tir_transform_force_narrow_index_to_i32.py::test_thread_axis2`, and it's been broken for quite some time.  [This merge commit](https://github.com/apache/tvm/commit/3de77f8de) from `main` into `unity` broke it, and it has been broken ever since.  There's some itervar-related changes in `src/script/ir_builder/tir/ir.cc` that I suspect to be the cause.  They remove an unconditional usage of `int32` dtype, which could be previously been assumed.

tests/python/tir-base/test_tir_renew_defs.py
slyubomirsky: Switching the strides value causes the test to fail later (the assert `matched_buffer1.strides[1] != matched_buffer2.strides[1]` fails), so I don't think that's the issue. I do not know what the intent of the test case is, but it may be necessary to make bigger changes to it.

tests/python/tir-base/test_tir_specialize.py
slyubomirsky: Hm, is this a legitimate test case then? I am not sure. It would be easy enough to add x as an argument, but would that still make sense as a test?

tests/python/tir-base/test_tir_specialize.py
```diff
@@ -275,7 +275,8 @@ def before(A: T.Buffer([16, 16], "float32"), B: T.Buffer([16, 16], "float32")):
         for i in range(256):
             B_flat[i] = A_flat[i] * 2.0
 
-    @T.prim_func(private=True)
+    # well-formed checker complains about multiple nested definitions of B_flat?
```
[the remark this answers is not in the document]
slyubomirsky: @Lunderberg do you think anything needs to be done about this case? Is it legitimate to have a well-formedness error here?

tests/python/tir-transform/test_tir_transform_convert_ssa.py
```diff
@@ -327,7 +327,8 @@ class TestDeDuplicateThreadIdxAcrossMultipleFunctions(BaseBeforeAfter):
     def before(self):
         threadIdx_x = tvm.tir.Var("threadIdx_x", "int32")
 
-        @I.ir_module
+        # complaints of duplicate definitions of threadIdx_x
+        @I.ir_module(check_well_formed=False)
```
slyubomirsky, line 331: Is there anything that can be done with this case to make it well-formed?

tests/python/tir-transform/test_tir_transform_lower_cross_thread_reduction.py
```diff
@@ -116,7 +116,8 @@ def no_normal_reduction(a: T.handle, b: T.handle) -> None:
                 B[vi] = B[vi] + A[vi, vk]
 
 
-@T.prim_func
```
slyubomirsky, line 119: Any idea what a fix would look like? I'd prefer not to have exceptions to the check without knowing what a well-formed version would look like

tests/python/tvmscript/test_tvmscript_roundtrip.py
```diff
@@ -27,8 +27,9 @@
 
 
 def opt_gemm_normalize():
-    @tvm.script.ir_module
+    @tvm.script.ir_module(check_well_formed=False)
```
slyubomirsky, line 30: Any idea for making this example well-formed? I tried replacing the hanging buffers with `decl_buffer` but it introduced a new error

tests/python/codegen/test_inject_ptx_ldg32.py
```diff
@@ -20,7 +20,8 @@
 import tvm.testing
 
 
-@T.prim_func
+# A_local is undefined
+@T.prim_func(check_well_formed=False)
```
slyubomirsky, line 24: Any suggestion for avoiding this undefined var? Would `decl_buffer` work?

tests/python/tvmscript/test_tvmscript_roundtrip.py
Lunderberg: I think this is a bug in the well-formed checker.  The behavior of a `BufferRealize` node depends on whether it is an externally-provided buffer or not.  For an externally-provided buffer, it indicates the region in which the buffer is accessed.  For other buffers, it indicates the region for which the buffer must be allocated.

So, the well-formed checker should treat `BufferRealize` as a point of definition if the buffer hasn't already been defined.  I've submitted https://github.com/apache/tvm/pull/16655, which resolves this issue as well as a few other failures for well-formed checks that have been reported.

tests/python/tir-transform/test_tir_transform_lower_cross_thread_reduction.py
Lunderberg: In the `T.evaluate(T.tvm_thread_allreduce(...))` call, `k` should be replaced with `vk`.

tests/python/tir-transform/test_tir_transform_convert_ssa.py
Lunderberg: I think this one should disable the well formed check.  This test is validating that the pass has correct behavior when resolving SSA violations across multiple functions.  So, the `before` should be ill-formed, but the `expected` should be well-formed.

tests/python/tir-transform/test_tir_transform_convert_ssa.py
```diff
@@ -350,7 +351,8 @@ def kernel_2(A: T.Buffer([256], "float32")):
         return mod
 
     def expected(self):
-        @I.ir_module
+        # complaints of duplicate definitions of threadIdx_x
```
Lunderberg, line 354: Can you verify that this occurs in the `expected` and not just the `before`?  The before case should be ill-formed, because `threadIdx_x` is deliberately defined outside the TIR function and reused, but the expected case defines `threadIdx_x` inside of each function.

slyubomirsky: It happens in both

tests/python/codegen/test_inject_ptx_ldg32.py
slyubomirsky: Good to know, I'll do that

tests/python/tvmscript/test_tvmscript_roundtrip.py
slyubomirsky: Excellent! I'm glad to know that this little change has uncovered real bugs

tests/python/tir-transform/test_tir_transform_lower_cross_thread_reduction.py
slyubomirsky: Thanks!

tests/python/tir-transform/test_tir_transform_convert_ssa.py
slyubomirsky: Good to know, I will update the explanation

tests/python/tir-transform/test_tir_transform_lower_cross_thread_reduction.py
slyubomirsky: Changing it causes the test to fail, so I think we need to change the transformation too.

tests/python/codegen/test_inject_ptx_ldg32.py
slyubomirsky: Hm, possible parsing bug? I get this error:
```python
ValueError: Block frame or PrimFunc frame not find. Please ensure 'T.alloc_buffer' is called under T.block() or T.prim_func()
 --> /home/slyubomirsky/code/tvm/tests/python/codegen/test_inject_ptx_ldg32.py:31:15
    |  
 31 |      A_local = T.alloc_buffer((32), "float32", scope="local")
    |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
It is being done inside a PrimFunc, so I don't know why we would get that error. I'll add a block and see if the example will work regardless.

Lunderberg: Whoops, you were right initially, and it should have been a `T.decl_buffer`.  I got it confused between the `T.decl_buffer` construct in TVMScript, and the `DeclBuffer` node in C++.  If `T.decl_buffer` has the default value of `data=None`, then it will expand to a `DeclBuffer` and the corresponding `Allocate` node.

The `T.alloc_buffer` is only valid within a block, and fills the `BlockNode::alloc_buffers` field.

tests/python/tvmscript/test_tvmscript_roundtrip.py
Lunderberg: Absolutely.  I really like enabling checks to run in all cases for that reason.  There's always some edge case that they find, which can lead to further improvement.

tests/python/tir-transform/test_tir_transform_lower_cross_thread_reduction.py
Lunderberg: Yeah, I think this pass pre-dates schedulable TIR, and doesn't check for mapping of `ForNode::loop_var` to `BlockNode::iter_vars`.  For now, I'd probably just marked as ill-formed, to avoid expanding the scope of this PR.

yongwww:
@tvm-bot rerun

slyubomirsky:
Another failing test, presumably unrelated to my changes: `tests/python/tir-transform/test_tir_transform_inject_ptx_async_copy.py::test_vectorize_cp_async_in_if_then_else`. It complains that the var `data_im2col_reindex_shared_dyn` is undefined, which appears wrong, since it is defined with an `alloc_buffer`.

slyubomirsky:
I've determined that the above failure was introduced by commit `ff0b99c5ce4371ec966cd4fa07ae36351faf2a5e`. In particular, reordering `MergeSharedMemoryAllocations` in `[src/driver/driver_api.cc](https://github.com/apache/tvm/commit/ff0b99c5ce4371ec966cd4fa07ae36351faf2a5e#diff-6e7a38d1bddf565ab5096fbd3a85f39c4b6002e75fbc1cea73d783d3a803a086)` triggers it--reverting those changes fixes it. However, I don't want to reverse the reordering because there is a comment on it that implies it was done for a reason, so I will see if there is a bug in that pass.

Lunderberg:
Taking a look at that commit, there's a couple of things that stand out to me.

* Related to `MergeSharedMemoryAllocations`
  * The `MergeSharedMemoryAllocations` is a copy/paste of `StorageRewrite` with modifications.  The `StorageRewrite` pass is applied earlier and already handled merging of shared memory allocations, so I'm not sure why the separate pass is needed.
    * It looks like there's some different descriptions of `StorageRewrite`.  [This 2021 comment](https://github.com/apache/tvm/pull/9341#issuecomment-949269726) states that `StorageRewrite` cannot merge allocations of different dtypes.  However, [this 2018 PR](https://github.com/apache/tvm/pull/9341#issuecomment-949269726) added memory-sharing of allocations of different dtypes for `StorageRewrite`.  This may be related to static vs dynamic shapes.
  * The `StorageRewrite` pass only merged allocations that were within the same `attr::thread_extent` annotation, and would be part of the same kernel.  Assuming that `MergeSharedMemoryAllocations` kept this behavior from `StorageRewrite`, moving it after `SplitHostDevice` shouldn't be necessary.
  * I'd have to dig into it to be sure, but using `MergeSharedMemoryAllocations` after `ThreadSync("warp")` seems like it could cause an issue.  Applying `MergeSharedMemoryAllocations` (merge two buffers into one) feels like it would undo the benefit of `ThreadSync("warp")` (synchronize to avoid contention of a single buffer).

* Related to `HoistIfThenElse`
  * The comment about requiring `HoistIfThenElse` to be before `UnrollLoop` doesn't make sense to me.  Any condition that could be statically proven for all indices prior to `UnrollLoop` could also be statically proven for a specific loop index.  The end result of `UnrollLoop` and `HoistIfThenElse` should be the same regardless of the ordering.
    * Slight caveat: Unless the `HoistIfThenElse` causes a different decision to be made by the automatic unroll heuristics in `UnrollLoop`.  However, those are disabled by default, and are not enabled by the unit test.
  * Moving `HoistIfThenElse` changed it's ordering relative to user-specified passes in the `"tir.add_lower_pass"` configuration.  While I don't know of any usage of this functionality, it's a bit worrying that it occurred without discussion.
  * The new function attribute `"tir.HoistIfThenElseExprWithBlock"` overrides the user's configuration for `"tir.HoistIfThenElse"`.  If the user had selected a stronger hoisting to be performed (e.g. hoisting `tir::IfThenElse` statements), it would no longer be applied.
  * The configuration used when the new function attribute `"tir.HoistIfThenElseExprWithBlock"` is set doesn't make sense for this pass's location in the lowering flow.  The `HoistIfThenElse` pass occurs after `ConvertBlocksToOpaque` has removed all block variables, so the `HoistedConditionals::kUsingBlockVar` flag shouldn't be required.

* Related to unit tests
  * The unit tests added in `test_gpu_low_batch_gemv.py` only exercise the meta-schedule functionality, and do not exercise any of the changes made to `driver_api.cc` or to `HoistIfThenElse`.

slyubomirsky:
Thanks for the notes. I will note that `HoistIfThenElse` does not affect the failing test case, as I tested those changes separately. Perhaps we should discuss with @jinhongyii, the author of that PR?

tests/python/relax/test_transform_rewrite_cuda_graph.py
```diff
@@ -709,7 +709,7 @@ def main():
 def test_static_args():
```
slyubomirsky, line 709: This test case seemed a little concerning to me because the transformation hides the potentially impure packed call in `Before`'s `main` function and also because the `builtin_with_ctx` is not considered pure (maybe we should add a pure override for cases like that)?

Lunderberg: Agreed, I feel like the `main` function should be marked as impure, because it entirely consists of side effects.

I could see a case for having a wrapper function marked as pure, even if it calls impure functions.  That's kind of what the `R.call_tir` builtin does, to provide a pure wrapper for an impure TIR function.  However, that wouldn't be the case here, as it the `cuda_graph.run_or_capture` function is entirely impure.

slyubomirsky:
Hm, the only remaining failure has to do with doc generation, as the doc generator finds multiple definitions of Target. One results from my previous changes (the `import *` in `python/tvm/tir/transform/__init__.py` will re-export the import of target), but I am not sure what is causing the others.

tqchen:
Unfortunately we find that the pr caused an outage of the MLC compilation, seems to relates to `MergeSharedMemoryAllocations` location change. 

We should also carve out a UT from failed TIR which could potentially helps future regressions.


 opened a #16769 as a temp measure to get things back for now. We can redo the PR ideally not changing the `MergeSharedMemoryAllocations`, or confirm that the carved out tir cases can pass


cc @slyubomirsky @jinhongyii

src/driver/driver_api.cc
```diff
@@ -607,9 +608,6 @@ transform::Sequential MixedModulePassManager(IRModule mixed_mod, Target target)
 
   mixed_pass_list.push_back(tir::transform::AnnotateDeviceRegions());
   mixed_pass_list.push_back(tir::transform::SplitHostDevice());
-  // MergeSharedMemoryAllocations must be applied after SplitHostDevice
-  // because the merged allocation site is at the beginning of each device function
```
tqchen, line 611: The main reason being here is because we cannot merge dyn share mem across kernel boundary i assume

tests/python/tir-transform/test_tir_transform_merge_dynamic_shared_memory_allocations.py
```diff
@@ -456,7 +456,7 @@ def func(
 class TestSimpleAllocNoReuse(tvm.testing.CompareBeforeAfter):
     """Test alloc and free within the same scope."""
 
-    transform = tvm.tir.transform.MergeDynamicSharedMemoryAllocations()
```
tqchen, line 459: I am not sure the change here is intended `MergeDynamicSharedMemoryAllocations`

slyubomirsky: I was getting an error that there is no such pass

PR #16571, opened 2024-02-14
[MISC] Fix compile warnings

Hzfengsy:
This PR fixes the following compile warnings with clang-17:
```
[build] warning: unknown command tag name [-Wdocumentation-unknown-command]
[build]    90 |    *     @R.function
[build]       |          ^~

[build] warning: 'tvm::codegen::CodeGenWebGPU::AddFunction' hides overloaded virtual function [-Woverloaded-virtual]
[build]    52 |   runtime::FunctionInfo AddFunction(const PrimFunc& f, bool skip_readonly_decl);  // NOLINT(*)
[build]       |
```

cc @Lunderberg @tqchen

### DIFF
--- a/src/target/source/codegen_webgpu.h
+++ b/src/target/source/codegen_webgpu.h
@@ -21,7 +21,7 @@
  * \file codegen_webgpu.h
  * \brief Generate WebGPU shaders in WGSL.
  *
- * This module generates WGSL shading langauge.
+ * This module generates WGSL shading language.
  * See https://www.w3.org/TR/WGSL/ for the language reference.
  */
 #ifndef TVM_TARGET_SOURCE_CODEGEN_WEBGPU_H_
@@ -40,7 +40,7 @@ namespace codegen {
  * \brief WebGPU code generator.
  *
  * Note WGSL have a different syntax from normal C.
- * We only leevrage the C for expression generation and
+ * We only leverage the C for expression generation and
  * write most of the language generations.
  */
 class CodeGenWebGPU final : public CodeGenC {
@@ -48,6 +48,7 @@ class CodeGenWebGPU final : public CodeGenC {
   explicit CodeGenWebGPU(Target target);
   // overrides
   std::string Finish() final;
+  using CodeGenC::AddFunction;
   runtime::FunctionInfo AddFunction(const PrimFunc& f, bool skip_readonly_decl);  // NOLINT(*)
   void InitFuncState(const PrimFunc& f) final;
   void PrintStorageSync(const CallNode* op) final;     // NOLINT(*)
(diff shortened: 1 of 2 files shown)


quic-sanirudh (approved):

src/target/source/codegen_webgpu.h
```diff
@@ -40,14 +40,15 @@ namespace codegen {
...
- * We only leevrage the C for expression generation and
+ * We only leverage the C for expression generation and
  * write most of the language generations.
  */
 class CodeGenWebGPU final : public CodeGenC {
  public:
   explicit CodeGenWebGPU(Target target);
   // overrides
   std::string Finish() final;
+  using CodeGenC::AddFunction;
```
Lunderberg, line 51: I assume this is for silencing a warning about shadowing `CodeGenC::AddFunction`?  If so, could we instead remove the `skip_readonly_decl` argument altogether?

The `CodeGenWebGPU::AddFunction` method is only ever called [here](https://github.com/apache/tvm/blob/main/src/target/source/codegen_webgpu.cc#L723), and it is always passed `false`.  I'm wondering if it should be part of the `CodeGenWebGPU` constructor instead.

Lunderberg (approved):
Overall, looks good, and thank you for the fix.  I have one question about a warning that should perhaps be fixed in a different manner, but that can be in a follow-up PR.

src/target/source/codegen_webgpu.h
Hzfengsy: Thanks for your comment, but I'm not sure if it helps. As [stack overflow](https://stackoverflow.com/questions/18515183/c-overloaded-virtual-function-warning-by-clang) says, the warning happens if the functions have the same names but different signatures. 

Indeed, there are different solutions, like renaming `CodeGenWebGPU::AddFunction`. Meanwhile, I think it's also fine to use `using CodeGenC::AddFunction;`, as the compiler will import the function automatically and implicitly. 

I have no personal preference. I would love to follow up if you have a strong opinion on a specific solution :)

Lunderberg: My main concern is because `CodeGenC::AddFunction` is a virtual function.  The `CodeGenWebGPU::AddFunction` shadows `CodeGenC::AddFunction`, but doesn't override it due to the different signature.  Calls made to this function through a base class pointer `CodegenC* my_codegen = get_codegen(); my_codegen->AddFunction(...)` would therefore call `CodeGenC::AddFunction` not `CodeGenWebGPU::AddFunction`.

I don't think there are any calls to `AddFunction` through a base class pointer, but its something that could cause unexpected breakage in the future.

PR #16572, opened 2024-02-14
[Doc] Remove MxNet related tutorials

Hzfengsy:
As mxnet is retired, we remove related tutorials and scripts first.

We will also remove mxnet frontend support in the future


Hzfengsy:
cc @tqchen

yongwww (approved):
LGTM! Many thanks for effort!

PR #16575, opened 2024-02-15
[Unity] GPU sampling

yongwww:
This PR introduces the `multinomial_from_uniform` into nn module for GPU case, and `vm.builtin.multinomial_from_uniform` for CPU. The `multinomial_from_uniform` returns a tensor wherein each row contains an index sampled from a multinomial probability distribution.

[The test case](https://github.com/apache/tvm/pull/16575/commits/abbe733e4af04ac1591066769d269d8c3aa95d56#diff-cd42596733b780d4b3ca6e4e66a002f6175c144821cdeabe8ebaa91bc6c16f67L907-L924) is for testing the build and run, it has been commented out to avoid overloading the nn module op tests.

----
Updated on Feb. 21, 2024

Add `sample_top_p_top_k_from_sorted_prob` and `renormalize_top_p_top_k_prob`

cc: @tqchen @vinx13 @MasterJH5574

### DIFF
--- a/python/tvm/relax/frontend/nn/op.py
+++ b/python/tvm/relax/frontend/nn/op.py
@@ -2057,26 +2057,31 @@ def cumsum(
     return wrap_nested(_op.cumsum(data._expr, axis, dtype, exclusive), name)
 
 
-def multinomial_from_uniform(prob: Tensor, uniform_sample: Tensor):
+def multinomial_from_uniform(prob: Tensor, uniform_sample: Tensor, dtype: str = "int64"):
     """Returns a tensor where each row contains the index sampled from the multinomial
     probability distribution located in the corresponding row of tensor prob.
 
     Notes
     -----
     For better cpu performance, use 'vm.builtin.multinomial_from_uniform'.
+    For accurate results, ensure probabilities are between 0 and 1 and sum to 1.
 
     Parameters
     ----------
     prob : Tensor
-        The probability tensor.
+        A 2-D tensor of shape (batch, vocab_size) representing probability distributions.
+        Each row is a distribution across vocabulary for a batch, where:
+        Values range from [0, 1], indicating the probability of each vocabulary item.
+        The sum of values in each row is 1, forming a valid distribution.
 
     uniform_sample : Tensor
-        The uniform sample.
+        The uniformly sampled 2-D tensor with the shape (batch, 1).
+        Values range from 0 to 1, indicating probabilities sampled uniformly.
 
     Returns
     -------
     result : Tensor
-        The computed result.
+        The computed tensor with shape (batch, 1).
 
     Examples
     --------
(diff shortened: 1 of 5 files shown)


python/tvm/relax/frontend/nn/op.py
```diff
@@ -1825,3 +1826,260 @@ def print_(tensor: Tensor):
...
+    prob : Tensor
+        The probability tensor.
+
+    uniform_sample : Tensor
+        The uniform sample.
+
+    Returns
+    -------
+    result : Tensor
+        The computed result.
```
MasterJH5574, line 2045: Shall we note the shapes of these tensors in the docstring? For example wen can say `prob` has shape `(b, n)`, `uniform_sample` has shape `(b, 1)`, and `result` has shape `(b, 1)`.

python/tvm/relax/frontend/nn/op.py
```diff
@@ -1825,3 +1826,260 @@ def print_(tensor: Tensor):
...
+        for ax0 in T.parallel(batch):
+            for ax1 in T.parallel(vocab_size):
+                with T.block("T_get_sample_index"):
+                    v_ax0, v_ax1 = T.axis.remap("SS", [ax0, ax1])
+                    T.writes(output_index[v_ax0, 0])
+                    if usample[v_ax0, T.int64(0)] < prob[v_ax0, v_ax1] or v_ax1 + 1 == vocab_size:
+                        if v_ax1 == 0:
+                            output_index[v_ax0, 0] = 0
+                        else:
+                            if not (usample[v_ax0, T.int64(0)] < prob[v_ax0, v_ax1 - 1]):
```
MasterJH5574, line 2077: Just curious: is there any difference between using `not (xxx < yyy)` and `xxx >= yyy` here?

python/tvm/relax/frontend/nn/op.py
```diff
@@ -1825,3 +1826,260 @@ def print_(tensor: Tensor):
...
+    probability distribution located in the corresponding row of tensor prob.
+
+    Notes
+    -----
+    For better cpu performance, use 'vm.builtin.multinomial_from_uniform'.
+
+    Parameters
+    ----------
+    prob : Tensor
+        The probability tensor.
```
MasterJH5574, line 2037: The TIR function does not require the input prob distribution to have a sum of 1 for each row and also does not hard require the samples to be in range `[0, 1]`. Shall we add these notes in the docstring? It might be good to just say the behavior is undefined when the prob distributions do not sum to 1 and samples not in range [0, 1].

python/tvm/relax/frontend/nn/op.py
```diff
@@ -1825,3 +1826,260 @@ def print_(tensor: Tensor):
...
+            [ 4,  9, 15]]
+
+        a = [1, 0, 1, 0, 1, 1, 0]  # a is a boolean array
+        cumsum(a, dtype=int32)  # dtype should be provided to get the expected results
+        -> [1, 1, 2, 2, 3, 4, 4]
+    """
+    return wrap_nested(_op.cumsum(data._expr, axis, dtype, exclusive), name)
+
+
+def multinomial_from_uniform(prob: Tensor, uniform_sample: Tensor):
```
MasterJH5574, line 2026: Do we have tests showing this op can build and correctly run on GPU?

MasterJH5574 (commented):
Thank you @yongwww! I just have a few comments that might help the clarity.

python/tvm/relax/frontend/nn/op.py
yongwww: good catch! The test case can be found [here](https://github.com/apache/tvm/pull/16575/commits/abbe733e4af04ac1591066769d269d8c3aa95d56#diff-cd42596733b780d4b3ca6e4e66a002f6175c144821cdeabe8ebaa91bc6c16f67L907-L924). The reason I didn't enable it was the nn module tests currently don't cover the build and run. I will add it back.

python/tvm/relax/frontend/nn/op.py
yongwww: They are logically equivalent. Just compared the cuda code for the two, the codes are also equivalent.

yongwww:
@MasterJH5574 I've addressed your comments and added two new operations, along with tests for the compile and run. Please take a look them at your convenience.

python/tvm/relax/frontend/nn/op.py
```diff
@@ -1825,3 +1827,509 @@ def print_(tensor: Tensor):
...
+    Returns
+    -------
+    result : Tensor
+        The filtered and nomalized tensor with the sampe shape as input prob.
+    """
+    dtype = prob.dtype
+    batch = sorted_prob.shape[0]
+
+    @T.prim_func(private=True)
+    def _get_renorm_cutoff(A: T.handle, B: T.handle, C: T.handle):
```
tqchen, line 2288: make cumsum_mask a function that takes in axis, then inline it via metaprogramming

```python
def cumsum_mask(cumsum_sorted, top_p, top_k, i, j):
     return _tir.all(cumsum_sorted[i, j] < top_p[i, 0], j + 1 < top_k[i, 0])
```

tqchen (approved):

PR #16579, opened 2024-02-15
[Dlight] Scheduling Low batch GEMM using GEMV-like rule

jinhongyii:
1. add a dlight rule LowBatchGEMV to schedule low-batch GEMM just like GEMV.
2. fix some issues when lowering low-batch GEMM


jinhongyii:
cc: @MasterJH5574

Hzfengsy:
Would be great if you could add testcases :)

spectrometerHBH (approved):

PR #16584, opened 2024-02-16
[Unity][TIR] Clear struct info when specializing PrimFunc

Lunderberg:
In rare cases, a `PrimFunc` may be annotated with `StructInfo`, to indicate that it is an impure function with specific shapes for the parameters.  If struct info is present, it is invalidated when specializing a `PrimFunc`, and should be cleared.

### DIFF
--- a/tests/python/tir-base/test_tir_specialize.py
+++ b/tests/python/tir-base/test_tir_specialize.py
@@ -16,6 +16,8 @@
 # under the License.
 # pylint: disable=missing-function-docstring, missing-module-docstring
 
+import pytest
+
 import tvm
 from tvm.script import tir as T
 from tvm.tir.schedule.testing import assert_structural_equal_ignore_global_symbol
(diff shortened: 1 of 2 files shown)


tests/python/tir-base/test_tir_specialize.py
```diff
@@ -324,5 +326,41 @@ def expected(A_data: T.handle("float32")):
     tvm.ir.assert_structural_equal(expected, after)
 
 
+def test_specialization_removes_struct_info():
+    """Reset struct info in specialization
+
+    While a PrimFunc usually doesn't have a `relax.StructInfo`, the
+    field can be populated in some edge cases.  If that PrimFunc is
```
slyubomirsky, line 333: Do you know of times when we're doing it? It should be something we do consistently or not at all.

slyubomirsky (commented):
The change seems straightforward but I would be more curious to hear about why it's necessary (i.e., what problems were being caused) and when the StructInfo is being set for PrimFuncs.

tests/python/tir-base/test_tir_specialize.py
Lunderberg: There's a couple of MLC ([example](https://github.com/mlc-ai/mlc-llm/blob/fb4c52c749c4b32594f1e9382113fecd11847893/mlc_llm/transform/fuse_split_rotary_embedding.py#L100)) where the a dynamic PrimFunc is annotated with the `relax.FuncStructInfo`.  This allows the `R.call_tir` shape inference to work correctly, even after the arguments are mutated.

IIRC, the primary reasons not to add `FuncStructInfo` to PrimFuncs was that primfuncs didn't follow the relax requirements for functional purity.  Since we now have the purity expressed within `FuncStructInfo`, I think that means we could add `FuncStructInfo` to every PrimFunc on construction, without making any incorrect statements about the functions.

slyubomirsky: Always specifying it sounds like a better approach, as long as it would not interfere with anything in TIR.

PR #16585, opened 2024-02-16
[Unity][Analysis] Include impure call in VerifyWellFormed errors

Lunderberg:
Prior to this commit, `VerifyWellFormed` would state that a dataflow block or pure function contained an impure call, but identifying which call was impure was left to the user.

This commit updates `VerifyWellFormed` to show the impure `relax::Call` as part of the error message.


slyubomirsky (approved):
A very good UX improvement, thanks a ton.

PR #16586, opened 2024-02-16
[Relax] Ignore non-relax functions in relax.transform.RunCodegen

Lunderberg:
The `relax.transform.RunCodegen` pass replaces calls to relax functions with the `"Codegen"` attribute with calls into a compiled module. Prior to this commit, while calls to relax functions without the `"Codegen"` attribute were ignored, calls to non-relax functions would raise an error.

This commit updates `relax.transform.RunCodegen` to also ignore calls to non-relax functions.

### DIFF
--- a/tests/python/relax/test_transform_codegen_pass.py
+++ b/tests/python/relax/test_transform_codegen_pass.py
@@ -40,7 +40,7 @@ has_tensorrt_runtime = pytest.mark.skipif(
 )
 
 # Global variable in pytest that applies markers to all tests.
-# pytestmark = [has_tensorrt_codegen, has_tensorrt_runtime]
+pytestmark = [has_tensorrt_codegen, has_tensorrt_runtime]
 
 # Target gpu
 target_str = "nvidia/nvidia-t4"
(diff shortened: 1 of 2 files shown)


tests/python/relax/test_transform_codegen_pass.py
```diff
@@ -40,7 +40,7 @@
 )
 
 # Global variable in pytest that applies markers to all tests.
-pytestmark = [has_tensorrt_codegen, has_tensorrt_runtime]
+# pytestmark = [has_tensorrt_codegen, has_tensorrt_runtime]
```
yongwww, line 43: I guess we will need this

tests/python/relax/test_transform_codegen_pass.py
```diff
@@ -350,6 +350,36 @@ def main(
 
     after = relax.transform.RunCodegen()(Before)
     tvm.ir.assert_structural_equal(after["main"], Expected["main"])
+    after.show()
```
yongwww, line 353: remove this printing

yongwww (approved):
good addition!

tests/python/relax/test_transform_codegen_pass.py
Lunderberg: Whoops, thank you for the catch.  This was commented out for debugging, and the PR is updated to remove the change.

tests/python/relax/test_transform_codegen_pass.py
Lunderberg: Thank you, and cleaned up!

PR #16590, opened 2024-02-16
[Unity][Transform] Check for permute_dims in ExpandMatmulOfSum

Lunderberg:
This pattern occurs whenever `relax.op.linear` is used.


slyubomirsky (approved):
The change is straightforward and it's good to handle a case that comes up often.

PR #16592, opened 2024-02-16
[Relax] Handle dynamic arguments in legalization of nn.attention

Lunderberg:
Prior to this commit, when using `causal_mask="BottomRight"` in `R.nn.attention`, the legalization would assume that the query and key/value sequence lengths were static integers.  This commit updates the legalization to allow dynamic shapes.


slyubomirsky (approved):
Good catch and it's a nice improvement. We should really have "always think about the dynamic case" as a motto.

Lunderberg:
> We should really have "always think about the dynamic case" as a motto.

I like the motto, as it's the most user-friendly option at a high level.  Since specialization can go from dynamic to static, but going from static to dynamic would be ambiguous, I really like "dynamic until specialized" as the default behavior.

PR #16594, opened 2024-02-16
[Relax][Transform] Preserve param names in LiftTransformParams

Lunderberg:
The `relax.transform.LiftTransformParams` pass splits apart a relax function, extracting the steps that could be performed at compile-time.  Prior to this commit, the transformed parameters were named `param0`, `param1`, and so on.

This commit updates the `LiftTransformParams` pass to preserve any human-readable parameter names.  The parameter names for the updated function are taken from the original parameter names, if no transformation is performed, or from the internal variable binding, if a transformation is applied.  This implementation uses `LambdaLift` internally, relying on the changes made in
https://github.com/apache/tvm/pull/16306.

### DIFF
--- a/src/relax/transform/canonicalize_bindings.cc
+++ b/src/relax/transform/canonicalize_bindings.cc
@@ -247,7 +247,8 @@ class BindingCanonicalizer : public ExprMutator {
   }
 
   Expr VisitExpr_(const VarNode* var) override {
-    Var new_var = GetRef<Var>(var);
+    Var new_var = Downcast<Var>(ExprMutator::VisitExpr_(var));
+
     while (auto opt = plan_.replace_usage.Get(new_var->vid)) {
       new_var = opt.value();
     }
--- a/src/relax/transform/canonicalize_bindings.cc
+++ b/src/relax/transform/canonicalize_bindings.cc
@@ -247,8 +247,7 @@ class BindingCanonicalizer : public ExprMutator {
   }
 
   Expr VisitExpr_(const VarNode* var) override {
-    Var new_var = Downcast<Var>(ExprMutator::VisitExpr_(var));
-
+    Var new_var = GetRef<Var>(var);
     while (auto opt = plan_.replace_usage.Get(new_var->vid)) {
       new_var = opt.value();
     }
--- a/src/relax/transform/lift_transform_params.cc
+++ b/src/relax/transform/lift_transform_params.cc
@@ -49,8 +49,8 @@ struct CollectInfo {
 
   /*! \brief Bindings that can be lifted out into a pre-processing
    *
-   * - All bindings in `liftable_bindings` are suitable for use in a
-   *   DataflowBlock.
+   * - All bindings in `computable_at_compile_time` are suitable for
+   *   use in a DataflowBlock.
    *
    * - Do not depend on any parameter prior to attr::kNumInput.
    *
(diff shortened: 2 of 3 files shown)


src/relax/transform/lift_transform_params.cc
```diff
@@ -37,405 +38,443 @@
...
+struct CollectInfo {
+  /* \brief The analyzed function */
+  Function orig_func;
+
+  /* \brief The number of parameters unknown until runtime */
+  size_t num_runtime_params;
+
+  /*! \brief Bindings that can be lifted out into a pre-processing
+   *
+   * - All bindings in `liftable_bindings` are suitable for use in a
```
slyubomirsky, line 52: Is assume `liftable_bindings` is an old name that should be updated?

src/relax/transform/lift_transform_params.cc
```diff
@@ -37,405 +38,443 @@
...
-        can_lift = false;
+      if (!liftable_vars_.count(var)) {
+        return false;
       }
     }
 
     // Cond 5. Do not lift declarations of external functions
-    if (binding->value.as<relax::ExternFuncNode>()) {
-      can_lift = false;
+    if (value.as<relax::ExternFuncNode>()) {
```
slyubomirsky, line 365: I'm curious when we ever have an ExternFunc as the RHS of a binding.

src/relax/transform/canonicalize_bindings.cc
```diff
@@ -247,7 +247,8 @@ class BindingCanonicalizer : public ExprMutator {
   }
 
   Expr VisitExpr_(const VarNode* var) override {
-    Var new_var = GetRef<Var>(var);
```
slyubomirsky, line 250: Was there any bug you encountered due to not doing this visit before? If so, that would be a good regression test to throw in.

src/relax/transform/lift_transform_params.cc
```diff
@@ -37,405 +38,443 @@
...
+    }
 
-      updates->Add(new_gvar, transform_func);
+    return mod;
+  };
+  return tvm::transform::CreateModulePass(pass_func, 1, "PartitionTransformParams", {});
+}
+Pass LiftTransformParams() {
+  // A post-proc utility to reproduce the previous behavior of
+  // LiftTransformParams.
```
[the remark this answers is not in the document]
Lunderberg: No, I completely agree with you there.  I should have kept references to "previous behavior" in the commit message instead, where there's a clear before/after.  Updated the comment to describe this as the third step in `LiftTransformParams`, and to include the names of the other two steps.

src/relax/transform/lift_transform_params.cc
```diff
@@ -37,405 +38,443 @@
...
+  //
+  // 1. Partition each function into a compile-time and run-time
+  // lambda functions.
```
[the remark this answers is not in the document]
Lunderberg: Thank you, and fixed!

src/relax/transform/lift_transform_params.cc
```diff
@@ -37,405 +38,443 @@
...
-    }
-    return ExprMutator::VisitExpr_(var);
   }
+};
 
-  // Remap the original parameters to TupleGetItem from the packed tuple of transformed parameters.
-  std::unordered_map<Var, Expr, ObjectPtrHash, ObjectPtrEqual> param_remap_;
-  // The plan of lifting the transform params
-  LiftTransformParamsInfoPlan lift_plan_;
+// Adapted from https://stackoverflow.com/a/2072890
```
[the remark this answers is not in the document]
Lunderberg: Thank you.  At some point, this may be useful as a member function of `tvm::String`, similar to how python has `str.endswith` and `str.startswith`.

src/relax/transform/lift_transform_params.cc
Lunderberg: Thank you, and that is exactly correct.  I replaced all references, but didn't grep the comments for spots that needed to be updated.  Fixed.

src/relax/transform/lift_transform_params.cc
Lunderberg: It happened more often prior to https://github.com/apache/tvm/pull/15900, but rarely happens now.

At some point, I'd like to allow a relax function to specify when `tvm::runtime::Registry::Get` is invoked.  It's legal for a `R.ExternFunc` to appear within the `IRModule::functions` table, on the RHS of a binding, or inline for a `relax::CallNode::op`.  Currently, all three of those are resolved when the `relax::VirtualMachine` is instantiated, but they could be used to represent binding at instantiation-time, at function start-time, and at function execution-time.  In that case, the ExternFunc would become more common as the RHS of a binding.

src/relax/transform/canonicalize_bindings.cc
Lunderberg: Hmm, this was definitely required during initial development, and was intended to handle cases where a replacement in `ExprMutator::var_remap_` occurs outside of the planned replacements in the `ReplacementPlan.replace_usage_`.  This can happen if the removal of trivial bindings triggers `InferStructInfo` in a downstream operation.  However, that doesn't reproduce the bug, nor does that bug occur in anything on my local dev branch.

I'm removing this line from the PR, and can make a separate PR (with test case!) if/when the bug resurfaces.

slyubomirsky: I don't think it's necessary to remove, especially since it seems like a correct change. I'm glad that there wasn't a bug that you encountered because of it.

src/relax/transform/lift_transform_params.cc
slyubomirsky: I guess ExternFuncs are permitted to be used in that way per the spec as first-class values, so this is good to handle.

PR #16595, opened 2024-02-16
[Transform] Check for zero-param operators in LiftTransformParams

Lunderberg:
Prior to this commit, `LiftTransformParams` would extract out all variable binding that have no runtime dependencies.  As a result, expressions such as `R.zeros([16], "int32")` would be extracted out into the parameter transformation, even though they do not depend on any parameters.

This commit updates `LiftTransformParams` to only output variables that depend on at least one compile-time parameter.

The unit test for this functionality also found that `relax::Call` was erroneously calling `MarkGraphNode` in `SEqualReduce` and `SHashReduce`.  This should only be called for nodes that have have reference equality, such as `relax::Var`, and not for composite objects.  This caused erroneous failures in the unit test when two instances of `R.zeros([16], "int32")` were being compared by reference equality in `StructuralEqual`.  These extra calls to `MarkGraphNode` have been removed.

### DIFF
--- a/include/tvm/relax/expr.h
+++ b/include/tvm/relax/expr.h
@@ -169,13 +169,11 @@ class CallNode : public ExprNode {
 
   bool SEqualReduce(const CallNode* other, SEqualReducer equal) const {
     // skip sinfo_args check for primitive ops.
-    equal->MarkGraphNode();
     return equal(op, other->op) && equal(args, other->args) && equal(attrs, other->attrs) &&
            equal(sinfo_args, other->sinfo_args) && equal(struct_info_, other->struct_info_);
   }
 
   void SHashReduce(SHashReducer hash_reduce) const {
-    hash_reduce->MarkGraphNode();
     hash_reduce(op);
     hash_reduce(args);
     hash_reduce(attrs);
--- a/src/relax/transform/lift_transform_params.cc
+++ b/src/relax/transform/lift_transform_params.cc
@@ -58,6 +58,15 @@ struct CollectInfo {
    */
   std::vector<Binding> computable_at_compile_time;
 
+  /*! \brief Variables that require a compile-time parameter
+   *
+   * Used to distinguish between computed tensors that depend on the
+   * model weights, and computed tensors that require neither model
+   * weights nor runtime arguments (e.g. `R.zeros([16], "float16")`).
+   */
+  std::unordered_set<Variant<relax::Var, tir::Var>, ObjectPtrHash, ObjectPtrEqual>
+      requires_compile_time_param;
+
   /*! \brief Variables that are required at runtime */
   std::unordered_set<Variant<relax::Var, tir::Var>, ObjectPtrHash, ObjectPtrEqual>
       required_at_runtime;
@@ -114,7 +123,8 @@ struct CollectInfo {
     // Any variable that is computed at compile-time, but is required
     // at runtime, must be provided as a parameter.
     for (const auto& binding : computable_at_compile_time) {
-      if (required_at_runtime.count(binding->var)) {
+      if (requires_compile_time_param.count(binding->var) &&
+          required_at_runtime.count(binding->var)) {
         params.push_back(binding->var);
       }
     }
@@ -182,16 +192,21 @@ struct CollectInfo {
 
     // Any binding that is computable at compile-time should be
     // suppressed at run-time.
-    struct SuppressCompileTime : ExprMutator {
-      std::unordered_set<Var, ObjectPtrHash, ObjectPtrEqual> to_suppress;
-      explicit SuppressCompileTime(const std::vector<Binding>& bindings) {
-        for (const auto& binding : bindings) {
-          to_suppress.insert(binding->var);
-        }
+    std::unordered_set<Var, ObjectPtrHash, ObjectPtrEqual> to_suppress;
+    for (const auto& binding : computable_at_compile_time) {
+      if (requires_compile_time_param.count(binding->var)) {
+        to_suppress.insert(binding->var);
       }
+    }
+
+    class SuppressCompileTime : public ExprMutator {
+     public:
+      explicit SuppressCompileTime(
+          const std::unordered_set<Var, ObjectPtrHash, ObjectPtrEqual>& to_suppress)
+          : to_suppress_(to_suppress) {}
 
       void VisitBinding(const Binding& binding) override {
-        if (!to_suppress.count(binding->var)) {
+        if (!to_suppress_.count(binding->var)) {
           ExprMutator::VisitBinding(binding);
         }
       }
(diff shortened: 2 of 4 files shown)


Lunderberg:
This PR depends on changes made in https://github.com/apache/tvm/pull/16594, and is marked as a draft until it lands.

vinx13:
`As a result, expressions such as R.zeros([16], "int32") would be extracted out into the parameter transformation, even though they do not depend on any parameters.
`
Does this affect the result? If a weight transformation depends on some values like `R.zeros`, such transformation will no longer be lifted if `R.zeros` is not lifted, maybe better we can check such dependency

Lunderberg:
> Does this affect the result? If a weight transformation depends on some values like `R.zeros`, such transformation will no longer be lifted if `R.zeros` is not lifted, maybe better we can check such dependency

Good question, and this case should be handled, by allowing zero-param operators to potentially appear in both functions.  (See [this unit test](https://github.com/apache/tvm/pull/16595/files#diff-0164f466641405b90811062a983b85caf8118561d094a7c89515922d1c34044fR822) for how this looks in practice.)  While the case isn't ever explicitly handled, it instead results from the overall lifting.

1. Add every variables binding that doesn't require runtime parameters to the lifted `transform_params` function.
2. For every variable binding that is present in `transform_params` and depends on the model weights, replace with the output from `transform_params`.
3. Run dead-code elimination on both functions.

Since the zero-param operators don't depend on runtime parameters, they appear in the `transform_params`.  Since the zero-param operators don't depend on the model weights, they aren't replaced in the original function.  If they are only required in one of the two functions, then the dead-code elimination will remove them from the other.

src/relax/transform/lift_transform_params.cc
```diff
@@ -37,405 +38,467 @@
...
+      info_.computable_at_compile_time.push_back(binding);
+      liftable_vars_.insert(binding->var);
+
+      auto upstream_vars = FreeVars(bound_value);
+      bool depends_on_compile_time_param = std::any_of(
+          upstream_vars.begin(), upstream_vars.end(),
+          [&](const Var& var) -> bool { return info_.requires_compile_time_param.count(var); });
+      if (depends_on_compile_time_param) {
+        info_.requires_compile_time_param.insert(binding->var);
+      }
```
slyubomirsky, line 344: Is this the logic that is meant to handle zero-param operators? I had a really hard time figuring out what the change was from #16594 and this was the only major area that came up in the diff.

slyubomirsky (commented):
I realize it's a draft but I had a look anyway. Since the code changes also included those from #16594, it was a little difficult to see what had changed. I didn't see anything to take issue with, though it was a little less than obvious to see how the delta from #16594 accomplished the stated purpose, but I think I follow it.

src/relax/transform/lift_transform_params.cc
```diff
@@ -37,405 +38,467 @@
...
-    Expr body = VisitWithNewScope(SeqExpr({block}, transformed_params), Array<Var>{params});
-    Function f_transform_params =
-        Function(/*params=*/{params}, /*body=*/body, /*ret_struct_info=*/NullOpt);
-    return {f_transform_params, output_to_index};
-  }
+    struct SuppressCompileTime : ExprMutator {
+      std::unordered_set<Var, ObjectPtrHash, ObjectPtrEqual> to_suppress;
+      explicit SuppressCompileTime(
+          std::unordered_set<Var, ObjectPtrHash, ObjectPtrEqual> to_suppress)
+          : to_suppress(to_suppress) {}
```
slyubomirsky, line 204: Question separate from the review: Are we going with the convention of having member var names end in an underscore? I'm not a partisan on that one, but we should be consistent. @tqchen

Lunderberg:
> I realize it's a draft but I had a look anyway. Since the code changes also included those from https://github.com/apache/tvm/pull/16594, it was a little difficult to see what had changed.

Apologies there.  Once #16594 lands, the "Files Changed" tab in this PR should update to only show the changes unique to this PR.  In the meantime, this PR branch has its changes in a separate commit ([link](https://github.com/apache/tvm/pull/16595/commits/a8ca845a883cfa7440941cba5ec35c11e2294bc8)), where they can be viewed separately from the #16594 changes.

include/tvm/relax/expr.h
```diff
@@ -169,13 +169,11 @@ class CallNode : public ExprNode {
 
   bool SEqualReduce(const CallNode* other, SEqualReducer equal) const {
     // skip sinfo_args check for primitive ops.
-    equal->MarkGraphNode();
```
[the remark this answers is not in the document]
Lunderberg: The `MarkGraphNode` function should be called for IR nodes that have reference equality, such as variables.  Using it with the `relax::CallNode` means that results from `StructuralEqual` will be dependent on how a `relax::Call` node was constructed, even if they have identical contents.

Relevant to this PR, if `R.zeros([16], "int32")` appears in both `main` and `transform_params`, the expected output generated by the TVMScript parser would have two different `relax::Call` objects, while the output of `LiftTransformParams` would use the same `relax::Call` object.  Because the `relax::Call` objects were being checked for analogous reference equality, this would cause `StructuralEqual` to erroneously report the test as failing.

I've added a unit test to specifically exercise this behavior, rather than implicitly relying on the tests for `LiftTransformParams`.

src/relax/transform/lift_transform_params.cc
Lunderberg: This section along with [this line](https://github.com/apache/tvm/pull/16595/commits/a8ca845a883cfa7440941cba5ec35c11e2294bc8#diff-dbc6646d7cc42ae403e97adb88f092c13fd91bc928595b7f44d54a3f218d714bR124), yes.  This section collects information to determine if this parameter depends, directly or indirectly, on one of the model weights.  The linked line uses the collected information to determine if the `transform_params` should output a variable.

I've updated the comment here to indicate why the additional information is being collected.

src/relax/transform/lift_transform_params.cc
Lunderberg: The TVM convention, as I understand it, is to have private members have a trailing underscore, while public members do not.  Often if I'm making quick subclasses of `ExprMutator`, I'll have a struct with public fields rather than a class with private fields.  Since the class definition already occurs within a function scope, the visibility of the entire class is already restricted, and additional visibility restrictions of private members isn't necessary.

That said, I've updated it to be a class with a private `to_suppress_`.  I figure that if a deviation from convention is big enough to raise a question, it's big enough to avoid the deviation altogether.

include/tvm/relax/expr.h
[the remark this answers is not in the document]
slyubomirsky: Good catch!

src/relax/transform/lift_transform_params.cc
slyubomirsky: I was just curious, I don't think it's worthy of holding up a review or anything like that. Having a different rule for public/private members is reasonable.

slyubomirsky:
Thank you for the changes. The new unit test and the new comment are both helpful.

Lunderberg:
With #16594 landed, I've rebased this PR on top of it, and it is now ready for review.

slyubomirsky (approved):
Thanks for responding to previous feedback.

PR #16597, opened 2024-02-16
[Transform] Allow explicit name of bundled model parameters

Lunderberg:
In `BundleModelParams`, allow the user to specify a name for the tuple parameters.  If unspecified, defaults to the previous name `"model_params"`.


slyubomirsky (approved):
Straightforward and a nice UX improvement. Thank you!

PR #16598, opened 2024-02-16
[Transform][Bugfix] Handle non-composite lambda functions in FuseOps

Lunderberg:
Prior to this commit, calling `FuseOpsByPattern` with `annotate_codegen=True` would cause an error when encountering a lambda function.  This was caused by the `CompositeFunctionAnnotator` asserting that all `relax::Function` encountered must have the `kComposite` attribute.  While this is true for all lambda functions produced by `FuseOpsByPattern`, the user may have defined other lambda functions as well.

This commit updates `CompositeFunctionAnnotator` to ignore lambda functions that do not have a `kComposite` attribute.

### DIFF
--- a/src/relax/transform/fuse_ops.cc
+++ b/src/relax/transform/fuse_ops.cc
@@ -1238,10 +1238,14 @@ class CompositeFunctionAnnotator : public ExprMutator {
 
   Expr VisitExpr_(const FunctionNode* func_node) final {
     Function f_inner = Downcast<Function>(ExprMutator::VisitExpr_(func_node));
-    auto composite_name = func_node->GetAttr<String>(attr::kComposite);
+
+    if (!func_node->GetAttr<String>(attr::kComposite)) {
+      // This lambda function doesn't have `attr::kComposite`, so it
+      // was not produced by FuseOps.
+      return std::move(f_inner);
+    }
 
     f_inner = WithoutAttr(std::move(f_inner), tvm::relax::attr::kPrimitive);
-    ICHECK(composite_name);
 
     Array<Var> param_vars;
     Array<Expr> params;
--- a/tests/python/relax/test_transform_fuse_ops_by_pattern.py
+++ b/tests/python/relax/test_transform_fuse_ops_by_pattern.py
@@ -530,6 +530,60 @@ def test_annotate_codegen():
     )
 
 
+@pytest.mark.parametrize("annotate_codegen", [True, False])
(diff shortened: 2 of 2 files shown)


src/relax/transform/fuse_ops.cc
```diff
@@ -1238,10 +1238,14 @@ class CompositeFunctionAnnotator : public ExprMutator {
 
   Expr VisitExpr_(const FunctionNode* func_node) final {
     Function f_inner = Downcast<Function>(ExprMutator::VisitExpr_(func_node));
-    auto composite_name = func_node->GetAttr<String>(attr::kComposite);
+
+    if (!func_node->GetAttr<String>(attr::kComposite)) {
```
vinx13, line 1242: Are non-composite functions visited? https://github.com/apache/tvm/blob/76c1708018d2bdcf4d5a207100ff8682be6cf09d/src/relax/transform/fuse_ops.cc#L1224 here we only visit composite functions

Lunderberg: The `PatternBasedPartitioner` only visits non-composite functions, produces a composite function for each pattern match, and updates the non-composite function to call the newly-generated composite function.  Afterwards, the call to [`CompositeFunctionAnnotator`](https://github.com/apache/tvm/blob/main/src/relax/transform/fuse_ops.cc#L1289) is called.  This visits only non-composite functions, finds any relax-to-relax function calls, and asserts that the callee is composite.

The callee will be composite for every function call generated by `PatternBasedPartitioner`, but that doesn't guarantee that all relax-to-relax function calls have a composite callee.  If the `IRModule` contains a relax-to-relax call prior to `PatternBasedPartitioner`, that callee may be non-composite.  This IRModule would be entirely legal, but would trigger the assert in `CompositeFunctionAnnotator`.

vinx13 (approved):

src/relax/transform/fuse_ops.cc
Lunderberg: So, the problem isn't with calls to inner functions as on line 1224, but with calls to other functions within the IRModule.

vinx13: I think the problem is if the callee is not a global var, the callee function will still be visited, so the fix makes sense to me

Lunderberg: Whoops, you're right on that one.  It's if there is a inner function in the input `IRModule`. (Apologies, trying to track too many PRs at one time.)

PR #16599, opened 2024-02-16
[Transform] De-duplicate MatchCast nodes in EliminateCommonSubexpr

Lunderberg:
Update the `relax.transform.EliminateCommonSubexpr` pass to handle `R.match_cast` bindings, where the argument of the `R.match_cast` has also been de-duplicated.

### DIFF
--- a/src/relax/transform/eliminate_common_subexpr.cc
+++ b/src/relax/transform/eliminate_common_subexpr.cc
@@ -90,10 +90,16 @@ class CommonSubexprEliminator : public ExprMutator {
   explicit CommonSubexprEliminator(bool call_only = false) : call_only_(call_only) {}
 
   BindingBlock VisitBindingBlock_(const DataflowBlockNode* block) override {
-    auto cache_exprs = expr_replacements_;
     auto cache_vars = var_remap_;
     auto output = ExprMutator::VisitBindingBlock_(block);
-    expr_replacements_ = cache_exprs;
+
+    for (auto& [key, replacements] : expr_replacements_) {
+      replacements.erase(
+          std::remove_if(replacements.begin(), replacements.end(),
+                         [](const Var& var) -> bool { return var->IsInstance<DataflowVarNode>(); }),
+          replacements.end());
+    }
+
     var_remap_ = cache_vars;
     return output;
   }
@@ -120,19 +126,22 @@ class CommonSubexprEliminator : public ExprMutator {
     } else if (ContainsImpureCall(bound_value)) {
       VLOG(1) << "Since the expression is impure, cannot de-duplicate " << bound_value;
 
-    } else if (auto it = expr_replacements_.find(lookup_key); it != expr_replacements_.end()) {
-      VLOG(1) << "Value " << bound_value << " has previously been bound as " << it->second
+    } else if (auto it = expr_replacements_.find(lookup_key);
+               it != expr_replacements_.end() && it->second.size()) {
+      VLOG(1) << "Value " << bound_value << " has previously been bound as " << it->second[0]
               << ".  The duplicate binding of this value to " << binding->var
               << " will be replaced with a trivial binding, "
-              << "and occurrences of " << binding->var << " will be replaced with " << it->second;
-      output_binding = VarBinding(binding->var, it->second);
-      var_remap_.insert({binding->var->vid, it->second});
+              << "and occurrences of " << binding->var << " will be replaced with "
+              << it->second[0];
+      output_binding = VarBinding(binding->var, it->second[0]);
+      var_remap_.insert({binding->var->vid, it->second[0]});
+      it->second.push_back(binding->var);
 
     } else {
       VLOG(1) << "Value " << bound_value << " is bound to " << binding->var
               << " and may be de-duplicated if it occurs again.";
 
-      expr_replacements_.insert({lookup_key, binding->var});
+      expr_replacements_[lookup_key].push_back(binding->var);
     }
 
     builder_->EmitNormalized(output_binding);
@@ -169,7 +178,7 @@ class CommonSubexprEliminator : public ExprMutator {
   }
 
   bool call_only_{false};
-  std::unordered_map<ReplacementKey, Var> expr_replacements_;
+  std::unordered_map<ReplacementKey, std::vector<Var>> expr_replacements_;
 };
 
 }  // namespace
--- a/src/relax/transform/eliminate_common_subexpr.cc
+++ b/src/relax/transform/eliminate_common_subexpr.cc
@@ -160,8 +160,8 @@ class CommonSubexprEliminator : public ExprMutator {
 
   Expr VisitExpr_(const IfNode* op) override {
     Expr cond = VisitExpr(op->cond);
-    Expr true_branch = VisitWithCleanScope(op->true_branch);
-    Expr false_branch = VisitWithCleanScope(op->false_branch);
+    Expr true_branch = VisitWithInnerScope(op->true_branch);
+    Expr false_branch = VisitWithInnerScope(op->false_branch);
     if (op->cond.same_as(cond) && op->true_branch.same_as(true_branch) &&
         op->false_branch.same_as(false_branch) &&
         VisitAndCheckStructInfoFieldUnchanged(op->struct_info_)) {
@@ -172,6 +172,15 @@ class CommonSubexprEliminator : public ExprMutator {
   }
 
  private:
+  Expr VisitWithInnerScope(Expr expr) {
+    auto cached_vars = var_remap_;
+    auto cached_exprs = expr_replacements_;
+    auto output = VisitExpr(expr);
+    var_remap_ = cached_vars;
+    expr_replacements_ = cached_exprs;
+    return output;
+  }
+
   Expr VisitWithCleanScope(Expr expr) {
     CommonSubexprEliminator clean_mutator(call_only_);
     return clean_mutator.VisitExpr(expr);
--- a/src/relax/transform/eliminate_common_subexpr.cc
+++ b/src/relax/transform/eliminate_common_subexpr.cc
@@ -29,51 +29,120 @@
 #include <tvm/relax/transform.h>
 #include <tvm/relax/utils.h>
 
+#include "../../support/utils.h"
+
 namespace tvm {
 namespace relax {
+namespace {
+/* \brief Lookup key for subexpression replacements
+ *
+ * The lookup key must contain the expression being bound, along with
+ * the struct info used for a match cast, if applicable.  Using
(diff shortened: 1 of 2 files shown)


src/relax/transform/eliminate_common_subexpr.cc
```diff
@@ -20,223 +20,162 @@
...
+/* \brief Lookup key for subexpression replacements
+ *
+ * The lookup key must contain the expression being bound, along with
+ * the struct info used for a match cast, if applicable.  Using
+ * `MatchCast` with StructuralEqual and StructuralHash would be almost
+ * correct, but acts as a point of definition for symbolic variables
+ * within the output struct info.  As a result, it would erroneously
+ * de-duplicate `R.match_cast(A, R.Tensor([m,n]))` and
+ * `R.match_cast(A, R.Tensor([p,q]))`, even though they define
+ * different symbolic variables.
```
slyubomirsky, line 46: Does it still do the wrong thing even if you set `map_free_vars` to false?

src/relax/transform/eliminate_common_subexpr.cc
```diff
@@ -20,223 +20,162 @@
...
-  void VisitBinding_(const MatchCastNode* binding) override {
-    // no need to visit var def because the struct info isn't going to change
-    Expr new_value = RegisterBoundValue(binding->var, binding->value);
-
-    // re-emit old binding if nothing changes
-    if (new_value.same_as(binding->value)) {
-      builder_->EmitNormalized(GetRef<MatchCast>(binding));
+  Expr VisitExpr_(const IfNode* op) override {
+    Expr cond = VisitExpr(op->cond);
+    Expr true_branch = VisitWithCleanScope(op->true_branch);
```
slyubomirsky, line 154: I don't think visiting with a clean scope is the correct approach, since there are potential correct replacements that could come from the outer scope. I think the right approach would be to cache the current state and restore it after visiting.

src/relax/transform/eliminate_common_subexpr.cc
```diff
@@ -20,223 +20,162 @@
...
-    // no need to visit var def because the struct info isn't going to change
-    Expr new_value = RegisterBoundValue(binding->var, binding->value);
-
-    if (new_value.same_as(binding->value)) {
-      builder_->EmitNormalized(GetRef<VarBinding>(binding));
+    // If we have accumulated any state, visit the function in a fresh
+    // copy of the mutator, to avoid replacing a child-scope
+    // expression with a parent-scope binding, or vice versa.
+    if (expr_replacements_.size() || var_remap_.size()) {
+      return VisitWithCleanScope(GetRef<Expr>(op));
```
slyubomirsky, line 146: I'm not sure using a clean scope in all cases is necessarily, since an inner function could capture vars from the outer scope and there might be legitimate substitutions that are possible. That said, this may not necessarily be desirable behavior since it could result in bigger closures (capturing more vars than you might want). We would also have to make sure we don't capture any DataflowVars, since that's not permitted. Any thoughts?

tests/python/relax/test_transform_cse.py
```diff
@@ -90,6 +88,12 @@ def foo() -> R.Tuple(R.Tensor((), dtype="int32"), R.Tensor((2, 2), dtype="int32"
 
 
 def test_repeated_inner_tuples():
+    """CSE is only applied at variable bindings
+
+    To remain consistent with the behavior of the normalizer, tuples
+    are kept as-is, even if they contain repeated sub-tuples.
+    """
```
slyubomirsky, line 95: Good observation, probably a good choice.

slyubomirsky (changes requested):
The redesign of the CSE pass seems generally sound and behaves more nicely than my previous approach of overriding the top-level VisitExpr and listing exceptions. I listed a couple of questions and concerns resulting from the substantial rewrites.

src/relax/transform/eliminate_common_subexpr.cc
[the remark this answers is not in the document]
Lunderberg: I believe so, though for a different reason.  Even if `map_free_vars` is false, `SEqualReducer::DefEqual` can override it to true for some IR nodes.  This is the function that is used for `MatchCastNode::struct_info`, in order to support `StructuralEqual` in cases where two match cast nodes are defining different variables, but should compare as structural equal.

Effectively, the `MatchCastNode::SEqualReduce` is implemented to allow TIR variable definitions, which is the opposite of what we need for the de-duplication.

src/relax/transform/eliminate_common_subexpr.cc
```diff
@@ -20,223 +20,162 @@
...
- private:
-  std::unordered_map<Expr, int, StructuralHash, StructuralEqual> count_map_;
-  ImpurityDetector impurity_detector_;
-};
+    } else if (auto it = expr_replacements_.find(lookup_key); it != expr_replacements_.end()) {
+      VLOG(1) << "Value " << bound_value << " has previously been bound as " << it->second
+              << ".  The duplicate binding of this value to " << binding->var
+              << " will be replaced with a trivial binding, "
+              << "and occurrences of " << binding->var << " will be replaced with " << it->second;
+      output_binding = VarBinding(binding->var, it->second);
```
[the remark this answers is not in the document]
Lunderberg: This is to handle cases where the variable remapping would not be allowed, but the variable binding is still valid.  If we the de-duplication results in remapping a non-dataflow variable to a dataflow variable, that remapping is only valid within the dataflow block.  Outside of that dataflow block, the variable remaps are reset, and any remaining usages would still need to be definition.

I've added the following unit test to test this behavior.

```python
def test_do_not_eliminate_impure():
    @I.ir_module
    class Before:
        @R.function(pure=False)
        def foo(x: R.Tensor((2, 3), dtype="float32"), y: R.Tensor((2, 3), dtype="float32")):
            # it's a repeated subexpression but it would be wrong to deduplicate it
            p1 = R.print(format="Message")
            p2 = R.print(format="Message")
            a1 = R.assert_op(R.const(False), format="Always fails")
            lv0 = R.add(x, y)
            lv1 = R.add(x, y)
            gv = R.multiply(lv0, lv1)
            a2 = R.assert_op(R.const(False), format="Always fails")
            return gv

    @I.ir_module
    class Expected:
        @R.function(pure=False)
        def foo(x: R.Tensor((2, 3), dtype="float32"), y: R.Tensor((2, 3), dtype="float32")):
            p1 = R.print(format="Message")
            p2 = R.print(format="Message")
            a1 = R.assert_op(R.const(False), format="Always fails")
            lv0 = R.add(x, y)
            lv1 = lv0
            gv = R.multiply(lv0, lv0)
            a2 = R.assert_op(R.const(False), format="Always fails")
            return gv

    verify(Before, Expected)
```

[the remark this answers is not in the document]
Lunderberg: That said, writing that test made me realize a limitation in the implementation.  The non-dataflow var `D = R.add(x,y)` can't be de-duplicated as the dataflow var `A`, but it still could be de-duplicated as the non-dataflow var `B`.

I've updated the implementation to have `std::unordered_map<ReplacementKey, std::vector<Var>> expr_replacements_`, tracking all variables that could be de-duplicated as each other.  That way, when exiting a DataflowBlock, the `expr_replacements_` could still track non-dataflow variables that could be used for de-duplication.

```python
@I.ir_module
class Expected:
    @R.function
    def main(x: R.Tensor((2, 3), dtype="float32"), y: R.Tensor((2, 3), dtype="float32")):
        with R.dataflow():
            A = R.add(x, y)
            B = A
            C = R.multiply(A, A)
            R.output(B, C)

        D = B
        return (B, C, B)
```

This way, such a function would only require CSE -> Canonicalize, whereas before it would require CSE -> Canonicalize -> CSE.

src/relax/transform/eliminate_common_subexpr.cc
```diff
@@ -20,223 +20,162 @@
...
 
-class CommonSubexprEliminator : public ExprMutator {
- public:
-  explicit CommonSubexprEliminator(
-      std::unordered_map<Expr, int, StructuralHash, StructuralEqual> count_map,
-      bool call_only = false)
-      : count_map_(std::move(count_map)), call_only_(call_only) {}
+    } else {
+      VLOG(1) << "Value " << bound_value << " is bound to " << binding->var
+              << " and may be de-duplicated if it occurs again.";
```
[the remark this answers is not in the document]
Lunderberg: I think so, since any future debugging would be useful.  They are disabled by default, and require both a build flag `-DTVM_LOG_DEBUG=1` and a runtime environment variable `TVM_LOG_DEBUG=eliminate_common_subexpr.cc=1` to enable.

(Trying to get in the habit of leaving `VLOG(1)` statements around after longer debugging sessions, rather than just removing the `LOG(DEBUG)` statements.)

src/relax/transform/eliminate_common_subexpr.cc
Lunderberg: Hmm.  From a lowering standpoint, my main concern would be to avoid introducing a captured variable.  If a user has an inner function without closure variables, it could be surprising if CSE prevents it from being hoisted out of the local function.

(Selfishly, I have a few plans to simplify `LambdaLift` when I have the time.  Currently, `LambdaLift` handles lifting of inner functions regardless of whether there are closure variables, which is a bit tricky to track.  I'd like to split that out into a `HoistClosureVariablesToParams` and `HoistInnerFunctions` passes.  The first would provide explicit arguments for any closure variables, and the second would hoist the inner function out to the `IRModule`.  If CSE is applied in-between those two passes, I'd like to avoid having it re-introduce closure variables.)

src/relax/transform/eliminate_common_subexpr.cc
Lunderberg: Good call.  I've updated to have propagate bindings from before a `if/else` branch into the body of either branch.  This functionality has three new unit tests, to validate (1) bindings before a branch are de-duped inside the branch, (2) bindings within a branch are not de-duped after the branch, and (3) bindings within a branch are not de-duped into its sibling branch.

tests/python/relax/test_transform_cse.py
Lunderberg: Thank you.  With how much discussion there was around the normalization with respect to `R.call_tir`, I figured it's best to match as much as possible to the normalization.

src/relax/transform/eliminate_common_subexpr.cc
slyubomirsky: Yeah I think you're right that capturing more vars might be an unwelcome surprising behavior.

slyubomirsky (approved):
Thank you for addressing my concerns.

PR #16601, opened 2024-02-16
[Relax][Transform] Canonicalize `let var = R.const` bindings

Lunderberg:
Prior to this commit, known tuples could be unwrapped into variables, but any constants would remain.  This commit updates `CanonicalizeBindings` to unwrap tuples containing constants.


Hzfengsy (approved):
LGTM

PR #16602, opened 2024-02-16
[Transform] Improvements to LazyTransformParams

Lunderberg:
* Handle non-bundled parameters in LazyTransformParams.

* Check for `"num_input"` attribute

* Handle relax.Const in LazyTransformParams

  Prior to this commit, `LazyTransformParams` would only output a call to the `fset_item` function if that element of the output had a corresponding `relax.Binding`.  If `relax.Const` appeared in the output, then the call to `fset_item` would be omitted.

  This commit updates `LazyTransformParams` to check for any non-`Var` elements of the output tuple.

### DIFF
--- a/python/tvm/relax/transform/lazy_transform_params.py
+++ b/python/tvm/relax/transform/lazy_transform_params.py
@@ -132,20 +132,22 @@ class LazyTransformParamsFuncCreator:
         self.extra_get_item_params = extra_get_item_params
         self.fset_item = fset_item
         self.extra_set_item_params = extra_set_item_params
-        # the only input param, which should be a Tuple
-        self.input_tuple_param = None
         self.input_params_set = None
         self.out_tuple_map = None
         self.out_tuple_var = None
         self.memory_free_insertion = None
 
     def transform(self, func: relax.Function) -> relax.Function:
-        self.input_tuple_param = func.params[0]
+        if func.attrs is not None and "num_input" in func.attrs:
+            num_input = func.attrs["num_input"].value
+        else:
+            num_input = 0
+
         seq_expr = func.body
         self.out_tuple_var = seq_expr.body
 
         # Step 1. collect out_tuple_map and input_params_set
-        forward_collector = ForwardCollector(self.out_tuple_var, self.input_tuple_param)
+        forward_collector = ForwardCollector(self.out_tuple_var, func.params[num_input])
         forward_collector.visit_expr(func)
         self.out_tuple_map = forward_collector.out_tuple_map
         # input_params_set is the set of binding var for var = params[i]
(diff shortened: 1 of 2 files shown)


python/tvm/relax/transform/lazy_transform_params.py
```diff
@@ -157,24 +159,60 @@ def transform(self, func: relax.Function) -> relax.Function:
...
+                            [relax.ObjectStructInfo()],
+                        ),
+                    )
+                    for expr, indices in leaf_outputs.items()
+                    for index in indices
+                ]
+                new_body = relax.SeqExpr(
+                    [*new_body.blocks, relax.BindingBlock(new_bindings)], new_body.body
+                )
+
```
slyubomirsky, line 189: I presume these additions are for handling the non-var case mentioned in the description?

tests/python/relax/test_transform_lazy_transform_params.py
```diff
@@ -602,5 +602,77 @@ def main_transform_params() -> R.Tuple:
...
+            B_sharded = R.strided_slice(
+                B, axes=[1], begin=[rank * 8], end=[(rank + 1) * 8], assume_inbound=True
+            )
+
+            return (A_sharded, B_sharded)
+
+    After = LazyTransformParams(fset_item=None)(Before)
+    tvm.ir.assert_structural_equal(After, Expected)
+
+
```
slyubomirsky, line 676: Are there any test cases that make use of extra parameters for get_item and set_item? If it's not tested, it should be. If there also isn't a case of a non-var output (I'm not sure exactly what that should look like, as I haven't used this pass), that would be good to add too.

python/tvm/relax/transform/lazy_transform_params.py
Lunderberg: That's correct.  Otherwise, a `R.const(...)` that occurs within the output tuple wouldn't produce a call to the `fset_item` function.  I've added a comment to clarify.

python/tvm/relax/transform/lazy_transform_params.py
```diff
@@ -157,24 +159,60 @@ def transform(self, func: relax.Function) -> relax.Function:
...
 
         # Step 5. Find all shape parameters that should be retained as
         # parameters.
         symbolic_vars = relax.analysis.defined_symbolic_vars(func)
         if symbolic_vars:
+
+            def unpack_sinfo(sinfo):
+                if isinstance(sinfo, relax.TupleStructInfo):
+                    for field in sinfo.fields:
+                        yield from unpack_sinfo(field)
```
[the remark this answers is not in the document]
Lunderberg: Thank you, and I really like using `yield from` for providing flattened results from a graph structures.  This way, the result of `unpack_sinfo(top_level.struct_info)` can be collected into a list, but there aren't any temporary lists made along the way.

python/tvm/relax/transform/lazy_transform_params.py
```diff
@@ -157,24 +159,60 @@ def transform(self, func: relax.Function) -> relax.Function:
...
+                        relax.Var("_", relax.ObjectStructInfo()),
+                        relax.Call(
+                            relax.ExternFunc(self.fset_item),
+                            [*self.extra_set_item_params, index, expr],
+                            None,
+                            [relax.ObjectStructInfo()],
+                        ),
+                    )
+                    for expr, indices in leaf_outputs.items()
+                    for index in indices
```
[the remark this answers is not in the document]
Lunderberg: Absolutely.  I think I prefer the chained iterators used in Rust or C#, but Pythons list/dict/set comprehensions are a close second.

slyubomirsky (approved):
Thank you for the additional tests.

PR #16604, opened 2024-02-18
[Relax][Frontend][Onnx] fix clip unsqueeze opset implement

chengven027-intellif:
Hi， tvm:
  This PR fixes a bug related to different opsets in the clip and unsqueeze operations.
  My model's opset default version is 10, and there is no implementation for opset below 10 for these two operators. Therefore, there might be some issues.


tqchen (approved):

tqchen:
Thanks @chengven027-intellif !

PR #16610, opened 2024-02-19
[Doc] Fixed Docstring usage example in `tvm.ir.make_node`

felix-ro:
The provided usage example for `tvm.ir.make_node` has become outdated. Creating an IR node of type `IntImm` requires the field `span`. 

Compare: [`make_node` Unit Test](https://github.com/apache/tvm/blob/main/tests/python/ir/test_node_reflection.py#L73)

### DIFF
--- a/python/tvm/ir/attrs.py
+++ b/python/tvm/ir/attrs.py
@@ -153,7 +153,7 @@ def make_node(type_key, **kwargs):
 
     .. code-block:: python
 
-       x = tvm.ir.make_node("IntImm", dtype="int32", value=10)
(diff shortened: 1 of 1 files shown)


quic-sanirudh (approved):
LGTM, but is there a reason for the `tvm.runtime.String` addition. the FFI call should automatically handle if I'm not mistaken.

felix-ro:
> LGTM, but is there a reason for the `tvm.runtime.String` addition. the FFI call should automatically handle if I'm not mistaken.

I just checked the FFI call, and it is handled there. I have updated the pr accordingly.

PR #16611, opened 2024-02-19
[AOT][Testing] Print output values on test failure

lhutton1:
This commit enhances the AOT test harness to print the "actual" and "reference" values when there is a mismatch. This helps when debugging a failing test. Sample output:
```
Actual, Reference
8.502946, 8.887751
9.810405, 9.108611
8.563767, 9.041000
10.019511, 9.190888
....
```

### DIFF
--- a/tests/python/relay/aot/test_crt_aot.py
+++ b/tests/python/relay/aot/test_crt_aot.py
@@ -93,6 +93,7 @@ def test_conv_with_params(interface_api, use_unpacked_api, test_runner):
         test_runner,
         interface_api,
         use_unpacked_api,
+        print_output_on_mismatch=True,
     )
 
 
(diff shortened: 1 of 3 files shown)


ekalda:
Thanks @lhutton1, would you mind adding a test for this change? There is currently no path that exercises this option.

tests/python/relay/aot/test_crt_aot.py
```diff
@@ -93,6 +93,7 @@ def test_conv_with_params(interface_api, use_unpacked_api, test_runner):
         test_runner,
         interface_api,
         use_unpacked_api,
+        print_output_on_mismatch=True,
```
ekalda, line 96: What is the reason for setting that option for this specific test?

ekalda (commented):
Thanks @lhutton1, LGTM, there was just one thing that was puzzling me, but I think in general it is good to go!

ekalda (approved):
Thanks @lhutton1, LGTM!

PR #16612, opened 2024-02-20
[SVE] Add support for scalable data type strings

lhutton1:
This commit adds support for representing scalable vectors using the string data type format. For example, "float32xvscalex4" may be used to represent the following scalable type: `DataType(kDLFloat, 32, /*lanes=*/4, /*is_scalable=*/true)`.

Co-authored-by: Elen Kalda <elen.kalda@arm.com>
Co-authored-by: Neil Hickey <neil.hickey@arm.com>

### DIFF
--- a/include/tvm/runtime/data_type.h
+++ b/include/tvm/runtime/data_type.h
@@ -27,6 +27,7 @@
 #include <tvm/runtime/c_runtime_api.h>
 #include <tvm/runtime/logging.h>
 
+#include <cstring>
 #include <string>
 #include <type_traits>
 
@@ -110,7 +111,7 @@ class DataType {
     return -lanes_as_int;
   }
   /*! \return whether type is a scalar type. */
-  bool is_scalar() const { return lanes() == 1; }
+  bool is_scalar() const { return !is_scalable_vector() && lanes() == 1; }
   /*! \return whether type is a scalar type. */
   bool is_bool() const { return code() == DataType::kUInt && bits() == 1; }
   /*! \return whether type is a float type. */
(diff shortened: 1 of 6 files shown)


lhutton1:
@tvm-bot rerun

include/tvm/runtime/data_type.h
```diff
@@ -110,7 +111,7 @@ class DataType {
     return -lanes_as_int;
   }
   /*! \return whether type is a scalar type. */
-  bool is_scalar() const { return lanes() == 1; }
+  bool is_scalar() const { return !is_scalable_vector() && lanes() == 1; }
```
ekalda, line 114: I think it would make more sense to use `!is_scalable_or_fixed_length_vector()` here. The current version is going to throw an error if applied on a scalable vector.

ekalda (changes requested):
Thanks @lhutton1 LGTM in a general, just one comment :)

include/tvm/runtime/data_type.h
ekalda: Ok yes, thinking about it, this probably is the most solid way to do it, even though it reads a bit awkward 👍

ekalda:
Thanks @lhutton1!

tqchen:
there is a regression due to concurrent merge, cc @Lunderberg

lhutton1:
thanks for raising @tqchen, I'm taking a look now

lhutton1:
Apologies for missing this previously, I've uploaded #16649 which I'd expect to fix the failing test case

ekalda:
Any idea why the CI didn't catch this?

lhutton1:
CI was last run on the patch yesterday (26th Feb) while the PR that caught the issue was merged a week ago https://github.com/apache/tvm/pull/16563. I didn't rebase before pushing, but had expected CI to do this automatically.

I _think_ it might have something to do with a missing call to `merge_with_main()`([link](https://github.com/apache/tvm/blob/997adcaaead1714b108fc7401197113fbb80ca54/ci/jenkins/templates/utils/Prepare.groovy.j2#L47)) in the [unity_jenkinsfile.groovy](https://github.com/apache/tvm/blob/a15fd1d6a21e9bedb622ffb9fd025c50624d734d/ci/jenkins/unity_jenkinsfile.groovy) file since the other jenkinsfiles used in CI make calls to this, which presumably updates the PR branch before CI is run

Issue #14869, opened 2023-05-17
[Bug][Relay] cannot squeeze axis with dimension not equal to 1 at Keras frontend


jikechao:
For the LSTM below, when `batch_size != 1` (i.e., the size of first dimension input), compiling will lead to an unexpected crash and throw `Check failed: *axis_ptr == 1 (2 vs. 1) : cannot squeeze axis with dimension not equal to 1`
Notice that, when `batch_size ==1`, TVM can run well. 

![image](https://github.com/apache/tvm/assets/29506758/7bd450b2-92e6-4765-b993-6be255cc3753)

### Question: 
For the LSTM model, why batch_size !=1 leads to a crash? 

Compile successfully

### Actual behavior
```
Traceback (most recent call last):
  File "test.py", line 18, in <module>
    model = relay.build_module.create_executor("vm", mod, tvm.cpu(0), 'llvm', params).evaluate()
  File "/workplace/software/tvm/tvm/python/tvm/relay/backend/interpreter.py", line 171, in evaluate
    return self._make_executor()
  File "/workplace/software/tvm/tvm/python/tvm/relay/backend/vm.py", line 219, in _make_executor
    self.executable = compile(self.mod, self.target)
  File "/workplace/software/tvm/tvm/python/tvm/relay/backend/vm.py", line 67, in compile
    compiler.lower(mod, target, target_host)
  File "/workplace/software/tvm/tvm/python/tvm/relay/backend/vm.py", line 126, in lower
    self._lower(mod, raw_targets)
  File "/workplace/software/tvm/tvm/python/tvm/_ffi/_ctypes/packed_func.py", line 237, in __call__
    raise get_last_ffi_error()
tvm._ffi.base.TVMError: Traceback (most recent call last):
  15: TVMFuncCall
  14: tvm::runtime::PackedFuncObj::Extractor<tvm::runtime::PackedFuncSubObj<tvm::relay::vm::VMCompiler::GetFunction(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, tvm::runtime::ObjectPtr<tvm::runtime::Object> const&)::$_0> >::Call(tvm::runtime::PackedFuncObj const*, tvm::runtime::TVMArgs, tvm::runtime::TVMRetValue*)
  13: tvm::relay::vm::VMCompiler::Lower(tvm::IRModule, tvm::runtime::Array<tvm::Target, void> const&)
[...]

PR #16615, opened 2024-02-21
[Relay] fix a bug about convert_lstm in keras.py

jikechao:
Fix the bug mentioned in the issue:https://github.com/apache/tvm/issues/14869

cc @echuraev


echuraev (commented):
Please, add unit tests

PR #16618, opened 2024-02-21
[Disco] Propagate structlog configuration to disco workers

Lunderberg:
Prior to this commit, `structlog.configure(...)` would only impact log statements generated in the main process, while workers started with `disco.session.ProcessSession` would not inherit the `structlog` configuration.  While `disco.session.ThreadedSession` would inherit the `structlog` configuration, it would also inherit process-specific CUDA variables.

This commit updates `disco.session.ProcessSession` to explicitly propagate any `structlog` configuration to child processes.  This implementation intentionally avoids introducing a new dependency for TVM.  If the `structlog` package is not available, the config propagation is skipped.

### DIFF
--- a/python/tvm/runtime/disco/session.py
+++ b/python/tvm/runtime/disco/session.py
@@ -17,6 +17,11 @@
 """This module defines a Session in Disco. Session is the primary interface that users interact
 with the distributed runtime.
 """
+
+import os
+import pickle
+
+
 from typing import Any, Callable, Optional, Sequence, Union
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


python/tvm/runtime/__init__.py
```diff
@@ -40,3 +40,4 @@
 )
 
 from . import executor
+from . import disco
```
Hzfengsy, line 43: Not sure if it's safe when runtime is not compiled with disco.

Lunderberg: Hmm, good point.  The `tvm._ffi.register_object("runtime.disco.*")` calls would raise an error if disco isn't available.

This was primarily to ensure that all packaged funcs generated by `tvm.runtime.disco` were available within the disco sub-process.  I'll try a build without disco enabled, and see what is required to make sure that the module import works either way.

masahi (approved):

PR #16619, opened 2024-02-21
[Bugfix][Cutlass] Check if function attributes is None

Lunderberg:
This commit updates the cutlass annotator to check if `relax.Function.attrs` is `None` before attempting to access it.


Lunderberg:
Unfortunately, I was unable to make a standalone unit test that would trigger this bug.  For the long term, I plan to change the default of `relax.Function.attrs` from `None` to an empty dictionary.  As it is, having `None` as the function attributes and having `{}` as the function attributes are semantically equivalent, but require an extra check at each usage.

masahi (approved):

Issue #16623, opened 2024-02-22
[Bug] Wrong default type in the Multinomial operaotor converter


jikechao:
According to the description on the [ONNX website](https://onnx.ai/onnx/operators/onnx__Multinomial.html), the default value of type is int32 rather than int64.

The TVM ONNX frontend mistakenly assigns the dtype as int64.
![image](https://github.com/apache/tvm/assets/29506758/28573fe0-2d64-4274-baf2-c547d7e012c3)

PR #16624, opened 2024-02-22
[Relay][ONNX] fix the wrong default value about dtype in Multinomial converter

jikechao:
Fix the bug: https://github.com/apache/tvm/issues/16623
the default type of attribute 'dtype' is int32 rather than int64


jikechao:
The bug can not produce a significant wrong result. Thus, I didn't add the unit test.

jikechao:
cc @echuraev @Hzfengsy

tqchen (approved):

PR #16628, opened 2024-02-22
[RUST] extend rust bindings for relay

cgerum:
It especially adds support to generate relay code from rust programmatically. 

The biggest part of these pull request are bindings to the packed funcs from `relay.op._make` also used for constructing relay from python. 

It further fixes some missing attributes in the relay attribute bindings. 

@gussmith23  @jroesch @lunderberg could you have a look?


tqchen:
Thanks @cgerum . Would also like to let you know that as a community we are moving towards using relax and TVM unity for [emerging needs like LLMs](https://discuss.tvm.apache.org/t/discuss-tvm-core-strategy-for-emerging-needs/15751), so there might not be a lot of effort further developing relay and it will remain mostly in maintain mode for now,  feel free to checkout latest relax related developed as they lands in main recently

cgerum:
@tqchen yes I am aware of TVM unity and relay, but unfortunately our projects are currently still stuck with relay. I will definitly try to find some time to also work on Rust bindings for relax.

PR #16638, opened 2024-02-23
[CMAKE][CUTLASS] Improve dependency management 

csullivan:
Improve dependency management with different cutlass versions.
 * Each cutlass-based submodule library now uses its own cutlass submodule dependency
 * TVM's cutlass submodule is decoupled from others and is bumped to v3.4.1 for H100 support
 * Add scaffold for new cutlass fp8 dequant gemm interface targeting TVM's cutlass submodule

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -368,6 +368,7 @@ tvm_file_glob(GLOB RUNTIME_SRCS
   src/runtime/minrpc/*.cc
   src/runtime/relax_vm/*.cc
 )
+set(TVM_RUNTIME_EXT_OBJS "")
 
 if(BUILD_FOR_HEXAGON)
   if(NOT BUILD_STATIC_RUNTIME)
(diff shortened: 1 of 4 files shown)


cmake/modules/contrib/CUTLASS.cmake
```diff
@@ -16,16 +16,61 @@
...
 
-  set(CUTLASS_DIR ${PROJECT_SOURCE_DIR}/3rdparty/cutlass)
+  ### Build cutlass runtime objects for fpA_intB_gemm using its cutlass submodule
   add_subdirectory(${PROJECT_SOURCE_DIR}/3rdparty/cutlass_fpA_intB_gemm)
+  target_include_directories(fpA_intB_gemm PRIVATE
+    ${PROJECT_SOURCE_DIR}/3rdparty/cutlass_fpA_intB_gemm
+    ${PROJECT_SOURCE_DIR}/3rdparty/cutlass_fpA_intB_gemm/cutlass/include
+  )
+  set(CUTLASS_FPA_INTB_RUNTIME_SRCS "")
+  list(APPEND CUTLASS_FPA_INTB_RUNTIME_SRCS src/runtime/contrib/cutlass/moe_gemm.cc)
```
vinx13, line 38: Is it needed? It’s already included in the submodule https://github.com/tlc-pack/cutlass_fpA_intB_gemm/blob/main/cutlass_kernels/CMakeLists.txt

CMakeLists.txt
```diff
@@ -594,26 +595,44 @@ add_library(tvm_libinfo_objs OBJECT ${LIBINFO_FILE})
 
 include(GNUInstallDirs)
 if(NOT BUILD_DUMMY_LIBTVM)
-  add_library(tvm SHARED $<TARGET_OBJECTS:tvm_objs> $<TARGET_OBJECTS:tvm_runtime_objs> $<TARGET_OBJECTS:tvm_libinfo_objs>)
+  add_library(tvm SHARED 
+    $<TARGET_OBJECTS:tvm_objs> 
+    $<TARGET_OBJECTS:tvm_runtime_objs> 
+    $<TARGET_OBJECTS:tvm_libinfo_objs>
+    ${CUTLASS_RUNTIME_OBJS}
```
vinx13, line 602: use `TVM_RUNTIME_EXT_OBJS`

cmake/modules/contrib/CUTLASS.cmake
csullivan: I see. I originally made my change against b1d7caac82c1cd6a9d142974f5e6dcbeff23a815 which has these src/runtime/contrib/cutlass/ sources (flash_decoding.cu, moe_gemm.cc), if these are duplicated now in the submodules and can be deleted then I can do that here. Though in my cursory read, I don't see that these are the same source files. Nonetheless, we can delete them if they are not needed. Let me know. https://github.com/apache/tvm/blob/b1d7caac82c1cd6a9d142974f5e6dcbeff23a815/cmake/modules/contrib/CUTLASS.cmake#L31-L33

vinx13:
There are some lint errors on CI

PR #16640, opened 2024-02-23
[TVMScript] Represent tir::builtin::ret() using python "return"

Lunderberg:
The TIR equivalent of python's `return` statement is the `tir::builtin::ret()` operator.  Prior to this commit, this was printed as `T.ret(value)`, and any use of `return` statement in TVMScript produced an error while parsing.

This commit updates the TVMScript parsing to produce `T.ret(value)` for the python `return value` statement.  If syntax sugar is enabled, the TIR `T.ret(value)` will be printed as `return value`.


slyubomirsky (approved):
A nice parsing improvement :slightly_smiling_face:

PR #16641, opened 2024-02-23
[TVMScript] Allow use of relax.Expr with void type as a statement

Lunderberg:
Prior to this commit, TVMScript required all relax expressions to be part of an explicit assignment or return statement.  While this matches the structure of the C++ IR types, this can be unexpected for functions that have no return value.  For example, needing to assign the result of `R.print(...)` to a variable.

This commit updates the TVMScript parser/printer to allow relax expressions to be used as statements, if they have a void return type. This allows use of `R.print(...)` and `R.assert_op(...)` to be called without assigning the result to an unused variable.

### DIFF
--- a/python/tvm/script/parser/relax/parser.py
+++ b/python/tvm/script/parser/relax/parser.py
@@ -274,7 +274,21 @@ def post_visit_local_function(self: Parser, node: doc.Expr) -> None:
 @dispatch.register(token="relax", type_name="Expr")
 def visit_expr_stmt(self: Parser, node: doc.Expr) -> None:
     value = self.eval_expr(node.value)
-    if value is not None:
+    if isinstance(value, relax.Expr):
+        var = R.emit(value)
+        IRBuilder.name("_", var)
+        is_void_value = (
+            isinstance(var.struct_info, relax.TupleStructInfo) and len(var.struct_info.fields) == 0
+        )
+
+        if not is_void_value:
+            self.report_error(
+                node,
+                f"Non-void relax expressions must be bound to a variable, "
+                f"but expression of type {var.struct_info} was used as a statement.",
+            )
+
+    elif value is not None:
         self.report_error(node, f"Unsupported Expr stmt type {value}.")
 
 
--- a/src/script/printer/relax/binding.cc
+++ b/src/script/printer/relax/binding.cc
@@ -69,6 +69,24 @@ TVM_STATIC_IR_FUNCTOR(IRDocsifier, vtable)
             Doc ret = d->AsDoc(n->value, n_p->Attr("value"));
             d->cfg->binding_names.pop_back();
             return ret;
+
+            // Uncommenting this section hides the variable binding
+            // when the StructInfo is void.  For example, printing
+            // `R.assert_op(expr)` instead of `_ = R.assert_op(expr)`.
+            // However, Relax represents void values as an empty
+            // tuple, and a void-type variable may still be used later
+            // in the function.  Hiding bindings of these void-type
+            // variables would result in use of an undefined variable.
+            //
+            // TODO(Lunderberg): Inline void-type variable to use
+            // `R.tuple()` during normalization.  This will avoid the
+            // cases that trigger the undefined variables, and allow
+            // this syntax sugar to be enabled.
+            //
+            // } else if (d->cfg->syntax_sugar && relax::HasVoidStructInfo(n->value) &&
+            //            relax::HasVoidStructInfo(n->var)) {
+            //   ExprDoc rhs = d->AsDoc<ExprDoc>(n->value, n_p->Attr("value"));
+            //   return ExprStmtDoc(rhs);
           } else {
             ExprDoc rhs = d->AsDoc<ExprDoc>(n->value, n_p->Attr("value"));
             Optional<ExprDoc> ann = StructInfoAsAnn(n->var, n_p->Attr("var"), d, n->value);
--- a/tests/python/relax/test_tvmscript_parser.py
+++ b/tests/python/relax/test_tvmscript_parser.py
@@ -1824,6 +1824,77 @@ def test_parse_multiple_pure_and_impure_funcs():
     _check(Mixture)
 
 
+def test_function_with_void_return_type_may_be_used_as_statements():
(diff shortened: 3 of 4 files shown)


python/tvm/script/parser/relax/parser.py
```diff
@@ -274,7 +274,21 @@ def post_visit_local_function(self: Parser, node: doc.Expr) -> None:
...
+        is_void_value = (
+            isinstance(var.struct_info, relax.TupleStructInfo) and len(var.struct_info.fields) == 0
+        )
+
+        if not is_void_value:
+            self.report_error(
+                node,
+                f"Non-void relax expressions must be bound to a variable, "
+                f"but expression of type {var.struct_info} was used as a statement.",
+            )
```
slyubomirsky, line 289: I wonder if we should even have this as a rule. Why not let users evaluate expressions without binding them regardless of their return type?

slyubomirsky (commented):
Thank you for making this PR, definitely a great quality-of-life addition. I think there's a test case we should check for this: Does this intersect with the quirky parsing for if-else? For example, if the value returned in an if-else is of void type. Would it be safe not to write out the return var? Would it still roundtrip?

python/tvm/script/parser/relax/parser.py
Lunderberg: At the moment, because I wanted to make the minimal change that would support common cases.  I think it would be good to remove the restriction altogether, but for the first step, I wanted to make the restriction be explicit.

There's a couple of concerns I could see with allowing non-void return value to be implicitly ignored.

* Prevent accidentally unused values.  If two IRModule instances An in-place operator that performs `a = a + b` may be represented as `cls.add(a, b)`.  This woul
* Round-trip TVMScript -> Relax -> TVMScript without a pre-processing pass.  Checking if a value has void type can done while printing the IR.  Checking whether a non-void variable could be omitted would require a pre-processing step to find any downstream users.

I don't think either of those are definitive arguments, but I figured I'd handle the unambiguous beneficial cases first, with a follow-up PR to relax the restriction.

Lunderberg:
> Does this intersect with the quirky parsing for if-else? For example, if the value returned in an if-else is of void type. Would it be safe not to write out the return var? Would it still roundtrip?

Good call on a test to add.  Looks like this change does cause an issue with round-trips when the elided binding is the last binding in an if/else block.  I've added a currently-failing unit test for the round-trip.

Lunderberg:
And after poking at it, it looks like there an ambiguity in the relax type system.  In most cases, a zero-field `TupleStructInfo` is used to represent a void type ([example](https://github.com/apache/tvm/blob/main/include/tvm/relax/struct_info.h#L457)).  However, in some cases, a zero-field `TupleStructInfo` is used to represent, well, a zero-field tuple ([example](https://github.com/apache/tvm/blob/main/tests/python/relax/test_tvmscript_parser.py#L1369)).  Eliding the variable binding for an actual void type make sense, as there are no valid uses of a void type.  However, a zero-field tuple can be treated as an object, and so removing its variable binding may result in undefined usage.

slyubomirsky:
Well, I wouldn't call that's not a type system ambiguity, but it could become one in terms of parser roundtripping. It's just that an empty tuple (unit value) cannot contain anything and is immutable, so its use is historically like a null value in functional languages. We do have the odd quirk that we have both the `null_value` operator that returns a void value and also the unit value. Personally, I think it would be better to use a unit value in all cases since there would be no reason to have an empty tuple at all.

I'm not sure what our stance would be on parser roundtripping: I'm fine with saying that is _permitted_ to omit the binding and that we should go with one way or the other for parser roundtripping.

python/tvm/script/parser/relax/parser.py
slyubomirsky: Those are good points. For the first case, we could have a warning (as C and other languages do with the right settings) for ignoring a return value. The second one is an interesting issue. I think it suggests that expecting an exact textual match for the parser roundtripping is too strict of a criterion for this situation, since a "statement" can always be written as `_ = ...` and it would be a choice as to whether to write it that way or use the friendlier syntax. It would make it harder to write automatic tests, true. For a systematic solution, maybe we could formalize the idea of a desugaring step for testing purposes?

Lunderberg:
True.  I'd see the mismatch as the degree to which relax should be compatible with the python environment.  For lisp, I'd expect `nil` to be the empty tuple, but for Python, I'd expect `None` to be distinct.from the empty tuple.

>I'm not sure what our stance would be on parser roundtripping: I'm fine with saying that is permitted to omit the binding and that we should go with one way or the other for parser roundtripping.

Unfortunately, this runs a little bit deeper.  Since a size-zero tuple may be used at a later point in the relax function, omitting the binding requires inspecting the function to see if that binding is being used.  Even with something like `dummy_var: R.Tuple() = R.print(...)`, the `R.Tuple()` struct info is insufficient to know if `dummy_var` is used later in the function.

slyubomirsky:
Ah I see, we don't really want the pretty printer to have to be checking if a var is later used. We could take the policy of always inlining a unit tuple value (or always doing it in the pretty printer), as this brings no harm. We could even have the normalizer do it.

Lunderberg:
> We could even have the normalizer do it.

I like this as a long-term solution.  I've updated this PR to only change the TVMScript parsing, and not the printing, along with disabling the related unit tests.  That way, they can be re-enabled in a follow-up PR.

Lunderberg:
Rebased onto main, as the CI failures seemed unrelated to this PR, and `main` has had some CI fixes recently.

Lunderberg:
All CI tests passing, ready for review/merge.

slyubomirsky (approved):
Thanks for responding to feedback. This will be a nice bit of syntactic sugar and along with #16658 will be a substantial improvement.

Edit: In principle, I'm in favor of always allowing a user to omit the var and giving a warning rather than an error if they do it for a statement with a non-void return. We can consider that as a later change if others are in favor.

Lunderberg:
> In principle, I'm in favor of always allowing a user to omit the var and giving a warning rather than an error if they do it for a statement with a non-void return. We can consider that as a later change if others are in favor.

I agree, and would like to add this in a later PR.

PR #16642, opened 2024-02-23
[Relax] Allow R.Prim('bool') in relax::If and assert_op 

Lunderberg:
Prior to this commit, the condition used for `relax::If` node and the `"relax.assert_op"` operator was required to be a scalar tensor.  This made it difficult to alter behavior based on a runtime shape parameter.  For example, delegating to a vectorized implementation based on a whether a tensor shape is divisible by the vector size.

This commit adds support for expressions of type `R.Prim('bool')` as the conditional for `relax::If` and `"relax.assert_op"`, to allow these use cases.

### DIFF
--- a/python/tvm/relax/transform/transform.py
+++ b/python/tvm/relax/transform/transform.py
@@ -489,9 +489,18 @@ def KillAfterLastUse() -> tvm.ir.transform.Pass:
 def ComputePrimValue() -> tvm.ir.transform.Pass:
     """Compute all R.prim_value instances
 
+    While high-level relax can include expressions in terms of its
+    symbolic variables, these expressions cannot natively be computed
+    within relax.  In order to provide values for symbolic expressions
+    (e.g. `R.prim_value(N*N)`, where `N` is a symbolic variable), this
+    pass generates a PrimFunc in which the expression can be computed.
+    The relax graph is then updated to include a call to that
+    PrimFunc, in place of the original `R.prim_value(expr)`.
+
     Returns
     -------
     ret : tvm.ir.transform.Pass
+
     """
     return _ffi_api.ComputePrimValue()  # type: ignore
 
--- a/python/tvm/relax/transform/transform.py
+++ b/python/tvm/relax/transform/transform.py
@@ -486,6 +486,16 @@ def KillAfterLastUse() -> tvm.ir.transform.Pass:
     return _ffi_api.KillAfterLastUse()  # type: ignore
 
 
+def ComputePrimValue() -> tvm.ir.transform.Pass:
+    """Compute all R.prim_value instances
+
+    Returns
+    -------
+    ret : tvm.ir.transform.Pass
+    """
+    return _ffi_api.ComputePrimValue()  # type: ignore
+
+
 def VMBuiltinLower() -> tvm.ir.transform.Pass:
     """Lowering generic intrinsic to VM intrinsics.
 
--- a/src/relax/transform/compute_prim_value.cc
+++ b/src/relax/transform/compute_prim_value.cc
@@ -1,0 +1,94 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <tvm/relax/expr_functor.h>
+#include <tvm/relax/transform.h>
+#include <tvm/tir/analysis.h>
+#include <tvm/tir/builtin.h>
+#include <tvm/tir/stmt_functor.h>
+
+namespace tvm {
+namespace relax {
+
+namespace {
+
+class PrimValueComputeInjector : public ExprMutator {
+ public:
+  IRModule Finalize() const { return builder_->Finalize(); }
+
+  using ExprMutator::VisitExpr_;
+
+  Expr VisitExpr_(const PrimValueNode* op) override {
+    auto node = Downcast<PrimValue>(ExprMutator::VisitExpr_(op));
+
+    if (node->value->IsInstance<tir::IntImmNode>() || node->value->IsInstance<tir::VarNode>()) {
+      return node;
+    }
+
+    auto ret_dtype = node->value->dtype;
+    auto param_vars = tir::UndefinedVars(node->value);
+    tir::Stmt body = tir::Evaluate(tir::Call(ret_dtype, tir::builtin::ret(), {node->value}));
+
+    tir::PrimFunc func(param_vars, body, PrimType(ret_dtype));
+    func = tir::RenewDefs(func);
+
+    auto callee = builder_->AddFunction(func, "compute_symbolic_expr");
+
+    return relax::Call(callee, param_vars.Map([](const tir::Var& tir_var) -> relax::Expr {
+      return relax::PrimValue(tir_var);
+    }));
+  }
+};
+
+}  // namespace
+
+namespace transform {
+
+Pass ComputePrimValue() {
+  runtime::TypedPackedFunc<IRModule(IRModule, PassContext)> pass_func =
+      [=](IRModule mod, PassContext pc) -> IRModule {
+    PrimValueComputeInjector mutator;
+
+    IRModule updates;
+    for (const auto& [gvar, base_func] : mod->functions) {
+      if (auto func = base_func.as<Function>()) {
+        auto updated = Downcast<Function>(mutator(func.value()));
+        if (!updates.same_as(base_func)) {
+          updates->Add(gvar, updated);
+        }
+      }
+    }
+
+    if (updates->functions.size()) {
+      auto write_ptr = mod.CopyOnWrite();
+      write_ptr->Update(updates);
+      write_ptr->Update(mutator.Finalize());
+    }
+
+    return mod;
+  };
+  return CreateModulePass(pass_func, 0, "ComputePrimValue", {});
+}
+
+TVM_REGISTER_GLOBAL("relax.transform.ComputePrimValue").set_body_typed(ComputePrimValue);
+
+}  // namespace transform
+
+}  // namespace relax
+}  // namespace tvm
--- a/src/tir/ir/function.cc
+++ b/src/tir/ir/function.cc
@@ -21,12 +21,52 @@
  * \file src/tir/ir/function.cc
  * \brief The function data structure.
  */
+#include <tvm/relax/struct_info.h>
 #include <tvm/runtime/registry.h>
+#include <tvm/tir/analysis.h>
 #include <tvm/tir/function.h>
 #include <tvm/tir/op.h>
 
 namespace tvm {
 namespace tir {
+namespace {
+relax::StructInfo InferStructInfo(const PrimFunc& prim_func) {
+  Array<relax::StructInfo> params;
+  for (const auto& param : prim_func->params) {
+    relax::StructInfo param_sinfo = [&]() -> relax::StructInfo {
+      if (auto opt_buf = prim_func->buffer_map.Get(param)) {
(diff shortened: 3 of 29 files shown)


python/tvm/relax/transform/transform.py
```diff
@@ -463,6 +463,16 @@ def KillAfterLastUse() -> tvm.ir.transform.Pass:
     return _ffi_api.KillAfterLastUse()  # type: ignore
 
 
+def ComputePrimValue() -> tvm.ir.transform.Pass:
+    """Compute all R.prim_value instances
```
slyubomirsky, line 467: I think this description should be more precise. I assume it's supposed to come late in the phase ordering since it inserts direct calls to PrimFuncs? (And so should probably come after we end purity checking?)

src/relax/transform/compute_prim_value.cc
```diff
@@ -0,0 +1,94 @@
...
+
+  Expr VisitExpr_(const PrimValueNode* op) override {
+    auto node = Downcast<PrimValue>(ExprMutator::VisitExpr_(op));
+
+    if (node->value->IsInstance<tir::IntImmNode>() || node->value->IsInstance<tir::VarNode>()) {
+      return node;
+    }
+
+    auto ret_dtype = node->value->dtype;
+    auto param_vars = tir::UndefinedVars(node->value);
```
slyubomirsky, line 45: Would this call know which TIR vars are in scope per the Relax scoping rules?

python/tvm/relax/transform/transform.py
Lunderberg: Good point on improving the docstring.

Regarding phase ordering, I don’t think we need to restrict its usage.  The calls to PrimFunc instances are valid in user-provided Relax functions, so this could occur early in the phase ordering.  The only limitation is that it must occur before `VMShapeLower`, as `VMShapeLower` expects all `R.prim_value(arg)` expressions to have int64 arguments.

src/relax/transform/compute_prim_value.cc
Lunderberg: This call isn’t aware of the Relax scoping rules, but I don’t think there’s a benefit of checking it at this point.  Any well-formed input that only uses in-scope TIR variables would produce well-formed output.  Any ill-formed input that uses out-of-scope TIR variables would produce ill-formed output that still uses the out-of-scope TIR variables.

Validating the relax scoping rules at this point would require additional tracking the in-scope variables, which would duplicate the functionality of the well-formed checker.  Since this pass wouldn’t make any ill-formed usage worse (and therefore harder to debug), I don’t think it’s worth duplicating the in-scope tracking here.

python/tvm/relax/transform/transform.py
slyubomirsky: Yes, but inserting the PrimFunc calls will likely change the purity of the functions where that happens. `call_tir` could be used to avoid that but then that will require using this before lowering `call_tir`.

src/relax/transform/compute_prim_value.cc
slyubomirsky: Yeah I think you're right that this would still work out just fine in that case.

python/tvm/relax/transform/transform.py
Lunderberg: Hmm.  I think this is another argument in favor of allowing `FuncStructInfo` annotations for `PrimFunc` objects, as that would allow the generated `PrimFunc` instances to be marked as pure functions.  I'll add a unit test to see how well that works for maintaining purity tracking when calling a pure PrimFunc from a pure Relax function.

slyubomirsky: Are they truly pure? No modification of external values?

Lunderberg: I was thinking an even simpler heuristic: All PrimFuncs are impure, unless explicitly annotated otherwise.  In this case, since the functions are being generated in a manner that requires purity, it could also provide the annotation.

For long-term, agreed, it would be good to have the TIR-level purity analysis.  I think I'd weaken the condition you mentioned slightly: A function is impure if it writes to a buffer that it didn't itself allocate.

slyubomirsky: Good point. I'm fine with requiring an annotation since the great majority of PrimFuncs are going to be impure.

tests/python/relax/test_transform_compute_prim_value.py
```diff
@@ -76,5 +76,29 @@ def compute_symbolic_expr(N: T.int64) -> T.bool:
...
+        @R.function
+        def main(_N: R.Prim(value="N"), _M: R.Prim(value="M")) -> R.Prim(value="N*M"):
+            N = T.int64()
+            M = T.int64()
+            out = Expected.compute_symbolic_expr(R.prim_value(N), R.prim_value(M))
+            return out
+
+        @T.prim_func(private=True)
+        def compute_symbolic_expr(N: T.int64, M: T.int64) -> T.int64:
+            T.ret(N * M)
```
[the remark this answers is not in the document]
Lunderberg: Good point.  The test case only passes right now because the `PrimFuncNode::struct_info_`, if it exists, isn't checked for structural equality.

src/relax/transform/compute_prim_value.cc
Lunderberg: Probably overkill, but I think it ended up being simpler to just generate the appropriate `FuncStructInfo` for a PrimFunc on construction, and to populate the currently-empty `struct_info_` field.  This includes inspecting the body to see if the PrimFunc is pure.

slyubomirsky:
Excellent, I'm glad to see an analysis for PrimFunc purity. Do you think it's necessary to have an override to assert one is pure? It's doubtful that we would need it right now but it might be a good thing to have in the back of our pocket (i.e., an escape hatch). I think it could be an analogue to `force_pure` in Relax (a special attribute).

Lunderberg:
> Do you think it's necessary to have an override to assert a PrimFunc is pure?

I think I prefer having the overrides at an individual operator level, rather than on the function as a whole.  So, a user could use `T.call_pure_packed` instead of `T.call_packed`, in order to make the overall function be marked as pure.  Especially when inlining/extracting regions of a function (e.g. `SplitHostDevice`), the top-down annotations can be ambiguous.  Relying on bottom-up annotations allows each extracted region to have correct annotations, without requiring the developer to explicitly propagate function-level annotations.

slyubomirsky:
That's a good idea. What would be the analogue to such annotations in a PrimFunc? Do we think we need it?

Lunderberg:
> That's a good idea. What would be the analogue to such annotations in a PrimFunc? Do we think we need it?

(Apologies, missed the notification on this one.)  For a `PrimFunc`, it already has a distinction between `T.call_extern` and `T.call_pure_extern`.  Looks like a PrimFunc only has `T.call_packed`, and doesn't have a corresponding `T.call_pure_packed`.  However, as these are largely for internal functions, where the packed interface isn't necessary, and `T.call_pure_extern` would be sufficient.

Lunderberg:
This PR needs an additional test case before merging.  On the main branch, the following IRModule is valid.  However, with this PR, it fails during parsing.

```python
@I.ir_module
class Module:
    @R.function
    def main(A: R.Tensor) -> R.Prim("bool"):
        return Module.is_bfloat16_dtype(A)

    @T.prim_func(private=True)
    def is_bfloat16_dtype(tensor: T.handle) -> T.bool:
        T.func_attr({"tir.is_scheduled": True, "tir.is_host_func": True})

        # From #include <tvm/tir/builtin.h>
        kArrTypeCode = T.meta_var(5)
        kArrTypeBits = T.meta_var(6)
        kArrTypeLanes = T.meta_var(7)

        # From #include <dlpack/dlpack.h>
        kDLBfloat = T.meta_var(4)

        type_code = T.tvm_struct_get(tensor, 0, kArrTypeCode, dtype="uint8")
        type_bits = T.tvm_struct_get(tensor, 0, kArrTypeBits, dtype="uint8")
        type_lanes = T.tvm_struct_get(tensor, 0, kArrTypeLanes, dtype="uint16")

        is_bfloat16: T.bool = (
            (type_code == kDLBfloat) and (type_bits == 16) and (type_lanes == 1)
        )
        return is_bfloat16
```

The failure occurs due to the inferred struct info of `R.Callable([R.Prim("handle")], R.Prim("bool"))`.  When the function is passed a `R.Tensor`, it fails relax's type check as `R.Tensor` is not compatible with `R.Prim("handle")`.

I think the fix will be to update `StructInfoBaseCheck` to recognize that `R.Prim("handle")` is a valid TIR representation of a `DLTensor*`, but I want to think on it first.

Lunderberg:
The additional test case has been added, and fixed, so this PR is ready for review/merge.

src/tir/ir/function.cc
```diff
@@ -21,12 +21,52 @@
...
+  for (const auto& param : prim_func->params) {
+    relax::StructInfo param_sinfo = [&]() -> relax::StructInfo {
+      if (auto opt_buf = prim_func->buffer_map.Get(param)) {
+        auto buf = opt_buf.value();
+        relax::ShapeExpr shape(
+            buf->shape.Map([](PrimExpr dim) { return cast(DataType::Int(64), dim); }));
+        return relax::TensorStructInfo(shape, buf->dtype);
+      }
+
+      if (auto prim_type = param->type_annotation.as<PrimTypeNode>();
```
slyubomirsky, line 44: By my understanding from the TIR spec, some Handle-typed vars sound like the would correspond more to Objects in Relax. I.e., void-typed values that are completely opaque

tests/python/relax/test_transform_compute_prim_value.py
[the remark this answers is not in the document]
slyubomirsky: Now that purity checking is implemented for PrimFuncs, this will roundtrip, correct?

slyubomirsky:
Thank you for your changes, I am very excited to have purity checking and more accurate StructInfo for PrimFuncs. I had one question remaining about the inference for argument types, see my comment on those lines.

src/tir/ir/function.cc
Lunderberg: That's a good call, and I've updated it to provide `R.Object` instead of `R.Tensor` in this case.  My main goal with this conditional was to avoid having `DLTensor*` parameters be labeled with `R.Prim(dtype='handle')`, as that would be incompatible with `R.Tensor` arguments.

In the future, the annotations could be improved by inspecting the PrimFunc body.  If the handle-typed `tir.Var` is used in a context that requires a `DLTensor*` argument (typically `builtin::tvm_struct_get`), then the annotation could be improved from `R.Object` to `R.Tensor`.  That would allow better type-checking at the Relax callsite, but isn't necessary for a first implementation.

Lunderberg:
Rebased onto main to resolve a merge conflict.

slyubomirsky (approved):
Thank you for responding to feedback. I am pleased to see the improved type-checking for PrimFuncs in Relax.

quic-sanirudh:
@Lunderberg After this pass introduced support for computing struct_info for PrimFuncs, I'm seeing a case where a primfunc whose output buffer is modified with the `sch.transform_layout` primitive and added back to the mod during legalize like what happens in the legalization of `R.layout_transform` causes a failure later in `call_tir_rewrite` pass.

A small test case that reproduces the issue is something like below:
```python
import tvm
from tvm.script import relax as R, tir as T, ir as I
from tvm import tir, relax

@I.ir_module
class Before:
    @R.function
    def main(x: R.Tensor((16,), dtype="float32"), y: R.Tensor((16,), dtype="float32")) -> R.Tensor((16,), dtype="float32"):
        with R.dataflow():
            lv: R.Tensor((4, 4), dtype="float32") = R.layout_transform(x, index_map=lambda i: (i // 4, i % 4), pad_value=None)
            gv: R.Tensor((4, 4), dtype="float32") = lv
            R.output(gv)
        return gv

if __name__ == '__main__':
    mod = Before
    mod = relax.transform.LegalizeOps()(mod)
    mod = relax.transform.CallTIRRewrite()(mod)
``` 

From what I've been able to understand, the `FuncStructInfo` for the PrimFunc is not recomputed when it's output buffer got changed, which happens through a schedule primitive in the case of the R.layout_transform legalization. I can try to fix it, but wanted to check whether the right fix would be to identify cases where a prim_func can change its struct_info through any means (scheduling in this case) and call InferStructInfo in all those cases, but I wanted to ask if there was any better solutions here.

Lunderberg:
Ooh, interesting.  I wasn't aware that `R.layout_transform` was implemented in terms of `Schedule.layout_transform`.  Can you take a look at https://github.com/apache/tvm/pull/16832 and see if it resolves your issue?

quic-sanirudh:
> Ooh, interesting. I wasn't aware that `R.layout_transform` was implemented in terms of `Schedule.layout_transform`. Can you take a look at #16832 and see if it resolves your issue?

Thanks a lot for the very quick fix. Yes it does seem to solve the issue.

PR #16644, opened 2024-02-26
[Relax][Frontend][Onnx]fix name supply bug

chengven027-intellif:
Hi,  tvm:
    From the description of the function of  _sanitize_name:
    ```
        If the name is None, returns a string input_0, input_1, etc.
        If the input is an empty string, returns empty_0, empty_1, etc.
        If the input is a string that does not start with a letter or underscore,
        returns input_<name>. Otherwise, returns an unique input name.
    ```.
   The input to the current model is "input.1".which will be converted into "input_1" by this function.
   This conversion is wrong and does not conform to the three situations described above.
   So I re-modified the relevant conversion conditions according to the above description. I am not sure whether it will affect other cases. @jwfromm @gigiblender @tqchen

### DIFF
--- a/python/tvm/relax/frontend/onnx/onnx_frontend.py
+++ b/python/tvm/relax/frontend/onnx/onnx_frontend.py
@@ -2059,12 +2059,10 @@ class ONNXGraphImporter:
         if name == "":
             return self._name_supply.fresh_name("empty_")
 
-        new_name = name.replace(".", "_")
-        if not new_name[0].isalpha() and new_name[0] != "_":
-            new_name = str(self._name_supply.fresh_name("input_" + new_name))
+        if not name[0].isalpha() and name[0] != "_":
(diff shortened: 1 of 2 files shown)


python/tvm/relax/frontend/onnx/onnx_frontend.py
```diff
@@ -2059,12 +2059,10 @@ def _sanitize_name(self, name: str) -> str:
         if name == "":
             return self._name_supply.fresh_name("empty_")
 
-        new_name = name.replace(".", "_")
```
yongwww, line 2062: it would be good to have a test case to cover this change. If this change is going to align with the description, one option is to refine the description. It would be good to avoid modifications to src/ir/supply.cc, given that it serves as the foundation for numerous cases.

chengven027-intellif:
sry, Its my fault. This is right.

PR #16649, opened 2024-02-27
[BugFix][FFI] Add a missing default for datatype lanes

lhutton1:
A default value for lanes was unintentionally removed in #16612, this PR fixes this which in turn fixes the test failure seen in `test_tensor_dtype_lanes` in CI.

cc @tqchen @Lunderberg @ekalda


ekalda (approved):
Thanks for the quick fix @lhutton1!

PR #16650, opened 2024-02-27
[Web] Revert back to the non-parallel version to avoid cache.add() error

DiegoCao:
To address the issue in https://github.com/mlc-ai/web-llm/issues/313, we revert back the parallel download to avoid the cache.error() on the webLLM side. Will dig further into the issue and come with the solution to it together with later indexDB  support.


CharlieFRuan (approved):

DavidGOrtega:
@DiegoCao Its not the real solution.
As I stated the cache can fail even if its just one by one. I have suffered that also.

Definitely HF CDN is not great. It should allow to do parallel requests without mayor problems.

Real Solution:
 1. Implement retry mechanism (like 3 retries or something)
 2. Parallellse n files (5 by default) instead of all of them

DavidGOrtega:
I can give it a shot tomorrow

tqchen (approved):

CharlieFRuan:
@DavidGOrtega Thanks for offering help! We found out that the issue was not due to parallel downloading the shards but parallelly processing the shards. We managed to keep the parallel downloads.

tqchen:
need rebase

tqchen (approved):

PR #16655, opened 2024-02-28
[TIR] Improve well-formed check's handling of match buffer

Lunderberg:
- The `T.match_buffer` at the start of a function may contain repeated use of the same data var.  For example, a function that must accept two `DLTensor` objects with the same backing allocation.

- The `"buffer_bind_scope"` is an older style of match buffer, and may be the point of definition for variables.

- A `BufferRealize` node may act either as an annotation of an external buffer (which indices are used), or as a point of definition for a local buffer (allocation extents of that buffer).

### DIFF
--- a/src/tir/ir/tir_visitor_with_path.cc
+++ b/src/tir/ir/tir_visitor_with_path.cc
@@ -78,31 +78,12 @@ void TIRVisitorWithPath::Visit(const PrimFunc& func, ObjectPath path) {
   // variable has occurred.  Therefore, to ensure that we only avoid
   // duplicate calls to VisitVarDef, these semantics need to be
   // checked.
-  std::unordered_set<Var, ObjectPtrHash, ObjectPtrEqual> defined_params;
   std::vector<std::variant<DefContext<Var>, DefContext<Buffer>>> context;
 
   auto ppath = path->Attr("params");
   for (size_t i = 0; i < func->params.size(); i++) {
     context.push_back(WithDef(func->params[i], ppath->ArrayIndex(i)));
-    defined_params.insert(func->params[i]);
   }
-
-  auto try_visit_implicit_var_def = [this, &defined_params, &context](const PrimExpr& expr,
-                                                                      ObjectPath path) {
-    if (auto opt = expr.as<Var>()) {
-      auto var = opt.value();
-      if (!defined_params.count(var)) {
-        context.push_back(WithDef(var, path));
-        defined_params.insert(var);
-      }
-    }
-  };
-  auto try_visit_implicit_var_def_array = [&try_visit_implicit_var_def](const Array<PrimExpr>& arr,
-                                                                        ObjectPath path) {
-    for (size_t i = 0; i < arr.size(); i++) {
-      try_visit_implicit_var_def(arr[i], path->ArrayIndex(i));
-    }
-  };
 
   auto buffer_map_path = path->Attr("buffer_map");
   for (size_t i = 0; i < func->params.size(); i++) {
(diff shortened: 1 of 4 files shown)


Lucien0:
Hi @Lunderberg , I tested my cases with this pr, and they are all working properly now. Thanks for your help!

src/tir/ir/tir_visitor_with_path.cc
```diff
@@ -199,12 +174,23 @@ void TIRVisitorWithPath::VisitStmt_(const LetStmtNode* op, ObjectPath path) {
...
 
-  std::optional<DefContext<IterVar>> context = std::nullopt;
+  std::vector<std::variant<DefContext<IterVar>, DefContext<Var>>> context;
   if (auto iter_var = op->node.as<IterVar>();
       iter_var && (op->attr_key == attr::thread_extent || op->attr_key == attr::virtual_thread)) {
     // Some attributes serve as a source of definition for the
     // tir::Var they annotate.
-    context = WithDef(iter_var.value(), path->Attr("node"));
+    context.push_back(WithDef(iter_var.value(), path->Attr("node")));
+  } else if (op->attr_key == attr::buffer_bind_scope) {
```
slyubomirsky, line 183: Probably worth commenting that this acts as an older form of MatchBuffer, per the PR description.

src/tir/ir/tir_visitor_with_path.cc
```diff
@@ -199,12 +174,23 @@ void TIRVisitorWithPath::VisitStmt_(const LetStmtNode* op, ObjectPath path) {
 void TIRVisitorWithPath::VisitStmt_(const AttrStmtNode* op, ObjectPath path) {
   Visit(op->value, path->Attr("value"));
 
-  std::optional<DefContext<IterVar>> context = std::nullopt;
+  std::vector<std::variant<DefContext<IterVar>, DefContext<Var>>> context;
```
slyubomirsky, line 177: Is this ever used? I don't see any reads from it.

src/tir/ir/tir_visitor_with_path.cc
```diff
@@ -250,7 +236,8 @@ void TIRVisitorWithPath::VisitStmt_(const BufferStoreNode* op, ObjectPath path)
 void TIRVisitorWithPath::VisitStmt_(const BufferRealizeNode* op, ObjectPath path) {
   Visit(op->condition, path->Attr("condition"));
   Visit(op->bounds, path->Attr("bounds"));
-  auto context = WithDef(op->buffer, path->Attr("buffer"));
+  auto context = WithDefIfUndefined(op->buffer->data, path->Attr("buffer")->Attr("data"));
```
slyubomirsky, line 239: I imagine this accounts for the case where a BufferRealize can act as a point of definition?

slyubomirsky (commented):
It's good to have a more systematic approach to a feature like points of definition in a MatchBuffer. I commented on some parts of the implementation that I couldn't entirely follow.

src/tir/ir/tir_visitor_with_path.cc
Lunderberg: Thank you, and added a comment with description.

src/tir/ir/tir_visitor_with_path.cc
Lunderberg: There aren't any reads from it, as it holds a scoped context manager.  On destruction, the `DefContext<T>` object removes items from `TIRVisitorWithPath::in_scope_definitions_`, and calls the `ExitDef` handler of the child class.

Also, thank you for pointing this one out.  When switching from `std::optional` to `std::vector`, I forgot to add a `while(context.size()) context.pop_back();` loop in case child classes rely on `ExitDef` being called in the reverse order from `EnterDef`.

src/tir/ir/tir_visitor_with_path.cc
Lunderberg: That's correct.  In cases where the buffer's backing allocation is defined externally, the `BufferRealize` is an annotation of the bounds where the external buffer is accessed.  Otherwise, `BufferRealize` is an allocation.  Prior to this commit, only the external backing allocation was handled.

src/tir/ir/tir_visitor_with_path.cc
```diff
@@ -195,6 +204,10 @@ void TIRVisitorWithPath::VisitStmt_(const AttrStmtNode* op, ObjectPath path) {
     Visit(expr.value(), path->Attr("node"));
   }
   Visit(op->body, path->Attr("body"));
+
+  while (context.size()) {
```
slyubomirsky, line 208: Thanks for the change. It's probably worth also putting in a line that this is to ensure that the defs expire, otherwise it seems like spooky action at a distance.

Lunderberg: Yeah, I tried to follow the existing `FooContext` naming structure (e.g. `arith::ConstraintContext`).

slyubomirsky (approved):
Thank you for the clarifications. The changes seem reasonable and they address issues that have arisen elsewhere.

PR #16658, opened 2024-02-29
[Relax] Normalize use of void-type variable to inline R.tuple() 

Lunderberg:
This is a follow-up commit to https://github.com/apache/tvm/pull/16641.  While parsing of relax expressions without a variable binding could be implemented at that point (e.g. `R.assert_op(condition)` instead of `dummy_var = R.assert_op(condition)`), the corresponding printing changes could not.  This was because a variable that satisfies `relax::HasVoidStructInfo(var)` could still be used later in the function, and removing its binding would result in use of an undefined variable.

This commit normalizes use of void-type variables to an in-line `R.tuple()`.  This simplifies the relax function, and also allows the binding of void-type variables to be hidden.

### DIFF
--- a/src/relax/ir/block_builder.cc
+++ b/src/relax/ir/block_builder.cc
@@ -547,7 +547,14 @@ class Normalizer : public BlockBuilderImpl, private ExprFunctor<Expr(const Expr&
     return GetRef<Var>(var);
   }
 
-  Expr VisitExpr_(const VarNode* var) final { return VisitVar_<Var>(var); }
+  Expr VisitExpr_(const VarNode* var_ptr) final {
+    auto var = VisitVar_<Var>(var_ptr);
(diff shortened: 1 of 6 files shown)


Lunderberg:
This commit includes https://github.com/apache/tvm/pull/16641 in its history, and is marked as a draft until https://github.com/apache/tvm/pull/16641 lands.

slyubomirsky (approved):
This will be good for avoiding the tricky situation that arises with #16641, mainly where it can be a problem to omit the binding of a unit-type variable if the var is used later. I think it's pretty harmless to simply inline all unit values.

Lunderberg:
> I think it's pretty harmless to simply inline all unit values.

Agreed, and it matches our existing behavior for things like `R.ExternFunc` and `R.const`.

PR #16660, opened 2024-02-29
[Bugfix][TIR] Fix duplicate AllocateConst in CacheReadWrite schedule primitive

Anndrey24:
When inserting a `cache_read` / `cache_write` stage, the `tir.AllocateConst` statement would be duplicated if its body was not a `tir.SeqStmt` node (e.g. `tir.For`), leading to compilation failures. This happened because `tir.AllocateConst` and `tir.DeclBuffer` statements are always re-attached to the statement's body after the `cache_read` / `cache_write` stage is inserted in it, but the stage was being appended to the whole statement (which already contains the `tir.AllocateConst`) and not just its body, causing duplications.

This commit also adds a test where the first `cache_read` stage is inserted into a statement whose body is a `tir.For`, while the second stage is added to a body that is `tir.SeqStmt` to check for regressions.  

cc @Lunderberg @ekalda @lhutton1

### DIFF
--- a/tests/python/tir-schedule/test_tir_schedule_cache_read_write.py
+++ b/tests/python/tir-schedule/test_tir_schedule_cache_read_write.py
@@ -1384,11 +1384,10 @@ def test_cache_read_allocate_const():
     def before(A: T.Buffer((8), "float32"), C: T.Buffer((8), "float32")):
         B = T.allocate_const([0.0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7], "float32", [8])
         B_buf = T.decl_buffer((8), dtype="float32", data=B)
-        for i in T.serial(128):
+        for i in range(8):
             with T.block("C"):
-                vi = T.axis.remap("S", [i])
+                vi = T.axis.spatial(8, i)
                 T.reads(A[vi], B_buf[vi])
-                T.writes(C[vi])
                 C[vi] = A[vi] + B_buf[vi]
 
     @T.prim_func
@@ -1400,20 +1399,15 @@ def test_cache_read_allocate_const():
         for ax0 in range(8):
             with T.block("A_global"):
                 v0 = T.axis.spatial(8, ax0)
-                T.reads(A[v0])
-                T.writes(A_global[v0])
                 A_global[v0] = A[v0]
         for ax0 in range(8):
             with T.block("B_buf_global"):
                 v0 = T.axis.spatial(8, ax0)
                 T.reads(B_buf[v0])
-                T.writes(B_buf_global[v0])
                 B_buf_global[v0] = B_buf[v0]
-        for i in range(128):
+        for i in range(8):
             with T.block("C"):
-                vi = T.axis.spatial(128, i)
-                T.reads(A_global[vi], B_buf_global[vi])
-                T.writes(C[vi])
+                vi = T.axis.spatial(8, i)
                 C[vi] = A_global[vi] + B_buf_global[vi]
 
     sch = tir.Schedule(before)
--- a/tests/python/tir-schedule/test_tir_schedule_cache_read_write.py
+++ b/tests/python/tir-schedule/test_tir_schedule_cache_read_write.py
@@ -1379,6 +1379,54 @@ def test_cache_read_fail_invalid_storage_scope(use_block_name):
         sch.cache_read(block_b, 0, "test_scope")
 
 
+def test_cache_read_allocate_const():
+    @T.prim_func
+    def before(A: T.Buffer((8), "float32"), C: T.Buffer((8), "float32")):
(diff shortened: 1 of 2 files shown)


tests/python/tir-schedule/test_tir_schedule_cache_read_write.py
```diff
@@ -1379,6 +1379,54 @@ def test_cache_read_fail_invalid_storage_scope(use_block_name):
...
 
+def test_cache_read_allocate_const():
+    @T.prim_func
+    def before(A: T.Buffer((8), "float32"), C: T.Buffer((8), "float32")):
+        B = T.allocate_const([0.0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7], "float32", [8])
+        B_buf = T.decl_buffer((8), dtype="float32", data=B)
+        for i in T.serial(128):
+            with T.block("C"):
+                vi = T.axis.remap("S", [i])
+                T.reads(A[vi], B_buf[vi])
```
Lunderberg, line 1390: Nitpick: The `T.reads` and `T.writes` annotations can be inferred in TVMScript parsing.  Unless the test case is specifically checking the values in these annotations, readability of the test case can be improved by removing them.

Lunderberg (approved):
Looks good, and thank you for the fix!

Long-term, we may want to avoid this class of bugs by changing the `const Stmt& stmt` argument of `InsertCacheState` to `Stmt body`.  This would avoid having two almost-but-not-quite identical variables within the same scope.

tests/python/tir-schedule/test_tir_schedule_cache_read_write.py
Anndrey24: Agreed! However, it seems that `B_buf` does not get automatically added to the `T.reads()` annotations so I still specify it myself in the two blocks where it is needed. This would be the error after removing all `T.reads()` and `T.writes()`:

```
ScheduleError: An error occurred in the schedule primitive 'cache-read'.
E       The IR with diagnostic is:
E       # from tvm.script import ir as I
E       # from tvm.script import tir as T
E       
E       @I.ir_module
E       class Module:
E           @T.prim_func
E           def main(A_handle: T.handle, C_handle: T.handle):
E               T.func_attr({"global_symbol": "before"})
E               A = T.match_buffer(A_handle, (8,))
E               C = T.match_buffer(C_handle, (8,))
E               with T.block("root"):
E                   T.reads()
E                   T.writes()
E                   B_buf = T.allocate_const([0, 0.10000000149011612, 0.20000000298023224, 0.30000001192092896, 0.40000000596046448, 0.5, 0.60000002384185791, 0.69999998807907104], "float32", [8])
E                   B_buf_1 = T.decl_buffer((8,), data=B_buf)
E                   for i in range(8):
E                       # tir.Block#0
E                       with T.block("C"):
E                       ^^^^^^^^^^^^^^^^^^
E                           vi = T.axis.spatial(8, i)
E                           ^^^^^^^^^^^^^^^^^^^^^^^^^
E                           T.reads(A[vi])
E                           ^^^^^^^^^^^^^^
E                           T.writes(C[vi])
E                           ^^^^^^^^^^^^^^^
E                           C[vi] = A[vi] + B_buf_1[vi]
E                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
E       Error message: The block tir.Block#0 has 1 read regions, so `buffer_index` is required to be in [0, 1). However, the input `buffer_index` is 1, which is out of the expected range.
```

Lunderberg: Thank you for cleaning it up, and that's an iteresting failure.  It looks like the `T.reads()` annotations are only generated for buffers that can be manipulated withing schedulable TIR.  The filtering is applied [here](https://github.com/apache/tvm/blob/main/src/tir/analysis/block_access_region_detector.cc#L300), and only contains buffers that were passed as arguments ([added here](https://github.com/apache/tvm/blob/main/src/tir/ir/script/script_complete.cc#L107)) or that were part of a `tir::Block` node ([added here](https://github.com/apache/tvm/blob/main/src/tir/ir/script/script_complete.cc#L52)).

I think this is a bug in the `GetBlockAccessRegion`, that it should track reads/writes in buffers declared by `DeclBuffer` nodes as well.

But that's for a later clean-up on my side, and this PR looks good!  Thank you!

lhutton1:
@tvm-bot rerun

lhutton1 (approved):
Thanks @Anndrey24, LGTM. Since #16663 was merged would you prefer to address the comment here or address in a follow up?

Anndrey24:
Oops, I hadn't noticed the fix was already merged. I've also removed the `T.reads()` now! :)

lhutton1:
@tvm-bot rerun

Anndrey24:
@tvm-bot rerun

lhutton1:
Thanks @Anndrey24 @Lunderberg!

Issue #16577, opened 2024-02-15
[Bug] Possible issue with the "simplify pass" using the "propagate_knowns_to_simplify_expressions" flag


sdalvi-quic:
I was trying to understand the use of **T.assume** in the simplification of expressions and I came across this example in the "**test_tir_transform_simplify.py**" file named as "**TestSimplifyUsingPartiallyKnownBufferExpression**". Here, I tried to modify the test and saw that, it was failing while using the "**propagate_knowns_to_simplify_expressions**" flag and T.assume statements. I have attached the test below. 
```c++
class TestSimplifyUsingPartiallyKnownBufferExpression(BaseBeforeAfter):
    """An assumption about buffer contents may apply to only part of a buffer

    Like TestSimplifyUsingPartiallyKnownBufferConditional, but the
    conditional is expressed as part of T.assume, instead of in the
    control flow.
    """

    # propagate_knowns_to_prove_conditional = True
    propagate_knowns_to_simplify_expressions = True

    def before(A: T.Buffer(16, "int32")):
        for i in T.serial(16):
            T.evaluate(T.assume(i < 14 or A[i] == 0))

        for i in T.serial(16):
            if i < 14:
                if A[i] == 0:
                    A[i] = 42

    def expected(A: T.Buffer(16, "int32")):
        for i in T.serial(16):
            T.evaluate(T.assume(i < 14 or A[i] == 0))

        for i in T.serial(16):
            if i < 14:
                if A[i] == 0:
                    A[i] = 42
``` 

Here given the assumption statements `T.assume(i < 14 or A[i] == 0)`, A[i]  will have to be 0 if the range of `i >= 14` but A[i] may or may not be 0 if `i < 14`. So, we should see the expected code as **above** where no statement is eliminated. But the IR returned after the pass eliminates the condition `if A[i] == 0` and returns the IR as **below** : 
```c++
def after(A: T.Buffer((16,), "int32")):
   T.func_attr({"global_symbol": "main"})
   for i in range(16):
       T.assume(i < 14 or A[i] == 0)
   for i in range(16):
      if i < 14:
           A[i] = 42
``` 

I see that the two flags "**propagate_knowns_to_simplify_expressions**" and "**propagate_knowns_to_prove_conditional** " are internally calling the function **SimplifyInContext** which is simplifying using the line :
```c++
expr = control_flow_block.known_at_block_start.SubstituteKnownBufferValues(
      std::move(expr), axis_var_lookup_, analyzer);
``` 

[...]

PR #16664, opened 2024-03-01
Fix for assume statement

sdalvi-quic:
The assume statement was not understanding the complete expression present inside. The predicate was not getting appended along with the buffer access. The additional predicate was added to the conditions list but was not appended while creating the buffer touch . This fix helps to resolve it. 

The issue is that the "TestSimplifyUsingPartiallyProvenBufferValueScatter" test is not completing after the fix. Rest of the tests are passing. I guess the issue is not with the fix but might be somewhere else. 

For reference : https://github.com/apache/tvm/issues/16577

 @Lunderberg, @jverma-quic and @quic-sanirudh Can you please review and point to the possible issue with the above test.

### DIFF
--- a/src/tir/analysis/control_flow_graph.cc
+++ b/src/tir/analysis/control_flow_graph.cc
@@ -473,9 +473,10 @@ class ControlFlowGraphBuilder final : public IRVisitorWithAnalyzer {
   void VisitAccess(const BufferAccess& node, BufferTouch::AccessType touch_type,
                    PrimExpr known_value_expr) {
     auto& current_block = out_->control_flow_.back();
-    current_block.current_predicate = CurrentScopePredicate();
+    PrimExpr current_predicate = CurrentScopePredicate();
     BufferTouch buffer_touch = current_block.MakeBufferTouch(out_, node->buffer, node->indices,
-                                                             touch_type, known_value_expr);
+                                                             touch_type, known_value_expr, 
+                                                             current_predicate);
     current_block.touch_points.push_back(buffer_touch);
   }
 
@@ -638,7 +639,7 @@ class ControlFlowGraphBuilder final : public IRVisitorWithAnalyzer {
 
 std::pair<BufferTouch, Map<Var, Range>> ControlFlowGraph::ControlFlowBlock::MakeBufferTouch(
     const tir::Buffer& buf, Array<Var> index_variables, Array<PrimExpr> indices,
-    BufferTouch::AccessType touch_type, PrimExpr known_value_expr) const {
+    BufferTouch::AccessType touch_type, PrimExpr known_value_expr, PrimExpr current_predicate) const {
   const auto& current_block = *this;
 
   Analyzer local_analyzer;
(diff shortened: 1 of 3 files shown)


quic-sanirudh:
@tvm-bot rerun

src/tir/analysis/control_flow_graph.cc
```diff
@@ -473,6 +473,7 @@ class ControlFlowGraphBuilder final : public IRVisitorWithAnalyzer {
   void VisitAccess(const BufferAccess& node, BufferTouch::AccessType touch_type,
                    PrimExpr known_value_expr) {
     auto& current_block = out_->control_flow_.back();
+    current_block.current_predicate = CurrentScopePredicate();
```
Lunderberg, line 476: Nitpick: Instead of adding a member variable, can we add this as an additional parameter to `current_block.MakeBufferTouch`?  That way, we avoid accidental re-use of `current_predicate` from other callsites.

tests/python/tir-transform/test_tir_transform_simplify.py
```diff
@@ -1325,6 +1326,34 @@ def expected(A: T.Buffer(16, "int32")):
...
+    def before(A: T.Buffer(16, "int32")):
+        for i in T.serial(16):
+            T.evaluate(T.assume(i < 14 or A[i] == 0))
+
+        for i in T.serial(16):
+            if i < 14:
+                if A[i] == 0:
+                    A[i] = 42
+
+    def expected(A: T.Buffer(16, "int32")):
```
Lunderberg, line 1347: Nitpick: Instead of duplicating the same function definition, I'd recommend `expected = before`.  That makes it clear to a reader that the simplification shouldn't cause any changes.

tests/python/tir-transform/test_tir_transform_simplify.py
```diff
@@ -1325,6 +1326,34 @@ def expected(A: T.Buffer(16, "int32")):
                 A[i] = 42
 
 
+class TestAssume(BaseBeforeAfter):
```
Lunderberg, line 1329: Nitpick: The `T.assume` function is tested in other cases as well, but calling this `TestAssume` implies this is validating all behavior of `T.assume`.  Can we rename the test case to `TestAssumeMayContainAdditionalPredicate`?

Lunderberg (commented):
Overall, looks good, just a number of nitpick changes.

sdalvi-quic:
Thanks @Lunderberg I will do the above changes recommended! 

I realized the issues that after this fix the "TestSimplifyUsingPartiallyProvenBufferValueScatter" test is not completing but, "TestSimplifyUsingPartiallyProvenBufferValueGather" test is passing. There does not seem to be an issue while generating the control flow graph but somewhere later after introducing the fix. Can you please point what might be the possible issue? 

Currently src/tir/transforms/simplify.cc and src/tir/transforms/reduce_branching_through_overcompute.cc pass works with lowered IR but is there any way how we can use it for T.blocks level implementation of the IR?

Thank you!

src/tir/analysis/control_flow_graph.cc
sdalvi-quic: Thanks! I added an additional parameter to MakeBufferTouch.

sdalvi-quic:
Hello @Lunderberg and @junrushao can you please review the above PR?

sdalvi-quic:
Since there is bug in the scatter test which is not yet resolved it might get difficult to upstream this PR. Closing it for now may raise the same PR later.

PR #16667, opened 2024-03-02
[Runtime][Builtin] Using float32 accumulation in attention kernel

MasterJH5574:
Prior to this PR, the TIR attention kernels does not cast matmul operands to fp32 before multiplying.
For models like Phi-2 which may have large Q/K/V data (at the level of a few hundreds), the fp16 multiplication exceeds the range of fp16, and lead to attention result being NAN sometimes.

This PR fixes this issue.


MasterJH5574:
@tvm-bot rerun

yongwww (approved):

PR #16672, opened 2024-03-04
[CLML] Fix build TVM with CLML on MacOS

echuraev:
The first fix related to the problem when we build TVM with `USE_CLML` option and OpenCL header files were not found in the system. The second fix is for building CLML graph executor in Android build. `find_library` on MacOS is looking for libraries with `dylib` extension.  This is why OpenCL libraries from CLML SDK were not found. To fix this problem we specify paths to the libraries manually in case if they were not found by `find_library`.

cc: @srkreddy1238


echuraev:
@tvm-bot rerun

echuraev:
@tvm-bot rerun

echuraev:
@tvm-bot rerun

echuraev:
@srkreddy1238 CI is green. Could you please review this PR?
cc: @masahi

PR #16674, opened 2024-03-05
[Relax] Eager free original weights in transform_params

vinx13:
This updates `LiftTransformParams` to always free original weight in the bundled weights tuple. This prevents the issue that `transform_params` consumes memory with double weight size. See discussion https://discuss.tvm.apache.org/t/inplacetransformparams/16323.

cc @tqchen @MasterJH5574 @Lunderberg

### DIFF
--- a/src/relax/transform/lift_transform_params.cc
+++ b/src/relax/transform/lift_transform_params.cc
@@ -38,6 +38,9 @@
 namespace tvm {
 namespace relax {
 
+constexpr const char* kLiftTransformConsumeParams = "relax.lift_transform_params.consume_params";
+TVM_REGISTER_PASS_CONFIG_OPTION(kLiftTransformConsumeParams, Bool);
+
 namespace {
 
 struct CollectInfo {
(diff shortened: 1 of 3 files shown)


src/relax/transform/lift_transform_params.cc
```diff
@@ -498,6 +540,7 @@ Pass LiftTransformParams() {
         if (ends_with(func_name, "transform_params")) {
           func = WithAttr(func, tvm::attr::kGlobalSymbol, gvar->name_hint);
           func = BundleModelParams(func);
+          func = Downcast<Function>(ConsumeBundledParams()(func));
```
tqchen, line 543: let us add this as an option, consume_params

Lunderberg:
Any idea why the `KillAfterLastUse` pass isn’t already handling this case?  It seems like it shouldn’t require special handling in `LiftTransformParams`.

tqchen:
@Lunderberg in this case we are also killing the elements within tuple which is not covered by killafter use

Lunderberg:
Ah, thank you, and that makes sense. That would also explain why the increased footprint isn’t seen when using `LazyTransformParams`, since the tuple doesn’t exist as a parameter.

If I understand correctly, we could achieve a similar effect by unpacking the entire tuple on its first use, though that could be less robust to later lowering passes.  Since we have the in-place updates, it makes sense to use them, and I like the change.

tqchen (approved):

PR #16677, opened 2024-03-05
[Dlight] LowBatchGemv rule only apply to function with spatial symbolic var

jinhongyii:


jinhongyii:
cc: @MasterJH5574

jinhongyii:
@tvm-bot rerun

PR #16678, opened 2024-03-05
 [Dlight] LowBatchGemv rule only apply to function with spatial symbolic var

jinhongyii:
LowBatchGemv rule only apply to function with spatial symbolic var, so that this rule skips matmul in attention kernel which has symbolic_var `kv_seq_len`


jinhongyii:
@tvm-bot rerun

jinhongyii:
@tvm-bot rerun

MasterJH5574 (approved):
LGTM!

jinhongyii:
@tvm-bot rerun

PR #16682, opened 2024-03-06
[Bugfix][TIR] Handle AttrStmt of upcoming tir.Var in ConvertSSA

Lunderberg:
In some cases, an `AttrStmt` may legally refer to a TIR variable that hasn't yet been defined.  For example, the
`"pragma_parallel_launch_point"` attribute, which annotates a variable that is about to occur in a ForNode.  Prior to this commit, `ConvertSSA` treated the `AttrStmt` as the usage of a variable, followed by a nested definition to be de-duplicated.  This resulted in the output `AttrStmt` containing a reference to an undefined variable.

This commit updates `ConvertSSA` to handle this case.  If an `AttrStmt` refers to a not-yet-defined variable, the body is visited before marking it as defined.

This implementation may be simplified in the future by moving "pragma_parallel_launch_point" to be an annotation on the `ForNode`, rather than an `AttrStmt`.

### DIFF
--- a/src/tir/transforms/ir_utils.cc
+++ b/src/tir/transforms/ir_utils.cc
@@ -358,6 +358,7 @@ class IRConvertSSA final : public StmtExprMutator {
       }
 
       Var var = iter_var->var;
+      bool delayed_define = false;
       if (auto it = function_scope_var_remap_.find(var.get());
           it != function_scope_var_remap_.end()) {
         var = it->second;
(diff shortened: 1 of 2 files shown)


Lunderberg:
@slyubomirsky This should unblock https://github.com/apache/tvm/pull/16682, as it resolves the issue found in `tests/python/tir-transform/test_tir_transform_hoist_if.py::test_hoisting_block_scope_4`.

slyubomirsky (approved):
Thank you for the change. Would this also work for attributes that _do_ act as points of definition?

Lunderberg:
> Would this also work for attributes that do act as points of definition?

It should, yes.  This behavior is tested in `TestThreadIdxReusedWithinAndAcrossFunctions`, where the `"thread_extent"` attribute does act as a point of definition.

PR #16684, opened 2024-03-06
[Bugfix][SLM] Produce well-formed Relax for nn.modules.KVCache

Lunderberg:
Prior to this commit, the `nn.modules.KVCache` implementations used `R.call_packed(...)` to call the `"vm.builtin.attention_*"` functions. Since `nn.Module` emits all relax functions within a `relax.DataflowBlock`, where impure expressions are forbidden, this is ill-formed.

This commit updates the implementations in `nn.modules.KVCache` to use `R.call_pure_packed` instead of `R.call_packed`.  This assertation that the callee is pure allows the call to occur within a `relax.DataflowBlock`.

### DIFF
--- a/python/tvm/relax/frontend/nn/exporter.py
+++ b/python/tvm/relax/frontend/nn/exporter.py
@@ -21,7 +21,7 @@ import typing
 from tvm import tir
 from tvm.ir import IRModule
 
-from ... import expr as rx
+from .... import relax as rx
 from ...block_builder import BlockBuilder
 from ...struct_info import ObjectStructInfo, ShapeStructInfo, TupleStructInfo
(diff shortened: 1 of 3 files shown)


tqchen:
one note: we are migrating into the latest unified kv so likely we might want to remove the legacy kv builtin soon in favor of the new kv

Lunderberg:
That makes sense with the migration.  I've been assuming that the translation layer from SLM to Relax would still exist with the unified KV cache, and so it would need a corresponding change either way.

Lunderberg:
All unit tests now passing, ready for review/merge.

tqchen (approved):

PR #16685, opened 2024-03-06
[Relax] Implement StructInfoPattern for dataflow pattern matching

Lunderberg:
This commit implements `StructInfoPattern`, which can be applied to any existing `DFPattern`, and requires the expression to have a specific struct info.  Any symbolic variables that occur in the struct info are treated as free parameters, to be defined by the match.


slyubomirsky (approved):
This sounds like a great feature to have. The implementation is also easy to follow.

PR #16686, opened 2024-03-06
[SLM] Implement pattern replacement 

Lunderberg:
Prior to this commit, if an optimized `nn.Module` implementation were defined, either the model definition needed to be rewritten to include the optimized implementation, or the user needed to write their own `nn.Mutator` implementation.

This commit implements `nn.replace_implementation`, which provides a `nn.Mutator` that replaces all instances of a `nn.Module` with an optimized implementation.  This allows a user to write optimized implementations as subclasses, such as shown below.

```python
class BaseImplementation(nn.Module):
    ...

class OptimizedImplementation(BaseImplementation):
    ...
```

After defining the optimized implementation, `nn.replace_implementation(OptimizedImplementation)` returns a `nn.Mutator` that can inject the optimized module into an existing end-to-end SLM model.  In addition, the SLM-to-SLM transformation can be converted into a Relax-to-Relax transformation, allowing an easy path for migrating optimized kernels into a Relax optimation pipeline.

### DIFF
--- a/python/tvm/relax/frontend/nn/__init__.py
+++ b/python/tvm/relax/frontend/nn/__init__.py
@@ -26,6 +26,7 @@ from .modules import (
     ConvTranspose1D,
     Embedding,
     GroupNorm,
+    Identity,
     IOEffect,
     KVCache,
     LayerNorm,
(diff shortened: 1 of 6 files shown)


Lunderberg:
This PR is currently marked as a draft, both because it depends on functionality introduced in https://github.com/apache/tvm/pull/16684 and https://github.com/apache/tvm/pull/16685, and to allow discussion on the feature.

Lunderberg:
Finally getting back around to this one, ready for review.

PR #16676, opened 2024-03-05
[Relax] Remove the legalization of cumsum/cumprob

yongwww:
Remove the legalization for cumsum/cumprob, DispatchSortScan in default pipeline should work for them

PR #16688, opened 2024-03-06
[Bugfix][Thrust] Mark relax.cumsum implementation as scheduled

Lunderberg:
Prior to this commit, the `tests/python/relax/test_frontend_nn_op.py::test_renormalize_top_p_top_k_prob` test is failing on the main branch.  This is due to use of `tir.transform.DefaultGPUSchedule` on a kernel that cannot be scheduled.

This commit updates `relax.backend.DispatchSortScan()` to mark its output `cumsum` function as scheduled.


yongwww:
Thanks for the patch. The test should be fixed in https://github.com/apache/tvm/pull/16676.

Lunderberg:
Sounds good, and thank you!

PR #16689, opened 2024-03-08
[DeviceAPI] Support "GetCurrentStream"

MasterJH5574:
This PR introduces a new function `GetCurrentStream`to device API, which returns the current stream of the given device.

Meanwhile, this PR updates the "CreateStream" of CUDA to creating a non-blocking stream, so that the execution on this stream can overlap with the execution of other streams.

This PR also changes the `GPUCopy` of CUDA device API to always using `cudaMemcpyAsync`.

### DIFF
--- a/include/tvm/runtime/device_api.h
+++ b/include/tvm/runtime/device_api.h
@@ -177,6 +177,12 @@ class TVM_DLL DeviceAPI {
    */
   virtual void SetStream(Device dev, TVMStreamHandle stream) {}
   /*!
+   * \brief Get the current stream
+   * \param dev The device to get stream.
+   * \return The current stream of the device.
+   */
+  virtual TVMStreamHandle GetCurrentStream(Device dev);
+  /*!
    * \brief Synchronize 2 streams of execution.
    *
    * An event is created in event_src stream that the second then
--- a/src/runtime/c_runtime_api.cc
+++ b/src/runtime/c_runtime_api.cc
@@ -210,6 +210,8 @@ TVMStreamHandle DeviceAPI::CreateStream(Device dev) { return nullptr; }
 
 void DeviceAPI::FreeStream(Device dev, TVMStreamHandle stream) {}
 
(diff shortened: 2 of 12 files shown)


tqchen (approved):

Lunderberg:
> This PR also changes the `GPUCopy` of CUDA device API to always using `cudaMemcpyAsync`.

I think this portion of the commit needs to be reverted.  Prior to this commit, the [`NDArray::CopyTo`](https://github.com/apache/tvm/blob/main/include/tvm/runtime/ndarray.h#L116) function could be called to transfer an array to/from the GPU and return the transferred array.   After this commit, there is no synchronization point after the `cudaMemcpyAsync`, before returning control to the caller of `NDArray::CopyTo`.

* The caller may read from the `NDArray` result immediately after it completes.  After this commit, this is a read from uninitialized memory.
* The caller may free the backing allocation of the `NDArray` argument immediate after `NDArray::CopyTo` completes.  After this commit, this causes CUDA to read from a dangling pointer.

This function is used in many locations which relied on the previous semantics.

* The [`""vm.builtin.to_device"`](https://github.com/apache/tvm/blob/main/src/runtime/relax_vm/builtin.cc#L418) PackedFunc, which is the lowered form of `R.to_device`.

tqchen:
Indeed agree that this makes things more relax. On the other hand, from the device api's pov, we don't really guarantee the sync behavior in other DeviceAPIs, e.g. in the case of metal API or vulkan 


- In most GPU APIs, both copy from/to host and across are async (e.g. in the case of metal or vulkan)
- The default CUDA sync behavior of copyfromto actually was mainly limited to the default stream. 

One possible middleground we could have is to update CopyTo to always enable a StreamSync before CopyTo ends, this would help us preserve original usage of CopyTo, but still allows low level device API to enable async copy behavior that would generally provide more optimizations opportunities.

tqchen:
actually it is great this topic get revealed! since the original logics would cause issues for backends like metal/vulkan due to the fact that these copies are async

tqchen:
https://github.com/apache/tvm/pull/16716 contains the followup

PR #16690, opened 2024-03-08
Use target name instead of node name as function name

sunzj:
torch change the description of the graph, function name isn't same as node name, cause can't find function if import torch model. So use target name instead of node name.

fix https://github.com/mlc-ai/web-stable-diffusion/issues/54


tqchen (approved):

sidsingla:
Hi, facing still the same error as https://github.com/mlc-ai/web-stable-diffusion/issues/54

PR #16693, opened 2024-03-09
[Contrib] Support NDArray cache taking generator

MasterJH5574:
This PR enhances the `dump_ndarray_cache` function to take generator as input. Previously it can only take a dictionary.

Sometimes, it is possible that the total ndarray size cannot fit the main CPU memory, in which case we may turn to using generators so we can free some NDArray memory on the fly. And this PR supports the NDArray cache dumping with generators.

### DIFF
--- a/python/tvm/contrib/tvmjs.py
+++ b/python/tvm/contrib/tvmjs.py
@@ -17,12 +17,14 @@
 """Namespace to store utilities for building web runtime."""
 import hashlib
 import json
+import math
 import os
 import shutil
 
 # pylint: disable=unused-import
 import sys
-from typing import Mapping, Union
+from types import GeneratorType
+from typing import Iterator, Mapping, Tuple, Union
 
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


python/tvm/contrib/tvmjs.py
```diff
@@ -212,12 +224,15 @@ def dump_ndarray_cache(
         shard_manager.append(data, name=k, shape=shape, dtype=dtype, encode_format=encode_format)
 
         counter += 1
-        last_cmd = "[%04d/%04d] saving %s" % (counter, total, k)
+        if from_generator:
+            total_num += 1
+        last_cmd = "[%04d/%04d] saving %s" % (counter, total_num, k)
```
tqchen, line 229: remove total_num from the progress if from generator since it can be strange

python/tvm/contrib/tvmjs.py
```diff
@@ -159,27 +164,32 @@ def dump_ndarray_cache(
...
 
     cache_dir: str
         The path to the cache
 
     encode_format: {"f32-to-bf16", "raw"}
         Encoding format.
 
-    meta_data: json-compatible-struct
-        Extra meta_data to be stored in the cache json file.
+    meta_data: json-compatible-struct or Callable[[], Any]
```
tqchen, line 179: add an option `show_progress=True`, which toggles progress print, since the caller may use other means(e.g. tqdm)

tqchen (commented):
some minor nitsm otherwise LGTM

tqchen (approved):

PR #16694, opened 2024-03-10
[WEB] Initial support for asyncify

tqchen:
This PR enables asyncify support for web runtime.

Asyncify is a feature to allow C++ to call async function in javascript. The emcc compiler will unwind and store the stack, returning control to JS runtime. The JS runtime needs to be able to await the promise and then call rewind to get to the original suspended point.

This feature can be potentially useful when we would like to call WebGPU sync in C++ runtime. As on web platform everything have to be non-blocking. Previously we simply forbid c++ runtime to call sync and have to return to javascript side and let javascript side to call async.

Because asyncify can increase the wams size by 2x, we don't enable it by default in emcc.py and still would need to pass in options.

We will confirm potential benefit tradeoffs before turning it on by default. Another catch is that as of now asyncify is not compatible with wasm exception, so we temporary turn wasm-exception it off for now. This is an item that is being worked on by emscripten so we might be able to turn it back on later.

The testcases are added.

reference: https://emscripten.org/docs/porting/asyncify.html

### DIFF
--- a/python/tvm/contrib/emcc.py
+++ b/python/tvm/contrib/emcc.py
@@ -42,7 +42,14 @@ def create_tvmjs_wasm(output, objects, options=None, cc="emcc"):
     cmd += ["-O3"]
     cmd += ["-std=c++17"]
     cmd += ["--no-entry"]
-    cmd += ["-fwasm-exceptions"]
+    # NOTE: asynctify conflicts with wasm-exception
+    # so we temp disable exception handling for now
+    #
+    # We also expect user to explicitly pass in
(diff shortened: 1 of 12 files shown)


tqchen:
cc @CharlieFRuan @MasterJH5574 @rickzx

tqchen:
To enable it, when calling `create_tvmjs_wasm`, add `["-s", "ASYNCIFY=1"]` to options

MasterJH5574 (approved):
LGTM! Thanks for the support.

PR #16697, opened 2024-03-11
[Relax] Implement relax.transform.TopologicalSort

Lunderberg:
This commit implements a utility `relax.transform.TopologicalSort`, which can re-order the bindings that occur in a
`relax.DataflowBlock`.  This is not intended for use in a general-purpose optimization pipeline, but instead as a utility that may be used as needed in specific cases.  For example, normalization of unit tests that should not depend on the order of variable binding.

### DIFF
--- a/python/tvm/relax/transform/transform.py
+++ b/python/tvm/relax/transform/transform.py
@@ -234,7 +234,7 @@ def ToNonDataflow() -> tvm.ir.transform.Pass:
 
 
 def TopologicalSort(order="depth-first", direction="from-inputs") -> tvm.ir.transform.Pass:
-    """Sort relax.Dataflow blocks
+    """Sort bindings in relax.Dataflow blocks in the order specified
 
     Parameters
     ----------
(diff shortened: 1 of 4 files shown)


python/tvm/relax/transform/transform.py
```diff
@@ -233,6 +233,29 @@ def ToNonDataflow() -> tvm.ir.transform.Pass:
     return _ffi_api.ToNonDataflow()  # type: ignore
 
 
+def TopologicalSort(order="depth-first", direction="from-inputs") -> tvm.ir.transform.Pass:
+    """Sort relax.Dataflow blocks
```
masahi, line 237: To be more precise, "Sort bindings in relax.Dataflow blocks with the specified order"?

masahi (approved):

python/tvm/relax/transform/transform.py
Lunderberg: Good call, and updated.

PR #16699, opened 2024-03-11
[Transform] Remove R.Object parameters after LazyTransformParams

Lunderberg:
Prior to this commit, the output of `relax.transform.LazyTransformParams` would include all parameters that are not `R.Tensor`, in case they defined symbolic variables. However, this added too many unnecessary parameters, such as `R.Object` which cannot define symbolic variables.  This commit updates `relax.transform.LazyTransformParams` to only retain `R.Prim` and `R.Shape` parameters, which can define symbolic variables.


masahi (approved):

masahi:
@Lunderberg Can you take a look at the failed test?

PR #16704, opened 2024-03-12
[Fix][Arith] Fix canonical simplification of LE

MasterJH5574:
PR #15471 enhances the simplification for LE, while missed a case where the upper bound `kPosInf` is divisible by a factor. Therefore, prior to this PR, when simplifying `x * 1024 + y < z * 7168`, it will fails with the error message
```
InternalError: Check failed: value < 1LL << (dtype.bits() - 1)
(8589934591 vs. 2147483648) : ValueError: Literal value 8589934591
exceeds maximum of int32
```

This is just because the upper bound 7 here divides `kPosInf` the maximum value of int64, which passes an "if" condition in #15471 unexpectedly.

This PR fixes the issue.


tqchen:
Great fix!

tqchen (approved):

PR #16706, opened 2024-03-12
[VM] [Hexagon] Add buffers to `dma_wait` builtin

abhikran-quic:
- While introducing dma operations at graph level, relax `KillAfterLastUse` pass introduces `kill_tensor` operation after `dma_copy`. This leads to memory being deallocated when asynchronous copy operation is in progress. Hence, moving the input/output buffers to `dma_wait` to ensure `kill_tensor` is introduced after `dma_wait` at the graph level.
- Also, the logic for size calculation is updated to use `GetDataSize` function. 
- The test case is updated to use offsets instead of allocating different storage in VTCM.

### DIFF
--- a/src/runtime/relax_vm/hexagon/builtin.cc
+++ b/src/runtime/relax_vm/hexagon/builtin.cc
@@ -52,8 +52,8 @@ TVM_REGISTER_GLOBAL("vm.builtin.hexagon.dma_copy")
     });
 
 TVM_REGISTER_GLOBAL("vm.builtin.hexagon.dma_wait")
-    .set_body_typed([](TVMArgValue vm_ptr, int queue_id, int inflight_dma, NDArray src_arr,
-                       NDArray dst_arr) {
+    .set_body_typed([](TVMArgValue vm_ptr, int queue_id, int inflight_dma,
+                       [[maybe_unused]] NDArray src_arr, [[maybe_unused]] NDArray dst_arr) {
       ICHECK(inflight_dma >= 0);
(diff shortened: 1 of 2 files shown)


abhikran-quic:
cc: @Hzfengsy @quic-sanirudh

src/runtime/relax_vm/hexagon/builtin.cc
```diff
@@ -53,7 +52,8 @@ TVM_REGISTER_GLOBAL("vm.builtin.hexagon.dma_copy")
     });
 
 TVM_REGISTER_GLOBAL("vm.builtin.hexagon.dma_wait")
-    .set_body_typed([](TVMArgValue vm_ptr, int queue_id, int inflight_dma) {
+    .set_body_typed([](TVMArgValue vm_ptr, int queue_id, int inflight_dma, NDArray src_arr,
+                       NDArray dst_arr) {
```
quic-sanirudh, line 56: For `src_arr` and `dst_arr`, perhaps we should add the [`[[maybe_unused]]`](https://en.cppreference.com/w/cpp/language/attributes/maybe_unused).

quic-sanirudh (approved):
LGTM, apart from the minor comment for unused argument. Thanks.

src/runtime/relax_vm/hexagon/builtin.cc
abhikran-quic: Thank you @quic-sanirudh ! I have fixed this in the latest patch.

PR #16712, opened 2024-03-13
WIP: [Rust] Fixup rust bindings

cgerum:
This is a rewort of    #16628 it does not introduce new bindings, but only adopts to some changes in the TVM API. 

By adding missing attribute values.


cgerum:
@Lunderberg @tqchen any chance you know somebody who could review?

rust/tvm/src/ir/attrs.rs
```diff
@@ -27,3 +27,19 @@ use tvm_macros::Object;
...
+
+impl BaseAttrsNode {
+    pub fn base<T: IsObject>() -> BaseAttrsNode {
+        BaseAttrsNode {
+            base: Object::base::<T>(),
+        }
+    }
+}
+
+impl Attrs {
```
Lunderberg, line 39: On the C++ side, most `Attrs` are constructed using their subclass, not as a `BaseAttrsNode`.  Is this a difference that we want to have between the C++ and Rust bindings?

rust/tvm/src/ir/expr.rs
```diff
@@ -94,9 +94,9 @@ external! {
     fn _as_text(object: ObjectRef, show_meta_data: i32, annotate: runtime::Function) -> TString;
 }
 
-pub fn as_text<T: IsObjectRef>(object: T) -> String {
+pub fn as_text<T: IsObjectRef>(object: T, show_meta_data: i32) -> String {
```
Lunderberg, line 97: 1. Does the new argument break existing callsites of `as_text`?  Since we can't add default parameters to preserve backwards compatibility, I'm wondering if this should be updated to use the builder pattern in the future.
2. Should the argument be `show_meta_data: bool` instead of `show_meta_data: i32`?  The FFI cannot currently distinguish between the two, but exposing it as `i32` in the Rust bindings seems odd.

rust/tvm/src/ir/tir.rs
```diff
@@ -56,12 +56,35 @@ impl From<i32> for IntImm {
...
+    }
+}
+
 impl From<i32> for PrimExpr {
     fn from(i: i32) -> PrimExpr {
         IntImm::from(i).upcast()
     }
 }
 
+impl Into<i64> for PrimExpr {
```
Lunderberg, line 71: I don't think we should provide `Into<i64>` for `PrimExpr`, because `Into` should only be used for infallible conversions.  Can we provide `TryInto<i64>` instead?

Lunderberg (commented):
I'm not too family with the TVM Rust bindings, but the changes look reasonable overall.  I've made a couple of comments, but it would be good to get eyes from somebody more familiar with the Rust bindings as well.  (@jroesch, maybe?)

rust/tvm/src/ir/expr.rs
cgerum: > Does the new argument break existing callsites of `as_text`? Since we can't add default parameters to preserve backwards compatibility, I'm wondering if this should be updated to use the builder pattern in the future.

You mean `object.with_metadata().as_text()`? Unfortunately I am not sure wether backwards compatibility in this case is important enough, to warrant the increased API complexity.

PR #16713, opened 2024-03-13
[RUNTIME][METAL] Provide richer runtime when error happens

tqchen:
This PR enhances metal runtime to include more error messages when error happens.


MasterJH5574 (approved):
LGTM, thanks! The error description is indeed more insightful.

PR #16715, opened 2024-03-13
[Disco] Propagate structlog/logging config to workers

Lunderberg:
This is a follow-up to #16618, which propagates the `structlog` configuration to disco worker processes.  For configurations that use `structlog.stdlib` to integrate `structlog` with the stdlib `logging` module, this integration must also be forwarded.

### DIFF
--- a/python/tvm/runtime/disco/session.py
+++ b/python/tvm/runtime/disco/session.py
@@ -18,6 +18,7 @@
 with the distributed runtime.
 """
 
+import logging
 import os
 import pickle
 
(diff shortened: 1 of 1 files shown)


slyubomirsky (approved):
This seems straightforward enough, though I'm not aware of the wider context.

Lunderberg:
> This seems straightforward enough, though I'm not aware of the wider context.

Thank you.  For context, some applications use `structlog` to provide more flexible logging than python's stdlib `logging`.  The `structlog` configuration determines what pre-processing is done for the log statements (e.g. appending contextual information to a log statement).  When starting child processes using `multiprocessing`, it would be useful for the child processes to format/save their logs in the same manner as the parent, but this doesn't occur by default.  In [PR#16618](https://github.com/apache/tvm/pull/16618), I added handling to forward the `structlog` configuration from the main process to the `tvm.runtime.disco` worker processes.

However, some configurations ([example from `structlog`'s documentation](https://www.structlog.org/en/stable/standard-library.html#rendering-using-structlog-based-formatters-within-logging) integrate `structlog` with the stdlib `logging`.  The previous implementation only forwarded the configuration held by `structlog`, and didn't forward the configuration within the stdlib `logging`.  This PR closes that gap.

PR #16717, opened 2024-03-14
Enable Shared Function in LiftTransformParam Pass

zxybazh:
This PR enables specifying a list of function names to extract shared transform parameters. A single parameter transformation function will be produced, containing the preprocessing steps common across each function whose name is in a given function name list. 

Unit tests are passing except the one that has no shared prepocessing (transpose), skipping for now, will follow-up in another PR.

Cherry-picked from @vinx13's working branch.

### DIFF
--- a/src/relax/transform/lift_transform_params.cc
+++ b/src/relax/transform/lift_transform_params.cc
@@ -700,38 +700,6 @@ class ConsumeBundledParams : public ExprMutator {
   std::unordered_map<int, Expr> param_remap_;
 };
 
-// std::pair<std::unordered_set<GlobalVar, ObjectPtrHash, ObjectPtrEqual>, std::vector<Function>>
-//  GetTargetFunctions(const IRModule& mod, const Array<String>& target_function_names) {
-//   std::unordered_set<GlobalVar, ObjectPtrHash, ObjectPtrEqual> target_gvars;
-//   std::vector<Function> target_functions;
-//   if (target_function_names.size()) {
-//     for (const auto& name : target_function_names) {
-//       auto gvar = mod->GetGlobalVar(name);
-//       target_gvars.insert(gvar);
-//       target_functions.push_back(Downcast<Function>(mod->Lookup(gvar)));
-//     }
-//   } else {
-//     // Get all the functions that have the `num_input` attribute.
-//     std::vector<GlobalVar> ordered_target_gvars;
-//     for (const auto& [gvar, func] : mod->functions) {
-//       if (func->IsInstance<FunctionNode>()) {
-//         auto opt_num_input = func->GetAttr<Integer>(attr::kNumInput);
-//         if (opt_num_input) {
-//           ordered_target_gvars.push_back(gvar);
-//         }
-//       }
-//     }
-//     std::sort(ordered_target_gvars.begin(), ordered_target_gvars.end(),
-//               [](const GlobalVar& lhs, const GlobalVar& rhs) { return lhs->name_hint <
-//               rhs->name_hint; });
-//     for (const auto& gvar : ordered_target_gvars) {
-//       target_functions.push_back(Downcast<Function>(mod->Lookup(gvar->name_hint)));
-//     }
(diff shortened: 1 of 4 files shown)


src/relax/transform/lift_transform_params.cc
```diff
@@ -494,22 +700,103 @@ class ConsumeBundledParams : public ExprMutator {
   std::unordered_map<int, Expr> param_remap_;
 };
 
+// std::pair<std::unordered_set<GlobalVar, ObjectPtrHash, ObjectPtrEqual>, std::vector<Function>>
```
vinx13, line 703: let’s remove this

src/relax/transform/lift_transform_params.cc
```diff
@@ -494,22 +700,71 @@ class ConsumeBundledParams : public ExprMutator {
...
+    for (const auto& [gvar, func] : target_functions) {
+      auto info = LocalLiftableBindingCollector::Collect(
+          func, global_collect_info.has_value() ? &global_collect_info.value() : nullptr);
+      local_collect_info[gvar] = info;
+    }
+
+    for (const auto& [gvar, info] : local_collect_info) {
+      auto new_runtime_func = info.MakeRuntimeFunction();
+      updates->Add(gvar, new_runtime_func);
+      if (!global_collect_info.has_value()) {
```
Lunderberg, line 759: Can this conditional be moved to an else block of `if (global_collect_info.has_value())`?  That would make it clearer to a reader that the runtime function is always generated, and that there's a choice between a single global transformation and several individual transformations.

```c++
if (global_collect_info.has_value()) {
  auto global_transform = global_collect_info.value().MakeCompileTimeFunc();
  updates->Add(GlobalVar("transform_params"), global_transform);
} else {
  for (const auto& [gvar, info] : local_collect_info) {
      updates->Add(GlobalVar(gvar->name_hint + "_transform_params"),
                     info.MakeCompileTimeFunction());
  }
}
```

src/relax/transform/lift_transform_params.cc
```diff
@@ -494,22 +700,71 @@ class ConsumeBundledParams : public ExprMutator {
...
-    PreprocessPartitioner mutator;
-
     IRModule updates;
-    for (const auto& [gvar, func] : mod->functions) {
-      if (auto opt = func.as<relax::Function>()) {
-        auto new_func = Downcast<Function>(mutator(opt.value()));
-        if (!new_func.same_as(func)) {
-          updates->Add(gvar, new_func);
-        }
+
```
Lunderberg, line 736: With accepting a `Variant<Bool, Array<String>>` for the `shared_transform` argument, we would then check whether we need to make the shared transform as follows:

```c++
bool generate_shared_transform = shared_transform.as<Bool>().value_or(Bool(true))->value;
```

src/relax/transform/lift_transform_params.cc
```diff
@@ -494,22 +700,71 @@ class ConsumeBundledParams : public ExprMutator {
   std::unordered_map<int, Expr> param_remap_;
 };
 
+std::vector<std::pair<GlobalVar, Function>> GetTargetFunctions(
+    const IRModule& mod, const Array<String>& target_function_names) {
+  std::vector<std::pair<GlobalVar, Function>> target_functions;
+  if (target_function_names.size()) {
```
Lunderberg, line 706: With accepting a `Variant<Bool, Array<String>>` for the shared transforms, that argument would then be forwarded directly to the `GetTargetFunctions` utility.

```c++
if(auto function_names = shared_transforms.as<Array<String>>()) {
    for(const auto& name: function_names.value()) {
        ...
     }
} else {
    ...
}
```

Lunderberg (commented):
Overall, looks good, and thank you for the improvement!  A couple of comments on it, then it should be good to go!

zxybazh:
@Lunderberg  thank you for the detailed review, comments are all addressed, please take another look when you got time, thanks a lot!

Lunderberg (approved):

PR #16727, opened 2024-03-15
[TIR] LowerTVMBuiltin may use device_type from PrimFunc annotation

Lunderberg:
If an allocation occurs within a host function, it may not have a device/host split, and may just be annotated with `"tir.is_host_func"`.

### DIFF
--- a/tests/python/tir-transform/test_tir_transform_lower_tvm_builtin.py
+++ b/tests/python/tir-transform/test_tir_transform_lower_tvm_builtin.py
@@ -260,11 +260,13 @@ class TestLowerCPUAllocation(tvm.testing.CompareBeforeAfter):
 
 
 class TestLowerAllocateRequiresDeviceID(tvm.testing.CompareBeforeAfter):
+    """If device id is missing, error."""
+
     transform = tvm.tir.transform.LowerTVMBuiltin()
 
     def before():
         T.func_attr({"target": T.target("llvm")})
-        T.attr("dummy", "device_id", 0)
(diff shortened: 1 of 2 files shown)


tests/python/tir-transform/test_tir_transform_lower_tvm_builtin.py
```diff
@@ -260,11 +260,13 @@ def expected():
...
 
 class TestLowerAllocateRequiresDeviceID(tvm.testing.CompareBeforeAfter):
+    """If device id is missing, error."""
+
     transform = tvm.tir.transform.LowerTVMBuiltin()
 
     def before():
         T.func_attr({"target": T.target("llvm")})
-        T.attr("dummy", "device_id", 0)
+        T.attr("dummy", "device_type", 2)  # kDLCuda
```
slyubomirsky, line 269: Is this constant defined anywhere on the Python side? Perhaps it should be, to avoid such magic numbers (if we expect that to happen again).

slyubomirsky (approved):
The change is simple and this seems like a good capability to have. Thank you for implementing it.

tests/python/tir-transform/test_tir_transform_lower_tvm_builtin.py
Lunderberg: Good question, and looks like it is defined in the `tvm.runtime.Device` struct.  I've updated the usage here, and throughout this unit test file.

PR #9842, opened 2022-01-05
Attempt to prevent concurrent update in Map

d-smirnov:
Calling Map::Set invalidates existing iterators to protect from
using already deleted data due to re-hashing

Related to the issue fixed [here](https://github.com/apache/tvm/commit/6aa959028b94ab67621d81c6a7f7e5281e7376d3)

PR #16728, opened 2024-03-15
[Fix] Introduce TVM_DEBUG_WITH_ABI_CHANGE to warn ABI changes in debug mode

Ubospica:
Currently the ABI of tvm::runtime::Map will change if TVM_LOG_DEBUG is enabled or not:
```
// In MapNode::iterator
#if TVM_LOG_DEBUG
    uint64_t state_marker;
    ...
#else
    ...
#endif  // TVM_LOG_DEBUG
```
When TVM_LOG_DEBUG is inconsistent between different libraries (e.g. in TVM and in MLC-LLM), the inconsistency in ABI will lead to an invalid read/write.

This PR introduces another macro `TVM_DEBUG_WITH_ABI_CHANGE` for such cases to warn users that this debug option can lead to changes in ABI. It is not enabled by default.

cc @areusch @tqchen


tqchen (approved):

Ubospica:
See also #9842

PR #16730, opened 2024-03-16
[Relax] Refactor PatternRewriter into separate Block/Expr mutators

Lunderberg:
Prior to this commit, the `PatternRewriter` mutator handled pattern rewriting at either the expression level (`rewrite_call`) or the dataflow block level (`rewrite_bindings`).  These two functionalities had different external APIs, defined diffierent member variables, and visited different IR nodes.  In effect, it had two entirely independent implementations, which just happened to be implemented within the same class.

This commit refactors the single `PatternRewriter` mutator into separate `BlockPatternRewriter` and `ExprPatternRewriter` mutators.


tqchen:
related to the class but not this PR necessarily would be good to think about complexity reduction for rewrite_bindings, which is source of slowdown in our overall compilation

slyubomirsky (approved):
This seems like a clean refactor, thank you for separating out the functionality.

PR #16732, opened 2024-03-18
[Relax] Allow composition of DFPattern replacements 

Lunderberg:
The `rewrite_call` function accepts a `DFPattern`, and a function to
rewrite expressions matching that pattern.  Often, the rewriting
function will perform additional validation that cannot be expressed
within the `DFPattern` itself.  If this additional validation fails,
the rewriter function will return the matched expression unmodified.

Prior to this commit, an `OrPattern` that matches on the first branch,
but whose rewriter function does not apply a modification, would
prevent the second branch from being checked.  This commit updates the
`ExprPatternRewriter` to check both branches of a `OrPattern`, if the
rewriter function of the first branch does not modify the result.

### DIFF
--- a/src/relax/ir/dataflow_matcher.cc
+++ b/src/relax/ir/dataflow_matcher.cc
@@ -1158,17 +1158,51 @@ class ExprPatternRewriter : ExprMutator {
   Expr VisitExpr(const Expr& expr) override {
     auto node = ExprMutator::VisitExpr(expr);
 
-    if (auto matches_opt = ExtractMatchedExpr(pattern_, node, bindings_)) {
-      Expr rewritten_expr = rewriter_func_(node, matches_opt.value());
-      if (!rewritten_expr.same_as(node)) {
-        return builder_->Normalize(rewritten_expr);
-      }
+    std::vector<DFPattern> matches_top_level;
+    if (auto rewritten = TryRewrite(node, pattern_, &matches_top_level)) {
+      return builder_->Normalize(rewritten.value());
     }
 
     return node;
   }
 
  private:
+  Optional<Expr> TryRewrite(const Expr& expr, const DFPattern& pattern,
+                            std::vector<DFPattern>* matches_top_level) {
+    ICHECK(matches_top_level);
+
+    // Special handling if the user-supplied pattern is a `OrPattern`.
+    // While the `ExtractMatchedExpr` can handle matching the
+    // `OrPattern`, it will return on the first match, even if the
+    // `rewriter_func_` doesn't apply a replacement.  Unpacking the
+    // `OrPattern` here allows the match to be resumed if
+    // `rewriter_func_` returns the original function unmodified.
+    // This is only valid for a top-level match.
+    if (auto or_pattern = pattern.as<OrPatternNode>()) {
+      matches_top_level->push_back(pattern);
+      Optional<Expr> output = TryRewrite(expr, or_pattern->left, matches_top_level);
+      if (!output.defined()) {
+        output = TryRewrite(expr, or_pattern->right, matches_top_level);
+      }
+      matches_top_level->pop_back();
+      return output;
+    }
+
+    if (auto opt_matches = ExtractMatchedExpr(pattern, expr, bindings_)) {
+      auto matches = opt_matches.value();
+      for (const auto& pat : *matches_top_level) {
+        matches.Set(pat, expr);
+      }
+
+      Expr rewritten_expr = rewriter_func_(expr, matches);
+      if (!rewritten_expr.same_as(expr)) {
+        return builder_->Normalize(rewritten_expr);
+      }
+    }
+
+    return NullOpt;
+  }
+
   /*! \brief The pattern for rewriting call nodes */
   DFPattern pattern_;
   /*!
--- a/tests/python/relax/test_dataflow_pattern.py
+++ b/tests/python/relax/test_dataflow_pattern.py
@@ -1889,5 +1889,68 @@ def test_wildcard_struct_info_with_symbolic_vars():
     tvm.ir.assert_structural_equal(expected, after)
 
 
+def test_backtrack_if_rewriter_returns_no_op():
+    """Rewriter participates in the pattern matching
+
+    Sometimes, the pattern-matching syntax is insufficient to check if
+    a replacement may be performed.  In this case, the `rewriter`
+    function may perform additional validation.  If this validation
+    fails, the `rewriter` function can return the original expression,
+    and no replacement is performed.
+
+    In addition, when the `rewriter` returns the original expression,
+    the pattern match should backtrack to determine if another branch
+    of the match may have produced a replacement.
+
+    This functionality allows pattern replacements to be composed.
+    """
+
+    pat_match_no_rewrite = is_op("relax.add")(wildcard(), wildcard())
+
+    pat_arg = wildcard()
+    pat_zeros = is_op("relax.zeros")(wildcard())
+    pat_add = is_op("relax.add")(pat_arg, pat_zeros)
+
+    # OR conditions are checked in the order that they occur.  Because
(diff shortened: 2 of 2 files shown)


Lunderberg:
This PR is currently marked as a draft, because it depends on the refactor done in https://github.com/apache/tvm/pull/16730.

src/relax/ir/dataflow_matcher.cc
```diff
@@ -1140,34 +1071,173 @@ class PatternRewriter : ExprMutator {
...
+    return node;
+  }
+
+ private:
+  Optional<Expr> TryRewrite(const Expr& expr, const DFPattern& pattern,
+                            std::vector<DFPattern>* matches_top_level) {
+    ICHECK(matches_top_level);
+
+    // Special handling if the user-supplied pattern is a `OrPattern`.
+    // While the `ExtractMatchedExpr` can handle match the
```
slyubomirsky, line 1175: Looks like a typo. I assume it's supposed to be "handle matching," correct?

slyubomirsky (approved):
The changes seem easy to understand given the PR this is built on and I think the new behavior is intuitive as well: If the first branch of the OR does not result in a rewrite, check the second. In principle, I think this should be noted in whatever top-level documentation there is for the pattern matching grammar.

src/relax/ir/dataflow_matcher.cc
Lunderberg: Whoops, this was a typo.  It should be "handle matching", and I've updated the PR with the correction.

Lunderberg:
The pre-requisite PR https://github.com/apache/tvm/pull/16730 has landed, so this PR is now rebased on top of `main` and marked as ready.  Thank you @slyubomirsky for the review, and so I think it's just waiting on CI now.

MasterJH5574:
Hi @Lunderberg, it seems that this PR introduces new regression. The original issue came from CI in MLC-LLM https://ci.mlc.ai/blue/organizations/jenkins/mlc-llm/detail/PR-2068/1/pipeline/. And I spent some time adapting your test `test_backtrack_if_rewriter_returns_no_op` and reducing it to the minimal I can get.

Running the test below
```python
def test_backtrack_if_rewriter_returns_no_op():
    pat_match_no_rewrite = is_op("relax.add")(wildcard(), wildcard())

    pat_arg = wildcard()
    pat_zeros = is_op("relax.zeros")(wildcard())
    pat_add = is_op("relax.add")(pat_arg, pat_zeros)

    pat = pat_match_no_rewrite | pat_add

    def rewriter(expr, matches):
        print(f"matching {pat} to {matches[pat]}")
        assert isinstance(matches[pat], rx.Call)
        if pat_match_no_rewrite in matches:
            return expr   ## <<== return the expr
        elif pat_add in matches:
            return expr   ## <<== also return the expr
        else:
            raise RuntimeError("Pattern matched, but neither branch matched")

    @R.function(private=True)
    def before():
        with R.dataflow():
            A = R.ones([64, 128], "int32")
            B = R.zeros([64, 128], "int32")
            C = R.add(A, B)

            R.output(C)
        return C

    after = rewrite_call(pat, rewriter, before)
```
will yield the following output:
```
matching OrPattern(Op(relax.add)(*, *) | Op(relax.add)(*, Op(relax.zeros)(*))) to R.add(A, B)
matching OrPattern(Op(relax.add)(*, *) | Op(relax.add)(*, Op(relax.zeros)(*))) to R.add(A, B)
matching OrPattern(Op(relax.add)(*, *) | Op(relax.add)(*, Op(relax.zeros)(*))) to C
[22:36:29] /home/ruihang/Workspace/tvm/src/relax/ir/block_builder.cc:65: Warning: BlockBuilder destroyed with remaining blocks!
  File "/home/ruihang/Workspace/tvm/tests/python/relax/test_dataflow_pattern.py", line 1949, in <module>
    test_backtrack_if_rewriter_returns_no_op()
  File "/home/ruihang/Workspace/tvm/tests/python/relax/test_dataflow_pattern.py", line 1945, in test_backtrack_if_rewriter_returns_no_op
    after = rewrite_call(pat, rewriter, before)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ruihang/Workspace/tvm/python/tvm/relax/dpl/rewrite.py", line 59, in rewrite_call
    return ffi.rewrite_call(pattern, rewriter, func)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "tvm/_ffi/_cython/./packed_func.pxi", line 332, in tvm._ffi._cy3.core.PackedFuncBase.__call__
  File "tvm/_ffi/_cython/./packed_func.pxi", line 263, in tvm._ffi._cy3.core.FuncCall
  File "tvm/_ffi/_cython/./packed_func.pxi", line 252, in tvm._ffi._cy3.core.FuncCall3
  File "tvm/_ffi/_cython/./base.pxi", line 182, in tvm._ffi._cy3.core.CHECK_CALL
  File "/home/ruihang/Workspace/tvm/python/tvm/_ffi/base.py", line 481, in raise_last_ffi_error
    raise py_err
  File "tvm/_ffi/_cython/./packed_func.pxi", line 56, in tvm._ffi._cy3.core.tvm_callback
  File "/home/ruihang/Workspace/tvm/tests/python/relax/test_dataflow_pattern.py", line 1921, in rewriter
    assert isinstance(matches[pat], rx.Call)
AssertionError
```

As you can see in the error message, with this PR, the pattern matcher mapped an `add` to a single Var. This breaks our assumption that what matched is always an `add`. Would you mind helping dig and fix this issue? Thanks in ahead.

MasterJH5574:
The particular `rewrite_call` that fails in MLC-LLM CI is this one https://github.com/mlc-ai/mlc-llm/blob/main/python/mlc_llm/compiler_pass/fuse_ft_dequantize_matmul_epilogue.py#L213-L225

Lunderberg:
Ooh, that's a weird one, and thank you for the test case.  I can reproduce it on my side, and will look in more detail tomorrow.  For now, the first thing I'm noticing is that the failure occurs on the third time that `rewriter` is called, where I'd expect it to only be called twice (once for each branch).

`matches[pat_match_no_rewrite]` is filled during the first call, and `matches[pat_add]` is filled during the second call.  During the third call, which fails the assert, `matches[pat_match_no_rewrite]` is filled again, which seems rather odd.

Lunderberg:
This is an interesting one.

1. The `DFPatternMatcher` will unwrap variables when attempting to find a match ([link](https://github.com/apache/tvm/blob/main/src/relax/ir/dataflow_matcher.cc#L62)), as part of `ExtractMatchedExpr`.
2. The `ExprPatternMatcher` will save the value of each variable binding is saved in order to ([link](https://github.com/apache/tvm/blob/main/src/relax/ir/dataflow_matcher.cc#L1152)).
3. The `ExprPatternMatcher` checks for a match in the override for `Expr VisitExpr(const Expr&)`.

So the call to `rewriter` with the assert fails is made when `ExprPatternMatcher` is checking the variable usage in the body of the `SeqExpr`.  This gets unwrapped by `DFPatternMatcher`, and identified as an expression whose value is equal to the match itself.

This potential bug existed all the way back to https://github.com/apache/tvm/pull/15578.  However, until https://github.com/apache/tvm/pull/16732, the `relax::Var` body of a `SeqExpr` would never be the first match encountered.  Any time a `relax::Var` would have matched, the expression to which it was bound would have matched first, hiding the later match.

I think the resolution is to update condition (1).  Where currently, we unwrap all variable bindings as part of a match, we should instead unwrap all variable bindings except for the top-level of the match.  This could be most easily implemented by having the `ExprPatternMatcher::VisitExpr` skip the call to `TryRewrite` for `relax::Var`.  I'll try this out and see if it resolves the issue.

MasterJH5574:
@Lunderberg Thanks for the swift fix! It works for e2e compilation now.

PR #16735, opened 2024-03-18
[Arith][Fixup] Require feature flag for tighter inequality bounds

Lunderberg:
This is a follow-up to https://github.com/apache/tvm/pull/16588.  Due to an incorrect rebase, the version that was merged into `main` had the tighter `ConstIntBounds` enabled by default, rather than having them implemented in `RewriteSimplifier`, gated behind a feature flag.


Lunderberg:
@vinx13 @tqchen Thank you both for the reviews on #16588, and apologies for accidentally replacing the updated version with an older PR branch.  This PR is a fixup, which changes the implementation to the approved version.

tqchen (approved):

PR #16737, opened 2024-03-19
[SLM] Allow TensorStructInfo to specify parameter in export

Lunderberg:
Prior to this commit, the parameter specification for SLM tensor needed to be passed as a `nn.spec.Tensor`.  As this object is only used to construct a `relax.TensorStructInfo`, and has the same fields as a `relax.TensorStructInfo`, this commit allows the parameter specification to be passed as a `relax.TensorStructInfo`.

### DIFF
--- a/python/tvm/relax/frontend/nn/exporter.py
+++ b/python/tvm/relax/frontend/nn/exporter.py
@@ -138,6 +138,7 @@ class Exporter:
         mod = self.builder.finalize()
         assert rx.analysis.well_formed(mod)
 
+        mod = rx.transform.CanonicalizeBindings()(mod)
         return mod, params, ext_mods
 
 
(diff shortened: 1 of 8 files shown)


tqchen:
As of now, the design rationale of SLM was to make this year reasonably self-contained, without having to leak through the layer below. 

I understand there are some levels of tradeoff here, but would be good to still center around most of the APIs for `nn.spec.Tensor` for now

Lunderberg:
That makes sense, and I agree for the majority of SLM functions.  For `nn.Module.export_tvm`, since the purpose is to translate to the Relax function.  Since it already depends on the Relax IR for the output of `export_tvm`, there's no additional leakage by having the arguments of `export_tvm` be expressed in Relax IR.

PR #16739, opened 2024-03-19
[BugFix][Ansor] Fixing Ansor Gradient Bug

thaisacs:
When Ansor does not find a schedule for a task in warm up, Ansor gradient gets stuck in this task, because there is no optimized schedule for this task. Hence, Ansor does not optimize any task.

Behavior before correction

bug

Behavior after correction

bug1

### DIFF
--- a/python/tvm/auto_scheduler/task_scheduler.py
+++ b/python/tvm/auto_scheduler/task_scheduler.py
@@ -358,6 +358,12 @@ class TaskScheduler:
         self.best_ct = self.ct
         self.best_score = self.cur_score
 
+        # put task without schedule on warm up to dead state
+        if self.strategy == "gradient":
+            for task_idx in range(len(self.tasks)):
+                if(self.best_costs[task_idx] == 1e10):
+                    self.dead_tasks.add(task_idx)
+
         # use the specific strategy to choose workload to tune
         task_idx = -1
         while self.ct < tune_option.num_measure_trials and len(self.dead_tasks) < len(self.tasks):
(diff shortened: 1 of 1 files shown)


thaisacs:
@merrymercy do you think that this make sense?

cbalint13:
Cc @comaniac , @merrymercy , @vinx13 , @jcf94

comaniac:
This change looks a bit hacky and unsafe as you override the cost before calculating the gradients and recover it back afterwards. Intuitively if a task has no schedules, should we just mark it as a dead task and never consider it in the rest of the process?

python/tvm/auto_scheduler/task_scheduler.py
```diff
@@ -358,6 +358,12 @@ def tune(
         self.best_ct = self.ct
         self.best_score = self.cur_score
 
+        # put task without schedule on warm up to dead state
+        if self.strategy == "gradient":
+            for task_idx in range(len(self.tasks)):
+                if(self.best_costs[task_idx] == 1e10):
+                    self.dead_tasks.add(task_idx)
```
comaniac, line 365:
```suggestion
            for task_idx, cost in enumerate(self.best_costs):
                if cost == 1e10:
                    self.dead_tasks.add(task_idx)
```

python/tvm/auto_scheduler/task_scheduler.py
```diff
@@ -358,6 +358,12 @@ def tune(
         self.best_ct = self.ct
         self.best_score = self.cur_score
 
+        # put task without schedule on warm up to dead state
+        if self.strategy == "gradient":
```
comaniac, line 362: I feel this logic applies to all strategies instead of just gradient, so this condition may not be necessary. Could you help confirm?

comaniac (approved):
LGTM

cbalint13:
@tvm-bot rerun

comaniac:
@tvm-bot rerun

comaniac:
Thanks @thaisacs

PR #16745, opened 2024-03-19
[IR] Default to empty attributes, instead of NULL

Lunderberg:
Prior to this commit, the default `DictAttrs` for an `IRModule`, `tir::PrimFunc`, `relax::Function`, and `relay::Function` was a null value.  At each callsite, the absence of a `DictAttrs` needed to be treated as equivalent to an empty `DictAttrs`.  In C++, this typically was done using the `foo->GetAttr` helper function, but in Python it needed to be checked explicitly.  That is, every callsite needed to check `if func.attrs is not None and attr_name in func.attrs`, rather than only checking `if attr_name in func.attrs`.

Since most functions would have at least one attribute to specify the global symbol, these bugs would often surface when working on unrelated changes.

This commit changes the default attribute dictionary from `NullValue<DictAttrs>()` to `DictAttrs()`.  This avoids having two separate representations of an object without any attributes, and allows the `if attr_name in func.attrs` pattern in the Python API.

### DIFF
--- a/include/tvm/ir/attrs.h
+++ b/include/tvm/ir/attrs.h
@@ -230,7 +230,7 @@ class DictAttrs : public Attrs {
    * \brief Consruct a Attrs backed by DictAttrsNode.
    * \param dict The attributes.
    */
-  TVM_DLL explicit DictAttrs(Map<String, ObjectRef> dict);
+  TVM_DLL explicit DictAttrs(Map<String, ObjectRef> dict = {});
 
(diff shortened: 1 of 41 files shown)


Lunderberg:
@tvm-bot rerun

quic-sanirudh (approved):
Thanks a lot for this change. I've run into this issue a couple times as well, so this makes a lot of sense.

PR #16747, opened 2024-03-19
[CI][AArch64] Enable ONNX and PyTorch tests on AArch64

Liam-Sturge:
This makes use of work done in enabling PyTorch and ONNX tests to be run on AArch64 to actually enable the tests to run as part of CI.

Co-Authored-By: Nicola Lancellotti [Nicola.Lancellotti@arm.com](mailto:Nicola.Lancellotti@arm.com)
Co-Authored-By: Leandro Nunes [leandro.Nunes@arm.com](mailto:Leandro.Nunes@arm.com)

cc @leandron @lhutton1 @NicolaLancellotti @tqchen @yongwww

### DIFF
--- a/ci/jenkins/generated/arm_jenkinsfile.groovy
+++ b/ci/jenkins/generated/arm_jenkinsfile.groovy
@@ -60,7 +60,7 @@
 // 'python3 jenkins/generate.py'
 // Note: This timestamp is here to ensure that updates to the Jenkinsfile are
 // always rebased on main before merging:
-// Generated at 2024-01-10T13:15:25.226391
+// Generated at 2024-03-01T17:24:13.590653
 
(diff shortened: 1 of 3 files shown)


tqchen:
I think we should actually avoid running exccessive integration tests across platforms.  The unit tests of frontend should only cover the ingestion part. While the UT of the backend should be constructed from the lower-level IR.

While integration coverage of e2e integration across platforms  is useful, one thing we can start doing is to put these tests as scheduled on nightly as opposed to trigger in each PR

Liam-Sturge:
OK, thanks for your feedback @tqchen. I see your point relating to overdoing the testing. Do we currently have any implementation of nightly tests set up that I could append these tests on to? I didn't see anything when I looked, but I may have missed something.

PR #16750, opened 2024-03-19
[Runtime] CUDA IPC Memory support and custom allreduce kernels

MasterJH5574:
This PR introduces the CUDA IPC memory support in TVM runtime. IPC memory allows multiple distribtued workers accessing the GPU memory of each other directly. This functionality is helpful for implementing customzied communication primitives across distributed workers.

In this PR, we bring the customized all-reduce implementation from TensorRT-LLM into 3rdparty. This all-reduce implementation makes use of the CUDA IPC memory. We expose the all-reduce function in global function under namespace `tvm::runtime::disco::cuda_ipc`.

One unit test for the customized all-reduce kernel over two workers is added.

---

Co-authored-by: Hongyi Jin <hongyij@andrew.cmu.edu>


tests/python/disco/test_custom_allreduce.py
```diff
@@ -0,0 +1,59 @@
...
+    d_input = sess.call_packed(falloc_tensor, d_storage, 0, ShapeTuple(shape), DataType(dtype))
+
+    array_1 = np.arange(12, dtype="float32").reshape(3, 4)
+    array_2 = np.arange(start=1, stop=-11, step=-1, dtype="float32").reshape(3, 4)
+    d_input.debug_copy_from(0, array_1)
+    d_input.debug_copy_from(1, array_2)
+    d_output = sess.empty((3, 4), "float32")
+
+    allreduce_strategy = 1
+    sess.call_packed(fallreduce, d_input, allreduce_strategy, d_output)
```
jinhongyii, line 50: we should test both one shot and two shot

src/runtime/disco/cuda_ipc/cuda_ipc_memory.cc
```diff
@@ -0,0 +1,212 @@
...
+using tensorrt_llm::MAX_RANKS_PER_NODE;
+using tvm::runtime::memory::Buffer;
+
+/*! \brief All-gather the IPC memory handles across all distributed workers. */
+std::vector<cudaIpcMemHandle_t> AllGatherIPCHandles(nccl::CCLThreadLocalContext* ctx,
+                                                    cudaIpcMemHandle_t local_handle) {
+  void *d_src, *d_dst;
+  CUDA_CALL(cudaMalloc(&d_src, CUDA_IPC_HANDLE_SIZE));
+  CUDA_CALL(cudaMalloc(&d_dst, CUDA_IPC_HANDLE_SIZE * ctx->worker->num_workers));
+  CUDA_CALL(cudaMemcpy(d_src, &local_handle, CUDA_IPC_HANDLE_SIZE, cudaMemcpyDefault));
```
tqchen, line 42: cudaMemcpyHostToDevice

src/runtime/disco/cuda_ipc/cuda_ipc_memory.cc
```diff
@@ -0,0 +1,212 @@
...
+namespace runtime {
+namespace cuda_ipc {
+
+using tensorrt_llm::MAX_RANKS_PER_NODE;
+using tvm::runtime::memory::Buffer;
+
+/*! \brief All-gather the IPC memory handles across all distributed workers. */
+std::vector<cudaIpcMemHandle_t> AllGatherIPCHandles(nccl::CCLThreadLocalContext* ctx,
+                                                    cudaIpcMemHandle_t local_handle) {
+  void *d_src, *d_dst;
```
tqchen, line 39: add comment about how we reuse the nccl kernel here

src/runtime/disco/cuda_ipc/custom_allreduce.cc
```diff
@@ -0,0 +1,80 @@
...
+
+namespace tvm {
+namespace runtime {
+namespace nccl {
+namespace cuda_ipc {
+
+using tvm::runtime::cuda_ipc::CUDAIPCMemory;
+
+/*! \brief Compute the size (i.e., number of elements) of the input tensor. */
+inline int TensorSize(const DLTensor* tensor) {
```
tqchen, line 35: use int64_t to avoid large buffer overflow

include/tvm/runtime/disco/cuda_ipc_memory.h
```diff
@@ -0,0 +1,102 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#ifndef TVM_RUNTIME_DISCO_CUDA_IPC_MEMORY_H_
+#define TVM_RUNTIME_DISCO_CUDA_IPC_MEMORY_H_
+
+#include <tvm/runtime/c_runtime_api.h>
```
yongwww, line 23: I have to add `#include <cuda_runtime.h>` to get it built probably because I am using different versions of cmake, etc.

MasterJH5574: Thanks! Included the headers in the .cc files.

tqchen (approved):

include/tvm/runtime/memory/memory_manager.h
```diff
@@ -157,7 +160,7 @@ class StorageObj : public Object {
 /*! \brief reference to storage. */
 class Storage : public ObjectRef {
  public:
-  TVM_DLL explicit Storage(Buffer buffer);
+  TVM_DLL explicit Storage(Buffer buffer, Allocator* allocator);
```
masahi, line 163: Is this constructor actually used? In this PR I only see manual attachment like `storage_obj->allocator = allocator;`.

We started to see a segfault like `free(): invalid pointer` when a python process that runs a script exits, after we integrated recent upstream changes. I'm investigating if the segfault is really due to some recent upstream changes, and which one if so.

tqchen: Yes, this is used so recycle goes back to original allocator, in this case we need to recycle to IPCMmeory in the special allocator

masahi: > Yes, this is used

Can you point out the code? In our fork, I don't see it used outside of `paged_kv_cache.cc`.

MasterJH5574: There are mixture usages of `Storage`/`StorageObj` in TVM. The only use of the constructor seems to be in `paged_kv_cache.cc` in TVM. For other constructions of `Storage`, they all use `make_object` and directly set the allocator
https://github.com/apache/tvm/blob/622bd150dd331780eb41a1c67c65aae802eb9b20/src/runtime/disco/cuda_ipc/cuda_ipc_memory.cc#L204-L210
https://github.com/apache/tvm/blob/622bd150dd331780eb41a1c67c65aae802eb9b20/src/runtime/relax_vm/builtin.cc
https://github.com/apache/tvm/blob/622bd150dd331780eb41a1c67c65aae802eb9b20/src/runtime/vm/vm.cc#L848-L859

There is a use of the constructor on MLC LLM side https://github.com/mlc-ai/mlc-llm/blob/main/cpp/serve/model.cc#L68-L69, where we use the constructor so that we don't need to expand it into the `make_object` lines.

masahi: What worries me is that the code is doing `allocator->Free(buffer);` in the destructor. But how can you be sure that it's safe if there is no guarantee that the allocator is always attached to the storage when it is created?

MasterJH5574: I see. Can we just check the allocator is defined by adding ICHECK in the destructor, or change all `make_object` to use the constructor and add ICHECK inside? If I remember correctly, I made a pass over all occurrences and attached the allocators.

masahi: > Can we just check the allocator is defined by adding ICHECK in the destructor

Yes, I tried that in our fork but the segfault is still there. So this PR is not a cause of our problem.

But still, this doesn't change my concern about this code. I'm sure that you checked all usages, but there is always a possibility to miss some. And you never know about the usages in new code or someone else's fork.

MasterJH5574: Yeah I agree with your concern. I can send a patch to add some checks for more safety later on.

PR #16757, opened 2024-03-20
[SLM] Allow modules to define pre-processing of weights

Lunderberg:
Prior to this commit, the weights used by `nn.Module` instances were required to be `nn.Parameter` instances.  This commit allows the weights to instead be `nn.Tensor` instances, defined in terms of other `nn.Parameter` weights.  This allows a model to define both the original weights that would be present in an external checkpoint (e.g. a Pytorch or Safetensors file), and the pre-processing that should be performed on those weights.

### DIFF
--- a/python/tvm/relax/frontend/nn/core.py
+++ b/python/tvm/relax/frontend/nn/core.py
@@ -591,7 +591,22 @@ def wrap_nested(expr: rx.Expr, name: str) -> Union[Tensor, Sequence[Tensor]]:
         The computed result.
     """
     if not isinstance(expr, rx.DataflowVar):
-        expr = BlockBuilder.current().emit(expr, name)
+        block_builder = BlockBuilder.current()
+        if block_builder is None:
+            # Normalize to make sure we have valid StructInfo, but
+            # wait until we are actually building the function to
+            # flatten nested expressions.
+            #
+            # TODO(Lunderberg): Make this easier to call.  Infering
+            # struct info for a nested expression should be doable in
+            # a free function, without requiring an active
+            # BlockBuilder and an active FunctionFrame.
+            builder = BlockBuilder()
+            with builder.function("dummy_scope", params=[]):
+                expr = builder.normalize(expr)
+                builder.emit_func_output([])
+        else:
+            expr = BlockBuilder.current().emit(expr, name)
     if isinstance(expr.struct_info_, TensorStructInfo):
         return Tensor(_expr=expr)
     if isinstance(expr.struct_info_, TupleStructInfo):
--- a/python/tvm/relax/frontend/nn/exporter.py
+++ b/python/tvm/relax/frontend/nn/exporter.py
@@ -111,7 +111,8 @@ class Exporter:
             return result
 
         # pylint: enable=protected-access
-        params = None
+
+        params = _params()
         effects = _effects()
         ext_mods = self.extern_mods
         with self:
@@ -121,7 +122,6 @@ class Exporter:
                         outputs = _emit_effect_init(self.builder, effects)
                     self.builder.emit_func_output(outputs, params=[])
             for method_name, method_spec in zip(spec.method_names, spec.method_specs):
-                params = _params()  # Re-initialize so symbolic shapes not shared across methods
                 len_args = len(method_spec.arg_specs)
                 len_effects = {
                     "packed": 1,
@@ -135,6 +135,14 @@ class Exporter:
                     with self.builder.dataflow():
                         outputs, inputs = _emit_method(self.builder, method_spec, params, effects)
                     self.builder.emit_func_output(outputs, inputs)
+
+                # TODO(Lunderberg): Make a `ir.transform.ConvertSSA`,
+                # similar to the existing `tir.transform.ConvertSSA`,
+                # that converts an entire module to SSA, including TIR
+                # variable definitions used in either TIR or Relax.
+                mod = self.builder.get()
+                mod[method_name] = rx.utils.copy_with_new_vars(mod[method_name])
+
         mod = self.builder.finalize()
         assert rx.analysis.well_formed(mod)
 
@@ -162,8 +170,6 @@ def _emit_method(  # pylint: disable=too-many-locals,too-many-branches,too-many-
     effects: typing.Optional[typing.List[typing.Tuple[str, core.Effect]]],
 ):
     # pylint: disable=protected-access
-    # symbolic shape's name mapping to its tir.Var for reuse
-    str2var_params: typing.Dict[str, tir.Var] = {}
 
     def _unwrap_ret(expr: typing.Any) -> typing.Any:
         if isinstance(expr, (core.Tensor, core.Object)):
@@ -194,22 +200,9 @@ def _emit_method(  # pylint: disable=too-many-locals,too-many-branches,too-many-
     def _params(mode: str) -> typing.List[rx.Var]:
         inputs: typing.List[rx.Var] = []
 
-        def _get_var(shape_var: tir.Var) -> tir.Var:
-
-            name = shape_var.name
-            if name in str2var_params:
-                return str2var_params[name]
-            var = tir.Var(name, "int64")
-            str2var_params[name] = var
-            return var
-
         for name, param in params:
-            # Make sure the a symbolic shape is not re-registered (same as _method_spec_to_inputs)
-            # e.g. we do not see `vocab_size` for `lm_head` and `vocab_size_1` for `embed_tokens`
-            new_shape = [_get_var(x) if isinstance(x, tir.Var) else x for x in param.shape]
-            var = core.Tensor.placeholder(new_shape, param.dtype, name)._expr
-            inputs.append(var)
-            param._expr = var
+            inputs.append(param._expr)
+
         if mode == "none":
             return []
         if mode == "plain":
--- a/python/tvm/relax/frontend/nn/exporter.py
+++ b/python/tvm/relax/frontend/nn/exporter.py
@@ -299,20 +299,16 @@ def _method_spec_to_inputs(
         return var
 
     def _convert_input(arg_name, arg_spec):
-        if isinstance(arg_spec, rx.TensorStructInfo):
-            return core.Tensor.from_struct_info(arg_spec, name=arg_name)
-        elif isinstance(arg_spec, rx.StructInfo):
-            return core.Object(_expr=rx.Var(arg_name, arg_spec), _name=arg_name)
-        elif isinstance(arg_spec, _spec.Int):
-            return _get_var(arg_name)
+        if isinstance(arg_spec, _spec.Int):
+            arg = _get_var(arg_name)
         elif isinstance(arg_spec, _spec.Tensor):
-            return core.Tensor.placeholder(  # pylint: disable=protected-access
+            arg = core.Tensor.placeholder(  # pylint: disable=protected-access
                 shape=[_get_var(x) if isinstance(x, str) else x for x in arg_spec.shape],
                 dtype=arg_spec.dtype,
                 name=arg_name,
             )
         elif isinstance(arg_spec, _spec.Object):
-            return arg_spec.object_type(_expr=rx.Var(arg_name, ObjectStructInfo()), _name=arg_name)
+            arg = arg_spec.object_type(_expr=rx.Var(arg_name, ObjectStructInfo()), _name=arg_name)
         elif isinstance(arg_spec, _spec.Tuple):
             elements = type(arg_spec.elements)(
                 [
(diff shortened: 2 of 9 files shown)


Lunderberg:
This PR is marked as a draft, as it is based on https://github.com/apache/tvm/pull/16737.  After https://github.com/apache/tvm/pull/16737 lands, this PR will be marked as ready.

Lunderberg:
Separated out the functionality of this PR from the portions of https://github.com/apache/tvm/pull/16737 that require additional discussion.  This PR is now independent, and ready for review.

python/tvm/relax/frontend/nn/core.py
```diff
@@ -591,7 +591,22 @@ def wrap_nested(expr: rx.Expr, name: str) -> Union[Tensor, Sequence[Tensor]]:
...
+            # flatten nested expressions.
+            #
+            # TODO(Lunderberg): Make this easier to call.  Infering
+            # struct info for a nested expression should be doable in
+            # a free function, without requiring an active
+            # BlockBuilder and an active FunctionFrame.
+            builder = BlockBuilder()
+            with builder.function("dummy_scope", params=[]):
+                expr = builder.normalize(expr)
+                builder.emit_func_output([])
```
vinx13, line 607: do we have use case using it outside block builder? emitting a function is not equivalent to emitting the expression itself, unless the function is evaluated

python/tvm/relax/frontend/nn/exporter.py
```diff
@@ -121,7 +122,6 @@ def _effects() -> typing.List[typing.Tuple[str, core.Effect]]:
                         outputs = _emit_effect_init(self.builder, effects)
                     self.builder.emit_func_output(outputs, params=[])
             for method_name, method_spec in zip(spec.method_names, spec.method_specs):
-                params = _params()  # Re-initialize so symbolic shapes not shared across methods
```
vinx13, line 124: personally this is preferable given its explicitness and simplicity (we don't need to do copy the whole function)

python/tvm/relax/frontend/nn/core.py
Lunderberg: The expression isn't actually being generated in the `builder.function("dummy_scope")` here, because it uses `builder.normalize` and not `builder.emit`.  The purpose is to infer the `StructInfo` to be used, but to skip the flattening of nested expressions.

The only reason why the function scope is entered is because the `builder.normalize` method requires an active scope.  This is something I hope to change in the future.

python/tvm/relax/frontend/nn/exporter.py
Lunderberg: Unfortunately, generating fresh Relax variables for the parameters breaks any expressions that have been defined in terms of the previous Relax variables.  (See [this test case](https://github.com/apache/tvm/pull/16757/files#diff-5c2e0984ad1260219560ecdc3bfccb77f92e31ba26246a3ac4e128810c931e71R316) for an example.)  The expressions held internally to represent the pre-processed weights are specified in terms of those initial Relax variables, so replacing them with fresh Relax variables leaves the original variables undefined.  Using `copy_with_new_vars` afterward provides unique symbolic shapes in each function, without breaking the expressions used within each function.

Long-term, I'm planning to expand the `tir.transform.ConvertSSA` pass to handle both TIR and Relax functions.  That way, it can be applied as a post-processing step to ensure all symbolic variables are unique to a single function, without needing to handle it before that.

vinx13 (approved):

python/tvm/relax/frontend/nn/exporter.py
```diff
@@ -176,35 +183,26 @@ def _unwrap_ret(expr: typing.Any) -> typing.Any:
...
+        else:
+            raise TypeError(f"Unsupported input type: {type(arg)}")
 
     def _params(mode: str) -> typing.List[rx.Var]:
         inputs: typing.List[rx.Var] = []
 
-        def _get_var(shape_var: tir.Var) -> tir.Var:
-            name = shape_var.name
-            if name in str2var_params:
-                return str2var_params[name]
```
tqchen, line 196: removing this would cause issues on handling symbolic shape tying remapping and which allows simplicity of mapping string to variables. We should preserve the behavior

python/tvm/relax/frontend/nn/exporter.py
tqchen: i also agree that the original logic was simpler. The intermediate computation sharing was a mechanism that we would like to introduce, would be good to discuss alternative ways to achieve the same goal. E.g. defining a seperate parameter mapping function explicitly rather than having things implicitly

yongwww:
The mixtral compile using mlc-llm with the PR failed, resulting in the following error: 

```
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/mlc-llm/python/mlc_llm/__main__.py", line 47, in <module>
    main()
  File "mlc-llm/python/mlc_llm/__main__.py", line 24, in main
    cli.main(sys.argv[2:])
  File "/mlc-llm/python/mlc_llm/cli/compile.py", line 128, in main
    compile(
  File "/mlc-llm/python/mlc_llm/interface/compile.py", line 232, in compile
    _compile(args, model_config)
  File "/mlc-llm/python/mlc_llm/interface/compile.py", line 178, in _compile
    args.build_func(
  File "/mlc-llm/python/mlc_llm/support/auto_target.py", line 242, in build
    relax.build(
  File "/tvm/python/tvm/relax/vm_build.py", line 335, in build
    mod = pipeline(mod)
          ^^^^^^^^^^^^^
  File "tvm/python/tvm/ir/transform.py", line 238, in __call__
    return _ffi_transform_api.RunPass(self, mod)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "vm/python/tvm/_ffi/_ctypes/packed_func.py", line 239, in __call__
    raise_last_ffi_error()
  File "vm/python/tvm/_ffi/base.py", line 481, in raise_last_ffi_error
    raise py_err
  File "mlc-llm/python/mlc_llm/compiler_pass/pipeline.py", line 172, in _pipeline
    mod = seq(mod)
          ^^^^^^^^
  File "tvm/python/tvm/ir/transform.py", line 238, in __call__
    return _ffi_transform_api.RunPass(self, mod)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "tvm/python/tvm/ir/transform.py", line 307, in _pass_func
    return inst.transform_module(mod, ctx)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "mlc-llm/python/mlc_llm/compiler_pass/dispatch_kv_cache_creation.py", line 94, in transform_module
    kwargs = extract_creation_args(creation_func)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "mlc-llm/python/mlc_llm/compiler_pass/dispatch_kv_cache_creation.py", line 16, in extract_creation_args
    assert len(func.body.blocks[0].bindings) == 2
```

python/tvm/relax/frontend/nn/exporter.py
Lunderberg: Can you describe the behavior that should be preserved?

The main functionality of this step was to ensure symbolic variables are not duplicated across multiple functions, and this functionality is still implemented through `copy_with_new_vars`.  The mapping of strings to variables is still handled [here](https://github.com/apache/tvm/blob/main/python/tvm/relax/frontend/nn/exporter.py#L294), in the `_get_var` function inside `_method_spec_to_input`.

MasterJH5574: Hi @Lunderberg, this PR causes the compilation failure in MLC LLM CI for the Phi model https://ci.mlc.ai/blue/organizations/jenkins/mlc-llm/detail/main/373/

I dug a bit into this. To describe the issue, say the Phi model has hidden size 2560, the final linear layer of the Phi model has weight shape `("vocab_size", 2560)`, and bias shape `("vocab_size",)`.

Prior to this PR, both `"vocab_size"` in the weight and bias are remapped to the same tir.Var, so that for any `x`, `op.matmul(x, op.transpose(weight)) + bias` can have the output shape successfully inferred, represented in `vocab_size`.

Since this PR, the `"vocab_size"` var in the weight and bias shape mismatch. Therefore, the result shape cannot be symbolically inferred, falling back to `R.Tensor(dtype="float16", ndim=3)`, and thus caused the assertion failure as you can see in the CI:

```
  File "/home/ruihang/Workspace/tvm/python/tvm/relax/frontend/nn/modules.py", line 139, in forward
    x = x + self.bias
        ~~^~~~~~~~~~~
  File "/home/ruihang/Workspace/tvm/python/tvm/relax/frontend/nn/_tensor_op.py", line 44, in __add__
    return _op().add(self, other)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ruihang/Workspace/tvm/python/tvm/relax/frontend/nn/op.py", line 105, in add
    return wrap_nested(_op.add(a._expr, b._expr), name)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ruihang/Workspace/tvm/python/tvm/relax/frontend/nn/core.py", line 614, in wrap_nested
    return Tensor(_expr=expr)
           ^^^^^^^^^^^^^^^^^^
  File "/home/ruihang/Workspace/tvm/python/tvm/relax/frontend/nn/core.py", line 117, in __init__
    _check_tensor(_expr)
  File "/home/ruihang/Workspace/tvm/python/tvm/relax/frontend/nn/core.py", line 112, in _check_tensor
    assert expr.struct_info.shape is not None
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError  File "/home/ruihang/Workspace/tvm/python/tvm/relax/frontend/nn/modules.py", line 139, in forward
    x = x + self.bias
        ~~^~~~~~~~~~~
  File "/home/ruihang/Workspace/tvm/python/tvm/relax/frontend/nn/_tensor_op.py", line 44, in __add__
    return _op().add(self, other)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ruihang/Workspace/tvm/python/tvm/relax/frontend/nn/op.py", line 105, in add
    return wrap_nested(_op.add(a._expr, b._expr), name)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ruihang/Workspace/tvm/python/tvm/relax/frontend/nn/core.py", line 614, in wrap_nested
    return Tensor(_expr=expr)
           ^^^^^^^^^^^^^^^^^^
  File "/home/ruihang/Workspace/tvm/python/tvm/relax/frontend/nn/core.py", line 117, in __init__
    _check_tensor(_expr)
  File "/home/ruihang/Workspace/tvm/python/tvm/relax/frontend/nn/core.py", line 112, in _check_tensor
    assert expr.struct_info.shape is not None
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

tqchen:
reverting this PR for now as it deletes the symbolic var remapping logic that caused some compatibility issues. We should followup with a test case that covers the name remapping so hopefully it won;t be broken again

python/tvm/relax/frontend/nn/exporter.py
Lunderberg: Thank you for digging into it, and I'll look into it on Monday.

My first impression is that we shouldn't support this behavior while exporting a module.  If we have a dynamic `Linear` module with dynamic shape parameters `input_features` and `output_features`, we should *not* assume that all such instances of the dynamic `Linear` have the same `input_features` and `output_features`.

I think we should normalize all `str` shapes to the same `tir.Var`, but anything already specified as distinct `tir.Var` shapes should remain separate.  This could later be moved to an improved struct inference, which could provide `R.assert_op` on symbolic shapes, rather than falling back to `R.Tensor(ndim=3)`.

Lunderberg:
Having an explicit remapping function could also work well.  I wanted to avoid having any changes to the `forward` function, and the `__init__` function seemed like the natural point for it.  We could have another function that is called after the parameters are overwritten, but before calling `forward`, if you'd prefer.

python/tvm/relax/frontend/nn/exporter.py
tqchen: having ability to have common dynamic variables like `vocab_size` is a pretty common ndeed, so likely we will need to support this behavior.

Lunderberg: @MasterJH5574 Thank you, and that helps to reproduce the issue.  Does the issue only involve `nn.Linear`, or would it also apply to dynamic shapes that span across multiple independent `nn.Module` subclasses?  If the issue is specific to `nn.Linear`, can you take a look at https://github.com/apache/tvm/pull/16781?  This avoids the duplicate default conversion from `str` to a fresh TIR variable in both the weights and bias.

Overall, I think de-duplicating `str` values makes sense, because Python `str` objects have value equality.  De-duplicating TIR variables by their name doesn't make sense, because TIR variables have reference equality.  I think we could have a good intermediate implementation if we de-duplicate `str` values prior to conversion into TIR variables.

Lunderberg: @MasterJH5574 Can you test against https://github.com/apache/tvm/pull/16785?  It re-implements the same functionality that I require on my side, while maintaining the handling of dynamic shapes specified by strings.

PR #16761, opened 2024-03-21
[SME] Add support for inserting processor state annotations

lhutton1:
Execution of SME instructions requires the processor be in a certain state. This functionality can be can be controlled using LLVM function level annotations such as "aarch64_pstate_sm_enabled" or "aarch64_pstate_za_new" (see arm_utils.py for more information).

This commit exposes this functionality for AArch64 schedules where SME intrinsics will be called. The attributes are intended to be added at the block-level around the compute definition. They are prepended with "pragma" to ensure they remain in the lowering.

In order to detect these attributes and convert them to the relevant LLVM function attributes, a new AArch64 LLVM codegen backend is added. This backend extends the functionality of `codegen_llvm` for AArch64 specific compilation.

Tests to check these attributes propagate correctly have been added.

### DIFF
--- a/python/tvm/topi/arm_cpu/pstate_attributes.py
+++ b/python/tvm/topi/arm_cpu/pstate_attributes.py
@@ -1,0 +1,84 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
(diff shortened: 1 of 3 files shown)


ekalda (approved):
Thanks @lhutton1 looks all good to me!

ekalda:
Thanks @lhutton1!

PR #16766, opened 2024-03-21
[BugFix][Target] Added null check to fix segfault at ->defined() in cpu.cc DetectSystemTriple()

OttoWRas:
I ran into a problem running a very simple ONNX compile, i would get a segfault at a FoldConstantExpr() call from TVMC. **This only happens if the compile flag `set(USE_LLVM OFF)` is OFF.**

```
Thread 1 "python3" received signal SIGSEGV, Segmentation fault.
0x00007fffc94ac78c in tvm::runtime::ObjectPtr<tvm::runtime::Object>::operator!=(decltype(nullptr)) const (this=0x0, null=<error reading variable: Attempt to dereference a generic pointer.>) at /home/otto/tvm/include/tvm/runtime/object.h:470
470       bool operator!=(std::nullptr_t null) const { return data_ != null; }
```

I had compiled TVM Using GCC:
```
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-linux-gnu/11/lto-wrapper
OFFLOAD_TARGET_NAMES=nvptx-none:amdgcn-amdhsa
OFFLOAD_TARGET_DEFAULT=1
Target: x86_64-linux-gnu
Configured with: ../src/configure -v --with-pkgversion='Ubuntu 11.4.0-1ubuntu1~22.04' --with-bugurl=file:///usr/share/doc/gcc-11/README.Bugs --enable-languages=c,ada,c++,go,brig,d,fortran,objc,obj-c++,m2 --prefix=/usr --with-gcc-major-version-only --program-suffix=-11 --program-prefix=x86_64-linux-gnu- --enable-shared --enable-linker-build-id --libexecdir=/usr/lib --without-included-gettext --enable-threads=posix --libdir=/usr/lib --enable-nls --enable-bootstrap --enable-clocale=gnu --enable-libstdcxx-debug --enable-libstdcxx-time=yes --with-default-libstdcxx-abi=new --enable-gnu-unique-object --disable-vtable-verify --enable-plugin --enable-default-pie --with-system-zlib --enable-libphobos-checking=release --with-target-system-zlib=auto --enable-objc-gc=auto --enable-multiarch --disable-werror --enable-cet --with-arch-32=i686 --with-abi=m64 --with-multilib-list=m32,m64,mx32 --enable-multilib --with-tune=generic --enable-offload-targets=nvptx-none=/build/gcc-11-XeT9lY/gcc-11-11.4.0/debian/tmp-nvptx/usr,amdgcn-amdhsa=/build/gcc-11-XeT9lY/gcc-11-11.4.0/debian/tmp-gcn/usr --without-cuda-driver --enable-checking=release --build=x86_64-linux-gnu --host=x86_64-linux-gnu --target=x86_64-linux-gnu --with-build-config=bootstrap-lto-lean --enable-link-serialization=2
Thread model: posix
Supported LTO compression algorithms: zlib zstd
gcc version 11.4.0 (Ubuntu 11.4.0-1ubuntu1~22.04) 
```

This was caused by a call to defined() from DetectSystemTriple() in cpu.cc that was added in #16513. When the previous call
```
auto pf = tvm::runtime::Registry::Get("target.llvm_get_system_triple");
```
would fail, and return null. The consecutive call to defined() would segfault after trying to dereference the null value. This commit adds a check to see if the function pointer is null. This might not be the best solution, but it worked for me, so it might also help someone else struggling with this. Please suggest a better solution, if you know one.

### DIFF
--- a/src/target/parsers/cpu.cc
+++ b/src/target/parsers/cpu.cc
@@ -29,13 +29,11 @@ namespace parsers {
 namespace cpu {
 
 Optional<String> DetectSystemTriple() {
-  #ifdef TVM_INFO_USE_LLVM
   auto pf = tvm::runtime::Registry::Get("target.llvm_get_system_triple");
-  if (pf->defined()) {
-    return (*pf)();
-  }
-  #endif
-  return {};
+  ICHECK (pf != nullptr) 
+  << "The target llvm_get_system_triple was not found, " 
+  "please compile with USE_LLVM = ON";
+  return (*pf)();
 }
 
 TargetJSON ParseTarget(TargetJSON target) {
--- a/src/target/parsers/cpu.cc
+++ b/src/target/parsers/cpu.cc
@@ -29,10 +29,12 @@ namespace parsers {
 namespace cpu {
 
 Optional<String> DetectSystemTriple() {
+  #ifdef TVM_INFO_USE_LLVM
   auto pf = tvm::runtime::Registry::Get("target.llvm_get_system_triple");
-  if (pf != 0 && pf->defined()) {
+  if (pf->defined()) {
     return (*pf)();
   }
+  #endif
   return {};
 }
 
--- a/src/target/parsers/cpu.cc
+++ b/src/target/parsers/cpu.cc
@@ -29,11 +29,13 @@ namespace parsers {
 namespace cpu {
 
 Optional<String> DetectSystemTriple() {
+#ifdef TVM_LLVM_VERSION
   auto pf = tvm::runtime::Registry::Get("target.llvm_get_system_triple");
-  ICHECK (pf != nullptr) 
-  << "The target llvm_get_system_triple was not found, " 
-  "please compile with USE_LLVM = ON";
+  ICHECK(pf != nullptr) << "The target llvm_get_system_triple was not found, "
+                           "please compile with USE_LLVM = ON";
   return (*pf)();
+#endif
+  return {};
 }
 
 TargetJSON ParseTarget(TargetJSON target) {
--- a/src/target/parsers/cpu.cc
+++ b/src/target/parsers/cpu.cc
@@ -30,7 +30,7 @@ namespace cpu {
 
 Optional<String> DetectSystemTriple() {
   auto pf = tvm::runtime::Registry::Get("target.llvm_get_system_triple");
-  if (pf->defined()) {
+  if (pf != 0 && pf->defined()) {
     return (*pf)();
   }
   return {};
--- a/tests/cpp/target_test.cc
+++ b/tests/cpp/target_test.cc
@@ -457,9 +457,13 @@ TEST(TargetCreation, DetectSystemTriple) {
   Target target = Target(config);
   ICHECK_EQ(target->kind, TargetKind::Get("llvm").value());
 
-  Optional<String> mtriple = target->GetAttr<String>("mtriple");
   auto pf = tvm::runtime::Registry::Get("target.llvm_get_system_triple");
-  ASSERT_TRUE(pf->defined());
+  if (pf == nullptr) {
+    GTEST_SKIP() << "LLVM is not available, skipping test";
+  }
+
+  Optional<String> mtriple = target->GetAttr<String>("mtriple");
+  ASSERT_TRUE(mtriple.value() == String((*pf)()));
 }
 
 #endif
--- a/tests/cpp/target_test.cc
+++ b/tests/cpp/target_test.cc
@@ -290,7 +290,7 @@ TEST(TargetCreation, ProcessStrings) {
   ASSERT_EQ(array7[1][1][0], "fred");
 }
 
-#ifdef TVM_INFO_USE_LLVM
+#ifdef TVM_LLVM_VERSION
 // Checks that malformed options cause an assertion.
 TEST(TargetCreation, LLVMCommandLineParseFatalDashDashDash) {
   tvm::codegen::LLVMInstance inst;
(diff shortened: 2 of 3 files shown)


lhutton1:
Thanks for fixing @OttoWRas, would it be possible to add a test somewhere around [here](https://github.com/apache/incubator-tvm/blob/c00cc031de929ba04f46689bb42ced1a3a44d3ef/tests/cpp/target_test.cc#L501) for when LLVM is not available?

OttoWRas:
I just tried to compile the `make cpptest` without LLVM and it even fails compiling, with a missing `tvm::codegen::LLVMTargetInfo` and `tvm::codegen::LLVMInstance`. So I added some compile guarding around the test including LLVM. This also seemed like a better solution for the initial problem, so i changed that fix to a compile guard. 

I'll look into adding a test [here](https://github.com/apache/tvm/blob/c00cc031de929ba04f46689bb42ced1a3a44d3ef/tests/cpp/target_test.cc#L501)

lhutton1:
Hi @OttoWRas, did you get chance to look at adding a test? I should have some time to take a look in the next couple of days if not

OttoWRas:
Hi @lhutton1, I just fixed the test, it actually seemed to already fail with out LLVM, but i just wasn't good enough at using the test system to realize. It seems like the test was misplaced, as there was a guarded area for LLVM dependent test, a few lines above the test. I added an `ASSERT`, instead of the previous check to make the test actually test something.

But looking back on this issue, it seems to me like, `defined()` was either used wrong here, or it should be changed to be able to handle `null` and not only `null_pointer`'s.

I hope this was what you meant, otherwise feel free to make your own changes.

src/target/parsers/cpu.cc
```diff
@@ -29,10 +29,12 @@ namespace parsers {
 namespace cpu {
 
 Optional<String> DetectSystemTriple() {
+  #ifdef TVM_INFO_USE_LLVM
   auto pf = tvm::runtime::Registry::Get("target.llvm_get_system_triple");
   if (pf->defined()) {
```
lhutton1, line 34: You're correct, `pf->defined()` here doesn't really make sense because there will be a segfault when accessing the method. I'm happy to stick with the `#ifdef TVM_LLVM_VERSION` check, but we should probably add an `ICHECK` here as well, similar to other uses in the codebase e.g. https://github.com/apache/tvm/blob/bc1597ce2b643c20b0713ab546b7447459708505/src/relax/backend/contrib/cutlass/codegen.cc#L280

tests/cpp/target_test.cc
```diff
@@ -448,6 +449,21 @@ TEST(TargetCreation, LLVMCommandLineSaveRestore) {
   ASSERT_FALSE(info.MatchesGlobalState());
 }
 
+TEST(TargetCreation, DetectSystemTriple) {
```
lhutton1, line 452: I think we should keep this test outside the `#ifdef TVM_INFO_USE_LLVM` scope so that we get coverage when LLVM is not available. We can use the skip functionality provided by gtest before making a call to LLVM dependent functions (see comment below). That way we're _somewhat_ able to test the changes made in `cpu.cc` this PR makes.

... I say somewhat because the "gpu other" build (the only build in CI that uses USE_LLVM=OFF) doesn't run the cpp test suite. Building and running the tests have been commented out: https://github.com/apache/tvm/blob/a15fd1d6a21e9bedb622ffb9fd025c50624d734d/ci/jenkins/templates/gpu_jenkinsfile.groovy.j2#L57. There is a hint here: https://github.com/apache/tvm/pull/13335/files#r1023416998. But we should probably investigate this further in a separate PR, so we don't block this fix.

tests/cpp/target_test.cc
```diff
@@ -448,6 +449,21 @@ TEST(TargetCreation, LLVMCommandLineSaveRestore) {
...
+  Map<String, ObjectRef> config = {
+      {"kind", String("llvm")},
+  };
+
+  Target target = Target(config);
+  ICHECK_EQ(target->kind, TargetKind::Get("llvm").value());
+
+  Optional<String> mtriple = target->GetAttr<String>("mtriple");
+  auto pf = tvm::runtime::Registry::Get("target.llvm_get_system_triple");
+  ASSERT_TRUE(pf->defined());
```
lhutton1, line 462: Apologies, I'm not too sure what happened to the final part of this test. In its current form, it gets `mtriple` then doesn't use it, that wasn't intentional 😅

I'd suggest updating the test to something like:
```suggestion
  auto pf = tvm::runtime::Registry::Get("target.llvm_get_system_triple");
  if (pf == nullptr) {
    GTEST_SKIP() << "LLVM is not available, skipping test";
  }

  Optional<String> mtriple = target->GetAttr<String>("mtriple");
  ASSERT_TRUE(mtriple.value() == String((*pf)()));
```
which is closer to the original intention of the test.

By skipping at this stage, rather than earlier in the test, we can test that a call to `Target(config)` doesn't cause a segfault when `USE_LLVM=OFF`.

lhutton1 (commented):
Thanks for persisting with this fix @OttoWRas! Overall your changes make sense to me - apologies for introducing the issues in the first place!

I had a closer look at the changes and left some comments which should help get it to pass CI. Let me know if they don't make sense, I'm happy to expand further on them.

src/target/parsers/cpu.cc
```diff
@@ -30,10 +30,10 @@ namespace cpu {
 
 Optional<String> DetectSystemTriple() {
   auto pf = tvm::runtime::Registry::Get("target.llvm_get_system_triple");
-  if (pf->defined()) {
-    return (*pf)();
-  }
-  return {};
+  ICHECK(pf != nullptr)
```
lhutton1, line 33: Apologies for the confusion, we'd still need to wrap this section in `#ifdef TVM_LLVM_VERSION` like before

OttoWRas: Sorry for the inconvenience, I got a bit confused, but i think i fixed it now.

lhutton1: No worries :)

PR #16767, opened 2024-03-22
[TIR][Driver] Use `BindTarget` to specify target for FP8 legalization

slyubomirsky:
This PR cleans up the interface for the FP8 legalization passes by having them use the target attributes that are added by the `BindTarget` pass. This not only removes the need to pass the target as a parameter to those passes but also, in principle, allows for PrimFuncs with different targets to coexist and be handled correctly by the passes. Thanks to @Lunderberg for giving the suggestion.

### DIFF
--- a/src/driver/driver_api.cc
+++ b/src/driver/driver_api.cc
@@ -569,14 +569,13 @@ transform::Sequential MixedModulePassManager(IRModule mixed_mod, Target target) 
 
   Array<Pass> mixed_pass_list;
 
-  mixed_pass_list.push_back(tir::transform::FP8ComputeLegalize(target));
+  mixed_pass_list.push_back(tir::transform::BindTarget(target));
+  mixed_pass_list.push_back(tir::transform::FP8ComputeLegalize());
 
   // VerifyVTCMLimit must occur before LowerVtcmAlloc
   mixed_pass_list.push_back(tir::transform::VerifyVTCMLimit(target));
   // LowerVtcmAlloc must occur after any transformations that modify memory allocation locations
   mixed_pass_list.push_back(tir::transform::LowerVtcmAlloc());
-
-  mixed_pass_list.push_back(tir::transform::BindTarget(target));
 
   mixed_pass_list.push_back(tir::transform::VerifyMemory());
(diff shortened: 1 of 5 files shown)


src/driver/driver_api.cc
```diff
@@ -569,15 +569,14 @@ transform::Sequential MixedModulePassManager(IRModule mixed_mod, Target target)
 
   Array<Pass> mixed_pass_list;
 
-  mixed_pass_list.push_back(tir::transform::FP8ComputeLegalize(target));
+  mixed_pass_list.push_back(tir::transform::BindTarget(target));
```
Lunderberg, line 572: Can we add a comment here specifying that the `BindTarget` should occur first in this sequence, so that later passes can rely on the target attribute being present?  (It looks like both `VerifyVTCMLimit` and `LowerVtcmAlloc` were implemented more recently than `BindTarget`, and probably should have been placed after `BindTarget` at that point.)

Lunderberg (approved):
Looks good to me, and thank you for the follow-up!

Lunderberg:
@csullivan @JosephTheOctonaut FYI on this change, as it relates to your work on FP8.  I don't expect it to impact your work, but wanting to make sure you're aware of it.

src/driver/driver_api.cc
slyubomirsky: Yeah will do

yongwww (approved):

src/driver/driver_api.cc
csullivan: >
> but also, in principle, allows for PrimFuncs with different targets to coexist and be handled correctly by the passes.

Here BindTarget is applying the target to all functions in the mixed IRModule, no?

slyubomirsky: Yeah that's what it does. In principle, we could set the attributes differently.

Lunderberg: Slight nitpick: All functions that do not already have the target attribute.  A function could already have the target attribute, such as defining a module that contains kernels for multiple different targets.

PR #16768, opened 2024-03-22
[RUNTIME][OPENCL] Bugfix for ciImage create with host ptr

srkreddy1238:
Added couple more tests for host ptr data validation


echuraev (approved):
LGTM. Thanks!

PR #16634, opened 2024-02-23
[BugFix] Fix Crash Cases Caused by "__tvm_meta__ = None"

Johnson9009:
Fix https://github.com/apache/tvm/issues/16633

The broken tests are below items of "tests/scripts/task_python_unittest.sh"

"tir-transform"
"tir-usmp"
The crashed cases：
tests/python/tir-transform/test_tir_transform_inject_rolling_buffer.py
tests/python/tir-usmp/test_tir_usmp_algo.py
tests/python/tir-usmp/test_tir_usmp_analysis_extract_bufferinfo.py
tests/python/tir-usmp/test_tir_usmp_transform_convert_pool_allocations_to_offsets.py
tests/python/tir-usmp/test_tir_usmp_utils.py

other failed cases:
tests/python/tir-transform/test_tir_transform_pointer_value_type_rewrite.py::TestRewriteToShuffle::test_compare
tests/python/tir-transform/test_tir_transform_force_narrow_index_to_i32.py::test_thread_axis2
tests/python/tir-transform/test_tir_transform_hoist_if.py::test_hoisting_block_scope_4
tests/python/tir-transform/test_transform_default_gpu_schedule.py::test_add_on_metal

PR #16769, opened 2024-03-22
Revert "[Unity][Parser] Check well-formedness in the parser"

tqchen:
Reverts apache/tvm#16569

### DIFF
--- a/src/tir/ir/data_type_rewriter.cc
+++ b/src/tir/ir/data_type_rewriter.cc
@@ -532,12 +532,6 @@ Stmt IndexDataTypeRewriter::VisitStmt_(const ForNode* op) {
     n->loop_var = new_loop_var;
     n->min = cast(new_loop_var.dtype(), min);
     n->extent = cast(new_loop_var.dtype(), extent);
-    if (op->thread_binding.defined()) {
-      auto old_thread_binding = op->thread_binding.value();
-      auto* ptr = old_thread_binding.CopyOnWrite();
-      ptr->var = old_thread_binding->var.copy_with_dtype(new_loop_var.dtype());
-      n->thread_binding = std::move(Optional<IterVar>(std::move(old_thread_binding)));
-    }
     n->body = new_body;
     return std::move(new_for);
   } else {
--- a/src/tir/transforms/default_gpu_schedule.cc
+++ b/src/tir/transforms/default_gpu_schedule.cc
@@ -113,8 +113,7 @@ bool IsScheduledOnGPU(const BaseFunc& func) {
 
   if (target.defined()) {
     int dev_type = target->GetTargetDeviceType();
-    if (!(dev_type == kDLCUDA || dev_type == kDLMetal || dev_type == kDLROCM ||
-          dev_type == kDLWebGPU)) {
+    if (dev_type != kDLCUDA) {
       return false;
     }
   }
--- a/python/tvm/relax/block_builder.py
+++ b/python/tvm/relax/block_builder.py
@@ -35,12 +35,11 @@ from .utils import gen_call_tir_inputs
 class FunctionScope(object):
     """Auxiliary scope for function"""
 
-    def __init__(self, block_builder, name, params, attrs, is_pure):
+    def __init__(self, block_builder, name, params, attrs):
         self._bb = block_builder
         self._name = name
         self._params = params
         self._attrs = attrs
-        self._is_pure = is_pure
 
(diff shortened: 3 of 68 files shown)


slyubomirsky:
Could we perhaps just revert the driver changes if that's what's causing the MLC failure? They are unrelated to the main functionality of the PR and I would rather have the checking for future tests

Lunderberg (changes requested):
@tqchen Can this PR also add a test case for the behavior that should be preserved?  From your comment [here](https://github.com/apache/tvm/pull/16569#issuecomment-2015458566), MLC is currently relying on untested functionality in TVM, and I don't think we should revert based solely on untested functionality.  Doing so would mean that there would neither be a path forward for re-applying the breaking change, nor any protection for the same breakage occurring in other PRs

I think it makes sense to revert commits that break important workflows, so long as there's a reproducible example of the failure.  Otherwise, the responsibility for reproducing that failure mode falls on a developer who wasn't actually seeing the failure.

Edit: Not requesting a perfect unit test that exercises the specific bug, just a high-level test that would trigger the bug in CI.

tqchen:
@slyubomirsky reverting the driver change also works, just wonder if that will break some of the existing tests

slyubomirsky:
The test should have already been broken, oddly! The error was not triggered by anything I added.

tqchen:
agree that we should have a carved out testcases here, will followup in comment soon, as well as in the original thread

I believe it was along the original lines of comment in the changed order, which i assume we say that  we cannot merge dyn smem beyond the kernel boundary, 

```c++
  // MergeSharedMemoryAllocations must be applied after SplitHostDevice
  // because the merged allocation site is at the beginning of each device function
  mixed_pass_list.push_back(tir::transform::MergeSharedMemoryAllocations());
```

slyubomirsky:
`tests/python/tir-transform/test_tir_transform_inject_ptx_async_copy.py::test_vectorize_cp_async_in_if_then_else` was the test that prompted me to change the ordering and it had been failing for several weeks before my PR, oddly. For some reason it was not triggering in CI though.

tqchen:
OK, given that case we should perhaps first revert the driver order only and disable(then look into the failed tests). 

We will still get a UT for the ordering case, which I assume would might relates to getting that dlight function compiled, and then reduce to a smaller case

src/tir/transforms/default_gpu_schedule.cc
```diff
@@ -113,8 +113,7 @@ bool IsScheduledOnGPU(const BaseFunc& func) {
 
   if (target.defined()) {
     int dev_type = target->GetTargetDeviceType();
-    if (!(dev_type == kDLCUDA || dev_type == kDLMetal || dev_type == kDLROCM ||
-          dev_type == kDLWebGPU)) {
+    if (dev_type != kDLCUDA) {
```
tqchen, line 116: Is this intended change?

slyubomirsky: Yes, this was causing a test failure. I will find the link to the discussion

slyubomirsky: https://github.com/apache/tvm/pull/16634#issuecomment-1979832554

src/tir/ir/data_type_rewriter.cc
```diff
@@ -532,12 +532,6 @@ Stmt IndexDataTypeRewriter::VisitStmt_(const ForNode* op) {
     n->loop_var = new_loop_var;
     n->min = cast(new_loop_var.dtype(), min);
     n->extent = cast(new_loop_var.dtype(), extent);
-    if (op->thread_binding.defined()) {
-      auto old_thread_binding = op->thread_binding.value();
-      auto* ptr = old_thread_binding.CopyOnWrite();
-      ptr->var = old_thread_binding->var.copy_with_dtype(new_loop_var.dtype());
-      n->thread_binding = std::move(Optional<IterVar>(std::move(old_thread_binding)));
-    }
```
slyubomirsky, line 540: This was to address https://github.com/apache/tvm/pull/16634#issuecomment-1973891325

slyubomirsky:
https://github.com/apache/tvm/pull/16770 Smaller-scale reversion

Lunderberg:
Thank you, @tqchen ,  for the follow-up, and for helping to make the more targeted reversion!

tqchen:
closing in favor of https://github.com/apache/tvm/pull/16770

PR #16770, opened 2024-03-22
[Hotfix] Revert driver API pass ordering that breaks MLC, mark failing test

slyubomirsky:
Following up from discussion in https://github.com/apache/tvm/pull/16769, this is a smaller-scale reversion that patches the MLC build by reverting only the changes that broke it (the driver API pass reordering). However, these changes came from https://github.com/apache/tvm/pull/16569 and were needed to fix failing tests. That test has been marked as skippable here but should be fixed along with MLC.

### DIFF
--- a/tests/python/tir-transform/test_tir_transform_inject_ptx_async_copy.py
+++ b/tests/python/tir-transform/test_tir_transform_inject_ptx_async_copy.py
@@ -482,6 +482,12 @@ def test_cp_async_in_if_then_else(postproc_if_missing_async_support):
     assert generated_code == expected_cuda_script
 
 
+@pytest.mark.skip(
+    reason="This test fails due to an ordering issue with MergeSharedMemoryAllocations "
+    "in device_driver_api.cc. However, fixing this causes failures in MLC. "
+    "This bug should be addressed. See discussion in https://github.com/apache/tvm/pull/16769 "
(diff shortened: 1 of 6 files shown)


tests/python/tir-transform/test_tir_transform_inject_ptx_async_copy.py
```diff
@@ -482,6 +482,12 @@ def simple_compute(
     assert generated_code == expected_cuda_script
 
 
+@pytest.mark.skip(
```
Lunderberg, line 485: I like the documentation, both to indicate what needs to be resolved in order to re-enable functionality, and where the existing discussions are

Lunderberg (approved):
Thank you for making the more targeted reversion, and for linking back to the discussions.  Those links are always a big help in tracing back changes after the fact.

slyubomirsky:
Tianqi clarified that only the driver API pass ordering caused the MLC failures, so that's the only change being reverted here now.

MasterJH5574 (approved):
Thanks so much @slyubomirsky @Lunderberg @tqchen. Just checked on my local side: this PR can fix the model compilation error.

tqchen (approved):

slyubomirsky:
The CI error looks to be spurious. We will have to retrigger it.

slyubomirsky:
@tvm-bot rerun

PR #16777, opened 2024-03-24
Revert "[SLM] Allow modules to define pre-processing of weights"

tqchen:
Reverts apache/tvm#16757


Lunderberg:
@tqchen Can you add a unit test demonstrating the behavior to be retained?  Is it only for the string names of symbolic variables, as mentioned in [this comment](https://github.com/apache/tvm/pull/16757#discussion_r1536871492)?

tqchen:
Likely i think it would be primarily two parameters with same `vocab_size` as the dynamic shape.

tqchen:
Happy to also test followup redos of the PR in case we miss other cases

MasterJH5574:
I replied with a test to the other PR before noticing the on-going thread here. https://github.com/apache/tvm/pull/16757#discussion_r1536887971

@tqchen so could you help pasting it into `test_frontend_nn_modules.py`? Then we don't need a separate PR for the test.

MasterJH5574 (approved):

MasterJH5574:
Merged without the unit test to keep the reversion standalone. Let's add the test case separately. Thank you folks!

Lunderberg:
I have a parametrized version of the test case, along with testing for cases beyond `nn.Linear`, that I'll open as a PR.

PR #16778, opened 2024-03-24
[Analysis] Allow calls to GlobalVar in @R.function

Lunderberg:
Prior to this commit, the post-parsing well-formed check performed by TVMScript allowed a call to `GlobalVar` in a `@R.function`, but only if it occurred within the context of a `@I.ir_module`.  If `@R.function` appeared on its own, calls to a `GlobalVar` would be treated as calls to an undefined function.


slyubomirsky (approved):
This is an interesting case, one that did not occur to me. I think it's fine to have a separate rule for analyzing a standalone function compared to an entire module. The restructuring also cleans up some of the code in the parser, so I'm for it.

PR #16780, opened 2024-03-24
[Fix] Fix numpy dtype map

MasterJH5574:
NumPy 2.0 removes the dtype `np.float_`, which may introduces compatibility issue for TVM.

### DIFF
--- a/python/tvm/_ffi/runtime_ctypes.py
+++ b/python/tvm/_ffi/runtime_ctypes.py
@@ -95,8 +95,9 @@ class DataType(ctypes.Structure):
         np.dtype(np.float16): "float16",
         np.dtype(np.float32): "float32",
         np.dtype(np.float64): "float64",
-        np.dtype(np.float_): "float64",
     }
+    if np.__version__.startswith("1."):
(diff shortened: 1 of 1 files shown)


mshr-h:
I think this is better.
```diff
-    if np.__version__.startswith("1."):
-        NUMPY2STR[np.dtype(np.float_)] = "float64"
+    if hasattr(np, "float_"):
+        NUMPY2STR[np.dtype(np.float_)] = "float64"
```

mshr-h:
This is more Pythonic?
```diff
-    if np.__version__.startswith("1."):
-        NUMPY2STR[np.dtype(np.float_)] = "float64"
+ try:
+     NUMPY2STR[np.dtype(np.float_)] = "float64"
+ except AttributeError:
+     pass
```

[LBYL](https://docs.python.org/3/glossary.html#term-LBYL)

tqchen (approved):

MasterJH5574:
@mshr-h Thank you for the great suggestion! I agree that the way I was doing might not be the best. Feel free to send the change you propose.

PR #16781, opened 2024-03-24
[SLM] Provide consistent dynamic shape when specified as python string

Lunderberg:
Resolve a breakage introduced in https://github.com/apache/tvm/pull/16757.  Prior to #16757, distinct TIR variables were unified if they had the same name.  This commit avoids using distinct TIR variables to represent the same user input.


Lunderberg:
This PR is superseded by https://github.com/apache/tvm/pull/16785.

PR #16696, opened 2024-03-11
[SVE] Add codegen support for scalable buffer accesses

lhutton1:
This commit adds support for generating code for scalable loads and stores. It also adds support for the creation of scalable broadcast operations.

Co-authored-by: Elen Kalda <elen.kalda@arm.com>
Co-authored-by: Neil Hickey <neil.hickey@arm.com>

PR #16782, opened 2024-03-25
[SVE] Support scalable vectors in LoopVectorizer

ekalda:
This patch add support for turning loops marked for vectorizing into scalable vectors if the extent of the loop is a vscale dependent expression in a correct form.

The testing for both scalable and fixed length vectors in test_tir_transform.py has been extended and most of the tests have been converted to TVMScript based testing against expected output.

### DIFF
--- a/src/tir/ir/expr.cc
+++ b/src/tir/ir/expr.cc
@@ -197,8 +197,7 @@ TVM_REGISTER_NODE_TYPE(StringImmNode);
 Cast::Cast(DataType t, PrimExpr value, Span span) {
   ICHECK(value.defined());
   ICHECK_EQ(t.get_lanes_or_vscale_factor(), value.dtype().get_lanes_or_vscale_factor());
-  ICHECK((t.is_scalable_vector() == value.dtype().is_scalable_vector()) ||
-         (!t.is_scalable_vector() && !value.dtype().is_scalable_vector()));
+  ICHECK(t.is_scalable_vector() == value.dtype().is_scalable_vector());
   ObjectPtr<CastNode> node = make_object<CastNode>();
   node->dtype = t;
   node->value = std::move(value);
--- a/src/tir/ir/expr.cc
+++ b/src/tir/ir/expr.cc
@@ -196,7 +196,9 @@ TVM_REGISTER_NODE_TYPE(StringImmNode);
 // Cast
 Cast::Cast(DataType t, PrimExpr value, Span span) {
   ICHECK(value.defined());
-  ICHECK_EQ(t.lanes(), value.dtype().lanes());
+  ICHECK_EQ(t.get_lanes_or_vscale_factor(), value.dtype().get_lanes_or_vscale_factor());
+  ICHECK((t.is_scalable_vector() == value.dtype().is_scalable_vector()) ||
+         (!t.is_scalable_vector() && !value.dtype().is_scalable_vector()));
   ObjectPtr<CastNode> node = make_object<CastNode>();
   node->dtype = t;
   node->value = std::move(value);
@@ -354,7 +356,8 @@ And::And(PrimExpr a, PrimExpr b, Span span) {
   ICHECK(a.dtype() == b.dtype()) << "TypeError: mismatched types";
 
   ObjectPtr<AndNode> node = make_object<AndNode>();
-  node->dtype = DataType::Bool(a.dtype().lanes());
+  node->dtype =
+      DataType::Bool(a.dtype().get_lanes_or_vscale_factor(), a.dtype().is_scalable_vector());
   node->a = std::move(a);
   node->b = std::move(b);
   node->span = std::move(span);
@@ -376,7 +379,8 @@ Or::Or(PrimExpr a, PrimExpr b, Span span) {
   ICHECK(a.dtype() == b.dtype()) << "TypeError: mismatched types";
 
   ObjectPtr<OrNode> node = make_object<OrNode>();
-  node->dtype = DataType::Bool(a.dtype().lanes());
+  node->dtype =
+      DataType::Bool(a.dtype().get_lanes_or_vscale_factor(), a.dtype().is_scalable_vector());
   node->a = std::move(a);
   node->b = std::move(b);
   node->span = std::move(span);
@@ -412,7 +416,9 @@ Select::Select(PrimExpr condition, PrimExpr true_value, PrimExpr false_value, Sp
   ICHECK(true_value.defined()) << "ValueError: true_value is undefined";
   ICHECK(false_value.defined()) << "ValueError: true_value is undefined";
   ICHECK(condition.dtype().is_bool());
-  ICHECK(condition.dtype().lanes() == true_value.dtype().lanes() || condition.dtype().lanes() == 1);
+  ICHECK(condition.dtype().get_lanes_or_vscale_factor() ==
+             true_value.dtype().get_lanes_or_vscale_factor() ||
+         condition.dtype().is_scalar());
   ICHECK(false_value.dtype() == true_value.dtype())
       << "TypeError: mismatched types. "
       << "False type: " << false_value.dtype() << "; True type: " << true_value.dtype();
--- a/src/tir/transforms/vectorize_loop.cc
+++ b/src/tir/transforms/vectorize_loop.cc
@@ -54,7 +54,7 @@ inline PrimExpr BroadcastTo(PrimExpr e, int lanes, bool is_scalable) {
   if (const BroadcastNode* op = e.as<BroadcastNode>()) {
     ICHECK(op->dtype.is_scalable_vector() == is_scalable)
         << "Can't broadcast between scalable and fixed length vectors.";
-    int e_lanes = is_scalable ? op->dtype.vscale_factor() : op->dtype.lanes();
+    int e_lanes = op->dtype.get_lanes_or_vscale_factor();
 
     if (lanes % e_lanes == 0) {
       return Broadcast(op->value, CreateNewLanes(is_scalable, lanes));
@@ -203,16 +203,17 @@ class Vectorizer : public StmtMutator, public ExprFunctor<PrimExpr(const PrimExp
       bool is_vec_b = b.dtype().is_scalable_or_fixed_length_vector();
       if (is_vec_a && is_vec_b) {
         // Let's not multiply scalable and fixed length vectors
-        ICHECK(a.dtype().is_scalable_vector() == b.dtype().is_scalable_vector());
+        ICHECK(a.dtype().is_scalable_vector() == b.dtype().is_scalable_vector())
+            << "Fixed length and scalable vectors can't be mixed in multiplication.";
       }
       if (is_vec_a || is_vec_b) {
         const RampNode* b_ramp = b.as<RampNode>();
         const RampNode* a_ramp = a.as<RampNode>();
-        if (a_ramp && !b_ramp && analyzer_.CanProve(b > 0)) {
+        if (a_ramp && b.dtype().is_scalar() && analyzer_.CanProve(b > 0)) {
           PrimExpr lanes = a_ramp->lanes;
           return Ramp(a_ramp->base * b, a_ramp->stride * b, lanes);
         }
-        if (b_ramp && !a_ramp && analyzer_.CanProve(a > 0)) {
+        if (b_ramp && a.dtype().is_scalar() && analyzer_.CanProve(a > 0)) {
           PrimExpr lanes = b_ramp->lanes;
           return Ramp(b_ramp->base * a, b_ramp->stride * a, lanes);
         }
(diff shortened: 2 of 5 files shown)


ekalda:
cc the usual suspects @lhutton1 @leandron @Anndrey24 @tqchen @Lunderberg

src/tir/transforms/vectorize_loop.cc
```diff
@@ -37,19 +37,36 @@
...
+    return e;
+
   if (const BroadcastNode* op = e.as<BroadcastNode>()) {
-    ICHECK(!e.dtype().is_scalable_vector());
-    int broadcast_lanes = static_cast<int>(Downcast<IntImm>(op->lanes)->value);
-    if (lanes % broadcast_lanes == 0) {
-      return Broadcast(op->value, lanes);
+    ICHECK(op->dtype.is_scalable_vector() == is_scalable)
+        << "Can't broadcast between scalable and fixed length vectors.";
+    int e_lanes = is_scalable ? op->dtype.vscale_factor() : op->dtype.lanes();
```
lhutton1, line 57: nit: `get_lanes_or_vscale_factor()`

src/tir/transforms/vectorize_loop.cc
```diff
@@ -182,21 +199,29 @@ class Vectorizer : public StmtMutator, public ExprFunctor<PrimExpr(const PrimExp
...
       return GetRef<PrimExpr>(op);
     } else {
-      // TODO(ekalda): P5 in https://github.com/apache/tvm/issues/16455
-      int lanes = std::max(a.dtype().lanes(), b.dtype().lanes());
-      if (lanes != 1) {
+      bool is_vec_a = a.dtype().is_scalable_or_fixed_length_vector();
+      bool is_vec_b = b.dtype().is_scalable_or_fixed_length_vector();
+      if (is_vec_a && is_vec_b) {
+        // Let's not multiply scalable and fixed length vectors
+        ICHECK(a.dtype().is_scalable_vector() == b.dtype().is_scalable_vector());
```
lhutton1, line 206: Worth adding a message to this?

Lunderberg:
The implementation looks reasonable, though I have one main question for it: What is the behavior of the updated pass for a target that doesn't support SVE?  Prior SVE-commits enabled the functionality, but didn't produce SVE in any of the default lowering passes.

From [this line](https://github.com/apache/tvm/pull/16696/files#diff-f61b04b100f5145f2681340c81d3f2af221239594ed01e2e24896522329ce92cR598-R600), versions of LLVM before 11.0 do not support SVE, nor from my brief reading of the CUDA codegen [here](https://github.com/apache/tvm/blob/main/src/target/source/codegen_cuda.cc#L253) does CUDA.

Since `VectorizeLoop` occurs after the `BindTarget` pass, we can check the function attribute to know which target will be executing each function.  I think we should have the loop vectorization apply only to fixed-extent loops by default, but enable the scalable vectorization for targets that support it.

src/tir/transforms/vectorize_loop.cc
```diff
@@ -182,21 +199,29 @@ class Vectorizer : public StmtMutator, public ExprFunctor<PrimExpr(const PrimExp
...
+      }
+      if (is_vec_a || is_vec_b) {
         const RampNode* b_ramp = b.as<RampNode>();
         const RampNode* a_ramp = a.as<RampNode>();
-        if (a_ramp && b.dtype().lanes() == 1 && analyzer_.CanProve(b > 0)) {
-          int lanes = static_cast<int>(Downcast<IntImm>(a_ramp->lanes)->value);
+        if (a_ramp && !b_ramp && analyzer_.CanProve(b > 0)) {
```
lhutton1, line 211: Would it be better to use `b.dtype().is_scalar()` here? (just because it's closer to the original statement... I'm unsure if `!b_ramp` relaxes this check)

src/tir/transforms/vectorize_loop.cc
```diff
@@ -182,21 +199,29 @@ class Vectorizer : public StmtMutator, public ExprFunctor<PrimExpr(const PrimExp
...
+          PrimExpr lanes = a_ramp->lanes;
           return Ramp(a_ramp->base * b, a_ramp->stride * b, lanes);
         }
-        if (b_ramp && a.dtype().lanes() == 1 && analyzer_.CanProve(a > 0)) {
-          int lanes = static_cast<int>(Downcast<IntImm>(b_ramp->lanes)->value);
+        if (b_ramp && !a_ramp && analyzer_.CanProve(a > 0)) {
```
lhutton1, line 215: same as above

src/tir/transforms/vectorize_loop.cc
```diff
@@ -433,20 +488,27 @@ class Vectorizer : public StmtMutator, public ExprFunctor<PrimExpr(const PrimExp
...
-      int index_lanes = other_index_lanes * indices[indices.size() - 1].dtype().lanes();
+      int lanes_in_last_index = indices[indices.size() - 1].dtype().get_lanes_or_vscale_factor();
+      int index_lanes = other_index_lanes * lanes_in_last_index;
 
       // The total number of lanes in this store operation.  Either
       // the index or the value will be broadcast out to this number
       // of lanes, depending on which has more lanes.
-      int total_lanes = std::max(index_lanes, value.dtype().lanes());
+      int value_dtype_lanes = value.dtype().get_lanes_or_vscale_factor();
+      bool is_last_index_scalable = indices[indices.size() - 1].dtype().is_scalable_vector();
```
lhutton1, line 510: nit: might be nicer to replace uses of `indices[indices.size() - 1].dtype()` with a `last_index_dtype` variable

src/tir/transforms/vectorize_loop.cc
```diff
@@ -635,19 +701,22 @@ class Vectorizer : public StmtMutator, public ExprFunctor<PrimExpr(const PrimExp
...
     } else {
-      int lanes = std::max(a.dtype().lanes(), b.dtype().lanes());
+      int a_lanes = a.dtype().get_lanes_or_vscale_factor();
+      int b_lanes = b.dtype().get_lanes_or_vscale_factor();
+      int lanes = std::max(a_lanes, b_lanes);
       if (lanes != 1) {
         const RampNode* b_ramp = b.as<RampNode>();
         const RampNode* a_ramp = a.as<RampNode>();
-        if (a.dtype().lanes() == 1 && b_ramp) {
+        if (!a.dtype().is_scalable_or_fixed_length_vector() && b_ramp) {
```
lhutton1, line 710: `is_scalar`?

src/tir/transforms/vectorize_loop.cc
```diff
@@ -635,19 +701,22 @@ class Vectorizer : public StmtMutator, public ExprFunctor<PrimExpr(const PrimExp
...
           return Ramp(fcompute(a, b_ramp->base),
                       fcompute(make_zero(b_ramp->stride.dtype()), b_ramp->stride), b_ramp->lanes);
         }
-        if (b.dtype().lanes() == 1 && a_ramp) {
+        if (!b.dtype().is_scalable_or_fixed_length_vector() && a_ramp) {
```
lhutton1, line 714: same as above

lhutton1 (commented):
Nice work, there are plenty of non-trivial changes here! Thanks for tidying up the tests as well :) The comments are largely just nitpicks 😅

ekalda:
Thank you for your feedback @Lunderberg, much appreciated!

> The implementation looks reasonable, though I have one main question for it: What is the behavior of the updated pass for a target that doesn't support SVE? Prior SVE-commits enabled the functionality, but didn't produce SVE in any of the default lowering passes.
> 
> From [this line](https://github.com/apache/tvm/pull/16696/files#diff-f61b04b100f5145f2681340c81d3f2af221239594ed01e2e24896522329ce92cR598-R600), versions of LLVM before 11.0 do not support SVE, nor from my brief reading of the CUDA codegen [here](https://github.com/apache/tvm/blob/main/src/target/source/codegen_cuda.cc#L253) does CUDA.

When it comes to targets that don't support SVE, I'd expect these targets to not trigger the creation of scalable vectors. In the current plan the creation of scalable vectors has to be intentional, i.e. it comes from splitting an axis by a `vscale` dependent expression in the (target dependent) schedules and vectorizing the resulting axis. If the `LoopVectorizer` is trying to create scalable vectors for target that doesn't support it, something has gone wrong and the compilation will fall over at some point:
* If by some mistake a schedule that doesn't support VLA programming contains `vscale`, it will fall over latest in a target dependent codegen
* If there is an attempt to vectorize loops with non-int extent that doesn't contain `vscale`, the "scalable ramp" creation will error since it expects the `PrimExpr lanes` in a form `vscale * int`. I realize though that this is a weird deviation from a current behaviour of
```
      if (!extent_as_int || extent_as_int->value < 1) {
        LOG(FATAL) << "Failed to vectorize loop with extent " << op->extent;
      }
```
so I'll modify the patch such that it checks for a target and fails as before if the extent is not an int.

> Since `VectorizeLoop` occurs after the `BindTarget` pass, we can check the function attribute to know which target will be executing each function. I think we should have the loop vectorization apply only to fixed-extent loops by default, but enable the scalable vectorization for targets that support it.

In principle I'm not against making vectorizing for scalable vectors functionality more explicitly target specific, but it is not obvious to me what that would mean in terms of code? `ICHECK`s for the appropriate targets at the places where scalable vectors are created?

src/tir/transforms/vectorize_loop.cc
```diff
@@ -37,19 +37,36 @@
...
+    return Mul(Call(DataType::Int(32), builtin::vscale(), {}), lanes_or_vscale_factor);
+  } else {
+    return lanes_or_vscale_factor;
+  }
+}
+
+inline PrimExpr BroadcastTo(PrimExpr e, int lanes, bool is_scalable) {
+  // Check if e is already in the expected form
+  if (e.dtype().get_lanes_or_vscale_factor() == lanes &&
+      e.dtype().is_scalable_vector() == is_scalable)
```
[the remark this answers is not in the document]
lhutton1: Ignore this comment

cbalint13:
Trying to reason loudly on this,

> When it comes to targets that don't support SVE, I'd expect these targets to not trigger the creation of scalable vectors. In the current plan the creation of scalable vectors has to be intentional, i.e. it comes from splitting an axis by a vscale dependent expression in the (target dependent) schedules and vectorizing the resulting axis. 

See, so the **schedules** are already target dependent but the underlaying execution of such schedule is not. 
Well, in this case there would be no need for more additional extra checks I think, but correct me please if I got this wrong.

Lunderberg:
That's a good point.  This only applies for cases that would currently produce an error, where a `ForKind::kVectorized` is applied to a loop with dynamic extent.

I'm convinced, and agree that this change doesn't introduce any errors that would not already have been errors from an upstream pass.  We may still want to have validation at this step, to avoid making errors be more complicated downstream, but that can be added later as needed.

src/tir/ir/expr.cc
```diff
@@ -196,7 +196,9 @@ TVM_REGISTER_NODE_TYPE(StringImmNode);
 // Cast
 Cast::Cast(DataType t, PrimExpr value, Span span) {
   ICHECK(value.defined());
-  ICHECK_EQ(t.lanes(), value.dtype().lanes());
+  ICHECK_EQ(t.get_lanes_or_vscale_factor(), value.dtype().get_lanes_or_vscale_factor());
+  ICHECK((t.is_scalable_vector() == value.dtype().is_scalable_vector()) ||
+         (!t.is_scalable_vector() && !value.dtype().is_scalable_vector()));
```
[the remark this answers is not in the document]
ekalda: Oops yes, done

tests/python/tir-transform/test_tir_transform_vectorize.py
```diff
@@ -64,28 +61,86 @@ def test_vectorize_vector():
...
+
+def test_vectorize_vector_scalable_error2():
+    @I.ir_module
+    class Module:
+        @T.prim_func
+        def main(A: T.Buffer((25,), "float32xvscalex4")):
+            for j in T.vectorized(4):
+                A[j] = T.Broadcast(T.float32(1), T.vscale() * 4)
+
+    with pytest.raises(tvm.error.InternalError):
```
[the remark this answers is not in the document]
ekalda: Done

src/tir/transforms/vectorize_loop.cc
ekalda: Done

src/tir/transforms/vectorize_loop.cc
ekalda: Yes let's keep it as close as the original behaviour as possible :)

src/tir/transforms/vectorize_loop.cc
ekalda: Done

src/tir/transforms/vectorize_loop.cc
ekalda: Done

src/tir/transforms/vectorize_loop.cc
ekalda: Done

src/tir/transforms/vectorize_loop.cc
ekalda: Done

lhutton1 (approved):
Thanks for the updates @ekalda, LGTM!

lhutton1:
Thanks @ekalda @Lunderberg @cbalint13! Let's investigate adding the target dependent error message in a follow-up

PR #16783, opened 2024-03-25
[Relax] Improve CanonicalizeBindings in DataflowVar edge case

Lunderberg:
If there is a trivial binding of `Var = DataflowVar`, but the non-dataflow variable is never used outside the dataflow block in which is is declared, then we should keep the name of the upstream `DataflowVar`, as it is more likely to be the human-readable name (e.g. a function parameter).

### DIFF
--- a/src/relax/transform/canonicalize_bindings.cc
+++ b/src/relax/transform/canonicalize_bindings.cc
@@ -96,15 +96,16 @@ class CanonicalizePlanner : public ExprVisitor {
         // Case 1: Var = Var
         // Case 2: DataflowVar = Var
         // Case 3: DataflowVar = DataflowVar
-        // Case 4a: Var = DataflowVar, but used outside this DataflowBlock
+        // Case 4a: Var = DataflowVar, where the Var is not used
+        //          outside the DataflowBlock containing the binding
         //
-        // For these three cases, the trivial binding can be
-        // unwrapped, using the bound variable directly at the point
-        // of use.
+        // For these four cases, the trivial binding can be unwrapped,
+        // using the bound variable directly at the point of use.
         plan.replace_usage.Set(bound_var->vid, bound_to);
         plan.bindings_to_remove.insert(bound_var->vid);
       } else {
-        // Case 4b: Var = DataflowVar, and used outside this DataflowBlock
+        // Case 4b: Var = DataflowVar, where the Var is used somewhere
+        //          outside the DataflowBlock containing the binding
         //
         // Replacing a Var with a DataflowVar could result in illegal
         // use of a DataflowVar outside of a DataflowBlock.  Instead,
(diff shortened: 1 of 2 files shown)


src/relax/transform/canonicalize_bindings.cc
```diff
@@ -91,18 +91,20 @@ class CanonicalizePlanner : public ExprVisitor {
...
 
-      if (bound_var.as<DataflowVarNode>() || !bound_to.as<DataflowVarNode>()) {
+      if (bound_var.as<DataflowVarNode>() || !bound_to.as<DataflowVarNode>() ||
+          !visitor.used_outside_home_dataflow_.count(bound_var)) {
         // Case 1: Var = Var
         // Case 2: DataflowVar = Var
         // Case 3: DataflowVar = DataflowVar
+        // Case 4a: Var = DataflowVar, but used outside this DataflowBlock
         //
         // For these three cases, the trivial binding can be
```
slyubomirsky, line 101:
```suggestion
        // For these four cases, the trivial binding can be
```
:wink:

slyubomirsky, line 101:
```suggestion
        // For these four cases, the trivial binding can be
```
:wink:

src/relax/transform/canonicalize_bindings.cc
```diff
@@ -91,18 +91,20 @@ class CanonicalizePlanner : public ExprVisitor {
         bound_to = opt.value();
       }
 
-      if (bound_var.as<DataflowVarNode>() || !bound_to.as<DataflowVarNode>()) {
+      if (bound_var.as<DataflowVarNode>() || !bound_to.as<DataflowVarNode>() ||
+          !visitor.used_outside_home_dataflow_.count(bound_var)) {
         // Case 1: Var = Var
         // Case 2: DataflowVar = Var
         // Case 3: DataflowVar = DataflowVar
+        // Case 4a: Var = DataflowVar, but used outside this DataflowBlock
```
slyubomirsky, line 99:
```suggestion
        // Case 4a: Var = DataflowVar, but not used outside this DataflowBlock
```
It appears the text was copied from case 4b. I also think it would be clearer to write out that it is the var that is not used outside the DF block

slyubomirsky (changes requested):
This is a nice improvement, though there's a significant typo that should be fixed before we merge :sweat_smile: Thank you!

src/relax/transform/canonicalize_bindings.cc
Lunderberg: Thank you, and updated the comment to be more explicit.  I've also changed "this DataflowBlock" to "the DataflowBlock containing the binding", since this function is called after the entire function is visited, not during the visit of any specific dataflow block.

Lunderberg:
@tvm-bot rerun

slyubomirsky (approved):
Thanks for the correction.

PR #16784, opened 2024-03-25
[SLM] Add unit tests for SLM to Relax exporter

Lunderberg:
Follow-up to https://github.com/apache/tvm/pull/16777, add unit tests demonstrating desired behavior.

### DIFF
--- a/tests/python/relax/test_frontend_nn_exporter.py
+++ b/tests/python/relax/test_frontend_nn_exporter.py
@@ -27,7 +27,7 @@ from tvm.script import ir as I, relax as R, tir as T
 
 
 def test_simple():
-    """A module may be exported from nn.Module to Relax"""
+    """The nn.modules.* may be exported from nn.Module to Relax"""
 
     slm_mod = nn.modules.ReLU()
     exported_mod, _ = slm_mod.export_tvm(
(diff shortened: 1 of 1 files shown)


tests/python/relax/test_frontend_nn_exporter.py
```diff
@@ -0,0 +1,632 @@
...
+                relu = R.nn.relu(x)
+                relu = relu
+                R.output(relu)
+            return relu
+
+    assert_structural_equal(exported_mod, Expected)
+
+
+def test_custom_module():
+    """A module may be exported from nn.Module to Relax"""
```
slyubomirsky, line 53: Perhaps the comment should differ in soem way from that in the above test.

slyubomirsky (approved):
It's good to have test cases, especially when they explain the intended functionality very well. A tutorial featuring based on these examples (perhaps literally generated from the test cases) might be a good investment of time as well, especially if it can be made more visible.

tests/python/relax/test_frontend_nn_exporter.py
Lunderberg: Good call, and I've updated the docstring to remove the copy/paste duplicate.

Lunderberg:
> It's good to have test cases, especially when they explain the intended functionality very well. A tutorial featuring based on these examples (perhaps literally generated from the test cases) might be a good investment of time as well, especially if it can be made more visible.

Thank you, and that was in part my goal with the sequence of unit tests.  Whenever possible, I prefer test cases that double as mini-tutorials, since then they are less likely to become stale than full tutorials.  (Though I agree that it would be beneficial to have a tutorial that follows user-focused flow, rather than the feature-focused flow of these unit tests.)

MasterJH5574 (approved):
Thank you @Lunderberg 🙏

PR #16790, opened 2024-03-26
[Fix] Fix build errors with VS2022

Jiawei-Shao:
This patch fixes all the build errors from VS2022. With this patch we can build tvm.dll successfully with VS2022.


Jiawei-Shao:
@tqchen PTAL, thanks!

tqchen (approved):

tqchen:
Thank you @Jiawei-Shao

PR #16795, opened 2024-03-26
[TIR] Ramp and Broadcast lanes fixed to int32 dtype

quic-sanirudh:
When Ramp and Broadcast nodes are created with fixed length lanes,
they're fixed to int32 dtype since DLDataType always supports only
uint16 lanes.

### DIFF
--- a/tests/python/arith/test_arith_rewrite_simplify.py
+++ b/tests/python/arith/test_arith_rewrite_simplify.py
@@ -75,6 +75,7 @@ class BaseCompare:
 
 class TestVector(BaseCompare):
     x, y, z = te.var("x"), te.var("y"), te.var("z")
+    x64 = te.var("x", dtype="int64")
     vx = te.var("vx", dtype="int32x2")
     vc = te.var("vc", dtype="uint1")
     test_case = tvm.testing.parameter(
@@ -88,6 +89,7 @@ class TestVector(BaseCompare):
         ),
         TestCase(y.astype("int32x2") + x.astype("int32x2"), (y + x).astype("int32x2")),
         TestCase(tvm.tir.Broadcast(0, 4) + y, tvm.tir.Broadcast(y, 4)),
+        # int64 lanes
         TestCase(
             tvm.tir.Broadcast(x, 4) + tvm.tir.Ramp(0, 1, tvm.tir.IntImm(dtype="int64", value=4)),
             tvm.tir.Ramp(x, 1, 4),
@@ -95,6 +97,11 @@ class TestVector(BaseCompare):
         TestCase(
             tvm.tir.Broadcast(x, tvm.tir.IntImm(dtype="int64", value=4)) + tvm.tir.Ramp(0, 1, 4),
             tvm.tir.Ramp(x, 1, 4),
+        ),
+        # int64 iterators with int32 lanes
+        TestCase(
+            tvm.tir.Broadcast(x64, 4) + tvm.tir.Ramp(tvm.tir.IntImm(dtype="int64", value=0), 1, 4),
+            tvm.tir.Ramp(x64, 1, 4),
         ),
         TestCase(
             tvm.tir.Broadcast(0, tir.vscale() * 8) + y, tvm.tir.Broadcast(y, tir.vscale() * 8)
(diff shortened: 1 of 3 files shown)


tqchen:
In this particular case, i think type do matter, so it would be great instead to fix the cases that would depend on the relaxed behavior.

quic-sanirudh:
> In this particular case (deep equality), i think type do matter, so it would be great instead to fix the cases that would depend on the relaxed behavior.
> 
> I know we had some i64/i32 issues, and general rule of thumb now is to try to be explicit as much as possible and that helps to reduce errors

Oh okay, thanks for the feedback @tqchen. The cases we started seeing was that some expressions were not getting simplified properly after [`RampNode` lanes were changed to PrimExpr](https://github.com/apache/tvm/commit/a6157a6369c184b6fa5f66654feb685e58726737#diff-046cdcb6494a6719465080bb9156cd4620828af4b18f7018e5b443d6c7c1c1d0L792-R792). I narrowed down the exact simplification that was failing was actually a [rewrite simplify rule here](https://github.com/apache/tvm/blob/main/src/arith/rewrite_simplify.cc#L401). 

I realized that the simplification was not happening because lanes between broadcast and RampNode in this case had different types, so a couple other solutions I thought would apply here is to either fix the RampNode constructor to stick to some fixed dtype for lanes (something like int32/int16, since `DLDataType` anyways only supports int16 dtype), or to update the simplify rules here to try the same rules with an `PVar<IntImm>` lanes type in case of fixed length vectors.

But if dtype does matter, then should we update the `PEqualChecked<IntImm>` to also check for dtypes?

tqchen:
ah oK, i think in this case we should try to come up with a rule for lanes. I think having a fixed dtype probably makes sense then we handle cast for related cases

quic-sanirudh:
> Yeah I think fixing the dtype is a good idea, it would hopefully avoid this kind of problems in the future as well. Out of interest, what were the mismatching dtypes of the two compared `IntImmNode`s that you observed @quic-sanirudh?

Thanks @ekalda. I'll update the PR to fix the dtypes in RampNode (and perhaps the broadcast node as well).

The dtypes in my case were `int32` and `int64`. The expression I saw was something like this (slightly simpler version)
`T.Broadcast(c, 128) + T.Ramp(T.int64(0), T.int64(1), T.int64(128))`

The RampNode seems to get the int64 lanes because the all the iterators in our case is by default int64, but the broadcast seems to be inserted during the [evaluation of AddNode in op.cc here](https://github.com/apache/tvm/blob/d43e1ab71d5d9e16bbc962d4d7952dcc7a1cdbca/src/tir/op/op.cc#L126-L139)

quic-sanirudh:
I've updated the PR to fix the dtype of lanes as we discussed. Let me know if this looks good.

ekalda (approved):
Looks good to me, thank you for the fix @quic-sanirudh! Hopefully we won't see these kind of silent simplification failures in the future now 🤞

quic-sanirudh:
Thanks for the review @ekalda

tqchen:
@quic-sanirudh can you add a test case about i64 iterator type with Ramp? I know fixing to i32 likely will resolve issues when most programs are i32 iterator, we should make sure it is i64 compatible

tests/python/arith/test_arith_rewrite_simplify.py
```diff
@@ -88,6 +88,14 @@ class TestVector(BaseCompare):
         ),
         TestCase(y.astype("int32x2") + x.astype("int32x2"), (y + x).astype("int32x2")),
         TestCase(tvm.tir.Broadcast(0, 4) + y, tvm.tir.Broadcast(y, 4)),
+        TestCase(
+            tvm.tir.Broadcast(x, 4) + tvm.tir.Ramp(0, 1, tvm.tir.IntImm(dtype="int64", value=4)),
+            tvm.tir.Ramp(x, 1, 4),
+        ),
+        TestCase(
+            tvm.tir.Broadcast(x, tvm.tir.IntImm(dtype="int64", value=4)) + tvm.tir.Ramp(0, 1, 4),
```
tqchen, line 96: I actually meant that the value being broadcasted and ramp base being i64. but lanes remains i32, this would be a more common case in our setting

tqchen (changes requested):

tests/python/arith/test_arith_rewrite_simplify.py
quic-sanirudh: ah okay, sorry for the confusion. I'll update the test cases to check that as well.

quic-sanirudh: @tqchen I've updated the test case, let me know if this looks good.

quic-sanirudh:
@tvm-bot rerun

tqchen (approved):

PR #16798, opened 2024-03-26
[Relax][Transform] Provide callback versions of LazyTransformParams

Lunderberg:
Prior to this commit, the `LazyTransformParams` function could be used to load model parameters on demand.  However, the function used to load or set parameters needed to be registered within the global registry of `PackedFunc`s.  This PR provides `LazyGetInput` and `LazySetOutput` transforms, which perform the lazy-loading through a `R.Callable` callback argument, rather than through a globally-registered `PackedFunc`.


Lunderberg:
The prerequisite https://github.com/apache/tvm/pull/16642 has now landed, so this PR is ready for review.

slyubomirsky:
Odd, the changes in this PR are still showing those that I presume were introduced in #16642. Maybe it's a Github thing.

src/relax/transform/lazy_transform_params.cc
```diff
@@ -0,0 +1,277 @@
...
+// Pass LazyTransformParams() {
+//   auto pass_func = [](Function func, IRModule, PassContext) -> Function {
+//     LazyInput mutator;
+//     return Downcast<Function>(mutator(func));
+//   };
+//   return CreateFunctionPass(/*pass_function=*/pass_func,
+//                             /*opt_level=*/0,
+//                             /*pass_name=*/"MutateOpsForTraining",
+//                             /*required=*/{});
+// }
```
slyubomirsky, line 273: I doubt we want to keep this much commented-out code.

python/tvm/relax/transform/transform.py
```diff
@@ -303,6 +303,26 @@ def LambdaLift() -> tvm.ir.transform.Pass:
     return _ffi_api.LambdaLift()
 
 
+def LazyGetInput() -> tvm.ir.transform.Pass:
+    """A pass that requests inputs lazily
```
slyubomirsky, line 307: I think it would be preferable to have more detail in the doc comment as to what exactly the output of the pass will look like (the same is true for `LazySetOutput` too).

src/relax/transform/lazy_transform_params.cc
```diff
@@ -0,0 +1,277 @@
...
+  Expr VisitExpr_(const FunctionNode* func) override {
+    if (plan_.has_value()) {
+      return ExprMutator::VisitExpr_(func);
+    }
+
+    int64_t num_input_params = GetNumInputParams(func).value_or(0);
+
+    std::unordered_map<Var, size_t, ObjectPtrHash, ObjectPtrEqual> param_lookup;
+    for (size_t i = 0; i < func->params.size(); i++) {
+      param_lookup.insert({func->params[i], i - num_input_params});
```
slyubomirsky, line 64: What is the second value supposed to be? If `i` starts at 0 and `num_imput_params` is non-negative, you're going to get a negative value with weird implicit conversions or just underflow on `size_t` (I don't know the C++ spec well enough to know what happens but I'm surprised there's no compiler warning).

slyubomirsky (changes requested):
Thank you for this change. Most of it was not hard to follow, though the indexing in one case raised an eyebrow for me. I think the pass descriptions should mention how the callback is supposed to work. We should probably also have small end-to-end tests that constructs examples of callbacks and uses them (the callbacks could be practically no-ops, just enough to verify that it works).

It might also be good to have an occasional line in the transformation passes explaining what is being constructed, though I was able to follow it.

src/relax/transform/lazy_transform_params.cc
Lunderberg: Thank you, and removing.  My intent is to replace the existing `LazyTransformParams` to be in terms of the C++ implementation, but didn't get around to it.

python/tvm/relax/transform/transform.py
Lunderberg: Good call, and I've updated the docstrings for each.

src/relax/transform/lazy_transform_params.cc
Lunderberg: The second parameter is the index of the model weight, ignoring any additional parameters that are provided at runtime.  So, you can have additional small parameters (e.g. `rank: R.Prim('int64')` or `num_gpus: R.Prim('int64')`), which are kept separate from the bundled parameters in the callback.

That said, I'm re-reading it and am not sure how this indexing worked in the first place.  The `i - num_input_params` would convert `i` to a signed integer, but that should cause some erroneous usage when used.  It's enough to raise my concerns on it as well, so that's enough to warrant changing it.

Updating the loop to start at `size_t i=num_input_params` instead of `size_t i=0`.

slyubomirsky (approved):
Thank you for addressing my comments. It would still be good to have some end-to-end tests that show what an example of such a callback would look like, if you're willing to do it.

Lunderberg:
Thank you, and good point on the end-to-end examples.  I'm planning to upstream a use case, where the callback would load data from a `.safetensors` file, and that would be a good place to add a practical example overall.

PR #16800, opened 2024-03-27
[BugTIR]fix error merging shared memory for ptx_cp_async

JackWeiw:
In driver_api.cc,  
```
// MergeSharedMemoryAllocations must be applied after SplitHostDevice
 // because the merged allocation site is at the beginning of each device function
```
howerver MergeSharedMemoryAllocations didn't take care of call op `ptx_cp_async`,
result in error
[script to reproduce](https://gist.github.com/JackWeiw/d7c804744dc7fa0fc17d6d6f69997057) 
image
as we can see it didn't successfully merge shared memory in `ptx_cp_async`
image


JackWeiw:
CC @vinx13

vinx13 (approved):

PR #16801, opened 2024-03-27
[Relax] Allow DeadCodeElimination within ApplyPassToFunction

Lunderberg:
The `tvm.ir.transform.ApplyPassToFunction` allows a transform to be applied selectively to some portions of a `IRModule`, without applying to the entire `IRModule`.  For example, to apply an optimization pass (e.g. `relax.transform.ExpandMatmulOfSum`) or an interface-altering pass (e.g. `relax.transform.BundleModelParams`) to specific functions.  It does so by generating an intermediate `IRModule` containing only the functions specified, applying the transform to that intermediate, then merging the results.

When using `ApplyPassToFunction` to apply `DeadCodeElimination`, or a pipeline containing `DeadCodeElimination`, this intermediate `IRModule` may contain calls to `GlobalVar` instances that are not within the intermediate `IRModule`.  Prior to this commit, this resulted in an error being thrown when collecting the call graph. This commit updates `DeadCodeElimination` to instead handle incomplete call-graph collection.

### DIFF
--- a/src/relax/transform/dead_code_elimination.cc
+++ b/src/relax/transform/dead_code_elimination.cc
@@ -50,12 +50,22 @@ class CallTracer : public ExprVisitor {
   explicit CallTracer(IRModule mod) : mod_{mod}, called_funcs_{}, visiting_{} {}
 
   void VisitExpr_(const GlobalVarNode* op) final {
-    called_funcs_.insert(GetRef<GlobalVar>(op));
-    auto func = mod_->Lookup(op->name_hint);
-    if (const auto* function_node = func.as<FunctionNode>()) {
-      VisitExpr(GetRef<Function>(function_node));
+    auto gvar = GetRef<GlobalVar>(op);
+    called_funcs_.insert(gvar);
+    if (auto func = mod_->functions.Get(gvar)) {
+      if (const auto* function_node = func.as<FunctionNode>()) {
+        VisitExpr(GetRef<Function>(function_node));
+      }
+      // else: Don't visit PrimFuncs -- we don't need to collect any tir.Calls therein.
+    } else {
+      // The GlobalVar is not contained in the IRModule.  While the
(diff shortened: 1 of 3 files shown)


slyubomirsky:
Hmmm. This approach seems fine for the specific case of DCE, but I wonder if the design of `ApplyPassToFunction` needs further thought, since the fact that it can output malformed modules could be an issue for other passes as well. Maybe the pass infrastructure/interfaces could try to accommodate the case of not processing the entire module? A shorter-term change might be ensuring that all module-level passes check whether a function is defined when looking up global vars (we could make it a helper function in the `ExprMutator` base, even, since the examples I found were using the `ExprMutator`'s `mod_` field for this purpose). Cursory search for other examples of directly calling `Lookup` on `mod_`: `BindSymbolicVars`, `FuseTIR`, `RewriteCudaGraph`, `MergeCompositeFunctions`, `LambdaLift`, `DetectRecursion` (analysis), with a few others that are related to training or autotuning.

We should be careful about introducing another possible pitfall to pass design (in this case having to accommodate a violated invariant).

Lunderberg:
Agreed, it would be good to avoid having too many special cases.  The intent was to allow optimization passes, or API-altering passes, to be part of an end-to-end pipeline while only impacting a few functions within the module.  (e.g. Applying `LazyGetInput` to the result of `LiftTransformParams`, but not to any of the inference functions.)

If it requires more special cases in general, I think the `ApplyPassToFunction` could be improved by replacing all non-matching functions with `relax::ExternFunc`, rather than filtering them out altogether.  That way, the inner transform could operate on a well-formed `IRModule`, but wouldn't be able to mutate other functions.

slyubomirsky (approved):
Might as well approve these changes since `ApplyPassToFunction` is already there and they address it. I don't think a huge redesign would necessarily be in order, but we should give some thought to the intended design for it.

PR #16804, opened 2024-03-27
[Runtime] Introduce MSCCLPP with NCCL equivalent interface

csullivan:
Currently only supports ncclAllreduce equivalent, with op=Sum up to 2**24 bytes as noted below.
* Adds a fast and simple AllReduce kernel (sum only) using mscclpp smChannel scratch for small reductions up to 2**24 bytes.

Plan to integrate with disco in a follow up PR.


tqchen (approved):

tqchen:
Thanks @csullivan . looking forward to the disco integration

PR #16805, opened 2024-03-27
[Bugfix][Relax] BlockBuilder may not assume unique input functions

Lunderberg:
Prior to this commit, the implementation of `relax::BlockBuilder::AddFunction` implicitly assumed that the input `IRModule` does not contain duplicate copies of the same function. This commit updates the implementation, removing the reliance on this assumption.  This commit resolves the error by tracking all `GlobalVar` that map to the same function, rather than an just one.

A well-formed IRModule may contain duplicate function definitions. This is rare, as most functions can be disambiguated by the the function attribute `tvm::attr::kGlobalSymbol`.  However, private functions do not have this attribute, and a well-formed IRModule may contain multiple copies of the same function.

The regression test added in this PR calls `BlockBuilder::UpdateFunc` and `BlockBuilder::AddFunc` in a specific order to reproduce this issue.  In practice, this failure was sporadic, depending on the order in which a transformation pass visited functions in a module.  This was first observed in `VMShapeLower`, with sporadic errors depending on the order of iteration over `mod->functions`.

### DIFF
--- a/src/relax/ir/block_builder.cc
+++ b/src/relax/ir/block_builder.cc
@@ -35,6 +35,7 @@
 
 #include <memory>
 #include <unordered_map>
+#include <unordered_set>
 #include <vector>
 
 #include "../../node/ndarray_hash_equal.h"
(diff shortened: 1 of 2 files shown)


tqchen:
In this case, I think we should perhaps fix the pass instead if that is a rare case. The main consideration is the added cost and complexity

Lunderberg:
I don't think there's any indication of this being a bug in the pass itself.  I also don't know if it's a rare case, only that I encountered it within one specific pass.  I think this is a bug in the general `relax::BlockBuilder` utility, as nothing in the API suggests that use of `AddFunc` and `UpdateFunc` within the same transform is unsupported.  If we want to avoid complexity in the `relax::BlockBuilder`, then we should remove the function-level de-duplication from `relax::BlockBuilder` altogether, and handle it in a dedicated transform.

tqchen (approved):

tqchen:
OK, after looking through again i agree, so merging this

PR #16806, opened 2024-03-27
[Debug] Improve error message in VMShapeLower

Lunderberg:
If `VMShapeLower` raises an error, specify which function produced the error.


slyubomirsky (approved):
Good to have better error messages.

PR #16807, opened 2024-03-27
[Disco] Support setting workers' CPU affinity

MasterJH5574:
This PR supports setting the CPU affinity for disco workers. Specifically, a global function `runtime.disco.bind_worker_to_cpu_core` is added to allow accepting a list of CPU ids, and then set the CPU affinity for workers.

This can potentially reduce the OS scheduling overhead that increases the disco worker pthread conditional waiting time before being waken up.


tqchen (approved):

MasterJH5574:
@tvm-bot rerun

tqchen:
let us disable the failed hexagon test first since it seems to be flaky

MasterJH5574:
> let us disable the failed hexagon test first since it seems to be flaky

@tqchen Thanks. Just disable.d

PR #16808, opened 2024-03-28
[Runtime][LLVM] Fix errors during loading of target tags

cbalint13:
This PR fixes TVM loading errors when lower LLVM version does not support TVM predefined target tags.

Report and related discussions are here: https://github.com/apache/tvm/pull/16425#issuecomment-2023292845

---

#### Rationale:

 * During tvm load/initialization [list_tags()](https://github.com/apache/tvm/blob/83e7e9b2eb8dbeeb16dcfdbaf3336caa81071877/python/tvm/target/tag.py#L71) checks all ```src/target/tag.cc``` target definitions
 * The target tags **should be protected** against improper LLVM versions at their declaration time

#### Other fixes:

* Enforce the ```mcpu=generic``` [fallback assignment](https://github.com/apache/tvm/pull/16808/files#diff-f13c79368edd27b0c05d044df267746187448906d4310d5617bce89003b8813bR230-R231), so LLVM will not noise us with improper invocation 
* Minor compilation fix for LLVM version ranges: ~~10~~ 11 < llvm < ~~16~~ 13 (introduced by SVE support)
* Add a clear hint with LLVM_VERSION_STRING to the error message for the user

Cc @Lunderberg , @lhutton1

### DIFF
--- a/src/target/llvm/codegen_llvm.cc
+++ b/src/target/llvm/codegen_llvm.cc
@@ -1871,7 +1871,11 @@ llvm::Value* CodeGenLLVM::VisitExpr_(const BroadcastNode* op) {
   value = builder_->CreateInsertElement(undef, value, zero);
 #if TVM_LLVM_VERSION >= 110
   llvm::ElementCount ec =
+#if TVM_LLVM_VERSION >= 120
       llvm::ElementCount::get(dtype.get_lanes_or_vscale_factor(), dtype.is_scalable_vector());
+#else
+      llvm::ElementCount(dtype.get_lanes_or_vscale_factor(), dtype.is_scalable_vector());
+#endif
   llvm::Constant* mask = llvm::ConstantVector::getSplat(ec, zero);
 #else
   ICHECK(!dtype.is_scalable_vector())
--- a/python/tvm/target/tag.py
+++ b/python/tvm/target/tag.py
@@ -67,9 +67,6 @@ def register_tag(name: str, config: Dict[str, Any], override: bool = False) -> O
     return None
 
 
-# To check the correctness of all registered tags, the call is made in library loading time.
-list_tags()
-
 # We purposely maintain all tags in the C++ side to support pure C++ use cases,
 # and the Python API is only used for fast prototyping.
 register_tag(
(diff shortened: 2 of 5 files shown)


src/target/llvm/codegen_llvm.cc
```diff
@@ -1871,7 +1871,11 @@ llvm::Value* CodeGenLLVM::VisitExpr_(const BroadcastNode* op) {
   value = builder_->CreateInsertElement(undef, value, zero);
 #if TVM_LLVM_VERSION >= 110
   llvm::ElementCount ec =
+#if TVM_LLVM_VERSION >= 160
       llvm::ElementCount::get(dtype.get_lanes_or_vscale_factor(), dtype.is_scalable_vector());
+#else
+      llvm::ElementCount(dtype.get_lanes_or_vscale_factor(), dtype.is_scalable_vector());
+#endif
```
lhutton1, line 1878: Curious about this change, I tried building with llvm-13 locally but it failed. Also it seems the build for ci-cpu (which uses llvm-15) also failed: https://ci.tlcpack.ai/blue/organizations/jenkins/tvm-cpu/detail/PR-16808/2/pipeline. What version of llvm caused the failure?

lhutton1 (commented):
Thanks for taking a look at this and the quick fix @cbalint13! I had a question, but otherwise looks good to me

src/target/llvm/codegen_llvm.cc
cbalint13: @lhutton1 

* Updated now.
```
+#if TVM_LLVM_VERSION >= 120
       llvm::ElementCount::get(dtype.get_lanes_or_vscale_factor(), dtype.is_scalable_vector());
+#else
+      llvm::ElementCount(dtype.get_lanes_or_vscale_factor(), dtype.is_scalable_vector());
+#endif
```

I double checked compilation with llvm={10,11,12,13,14,15,16,17,18} .

lhutton1: Thanks, that works now :)

cbalint13: @lhutton1 

> Thanks, that works now :)

Also thanks for the catch !

Sorry for the confusion my double checks took **some longer time**
The threshold of  ```getter``` turned to be at ```>= 120``` after exhaustive checks.

lhutton1 (approved):
LGTM, I'll defer merging to @Lunderberg

PR #16811, opened 2024-03-28
[TVMScript] Do not throw error for duplicate definitions

Lunderberg:
TVM's IR dialects require single-site assignment. However, the printer is different from most utilities, as it is often used for debugging when runtime assumptions are violated.  Therefore, it may neither assume that its input is well-formed, nor may it throw an exception if the input is ill-formed, and must print an accurate representation of the underlying IR for use in debugging.  When debugging, being unable to print an ill-formed `IRModule` prevents a developer from determining why the IRModule is ill-formed.

### DIFF
--- a/src/script/printer/ir_docsifier.cc
+++ b/src/script/printer/ir_docsifier.cc
@@ -30,7 +30,21 @@ namespace script {
 namespace printer {
 
 IdDoc IRDocsifierNode::Define(const ObjectRef& obj, const Frame& frame, const String& name_hint) {
-  ICHECK(obj2info.find(obj) == obj2info.end()) << "Duplicated object: " << obj;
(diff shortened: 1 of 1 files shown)


slyubomirsky (approved):
Good idea and good point about treating the printer as a debugging tool, though I think logging a warning or some kind of indication that something is wrong would be useful to add.

Lunderberg:
I went back and forth on it, as the existing `ICHECK` could have been changed to a `LOG(WARNING)`.  However, that would prevent it from generating clean error messages for cases that are known to be ill-formed.  (e.g. Highlighting the ill-formed output, as it done for the TIR well-formed checker and which I want to bring over to the Relax well-formed checker.)

slyubomirsky:
Glad to hear there's a plan.

PR #16814, opened 2024-03-28
Fix includes of custom allreduce kernel

vinx13:
cc @yongwww


yongwww (approved):
LGTM, thanks for the fix!

PR #16822, opened 2024-03-29
[VM] Recycle VMFrame

tqchen:
This PR recycles the VMFrame in VM which can help a bit when function involves large frames.


MasterJH5574 (approved):
LGTM.

PR #16815, opened 2024-03-29
[Relax] Enable capturing symbolic shapes in cuda graph

vinx13:
cc @tqchen

PR #16823, opened 2024-03-29
[Relax][VM] Refactor CUDA graph builtins as VM extension

vinx13:
This PR introduce VM extension, which allow users to define target related runtime extensions. The VM holds the reference to the extension once created to ensure it is destructed at VM destruction.

CUDA graph builtins are refactored as VM extension to make the cuda graph cache per-VM instead of per-thread, to allow VM to be run on different threads (non-concurrently).

This PR need rebase after #16815 

cc @tqchen


tqchen:
@vinx13 let us disable the failed hexagon test first, since it seems to be a flaky one

tqchen (approved):

PR #16824, opened 2024-03-30
[KVCache] Introducing auxiliary data manager

MasterJH5574:
This PR introduces class `PagedKVCacheAuxDataManager` for PagedKVCache. This class manages all the integer auxiliary data required for paged attention and other KV cache operations, such as page table arrays, position arrays, etc..

The purpose of introducing this class is because prior to this PR, for each auxiliary array we issue a host-to-device copy. This may cause extra overhead, since these auxiliary array are usually lightweight. One simple idea is to "merge" all the auxiliary arrays into a single one, and taking slices of this large array for each original auxiliary array. By doing this, we enable to issue only one single host-to-device copy for the auxiliary arrays altogether.

The intrduction of `PagedKVCacheAuxDataManager` abstracts the interface that PagedKVCache copies host arrays to device arrays, enabling us to support both the previous way of copying and the new way.

To support slicing for attention-related TIR functions, we introduce `elem_offset` match in TIR functions in this PR.
This PR also bumps FlashInfer to support the auxiliary array slicing.


MasterJH5574:
Depending on flashinfer-ai/flashinfer#193. Mark as draft for now.

src/runtime/relax_vm/paged_kv_cache.cc
```diff
@@ -636,9 +938,17 @@ class PagedAttentionKVCacheObj : public AttentionKVCacheObj {
...
+      // Set the copy stream for copy.
+      DeviceAPI::Get(device_)->SetStream(device_, copy_stream_);
+    }
     for (int layer = 0; layer < num_layers_; ++layer) {
       f_copy_single_page_(pages_[layer], src_page_id, tgt_page_id, copy_length);
     }
+    if (copy_stream_ != compute_stream_) {
+      // Set the compute stream back.
+      DeviceAPI::Get(device_)->SetStream(device_, compute_stream_);
+    }
```
MasterJH5574, line 951: In this way we execute the copy on a separate copy stream. cc @cyx-6

tqchen (approved):

PR #16831, opened 2024-04-02
[KVCache] Reducing CacheAuxDataManager copy size

MasterJH5574:
The cached KV cache auxiliary data manager turns out introducing much extra copy size due to improper handling of array offsets. Specifically, prior to this PR, the manager always align the start of each offset to the largest possible. As a result, in each copy there are quite a lot of unnecessary elements getting copied.

This PR reduces the copy size to the minimal by aligning properly. This significantly reduces the copy size.


MasterJH5574:
cc @vinx13

tqchen (approved):

vinx13 (approved):

PR #16833, opened 2024-04-03
[Relax][Analysis] Validate FuncStructInfo correctness

Lunderberg:
Prior to this PR, the well-formed checker did not validate the `FuncStructInfo` annotations.  While these `FuncStructInfo` annotations are generated during construction, they can be invalidated by `CopyOnWrite()`.  This commit adds validation of the `FuncStructInfo` to the well-formed checker.

### DIFF
--- a/src/relax/analysis/well_formed.cc
+++ b/src/relax/analysis/well_formed.cc
@@ -90,11 +90,14 @@ class WellFormedChecker : public relax::ExprVisitor,
         WellFormedChecker(obj.as<IRModule>(), check_struct_info);
 
     if (const auto* mod = obj.as<IRModuleNode>()) {
-      for (const auto& it : mod->functions) {
+      for (const auto& [gvar, base_func] : mod->functions) {
+        well_formed_checker.CheckGlobalVarStructInfoConsistency(gvar, base_func);
+        well_formed_checker.CheckParamStructInfoConsistency(gvar, base_func);
+
         // visit relax.Function
-        if (auto* n = it.second.as<FunctionNode>()) {
-          Function func = GetRef<Function>(n);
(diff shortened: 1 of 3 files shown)


slyubomirsky (approved):
Good catch, also good to have regression tests.

Lunderberg:
Thank you!

CI has found a surprising number of failures for this check.  It looks like the new check requiring the same `GlobalVar` and `Function` struct info has caught an issue with TVMScript.  The GlobalVar instances are generated based solely on the function signature, where the `Function` instances can also infer the return type from the body.

For now, I think I'll make the check looser, using `IsBaseOf` instead of `StructuralEqual`.  That would require the `GlobalVar` to be an accurate representation of the function, without requiring it to be a complete representation of the function.

Lunderberg:
Well, a check that finds a large number of errors is a good check to have, and a good source of cleanup.  Unfortunately, I think it means that this PR will need to sit until I have time to clean them up.

The current CI failures are due to transforms that update a function's signature, but do not propagate that change to the `GlobalVar`.  There's enough failures that it should have a common utility to resolve the mismatch, updating the `GlobalVar` of the function, and replacing it at all call sites.  Resolving this cannot be done as part of the block builder's normalizer, because it would depend on the order in which the caller/callee are visited.

Lunderberg:
This PR has been stalled for a while, in part because there are a large number of passes that will need to be updated.  The `ReplaceGlobalVar` function, introduced in https://github.com/apache/tvm/pull/17202, should help to resolve these breakages.  That way, when a function's signature is changed, we'd be able to easily update the `GlobalVar` to have the new `StructInfo`.

PR #16836, opened 2024-04-03
[Runtime] Allow inspection of function names from a compiled .so

Lunderberg:
Prior to this commit, the `LLVMModuleNode` provided a `"get_func_names"` function that would return the names of functions available within the result of `tvm.build`.  However, this utility was not preserved across a round-trip through `mod.export_library` and `tvm.runtime.load_module`.

This commit adds a similar `"__get_func_names"` function to a `DSOLibrary`, which returns the symbols available for use in the library.  This is exposed in the `Module.keys()` method, mimicking the interface that a Python user would expect.

In addition, this is used to improve the error message shown when a library does not contain the requested function.  The `difflib` module (from Python's stdlib) is used to find functions that are contained in the module, and have similar names to the one requested.

### DIFF
--- a/src/runtime/dso_library.cc
+++ b/src/runtime/dso_library.cc
@@ -32,8 +32,6 @@
 #include <windows.h>
 #else
 #include <dlfcn.h>
-#include <elf.h>
-#include <link.h>
 #endif
 
 #if defined(__hexagon__)
@@ -53,7 +51,6 @@ namespace runtime {
 class DSOLibrary final : public Library {
  public:
   ~DSOLibrary();
-
   /*!
    * \brief Initialize by loading and storing
    * a handle to the underlying shared library.
@@ -61,7 +58,6 @@ class DSOLibrary final : public Library {
    * shared library over which to initialize.
    */
   void Init(const std::string& name);
-
   /*!
    * \brief Returns the symbol address within
    * the shared library for a given symbol name.
@@ -70,40 +66,18 @@ class DSOLibrary final : public Library {
    */
   void* GetSymbol(const char* name) final;
 
-  /*! \brief List symbols available within the module
-   *
-   * \param callback The callback to be executed for each symbol in
-   *    the library.
-   */
-  void ListSymbols(std::function<void(const char*)> callback) final;
-
  private:
   /*! \brief Private implementation of symbol lookup.
-   *
-   * Implementation is operating system dependent.
-   *
-   * \param name The name of the symbol.
-   *
+   *  Implementation is operating system dependent.
+   *  \param The name of the symbol.
    * \return The symbol.
    */
   void* GetSymbol_(const char* name);
-
-  /*! \brief Private implementation of symbol lookup.
-   *
-   * Implementation is operating system dependent.
-   *
-   * \param callback The callback for each symbol located.
-   *
-   * \return The symbol.
-   */
-  void ListSymbols_(std::function<void(const char*)> callback);
-
   /*! \brief Implementation of shared library load.
    *  Implementation is operating system dependent.
    *  \param The name/path of the shared library.
    */
   void Load(const std::string& name);
-
   /*! \brief Implementation of shared library unload.
    *  Implementation is operating system dependent.
    */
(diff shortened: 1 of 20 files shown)


tqchen:
seems this behavior was platform dependent, and ideally we should not introduce behavior that only works on one platform but not others. Keep the minimal necessary functionalities is preferred here

src/runtime/dso_library.cc
```diff
@@ -136,6 +166,50 @@ void DSOLibrary::Load(const std::string& name) {
 
 void* DSOLibrary::GetSymbol_(const char* name) { return dlsym(lib_handle_, name); }
 
+void DSOLibrary::ListSymbols_(std::function<void(const char*)> callback) {
+  // Adapted from https://stackoverflow.com/a/62205128
+
+  struct link_map* map = nullptr;
+  dlinfo(lib_handle_, RTLD_DI_LINKMAP, &map);
+
+  Elf64_Sym* symbol_table = nullptr;
```
tqchen, line 175: I am not sure if this will work on windows

tqchen (changes requested):

src/runtime/dso_library.cc
Lunderberg: This particular implementation definitely wouldn't work on Windows, and is in the `#ifdef` for the linux-specific code.  Currently, the Windows implementation returns an empty list.

Lunderberg:
> seems this behavior was platform dependent, and ideally we should not introduce behavior that only works on one platform but not others.

My goal was partly to avoid needing to alter an existing `IRModule` to support this functionality.  Since the `Library` interface already abstracts over the platform-specific behavior, it makes sense for additional platform-specific to be implemented by extending the `Library` interface.

Lunderberg:
As an alternative implementation for the same end-user functionality, what would you think about a lowering pass that is applied just before `tir.transform.MakePackedAPI()`?  That could generate the `__get_func_names` function within the `IRModule`, and wouldn't require platform-specific support to read symbols from the shared library.

tqchen:
the only main issue about generate approach is that it may not be comprehensive if we in the end link multiple exports together, which is a less frequent usecase as in static library case(where we append the symbol prefix).

I think this is more of an optional feature rather than mandatory feature, so we should avoid exposing it as things like `keys`, but maybe use some of the info if available for error reporting

Lunderberg:
> the only main issue about generate approach is that it may not be comprehensive if we in the end link multiple exports together, which is a less frequent usecase as in static library case(where we append the symbol prefix).

That's a good point, though I believe we'll have the correct behavior by default.  For both the `mod.imported_modules` and the static libraries, the user should only be calling the functions through the externally-exposed wrappers in `mod`.  These are the names present in the `IRModule` itself.

(As an extension, we could indicate which functions are present within the static and/or imported modules, but that would be a later extension.)

> I think this is more of an optional feature rather than mandatory feature, so we should avoid exposing it as things like `keys`, but maybe use some of the info if available for error reporting

I think we want to expose it to the user, for use debugging.  If a user receives an error message with suggested typo corrections, or stating that there no similarly named functions, the obvious next step is to look at what functions actually are available.  Since `runtime.Module` already provides `__getitem__` like a python dictionary, other dictionary-like methods such as `.keys()` would be the obvious place for a new developer to look.

Lunderberg:
I've reverted the implementation of this PR, but kept the same unit tests.  The platform-dependent implementation is entirely removed, replaced with a TIR lowering pass.

Lunderberg:
Looks like the re-implemented version has trouble with RPC devices and with microtvm runtime.  In order to compile the metadata function, I needed to add the ability to return a `tir::StringImm` from a `PrimFunc`, where previously this was only used to provide `const char*` arguments.  There's a difference in how `TVMArgValue` and `TVMRetValue` represent `kTVMStr`: in `TVMArgValue` it stores a non-owning `const char*` in `TVMValue::v_str`, while in `TVMRetValue` it stores an owning `std::string*` in `TVMValue::v_handle`.

There is no straightforward way to perform the equivalent of `rv.v_handle = new std::string("string literal");` in generated LLVM code, and it would be very implementation-dependent.  Instead, I implemented `T.ret("string literal")` by delegating to a PackedFunc to allocate the `TVMRetValue`, then returning that value.  This works very nicely for most use cases, but fails for the microtvm runtime (because the backend helper function is not present) and for RPC devices (because the string's buffer doesn't seem to be returned).

I think the ability to query a module's contents is too useful to set aside entirely, but this PR is going on the back burner for me until I can find a better way to implement it.

PR #16841, opened 2024-04-03
[Relax] Provide well-formed output in `transform.LazyGetInput`

Lunderberg:
Prior to this commit, symbolic variables inferred from the parameters were retained in the output function's `ret_struct_info`.  This is ill-formed, as the parameters from which these symbolic variables are inferred are no longer part of the function signature.

This commit updates `LazyGetInput` to use `EraseToWellDefined` to remove any symbolic variables from `ret_struct_info` that cannot be inferred from the remaining arguments.


Lunderberg:
@slyubomirsky This is the follow-up to https://github.com/apache/tvm/pull/16798, which I ended up making much sooner than I had initially planned.

slyubomirsky (approved):
Good catch; I didn't realize that we were losing shape vars from the signature

PR #16843, opened 2024-04-03
[IR] Provide well-formed intermediate in ApplyPassToFunction

Lunderberg:
Prior to this commit, `ApplyPassToFunction` removed functions from the `IRModule` to hide them from the inner `ir.transform.Pass`.  The dangling `GlobalVar` references to those functions meant that the intermediate `IRModule` was ill-formed This commit updates the `ApplyPassToFunction` utility to instead replace the functions with `ExternFunc` nodes.  This still prevents the inner `ir.transform.Pass` from having visibility into functions that should not be mutated, but provides a well-formed `IRModule`.

### DIFF
--- a/src/ir/apply_pass_to_function.cc
+++ b/src/ir/apply_pass_to_function.cc
@@ -1,0 +1,136 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/*!
+ * \file src/ir/apply_pass_to_function.cc
+ * \brief Utility transformation that applies an inner pass to a subset of an IRModule
+ */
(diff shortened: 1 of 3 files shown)


Lunderberg:
A follow-up commit to https://github.com/apache/tvm/pull/16801

slyubomirsky:
Are the substituted global vars restored afterwards? I couldn't find a place where they were put back. I would be worried about these extern funcs ending up in the final module.

Lunderberg:
The `GlobalVar` instances themselves don't need to be changed, only the functions to which they point.  This is why the pre-transform steps don't need to modify any function bodies, because the `GlobalVar` is still present in the `IRModule`.  After running the transform, the values from `new_subset` are only copied back to the original `mod` if they weren't one of the dummy `ExternFunc` instances.

Lunderberg:
The CI failure here will be resolved with https://github.com/apache/tvm/pull/16843.

slyubomirsky:
Wait, a global var can be mapped to an `ExternFunc`? That's wild.

slyubomirsky (approved):
I had no idea that it was possible to have an `ExternFunc` as a module entry. If this is behavior used elsewhere, I should probably mention this in the specification. Otherwise, this "should" be safe because most accesses of global vars have to check if the value is a Relax `Function` since it could be a `PrimFunc`... then again, there might be hazards if a pass assumes (reasonable) that those are the _only_ two possibilities. In any case, this is an improvement compared to passing a potentially malformed module entirely.

Lunderberg:
Yup, `relax::ExternFunc` is a subclass of `BaseFunc`, and can be stored directly in an `IRModule`.  At some point, I want to use this to specify when the lookup of a `PackedFunc` in the global registry occurs.  Currently, there's a discrepency between Relax and TIR, where Relax performs the lookup when the VirtualMachine is initialized, while TIR performs the lookup just before the call is performed.  I'd like to make these more similar, with an `ExternFunc` being resolved either at initialization (declared in IRModule), at the start of a function (ExternFunc bound to a variable), or at the callsite (ExternFunc used inline).

But the main impetus for that would have been the `LazyTransformParams` pass, and with the `LazyGetInput` accepting a callback parameter, there isn't as much need for it.

PR #16844, opened 2024-04-03
[TVMScript] Produce empty DictAttrs when R.func_attrs is absent

Lunderberg:
A follow-up to https://github.com/apache/tvm/pull/16745.  For Relax functions produced in TVMScript, when `R.func_attrs` was not present, the default was set to `None` instead of an empty dictionary.


slyubomirsky:
What is the specific reason to have this change? It might be a good invariant to have, just wondering if there was a specific motivation for it.

slyubomirsky (approved):
I think I support your reasoning, this is an easily averted source of errors.

PR #16847, opened 2024-04-04
[relay][feature] save relay IR as onnx for visualize

ShawnZhuang:
Prior to this commit, relay ir may not be easy to be viewed in netron. Awareness of distinct Attr types is required in some solutions.

this commit provide a method to serializing relay ir to protobuf for visualization.
1. it provide a general method to retrieve member variables of any Attrs, which reference the implementation of relay_text_printer
2. the protobuf use the defination of onnx-ml.pb, and could be opened by netron naturally.


tqchen:
thank you @ShawnZhuang , one thing that I think worth clarifying is that whether it serves as a onnx exporter(which i believe there were some work in https://github.com/apache/tvm/blob/main/python/tvm/contrib/target/onnx.py

 as a community we are migrating a lot of effort to relax, so if you are interested in bringing onnx translation to relax, that could also be interesting

ShawnZhuang:
> thank you @ShawnZhuang , one thing that I think worth clarifying is that whether it serves as a onnx exporter(which i believe there were some work in https://github.com/apache/tvm/blob/main/python/tvm/contrib/target/onnx.py) seems the purpose was different(mainly to export to netron compatible format, is it possible to support a variant of format that netron directly take?
> 
> as a community we are migrating a lot of effort to relax, so if you are interested in bringing onnx translation to relax, that could also be interesting

Yes, the mainly purpure is to export  to netron compatible format  for visualize.    

According  to  https://github.com/lutzroeder/netron ,  the onnx supports tf, caffe, keras,mxnet,onnx and darknet ,  but  a directly 
 netron format  has not been found.

Supporting  netron directly  may takes two steps, Firstly a tvm.proto should be defined,  secondly,  netron should support the new proto(which needs front-end efforts in netron)

ShawnZhuang:
ok,  i would  start bringing onnx translation to relax  after this PR  work is finished.

ShawnZhuang:
@tqchen  it seam that  google protobuf  is required,   and should be added to dockfile.  Do you  
any sugguestion on  editing the CmakeList and Dockerfile,  since lots of docker file are found in  tvm/docker

tqchen:
seems a better approach to write something like https://github.com/lutzroeder/netron/blob/main/source/caffe.js in netron. That file would parse the json file exported by the tvm relax, and construct the graph. Alternatively, we can add a visualization spec json for tvm, and add netron support.

This would remove the need of protobuf and allows more direct export

PR #16848, opened 2024-04-04
[DLight] Fix a corner case for reduction rule

Hzfengsy:
The current rule will fail when the output shape is only one element, because of missing `preserve_unit_loops`. This PR fixes it and adds a test case.


echuraev (approved):
LGTM!

PR #1557, opened 2018-08-05
[CONTRIB] PyTorch bridge

eqy:
This PR uses @jwfromm's implementation #658 as a bridge for PyTorch tensors, similar to the existing MxNet bridge.

PR #16854, opened 2024-04-08
[OpenCL] Add OpenCL device for automatic target detection.

mengshyu:
This PR adds OpenCL device for automatic target detection.


echuraev:
Please, add CI tests.

tests/python/target/test_target_target.py
```diff
@@ -547,5 +547,17 @@ def test_target_from_device_rocm(input_device):
     )
 
 
+@tvm.testing.requires_opencl
+@pytest.mark.parametrize("input_device", ["opencl", tvm.opencl()])
+def test_target_from_device_rocm(input_device):
```
echuraev, line 552:
```suggestion
def test_target_from_device_opencl(input_device):
```

python/tvm/target/detect_target.py
```diff
@@ -58,6 +58,17 @@ def _detect_rocm(dev: Device) -> Target:
     )
 
 
+def _detect_opencl(dev: Device) -> Target:
+    return Target(
+        {
+            "kind": "opencl",
+            "max_shared_memory_per_block": dev.max_shared_memory_per_block,
+            "max_threads_per_block": dev.max_threads_per_block,
+            "thread_warp_size": dev.warp_size,
```
echuraev, line 67: What about other attributes which are specified in `target_kind.cc`?

python/tvm/target/detect_target.py
```diff
@@ -58,6 +58,17 @@ def _detect_rocm(dev: Device) -> Target:
     )
 
 
+def _detect_opencl(dev: Device) -> Target:
+    return Target(
+        {
+            "kind": "opencl",
```
echuraev, line 64: In some cases, we can initialize OpenCL target with specifying specific device, e.g.:
```
target = Target("opencl -device=mali", host="llvm -mtriple=aarch64-linux-gnu")
or
target = Target("opencl -device=adreno", host="llvm -mtriple=arm64-linux-android")
```
Can we handle these cases somehow?

python/tvm/target/detect_target.py
mengshyu: The rest of attributes will use the default value.  meta, cuda, and rocm are similar ways to initialize.

python/tvm/target/detect_target.py
mengshyu: Yes, I have verified on Orange Pi.

Screenshot from 2024-04-09 12-56-06

tests/python/target/test_target_target.py
mengshyu: I'll fix it, thanks.

python/tvm/target/detect_target.py
echuraev: Sorry, probably my question wasn't clear. Yes, I know that we can get such information from target. But I'm curious if we can get this information from the device somehow? 

Probably, we can specify by an addition parameter to `Target.from_device(input_device)`. For example: `Target.from_device(input_device, '-device=adreno')` will specify target as `"kind": "opencl -device=adreno"`. Does it make sense?

mengshyu: It seems there is no argument for '-device=adreno' in API [Target.from_device.](https://github.com/apache/tvm/blob/main/python/tvm/target/target.py#L154)

This PR intends to fix weight conversion for MLC LLM on OrangePi ([#1557](https://github.com/mlc-ai/mlc-llm/issues/1557)), assuming there are no other issues. I would like to merge it first. If you have any suggestions, perhaps we can create another issue to discuss. Thanks.

echuraev (approved):
In general LGTM. Thanks!

PR #16859, opened 2024-04-09
[QoL][Relax] Use SeqExpr in IR types when SeqExpr is required

Lunderberg:
The Relax IR requires the `FunctionNode::body`, `IfNode::true_branch`, and `IfNode::false_branch` to be instances of `relax::SeqExpr`. If these Relax requirements are violated, correctly-implemented transformations may raise an exception
(e.g. from `Downcast` in `Downcast<SeqExpr>(func->body)->blocks`), or even segfault (e.g. when `.as` returns a nullptr in `func->body.as<SeqExprNode>()->blocks`).  Debugging these failure modes is also difficult, as even the TVMScript printer relies on the body of the function being a `SeqExprNode`.

This commit updates the C++ type of `FunctionNode::body`, `IfNode::true_branch`, and `IfNode::false_branch` to be `relax::SeqExpr` instead of `relax::Expr`.  This does not impact any well-formed Relax IR, and allows this type of ill-formed Relax IR type to be checked at compile-time.  A large number of checks applied during TVM runtime can now be removed, as they duplicate the new compile-time check.

To maintain backwards compatibility, this commit adds a new constructor to `relax::SeqExpr`, which accepts a single `Expr body` argument.  This constructor provides either an additional reference to the same underlying `relax::SeqExprNode`, if `body` already contains a `relax::SeqExprNode`, and otherwise wraps the body in a `relax::SeqExpr`.  For implementations that previously produced well-formed Relax IR, this change has no effect.  For implementations that previously produced ill-formed Relax IR, this change results in the equivalent well-formed Relax IR.

Alternate implementations considered:

* Perform the backwards-compatibility wrapping within the `relax::Function` and `relax::If` constructors.  While this would provide the intended conversion when these constructors are used, Relax transforms make frequent use of copy-on-write (e.g. `func.CopyOnWrite()->body = new_body`), which does not use the constructor.  Maintaining backwards compatibility for this usage requires the implicit conversion constructor that was chosen for this PR.

* Remove the Relax IR requirement for these expressions to be `SeqExpr`.  While this would make Relax more internally consistent, such a change would break backwards compatibility that relies on `SeqExpr` being present.  While the callsites within TVM could be updated to resolve this breakage, callsites outside of TVM (e.g. MLC-LLM) could not.  Exposing the special case within the C++ type, as done in this PR, maintains backwards compatibility.

### DIFF
--- a/include/tvm/relax/expr.h
+++ b/include/tvm/relax/expr.h
@@ -213,78 +213,6 @@ Call WithFields(Call call, Optional<Expr> opt_op = Optional<Expr>(),
                 Optional<Array<StructInfo>> opt_sinfo_args = Optional<Array<StructInfo>>(),
                 Optional<Span> opt_span = Optional<Span>());
 
-/*!
- * \brief Condition expression
- *
- * Unlike traditional statement `if`s, the if evalutes
- * to the result of the branch taken.
- *
- * x = if (true) { 1 } else { 0 }; // x is 1
- * y = if (false) { 1 } else { 0 }; // y is 0
- *
- * \note This is similar to C's ternary operator.
- */
-class IfNode : public ExprNode {
- public:
-  /*! \brief The condition. */
-  Expr cond;
-  /*! \brief The expression evaluated when condition is true. */
-  Expr true_branch;
-  /*! \brief The expression evaluated when condition is false */
-  Expr false_branch;
-
-  void VisitAttrs(tvm::AttrVisitor* v) {
-    v->Visit("cond", &cond);
-    v->Visit("true_branch", &true_branch);
-    v->Visit("false_branch", &false_branch);
-    v->Visit("_checked_type_", &checked_type_);
-    v->Visit("struct_info_", &struct_info_);
-    v->Visit("span", &span);
-  }
-
-  bool SEqualReduce(const IfNode* other, SEqualReducer equal) const {
-    equal->MarkGraphNode();
-    return equal(cond, other->cond) && equal(true_branch, other->true_branch) &&
-           equal(false_branch, other->false_branch) && equal(struct_info_, other->struct_info_);
-  }
-
-  void SHashReduce(SHashReducer hash_reduce) const {
-    hash_reduce->MarkGraphNode();
-    hash_reduce(cond);
-    hash_reduce(true_branch);
-    hash_reduce(false_branch);
-    hash_reduce(struct_info_);
-  }
-
-  static constexpr const char* _type_key = "relax.expr.If";
-  TVM_DECLARE_FINAL_OBJECT_INFO(IfNode, ExprNode);
-};
-
-class If : public Expr {
- public:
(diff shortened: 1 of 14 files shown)


Lunderberg:
This PR is currently marked as a draft, to allow time for community discussion.  A forum thread ([link](https://discuss.tvm.apache.org/t/relax-ir-fields-that-require-seqexpr/17055)) for discussion has also been started for this change.

include/tvm/relax/expr.h
```diff
@@ -915,18 +843,113 @@ class SeqExprNode : public ExprNode {
...
+ *
+ * Unlike traditional statement `if`s, the if evalutes
+ * to the result of the branch taken.
+ *
+ * x = if (true) { 1 } else { 0 }; // x is 1
+ * y = if (false) { 1 } else { 0 }; // y is 0
+ *
+ * \note This is similar to C's ternary operator.
+ */
+class IfNode : public ExprNode {
```
quic-sanirudh, line 875: Do we have to move the whole definition of `IfNode` to a new location, because this would make the git history a bit harder to read later on. Could we perhaps do a forward declaration of `SeqExpr` class and change the definition of `IfNode` in place?

quic-sanirudh (approved):
Thanks a lot for this. Looks good to me apart from a minor comment about possibly modifying `IfNode` in place to improve git history maintenance to just highlight the changed lines instead of the whole class.

include/tvm/relax/expr.h
Lunderberg: Unfortunately, I wasn't able to do so.  Declaring an member variable in a class requires the size of that member variable to be defined when the member variable is declared.  For cases where the size of a member variable is independent of the size of `T` (e.g. `T*`, `const T&`, or some templates such as `std::unique_ptr<T>`), a forward-declaration can be used.  Since the `relax::SeqExpr` is held by value, it cannot be defined with a forward-reference, and so the definitions of `relax::Function` and `relax::If` must occur after `relax::SeqExpr`.

(Even though all `ObjectRef` subclasses store their data in a reference-managed pointer, and have the exact same size, I'm not aware of any way to expose this to the compiler in a forward declaration.  A few years ago, I looked into what it would take to turn `ObjectRef` into a template over the internal object (e.g. `using SeqExpr = ObjectRefImpl<SeqExprNode>;`), and decided that it would be a bit more effort than I'd like.  It may be worth doing so, with care to ensure that a forward reference of `SeqExprNode` would be sufficient to define `ObjectRefImpl<SeqExprNode>`, but that would be a much larger change.)

quic-sanirudh: Ah yes, I missed that it was declared by value. Sorry about that and thanks for the clarification and yes, I agree that "templatizing" the `ObjectRef` would be a much bigger change, and probably not needed for this since the concern was more of a stylistic one to preserve git history.

PR #16863, opened 2024-04-09
[Picojson] Let the key of objects in json be ordered by default

Ubospica:
Previously picojson define `object` as an alias of `std::unordered_map`. That means when parsing json, the order of keys in objects are uncertain and dependent on implementation. This makes it inconvenient for certain applications, e.g. in LLM generation output, we wish the order of keys the same as the order in the json file.

This PR implements a ordered hashmap `ordered_hashmap` that 1) maintains the order in which the elements are inserted, and 2) have the same interface as `std::unordered_map`. Picojson will define object as an alias of `ordered_hashmap`, so the order of the input json is maintained when parsing.

Macro `PICOJSON_USE_ORDERED_OBJECT` controls whether object uses the ordered version or the unordered version. It is set by default.

### DIFF
--- a/3rdparty/picojson/picojson.h
+++ b/3rdparty/picojson/picojson.h
@@ -26,12 +26,21 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 #pragma once
+
 #ifndef PICOJSON_USE_INT64
 #define PICOJSON_USE_INT64
 #define __STDC_FORMAT_MACROS 1
 #endif
 
+// If PICOJSON_USE_ORDERED_OBJECT is set, picojson uses object_with_ordered_keys, which maintains
+// the insertion order of keys, i.e. the order of keys in the json string.
+// This macro is set by default.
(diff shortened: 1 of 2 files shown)


3rdparty/picojson/picojson.h
```diff
@@ -137,10 +146,177 @@ enum { INDENT_WIDTH = 2 };
...
+  const_iterator end() const { return {static_cast<int>(order.size()), this}; }
+  const_iterator cbegin() const { return {0, this}; }
+  const_iterator cend() const { return {static_cast<int>(order.size()), this}; }
+
+  using std::unordered_map<Key, T>::empty;
+  using std::unordered_map<Key, T>::size;
+  using std::unordered_map<Key, T>::at;
+
+  T& operator[](const Key& key) {
+    if (count(key) == 0) {
```
tqchen, line 230: use find, so you only do lookup once

PR #16865, opened 2024-04-09
Feat/fp8 broadcast

vinx13:
This added missing handling of broadcast and const for fp8 codegen. It relies on the cast operator from float because there's no direct ways to create fp8 literal.

cc @csullivan @tqchen

### DIFF
--- a/src/target/source/codegen_cuda.cc
+++ b/src/target/source/codegen_cuda.cc
@@ -1260,6 +1260,20 @@ void CodeGenCUDA::VisitExpr_(const BroadcastNode* op, std::ostream& os) {  // NO
     return;
   }
 
+  if (op->dtype.is_float8()) {
+    int lanes = op->dtype.lanes();
+    ICHECK(lanes == 1 || lanes == 2 || lanes == 4);
+    std::string v = PrintExpr(op->value);
+    PrintType(op->dtype, os);
+    os << "(make_float" << lanes << "(";
+    for (int i = 0; i < lanes; ++i) {
+      if (i != 0) os << ", ";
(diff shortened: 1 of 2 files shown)


src/target/source/codegen_cuda.cc
```diff
@@ -1260,6 +1260,20 @@ void CodeGenCUDA::VisitExpr_(const BroadcastNode* op, std::ostream& os) {  // NO
     return;
   }
 
+  if (op->dtype.is_float8()) {
+    int lanes = op->dtype.lanes();
+    ICHECK(lanes == 1 || lanes == 2 || lanes == 4);
+    std::string v = PrintExpr(op->value);
+    PrintType(op->dtype, os);
```
csullivan, line 1267: nit, only because it took me more than 30 seconds to digest this
```suggestion
    # Implicit conversion from float back to fp8 
    PrintType(op->dtype, os);
```

csullivan (approved):
LGTM, thank you @vinx13. One nit, up to you if you want the documentation or not.

src/target/source/codegen_cuda.cc
```diff
@@ -1264,6 +1264,7 @@ void CodeGenCUDA::VisitExpr_(const BroadcastNode* op, std::ostream& os) {  // NO
     int lanes = op->dtype.lanes();
     ICHECK(lanes == 1 || lanes == 2 || lanes == 4);
     std::string v = PrintExpr(op->value);
+    # Implicit conversion from float back to fp8
```
csullivan, line 1267:
```suggestion
    // Implicit conversion from float back to fp8 
```

PR #16873, opened 2024-04-11
[Thrust] Fix thrust workspace allocation

vinx13:
cc @tqchen @MasterJH5574


MasterJH5574:
Thanks! Could you also enlarge the workspace estimation? 2n is not enough for sort based on our observation. Likely at least 8n with additional 4MB is needed.

tqchen (approved):

PR #16876, opened 2024-04-12
[TVMScript] Optionally use `ruff format` instead of `black`

Lunderberg:
The `ruff format` tool is significantly faster than the `black` formatter.  For some particularly long TVMScript modules, using it can reduce the time required to show a formatted module from ~5 minutes to ~1 minute.  This commit updates the `.show()` function to apply the optionally formatting using `ruff format` if available, falling back to `black` otherwise.

### DIFF
--- a/python/tvm/script/highlight.py
+++ b/python/tvm/script/highlight.py
@@ -17,7 +17,10 @@
 """Highlight printed TVM script.
 """
 
+import functools
 import os
+import shutil
+import subprocess
 import sys
 import warnings
 from typing import Any, Optional, Union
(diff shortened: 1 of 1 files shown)


python/tvm/script/highlight.py
```diff
@@ -92,7 +95,73 @@ def cprint(
...
+        def formatter(code_str):
+            return black.format_str(code_str, mode=black.FileMode())
+
+        return formatter
+
+    def get_fallback_formatter():
+        def formatter(code_str):
+            with warnings.catch_warnings():
+                warnings.simplefilter("once", UserWarning)
+                ruff_install_cmd = sys.executable + " -m pip install ruff"
```
quic-sanirudh, line 132: Should we restrict to a specific version of ruff similar to a specific black version we've specified below?

quic-sanirudh (approved):
This a great addition, thanks a lot. Just had one minor doubt about whether constraining to specific version of ruff is needed, but otherwise this will be really useful.

python/tvm/script/highlight.py
Lunderberg: I'm uncertain, but would lean instead toward removing the specific version of `black`, if we want consistency.  Specifying a specific version of `black` is useful, since then it can be matched to the CI's version, in case a developer also uses `black` either for format-on-save or as a pre-commit hook.  However, the version number in the warning has become out of sync with the CI's version (warning message was added in November 2022, but the CI was bumped from `22.3.0` to `22.12.0` in March 2023).

Since there isn't really a good way to keep them in sync (nor any `ruff` version in the CI with which to be in sync), I'd lean toward removing the versions from the warning message altogether.

quic-sanirudh: Right yeah, that makes more sense. It is very hard to keep them in sync. Perhaps after we remove the version info from the warning message, we can just add a line saying something like **"For version compatibility with CI, please check docker/Dockerfile.ci_lint"** if the version info has to be specified.

Thanks for the response.

PR #16886, opened 2024-04-16
[Relax][Frontend] Fix sort, argsort and topk in nn module

Hzfengsy:
Fixes errors introduced in #16851 and add test cases.


Hzfengsy:
cc @vinx13

echuraev (approved):

PR #16888, opened 2024-04-16
[CUBLAS][FP8] Support e4m3 gemm in cuBLAS BYOC

ibsidorenko:
This is attempt to bring [PR#63](https://github.com/octoml/tvm/pull/63) from `octoml/tvm` to `apache/tvm`.
This commit enables FP8 e4m3 matmul offloading to cuBLAS.

сс @vinx13 @JosephTheOctonaut


vinx13 (approved):
we also need to check it's TN layout as it's the only one supported in cublas. I'll follow up on it

PR #16893, opened 2024-04-16
[SVE] Check for SVE target in VectorizeLoop

ekalda:
Check that we are compiling for an SVE enabled target when the extent of a loop marked for vectorizing is a vscale dependent expression.

The extent of a loop should be either a positive integer or an vscale dependent expression, in the latter case we'd expect the target to have `has_sve` feature.

### DIFF
--- a/src/tir/transforms/vectorize_loop.cc
+++ b/src/tir/transforms/vectorize_loop.cc
@@ -731,15 +731,12 @@ class LoopVectorizer : public StmtMutator {
   Stmt VisitStmt_(const ForNode* op) final {
     if (op->kind == ForKind::kVectorized) {
       auto* extent_as_int = op->extent.as<IntImmNode>();
+
       if (!extent_as_int || extent_as_int->value < 1) {
-        Target current_target = Target::Current();
-        bool has_sve{false};
-        if (current_target.defined()) {
-          has_sve = current_target->GetFeature<Bool>("has_sve").value_or(Bool(false));
-        }
         bool is_scalable_expr = CheckContains::ExprContains(op->extent, arith::IsVScaleCall);
-        ICHECK(is_scalable_expr && has_sve) << "Failed to vectorize loop with extent " << op->extent
-                                            << " for target " << current_target;
+        ICHECK(is_scalable_expr && arith::TargetHasSVE())
+            << "Failed to vectorize loop with extent " << op->extent << " for target "
+            << Target::Current();
       }
       ICHECK(is_zero(op->min));
       return Vectorizer(op->loop_var, op->extent)(op->body);
--- a/src/tir/transforms/vectorize_loop.cc
+++ b/src/tir/transforms/vectorize_loop.cc
@@ -34,6 +34,9 @@
 #include <unordered_map>
 #include <vector>
 
+#include "../../src/arith/scalable_expression.h"
+#include "../../tir/analysis/check_contains.h"
+
 namespace tvm {
 namespace tir {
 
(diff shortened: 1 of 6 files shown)


src/tir/transforms/vectorize_loop.cc
```diff
@@ -725,17 +728,33 @@ class Vectorizer : public StmtMutator, public ExprFunctor<PrimExpr(const PrimExp
...
+      }
       ICHECK(is_zero(op->min));
       return Vectorizer(op->loop_var, op->extent)(op->body);
     } else {
       return StmtMutator::VisitStmt_(op);
     }
   }
-};
 
-Stmt VectorizeLoop(Stmt stmt) { return LoopVectorizer()(std::move(stmt)); }
```
ekalda, line 738: Is this function necessary?

ekalda:
@Lunderberg @lhutton1 

This is a follow up from https://github.com/apache/tvm/pull/16782, it makes sure that scalable vectors are created only for SVE target. In this implementation we check for target in function attributes, so we'd rely on `BindTarget` having run before. Alternatively we could access the "global" target via `Target::Current()`, which would be simpler, but would make it harder to mix SVE and non-SVE targets. I'm happy to go with either option :)

ekalda:
I moved the implementation to use the current target since changing the graph executor's pipeline is breaking lots of CUDA tests.

tests/python/tir-transform/test_tir_transform_vectorize.py
```diff
@@ -99,26 +105,29 @@ def main(A: T.Buffer((25,), "float32")):
...
+        with tvm.target.Target(sve_target):
+            tvm.tir.transform.VectorizeLoop()(Module)
 
 
 def test_vectorize_vector_scalable_error4():
     @I.ir_module
     class Module:
         @T.prim_func(private=True)
         def main(A: T.Buffer((25,), "float32")):
+            T.func_attr({"target": sve_target})
```
lhutton1, line 117: Is this still needed?

src/tir/transforms/vectorize_loop.cc
```diff
@@ -727,6 +730,17 @@ class LoopVectorizer : public StmtMutator {
  public:
   Stmt VisitStmt_(const ForNode* op) final {
     if (op->kind == ForKind::kVectorized) {
+      auto* extent_as_int = op->extent.as<IntImmNode>();
+      if (!extent_as_int || extent_as_int->value < 1) {
+        Target current_target = Target::Current();
+        bool has_sve{false};
+        if (current_target.defined()) {
+          has_sve = current_target->GetFeature<Bool>("has_sve").value_or(Bool(false));
```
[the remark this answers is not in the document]
ekalda: Good idea, I'll create a utility for it, it probably won't be the last time we would need to do that check :) There is another new utility function in @Anndrey24's [patch that is in progress ](https://github.com/apache/tvm/pull/16899/files#diff-a31e6da5360b6615078023bb62667d79ff77fad68c29dda820a9267a918d9c01L44) for checking whether an expression contains `vscale`, so once this goes in, we should unify the uses for that case as well.

tests/python/tir-transform/test_tir_transform_vectorize.py
ekalda: Oops no, will remove.

lhutton1 (approved):
Thanks @ekalda, LGTM!

lhutton1:
Thanks @ekalda!

PR #16894, opened 2024-04-16
[Dlight] Enhance vectorization for gpu matmul

vinx13:
This fixed dlight matmul rule to make sure vectorization is always applied to the innermost loop

cc @Hzfengsy @tqchen


Hzfengsy (approved):

tqchen:
@vinx13 please fix the testcase

PR #16895, opened 2024-04-16
[BYOC] Add layout check and update shape check for cublas FP8 BYOC

vinx13:
This adds additional check for cublas BYOC as it only supports TN layout. It also change the shape requirement to multiple of 16 bytes.

cc @tqchen @ibsidorenko

### DIFF
--- a/python/tvm/relax/backend/contrib/cublas.py
+++ b/python/tvm/relax/backend/contrib/cublas.py
@@ -20,6 +20,7 @@ import operator
 from functools import reduce
 
 import tvm
+from tvm import DataType
 from tvm.relax import transform
 from tvm.relax.transform import PatternCheckContext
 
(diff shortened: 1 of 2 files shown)


python/tvm/relax/backend/contrib/cublas.py
```diff
@@ -68,11 +69,30 @@ def _check_matmul(context: PatternCheckContext) -> bool:
...
-        # docs, but it was observed during testing.
-        if not isinstance(rhs_shape[-1], (tvm.tir.expr.IntImm, int)) or rhs_shape[-1] % 16 != 0:
+        matmul_rhs_var = matmul_call.args[1]
+        rhs_transposed = False
+        if matmul_rhs_var in context.matched_bindings:
+            matmul_rhs_call = context.matched_bindings[matmul_rhs_var]
+            assert (
+                isinstance(matmul_rhs_call, tvm.relax.Call)
+                and matmul_rhs_call.op.name == "relax.permute_dims"
+            )
```
ibsidorenko, line 79: I am Ok, thank you! Just a nit question: do we need here assert for the case when rhs_call is something but not `permute_dims`? Just to leave rhs_transposed == False and return False in the next IF (without crash):
```
        if not rhs_transposed:
            # cuBLAS FP8 operations require rhs being transposed
            return False
```

ibsidorenko (approved):

python/tvm/relax/backend/contrib/cublas.py
vinx13: `if matmul_rhs_var in context.matched_bindings:` this condition implies rhs is transposed (it's the only pattern that rhs is another binding being matched), so I added an assertion here, it won't crash if we have non-transposed rhs

PR #16897, opened 2024-04-17
[Target] Don't register AArch64 target tags without LLVM compiler support

lhutton1:
This commit aims to fix the issue described here: https://github.com/apache/tvm/pull/16425#issuecomment-2059781680 by conditionally registering the target tags based on the availability of the LLVM AArch64 backend. It's possible to extract the targets LLVM has been compiled for using `llvm-config --targets-built`.

cc @tqchen


mvermeulen:
I retested with the ROCm build and no longer see warning messages about a missing ARM configuration when running ROCm.

leandron (approved):
LGTM, thanks @lhutton1

PR #16903, opened 2024-04-18
[Bugfix] CudaDeviceAPI::GetAttr may check kExist when GPUs absent

Lunderberg:
This commit resolves a bug that was introduced in
https://github.com/apache/tvm/pull/16377.  If no CUDA-capable GPUs are present, the call to `cudaGetDeviceCount` will return an error, which will be raised as an exception by the `CUDA_CALL` macro.  However, checking the `kExist` flag is valid even if no GPUs are present.

This commit removes the use of `CUDA_CALL`, and instead returns false in this case.

### DIFF
--- a/src/runtime/cuda/cuda_device_api.cc
+++ b/src/runtime/cuda/cuda_device_api.cc
@@ -41,11 +41,12 @@ class CUDADeviceAPI final : public DeviceAPI {
   void GetAttr(Device dev, DeviceAttrKind kind, TVMRetValue* rv) final {
     int value = 0;
     switch (kind) {
-      case kExist:
+      case kExist: {
         int count;
-        CUDA_CALL(cudaGetDeviceCount(&count));
-        value = static_cast<int>(dev.device_id < count);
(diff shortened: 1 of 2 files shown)


tqchen (approved):

vinx13:
I found the issue is not about `cudaGetDeviceCount`, `cudaGetDeviceCount` doesn't return error, however, the issue is it doesn't check `CUDA_VISIBLE_DEVICES` setting but like other APIs do. When you set `CUDA_VISIBLE_DEVICES` to empty, checking `kExist` will return true, and other attrs will checked and cause no device error. I think we should recover the original check before #16377 that uses `cudaDeviceGetAttribute` and clear the error code if not exists

tqchen:
@vinx13 do you mind send a followup pr on this?

vinx13:
it's not because `cudaGetDeviceCount`, when I ran the test, I found that there are multiple calls to `GetAttr`, the first call is to check `kExist` and it always return true, the subsequent call fails because of setting `CUDA_VISIBLE_DEVICES=''`

PR #16910, opened 2024-04-21
[Web] Support string[] in setPackedFunc() and exceptionally long arrays

CharlieFRuan:
There are two changes in this PR.

#### Change 1: Support `string[]` in `setPackedFunc()`
Prior to this PR, we cannot pass in `string[]` from typescript to a TVM PackedFunc and need to convert it to `TVMArray<TVMString>` (for instance in `getParamsFromCacheByName()`). This may not be the most convenient if the PackedFunc's caller is not internal to tvmjs. Thus, this PR moves the conversion to `setPackedFunc()` instead.

#### Change 2: Support exceptionally long TVM arrays
The second change is dealing with exceptionally long TVM arrays. In cases like passing in a token table, we need to pass in a long `string[]` (in Llama-3's case, of size 128000), leading to JS error `RangeError: Maximum call stack size exceeded` since we treat each string element as an argument, shown in `this.ctx.arrayMake(...inputs)`. 

This PR sets an empirical call stack limit of 30000 and chunks the array elements in `makeTVMArray()`, converting each chunk to its own TVMArray. Then we concatenate them with the newly implemented `runtime.ArrayConcat` that concatenates N TVMArrays.

Tested end-to-end in WebLLM.

### DIFF
--- a/src/runtime/container.cc
+++ b/src/runtime/container.cc
@@ -49,23 +49,6 @@ TVM_REGISTER_GLOBAL("runtime.Array").set_body([](TVMArgs args, TVMRetValue* ret)
   *ret = Array<ObjectRef>(data);
 });
 
-// Concatenate n TVMArrays
-TVM_REGISTER_GLOBAL("runtime.ArrayConcat").set_body([](TVMArgs args, TVMRetValue* ret) {
-  std::vector<ObjectRef> data;
-  for (int i = 0; i < args.size(); ++i) {
-    // Get i-th TVMArray
-    ICHECK_EQ(args[i].type_code(), kTVMObjectHandle);
-    Object* ptr = static_cast<Object*>(args[i].value().v_handle);
-    ICHECK(ptr->IsInstance<ArrayNode>());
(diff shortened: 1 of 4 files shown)


src/runtime/container.cc
```diff
@@ -49,6 +49,23 @@ TVM_REGISTER_GLOBAL("runtime.Array").set_body([](TVMArgs args, TVMRetValue* ret)
   *ret = Array<ObjectRef>(data);
 });
 
+// Concatenate n TVMArrays
+TVM_REGISTER_GLOBAL("runtime.ArrayConcat").set_body([](TVMArgs args, TVMRetValue* ret) {
```
tqchen, line 53: given this is not as a commonly used function, let us move it to https://github.com/apache/tvm/blob/main/web/emcc/wasm_runtime.cc

CharlieFRuan: Makes sense; just moved and renamed to `tvmjs.runtime.ArrayConcat`.

MasterJH5574 (approved):
LGTM!

PR #16921, opened 2024-04-24
[SME] Introduce scalable fp32 dense schedule

lhutton1:
This commit adds a new scalable fp32 dense schedule that calls SME intrinsics according to the SME RFC: https://github.com/apache/tvm-rfcs/pull/107.

Currently the schedule does not make use of predication, meaning the output from the matmul compute must be copied in a subsequent compute stage. This will be removed once support for predication is added.

cc @ekalda @Anndrey24 @tqchen @cbalint13 @Lunderberg

### DIFF
--- a/src/arith/const_int_bound.cc
+++ b/src/arith/const_int_bound.cc
@@ -371,7 +371,7 @@ class ConstIntBoundAnalyzer::Impl
     } else if (op->op.same_as(tir::builtin::bitwise_and())) {
       return VisitBitwiseAnd(op);
     } else if (op->op.same_as(tir::builtin::vscale()) && TargetHasSVE()) {
-      return MakeBound(1, 16);
+      return MakeBound(1, kAArch64VScaleValues.size());
     } else {
       return Everything(op->dtype);
     }
--- a/python/tvm/micro/testing/aot_test_utils.py
+++ b/python/tvm/micro/testing/aot_test_utils.py
@@ -65,6 +65,16 @@ AOT_USMP_CORSTONE300_RUNNER = AOTTestRunner(
     },
 )
 
+AOT_APROFILE_AEM_RUNNER = AOTTestRunner(
+    makefile="aprofile_aem",
+    includes=[],
+    pass_config={
+        "tir.usmp.enable": False,
+        # AOT test infra generates 'fake' tensor inputs which fails asserts
+        "tir.disable_assert": True,
+    },
+)
+
 
 def parametrize_aot_options(test):
     """Parametrize over valid option combinations"""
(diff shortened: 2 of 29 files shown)


lhutton1:
@tvm-bot rerun

lhutton1:
Retriggering CI since it hasn't been run for a while

src/arith/const_int_bound.cc
```diff
@@ -371,7 +371,7 @@ class ConstIntBoundAnalyzer::Impl
     } else if (op->op.same_as(tir::builtin::bitwise_and())) {
       return VisitBitwiseAnd(op);
     } else if (op->op.same_as(tir::builtin::vscale()) && TargetHasSVE()) {
-      return MakeBound(1, 16);
+      return MakeBound(1, kAArch64VScaleValues.size());
```
Anndrey24, line 374: This is a lot of great work which will also serve as a nice foundation and example for the rest of the SME schedules. Thanks, @lhutton1! 

I just have a quick comment about the bounds of the "vscale" builtin. I understand that here we are making the assumption that the length of the vector containing all possible vscale values will be equal to the vector's maximum value (i.e. the vector will include all elements 0 < i <= max_vscale). Could an issue arise in the future from someone modifying the list of values without being aware of this assumption?

For example, I see other sources that mention vscale taking only powers of 2 as possible values:
- [LLVM](https://llvm.org/docs/LangRef.html#llvm-vscale-intrinsic:~:text=This%20function%20attribute%20indicates%20vscale%20is%20a%20power%2Dof%2Dtwo%20within%20a%20specified%20range): 

> This function attribute indicates vscale is a power-of-two within a specified range. 
- [Reference manual](https://developer.arm.com/documentation/ddi0487/latest/):

> For each IMPLEMENTATION DEFINED vector length, all the following apply:
> • The smallest architecturally defined maximum length is 128 bits.
> • The largest architecturally defined maximum length is 2048 bits.
> • The length is a power of two.

This might mean checking just the 1, 2, 4, 8, 16 values could be enough (slightly speeding up the analyzer), in which case the length of the `kAArch64VScaleValues` vector would no longer match its largest element. However, I'm happy to discuss and maybe address this in a separate PR later since it doesn't have anything to do with the rest of the schedule :)

lhutton1: Thanks for taking a look @Anndrey24. I chose to use `.size()` here over something like `std::max_element()` due to speed (although admittedly it probably doesn't have much of an impact) in the analyzer, but I agree with your comments that it won't be the most maintainable going forwards. 

Regarding the range of values, I think you're right. I'm trying to recall the reason for going with 1, .., 16 but that escapes me for now. I think it would be best to discuss this further in another PR.

leandron (approved):
LGTM, thanks @lhutton1

lhutton1:
thanks @Anndrey24 @leandron!

PR #16922, opened 2024-04-25
[BugFix][Relax] change FuseOpsByPattern strategy to pattern-match maximal subgraph

wanghuibin0:
At first, FuseOpsByPattern for Relax used a strategy to match maximal subgraph. 
But this was changed in a previous PR https://github.com/apache/tvm/pull/15308. 
The intention was to fix a bug: in FuseOpsByPattern, when two matched subgraph are intersected, they are incorrectly merged together. 
The bug was fixed by returning only the first subgraph if multiple subgraph are intersected. 
The solution successfully solved the problem described there, but unfortunately, resulted in a minimal subgraph match strategy, probably unexpected. 

Let me clarify with a compute graph as an example:
```
lv0 = R.matmul(x, y)
lv1 = R.add(lv0, bias)
lv2 = R.clip(lv1, -128, 127)
```
Consider this pattern: `matmul->add | matmul->add->clip`. It should match both `lv0+lv1` and `lv0+lv1+lv2`. The solution in the aforementioned PR matches only `lv0+lv1`. But a stategy to match all `lv0+lv1+lv2` is expected. 

This bug has been fixed by checking, when two matched subgraph are intersected, whether they have an inclusion relationship(i.e. whether subgraph A is contained in subgraph B), and if they have, the larger one is retained.

cc @wrongtest-intellif

### DIFF
--- a/src/relax/transform/fuse_ops.cc
+++ b/src/relax/transform/fuse_ops.cc
@@ -1073,7 +1073,11 @@ class PatternBasedPartitioner : ExprVisitor {
     current_block_use_def_ = {};
   }
 
-  void VisitVarDef(const Var& var) final { group_map_[var.get()] = arena_->make<Group>(); }
+  void VisitVarDef(const Var& var) final {
+    Group* g = arena_->make<Group>();
+    group_map_[var.get()] = g;
+    vars_in_group_[g].push_back(var);
+  }
 
   void VisitBinding_(const VarBindingNode* binding) final {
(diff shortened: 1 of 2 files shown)


wrongtest-intellif:
hi, could you kindly take a look at this change?  I understand it is one more patch to https://github.com/apache/tvm/pull/15308.  thanks!  @masahi

masahi (approved):
Makes sense, thanks!

wrongtest-intellif (approved):

PR #16926, opened 2024-04-25
[TIR] Use enable_check in cache_read and add some boundary check

qsqqsqqsq-intellif:
1. Do not check region cover for cache_read and cache_inline when schedule enable_check set to false. We don't need region cover check in some case.
2. Add some boundary check for ProducerCoversConsumer when intset is nothing. It will raise error when compare to a empty intset.


wrongtest-intellif:
The testcases are required.

PR #16929, opened 2024-04-25
[VM][OPENCL] Take advantage of OpenCL host ptr for improved copy

srkreddy1238:
We can use OpenCL mapped pointer for these copies for improved performance.


MasterJH5574 (approved):
LGTM. Thank you @srkreddy1238!

PR #16930, opened 2024-04-25
[UnitTest] Use pytest's scope='session' for tvm.testing.parameter

Lunderberg:
Prior to this commit, the `tvm.testing.parameter` utility defined a fixture with the default `scope="function"`.  However, this prevents use of these parameters as arguments for other fixtures that are themselves cached using pytest.  Since these are parameters, not large values that would be expensive to compute, there is no downside to caching them at the pytest level.

This commit updates the scope of fixtures generated using `tvm.testing.parameter` to use `scope="session"` instead of the default `scope="function"`.


Lunderberg:
@tvm-bot rerun ci

masahi (approved):

PR #16933, opened 2024-04-26
[CLML] Fix in clml pattern check condition

krishnaraj36:
Added more check condition to make clml path more robust.
1. Depth_to_space - CLML path only supported for mode="DCR" and NCHW layout
2. Default checks -  CLML supports less than 4D tensor dimension and with batch size =1.


krishnaraj36:
@srkreddy1238 : Can you please take a look in this PR

srkreddy1238 (approved):
LGTM

PR #16937, opened 2024-04-26
[Thrust] Increase static workspace size

MasterJH5574:
This PR increases the thrust workspace size, since in practice we found that the current workspace size can still be insufficient. Thrust sort may require larger workspace when the number of elements being sorted is large (e.g., in Llama3 that is 128k).


yongwww (approved):

tqchen:
@tvm-bot rerun

PR #16949, opened 2024-04-28
[TVMScript][Relax] Use tir.SizeVar for shape variables

Lunderberg:
As reported in #16877, shape inference performed during a Relax transformation may produce different results than shape inference performed during TVMScript parsing.  While Relax transformations call `Analyzer::MarkGlobalNonNegValue` for each shape expression, this is not provided during TVMScript parsing.  As a result, output shapes that are conditional on the sign of a variable may produce different results when simplifying.

This commit provides a partial resolution for this issue.  Where prior to this commit, the TVMScript parser generated a `tir.Var` for each symbolic variable, the TVMScript parser now provides a `tir.SizeVar` for symbolic variables that are defined in contexts that require a non-negative size.

### DIFF
--- a/python/tvm/script/parser/relax/entry.py
+++ b/python/tvm/script/parser/relax/entry.py
@@ -150,6 +150,9 @@ class StructInfoProxy(ObjectGeneric):
     def get_symbolic_vars(self) -> Set[str]:
         return {}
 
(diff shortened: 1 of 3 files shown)


tqchen:
This is a nice PR that would be good to land it. cc @yongwww

PR #16959, opened 2024-04-30
[Relax][Transform] Handle identical PrimFunc with distinct VDevice

Lunderberg:
Prior to this commit, if an `IRModule` contained two expressions, where the types of the arguments differed only by the `VDevice`, these would be legalized to produce a single PrimFunc.  This PrimFunc would have the a `tvm::attr::kTarget` annotation specific to one of those expressions, and would be incorrect for use in the other location.

This commit updates the `LegalizeOps` transform to handle this case, producing multiple TIR PrimFuncs if required by the `VDevice` annotations.

### DIFF
--- a/src/relax/transform/legalize_ops.cc
+++ b/src/relax/transform/legalize_ops.cc
@@ -28,6 +28,7 @@
 #include <tvm/relax/op_attr_types.h>
 #include <tvm/relax/struct_info.h>
 #include <tvm/relax/transform.h>
+#include <tvm/tir/transform.h>
 
 namespace tvm {
 namespace relax {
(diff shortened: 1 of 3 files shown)


Lunderberg:
CI is stalled on `tests/python/relax/test_transform_few_shot_tuning.py::test_funcs[True-Softmax]`, with no updates shown in the last 2 days.  This test was not modified by this PR, and passes when run locally, so it looks like an issue with the CI environment.

Restarting CI to see if it resolves the issue.

Lunderberg:
@tvm-bot rerun

masahi (approved):

PR #16962, opened 2024-05-01
[SVE] Add codegen support for `vscale_range()` function attribute

Anndrey24:
This commit adds support for the `vscale_range()` LLVM function attribute to be generated for SVE and SME targets. 

Some LLVM optimisation passes make use of the `vscale_range()` function attribute when scalable vectors are present (e.g. BasicAA llvm/llvm-project/pull/80445), so we include it alongside the "target_cpu" and "target-features" attributes.  

cc @lhutton1

### DIFF
--- a/src/target/llvm/codegen_llvm.cc
+++ b/src/target/llvm/codegen_llvm.cc
@@ -104,7 +104,6 @@
 #include <vector>
 
 #include "../../arith/pattern_match.h"
-#include "../../arith/scalable_expression.h"
 #include "../build_common.h"
 #include "../func_registry_generator.h"
 #include "codegen_params.h"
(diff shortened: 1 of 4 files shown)


lhutton1 (commented):
Thanks @Anndrey24, looks great!

src/target/llvm/codegen_llvm.cc
```diff
@@ -1127,6 +1128,13 @@ void CodeGenLLVM::SetTargetAttributes(llvm::Function* func) {
   if (!features.empty()) {
     func->addFnAttr("target-features", features);
   }
+#if TVM_LLVM_VERSION >= 130
+  // Add vscale_range() function attribute when appropriate.
+  if (llvm_target_->TargetHasCPUFeature("sve") || llvm_target_->TargetHasCPUFeature("sme")) {
```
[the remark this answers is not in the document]
Anndrey24: Yes, I was actually a bit unsure how to proceed with this, but it makes more sense to only check for "sve".

Interestingly, just "+sme" by itself does not seem to include "sve" as part of the features returned by `GetAllLLVMCpuFeatures()`: `bf16, trbe, use-scalar-inc-vl, use-postra-scheduler, fuse-aes, fp-armv8, neon, enable-select-opt, ete, sme, fuse-adrp-add`, but "+v9a" does, so we would just need to always add them both to the target string (e.g. "mattr=+v9a,+sme").

lhutton1:
@tvm-bot rerun

src/target/llvm/codegen_llvm.cc
lhutton1, line 1133: Apologies, I had another thought about this. Since it's AArch64 specific, perhaps we should move it to `codegen_aarch64.cc`?

Anndrey24: Done!

lhutton1:
Thanks @Anndrey24!

PR #16965, opened 2024-05-03
[SVE] Add get_active_lane_mask builtin

lhutton1:
Adds a `get_active_lane_mask` builtin and lowering to `llvm.get.active.lane.mask` intrinsic. This will be used in subsequent patches for expressing predicated buffer loads/stores in TIR. Further information can be found in the [RFC](https://github.com/apache/tvm-rfcs/blob/main/rfcs/0104-scalable-vectors-in-tir.md#predication).

Co-authored-by: Elen Kalda <elen.kalda@arm.com>
Co-authored-by: Neil Hickey <neil.hickey@arm.com>

PR #16966, opened 2024-05-03
[SVE] Add support for representing and creating buffer-level predicates

lhutton1:
### Representation
This commit extends `BufferLoad` and `BufferStore` to accept a predicate mask argument indicating which lanes in a vectorized buffer load/store should be read/written.

As a simple example, we can load all lanes:
```
tir.BufferLoad(buf, [tir.Ramp(0, 1, 8)], predicate=tir.Broadcast(1, 8))
```

Or disable loading all lanes:
```
tir.BufferLoad(buf, [tir.Ramp(0, 1, 8)], predicate=tir.Broadcast(0, 8))
```

In TVMScript, buffer loads and stores are currently displayed using a "short-hand" notation e.g. `A[0:4]`, but there was no clear path for extending this notation to support predicates. Therefore, the vload/vstore notation is used e.g. `A.vload([T.Ramp(0, 1, 4)], predicate=...)`. The TVMScript printer falls back to the vload/vstore notation whenever predicates are specified.

### Creation
Buffer-level predication becomes more motivating when combined with the `tir.get_active_lane_mask` intrinsic. It can be used to mask off lanes when the vectorized axis is not divisible by the vector length. A detailed example and rationale can be found in the [RFC](https://github.com/apache/tvm-rfcs/blob/main/rfcs/0104-scalable-vectors-in-tir.md#predication).

Predicated buffer load/stores are created in the `VectorizeLoop` pass via `TryPredicateBufferAccesses`. This pass aims to convert block-level predicates e.g.
```
for i_0 in T.serial(4):
    for i_1 in T.vectorized(4):
        if i_0 * 4 + i_1 < 14:
            B[i_0 * 4 + i_1] = A[i_0 * 4 + i_1] + 1.0
```
to buffer-level predicates, e.g.
```
for i_0 in T.serial(4):
    predicate = T.get_active_lane_mask("int1x4", i_0 * 4, 14)
    A_load = T.meta_var(A.vload([T.Ramp(i_0 * 4, 1, 4)], predicate=predicate))
    B.vstore([T.Ramp(i_0 * 4, 1, 4)], A_load, predicate=predicate)
```
It takes a conservative approach for now, focussing only on expressions produced by the split scheduling primitive, but more complex expressions could be supported in the future.

`TryPredicateBufferAccesses` can be explicitly enabled/disabled with the `tir.enable_buffer_level_predication` pass context option. By default it will be disabled, unless the target supports SVE, in which case it will be enabled by default.

~Note: this commit depends on https://github.com/apache/tvm/pull/16965, so also contains the contents of https://github.com/apache/tvm/pull/16965.~

Co-authored-by: Elen Kalda <elen.kalda@arm.com>
Co-authored-by: Neil Hickey <neil.hickey@arm.com>

### DIFF
--- a/include/tvm/script/ir_builder/tir/ir.h
+++ b/include/tvm/script/ir_builder/tir/ir.h
@@ -411,10 +411,11 @@ Var EnvThread(String thread_tag, DataType dtype = DataType::Int(32));
  * \param buffer The buffer.
  * \param value The value to be stored.
  * \param indices The indices location to be stored.
- * \param predicate A vector mask of int1 values indicating which lanes of a vector are to be
- * stored.
+ * \param predicate A vector mask of boolean values indicating which lanes of a vector are to be
+ * stored. The number lanes of the mask must be equal to the number of lanes in value.
  */
-void BufferStore(Buffer buffer, PrimExpr value, Array<PrimExpr> indices, PrimExpr predicate);
+void BufferStore(Buffer buffer, PrimExpr value, Array<PrimExpr> indices,
+                 Optional<PrimExpr> predicate);
 
 /*!
  * \brief The prefetch hint for a buffer
--- a/include/tvm/script/ir_builder/tir/ir.h
+++ b/include/tvm/script/ir_builder/tir/ir.h
@@ -411,8 +411,10 @@ Var EnvThread(String thread_tag, DataType dtype = DataType::Int(32));
  * \param buffer The buffer.
  * \param value The value to be stored.
  * \param indices The indices location to be stored.
+ * \param predicate A vector mask of int1 values indicating which lanes of a vector are to be
+ * stored.
  */
-void BufferStore(Buffer buffer, PrimExpr value, Array<PrimExpr> indices);
+void BufferStore(Buffer buffer, PrimExpr value, Array<PrimExpr> indices, PrimExpr predicate);
 
 /*!
  * \brief The prefetch hint for a buffer
--- a/include/tvm/tir/buffer.h
+++ b/include/tvm/tir/buffer.h
@@ -209,14 +209,20 @@ class Buffer : public ObjectRef {
    * \brief Create an Expr that does a vector load at begin index.
    * \param begin The beginning index
    * \param dtype The data type to be loaded.
+   * \param predicate A vector mask of boolean values indicating which lanes of a vector are to be
+   * stored. The number lanes of the mask must be equal to the number of lanes in value.
    */
-  TVM_DLL PrimExpr vload(Array<PrimExpr> begin, DataType dtype) const;
+  TVM_DLL PrimExpr vload(Array<PrimExpr> begin, DataType dtype,
+                         Optional<PrimExpr> predicate = NullOpt) const;
   /*!
    * \brief Create a Stmt that does a vector store at begin index.
    * \param begin The beginning index
    * \param value The value to be stored.
+   * \param predicate A vector mask of boolean values indicating which lanes of a vector are to be
+   * stored. The number lanes of the mask must be equal to the number of lanes in value.
    */
-  TVM_DLL Stmt vstore(Array<PrimExpr> begin, PrimExpr value) const;
+  TVM_DLL Stmt vstore(Array<PrimExpr> begin, PrimExpr value,
+                      Optional<PrimExpr> predicate = NullOpt) const;
 
   /*!
    * \brief Get a flattened version of the buffer
--- a/include/tvm/tir/buffer.h
+++ b/include/tvm/tir/buffer.h
@@ -210,7 +210,7 @@ class Buffer : public ObjectRef {
    * \param begin The beginning index
    * \param dtype The data type to be loaded.
    * \param predicate A vector mask of boolean values indicating which lanes of a vector are to be
-   * stored. The number lanes of the mask must be equal to the number of lanes in value.
+   * loaded. The number lanes of the mask must be equal to the number of lanes in being loaded.
    */
   TVM_DLL PrimExpr vload(Array<PrimExpr> begin, DataType dtype,
                          Optional<PrimExpr> predicate = NullOpt) const;
--- a/include/tvm/tir/expr.h
+++ b/include/tvm/tir/expr.h
@@ -630,11 +630,14 @@ class BufferLoadNode : public PrimExprNode {
   Buffer buffer;
   /*! \brief The indices location to be loaded. */
   Array<PrimExpr> indices;
+  /*! \brief The predicate mask for loading values. */
+  PrimExpr predicate;
 
   void VisitAttrs(AttrVisitor* v) {
     v->Visit("dtype", &(this->dtype));
     v->Visit("buffer", &buffer);
     v->Visit("indices", &indices);
+    v->Visit("predicate", &predicate);
     v->Visit("span", &span);
   }
 
@@ -647,6 +650,7 @@ class BufferLoadNode : public PrimExprNode {
     hash_reduce(dtype);
     hash_reduce(buffer);
     hash_reduce(indices);
+    hash_reduce(predicate);
   }
 
   static constexpr const char* _type_key = "tir.BufferLoad";
@@ -675,7 +679,8 @@ class BufferLoadNode : public PrimExprNode {
  */
 class BufferLoad : public PrimExpr {
  public:
-  TVM_DLL explicit BufferLoad(Buffer buffer, Array<PrimExpr> indices, Span span = Span());
+  TVM_DLL explicit BufferLoad(Buffer buffer, Array<PrimExpr> indices,
+                              PrimExpr predicate = PrimExpr(), Span span = Span());
   TVM_DEFINE_OBJECT_REF_METHODS(BufferLoad, PrimExpr, BufferLoadNode);
   TVM_DEFINE_OBJECT_REF_COW_METHOD(BufferLoadNode);
 };
--- a/include/tvm/tir/expr.h
+++ b/include/tvm/tir/expr.h
@@ -631,7 +631,7 @@ class BufferLoadNode : public PrimExprNode {
   /*! \brief The indices location to be loaded. */
   Array<PrimExpr> indices;
   /*! \brief The predicate mask for loading values. */
-  PrimExpr predicate;
+  Optional<PrimExpr> predicate;
 
   void VisitAttrs(AttrVisitor* v) {
     v->Visit("dtype", &(this->dtype));
@@ -680,7 +680,7 @@ class BufferLoadNode : public PrimExprNode {
 class BufferLoad : public PrimExpr {
  public:
   TVM_DLL explicit BufferLoad(Buffer buffer, Array<PrimExpr> indices,
-                              PrimExpr predicate = PrimExpr(), Span span = Span());
+                              Optional<PrimExpr> predicate = NullOpt, Span span = Span());
   TVM_DEFINE_OBJECT_REF_METHODS(BufferLoad, PrimExpr, BufferLoadNode);
   TVM_DEFINE_OBJECT_REF_COW_METHOD(BufferLoadNode);
 };
--- a/include/tvm/tir/stmt.h
+++ b/include/tvm/tir/stmt.h
@@ -231,11 +231,14 @@ class BufferStoreNode : public StmtNode {
   PrimExpr value;
   /*! \brief The indices location to be stored. */
   Array<PrimExpr> indices;
+  /*! \brief The predicate mask for storing values. */
+  PrimExpr predicate;
 
   void VisitAttrs(AttrVisitor* v) {
     v->Visit("buffer", &buffer);
     v->Visit("value", &value);
     v->Visit("indices", &indices);
+    v->Visit("predicate", &predicate);
     v->Visit("span", &span);
   }
 
@@ -248,6 +251,7 @@ class BufferStoreNode : public StmtNode {
     hash_reduce(buffer);
     hash_reduce(value);
     hash_reduce(indices);
+    hash_reduce(predicate);
   }
 
   static constexpr const char* _type_key = "tir.BufferStore";
@@ -261,7 +265,7 @@ class BufferStoreNode : public StmtNode {
 class BufferStore : public Stmt {
  public:
   TVM_DLL explicit BufferStore(Buffer buffer, PrimExpr value, Array<PrimExpr> indices,
-                               Span span = Span());
+                               PrimExpr predicate = PrimExpr(), Span span = Span());
 
   TVM_DEFINE_OBJECT_REF_METHODS(BufferStore, Stmt, BufferStoreNode);
   TVM_DEFINE_OBJECT_REF_COW_METHOD(BufferStoreNode);
--- a/include/tvm/tir/stmt.h
+++ b/include/tvm/tir/stmt.h
@@ -232,7 +232,7 @@ class BufferStoreNode : public StmtNode {
   /*! \brief The indices location to be stored. */
   Array<PrimExpr> indices;
   /*! \brief The predicate mask for storing values. */
-  PrimExpr predicate;
+  Optional<PrimExpr> predicate;
 
   void VisitAttrs(AttrVisitor* v) {
     v->Visit("buffer", &buffer);
@@ -265,7 +265,7 @@ class BufferStoreNode : public StmtNode {
 class BufferStore : public Stmt {
  public:
   TVM_DLL explicit BufferStore(Buffer buffer, PrimExpr value, Array<PrimExpr> indices,
-                               PrimExpr predicate = PrimExpr(), Span span = Span());
+                               Optional<PrimExpr> predicate = NullOpt, Span span = Span());
 
   TVM_DEFINE_OBJECT_REF_METHODS(BufferStore, Stmt, BufferStoreNode);
   TVM_DEFINE_OBJECT_REF_COW_METHOD(BufferStoreNode);
--- a/python/tvm/tir/buffer.py
+++ b/python/tvm/tir/buffer.py
@@ -141,6 +141,57 @@ class Buffer(Object, Scriptable):
         begin = (begin,) if isinstance(begin, (int, PrimExpr)) else begin
         return _ffi_api.BufferVStore(self, begin, value)  # type: ignore
 
+    def load(self, indices, predicate=None):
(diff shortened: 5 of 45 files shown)


lhutton1:
cc @ekalda @tqchen @Lunderberg @cbalint13 @Anndrey24

include/tvm/script/ir_builder/tir/ir.h
```diff
@@ -411,8 +411,10 @@ Var EnvThread(String thread_tag, DataType dtype = DataType::Int(32));
  * \param buffer The buffer.
  * \param value The value to be stored.
  * \param indices The indices location to be stored.
+ * \param predicate A vector mask of int1 values indicating which lanes of a vector are to be
```
Lunderberg, line 414: Can we specify that the number of lanes for the `predicate` must be equal to the number of lanes in the `value`?

include/tvm/tir/expr.h
```diff
@@ -675,7 +679,8 @@ class BufferLoadNode : public PrimExprNode {
  */
 class BufferLoad : public PrimExpr {
  public:
-  TVM_DLL explicit BufferLoad(Buffer buffer, Array<PrimExpr> indices, Span span = Span());
+  TVM_DLL explicit BufferLoad(Buffer buffer, Array<PrimExpr> indices,
+                              PrimExpr predicate = PrimExpr(), Span span = Span());
```
Lunderberg, line 683: Using `PrimExpr()` as the default would have an undefined object as the predicate.  While this is legal, it is rather unexpected.  While changing `PrimExpr` to be non-nullable would involve more breakage than I'd want to handle at the moment, we should avoid introducing additional cases where `PrimExpr` contains null.

Changing this to `Optional<PrimExpr> predicate = NullOpt` would avoid this issue.

include/tvm/tir/stmt.h
```diff
@@ -261,7 +265,7 @@ class BufferStoreNode : public StmtNode {
 class BufferStore : public Stmt {
  public:
   TVM_DLL explicit BufferStore(Buffer buffer, PrimExpr value, Array<PrimExpr> indices,
-                               Span span = Span());
+                               PrimExpr predicate = PrimExpr(), Span span = Span());
```
Lunderberg, line 268: Same request here, `Optional<PrimExpr> predicate = NullOpt`.

python/tvm/tir/buffer.py
```diff
@@ -141,6 +141,57 @@ def vstore(self, begin, value):
         begin = (begin,) if isinstance(begin, (int, PrimExpr)) else begin
         return _ffi_api.BufferVStore(self, begin, value)  # type: ignore
 
+    def load(self, indices, predicate=None):
```
Lunderberg, line 144: These look very similar to the existing `vload` and `vstore` methods.  Should they be extended instead with a `predicate` argument instead, rather than introducing new methods?

python/tvm/tir/expr.py
```diff
@@ -1093,20 +1093,27 @@ class BufferLoad(PrimExprWithOp):
...
 
     indices : List[PrimExpr]
-        The buffer indices.
+        The buffer indices to load values from.
 
     span : Optional[Span]
         The location of this expression in the source code.
+
+    predicate : Optional[PrimExpr]
+        A vector mask of int1 values indicating which lanes of a vector are to be loaded.
```
Lunderberg, line 1102: At first glance, the `int1` reads like a typo.  Can we explicitly call it `boolean values` instead?

python/tvm/tir/stmt.py
```diff
@@ -224,24 +224,29 @@ class BufferStore(Stmt):
     indices : List[PrimExpr]
         The indices location to be stored.
 
+    predicate : Optional[PrimExpr]
+        A vector mask of int1 values indicating which lanes of a vector are to be stored.
```
Lunderberg, line 228: Same nitpick here, using `boolean` instead of `int1`.

src/tir/transforms/vectorize_loop.cc
```diff
@@ -72,6 +72,126 @@ inline PrimExpr BroadcastTo(PrimExpr e, int lanes, bool is_scalable) {
   return Broadcast(e, CreateNewLanes(is_scalable, lanes));
 }
 
+bool EnableBufferLevelPredication() {
```
Lunderberg, line 75: I like the fallback, where it checks whether a target supports SVE in order to apply the vectorization if not specified in the pass context.  However, this function should accept a `Target` argument from the calling scope rather than using `Target::Current()`, because the `Target::Current()` may be overridden. 

The override can be either at the function level (`prim_func->GetAttr<Target>(tvm::attr::kTarget)`), or for individual blocks of TIR (an `AttrStmt` with `attr_stmt->attr_key==tvm::attr::kTarget`).

Lunderberg: This will require updating `arith::TargetHasSVE()` as well.  Apologies for not catching it earlier in #16893.

tests/python/tir-transform/test_tir_transform_vectorize.py
```diff
@@ -488,5 +529,170 @@ def main(A: T.Buffer((16,), "float32")):
...
+        for i_0, i_1_s in T.grid(4, 4):
+            if i_0 * 4 + i_1_s < 14:
+                B[i_0 * 4 + i_1_s] = A[i_0 * 4 + i_1_s] + T.float32(1)
+
+    mod = tvm.IRModule.from_expr(before)
+    with tvm.transform.PassContext(config={"tir.enable_buffer_level_predication": False}):
+        with tvm.target.Target("llvm -mtriple=aarch64-linux-gnu -mattr=+sve"):
+            after = tvm.tir.transform.VectorizeLoop()(mod)["main"]
+    tvm.ir.assert_structural_equal(after, expected)
+
```
Lunderberg, line 695: Can we also add test cases for the target override both in the function attributes (`T.func_attr({'target': T.target(...)})`) and in an attr stmt (`T.attr(T.target(...), "target", 0)`)?

([Link](https://github.com/apache/tvm/blob/main/tests/python/tir-transform/test_tir_transform_annotate_device_regions.py#L27) to example for the TVMScript syntax.)

src/target/llvm/codegen_llvm.cc
```diff
@@ -1768,11 +1774,17 @@ llvm::Value* CodeGenLLVM::VisitExpr_(const BufferLoadNode* op) {
...
-                                  bool is_volatile) {
+  auto make_load = [this, &loads](TypedPointer buffer_ptr, int /* subelement_i */,
+                                  llvm::Value* predicate, int alignment, bool is_volatile) {
 #if TVM_LLVM_VERSION >= 110
-    auto load = builder_->CreateAlignedLoad(buffer_ptr.type, buffer_ptr.addr,
-                                            llvm::Align(alignment), is_volatile);
+    llvm::Instruction* load = nullptr;
+    if (predicate != NULL) {
+      load = builder_->CreateMaskedLoad(buffer_ptr.type, buffer_ptr.addr, llvm::Align(alignment),
+                                        predicate);
```
Lunderberg, line 1783: It looks like this ignores the `is_volatile` argument.  Can we provide `is_volatile` to `CreateMaskedLoad`?  If not, we should specify that these two arguments are incompatible and throw an exception if they are both provided.

src/target/llvm/codegen_llvm.cc
```diff
@@ -1768,11 +1774,17 @@ llvm::Value* CodeGenLLVM::VisitExpr_(const BufferLoadNode* op) {
...
+    } else {
+      load = builder_->CreateAlignedLoad(buffer_ptr.type, buffer_ptr.addr, llvm::Align(alignment),
+                                         is_volatile);
+    }
 #elif TVM_LLVM_VERSION >= 80
     auto load =
```
Lunderberg, line 1789: The PR only adds `CreateMaskedLoad` when `TVM_LLVM_VERSION >= 110`.  If somebody is using an older version of LLVM, it would silently ignore the predicate for the load/store.  We should either support it, or throw an exception.

It looks like `CreateMaskedLoad` has been supported in LLVM since [this commit](https://github.com/llvm/llvm-project/commit/f1de34b84dea91b5060ee0fafbadaad5deaf199c), so I'd lean toward adding it in the other `#elif` branches.

tests/python/codegen/test_target_codegen_aarch64.py
```diff
@@ -700,5 +700,31 @@ def before(a: T.handle):
     assert "get.active.lane.mask" in ll
 
 
+@pytest.mark.skipif(
```
Lunderberg, line 703: This validates that we have the correct output for architectures that support SVE, but it doesn't test the behavior of other targets that do not (yet) support predicated loads/stores.  While the `VectorizeLoop` pass would only insert a predicated load/store for targets that support it, the predicated load/store could still be generated in hand-written kernels, or through other transforms in the future.

Can we add a test, parametrized over each target tested in CI, which attempts to compile a PrimFunc containing predicated loads/stores?  For each target that supports sve, `tvm.build` should compile without error, and for each target that does not, `tvm.build` should raise an exception.

src/tir/ir/stmt.cc
```diff
@@ -458,7 +458,8 @@ TVM_REGISTER_GLOBAL("tir.Evaluate").set_body_typed([](PrimExpr value, Span span)
 TVM_REGISTER_NODE_TYPE(EvaluateNode);
 
 // BufferStore
-BufferStore::BufferStore(Buffer buffer, PrimExpr value, Array<PrimExpr> indices, Span span) {
+BufferStore::BufferStore(Buffer buffer, PrimExpr value, Array<PrimExpr> indices, PrimExpr predicate,
+                         Span span) {
   ICHECK_EQ(buffer->shape.size(), indices.size())
```
Lunderberg, line 463: We should add an assert here, that if `predicate` is defined, the number of lanes in `value` and in `predicate` are identical.

Lunderberg (changes requested):
I like the functionality overall.  I have a couple of requested changes, which mostly fall under the categories below.

* TIR edge cases, when `Target::Current()` may be overridden.
* Using `Optional<PrimExpr>` instead of `PrimExpr`.
* Validating that `!predicate.defined()` in any target that does not support it.

python/tvm/tir/buffer.py
lhutton1: We did look at reusing these but decided against it since the vload/vstore api doesn't let you specify the "indices" argument of the underlying buffer load/store which can limit expressing more complex (e.g. non-consecutive) accesses

src/tir/transforms/vectorize_loop.cc
lhutton1: Thanks, that makes sense. Do you know if there is any general infrastructure for keep track of the current target (which takes into account this override functionality) from within a pass? Otherwise I feel we will be duplicating this functionality in multiple places. I was thinking something similar to: `LexicalOnDeviceMixin` (assuming I understood it correctly)

lhutton1 (commented):
Thanks for taking the time to review @Lunderberg, I'm working through the comments but wanted to ask a couple of questions / respond to some of the comments before I continue

src/tir/transforms/vectorize_loop.cc
Lunderberg: To my knowledge, there isn't a general utility that would provide it, though I definitely agree on its usefulness.

There's currently a couple similar utilities, such as [`IRMutatorWithAnalyzer`](https://github.com/apache/tvm/blob/main/src/arith/ir_mutator_with_analyzer.h#L45) and [`IRVisitorWithAnalyzer`](https://github.com/apache/tvm/blob/main/src/arith/ir_visitor_with_analyzer.h#L35), which allow context-dependent algebraic simplifications. The easiest way to implement such a utility for target-aware simplifications would be to extract out the context-dependent target tracking from an existing visitor/mutator (e.g. [here](https://github.com/apache/tvm/blob/main/src/tir/transforms/lower_device_kernel_launch.cc#L65), used in the `LowerDeviceKernelLaunch` transform) into a `IRVisitorWithTarget` and `IRMutatorWithTarget`, which is then subclassed again by each transform that requires it.

For long-term maintainability, I like your suggestion of mixins much more than the subclasses, because they could be composed together much more easily.  Implemented with subclasses, a hypothetical transform that requires both the context-dependent target and context-dependent algebraic simplifications would have incorrect behavior from multiple inheritance, and would need to re-implement one or the other functionalities.

Unfortunately, I don't think the current `tir::StmtVisitor` API could be used to support mixins, either through multiple inheritance or through composition.  The main issue is that the implementable virtual functions (e.g. `tir::StmtVisitor::VisitStmt_(const tir::ForNode*)`) aren't granular enough.  In order to compose separate transforms that provide context-dependent information, it would require separate virtual functions for pre-visit and post-visit.  I've toyed around with different variations of the TIR stmt/expr visitor/mutator implementations, but haven't had the time to make anything PR-able.

TL;DR: It doesn't currently exist, would be very useful to have, and a subclass would be sufficient for this PR, though long-term maintainability concerns may appear at some point in the future.

python/tvm/tir/buffer.py
Lunderberg: Ah, I see.  I think this is an unfortunate confusion resulting from the name `begin` for the parameter in `vload` and `vstore`.  This parameter would be better named `indices`, because it is used to set the `indices` argument of the buffer load/store.

While there's some additional logic to handle automatic promotion from a scalar index to a vector index when needed ([link](https://github.com/apache/tvm/blob/main/src/tir/ir/buffer.cc#L410)), a vector index can also be provided directly.

src/target/llvm/codegen_llvm.cc
Lunderberg: Thank you, and throwing the exception makes sense.  (And we should probably also add a unit test to verify that the exception is thrown.)

Lunderberg:
No problem, and thank you on the revisions!

python/tvm/tir/expr.py
lhutton1: Done. There was a bit of confusion here since LLVM intrinsics expect predicate masks of `int1` values. However, boolean types are already converted to `int1` in the codegen backend. So I have required `uint1`/boolean types in TVM.

src/target/llvm/codegen_llvm.cc
lhutton1: Done

src/tir/transforms/vectorize_loop.cc
lhutton1: Thanks for the detailed description. To save this PR exploding further, I've added the minimum changes necessary for the vectorizer pass. Perhaps we can clean this up and fix the target checks in the analyzer in a separate PR?

tests/python/tir-transform/test_tir_transform_vectorize.py
lhutton1: Done

tests/python/codegen/test_target_codegen_aarch64.py
lhutton1: Done

src/tir/ir/stmt.cc
lhutton1: Done

ekalda (commented):
Thanks @lhutton1 for all the work on this (it's a lot of work) and @Lunderberg for constructive reviews! I've only got some minor nits.

src/tir/ir/stmt.cc
```diff
@@ -476,29 +477,39 @@ BufferStore::BufferStore(Buffer buffer, PrimExpr value, Array<PrimExpr> indices,
...
 
   ICHECK_EQ(index_lanes * buffer_lanes, value_dtype_lanes)
       << "Cannot store value with " << value_dtype_lanes << ", expected value with "
       << index_lanes * buffer_lanes << " (" << index_lanes << " index lanes * " << buffer_lanes
       << " buffer element lanes)";
 
+  if (predicate.defined()) {
+    DataType predicate_dtype = predicate.value().dtype();
+    int predicate_dtype_lanes = predicate_dtype.get_lanes_or_vscale_factor();
+    ICHECK_EQ(value_dtype_lanes, predicate_dtype_lanes)
```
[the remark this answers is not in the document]
lhutton1: Good catch, yes we should

include/tvm/tir/buffer.h
```diff
@@ -209,14 +209,20 @@ class Buffer : public ObjectRef {
    * \brief Create an Expr that does a vector load at begin index.
    * \param begin The beginning index
    * \param dtype The data type to be loaded.
+   * \param predicate A vector mask of boolean values indicating which lanes of a vector are to be
+   * stored. The number lanes of the mask must be equal to the number of lanes in value.
```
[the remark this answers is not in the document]
lhutton1: Oops thanks, result of copy and pasting

ekalda (approved):
Thanks @lhutton1 LGTM! 🚀

lhutton1:
@tvm-bot rerun

lhutton1:
friendly ping @Lunderberg if you have some free time

leandron (approved):
This patch is waiting here for more than a week now. I think we should merge it before it starts creating merge-conflicts, so I'm going to merge it now.

@Lunderberg, I'm sure @lhutton1 will be happy to take any other outstanding comment in follow-up patches, please reply when you can. Thanks.

tests/python/codegen/test_target_codegen_aarch64.py
```diff
@@ -780,5 +780,31 @@ def before(a: T.handle):
...
+    def before(a: T.handle, b: T.handle):
+        A = T.match_buffer(a, (16,), "float32")
+        B = T.match_buffer(b, (16,), "float32")
+        T.func_attr({"global_symbol": "main", "tir.noalias": True})
+        for i_0 in T.serial(T.ceildiv(16, 4 * T.vscale())):
+            for i_1 in T.vectorized(4 * T.vscale()):
+                if i_0 * 4 * T.vscale() + i_1 < 14:
+                    B[i_0 * 4 * T.vscale() + i_1] = A[i_0 * 4 * T.vscale() + i_1] + 1.0
+
+    with tvm.target.Target(target):
```
tqchen, line 800: we should avoid this style of code(this was created for legacy usage), instead, attach the target attribute to the PrimFunc itself

Lunderberg:
Apologies for not getting back to the review, and thank you for making the changes!

PR #16975, opened 2024-05-07
[CI] Enable unittests for disco

Lunderberg:
The refactor of the unittest folder in #16110 did not include the `tests/python/disco` folder in the list of folders to run in CI.

### DIFF
--- a/tests/scripts/task_python_unittest.sh
+++ b/tests/scripts/task_python_unittest.sh
@@ -39,6 +39,7 @@ TEST_FILES=(
   "auto_scheduler"
   "autotvm"
   "codegen"
+  "disco"
   "ir"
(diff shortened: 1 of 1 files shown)


yongwww (approved):

Lunderberg:
Looks like there's some failing tests, which seem to have crept in while the tests were disabled.

* There is no `tvm.testing.disco` that can be imported.  Looks like it should instead import the `exec_worker` to get the PackedFunc definitions from that file.

* Ill-formed TIR, which is now validated when parsing TVMScript.

* NCCL not currently installed in CI environment, and not all NCCL-dependent tests are marked with `tvm.testing.requires_nccl`.

PR #16978, opened 2024-05-08
[Disco] Implement `num_workers` property for `disco.Session`

Lunderberg:
Prior to this commit, while the `num_workers` argument was provided to the `disco.Session` object, it could not be determined from an existing `disco.Session` object.  As a result, functions that interacted with a multi-GPU setup frequently required separate `num_workers` and `disco_session` argument, which could erroneously be out-of-sync (e.g. passing the incorrect `num_workers`, or omitting the `disco_session` argument when `num_workers>1`).

To remove this class of errors, this commit adds a `disco.Session.num_workers` property.  The separate `num_workers` argument is no longer necessary, as it can be determined from the `disco.Session` instance.

### DIFF
--- a/include/tvm/runtime/disco/session.h
+++ b/include/tvm/runtime/disco/session.h
@@ -197,6 +197,8 @@ class SessionObj : public Object {
    * The thirtd element is the function to be called.
    */
   TVM_DLL virtual DRef CallWithPacked(const TVMArgs& args) = 0;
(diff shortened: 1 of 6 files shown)


masahi (approved):

masahi:
@Lunderberg Need to resolve the conflict.

Lunderberg:
Thank you, and conflict resolved!

PR #16988, opened 2024-05-12
[KVCache] Fix the aux data syncing order of paged KV cache

rickzx:
Fix the aux data syncing order of paged KV cache, so that `q_rope_position_map` is the first one to be synced and has a 0 byte_offset.


tqchen (approved):

MasterJH5574 (approved):
Thanks for the fix!

PR #16992, opened 2024-05-13
[Bugfix][Disco] Handle NDArray larger than OS buffer for pipe

Lunderberg:
Prior to this commit, using `disco.Session` methods to transfer `NDArray` instances to workers could raise an exception if the `NDArray` is larger than the buffer allocated by the OS for the controller/worker pipe.  In these case, the first call to the `Read` method of `tvm::support::Pipe` would successfully return, but only with the initial bytes of the `NDArray`.  Receiving the full `NDArray` requires repeatedly calling the POSIX `read` function.

This commit updates the `Read` and `Write` methods of `tvm::support::Pipe` to repeatedly call the underlying read/write methods, until the full `NDArray` has been transferred.

This commit does not add any unit tests, as the existing unit test `tests/python/disco/test_ccl.py::test_attention[nccl-ProcessSession]` requires this change to pass.

### DIFF
--- a/src/support/pipe.h
+++ b/src/support/pipe.h
@@ -86,8 +86,19 @@ class Pipe : public dmlc::Stream {
     DWORD nread = static_cast<DWORD>(RetryCallOnEINTR(fread, GetLastErrorCode));
     ICHECK_EQ(static_cast<size_t>(nread), size) << "Read Error: " << GetLastError();
 #else
-    ssize_t nread = RetryCallOnEINTR([&]() { return read(handle_, ptr, size); }, GetLastErrorCode);
-    ICHECK_GE(nread, 0) << "Write Error: " << strerror(errno);
+    size_t nread = 0;
+    while (size) {
+      ssize_t nread_chunk =
+          RetryCallOnEINTR([&]() { return read(handle_, ptr, size); }, GetLastErrorCode);
+      ICHECK_NE(nread_chunk, -1) << "Write Error: " << strerror(errno);
+
+      ICHECK_GT(nread_chunk, 0) << "Was unable to read any data from pipe";
+      ICHECK_LE(nread_chunk, size) << "Read " << nread_chunk << " bytes, "
+                                   << "but only expected to read " << size << " bytes";
+      size -= nread_chunk;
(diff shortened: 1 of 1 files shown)


masahi (approved):

tqchen:
Thanks @Lunderberg ! 
In this case, i think it is better to introduce a ReadAll and WriteAll function, in pairing with https://github.com/apache/tvm/blob/main/src/support/socket.h#L492,  and then we call these functions instead

Just to keep Socket and pipe behavior consistent with each other, low-level read/write can remain partial while readall and writeall contains convenient method for the intended behavior

Lunderberg:
I agree.  This is a stop-gap measure, as the longer-term fix will require updating the `dmlc::Stream` API to return `size_t` from `Write`, similar to `Read`.  Without that API change, the calling scope cannot correctly implement a `WriteAll` method.

Lunderberg:
@tqchen The changes for the long-term fix are implemented in https://github.com/dmlc/dmlc-core/pull/686, with TVM compatibility changes implemented in https://github.com/apache/tvm/pull/16998.  Once those changes land, we will be able to remove the stop-gap implementation, and provide `ReadAll` and `WriteAll` helper methods.

tqchen:
Ah I see, the Stream interface was actually created to be in analogy of FileStream interface(which do not have partial writes). After seeing your fix, i realized that this PR actually did the right thing given pipe inheritated from the stream interface

tqchen:
The socket interface was mainly meant to accomodate for the non-blocking case where full read is not always possible

Lunderberg:
> Ah I see, the Stream interface was actually created to be in analogy of FileStream interface(which do not have partial writes).

Thank you, and that history makes a lot of sense.  I'm thinking it would still be a good change (though low priority) to expose partial writes in `dmlc::Stream` interface, because the `Stream` interface is also used for pipes and TCP sockets.

> The socket interface was mainly meant to accomodate for the non-blocking case where full read is not always possible

Good point.  Looking into it, it looks like the `Socket` interface also includes functionality that isn't appropriate for a pipe, such as the address/port to listen on, accepting new connections, etc.

PR #17000, opened 2024-05-15
[BugFix][MSC] split name_string with index by colon from the right

psunn:
Fixes a naming mismatch in MSCGraph where tensor_name could formatted as 'string:index:index',and the corresponding node.name is 'string:index'. Splitting tensor_name from the right aligns it correctly.

For example, the TFLite default input name 'serving_default_input:0' becomes 'serving_default_input:0:0' in MSCGraph, while node.name remains 'serving_default_input:0'.

### DIFF
--- a/src/contrib/msc/core/utils.h
+++ b/src/contrib/msc/core/utils.h
@@ -142,7 +142,7 @@ class StringUtils {
    */
   TVM_DLL static const std::tuple<String, String> SplitOnce(const String& src_string,
                                                             const String& sep,
-                                                            bool from_left = true);
+                                                            bool from_left = false);
 
   /*!
    * \brief Get the tokens between left and right.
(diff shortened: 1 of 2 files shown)


lhutton1 (commented):
Thanks for the fix @psunn, would it be possible to add a test case?

Hzfengsy:
cc @Archermmt

psunn:
> Thanks for the fix @psunn, would it be possible to add a test case?

@lhutton1 Thanks for the review. I've added a test case.

psunn:
Hi @Archermmt, could you help to review this PR? thanks in advance.

Archermmt:
Seem ok for me. The only change in this PR is the split direction, which has no influence on current test. I think the main reason is that: in some corner cases like TFLite model, the node name looks like "NODE:IDX", which is normally the tensor name format.

lhutton1 (approved):

lhutton1:
Thanks @psunn @Archermmt @Hzfengsy!

PR #17001, opened 2024-05-15
[SVE] Use only powers of two as possible vscale values

lhutton1:
When analyzing scalable expressions, the analyzer will iterate over a series of known vscale values in the range 1-16. However, we can tighten this range to only values that are a power of two, as stated in the [LLVM lang ref](https://llvm.org/docs/LangRef.html#llvm-vscale-intrinsic:~:text=This%20function%20attribute%20indicates%20vscale%20is%20a%20power%2Dof%2Dtwo%20within%20a%20specified%20range) and more generally the [reference manual](https://developer.arm.com/documentation/ddi0487/latest/).

This comes from a discussion in https://github.com/apache/tvm/pull/16921#discussion_r1600048788


lhutton1:
cc @ekalda @Anndrey24

lhutton1:
Credit here is to @Anndrey24 ;)

PR #17004, opened 2024-05-16
Pass default values to target config/options

MNGanesan:
BYOC Compiler Config node defines the target compiler's command line options, along with default values. 
This change extract the default values while constructing target options for codegen/target compiler.
Added test case for this feature as well.


MNGanesan:
For my PR,  cortexm/pr-head test is failing, on investigation I find 

boolean type argument  of CMSIS-NN is not passed with numeric value and got below error

[2024-05-16T14:29:37.357Z] tvmc compile: error: argument --target-cmsis-nn-debug_last_error: invalid 

Argument passed: 

"--target-cmsis-nn-debug_last_error=false" 

Based on the python argument parse, above error looks to be valid, however it is not observed other PR.
I could not decipher the link between this error and my PR. Could some one put a light on it ? How it is passing otherwise when python argument parser looks to be valid ?

Appreciate your insights!

PR #17006, opened 2024-05-17
[CUBLAS][FP8] Enable fusing astype operation for matmul multiply pattern

valmat07:
This PR adds fusing of the astype operation to matmul for cublas. This change is needed to improve the performance for fp8.


vinx13:
do we need to update cublas codegen or runtime to support the cast?

PR #17022, opened 2024-05-23
[Relax][Bugfix] Apply FuseOps to nested DataflowBlock

Lunderberg:
While it is ill-formed for control-flow to occur within a `DataflowBlock`, it is legal for a `DataflowBlock` to be contained within a control-flow.  Prior to this commit, the `FuseOps` and `FuseOpsByPattern` transforms erroneously skipped `DataflowBlock` instances that were contained within a `relax::If` node.

This commit updates `FuseOps` to apply operator fusion to any dataflow block, regardless of whether it is found at the top level of a a Relax function.

### DIFF
--- a/src/relax/transform/fuse_ops.cc
+++ b/src/relax/transform/fuse_ops.cc
@@ -108,9 +108,16 @@ class GraphCreator : public ExprVisitor {
   static IndexedForwardGraph Create(IRModule mod, support::Arena* arena) {
     GraphCreator creator(mod, arena);
     for (const auto& it : mod->functions) {
-      // Only visit Relax function without attr kPrimitive.
+      // Only visit Relax functions with neither attr::kPrimitive nor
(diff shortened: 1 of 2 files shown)


src/relax/transform/fuse_ops.cc
```diff
@@ -262,16 +262,11 @@ class GraphCreator : public ExprVisitor {
     IndexedForwardGraph::Node* leaf_node = nullptr;
     if (it != graph_.node_map.end()) {
       leaf_node = it->second;
-    } else if (leaf_expr->IsInstance<ConstantNode>() || leaf_expr->IsInstance<ShapeExprNode>() ||
-               leaf_expr->IsInstance<PrimValueNode>() || leaf_expr->IsInstance<StringImmNode>() ||
```
tqchen, line 266: Is this intended change? Seems that we should consider sass where constant value(and there may not be a binding pt)

tqchen (changes requested):

PR #17023, opened 2024-05-23
[Relax][Bugfix] Bind symbolic variables in R.match_cast

Lunderberg:
Prior to this commit, variable replacement by `BindSymbolicVars` would fail to replace variables that occur within a `relax::MatchCast` node. This pattern is rare, because the `bind_symbolic_vars` method can only replace variables that are exposed as part of the function signature, and most uses of `relax::MatchCast` act as a definition for symbolic variables that are not exposed through the function signature.  This pattern is well-formed, though, since the `relax::MatchCast` node can also act as a user of previously-defined symbolic variables.

The root cause for this bug was in the `ExprMutator` visitor for `relax::MatchCast`, which did not visit the struct info field.  As a result, the virtual `ExprMutator::VisitPrimExpr` function was not called for expressions that occur within the `StructInfo` of a `relax::MatchCast`.  This commit updates `ExprMutator` to resolve this bug, and applies an analogous fix for `ExprVisitor`.

### DIFF
--- a/src/relax/ir/expr_functor.cc
+++ b/src/relax/ir/expr_functor.cc
@@ -257,6 +257,7 @@ RELAX_EXPR_VISITOR_VISIT_BINDING_IMPL(DataTypeImmNode);
 
 void ExprVisitor::VisitBinding_(const MatchCastNode* binding) {
   this->VisitExpr(binding->value);
+  this->VisitExprDepStructInfoField(binding->struct_info);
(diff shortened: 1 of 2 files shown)


tqchen (approved):

Lunderberg:
Closing to resubmit as a new PR (same issue as mentioned [here](https://github.com/apache/tvm/pull/17024#issuecomment-2132387858)), due to CI issues.

PR #17024, opened 2024-05-23
[Relax][Bugfix] Annotate ComputePrimValue output as host function

Lunderberg:
The `ComputePrimValue` transform is used to compute the value of symbolic expressions that may appear within a Relax function.  For example, to compute a boolean condition used for a `relax::If` node. These functions are used for small host-side computations, prior to launching a device kernel.

This commit updates `ComputePrimValue` to annotate the generated `PrimFunc` with `tir::attr::kIsHostFunc`.  This annotation is required for correct behavior in `tvm.dlight.ApplyDefaultSchedule`, to avoid erroneous scheduling of this function for the GPU, and for `tir::transform::BindTarget`, to ensure that the function is compiled for execution on the host.

### DIFF
--- a/src/relax/transform/compute_prim_value.cc
+++ b/src/relax/transform/compute_prim_value.cc
@@ -45,7 +45,8 @@ class PrimValueComputeInjector : public ExprMutator {
     auto param_vars = tir::UndefinedVars(node->value);
     tir::Stmt body = tir::Evaluate(tir::Call(ret_dtype, tir::builtin::ret(), {node->value}));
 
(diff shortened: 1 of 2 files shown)


tqchen (approved):

tqchen:
@tvm-bot rerun

tqchen:
due to recent ci restart, likely we need to close and re create another pr for the ci to get retriggerd

Lunderberg:
Thank you for looking into the CI failures, and I'll resubmit as a new PR.  Looks like #17022 and #17023 have the similar CI failure, as they were all submitted while the CI was down.

PR #17033, opened 2024-05-28
[Relax][Bugfix] Apply FuseOps to nested DataflowBlock

Lunderberg:
While it is ill-formed for control-flow to occur within a `DataflowBlock`, it is legal for a `DataflowBlock` to be contained within a control-flow.  Prior to this commit, the `FuseOps` and `FuseOpsByPattern` transforms erroneously skipped `DataflowBlock` instances that were contained within a `relax::If` node.

This commit updates `FuseOps` to apply operator fusion to any dataflow block, regardless of whether it is found at the top level of a a Relax function.

### DIFF
--- a/src/relax/transform/fuse_ops.cc
+++ b/src/relax/transform/fuse_ops.cc
@@ -108,9 +108,16 @@ class GraphCreator : public ExprVisitor {
   static IndexedForwardGraph Create(IRModule mod, support::Arena* arena) {
     GraphCreator creator(mod, arena);
     for (const auto& it : mod->functions) {
-      // Only visit Relax function without attr kPrimitive.
+      // Only visit Relax functions with neither attr::kPrimitive nor
+      // attr::kCodegen.  Relax functions with `attr::kPrimitive` are
+      // previously fused functions, potentially from a previous use
+      // of `FuseOps` or `FuseOpsByPattern`.  Relax functions with
+      // `attr::kCodegen` are previously fused functions from
+      // `FuseOpsByPattern`, when the `annotate_codegen` option is
+      // true.
       const auto* func = it.second.as<FunctionNode>();
-      if (func == nullptr || func->HasNonzeroAttr(attr::kPrimitive)) {
+      if (func == nullptr || func->HasNonzeroAttr(attr::kPrimitive) ||
+          func->GetAttr<String>(attr::kCodegen).defined()) {
         continue;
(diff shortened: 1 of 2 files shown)


Lunderberg:
This PR is identical to the now-closed PR https://github.com/apache/tvm/pull/17022, as a CI restart prevented the CI from running on the original PR.

src/relax/transform/fuse_ops.cc
```diff
@@ -262,16 +262,11 @@ class GraphCreator : public ExprVisitor {
     IndexedForwardGraph::Node* leaf_node = nullptr;
     if (it != graph_.node_map.end()) {
       leaf_node = it->second;
-    } else if (leaf_expr->IsInstance<ConstantNode>() || leaf_expr->IsInstance<ShapeExprNode>() ||
-               leaf_expr->IsInstance<PrimValueNode>() || leaf_expr->IsInstance<StringImmNode>() ||
-               leaf_expr->IsInstance<DataTypeImmNode>()) {
+    } else {
```
Lunderberg, line 265: @tqchen Responding to the comment [here](https://github.com/apache/tvm/pull/17022#discussion_r1613292617), with conversation migrated to the new PR.

> Is this intended change? Seems that we should consider sass where constant value(and there may not be a binding pt)

This change is intentional, and maintains the same behavior for constants as before.  Prior to this change, the `else if` condition was entered for five of the six child classes of `relax::LeafExprNode`, and the `else` branch with the `LOG(FATAL)` was only entered for a `relax::Var`.

With this change, the behavior of all LeafExpr subclasses is the same: If encoutered without being previously assigned to a group, they are treated as a group of size one.

Lunderberg: This resolves an issue that results from the use of `PostOrderVisit` [here](https://github.com/apache/tvm/blob/main/src/relax/transform/fuse_ops.cc#L230), in `VisitUnsupportedNode`.  It looks like this function is intended to collect all `Var`/`Constant` nodes that are part of the RHS of a binding (e.g. extracting `A` and `B` from `C = R.add(A,B)`).  However, it doesn't collect any bindings that may be part of a nested expression.  As a result, variable bindings that are part of a conditional's body would erroneously trigger the `LOG(FATAL)` in `VisitLeaf`.

tqchen (approved):

PR #17036, opened 2024-05-28
[Runtime] Use preferred host memory (pinned memory) in KV cache

MasterJH5574:
This PR updates the PagedKVCache with the pinned memory support, which can reduce the copy overhead between CPU and GPU.

This PR also bumps FlashInfer version, which now supports
* specifying kernels to build via cmake,
* pinned memory as host memory.

We also update CMakeLists.txt and config.cmake to include the FlashInfer compile options. Prior to this PR, the kernels being built is hardcoded in FlashInfer header files.


tqchen:
@tvm-bot rerun

tqchen (approved):

PR #17039, opened 2024-05-29
[TIR] Fix Bug in VectorizeLoop 

CharlieFRuan:
This PR fixes a bug in vectorize loop introduced related to https://github.com/apache/tvm/pull/16966

The visit to condition can write need scalarize to true then the followup visit to then case can trigger an ICHECK.

The visit to let value can also write need scalarize flag in which case we need to immediately scalarize.


lhutton1 (approved):
Thanks for the fix, apologies for missing it previously!

tqchen:
@tvm-bot  rerun

tqchen:
@tvm-bot rerun

ekalda:
Thanks @tqchen and @CharlieFRuan for the fix!

PR #17038, opened 2024-05-28
Revert "[SME][TOPI] Add conv2d NHWC SME fp32 schedule (#17003)"

tqchen:
This reverts commit cab54e0dee82f84d94cd65f8fe0432ee1c2f2e22.

PR #16981, opened 2024-05-08
[SME] Add scalable fp16->fp32 dense schedule

lhutton1:
This commit extends the functionality of the SME dense and matmul schedules to support operations with fp16 inputs and an fp32 output, where `transpose_a=False` and `transpose_b=True`.

For convenience, it also adds a utility called `get_vscale_factor` which created the correct multiplier for `vscale` given a data type, reflecting ideas from an early design of the [SVE](https://github.com/apache/tvm-rfcs/pull/104) RFC.

Note: this commit depends on https://github.com/apache/tvm/pull/16921 so also contains the contents of https://github.com/apache/tvm/pull/16921.

PR #17040, opened 2024-05-29
[TOPI] Fix SME conv2d schedule import and intrin argument

lhutton1:
Fixes a merge conflict between #16981 and #17003.

Closes https://github.com/apache/tvm/pull/17038

cc @ekalda @Anndrey24


Lunderberg (approved):
Thank you for the fix!

Looking at the breakage, it looks like it was caused by the addition of a required function parameter in https://github.com/apache/tvm/pull/16981, along with the addition of new calls to that function in https://github.com/apache/tvm/pull/17003.  This breakage could have been avoided if the new function parameter had been optional instead of required.

Unless there is a significant reason not to do so, I think new function parameters should be optional, with a default value that reproduces the previous behavior.  Doing so would avoid this type of breakage in the future, and would also improve backwards compatibility with external callers, when the function is part of a public-facing API

PR #17046, opened 2024-05-30
[Arith][SVE] Add rewrite rules for indices split by scalable expressions

Anndrey24:
This commit introduces rewrite rules for indices which can arise from splitting axes by scalable factors (e.g. `xo, xi = sch.split(x, factors = [None, 8 * T.vscale()])`):

```
(v_x_o * T.Cast("int64", T.vscale()) * T.int64(8) + v_x_i) // (T.Cast("int64", T.vscale()) * T.int64(8)) == v_x_o
(v_x_o * T.Cast("int64", T.vscale()) * T.int64(8) + v_x_i) % (T.Cast("int64", T.vscale()) * T.int64(8)) == v_x_i
```

The rewrites help prove checks needed by `sch.tensorize()` (e.g. CompareBufferRegion).

cc @ekalda @lhutton1

### DIFF
--- a/src/arith/rewrite_simplify.cc
+++ b/src/arith/rewrite_simplify.cc
@@ -1136,8 +1136,15 @@ PrimExpr RewriteSimplifier::Impl::VisitExpr_(const FloorDivNode* op) {
                        x + floordiv(y, z), CanProveGreaterEqual(z.Eval(), 0));
     TVM_TRY_REWRITE_IF(matches_one_of(floordiv(y + x * z, z), floordiv(y + z * x, z)),
                        floordiv(y, z) + x, CanProveGreaterEqual(z.Eval(), 0));
+    TVM_TRY_REWRITE_IF(floordiv(x * z * c1 + y, z * c1), x + floordiv(y, z * c1),
+                       CanProveGreaterEqual(z.Eval() * c1.Eval(), 0));
 
     TVM_TRY_REWRITE_IF(floordiv(x - floormod(x, c1), c1), floordiv(x, c1), c1.Eval()->value != 0);
+
+    // Scalable divisor
+    TVM_TRY_REWRITE_IF(floordiv(x, y), ZeroWithTypeLike(x),
+                       ContainsVscaleCall(y.Eval()) && CanProveGreaterEqual(x.Eval(), 0) &&
+                           CanProveGreaterEqual(y.Eval(), 0) && CanProve(x.Eval() < y.Eval()));
   }
   return ret;
 }
@@ -1229,6 +1236,14 @@ PrimExpr RewriteSimplifier::Impl::VisitExpr_(const FloorModNode* op) {
         matches_one_of(floormod(x - floormod(x, z), y), floormod(floormod(x, z) - x, y)),
         ZeroWithTypeLike(x),
         CanProveEqual(y.Eval() - z.Eval(), 0) || CanProveEqual(y.Eval() + z.Eval(), 0));
+
+    TVM_TRY_REWRITE_IF(floormod(x * z * c1 + y, z * c1), floormod(y, z * c1),
+                       CanProveGreaterEqual(z.Eval() * c1.Eval(), 0));
+
+    // Scalable divisor
+    TVM_TRY_REWRITE_IF(floormod(x, y), x,
+                       ContainsVscaleCall(y.Eval()) && CanProveGreaterEqual(x.Eval(), 0) &&
+                           CanProveGreaterEqual(y.Eval(), 0) && CanProve(x.Eval() < y.Eval()));
 
     if (floormod(x, c1).Match(ret)) {
       int64_t c1val = c1.Eval()->value;
--- a/src/arith/rewrite_simplify.h
+++ b/src/arith/rewrite_simplify.h
@@ -229,6 +229,8 @@ class RewriteSimplifier::Impl : public IRMutatorWithAnalyzer {
     // TODO(tqchen) refer back to super-analyzer.
     return TryCompare(x, val) == CompareResult::kEQ;
   }
+  // Whether x is true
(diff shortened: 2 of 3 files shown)


src/arith/rewrite_simplify.cc
```diff
@@ -1136,8 +1136,15 @@ PrimExpr RewriteSimplifier::Impl::VisitExpr_(const FloorDivNode* op) {
...
                        floordiv(y, z) + x, CanProveGreaterEqual(z.Eval(), 0));
+    TVM_TRY_REWRITE_IF(floordiv(x * z * c1 + y, z * c1), x + floordiv(y, z * c1),
+                       CanProveGreaterEqual(z.Eval() * c1.Eval(), 0));
 
     TVM_TRY_REWRITE_IF(floordiv(x - floormod(x, c1), c1), floordiv(x, c1), c1.Eval()->value != 0);
+
+    // Scalable divisor
+    TVM_TRY_REWRITE_IF(floordiv(x, y), ZeroWithTypeLike(x),
+                       ContainsVscaleCall(y.Eval()) && CanProveGreaterEqual(x.Eval(), 0) &&
+                           CanProveGreaterEqual(y.Eval(), 0) && CanProve(x.Eval() < y.Eval()));
```
ekalda, line 1147: Nit: `CanProveGreaterEqual(y.Eval(), 0)` seems redundant here (but I don't oppose keeping it for the sake of clarity)

src/arith/rewrite_simplify.cc
```diff
@@ -1136,8 +1136,15 @@ PrimExpr RewriteSimplifier::Impl::VisitExpr_(const FloorDivNode* op) {
                        x + floordiv(y, z), CanProveGreaterEqual(z.Eval(), 0));
     TVM_TRY_REWRITE_IF(matches_one_of(floordiv(y + x * z, z), floordiv(y + z * x, z)),
                        floordiv(y, z) + x, CanProveGreaterEqual(z.Eval(), 0));
+    TVM_TRY_REWRITE_IF(floordiv(x * z * c1 + y, z * c1), x + floordiv(y, z * c1),
+                       CanProveGreaterEqual(z.Eval() * c1.Eval(), 0));
```
ekalda, line 1140: If `z * c1 == 0` we get a division by zero? I hope it would not be allowed in constructing the `floordiv`, but maybe we shouldn't explicitly allow it in rewrite rules.

src/arith/rewrite_simplify.cc
```diff
@@ -1230,6 +1237,14 @@ PrimExpr RewriteSimplifier::Impl::VisitExpr_(const FloorModNode* op) {
         ZeroWithTypeLike(x),
         CanProveEqual(y.Eval() - z.Eval(), 0) || CanProveEqual(y.Eval() + z.Eval(), 0));
 
+    TVM_TRY_REWRITE_IF(floormod(x * z * c1 + y, z * c1), floormod(y, z * c1),
```
ekalda, line 1240: Same as before, unless I'm missing something, we're allowing modulo 0 here.

src/arith/rewrite_simplify.cc
Anndrey24: Oh right, I didn't consider that `x >= 0` and `x < y` would already necessitate that `y >= 0`. I wouldn't mind removing it, but even apart from helping with readability like you mentioned, it might save the analyzer calling the `CanProve()` function when `y < 0` could be proven with a quick `const_int_bound` check in `CanProveGreaterEqual()`, although it would just be redundant in the `y >= 0` case.  
I'm alright either way too, to be honest.

src/arith/rewrite_simplify.cc
Anndrey24: Yes, I wasn't sure what to do about it, especially since the earlier rewrite rules don't seem to address this case either. 
I've tried out a few more testcases and at least setting the constant `c1` to 0 does make it fail during `TestCase(fld(x * z * 0 + y, z * 0), x + fld(y, z * 0))` with `InternalError: Check failed: pb->value != 0 (0 vs. 0) : Divide by zero` in `tvm::arith::TryConstFold`, but this similar case doesn't fail although it preferably should: `TestCase(fld(x * z * 2 + y, z * 2), x + fld(y, z * 2), [z >= 0, z <= 0])`.

Since there is no `CanProveGreater()` function, I think we could either:

1. use `CanProveLess(-1 * z.Eval() * c1.Eval(), 0)` which uses the same simple `const_int_bound` check but looks a bit awkward
2. use `CanProve(z.Eval() * c1.Eval() > 0)` which is clear but _might_ be a costlier check
3. write a new `CanProveGreater()` function in `analyzer.cc`

ekalda: I'd lean towards option 3, it's pretty straightforward to implement (it can even just be a wrapper around 1). WDYT? 

To me it is a bit arbitrary that `CanProveLess` and `CanProveGreaterEqual` don't have same powers as `CanProve` and `CanProveEqual`, but I suppose the tradeoffs there are a topic of another discussion :)

src/arith/rewrite_simplify.cc
ekalda: Let's keep it :)

lhutton1:
cc @Lunderberg

src/arith/rewrite_simplify.cc
Anndrey24: I went ahead and implemented the `CanProveGreater()` function and took the opportunity to change all other cases of possible division by zero that I could find in the `rewrite_simplify.cc` rules to use it too.

ekalda:
We decided to roll back to the initial version of this patch since a large number of Relax tests fail when division by zero is disabled in the rewrite rules (so dealing with this is out of scope for this patch).

ekalda:
Thanks @Anndrey24!

PR #17048, opened 2024-05-30
[SME][TOPI] Add conv2d NHWC SME fp16->fp32 schedule

Anndrey24:
This commit extends the SME conv2d NHWC schedule to support convolutions with float16 inputs (data and kernel) and a float32 output using the tensor intrinsics added in #16981.

cc @ekalda @lhutton1

PR #17054, opened 2024-05-31
[SME] Utilize predication in fp32 matmul and conv2d schedules

lhutton1:
Prior to this commit, the matmul and conv2d schedules required padding of the inputs to some multiple of vscale and a final "unpadding" stage.

Instead, we can leverage predicated operations to avoid the the requirement for padding. Both the transpose interleave and outer product fp32 intrinsics are updated to use predication. The `get_active_lane_mask` intrinsic is utilized to generate a variably sized mask of active lanes depending on the global position the tensor intrinsic is operating on.

For now this relies on using `offset_of` and `stride` information from the tensor we're predicating an access on. Likely we will want to build on this in the future with a more intuitive API for determining the current tile location.

Support for batched conv2d was removed since this causes numerical issues which is suspected to be due to how the current tile is determined (paragraph above).

~Note: this should not be merged until after https://github.com/apache/tvm/pull/17048~

cc @ekalda @Anndrey24

### DIFF
--- a/python/tvm/relay/op/strategy/arm_cpu.py
+++ b/python/tvm/relay/op/strategy/arm_cpu.py
@@ -110,6 +110,8 @@ def conv2d_strategy_arm_cpu(attrs, inputs, out_type, target):
     """conv2d arm cpu strategy"""
     strategy = _op.OpStrategy()
     data, kernel = inputs
+    data_shape = data.shape
+    kernel_shape = kernel.shape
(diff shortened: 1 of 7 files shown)


Anndrey24:
LGTM, too! Seems there's just a small merge conflict that has come up.

ekalda:
Thanks @lhutton1 and @Anndrey24!

PR #17060, opened 2024-06-03
Build jetson runner

cgerum:


leandron (changes requested):
Hi @cgerum, thanks for the PR.

I see this has 250 different commits, spanning many areas of the project, which makes it quite hard to review. Can this be split into intelligible changes so that we can review them in isolation, as well as making in traceable for the future? Thanks.

PR #17075, opened 2024-06-10
[Transform] Modify FuseTIR pass to propagate buffer attributes

quic-sanirudh:
Arguments of a fused TIR PrimFunc generated from a fused relax function do not retain all the buffer attributes from their original PrimFuncs as the buffers are created from the StructInfo of the Relax vars. This patch collects a mapping of relax vars to its corresponding TIR buffers in a fused relax function and uses that info to propagate its buffer attributes such as `axis_separators` and `storage_scope`


quic-sanirudh:
cc @Hzfengsy @Lunderberg @slyubomirsky

src/relax/transform/fuse_tir.cc
```diff
@@ -362,6 +362,99 @@ class BlockNameDeduplicator : public tir::StmtMutator {
...
+        CollectVarMapping(call, lhs_var);
+      }
+    }
+  }
+
+ private:
+  /*! \brief The IRModule */
+  const IRModule& mod_;
+  // size_t call_num_inputs_ = -1;
+  Map<Var, tir::Buffer> relax_to_tir_var_map_;
```
Lunderberg, line 455: This data structure assumes that there is a 1:1 mapping from `relax::Var` to `tir::Buffer` across the entire fused function.  This would have incorrect results for cases where the same tensor is used as multiple arguments (e.g. `R.add(A, A)`), or where the same tensor is used as an argument to more than one function (e.g. The tensor `A` corresponds to two different TIR buffers in the sequence `mean = R.mean(A); norm = R.sqrt(mean); A_norm = R.divide(A, norm)`).

src/relax/transform/fuse_tir.cc
```diff
@@ -362,6 +362,99 @@ class BlockNameDeduplicator : public tir::StmtMutator {
...
+      ret.push_back(Integer(last_idx));
+      last_idx++;
+    }
+  }
+
+  return ret;
+}
+
+class RelaxToTIRVarMapCollector : public ExprVisitor {
+  void CollectVarMapping(const CallNode* call, const Expr& lhs_var, bool in_place = false) {
```
Lunderberg, line 383: Nitpick: For readability, having the public-facing interface at the top of the class makes it easier to find the entry point.  Can the `static Map<Var, tir::Buffer> Collect` function be moved to the top of `RelaxToTIRVarMapCollector`?

src/relax/transform/fuse_tir.cc
```diff
@@ -362,6 +362,99 @@ class BlockNameDeduplicator : public tir::StmtMutator {
...
+    } else {
+      for (size_t i = num_inputs; i < num_inputs + num_outputs; i++) {
+        output_idxs.push_back(i);
+      }
+    }
+    for (size_t i = 0; i < tir_args.size(); ++i) {
+      const auto& tir_var = Downcast<tir::Var>(tir_args[i]);
+      if (i < num_inputs) {
+        const auto& relax_var = Downcast<Var>(relax_args[i]);
+        relax_to_tir_var_map_.Set(relax_var, buffer_map[tir_var]);
```
Lunderberg, line 416: The `buffer_map` does not necessarily contain an entry for `tir_var`.  For example, the `relax_var` could have `PrimStructInfo` to pass a primitive scalar to the TIR funciton.  Even if `relax_var` has `TensorStructInfo`, the TIR function may treat the `DLTensor*` as an opaque pointer, passing it to a `PackedFunc` without having an entry in the `buffer_map`.

The best way to handle these cases is to wrap this line in a `if(auto tir_buffer = buffer_map.Get(tir_var))` conditional, and then use `tir_buffer.value()` inside the conditional instead of `buffer_map[tir_var]`.

src/relax/transform/fuse_tir.cc
```diff
@@ -362,6 +362,99 @@ class BlockNameDeduplicator : public tir::StmtMutator {
...
+      const auto* attrs = call->attrs.as<CallTIRInplaceAttrs>();
+      CHECK(attrs) << "Must have CallTIRInplaceAttrs for an in-place call";
+      output_idxs = GetInplaceOutputIndices(attrs->inplace_indices, num_inputs);
+    } else {
+      for (size_t i = num_inputs; i < num_inputs + num_outputs; i++) {
+        output_idxs.push_back(i);
+      }
+    }
+    for (size_t i = 0; i < tir_args.size(); ++i) {
+      const auto& tir_var = Downcast<tir::Var>(tir_args[i]);
```
Lunderberg, line 413: This `Downcast<tir::Var>` is unnecessary, because `prim_func->params` is already an array of `tir::Var`.

src/relax/transform/fuse_tir.cc
```diff
@@ -362,6 +362,99 @@ class BlockNameDeduplicator : public tir::StmtMutator {
...
+      if (i < num_inputs) {
+        const auto& relax_var = Downcast<Var>(relax_args[i]);
```
Lunderberg, line 415: This `Downcast<Var>` is not guaranteed to work.  While the normalizer will pull most `relax.Var` instances out to their own variable binding, `R.const` arguments may still appear inline.

src/relax/transform/fuse_tir.cc
```diff
@@ -362,6 +362,99 @@ class BlockNameDeduplicator : public tir::StmtMutator {
...
+
+ public:
+  explicit RelaxToTIRVarMapCollector(const IRModule& mod) : mod_(mod) {}
+  static Map<Var, tir::Buffer> Collect(const IRModule& mod, const Function& func) {
+    RelaxToTIRVarMapCollector visitor(mod);
+    visitor(func->body);
+    return visitor.relax_to_tir_var_map_;
+  }
+  void VisitBinding_(const VarBindingNode* binding) final {
+    const auto& lhs_var = binding->var;
```
Lunderberg, line 434: As written, this would only collect the mapping for relax variables whose binding occurs in the outermost `SeqExpr`.  Nested `SeqExpr` may occur if `binding->value` is a `relax::If` node, where each branch then contains a `SeqExpr`.

To resolve this, I'd recommend either adding `ExprVisitor::VisitBinding_(binding);` or `VisitExpr(binding->value)` to this method.

src/relax/transform/fuse_tir.cc
```diff
@@ -391,10 +484,15 @@ class FusedTIRConstructor : public ExprVisitor {
       : mod_(mod), func_name_(func_name) {}
 
   void VisitExpr_(const FunctionNode* func) final {
+    auto relax_to_tir_var_map = RelaxToTIRVarMapCollector::Collect(mod_, GetRef<Function>(func));
     std::vector<Variant<tir::Var, tir::Buffer>> prim_func_params;
     for (const Var& relax_param : func->params) {
       size_t size_before = prim_func_params.size();
-      CollectPrimFuncParams(relax_param, &prim_func_params);
+      if (relax_to_tir_var_map.count(relax_param)) {
```
Lunderberg, line 491: Nitpick: Instead of having the conditional, both branches could be written as `CollectPrimFuncParams(relax_param, &prim_func_params, relax_to_tir_var_map.Get(relax_param))`.  The `Map::Get` method returns an `Optional<tir::Buffer>`, which is `NullOpt` when the key is absent from the map.

tests/python/relax/test_transform_fuse_tir.py
```diff
@@ -2314,5 +2314,88 @@ def take(
...
+        def fused_function(
+            x: R.Tensor([T.int64(16), T.int64(32)], "float32"),
+            y: R.Tensor([T.int64(16), T.int64(32)], "float32"),
+            z: R.Tensor([T.int64(16), T.int64(32)], "float32"),
+        ) -> R.Tensor([T.int64(16), T.int64(32)], dtype="float32"):
+            R.func_attr({"Primitive": 1})
+            cls = Before
+            with R.dataflow():
+                w = R.call_tir(
+                    cls.add, [x, y], out_sinfo=R.Tensor([T.int64(16), T.int64(32)], "float32")
```
Lunderberg, line 2341: Can we add a test case for incompatible usage of a single Relax var?  As currently written, we could have a single Relax variable that is used in two separate `R.call_tir` statements, where the function being called imposes different restrictions on it.  For example, if `x` were used in `cls.add1`, which requires `axis_separators=[1]`, and `cls.add2`, which requires `axis_separators=[]`.  We should be able to identify this case and raise an error when it occurs.

(Ideally, that should never happen, but this would be the last point at which we'd have enough information to catch this failure mode at compile-time.)

Lunderberg (changes requested):
I like the fix overall, thank you!  It looks like there's some implicit assumptions about the Relax/TIR being consumed, which may not hold in general.

src/relax/transform/fuse_tir.cc
```diff
@@ -362,6 +362,99 @@ class BlockNameDeduplicator : public tir::StmtMutator {
...
+static Array<Integer> GetInplaceOutputIndices(const Array<Integer>& inplace_indices,
+                                              int num_inputs) {
+  Array<Integer> ret;
+  int last_idx = num_inputs;
+  for (auto idx : inplace_indices) {
+    int i = idx.IntValue();
+    if (i >= 0) {
+      ret.push_back(Integer(i));
+    } else {
+      ret.push_back(Integer(last_idx));
```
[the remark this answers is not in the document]
quic-sanirudh: Done

src/relax/transform/fuse_tir.cc
quic-sanirudh: Yes, I did think about this issue, but I assumed that even though the same relax var might map to different buffers, it should have the same buffer attributes (since its source is the same relax var). I've also added a validation ICHECK to verify that the buffer attributes match (using structural equal).

I've also added a testcase to verify this use case as suggested in the below comment.

src/relax/transform/fuse_tir.cc
quic-sanirudh: Done

src/relax/transform/fuse_tir.cc
quic-sanirudh: Thanks for the catch. I did not consider this case. Fixed.

src/relax/transform/fuse_tir.cc
quic-sanirudh: Done, Thanks

src/relax/transform/fuse_tir.cc
quic-sanirudh: Thanks, yes constants are possible. I've updated output as a map from `Expr` to `Buffer` instead of `Var`.

src/relax/transform/fuse_tir.cc
quic-sanirudh: I've added the visit, thanks.

src/relax/transform/fuse_tir.cc
quic-sanirudh: Yes, that's a much better solution.

tests/python/relax/test_transform_fuse_tir.py
quic-sanirudh: I've added the test case to check possible inconsistencies.

quic-sanirudh:
@Lunderberg I think I've addressed your comments. When you get a chance, could you please take a look?

Lunderberg (approved):
Thank you for making the changes, and looks good!

quic-sanirudh:
Thanks for taking the time to review and provide feedback.

onedayherocoming:
I found that this MR would change `finfo->IsOpaque` result in struct_info_analysys.cc, which would cause error.

PR #17076, opened 2024-06-10
[SME] Extract gemm block correctly when fused with bias

lhutton1:
Prior to this commit, the scheduling assumed the gemm block would be the second to last block in the function ("unpadding" step is the final block). However, when dense is fused with a bias or activation the gemm block is no longer the second to last block. This commit instead searches a single reduction block to use as the gemm block.

cc @ekalda @Anndrey24


ekalda (commented):
Looks good to me in general, the only thing I was wondering is that the commit message says that this fix addresses case where there is fused bias or activation (or I suppose both?) but the testing only covers bias case? I suppose the TIR looks similar for all of these cases though, so I won't block the patch for that.

ekalda (approved):
LGTM, thanks @lhutton1!

PR #17077, opened 2024-06-10
[AOT] Correctly calculate workspace for vector types

lhutton1:
When calculating the size of the workspace for a given prim func, the lanes of the data type was not being considered, meaning sizes calculated for dtypes such as "float32x4" were smaller than what they should be. This commit also considers lanes in the calculation.


ekalda (approved):
Thanks @lhutton1 LGTM!

PR #16637, opened 2024-02-23
[Bugfix][Transform] Preserve symbolic variables in FuseOps

Lunderberg:
Prior to this commit, the `CompositeFunctionAnnotator` visited the body of functions without the parameters being considered in-scope. As a result, `EraseToWellDefined` would remove known shapes from the function body's `StructInfo`.

PR #17080, opened 2024-06-10
[Relax] Implement relax.transform.RemoveSymbolicExpressionsInSubroutine

Lunderberg:
This is a follow-up commit to
https://github.com/apache/tvm/pull/16637, which updated `relax.transform.FuseOps` to provide additional parameters defining symbolic variables required by the fused functions.  While this ensures that `relax.transform.FuseOps` produces well-formed Relax functions, these additional arguments can break some kernel implementations.

This commit implements a new transform
`RemoveSymbolicExpressionsInSubroutine` to resolve this issue.  This transform identifies function arguments whose sole purpose is to compute a symbolic expression, when that symbolic expression could be inferred from tensor shapes.

For example, consider the following Relax function:

```python
@R.function
def func(
    data: R.Tensor(["batch_size * seq_len", "hidden_size"]),
    weights: R.Tensor(["hidden_size", "intermediate_size"]),
    dummy_arg: R.Shape(["batch_size", "seq_len"]),
  ) -> R.Tensor(["batch_size * seq_len", "intermediate_size"]):

    batch_size = T.int64()
    seq_len = T.int64()
    intermediate_size = T.int64()
    hidden_size = T.int64()

    output: R.Tensor([batch_size * seq_len, intermediate_size]) = R.matmul(data, weights)
    return output
```

The `data` tensor may be used to infer `hidden_size`, but cannot be used to infer `batch_size` or `seq_len`.  The `R.Shape` parameter exists solely to define `batch_size` and `seq_len`, since all symbolic variables must be defined.  However, neither `batch_size` nor `seq_len` are ever used outside of the expression `batch_size * seq_len`, and the value of `batch_size * seq_len` could be inferred from the shape of the `data` tensor.

This new transform identifies cases where an argument is otherwise unnecessary, and replaces the symbolic expression with a new argument.  This makes the `dummy_arg: R.Shape` be entirely unused, so a later use of `relax.transform.RemoveUnusedParameters()` can remove the parameter altogether.

```python
@R.function
def func(
    data: R.Tensor(["data_dim0", "hidden_size"]),
    weights: R.Tensor(["hidden_size", "intermediate_size"]),
    dummy_arg: R.Shape(["batch_size", "seq_len"]),
  ):

    data_dim0 = T.int64()
    intermediate_size = T.int64()
    hidden_size = T.int64()

    output: R.Tensor([data_dim0, intermediate_size]) = R.matmul(data, weights)
    return output
```

### DIFF
--- a/include/tvm/relax/transform.h
+++ b/include/tvm/relax/transform.h
@@ -307,6 +307,23 @@ TVM_DLL Pass RemoveUnusedParameters();
  */
 TVM_DLL Pass RemoveUnusedOutputs();
 
+/*! \brief Remove unnecessary symbolic expressions in subroutines
+ *
+ * If all occurrences of a symbolic variable within a subroutine
+ * occur within the same symbolic expression, then the subroutine
+ * could be simplified to be in terms of that expression.
+ *
+ * For example, if a subroutine accepts symbolic shape parameters `N`
+ * and `M`, and the variables `N` and `M` are only ever used to
+ * compute `N*M`, then the subroutine could instead accept a symbolic
+ * shape parameter `new_var = N*M`.  This can allow shape parameters
+ * to be inferred from tensor shapes, rather than requiring additional
+ * arguments.
+ *
+ * \return The pass
+ */
+TVM_DLL Pass RemoveSymbolicExpressionInSubroutine();
+
 /*!
  * \brief Annotate Op Pattern Kind for TIR functions, which is used in FuseOps.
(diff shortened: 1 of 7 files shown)


Lunderberg:
This transform is intended to be used in the implementation of https://github.com/apache/tvm/pull/16450, as recommended [here](https://github.com/apache/tvm/pull/16450#issuecomment-1904554188).

PR #17086, opened 2024-06-12
[Relax] Ignore dynamic parameters in RewriteDataflowReshape

Lunderberg:
The Relax transform `RewriteDataflowReshape` identifies TIR functions that are equivalent to `relax.op.reshape`, and replaces them with calls to `relax.op.reshape`.  This is used as a precursor for simplifications that rely on the high-level knowledge that an operator is a reshape, but also require the low-level knowledge of the adjacent TIR PrimFuncs.

Prior to this commit, the `RewriteDataflowReshape` pass would only recognize static shapes, or dynamic shapes that could be inferred from the shapes of tensor arguments.  This commit updates `RewriteDataflowReshape` to recognize cases where an extra symbolic variable has been provided.

### DIFF
--- a/src/relax/analysis/tir_op_pattern_kind.cc
+++ b/src/relax/analysis/tir_op_pattern_kind.cc
@@ -517,19 +517,23 @@ bool HasReshapePattern(const PrimFunc& func) {
     arith::Analyzer ana_;
   };
 
-  if (func->params.size() < 2) {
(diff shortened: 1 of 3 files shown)


masahi:
cc @MasterJH5574

MasterJH5574 (approved):

PR #17094, opened 2024-06-14
[Utility][Container] Support non-nullable types in Array::Map

Lunderberg:
Prior to this commit, the `Array::Map` member function could only be applied to nullable object types.  This was due to the internal use of `U()` as the default value for initializing the output `ArrayNode`, where `U` is the return type of the mapping function.  This default constructor is only available for nullable types, and would result in a compile-time failure for non-nullable types.

This commit replaces `U()` with `ObjectRef()` in `Array::Map`, removing this limitation.  Since all items in the output array are overwritten before returning to the calling scope, initializing the output array with `ObjectRef()` does not violate type safety.


Lunderberg:

masahi (approved):

PR #17095, opened 2024-06-14
[Cleanup] Accept Variant<...> instead of ObjectRef when possible

Lunderberg:
Prior to the implementation of `Variant<...>` in https://github.com/apache/tvm/pull/15672, functions that were polymorphic over an argument type would typically accept an `ObjectRef` argument, then downcast to an allowed type.  This delays the catching of an error, and can accidentally omit automatic conversions applied by the FFI.

This commit updates several locations using this pattern to instead accept a `Variant`, templated over the allowed types.  This enables C++ type checking for C++ callers, standardizes the type-checking in the FFI for non-C++ callers, and ensures that FFI type conversions are uniformly applied.

### DIFF
--- a/include/tvm/runtime/container/variant.h
+++ b/include/tvm/runtime/container/variant.h
@@ -82,7 +82,7 @@ class Variant : public ObjectRef {
  public:
   /* \brief Helper utility to check if the type is part of the variant */
   template <typename T>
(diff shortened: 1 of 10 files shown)


Lunderberg:

Lunderberg:
Closed, as this was a subset of #16183, which has merged.

PR #17096, opened 2024-06-14
[Bugfix] Update FAttrsGetter to return Map<String, ObjectRef>

Lunderberg:
Prior to this commit, `FAttrsGetter` was defined as a function that returned `Map<String, String>`.  However, it is used to define attributes in a `Map<String, ObjectRef>`, and in some cases is used to define attributes whose value is a dictionary (e.g. `msc_attrs_getter` in `python/tvm/contrib/msc/core/transform/pattern.py`).

This commit updates the type signature of `FAttrsGetter` to match its usage, returning a `Map<String, ObjectRef>`.

### DIFF
--- a/include/tvm/relax/transform.h
+++ b/include/tvm/relax/transform.h
@@ -367,7 +367,7 @@ class FusionPatternNode : public Object {
    * \brief The function to get attributes for fused function
    *
    * It should have signature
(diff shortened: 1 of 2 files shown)


Lunderberg:

yongwww (approved):

PR #17097, opened 2024-06-14
[Bugfix][CRT] Return error code on error from ModuleGetFunction

Lunderberg:
Prior to this commit, `ModuleGetFunction` returned zero if called with an incorrect number of arguments, or with incorrect type codes.  This incorrectly indicated that the module was inspected, and did not contain the requested function.

This commit corrects the implementation of `ModuleGetFunction` to instead call set an error message with `TVMAPISetLastError`, then to return an appropriate error code.


Lunderberg:

yongwww (approved):
lgtm

PR #17103, opened 2024-06-17
[TIR][FFI] Allow string in PrimFunc return statement

Lunderberg:
Prior to this PR, a TIR PrimFunc could return an `int64`, `float64`, or void.  Returning any other type, even if supported by TVM's FFI, would raise an exception during `MakePackedAPI`.  With this PR, string literals can be returned from TIR PrimFuncs.

Enabling this TIR functionality requires a change to the TVM FFI.  Previously, a `TVMRetValue` holding a string would allocate a
`std::string*` and store it to `value_.v_handle`.  This return value depends on the C++ runtime, and cannot be constructed on all backends that only require the C API.  With this commit, this is instead represented as a `const char*` in `value_.v_str`, matching the representation in `TVMArgValue`.  Unlike `TVMArgValue`, since p`TVMRetValue` may own its return value, two new member variables `void (*f_deleter_)(void*)` and `void* f_deleter_arg_`.  These are designed to be easy to call from a C API, and are set to a deleter function that will free the allocation (if required), and an argument to be passed into the deleter function.

With this new representation, the return value for different use cases would be set as follows:

* Return string from C++ PackedFunc.  The backing allocation is made using `new std::string(std::move(value))`.  This heap allocation is stored in `f_deleter_arg_`, so that `f_deleter_` can then call `delete static_cast<std::string*>(arg)`.  Finally, `new_str->data()` is stored as `value.v_str`.

* Return pointer to static string from C.  A pointer to the static string is stored in `v_str`.  Both `f_deleter_` and `f_deleter_arg_`are left with their default value of `nullptr`, indicating that the `TVMRetValue` does not need to free any memory on destruction.

  This is the return value used by `T.ret(T.StringImm("my_string"))` in a TIR PrimFunc.

* Return pointer to heap allocation made in C.  Assuming the allocation is made using `malloc()`, a pointer to the heap allocation is stored in both `value_.v_str` and `f_deleter_arg_`. The `f_deleter_` is set to `free`, indicating that `free(f_deleter_arg_)` should be called when the `TVMRetValue` is destructed.

  This functionality is possibly within the updated `TVMRetValue` API, but is not currently used.

This commit adds unit tests for the return of strings from compiled TIR PrimFuncs, when either running the function locally, or over a RPC connection.

### DIFF
--- a/include/tvm/runtime/packed_func.h
+++ b/include/tvm/runtime/packed_func.h
@@ -27,6 +27,7 @@
 #include <tvm/runtime/c_runtime_api.h>
 #include <tvm/runtime/container/array.h>
 #include <tvm/runtime/container/map.h>
+#include <tvm/runtime/container/string.h>
 #include <tvm/runtime/container/variant.h>
 #include <tvm/runtime/data_type.h>
 #include <tvm/runtime/logging.h>
@@ -585,31 +586,6 @@ class TVMPODValue_ {
     TVM_CHECK_TYPE_CODE(type_code_, kTVMOpaqueHandle);
     return value_.v_handle;
   }
-  operator std::string() const {
-    if (type_code_ == kTVMDataType) {
-      return DLDataType2String(operator DLDataType());
-    } else if (type_code_ == kTVMBytes) {
-      TVMByteArray* arr = static_cast<TVMByteArray*>(value_.v_handle);
-      return std::string(arr->data, arr->size);
-    } else if (type_code_ == kTVMStr) {
-      return std::string(value_.v_str);
-    } else {
-      return AsObjectRef<tvm::runtime::String>().operator std::string();
-    }
-  }
-  operator TVMByteArray() const {
-    if (type_code_ == kTVMBytes) {
-      return *static_cast<TVMByteArray*>(value_.v_handle);
-    } else {
-      LOG(FATAL) << "Expected "
-                 << "TVMByteArray but got " << ArgTypeCode2Str(type_code_);
-    }
-  }
-
-  inline operator DLDataType() const;
-
-  operator DataType() const { return DataType(operator DLDataType()); }
-
   operator DLTensor*() const {
     if (type_code_ == kTVMDLTensorHandle || type_code_ == kTVMNDArrayHandle) {
       return static_cast<DLTensor*>(value_.v_handle);
@@ -664,16 +640,13 @@ class TVMPODValue_ {
   friend class TVMArgsSetter;
   friend class TVMRetValue;
   friend class TVMMovableArgValue_;
-  friend PackedFunc WrapPackedFunc(int (*)(TVMValue*, int*, int, TVMValue*, int*, void*),
-                                   const ObjectPtr<Object>&);
-
-  TVMPODValue_() : value_{0}, type_code_(kTVMNullptr) {}
+  TVMPODValue_() : type_code_(kTVMNullptr) {}
   TVMPODValue_(TVMValue value, int type_code) : value_(value), type_code_(type_code) {}
 
   /*! \brief The value */
-  TVMValue value_{0};
+  TVMValue value_;
   /*! \brief the type code */
-  int type_code_{kTVMNullptr};
+  int type_code_;
 };
 
 /*!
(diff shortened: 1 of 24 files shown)


Lunderberg:
This change is intended to allow compile-time information to be embedded into the compiled module, such as for use in https://github.com/apache/tvm/pull/16836.

This PR is marked as a draft because (1) it depends on https://github.com/apache/tvm/pull/17098, and (2) it most likely will conflict with https://github.com/apache/tvm/pull/16183 and I'd rather land https://github.com/apache/tvm/pull/16183 first.

include/tvm/runtime/packed_func.h
```diff
@@ -833,35 +859,17 @@ class TVMRetValue : public TVMPODValue_ {
...
   // Assign operators
   TVMRetValue& operator=(TVMRetValue&& other) {
-    this->Clear();
-    value_ = other.value_;
-    type_code_ = other.type_code_;
-    other.type_code_ = kTVMNullptr;
+    std::swap(value_, other.value_);
+    std::swap(type_code_, other.type_code_);
+    std::swap(f_deleter_, other.f_deleter_);
+    std::swap(f_deleter_arg_, other.f_deleter_arg_);
```
tqchen, line 872: I think we are doing a bit too much special casing for this case. Instead, a simpler approach that might work is to leverage the Object system, which might align with our future directions of better FFI surrounding the object system

tqchen (changes requested):

tqchen:
We need to be careful with FFI updates. As of now, TIR function mostly relies on destination passing, so not involving too much heap allocations. 

For the return value in generally, I think we would like to move toward a world where we rely on Object system for most cases, including the case of string return

Lunderberg:
Using the Object system requires a C++ compiler, and a C++ runtime.  The FFI supports targets that only have a C compiler and a C runtime.  The initial implementation in https://github.com/apache/tvm/pull/16836 used the Object system, and failure on C-only backends was caught by CI.  See [this comment](https://github.com/apache/tvm/pull/16836#issuecomment-2077686494) for details.

This commit does not introduce any heap allocations that were not already present, and in fact does the complete opposite.  It adds the ability to express a return value that doesn't require a heap allocation.

tqchen:
Object system is not necessarily tied to c++ from the ABI level (you can view the deleter etc in the layout), the main complexity here is that the return value container is adding more complexity that already is part of object system
 
Seems returning a string is more advanced feature that we still would want to be careful with, so maybe we don't have to strictly enable it in pure C backend.

Lunderberg:
> Object system is not necessarily tied to c++ from the ABI level 

The Object base class uses `std::atomic<int32_t>` to store the reference count.  This is an opaque type, and is not guaranteed to have an internal `int32_t`.  In C++23, the `_Atomic` macro was introduced to `stdatomic.h` which could make an equivalent declaration in C, but it is still not guaranteed to have an internal `int32_t`.  The ABI is tied to C++.

> Seems returning a string is more advanced feature that we still would want to be careful with, so maybe we don't have to strictly enable it in pure C backend.

Operations that are difficult to write for a C backend will also be difficult to write in the LLVM codegen.  The main difficulty is due to the current use of a C++ type `std::string` to store a returned string value.

> the main complexity here is that the return value container is adding more complexity that already is part of object system

I suppose we could strip out `TVMRetValue` entirely and return a `Object*` directly, but that seems like a far more wide-reaching change.  Since we already had the distinction between `TVMRetValue` (directly contains primitive types for ease of ABI) and `Object*`, I thought it best to preserve that distinction.

Lunderberg:
I will attempt an additional implementation, in which `kTVMStr` and `kTVMBytes` use `value_.v_handle` to store a `tvm::runtime::StringObj*`, with the nested `Object*` constructed from the compiled module.  This will run the risk of incompatibility between `int32_t` and `std::atomic<int32_t>`, but there's no way to avoid that undefined behavior when constructing it from C.

tqchen:
The  `std::atomic<int32_t>` part is not too hard, since we can use the C counterpart. Still think it is something we should think a bit, since returning string from low-level is indeed more advanced

Lunderberg:
And re-implemented so that the codegen produces a `tvm::runtime::StringObj` in order to return a `StringImm`.

PR #17108, opened 2024-06-18
[Relax] Apply DefaultGPUSchedule() in default build pipeline

Lunderberg:
This is a follow-up to https://github.com/apache/tvm/pull/15864, which added `LegalizeOps` to the default Relax build pipeline.  Since legalization may produce additional TIR PrimFuncs that require scheduling, the output of `LegalizeOps` typically must also be passed through `tir.transform.DefaultGPUSchedule()`.  This PR adds `DefaultGPUSchedule()` to the relax build pipeline to handle these cases.

Scheduled PrimFunc have the `"tir.is_scheduled"` attribute set to true, and are ignored by `DefaultGPUSchedule()`.  In addition, the `DefaultGPUSchedule` transform has no effect on non-GPU targets. Therefore, this change should only impact `tvm.relax.build` calls that previously resulted in an error due to unscheduled GPU functions, and should not have any impact on existing calls.

### DIFF
--- a/python/tvm/relax/pipeline.py
+++ b/python/tvm/relax/pipeline.py
@@ -60,6 +60,7 @@ def zero_pipeline(*, enable_warning: bool = False):
         seq = tvm.transform.Sequential(
             [
                 transform.LegalizeOps(enable_warning=enable_warning),
+                tvm.tir.transform.DefaultGPUSchedule(),
                 transform.AnnotateTIROpPattern(),
                 transform.FoldConstant(),
                 transform.FuseOps(),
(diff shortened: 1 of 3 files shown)


yongwww:
Seems we discussed this before in the phase ordering discussion in open dev meeting, can't remember the reason why we didn't add this pass in default pipeline, need to take a look at the meeting note

Lunderberg:
I don't remember that discussion, but I'd be interested in seeing the minutes from it.  The main roadblocks to it are ones that I've resolved in previous PRs:

* Accidentally scheduling a host function, fixed in https://github.com/apache/tvm/pull/16958
* Accidentally scheduling the output of `ComputePrimValue`, fixed in https://github.com/apache/tvm/pull/17032.
* Accidentally de-duplicating functions that differ by vdevice, only one of which should be scheduled for GPU, fixed in https://github.com/apache/tvm/pull/16959.

With those fixes applied, I think it's now a pretty big usability benefit to include it in the lowering pipeline.

MadFunMaker:
LGTM - I think it's clear usability gain without side effects and significant extra burden on build pipeline.

MadFunMaker (approved):

Lunderberg:
Rebased onto main to avoid having stale CI results.

tqchen:
moved toward https://github.com/apache/tvm/pull/17626

PR #17113, opened 2024-06-24
[CudaGraph] Handle exceptions thrown while capturing cuda graph

Lunderberg:
Prior to this commit, an exception thrown during the capture of a cuda graph would result in `std::terminate` being called.  This commit updates the implementation of `"vm.builtin.cuda_graph.run_or_capture"` such that a thrown exception can be recovered from, and does not cause any changes to the state of TVM's cuda graph cache.

- Call to `cudaStreamDestroy` was previously skipped, now moved to a RAII-style destructor in a `ScopedCUDAStream` class.

- Call to `cudaStreamEndCapture` was previously skipped, end of cuda graph capture now performed as part of RAII-style destructor for `CUDACaptureStream` class.

- Restoration of `CUDAThreadEntry::ThreadLocal()->stream` was previously skipped, now restored as part of RAII-style destructor for `CUDACaptureStream` class.

- Previously, an error raised from `cudaGraphInstantiate` would leave the `capture_cache_` in an ill-formed state.  Now, the `capture_cache_` is only updated after a valid `CUDAGraphCapturedState` has been fully constructed.

### DIFF
--- a/src/runtime/relax_vm/cuda/cuda_graph_builtin.cc
+++ b/src/runtime/relax_vm/cuda/cuda_graph_builtin.cc
@@ -32,6 +32,8 @@ namespace tvm {
 namespace runtime {
 namespace relax_vm {
 
+namespace {
+
 struct CUDAGraphCaptureKey {
   // The unique index of the capture function within the module
   int64_t index;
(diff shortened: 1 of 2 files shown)


src/runtime/relax_vm/cuda/cuda_graph_builtin.cc
```diff
@@ -82,6 +96,43 @@ struct CUDAGraphCapturedState {
...
+class CUDACaptureStream {
+ public:
+  explicit CUDACaptureStream(cudaGraph_t* graph)
+      : prev_default_stream_(CUDAThreadEntry::ThreadLocal()->stream), output_graph_(graph) {
+    CUDAThreadEntry::ThreadLocal()->stream = capture_stream_;
+
+    CUDA_CALL(cudaStreamBeginCapture(capture_stream_, cudaStreamCaptureModeGlobal));
+  }
+  ~CUDACaptureStream() {
+    cudaStreamEndCapture(capture_stream_, output_graph_);
```
vinx13, line 123: this should be enclosed with `CUDA_CALL` to check return code

Lunderberg: I tested it with `CUDA_CALL` to verify, and using the `CUDA_CALL` would re-introduce the same bug that this change is intended to fix.  If the stack is unwinding due to a thrown exception, then throwing another exception would result in `std::terminate` being called.  To avoid this, destructors shouldn't throw exceptions ([stackoverflow link](https://stackoverflow.com/a/130123)).

There are some ways to use `std::uncaught_exceptions` to determine whether an exception is being unwound, and to conditionally throw an exception if it isn't already the case.  However, those tend to be pretty context-dependent, and probably aren't worth using in this case.

vinx13 (approved):

PR #17128, opened 2024-07-01
[Compute-inline] Prefer T.where for reverse compute-inlined block with predicate

wrongtest-intellif:
Hi, we found a kind of corner bad cases for reverse inline primitive. Currently we encode predicate produced by inlining as block compute body, like
```python
with T.block():
    if pred:
        body()
```

While other primitives (typically `cache_write`) currently may only respect the predicate of `T.where` for region estimation. The newly created blocks take the risk of region out-of-bound write accesses, lead to segfaults even in cpu target computations.

The change proposes to use equavalent `T.where` when possible to make it more consistent with s-tir system, like
```python
with T.block():
    T.where(pred_on_loop_vars)
    body()
```


wrongtest-intellif:
@tvm-bot re-run

Hzfengsy (approved):

PR #17131, opened 2024-07-01
[TVMScript][Relax] Allow return statement in DataflowBlock

Lunderberg:
Prior to this commit, TVMScript required the return value of a Relax to be specified outside of any `with R.dataflow()` blocks.  This resulted in a common pattern, where the return value of a function was first called with `R.output(ret_value)`, to mark `ret_value` as a `tvm::relax::Var` instead of a `tvm::relax::DataflowVar`, followed immediately by a `return ret_value` statement.

This commit updates the TVMScript parser to allow a `return` statement inside a `with R.dataflow()` block.  This is syntactic sugar that is equivalent to calling `R.output`, followed by a `return`.

With this change, the following two TVMScript examples are now equivalent.  (Prior to this change, the `return_inside_dataflow` example would raise an error during parsing.)

```python
@R.function(private=True)
def output_then_return(A: R.Tensor):
    with R.dataflow():
        B = R.add(A, A)
        C = R.multiply(B, B)
        R.output(C)

    return C

@R.function(private=True)
def return_inside_dataflow(A: R.Tensor):
    with R.dataflow():
        B = R.add(A, A)
        C = R.multiply(B, B)
        return C
```

### DIFF
--- a/src/script/ir_builder/relax/frame.cc
+++ b/src/script/ir_builder/relax/frame.cc
@@ -118,36 +118,23 @@ void BlockFrameNode::EnterWithScope() {
   }
 }
 
-class DataflowBlockRewriter : public tvm::relax::ExprMutator {
+class VarReplacer : public tvm::relax::ExprMutator {
  public:
-  static tvm::relax::DataflowBlock Rewrite(const tvm::relax::DataflowBlock& block,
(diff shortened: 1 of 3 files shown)


tqchen (approved):

Lunderberg:
Rebased onto main to resolve conflict.

PR #17115, opened 2024-06-25
[Relax] Support `input_axis_separator` to allow 2D to 1D conversion

abhikran-quic:
- Introduce `input_axis_separator` in `relax.transform_layout` op for 2D input buffer to allow conversion of 2D buffers to 1D buffers. The conversion from 2D->1D is handled while lowering of `transform_layout` operator. 
- This is needed because once the tensor is converted into 2D via `axis_separator`, the `input_axis_separator` helps in identifying the buffers with 2 dimensions especially when the conversion is needed to 1D operation.
- Also introducing support for `input_axis_separator` in `AlterOpImpl` pass.

PR #17134, opened 2024-07-03
[TIR][Schedule] Remove `@type_check` for `set_axis_separator`

abhikran-quic:
- The decorator is not allowing types like `Array` to be passed to set_axis_separator directive. 
- The FFI has a type checking implemented internally making this redundant.
- The discussion for this change came up in a review comment in https://github.com/apache/tvm/pull/17115

### DIFF
--- a/python/tvm/relax/transform/legalize_ops/manipulate.py
+++ b/python/tvm/relax/transform/legalize_ops/manipulate.py
@@ -213,7 +213,6 @@ def _layout_transform(bb: BlockBuilder, call: Call) -> Expr:
     sch.transform_layout(primfunc_name, ("write", 0), index_map, pad_value)
     set_axis_sep(axis_separators, sch, "write")
     if input_axis_separators is not None:
-        input_axis_separators = [int(sep) for sep in input_axis_separators]
         set_axis_sep(input_axis_separators, sch, "read")
     gvar = bb.add_func(sch.mod["main"], primfunc_name)
     output_shape = index_map.map_shape(list(call_args[0].struct_info.shape))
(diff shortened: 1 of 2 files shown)


abhikran-quic:
@tvm-bot rerun

abhikran-quic:
cc: @Lunderberg 
I've tried to address a review comment in https://github.com/apache/tvm/pull/17115 where we discussed about `@type_check` in `set_axis_separator`.
Please help in reviewing this change. Thank you!

abhikran-quic:
> Thank you for the follow-up, and this change looks good! I like the consolidation of the type-checking into the FFI.

Thank you @Lunderberg for reviewing this change! 
I saw other scheduling directives using `type_check` . Do you recommend cleaning up the others as well ? If yes, I can volunteer to do it.

Lunderberg:
> I saw other scheduling directives using type_check . Do you recommend cleaning up the others as well ? If yes, I can volunteer to do it.

If they are being checked in the FFI, then that would be fantastic, thank you!  The main thing to watch out for is C++ functions that use `set_body_typed` and accept `ObjectRef`, or which use `set_body` to interact with the PackedFunc interface directly.  Neither of these forms apply a strict type-checking during the FFI, so the python-side checks might still be useful there.

abhikran-quic:
> > I saw other scheduling directives using type_check . Do you recommend cleaning up the others as well ? If yes, I can volunteer to do it.
> 
> If they are being checked in the FFI, then that would be fantastic, thank you! The main thing to watch out for is C++ functions that use `set_body_typed` and accept `ObjectRef`, or which use `set_body` to interact with the PackedFunc interface directly. Neither of these forms apply a strict type-checking during the FFI, so the python-side checks might still be useful there.

Sure. I will work on this and try to check how much more type checks can be moved to FFI in a new PR. Would it be okay to merge this PR ?
Thank you!

PR #17145, opened 2024-07-09
[Relax] Implement R.ensure_aligned and update memory planning for R.view

vinx13:
Previously, `R.view` was legalized to extern call to `runtime.TVMArrayCreateView` during `LegalizeOps`. This call to extern func can't be properly handled by `StaticBlockPlanMemory` because it assumes the extern func does not retain the input buffer. Extern func returning a view of the input would break the ref count of the buffer. This PR defers the legalization of `R.view` so that it can be explicitly handled by memory planning.

A new op `R.ensure_aligned` is added as discussed in #16955 

cc @tqchen @yongwww @Lunderberg


Lunderberg (changes requested):
Thank you for making this follow-up.  I have a couple of straight

It looks like the removal of the legalization for `R.memory.view` is to avoid a phase-order issue, where `StaticPlanBlockMemory` must be able to identify operators that may alias.  Is that understanding correct?

Rather than moving some of the legalization steps into `LowerVMBuiltin`, I propose we instead add a `legalization_level` for each operator, and to `LegalizeOps`.  That way, we can distinguish between higher-abstraction operators (legalize before `StaticPlanBlockMemory`) and lower-abstraction operators (legalize after `StaticPlanBlockMemory`).

* If not specified, an operator would have legalization level of 10.  The `R.memory.view` and `R.memory.ensure_aligned` operators would have legalization level of 0.
* `LegalizeOps` would default to a legalization level of 10.  Any operator whose legalization level is less than the `LegalizeOps` level would be skipped.
* An additional pass of `LegalizeOps` would occur at the end of the Relax lowering pipeline, with legalization level of zero.

Lunderberg:
Also, if you're interested, I have a partial implementation in [this dev branch](https://github.com/apache/tvm/compare/main...Lunderberg:temp_relax_implement_ensure_aligned_operator) that includes the device-type validation and a TIR legalization.  If you'd like to pull any of it over, you're welcome to it, as I've had it on the back-burner for far too long.

tqchen:
Based on the current grouping, seems quite a bit of the runtime function dispatchings happens in LowerBuiltin, while legalizeOps primarily focused on lowering to TIR related functions. 

I think such distinction is still helpful, so that can be a factor considering moving the view legalization into the VMBuiltin.

Lunderberg:
> Based on the current grouping, seems quite a bit of the runtime function dispatchings happens in LowerBuiltin, while legalizeOps primarily focused on lowering to TIR related functions.

I don't think distinguishing between the style of implementation is a useful distinction to make.  The important distinction is what functionality must still be observable outside of the operator, not the functional form of the legalized expression.

tqchen:
There are different ways to look at this particular case.

For this particular case, given the view was lowered to runtime function, it was primarily focused for the VM itself. One can also envision in future we have a codegen approach to get a view function that get inlined which is not needed in the VM approach.

Introducing legalize ops with different levels can also have extra issues, as we need to run default scheduling of some of the ops. But for the certain legalization level we do not have to. 

Perhaps one way to make it more clear is to rename LowerVMBuiltin to LowerRuntimeBuiltin, which have clear indication that that is a pass which takes charge of lowering all implementaitons of runtime builtin functions.

Lunderberg:
> Perhaps one way to make it more clear is to rename LowerVMBuiltin to LowerRuntimeBuiltin, which have clear indication that that is a pass which takes charge of lowering all implementaitons of runtime builtin functions.

I like this idea, but I don't think we should move the definition of the legalized form into the `LowerRuntimeBuiltin`.  What if we were to instead add a new attribute, which has the same signature as `FLegalize`, but would be applied at the later point.  This would allow `LowerRuntimeBuiltin` to replace anything that has the `FLowerBuiltin` attribute, and wouldn't require a distinction between different levels of `FLegalize`.

That would also allow `FLowerBuiltin` to only run after `ToNonDataflow`, and to be implemented in terms of impure functions.  By constrast, since `FLegalize` may replace a call within a dataflow block, the implementation cannot be in terms of an impure call.

> One can also envision in future we have a codegen approach to get a view function that get inlined which is not needed in the VM approach.

I like this idea, and have been toying around with some TIR implementations.  The key limitation at the moment is the inability to construct and return a new `NDArray` if required.  (Similar to the difficulties in returning a string that are blocking https://github.com/apache/tvm/pull/16836 and https://github.com/apache/tvm/pull/17103.)

tqchen:
> I like this idea, but I don't think we should move the definition of the legalized form into the LowerRuntimeBuiltin. What if we were to instead add a new attribute, which has the same signature as FLegalize, but would be applied at the later point. This would allow LowerRuntimeBuiltin to replace anything that has the FLowerBuiltin attribute, and wouldn't require a distinction between different levels of FLegalize.

I think having `FLowerBuiltin` builtin attribute is great. lets go with that

src/relax/op/memory/view.cc
```diff
@@ -334,13 +334,12 @@ Expr LegalizeView(const BlockBuilder& bb, const Call& call) {
...
-  auto runtime_view_sinfo = FuncStructInfo::OpaqueFunc(infer_sinfo_env_func, true);
-
-  ExternFunc runtime_view_func("runtime.TVMArrayCreateView", runtime_view_sinfo);
+  if (shape.same_as(call->args[1]) && dtype.same_as(call->args[2]) &&
+      relative_byte_offset.same_as(call->args[3])) {
+    return call;
+  }
 
-  return Call(runtime_view_func, {data, shape, dtype, relative_byte_offset});
+  return Call(call->op, {data, shape, dtype, relative_byte_offset});
```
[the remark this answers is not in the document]
vinx13: This only does the inference of `void` type args and leave the lowering to the later pass.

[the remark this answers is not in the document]
Lunderberg: I don't think we want to do the inference of the shape/dtype prior to lowering, because it could result in unexpected StructInfo inference later on.

Suppose we have `R.memory.view(arg, shape=[16])`.  This returns a view into the first 16 elements of `arg`, without changing the dtype.  If an `IRModule` pass updates the datatype of `arg`, then that new datatype should also propagate to the view.  However, legalizing it to `R.memory.view(arg, shape=[16], dtype="float16")` would return a view into `arg`, interpreting the first 32 bytes as if they were `"float16"`.  Now, if an `IRModule` pass updates the datatype of `arg`, the view would still be `"float16"`.  To avoid this issue, the unknown arguments shouldn't be filled in until the lowering is about to occur.

What if we were to remove `.set_attr<FLegalize>` altogether, and only have `.set_attr<FLowerBuiltin>`?  That way, we preserve the `R.memory.view` as-is until it is ready to be lowered.  The `LegalizeOps` pass would then be a no-op for `R.memory.view`, and only the `LowerRuntimeBuiltin` pass would change it at all.

include/tvm/runtime/device_api.h
```diff
@@ -240,6 +240,10 @@ class TVM_DLL DeviceAPI {
     return device_type != kDLCPU && device_type != kDLMicroDev;
   }
 
+  static bool SupportsPointerArithmetics(int device_type) {
```
Lunderberg, line 243: Since we already have a vtable for `DeviceAPI`, this should be a virtual function instead of a static boolean.  That would allow individual `DeviceAPI` implementations to independently mark that they support the pointer-arithmetic.  (It would also allow checking for driver-dependent support, such as vulkan support for the optional `VK_KHR_buffer_device_address` feature.)

Since host-side pointer arithmetic is not the default behavior for `DLTensor::data`, the default implementation in `DeviceAPI` would return false, and it could be overridden in `CPUDeviceAPI` and `CUDADeviceAPI` to return true.

Lunderberg: Also, a nitpick: This isn't whether the device supports pointer arithmetic, but whether pointer arithmetic of a device-owned `void* DLTensor::data` may be performed on the host.  The TVM backends for both Vulkan and OpenCL support pointer arithmetic, but only within the generated kernels.  Neither support pointer arithmetic being performed on the host.

src/runtime/relax_vm/builtin.cc
```diff
@@ -545,6 +545,23 @@ TVM_REGISTER_GLOBAL("vm.builtin.tensor_to_shape").set_body_typed([](NDArray data
   return ShapeTuple(out_shape);
 });
 
+TVM_REGISTER_GLOBAL("vm.builtin.ensure_zero_offset").set_body_typed([](NDArray data) {
+  if (data->byte_offset == 0) {
+    return data;
+  }
+  if (DeviceAPI::SupportsPointerArithmetics(data->device.device_type)) {
```
Lunderberg, line 552: Can we also restrict this to `data->byte_offset % kAllocAlignment == 0`?  The alignment guarantee from TVM's allocators ensure that `NDArray->data % kAllocAlignment == 0`.  This is checked when taking ownership of a `DLManagedTensor*`, and would raise an error in `NDArray::FromDLPack` if it fails.

Lunderberg:
@vinx13 I took a look at the current CI failures, and it looks like it pretty close to passing.  If you'd like, applying the diff below should resolve the last 4 failing tests in CI.

[pr_17145_diff.txt](https://github.com/user-attachments/files/16459245/pr_17145_diff.txt)

tqchen (approved):

python/tvm/relax/transform/__init__.py
```diff
@@ -84,14 +85,13 @@
...
 
+from .attach_external_modules import AttachExternModules
+from .fast_math import FastMathTransform
 from .ipc_allreduce_rewrite import IPCAllReduceRewrite
 from .lazy_transform_params import LazyTransformParams
 from .lower_gpu_ipc_alloc_storage import LowerGPUIPCAllocStorage
 from .optimize_layout_transform import OptimizeLayoutTransform
 from .remove_redundant_reshape import RemoveRedundantReshape
-from .fast_math import FastMathTransform
-from .fuse_transpose_matmul import FuseTransposeMatmul
```
Lunderberg, line 93: Looks like the import of `FuseTransposeMatmul` was dropped when sorting the imports.  The last failing unit test is caused by the dropped import.

tqchen:
cc @Lunderberg let us merge this in

Lunderberg (approved):
Thank you for making the changes (and for pinging me on it).  Looks good!

PR #17149, opened 2024-07-10
 [Relax] Implement Rewriter class for pattern-rewrite

Lunderberg:
Prior to this commit, the pattern to be matched and the rewrite to be performed were provided as separate arguments to `rewrite_call`.  This commit introduces a new class `ExprRewriter`, which contains both parts.  This abstraction makes it easier to combine multiple independent rewrite rules, then apply them all in a IRModule pass.

This implementation also allows pattern rewrite rule to be written using TVMScript, by representing both the pattern to be matched, and the replacement to be applied as functions.  For example:

```python
@R.rewriter
class Rewriter:
    @R.function
    def pattern(
        A: R.Tensor([16], "float32"),
        B: R.Tensor([16], "float32"),
    ):
        C = R.add(A, B)
        return C

    @R.function
    def replacement(
        A: R.Tensor([16], "float32"),
        B: R.Tensor([16], "float32"),
    ):
        C = R.call_pure_packed(
            "my_optimized_add_impl", A, B,
            sinfo_args=R.Tensor([16], "float32"),
        )
        return C
```

### DIFF
--- a/include/tvm/relax/block_builder.h
+++ b/include/tvm/relax/block_builder.h
@@ -163,7 +163,7 @@ class BlockBuilderNode : public Object {
   /*!
    * \brief Append a definition to the cuurrent scope.
    *
-   * \param Var A variable within the current scope.
+   * \param var A variable within the current scope.
    *
    * \note This function should be called when a new variable is
    *       defined that may impact struct inference (e.g. MatchCast)
--- a/include/tvm/relax/block_builder.h
+++ b/include/tvm/relax/block_builder.h
@@ -161,7 +161,7 @@ class BlockBuilderNode : public Object {
   virtual void BeginInnerScope() = 0;
 
   /*!
-   * \brief Append a definition to the cuurrent scope.
+   * \brief Append a definition to the current scope.
    *
    * \param var A variable within the current scope.
    *
--- a/include/tvm/relax/block_builder.h
+++ b/include/tvm/relax/block_builder.h
@@ -133,15 +133,46 @@ class BlockBuilderNode : public Object {
    * \brief Begin a new scope, with optional parameters that
    *        are visible within the scope.
    *
+   * Symbolic variables from the parent scope are not available.
+   *
    * \param params Parameters that are visible within the scope.
    *
    * \note This function should be called when new scope is introduced
-   *       (function, seq) to properly track the variable availability
-   *       and help the best effort deduction.
+   *       (e.g. function bodies) to properly track the variable
+   *       availability and help the best effort deduction.
    *
    * \sa EndScope
    */
   virtual void BeginScope(Optional<Array<Var>> params) = 0;
+
+  /*!
+   * \brief Begin a new scope, which inherits visible parameters from
+   * its parent scope.
+   *
+   * Symbolic variables from the parent scope are available.
+   *
+   * \note This function should be called when an inner scope is
+   *       introduced (e.g. conditional branches) to properly track
+   *       the variable availability and help the best effort
+   *       deduction.
+   *
+   * \sa EndScope
+   */
+  virtual void BeginInnerScope() = 0;
+
+  /*!
+   * \brief Append a definition to the cuurrent scope.
+   *
+   * \param Var A variable within the current scope.
+   *
+   * \note This function should be called when a new variable is
+   *       defined that may impact struct inference (e.g. MatchCast)
+   *       to properly track the variable availability and help the
+   *       best effort deduction.
+   *
+   * \sa EndScope
+   */
+  virtual void AddDefinitionToScope(Var var) = 0;
 
   /*! \brief End the previously defined scope. */
   virtual void EndScope() = 0;
--- a/python/tvm/relax/dpl/rewrite.py
+++ b/python/tvm/relax/dpl/rewrite.py
@@ -15,16 +15,75 @@
 # specific language governing permissions and limitations
 # under the License.
 """APIs for pattern-based rewriting."""
-from typing import Dict, Callable
+
+from typing import Dict, Callable, Union
 from .pattern import DFPattern
 from .context import PatternContext
 
+from tvm.ir import IRModule
+from tvm.runtime import Object
+from tvm._ffi import register_object
 from ..expr import Expr, Function, Var
 from . import _ffi as ffi
 
 
+@register_object("relax.dpl.ExprRewriter")
+class ExprRewriter(Object):
+    @staticmethod
+    def from_pattern(
+        pattern: DFPattern,
+        func: Callable[[Expr, Dict[DFPattern, Expr]], Expr],
+    ) -> "ExprRewriter":
+        return ffi.ExprRewriterFromPattern(
+            pattern,
+            func,
+        )  # type: ignore
+
+    @staticmethod
+    def from_module(mod: IRModule) -> "ExprRewriter":
+        return ffi.ExprRewriterFromModule(mod)  # type: ignore
+
+    def __call__(self, obj: Union[Expr, IRModule]) -> Union[Expr, IRModule]:
+        return ffi.ExprRewriterApply(self, obj)
+
+    def __or__(self, other: "ExprRewriter") -> "ExprRewriter":
+        return OrRewriter(self, other)
+
+
+@register_object("relax.dpl.PatternRewriter")
+class PatternRewriter(ExprRewriter):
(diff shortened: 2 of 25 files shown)


Lunderberg:
@sunggg @csullivan

include/tvm/relax/block_builder.h
```diff
@@ -133,16 +133,47 @@ class BlockBuilderNode : public Object {
...
+   *       introduced (e.g. conditional branches) to properly track
+   *       the variable availability and help the best effort
+   *       deduction.
+   *
+   * \sa EndScope
+   */
+  virtual void BeginInnerScope() = 0;
+
+  /*!
+   * \brief Append a definition to the cuurrent scope.
```
sunggg, line 164:
```suggestion
   * \brief Append a definition to the current scope.
```

python/tvm/relax/dpl/rewrite.py
```diff
@@ -15,16 +15,154 @@
...
+
 from .pattern import DFPattern
 from .context import PatternContext
-
 from ..expr import Expr, Function, Var
 from . import _ffi as ffi
 
 
+@register_object("relax.dpl.ExprRewriter")
+class ExprRewriter(Object):
```
sunggg, line 32: Can we rename it to something like `ExprPatternMatchingRewriter`? `ExprRewriter` sounds too generic even though it is only about the pattern matching unless I'm missing something.

python/tvm/relax/dpl/rewrite.py
```diff
@@ -15,16 +15,154 @@
...
+
+    @staticmethod
+    def from_module(mod: IRModule) -> "ExprRewriter":
+        """Construct a rewriter from an IRModule
+
+        Parameters
+        ----------
+        mod: IRModule
+
+            A module with `pattern` and `replacement` functions,
```
sunggg, line 78: Since this is new feature, maybe good to add an example in the comment?

tests/python/relax/test_dataflow_rewriter.py
```diff
@@ -0,0 +1,1370 @@
...
+    with pytest.raises(ValueError, match="must have the same signature"):
+
+        @R.rewriter
+        class Rewriter:
+            @R.function
+            def pattern(A: R.Tensor([32])):
+                return A
+
+            @R.function
+            def replacement(A: R.Tensor([16])):
```
sunggg, line 133: Out of curiosity, do we catch these errors during parsing? or do we defer until we run the replacement?

tests/python/relax/test_dataflow_rewriter.py
```diff
@@ -0,0 +1,1370 @@
...
+        concat_AB = R.concat([A, B])
+        proj_concat = R.matmul(concat_AB, state)
+        proj_A = R.strided_slice(proj_concat, axes=[0], begin=[0], end=[16])
+        proj_B = R.strided_slice(proj_concat, axes=[0], begin=[16], end=[32])
+        out = proj_A + proj_B
+        return out
+
+    after = Rewriter(before)
+    tvm.ir.assert_structural_equal(expected, after)
+
```
sunggg, line 1302: Can we add a test case for matching with "any" datatype?

sunggg (commented):
Hi, @Lunderberg! Thank you for the cool addition! 
I haven't taken a deeper look at the detailed mechanics yet, but I love how this PR allows us to manage the pattern and rewrite conveniently with better readability :) 

My main question for you is if this PR is aiming the full support for the existing pattern language. 
It is okay if there are some rough edges that we don't currently support, but I would like to clearly document them somewhere to avoid the potential misusage. 

Here are some of the cases I wonder how these will be supported with your PR:
* `~` not pattern: https://github.com/apache/tvm/blob/main/tests/python/relax/test_dataflow_pattern.py#L222
* If we want to match with any placeholder (`wildcard()`), which could be tensor, constant, etc., how should we express this?
* Can we match with any number of function arguments?

include/tvm/relax/block_builder.h
Lunderberg: Thank you, and fixed!

python/tvm/relax/dpl/rewrite.py
Lunderberg: Hmm.  I definitely agree on it being too generic of a name, but I think that specifying it as `PatternMatching` would be a bit too specific of a name.

For example, the existing `CombineParallelMatmul` pass couldn't be expressed as a pattern-matching rule (cannot express arbitrarily many matmuls with a shared RHS), but could be expressed as a subclass of `ExprRewriter`.  Doing so would then allow `CombineParallelMatmul` to be composed together with other `ExprRewriter` subclasses into a single optimization pass.

tests/python/relax/test_dataflow_rewriter.py
Lunderberg: These errors are caught while converting the `IRModule`, as part of the `ExprRewriter::FromModule` function, before the `@R.rewriter` annotator returns the rewriter.  So, technically it occurs after parsing (since the IRModule is fully defined) but before replacement (since the `ExprRewriter` is validated on construction).

tests/python/relax/test_dataflow_rewriter.py
Lunderberg: Can do.  I added a test case with a pattern-match against arbitrary datatype (last slice of a 2-d tensor), which can be applied to any known dtype, or an unknown dtype.

Lunderberg:
> My main question for you is if this PR is aiming the full support for the existing pattern language. It is okay if there are some rough edges that we don't currently support, but I would like to clearly document them somewhere to avoid the potential misusage.

I've been aiming to support as much of the existing pattern language as possible, but not the entire syntax.  There are some portions of the pattern language that don't have clear Relax equivalents, and those wouldn't be supported.

> Here are some of the cases I wonder how these will be supported with your PR:
> 
>     * `~` not pattern: https://github.com/apache/tvm/blob/main/tests/python/relax/test_dataflow_pattern.py#L222

This is probably the biggest one that doesn't have a relax-level equivalent.  The main usage would be to avoid aliasing in a pattern-match.  (e.g. An implementation that is valid so long as two wildcard patterns do not refer to the same tensor.)

>     * If we want to match with any placeholder (`wildcard()`), which could be tensor, constant, etc., how should we express this?

If the pattern has an argument `var_name: R.Object`, then it would match against any Relax value.  This is converted to `StructInfoPattern(WildcardPattern(), ObjectStructInfo())` when producing the pattern rule.  When the `StructInfoPattern` validates a match, it does so using `StructInfoBaseChecker`.  Since `ObjectStructInfo()` is the base of all other relax types, this validation always passes.

There are some edge cases where this doesn't work, mostly arising from struct inference applied within the pattern or replacement.  For example, suppose the pattern has wildcard `A: R.Object` and returns `A + A`.  This would raise an error, because Relax requires binary operations to have either `R.Prim` or `R.Tensor` struct info.  Resolving this edge case would require re-visiting [this discussion](https://github.com/apache/tvm/pull/16827#discussion_r1546888040) on whether the struct inference should throw an error for provably-incorrect usage, or should throw an error for not-provably-correct usage.

>     * Can we match with any number of function arguments?

We can have arbitrarily many function arguments for the pattern/replacement.  However, each pattern must have a fixed number of arguments.

sunggg:
Thanks, @Lunderberg for the clarification! I think it is absolutely okay to have some rough edges that we cannot support yet, but I hope we clearly document in the code so that people won't get confused.

python/tvm/relax/dpl/rewrite.py
Lunderberg: Thinking on it a bit more, I think `PatternMatchingRewriter` would be a reasonable name.  Because the C++ API must return a `Map<Var,Expr>`, it is restricted to some sort of pattern-match.  That pattern may be represented in a specific `DFPattern` object, or it could be implicit in the function used to identify a match, but they're still restricted to pattern-matching.

Lunderberg: And names are all updated, running through CI.

Lunderberg:
> I think it is absolutely okay to have some rough edges that we cannot support yet, but I hope we clearly document in the code so that people won't get confused.

Agreed.  Any unsupported functionality should raise an error during the conversion, which (hopefully) will make it as clear as possible.  I'm a big fan of error-checking as early as possible, especially when the input is relatively unstructured.

PR #17155, opened 2024-07-12
[CI] Remove lint step from `unity/pr-head` step

Lunderberg:
This step should only be performed as part of the `lint/pr-head` CI step.  It was included as part of the unity-specific CI steps prior to merging of unity into main.  It is no longer necessary as part of `unity/pr-head`.


Lunderberg:
@tvm-bot rerun

quic-sanirudh (approved):
This will definitely help, thanks.

Not strictly related to this PR, but as a general question, do we need to keep the unity/pr-head separate anymore? I see that it builds and runs some cpu and gpu tests, couldn't we just move them to the cpu and gpu heads and just remove this. It might help avoid building for cpu and gpu in 2 separate builds, and avoid running some of these tests multiple times?

Lunderberg:
> do we need to keep the unity/pr-head separate anymore?

I don’t know of any reason why they’d need to remain separate, and that would definitely help make the test summaries more readable.

PR #17157, opened 2024-07-15
[Relax] Integrate cuDNN attention

vinx13:
This integrates cuDNN attention kernels to BYOC.
A dependency of cudnn_frontend is added.
The cuDNN attention kernel supports fused qkv in BS3NH and SBN3H layouts.

cc @sunggg @masahi @yongwww @tqchen


yongwww (approved):

python/tvm/contrib/cutlass/gen_tensor_op.py
```diff
@@ -745,8 +745,8 @@ def get_batch_on_arg(arg_name, arg_shape):
             attrs["qkv"] = func_args[0]
             attrs["num_queries"] = s = annotations["num_queries"]
             attrs["num_keys"] = annotations["num_keys"]
-            if len(func_args) > 5 and not is_var_len:  # +1 for workspace, the last arg
-                attrs["bias"] = func_args[4]
+            if len(func_args) > 2 and not is_var_len:  # +1 for workspace, the last arg
```
sunggg, line 748: I guess cutlass side is mostly the interface change? 
Or is there any functional change?

sunggg (commented):
Thank you @vinx13 for the new addition! 
Overall, looks good to me.
Would you describe high-level strategy for attention somewhere? (e.g., when to offload cudnn, cutlass, TIR, etc.)
If this is about landing machinery rather than such offloading decision, would appreciate if you can provide some recommendations.

vinx13:
The new attention can be applied via cudnn BYOC. The decision of which BYOC backend (cudnn, cutlass) to use is left to the users. cudnn is likely to perform better on H100 as it has specific optimizations

python/tvm/contrib/cutlass/gen_tensor_op.py
vinx13: FuseOps is updated to inline the constant shape parameters. So there are fewer func_args here

masahi (approved):
I remember cuDNN attention supports fp8, would be interesting to support that too.

PR #17158, opened 2024-07-15
[TIR][Analyzer] Simplify `x==x` expressions for all dtypes

Lunderberg:
Prior to this commit, there was no rule to simplify `x == x` into `True`.  In some cases, despite not having an explicit rewrite rule in `RewriteSimplifier`, the `RewriteSimplifier::CanProve` function would check if `x-x` simplifies to zero, relying on the rewrite rules used for `tir::Sub`.  However, the rule to rewrite `x-x` into zero was only enabled for `int32`, `int64`, and floating-point types, so relying on this behavior was inconsistent.

This commit updates the rewrite rules for both `tir::EQ` and `tir::Sub` to check for simplification of `x-x` or `x==x`, regardless of the datatype.  This change preserves the fast-path for index data-types, in which `int32` and `int64` expressions may be simplified without checking for side effects.  For all other dtypes, the cancellation only applies when evaluating `x` has no side effects.

### DIFF
--- a/src/arith/rewrite_simplify.cc
+++ b/src/arith/rewrite_simplify.cc
@@ -701,6 +701,12 @@ PrimExpr RewriteSimplifier::Impl::VisitExpr_(const SubNode* op) {
   } else {
     // Cancellation rules.  Deliberately off of the integer path, to
     // avoid introducing checks on the side effects for the fast path.
+    //
+    // These simplifications do not preserve NaN/Inf that may occur in
+    // the inputs.  For IEEE floats, `NaN - NaN` is `NaN`, and does
+    // not cancel out.  However, since models should not encounter NaN
+    // in the first place, this allows better simplification for the
+    // supported path.
     TVM_TRY_REWRITE_IF(x - x, ZeroWithTypeLike(x),
                        SideEffect(x.Eval()) <= CallEffectKind::kReadState);
     TVM_TRY_REWRITE_IF((x + y) - y, x, SideEffect(y.Eval()) <= CallEffectKind::kReadState);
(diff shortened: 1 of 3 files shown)


src/arith/rewrite_simplify.cc
```diff
@@ -697,7 +698,7 @@ PrimExpr RewriteSimplifier::Impl::VisitExpr_(const SubNode* op) {
     TVM_TRY_RECURSIVE_REWRITE(x - (y + c1), (x - y) + (0 - c1));
     TVM_TRY_RECURSIVE_REWRITE(x - (y - z), (x + z) - y);
     TVM_TRY_RECURSIVE_REWRITE(x - y * c1, x + y * (0 - c1));
-  } else if (op->dtype.is_float()) {
+  } else {
```
sunggg, line 701: For currently unsupported integer types, is there any reason to take this path rather than path for the index types? They are also integers and this path has more rules defined. But, I'm not currently sure if it supports negative index but if it does not, it could be effectively same with unsigned integer?

sunggg (commented):
Great catch, @Lunderberg! 
LGTM, one question.

src/arith/rewrite_simplify.cc
Lunderberg: Some of the rewrite rules in the `IsIndexDType` branch assume that overflow won't occur.  While this isn't guaranteed to be true, we're more likely to get overflow for small integer types than for the 32/64-bit signed integers used for index dtypes.

For the uint16/uint8 types seen in https://github.com/apache/tvm/pull/17149, it would be a lot easier to get accidental overflow.

src/arith/rewrite_simplify.cc
```diff
@@ -1698,6 +1700,11 @@ PrimExpr RewriteSimplifier::Impl::ApplyRewriteRules(EQ ret) {
     TVM_TRY_REWRITE(c1 - x == c2, x == c1 - c2);
     TVM_TRY_REWRITE(x + c1 == c2, x == c2 - c1);
     TVM_TRY_RECURSIVE_REWRITE(x * y == 0, x == 0 || y == 0);
+    TVM_TRY_REWRITE(x == x, ctrue);
+  } else {
```
wrongtest-intellif, line 1704: Hi, I have one question on that generally `NaN == NaN` evaluates to False. I think it would be great to leave some explaination on the case. For example, could we just define it as some "undefined" behaviour?

Lunderberg: Good point, and I'll add a comment on it.  In general, the numeric simplifications assume that the values are neither `NaN` nor `Inf`, and that expressions do not rely on integer overflow for correctness.  That is, integer types and floating-point types are simplified as if they are mathematical integer/real numbers, rather than IEEE floats.

wrongtest-intellif (approved):

PR #17160, opened 2024-07-16
[Relax] Fix fuseOps via pattern 

Hzfengsy:
The pass would remove functions when iterating the whole IRModule, which would cause function not found issue. 
This PR fixes this one

cc @tqchen


quic-sanirudh (approved):
LGTM, Thanks.

tqchen (approved):

PR #17170, opened 2024-07-18
Pass to eliminate redundant branch and overcompute

sdalvi-quic:
This C++ pass iterates through the primfunc and finds if there is an if_then_else branch related to layout specific padding where we can leverage from assume statement, remove the extra branch and over-compute.
This pass works if the branch has the buffer present in its then_clause or else_clause for which buffer assumption exists. If there are intermediate buffers, then we need to inline the code to use the pass. This pass first parses the assume statement and then uses the information to simplify the expression in the then_clause and else_clause in then_condition_context and else_condition_context. It then decides to eliminate the branch if then_clause and else_clause evaluates to same value in then_condition_context (since then_clause will be true in this context and if else_clause is also evaluating to true then we can directly replace it with else_clause), it returns the then_clause if we are in the else_condition_context and then_clause and else_clause evaluates to same value.

Please find the image for reference :
image
image

### DIFF
--- a/include/tvm/tir/transform.h
+++ b/include/tvm/tir/transform.h
@@ -835,7 +835,7 @@ TVM_DLL Pass InstrumentProfileIntrinsics();
 TVM_DLL Pass DefaultGPUSchedule();
 
 /*!
- * \brief This pass analyzes primfunc and eliminates branch introdued due to layout specific padding.
+ * \brief This pass analyzes primfunc & eliminates branch introdued due to layout specific padding.
  *  It leverages from the buffer assumptions and use the information to eliminate the branch.
  * \note This creates more opportunity to vectorize the code.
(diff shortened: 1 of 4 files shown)


sdalvi-quic:
@jverma-quic and @quic-sanirudh can you please review the PR?

sdalvi-quic:
After making changes in the way we identify the elementwise op (i.e from the op pattern), I have tested the pass with the test cases only. In resent50 the op_pattern value for elementwise ops is set to 8 (i.e kOpaque), because of which when we check for an op to be elementwise op using op_pattern it may not give correct results. Hence, there is slight change in downstream where we identify an op by the path op_attrs -> op_name -> op_kind.

quic-sanirudh (approved):
LGTM, Thanks.

PR #17177, opened 2024-07-19
[Relay][FQ2I]: Use appropriate dtype while quantizing relay.op.nn.pad…

arangasa:
…'s constant pad value

The test case fails without this change.

@mbrookhart @tqchen


quic-sanirudh (approved):
LGTM Thanks.

PR #17180, opened 2024-07-21
[Disco] Group-wise operation

cyx-6:
This PR introduces the group attribute into Disco, so that group wise allreduce and allgather is enabled.

cc: @MasterJH5574 @vinx13


MasterJH5574 (approved):
Thanks for the comprehensive enhancement!

PR #17182, opened 2024-07-21
[Disco] Implement SocketSession

vinx13:
Implements SocketSession that connects multiple local worker processes/threads over multiple distributed nodes via TCP socket.

cc @tqchen @cyx-6

### DIFF
--- a/include/tvm/runtime/disco/disco_worker.h
+++ b/include/tvm/runtime/disco/disco_worker.h
@@ -54,6 +54,7 @@ class DiscoWorker {
       : worker_id(worker_id),
         num_workers(num_workers),
         num_groups(num_groups),
+        local_worker_id(worker_id),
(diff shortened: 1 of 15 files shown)


include/tvm/runtime/disco/disco_worker.h
```diff
@@ -68,6 +69,10 @@ class DiscoWorker {
   int worker_id;
   /*! \brief Total number of workers */
   int num_workers;
+  /*! \brief The local worker id. This can be different from `worker_id` if the session is
+   * consisted of multiple distritributed sub-sessions.
+   */
+  int local_worker_id;
```
MasterJH5574, line 75: We introduced `int num_groups` in #17180 which causes some conflicts. `local_worker_id` now equals to `worker_id % (num_workers / num_groups)`.

tqchen (approved):

PR #17203, opened 2024-07-26
[Relax][Analysis] Validate global_symbol on non-Relax functions

Lunderberg:
Prior to this commit, the well-formed checker verified that the `"global_symbol"` attribute, if present, matches the name of the `GlobalVar`.  However, this check was only applied for Relax functions.  As a result, discrepencies between the `"global_symbol"` and the `gvar->name_hint` could result in unexpected bugs.  (For example, https://github.com/apache/tvm/issues/17176.)

This commit updates the well-formed checker to verify `"global_symbol"` on all functions in an `IRModule`.


Lunderberg:
I expect this PR to fail CI, as https://github.com/apache/tvm/pull/17202 resolves a failure mode that this change is designed to detect.  (Though that isn't necessarily the only case that this check will catch.)

PR #17204, opened 2024-07-26
[Hexagon] Fix LWP assembly handler (predicate register)

FranckQC:
This PR solves a known issue with Lightweight Profiling (LWP) on Hexagon that we have fixed in our downstream repo and that I'd like to upstream for the community. The issue appeared with the operator maxpool. When using LWP it gave incorrect results (i.e. with the accuracy terribly bad).

The problem was that the LWP handler was forgetting to save p0 (used by the handler). This predicate register needs to be saved too, just like r0-r5, as it had been decided that it was the responsibility of the handler to save everything (even these theoretically caller-saved registers). Said differently, since it had been decided that calling the LWP handler would not follow the normal ABI, and that the LWP handler would save everything it touches (even normally caller-saved registers like r0-r15 and p0-3), then it absolutely needs to save the predicate registers too (in particular p0, which was causing the issue).

The issue appeared only with maxpool because it's the only one that had a state saved in p0 before calling the LWP handler. And this call destroyed the content of what it had saved, making it subsequently branch to different portions of the code.

Fix: Allocate 32 bytes (instead of 24 previously), in order to save p3:0, and I save those at the bottom of the stack. Restore it at the end of the LWP handler.

### DIFF
--- a/src/runtime/hexagon/profiler/lwp_handler.S
+++ b/src/runtime/hexagon/profiler/lwp_handler.S
@@ -50,7 +50,7 @@ handler itself.
   .falign
   .type  lwp_handler,@function
 lwp_handler:
-  { 
+  {
     allocframe(#32)                            // Allocate 32 bytes on the stack to save R0-R5 registers (6*4bytes) and P0-P3 (4*1byte) + 4 unused bytes as the stack has to be 8-bytes aligned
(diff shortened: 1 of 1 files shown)


quic-sanirudh (approved):
Thanks @FranckQC

PR #17201, opened 2024-07-26
[Transform][Relax] Handle `is_group` argument in IPC AllReduce

Lunderberg:
The `relax.transform.IPCAllReduceRewrite` pass rewrites calls to `"runtime.disco.allreduce"` to instead call an optimized `"runtime.disco.cuda_ipc.custom_allreduce"` version.  When the legalization of `R.ccl.allreduce` was updated in https://github.com/apache/tvm/pull/17180 to provide an `in_group` argument, the `IPCAllReduceRewrite` pass was not updated.

This commit updates the `IPCAllReduceRewrite` to be handle the additional `in_group` argument.

PR #17206, opened 2024-07-26
[CI] Update dummy-variable regex for pylint

Lunderberg:
Prior to this commit, the regex used for pylint to identify dummy variables would correctly identify variables that start with an underscore (e.g. `_scale`), unless they have an underscore elsewhere in the name (e.g. `_scale_factor`).  This leads to false positives from pylint for unused variables, as prefixing a variable with an underscore should mark a variable as intentionally unused.

This commit updates the regex in TVM's `pylintrc` to match the current default value for `dummy-variables-rgx`, to allow unused variables to be named with a leading underscore, even if they also contain another underscore.

### DIFF
--- a/tests/lint/pylintrc
+++ b/tests/lint/pylintrc
@@ -252,7 +252,7 @@ init-import=no
 
 # A regular expression matching the name of dummy variables (i.e. expectedly
 # not used).
-dummy-variables-rgx=(_+[a-zA-Z0-9]*?$)|dummy
+dummy-variables-rgx=_+$|(_[a-zA-Z0-9_]*[a-zA-Z0-9]+?$)|dummy|^ignored_|^unused_
 
 # List of additional names supposed to be defined in builtins. Remember that
 # you should avoid to define new builtins when possible.


Lunderberg:
I ran into this false positive for https://github.com/apache/tvm/pull/17201 for a variable named `_in_group` (screenshot below).  While I could work around it by renaming to `_ingroup`, it's also good to have a long-term fix for the lint rule.

image

quic-sanirudh (approved):
Since we're already using names starting with `_` to represent dummy variables, it makes sense to allow `_` in other places as part of the name.

One thing though is that [PEP 8 style guide](https://peps.python.org/pep-0008/#descriptive-naming-styles) suggests using names starting with `_` to be used as "weak" internal use variables, and the convention I've used till now is that unused variables are named with just an `_`  like `x, _ = get_xy()`. So, my question is why do we want to name unused variables in general? Could we not just use `_`?

quic-sanirudh:
> > So, my question is why do we want to name unused variables in general? Could we not just use `_`?
> 
> Mainly to indicate to a reader which return values are being ignored. As an example, suppose there's a function that produces a
> 
> ```python
> # A reader may not know what the first two return values are.
> _, _, attrs = define_function(relax_args)
> 
> # The first two return values are a function and parameters to 
> # that function.
> _func, _params, attrs = define_function(relax_args)
> ```
> 
> I've run into a few cases where the same return value was re-generated, because a later developer wasn't aware that it was already being generated. By explicitly naming the unused return value, this becomes less likely. (e.g. Seeing `_func` instead of `_`, and checking whether it is already the function that should be used.)
> 
> Obviously, this shouldn't be applied in all cases, as `_, y = get_xy()` is already perfectly readable, but it can help in cases where it isn't as clear.

Makes sense, thanks for your reply.

PR #17214, opened 2024-07-30
Replacing unary ops with LookUpTable and Take op to improve performance

sdalvi-quic:
**Motivation :** Unary ops calculate the value during runtime which is costly. Since we are handling quantized data hence, it is better to pre compute the value during compile time (for 0 -256 value) and store in a Look Up Table which is accessed during the runtime.

**Implementation :** This pass generates the Look Up Table for each unary op during compile time. It iterates over the original graph and replaces the unary ops call node, (example tanh, sqrt) to a call node that accesses a Look Up Table corresponding to the unary op, and return the value.

**Testing :** The test_take checks for both value correctness and structural correctness of the output generated after the pass.

Reference: https://github.com/apache/tvm/pull/15614

### DIFF
--- a/python/tvm/contrib/hexagon/generate_take_op.py
+++ b/python/tvm/contrib/hexagon/generate_take_op.py
@@ -1,0 +1,86 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+# pylint: disable=missing-docstring, invalid-name, unnecessary-comprehension, unused-argument
+
+import tvm
+import tvm.testing
+from tvm import relax
+from tvm.contrib.hexagon import hexagon_unary_ops
+
+
+def op_replace(call_node):
+    def is_op(op_name: str, call_node: relax.Call) -> bool:
+        if not isinstance(call_node, relax.Call):
+            return False
+        call_tir_op = tvm.ir.Op.get("relax.call_tir")
+        if call_node.op != call_tir_op:
+            return False
+        global_var = call_node.args[0]
+        return op_name in global_var.name_hint
+
+    ops = ["tanh", "sqrt", "rsqrt", "exp", "erf", "sigmoid", "hardswish", "log", "abs"]
+    for op in ops:
+        if is_op(op, call_node):
+            return True
+    return False
+
+
+@relax.expr_functor.mutator
+class Tanh2TakeReplace(tvm.relax.PyExprMutator):
+    def __init__(self, mod: tvm.IRModule) -> None:
+        super().__init__(mod)
+        self.mod_ = mod
+
+    def transform(self) -> tvm.IRModule:
+        # Iterate over all the nodes to check for the node replaceable
+        for global_var, func in self.mod_.functions.items():
+            # Skip non-relax functions
+            if not isinstance(func, relax.Function):
+                continue
+            updated_func = self.visit_expr(func)
+            self.builder_.normalize(updated_func)
+            self.builder_.update_func(global_var, updated_func)
+        # At the end of the transformation we return the updated IRModule from the BlockBuilder.
+        return self.builder_.get()
+
+    def visit_call_(self, call_node: relax.Call) -> relax.Call:
+        if call_node.args[1][0].struct_info.dtype == "uint8":
+            if op_replace(call_node):
+                inp, inp_scale, inp_zp, out_scale, out_zp = [x for x in call_node.args[1]]
+                # LUT node creation
+                LUT = hexagon_unary_ops.LUT_generation(
+                    inp_scale, inp_zp, out_scale, out_zp, call_node.args[0].name_hint
+                )
+                # Take operation node creation
+                take_func = hexagon_unary_ops.generate_take_primfunc(inp, call_node.struct_info)
+                take_func = take_func.without_attr("global_symbol")
+                take_func_gv = self.builder_.add_func(take_func, "take")
+                take_node = relax.call_tir(
+                    take_func_gv,
+                    relax.expr.Tuple(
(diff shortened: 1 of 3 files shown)


sdalvi-quic:
@jverma-quic and @quic-sanirudh can you please review the PR? 
I had implemented this sometime back but, could not merge it then, and the PR was closed.
Reference : https://github.com/apache/tvm/pull/15614.

sdalvi-quic:
@tvm-bot rerun

python/tvm/contrib/hexagon/generate_take_op.py
```diff
@@ -0,0 +1,86 @@
...
+
+def op_replace(call_node):
+    def is_op(op_name: str, call_node: relax.Call) -> bool:
+        if not isinstance(call_node, relax.Call):
+            return False
+        call_tir_op = tvm.ir.Op.get("relax.call_tir")
+        if call_node.op != call_tir_op:
+            return False
+        global_var = call_node.args[0]
+        return op_name in global_var.name_hint
```
quic-sanirudh, line 33: Should we use a better solution than looking at the global_var name to determine the type of op? Names might not be the most reliable way. Perhaps look for an attribute like `operator_name` similar to what is used by [`AlterOpImpl` pass](https://github.com/apache/tvm/blob/main/tests/python/relax/test_transform_alter_op_impl.py#L53)?

python/tvm/contrib/hexagon/generate_take_op.py
```diff
@@ -0,0 +1,86 @@
...
+            if not isinstance(func, relax.Function):
+                continue
+            updated_func = self.visit_expr(func)
+            self.builder_.normalize(updated_func)
+            self.builder_.update_func(global_var, updated_func)
+        # At the end of the transformation we return the updated IRModule from the BlockBuilder.
+        return self.builder_.get()
+
+    def visit_call_(self, call_node: relax.Call) -> relax.Call:
+        if call_node.args[1][0].struct_info.dtype == "uint8":
```
quic-sanirudh, line 61: Should we verify whether the call_node is a `relax.call_tir` op before accessing the args?

python/tvm/contrib/hexagon/generate_take_op.py
```diff
@@ -0,0 +1,86 @@
...
+            if op_replace(call_node):
+                inp, inp_scale, inp_zp, out_scale, out_zp = [x for x in call_node.args[1]]
+                # LUT node creation
+                LUT = hexagon_unary_ops.LUT_generation(
```
Lunderberg, line 65: When is this pass intended to be applied?  If it's moved to before `LegalizeOps`, then that would make it easier to define the lookup table as a Relax expression, which would then be handled by the `FoldConstant` pass.

That would also allow the pattern-matching to be done based on the Relax operations themselves, rather than their lowered names.

python/tvm/contrib/hexagon/generate_take_op.py
jverma-quic: > Should we verify whether the call_node is a `relax.call_tir` op before accessing the args?

@quic-sanirudh: wouldn't it be guaranteed since we're only visiting the call nodes?

python/tvm/contrib/hexagon/generate_take_op.py
jverma-quic: I agree with you that relying on the global_var is not the best way to identify the operators for this transformation. However, I don't really think that operator_name will be much better. The problem here is that we lower the graph to Relay first and then during translation to Relax, the operator knowledge is lost. @Lunderberg's suggestion would have worked very well if we could have imported the graph directly to Relax and then before legalizing it, we could have replaced R.tanh with R.take(..).

quic-sanirudh: @jverma-quic I agree that `operator_name` is not the best solution either, and yeah @Lunderberg's solution is certainly the best, but yes, since we translate a relay graph to relax, we don't have that option. My only concern was that, atleast with an attribute like `operator_name`, it has to be deliberately annotated if the pass has to work. If we use the function name instead, there could potentially be a case where someone could apply the pass on some fused op that has say something like `add_tanh` and that might get replaced and result in correctness errors.

That being said, since this pass isn't added to any pipeline, and has to be consciously applied by someone, I think this might also be okay.

As for @Lunderberg's suggestion, we could perhaps support both types of ops (legalized and direct relax ops).

python/tvm/contrib/hexagon/generate_take_op.py
quic-sanirudh: Call nodes can be any relax call I think, and call_tir is just one type of call node right? What if there's a builtin call or some direct relax op call, etc.

python/tvm/contrib/hexagon/generate_take_op.py
Lunderberg: Thank you on clarifying the use case, that that makes sense with why it would only have access to the legalized operators.

jverma-quic: > @jverma-quic I agree that `operator_name` is not the best solution either, and yeah @Lunderberg's solution is certainly the best, but yes, since we translate a relay graph to relax, we don't have that option. My only concern was that, atleast with an attribute like `operator_name`, it has to be deliberately annotated if the pass has to work. If we use the function name instead, there could potentially be a case where someone could apply the pass on some fused op that has say something like `add_tanh` and that might get replaced and result in correctness errors.
> 
> That being said, since this pass isn't added to any pipeline, and has to be consciously applied by someone, I think this might also be okay.
> 
> As for @Lunderberg's suggestion, we could perhaps support both types of ops (legalized and direct relax ops).

sdalvi-quic: Thank you for the suggestions! @Lunderberg, @jverma-quic, @quic-sanirudh

I had few questions, this pass works for quantized data, does relax support quantized unary ops? I couldn't find quantized direct relax ops. 
In that case we may not be able to do the optimization for direct relax ops yet. 

Please let me know your suggestions on the same.

python/tvm/contrib/hexagon/generate_take_op.py
sdalvi-quic: To be safer I will add check for call_tir before invoking the pass. Thank you.

quic-sanirudh (approved):
LGTM, Thanks @sdalvi-quic

quic-sanirudh:
@Lunderberg As @sdalvi-quic has mentioned above in her comment, since this pass is to replace quantized (qnn) ops with a lookup table, and since relax does not have quantized ops yet (AFAIK, pls correct me if I'm wrong about that), I think this version is probably the best we'll be able to do for now.

jverma-quic (approved):
LGTM

Lunderberg:
Agreed, since the quantized ops have already been converted to Relax by the point this pass is applied, the high-level Relax information would no longer be present.

quic-sanirudh:
Thanks @sdalvi-quic @Lunderberg @jverma-quic.

Issue #17217, opened 2024-07-30
[Bug] [Relax] VMBuiltinLower expects bound value to be a ShapeExpr


Cookiee235:
### Actual behavior

```
Traceback (most recent call last):
  File "test_simple.py", line 46, in <module>
    ex = relax.build(mod, target='llvm') # crash here!
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/software/tvm/python/tvm/relax/vm_build.py", line 335, in build
    mod = pipeline(mod)
          ^^^^^^^^^^^^^
  File "/software/tvm/python/tvm/ir/transform.py", line 265, in __call__
    return _ffi_transform_api.RunPass(self, mod)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/software/tvm/python/tvm/_ffi/_ctypes/packed_func.py", line 239, in __call__
    raise_last_ffi_error()
  File "/software/tvm/python/tvm/_ffi/base.py", line 481, in raise_last_ffi_error
    raise py_err
  File "/software/tvm/python/tvm/relax/pipeline.py", line 101, in _pipeline
    mod = seq(mod)
          ^^^^^^^^
  File "/software/tvm/python/tvm/ir/transform.py", line 265, in __call__
    return _ffi_transform_api.RunPass(self, mod)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/software/tvm/python/tvm/_ffi/_ctypes/packed_func.py", line 239, in __call__
    raise_last_ffi_error()
  File "/software/tvm/python/tvm/_ffi/base.py", line 481, in raise_last_ffi_error
    raise py_err
tvm._ffi.base.TVMError: Traceback (most recent call last):
[...]

PR #17218, opened 2024-07-30
[Relax] Allow dynamic shape argument to R.reshape

Lunderberg:
Prior to this commit, the `shape` argument to `R.reshape` was required to either be an in-line `relax::ShapeExpr`, or a variable that had been bound to a `relax::ShapeExpr` within the current function.  As a result, shapes that were provided as function arguments or that were produced by another operation (e.g. `R.tensor_to_shape`) would unnecessarily trigger an error.

This commit updates the `VMBuiltinLower` pass to instead check that the argument has `relax::ShapeStructInfo`.

Closes https://github.com/apache/tvm/issues/17217


tqchen (approved):

Lunderberg:
Thank you for the review, and rebased onto main to resolve conflicts.

PR #17224, opened 2024-07-31
[Relax][Analysis] Handle recursive functions in CollectVarUsage

Lunderberg:
Prior to this commit, the `relax::analysis::CollectVarUsage` utility treated a local function definition as in-scope after visiting the body of the local function.  As a result, recursive calls from a local function were incorrectly identified as calls to an undefined variable.

This commit updates the `CollectVarUsage` to treat a local function definition as in-scope when inspecting the function body.  This change is similar to the change made for structural equality in https://github.com/apache/tvm/pull/16756.


yongwww (approved):
LGTM

PR #17208, opened 2024-07-27
[Runtime] Allow aborting fetchNDArray through AbortSignal

Neet-Nestor:
This PR changes the interface of function `fetchNDArrayCache` of TVM runtime to accept an additional optional parameter of `signal: AbortSignal` to allow clients to use `AbortController` to abort a given fetch.

https://developer.mozilla.org/en-US/docs/Web/API/AbortController

Related issues:
https://github.com/mlc-ai/web-llm/issues/484
https://github.com/mlc-ai/web-llm/issues/499

PR #17227, opened 2024-08-01
[Runtime] Allow aborting fetchWithCache through AbortSignal

Neet-Nestor:
This is a follow-up for a previous change https://github.com/apache/tvm/pull/17208.

This Pull Request updates function `fetchWithCache()` in TVM runtime class to accept an optional parameter `signal: AbortSignal`, so that users could use `AbortController` to abort the fetch process if needed.

https://developer.mozilla.org/en-US/docs/Web/API/AbortController

Related issues:
https://github.com/mlc-ai/web-llm/issues/484
https://github.com/mlc-ai/web-llm/issues/499


Neet-Nestor:
cc. @tqchen

tqchen (approved):

PR #17229, opened 2024-08-01
[Bugfix][Cutlass] fix cutlass instantiate attention template bugs

senlyu163:
Fixed a bug in cutlass byoc during instantiation of attention template, which would cause CUDA ERROR ILLEGAL ADDRESS error.


senlyu163:
cc @masahi @sunggg @tqchen

masahi (approved):

PR #17237, opened 2024-08-02
[Runtime] Reorganize PagedKVCache attn kernel invocation

MasterJH5574:
This PR reorganizes the attention kernel invocation logic in the PagedKVCache, so that in cases of sequence fork, we can effectively merge one ragged-prefill kernel and a decode kernel into a single decode kernel.


MasterJH5574:
Depending on #17236.

tqchen (approved):

PR #17240, opened 2024-08-05
[FFI][Runtime] Use TVMValue::v_int64 to represent boolean values

Lunderberg:
This is a follow-up to https://github.com/apache/tvm/pull/16183, which added handling of boolean values in the TVM FFI.  The initial implementation added both a new type code (`kTVMArgBool`) and a new `TVMValue::v_bool` variant.  This commit removes the `TVMValue::v_bool` variant, since the `kTVMArgBool` type code is sufficient to handle boolean arguments.

Removing the `TVMValue::v_bool` variant also makes all `TVMValue` variants be 64-bit (assuming a 64-bit CPU).  This can simplify debugging in some cases, since it prevents partial values from inactive variants from being present in memory.


Lunderberg:
Rebased onto main, to rerun CI with the extra fixes after the revert/reapply of the FFI changes in https://github.com/apache/tvm/pull/17252 and https://github.com/apache/tvm/pull/17257.

Lunderberg:
Fixed last two failing unit tests, since `MakePackedAPI` no longer needs to conditional load/convert a boolean to `int64`.

tqchen (approved):

PR #17241, opened 2024-08-05
[Draft][FFI] Remove backwards-compatibility unwrapping of IntImm

Lunderberg:
This is a follow-up PR to https://github.com/apache/tvm/pull/16183, which updated the FFI with explicit integer types.  As part of that change, many internal functions were updated to accept non-IR types (e.g. `Array<runtime::Int>` instead of `Array<IntImm>`).  For backwards compatibility with callees that provided the IR types, a specialization of `PackedFuncValueConverter` unwrapped the `IntImm` into a `runtime::Int`.

This commit removes the backwards-compatibility specialization of `PackedFuncValueConverter`.  Breakages that are found in CI as a result will then be updated at the caller side, removing the need for the backwards-compatibility handler altogether.

### DIFF
--- a/include/tvm/ir/expr.h
+++ b/include/tvm/ir/expr.h
@@ -866,28 +866,6 @@ struct PackedFuncValueConverter<tvm::FloatImm> {
   }
 };
 
-/* \brief Backwards compatibility wrapper for IntImm arguments
- *
- * In previous versions of TVM, IntImm was the default FFI type for
- * integer arguments, instead of runtime::Int.  For backwards
- * compatibility where the callee has been updated to expected a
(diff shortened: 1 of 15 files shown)


tqchen:
would be great to followup on this one

Lunderberg:
Agreed, though unfortunately I haven't had time to do so.  It looks like the current failures are in the relay/collage C++ testing, but I haven't looking into it.

tqchen:
Given these are legacy. I think we can remove those tests to move forward

tqchen:
would be great to check the remaining errors and get this in

tqchen:
superseded by #17920

PR #17246, opened 2024-08-06
[Cleanup] Remove `using namespace tvm::runtime` from headers

Lunderberg:
Prior to this commit, various header files had `using namespace tvm::runtime`, which imports all names from `tvm::runtime` into the current namespace.

These imports can cause compilation errors depending on the order of `#include` statements.  For example, the `#include <tvm/relay/attrs/transform.h>` file uses the unqualified name `Bool` to refer to `::tvm::Bool`, a subclass of `PrimExpr`.  If a different header file specifies `using namespace tvm::runtime` within the `tvm::relay` namespace, then the unqualified name `Bool` ambiguously refers to either `::tvm::Bool` or `::tvm::runtime::Bool`.

In MSVC, this can cause even further compilation errors.  By default, MSVC does not follow the C++ standard for name resolution in templates.  The standard requires that any names in a template that do not depend on template parameters be resolved when the template is declared.  However, MSVC instead resolves these names when the template is instantiated.  As a result, the same `using namespace tvm::runtime` may cause a compilation error if it occurs after the template's declaration, but before the template's usage.

(TVM provides the `/permissive-` flag to MSVC builds specifically to disable MSVC's non-standard name resolution, so this only impacts downstream forks that disable this flag.  See https://github.com/apache/tvm/pull/16343 for more details.)

This commit removes `using namespace tvm::runtime`, replacing them with explicit `using tvm::runtime::SOME_SPECIFIC_SYMBOL` where necessary.  This resolves both the include-order dependency for standards-compliant compilers, and the compilation errors for MSVC's default build.

### DIFF
--- a/src/contrib/msc/core/ir/graph_builder.h
+++ b/src/contrib/msc/core/ir/graph_builder.h
@@ -51,7 +51,8 @@ namespace msc {
 using Expr = tvm::RelayExpr;
 using RelaxExprVisitor = tvm::relax::ExprVisitor;
 using RelayExprVisitor = tvm::relay::ExprVisitor;
-using namespace tvm::runtime;
+
+using tvm::runtime::NDArray;
 
 /*!
  * \brief Config for building MSCGraph.
--- a/src/relay/backend/vm/compiler.h
+++ b/src/relay/backend/vm/compiler.h
@@ -51,7 +51,8 @@ namespace tvm {
 namespace relay {
 namespace vm {
 
-using namespace tvm::runtime;
+using tvm::runtime::ModulePropertyMask;
+using tvm::runtime::NDArray;
 using namespace tvm::runtime::vm;
 using namespace relay::transform;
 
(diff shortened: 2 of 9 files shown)


tqchen (approved):

tqchen (changes requested):
temp block this in light of https://github.com/apache/tvm/pull/16183#issuecomment-2273212033

Lunderberg:
> temp block this in light of [#16183 (comment)](https://github.com/apache/tvm/pull/16183#issuecomment-2273212033)

Good point.  I think this change is beneficial regardless of #16183, but the FFI changes touch so many parts of TVM that I wouldn't want to declare them 100% independent.

Lunderberg:
Rebased these changes onto main, and this PR should be land-able now that the performance regressions from #16183 have been resolved.

tqchen (approved):

PR #17250, opened 2024-08-06
[WebGPU] Fix unexpected device lost error when intentional dispose

CharlieFRuan:
This PR fixes an issue introduced in https://github.com/apache/tvm/pull/17005. `deviceLostIsError` was introduced to make sure intentional `dispose()` (hence webgpu's `destroy()`) does not cause device lost callback to treat it as an error. However, we cannot set `deviceLostIsError` immediately to true after calling `this.lib.dispose()` (which calls `device.destroy()`) because WebGPU is asynchronous. Otherwise, we would trigger the device lost callback when calling `Instance.dispose()` to destroy device intentionally.

### DIFF
--- a/web/src/runtime.ts
+++ b/web/src/runtime.ts
@@ -1122,7 +1122,7 @@ export class Instance implements Disposable {
     // ctx release goes back into lib.
     this.ctx.dispose();
     this.lib.dispose();
-    this.deviceLostIsError = true;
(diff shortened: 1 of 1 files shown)


CharlieFRuan:
@tvm-bot rerun

CharlieFRuan:
@tvm-bot rerun

tqchen (approved):

PR #17251, opened 2024-08-07
[Web] Add TVMArgBool to ArgTypeCode

CharlieFRuan:
In a recent PR https://github.com/apache/tvm/pull/16183, `TVMArgTypeCode` introduced `TVMArgBool`. This PR is needed for the web runtime when a TVM packed function returns a bool.

### DIFF
--- a/web/src/runtime.ts
+++ b/web/src/runtime.ts
@@ -2473,6 +2473,7 @@ export class Instance implements Disposable {
     switch (tcode) {
       case ArgTypeCode.Int:
       case ArgTypeCode.UInt:
+      case ArgTypeCode.TVMArgBool:
         return this.memory.loadI64(rvaluePtr);
       case ArgTypeCode.Float:
         return this.memory.loadF64(rvaluePtr);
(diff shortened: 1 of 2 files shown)


web/src/runtime.ts
```diff
@@ -2473,6 +2473,7 @@ export class Instance implements Disposable {
     switch (tcode) {
       case ArgTypeCode.Int:
       case ArgTypeCode.UInt:
+      case ArgTypeCode.TVMArgBool:
         return this.memory.loadI64(rvaluePtr);
```
CharlieFRuan, line 2477: cc @tqchen, would `loadI64()` be the right function to use for bool?

tqchen:
temp block this in light of https://github.com/apache/tvm/pull/16183#issuecomment-2273212033

web/src/runtime.ts
tqchen: this would need to wait until https://github.com/apache/tvm/pull/17240, but once the new bool type lands yes this would be the case

Lunderberg:
#17240 has now landed, so this PR should be able to be revisited.

tqchen (approved):

PR #17252, opened 2024-08-07
Revert "[FFI][RUNTIME] Introduce runtime boxed types for int/float/bool"

tqchen:
Reverts apache/tvm#16183

see https://github.com/apache/tvm/pull/16183#issuecomment-2273212033 for context


Lunderberg (approved):
Revert approved, and crossing fingers that resolving the performance overhead won't be too onerous.

PR #17257, opened 2024-08-08
 [FFI] Re-introduce the boxed primitive values

Lunderberg:
Initially introduced in https://github.com/apache/tvm/pull/16183, these changes were reverted in https://github.com/apache/tvm/pull/17252 due to performance degredation in some Relax models.  This could occur when a model contained a large number of calls to `"vm.builtin.tuple_getitem"`, which may occur when model weights are provided as a tuple.

This PR re-applies the changes from https://github.com/apache/tvm/pull/16183, but with the performance degredation resolved.  The root cause was unnecessary type-checking when converting from an untyped `tvm::ArrayNode*` to the typed `tvm::Array<T>`, in the case where `T` is `ObjectRef`.

### DIFF
--- a/include/tvm/ir/attrs.h
+++ b/include/tvm/ir/attrs.h
@@ -265,7 +265,16 @@ class DictAttrs : public Attrs {
 
     auto it = node->dict.find(attr_key);
     if (it != node->dict.end()) {
-      return Downcast<Optional<TObjectRef>>((*it).second);
+      // For backwards compatibility, return through TVMRetValue.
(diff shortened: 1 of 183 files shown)


tqchen:
cc @vinx13 , would be great to confirm LLM perf before we merge this

vinx13:
Confirmed the performance is good

Lunderberg:
Thank you on the confirmation!  With how many parts could be impacted by this change, I very much appreciate the QC check.

PR #17260, opened 2024-08-09
[WINDOWS] Compiler options for non x86 targets

krishnaraj36:
Windows compiler options for targets

### DIFF
--- a/python/tvm/contrib/cc.py
+++ b/python/tvm/contrib/cc.py
@@ -92,8 +92,8 @@ def create_shared(output, objects, options=None, cc=None, cwd=None, ccache_env=N
 
     if _is_linux_like():
         _linux_compile(output, objects, options, cc, cwd, ccache_env, compile_shared=True)
-    elif _is__like():
-        __compile(output, objects, options, cwd, ccache_env)
+    elif _is_windows_like():
+        _windows_compile(output, objects, options, cwd, ccache_env)
     else:
         raise ValueError("Unsupported platform")
 
@@ -173,8 +173,8 @@ def create_executable(output, objects, options=None, cc=None, cwd=None, ccache_e
 
     if _is_linux_like():
         _linux_compile(output, objects, options, cc, cwd, ccache_env)
-    elif _is__like():
-        __compile(output, objects, options, cwd, ccache_env)
+    elif _is_windows_like():
+        _windows_compile(output, objects, options, cwd, ccache_env)
     else:
         raise ValueError("Unsupported platform")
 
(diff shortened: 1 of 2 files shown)


krishnaraj36:
@srkreddy1238  Can you please look into this PR

python/tvm/contrib/cc.py
```diff
@@ -373,7 +373,11 @@ def _linux_compile(
 
 def _windows_compile(output, objects, options, cwd=None, ccache_env=None):
     cmd = ["clang"]
+    compiler = os.getenv("TVM_WIN_CC", default="clang")
+    win_target = os.getenv("TVM_WIN_TARGET", default="x86_64")
+    cmd = [compiler]
```
cbalint13, line 378: This overwrite initial ```cmd = [clang]``` initialization, in a counterintuitive logic, could we condition the new assignment here to ```TVM_WIN_TARGET``` presence ?

python/tvm/contrib/cc.py
```diff
@@ -373,7 +373,11 @@ def _linux_compile(
...
     cmd += ["-O2"]
+    cmd += ["--target=" + win_target]
```
cbalint13, line 380: Let's also condition these too within the above comment.

cbalint13 (changes requested):
Thank you @krishnaraj36 for the contribution !

Looks good to me, lets have the assignment logic and the CI lint issue fixed in order to get this merged.

python/tvm/contrib/cc.py
cbalint13: Also lets fix ```cwd=None``` typo here, perhaps we could have ```cmd=[clang]``` (if you would like better) and then just override with windows env related conditional.

krishnaraj36: @cbalint13   make sense, we don't require this assignment as we are consuming from environment variables.

srkreddy1238 (approved):
LGTM, let the CI pass.

python/tvm/contrib/cc.py
cbalint13: This is fine too (removing), I am sure the original would worked but its logic was strange. 

Thank you !

 Apollogise for confusion of ```cwd``` vs. ```cmd``` function parameter, it is not the case to involve that too.

cbalint13 (approved):

python/tvm/contrib/cc.py
```diff
@@ -173,8 +173,8 @@ def create_executable(output, objects, options=None, cc=None, cwd=None, ccache_e
 
     if _is_linux_like():
         _linux_compile(output, objects, options, cc, cwd, ccache_env)
-    elif _is_windows_like():
-        _windows_compile(output, objects, options, cwd, ccache_env)
+    elif _is__like():
+        __compile(output, objects, options, cwd, ccache_env)
```
cbalint13, line 177: Does changes to ```_wimdows_compile()``` still make sense since now it is switched to ```_compile()```  ?

PR #17265, opened 2024-08-10
[CompileBugfix][contrib] meet 'base64.h: No such file or directory' and '‘tvm::runtime::vm::AllocatorType’ has not been declared' while compiling

DDDVE:
1. modify the relative path of base64.h
2. modify namespace of base64 and AllocatorType


DDDVE:
By the way, I encountered this bug when I was compiling on Ubuntu 20.04 with cmake 3.22.1, clang 10.0.0

tqchen (approved):

tqchen:
Thanks @DDDVE

PR #17271, opened 2024-08-12
[CI] Resolve CI compilation failures on MacOSX

Lunderberg:
The build on main failed [link](https://github.com/apache/tvm/actions/runs/10351996240/job/28661217963) with the following error message.  This first occurred in the post-merge build of #17257, but the failure mode (error message below) doesn't appear related to the changes in that commit.

```
conda.exceptions.CondaValueError: You have chosen a non-default solver backend (libmamba) but it was not recognized.
```

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -33,3 +33,5 @@ runs:
     run: |
       conda info
       conda list
+      conda info --envs
+      conda list --name base
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -15,7 +15,7 @@ runs:
       channel-priority: strict
       environment-file: conda/build-environment.yaml
       auto-activate-base: false
-      auto-update-conda: true
+      conda-solver: classic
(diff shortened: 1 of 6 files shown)


Lunderberg:
After investing, it looks like the failures were due to `conda` attempting to use a dependency solver that wasn't actually installed in the `conda` environment.  This seems to result from some mix of the following:

1) The most recent release of `conda-build` now defaults to using `libmamba` for dependency solving.
2) The current `conda-build` only requires `conda >= 23.7.0`
3) The `libmamba` library is not added as a dependency  until `conda >= 23.10`.
4) Some combination of the python3 versions and the cmake versions that are pinned in TVM requires that `conda < 23.10`.
5) The configuration for `conda-incubator/setup-miniconda@v3` applies the default value of `conda-solver` based on the conda version in the base environment.  This is then specified as an explicit config to be used by the conda version in the `tvm-build` environment.

If we explicitly specify `conda-solver: classic`, then we avoid using the missing `libmamba` at all.

tqchen (approved):

PR #17274, opened 2024-08-14
[TOPI][ADRENO] Add Group Conv2d texture schedule

krishnaraj36:
Implemented the adreno schedule for group conv2d op with texture enablement.

### DIFF
--- a/python/tvm/topi/adreno/group_conv2d_nchw.py
+++ b/python/tvm/topi/adreno/group_conv2d_nchw.py
@@ -130,10 +130,11 @@ def group_conv2d_nchwc(cfg, Input, Filter, stride, padding, dilation, out_dtype)
     else:
         out_channel_chunks, in_filter_channels, kernel_h, kernel_w, out_channel_block = Filter.shape
         out_channles = out_channel_chunks * out_channel_block
-    
+
     assert in_channels % in_filter_channels == 0
     groups = in_channels // in_filter_channels
     in_group_channels = in_filter_channels
+    assert out_channles % groups == 0
     out_group_channels = out_channles // groups
 
     out_height_orig, out_height, out_width_orig, out_width = expand_spatial_dimensions(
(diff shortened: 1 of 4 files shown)


python/tvm/topi/adreno/group_conv2d_nchw.py
```diff
@@ -0,0 +1,386 @@
...
+    )
+    cfg.define_split(
+        "tile_x",
+        x,
+        num_outputs=3,
+        filter=lambda entity: entity.size[1] <= 8 and entity.size[2] <= 16,
+    )
+
+    cfg.define_split("tile_rcc", rcc, num_outputs=2)
+    cfg.define_split("tile_ry", ry, num_outputs=2)
```
tqchen, line 279: Thanks @krishnaraj36 , as we are moving towards meta-schedule and relax, we are moving away from legacy pipelines in updated flows, would be great to bring the new schedules under TensorIR, and dlight pipeline instead

tqchen (changes requested):

python/tvm/topi/adreno/group_conv2d_nchw.py
krishnaraj36: @tqchen  - So that mean relay legacy pipeline will be deprecated ?
@srkreddy1238 FYI

tqchen: ideally we would like to move toward the relax-based pipeline,  that would help us to leverage the latest features also simplify some of the developments

tqchen (approved):

python/tvm/topi/adreno/group_conv2d_nchw.py
tqchen: I am merging in this one for now, but would be good to do gthe transition for future ones :)

PR #17277, opened 2024-08-16
[Relay][Pytorch] Add support for `aten::tile`

mshr-h:
Support [torch.tile](https://pytorch.org/docs/stable/generated/torch.tile.html).
cc @vinx13 @yongwww


tqchen:
Thanks @mshr-h , we are moving more towards relax based pipeline whose tests won't depends on running the code, which has increasing pressure on the ci. Would be great to bring support for relax

tqchen (approved):

PR #17278, opened 2024-08-16
[Codegen] Emit `tir::Let` as var assignment explicitly

MasterJH5574:
Prior to this PR, the PrimExpr `tir::Let` is treated as inlining during codegen, which makes any common subexpression elimination (CSE) efforts using `tir::Let` at TIR level effectless.

This PR updates codegen so that the `tir::Let` will have an explicit var assignment and thus can effectively reflect the CSE efforts.


tqchen:
@tvm-bot rerun

tqchen (approved):

Lunderberg:
I'm a little bit late on the review, but I like the fix, thank you!  I double-checked `codegen_llvm.cc`, and it tracks `tir::Let` expressions by their `llvm::Value*`, and wouldn't be susceptible to the same issue.

PR #17285, opened 2024-08-19
[Relax] Require correct input/output shapes `R.call_tir`

Lunderberg:
Prior to this commit, the Relax well-formed checker validated arguments provided to Relax functions, but did not validate arguments provided to `R.call_tir`.  As a result, incorrect arguments from Relax to TIR would not be checked until runtime, if at all.

This commit updates the well-formed checker to verify that `R.call_tir` has received the correct arguments, and has the correct output shape specified in the `out_sinfo` parameter.

### DIFF
--- a/include/tvm/relax/op_attr_types.h
+++ b/include/tvm/relax/op_attr_types.h
@@ -56,12 +56,39 @@ using FCallPacked = String;
  * expressed in multiple syntactically valid and semantically
  * equivalent forms, to normalize to a single representation.
  *
+ * Note: `FNormalize` is applied for each expression as part of the
+ *    `relax::BlockBuilder`.  While operator-specific validation may
+ *    be performed within the `FNormalize` implementation, ensuring
+ *    that errors are caught as early as possible, this should only be
+ *    used when validation is fast to apply.  If the validation logic
+ *    may be slow, it should instead be implemented in `FValidate`,
+ *    which is only run as part of the well-formed checker.
+ *
  * \param bb The BlockBuilder context.
  *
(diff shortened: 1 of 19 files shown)


Lunderberg:
The implementation in this PR is similar to https://github.com/apache/tvm/pull/17216, but with a few key changes:

* Unlike #17216, this PR raises an error from the well-formed checker if the shape or dtype of `R.call_tir` arguments mismatch.
* Unlike #17216, this PR still requires a user to explicitly provide the `out_sinfo` argument to `R.call_tir`.

Lunderberg:
All CI tests are passing, with the exception of the `CI / Windows` step, which has the same failures as seen across all PRs (currently being debugged in https://github.com/apache/tvm/pull/17283).  Since the `CI / Windows` is not a required step, this PR is ready for review/merge.

Lunderberg:
@tvm-bot rerun

tqchen (approved):

Lunderberg:
Looks like the failure on `hexagon/pr-head` is due to a timeout, not an error introduced in this PR.  Of the 8 shards that run Hexagon unit tests, 7 of them finish in about 1h20m.  The 8th node looks like it's being killed after 2 hours, and spends about 50 minutes running `tests/python/contrib/test_hexagon/test_run_unit_tests.py::test_run_unit_tests[]`.

https://github.com/apache/tvm/pull/17334 should resolve the hexagon CI issues by breaking up the single gigantic test, allowing them to be sharded out.

Lunderberg:
@tvm-bot rerun

Since #17334 has landed, the hexagon tests should be better spread out across the 8 shards.

PR #17286, opened 2024-08-20
[Docs] Introduce Relax API and move legacy part to standalone page

Hzfengsy:
As the TVM project evolves, the Unity strategy has been the recommended way to use Apache TVM applications. Hence, we are pushing documentation for the Relax API to the forefront and moving the legacy part to a standalone page, which may be removed in the future.

cc @tqchen


Hzfengsy:
an online preview at http://8.130.134.126:58000

quic-sanirudh:
Hi, this is great, thanks for this. 

One thing I noticed was that the **Design and Architecture** section has also gone into legacy part. This contains some very useful documentation about the TVM Runtime, PackedFunc, TVM Object stack, etc. which could be very useful for beginners. 

It also contains relay specific details which is probably legacy, but perhaps we could clean that up and move the runtime specific details out of legacy part? It could be done as a follow-up task as well, but I just noticed that and thought it might be useful to retain those docs outside legacy part.

tqchen:
thanks @quic-sanirudh  for great suggestion, yes, i think we can followup to bring up some of the FFI and Object systems

Hzfengsy:
@quic-sanirudh Thanks for the great suggestion. I agree that some parts are still valuable. However, the current doc is coupled with Relay, and it is not easy to clean it out. I will get them back in a following PR.

PR #17294, opened 2024-08-22
[Runtime] Support KV cache with RoPE extension factor array

MasterJH5574:
This PR enhances the KV cache with the RoPE extensio factor support. With this PR, the KV cache can support models like Phi3.5 which comes with the extension factor.


tqchen (approved):

tqchen (approved):

MasterJH5574:
@tvm-bot rerun

PR #17293, opened 2024-08-22
[Rust] Remove mxnet dependency and re-enable rust example

mshr-h:
Replace mxnet's resnet18 with torchvision.
Clean up import statements in `rust/tvm/examples/resnet/src/build_resnet.py`.
Related PR and Issue:
- #16546 
- #16572 
- #16547 

cc: @Hzfengsy @tqchen @cgerum

Issue #16547, opened 2024-02-08
[Tracking Issue] Remove MXNet Dependency


tqchen:
MXNet has been archived and likely future versions of tvm should not depend on it. This is a tracking issue for us to remove the related stale testcases and documentations that depends on mxnet

Lunderberg:
I have a rough list of the impacted files that will need to be updated, from putting together #16546

- [ ] Depend on mxnet's AWS
  - [ ] `gallery/how_to/compile_models/from_mxnet.py`
  - [ ] `gallery/how_to/deploy_models/deploy_model_on_nano.py`
  - [ ] `gallery/how_to/deploy_models/deploy_model_on_rasp.py`
  - [ ] `gallery/how_to/deploy_models/deploy_quantized.py`
  - [ ] `gallery/how_to/extend_tvm/bring_your_own_datatypes.py`
  - [ ] `gallery/how_to/tune_with_autoscheduler/tune_network_arm.py`
  - [ ] `gallery/how_to/tune_with_autoscheduler/tune_network_cuda.py`
  - [ ] `gallery/how_to/tune_with_autoscheduler/tune_network_mali.py`
  - [ ] `gallery/how_to/tune_with_autoscheduler/tune_network_x86.py`
  - [ ] `rust/tvm/examples/resnet`
  - [ ] `tests/python/frontend/mxnet/test_forward.py`
  - [ ] `vta/scripts/tune_resnet.py`
  - [ ] `vta/tutorials/autotvm/tune_alu_vta.py`
  - [ ] `vta/tutorials/autotvm/tune_relay_vta.py`
  - [ ] `vta/tutorials/frontend/deploy_classification.py`
  
  
- [ ] Depend on mxnet's AWS (commented out for CI)
  - [ ] `gallary/how_to/tune_with_autoscheduler/tune_network_cuda.py`
  - [ ] `gallary/how_to/tune_with_autoscheduler/tune_network_mali.py`
  - [ ] `gallary/how_to/tune_with_autoscheduler/tune_sparse_x86.py`
  - [ ] `gallary/how_to/tune_with_autoscheduler/tune_network_arm.py`
  - [ ] `gallary/how_to/tune_with_autoscheduler/tune_conv2d_layer_cuda.py`
  - [ ] `gallary/how_to/tune_with_autoscheduler/tune_conv2d_layer_cuda.py`
  - [ ] `gallary/how_to/tune_with_autoscheduler/tune_network_x86.py`
[...]

PR #17297, opened 2024-08-23
[Apps] Remove mxnet dependency from /apps/android_camera/models

mshr-h:
As per title.
I'll leave mobilenet_v2 part as it is but we could remove keras dependency since torchvision provides a mobilenet_v2 pretrained model. Not sure which is better.

Related issue and pr:
- #16547 
- #17293 

cc: @Hzfengsy @tqchen


mshr-h:
@tvm-bot rerun

Hzfengsy (approved):

PR #17305, opened 2024-08-27
[Python][Relax] Rotary positional embedding scaling

MasterJH5574:
This PR introduces two styles of RoPE scaling: the llama3 style and the longrope scale.


MasterJH5574:
cc @vinx13 @tqchen

tqchen (approved):

Issue #17200, opened 2024-07-26
[Bug] [Relax] InternalError: Check failed: (*it).second == var


MellowArtisan:
### Actual behavior
```
Traceback (most recent call last):
  File "/share_container/optfuzz/res/bugs/18_check.py", line 35, in <module>
    mod = tvm.transform.Sequential([relax.transform.LiftTransformParams(), relax.transform.LiftTransformParams()])(mod)  # crash here
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/software/tvm/python/tvm/ir/transform.py", line 238, in __call__
    return _ffi_transform_api.RunPass(self, mod)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/software/tvm/python/tvm/_ffi/_ctypes/packed_func.py", line 239, in __call__
    raise_last_ffi_error()
  File "/software/tvm/python/tvm/_ffi/base.py", line 481, in raise_last_ffi_error
    raise py_err
tvm.error.InternalError: Traceback (most recent call last):
  11: tvm::runtime::PackedFuncObj::Extractor<tvm::runtime::PackedFuncSubObj<tvm::runtime::TypedPackedFunc<tvm::IRModule (tvm::transform::Pass, tvm::IRModule)>::AssignTypedLambda<tvm::transform::{lambda(tvm::transform::Pass, tvm::IRModule)#7}>(tvm::transform::{lambda(tvm::transform::Pass, tvm::IRModule)#7}, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >)::{lambda(tvm::runtime::TVMArgs const&, tvm::runtime::TVMRetValue*)#1}> >::Call(tvm::runtime::PackedFuncObj const*, tvm::runtime::TVMArgs, tvm::runtime::TVMRetValue*)
  10: tvm::transform::Pass::operator()(tvm::IRModule) const
  9: tvm::transform::Pass::operator()(tvm::IRModule, tvm::transform::PassContext const&) const
  8: tvm::transform::SequentialNode::operator()(tvm::IRModule, tvm::transform::PassContext const&) const
  7: tvm::transform::Pass::operator()(tvm::IRModule, tvm::transform::PassContext const&) const
[...]

PR #17314, opened 2024-08-28
[Relax][Transform] Compose preproc functions in LiftTransformParams 

Lunderberg:
The `LiftTransformParams` pass produces additional functions, either named `$FOO_transform_params` when generating one transformation function per inference function, or `transform_params` when generating a single shared transformation function.  Prior to this commit, if the `IRModule` already contained a function with that name, an error would be raised.

After this commit, the `LiftTransformParams` pass will instead check for existing functions, and compose the previous transformation function with the newly-lifted transformation.  This allows `LiftTransformParams` to be used alongside a hand-written parameter transformation.

Closes https://github.com/apache/tvm/issues/17200

### DIFF
--- a/src/relax/transform/lift_transform_params.cc
+++ b/src/relax/transform/lift_transform_params.cc
@@ -119,7 +119,10 @@ struct BaseCollectInfo {
     Function func(params, body, GetStructInfo(tuple_var));
     func = WithAttr(func, attr::kNumInput, Integer(0));
     func = CopyWithNewVars(func);
+    func = BundleModelParams(func);
     func = Downcast<Function>(CanonicalizeBindings(func));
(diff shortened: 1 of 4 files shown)


Lunderberg:
This PR relies on bugfixes implemented in https://github.com/apache/tvm/pull/17312 and https://github.com/apache/tvm/pull/17313.  The PR branch contains these bugfixes in its history, to allow CI to run, but will need to be rebased after these per-requisites land in order to maintain a linear commit history on main.

This PR is currently marked as a draft until the pre-requisites land.

Lunderberg:
The pre-requisite PRs have landed, this PR is rebased on top of `main`, and is now ready for review.

vinx13 (approved):

PR #17321, opened 2024-08-30
[Target] Refine equality check on TargetKind instances

wrongtest-intellif:
The change propose to ensure the TargetKind instances to be singleton on serialization.

We encounter the issue when we try to establish distributed tools to accelerate the lowering. Though `TargetKind` are generally singleton instances for each registered kind, the workload serialization would also create new `TargetKind` instances (which mostly lay in `kTarget` func attr). 

One simple test workflow is like below:
```python
target = Target("llvm")
mod = tvm.IRModule({"main": func})

# in practice, module dict is collected from many lowerred workloads, with target deserialized.
target2 = tvm.ir.load_json(tvm.ir.save_json(target))
lib = tvm.build({target2: mod}, target_host=target)
lib["func"]()  # oops, no "func" found in the host module.
```

### DIFF
--- a/src/target/target_kind.cc
+++ b/src/target/target_kind.cc
@@ -35,7 +35,20 @@
 
 namespace tvm {
 
-TVM_REGISTER_NODE_TYPE(TargetKindNode);
+// helper to get internal dev function in objectref.
+struct TargetKind2ObjectPtr : public ObjectRef {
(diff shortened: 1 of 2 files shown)


wrongtest-intellif:
@tqchen Hi, could you kindly take a look at the change?  `CI / Windows (pull_request)` seems to be broken but I can't find why.

tqchen (approved):

tqchen:
thank you @wrongtest-intellif !

PR #17323, opened 2024-08-30
[Relax] Easier use of Function.bind_params with relax.op.create.*

Lunderberg:
The `Function.bind_params` method requires all bound values to have `StructInfo` annotations.  While these are generated on construction for `relax::Constant` and `relax::PrimValue`, the most common cases of bound values, other useful expression such as `relax.op.zeros(shape,dtype)` do not have `StructInfo` until they are normalized.  In addition, because `Function.bind_params` may be used in contexts that do not have a `relax.BlockBuilder` available, performing this normalization is not straightforward.

This commit updates the constructors for Relax operations within the `tvm.relax.op.create` namespace to infer their `StructInfo` on construction, if it is possible to do so.  This allows these operations to be used inside `Function.bind_params`.


Lunderberg:
Rebase onto main, since the CI failures were caused by https://github.com/apache/tvm/pull/17371.

PR #17326, opened 2024-09-02
[KVCache] Add tree attention with paged cache support

vinx13:
This adds support for tree attention with tokens in the kv cache. KVCache is now stateful w.r.t the tree state, multiple decoding rounds can be performed with tree mask and the accepted tokens can be committed at once at the end.

cc @MasterJH5574 @tqchen


MasterJH5574 (approved):
Thanks for the enhancement!

PR #17332, opened 2024-09-03
[Relax] Validate StructInfo of variable bindings

Lunderberg:
In Relax, both the variable and the expression in a `VarBinding` may contain `StructInfo` annotations.  Prior to this commit, these `StructInfo` annotations could be inconsistent, assigning an expression to a variable of incompatible type.

This commit updates the Relax well-formed checker to verify that the `StructInfo` of Relax variables accurately describes their contents.

### DIFF
--- a/src/relax/analysis/struct_info_analysis.cc
+++ b/src/relax/analysis/struct_info_analysis.cc
@@ -982,10 +982,25 @@ class StructInfoLCAFinder
   StructInfo VisitStructInfo_(const PrimStructInfoNode* lhs, const StructInfo& other) final {
     auto* rhs = other.as<PrimStructInfoNode>();
     if (rhs == nullptr) return ObjectStructInfo(lhs->span);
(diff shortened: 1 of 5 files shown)


Lunderberg:
This validation exposed a bug in `StructInfoLCA`, which only inspected the datatype of `PrimStructInfo` and not the known value.  This bug was introduced in https://github.com/apache/tvm/pull/15577, which added tracking fo known `PrimStructInfo`, but didn't update `StructInfoLCA`.  This bug would result in incorrect known value annotations in the `StructInfo`, after a conditional statement.  Since these annotations are later used to simplify known expressions, this bug could have resulted in incorrect simplifications.

The bug in `StructInfoLCA` is fixed as part of this PR.

Lunderberg:
Rebased onto main to resolve conflicts in `test_analysis_well_formed.py`.

tqchen:
cc @yongwww

yongwww (approved):
Looks good to me!

PR #17334, opened 2024-09-04
[CI][Hexagon] Forward gtest tests into pytest as separate tests

Lunderberg:
Prior to this commit, all Hexagon test cases in `tests/cpp-runtime/hexagon` were executed as part of a single unit test in pytest.  This can take a significant portion of the total timeout in CI (~50 minutes out of a 2-hour timeout).  While the hexagon tests are split out onto 8 separate runners, having a single large test can cause timeouts on whichever runner happens to receive it.

This commit exposes each unit test from `tests/cpp-runtime/hexagon` into a separate unit test in pytest, to avoid these timeouts.

### DIFF
--- a/tests/python/contrib/test_hexagon/test_run_unit_tests.py
+++ b/tests/python/contrib/test_hexagon/test_run_unit_tests.py
@@ -15,6 +15,8 @@
 # specific language governing permissions and limitations
 # under the License.
 
+# pylint: disable=redefined-outer-name
+
 """capture gtest output and return over FFI"""
 
 import tvm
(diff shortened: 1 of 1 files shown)


quic-sanirudh (approved):
Nice idea to use the gtest filter to split these tests. Thanks for this.

Lunderberg:
Thank you!  It looks like the hexagon tests now have a narrower spread in runtimes across the shards, though there's still some variation (about 1h20m to 1h50m), probably depending on which shard happens to get the long runtime test cases.

For now, hopefully this will help avoid having timeouts during CI.

PR #17339, opened 2024-09-05
[relay][qnn]: Fix qnn.avg_pool2d layout inference

f2013519:
**Issue:**
Currently the qnn.avg_pool2d operator does not take the layout specified with the Convert Layout pass.
 
**Cause:**
The layout of the zero point and scale tensors are specified as "undefined". This causes the Convert Layout pass to bail and maintain the original layout.

**Fix:**
The fix updates the layouts for the scale and zero point tensors to be "channel" layout, similar to how it is done for qnn.conv2d.

**Testing:**
Added unit test


f2013519:
@masahi @kparzysz-quic @quic-sanirudh 

PTAL, thanks

cbalint13 (approved):
LGTM

cbalint13:
Thanks @f2013519 for the fix !

PR #17349, opened 2024-09-09
Fix the connection between server and tracker when using WSL2

gaoren002:
There is a bug when we use WSL2 or other virtual machine in NAT mode as host, the host can find the device easily, but when device respond to the host, it will lose the device ip address, in this case, host will just try to find a IP that is used for visiting from virtual machine(e,g, WSL2, 172.xxx.xxx.xxx) to physical host, this IP cannot be used to visit device, this will make the server in device  print"no incoming connections ", because the host cannot find the device

To fix the bug, I just pass the real IP of device to host,which can solve this problem. But this fix need Internet to get device IP, I wonder if there is a better way. 
image

### DIFF
--- a/python/tvm/exec/rpc_server.py
+++ b/python/tvm/exec/rpc_server.py
@@ -19,6 +19,21 @@
 import argparse
 import logging
 from .. import rpc
+import socket
+
+
+def get_local_ip():
+    try:
+        # create UDP socket
+        s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+        # connect to a outer server, but we don't send data
+        s.connect(("8.8.8.8", 80))
+        # get local ip
+        local_ip = s.getsockname()[0]
+        s.close()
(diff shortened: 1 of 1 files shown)


gaoren002:
@tvm-bot re-run

gaoren002:
https://discuss.tvm.apache.org/t/lots-of-key-mismatches-when-autotuning-with-rpc/3180
This bug remains for a long time, can anyone help to fix it?
@tqchen

tqchen:
@Hzfengsy  do u mind take a look?

Hzfengsy:
Thanks for pinging me. However, I'm not familiar with Windows and WSL. I'm afraid that I'm not able to review this one

gaoren002:
> Thanks for pinging me. However, I'm not familiar with Windows and WSL. I'm afraid that I'm not able to review this one

actually Any VM could encouter this problem

PR #17350, opened 2024-09-09
[Relax][Frontend][Onnx] fix params name bug in onnx frontend

HongHongHongL:
1. In onnx_frontend.py, for parameters with names starting with "onnx::", a strip processing is performed. In self._params, they should be stored as the new var_name.
```python
if self._keep_params_in_input:
    # Pytorch sometimes inserts silly weight prefix. Remove it.
    var_name = init_tensor.name.strip("onnx::")
    init_var = self._new_var(var_name, shape=array.shape, dtype=array.dtype)
    self._nodes[init_tensor.name] = init_var
    # We need to keep track of both the real value and variable for this variable.
    self._params[init_tensor.name] = (init_var, array)
```
2. In ONNX models, a param can be used many times. As a result, we should not use pop in get_constant.
```python
# Params is actually both the graph nodes and param dictionary, unpack them.
graph_nodes, params = params
# Convert if possible
if isinstance(var, relax.Var) and var.name_hint in params:
    # When converting a parameter to a constant, update references to it as well.
    _, value = params.pop(var.name_hint)
    const_value = relax.const(value)
    graph_nodes[var.name_hint] = const_value
    return const_value
```

### DIFF
--- a/python/tvm/relax/frontend/onnx/onnx_frontend.py
+++ b/python/tvm/relax/frontend/onnx/onnx_frontend.py
@@ -91,7 +91,7 @@ def get_constant(
     # Convert if possible
     if isinstance(var, relax.Var) and var.name_hint in params:
         # When converting a parameter to a constant, update references to it as well.
-        _, value = params.pop(var.name_hint)
+        _, value = params[var.name_hint]
         const_value = relax.const(value)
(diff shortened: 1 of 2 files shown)


Hzfengsy:
Please add a regression test, thank!

yongwww (approved):

PR #17360, opened 2024-09-10
[FIX] fix bug when normalize iter with different lower bounds

Karbit:
If an iter has been normalized with a lower bound, and then try to normalize with a new lower bound, the iter_min need to be updated only when the new lower bound is smaller than the original one.


tqchen (approved):

tqchen:
thanks @Karbit

PR #17363, opened 2024-09-11
[DLight] Fix Matmul rule for Conv3D

Hzfengsy:
Currently, the matmul rule for Conv3D is incorrect, due to the incorrect reindexing of the input tensor. This commit fixes the issue by correctly

The `index map` of `transform_layout` should be calculated after the `reindex` process

cc @MasterJH5574


tqchen:
Thanks @Hzfengsy , please fix lint

tqchen (approved):

PR #17373, opened 2024-09-13
[TVMScript] Avoid segfault from invalid TVMScript

Lunderberg:
Prior to this commit, after the `DiagnosticContext` prints its error, it overwrites the `DiagnosticRenderer` with a NULL renderer.  If a second call to `DiagnosticContext::Render` occurs, it will segfault. This appears to be intended to prevent double-printing of error messages, but double-printing error messages is much worse than a segfault.

In addition, `DiagnosticContext::Render` should only be called once. There's a common pattern in the parser where it will wrap exceptions in `DiagnosticError`, but re-raise exceptions that are already a `DiagnosticError`.  This requires every such location to include `except DiagnosticError: raise`, and can easily be missed.

This PR makes two changes: First, the `DiagnosticRenderer` is updated to have a no-op callback rather than a NULL callback.  Second, the re-raising of `DiagnosticError` is moved to `Parser.report_error`, so that it does not need to be handled separately at several independent locations in the TVMScript parser.

### DIFF
--- a/python/tvm/script/parser/core/evaluator.py
+++ b/python/tvm/script/parser/core/evaluator.py
@@ -465,6 +465,7 @@ def eval_assign(
         return _eval_assign(target, source)
     except Exception as err:  # pylint: disable=broad-except
         parser.report_error(target, err)
+        raise
 
 
 def _eval_expr(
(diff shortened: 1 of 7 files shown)


tqchen (approved):

tqchen:
@tvm-bot rerun

PR #17374, opened 2024-09-14
Added bit-serial operators for AArch64

TBucher23:
The current implementation does not support bit-serial operators for AArch64 processor architecture. I added the functions in the folder  python/tvm/topi/arm_cpu  and adjusted the functions bitserial_conv2d_strategy_arm_cpu and schedule_bitserial_dense_arm_cpu in the folder python/tvm/relay/op/strategy to call the appropriate functions according to the target processor.

### DIFF
--- a/python/tvm/relay/op/strategy/arm_cpu.py
+++ b/python/tvm/relay/op/strategy/arm_cpu.py
@@ -601,11 +601,18 @@ def bitserial_conv2d_strategy_arm_cpu(attrs, inputs, out_type, target):
             name="bitserial_conv2d_nchw.arm_cpu",
         )
     elif layout == "NHWC":
-        strategy.add_implementation(
-            wrap_compute_bitserial_conv2d(topi.arm_cpu.bitserial_conv2d_nhwc),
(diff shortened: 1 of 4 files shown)


cbalint13 (commented):
Hi @TBucher23  ,

Thank you for this work in order to enhance the bitserial operators !

* Could we just reuse the existing [bitserial_conv2d.py](https://github.com/apache/tvm/blob/main/python/tvm/topi/arm_cpu/bitserial_conv2d.py) and [bitserial_dense.py](https://github.com/apache/tvm/blob/main/python/tvm/topi/arm_cpu/bitserial_dense.py)  infra  instead of the new files ?
* Also this will need some testcases to make sure operators are still numericaly correct, and also it emmits the expected llvm intrinsics for ```aarch64``` ([a hint](https://github.com/apache/tvm/blob/0d338828eebaa3ff705e8521f2a1b3530f73dc7d/tests/python/relay/test_op_level1.py#L857-L888)).


The TOPI will be phased out at some point, would be awesome to have these  bitserial/bitpack operators in the metaschedule flow.

PR #17377, opened 2024-09-16
[DOCS] Update document to include security model of RPC server

tqchen:
This PR update the documents to include the security model of the RPC server.


docs/reference/security.rst
```diff
@@ -34,10 +34,15 @@ The private security mailing address is: `security@apache.org <security@apache.o
...
 
-Considerations
+Security Model
 --------------
 The default binary generated by TVM only relies on a minimum runtime API.
 The runtime depends on a limited set of system calls(e.g. malloc) in the system library.
 
+
+TVM RPC server assumes that the user is trusted and needs to be used in a trusted network environment
+and encrypted channels. It allows writings of arbitrary files into the server for benchmarking purposes.
```
raboof, line 44: This seems like a good addition. Perhaps to make it really clear to users we should also mention that writing arbitrary files typically also leads to full remote code execution capabilities to anyone who can access this API?

tqchen: great suggestion, just updated to include explicit discussion about RCE

MasterJH5574 (approved):

Issue #17050, opened 2024-05-31
[Bug] [Relay] [ONNX] Incorrect shape inference of `Squeeze` in `DynamicToStatic`


shaoyuyoung:
#### Description
This torch model has only two ops: `ReflectionPad3d` and `squeeze`

Firstly, I try to export the ``torch model`` to ``onnx model``.
Then I get the below.
![model onnx](https://github.com/apache/tvm/assets/100203773/ad6dd940-1db0-429a-93f9-75ff36077af5)

Onnx does its unique operation on the model.
We can find that this is a **dynamic graph**  which **contains `if branch` structure** because of the **squeeze operator**. 

ONNX thinks **this model is valid**.
However, When I used relay to convert the model, I met `shape mismatch` error. The correct shape should be **Tensor[(13, 1, 1, 1), float32]** but TVM got **Tensor[(13, 13, 1, 1), float32]**.

(I think maybe) TVM has some bugs in the **DynamicToStatic** :(

#### Code
```python
import onnx
import torch
import torch.nn as nn
import torch.onnx
from tvm import relay, relax


def get_onnx_shape(onnx_model):
    input_shapes = {}
    for input in onnx_model.graph.input:
        shape = []
        for dim in input.type.tensor_type.shape.dim:
            if dim.dim_value > 0:
                shape.append(dim.dim_value)
            else:
                shape.append(1)

        input_shapes[input.name] = shape
    return input_shapes


class Model(nn.Module):
    def __init__(self):
        super(Model, self).__init__()
        self.pad = nn.ReflectionPad3d((0, 0, -43, 0, 0, -46))

    def forward(self, x):
        x = self.pad(x)
        x = torch.squeeze(x, dim=1)
        return x



model = Model()

input_tensor = torch.randn(13, 47, 44, 1)

onnx_file_path = "model.onnx"
torch.onnx.export(model,
                  input_tensor,
                  onnx_file_path,
                  export_params=True,
                  opset_version=14,
                  do_constant_folding=False,
                  input_names=['input'],
                  output_names=['output']
                  )

onnx_model = onnx.load("model.onnx")
shape_dict = get_onnx_shape(onnx_model)

mod, params = relay.frontend.from_onnx(
    onnx_model, shape_dict, freeze_params=True
)
```

#### Error Log
<details>
  <summary>click to see the error log</summary>

```
TVMError: Traceback (most recent call last):
[...]

Issue #16898, opened 2024-04-17
[Bug] Graph optimization model compilation error involving `Pad` operator


shaoyuyoung:
I am trying to compile an ONNX (graph below) model using TVM.
![5618520ff3e8817d39d4422c547eaf9](https://github.com/apache/tvm/assets/100203773/0a00c2d4-70e1-4e66-b7ae-96737e30d5b3)

Of course, this is a complicated graph, but we can simplify it as below.
![image](https://github.com/apache/tvm/assets/100203773/10fd68b9-168f-4db7-b593-a88bd410e612)

These two graphs are equal. When I try to compile them using TVM. The original ONNX model fails but the simplified ONNX model passes. It is very strange!

This seems to involve the `Pad` operator shape-checking problem.

In theory, I think **TVM should have strong compatibility with the native ONNX model**. However, the truth is not satisfactory. 

It seems that only simplified, simple models are acceptable to TVM
### Expected behavior
ONNX compilation passes

### Actual behavior

```
onnx fail
Traceback (most recent call last):
  18: tvm::runtime::PackedFuncObj::Extractor<tvm::runtime::PackedFuncSubObj<tvm::runtime::TypedPackedFunc<tvm::IRModule (tvm::transform::Pass, tvm::IRModule)>::AssignTypedLambda<tvm::transform::__mk_TVM9::{lambda(tvm::transform::Pass, tvm::IRModule)#1}>(tvm::transform::__mk_TVM9::{lambda(tvm::transform::Pass, tvm::IRModule)#1}, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >)::{lambda(tvm::runtime::TVMArgs const&, tvm::runtime::TVMRetValue*)#1}> >::Call(tvm::runtime::PackedFuncObj const*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, tvm::runtime::TVMRetValue)
  17: tvm::transform::Pass::operator()(tvm::IRModule) const
  16: tvm::transform::Pass::operator()(tvm::IRModule, tvm::transform::PassContext const&) const
  15: tvm::relay::transform::FunctionPassNode::operator()(tvm::IRModule, tvm::transform::PassContext const&) const
  14: _ZN3tvm7runtime13PackedFun
[...]

PR #17383, opened 2024-09-18
[Bugfix][ONNX] Skip constant If node generated by PyTorch

xhmelon:
This PR adds a check for If nodes for ONNX frontend of Relay to skip the broadcast if the predicate is constant.
Sometimes PyTorch to ONNX inserts silly if nodes that produce dynamic ranks, and ONNX frontend of TVM would broadcast the lower dimensions between branches, which is irrational for some cases, e.g. `5×5×3×4` to `5×5×3×4×1`. The predicate of silly if might be constant and reasonable to skip to avoid the broadcast problem.
This commit was captured in #16898 and #17050.


shaoyuyoung:
cc @Hzfengsy

Hzfengsy:
Thanks @xhmelon for the contribution. As we are moving towards the Unity strategy with Relax, we may deprecate Relay in the future ([ref](https://discuss.tvm.apache.org/t/phasing-out-legacy-components/17703/14)). We encourage the community to move to Relax stack.

Meanwhile, this PR is self-contained, we can get this in

Hzfengsy (approved):

PR #17395, opened 2024-09-20
[TIR, TVMScript] Add TIR - Triton integration

vinx13:
Added a macro `T.call_triton` in TIR script parser, which expands to AOT compilation of the kernel and the host TIR code to launch the kernel.

cc @tqchen @cyx-6

### DIFF
--- a/python/tvm/relax/vm_build.py
+++ b/python/tvm/relax/vm_build.py
@@ -243,13 +243,25 @@ def _vmlink(
     if ext_libs is None:
         ext_libs = []
     lib = None
+    relax_ext_libs = []
+    tir_ext_libs = []
     if tir_mod is not None and len(tir_mod.get_global_vars()) > 0:
         lib = tvm.build(
             tir_mod,
             target=target,
             runtime=_autodetect_system_lib_req(target, system_lib),
         )
-    return Executable(_ffi_api.VMLink(builder, target, lib, ext_libs, params))  # type: ignore
+    for ext_mod in ext_libs:
+        if ext_mod.type_key == "cuda":
+            tir_ext_libs.append(ext_mod)
+        else:
+            relax_ext_libs.append(ext_mod)
+    if lib is not None:
+        for mod in tir_ext_libs:
+            lib.import_module(mod)
+    elif len(tir_ext_libs) > 0:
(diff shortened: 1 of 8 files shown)


cyx-6 (approved):

tqchen:
This is a great mechanism to integrate kernel generators. Some notes on design that might help generalize it  abit


Would be great to change the intrinsic to T.call_kernel, which checks the first parameter for kernel types and dispatch accordingly.

- Let us think of a base class tir.kernel.BaseKernel which implements the base methods needed
    - compile_to_device_module(
        -   calls get_meta_data to get the meta_data
    - have a registry from class name to Kernel constructor(can be in next PR)
        - So the construct and import of related kernel can be done only when we see the related class name
         - call_kernel will lookup registry and construct related kernels if needed
    - TritonKernel subclasses that, call_kernel will lookup the constructor, construct it and  call related class method to generate the necessary downstream classes.
- Let us also add a tir.kernel.CUDAKernel as an example extension pt
    - CUDAKernel takes in cuda C source device code, and ensure call_kernel mechanism works the same

tests/python/contrib/test_tir_triton_integration.py
```diff
@@ -0,0 +1,79 @@
...
+import pytest
+
+try:
+    import triton
+    import triton.language as tl
+except ImportError:
+    pytestmark = pytest.skip("Triton is not available", allow_module_level=True)
+
+
+@tvm.testing.requires_cuda
```
tqchen, line 20: add a testcase of rewrite usecase

python/tvm/relax/vm_build.py
```diff
@@ -243,13 +243,25 @@ def _vmlink(
...
+    tir_ext_libs = []
     if tir_mod is not None and len(tir_mod.get_global_vars()) > 0:
         lib = tvm.build(
             tir_mod,
             target=target,
             runtime=_autodetect_system_lib_req(target, system_lib),
         )
-    return Executable(_ffi_api.VMLink(builder, target, lib, ext_libs, params))  # type: ignore
+    for ext_mod in ext_libs:
+        if ext_mod.type_key == "cuda":
```
tqchen, line 255: as a followup, add a function to check if the module is_device_module, this should include cuda, rocm, webgpu, vulkan, opencl

Issue #17346, opened 2024-09-08
[Relax][PyTorch] Support `torch.export.ExportedProgram` in Relax PyTorch Frontend


mshr-h:
This is a tracking issue for supporting PyTorch's `torch.export.ExportedProgram` in the Relax PyTorch frontend.

TODOs:
- [x] Cleanup the existing `TorchFXImporter`
  - [x] Cleanup unary op converters #17356 
  - [x] Cleanup binary op converters #17366
  - [x] Cleanup Neural Network op converters #17369 
  - [x] Cleanup Statistical op converters #17372
  - [x] Cleanup Search op converters #17372
  - [x] Cleanup Tensor Manipulation op converters #17376
  - [x] Cleanup Tensor Creation op converters #17376
  - [x] Cleanup DataType op converters #17372
  - [x] Cleanup Other op converters #17376
- [x] Upgrade PyTorch #17338 
- [x] Upgrade CI image
  - [x] #17397 
  - [x] #17410 
- [x] Introduce `ExportedProgramImporter` #17396
- [x] Expand ops supports
  - [x] unary ops #17421
  - [x] binary ops #17424
  - [x] neural network ops #17426
  - [x] statistical ops #17424
  - [x] search ops #17424
  - [x] tensor manipulation ops #17429 
  - [x] tensor creation ops #17429 
- [x] Documentation #17436 

refs:
- #16531 
- [torch.export IR Specification](https://pytorch.org/docs/stable/export.ir_spec.html)
- [torch.export Nightly Tutorial](https://pytorch.org/tutorials/intermediate/torch_export_nightly_tutorial.html)
- [Definition of the Core ATen Operator Set](https://pytorch.org/executorch/stable/ir-ops-set-definition.html)
- [IRs](https://pytorch.org/docs/stable/torch.compiler_ir.html)

cc @junrushao @shingjan @yelite

mshr-h:
This is a comparison of the number of lines of source code.

Current `fx_graph_translator`.
```
$ tokei fx_translator.py 
===============================================================================
 Language            Files        Lines         Code     Comments       Blanks
===============================================================================
 Python                  1         1947         1691           48          208
===============================================================================
 Total                   1         1947         1691           48          208
===============================================================================
```

New `fx_graph_translator` and`exportedprogram_translator`
```
$ tokei base_fx_graph_translator.py exportedprogram_translator.py fx_translator.py 
===============================================================================
 Language            Files        Lines         Code     Comments       Blanks
[...]

PR #17337, opened 2024-09-05
[CI] Disable NNPACK build and fix error on Android SDK installaion

mshr-h:
Fix #17070

PR #17397, opened 2024-09-20
[CI] Update image tag to 20240917-153130-9f281758

mshr-h:
As per title.
Part of #17346

### DIFF
--- a/ci/jenkins/docker-images.ini
+++ b/ci/jenkins/docker-images.ini
@@ -17,13 +17,13 @@
 
 # This data file is read during when Jenkins runs job to determine docker images.
 [jenkins]
-ci_arm: tlcpack/ci-arm:20240428-060115-0b09ed018
-ci_cortexm: tlcpack/ci-cortexm:20240428-060115-0b09ed018
-ci_cpu: tlcpack/ci_cpu:20240428-060115-0b09ed018
-ci_gpu: tlcpack/ci-gpu:20240428-060115-0b09ed018
-ci_hexagon: tlcpack/ci-hexagon:20240428-060115-0b09ed018
-ci_i386: tlcpack/ci-i386:20240428-060115-0b09ed018
-ci_lint: tlcpack/ci-lint:20240428-060115-0b09ed018
-ci_minimal: tlcpack/ci-minimal:20240428-060115-0b09ed018
-ci_riscv: tlcpack/ci-riscv:20240428-060115-0b09ed018
-ci_wasm: tlcpack/ci-wasm:20240428-060115-0b09ed018
+ci_arm: tlcpack/ci-arm:20240917-153130-9f281758
+ci_cortexm: tlcpack/ci-cortexm:20240917-153130-9f281758
+ci_cpu: tlcpack/ci_cpu:20240917-153130-9f281758
+ci_gpu: tlcpack/ci-gpu:20240917-153130-9f281758
+ci_hexagon: tlcpack/ci-hexagon:20240917-153130-9f281758
+ci_i386: tlcpack/ci-i386:20240917-153130-9f281758
+ci_lint: tlcpack/ci-lint:20240917-153130-9f281758
+ci_minimal: tlcpack/ci-minimal:20240917-153130-9f281758
+ci_riscv: tlcpack/ci-riscv:20240917-153130-9f281758
+ci_wasm: tlcpack/ci-wasm:20240917-153130-9f281758
--- a/tests/python/frontend/pytorch/test_fx_quant.py
+++ b/tests/python/frontend/pytorch/test_fx_quant.py
@@ -87,6 +87,9 @@ def test_deeplab_v3():
     quantize_and_build(model, 300)
 
 
+@pytest.mark.skip(
+    reason="Model binary isn't uploaded to S3. See https://github.com/apache/tvm/pull/17397"
+)
 def test_imagenet():
     for model_func in [resnet50, efficientnet_b4]:
         quantize_and_build(model_func(pretrained=True).eval(), 224)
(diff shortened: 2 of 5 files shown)


tqchen (approved):

yongwww (approved):

tqchen:
looks like a flaky error, we should increas atol and rtol for the case

mshr-h:
- https://ci.tlcpack.ai/blue/organizations/jenkins/tvm-unity/detail/PR-17397/1/pipeline/
  - `tests/python/relax/test_frontend_onnx.py:test_all_reduce_funcs` is failed due to flakiness
  - increased rtol from 1e-7 to 1e-4
- https://ci.tlcpack.ai/blue/organizations/jenkins/tvm-gpu/detail/PR-17397/1/tests/
  - `tests/python/frontend/pytorch/test_fx_quant.py:test_imagenet` is failed
  - As we upgraded torchvision, the model binary is also updeted but new one isn't uploaded to S3
- https://ci.tlcpack.ai/blue/organizations/jenkins/tvm-hexagon/detail/PR-17397/1/pipeline/
  - `./tests/scripts/task_build_hexagon_api.sh: line 44: ANDROID_NDK_HOME: unbound variable`
  - Failed to build because we removed Android SDK in #17337

tqchen:
here is the error 
```
[2024-09-21T07:06:36.974Z] self = <FixtureDef argname='test_case' scope='session' baseid='tests/python/tir-transform/test_tir_transform_simplify.py::TestMostRestrictiveConditional'>
[2024-09-21T07:06:36.974Z] request = <SubRequest 'test_case' for <Function test_compare[test_case2]>>
[2024-09-21T07:06:36.974Z] 
[2024-09-21T07:06:36.974Z]     def execute(self, request: SubRequest) -> FixtureValue:
[2024-09-21T07:06:36.974Z]         """Return the value of this fixture, executing it if not cached."""
[2024-09-21T07:06:36.974Z]         # Ensure that the dependent fixtures requested by this fixture are loaded.
[2024-09-21T07:06:36.974Z]         # This needs to be done before checking if we have a cached value, since
[2024-09-21T07:06:36.974Z]         # if a dependent fixture has their cache invalidated, e.g. due to
[2024-09-21T07:06:36.974Z]         # parametrization, they finalize themselves and fixtures depending on it
[2024-09-21T07:06:36.974Z]         # (which will likely include this fixture) setting `self.cached_result = None`.
[2024-09-21T07:06:36.974Z]         # See #4871
[2024-09-21T07:06:36.974Z]         requested_fixtures_that_should_finalize_us = []
[2024-09-21T07:06:36.974Z]         for argname in self.argnames:
[2024-09-21T07:06:36.974Z]             fixturedef = request._get_active_fixturedef(argname)
[2024-09-21T07:06:36.974Z]             # Saves requested fixtures in a list so we later can add our finalizer
[2024-09-21T07:06:36.974Z]             # to them, ensuring that if a requested fixture gets torn down we get torn
[2024-09-21T07:06:36.974Z]             # down first. This is generally handled by SetupState, but still currently
[2024-09-21T07:06:36.974Z]             # needed when this fixture is not parametrized but depends on a parametrized
[2024-09-21T07:06:36.974Z]             # fixture.
[2024-09-21T07:06:36.974Z]             if not isinstance(fixturedef, PseudoFixtureDef):
[2024-09-21T07:06:36.974Z]                 requested_fixtures_that_should_finalize_us.append(fixturedef)
[2024-09-21T07:06:36.974Z]     
[2024-09-21T07:06:36.974Z]         # Check for (and return) cached value/exception.
[2024-09-21T07:06:36.974Z]         if self.cached_result is not None:
[2024-09-21T07:06:36.974Z]             request_cache_key = self.cache_key(request)
[2024-09-21T07:06:36.974Z]             cache_key = self.cached_result[1]
[2024-09-21T07:06:36.974Z]             try:
[2024-09-21T07:06:36.974Z]                 # Attempt to make a normal == check: this might fail for objects
[2024-09-21T07:06:36.974Z]                 # which do not implement the standard comparison (like numpy arrays -- #6497).
[2024-09-21T07:06:36.974Z] >               cache_hit = bool(request_cache_key == cache_key)
[2024-09-21T07:06:36.974Z] E               TypeError: __bool__ should return bool, returned NotImplementedType
```

tqchen:
My guess is that it is likely due to the python used in ci by default is too low(it was 3.8) and we should have larger version that runs compatible with the latest pytest. cc @Lunderberg

tqchen:
Some guess about what happens to the `tests/python/tir-transform/test_tir_transform_simplify.py:TestMostRestrictiveConditional`

This is the only test where all parameters appears as bool expression and also do not contain a bool constant, not sure how that impacts the comparison impl. Note note the python version being used is python3.8 which likely is a bit too old here.

tqchen:
@yongwww can you help a bit with model binary

mshr-h:
Python 3.8 is going to EoL in 2024-10, so maybe it's a good time to move to 3.9.

tqchen:
Update about my finding. I find that i can reproduce the same issue on my python3.10, then i digged a bit into pytest code. Seems it is a versioning problem. The pytest version was too old. Likely the reliance of `bool(cached_key == requested_key)` was fixed by a later version of pytest. So the main reason here likely was pytest version issue

mshr-h:
The test passes with `pytest==8.2.0`, but it fails with `pytest==8.3.3`. Maybe there's a degradation in pytest?
We can temporarily work around this issue by pinning `pytest==8.2.0`.

tqchen:
Ah actually it has to do with newer pytest version. I see, then we should up find ways to fix the testcase

tqchen:
I think one solution  would be to add a wrapper class for the parameters and define our own equal operator, so it won't rely on tuple equal. @mshr-h do you mind try that?

tqchen:
if we can work around the torch model issue by skipping first, we can aim to merge first

mshr-h:
CI has passed.
Can you take a look at it again? Thanks! @tqchen @yongwww

PR #17398, opened 2024-09-21
[Disco] Enable float8 data type in disco

cyx-6:
This PR enables the float8 data type in disco, except all reeuce operation. Since in this PR, we pretend float8 to be uint8.


MasterJH5574 (approved):
Thanks for the fix!

PR #17411, opened 2024-09-24
[TIR][NarrowDataType] Bufferload's index should not inherit bits constraint of value

wrongtest-intellif:
Fix an issue in `NarrowDataType` pass. When we perform narrowing on buffer indices, the integer bits constraint is irrelevant to buffer load value's integer bits. Thus it looks like we should reset the context bit info when traverse into the indices field.

Another small change is in cast node rewrite, skip node construction when dtype actually matches.

### DIFF
--- a/src/tir/transforms/narrow_datatype.cc
+++ b/src/tir/transforms/narrow_datatype.cc
@@ -97,6 +97,13 @@ class DataTypeVisitor final : public StmtExprVisitor {
     }
   }
 
+  void VisitExpr_(const BufferLoadNode* op) {
+    int tmp = bits_;
+    bits_ = target_bits_;
+    StmtExprVisitor::VisitExpr_(op);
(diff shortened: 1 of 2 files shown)


tqchen:
do u mind leave a quick comment about what will happen without this PR

wrongtest-intellif:
> do u mind leave a quick comment about what will happen without this PR

In the main branch, this case's indices would not change after narrowing with i32.
```python
@T.prim_func
def before(A: T.Buffer((16,), "int64")):
    for i in range(T.int64(15)):
          A[i + T.int64(1)] = A[i] + T.int64(1)
```

The deduction chain is like below:
1. A[i] + T.int64(1) -> 64b
2. A[i] -> 64b
3. indices of load A -> 64b (here is the issue)
4. i -> requires 64b

tqchen (approved):

PR #17419, opened 2024-09-25
[FFI][BUGFIX] Grab GIL when check env signals

tqchen:
This PR updates the CheckSignals function to grab GIL. This is needed because we now explicitly release gil when calling any C functions. GIL will need to be obtained otherwise we will run into segfault when checking the signal.

The update now enables us to run ctrl + C in long running C functions.


tqchen:
run the following code and press control + C to check if signal is intercepted in python

```python
import numpy as np
import tvm


def main():
    print("start")
    tvm.get_global_func("testing.check_signal")(0.01)
    print("good")

main()
```

MasterJH5574 (approved):

MasterJH5574:
@tvm-bot rerun

PR #17371, opened 2024-09-13
[WEBGPU] Update runtime to remove deprecated API

tqchen:
This PR updates webgpu runtime code to remove
deprecated API. unblocks the CI.

PR #17420, opened 2024-09-26
[Web] Allow deprecated API requestAdapterInfo with any cast

CharlieFRuan:
The WebGPU API `requestAdapterInfo()` was deprecated as described here: https://developer.mozilla.org/en-US/docs/Web/API/GPUAdapter/requestAdapterInfo#browser_compatibility. This caused `npm run build` under `web/` to cause error since it cannot find `requestAdapterInfo`, hence the PR: https://github.com/apache/tvm/pull/17371

However, downstream users may still have older browser version and would still rely on `requestAdapterInfo()`, see https://github.com/mlc-ai/web-llm/issues/572 for an example.

This PR casts `adapter` to `any` type when calling `requestAdapterInfo()`, so it does not complain about not being able to find `requestAdapterInfo()` during compile time, hence no issue should be observed during runtime on older browsers, achieving backward compatibility.

Bump TVMjs version from `0.18.0-dev0` to `0.18.0-dev2` (skipped dev1 due to a mistake)


CharlieFRuan:
@tvm-bot rerun

tqchen (approved):

ysh329:
Hi @CharlieFRuan , recently release starts, but I found WebGPU part has some modification about version number. Is it required to change version in `web/package-lock.json` with same version number?

Each modification includes fixed five file modifications: https://github.com/apache/tvm/pull/17156. If there are any modifications related to 'web/package-lock. json', I will add them to the document and make the necessary changes this time if requied.

CharlieFRuan:
Hi @ysh329, thanks for the note! Just to check my understanding, does the version in `web/package-lock.json` have to match all five files, including the `dev` suffix?

ysh329:
@CharlieFRuan I'm not sure the `dev` suffix in `web/package-lock.json` whether should be same as all five files. However, I think the big version (before `dev` part) should be same. Then, it seems it's required to change `web/package-lock.json` to keep same with other five files.

Besides, I wann know in this patch why change the `dev0` to `dev2`, but not `dev1`?

CharlieFRuan:
I see, then I think it's easier for me to revert the version in `web/package-lock.json`. Shall I make a PR to revert back to `0.18.0-dev0`? Thanks

ysh329:
@CharlieFRuan Don't worry. Let me fix this minor problem.

CharlieFRuan:
Thanks a lot!

PR #17434, opened 2024-10-01
[TVMScript][TIR] Add source kernel intetration via call_kernel

vinx13:
Similar to triton integration, this adds integration with cuda source kernel via `call_kernel` in TIR

cc @tqchen


tqchen (approved):

tqchen:
@vinx13 please fix lint

Issue #17375, opened 2024-09-14
[Bug] Improper touched buffer assignment of Pass `MergeSharedMemoryAllocations` 


LeiWang1999:
Lead to Suboptimal Shared Memory Reuse.

pr #9341 introduced liveness analysis to merge the shared memory allocations , places touched buffer records at the outermost scope (e.g., outer loops) rather than at the innermost possible scope (e.g., inner loops or conditional branches). This can lead to incorrect liveness analysis. This approach luckily works well for some cases, such as the GEMM kernel, it fails in more complex scenarios, like the batched GEMM case or more complex algos. (as the outermost loop is always the single for loop node, that lead to incorrect gen kill point for each buffer).

One solutions I'm applying is to replace:

```cpp
if (it != alloc_info_.end() && it->second.alloc) {
  ICHECK_LT(it->second.level, scope_.size());
  if (IsAppropriateSharedMemory(GetRef<Var>(buf))) {
    scope_[it->second.level].touched.push_back(buf);
  }
}
```

into 

```cpp
if (IsAppropriateSharedMemory(GetRef<Var>(buf))) {
   scope_[scope_.size() - 1].touched.push_back(buf);
}
```

more detailed analysis can be found at [TVM Shared Memory Reuse Analysis](https://leiblog.wang/TVM%E4%B8%AD%E7%9A%84Shared-Memory-Reuse-Pass-%E5%88%86%E6%9E%90/)

If you think this analysis is correct, I can submit a PR then. :)

spectrometerHBH:
LGTM

LeiWang1999:
I made a simple fix, CC @spectrometerHBH

LeiWang1999:
Closed, as it doesn’t need to be addressed for now.

PR #17438, opened 2024-10-04
[Bugfix] Fix improper touched buffer assignment of Pass MergeSharedMemoryAllocations

LeiWang1999:
As discussed in issue #17375, the current rule for assigning touched buffers is not appropriate. Consider the following example:

```python
code_block_0
for k in range(0, 10): # (the gen point of A_shared and B_shared will be injected into this for expression)
    for i in range(0, 10):
          A_shared <- A
    for i in range(0, 10):
          B_shared <- B
    code_block_1 (consume A_shared and B_shared)
code_block_2 (produce and consume C_shared)
```

This setup works by chance in simple GEMM scenarios. However, the correct approach should be

```python
code_block_0
for k in range(0, 10): 
    for i in range(0, 10):
          A_shared <- A # (the gen point of A_shared should be bind into this BufferStore Node)
    for i in range(0, 10):
          B_shared <- B # (the gen point of B_shared be bind into this BufferStore Node)
    code_block_1 (consume A_shared and B_shared)
code_block_2 (produce and consume C_shared)
```

This approach works correctly even in more complex scenarios, such as batched GEMM, where the naive template would fail.

This pull request made a simple modification for `MergeSharedMemory` Pass to enable the right analysis, and always disable the naive naive shared memory buffer fuse if kernel with dynamic in `StorageRewrite` Pass

### DIFF
--- a/src/tir/transforms/merge_shared_memory_allocations.cc
+++ b/src/tir/transforms/merge_shared_memory_allocations.cc
@@ -42,12 +42,12 @@ namespace tir {
 using runtime::StorageRank;
 using runtime::StorageScope;
 
-bool IsDynamicSharedMemory(Var buffer_var) {
+static bool IsDynamicSharedMemory(Var buffer_var) {
   StorageScope storage_scope = runtime::StorageScope::Create(GetPtrStorageScope(buffer_var));
   return storage_scope.rank == runtime::StorageRank::kShared && storage_scope.tag == ".dyn";
 }
 
-bool IsStaticSharedMemory(Var buffer_var) {
+static bool IsStaticSharedMemory(Var buffer_var) {
   StorageScope storage_scope = runtime::StorageScope::Create(GetPtrStorageScope(buffer_var));
   return storage_scope.rank == runtime::StorageRank::kShared && storage_scope.tag == "";
 }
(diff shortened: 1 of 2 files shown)


src/tir/transforms/merge_shared_memory_allocations.cc
```diff
@@ -156,29 +156,18 @@ class SharedMemLinearAccessPatternFinder final : public StmtExprVisitor {
     if (it != alloc_info_.end() && it->second.alloc) {
       ICHECK_LT(it->second.level, scope_.size()) << "Load memory in places other than store.";
       if (IsAppropriateSharedMemory(GetRef<Var>(buf))) {
-        scope_[it->second.level].touched.push_back(buf);
+        scope_[scope_.size() - 1].touched.push_back(buf);
       }
     }
   }
 
-  void VisitExpr_(const CallNode* op) final {
```
LeiWang1999, line 164: remove this visit function, as it only allow visit indices, which will lead to some buffer load statement not be traced.

src/tir/transforms/storage_rewrite.cc
```diff
@@ -1755,7 +1793,7 @@ Pass StorageRewrite() {
     // padded out to 32 bits) would require either rewriting
     // AllocateConst::data, or would require the code generators to
     // handle vectorized constants.
-    return PointerValueTypeRewrite(std::move(f), true, false, false, true, true, true, false,
+    return PointerValueTypeRewrite(std::move(f), true, false, false, false, true, true, false,
```
LeiWang1999, line 1796: The fourth condition must be false, as the vectorized buffer merge (for example, merge half B_shared[1024] into halfx8 B_shared[128]) will occasional lead to a unhandled behavior during async copy lowering phase.

tqchen:
@LeiWang1999 do u mind close and and send a new version of PR to main? with some updated testcases

PR #17171, opened 2024-07-18
[MetaSchedule]Add a testcase for padded conv2d in meta_schedule

YXY-0922:
For a `conv2d` workload that requires padding:

### Details

Test Script: `test_padded_conv`
Location: `tests/python/meta_schedule/test_meta_schedule_schedule_rule_mlt_tc.py` 
```Python
mod = te.create_prim_func(
    te_workload.conv2d_nhwc(
        1,
        224,
        224,
        3,
        64,
        7,
        2,
        3,
        in_dtype="float16",
        out_dtype="float32",
    )
)
```
This configuration sets up a conv2d workload where the product of reduce axes is 147. Therefore, padding is needed when applying the rule multi_level_tiling_tensorcore.

Issue #16614, opened 2024-02-21
[Bug] Tensorization Failure During Multilevel Tiling with Tensor Intrin


zxybazh:
### Expected behavior

MetaSchedule Tuning Works for the given Conv2d workload

### Actual behavior

Triggers an error `ValueError: The block no longer exists in the IRModule` during application of schedule rule Multi-level tiling with tensor intrin. I notcied that `state->tensor_core_reindex_store` would point to a block that is already merged into another block via ComputeInline during application of `TileWithTensorIntrin`. 

### Environment

Latest TVM Main

### Steps to reproduce

```
import tvm
from tvm.script import tir as T
from tvm import meta_schedule as ms

@T.prim_func(private=True)
def fused_conv2d_add1(reshape3: T.Buffer((T.int64(50), T.int64(8), T.int64(72), T.int64(128)), "float16"), conv_in_weight: T.Buffer((T.int64(320), T.int64(8), T.int64(3), T.int64(3)), "float16"), lv23: T.Buffer((T.int64(1), T.int64(320), T.int64(1), T.int64(1)), "float16"), T_add_intermediate: T.Buffer((T.int64(50), T.int64(320), T.int64(72), T.int64(128)), "float16")):
    T.func_attr({"tir.noalias": T.bool(True)})
    # with T.block("root"):
    pad_temp = T.alloc_buffer((T.int64(50), T.int64(8), T.int64(74), T.int64(130)), "float16")
    conv2d_nchw_intermediate = T.alloc_buffer((T.int64(50), T.int64(320), T.int64(72), T.int64(128)), "float16")
    for i0, i1, i2, i3 in T.grid(T.int64(50), T.int64(8), T.int64(74), T.int64(130)):
        with T.block("pad_temp"):
            v_i0, v_i1, v_i2, v_i3 = T.axis.remap("SSSS", [i0, i1, i2, i3])
            T.reads(reshape3[v_i0, v_i1, v_i2 - T.int64(1), v_i3 - T.int64(1)])
            T.writes(pad_temp[v_i0, v_i1, v_i2, v_i3])
            pad_temp[v_i0, v_i1, v_i2, v_i3] = T.if_then_else(T.int64(1) <= v_i2 and v_i2 < T.int64(73) and T.int64(1) <= v_i3 and v_i3 < T.int64(129), reshape3[v_i0, v_i1, v_i2 - T.int64(1), v_i3 - T.int64(1)], T.float16(0))
[...]

PR #16239, opened 2023-12-14
[TIR][Schedule] TileWithTensorIntrin skip incorrect ComputeInline for input-padding

XFPlus:
Previously, TileWithTensorIntrin tried to inline all producers into padding blocks once padding happened even if only one producer was padded.
This PR fixes that.

Possibly related issue: https://github.com/apache/tvm/issues/15446

cc @tqchen  @junrushao

Issue #15505, opened 2023-08-08
[Bug] [Relay] [MetaSchedule] ValueError: The block no longer exists in the IRModule


Civitasv:
I'm trying to use MetaSchedule to tune vae_decoder model. But it failed to extract tasks.

### Expected behavior

It should successful.

### Actual behavior

```txt
  File "/home/xxx/xxx/tvm_tunning/tunningx/tunning/meta_schedule.py", line 67, in do_all_tune
    database = ms.relay_integration.tune_relay(
  File "/home/xxx/xxx/tvm/python/tvm/meta_schedule/relay_integration.py", line 352, in tune_relay
    return tune_tasks(
  File "/home/xxx/xxx/tvm/python/tvm/meta_schedule/tune.py", line 118, in tune_tasks
    task_scheduler.tune(
  File "/home/xxx/xxx/tvm/python/tvm/meta_schedule/task_scheduler/task_scheduler.py", line 132, in tune
    _ffi_api.TaskSchedulerTune(  # type: ignore # pylint: disable=no-member
  File "/home/xxx/xxx/tvm/python/tvm/_ffi/_ctypes/packed_func.py", line 238, in __call__
    raise get_last_ffi_error()
ValueError: Traceback (most recent call last):
  11: TVMFuncCall
  10: _ZN3tvm7runtime13PackedFunc
[...]

PR #17440, opened 2024-10-04
[BugFix][TIR][Schedule] TileWithTensorIntrin skip ComputeInline if bu…

Yongqi-Zhuo:
…ffer not padded by PadEinsum

`TileWithTensorIntrin` in `src/tir/schedule/transform.cc` calls `PadEinsum` and inlines the padded input (output) to its original producer (consumer). However it is possible that one of the inputs/outputs does not need to be padded, in which case that producer (consumer) is not padded by `PadEinsum`. This means that `TileWithTensorIntrin` may inline blocks that are irrelevant to padding and must not be inlined.

This has led to multiple bug reports and (temporary) fixes as in #17171, #16614, #16239 and #15505. Unfortunately #16239 and #17171 can only prevent TVM from crashing when the padded buffer is an input/output buffer, and still some incorrect inlining may be performed. The [workaround](https://github.com/nautasolva/tvm/commit/4e0191e3e303daa62edb8050df444d6b99df1048) in #15505 tried to handle this bug by extra checking in the `MultiLevelTilingTensorCore` rule, which is logically incorrect. This PR aims to provide a one-and-for-all fix for this.

Credit to @XFPlus for the bug reproduce example in #16239.

### DIFF
--- a/src/tir/schedule/transform.cc
+++ b/src/tir/schedule/transform.cc
@@ -326,23 +326,62 @@ Optional<LoopRV> TileWithTensorIntrin(const tir::Schedule& sch, const tir::Block
   if (!opt_tensorize_info) return NullOpt;
   const tir::TensorizeInfoNode* info = opt_tensorize_info.value().get();
   if (info->block_iter_paddings.defined()) {
+    // We have to track whether each producer or consumer is padded.
+    // To do so, we first record all the Block's.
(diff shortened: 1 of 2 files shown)


Yongqi-Zhuo:
cc @tqchen @junrushao

vinx13 (approved):

PR #17442, opened 2024-10-05
[TIR] Enhance `LowerThreadAllreduce` pass to automatically infer shared memory scope

LeiWang1999:
The pass `LowerThreadAllreduce` enables efficient block reduction. However, block reduction often requires a large amount of shared memory space. The current implementation of `LowerThreadAllreduce` only enable static shared memory reduce buffer allocation and prevents the shared memory merging when another shared memory scope is defined as `shared.dyn`.

```cpp
A_shared: "shared.dyn"
B_shared: "shared.dyn"
C_shared: "shared.dyn"
red: "shared" (can not be merged into the union shared memory pool)
```

This pull request addresses this issue by first collecting buffer allocations, and then determining the memory scope of the reduction buffer, allowing for memory space fusion in the following `MergeSharedMemoryAllocations` pass.

### DIFF
--- a/src/tir/transforms/lower_thread_allreduce.cc
+++ b/src/tir/transforms/lower_thread_allreduce.cc
@@ -38,12 +38,49 @@
 namespace tvm {
 namespace tir {
 
+using runtime::StorageRank;
+using runtime::StorageScope;
+
+/*!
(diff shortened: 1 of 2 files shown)


Hzfengsy:
Please add a testcase for the enhancement

PR #17444, opened 2024-10-06
[CI][Docs] Upgrade Sphinx

mshr-h:
As per title.

### DIFF
--- a/docker/install/ubuntu_install_sphinx.sh
+++ b/docker/install/ubuntu_install_sphinx.sh
@@ -20,14 +20,14 @@ set -e
 set -u
 set -o pipefail
 
-# NOTE: install docutils < 0.17 to work around https://github.com/readthedocs/sphinx_rtd_theme/issues/1115
+# pin sphinx<7 because https://github.com/readthedocs/sphinx_rtd_theme/issues/1465
 pip3 install \
     autodocsumm \
-    "commonmark>=0.7.3" \
-    "docutils>=0.11,<0.17" \
+    commonmark \
+    docutils \
     Image \
     matplotlib \
-    sphinx==4.2.0 \
+    "sphinx<7" \
     sphinx_autodoc_annotation \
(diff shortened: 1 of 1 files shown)


docker/install/ubuntu_install_sphinx.sh
```diff
@@ -20,14 +20,14 @@ set -e
 set -u
 set -o pipefail
 
-# NOTE: install docutils < 0.17 to work around https://github.com/readthedocs/sphinx_rtd_theme/issues/1115
+# pin sphinx<7 because https://github.com/readthedocs/sphinx_rtd_theme/issues/1465
```
Hzfengsy, line 23: Sphinx 7 should work with https://github.com/tlc-pack/tlcpack-sphinx-addon/pull/11

docker/install/ubuntu_install_sphinx.sh
```diff
@@ -20,14 +20,14 @@ set -e
...
 pip3 install \
     autodocsumm \
-    "commonmark>=0.7.3" \
-    "docutils>=0.11,<0.17" \
+    commonmark \
+    docutils \
     Image \
     matplotlib \
-    sphinx==4.2.0 \
+    "sphinx<7" \
```
Hzfengsy, line 30: It would be great if you could pin to a specific version (via ==x.y.z) rather than `<7.0`

docker/install/ubuntu_install_sphinx.sh
mshr-h: installing https://github.com/tlc-pack/tlcpack-sphinx-addon solved the problem. Thanks a lot!

docker/install/ubuntu_install_sphinx.sh
mshr-h: unpinned sphinx version since docs build works

tqchen (approved):

PR #17446, opened 2024-10-06
[Relax] Enhance global var struct info update in BlockBuilder

Hzfengsy:
The current BlockBuilder does not update the global var struct info after a function is updated. This commit enhances the BlockBuilder to update the global var struct info after a function is updated.

### DIFF
--- a/src/relax/ir/block_builder.cc
+++ b/src/relax/ir/block_builder.cc
@@ -87,18 +87,7 @@ class BlockBuilderImpl : public BlockBuilderNode {
       }
       GlobalVar gvar(func_name);
 
-      StructInfo finfo;
-      if (func->struct_info_.defined()) {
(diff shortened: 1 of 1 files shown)


tqchen (approved):

src/relax/ir/block_builder.cc
```diff
@@ -141,6 +130,8 @@ class BlockBuilderImpl : public BlockBuilderNode {
     }
 
     context_mod_->Update(gv, function);
+    // Force update struct info
+    gv->struct_info_ = GetStructInfo(function);
```
tqchen, line 134: one potential issue here is that the caller still relies on the assumption of previous function sigature, so automatically updating this may not always be desirable, is there any motivating usecase here?

tqchen (changes requested):

PR #17455, opened 2024-10-09
Compiled with Default Target(LLVM) and Built with USE_MRVL=ON

MNGanesan:
Please review the fix:

It is a unique use-case to check the default target(LLVM), though TVM is built with BYOC(MRVL-ON)
The config of Codegen(BYOC) contains default values for configuration/optons, it is extracted
during _generate_codegen_args. In command line processing, validate_target_args checks if there are add-on options and it expects that particular target to be given explicitly in command line.  Here, it is test for default (LLVM) path only, hence validate_target_args need to ignore the codegen's configuration


MNGanesan:
@Hzfengsy @lhutton1 

Could you please review this patch ?

Thanks

Hzfengsy:
cc TVMC maintainer @leandron

tests/python/driver/tvmc/test_target_options.py
```diff
@@ -86,6 +86,19 @@ def test_default_arg_for_mrvl_hybrid():
...
+# Test for default(LLVM) target, when built with USE_MRVL=ON
+def test_mrvl_build_with_llvm_only_target():
+    parser = argparse.ArgumentParser()
+    generate_target_args(parser)
+    parsed, _ = parser.parse_known_args(
+        [
+            "--target=llvm",
+        ]
+    )
+    assert parsed.target == "llvm"
```
leandron, line 99: I'm trying to understand what would be expected from this test case if TVM is not built with `USE_MRVL=ON`?

tests/python/driver/tvmc/test_target_options.py
```diff
@@ -86,6 +86,19 @@ def test_default_arg_for_mrvl_hybrid():
...
 
 
+@tvm.testing.requires_mrvl
+# Test for default(LLVM) target, when built with USE_MRVL=ON
+def test_mrvl_build_with_llvm_only_target():
+    parser = argparse.ArgumentParser()
+    generate_target_args(parser)
+    parsed, _ = parser.parse_known_args(
+        [
+            "--target=llvm",
```
leandron, line 96: Shouldn't this be `--target=mrvl` if I'm understanding this correctly?

python/tvm/driver/tvmc/target.py
```diff
@@ -181,6 +181,12 @@ def validate_targets(parse_targets, additional_target_options=None):
     if additional_target_options is not None:
         for target_name in additional_target_options:
             if not any([target for target in parse_targets if target["name"] == target_name]):
+                # When built with USE_MRVL=ON, add-on target options are passed from MRVL codegen's
+                # config which has pass_default=True and compiled with default target, don't error
+                # Use case: --target="llvm" cnn.onnx
+                if (len(tvm_targets) == 1) and (target_name == "mrvl"):
```
leandron, line 187: TVMC would allow you to use multi-backend in many cases, so this conditional here seems too specific on a generic function. Can you propose a more generic solutionf?

leandron (changes requested):

python/tvm/driver/tvmc/target.py
MNGanesan: Thanks Leandron, I've taken your feedback , I'll be revisiting with a generic fix

MNGanesan:
@leandron 

Please review my updated patch which has generic fix.,
The purpose of this fix is not to extract any options of codegen(BYOC) when only default target is specified by the user.

Eg:
python3 -m tvm.driver.tvmc compile  --target="llvm" model.onnx 

Thanks

tests/python/driver/tvmc/test_target_options.py
MNGanesan: Here, we want to check the default target for the particular BYOC/codegen, which is now defined in the composite_target.py,  so use-case is --target="llvm"

tests/python/driver/tvmc/test_target_options.py
MNGanesan: The test will be skipped by the decorator when USE_MRVL=OFF i.e @tvm.testing.requires_mrvl

leandron (approved):
OK, thanks, it looks very elegant and general now.

MNGanesan:
@leandron 

Could you please merge this PR?

Thanks

leandron:
> @leandron
> 
> Could you please merge this PR?
> 
> Thanks

Yes, I just forgot this step after the review. This is merged now - thanks for your PR.

MNGanesan:
@leandron 

Thanks a lot for your quick review feedback and response.

PR #17467, opened 2024-10-15
[FIX][RELAX][ONNX] Fix typo in onnx frontend 

Husenap:
Fixed a typo in the BinaryBase class that was introduced in #17427


Hzfengsy (approved):

Hzfengsy:
Great catch and thanks for the fix

PR #17490, opened 2024-10-24
[FIX][ONNX][RELAX] Add support for dynamic ShapeExpr in Slice, Squeeze and Flatten

PatrikPerssonInceptron:
# Squeeze
I added a tuple wrapping around axis when sending it to _np.squeeze since it does not support list objects. I added an additional check to see if the data is a ShapeExpr. If this is the case I perform the inverse of the operations done in the Unsqueeze operation. 

# Slice 
I fixed a bug where the dimensions were assumed to be of IntImm data type by adding an additional check to handle the constant and dynamic case separately.

# Flatten
I fixed a bug where the dimensions were assumed to be of IntImm data type by adding an additional check to handle the constant and dynamic case separately.

# Split
I replaced asnumpy with numpy since asnumpy is deprecated. I added a call to .item() to convert from numpy data types to python primitives. 

# Regression Tests
I extracted generate_random_value function from generate_random_inputs.
I added isinstance(tvm_out, (int, float, bool)) in check_correctness since the VM can return python primitive types

## Squeeze
I added a test case for squeeze with constant input since it was missing
I added a test case for squeeze with dynamic input shape
I added a test case for squeeze with dynamic shape expression

## Slice
I added a test case for slice with dynamic shape expression

## Split
I fixed a bug in test_split which caused the constant split case never to occur

### DIFF
--- a/python/tvm/relax/frontend/onnx/onnx_frontend.py
+++ b/python/tvm/relax/frontend/onnx/onnx_frontend.py
@@ -1203,18 +1203,19 @@ class Squeeze(OnnxOpConverter):
         axis = get_constant(inputs[1], params)
         if isinstance(axis, relax.Constant):
             axis = [int(x) for x in axis.data.numpy()]
-        # If data is constant, perform computation directly.
-        if isinstance(data, relax.Constant):
-            out_data = _np.squeeze(data.data.numpy(), tuple(axis))
-            return relax.const(out_data, data.struct_info.dtype)
+            
+            # If data is constant, perform computation directly.
+            if isinstance(data, relax.Constant):
+                out_data = _np.squeeze(data.data.numpy(), tuple(axis))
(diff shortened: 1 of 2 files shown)


tqchen:
cc @Hzfengsy can you help to take a look

Hzfengsy:
Thanks for the contribution, please add a regression test :)

PatrikPerssonInceptron:
My pleasure, were exactly in the code should I add the regression tests? :)

Hzfengsy:
`tests/python/relax/test_frontend_onnx.py` would be great:)

Hzfengsy (approved):

Hzfengsy:
LGTM. Thanks for the contribution!!!

PatrikPerssonInceptron:
Happy to help!

Issue #17488, opened 2024-10-24
[Bug] InternalError: Check failed: (!block_stack_.empty()) is false in StaticPlanBlockMemory with Dataflow


Thrsu:
The code defines a relax function that allocates a tensor using R.builtin.alloc_tensor inside a dataflow block. When applying the StaticPlanBlockMemory transformation, it results in an internal error with the message: `Check failed: (!block_stack_.empty()) is false`. However, if the dataflow block is removed, the transformation completes without error.

### Expected behavior

The `StaticPlanBlockMemory` transformation should handle functions with dataflow blocks correctly, allowing the memory planning without causing internal errors.

### Actual behavior
```
File "/software/tvm/src/relax/transform/static_plan_block_memory.cc", line 597
InternalError: Check failed: (!block_stack_.empty()) is false:
```

### Steps to reproduce
```python
import tvm
from tvm import relax
from tvm.script import ir as I
from tvm.script import relax as R

@I.ir_module
class Module:
    @R.function
    def main() -> R.Tensor((10,), dtype="float32"):
        with R.dataflow():
            gv: R.Tensor((10,), dtype="float32") = R.builtin.alloc_tensor(
                R.shape([10]), R.dtype("float32"), R.prim_value(0), R.str("global")
            )
            R.output(gv)
        return gv

mod = Module
mod_seq = tvm.transform.Sequential([relax.transform.StaticPlanBlockMemory()])(mod)
```

This issue may indicate a problem with how dataflow blocks are processed within the StaticPlanBlockMemory pass. Any guidance or fixes to resolve this would be appreciated.

PR #17501, opened 2024-11-01
Fix InternalError in StaticPlanBlockMemory when visiting DataflowBlockNode

Thrsu:
This PR fixes an internal error #17488 

This error happens because the visitor class StorageAllocatorBaseVisitor does not correctly handle DataflowBlockNode instances. Specifically, the VisitBindingBlock_ method is not overridden for DataflowBlockNode, leading to an empty block_stack_ when it is expected to contain the current block.

To fix this issue, we need to override the VisitBindingBlock_ method for const DataflowBlockNode* in the StorageAllocatorBaseVisitor class. By doing so, we ensure that the block_stack_ is correctly managed when visiting dataflow blocks, similar to how it is managed for regular binding blocks.

### DIFF
--- a/src/relax/transform/static_plan_block_memory.cc
+++ b/src/relax/transform/static_plan_block_memory.cc
@@ -730,6 +730,15 @@ class StorageAllocator : public StorageAllocatorBaseVisitor {
     }
   }
 
+  void VisitBindingBlock_(const DataflowBlockNode* block) override {
+    // We maintain a block stack for token allocation-site and use-site check.
+    block_stack_.push_back(block);
+    ExprVisitor::VisitBindingBlock_(block);
+    ICHECK(!block_stack_.empty());
+    ICHECK(block_stack_.back() == block);
+    block_stack_.pop_back();
+  }
+
   void VisitBinding_(const VarBindingNode* binding, const CallNode* call) final {
     static const Op& alloc_tensor_op = Op::Get("relax.builtin.alloc_tensor");
     if (call->op == alloc_tensor_op) {
--- a/src/relax/transform/static_plan_block_memory.cc
+++ b/src/relax/transform/static_plan_block_memory.cc
@@ -314,6 +314,15 @@ class StorageAllocatorBaseVisitor : public ExprVisitor {
     SetTokens(binding->var.get(), token_map_[binding->value.get()]);
   }
 
+  void VisitBindingBlock_(const DataflowBlockNode* block) override {
+    // We maintain a block stack for token allocation-site and use-site check.
+    block_stack_.push_back(block);
+    ExprVisitor::VisitBindingBlock_(block);
+    ICHECK(!block_stack_.empty());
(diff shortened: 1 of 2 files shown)


Thrsu:
Could you please help review this pull request? @tqchen @yongwww

MasterJH5574:
Hi @Thrsu, thanks for contributing! Actually we expect all dataflow blocks to have been removed in order to properly apply memory planning, given we no longer need the dataflow block information in subsequent lowering passes. Here is our lowering pipeline and you can see that we apply `ToNonDataflow` before other lowering passes, including the memory planning https://github.com/apache/tvm/blob/3e386fd3732778d5eb7998f23e2d940741102d9a/python/tvm/relax/pipeline.py#L89-L92

So if possible, do you mind sharing the reason why you want the memory planning pass to work with dataflow blocks? A code example may be sufficient, thanks!

Thrsu:
> Hi @Thrsu, thanks for contributing! Actually we expect all dataflow blocks to have been removed in order to properly apply memory planning, given we no longer need the dataflow block information in subsequent lowering passes. Here is our lowering pipeline and you can see that we apply `ToNonDataflow` before other lowering passes, including the memory planning
> 
> https://github.com/apache/tvm/blob/3e386fd3732778d5eb7998f23e2d940741102d9a/python/tvm/relax/pipeline.py#L89-L92
> 
> So if possible, do you mind sharing the reason why you want the memory planning pass to work with dataflow blocks? A code example may be sufficient, thanks!

Thanks for your response!
I want to customize a pass pipeline for exploring the possibility of improving the optimization performance. However, it crashed unexpectedly due to the assumptions about the pass. This PR aims to enable the StaticPlanBlockMemory optimization to run correctly in such conditions.

MasterJH5574:
@Thrsu I see. Thanks for elaborating! The memory planning pass also has a dependency on the `transform.CallTIRRewrite` pass, which you may also need to be careful about. Could you please add a unit test to this file, showing that the memory planning works for dataflow blocks? With the unit test we can merge this PR. Thanks!

https://github.com/apache/tvm/blob/main/tests/python/relax/test_transform_static_plan_block_memory.py

Thrsu:
> @Thrsu I see. Thanks for elaborating! The memory planning pass also has a dependency on the `transform.CallTIRRewrite` pass, which you may also need to be careful about. Could you please add a unit test to this file, showing that the memory planning works for dataflow blocks? With the unit test we can merge this PR. Thanks!
> 
> https://github.com/apache/tvm/blob/main/tests/python/relax/test_transform_static_plan_block_memory.py

@MasterJH5574 Thank you for the guidance! I've added a unit test to the specified file, demonstrating that the memory planning pass works for dataflow blocks as requested. Could you please review the latest changes? Thank you for your time and assistance!

MasterJH5574 (approved):

MasterJH5574:
@Thrsu Thank you for adding the unit test! It looks good to me :-)

PR #17502, opened 2024-11-02
[FIX][TOPI][strided_slice] Fix topi.strided_slice output shape

PatrikPerssonInceptron:
Fixed inconsistency between relax.strided_slice output shape and topi.strided_slice output shape when assume_inbound 
is set to false.

I added an assume_inbound flag to topi.strided_slice. If it is false then the begin and end indices are canonicalized before the dimension length is calculated. To avoid introducing duplicate code i removed the corresponding functions from relax/../index.cc and instead reference the new functions defined in include/tvm/topi/transform.h.

I added a regression test for the case with assume_inbound = false when the end of the slice is at infinity, e.g. v[begin:], which previously caused an incorrect output shape.

### DIFF
--- a/tests/python/relax/test_op_index.py
+++ b/tests/python/relax/test_op_index.py
@@ -21,6 +21,7 @@ from tvm import relax, tir
 from tvm import TVMError
 from tvm.ir import Op, VDevice
 from tvm.script import ir as I, relax as R, tir as T
+import numpy as np
 
 
 def test_op_correctness():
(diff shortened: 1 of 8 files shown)


tests/python/relax/test_op_index.py
```diff
@@ -1010,5 +1011,81 @@ def strided_slice(
     tvm.ir.assert_structural_equal(expected, after)
 
 
+def test_legalize_dynamic_begin_inf_end():
+    """relax.op.strided_slice FLegalize must support dynamic begin/end"""
+
+    @I.ir_module
+    class before:
```
Hzfengsy, line 1018: You can use `# fmt: off` and `# fmt: on` to disable black for specific code (e.g. TVMScript)

Hzfengsy (approved):
General LGTM

PR #17504, opened 2024-11-03
[RELAX][ONNX] Add support for dynamic shape expression in Expand

PatrikPerssonInceptron:
# Expand
Updated Expand converter in onnx frontend to to handle dynamic shape expressions by matching data and target_shape
dimension lengths and converting any target shape elements assigned to -1 to the data shape dimension.

Added regression test to onnx frontend testing Expand with a dynamic shape 

# Additional small fixes
I replaced int(k.data.numpy()[0]) with int(k.data.numpy().item()) in Trilu since sometimes the data shape is a zero dimensional scalar

I added a function converting from PrimExpr to PrimValue before sending in starts, ends, steps to relax.op.strided_slice since it does not accept PrimeExpr as inputs.


Hzfengsy (approved):
LGTM

PR #17505, opened 2024-11-05
[RELAX][ONNX][FIX] add a parser to handle expression in the shape dim names 

PatrikPerssonInceptron:
# Problem
The shape names in an ONNX model can contain expressions such as the shape `int64[batch_size,past_sequence_length + sequence_length]` of the attention mask of an LLM. In this case, the second dimension contains an expression `past_sequence_length + sequence_length` where `past_sequence_length` and `sequence_length` should be individual variables added together. However, currently, a new variable named `"past_sequence_length + sequence_length"` is instead created when translating the graph. 

# Fix
I added a simple parser that creates individual size variables for the variable names and generates the resulting prim expression. Note, in order to keep the parser simple, it evaluates expressions left to right. Not accounting for operator precedence. 

# Test
I added regression tests to verify that the onnx shape dim expression are evaluated correctly.

# Additional small fixes
In the case when PrimValues are encountered in the BinaryBase, they are not always fully extracted before turning them into numpy arrays. I added an additional check that extracts the value from IntImm and FloatImm types before converting them to numpy arrays.

### DIFF
--- a/python/tvm/relax/frontend/onnx/onnx_frontend.py
+++ b/python/tvm/relax/frontend/onnx/onnx_frontend.py
@@ -35,6 +35,8 @@ Not all TVM kernels currently support dynamic shapes, please file an issue on
 github.com/apache/tvm/issues if you hit an error with dynamic kernels.
 """
 import math
+import operator
+import re
 import warnings
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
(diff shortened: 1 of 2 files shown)


Hzfengsy (approved):

Hzfengsy:
Thanks for the improvements

PR #17510, opened 2024-11-10
[Web][PROPOSAL]: implement ArtifactCache using OPFS

KarolNov:
**Motivation**
[OPFS](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API/Origin_private_file_system) has better persistence, performance, and consistent support for large files across browsers.
Personally I would like to see this caching strategy for https://github.com/mlc-ai/web-llm.

This PR doesn't implement all required changes and hasn't been proved to work yet - I just created a PoC class implementing the `ArtifactCacheTemplate`.
Potentially a problem with this approach would be that this cache uses OPFS on main thread with async APIs. A better solution could be to use Web Worker and sync file handlers - https://developer.mozilla.org/en-US/docs/Web/API/File_System_API/Origin_private_file_system#manipulating_the_opfs_from_a_web_worker. Both approaches could be valid for different use cases, so I propose creating 2 new caching strategies - `opfs` and `opfs-worker`.
I could finish the required work if we find it useful. Otherwise I would just move caching higher up in my stack and just do it for my specific use case.

### DIFF
--- a/web/src/artifact_cache.ts
+++ b/web/src/artifact_cache.ts
@@ -353,6 +353,138 @@ export class ArtifactIndexedDBCache implements ArtifactCacheTemplate {
   }
 }
 
+/**
+ * Cache by storing results as files in OPFS file system
+ * https://developer.mozilla.org/en-US/docs/Web/API/File_System_API/Origin_private_file_system
+ */
+export class ArtifactOPFSCache implements ArtifactCacheTemplate {
+  private dirName?: string;
+  private fileVersion = 1;
+  private dirHandle: FileSystemDirectoryHandle | null = null;
+
+  constructor(dirName: string) {
+    // use the version suffix to allow upgrade if needed
+    this.dirName = `${dirName}-v${this.fileVersion}`;
+  }
+
+  /**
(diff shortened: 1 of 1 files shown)


web/src/artifact_cache.ts
```diff
@@ -353,6 +353,138 @@ export class ArtifactIndexedDBCache implements ArtifactCacheTemplate {
...
+ * https://developer.mozilla.org/en-US/docs/Web/API/File_System_API/Origin_private_file_system
+ */
+export class ArtifactOPFSCache implements ArtifactCacheTemplate {
+  private dirName?: string;
+  private fileVersion = 1;
+  private dirHandle: FileSystemDirectoryHandle | null = null;
+
+  constructor(dirName: string) {
+    // use the version suffix to allow upgrade if needed
+    this.dirName = `${dirName}-v${this.fileVersion}`;
```
KarolNov, line 367: I'm new to this repo, but it seems versioning is not really needed here and it's up for user to version the urls instead. However storing the `-vX` suffix would allow to do internal breaking changes in the file structure by just creating a new OPFS cache handler that uses different dir name `-v2`.

PR #17516, opened 2024-11-11
[RUNTIME][CLML] Dynamic backward compatibility

srkreddy1238:
We can now build one binary and use across targets Compiler pass config to choose target clml support version

Introducing compiler config "relay.ext.clml.target_version".
Partition pass will choose offloading ops based on target support.
This config enables choosing target version on python api as well as tvmc.

### DIFF
--- a/docker/bash.sh
+++ b/docker/bash.sh
@@ -247,6 +247,16 @@ while (( $# )); do
             shift 2
             ;;
 
+        -e)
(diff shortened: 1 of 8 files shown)


tqchen (approved):

tqchen:
Thanks @srkreddy1238 , as noted in other threads, we should phase out relay and move towards relax based pipelines, CLML could be a great candidate there

PR #17533, opened 2024-11-16
[Contrib] Remove CLML version print

MasterJH5574:
This PR removes the CLML version print during importing tvm.


MasterJH5574:
@srkreddy1238 Hi Siva, could you please take a look? I assume that print was unintentional.

tqchen (approved):

PR #17536, opened 2024-11-19
[Relax][Frontend][Onnx] Add auto_pad support for conv

HongHongHongL:
Add auto_pad support for conv.


tests/python/relax/test_frontend_onnx.py
```diff
@@ -1022,6 +1022,57 @@ def _verify_conv(input_shape, weight_shape):
     _verify_conv([3, 4, 32, 32, 32], [2, 4, 3, 3, 3])  # group=2
 
 
+@pytest.mark.parametrize("stride", [1, 2])
```
Hzfengsy, line 1025: Can we combine the tests into `test_conv`? Looks like they are almost the same:)

Hzfengsy: That's good

Hzfengsy (approved):

PR #17545, opened 2024-11-27
[Web]Allows setting powerPreference on webgpu

marschr:
This PR adds the `powerPreferece` argument to `detetectGPUDevice()`, while keeping the previous functionality as default, without breaking older code.

In hybrid GPU systems (mainly laptops) is sometimes needed to select the low-power integrated GPU in order to test model inference on https://github.com/mlc-ai/web-llm/ for other low-power devices or just to save on battery.

Here's the related line of code on the `web-llm` repository that references this function:
https://github.com/mlc-ai/web-llm/blob/082f04e4941ff4f6ef70731d244c69228948c7a1/src/service_worker.ts#L117


tqchen (approved):

tqchen:
Thanks @marschr !

PR #17548, opened 2024-11-29
[Relax] Expose BlockBuilder's Analyzer instance in Python

oskar-inceptron:
It is currently not possible to access a `BlockBuilder`'s `Analyzer` instance from Python code, which would be useful e.g. when converting from front-end representations to Relax.

This PR addresses that by adding a `get_analyzer` method to the Python `BlockBuilder` class. Behind the scenes, it uses the same FFI and code paths as when creating a standalone Python `Analyzer`, the only difference being where the underlying C++ instance originates from. To avoid object lifetime issues in Python code (which would be surprising given that it is a memory managed language), a C++ `BlockBuilder`'s `Analyzer` field is now a shared pointer.

### DIFF
--- a/include/tvm/arith/analyzer.h
+++ b/include/tvm/arith/analyzer.h
@@ -779,6 +779,13 @@ class TVM_DLL Analyzer {
    * \note Analyzer will call into sub-analyzers to get the result.
    */
   PrimExpr Simplify(const PrimExpr& expr, int steps = 2);
+
+  /*!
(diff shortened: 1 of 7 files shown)


Hzfengsy:
cc @tqchen

tqchen:
If we want to go towards this directly, a better approach is to make Analyzer an Object, so we can use ObjectPtr for this

PR #17571, opened 2024-12-24
[OPENCL][TEXTURE] Improved texture memory planning

srkreddy1238:
Motivated form the fact that textures can be allocated over a clBuffer object and the size of backing clBuffer can be computed based on hardware image pitch alignment.

This optimizes the overall memory allocation on device and helps greately the models with large memory requirements.

Improvised the graph memory planner to not differentiate buffer and texture storage tokens and reuse them across. The texture pool in OpenCL runtime is rebranded as memory pool that handles allocation for both buffer and image objects.

NDArray to DeviceAPI interface is extended with AllocDataSpaceView and FreeDataSpaceView. These new API's acommodates accessing same physical memory as clBuffer / clImage objects.


srkreddy1238:
Except the relay backend changes everything is reusable for Texture support in Relax.

srkreddy1238:
@tvm-bot rerun

srkreddy1238:
@tvm-bot rerun

include/tvm/runtime/device_api.h
```diff
@@ -135,12 +136,32 @@ class TVM_DLL DeviceAPI {
...
+  /*!
+   * \brief Create a new view with given spec over existing tensor.
+   * \param dev The device device to perform operation.
+   * \param data The source array.
+   * \param shape The shape of allocated tensor.
+   * \param dtype The type of elements.
+   * \param mem_scope The memory scope of allocated tensor.
+   * \return The allocated device pointer.
+   */
+  virtual void* AllocDataSpaceView(Device dev, void* data, ShapeTuple shape, DLDataType dtype,
```
tqchen, line 149: given this is specific to OpenCL, let us still strive to keep it within opencl allocator interface instead of going through the DeviceAPI

tqchen (changes requested):

include/tvm/runtime/device_api.h
srkreddy1238: I see this is the clean way to keep minimal changes in other modules (graph runtime, ndarray, memory manager ..etc). 

In Relax also , I am mapping alloc_storage to allocate cl_buffer and alloc_tensor does a view over it by this device API call.  (WIP Ref. https://github.com/srkreddy1238/tvm/commit/a6376b94799fd9c546334073fcbaeb468c3884df#diff-847ee73fb0b77db96cce920da6cbae223f6bdb026ea125514122e96630356c9b)

Later, this also allows easy path for CLML memory management going through TVM memory_manager interface and also features like GMEM (on chip memory of AdrenoGPU) support for TVM  ..etc.

Let me know if you have different advice, I can explore the possibilities.

srkreddy1238: In case of VM (Relax)

alloc_storage : Allocator->Alloc (scope global always) => DeviceAPI::AllocDataSpace =>StorageObj
alloc_tensor : StorageObj::AllocNDArrayScoped => DeviceAPI::AllocDataSpace => DeviceAPI::AllocDataSpaceView => NDArray

Ref. AllocNDArrayScoped  with destructor that calls FreeDataSpaceView for clean up.
https://github.com/srkreddy1238/tvm/commit/a6376b94799fd9c546334073fcbaeb468c3884df#diff-847ee73fb0b77db96cce920da6cbae223f6bdb026ea125514122e96630356c9b

tqchen: We might be able to in tihs case have StorageObj to refer back to a raw ptr of allocator, in this case  

`StorageObj => AllocNDArrayWithScope => allocator_-> CreateView(Storage storage)`. The main thing is you need an allocator specific dispatch to create such view, while previously we don't have to

srkreddy1238: Understood and

From ``CreateView`` to ``clCreateImage`` do we go through a packed helper function similar to https://github.com/apache/tvm/blob/ba6c743aa17494610ddb6be7db754f7d6e15fcc1/src/runtime/opencl/opencl_device_api.cc#L598 or ``DeviceAPI::AllocDataSpaceView`` ?

srkreddy1238: Another question,

Device allocator types are specified externally (from like VM Init as kPooled or kNaive). How could we enforce OpenCL device always use our new allocator ?

Or 

Do I use existing allocators (NaiveAllocator, PooledAllocator)  and CreateView will be specialized for device type kDLOpenCL alone ?

srkreddy1238: @tqchen 

Up for another review where, we have new allocator for OpenCL and now the allocation goes by the suggested flow

Have added a packed helper function like ``DeviceAllocator.opencl`` where the memory manager can query if the the device has specialized Allocator for a given type like Naive or Pooled.

We are now clear from DeviceAPI interface modification. Let me know how does it look now.

srkreddy1238:
@tvm-bot rerun

srkreddy1238:
@tvm-bot rerun

srkreddy1238:
@tvm-bot rerun

srkreddy1238:
@tqchen  good for review now :)

tqchen (approved):

tqchen:
thanks @srkreddy1238 , merging so we can have changes in before relay phasing out

PR #17584, opened 2025-01-07
[TIR][LoopPartition] enforcement on loop partition control

wrongtest-intellif:
The change wants to merge set of finegrained control improvement on loop partition pass.

1) Add an annotation intrinsic `T.ignore_loop_partition`, which mark the wrapped predicate not to take part in the loop partition phase. Since in sometimes loop partition would enlarge the generated code size and we want to preserve some predicates intentionally.

2) Add patch for EQ predicate partition. Try solve `a <= b && b <= a` instead if solve `a==b` directly failed.

3) Fix `TryPartition`'s return statement. Since the caller handle the return value by
  
    - If undef, try recursive solving.
    - If defined, treat as a successfully partioned result stmt.

### DIFF
--- a/include/tvm/tir/builtin.h
+++ b/include/tvm/tir/builtin.h
@@ -970,6 +970,9 @@ TVM_DLL const Op& vscale();
  */
 TVM_DLL const Op& get_active_lane_mask();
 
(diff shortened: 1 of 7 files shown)


tqchen:
@tvm-bot rerun

tqchen (approved):

PR #17585, opened 2025-01-08
[Relay]Disable InferType if it was done and no changes after previous pass

elvin-n:
This optimization allows to speedup PatternRewriter transformations by reusing of previous type inferred expression instead of perform InferType multiple times

The test_pass_simplify_expr.py tests on my machine started to pass for 40 seconds vs 116 seconds main branch state

### DIFF
--- a/src/relay/ir/dataflow_matcher.cc
+++ b/src/relay/ir/dataflow_matcher.cc
@@ -851,24 +851,32 @@ Expr PatternRewriter::Rewrite(const Array<DFPatternCallback>& callbacks, const E
   std::unordered_map<DFPatternCallback, bool, ObjectPtrHash, ObjectPtrEqual> done;
   do {
     last = post;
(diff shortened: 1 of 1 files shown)


echuraev (approved):
LGTM! Thanks!

echuraev:
@tvm-bot rerun

yongwww:
@tvm-bot rerun

tqchen:
@tvm-bot rerun

tqchen:
a minor note, we are moving towards phasing out relay and use relax

echuraev:
@tvm-bot rerun

elvin-n:
> a minor note, we are moving towards phasing out relay and use relax

It is true, and we must switch to relax and we are moving toward it. At the same time legacy relay flow sometimes has deeper and wider support. Do we know if relax flow covered all frontend capabilities which were covered by relay? Do we know which backed switched and have same support from performance point of view?

PR #17593, opened 2025-01-15
Fix GPU detection in PerStoreFeatureNode

moonquekes:
**Bug Description: Incorrect Assignment of `is_gpu` Property**

**Summary:**
In the context of feature extraction using `meta_schedule`, the `is_gpu` property was incorrectly assigned by directly comparing the target kind name to `"cuda"`. This approach is not only too specific but also potentially unsafe, as it does not account for different GPU types or handle cases where the target might be undefined.

**Incorrect Code:**
https://github.com/apache/tvm/blob/567eeed38bdbcefb68e36328af6ab1501a81d51e/src/meta_schedule/feature_extractor/per_store_feature.cc#L1395
```cpp
bool is_gpu = tune_context->target.value()->kind->name == "cuda";
```

**Problems with the Incorrect Code:**
1. **Lack of Generality:** Not all GPUs use CUDA (e.g., AMD GPUs may use ROCm or OpenCL).
2. **Specificity Issue:** Even on NVIDIA hardware, the target kind name isn't guaranteed to be `"cuda"` and can vary based on library versions or configurations.

**Correct Implementation:**
```cpp
auto& target_keys = tune_context->target.value()->keys;
bool is_gpu = std::find(target_keys.begin(), target_keys.end(), "gpu") != target_keys.end();
```

**Explanation of Correct Code:**
The corrected code checks if `"gpu"` exists within the list of keys associated with the target. This method is more robust because:
- It correctly identifies any target that includes `"gpu"` in its keys, regardless of the underlying hardware or software platform.
- It avoids potential null pointer dereferencing by not directly accessing `kind->name`.
- It is more flexible and can accommodate various types of GPU targets without hardcoding specific names.

### DIFF
--- a/src/meta_schedule/feature_extractor/per_store_feature.cc
+++ b/src/meta_schedule/feature_extractor/per_store_feature.cc
@@ -1392,7 +1392,8 @@ class PerStoreFeatureNode : public FeatureExtractorNode {
 
   Array<runtime::NDArray> ExtractFrom(const TuneContext& tune_context,
                                       const Array<MeasureCandidate>& candidates) {
-    bool is_gpu = tune_context->target.value()->kind->name == "cuda";
+    auto& target_keys = tune_context->target.value()->keys;
+    bool is_gpu = std::find(target_keys.begin(), target_keys.end(), "gpu") != target_keys.end();
     std::vector<runtime::NDArray> results;
     results.resize(candidates.size());
     std::unique_ptr<tir::group6::Feature> feature_group6 = nullptr;


moonquekes:
hi，@Hzfengsy.
Please review the code I submitted, thanks!

Hzfengsy (approved):
Thanks for the fix

yongwww:
@tvm-bot rerun

yongwww:
@tvm-bot rerun

yongwww (approved):

PR #17598, opened 2025-01-20
[TIR][FIX] update FlopEstimator to include missing nodes

PatrikPerssonInceptron:
Updated FlopEstimator to include missing nodes that would otherwise cause tuning to fail for certain TIR functions


src/tir/analysis/estimate_flops.cc
```diff
@@ -138,6 +138,7 @@ class FlopEstimator : private ExprFunctor<TResult(const PrimExpr& n)>,
   }
 
   TResult VisitExpr_(const BufferLoadNode* op) override { return TResult(); }
+  TResult VisitStmt_(const AttrStmtNode* op) override { return TResult(); }
```
wrongtest-intellif, line 141: Do we need `VisitStmt(op->body)` for attr annotation since there maybe valid mops under attr body?

wrongtest-intellif (changes requested):

src/tir/analysis/estimate_flops.cc
PatrikPerssonInceptron: My original thought was that the attribute node would only server to hold values for downstream passes such as codegen and would not be converted to executable code.

wrongtest-intellif (approved):

wrongtest-intellif:
@tvm-bot re-run

PR #17607, opened 2025-01-26
[FIX][TVMC] Fix the mixed precision conversion pipeline

srkreddy1238:
Fixed the mixed precision conversion pipeline issue.


tqchen:
Thanks @srkreddy1238 as we move towards relax based pipelines, likely we are also moving towards python API based pipeline for specific models as a single cli may not cover all usecases, so just want to note here

PR #17637, opened 2025-02-10
[PYTHON] Build cython by default

tqchen:
Historically we have been maintaining cython and ctypes for FFI. One reason was cython was not that available in all platforms then. Cython have quite good availability now and also enjoys great perf benefit. This PR changes the CMake to make cython as required dependency of the python package and build cython by default. It also removes ctypes testcases in favor of all cython tests.

If things go well, we can consider remove ctypes and focus on cython support onward.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -819,6 +819,18 @@ endif()
 # Custom targets
 add_custom_target(runtime DEPENDS tvm_runtime)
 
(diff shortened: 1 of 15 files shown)


tqchen:
tested on windows, which is a particular platform to watch out for

spectrometerHBH (approved):

PR #17641, opened 2025-02-11
Pick up vector length from 'zvlXXXb' (RVV) mattr for riscv

cbalint13:
This PR is an enhancement for #17631.

---

New functionality:

* This also parses optional 'zvlXXXb' attribute from ```-mattr``` arguments, if user would supply such.

Enhanchements:

* Separate ```GetVectorWidth()``` @LLVMTargetInfo a shared **public function**, reachable from python side too.
* A basic **codegen testcase** for riscv{32,64} targets is introduced, for now it checks RVV presence/absence.
* A target **parser testcase** for riscv{32,64} is also introduced, it checks sanity and VLEN relations.

---

This not alter any behavior except riscv targets.
This work here is a preliminary before enable TVM's SVE for riscv too.

### DIFF
--- a/tests/python/codegen/test_target_codegen_riscv.py
+++ b/tests/python/codegen/test_target_codegen_riscv.py
@@ -1,0 +1,49 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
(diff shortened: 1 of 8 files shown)


tests/python/codegen/test_target_codegen_riscv.py
```diff
@@ -0,0 +1,56 @@
...
+        A = te.placeholder((K, N), dtype="int8", name="A")
+        B = te.placeholder((K, N), dtype="int8", name="B")
+        k = te.reduce_axis((0, K))
+        C = te.compute(
+            (N,),
+            lambda n: te.sum(A[k, n].astype("int32") * B[k, n].astype("int32"), axis=[k]),
+            name="C",
+        )
+        s = te.create_schedule(C.op)
+        s[C].vectorize(s[C].op.axis[0])
```
tqchen, line 41: we are moving away from te schedule very soon, so let us instead use tir schedule

cbalint13: Reworked, it is even more simple now, I was expecting ```relay``` to go, not realizing the same about ```te```.

cbalint13:
This is ready for review.
Cc (LLVM folks) @quic-sanirudh , @srkreddy1238

tests/python/codegen/test_target_codegen_riscv.py
tqchen: te compute will stay, but we wanted to remove te.schedule

cbalint13:
@tvm-bot rerun

tqchen (approved):

PR #17642, opened 2025-02-11
[ONNX][RELAX] replace topi.split with relax.op.split in the onnx frontend

PatrikPerssonInceptron:
Replaces topi.split with relax.op.split in the ONNX frontend to align with the Relax operator API.

Fixes related ONNX frontend unit tests by:
- Updating InferStructInfoSplit to return a TensorStructInfo instead of a TupleStructInfo if the operator produces only one split, to be consistent with topi.split
- Updating the relax.split legalize callback to only perform the modulo != 0 check when the modulo result is a constant expression

Fixes split unit tests when only one tensor is returned by changing the expected struct info from a TupleStructInfo to a TensorStructInfo.

The reason why it is necessary to update InferStructInfoSplit to return a TensorStructInfo instead of a TupleStructInfo when the  split operator only returns a single tensor is because the MakeCallTIR api will unpack any tuple of struct info containing only one element. See  tvm/src/relax/op/op.cc line 577:

```
Expr MakeCallTIR(Expr func, Tuple args, Array<TensorStructInfo> out_sinfo_list,
                 Optional<Expr> packed_ints) {
  ...
  StructInfo out_sinfo{nullptr};
  if (out_sinfo_list.size() == 1) {
    out_sinfo = out_sinfo_list[0];
  } else {
    out_sinfo = TupleStructInfo({out_sinfo_list.begin(), out_sinfo_list.end()});
  }
```

### DIFF
--- a/python/tvm/relax/frontend/onnx/onnx_frontend.py
+++ b/python/tvm/relax/frontend/onnx/onnx_frontend.py
@@ -1717,7 +1717,7 @@ class Split(OnnxOpConverter):
         # When splits isnt specified divide evenly over axis.
         else:
             indices = attr["tvm_custom"]["num_outputs"]
-        return bb.emit_te(topi.split, inputs[0], indices, attr.get("axis", 0))
+        return relax.op.split(inputs[0], indices, attr.get("axis", 0))
 
     @classmethod
     def _impl_v13(cls, bb, inputs, attr, params):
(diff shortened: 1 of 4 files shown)


Hzfengsy (approved):

Hzfengsy:
Thanks @PatrikPerssonInceptron

PR #17654, opened 2025-02-14
[RELAX][BYOC] OpenCLML offload support for Relax

srkreddy1238:
This brings in OpenCLML offloading via BYOC path with available operators in Relax.
Adds Codegen tests for Mainline CI.
Also brings in pipeline definitions for Adreno targets.

### DIFF
--- a/cmake/modules/contrib/CLML.cmake
+++ b/cmake/modules/contrib/CLML.cmake
@@ -16,10 +16,12 @@
 # under the License.
 
 if(USE_CLML)
-    file(GLOB CLML_RELAY_CONTRIB_SRC src/relay/backend/contrib/clml/*.cc src/relax/backend/contrib/clml/*.cc)
+    file(GLOB CLML_RELAY_CONTRIB_SRC src/relay/backend/contrib/clml/*.cc)
+    file(GLOB CLML_RELAX_CONTRIB_SRC src/relax/backend/contrib/clml/*.cc)
     file(GLOB CLML_RUNTIME_MODULE src/runtime/contrib/clml/clml_runtime.cc)
     include_directories(SYSTEM "3rdparty/OpenCL-Headers")
     list(APPEND COMPILER_SRCS ${CLML_RELAY_CONTRIB_SRC})
+    list(APPEND COMPILER_SRCS ${CLML_RELAX_CONTRIB_SRC})
     if(NOT USE_CLML_GRAPH_EXECUTOR)
         list(APPEND COMPILER_SRCS ${CLML_RUNTIME_MODULE})
     endif()
--- a/cmake/modules/contrib/CLML.cmake
+++ b/cmake/modules/contrib/CLML.cmake
@@ -16,7 +16,7 @@
 # under the License.
 
 if(USE_CLML)
-    file(GLOB CLML_RELAY_CONTRIB_SRC src/relay/backend/contrib/clml/*.cc)
+    file(GLOB CLML_RELAY_CONTRIB_SRC src/relay/backend/contrib/clml/*.cc src/relax/backend/contrib/clml/*.cc)
     file(GLOB CLML_RUNTIME_MODULE src/runtime/contrib/clml/clml_runtime.cc)
     include_directories(SYSTEM "3rdparty/OpenCL-Headers")
     list(APPEND COMPILER_SRCS ${CLML_RELAY_CONTRIB_SRC})
--- a/cmake/modules/contrib/CLML.cmake
+++ b/cmake/modules/contrib/CLML.cmake
@@ -16,11 +16,9 @@
 # under the License.
 
 if(USE_CLML)
-    file(GLOB CLML_RELAY_CONTRIB_SRC src/relay/backend/contrib/clml/*.cc)
     file(GLOB CLML_RELAX_CONTRIB_SRC src/relax/backend/contrib/clml/*.cc)
     file(GLOB CLML_RUNTIME_MODULE src/runtime/contrib/clml/clml_runtime.cc)
     include_directories(SYSTEM "3rdparty/OpenCL-Headers")
-    list(APPEND COMPILER_SRCS ${CLML_RELAY_CONTRIB_SRC})
     list(APPEND COMPILER_SRCS ${CLML_RELAX_CONTRIB_SRC})
     if(NOT USE_CLML_GRAPH_EXECUTOR)
         list(APPEND COMPILER_SRCS ${CLML_RUNTIME_MODULE})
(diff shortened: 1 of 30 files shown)


tests/python/relax/backend/clml/utils.py
```diff
@@ -0,0 +1,122 @@
...
+    out = build_and_run(
+        clml_run_mod,
+        inputs,
+        tvm.target.adreno(clml=True),
+        rpc=rpc,
+        params_np=params_np,
+        load_path="vm_library_clml.so",
+        clml_enable=True,
+        clml_codegen=clml_codegen,
+    )
```
tqchen, line 121: To reduce CI time. Let us focus on codegen struct compare test and skip run compare tests for now. Perhaps we can introduce a special test mark for all run and compare tests which are integration

python/tvm/relax/backend/adreno/pipeline.py
```diff
@@ -0,0 +1,88 @@
...
+            dl.gpu.GeneralReduction(),
+            dl.gpu.Fallback(),
+        ),
+    ]
+
+
+def dataflow_lower_passes(target: tvm.target.Target):  # pylint: disable=unused-argument
+    """The default dataflow lowering passes for Adreno GPU backend."""
+    return [
+        relax.transform.RewriteDataflowReshape(),
```
tqchen, line 53: for settings that are same as gpu_generic. dataflow_lower_passes

consider do return gpu_generic.dataflow_lower_passes()

python/tvm/relax/backend/adreno/pipeline.py
```diff
@@ -0,0 +1,88 @@
...
+        relax.transform.RemovePurityChecking(),
+        relax.transform.CallTIRRewrite(),
+    ]
+
+
+def finalize_passes(target: tvm.target.Target):  # pylint: disable=unused-argument
+    """The default finalization passes for Adreno GPU backend."""
+    return [
+        relax.transform.StaticPlanBlockMemory(),
+        relax.transform.LowerAllocTensor(),
```
tqchen, line 64: for settings that are same as gpu_generic consider redirect to these calls

tests/python/relax/backend/clml/utils.py
```diff
@@ -0,0 +1,122 @@
...
+def run_compare(mod, inputs, params_np, rpc=None, clml_codegen=None):
+    clml_mod = copy.deepcopy(mod)
+    clml_run_mod = copy.deepcopy(mod)
+    mod = tvm.relax.transform.BindParams("main", params_np)(mod)
+    clml_mod = tvm.relax.transform.BindParams("main", params_np)(clml_mod)
+    clml_run_mod = tvm.relax.transform.BindParams("main", params_np)(clml_run_mod)
+
+    # Verify codegen
+    clml_mod = OpenCLMLOffLoad()(clml_mod)
+    verify_codegen(clml_mod, clml_codegen)
```
srkreddy1238, line 98: We do codegen check here (JOSN comparison).

tests/python/relax/backend/clml/utils.py
```diff
@@ -0,0 +1,122 @@
...
+
+    # On Mainline CI
```
srkreddy1238, line 100: Mainline will not have rpc, Hence we don't proceed beyond this point

tests/python/relax/backend/clml/utils.py
srkreddy1238: Done

srkreddy1238:
@Hzfengsy  for the ``OptimizeBatchnorm`` pass

``DecomposeOpsForInference`` couldn't help here as it results Conv2D+few elem_wise ops.

``OptimizeBatchnorm``  does folds batchnorm attributes into Conv2D weight and Bias, which we can offload as Fused Conv2D+Bias op of CLML.

Hzfengsy:
Thanks for pinging me. @srkreddy1238 

> OptimizeBatchnorm does folds batchnorm attributes into Conv2D weight and Bias, which we can offload as Fused Conv2D+Bias op of CLML.

IIUC it works as a CLML-specific pass. If so, could you please move it under adreno or clml folders, or at least add comments saying that it works only for CLML.

srkreddy1238:
> IIUC it works as a CLML-specific pass.

This is not CLML specific. Can be used for fusing Conv2D+BN -> Conv2D (Updated weight, bias) for inference case.

cmake/modules/contrib/CLML.cmake
```diff
@@ -17,9 +17,11 @@
 
 if(USE_CLML)
     file(GLOB CLML_RELAY_CONTRIB_SRC src/relay/backend/contrib/clml/*.cc)
```
tqchen, line 19: consider remove relay

python/tvm/relax/transform/optimize_batchnorm.py
```diff
@@ -0,0 +1,100 @@
...
+from tvm.ir.transform import PassContext
+from tvm.relax import Expr
+from tvm.relax.dpl import is_op, rewrite_call, wildcard, is_const, TupleGetItemPattern
+from tvm import relax, tir
+
+from . import function_pass
+
+
+@function_pass(opt_level=0)
+class OptimizeBatchnorm:
```
tqchen, line 29: fold_batch_norm_to_conv2d_for_inference.py

Please add a UT

tests/python/relax/backend/clml/test_ops.py
```diff
@@ -0,0 +1,1184 @@
...
+        groups=groups,
+        dtype=dtype,
+        has_bias=has_bias,
+        has_bn=has_bn,
+        has_activation=has_activation,
+        has_pad=has_pad,
+        is_depthwise=is_depthwise,
+    )
+
+    run_compare(mod, inputs, params_np, rpc, clml_codegen)
```
tqchen, line 299: i know that run_compare won't run if rpc do not exist, however, it is not obvious from the testcase. Would be great to split into two files

test_clml_codegen.py that calls into verify_codegen only and structural comparison.

test_op_exec_clml_codegen.py that calls into runs the ops. 

This would help us to further categorize tests. 

Thank you!

python/tvm/relax/transform/optimize_batchnorm.py
tqchen: This was an optimization can indeed can apply more broadly. One note is that this optimization is a special case of FoldScaleAxis optimization that folds scale into weights, so we want to add a comment here that we can replace by general FoldScaleAxis in future

tqchen (changes requested):

tqchen:
Thanks @srkreddy1238 for great effort, glad to see the new target aware pipeline helps to simplify the flow here. also cc @MasterJH5574

tqchen (approved):

srkreddy1238:
The codegen tests are added in unity->gpu. But I see this CI pipeline only builds GPU but doesn't run tests. Is this expected to be enabled soon ?

tqchen:
https://github.com/apache/tvm/blob/main/ci/jenkins/unity_jenkinsfile.groovy#L356 the build stage of unity pipeline does run relax tests. 

We can also followup to fold unity pipeline into the normal gpu pipeline

PR #17652, opened 2025-02-13
[Refactor] Phaseout python dependency package `decorator`

LeiWang1999:
This pull request focuses on removing the `decorator` dependency from various parts of the codebase and updating the relevant files accordingly as the decorator package consists of a single file and has not been maintained for a long time.

Codebase updates:

* [`python/gen_requirements.py`](diffhunk://#diff-e90cbe77cd3691891521000b8c60d3ed4dab7b2c8f7a923fd558bcfe0088c43aL67): Removed `decorator` from the base requirements and the list of additional requirements. [[1]](diffhunk://#diff-e90cbe77cd3691891521000b8c60d3ed4dab7b2c8f7a923fd558bcfe0088c43aL67) [[2]](diffhunk://#diff-e90cbe77cd3691891521000b8c60d3ed4dab7b2c8f7a923fd558bcfe0088c43aL244)
* [`python/tvm/_ffi/base.py`](diffhunk://#diff-f8e830321e516a385daf2b8b1a6011839398f1b33733bbd47a6031db853e13a2L135-R137): Replaced the usage of the external `decorator` package with an internal implementation of the `decorate` function.

* [`conda/recipe/meta.yaml`](diffhunk://#diff-eba2917991959691cc8fe3f70bcc7a181ec4b806191e3b3ded427f067fcd5a30L85): Removed `decorator` from the list of runtime dependencies.
* [`docker/Dockerfile.demo_opencl`](diffhunk://#diff-a8448388ae4a38c373d70794a600c0ea96e2bb5042bdf9831f6f072949ca7643L43-R43): Removed `decorator` from the list of Python packages installed via `pip3`.
* [`docker/install/ubuntu2004_install_python_package.sh`](diffhunk://#diff-7e95c1dfcc4947180381cc3a845e3f64367a680cd08b8731924912f836001aa7L28): Removed `decorator` from the list of Python packages installed via `pip3`.
* [`docker/install/ubuntu_install_python_package.sh`](diffhunk://#diff-6a76ef512408162f24fa70f390ce27e150c1cef7714c678b3f75b3ba41e4bce2L28): Removed `decorator` from the list of Python packages installed via `pip3`.
* [`docs/install/from_source.rst`](diffhunk://#diff-fb60f53a75475f5774d6d0a2ec8c35c89b1b34255c19b2c6bf0a0404775efce8L212-R212): Updated the installation instructions to remove `decorator` from the list of extra Python dependencies.

PR #17661, opened 2025-02-17
[Refactor] Phrase out python dependency `decorator`

LeiWang1999:
Remove the 'decorator' package from various installation scripts, requirements, and documentation. Update the FFI base module to use a local decorator implementation instead of the external package.

Previous pull request #17652


LICENSE
```diff
@@ -222,6 +222,7 @@ BSD 2-clause License
 
 3rdparty/picojson
 3rdparty/dmlc-core/include/dmlc/concurrentqueue.h
+python/tvm/_ffi/decorator.py
```
tqchen, line 225: sorry i missed this part. Can we simply remove decorator dep via functools or do we still need the original decorator impl here? Would be good to understand, ideally we either remove, or keep the thirdparty dep for a bit

tqchen (changes requested):

LICENSE
tqchen: would be good to get a testcases that altnerative approach failed to understand here

python/tvm/_ffi/base.py
```diff
@@ -132,9 +132,11 @@ def decorate(func, fwrapped):
     fwrapped : function
         The wrapped function
     """
-    import decorator
+    @functools.wraps(func)
```
tqchen, line 135: if we confirm functools.wraps can do the job, perhaps we can simply replace all the callsite with functools and remove this function

LeiWang1999:
The CI test failed, but it was likely not caused by this pr.

python/tvm/_ffi/base.py
LeiWang1999: The latest commit has replaced all decorator call sites with `functools.wraps`, cc @tqchen

tqchen (approved):

tqchen:
Thanks @LeiWang1999 for great work, this is now merged!

PR #17666, opened 2025-02-19
[Refactor] Improve TargetHasSVE function with optional target handling

LeiWang1999:
pull request #16893 refactored the TargetHasSVE function to support target assignment instead of using Target::Current(). This introduces a breaking API change for older projects that rely on TVM. However, the issue can be easily fixed by adding an Optional tag and a default value.


cbalint13 (approved):

cbalint13:
Thanks @LeiWang1999 , it looks good to me. 
I also planed some similar plus to enable it for riscv if "v" feature is present.

PR #17668, opened 2025-02-20
[Relax][PyTorch] Add support for and_, lshift, min, or_, rshift, xor ops

tlopex:
This PR supports several Pytorch binary ops such as lshift and rshift for Relax


tqchen:
@tlopex given PT is moving heavily towards supporting torch.export, would be great for us to ensure coverage of both path onwards

tqchen:
aka both with respective testcases
- https://github.com/apache/tvm/blob/main/python/tvm/relax/frontend/torch/exported_program_translator.py
- https://github.com/apache/tvm/blob/main/python/tvm/relax/frontend/torch/fx_translator.py

with testcases on them, ideally some common impl can be bought onto

https://github.com/apache/tvm/blob/main/python/tvm/relax/frontend/torch/base_fx_graph_translator.py

tqchen:
also cc @Hzfengsy  and @mshr-h

Hzfengsy (approved):

PR #17676, opened 2025-02-24
[Docker] Use Torch GPU on gpu device

yongwww:
Use torch GPU version on GPU docker.


yongwww:
cc: @tqchen @jinhongyii

PR #17680, opened 2025-02-25
[CI] update images to 20250225-035137-aeadc31c

yongwww:
Update the CI Docker images to the latest built versions, and the torch gpu is supposed to be installed on gpu container.

cc: @MasterJH5574 @tqchen @mshr-h

### DIFF
--- a/ci/jenkins/docker-images.ini
+++ b/ci/jenkins/docker-images.ini
@@ -17,10 +17,10 @@
 
 # This data file is read during when Jenkins runs job to determine docker images.
 [jenkins]
-ci_arm: tlcpack/ci-arm:20250214-034537-bd1411f8
-ci_cpu: tlcpack/ci_cpu:20250214-034537-bd1411f8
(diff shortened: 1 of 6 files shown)


tqchen (approved):

mshr-h:
The unity errors may be caused by the hidden changes in the Core Aten IR in PyTorch.
The temporal solution would be to stick to torch==2.4.1.

MasterJH5574:
Just a random thought: could we have different expected IRModules for different torch versions? Since torch version is available by `torch.__version__`, I'm wondering whether providing different expected IRModules for different versions can enable a bit better compatibility.

mshr-h:
It's possible to support multiple versions of torch. But I don't think we need to maintain the compatibility with older versions of torch since Core Aten Ops is under active development so we only support the latest one to reduce the maintenance cost.

yongwww:
@MasterJH5574 @mshr-h  Thank you for the extremely helpful input! I’ve skipped several tests in this PR, and we can re-enable them in follow-up PRs. It would be great to unblock the Docker images in this PR.

PR #17682, opened 2025-02-26
[Relax] Add support for func attr inheritance in SplitLayoutRewritePreproc

HongHongHongL:
Add support for func attr inheritance in SplitLayoutRewritePreproc.


tests/python/relax/test_transform_split_layout_rewrite_preproc.py
```diff
@@ -77,6 +78,7 @@ def tir_func_weight_prepack(
             W: T.Buffer((224, 224), "float32"),
             W_rewrite: T.Buffer((4, 4, 56, 56), "float32"),
         ):
+            T.func_attr({"layout_free_buffers": [1]})
```
Hzfengsy, line 81: `layout_free_buffers` should be removed in this pass

PR #17688, opened 2025-02-27
Fix relax.ccl.scatter_from_worker0 assert

oskar-inceptron:
The current code asserts that floormod(input_dims[0], num_shards) can not be proven to be true, which has two problems:

- It is unclear what it means to prove floormod(..). Prove that floormod's return value is truthy, i.e. non-zero?
- It always checks the 0th dimension of the input shape, but the dimension index to be sharded is given by the "axis" parameter.
 
This commit fixes both of the above.


Hzfengsy:
cc @MasterJH5574 @tqchen

MasterJH5574 (approved):
Thank you @oskar-inceptron!

PR #17693, opened 2025-02-28
updated the assert in BindParams to allow tvm.relax.Constant

PatrikPerssonInceptron:
The c++ implementation of BindParams can accept tvm.relax.Constants, however the python interface assert only allows ndarrays. I added tvm.relax.Constant to the assert.

By allowing tvm.relax.Constants as input arguments to BindParams it is possible to avoid duplicating weights if there are several functions in the module that are using the same weight as an input parameter, such as prefill and decode in an LLM.


yongwww (approved):
pls fix the ci lint issue

PR #17695, opened 2025-03-02
[WASM] Update wasm include in accordance to kv cache revamp

CharlieFRuan:
Update the include for `wasm_runtime.cc` to reflect:
- https://github.com/apache/tvm/issues/17674


MasterJH5574 (approved):

CharlieFRuan:
Note: this does not require a rebuild of the TVMjs runtime library, nor a rebuild of all previous model WASMs. This is only needed to compile new model WASMs.

PR #17699, opened 2025-03-03
[WebGPU] Support warp-level shuffle primitives with subgroup

CharlieFRuan:
This PR supports warp-level shuffle primitives using the newly introduce `subgroup` in WebGPU. We then use them in the implementation of allreduce lowering.

The introduced primitives are:
- `subgroupShuffle()`
- `subgroupShuffleUp()`
- `subgroupShuffleDown()`

This PR largely follows the Metal counterpart:
- https://github.com/apache/tvm/pull/15401

Tested with Llama3.2-1B-q4f16_1 E2E with WebLLM. The dumped WebGPU kernel indeed contains subgroup shuffle primitives: https://gist.github.com/CharlieFRuan/cb54a8db0513ecbbc16c5de8df5ab845

Remaining TODOs:
- [ ] Measure speedup
- [ ] Be able to parameterize whether to use subgroup or not when targeting WebGPU, since not all devices support it
- [ ] Check `GPUFeatureName`'s inclusion of `subgroups` in `@webgpu/types`

Resources:
- https://github.com/gpuweb/gpuweb/blob/859fdd4a803a11e7b8de70483aa75c365be18b0e/proposals/subgroups.md
- https://www.w3.org/TR/WGSL/#subgroup-builtin-functions
- https://developer.chrome.com/blog/new-in-webgpu-134?hl=en

### DIFF
--- a/web/src/webgpu.ts
+++ b/web/src/webgpu.ts
@@ -110,10 +110,16 @@ export async function detectGPUDevice(powerPreference: "low-power" | "high-perfo
       );
     }
 
-    const requiredFeatures: GPUFeatureName[] = [];
+    // TODO(Charlie): cannot type annotate because @webgpu/types
+    // does not have "subgroups" as GPUFeatureName yet
+    // const requiredFeatures: GPUFeatureName[] = [];
+    const requiredFeatures = [];
     // Always require f16 if available
     if (adapter.features.has("shader-f16")) {
       requiredFeatures.push("shader-f16");
+    }
+    if (adapter.features.has("subgroups")) {
+      requiredFeatures.push("subgroups");
     }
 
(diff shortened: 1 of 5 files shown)


web/src/webgpu.ts
```diff
@@ -110,11 +110,17 @@ export async function detectGPUDevice(powerPreference: "low-power" | "high-perfo
       );
     }
 
-    const requiredFeatures: GPUFeatureName[] = [];
+    // TODO(Charlie): cannot type annotate because @webgpu/types
```
beaufortfrancois, line 113: @webgpu/types 0.1.55 should work now. See https://github.com/gpuweb/types/pull/167

CharlieFRuan: Great, thanks!

PR #17700, opened 2025-03-03
[relax] Fix tree attention for Qwen2-1.5 models

Hzfengsy:
Fix the compilation error(https://github.com/mlc-ai/mlc-llm/issues/3143) for Qwen2-1.5 models in the tree attention implementation for vulkan backend.

cc @spectrometerHBH @vinx13


Hzfengsy:
One additional note: this PR provides an immediate fix for the issue, but it doesn't address the underlying problem - the simplifier can potentially cause integer overflow. For illustration, here's a minimal reproducible example:

```python
import tvm

x = tvm.tir.Var("x", "int32")
# Creating an expression that triggers integer overflow during simplification
expr = (tvm.tir.Div(x + 1073741826, 3) - 357913942) * 1536
ana = tvm.arith.Analyzer()
print(ana.simplify(expr))
```

cc @tqchen

vinx13 (approved):

MasterJH5574 (approved):
Thanks for the fix!

PR #17704, opened 2025-03-04
[Runtime][Target] Add C module for host

Hzfengsy:
This PR adds a C host module for simple host tasks without LLVM dependency. LLVM dependency is one of the most troublesome for TVM distribution, while it's not necessary for simple host tasks.

This PR plays a role as the first step towards the goal by supporting the c runtime.

cc @LeiWang1999 @tqchen

### DIFF
--- a/python/tvm/contrib/cc.py
+++ b/python/tvm/contrib/cc.py
@@ -21,7 +21,11 @@ import subprocess
 
 # pylint: disable=invalid-name
 import sys
+import tempfile
 from typing import Dict
+
+import tvm
+from tvm._ffi.libinfo import find_include_path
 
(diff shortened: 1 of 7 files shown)


tqchen:
thanks @Hzfengsy , just want to also cross link the executable intetface here https://discuss.tvm.apache.org/t/unified-python-first-compilation-flow-through-tvm-compile/18096 which should in theory help us to directly leverage the original c source module, and the jit happens during executable.jit, would be good if we can consolidate toward that approach

Hzfengsy:
Close it as https://github.com/apache/tvm/pull/17710 and interface `tvm.compile` could handle c host as a more generic way.

PR #17711, opened 2025-03-06
[Fix][dlight] add an explicit reduction loop check in Reduce

PatrikPerssonInceptron:
Added an explicit check in DLight Reduction to verify that the block has a reduction loop since this is assumed  in later stages.


Hzfengsy:
Thanks for the fix. Could you please add regression test for it?

Hzfengsy (approved):

Hzfengsy:
Thanks @PatrikPerssonInceptron !!

PR #17714, opened 2025-03-06
[FFI] Phase out ctypes fallback in favor of cython

tqchen:
In the past we maintained both ctypes fallback and cython path for our FFI. One reason was cython was not that widely available then. The situation has changed, so this PR removes the ctypes fallback so we can focus on building cython based core for FFI.

Notably, the cython core still leverages ctypes, especially for some of the features that do not need fast path, and primarily uses cython to speedup core FFI features with fast path.

With cython, we can get most FFI costs down to level of roughtly 1e-6 sec, which aligns with most of the costs like overhead of a numpy operation.


tqchen:
@tvm-bot rerun

yongwww (approved):

PR #17716, opened 2025-03-07
Fix the get_target_compute_version for sm >= 100

yongwww:
The sm_100 or larger will be parsed to 1.0.0, breaking some dependent passes. The expected major version should be 10.
ref: the GPU Compute Capability: https://developer.nvidia.com/cuda-gpus

cc: @MasterJH5574 @cyx-6


MasterJH5574 (approved):
LGTM.

PR #17717, opened 2025-03-08
[Relax][PyTorch] Add support for ndarry_size ops

tlopex:
Add ndarry_size operators to Relax and corresponding Pytorch frontend.


Hzfengsy:
I wonder if a new relax operator is needed. In this case, we can directly extract information from struct info

PR #17720, opened 2025-03-08
Allow softmax to work on a large tensor when dimension is not the last one

hugolatendresse:


tqchen:
let us instead to work and allow scheduling to work correctly for non-last dimension cases

hugolatendresse:
> let us instead to work and allow dlight to work correctly for non-last dimension cases

Sounds good, closing the PR

MasterJH5574:
Fixed in #17754

PR #17741, opened 2025-03-12
[Codegen][CUDA] Fix codegen of cast among vector bfloat16, fp8 and fp4

MasterJH5574:
This PR fixes the CUDA code generation for fp8 (also fp4) and bfloat16. We added a few vector data conversion util functions.


cyx-6 (approved):
LGTM, verified on B100:)

PR #17743, opened 2025-03-12
Fix: Change variable i to x in split operation in cross_compilation_and_rpc.py

HGboda:
Changes Made
Fixed the variable name in the split operation in cross_compilation_and_rpc.py, changing the undefined variable 'i' to the correct variable 'x'.

Problem
The original code was using an undefined variable 'i' in the split operation, which could cause runtime errors when executing the code.

Solution
Changed the variable 'i' to the correct variable 'x' to ensure the code functions properly.


tlopex:
LGTM. Thanks! @HGboda 
Could you please resolve the conflict so that this PR can pass CI.

tqchen (approved):

PR #17749, opened 2025-03-15
[Relax] Move TIR backend to gpu_generic

Hzfengsy:
This PR moves the TIR backend code to the gpu_generic backend directory structure. The main changes include:
- Relocating TIR implementation files from relax/backend_tir to relax/backend/gpu_generic/tir
- Updating import paths in dispatch_sampling.py and dispatch_sort_scan.py
- Removing unused files (pattern.py, cutlass.py and related tests)


Hzfengsy:
I removed tir_cutlass backend as there is no code that uses it. Please let me know if it is still useful. cc @spectrometerHBH

tqchen:
thanks @Hzfengsy , given gpu_generic likely already uses tir by default, we don't need the tir subnamespace here, otherwise LGTM

tqchen (approved):

PR #17757, opened 2025-03-16
[Relax] Tensor.split with uneven tensors

hugolatendresse:
We used to have an assertion that when using torch.split, the dimension to be split must be an multiple of the split size. This is not required in pytorch: the last tensors will simply be smaller. This PR allows that capability.


hugolatendresse:
cc: @MasterJH5574 this is ready for your review

MasterJH5574 (approved):
LGTM!

PR #17761, opened 2025-03-18
[FIX][RELAX] fix fusion of transpose + matmul when constant  weight 

PatrikPerssonInceptron:
Applying the FuseTransposeMatmul pass on a matmul with a constant weight  would fail since the bind_constants defaults to true in the FuseOpsByPattern pass . This would lead the constant to be moved in to the fused function instead of being an input param, however on line 168 in /tvm/relax/transform/fuse_transpose_matmul.py

```pyhton
out_dtype = function.ret_struct_info.dtype
                return self.builder_.call_te(
                    te_transposed_matmul,
                    call.args[1],
                    call.args[0],
                    primfunc_name_hint="NT_matmul",
                )
```

, called later in the FuseTransposeMatmul pass, both input and weight are expected to be inputs to the fused function leading to a crash.

I fixed it by explicitly setting bind_constants to false and I added a unit test to verify correct behavior when the weight is a constant.

### DIFF
--- a/python/tvm/relax/transform/fuse_transpose_matmul.py
+++ b/python/tvm/relax/transform/fuse_transpose_matmul.py
@@ -41,7 +41,8 @@ class FuseTransposeMatmul:  # pylint: disable=too-few-public-methods
                     "transpose_matmul_fuse",
                     *_pattern(),
                 ),
-            ]
+            ],
+            bind_constants=False,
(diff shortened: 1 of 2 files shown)


Hzfengsy (approved):
Thanks @PatrikPerssonInceptron

PR #17773, opened 2025-03-23
[Relax] Support TupleWrapper for Relax Expr

tlopex:
This PR supports TupleWrapper for Relax Expr.


Hzfengsy (changes requested):
I wonder if it's necessary, as:
- it is only a helper function but not the data structure for core IR
- the current relax tuple has already support `__len__` and `__getitem__`

tlopex:
Thanks. I'll close this PR @Hzfengsy

PR #17778, opened 2025-03-25
[Relax][PyTorch] Support log2, log10 and log1p ops for ExportedProgram importer

tlopex:
This pr supports `log2`, `log10` and `log1p` ops for ExportedProgram importer


MasterJH5574 (approved):
LGTM.

PR #17787, opened 2025-03-28
[Relax][PyTorch] Add support for index_select

hugolatendresse:
Adding support in the exported translator for index_select.default, which was already supported by the FX translator. This is in the hope that index_select will be useful for index.Tensor, which is more complex and not currently supported by the FX translator.


hugolatendresse:
@tvm-bot rerun

hugolatendresse:
I deleted this branch by accident, which closed the PR. I reopened the same PR here 17790

PR #17806, opened 2025-04-03
[Relax][PyTorch] Add Softplus Op Support for Exported Program and FX graph

Deivanayaki-S:
This PR adds support for the Softplus operation in both the Exported Program and FX graph Relax frontend pipeline.

### DIFF
--- a/include/tvm/relax/attrs/nn.h
+++ b/include/tvm/relax/attrs/nn.h
@@ -461,8 +461,8 @@ struct SoftplusAttrs : public tvm::AttrsNode<SoftplusAttrs> {
   double threshold;
 
   TVM_DECLARE_ATTRS(SoftplusAttrs, "relax.attrs.SoftplusAttrs") {
-    TVM_ATTR_FIELD(beta).describe("It controls the curvature; higher values make the transition sharper.");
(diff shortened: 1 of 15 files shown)


Hzfengsy (commented):
Thanks for the PR. As you added a new relax op, please also expose it in tvm script frontend. To be sepefic: at file https://github.com/apache/tvm/blob/main/python/tvm/script/ir_builder/relax/ir.py.

Also please add unit tests:
1. op tests, in tests/python/relax/test_op_nn.py
2. tvm sciprt tests, in tests/python/relax/test_frontend_nn_op.py:394

Deivanayaki-S:
I've added the unit tests in test_op_nn.py and test_frontend_nn_op.py as requested. However, regarding exposing the op in the TVMScript frontend (in tvm/script/ir_builder/relax/ir.py), I noticed that only core ops are defined there, and nn ops don't seem to be exposed directly in that file.
Could you please clarify whether nn ops like softplus should also be registered there?

Hzfengsy:
> Could you please clarify whether nn ops like softplus should also be registered there?

Thanks for pointing it out. I apologize that nn ops do not need to be explicitly exposed in that file, as the `nn` namespace is exposed as a whole unit.

Hzfengsy (approved):
LGTM

PR #17808, opened 2025-04-04
[Install] Fix error during python/tvm installation

vacu9708:
# Expected behavior
"pip install -e /path-to-tvm/python" successfully installs the TVM package as documented on https://tvm.apache.org/docs/install/from_source.html#

# Issue
~~If "libtvm.so" and "libtvm_runtime.so" remain in "python/tvm" for some reason,~~
If a previous pip installation was unsuccessful and the cleanup was not performed, running "pip install -e /path-to-tvm/python" fails because shutil.copy() and shutil.copytree() raise exceptions due to the remaining temporary files.

# How I encountered the issue
I ran "pip install" without "sudo" first, and it failed without "sudo". (probably because I was not in a virtual environment)
Here the cleanup code couldn't be executed

# Error log
```
  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [21 lines of output]
      Use git describe based version 0.20.dev125+g8ad9bffb1
        File "/usr/lib/python3/dist-packages/pip/_vendor/pep517/in_process/_in_process.py", line 363, in <module>
          main()
        File "/usr/lib/python3/dist-packages/pip/_vendor/pep517/in_process/_in_process.py", line 345, in main
          json_out['return_val'] = hook(**hook_input['kwargs'])
        File "/usr/lib/python3/dist-packages/pip/_vendor/pep517/in_process/_in_process.py", line 130, in get_requires_for_build_wheel
          return hook(config_settings)
        File "/usr/lib/python3/dist-packages/setuptools/build_meta.py", line 162, in get_requires_for_build_wheel
          return self._get_build_requires(
        File "/usr/lib/python3/dist-packages/setuptools/build_meta.py", line 143, in _get_build_requires
          self.run_setup()
        File "/usr/lib/python3/dist-packages/setuptools/build_meta.py", line 158, in run_setup
          exec(compile(code, __file__, 'exec'), locals())
        File "setup.py", line 197, in <module>
          shutil.copy(path, os.path.join(CURRENT_DIR, "tvm"))
        File "/usr/lib/python3.10/shutil.py", line 417, in copy
          copyfile(src, dst, follow_symlinks=follow_symlinks)
        File "/usr/lib/python3.10/shutil.py", line 234, in copyfile
          raise SameFileError("{!r} and {!r} are the same file".format(src, dst))
      shutil.SameFileError: '/home/youngsik/Documents/open_source/AI/tvm/python/tvm/libtvm.so' and 'tvm/libtvm.so' are the same file
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: subprocess-exited-with-error
```

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -129,6 +129,12 @@ def _remove_path(path):
 LIB_LIST, __version__ = get_lib_path()
 __version__ = git_describe_version(__version__)
 
+if not CONDA_BUILD and not INPLACE_BUILD:
+    # Wheel cleanup
+    for path in LIB_LIST:
+        libname = os.path.basename(path)
+        _remove_path(f"tvm/{libname}")
+
 
 def config_cython():
     """Try to configure cython and return cython configuration"""
@@ -260,5 +266,5 @@ if not CONDA_BUILD and not INPLACE_BUILD:
     # Wheel cleanup
     os.remove("MANIFEST.in")
     for path in LIB_LIST:
-        _, libname = os.path.split(path)
+        libname = os.path.basename(path)
         _remove_path(f"tvm/{libname}")
--- a/python/tvm/_ffi/libinfo.py
+++ b/python/tvm/_ffi/libinfo.py
@@ -143,8 +143,18 @@ def find_lib_path(name=None, search_path=None, optional=False):
             ]
 
         name = lib_dll_names + runtime_dll_names + ext_lib_dll_names
-        lib_dll_path = [os.path.join(p, name) for name in lib_dll_names for p in dll_path]
-        runtime_dll_path = [os.path.join(p, name) for name in runtime_dll_names for p in dll_path]
+        lib_dll_path = [
+            os.path.join(p, name)
+            for name in lib_dll_names
+            for p in dll_path
+            if not p.endswith("python/tvm")
+        ]
+        runtime_dll_path = [
+            os.path.join(p, name)
+            for name in runtime_dll_names
+            for p in dll_path
+            if not p.endswith("python/tvm")
+        ]
         ext_lib_dll_path = [os.path.join(p, name) for name in ext_lib_dll_names for p in dll_path]
     if not use_runtime:
         # try to find lib_dll_path


Hzfengsy:
cc @tqchen

mshr-h:
@vacu9708 Thanks.
I hit the same error before, so solving it makes sense to me. Can we just overwrite the file/directory instead of skipping it?

vacu9708:
@mshr-h Your opinion makes sense. Overwriting instead of skipping would be the correct modification. I'll make the change shortly.

cchung100m (approved):
LGTM, thanks to @vacu9708 😄

tqchen (approved):

tqchen:
@mshr-h can you confirm if the patch fixes your error?

mshr-h:
Still fails...

Here's my repro.

```bash
$ python3 -m venv .venv
$ source .venv/bin/activate
$ pip3 install cmake ninja setuptools cython
$ cmake -S . -B build -G Ninja
$ cmake --build build
$ cp build/libtvm_runtime.so python/tvm/ # copy the shared library to raise shutil.SameFileError
$ cd python
$ python3 setup.py develop
Use git describe based version 0.20.dev158+gbf6121656
  File "/home/ubuntu/data/project/tvm_work/tvm/python/setup.py", line 197, in <module>
    shutil.copy(path, os.path.join(CURRENT_DIR, "tvm"))
  File "/usr/lib/python3.12/shutil.py", line 435, in copy
    copyfile(src, dst, follow_symlinks=follow_symlinks)
  File "/usr/lib/python3.12/shutil.py", line 240, in copyfile
    raise SameFileError("{!r} and {!r} are the same file".format(src, dst))
shutil.SameFileError: '/home/ubuntu/data/project/tvm_work/tvm/python/tvm/libtvm_runtime.so' and '/home/ubuntu/data/project/tvm_work/tvm/python/tvm/libtvm_runtime.so' are the same file
```

mshr-h:
It’s probably due to the search priority of `dll_path` in the `get_dll_directories()` function in [`python/tvm/_ffi/libinfo.py`](https://github.com/apache/tvm/blob/main/python/tvm/_ffi/libinfo.py#L43). For some reason, when a shared object already exists under `python/tvm/`, that shared object is chosen as the source for copying that causes the SameFileError.
It worked by increasing the priority of the build directory.

Changin from this
```python
    # Pip lib directory
    dll_path.append(os.path.join(ffi_dir, ".."))
    # Default cmake build directory
    dll_path.append(os.path.join(source_dir, "build"))
    dll_path.append(os.path.join(source_dir, "build", "Release"))
    # Default make build directory
    dll_path.append(os.path.join(source_dir, "lib"))
```
to this (prioritize build directory over pip lib directory)

```python
    # Default cmake build directory
    dll_path.append(os.path.join(source_dir, "build"))
    dll_path.append(os.path.join(source_dir, "build", "Release"))
    # Default make build directory
    dll_path.append(os.path.join(source_dir, "lib"))
    # Pip lib directory
    dll_path.append(os.path.join(ffi_dir, ".."))
```

vacu9708:
@mshr-h 
In my environment, adding the code for cleaning up the remaining files at the beginning resolved the issue.
So, I am afraid I cannot test your error case.

Would it be okay if I reflect your change to the commit apart from the cleanup part?

vacu9708:
@mshr-h Thank you for your analysis.
Based on your report, **I checked :**
- The problematic "python/tvm/libtvm_runtime.so" and "python/tvm/libtvm.so" are appended at line 146 and 147 in libinfo.py.
    - Appending these paths seems like a an unintended behavior because it is libraries from "build/" that should be used for installation, not libraries from "python/tvm/".
    - As a result, "lib_path" at line 50 in setup.py contains ['python/tvm/libtvm.so', 'build/libtvm.so', 'python/tvm/libtvm_runtime.so', 'build/libtvm_runtime.so']
    - The library choice logic in get_lib_path() chooses elements that appear earlier from the library list. Here problematic libraries in python/tvm are chosen.

**In my opinion**
- It looks more valid to just prevent the unintended libraries from being appended to the library list, instead of changing the order of being appended to the library list as you proposed.
    - This will resolve the shutil.SameFileError exception that shutil.copy() raises.
- Adding the cleanup logic earlier in case it is not executed due to a previous installation failure also looks valid to me
    - This will resolve the FileExistsErroexception that shutil.copytree() raises.

Would these two changes be valid to resolve the issue?

python/tvm/_ffi/libinfo.py
```diff
@@ -143,8 +143,18 @@ def find_lib_path(name=None, search_path=None, optional=False):
...
+            os.path.join(p, name)
+            for name in lib_dll_names
+            for p in dll_path
+            if "python/tvm" not in p
+        ]
+        runtime_dll_path = [
+            os.path.join(p, name)
+            for name in runtime_dll_names
+            for p in dll_path
+            if "python/tvm" not in p
```
mshr-h, line 156: `if not p.startswith("python/tvm")` would be better.

python/tvm/_ffi/libinfo.py
```diff
@@ -143,8 +143,18 @@ def find_lib_path(name=None, search_path=None, optional=False):
             ]
 
         name = lib_dll_names + runtime_dll_names + ext_lib_dll_names
-        lib_dll_path = [os.path.join(p, name) for name in lib_dll_names for p in dll_path]
-        runtime_dll_path = [os.path.join(p, name) for name in runtime_dll_names for p in dll_path]
+        lib_dll_path = [
+            os.path.join(p, name)
+            for name in lib_dll_names
+            for p in dll_path
+            if "python/tvm" not in p
```
mshr-h, line 150: `if not p.startswith("python/tvm")` would be better.

mshr-h (changes requested):
Overall looks good to me. Just minor changes are needed.

python/setup.py
```diff
@@ -129,6 +129,12 @@ def _remove_path(path):
 LIB_LIST, __version__ = get_lib_path()
 __version__ = git_describe_version(__version__)
 
+if not CONDA_BUILD and not INPLACE_BUILD:
+    # Wheel cleanup
+    for path in LIB_LIST:
+        _, libname = os.path.split(path)
```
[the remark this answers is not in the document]
vacu9708: I agree this makes the line more readable.
I copied this code from the cleanup logic at the end of setup.py, so I'll change this one too.

python/tvm/_ffi/libinfo.py
vacu9708: Your opinion sounds good for better readability.
dll_path contains absolute paths, so "python/tvm" is at the end of the path.
I used endswith() instead.

mshr-h:
Thanks a lot! @vacu9708

PR #17826, opened 2025-04-11
[Relax] Fix Torch frontends to report all the missing ops

Deivanayaki-S:
This PR fixes error handling in from_fx and from_exported_program to detect unsupported ops more effectively. Instead of stopping at the first missing op due to an assertion error, it now collects all missing function-type ops first, which helps identify all unsupported operations at once.


hugolatendresse:
cc @MasterJH5574 @tlopex

mshr-h (approved):

mshr-h:
We can put the check logic into `base_fx_graph_translator.py` since they're almost the same. But it's okay to do it in the following PR.

PR #17838, opened 2025-04-15
[BugFix][TIR] Schedule support reverse-inline with reduction blocks

MasterJH5574:
This PR fixes a bug in reverse-compute-inline of tir Schedule, which generates incorrect TIR after inlining a transpose block into a reduction block.


MasterJH5574:
cc @hugolatendresse

PR #17841, opened 2025-04-15
[Relax][PyTorch] Add support for norm op

tlopex:
This PR supports Pytorch `norm` op for Relax

### DIFF
--- a/tests/python/relax/test_frontend_from_fx.py
+++ b/tests/python/relax/test_frontend_from_fx.py
@@ -4518,7 +4518,7 @@ def test_norm():
 
     class Norm1(Module):
         def forward(self, x):
-            return torch.norm(x, p=float('inf'), dim=None, keepdim=False)
+            return torch.norm(x, p=float("inf"), dim=None, keepdim=False)
 
     @tvm.script.ir_module
     class Expected1:
@@ -4534,7 +4534,7 @@ def test_norm():
 
     class Norm2(Module):
         def forward(self, x):
-            return torch.norm(x, p=float('-inf'), dim=None, keepdim=False)
+            return torch.norm(x, p=float("-inf"), dim=None, keepdim=False)
 
     @tvm.script.ir_module
     class Expected2:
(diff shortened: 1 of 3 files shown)


tests/python/relax/test_frontend_from_fx.py
```diff
@@ -4513,5 +4513,143 @@ def main(
     verify_model(Narrow(), [([5, 3], "float32")], {}, Expected)
 
 
+def test_norm():
+    input_info = [([1, 3, 5, 3], "float32")]
+
+    class Norm1(Module):
```
mshr-h, line 4519: Can we define something like a more generic norm module and pass `p`, `dim`, and `keepdim` when instantiating?

```python
    class Norm(Module):
        def __init__(self, p, dim=None, keepdim=False):
            self.p = p
            self.dim = dim
            self.keepdim = keepdim
        def forward(self, x):
            return torch.norm(x, p=self.p, dim=self.dim, keepdim=self.keepdim)
```

tlopex: Sure, I'll modify the code like that soon. Thanks!

tlopex:
@tvm-bot rerun

tlopex:
@tvm-bot rerun

tests/python/relax/test_frontend_from_fx.py
```diff
@@ -4513,5 +4513,134 @@ def main(
...
+            with R.dataflow():
+                lv: R.Tensor((1, 3, 5, 3), dtype="float32") = R.multiply(inp_0, inp_0)
+                lv1: R.Tensor((), dtype="float32") = R.sum(lv, axis=None, keepdims=False)
+                lv2: R.Tensor((), dtype="float32") = R.sqrt(lv1)
+                gv: R.Tensor((), dtype="float32") = lv2
+                R.output(gv)
+            return gv
+
+    norms = [
+        (float("inf"), None, False),
```
mshr-h, line 4629: just for your info, you can define something like this for test cases.
```python
    norms = [
        ((float("inf"), None, False), Expected1),
        '(float("-inf"), None, False), Expected2),
```

mshr-h (approved):
LGTM! Thanks!

tests/python/relax/test_frontend_from_fx.py
tlopex: Got it! Thanks!!

PR #17852, opened 2025-04-17
[Relax][PyTorch] Sort.default

hugolatendresse:
Add support for sort.default in exported program translator. There was an existing _sort() function in base_fx_graph_translator.py, but it would return values only. Pytorch returns a tuple of values and indices, so that was corrected


hugolatendresse:
cc @MasterJH5574 @tlopex this is ready for review

hugolatendresse:
@tlopex @MasterJH5574 the conflicts have been resolved

MasterJH5574 (approved):

PR #17857, opened 2025-04-18
[Relax][PyTorch] Refactor norm op for ExportedProgram importer

tlopex:
This pr refactor linalg_vector_norm op for ExportedProgram importer and add tests for it.
This pr also refactors FX side tests.


tlopex:
cc @mshr-h

python/tvm/relax/frontend/torch/fx_translator.py
```diff
@@ -422,6 +422,48 @@ def _lerp(self, node: fx.Node) -> relax.Var:
             relax.op.add(start, relax.op.multiply(weight, relax.op.subtract(end, start)))
         )
 
+    ########## Statistical ##########
+
+    def _norm(self, node: fx.Node) -> relax.Var:
```
mshr-h, line 427: Looks like this function is exactly the same as the `_linalg_vector_norm` in exported program frontend. Can we create a shared func between fx and exported program?

mshr-h (changes requested):

python/tvm/relax/frontend/torch/fx_translator.py
tlopex: I separated them since the `_linalg_vector_norm` in the exported program doesn't support frobenius_norm but `torch.norm` in the fx does.

PR #17858, opened 2025-04-18
[Relax][PyTorch] Add copy_ op support in fxGraph

PhyG0:
This PR adds support for copy_ operation in fx graph and test scripts to validate the relax modules.
Adding this support extends the compatibility with YOLOv7 Model.

### DIFF
--- a/python/tvm/relax/frontend/torch/fx_translator.py
+++ b/python/tvm/relax/frontend/torch/fx_translator.py
@@ -476,6 +476,11 @@ class TorchFXImporter(BaseFXGraphImporter):
         self.env[node.args[0]] = filled
         return filled
 
+    def _inplace_copy(self, node: fx.Node) -> relax.Var:
+        dst = self.env[node.args[1]]
+        self.env[node.args[0]] = dst
+        return dst
+
     def _masked_scatter(self, node: fx.Node) -> relax.Var:
         x = self.env[node.args[0]]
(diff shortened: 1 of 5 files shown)


python/tvm/relax/frontend/torch/fx_translator.py
```diff
@@ -476,6 +476,11 @@ def _inplace_fill(self, node: fx.Node) -> relax.Var:
         self.env[node.args[0]] = filled
         return filled
 
+    def _inplace_copy(self, node: fx.Node) -> relax.Var:
+        dst = self.env[node.args[1]]
```
mshr-h, line 480: According to the official doc, the first arg is src. The logic seems correct but the naming of the intermediate variable is misleading.
https://pytorch.org/docs/stable/generated/torch.Tensor.copy_.html

mshr-h, line 480: According to the official doc, the first arg is src. The logic seems correct but the naming of the intermediate variable is misleading.
https://pytorch.org/docs/stable/generated/torch.Tensor.copy_.html

mshr-h (approved):

python/tvm/relax/frontend/torch/fx_translator.py
PhyG0: Thanks for pointing out. I've changed the variable name to 'src'.

PR #17864, opened 2025-04-20
[Relax][PyTorch] Support eye op for ExportedProgram importer

tlopex:
This pr supports `eye` op for ExportedProgram importer.

### DIFF
--- a/python/tvm/relax/frontend/torch/base_fx_graph_translator.py
+++ b/python/tvm/relax/frontend/torch/base_fx_graph_translator.py
@@ -1416,6 +1416,19 @@ class BaseFXGraphImporter(metaclass=abc.ABCMeta):
         x = self.env[node.args[0]]
         return self.block_builder.emit(relax.op.zeros_like(x))
 
+    def _eye(self, node: fx.Node) -> relax.Var:
+        args = self.retrieve_args(node)
+        if len(args) == 1:
+            n = args[0]
+            m = n
+        elif len(args) == 2:
+            n = args[0]
+            m = args[1]
+        else:
+            raise ValueError("Invalid number of arguments for eye")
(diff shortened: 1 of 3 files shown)


python/tvm/relax/frontend/torch/base_fx_graph_translator.py
```diff
@@ -1416,6 +1416,19 @@ def _empty_like(self, node: fx.Node) -> relax.Var:
...
 
+    def _eye(self, node: fx.Node) -> relax.Var:
+        args = self.retrieve_args(node)
+        if len(args) == 1:
+            n = args[0]
+            m = n
+        elif len(args) == 2:
+            n = args[0]
+            m = args[1]
+        else:
```
mshr-h, line 1427: I don't think we need to take care of this condition because the pytorch exporter does that.
So we can assume that the graph in the exported program doesn't contain wrong nodes.
Probably this is enough?
```python
n = args[0]
m = args[1] if len(args) > 1 else n
```

cc @Hzfengsy

tests/python/relax/test_frontend_from_exported_program.py
```diff
@@ -4377,5 +4377,26 @@ def main(
     verify_model(Narrow(), example_args, {}, Expected)
 
 
+def test_eye():
+    class Eye(Module):
+        def forward(self, input):
+            return torch.eye(3, 5, dtype=torch.float32)
```
mshr-h, line 4383: Need a testcase where only `n` is given.

mshr-h (changes requested):

tests/python/relax/test_frontend_from_exported_program.py
tlopex: Okay, I'll add that soon.

mshr-h (approved):

PR #17871, opened 2025-04-21
[Relax][Pytorch] Add support for bitwise_or op support

kavin-sai-krishna:
This PR adds support for the bitwise OR operation used in the Mistral/Mistral-3B-Instruct model.


kavin-sai-krishna:
If you'd like support for other bitwise operators as well, I'd be happy to add them.

mshr-h (approved):

PR #17875, opened 2025-04-22
[BugFix][Relax][Pytorch] Incorrect Handling of In-Place Ops in FX-Based TVM Frontend

### LINKED ISSUE #17874: [Bug] [FRONTEND][PYTORCH][FxGraph] Incorrect Handling of In-Place Ops in FX-Based TVM Frontend
### Expected behavior

When using `torch.fx` to convert a PyTorch model containing in-place operations (e.g., `bitwise_or_`), the resulting IR in TVM should accurately reflect the updated tensor and return the modified value.

### Actual behavior

Currently, the FX-based tracing results in incorrect IR where the original input tensor is returned, instead of the tensor updated by the in-place operation. This leads to a semantic mismatch.

Example:

```python
# Original PyTorch model
class Model(Module):
    def forward(self, input: torch.Tensor, other: torch.Tensor):
        input.bitwise_or_(other)
        return input
```

Produces this incorrect FX-derived IR:

```python
@R.function
def main(inp_0: R.Tensor((128, 128), dtype="int32"), inp_1: R.Tensor((128, 128), dtype="int32")) -> R.Tensor((128, 128), dtype="int32"):
    with R.dataflow():
        lv: R.Tensor((128, 128), dtype="int32") = R.bitwise_or(inp_0, inp_1)
        gv: R.Tensor((128, 128), dtype="int32") = inp_0  # Incorrect: should return lv
        R.output(gv)
    return gv
```

Whereas using `exported_program` gives the correct representation:

```python
@R.function
def main(input: R.Tensor((128, 128), dtype="int32"), other: R.Tensor((128, 128), dtype="int32")) -> R.Tuple(R.Tensor((128, 128), dtype="int32")):
    with R.dataflow():
        lv: R.Tensor((128, 128), dtype="int32") = R.bitwise_or(input, other)
        gv: R.Tuple(R.Tensor((128, 128), dtype="int32")) = (lv,)
        R.output(gv)
    return gv
```

* frontend
* bug

cc @shingjan

kavin-sai-krishna:
This PR resolves #17874. Currently, only one in-place operation (bitwise_or_) is affected, and it has been remapped to use self._binary_op_inplace instead of self._binary_op. Going forward, any additional in-place operations (e.g., bitwise_and_, bitwise_xor_, etc.) should also be explicitly mapped to self._binary_op_inplace to ensure correct handling.

### DIFF
--- a/python/tvm/relax/frontend/torch/fx_translator.py
+++ b/python/tvm/relax/frontend/torch/fx_translator.py
@@ -161,12 +161,16 @@ class TorchFXImporter(BaseFXGraphImporter):
                 return output
 
             elif isinstance(lhs, relax.expr.Constant):
-                output = call_binary_op(relax_op, lhs, relax.const(rhs, dtype=lhs.struct_info.dtype))
+                output = call_binary_op(
+                    relax_op, lhs, relax.const(rhs, dtype=lhs.struct_info.dtype)
+                )
                 self.env[node.args[0]] = output
                 return output
 
             elif isinstance(rhs, relax.expr.Constant):
-                output = call_binary_op(relax_op, relax.const(lhs, dtype=rhs.struct_info.dtype), rhs)
+                output = call_binary_op(
+                    relax_op, relax.const(lhs, dtype=rhs.struct_info.dtype), rhs
+                )
                 self.env[node.args[0]] = output
                 return output
 
--- a/python/tvm/relax/frontend/torch/fx_translator.py
+++ b/python/tvm/relax/frontend/torch/fx_translator.py
@@ -159,14 +159,17 @@ class TorchFXImporter(BaseFXGraphImporter):
                 output = call_binary_op(relax_op, lhs, rhs)
                 self.env[node.args[0]] = output
                 return output
+
             elif isinstance(lhs, relax.expr.Constant):
                 output = call_binary_op(relax_op, lhs, relax.const(rhs, dtype=lhs.struct_info.dtype))
                 self.env[node.args[0]] = output
                 return output
+
             elif isinstance(rhs, relax.expr.Constant):
                 output = call_binary_op(relax_op, relax.const(lhs, dtype=rhs.struct_info.dtype), rhs)
                 self.env[node.args[0]] = output
                 return output
+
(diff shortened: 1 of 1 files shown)


python/tvm/relax/frontend/torch/fx_translator.py
```diff
@@ -132,6 +132,54 @@ def convert(node: fx.Node) -> relax.Var:
...
+                    return relax.const(lhs, rhs.struct_info.dtype), rhs
+                else:
+                    assert False
+
+            def call_binary_op(op, lhs, rhs):
+                lhs, rhs = promote_binary_op_args(lhs, rhs)
+                return self.block_builder.emit(op(lhs, rhs))
+
+            lhs, rhs = self.retrieve_args(node)
+            if isinstance(lhs, relax.Var) or isinstance(rhs, relax.Var):
```
mshr-h, line 158: Can we avoid implementing separate converters for in-place and non-in-place ops? I think the following logic should handle both cases. What do you think? @kavin-sai-krishna 
```python
            if isinstance(lhs, relax.Var) or isinstance(rhs, relax.Var):
                output = call_binary_op(relax_op, lhs, rhs)
            elif isinstance(lhs, relax.expr.Constant):
                output = call_binary_op(
                    relax_op, lhs, relax.const(rhs, dtype=lhs.struct_info.dtype)
                )
            elif isinstance(rhs, relax.expr.Constant):
                output = call_binary_op(
                    relax_op, relax.const(lhs, dtype=rhs.struct_info.dtype), rhs
                )
            else:
                output = intrinsic_op(lhs, rhs)

            self.env[node.args[0]] = output
            return output
```

kavin-sai-krishna: You wouldn't do `self.env[node.args[0]] = output` for non-inplace operations, right? Also, having separate implementations for inplace and non-inplace ops would make the code easier to understand. What do you think?

mshr-h: Ah, right. I misunderstood.

mshr-h (approved):

mshr-h:
LGTM. Thanks!

PR #17587, opened 2025-01-09
Resolving inconsistency between attention/attention_bias

parsifal-47:
This is a fix for:
https://github.com/apache/tvm/issues/17486

Inconsistency between attention/attention_bias operations, after the fix, the script from https://github.com/apache/tvm/issues/17486
is printing proper IR.
Thank you!

PR #17883, opened 2025-04-23
fix test_attention

xinxilwl:
I fixed the bug below:
[17587](https://github.com/apache/tvm/pull/17587) fails on tvm.ir.assert_structural_equal in tests/python/relax/test_transform_legalize_ops_nn.py


yongwww:
cc: @MasterJH5574

tqchen:
@cyx-6 can you also help to cross check this

tqchen:
sorry for the delayed note, i take a closer look. In principle we should not pass Expr as part of Attrs and always treat Attrs as constant, as a result, we cannot take this change as it is. We can however, require the attention funciton to optionally takes a bias argument, and the printer should be updated to be able to print out the bias or remove bias as needed.

So the correct fix here would be to fix the printer and deduction function of attention op, if we allow optional bias in the attention operator

PR #17913, opened 2025-04-30
[Triton] Support latest `triton.compile` interface

MasterJH5574:
This PR updates the Triton JIT compilation so that it supports the latest `triton.compile` interface introduced in Triton 3.3.0.


MasterJH5574:
@tvm-bot rerun

Hzfengsy (approved):

Rhys-Q:
I noticed that there will be an issue with argument passing when constexpr is not placed last. For example:
```
def add(
    batch_size: tl.constexpr,
    x,
    y,
):
    pass
```

PR #17918, opened 2025-05-05
[Relax][PyTorch] Add ReLU6 Op Support for Exported Program and FX graph

Deivanayaki-S:
This PR adds support for the relu6 operation in exported program and fx graph translator torch frontend. By adding this support, the following torchvision models are now able to run successfully:

1. ofa_supernet_mbv3_w10
2. ofa_supernet_mbv3_w12
3. GhostNet

### DIFF
--- a/python/tvm/relax/frontend/torch/fx_translator.py
+++ b/python/tvm/relax/frontend/torch/fx_translator.py
@@ -639,9 +639,7 @@ class TorchFXImporter(BaseFXGraphImporter):
             nn.LogSoftmax: self._log_softmax_module,
             nn.PReLU: self._prelu_module,
             nn.ReLU: self._unary_op(relax.op.nn.relu),
-            nn.ReLU6: lambda node: self.block_builder.emit(
-                relax.op.clip(self.env[node.args[0]], 0, 6)
-            ),
+            nn.ReLU6: self._unary_op(relax.op.nn.relu6),
             nn.Sigmoid: self._unary_op(relax.op.sigmoid),
             nn.SELU: self._unary_op(relax.op.nn.selu),
             nn.SiLU: self._unary_op(relax.op.nn.silu),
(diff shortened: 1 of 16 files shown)


Deivanayaki-S:
@Hzfengsy Could you please review this PR?

python/tvm/relax/frontend/torch/fx_translator.py
```diff
@@ -639,9 +639,7 @@ def create_convert_map(
             nn.LogSoftmax: self._log_softmax_module,
             nn.PReLU: self._prelu_module,
             nn.ReLU: self._unary_op(relax.op.nn.relu),
-            nn.ReLU6: lambda node: self.block_builder.emit(
-                relax.op.clip(self.env[node.args[0]], 0, 6)
-            ),
+            nn.ReLU6: self._unary_op(relax.op.nn.relu6),
```
Hzfengsy, line 642: I wonder why we need relu6 op defined in relax? In other words, why we need to change the current implementation?

Deivanayaki-S: @Hzfengsy Since the current clip mapping only handled the FX translator’s nn.ReLU6, it left other mappings (relu6.default, relu6_.default, and functional.relu6 in FX) unaddressed. Therefore, I added ReLU6 as a Relax op to unify the implementation.

Hzfengsy: Thanks for explaining it. My concern is that relu6 should not be a relax operator as it can be represented by other ops (i.e. clip). 

We can keep the `relax.op.nn.relu6` interface as a sugar, but not register a new relax op. 
To be clear, it's better to remove the new definition in:
- src/relax/op/nn/nn.cc
- src/relax/op/nn/nn.h
- src/contrib/msc/framework/tvm/relax_opcode.cc
- python/tvm/topi/nn/elemwise.py
- python/tvm/relax/transform/legalize_ops/

And change the codes in python/tvm/relax/op/nn/nn.py
```python
return relax.op.clip(data, 0, 6)
```

tqchen:
agree with @Hzfengsy , we should simplify relax core op set, and not define relu6 as relax op

Deivanayaki-S:
@Hzfengsy @tqchen Thanks for pointing it out :), I've updated the code accordingly.

Hzfengsy (approved):

PR #17928, opened 2025-05-07
KV Cache Per Layer Sliding Window

joshua-j-hong:
Adds per layer sliding window functionality to the KV Cache. Currently there are some major bugs that should be resolved before merging. 

1. Throughput slows down significantly, even before the sliding window length is reached
2. Correctness issues past the sliding window length, potentially due to page indexing or some other KV cache issue


joshua-j-hong:
With some further testing and investigation, there is an additional MLC-LLM/TVM bug that is related to excessive prefilling (even without the per-layer sliding window changes outlined here) that is related to inference slowdown.

MasterJH5574:
Just see some conflicts with upstream. Likely we need to do a rebase.  Related changes are the recent FFI refactor and a namespace rename from `relax_vm` to `vm`.  I'll check the PR again after updating.

joshua-j-hong:
Conflicts and tests are fixed. Current plans are for a future change that will add optional parameters to the KVCache, some of which will be for layer sliding window. This will ensure that no hardcoding of values will be needed and the KVCache will be backwards compatible.

MasterJH5574 (approved):

PR #17932, opened 2025-05-09
[Relax][PyTorch] Add UpSample Bicubic Op Support for Exported Program and FX graph

Deivanayaki-S:
This PR introduces support for the upsample_bicubic2d operation in PyTorch, expanding the existing upsampling functionalities to include bicubic interpolation. The implementation aligns with the default behavior of PyTorch's torch.nn.functional.interpolate function, which utilizes a cubic convolution kernel with an alpha value of -0.75. Notably, the default alpha value has been adjusted from -0.5 to -0.75 to match the exact results produced by PyTorch's implementation. By adding this support, the following models are now able to run successfully.

1. hustvl/yolos-tiny
2. depth-anything/Depth-Anything-V2-Small-hf
3. yolo_wgisd
4. facebook/dinov2-base
5. OpenGVLab/InternViT-300M-448px


tests/python/relax/test_frontend_from_fx.py
```diff
@@ -3250,7 +3250,7 @@ def main(
                     method="nearest_neighbor",
                     coordinate_transformation_mode="asymmetric",
                     rounding_method="round",
-                    cubic_alpha=-0.5,
+                    cubic_alpha=-0.75,
```
Hzfengsy, line 3253: Why need to change the existing tests?

Hzfengsy (approved):
Overall LGTM, but a minor question

tests/python/relax/test_frontend_from_fx.py
Deivanayaki-S: @Hzfengsy Since the default value of cubic_alpha has changed, the corresponding test scripts for other operations may need to be updated to reflect this change.

PR #17939, opened 2025-05-09
[REFACTOR] Phase out legacy rust ffi

tqchen:
This PR phases out te legacy rust FFI as so we can focus on the new FFI, and phase out legacy C APIs.
The change will allow us to move away from the legacy FFI API and simplify maintainace. A new bindings can be built on top of new FFI later if there is momentum on the direction.


tqchen:
@tvm-bot rerun

Hzfengsy (approved):

Issue #17941, opened 2025-05-09
[Bug] tvm._ffi.base.TVMError: std: :bad_alloc


Cookiee235:
Currently, convolution ops (conv1d/conv2d) in the topi namespace can produce negative output dimensions with certain parameter combinations (e.g. large kernel/dilation). 
However, such invalid behaviours cannot be captured in the `topi` and `tvm.build`. It seems we should add validation to fail early with clear error messages when parameters would generate invalid output shapes.

### Actual behavior

```
Traceback (most recent call last):
  File "/data/qshenaf/remote_pc/TirFuzz/bugs/05-01_00-45/topi.nn.conv1d_4.py", line 11, in <module>
    tvm.nd.empty(op_output.shape, dtype=op_output.dtype, device=tvm.cpu(0))  # crash
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/qshenaf/envs/tvm/python/tvm/runtime/ndarray.py", line 444, in empty
    arr = _ffi_api.TVMArrayAllocWithScope(shape, dtype, device, mem_scope)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "tvm/_ffi/_cython/./packed_func.pxi", line 339, in tvm._ffi._cy3.core.PackedFuncBase.__call__
  File "tvm/_ffi/_cython/./packed_func.pxi", line 284, in tvm._ffi._cy3.core.FuncCall
  File "tvm/_ffi/_cython/./base.pxi", line 185, in tvm._ffi._cy3.core.CHECK_CALL
  File "/data/qshenaf/envs/tvm/python/tvm/_ffi/base.py", line 468, in raise_last_ffi_error
    raise py_err
tvm._ffi.base.TVMError: std: :bad_alloc
```

### Environment

Any environment details, such as: Operating System, TVM version, etc

### Steps to reproduce

```
import tvm
from tvm import te, topi, tir


data = te.placeholder((4, 2048, 32), dtype='int8', name='data')
kernel = te.placeholder((3, 2048, 1024), dtype='int8', name='kernel')
[...]

PR #17942, opened 2025-05-09
[TOPI] Add shape validation to prevent negative dimensions in conv operations

jikechao:
Fix the bug in https://github.com/apache/tvm/issues/17941


Cookiee235:
@tvm-bot rerun

Hzfengsy (approved):

Issue #17959, opened 2025-05-12
[Bug]  Unsupported Function Type 'frac.default' in ExportedProgram Translator


Pratheesh-04-MCW:
When attempting to convert a PyTorch exported program to TVM's Relax format using from_exported_program(), the process fails with an AssertionError indicating that the function type 'frac.default' is not supported.

### Environment
- tvm 0.21.dev26+g2ca6ec8a5
- torch 2.6.0

### Error Message
  warnings.warn("Can't initialize NVML")
Traceback (most recent call last):
  File "/home/pratheesh04/tvm/examples/relax_support/frac.py", line 16, in <module>
    relax_mod = from_exported_program(exported_program)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/pratheesh04/tvm/python/tvm/relax/frontend/torch/exported_program_translator.py", line 679, in from_exported_program
    return ExportedProgramImporter().from_exported_program(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/pratheesh04/tvm/python/tvm/relax/frontend/torch/exported_program_translator.py", line 544, in from_exported_program
    self._check_unsupported_func_type(nodes)
  File "/home/pratheesh04/tvm/python/tvm/relax/frontend/torch/base_fx_graph_translator.py", line 114, in _check_unsupported_func_type
    assert not missing_func_types, f"Unsupported function types {missing_func_types}"
           ^^^^^^^^^^^^^^^^^^^^^^
AssertionError: Unsupported function types ['frac.default']

### Expected behavior

Should generate the relax IR:

# from tvm.script import ir as I
# from tvm.script import relax as R

@I.ir_module
class Module:
    @R.function
    def main(x: R.Tensor((4,), dtype="float32")) -> R.Tuple(R.Tensor((4,), dtype="float32")):
        with R.dataflow():
            lv: R.Tensor((4,), dtype="float32") = R.frac(x)
            gv: R.Tuple(R.Tensor((4,), dtype="float32")) = (lv,)
            R.output(gv)
        return gv

### Steps to reproduce
import torch
from torch.export import export
from tvm.relax.frontend.torch import from_exported_program
import tvm
from tvm import relax

class FracModel(torch.nn.Module):
    def forward(self, x):
        return torch.frac(x)

# Create input tensor
[...]

PR #17962, opened 2025-05-12
added op support for frac op

Pratheesh-04-MCW:
This PR adds support for the frac op in exported program and fx graph translator torch frontend. By adding this PR we can fix the issue [#17959](https://github.com/apache/tvm/issues/17959)


Pratheesh-04-MCW:
@tvm-bot rerun

Issue #17899, opened 2025-04-27
[Bug]


peanutPod:
Thanks for participating in the TVM community! We use https://discuss.tvm.ai for any general usage questions and discussions. The issue tracker is used for actionable items such as feature proposals discussion, roadmaps, and bug tracking.  You are always welcomed to post on the forum first :smile_cat:

Issues that are inactive for a period of time may get closed. We adopt this policy so that we won't lose track of actionable issues that may fall at the bottom of the pile. Feel free to reopen a new one if you feel there is an additional problem that needs attention when an old one gets closed.

### Expected behavior

What you were expecting:
https://tvm.apache.org/docs/how_to/tutorials/e2e_opt_model.html#sphx-glr-download-how-to-tutorials-e2e-opt-model-py

### Actual behavior

What actually happened 

![Image](https://github.com/user-attachments/assets/37fd6cf1-9304-4d75-bb30-d3b6da5e20a6)

### Environment

Any environment details, such as: Operating System, TVM version, etc
tvm version: main

### Steps to reproduce

Preferably a minimal script to cause the issue to occur.
python e2e_opt_model.py

### Triage

Please refer to the list of label tags [here](https://github.com/apache/tvm/wiki/Issue-Triage-Labels) to find the relevant tags and add them below in a bullet format (example below).

* needs-triage

erdayang:
I've encountered the same problem and have no idea how to solve it either.

vacu9708:
I resolved the "Did you forget to bind" error which occurs in the pipeline "static_shape_tuning". But gpu_out at the end of the tutorial code also causes an error. I'll make changes after resolving this another error too.

PR #17971, opened 2025-05-13
[Doc][Relax][MetaSchedule] Fix errors in the e2e_opt_model.py tutorial

vacu9708:
# Summary
This PR resolves issue [#17899](https://github.com/apache/tvm/issues/17899) where the end-to-end tutorial `e2e_opt_model.py` fails with errors and cannot be executed successfully.

# Bug fixes
- Memory binding error("Did you forget to bind?")
This error occurs when the compiler attempts to use operations that are not properly bound to GPU memory, due to occasionally generated invalid GPU schedules that result in memory misalignment. (probably due to vectorization)
	- **Fix**: Added GPU code verification to SpaceGenerator() that filters out schedules involving memory misalignment.
- <class 'tvm.ir.container.Array'> has no attribute numpy
This error occurs because vm\["main"\]() returns an Array of NDArray
	- **Fix**: Extracted the NDArray out of the returned Array before calling .numpy().

### DIFF
--- a/jvm/README.md
+++ b/jvm/README.md
@@ -39,7 +39,7 @@ TVM4J contains three modules:
 - core
     * It contains all the Java interfaces.
 - native
-    * The JNI native library is compiled in this module. It does not link TVM runtime library (libtvm\_runtime.so for Linux and libtvm\_runtime.dylib for OSX). Instead, you have to specify `libtvm.so.path` which contains the TVM runtime library as Java system property.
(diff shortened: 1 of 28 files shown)


vacu9708:
Is this the intended behavior?

The comment in tvm.relax.static_shape_tuning_pipeline() says the expected output is in the form: vm["main"](input_data, *params).numpy(). However, gpu_out at the end of the code has numpy() at index 0 of the Array.

PR #17972, opened 2025-05-14
Fix g.costs

thaisacs:
Before this PR, I encountered the following error when tuning a resnet_18 task:

`
/home/thais/Dev/tvm/python/tvm/meta_schedule/cost_model/xgb_model.py:543: RuntimeWarning: invalid value encountered in divide
  [g.min_cost / g.costs for g in self.data.values()],
`

### DIFF
--- a/python/tvm/meta_schedule/cost_model/xgb_model.py
+++ b/python/tvm/meta_schedule/cost_model/xgb_model.py
@@ -537,13 +537,17 @@ class XGBModel(PyCostModel):
         self.last_train_size = self.data_size
 
         # Step 5. Re-train the model
-        self._train(
-            xs=list(itertools_chain.from_iterable([g.features for g in self.data.values()])),
-            ys=np.concatenate(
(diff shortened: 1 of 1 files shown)


cbalint13 (approved):
@thaisacs 

LGTM thanks for the fix !

cbalint13:
@thaisacs ,

This proposed fix here omits from the ratio any div by zero positions that before would happened and omitted anyway.

This looks to me equivalent in previous functionality, minus the undesired warning message :+1: 

---

A initial tempting idea of mine would be to give statistical mean to those '0' zeros, by replacing with something like```np.finfo(np.float32).max)``` , but this would run in machine overflows and choosing a "right" somewhat large but not at machine boundaries would become another problem, so :-1: for this idea.

cbalint13:
cc @Hzfengsy @spectrometerHBH @vinx13 @MasterJH5574

MrG9090 (approved):

PR #17974, opened 2025-05-14
Fix FLOP estimation for EvaluateNode by implementing VisitStmt_ handler

jikechao:
### issue desription
This pr aimed to fix https://github.com/apache/tvm/issues/17973. MetaSchedule's FLOP estimator fails when encountering EvaluateNode (e.g., from T.call_packed), throwing TVMError: Do not have a default for `tir.Evaluate`. This occurs because FlopEstimator lacked a handler for EvaluateNode, causing tuning to abort.

### Fix strategy

EvaluateNode typically wraps side-effect operations (e.g., T.call_packed for memory copies or external calls) that contribute negligible FLOPs. Returning 0 avoids overcounting while allowing tuning to proceed rather than crash.

cc @Hzfengsy @tqchen @vinx13


jikechao:
cc @merrymercy @jcf94 @comaniac

PR #17979, opened 2025-05-14
[FFI][REFACTOR] Update to distinguish as and cast

tqchen:
This PR updates the Any system to distinguish as and cast

- as function will run strict check and won't do any type conversion
- try_cast/cast will try to run the type conversion

We also updated the type traits to be consistent with the naming


tqchen:
cc @spectrometerHBH

spectrometerHBH (approved):

Issue #17965, opened 2025-05-13
[Bug] inconsistent results for the CPU and CUDA targets.


coffezhou:
### Expected behavior

TVM should output consistent results for the CPU and GPU targets.

### Actual behavior
For the following model:

![Image](https://github.com/user-attachments/assets/a95460b1-d289-4885-9d82-8ef27f993927)

when compile the model for the CPU target, the output is:
```c
cpu:  [[[[nan nan nan]
   [nan nan nan]
   [nan nan nan]]

  [[nan nan nan]
   [nan nan nan]
   [nan nan nan]]

  [[nan nan nan]
   [nan nan nan]
   [nan nan nan]]]]
```

However, when the target is CUDA, the output is:
```c
gpu:  [[[[ 9.5653236e-01  8.9820576e-01  8.9820576e-01]
   [ 9.5653236e-01 -3.4028231e+38 -3.4028231e+38]
   [-3.4028231e+38 -3.4028231e+38 -3.4028231e+38]]

  [[ 9.5653236e-01  8.9820576e-01  8.9820576e-01]
   [ 9.5653236e-01 -3.4028231e+38 -3.4028231e+38]
   [-3.4028231e+38 -3.4028231e+38 -3.4028231e+38]]

  [[ 9.5653236e-01  8.9820576e-01  8.9820576e-01]
   [ 9.5653236e-01 -3.4028231e+38 -3.4028231e+38]
   [-3.4028231e+38 -3.4028231e+38 -3.4028231e+38]]]]
```

### Environment

OS: Ubuntu 20.04
TVM: 0.21.dev0(bcb68b130)
CUDA: 11.8
GPU: NVIDIA GeForce RTX 3080

### Steps to reproduce
This bug can be reproduced by the following code with the model in the attachment.
```python
import sys

import numpy as np
import onnx
import onnxruntime

import tvm
import tvm.testing
from tvm import relax
from tvm.relax.frontend.onnx import from_onnx

import pickle

            
def main():
    onnx_model = onnx.load("a249.onnx")
    
    with open("inputs.pkl", "rb") as fp:
        inputs = pickle.load(fp)
       
    # Convert the onnx model into relax through the onnx importer.
    tvm_model = from_onnx(onnx_model, keep_params_in_input=True)
    # Convert operators for inference mode.
    tvm_model = relax.transform.DecomposeOpsForInference()(tvm_model)
    # Legalize any relax ops into tensorir.
[...]

PR #17985, opened 2025-05-18
[Codegen] Resolve issue #17965 where the same model produces different outputs on the LLVM (CPU) and CUDA (GPU) backends

vacu9708:
# Summary
This PR resolves issue [#17965](https://github.com/apache/tvm/issues/17965) where the same model produces different outputs on the LLVM (CPU) and CUDA (GPU) backends.<br>
The problem lies with inverse trigonometric functions and `MaxPool` ops in this issue.

# Changes
Two root causes were identified and addressed:
1. **Missing `asin` domain check**
The Taylor-series approximation of `asin` did not validate its input domain, allowing values outside [-1, 1] to silently produce numeric results.
    - **Fix:**  Updated `tir.asin` to return quiet-NaN if the input is outside of [-1, 1].
2. **LLVM backend and CUDA backend follow different NaN policies for float `max`/`min`**
    - **LLVM:** propagates NaNs (if any operand is NaN, the result is NaN)
    - **CUDA:** treats NaN as missing data as described in the [CUDA API manual](https://docs.nvidia.com/cuda/cuda-math-api/cuda_math_api/group__CUDA__MATH__DOUBLE.html):
        - If one operand is NaN and the other is a number, choose the numeric value.
        - If both are NaN, return NaN.
    - **Fix:** Updated `max`/`min` ops on LLVM to match CUDA’s behavior.

# Notes
- I also tried to align the CUDA behavior with LLVM but CUDA does not appear to support the propagating-NaN behavior.
    - Regarding this, I'd appreciate maintainers' feedback on whether to keep these different NaN policies.
- The lint check seems to have a bug. It flags the following as correct:
```cpp
PrimExpr out_range = Or(x<lower, x> upper);
```
However, the correct forms would be either:
```cpp
PrimExpr out_range = Or(x < lower, x > upper);
```
or
```cpp
PrimExpr out_range = Or(x<lower, x>upper);
```

### DIFF
--- a/src/target/llvm/intrin_rule_llvm.cc
+++ b/src/target/llvm/intrin_rule_llvm.cc
@@ -30,6 +30,8 @@
 #include <tvm/tir/op.h>
 #include <tvm/tir/op_attr_types.h>
 
+#include <limits>
+
 #include "../intrin_rule.h"
 
 namespace tvm {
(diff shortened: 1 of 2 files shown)


vacu9708:
@tvm-bot rerun

vacu9708:
@tvm-bot rerun

vacu9708:
@tvm-bot rerun

vacu9708:
@tvm-bot rerun

vacu9708:
@tvm-bot rerun

vacu9708:
@tvm-bot rerun

vacu9708:
@tvm-bot rerun

tqchen:
For this particular case, it should be the issue of the input itself? I think it is fine for different platforms to have different Nan policies as long as they are efficient, given efficiency outweights other parts

vacu9708:
@tqchen Thanks for your feedback.
I removed the changes for the different NaN policies on both backends.
I maintained the changes for the domain limit check.

vacu9708:
@tvm-bot rerun

vacu9708:
@tvm-bot rerun

vacu9708:
@tvm-bot rerun

PR #17992, opened 2025-05-19
Add registion for the operator erf in llvm 

Cookiee235:
The PR is designed to fix https://github.com/apache/tvm/issues/17975 by supporting the operators of erf in the LLVM backend.
It was implemented by using the approximation for erf(x)

as shown in https://en.wikipedia.org/wiki/Error_function

### DIFF
--- a/src/target/llvm/intrin_rule_llvm.cc
+++ b/src/target/llvm/intrin_rule_llvm.cc
@@ -232,6 +232,18 @@ TVM_REGISTER_OP("tir.atanh")
       return (log(one + x) - log(one - x)) * make_const(x.dtype(), 0.5);
     });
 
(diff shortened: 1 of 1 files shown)


Cookiee235:
@tvm-bot rerun

jikechao (approved):
LGTM,  It would be better to implement a higher-precision version of erf() in a future update.

cbalint13:
@Cookiee235 , @jikechao 

This proposal here was stalled and we got  a ERF proposal merged by #18104 (Abramowitz & Stegun variant).
Not an expert on this (as myself), but improvements are always welcome if there are better implementations.

Since now this is a duplicate, it could be closed if there are no objections.

Issue #17991, opened 2025-05-19
[Bug] [FRONTEND][ONNX] The onnx frontend incorrectly recognized two arguments for BitwiseNot operator that only has one arguement


coffezhou:
### Expected behavior

The onnx frontend should import the model correctly.

### Actual behavior

When importing the onnx model, tvm crashes as follows:
```c
Error converting operator BitwiseNot, with inputs: [R.astype(lv, dtype="int32")]
Traceback (most recent call last):
  File "/home/carla/Documents/test/test.py", line 36, in <module>
    main()
  File "/home/carla/Documents/test/test.py", line 30, in main
    tvm_model = from_onnx(onnx_model, keep_params_in_input=True)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/carla/Documents/tvm/python/tvm/relax/frontend/onnx/onnx_frontend.py", line 3696, in from_onnx
    return g.from_onnx(graph, opset)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/carla/Documents/tvm/python/tvm/relax/frontend/onnx/onnx_frontend.py", line 3327, in from_onnx
    self._construct_nodes(graph)
  File "/home/carla/Documents/tvm/python/tvm/relax/frontend/onnx/onnx_frontend.py", line 3507, in _construct_nodes
    raise err
  File "/home/carla/Documents/tvm/python/tvm/relax/frontend/onnx/onnx_frontend.py", line 3502, in _construct_nodes
    op = self._convert_operator(op_name, inputs, attr, self.opset)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/carla/Documents/tvm/python/tvm/relax/frontend/onnx/onnx_frontend.py", line 3602, in _convert_operator
    sym = op_function(self.bb, inputs, attrs, [self._nodes, self._params])
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/carla/Documents/tvm/python/tvm/relax/frontend/onnx/onnx_frontend.py", line 575, in _impl_v18
    return cls.base_impl(bb, inputs, attr, params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #18001, opened 2025-05-20
[Relax][Frontend][ONNX] Fix: bitwise_not misclassified as binary (is …

eaten-cake:
Addresses issue https://github.com/apache/tvm/issues/17991.

This PR corrects the classification of the `bitwise_not` operation from binary to unary by modifying the `BitwiseNot` implementation class.
`BitwiseNot` previously inherited from this specific class, which was a subclass of `BinaryBase`.
https://github.com/apache/tvm/blob/ffa82b30eccab8099bd4819ddc23af0f28e2d7e7/python/tvm/relax/frontend/onnx/onnx_frontend.py#L518-L531
However, BinaryBase needs two operands in this case.
https://github.com/apache/tvm/blob/ffa82b30eccab8099bd4819ddc23af0f28e2d7e7/python/tvm/relax/frontend/onnx/onnx_frontend.py#L348

### DIFF
--- a/python/tvm/relax/frontend/onnx/onnx_frontend.py
+++ b/python/tvm/relax/frontend/onnx/onnx_frontend.py
@@ -564,15 +564,14 @@ class BitwiseXor(BitwiseBase):
         return cls.base_impl(bb, inputs, attr, params)
 
 
-class BitwiseNot(BitwiseBase):
+class BitwiseNot(OnnxOpConverter):
     """Converts an onnx BitwiseNot node into an equivalent Relax expression."""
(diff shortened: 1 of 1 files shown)


Hzfengsy:
Thanks for the fix, please add a regression test case

eaten-cake:
> Thanks for the fix, please add a regression test case

The regression tests have been added. Since the previous CI did not support opset18, I have removed the skip statements. The current tests can now cover the issue.

Hzfengsy (approved):

Issue #17947, opened 2025-05-11
[Bug] Meta schedule produce inconsistant inference results


Cookiee235:
When I optimized the model using MetaSchedule, the modified model produced inconsistent inference results compared to the original model for the same inputs. This suggests a potential bug in MetaSchedule's implementation. In my view, MetaSchedule should always ensure semantically equivalent transformations of the model.

### Actual behavior

```
Traceback (most recent call last):
  File "/data/qshenaf/remote_pc/TirFuzz/bugs/05-11_03-53/topi.expand_like_1.py", line 31, in <module>
    np.testing.assert_allclose(
  File "/data/qshenaf/miniconda3/envs/tvm/lib/python3.12/site-packages/numpy/testing/_private/utils.py", line 1715, in assert_allclose
    assert_array_compare(compare, actual, desired, err_msg=str(err_msg),
  File "/data/qshenaf/miniconda3/envs/tvm/lib/python3.12/site-packages/numpy/testing/_private/utils.py", line 921, in assert_array_compare
    raise AssertionError(msg)
AssertionError:
Not equal to tolerance rtol=1e-05, atol=1e-05
An Inconsistent ouputs
Mismatched elements: 9 / 60 (15%)
Max absolute difference among violations: 502
Max relative difference among violations: inf
 ACTUAL: array([[[  0,   0,   0],
        [  0,   0,   0],
        [  0,   0,   0],...
 DESIRED: array([[[0, 0, 0],
        [0, 0, 0],
        [0, 0, 0],...

```

### Environment

tvm-0.21-dev0

### Steps to reproduce

```
import tvm
from tvm import te, topi, tir
from tvm import meta_schedule as ms
import numpy as np


def compile_mod(mod, np_input_list, output_shape, output_type, opt_level=3):
    with tvm.transform.PassContext(opt_level):
        ref_mod = tvm.build(mod, target='llvm')
    mod_output = tvm.nd.empty(output_shape, dtype=output_type, device=tvm.cpu(0))

    tvm_inputs = [tvm.nd.array(x) for x in np_input_list]
    ref_mod(*tvm_inputs, mod_output)
    return mod_output


a = te.placeholder((5, 1), dtype='int16', name='a')
shape_like = te.placeholder((5, 4, 3), dtype='float64', name='shape_like')
op_output = topi.expand_like(a, shape_like=shape_like, axis=[2])
[...]

PR #18006, opened 2025-05-22
[TOPI] Fix index handling in expand_like operator for axis expansion

jikechao:
Fix https://github.com/apache/tvm/issues/17947

This PR fixes an index calculation bug in `topi.expand_like` where incorrect indexing could lead to out-of-bounds access when expanding along specified axes.

If `a.shape[src_dim] == 1`, we use `indice=0` because broadcasting rules require replicating the single value across the expanded dimension, and indexing with 0 ensures correct behavior without bounds violations.

### DIFF
--- a/python/tvm/topi/transform.py
+++ b/python/tvm/topi/transform.py
@@ -96,9 +96,7 @@ def expand_like(a, shape_like, axis):
         axis_index = 0
         for i in range(0, len(idxs)):
             if i not in real_axis:
-                dim = tvm.tir.if_then_else(
-                    a.shape[len(indices)] !=1, idxs[i], 0
-                )
+                dim = tvm.tir.if_then_else(a.shape[len(indices)] != 1, idxs[i], 0)
                 indices.append(dim)
                 axis_index += 1
         return a(*indices)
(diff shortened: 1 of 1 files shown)


jikechao:
@tvm-bot rerun!

jikechao:
@tvm-bot rerun

jikechao:
cc @Hzfengsy @MasterJH5574 @mshr-h @tqchen

mshr-h (approved):

sunzj:
@jikechao  is that issue happens when apply the schedule or generate the schedule? i mean, if i already tuned the schedule, do i need re-tune the schedule after apply the patch, or just re-generate the model library?

jikechao:
@sunzj, you need not re-tune the schedule after applying this patch. Just regenerating the model library is enough.

jikechao:
@sunzj BTW, this patch specifically addresses a special case: the topi.expand_like operator when a.shape[src_dim] == 1. If your model doesn't encounter this edge case, you can safely ignore this patch, as it won't impact your model's behavior.

PR #18019, opened 2025-05-27
Add op support for slice_scatter

kavin-sai-krishna:
This PR adds support for the slice_scatter op in exported program and fx graph translator torch frontend.


tlopex (approved):
LGTM! Thanks!

tqchen:
Some checklist for adding a new op. 

- C0: Can the operator be decomposed into smaller ops (if so, then it may not be necessary to introduce the high-level op). It is helpful to reuse existing ops when possible so we don't have to introduce C1/C2, otherwise we need C1/C2
- C1: Introduce legalization rules so the op at least compiles on CPU
- C2: Make sure the op compiles on CUDA

tqchen:
it would be good to followup to check the above check list

PR #18025, opened 2025-05-30
[ARITH] Fix canonical simplify for LE with incorrect range assumptions

Hzfengsy:
Fix a bug in canonical simplification of less-than expressions where the algorithm incorrectly assumed variables could have negative values when simplifying expressions of the form `ax + b < c`.

The previous implementation checked if `-d < xn < d` before simplifying, but this was incorrect when variables are constrained to non-negative ranges. For example, with constraints `0 < x, y < 2` and expression `2x + y < 8`, the algorithm would incorrectly check if `-2 < y < 2` and then simplify to `x < 4`. However, when x=4 and y=-1, we get 2*4 + (-1) = 7 < 8, which satisfies the original constraint but violates the intended variable bounds.

The fix changes the range check to `0 <= xn < d`, ensuring that simplification only occurs when variables are properly bounded from below at zero.


tqchen (approved):

Hzfengsy:
@tvm-bot rerun

PR #18027, opened 2025-05-31
[Dtype] Low-precision Blackwell Datatype Support

Kathryn-cat:
This PR focuses on supporting FP4/FP8 data types introduced in Blackwell architectures (sm_100).

TVM nd array stores subbyte data types in compact format, thus two FP4 would be stored in 1 byte. The size calculator for array allocator is modified accordingly.

---

**Subtype arithmetic**

The type `__nv_fp4_e2m1` from `<cuda_fp4.h>` is a tag type and does not support pointer arithmetic. Accordingly, the compiler does not support index operations on an array declared with `__nv_fp4_e2m1` directly. If any index operations like `arr[0] + arr[1]` is desired, user should declare the array as vector type like `__nv_fp4x2_e2m1`.

For example, suppose user creates a TVM array of type `__nv_fp4_e2m1` with values

[-1 2 0.5 -6 -6 -2 2 3 4 1 -3 4 -2 2...]

```
extern "C" __global__ void __launch_bounds__(32) add_kernel(__nv_fp4_e2m1* __restrict__ A, __nv_fp4_e2m1* __restrict__ C) {
  C[((int)threadIdx.x)] = (__nv_fp4_e2m1)(((half)A[((int)threadIdx.x)]) + ((half)B[((int)threadIdx.x)]));
}
```

Printing out values of A[0], A[1], ... will show

```
A[0]: 2.000000
A[1]: -6.000000
A[2]: -2.000000
A[3]: 3.000000
```

This is because `__nv_fp4_e2m1` is only a tag type. When it advances pointer, it advance by 1-byte at a time, yielding the upper 4 bits in the packed memory buffer. As a result, we should avoid directly doing indexing on `__nv_fp4_e2m1` for arithmetic operations.

If user passes in `__nv_fp4_e2m1` nd array and perform indexing, we can convert it to `__nv_fp4x2_e2m1` and recalculate the indexing if possible, but this requires more careful handling in the lowering process.

Thus, the original corresponding test case in `test_target_codegen_cuda_fp4.py` is removed.


include/tvm/runtime/data_type.h
```diff
@@ -261,20 +294,80 @@ class DataType {
...
+   * \return The constructed data type.
+   */
+  static DataType NVFloat8E4M3FN(int lanes = 1) { return DataType(kFloat8_e4m3fn, 8, lanes); }
+
+  /*!
+   * \brief Construct NV float8 e4m3fnuz datatype.
+   * \param lanes The number of lanes
+   * \return The constructed data type.
+   */
+  static DataType NVFloat8E4M3FNUZ(int lanes = 1) { return DataType(kFloat8_e4m3fnuz, 8, lanes); }
```
MasterJH5574: CUDA only has `__nv_fp8_e4m3`, `__nv_fp8_e5m2` and `__nv_fp8_e8m0`, and their `__nv_fp8_e4m3` is `kFloat8_e4m3fn` in DLPack.  There is no `NVFloat8E4M3B11FNUZ`, `NVFloat8E4M3FNUZ`, `NVFloat8E3M4`.

DerrickYLJ: In this case, should we directly delete these datatypes since `__nv_fp8_e4m3`, `__nv_fp8_e5m2`, `__nv_fp8_e8m0`, and `__nv_fp8_e4m3` already exist?

src/runtime/device_api.cc
```diff
@@ -111,7 +111,7 @@ size_t DeviceAPI::GetDataSize(const DLTensor& arr, Optional<String> mem_scope) {
     for (int i = 0; i < arr.ndim; ++i) {
       size *= static_cast<size_t>(arr.shape[i]);
     }
-    size *= (arr.dtype.bits * arr.dtype.lanes + 7) / 8;
+    size = (size * arr.dtype.bits * arr.dtype.lanes + 7) / 8;
```
MasterJH5574: Is this calculation correct for sub-byte dtypes such as fp4 and fp6?

DerrickYLJ: Yes, this calculation should be correct for fp4 and fp6. It first calculates **total bits** and **ceil the bytes**, making sure it rounds up to the next full byte. 

For example, 3 elements of fp4 (4 bits each, 1 lane) would be `(3 * 4 * 1 + 7) / 8 = (12 + 7) / 8 = 19 / 8 = 2` bytes. This is correct, as 12 bits require 2 bytes of storage, 4 bits for padding. Same calculation for fp6. 

The previous method would have calculated bytes per element (rounded up) and then multiplied, which isn't optimal for packed sub-byte types (e.g., it would allocate 1 byte per fp4 element, instead of packing two fp4 elements into a single byte).

Kathryn-cat: Yes, I think so too. The calculation rounds up to the whole byte for the last byte.

Kathryn-cat: Yes, I agree. The calculation rounds up to the whole byte for the last byte.

tqchen: consider redirect to https://github.com/apache/tvm/blob/main/ffi/include/tvm/ffi/container/ndarray.h#L96

include/tvm/runtime/data_type.h
tqchen: Let us now remove the NV prefix for all these dtypes since they are DLPack convention now, previously we don't have these values in DLPack

tqchen:
cc @Hzfengsy

Hzfengsy:
Thanks for reminding me, I will take a close look tomorrow. also cc @LeiWang1999

src/runtime/device_api.cc
Kathryn-cat: Resolved.

include/tvm/runtime/data_type.h
Kathryn-cat: Resolved.

Kathryn-cat:
@tvm-bot rerun

tqchen:
@tvm-bot rerun

Kathryn-cat:
@tvm-bot rerun

MasterJH5574 (approved):
LGTM.

spectrometerHBH (approved):

src/tir/transforms/dtype_conversion.h
```diff
@@ -121,12 +122,43 @@ class FloatConfig {
...
+          // E4M3 variant with b11 encoding, not consistent with IEEE-754
+          return FloatConfig(4, 3, 7, InftyStyle::kNone, NaNStyle::kAllOnes);
         case DataType::kFloat8_e4m3fn:
           // E4M3 format, not consistent with IEEE-754
           return FloatConfig(4, 3, 7, InftyStyle::kNone, NaNStyle::kAllOnes);
-        default:
-          // E5M2 format, consistent with IEEE-754
+        case DataType::kFloat8_e4m3fnuz:
+          // UE4M3 format, not consistent with IEEE-754
+          return FloatConfig(4, 3, 7, InftyStyle::kNone, NaNStyle::kAllOnes);
```
FredJia-intellif, line 139: @Kathryn-cat , hi, why are the FloatConfig of e4m3, e4m3b11fnuz, e4m3fn and e4m3fnuz exactly the same?

PR #18033, opened 2025-06-02
[CUTLASS] Add GeMM kernels for Blackwell GPUs

MasterJH5574:
This PR introduces CUTLASS gemm kernels, groupwise-scaled gemm kernels and group gemm kernels for Blackwell GPUs.

Files are reorganized a bit so that the exposed global functions are now architecture agnostic.  Prior to this PR, our global function names for CUTLASS kernels usually end with `"_sm90"`, which brings extra complexity when the frontend compiler decides to dispatch kernels when there are multiple supported architectures, such as Hopper and Blackwell.

Therefore, this PR renames those global function so that the function names are arch agnostic. During the build time, only the kernels that the specific architecture supports will be built.


MasterJH5574:
@tvm-bot rerun

MasterJH5574:
@tvm-bot rerun

MasterJH5574:
@tvm-bot rerun

tqchen (approved):

PR #18034, opened 2025-06-03
[Metal] Fix `GetFunction` of metal runtime

MasterJH5574:
This PR fixes a bug in `MetalModuleNode::GetFunction`.

The lambda passed to autoreleasepool is supposed to return nothing. Prior to this PR, it returns a function at an early exit branch, which may cause the error of `EXC_BREAKPOINT` in Metal.


tqchen (approved):

tqchen:
@tvm-bot rerun

Issue #17486, opened 2024-10-23
[Bug] 'tvm.relax.op.nn' has no attribute 'attention_bias'


Cookiee235:
### Actual behavior

```
@I.ir_module
class Module:
    @R.function
    def main(q: R.Tensor((4, 16, 32, 8), dtype="float32"), k: R.Tensor((4, 8, 32, 8), dtype="float32"), v: R.Tensor((4, 8, 32, 16), dtype="float32"), bias: R.Tensor((4, 32, 16, 8), dtype="float32")) -> R.Tensor((4, 16, 32, 16), dtype="float32"):
        gv: R.Tensor((4, 16, 32, 16), dtype="float32") = R.nn.attention_bias(q, k, v, bias, scale=T.float32(0.10000000000000001), causal_mask="TopLeft", window_size=None)
        return gv

error: module 'tvm.relax.op.nn' has no attribute 'attention_bias'
 --> <str>:9:58
   |  
 9 |          gv: R.Tensor((4, 16, 32, 16), dtype="float32") = R.nn.attention_bias(q, k, v, bias, scale=T.float32(0.10000000000000001), causal_mask="TopLeft", window_size=None)
   |                                                           ^^^^^^^^^^^^^^^^^^^ 
```

### Steps to reproduce
```
irs= """# from tvm.script import ir as I
# from tvm.script import tir as T
# from tvm.script import relax as R

@I.ir_module
class Module:
    @R.function
    def main(q: R.Tensor((4, 16, 32, 8), dtype="float32"), k: R.Tensor((4, 8, 32, 8), dtype="float32"), v: R.Tensor((4, 8, 32, 16), dtype="float32"), bias: R.Tensor((4, 32, 16, 8), dtype="float32")) -> R.Tensor((4, 16, 32, 16), dtype="float32"):
        gv: R.Tensor((4, 16, 32, 16), dtype="float32") = R.nn.attention(q, k, v, bias, scale=T.float32(0.10000000000000001), causal_mask="TopLeft", window_size=None)
        return gv

import tvm
mod = tvm.script.from_source(irs)
mod.show()
mod_new = tvm.script.from_source(mod.script())   # crash!
"""
```

cc @Lunderberg @junrushao @Hzfengsy @tqchen

parsifal-47:
@Cookiee235 yes, will be happy to

cchung100m:
Hi @tlopex 

It seems that the related PR is merged, we might closed this as completed. 😄

PR #18038, opened 2025-06-04
Resolving inconsistency between attention/attention_bias

xinxilwl:
Given suggestions from Dr. Chen in #17883 , some commits are withdrawn. And #17486  is solved.


tqchen:
Sorry i take a closer look, seems a better fix would be to refactor and remove attention_bias op, and ensure that the attention operator would accept three or four arguments, so it would either take three or four argument here.

cc @cyx-6

tqchen:
actually seems there is a need to introduce attention_bias variant here, so let us first go with this version

cyx-6 (approved):
Thanks for fixing this! This fix looks good to me.

PR #18041, opened 2025-06-05
[DTYPE] Fix dtype functions after dtype refactor

tqchen:
This PR fixes dtype function compilation in NCCL after dtype refactor.


Kathryn-cat:
hi @tqchen , previously `NVFloat8E4M3` refers to `Float8E4M3FN`, so should be changed to `Float8E4M3FN` instead of `Float8E4M3`, since __nv_fp8_e4m3 actually refers to the finite version. Hope this helps!

MasterJH5574 (approved):

PR #18057, opened 2025-06-13
[TIR] Phase out ProducerStore, ProducerRealize and Prefetch

Hzfengsy:
As these three nodes are no longer used in the TVM code base, we phased them out.

In the future, we may remove/refactor the following nodes:
- ProducerLoad
- BufferRealize


Hzfengsy:
cc @tqchen

tqchen (approved):

tqchen:
needs to fix lint

Issue #18004, opened 2025-05-21
[Bug] TVM FFI encountered a Segfault when calling the invoke_stateful function


coffezhou:
### Expected behavior

TVM should run the model correctly.

### Actual behavior

When compiling and running the model, TVM crashes:
```c
!!!!!!! TVM FFI encountered a Segfault !!!!!!!
  File "<unknown>", in __pyx_pw_3tvm_3ffi_4core_8Function_1__call__(_object*, _object* const*, long, _object*)
  File "<unknown>", in tvm::ffi::FunctionObj::SafeCall(void*, TVMFFIAny const*, int, TVMFFIAny*)
  File "<unknown>", in tvm::runtime::relax_vm::VirtualMachineImpl::GetFunction(tvm::ffi::String const&, tvm::ffi::ObjectPtr<tvm::ffi::Object> const&)::{lambda(tvm::ffi::PackedArgs, tvm::ffi::Any*)#4}::operator()(tvm::ffi::PackedArgs, tvm::ffi::Any*) const
  File "<unknown>", in tvm::runtime::relax_vm::VirtualMachineImpl::_InvokeClosureStateful(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >)
  File "<unknown>", in tvm::runtime::relax_vm::VirtualMachineImpl::InvokeClosureInternal(tvm::ffi::ObjectRef const&, std::vector<tvm::ffi::Any, std::allocator<tvm::ffi::Any> > const&)
  File "<unknown>", in tvm::runtime::relax_vm::VirtualMachineImpl::GetClosureInternal(tvm::ffi::String const&, bool)::{lambda(tvm::ffi::PackedArgs, tvm::ffi::Any*)#1}::operator()(tvm::ffi::PackedArgs, tvm::ffi::Any*) const [clone .isra.0]
  File "<unknown>", in tvm::runtime::relax_vm::VirtualMachineImpl::InvokeBytecode(long, std::vector<tvm::ffi::Any, std::allocator<tvm::ffi::Any> > const&)
  File "<unknown>", in tvm::runtime::relax_vm::VirtualMachineImpl::RunLoop()
  File "<unknown>", in tvm::runtime::relax_vm::VirtualMachineImpl::RunInstrCall(tvm::runtime::relax_vm::VMFrame*, tvm::runtime::relax_vm::Instruction)
[...]

PR #18061, opened 2025-06-14
[Relax][ONNX][Transform] Add mode choice, new mode, and warning for take()

vacu9708:
# Summary
I was trying to resolve https://github.com/apache/tvm/issues/18004, where an ONNX model causes a segmentation fault in TVM but not in onnxruntime.
**Why the seg fault occurs**
This occurs because `take()`(used in `Compress`)defaults to `fast` mode, which deliberately segfaults on out-of-bounds indices. (I guess for the sake of fast speed)
**Solution**
onnxruntime's `Compress` ignores out-of-bounds indices, so I initially considered matching that behavior in TVM. However, the ONNX specification doesn’t actually mandate how to handle out-of-bounds indices
**Question for maintainers**
Would you prefer a follow-up change to `relax.nonzero()`(used by `Compress`), adding an optional "max_index" param to slice out invalid indices to mirror onnxruntime? although I guess this is unnecessary.

Instead of resolving this seg-fault, for now, this PR focuses just on enhancing `take()`.

# Updates
- Add a `mode` parameter to`take()` in Relax
  - > There are already several modes implemented in the transform layer, but Relax’s `take()` defaults to `fast` with no choice. I've exposed all modes so future use cases can pick as needed down the road.
- Add `NaN` mode to `take()`
  - > I was initially adding a `raise` mode like `numpy.take()`, but since there are no precedents raising runtime errors in the transform layer, I added an `NaN` mode that returns `NaN` for any out-of-bounds index.
- Add unit tests covering all `take()` modes
  - > The tests for take() modes had been lost during a refactor.
- Add a warning log for `fast` mode along an axis
  - > take() along a flattened input shows a warning that lets users know about the potential seg fault, while `take()` along an axis does not. I think both `take()` need to warn about it.
- Unify default modes in lower layers to `fast` for consistency with Relax
  - > The default mode for `take()` in Relax is currently `fast`, I’ve made the lower layers use `fast` mode as well for consistency.


Hzfengsy:
cc @tlopex

Hzfengsy (approved):

Hzfengsy:
Sorry for the late response. Please rebase the code, and we can get it in

vacu9708:
@Hzfengsy I've rebased the code according to the newly introduced `reflection` mechanism.

PR #18079, opened 2025-06-18
[Script] Add support for merging block annotations

Hzfengsy:
This commit introduces functionality to merge block annotations in TVM script. The implementation includes:

- MergeAnnotations function that recursively merges annotation dictionaries
- Support for nested dictionary merging with new values overriding old ones
- Error handling for conflicting annotation values
- BlockAttrs function that uses the merging logic to combine multiple T.block_attr() calls within the same block

The feature allows users to specify block attributes incrementally using multiple T.block_attr() calls, which will be automatically merged together.


tqchen (approved):

src/script/ir_builder/tir/ir.cc
```diff
@@ -219,12 +219,47 @@ void Writes(Array<ObjectRef> buffer_slices) {
...
+    auto new_dict = value.try_cast<Map<String, Any>>();
+    if (old_dict && new_dict) {
+      // Recursively merge the two dicts
+      auto merged_dict = MergeAnnotations(*old_dict, *new_dict);
+      result.Set(key, merged_dict);
+      continue;
+    }
+    // Case 2.2: the values are not both dicts, check if the keys are the same
+    if (!ffi::AnyEqual()(old_value.value(), value)) {
+      LOG(FATAL) << "ValueError: Try to merge two annotations with different values for key `"
```
tqchen, line 246: might be useful to cross check error reporting here, to make sure when error happens, we can locate the span and have right `^^^^^` to point at the code location. @Hzfengsy can you confirm?

Hzfengsy: Confirmed it works well. Try with testcase:
```python

@T.prim_func
def func3():
    with T.block():
        T.block_attr({"key1": "block1"})
        T.block_attr({"key1": "block2"})
        T.evaluate(0)
```
get output:
```
error: Try to merge two annotations with different values for key `key1`, previous one is "block1", new one is "block2"
 --> /path/to/tvm/t.py:7:9
   |  
 7 |          T.block_attr({"key1": "block2"})
   |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
note: run with `TVM_BACKTRACE=1` environment variable to display a backtrace.
```

Hzfengsy: Also I added a test case to test the error in tests/python/tvmscript/test_tvmscript_error_report.py

tqchen (approved):

PR #18084, opened 2025-06-19
added support for logsigmoid op

logeshwaranmcw:
Added logsigmoid operator support in the TVM Relax exported program and test script code to validate the relax module.


tlopex:
@logeshwaranmcw Thanks for your PR! Could you please have a check on the CI error and fix that?

PR #18086, opened 2025-06-20
[SCRIPT] Bump Python minimum version to 3.9 and update AST compatibility

Hzfengsy:
- Set Python minimal version to 3.9 as Python 3.8 has reached end of life
- Update AST handling to match Python 3.9+ API changes
- Fix deprecation warning: "Support for arbitrary keyword arguments is deprecated and will be removed in Python 3.15"
- Update TVMScript parser components to use modern AST interfaces
- Adjust linting and test infrastructure for Python 3.9+ compatibility


Hzfengsy:
cc @tqchen

PR #18087, opened 2025-06-20
fix exception when tvm not built with llvm support

ddwolf:
the exception is like this:
```
  File "/home/rss/duxiutao/repos/tvm/python/tvm/target/codegen.py", line 80, in llvm_lookup_intrinsic_id
    return _ffi_api.llvm_lookup_intrinsic_id(name)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: module 'tvm.target._ffi_api' has no attribute 'llvm_lookup_intrinsic_id'
```


python/tvm/tir/tensor_intrin/__init__.py
```diff
@@ -16,4 +16,10 @@
 # under the License.
 # pylint: disable=unused-import
 """Intrinsics for tensorization."""
-from . import arm_cpu, cuda, rocm, x86, hexagon
+import tvm
+from . import cuda
+
+if hasattr(tvm.target.codegen, "llvm_version_major"):
```
cbalint13: Something safer & better check would be ```tvm.runtime.enabled("llvm")```
https://github.com/apache/tvm/blob/main/python/tvm/tir/build.py#L153C29-L153C56

ddwolf: Thanks for your suggestions, Done it.

python/tvm/tir/tensor_intrin/__init__.py
```diff
@@ -16,4 +16,8 @@
 # under the License.
 # pylint: disable=unused-import
 """Intrinsics for tensorization."""
-from . import arm_cpu, cuda, rocm, x86, hexagon
+import tvm
```
cbalint13: One more small nit,
Let's relax a bit the dependency in this hotpath with: ```from tvm.runtime import enable```
Like here (tir init): https://github.com/apache/tvm/blob/main/python/tvm/tir/__init__.py#L20

cbalint13 (changes requested):

cbalint13 (approved):
Thanks @ddwolf !

PR #18091, opened 2025-06-24
[FFI] Provide Field Visit bridge so we can do gradual transition

tqchen:
This PR provides functions that adapts old VisitAttrs reflection utilities to use new reflection mechanism when available.

These adapter would allow us to gradually transition the object def from old VisitAttrs based mechanism to new mechanism.

- For all objects
  - Replace VisitAttrs with static void RegisterReflection() that registers the fields
  - Call T::ReflectionDef() in TVM_STATIC_INIT_BLOCK in cc file
- For subclass of AttrsNode<T>: subclass AttrsNodeReflAdapter<T> instead
  - Do the same steps as above and replace TVM_ATTRS
  - Provide explicit declaration of _type_key and TVM_FFI_DEFINE_FINAL_OBJECT_INFO

We will send followup PRs to do the gradual transition. Once all transition is completed, we will remove AttrsVisitor and only go through the new mechanism.

### DIFF
--- a/ffi/include/tvm/ffi/reflection/reflection.h
+++ b/ffi/include/tvm/ffi/reflection/reflection.h
@@ -404,6 +404,8 @@ inline Function GetMethod(std::string_view type_key, const char* method_name) {
  */
 template <typename Callback>
 inline void ForEachFieldInfo(const TypeInfo* type_info, Callback callback) {
+  using ResultType = decltype(callback(type_info->fields));
+  static_assert(std::is_same_v<ResultType, void>, "Callback must return void");
   // iterate through acenstors in parent to child order
   // skip the first one since it is always the root object
   for (int i = 1; i < type_info->type_depth; ++i) {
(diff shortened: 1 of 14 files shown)


tqchen:
Example
```c++

// can he in header file
struct TestAttrs : public AttrsNodeReflAdapter<TestAttrs> {
  int axis;
  String name;
  Array<PrimExpr> padding;
  TypedEnvFunc<int(int)> func;

  // replaces VisitAttrs
  static void RegisterReflection() {
    namespace refl = tvm::ffi::reflection;
    refl::ObjectDef<TestAttrs>()
        .def_ro("axis", &TestAttrs::axis, "axis field", refl::DefaultValue(10))
        .def_ro("name", &TestAttrs::name, "name")
        .def_ro("padding", &TestAttrs::padding, "padding of input",
                refl::DefaultValue(Array<PrimExpr>({0, 0})))
        .def_ro("func", &TestAttrs::func, "some random env function",
                refl::DefaultValue(TypedEnvFunc<int(int)>(nullptr)));
  }

  static constexpr const char* _type_key = "attrs.TestAttrs";
  TVM_FFI_DECLARE_FINAL_OBJECT_INFO(TestAttrs, BaseAttrsNode);
};

// in cc file
TVM_FFI_STATIC_INIT_BLOCK({ 
      // can group multiple such registration
     TestAttrs::RegisterReflection();
});

// still needed for now for the shash/equal mechanism
TVM_REGISTER_NODE_TYPE(TestAttrs);
```

MasterJH5574 (approved):

PR #18092, opened 2025-06-25
[Fix] Fix ExecBuilderDeclareFunction method name in exec_builder.py

ConvolutedDog:
Align it with the backend implementation in https://github.com/apache/tvm/blob/main/src/relax/backend/vm/exec_builder.cc#L349


ConvolutedDog:
@mshr-h Request a review.

mshr-h (approved):

PR #18093, opened 2025-06-25
[NVSHMEM] Extend CUDA backend to compile and link TIR modules with NVSHMEM

Kathryn-cat:
This PR enables the native NVSHMEM compilation support.


CMakeLists.txt
```diff
@@ -384,6 +384,8 @@ if (USE_CUDA AND USE_NVSHMEM)
   if (NOT NVSHMEM_FOUND)
     message(FATAL_ERROR "Cannot find NVSHMEM, USE_NVSHMEM=" ${USE_NVSHMEM})
   endif()
+  set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -I${NVSHMEM_INCLUDE_DIR} -L${NVSHMEM_LIB_DIR}")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -I${NVSHMEM_INCLUDE_DIR} -L${NVSHMEM_LIB_DIR}")
```
tqchen: use cmake primitives target_include_directories instead of setting CXX and CUDA_FLAGS

tqchen (changes requested):

python/tvm/runtime/disco/session.py
```diff
@@ -302,8 +306,6 @@ def load_vm_module(
         module : DModule
             The loaded VM module.
         """
-        if device is None:
-            device = Device(device_type=0, device_id=0)
```
tqchen, line 306: is this change intentional?

tqchen:
please include one testcase that tests the basic functionality such as just calling the get worker id

python/tvm/runtime/disco/session.py
Kathryn-cat: Yes, in the latest dlpack, `DLDeviceType` has enum values 1 to 17. The value`Device(device_type=0, device_id=0)` would raise an error of unrecognized device type. Since it is meant to indicate a Null value, I replace the subsequent usage with `Optional<Device>` type, see the changes of `UseDefaultDeviceIfNone`.

src/runtime/contrib/nvshmem/init.cc
```diff
@@ -106,12 +106,26 @@ void InitNVSHMEMWrapper(String args) {
   InitNVSHMEM(uid_64, num_workers, worker_id_start);
 }
 
+void NVSHMEMXCumoduleInit(void* cuModule) {
+  CUmodule mod = static_cast<CUmodule>(cuModule);
+  auto status = nvshmemx_init_status();
+  // The NVSHMEM library must have completed device initialization prior to
+  // nvshmemx_cumodule_init. If not, we skip the cumodule initialization.
```
jinhongyii, line 113: if not device initialized, we should return with error

Kathryn-cat: The design here is to enable NVSHMEM compilation and linking broadly for every kernel, including those whose NVSHMEM context is not initialized and do not use NVSHMEM in their kernels.

In such case, `nvshmemx_init_status()` is used to check whether we need to call `nvshmemx_cumodule_init` or not. If not device initialized, we just skip `nvshmemx_cumodule_init`.

src/target/source/codegen_cuda.cc
```diff
@@ -297,19 +297,10 @@ std::string CodeGenCUDA::Finish() {
   decl_stream << "#define TVM_ENABLE_L2_PREFETCH 0\n";
   decl_stream << "#endif\n";
 
-  decl_stream << "\n#ifdef _WIN32\n";
```
spectrometerHBH, line 300: Why change this?

Kathryn-cat: It's because NVSHMEM contains `#include <cstdint>`, which is in conflict with the original `#define int64_t long long` and could lead to CUDA compilation error. The `#define` semantics is quite error prone, so I remove it and just do `using`.

CMakeLists.txt
Kathryn-cat: Changed.

Kathryn-cat:
> please include one testcase that tests the basic functionality such as just calling the get worker id

Added a test case under `test_nvshmem.py`

Kathryn-cat:
@tvm-bot rerun

Kathryn-cat:
@tvm-bot rerun

tqchen:
@tvm-bot rerun

tqchen (approved):

PR #18097, opened 2025-06-27
[Runtime] CutensorMap support

spectrometerHBH:
This pull request enables support for CutensorMap in both the IR and runtime layers.

### DIFF
--- a/src/runtime/pack_args.h
+++ b/src/runtime/pack_args.h
@@ -64,14 +64,15 @@ union ArgUnion64 {
  *
  * \param f with signiture (ffi::PackedArgs args, ffi::Any* rv, void* void_args)
  * \param arg_types The arguments type information.
- * \param arg_is_tensormap mark whether the argument is a tensormap
+ * \param arg_extra_tags extra tags for the arguments
  * \tparam F the function type
  *
  * \return The wrapped packed function.
  */
 template <typename F>
-inline ffi::Function PackFuncVoidAddr(F f, const std::vector<DLDataType>& arg_types,
-                                      const std::vector<int>& arg_is_tensormap = {});
+inline ffi::Function PackFuncVoidAddr(
+    F f, const std::vector<DLDataType>& arg_types,
+    const std::vector<FunctionInfo::ArgExtraTags>& arg_extra_tags = {});
 /*!
  * \brief Create a packed function that from function only packs buffer arguments.
  *


tqchen:
cc @Hzfengsy

src/runtime/cuda/cuda_device_api.cc
```diff
@@ -357,5 +357,188 @@ TVM_DLL int GetCudaDeviceCount() {
 
 TVM_FFI_REGISTER_GLOBAL("runtime.GetCudaDeviceCount").set_body_typed(GetCudaDeviceCount);
 
+TVM_FFI_REGISTER_GLOBAL("runtime.cuTensorMapEncodeTiled")
```
tqchen, line 383: add comments about the calling convention

include/tvm/ir/type.h
```diff
@@ -312,5 +312,26 @@ class FuncType : public Type {
   TVM_DEFINE_OBJECT_REF_METHODS(FuncType, Type, FuncTypeNode);
 };
 
+class TensorMapTypeNode : public TypeNode {
+ public:
+  void VisitAttrs(AttrVisitor* v) { v->Visit("span", &span); }
```
tqchen, line 321: does not have to be now, see https://github.com/apache/tvm/pull/18098 to transition to new reflection, i can do that after erge as well

include/tvm/ir/type.h
```diff
@@ -312,5 +312,26 @@ class FuncType : public Type {
   TVM_DEFINE_OBJECT_REF_METHODS(FuncType, Type, FuncTypeNode);
 };
 
+class TensorMapTypeNode : public TypeNode {
```
tqchen, line 319: add docstring about this type

tqchen (commented):
overall looks good, please fix lint

Hzfengsy:
cc @LeiWang1999

tqchen (approved):

src/runtime/meta_data.h
```diff
@@ -58,6 +58,7 @@ struct FunctionInfo {
   std::string name;
   std::vector<DLDataType> arg_types;
   std::vector<std::string> launch_param_tags;
+  std::vector<int> arg_is_tensormap;
```
tqchen: update this to arg_extra_tags;

Make 
```c++
enum class ArgExtraTags : int {
   kNone = 0,
   kTensorMap = 1
}
```

tqchen (changes requested):

tqchen (approved):

w1049:
Should we add a version check? `CUtensorMap` causes compilation failures on CUDA versions below 12.

tqchen:
@w1049 can you elaborate on the failure cases?

w1049:
> @w1049 can you elaborate on the failure cases?

The `CUtensorMap` feature was introduced in CUDA 12. When attempting to compile with CUDA 11.8, the build fails because `CUtensorMap` is not available in `cuda.h`. The error message appears as:
```
tvm/src/runtime/cuda/cuda_device_api.cc:525: undefined reference to `cuTensorMapEncodeTiled'
```
Similar issue: https://github.com/llvm/llvm-project/issues/64529

tqchen:
i see, seems we should guard the registration of `runtime.cuTensorMapEncodeTiled` with `#if (CUDA_VERSION >= 12000)` macro.

tqchen:
fixed by https://github.com/apache/tvm/pull/18107 thanks @w1049

PR #18102, opened 2025-06-30
[RELAX] Fix rotary embedding buffer size calculation

harrywhoo:
- Change head_dim//2 to rotary_dim//2 in LongRope scaling
- Fixes buffer size when rotary_dim differs from head_dim when partial_rotary_factor < 1.0


MasterJH5574 (approved):
Thanks @harrywhoo for the fix!

PR #18107, opened 2025-07-01
fix: guard tensormap with cuda version check

spectrometerHBH:


spectrometerHBH:
@tvm-bot rerun

src/runtime/cuda/cuda_device_api.cc
```diff
@@ -357,6 +357,8 @@ TVM_DLL int GetCudaDeviceCount() {
 
 TVM_FFI_REGISTER_GLOBAL("runtime.GetCudaDeviceCount").set_body_typed(GetCudaDeviceCount);
 
+#ifdef __CUDACC__
```
tqchen: we don't need `__CUDACC__` likely, since this cc file is not compiled using nvcc, instead just a normal c compiler

spectrometerHBH:
@tvm-bot rerun

tqchen:
@tvm-bot rerun

tqchen (approved):

PR #18110, opened 2025-07-03
[release][Dont Squash] Update version to 0.21.0 and 0.22.0.dev on main branch

ysh329:
Make version modification twice as two commits: 
- First, to v0.21.0 on main branch.
- Second, to v0.22.dev0 on main branch in one PR.

**NOTE: Don't squash when merge.**

cc: @Hzfengsy @vinx13

### DIFF
--- a/conda/recipe/meta.yaml
+++ b/conda/recipe/meta.yaml
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-{% set version = '0.21.dev0' %}
+{% set version = '0.21.0' %}
 {% set pkg_name = 'tvm' %}
 {% set cuda_tag = cuda_version | replace('.', '') %} # [cuda]
 {% set pkg_name = pkg_name + '-cu' + cuda_tag %} # [cuda]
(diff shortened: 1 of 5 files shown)


Johnson9009:
@tqchen It seems recently the reflection refactor haven't finished, I think these kind of refactor relevant to the heart of the project, is it can be finished during a few of days? @ysh329 If so, I think we should merge this PR after the refactor, so this release can contains all code of the new mechanism.

ysh329:
I'd like to know what work remains for this refactor, how much longer it will take, and whether it can be completed this month. If it can, then we can wait.
cc @Johnson9009 @tqchen

tqchen:
Thanks @ysh329 given we are on a regular cadence with fixed timing pattern, I think we don't have to delay until refactor finishes, as they can catch in next release. Feel free to get it in. 

About the current state of refactor, likely we still will do some followup refactors, but major change like calling conventions are done. Most refactor PRs should make sure things are self-contained so code can be compiled at that time pt.

tqchen (approved):

Hzfengsy:
I agree @tqchen. It would be a good time to release :)

ysh329:
Alright, then let's proceed with the release according to our schedule.

PR #18113, opened 2025-07-04
[CMake] Refine C++/CUDA standard settings in CMakeLists.txt

PragmaTwice:
We can use `CMAKE_CXX_STANDARD` combined with `CMAKE_CXX_STANDARD_REQUIRED` to replace the old check logic, to make it more modern : )


Hzfengsy (approved):

Hzfengsy:
Thanks for the enhancement

PR #18114, opened 2025-07-04
[FFI] Replace `__attribute__` with C++ standard attributes

PragmaTwice:
Some changes:
- `__attribute__((unused))` is replaced by C++ standard attribute `[[maybe_unused]]` (https://en.cppreference.com/w/cpp/language/attributes/maybe_unused.html)
- `__forceinline` is replaced by `[[msvc::forceinline]]` (MSVC: https://learn.microsoft.com/en-us/cpp/cpp/attributes?view=msvc-170#msvcforceinline, NOTE that `inline` is added which is a slight difference with `__forceinline`)
- `__attribute__((always_inline))` is replaced by `[[gnu::always_inline]]` (Clang: https://clang.llvm.org/docs/AttributeReference.html#always-inline-force-inline, GCC: https://gcc.gnu.org/onlinedocs/gcc/Attribute-Syntax.html)
- `__declspec(noinline)` is replaced by `[[msvc::noinline]]` (MSVC: https://learn.microsoft.com/en-us/cpp/cpp/attributes?view=msvc-170#msvcnoinline)
- `__attribute__((noinline))` is replaced by `[[gnu::noinline]]` (Clang: https://clang.llvm.org/docs/AttributeReference.html#noinline, GCC: https://gcc.gnu.org/onlinedocs/gcc/Attribute-Syntax.html)

The specifier sequences in many declarations are reordered since:
```c++
[[attr]] static inline int a = 0; // OK
static inline int a [[attr]] = 0; // OK

static [[attr]] inline int a = 0; // NOT OK, mixed specifiers
static inline int [[attr]] a = 0; // NOT OK, attr is applied to the type
```

Reviewers can focus on `ffi/include/tvm/ffi/base_details.h` to avoid being overwhelmed by many trivial changes : )

### DIFF
--- a/ffi/include/tvm/ffi/any.h
+++ b/ffi/include/tvm/ffi/any.h
@@ -175,7 +175,7 @@ class AnyView {
    * \return Create an AnyView from TVMFFIAny
    * \param data the underlying ffi data.
    */
-  static TVM_FFI_INLINE AnyView CopyFromTVMFFIAny(TVMFFIAny data) {
+  TVM_FFI_INLINE static AnyView CopyFromTVMFFIAny(TVMFFIAny data) {
     AnyView view;
     view.data_ = data;
     return view;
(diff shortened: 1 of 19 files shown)


Hzfengsy:
cc @tqchen

PR #18116, opened 2025-07-05
[FFI] Use fold expression to simplify for_each

PragmaTwice:
Currently the `for_each` is implemented recursively, we can replace it by fold expression to make it more simple : )


Hzfengsy (approved):
LGTM

PR #18117, opened 2025-07-05
[FFI] Replace Arg2Str with a more powerful `for_each`

PragmaTwice:
Currently `for_each` accepts a callable with parameters typed `(size_t, Arg)`.

In this PR we extend it to three kinds of parameter types:
- `(size_t, Arg)`: same as original one
- `(Arg)`: `size_t` can be omitted if possible
- `(std::integral_constant<size_t, I>, Arg)`: the index is passed as a `constexpr` value so that it can be used in template arguments (e.g. for `std::get<I>`) and `if constexpr`s.

Also as a demonstration we replace the `struct Arg2Str` with a simple `for_each` call.

### DIFF
--- a/ffi/include/tvm/ffi/base_details.h
+++ b/ffi/include/tvm/ffi/base_details.h
@@ -29,6 +29,7 @@
 #include <tvm/ffi/endian.h>
 
 #include <cstddef>
+#include <type_traits>
 #include <utility>
 
 #if defined(_MSC_VER)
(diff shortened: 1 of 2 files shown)


Hzfengsy (approved):

tqchen:
Thanks for the contribution, for this particular case, let us remain as simplified as original version, mainly for the reason of simplicity and readability. We want to avoid indirections and polymorphism in template overloading, likely also simplifies compile time

tqchen:
followup on https://github.com/apache/tvm/pull/18121

Issue #17964, opened 2025-05-13
[Bug] LegalizeOps fails on valid onnx model: AssertionError: emit_te now only supports Tensor that has ShapeExpr shape


coffezhou:
### Expected behavior

TVM should compile the model correctly.

### Actual behavior

When compiling the model with the default relax optimization pipeline, TVM crashes as follows:
```c
Traceback (most recent call last):
  File "/home/carla/Documents/test/test.py", line 51, in <module>
    main()
  File "/home/carla/Documents/test/test.py", line 45, in main
    ex = relax.build(tvm_model, target="llvm", relax_pipeline=relax_pipeline)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/carla/Documents/tvm/python/tvm/relax/vm_build.py", line 253, in build
    mod = relax_pipeline(mod)
          ^^^^^^^^^^^^^^^^^^^
  File "/home/carla/Documents/tvm/python/tvm/ir/transform.py", line 238, in __call__
    return _ffi_transform_api.RunPass(self, mod)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "tvm/ffi/cython/./function.pxi", line 212, in tvm.ffi.core.Function.__call__
  File "tvm/ffi/cython/./function.pxi", line 265, in tvm.ffi.core.tvm_ffi_callback
  File "/home/carla/Documents/tvm/python/tvm/relax/backend/cpu_generic/pipeline.py", line 73, in _pipeline
    mod = seq(mod)
  File "/home/carla/Documents/tvm/python/tvm/ir/transform.py", line 238, in __call__
    return _ffi_transform_api.RunPass(self, mod)
  File "tvm/ffi/cython/./function.pxi", line 212, in tvm.ffi.core.Function.__call__
  File "tvm/ffi/cython/./function.pxi", line 265, in tvm.ffi.core.tvm_ffi_callback
  File "/home/carla/Documents/tvm/python/tvm/relax/transform/legalize_ops/nn.py", line 509, in _nn_gelu
    return bb.call_te(te_gelu, call.args[0], primfunc_name_hint="gelu")
  File "/home/carla/Documents/tvm/python/tvm/relax/block_builder.py", line 356, in call_te
[...]

PR #18119, opened 2025-07-06
[Relax] Fix a bug that occurred due to shape inference not handling static dim vs symbolic dim

vacu9708:
# Summary
This PR fixes https://github.com/apache/tvm/issues/17964 that occurs through the following process:
1. The Compress node produces a symbolic dimension "num_nonzero"
2. The relax.op.add() in the BiasGelu node adds shape=[2, num_nonzero] and shape=[3]
3. The current shape inference function does not take into account static vs symbolic dimension cases and produces "None" shape, which leads to the error.

# Changes
- Add comments on each case
- Add support for comparing static and symbolic dimensions in shape inference

# Notes
The current shape inference takes into account the following cases

| Case | Error? | Example | expected output dim |
|----------------------------------|---------|-------------------|---|
| static dim(1) | No | (2, 3) + (1, 3)   | (2, 3) |
| equal static dims | No | (2, 3) + (2, 3)   | (2, 3) |
| equal symbolic dims | No | (n, m) + (n, m)   | (n, m) |

However, it does not take into account the following cases
| Case | Error? | Example | expected output dim |
|----------------------------------|---------|-------------------|---|
| static dim vs symbolic dim | Yes | (2, 3) + (2, n)   | (2, 3) because the 2nd dim must be 3 regardless of the symbolic dim(n) |
| different symbolic dims | Yes | (2, n) + (2, m)   | (2, n) or (2, m) because output dim cannot be determined in compile time |

I was going to add support for the "different symbolic dims" case by introducing a new symbolic dim representing the output dim as follows:
```cpp
// different symbolic dims whose output dim cannot be determined in compile time
static int _bcast_counter = 0;
std::string name = "bcast_dim_" + std::to_string(_bcast_counter++);
PrimExpr new_dim = tir::Var(name, DataType::Int(64));
output_shape.push_back(new_dim);
```
However, the current test code assumes that output dimensions are always derived from input dimensions, so I didn't add support for that case at this time.


Hzfengsy:
I disagree with the premise that (2, 3) + (2, n) can be statically inferred to have the shape (2, 3). The validity of this operation depends on the runtime value of `n`.

Specifically:

- If `n=1` or `n=3`, the operation is valid due to broadcasting, and the output shape is indeed `(2, 3)`.
- If `n` is any other value, the shapes are incompatible, and the operation should raise a runtime error.

To handle this correctly, we need a mechanism for runtime shape validation. I suggest the following steps:

1. Assume the valid shape: For this specific operation, we can tentatively define the output shape as `(2, 3)` (as this pull request does), since it's the only possible shape for a valid operation.
2. Insert a runtime check: We MUST add an assertion to the model to verify at runtime that `n` is either 1 or 3.
3. (Optional) Hoist the checks: We could write an optimization pass to lift all such runtime assertions to the beginning of the model and potentially combine them for efficiency.

cc @tqchen

vacu9708:
@Hzfengsy Thanks for your feedback.

I'll review the mechanism you proposed, try to implement it, and get back to you.

vacu9708:
I inspected operator implementations in the source code and couldn't find any precedent for asserting or validating dynamic tensors and raising a runtime error.
I’m afraid I won’t be able to implement your suggestion until such a precedent exists given my currently limited experience with TVM.
I'll try to implement it in the future.

Issue #18135, opened 2025-07-11
[Bug] Error converting operator CumSum when 'reverse=1': (ptr) is false: The struct_info is not populated, check if you have normalized the expr


coffezhou:
### Expected behavior

TVM should run the model correctly.

### Actual behavior

For the following model,

<img width="156" height="361" alt="Image" src="https://github.com/user-attachments/assets/3408978d-39df-4509-ab3a-590af91dfc1b" />

it can be executed by onnxruntime, the results are as follows:
```c
ONNXRuntime:
 [array([[3.2746487 , 2.0243466 , 0.8304557 , 0.55226177],
       [0.48739833, 0.47312835, 0.3515296 , 0.19696969],
       [9.719148  , 7.0277977 , 4.5064907 , 0.13069437]], dtype=float32)]
```
However, the onnx frontend of TVM cannot import it:
```c
File "/home/carla/Documents/tvm/python/tvm/relax/frontend/onnx/onnx_frontend.py", line 3925, in from_onnx
    return g.from_onnx(graph, opset)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/carla/Documents/tvm/python/tvm/relax/frontend/onnx/onnx_frontend.py", line 3556, in from_onnx
    self._construct_nodes(graph)
  File "/home/carla/Documents/tvm/python/tvm/relax/frontend/onnx/onnx_frontend.py", line 3736, in _construct_nodes
    raise err
  File "/home/carla/Documents/tvm/python/tvm/relax/frontend/onnx/onnx_frontend.py", line 3731, in _construct_nodes
    op = self._convert_operator(op_name, inputs, attr, self.opset)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/carla/Documents/tvm/python/tvm/relax/frontend/onnx/onnx_frontend.py", line 3831, in _convert_operator
    sym = op_function(self.bb, inputs, attrs, [self._nodes, self._params])
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/carla/Documents/tvm/python/tvm/relax/frontend/onnx/onnx_frontend.py", line 1343, in _impl_v14
    data = bb.emit_te(topi.flip, data, axis=axis if axis else 0)
[...]

PR #18137, opened 2025-07-12
[Fix][ONNX] Fix CumSum conversion when loading ONNX model

ConvolutedDog:
- Fix the ONNX CumSum converter to properly handle reversed cumsum operations by:
  - Flip the input before cumsum if reverse=True
  - Flip the output back to maintain original dimension order
  - Add proper normalization of the output of relax.op.cumsum (Fix https://github.com/apache/tvm/issues/18135)
  - Add reverse=True test case to verify reversed cumsum operation


ConvolutedDog:
@mshr-h Request a review.

yongwww (approved):

PR #18148, opened 2025-07-15
[FFI][REFACTOR] Phase out TVM_FFI_REGISTER_GLOBAL in favor of GlobalDef

tqchen:
This PR migrates the remaining global def reg to use the new mechanism. It also phases out the TVM_FFI_REGISTER_GLOBAL macro in favor of the GlobalDef mechanism.


Johnson9009:
In the previous reflection design and implementation it can be used in runtime, because its runtime overhead is too big to runtime code. 

Does the new reflection can be used in runtime code or not? If so does it introduce extra obvious runtime overhead？Thanks.

tqchen:
thanks @Johnson9009 yes, it is effectively the same thing, just different syntax sugar. so it will have the same overhead as original one

MasterJH5574 (approved):
LGTM!

PR #18154, opened 2025-07-18
[TIR] Fix trivial index map [] -> [0]

wrongtest-intellif:
Just a corner case fix.

```python
index_map = IndexMap.from_func(lambda: [tvm.tir.IntImm("int32", 0)], ndim=0, index_dtype="int32")
index_map.map_shape([])
```


tqchen:
the error message is interesting, @wrongtest-intellif do you mind dig a bit?

wrongtest-intellif:
> the error message is interesting, @wrongtest-intellif do you mind dig a bit?

OK, trying to reproduce locally.

wrongtest-intellif:
The fault case `tests/python/relax/test_backend_dispatch_sort_scan.py::test_dispatch_scanop_cuda` is as follows:

In `NormalizePrimFunc` it make a `TransformBlockLayout` trial. Previously, it throws on mapping `[] -> 0`, then after fix, it throws on `state->Replace()` internally, because it tries to transform an opaque block to a block wrapped by dummy loop (induced by the trivial index mapping). But since debug mask may not set, the failure is internally which leave the block invalidated and the usage to this block rv crash.

```c++
try {
  sch->TransformBlockLayout(block, IndexMap(index_map_inputs, index_map_outputs));
} catch (tvm::runtime::Error& e) {
  // Skip layout transformation when not transformable.
}
block_loops.push_back(sch->GetLoops(block));
```

The following check is not executed. 
```c++
if (this->debug_mask != 0) {
    const StmtNode* src_stmt = _src_sref->stmt;
    bool input_correct =
        (src_stmt->IsInstance<ForNode>() && tgt_stmt->IsInstance<ForNode>()) ||
        (src_stmt->IsInstance<ForNode>() && tgt_stmt->IsInstance<BlockRealizeNode>()) ||
        (src_stmt->IsInstance<BlockNode>() && tgt_stmt->IsInstance<BlockNode>());
    if (!input_correct) {
      LOG(FATAL) << "TypeError: src_stmt has type: " << src_stmt->GetTypeKey()
                 << ". tgt_stmt has type: " << tgt_stmt->GetTypeKey() << ".\nsrc_stmt:\n"
                 << GetRef<Stmt>(src_stmt) << "\ntgt_stmt:\n"
                 << tgt_stmt;
    }
  }
```

A simple fix is check the invalid replacement in the primitive itself.

PR #18159, opened 2025-07-21
[FRONTEND] Add max_trials_per_task parameter to static_shape_tuning_pipeline

ConvolutedDog:
This PR introduces the max_trials_per_task parameter to the static_shape_tuning_pipeline, giving users explicit control over the maximum number of trials allocated to each task during MetaSchedule tuning.

Currently, the default value is 64, so when tuning on the GPU, when the max_trials_per_task value given by the user is less than the number of tasks x 64 but greater than 64, some TIR funcs are not tuned and not bound to threads.

### DIFF
--- a/python/tvm/relax/pipeline.py
+++ b/python/tvm/relax/pipeline.py
@@ -21,7 +21,7 @@ This namespace offers a pre-defined collection that can be used
 as it is or serves as a basis to do further composition.
 """
 # pylint: disable=unused-argument
-from typing import Union
+from typing import Union, Optional
 
 import tvm
 from tvm import meta_schedule as ms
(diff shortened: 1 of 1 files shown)


docs/how_to/tutorials/e2e_opt_model.py
```diff
@@ -119,6 +138,7 @@
     # Need to allocate data and params on GPU device
     gpu_data = tvm.nd.array(np.random.rand(1, 3, 224, 224).astype("float32"), dev)
     gpu_params = [tvm.nd.array(p, dev) for p in params["main"]]
-    gpu_out = vm["main"](gpu_data, *gpu_params).numpy()
+    gpu_out = vm["main"](gpu_data, *gpu_params)
 
+    gpu_out = np.array([gpu_out[0].numpy()[0][j] for j in range(1000)])
```
Hzfengsy: What's the purpose of this change?

ConvolutedDog: I have simplified the numpy conversion.

ConvolutedDog: After the new FFI was introduced, `vm["main"](...)` returns an object of `<class 'tvm.ffi.container.Array'>`, so a `[0]` needs to be added.

PR #18165, opened 2025-07-25
[FFI][REFACTOR] Enable custom s_hash/equal

tqchen:
This PR enables custom shash equal via TypeAttr,
also enhances the Var comparison by checking content so we can precheck type signatures.


tqchen:
@tvm-bot rerun

Hzfengsy (approved):

PR #18184, opened 2025-08-01
[FFI][REFACTOR] Hide StringObj/BytesObj into details

tqchen:
This PR hides StringObj/BytesObj into details and bring implementations to directly focus on the String/Bytes.

This change will prepare us for future changes such as SmallStr support. Also moves more ObjectRef into Any in RPC.


MasterJH5574 (approved):
LGTM

PR #18188, opened 2025-08-05
[FFI][REFACTOR] Migrate the Save/Load JSON to the new reflection

tqchen:
This PR migrates the Save/Load JSON to the new reflection based mechanism.

This is a breaking change that updates the the JSON format to ffi/extra/serialization to handle the serialization, see the json graph schema comment in ffi/extra/serialization.h for the format, which roughly aligns with the old style.

After this change, we no longer need node/reflection and reflection vtable. We can also phase out TVM_REGISTER_NODE and TVM_REGISTER_OBJECT to have a single place that defines the reflection.


tqchen:
Notes for migration
- For IR that are saved in json, the easiest way is to dump out tvmscript and parse back
- node/reflection.h is no longer needed
- Remove TVM_REGISTER_NODE_TYPE and TVM_OBJECT_TYPE as they are no longer needed
- For very rare cases that needs special handling of save/load from json, need to overload attributes `__data_from_json__` and `__data_to_json__`.

Hzfengsy (approved):

PR #18189, opened 2025-08-06
[FFI][REFACTOR] Phase out getattr based attribute handling

tqchen:
This PR phases out getattar based attribute handling as they are slower and introduces extra code path.

This does mean that if an Object is not explicitly registered in python side, we will no longer be able to access the field by name. Likely this is also desirable as we would like to enable faster use that updates the python end and do not rely on these behavior.


tqchen:
Note for upgrade

- If an object is not registered in python side and returned, it will results in a warning, and very likely we no longer will be able to access its field in python side
   -  Note that previous fallback access is very slow and is not desirable
- Solution is to expose the object(and the base object if they also have members) via `@register_object`

PR #18190, opened 2025-08-06
[Fix] Codegen fix for relax cutlass

jcf94:
Seems the migration from TVMValue to PackedArgs/AnyView missed the codegen part when I tried to test relax cutlass.
With this fix, now the codegened cutlass .cu can pass nvcc compile.

BTW, the relax/cutlass part has not been guarded by CI?

### DIFF
--- a/src/relax/backend/contrib/cutlass/codegen.cc
+++ b/src/relax/backend/contrib/cutlass/codegen.cc
@@ -67,6 +67,7 @@ runtime::Module Finalize(const std::string& code, const Array<String>& func_name
   default_headers << "#include <cutlass/coord.h>\n";
   default_headers << "#include <cutlass/tensor_ref.h>\n";
   default_headers << "#include <cutlass/util/host_tensor.h>\n";
+  default_headers << "#include <tvm/runtime/logging.h>\n";
 
   const auto pf = tvm::ffi::Function::GetGlobalRequired("runtime.CSourceModuleCreate");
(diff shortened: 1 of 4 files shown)


src/relax/backend/contrib/cutlass/codegen.cc
```diff
@@ -67,6 +67,7 @@ runtime::Module Finalize(const std::string& code, const Array<String>& func_name
   default_headers << "#include <cutlass/coord.h>\n";
   default_headers << "#include <cutlass/tensor_ref.h>\n";
   default_headers << "#include <cutlass/util/host_tensor.h>\n";
+  default_headers << "#include <tvm/runtime/logging.h>\n";
```
tqchen: To reduce overall dependency, assuming the codegen part did not use ICHECK or (or change to TVM_FFI_ICHECK), LOG(INFO), instead to `#include <tvm/ffi/function.h>` to minimize deps

tqchen:
Thanks @jcf94 ! some minor nit

tqchen:
@jcf94  would be great to do the followup minor change and see if we can bring it in. thank you

jcf94:
> @jcf94 would be great to do the followup minor change and see if we can bring it in. thank you

Sorry for just seeing your comment. 😂
I have rewritten the codegen code from `CHECK` to `TVM_FFI_ICHECK`.

tqchen:
Thanks @jcf94 I made a minor change and once it is green we can merge it in

PR #18200, opened 2025-08-09
[FFI][REFACTOR] Update Map ABI to enable flexible smallMap switch

tqchen:
This PR updates the Map ABI to use MSB in slots_ to indicate SmallMap. The change would open doors for future changes to small map boundary switch.


tqchen:
cc @yongwww @MasterJH5574 This PR updates the Map ABI so we need to sync changes in downstreams like mlc

MasterJH5574 (approved):

PR #18203, opened 2025-08-11
[Relax][Op] Added GQA Attention Resolved  #18196

### LINKED ISSUE #18196: [Tracking Issue] Need support for GQA Attention in Relax
This feature is critical for modern LLM compilation workflows and is currently not available in Relax. Adding native support for GQA in Relax will enable better performance and compatibility with transformer-based models exported from PyTorch, HuggingFace, or ONNX formats.

https://arxiv.org/abs/2305.13245

hamzaqureshi5:
Implemented native GQA support in Relax, enabling efficient compilation and execution of modern transformer architectures such as LLaMA.

[GQA Paper](https://arxiv.org/abs/2305.13245)

This issue is now resolved and can be closed.

### DIFF
--- a/ffi/scripts/benchmark_dlpack.py
+++ b/ffi/scripts/benchmark_dlpack.py
@@ -36,6 +36,7 @@ Summary of some takeaways:
 -
 
 """
+import os
 import torch
 import numpy as np
 from tvm import ffi as tvm_ffi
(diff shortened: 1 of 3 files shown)


hamzaqureshi5:
Please review @tqchen

tqchen:
Thanks, we are integrating GQA from flashinfer integration and not sure if we need it for standard attention interface, would be good to have a look at the current GQA interface and see if boolean flag is the right way, since we may need group size

PR #18206, opened 2025-08-12
[CODEGEN][REFACTOR] tir.call_llvm_intrin to remove nargs

tqchen:
This PR refactors the tir.call_llvm_intrin to omit the first nargs argument in the beginning. Previously the nargs was introduced when prefetch have different number of signature. The previous reason no longer stands as of now, and it is less intuitive to attach nargs for the call_llvm_intrin, where nargs directly appears in number of argument.

After the update, tir.call_llvm_intrin can directly pass in the arguments as it is.

### DIFF
--- a/include/tvm/tir/builtin.h
+++ b/include/tvm/tir/builtin.h
@@ -225,7 +225,7 @@ TVM_DLL const Op& call_spirv_pure_glsl450();
 // TODO(tvm-team) revisit the builtins below
 // some of them can simply become ops with special codegen attr.
 /*!
- * \brief Prefetch a cacheline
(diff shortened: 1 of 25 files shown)


tqchen:
##

Before
```python
T.call_llvm_intrin("type", "llvm.intrin_name", nargs, arg0, arg1, ...)
```

After
```python
T.call_llvm_intrin("type", "llvm.intrin_name", arg0, arg1, ...)
```

cbalint13 (approved):
This is really straightforward.
Never investigated why was there, personally I thought it is a IR parsing hint.

PR #18207, opened 2025-08-14
[Fix][ONNX] No precision widening for numpy binary operations

cbalint13:
This PR enforce ONNX constraint compliance for binary operations.

---

#### Changes

* Overides the behaviour of numpy precision widening, e.g. ```np.div(int64,in64)->float64```
* This **now complies** with the [onnx.Div](https://onnx.ai/onnx/operators/onnx__Div.html) output constraint: ```C: Result, has same element type as two inputs```.
* Use an more appropriate public helper from API for the numpy type conversion (now it works with onnx 0.20 too).

---

#### Failure example

   After shape sampling and division in the graph (excerpt from a custom onnx11n model) ```np.div(5,2)``` results in a ```float64``` (as per numpy behaviour) and then the input parameter of ```Slice``` fails upon its ```int64``` constraint.

  onnx-div-widening-issue

  ```
  tvm.error.InternalError: Check failed: (is_base_of_tuple_of_int64(sinfo)) is false: 
  Operator Op(relax.strided_slice) requires the end argument to be a tuple of int64 PrimValues.  
  However, in expression {...}, the end argument (R.prim_value(T.float64(2.5)),) has struct info
  R.Tuple(R.Prim(value=T.float64(2.5)))

  ```


cbalint13:
Cc @Hzfengsy

PR #18209, opened 2025-08-14
[ONNX][FRONTEND][Fix] Update Resize to accept ShapeExpr

cbalint13:
This updates ONNX Resize to work with relax.expr.ShapeExpr parameters.

---

#### Sample:
I attach a use case [nnblur.onnx.gz](https://github.com/user-attachments/files/21781486/nnblur.onnx.gz) that produces relax.expr.ShapeExpr as ```Resize``` parameter.

#### The error:
```
ValueError: Node /Resize cannot handle ShapeExpr inputs.
```

Cc: @Hzfengsy


tlopex (approved):
LGTM! Thanks!

PR #18211, opened 2025-08-15
[TARGET]add target for nvidia rtx 5060ti

thenumberouscode:
Added NVIDIA RTX 5060 Ti as a supported target device，so we can use it like the following code
```python
target = tvm.target.Target("nvidia/geforce-rtx-5060-ti")
```


cbalint13 (approved):
LGTM

cbalint13:
Thank you @thenumberouscode !

PR #18213, opened 2025-08-16
[FFI] Formalize ffi.Module

tqchen:
This PR formalizes original runtime::Module into ffi as ffi.Module and cleans the APIs around it.

The goal is to stablize the Module API as extra API that can benefit the overall ffi interactions. We also refactors the c++ code that depends on the Module.


tqchen:
Update note:

Make bytes serialization more self-contained across ffi
   - SaveBinary => SaveToBytes 
   - LoadFromBinary =>LoadFromBytes
   - Global registration keys are updated accordingly
   - `ffi.Module.load_from_bytes.<kind>`
   - `ffi.Module.load_from_bytes.<kind>`
Clarify API naming
  - Save => WriteToFile
  - type_ket =>kind (to avoid confusion with object system type key)
 
Update API convention
   - Remove sptr_to_self as we can obtain it in the Obj
   - Move most functions to ModuleObj

Hzfengsy (approved):

PR #18226, opened 2025-08-24
[FFI][REFACTOR] Establish tvm_ffi python module

tqchen:
This PR establishes tvm_ffi as a standalone python module. The ffi is structured as a minimal pip module that can be directly install by path or url.

examples/get_started provided a minimal example.
This is a major change as we are decoupling tvm_ffi as a separate package, users need to install tvm_ffi separately.

Thanks to its minimal dependency, tvm_ffi can be easily installed even just from the source by pip install ./ffi

This change would enable future improvement for library plugins to have lightweight dependencies by just working on top of the tvm_ffi, while the main compiler toolchain and runtime can be layered on top.


yongwww (approved):

tqchen:
In #17920 , we introduce the new TVM FFI system, after various evolutions and refactoring, we are getting to a state that the ffi component is ready to be isolated out. This marks a major milestone of our ffi refactoring.

Summary of the Change

- Now all ffi related components(including python binding) are isolated under the ffi folder
- The ffi part is now moves into its own package minimal  that contains the core ABI hopefully we start to become stable over time
    - We modernize the package using `scikit-build-core` and the package can be readily installed via ```pip install ./ffi```
    - An alpha sdist convenient binary can also be tried via `pip install apache-tvm-ffi` 
- TVM compiler and runtime will depend on the the new ffi module
- For extensions that only needs to expose certain functions, they can directly depend on tvm-ffi only
- Dynamic libraries(such as libtvm) that build on top of tvm-ffi will not be impacted by the python ABI or torch ABI, so wheels can be created independent from python/torch version

We will continue to evolve tvm-ffi a bit, but it should be at a reasonable stable state. The change would pave path for us to further modularize the stable part of components that can eable broad interactions, while continuously improving the compiler.

tqchen:
https://discuss.tvm.apache.org/t/discuss-bring-tvm-ffi-into-a-seperate-repo/18566

PR #18229, opened 2025-08-25
[Relax] Add Python function support and BasePyModule for PyTorch integration

tlopex:
### **Overview**

This PR implements native Python function support in TVM Relax through the `@I.pyfunc` decorator and `BasePyModule`, which enable seamless integration between TVM's compilation pipeline and Python/PyTorch runtime environments. This enhancement allows users to write Python functions directly in TVMScript that can interoperate with Relax and TIR functions that provides enhanced debugging capabilities and leveraging existing PyTorch operator libraries.

### **Key Features**
**TVMScript Parser Enhancement**
- `@I.pyfunc` decorator: Marks Python functions for integration into IRModules
- Dual storage format: Stores both raw string representation (for TVMScript printing) and captured PackedFunc (for runtime execution)
- ExternFunc representation: Each Python function is represented as an ExternFunc node with attributes storing source code and runtime wrapper

**Complete BasePyModule Implementation**
- DLPack-based tensor conversion: Seamless conversion between PyTorch tensors and TVM NDArrays
- Cross-function interoperability: Python functions can call Relax/TIR functions and vice versa
- JIT compilation: Delays compilation until module instantiation for flexible late-stage modifications
- Dynamic function registration: Supports runtime addition of Python functions

### Future Work
- TVMScript printer for IRModules with Python functions: Print IRModules in proper format with high-level operator mapping from Relax ops to PyTorch ops, handling symbolic shapes
- R.call_py_func primitive: Introduce Relax primitive to invoke corresponding PackedFunc of specified Python functions at runtime


tlopex:
cc @MasterJH5574 
Could you please have a look at CI.
Why it said [2025-08-25T05:02:07.869Z] ImportError: Error importing plugin "tvm.testing.plugin": No module named 'torch'

MasterJH5574:
> Could you please have a look at CI. Why it said [2025-08-25T05:02:07.869Z] ImportError: Error importing plugin "tvm.testing.plugin": No module named 'torch'

@tlopex Likely the docker image for wasm CI doesn't come with torch.  Can we change to lazy import torch, i.e. import torch inside where it is used?

python/tvm/relax/base_py_module.py
```diff
@@ -0,0 +1,367 @@
...
+        if isinstance(tvm_array, torch.Tensor):
+            return tvm_array
+        if not isinstance(tvm_array, NDArray):
+            return torch.tensor(tvm_array)
+        try:
+            dlpack = tvm_array.to_dlpack()
+            return torch.from_dlpack(dlpack)
+        except (tvm.TVMError, RuntimeError) as error:
+            print(f"Warning: DLPack conversion from TVM failed ({error}), using numpy fallback")
+            numpy_array = tvm_array.asnumpy()
```
MasterJH5574:
```suggestion
            numpy_array = tvm_array.numpy()
```
`asnumpy` is formally removed by the latest commits. Let's use `.numpy()`.  Likely we also need to rebase this PR onto the latest main branch.

tests/python/relax/test_dlpack_integration.py
```diff
@@ -0,0 +1,284 @@
...
+        tvm_ndarray_numpy = tvm.nd.array(numpy_array)
+        numpy_time = time.time() - start_time
+
+        # Verify both methods produce same result
+        result_dlpack = torch.from_dlpack(tvm_ndarray)
+        result_numpy = torch.from_numpy(tvm_ndarray_numpy.numpy())
+        assert torch.allclose(result_dlpack, result_numpy, atol=1e-5)
+
+        # DLPack should be faster (this is a basic check)
+        assert dlpack_time < numpy_time * 2, "DLPack should be reasonably fast"
```
MasterJH5574: Just curious: what's the observed time difference between dlpack and numpy conversions?

tlopex: DLPack is expected to be at least 2x faster than numpy conversion. And according to the test, the performance difference is much more dramatic than expected. 
On my system:
- 100K elements: DLPack is 32x faster
- 1M elements: DLPack is 310x faster  
- 10M elements: DLPack is 3500x faster

tqchen: let us avoid timing comparison here as the perf is expected but UT should not contain perf bench

tlopex: Okay, I'll remove them

python/tvm/script/parser/ir/parser.py
```diff
@@ -49,21 +52,39 @@ def _visit_class_def(self: Parser, node: doc.ClassDef) -> None:
...
+                )
+                # Store this information in the IRModule for later use
+                I.module_attrs({"base_py_module": True})
+                # Set the parser context to allow Python functions
+                self.set_class_context(node.name, True)
+            else:
+                print(
+                    f"ℹ Class '{node.name}' does not inherit from BasePyModule - "
+                    f"Python functions not allowed"
+                )
```
MasterJH5574: Are these prints intentional? It looks like they are debug prints that can be removed.

python/tvm/script/parser/core/entry.py
```diff
@@ -86,12 +87,15 @@ def parse(
         extra_vars = _default_globals()
 
     ann = {}
+    all_pyfuncs = {}
     if inspect.isfunction(program):
         ann = {program.__name__: program.__annotations__}
     elif inspect.isclass(program):
         for name, func in program.__dict__.items():
             if inspect.isfunction(func):
+                print(f"name: {name}, func: {func}, annotations: {func.__annotations__}")
```
MasterJH5574: May be removed as well.

MasterJH5574 (commented):
Overall looks very good, thank you @tlopex!  Would be great to add an example of “IRModule with pyfunc” in the PR description, and you can also reply the example to the forum discussion thread after this PR is merged.

python/tvm/script/parser/ir/parser.py
tlopex: Okay, removed

python/tvm/script/parser/core/entry.py
tlopex: Okay, removed too

MasterJH5574 (approved):

Issue #18228, opened 2025-08-25
[Bug] assert_structural_equal fails when comparing identical IRModule


guocju:
### Expected behavior

```
import tvm
from tvm.script import tir as T, ir as I

@I.ir_module
class Module:
    @T.prim_func
    def add_one(A: T.handle, B: T.handle):
        A_buf = T.match_buffer(A, (32,), "float32")
        B_buf = T.match_buffer(B, (32,), "float32")
        for i in range(32):
            B_buf[i] = A_buf[i] + T.float32(1.0)

mod = Module
# Expect: no error
tvm.ir.assert_structural_equal(mod, mod)
```
Expect: no error
### Actual behavior

ValueError: StructuralEqual check failed, caused by lhs at <root>.functions[I.GlobalVar("add_one")].params[0].dtype:
and rhs at <root>.functions[I.GlobalVar("add_one")].params[0].dtype:
```
# from tvm.script import ir as I
# from tvm.script import tir as T

@I.ir_module
class Module:
    @T.prim_func
    ^^^^^^^^^^^^
    def add_one(A: T.handle, B: T.handle):
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        A_buf = T.match_buffer(A, (32,))
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        B_buf = T.match_buffer(B, (32,))
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        for i in range(32):
        ^^^^^^^^^^^^^^^^^^^
            B_buf[i] = A_buf[i] + T.float32(1.0)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

### Environment

TVM version: a7a0168be  [FFI][REFACTOR] Establish tvm_ffi python module (#18226)
Python version: 3.10.12
OS: (Linux/Ubuntu 22.04 )

### Triage

Please refer to the list of label tags [here](https://github.com/apache/tvm/wiki/Issue-Triage-Labels) to find the relevant tags and add them below in a bullet format (example below).

* needs-triage

tqchen:
unfortunately i am not able to reproduce the error, can you try to rebuild and cross confirm

guocju:
> unfortunately i am not able to reproduce the error, can you try to rebuild and cross confirm

Yeah, I have checked and found that the problem isn't with the TVM code. It's actually the CMake configuration.

[...]

PR #18237, opened 2025-08-26
[FFI][BUGFIX] Fix type_traits on DataType after SmallStr update

tqchen:
This PR fixes the type_traits on DataType after SmallStr update. We need to explicitly zero out the FFFIAny data structure to allow fast comparison of FFIAny based on bytes values.


tqchen:
Fixes #18228 Thanks to @guocju for reporting

PR #18239, opened 2025-08-27
[Build] Migrate Python packaging to pyproject.toml with scikit-build-core

tlopex:
This pr migrates the TVM Python packaging system from the setup.py flow to the modern, PEP 517/518 compliant pyproject.toml standard, which allows us to produce a single, Python-version-agnostic wheel. 
This change streamlines the process for both developers and users. For local development, you can now set up a fully-functional editable environment with a single command: `pip install -e .`. To create the distributable package for release, simply run `pip wheel -w dist .`, which will produce a universal wheel in the dist/ folder. This ensures that end-users can reliably install TVM with a standard pip install tvm, regardless of their specific Python 3 version.


tlopex:
cc @tqchen

CMakeLists.txt
```diff
@@ -818,3 +820,123 @@ if(USE_ROCM AND USE_RCCL)
...
+  install(
+    DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/ffi/3rdparty/dlpack/include"
+    DESTINATION "tvm/3rdparty/dlpack"
+  )
+
+  # Install libbacktrace only if enabled
+  if(USE_LIBBACKTRACE)
+    install(
+      DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/ffi/3rdparty/libbacktrace"
+      DESTINATION "tvm/3rdparty"
```
tqchen: we no longer need to install libbacktrace

CMakeLists.txt
```diff
@@ -818,3 +820,123 @@ if(USE_ROCM AND USE_RCCL)
...
+    DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/cmake/utils"
+    DESTINATION "tvm/cmake/utils"
+    FILES_MATCHING
+    PATTERN "*.cmake"
+  )
+
+  # Install minimal third-party files
+  install(
+    DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/3rdparty/dlpack/include"
+    DESTINATION "tvm/3rdparty/dlpack"
```
tqchen: we no longer need to install dlpack

CMakeLists.txt
tqchen: they are shipped as part of ffi

CMakeLists.txt
```diff
@@ -818,3 +820,123 @@ if(USE_ROCM AND USE_RCCL)
...
+      DESTINATION "tvm/3rdparty/cutlass"
+      FILES_MATCHING
+      PATTERN "*.h"
+      PATTERN "*.hpp"
+    )
+  endif()
+
+  # Install other essential third-party headers
+  install(
+    DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/3rdparty/dmlc-core/include"
```
tqchen: they are shipped as part of ffi

CMakeLists.txt
```diff
@@ -818,3 +820,123 @@ if(USE_ROCM AND USE_RCCL)
...
+    DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/include/tvm/runtime"
+    DESTINATION "tvm/include/tvm/runtime"
+    FILES_MATCHING
+    PATTERN "*.h"
+  )
+
+  # Install DLPack headers
+  install(
+    DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/ffi/3rdparty/dlpack/include"
+    DESTINATION "tvm/3rdparty/dlpack"
```
tqchen: we no longer need to ship dlpack

pyproject.toml
```diff
@@ -14,6 +14,199 @@
...
+build-dir = "build"
+
+# CMake configuration - ensure proper installation paths
+cmake.args = [
+    "-DTVM_BUILD_PYTHON_MODULE=ON",
+    "-DTVM_FFI_BUILD_PYTHON_MODULE=ON",
+    "-DTVM_USE_CUTLASS=ON",
+    "-DTVM_USE_FLASH_ATTN=ON",
+    "-DTVM_USE_LLVM=ON",
+    "-DTVM_USE_CUDA=OFF",  # Set based on your needs
```
tqchen: do not set these configs by default, instead they can be picked up by config.cmake

pyproject.toml
```diff
@@ -14,6 +14,199 @@
...
+    "-DTVM_USE_OPENGL=OFF",
+    "-DTVM_USE_RPC=ON",
+    "-DTVM_USE_GRAPH_EXECUTOR=ON",
+    "-DTVM_USE_PROFILER=ON",
+    "-DTVM_USE_UTILS=ON",
+]
+
+# Ensure dependent libraries install to same location as libtvm.so
+cmake.define.CMAKE_INSTALL_PREFIX = "python/tvm"
+cmake.define.CMAKE_INSTALL_LIBDIR = "lib"
```
tqchen: we should not set cmake configs here, instead, we should set https://github.com/apache/tvm/blob/main/ffi/pyproject.toml#L77

pyproject.toml
tqchen: we only need to set -DTVM_BUILD_PYTHON_MODULE=ON,  given we are shipping ffi separately, we should not set `-DTVM_FFI_BUILD_PYTHON_MODULE=ON`

src/support/libinfo.cc
```diff
@@ -106,6 +106,10 @@
 #define TVM_INFO_USE_HEXAGON_GTEST "NOT-FOUND"
 #endif
 
+#ifndef TVM_INFO_TVM_BUILD_PYTHON_MODULE
+#define TVM_INFO_TVM_BUILD_PYTHON_MODULE "NOT-FOUND"
```
tqchen: this does not have to goto libinfo, as the flag is not as useful

tests/lint/pylintrc
```diff
@@ -124,7 +124,8 @@ disable=
     use-list-literal,
     arguments-renamed,
     super-init-not-called,
-    c-extension-no-member
+    c-extension-no-member,
+    wrong-import-order
```
tqchen: disable import order in each file instead of globally

CMakeLists.txt
```diff
@@ -818,3 +820,123 @@ if(USE_ROCM AND USE_RCCL)
...
+
+# Python package installation configuration
+# This section ensures that all necessary files are installed for the Python wheel
+if(TVM_BUILD_PYTHON_MODULE)
+  message(STATUS "Configuring Python package installation")
+
+  # Install Python source files
+  install(
+    DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/python/tvm"
+    DESTINATION "tvm"
```
tqchen: are we sure we want to install to the tvm subfolder? usually we need to install into the . folder in a python package, please cross confirm the wheel in pip install mode

Cross ref https://github.com/apache/tvm/blob/main/ffi/CMakeLists.txt for an example

tqchen (changes requested):

tqchen:
also cc @MasterJH5574 , would be great to test the wheel for specifically cuda related usecases like cutlass(maybe we need to also update mlc),

pyproject.toml
```diff
@@ -14,6 +14,199 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+
+[build-system]
+requires = ["scikit-build-core>=0.7.0"]
```
MasterJH5574: I saw ffi requires the minimal version of 0.10.0.  Do we want to align the minimal version requirement?

https://github.com/apache/tvm/blob/472b2fcd22c4147c35cbdf884d232f69bd86a23b/ffi/pyproject.toml#L51-L53

pyproject.toml
```diff
@@ -14,6 +14,199 @@
...
+build-backend = "scikit_build_core.build"
+
+[project]
+name = "tvm"
+version = "0.16.0.dev0"
```
MasterJH5574: The latest version might be 0.22.0.dev0 according to https://github.com/apache/tvm/tags

pyproject.toml
```diff
@@ -14,6 +14,199 @@
...
+]
+keywords = ["machine learning", "compiler", "deep learning", "inference"]
+classifiers = [
+    "Development Status :: 4 - Beta",
+    "Intended Audience :: Developers",
+    "Intended Audience :: Education",
+    "Intended Audience :: Science/Research",
+    "License :: OSI Approved :: Apache Software License",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.8",
```
MasterJH5574: FFI has a minimal Python version requirement of 3.9.  Likely we need to align with that https://github.com/apache/tvm/blob/472b2fcd22c4147c35cbdf884d232f69bd86a23b/ffi/pyproject.toml#L34

MasterJH5574:
> would be great to test the wheel for specifically cuda related usecases like cutlass(maybe we need to also update mlc),

Sure. Happy to test when this PR gets ready.

pyproject.toml
tlopex: I think so. I'll align all of them later.

pyproject.toml
```diff
@@ -14,6 +14,190 @@
...
 # specific language governing permissions and limitations
 # under the License.
+
+[build-system]
+requires = ["scikit-build-core>=0.10.0"]
+build-backend = "scikit_build_core.build"
+
+[project]
+name = "tvm"
+version = "0.22.0.dev0"
```
tqchen, line 25: Add a comment that version.py needs to be called to update the version before we build the wheel, update 

https://github.com/apache/tvm/blob/main/version.py to replace the version here

tqchen (approved):
one final comment, please cross check locally

tqchen:
@tlopex @MasterJH5574 please help cross check once ready

pyproject.toml
tlopex: Okay, added

tlopex:
cc @MasterJH5574 
I think this is ready for checking

MasterJH5574 (approved):

MasterJH5574:
> I think this is ready for checking

Great, I'll find a time in the next few days.   BTW there is a windows CI error that we need to fix.

tlopex:
> > I think this is ready for checking
> 
> Great, I'll find a time in the next few days. BTW there is a windows CI error that we need to fix.

How could we fix it? Just by committing again to restart CI?

tqchen:
windows should be fixed already by https://github.com/apache/tvm/pull/18245. we don't need to be gree to merge

We consider should followup and change the conda based windows ci to use the pip wheel based now that we have the scikit-build-core

PR #18244, opened 2025-08-27
[Fix] Update FlashInfer JIT header lookup

MasterJH5574:
This PR fixes the tvm/dlpack/dmlc header lookup in the FlashInfer kernel JIT compilation.

Prior to this fix, the JIT compilation assumes the environment variable `TVM_SOURCE_DIR` is always defined, which is not always true. This PR fixes the behavior and considers multiple cases, including TVM source builds and pip-installed packages.

### DIFF
--- a/python/tvm/libinfo.py
+++ b/python/tvm/libinfo.py
@@ -195,7 +195,9 @@ def find_include_path(name=None, search_path=None, optional=False):
     include_path : list(string)
         List of all found paths to header files.
     """
-    if os.environ.get("TVM_HOME", None):
+    if os.environ.get("TVM_SOURCE_DIR", None):
+        source_dir = os.environ["TVM_SOURCE_DIR"]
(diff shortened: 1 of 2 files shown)


yongwww:
Lint has failed. BTW, it might be worth considering adding a `pre-commit` setup (as other communities do). That way, running a single command could automatically fix all lint errors. @tqchen @Hzfengsy @yzh119

MasterJH5574:
> Thanks for the fix! It is helpful. Yeah, requiring both TVM_SOURCE_DIR and TVM_HOME was a bit confusing.

@yongwww I remember we renamed all `TVM_HOME` to `TVM_SOURCE_DIR` a while ago.  I put both here just in case anyone didn't rename their local environment variables.

yongwww:
I see. Thanks, @MasterJH5574 !

PR #18250, opened 2025-08-28
Revert "[FFI][CMAKE] Add missing download path for libbacktrace"

MasterJH5574:
This PR reverts apache/tvm#18246.

As mentioned in https://github.com/apache/tvm/pull/18246#issuecomment-3234991244, it seems that libbacktrace will be re-configured every time and `ffi/src/ffi/traceback.cc` will be compiled when we run `make`, even if there is no changes. This repetitive compilation increases the compilation time by a lot.


MasterJH5574:
Closed in observance of #18249.

PR #18252, opened 2025-08-29
[Build] Complete TVM wheel building migration

tlopex:
Following #18239, this PR removes the legacy setup.py and updates CI workflows to use pip wheel instead of conda build. We now use conda only for system-level dependencies.


tlopex:
cc @tqchen

PR #18253, opened 2025-08-30
[Relax] Building TVMScript printer for IRModules with Python functions

tlopex:
This PR implements TVMScript printer to format IRModules containing @I.pyfunc decorated Python functions.

Example:
```
@I.ir_module
class MyModule(BasePyModule):
    @I.pyfunc
    def python_func(self, x, y):
        x_tvm = self._convert_pytorch_to_tvm(x)
        y_tvm = self._convert_pytorch_to_tvm(y)
        result = self.call_tir(self.add_tir, [x_tvm, y_tvm], 
                             out_sinfo=R.Tensor((5,), "float32"))
        return self._convert_tvm_to_pytorch(result)
    
    @T.prim_func
    def add_tir(a: T.handle, b: T.handle, c: T.handle):
        A = T.match_buffer(a, (5,), "float32")
        B = T.match_buffer(b, (5,), "float32")
        C = T.match_buffer(c, (5,), "float32")
        for i in range(5):
            C[i] = A[i] + B[i]

# Usage:
print(MyModule.script())  # Print formatted TVMScript
MyModule.show()           # Display formatted output
```

### DIFF
--- a/tests/python/relax/test_base_py_module_printer.py
+++ b/tests/python/relax/test_base_py_module_printer.py
@@ -27,133 +27,138 @@ from tvm.script import relax as R
 @I.ir_module
 class SimplePyFuncModule(BasePyModule):
     """Test simple Python functions with basic operations."""
-    
+
     @I.pyfunc
     def add(self, x, y):
         """Simple addition function."""
         x_tvm = self._convert_pytorch_to_tvm(x)
         y_tvm = self._convert_pytorch_to_tvm(y)
-        result = self.call_tir(self.add_tir, [x_tvm, y_tvm], 
-                             out_sinfo=R.Tensor((5,), "float32"))
+        result = self.call_tir(self.add_tir, [x_tvm, y_tvm], out_sinfo=R.Tensor((5,), "float32"))
         return self._convert_tvm_to_pytorch(result)
-    
+
     @I.pyfunc
     def multiply(self, x, y):
         """Simple multiplication function."""
         x_tvm = self._convert_pytorch_to_tvm(x)
         y_tvm = self._convert_pytorch_to_tvm(y)
-        result = self.call_tir(self.multiply_tir, [x_tvm, y_tvm], 
-                             out_sinfo=R.Tensor((5,), "float32"))
+        result = self.call_tir(
+            self.multiply_tir, [x_tvm, y_tvm], out_sinfo=R.Tensor((5,), "float32")
+        )
         return self._convert_tvm_to_pytorch(result)
-    
+
     @T.prim_func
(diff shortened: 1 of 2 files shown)


tlopex:
cc @MasterJH5574

tests/python/relax/test_base_py_module_printer.py
```diff
@@ -0,0 +1,719 @@
...
+
+def test_relax_functions():
+    """Test that Relax functions are properly defined."""
+    ir_mod = SimplePyFuncModule
+    device = tvm.cpu()
+    module = BasePyModule(ir_mod, device)
+
+    # Note: Relax functions may not be available due to TVMScript compilation issues
+    # This is acceptable for testing as the focus is on Python functions
+    print("Note: Relax functions may not be available due to TVMScript compilation")
```
MasterJH5574: Just want to get more clarity here: do you mean the relax function in this testing TVMScript is dummy and can't compile?   We can add the note such as "this particular tvmscript is for testing purpose only, and cannot compile"

tests/python/relax/test_base_py_module_printer.py
```diff
@@ -0,0 +1,719 @@
...
+                Output[i] = T.sqrt(Data[i])
+            else:
+                Output[i] = 0.0
+
+
+if __name__ == "__main__":
+    # This allows the file to be run directly for debugging
+    # In normal pytest usage, these classes are automatically tested by TVMScript
+    print("All test modules defined successfully!")
+    print("TVMScript will automatically validate these modules during testing.")
```
MasterJH5574, line 427: How is the printer tested if we just run this file with `python tests/python/relax/test_base_py_module_printer.py`?  It looks to me that the test modules will only be parsed and won't be printed. So I was wondering if I have any misunderstanding.

tlopex: You are right. If we run the file directly with `python tests/python/relax/test_base_py_module_printer.py`, the printer functionality won't be tested.

tests/python/relax/test_base_py_module_printer.py
tlopex: Sure, this is only for testing. I think I should update it.

tests/python/relax/test_base_py_module_printer.py
MasterJH5574: I see. Is it possible to make `__main__` also run them? Given this is a printer test.  I assume that maybe `tvm.testing.main()` could work out of box.

tlopex: Okay, let me try it

tlopex:
cc @MasterJH5574

PR #18259, opened 2025-08-31
[FFI][ABI] Introduce weak rc support

tqchen:
This PR adds weak ref counter support to the FFI ABI. Weak rc is useful when we want to break cyclic dependencies.

- When a strong rc goes to zero, we call the destructor of the object, but not freeing the memory
- When both strong and weak rc goes to zero, we call the memory free operation

The weak rc mechanism is useful when we want to break cyclic dependencies in object, where the weak rc can keep memory alive but the destructor is called. As of now, because we deliberately avoid cyles in codebase, we do not have strong use-case for weak rc.

However, given weak rc is common practice in shared_ptr, Rust RC, and also used in torch's intrusive_ptr. It is better to make sure the ABI is future compatible to such use-cases before we freeze.

This PR implements weak rc as a u32 counter and strong rc as a u64 counter, with the following design consideration.

- Weak rc is very rarely used and u32 is sufficient.
- Keeping weak rc in u32 allows us to keep object header size to 24 bytes, saving extra 8 bytes(considering alignment)

We also need to update deleter to take flags that consider both weak and strong deletion events. The implementation tries to optimize common case where both strong and weak goes to 0 at the same time and call deleter once with both flags set.


yongwww:
@tvm-bot  rerun

mshr-h (approved):

PR #18275, opened 2025-09-06
[FFI][REFACTOR][ABI] Rename NDArray to Tensor

tqchen:
This PR Updates the NDArray => Tensor.

Both tensor and ndarray are commonly used terms.

Because the term Tensor is getting more common in the context of ML, we do the rename to stay more aligned with torch.Tensor and DLTensor.


tqchen:
## Upgrade Guide

- `tvm.nd.empty => tvm.runtime.empty`
- `tvm.nd.NDArray=>tvm.runtime.Tensor`
- `tvm.nd.from_dlpack=>tvm.runtime.from_dlpack`

Try to also look for cases where runtime::Tensor and te::Tensor gets into same namespace that may cause conflict, add prefix

PR #18284, opened 2025-09-08
[FFI][REFACTOR] Introduce UnsafeInit and enhance ObjectRef null safety

tqchen:
This PR enhances the nullptr and general type-safe of ObjectRef types. Previously ObjectRef relies on constructor from ObjectPtr<Object> for casting and initialize from nullptr.

We introduce a tag ffi::UnsafeInit, which explicitly states the intent that the initialization is unsafe and may initialize non-nullable Ref to null. Such tag should only be used in controlled scenarios.

Now the general RefType(ObjectPtr<Object>) is removed. We still keep RefType(ObjectPtr<ContainerType>) for nullable objects, but removes the default definition from non-nullable types, knowing that user can always explicitly add it to class impl (ensuring null checking).

The resulting code would be more null safe for non-nullable types.

### DIFF
--- a/ffi/include/tvm/ffi/cast.h
+++ b/ffi/include/tvm/ffi/cast.h
@@ -44,18 +44,20 @@ namespace ffi {
  */
 template <typename RefType, typename ObjectType>
 inline RefType GetRef(const ObjectType* ptr) {
-  static_assert(std::is_base_of_v<typename RefType::ContainerType, ObjectType>,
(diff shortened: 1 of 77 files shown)


tqchen:
## Upgrade Guide

- Use Optional<RefType> in places where nullptr is needed
- If RefType(ObjectPtr<ObjectType>) constructor is needed, add a notnull check

PR #18285, opened 2025-09-08
[FFI][Bugfix] Enable `load_inline` on macos

yaoyaoding:
This PR fix the bug to enable `tvm_ffi.cpp.load_inline` on macos. We need to link the `libtvm_ffi.dylib` to the custom module.


tqchen:
CI failure was due to the concurrent merge and register_func->register_global_func renaming

PR #18288, opened 2025-09-09
[Relax] Add symbolic shape support to BasePyModule for dynamic tensor operations

tlopex:
This PR adds symbolic shape support to `BasePyModule`, which enables dynamic tensor operations with runtime shape inference. This allows users to use Relax's symbolic shape functionality in Python function calls through BasePyModule, with dimensions automatically resolved at execution time based on input tensor shapes.

## Usage Example
```python
import tvm
from tvm.script import ir as I, relax as R
from tvm.relax.base_py_module import BasePyModule
import numpy as np

@I.ir_module
class VectorAddModule(BasePyModule):
    @R.function
    def add(x: R.Tensor(("n",), "float32"), 
            y: R.Tensor(("n",), "float32")) -> R.Tensor(("n",), "float32"):
        return R.add(x, y)

module = VectorAddModule(device=tvm.cpu(0), target="llvm")

a = np.array([1.0, 2.0, 3.0], dtype="float32")
b = np.array([4.0, 5.0, 6.0], dtype="float32")
result = module.add(a, b)  # Result: [5.0, 7.0, 9.0]
```

### DIFF
--- a/python/tvm/relax/base_py_module.py
+++ b/python/tvm/relax/base_py_module.py
@@ -198,7 +198,7 @@ class BasePyModule:
             )
         func = self.compiled_tir_funcs[func_name]
 
-        out = self._create_output_tensors(out_sinfo)
+        out = self._create_output_tensors(out_sinfo, args)
         tvm_args = self._convert_pytorch_to_tvm(args)
         tvm_out = self._convert_pytorch_to_tvm(out)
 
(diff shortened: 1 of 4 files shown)


tlopex:
cc @MasterJH5574

tlopex:
@tvm-bot rerun

tlopex:
@tvm-bot rerun

PR #18293, opened 2025-09-09
[FFI] Record stream and device type for Paddle context

SigureMo:
Record device context when user installed PaddlePaddle. The API used in this PR, `paddle._C._cuda_getCurrentRawStream`, was introduced in PaddlePaddle/Paddle#75153, so this only runs on the latest develop branch (the latest released version `3.2.0` does not work).

If you have any questions, feel free to reach out — I’m happy to help and will actively support any further adaptation work.

### DIFF
--- a/ffi/python/tvm_ffi/cython/function.pxi
+++ b/ffi/python/tvm_ffi/cython/function.pxi
@@ -19,14 +19,30 @@ import os
 from numbers import Real, Integral
 
 
-if os.environ.get("TVM_FFI_BUILD_DOCS", "0") == "0":
+def is_building_docs():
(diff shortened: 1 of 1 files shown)


tqchen:
Thanks @SigureMo , i am thinkning about more scalable ways to add support without doing specific switch case. One idea is to enable functions like `tensor.__tvm_ffi_env_stream__`  for stream getter in current tensor.

tqchen:
https://github.com/apache/tvm/pull/18295 should implement this

SigureMo:
Thanks @tqchen! That's indeed a more elegant solution. I will try to push this protocol forward within PaddlePaddle as soon as possible (maybe PaddlePaddle/Paddle#75193).

PR #18304, opened 2025-09-12
Clear ext_lib_dll_names for macOS platform

oraluben:
Removed external library DLL names for macOS.

found during https://github.com/tile-ai/tilelang/pull/799

cc @LeiWang1999


FrozenGene:
as [cutlass](https://github.com/NVIDIA/cutlass?tab=readme-ov-file#operating-systems) only tested on Ubuntu and Windows , I think remove macOS is reasonable.

FrozenGene:
Thanks @oraluben It is merged now

PR #18313, opened 2025-09-13
[Relax] Introduce R.call_py_func operator for calling Python functions from Relax IR

tlopex:
This PR allows calling Python functions directly from Relax IR, where integration between Relax computations and Python/PyTorch operations can be supported.

### Usage Example
```python
@I.ir_module
class MyModule(BasePyModule):
    @I.pyfunc
    def pytorch_add(self, x, y):
        return x + y
    
    @R.function
    def compute(x: R.Tensor((5,), "float32"), y: R.Tensor((5,), "float32")) -> R.Tensor((5,), "float32"):
        result = R.call_py_func("pytorch_add", (x, y), out_sinfo=R.Tensor((5,), "float32"))
        return result
```

**THIS PR** represents the final milestone in the Relax Python integration design, and completes the full feature set for Python/Relax interoperability. The complete Relax-Python integration ecosystem has been built upon the following key design principles:

### Cross-level Calls
- **Two-way interoperability**: Python functions can invoke Relax/TIR/packed functions, and Relax functions can invoke Python functions via `R.call_py_func`
- **Seamless data conversion**: DLPack enables efficient conversion between TVM Tensors and PyTorch Tensors with minimal overhead

### Just-in-time (JIT) Compilation  
- **Delayed compilation**: TIR and Relax functions are compiled only when the IRModule is instantiated
- **Flexible integration**: Allows late-stage modifications and seamless integration with the Python runtime
- **Relax VM execution**: Compiled Relax functions are executed using a Relax VM created at instantiation time

### Conversion between Relax and Python Functions
- **IRModule printer**: Converts Relax functions into executable Python code for debugging and deployment
- **Operator mapping**: High-level Relax operators (e.g., `R.nn.relu`) are mapped to corresponding PyTorch APIs (e.g., `F.relu`)
- **Multi-stage conversion**: Can happen at any stage of compilation, from early Relax functions to fully lowered TIR modules
- **DLPack integration**: Handles `call_tir` and `call_dps_packed` by converting PyTorch tensors to/from DLPack format

### DIFF
--- a/python/tvm/script/ir_builder/relax/ir.py
+++ b/python/tvm/script/ir_builder/relax/ir.py
@@ -30,6 +30,7 @@ from tvm.relax import (
     Expr,
     ExternFunc,
     ShapeExpr,
+    StringImm,
     TupleGetItem,
     Var,
     VarBinding,
@@ -64,6 +65,7 @@ from tvm.relax.op import (
     call_dps_packed,
     call_inplace_packed,
     call_pure_packed,
+    call_py_func as _call_py_func,
     call_tir,
     call_tir_inplace,
     call_tir_with_grad,
@@ -453,15 +455,15 @@ def call_packed(
 
 @args_converter.auto
 def call_py_func(
-    func_name: py_str,
+    py_func_name: py_str,
     *args: Expr,
     out_sinfo: Union[StructInfo, List[StructInfo]],
 ) -> Call:
     """Create a relax Call, which calls a Python function.
-    
+
     Parameters
     ----------
-    func_name: str
+    py_func_name: str
         The name of the Python function to call. This should correspond to a function
         in the IRModule's pyfuncs attribute.
     *args : Expr
@@ -476,9 +478,6 @@ def call_py_func(
     call: Call
         The created Relax Call for call_py_func operator.
     """
-    from tvm.relax.op import call_py_func as _call_py_func
-    from tvm.relax.expr import StringImm
-    
     if isinstance(out_sinfo, py_tuple):  # type: ignore
         out_sinfo = list(out_sinfo)
     elif not isinstance(out_sinfo, list):
@@ -497,9 +496,9 @@ def call_py_func(
 
     # Convert string to StringImm
     try:
-        func_name_imm = StringImm(func_name) if hasattr(func_name, 'strip') else func_name
-    except:
-        func_name_imm = StringImm(func_name)
+        func_name_imm = StringImm(py_func_name) if hasattr(py_func_name, "strip") else py_func_name
+    except Exception:
+        func_name_imm = StringImm(py_func_name)
     return _call_py_func(func_name_imm, args, out_sinfo)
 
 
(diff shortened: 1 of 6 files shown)


tlopex:
cc @tqchen @MasterJH5574

python/tvm/script/ir_builder/relax/ir.py
```diff
@@ -451,6 +453,55 @@ def call_packed(
...
+            else sinfo.asobject()
+            if isinstance(sinfo, ObjectConvertible)
+            else sinfo
+        )
+        for sinfo in out_sinfo
+    ]
+
+    # Convert string to StringImm
+    try:
+        func_name_imm = StringImm(py_func_name) if hasattr(py_func_name, "strip") else py_func_name
```
MasterJH5574: Is `hasattr(py_func_name, "strip")` intended to check whether `py_func_name` is a Python string? If so, can we just use `isinstance(py_func_name, str)` instead?

tests/python/relax/test_base_py_module_printer.py
```diff
@@ -758,3 +758,213 @@ def test_python_functions_in_irmodule():
...
+    module = ValidationTestModule(device)
+
+    # Test that calling non-existent function raises error
+    x = torch.randn(5, dtype=torch.float32)
+
+    with pytest.raises(ValueError, match="Python function 'non_existent_func' not found"):
+        module.call_py_func("non_existent_func", [x])
+
+
+def test_call_py_func_in_relax_function():
```
MasterJH5574, line 792: Quick question: how is this test different from `test_call_py_func_operator`?

tests/python/relax/test_base_py_module_printer.py
MasterJH5574, line 1: A high-level comment: let's try to reduce the tests that actually runs (maybe sufficient to keep only one and remove other actual execution). Since runtime execution may increase the test time, we may want to keep execution as few as possible to make the tests "unit".

python/tvm/script/ir_builder/relax/ir.py
tlopex: Okay, I think we can use `isinstance(py_func_name, str)`

tests/python/relax/test_base_py_module_printer.py
tlopex: Two tests are for different level func: one is simple ops and another is a little bit complex

tests/python/relax/test_base_py_module_printer.py
tlopex: I see. So you mean we have to reduce as much as we can? Since I'm afraid whether only one is not enough

tests/python/relax/test_base_py_module_printer.py
MasterJH5574: In this case I think the “complex” one is sufficient.

tests/python/relax/test_base_py_module_printer.py
MasterJH5574: Right, keep only one or get to a minimal set of tests that need execution.

tlopex:
cc @MasterJH5574

MasterJH5574 (approved):

PR #18320, opened 2025-09-16
[TIR][CUDA] Enhance Precision CUDA Codegen

LeiWang1999:
Previously, `float` constants in codegen were always emitted in **scientific decimal format**, e.g.:

```cpp
bfloat16_t(3.487723e-05f);
```

This could introduce slight **rounding differences** compared to the actual binary representation, since the constant is printed and then re-parsed in decimal. we now emit the value in **hexadecimal floating-point format** (`std::hexfloat`) to preserve the exact binary value, and additionally include the decimal form as a comment for readability:

```cpp
bfloat16_t(0x1.2492492492492p-15f /*3.487723e-05*/)
```

### DIFF
--- a/src/target/source/codegen_cuda.cc
+++ b/src/target/source/codegen_cuda.cc
@@ -1615,13 +1615,17 @@ inline void PrintConst(const FloatImmNode* op, std::ostream& os, CodeGenCUDA* p)
   // Type code is kBFloat
   if (op->dtype.is_bfloat16()) {
     os << "__float2bfloat16_rn";
-    os << '(' << std::scientific << op->value << 'f' << ')';
+    os << '(' << std::hexfloat << op->value << 'f';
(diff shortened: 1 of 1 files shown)


tqchen:
need to update the testcase and we are good to go

LeiWang1999:
> need to update the testcase and we are good to go

added a test case at `tests/python/codegen/test_target_codegen_cuda.py`: https://github.com/apache/tvm/pull/18320/files#diff-de74af0a4254367528294a58410a1c2488d014d3fafdbd8446f88b621358ad4dR804-R820

tqchen:
i meant that we should fix https://ci.tlcpack.ai/blue/organizations/jenkins/tvm-gpu/detail/PR-18320/2/pipeline which likely is caused by expected code in the original form

tqchen (approved):

PR #18323, opened 2025-09-19
Add support for conditional expressions in TVMScript

Hzfengsy:
This PR adds support for conditional expressions in TVMScript parser, which allows developers to use Python-style conditional expressions

```python
@T.prim_func
def func(A: T.buffer((128, 128), "float32")):
    for i, j in T.grid(128, 128):
        A[i, j] = i if i < j else j

@T.prim_func
def expected(A: T.buffer((128, 128), "float32")):
    for i, j in T.grid(128, 128):
        A[i, j] = T.if_then_else(i < j, i, j)
```


MasterJH5574 (approved):
Thanks for introducing the sugar!

PR #18325, opened 2025-09-19
[flashinfer] Support directing JIT to FlashInfer GroupedGemm kernels

neurusL:
in tvm/python/tvm/relax/backend/cuda/flashinfer.py added a gen_grouped_gemm_module
in tvm/tests/python/relax/test_group_gemm_flashinfer.py added tests for different combinations of

- input and output types: ("float8_e4m3fn", "float8_e4m3fn", "bfloat16"), ("float8_e4m3fn", "float8_e4m3fn", "float16"),
- scale granularity of m, n, k: (1, 128, 128),
- scale major mode: "MN", "K"
- mma_sm: 1, 2
- different batch sizes and m_sizes

This PR replaces closed PR  #18322

### DIFF
--- a/tests/python/relax/test_group_gemm_flashinfer.py
+++ b/tests/python/relax/test_group_gemm_flashinfer.py
@@ -1,0 +1,500 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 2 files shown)


tests/python/relax/test_group_gemm_flashinfer.py
```diff
@@ -0,0 +1,500 @@
...
+# specific language governing permissions and limitations
+# under the License.
+
+"""Test for FlashInfer GroupedGemm TVM integration"""
+
+import math
+import numpy as np
+import pytest
+import torch
+from einops import rearrange, reduce, repeat
```
MasterJH5574: There is a CI error about `ModuleNotFoundError: No module named 'einops'`, likely because the CI docker image doesn't have einops installed.

In this case, could you help check if we can replace `einops` operations with torch operations (so we can avoid using einops). I assume that is doable. And in case it isn't, we can move this import into the functions that call these operations, so that einops isn't imported unless executing those functions.

PR #18331, opened 2025-09-22
[Relax] Update BasePyModule with faster DLPack converter for tensor conversion

tlopex:
This PR enhances `BasePyModule` by integrating a faster DLPack converter for efficient tensor conversion between TVM and PyTorch following #18306


tlopex:
cc @MasterJH5574

Issue #18094, opened 2025-06-26
[Bug] Shared Memory Overflow in Generated CUDA Kernel When Compiling ONNX Model via Relay Build


duduhedangdang:
The error occurs during the `relay.build()` phase.

---

### Environment

- OS: Ubuntu 20.04
- Python: 3.8
- TVM Version: v0.9.0 (built from source)
- CUDA Version: 11.1
- GPU: NVIDIA RTX 2080Ti
- ONNX: 1.14.1
- ONNX model: See attached (simplified) model

---

### Steps to reproduce

```python
import onnx
import tvm
from tvm import relay

# Load the uploaded ONNX model
model = onnx.load("original.onnx")

# Provide input shape
shape_dict = {"input": (1, 32, 28, 28)}

# Convert ONNX to Relay
mod, params = relay.frontend.from_onnx(model, shape_dict)

# Try to compile for CUDA
target = "cuda"
with tvm.transform.PassContext(opt_level=3):
    lib = relay.build(mod, target=target, params=params)

[error.txt](https://github.com/user-attachments/files/20922208/error.txt)

[input_info.txt](https://github.com/user-attachments/files/20922217/input_info.txt)

[relay.txt](https://github.com/user-attachments/files/20922219/relay.txt)


### Model Attachment

Due to GitHub not supporting `.onnx` uploads in Issues, the ONNX model has been attached in base64 format as `original.onnx.txt`. You can decode it as follows:

```python
import base64

with open("original.onnx.txt", "r") as f:
    b64 = f.read()

with open("original.onnx", "wb") as f:
    f.write(base64.b64decode(b64))

[original.onnx.txt](https://github.com/user-attachments/files/20922261/original.onnx.txt)

Johnson9009:
The relay is deprecated and deleted thoroughly, please use Relax, the next generation of Graph Level IR of TVM.

PR #18347, opened 2025-09-27
tests(cuda): deterministic VerifyGPUCode test for shared-memory overflow (64KB vs 48KB)

Aminsed:
This PR adds a deterministic, schedule-free CUDA test to validate shared memory limits using tir.analysis.verify_gpu_code.\n\n- Adds a direct TIR kernel test that allocates a 64 KB shared buffer (T.launch_thread + T.allocate) and asserts verification fails when max_shared_memory_per_block=48 KB.\n- Keeps changes scoped to tests; no default behavior altered.\n- Documents an opt-in pipeline toggle (off by default) and a per-device cap hint in tvm/python/tvm/tir/pipeline.py.\n  Usage:\n  \n\nReferences: #18094

### DIFF
--- a/python/tvm/tir/pipeline.py
+++ b/python/tvm/tir/pipeline.py
@@ -120,6 +120,11 @@ def default_tir_pipeline():
                 tir.transform.LowerDeviceKernelLaunch(),
             ]
         )
+        # Optional GPU verification based on PassContext configuration.
+        if bool(config.get("tir.verify_gpu_code", False)):
+            # Conservative default cap; users can override per device via PassContext config.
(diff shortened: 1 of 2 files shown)


tqchen:
given this is something that can be composed (after running pipeine), we will likely not need this in the default pipeline to reduce complexity there

Aminsed:
@tqchen Thanks for the guidance. I have removed the optional `VerifyGPUCode` hook from the default TIR pipeline to keep the pipeline minimal.

tqchen:
Thanks, given the remaining changes can be done in downstream, we will close this for now

Issue #18338, opened 2025-09-23
[Bug][Relax][Torch] from_exported_program fails on unbind producing single-element tuple


tinywisdom:
### Expected behavior

+ `unbind` producing a single-element tuple should be represented as a tuple in Relax IR, not misinterpreted as a tensor.

+ The frontend should either:

  + Correctly lower to Tuple with one element (each tensor shaped `(3,)` in this case), or

  + Gracefully reject with a clear Python exception, not an internal assertion failure.

### Actual behavior

When importing a PyTorch `torch.exported` program into TVM Relax, if the model applies `unbind(dim=0)` to a tensor with dimension=1 along that axis, the frontend crashes with:
```
Check failed: (opt) is false: The struct info of Tuple must be TupleStructInfo,
but expression lv3 has struct info R.Tensor((1, 3), dtype="float32")
```
This indicates a mismatch between expected `TupleStructInfo` vs. actual `TensorStructInfo`.

```
[INFO] start importing exported program into TVM Relax...
[REPRODUCED] Caught exception while importing:
Check failed: (opt) is false: The struct info of Tuple must be TupleStructInfo,
but expression lv3 has struct info R.Tensor((1, 3), dtype="float32")

tvm.error.InternalError: Check failed: (opt) is false: The struct info of Tuple must be TupleStructInfo, but expression lv3 has struct info R.Tensor((1, 3), dtype="float32")
[...]/src/relax/ir/block_builder.cc:64: Warning: BlockBuilder destroyed with remaining blocks!
```

### Environment

+ OS: (Ubuntu 22.04.4 LTS (x86_64))
+ TVM version: (release v0.21.0)
+ Python: (3.10.16)
+ LLVM: (17.0.6)

### Steps to reproduce

```python
import torch
import torch.nn as nn

# Minimal model: gather -> max -> unbind(0)
class MyModel(nn.Module):
    def forward(self, x, y):
        # x: (2, 3), y: (1, 3)
        indices = torch.zeros((1,) + x.size()[1:], dtype=torch.long, device=x.device)
        x_gathered = torch.gather(x, 0, indices)   # (1, 3)
        compared = torch.max(x_gathered, y)        # (1, 3)
        outs = compared.unbind(0)                  # tuple of length 1, each (3,)
        return outs

def get_inputs():
    torch.manual_seed(0)
    x = torch.randn(2, 3)
    y = torch.randn(1, 3)
    return x, y

if __name__ == "__main__":
[...]

PR #18351, opened 2025-09-28
[Relax][Frontend][Torch] Fix parsing error when input dimension of unbind is 1

eaten-cake:
fix issue #18338 
When the input dimension of unbind is 1, we should directly apply squeeze instead of performing split followed by squeeze. This issue has been fixed, and a corresponding regression test has been added.
Previously, the error occurred because when the output of split contains only a single tensor, the `struct_info` of the expression is `TensorStructInfo` rather than `TupleStructInfo`. As a result, the `TupleGetItem` node generated at https://github.com/apache/tvm/blob/f68651f035d08024c05f218182b5c003ad814eb5/python/tvm/relax/frontend/torch/base_fx_graph_translator.py#L1279-L1280 would fail.

### DIFF
--- a/tests/python/relax/test_frontend_from_exported_program.py
+++ b/tests/python/relax/test_frontend_from_exported_program.py
@@ -3254,7 +3254,9 @@ def test_unbind():
     @tvm.script.ir_module
     class expected3:
         @R.function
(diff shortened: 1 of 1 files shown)


eaten-cake:
resolve #18338

Issue #18356, opened 2025-10-02
[Bug] [feature][relax.frontend.torch] from_exported_program rejects fused GRU (aten::gru.input) with “Unsupported function types ['gru.input']”


tinywisdom:
### Expected behavior

Translating a tiny torch.exported model that contains nn.GRU fails in TVM Relax Torch frontend with:
```
AssertionError: Unsupported function types ['gru.input']
```
This looks like missing coverage for the fused RNN op (aten::gru overload gru.input) in from_exported_program. Even if unsupported today, it would be helpful to either (a) add lowering for GRU, or (b) fail with a more actionable message / doc pointer describing the current RNN support scope and workarounds.

### Actual behavior

```
=== Versions ===
Python : 3.10.16 | packaged by conda-forge | (main, Apr  8, 2025, 20:53:32) [GCC 13.3.0]
Torch  : 2.8.0+cu128
TVM   : 0.21.0
================
torch.export: OK
TVM from_exported_program: FAILED as expected
Error: Unsupported function types ['gru.input']
Traceback (most recent call last):
  ...
  File ".../tvm/relax/frontend/torch/base_fx_graph_translator.py", line 116, in _check_unsupported_func_type
    assert not missing_func_types, f"Unsupported function types {missing_func_types}"
AssertionError: Unsupported function types ['gru.input']
```

### Environment

- OS: (Ubuntu 22.04.4 LTS (x86_64))
- TVM version: (release v0.21.0)
- Python: (3.10.16)
- LLVM: (17.0.6)

### Steps to reproduce

```python
# minimal_gru_from_exported_program_repro.py
import torch
import torch.nn as nn
import sys

def print_env():
    print("=== Versions ===")
    print("Python :", sys.version.replace("\n", " "))
    print("Torch  :", torch.__version__)
    try:
        import torchaudio, torchvision
        print("torchaudio:", getattr(torchaudio, '__version__', 'unknown'))
        print("torchvision:", getattr(torchvision, '__version__', 'unknown'))
    except Exception:
        pass
    try:
        import tvm
        print("TVM   :", tvm.__version__)
    except Exception as e:
        print("TVM   : import error ->", e)
    print("================")

class M(nn.Module):
    """
    Tiny model that triggers a fused GRU op (aten::gru/gru.input)
    in the ExportedProgram graph.
    """
    def __init__(self):
        super().__init__()
        self.proj = nn.Linear(12, 10)   # map to GRU input_size=10
[...]

PR #18360, opened 2025-10-06
[Relax][PyTorch] Support gru op for ExportedProgram importer

tlopex:
This pr supports `gru.input` for ExportedProgram importer.
This links to issue #18356

### DIFF
--- a/python/tvm/relax/frontend/torch/base_fx_graph_translator.py
+++ b/python/tvm/relax/frontend/torch/base_fx_graph_translator.py
@@ -102,6 +102,9 @@ class BaseFXGraphImporter(metaclass=abc.ABCMeta):
             return [self._retrieve_args(x) for x in node]
         elif isinstance(node, dict):
             return {self._retrieve_args(k): self._retrieve_args(v) for k, v in node.items()}
+        elif node is None:
+            # Convert None to a Relax null_value expression
+            return relax.op.null_value()
         else:
             return node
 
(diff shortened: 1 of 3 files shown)


mshr-h:
@tlopex 
The root problem is in https://github.com/apache/tvm/blob/f30b29c2c5e35eb975ae8926fb7ebfae4d817a50/python/tvm/relax/frontend/torch/exported_program_translator.py#L964, where `run_decompositions()` isn't an in-place operation. We need to change to something like this so that the decomposed ep is used.
```python
    decomposed_ep = exported_program.run_decompositions()

    return ExportedProgramImporter().from_exported_program(
        decomposed_ep,
        keep_params_as_input,
        unwrap_unit_return_tuple,
        no_bind_return_tuple,
    )
```

I wrote the original code and I think it was wrong.

mshr-h:
With proper decomposition, almost 40% (75 failed, 101 passed, 2 warnings) of the exported program frontend tests will fail so I think it might be easier to migrate gradually.

tlopex:
@mshr-h Got it! I'll consider how to update it. Shall we merge this pr first and then get on it?

mshr-h:
@tlopex Okay!

tlopex:
@mshr-h  Sorry for late reply. I checked the code and found there will be 40% of the exported program frontend tests will fail if modified correctly. Maybe the first step is try to support unsupported ops because of decomposition step by step? But during the fixing, the tests may remain fail. If you think this is good to do, I can get on it this week

mshr-h:
@tlopex Thanks.
One possible approach is to add a new importer flag, such as `run_ep_decomposition`, defaulting to `False`. And then gradually switch it to `True` in unittests. Once every unittests has been set to `True`, we can remove the flag entirely.

tlopex:
Get it! Let me start fixing it @mshr-h

PR #18382, opened 2025-10-21
Support integer for exp operator

jikechao:
This PR addresses the issue where tvm.tir.exp does not support integer types (e.g., int32, int64), causing an InternalError during LLVM code generation with the message, The issue arises because the `llvm.exp` intrinsic expects floating-point inputs, but no type conversion is performed for integer inputs.

I opened this PR to solve it via type conversion. This change aligns the behavior of tir.exp with libraries like PyTorch and NumPy, which implicitly convert integer inputs to floating-point types for their exponential functions.


jikechao:
@tvm-bot rerun

PR #18385, opened 2025-10-22
[Web] Upgrade web runtime to new FFI

akaashrp:
Fix issues arising from new FFI and renaming NDArray to Tensor


akaashrp:
Please let me know if I should not be pushing the updated package-lock.json

tqchen (approved):

Issue #18362, opened 2025-10-08
[bug/feature][relax.frontend.torch] from_exported_program rejects randn.default (blocks repro that stresses advanced-indexing write + tuple output)


tinywisdom:
### Actual behavior

Importing an exported PyTorch program fails early in TVM Relax Torch frontend with:
```
AssertionError: Unsupported function types ['randn.default']
```
The model is intentionally small and uses advanced indexing write (two index tensors on the same dims) and returns a tuple (y, L) to exercise relax::Tuple creation. However, the importer aborts first due to missing coverage for aten::randn.default, so we cannot reach the later code path.

### Environment

- OS: (Ubuntu 22.04.4 LTS (x86_64))
- TVM version: (release v0.21.0)
- Python: (3.10.16)
- LLVM: (17.0.6)
- Pytorch: (2.8.0)

### Steps to reproduce

```python
# Purpose: Trigger importer on advanced indexing (two index tensors) + tuple outputs.
# Current blocker: Unsupported function types ['randn.default'].

import torch
import torch.nn as nn
import torch.nn.functional as F

def main():
    print("==== Versions ====")
    import tvm
    print("PyTorch:", torch.__version__)
    print("TVM    :", tvm.__version__)

    torch.manual_seed(0)
    N = 5

    class M(nn.Module):
        def forward(self, x):
            L = torch.zeros(N, N, dtype=x.dtype, device=x.device)
            idx = torch.arange(N, device=x.device)
            v = torch.randn(N, device=x.device)       # <-- aten::randn.default
            v = F.elu(v) + 1.0 + 1e-8                # keep values > 0
            L[idx, idx] = v                           # advanced indexing write
            y = x + 1
            return y, L                                # tuple output -> relax::Tuple

    m = M().eval()
    ex_in = torch.randn(2, N)

    # 1) export OK
    from torch.export import export as torch_export
    ep = torch_export(m, (ex_in,))
    print("torch.export: OK")

    # 2) TVM import -> fails on unsupported randn
    from tvm.relax.frontend.torch import from_exported_program
    print("about to call from_exported_program(ep)...")
    mod = from_exported_program(ep)  # <-- aborts with Unsupported function types ['randn.default']
    print("from_exported_program: OK (if you see this, the randn support may have landed)")

if __name__ == "__main__":
    main()
```

### Output
```
==== Versions ====
PyTorch: 2.8.0a0+gitba56102
TVM    : 0.21.0
torch.export: OK
[...]

PR #18396, opened 2025-10-25
[Relax][PyTorch] Fix AssertionError: Unsupported function types `randn.default`

cchung100m:
This PR is trying to fix issues https://github.com/apache/tvm/issues/18362.

### Solution:
For now, return zeros with the correct shape since random initialization [randn](https://docs.pytorch.org/docs/stable/generated/torch.randn.html#torch-randn) is mainly used during training, not inference. This is a temporary solution until Relax adds proper random number generation support.

### Future Enhancement:
A proper implementation would require adding:
- A `relax.op.random_normal(shape, mean=0, std=1, dtype, seed)` operator
- TIR lowering rules for CPU/GPU backends
- State management for reproducible random sequences

### DIFF
--- a/python/tvm/relax/frontend/torch/base_fx_graph_translator.py
+++ b/python/tvm/relax/frontend/torch/base_fx_graph_translator.py
@@ -2275,6 +2275,14 @@ class BaseFXGraphImporter(metaclass=abc.ABCMeta):
         x = self.env[node.args[0]]
         return self.block_builder.emit(relax.op.zeros_like(x))
 
+    def _randn(self, node: fx.Node) -> relax.Var:
+        import torch
+
+        dtype = self._convert_data_type(
+            node.kwargs.get("dtype", torch.get_default_dtype()), self.env
+        )
+        return self.block_builder.emit(relax.op.zeros(node.args[0], dtype))
+
     def _eye(self, node: fx.Node) -> relax.Var:
         args = self.retrieve_args(node)
         n = args[0]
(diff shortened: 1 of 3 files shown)


cchung100m:
Hi @tlopex @mshr-h 

Any suggestions would be appreciated if you are available.

python/tvm/relax/frontend/torch/base_fx_graph_translator.py
```diff
@@ -2275,6 +2275,14 @@ def _empty_like(self, node: fx.Node) -> relax.Var:
         x = self.env[node.args[0]]
         return self.block_builder.emit(relax.op.zeros_like(x))
 
+    def _randn(self, node: fx.Node) -> relax.Var:
+        import torch
+
+        dtype = self._convert_data_type(
+            node.kwargs.get("dtype", torch.get_default_dtype()), self.env
+        )
+        return self.block_builder.emit(relax.op.zeros(node.args[0], dtype))
```
tlopex, line 2284: Could you add a `TODO` note here indicates that we still need to implement it properly?

mshr-h:
I don't know if we really need a temp support for rand. As we added custom op converter in #18544 , users can now define randn conveter **in their codebase** instead of editing the tvm torch frontend files.

cchung100m:
Hi @tlopex @mshr-h 
Thanks for the prompt reply. I have no other thoughts for #18362 if  users can now define randn conveter by custom op converter. I can close this PR if no further actions needed.

PR #18400, opened 2025-10-26
[Relax][PyTorch] Add support for decomposed operators in extended unary ops tests

tlopex:
This PR adds support for decomposed operators that ops needed in the extended unary ops tests.


tlopex:
cc @mshr-h

PR #18414, opened 2025-11-02
[Relax][PyTorch] Add support for decomposed operators and fix IR of ops tests(4)

tlopex:
This pr fixes ops like`prod`, `var`, `std`, `take`, `flip`, `unflatten` and so on


tlopex:
cc @mshr-h

PR #18415, opened 2025-11-02
[Web] Fix arrayDecodeStorage scope issue for q0f32 models

akaashrp:
This PR fixes scope issues caused by the different flow for q0f32 models in `ArrayDecodeStorage`. Currently, for q0f32 models, during the `fetchTensorCacheInternal` call in runtime.ts, the loop for loading shards from the cache fails on the `arrayDecodeStorage` JS call due to stack pollution. In particular, the type_index for the return value is read as 0x3ad70000 = 987168768 despite pre-storing the result on the stack to be null when creating packed functions:
```
stack.storeI32(retOffset, TypeIndex.kTVMFFINone);
// clear off the extra zero padding before ptr storage
stack.storeI32(retOffset + SizeOf.I32, 0);
stack.commitToWasmMemory();
```
This causes it to go through the default case in `retValueToJS`, which attempts to attach the object to the current scope. This leads to scope issues since the return type of ArrayDecodeStorage is void, and thus, the JS code does not explicitly handle this. The resolution for this is to define ArrayDecodeStorage using `def_packed` and manually set the `type_index` of the return value to 0 before returning from ArrayDecodeStorage to ensure that `retValueToJS` sees a None type when performing its conversion logic.

### DIFF
--- a/web/emcc/wasm_runtime.cc
+++ b/web/emcc/wasm_runtime.cc
@@ -148,20 +148,20 @@ void ArrayDecodeStorage(Tensor cpu_arr, std::string bytes, std::string format, s
 
 TVM_FFI_STATIC_INIT_BLOCK() {
   namespace refl = tvm::ffi::reflection;
-  refl::GlobalDef().def_packed("tvmjs.array.decode_storage",
-                               [](ffi::PackedArgs args, ffi::Any* ret) {
-                                 Tensor cpu_arr = args[0].cast<Tensor>();
-                                 auto bytes = args[1].cast<ffi::Bytes>();
-                                 std::string format = args[2].cast<ffi::String>().operator std::string();
-                                 std::string dtype = args[3].cast<ffi::String>().operator std::string();
-                                 ArrayDecodeStorage(cpu_arr, bytes, format, dtype);
-                                 if (ret != nullptr) {
-                                  auto* ret_data = reinterpret_cast<TVMFFIAny*>(ret);
-                                  ret_data->type_index = TVMFFITypeIndex::kTVMFFINone;
-                                  ret_data->zero_padding = 0;
-                                  ret_data->v_int64 = 0;
-                                 }
-                               });
+  refl::GlobalDef().def_packed(
+      "tvmjs.array.decode_storage", [](ffi::PackedArgs args, ffi::Any* ret) {
+        Tensor cpu_arr = args[0].cast<Tensor>();
+        auto bytes = args[1].cast<ffi::Bytes>();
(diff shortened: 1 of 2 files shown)


web/src/runtime.ts
```diff
@@ -1396,7 +1396,18 @@ export class Instance implements Disposable {
           });
           const recSource = buffer.slice(rec.byteOffset, rec.byteOffset + rec.nbytes);
           // first sync copy to cpu.
-          this.ctx.arrayDecodeStorage(cpu_arr, new Uint8Array(recSource), rec.format, rec.dtype);
+          this.withNewScope(() => {
+            const retValue = this.ctx.arrayDecodeStorage(
```
tqchen: would be good to document the rationale here, any better ways than try catch?

akaashrp: I've documented the rationale in the PR description. If you'd like, I can add a comment in the code too. I've updated the error handling in the catch block to throw when the error is not something related to the object not being found in the current scope. I don't think there's a better way than a try-catch to avoid throwing an error in the aforementioned case - maybe creating an if block that mirrors the if-else conditions in ArrayDecodeStorage (wasm_runtime.cc)?

tqchen:
Ok, reading https://github.com/apache/tvm/blob/main/web/emcc/wasm_runtime.cc#L125, seems the function does not have a return value. Would be important to double check in here on what is going on, maybe related to a bug.

web/emcc/wasm_runtime.cc
```diff
@@ -145,7 +148,20 @@ void ArrayDecodeStorage(Tensor cpu_arr, std::string bytes, std::string format, s
...
   namespace refl = tvm::ffi::reflection;
-  refl::GlobalDef().def("tvmjs.array.decode_storage", ArrayDecodeStorage);
+  refl::GlobalDef().def_packed(
+      "tvmjs.array.decode_storage", [](ffi::PackedArgs args, ffi::Any* ret) {
+        Tensor cpu_arr = args[0].cast<Tensor>();
+        auto bytes = args[1].cast<ffi::Bytes>();
+        std::string format = args[2].cast<ffi::String>().operator std::string();
+        std::string dtype = args[3].cast<ffi::String>().operator std::string();
+        ArrayDecodeStorage(cpu_arr, bytes, format, dtype);
+        if (ret != nullptr) {
```
tqchen, line 158: This appears to be an workaround but we did not yet know why the return value is polluted, would be good to know

tqchen:
Merge as a temp workaround

Issue #18425, opened 2025-11-07
[Bug] Cannot find intrinsic declaration, possible type mismatch: llvm.log2


Cookiee235:
### Actual behavior

```
Traceback (most recent call last):
  File "test", line 37, in <module>
    ref_output = compile_mod(sch.mod, np_inputs, op_outputs, opt_level=0)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/share_container/LLMFuzz/TirFuzz/bugs/10-24_20-21/topi.log2_1.py", line 10, in compile_mod
    ref_mod = tvm.build(mod, target='llvm')
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/software/tvm-latest/python/tvm/driver/build_module.py", line 59, in build
    return tvm.tir.build(mod, target, pipeline)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/software/tvm-latest/python/tvm/tir/build.py", line 239, in build
    return tir_to_runtime(host_mod, device_mod_dict, target_host)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/software/tvm-latest/python/tvm/tir/build.py", line 149, in tir_to_runtime
    mhost = codegen_build(mhost_all, target_host)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/software/tvm-latest/python/tvm/tir/build.py", line 131, in codegen_build
    return bf(mod, target)
           ^^^^^^^^^^^^^^^
  File "python/tvm_ffi/cython/function.pxi", line 758, in core.Function.__call__
  File "<unknown>", line 0, in tvm::codegen::LLVMModuleNode::Init(tvm::IRModule const&, tvm::Target const&)
  File "<unknown>", line 0, in tvm::codegen::CodeGenCPU::AddFunction(tvm::GlobalVar const&, tvm::tir::PrimFunc const&)
  File "<unknown>", line 0, in tvm::codegen::CodeGenLLVM::AddFunctionInternal(tvm::GlobalVar const&, tvm::tir::PrimFunc const&)
  File "<unknown>", line 0, in tvm::codegen::CodeGenCPU::VisitStmt_(tvm::tir::AssertStmtNode const*)
[...]

PR #18426, opened 2025-11-07
Support integer type input for log and log2

jikechao:
Adds support for integer inputs in `topi.log` and `topi.log2` by automatically converting them to float32, aligning with NumPy's implicit float promotion behavior.

Fix https://github.com/apache/tvm/issues/18425


jikechao:
cc @tqchen @Hzfengsy @tlopex

PR #18428, opened 2025-11-10
[Relax][PyTorch] Add support for decomposed operators and fix IR of ops tests(8)

tlopex:
This pr fixes ops like `pixel_shuffle`.


tlopex:
cc @mshr-h

Issue #17818, opened 2025-04-09
[Tracking Issue] [Relax][PyTorch] Dynamic shape support in ExportedProgram


mshr-h:
### This issue tracks the progress of dynamic shapes support in ExportedProgram
- [ ] Basic support for dynamic shape #17817 
- [ ] Support for ange constraints

demoncoder-crypto:
Can I get assigned on this please if its open to community contributions?

mshr-h:
@demoncoder-crypto 
I haven't started to work on range constraints support due to my bandwidth. So feel free to take the issue!

demoncoder-crypto:
#17898 Check implementation, If my direction is correct I will iteratively improve everything

PR #17898, opened 2025-04-26
feat(relax/frontend/torch): Add basic range constraint support 

demoncoder-crypto:
Addresses #17818

PR #18429, opened 2025-11-10
[Relax][PyTorch] Support basic range constraints

guan404ming:
## Related Issue

related to #17818

## Why 

- Add support for PyTorch range constraints in Relax frontend
- Enables proper handling of dynamic shapes with min/max bounds

## How

- Extract range constraints from PyTorch ExportedProgram and store in `shape_var_constraints` function attribute
- Add `test test_dynamic_shape_with_range_constraints` to verify constraint extraction works correctly

### DIFF
--- a/tests/python/relax/test_frontend_from_exported_program.py
+++ b/tests/python/relax/test_frontend_from_exported_program.py
@@ -6536,13 +6536,13 @@ def test_dynamic_shape_with_range_constraints():
     class Expected:
         @R.function
         def main(
-            x1: R.Tensor(("s24", 4), dtype="float32"), x2: R.Tensor(("s24", 4), dtype="float32")
-        ) -> R.Tuple(R.Tensor(("s24", 4), dtype="float32")):
-            s24 = T.int64(is_size_var=True)
-            R.func_attr({"tir_var_upper_bound": {"s24": 64}})
+            x1: R.Tensor(("s0", 4), dtype="float32"), x2: R.Tensor(("s0", 4), dtype="float32")
+        ) -> R.Tuple(R.Tensor(("s0", 4), dtype="float32")):
+            s0 = T.int64(is_size_var=True)
+            R.func_attr({"tir_var_upper_bound": {"s0": 64}})
             with R.dataflow():
-                lv: R.Tensor((s24, 4), dtype="float32") = R.add(x1, x2)
-                gv: R.Tuple(R.Tensor((s24, 4), dtype="float32")) = (lv,)
+                lv: R.Tensor((s0, 4), dtype="float32") = R.add(x1, x2)
+                gv: R.Tuple(R.Tensor((s0, 4), dtype="float32")) = (lv,)
                 R.output(gv)
             return gv
 


guan404ming:
cc @mshr-h

python/tvm/relax/frontend/torch/exported_program_translator.py
```diff
@@ -1121,13 +1136,19 @@ def create_input_vars(
...
                 torch_dtype = exported_program.state_dict[spec.target].dtype
 
-            # TODO(mshr-h): Support range constraints
-            relax_shape = [
-                torch_symbol_to_relax_var.setdefault(str(s), tvm.tir.SizeVar(str(s), "int64"))
-                if isinstance(s, torch.SymInt)
-                else s
-                for s in torch_shape
-            ]
+            # Create TIR variables for symbolic dimensions
```
mshr-h: Looks like there's no functional changes. Any reason for the change?

python/tvm/relax/frontend/torch/exported_program_translator.py
```diff
@@ -1149,14 +1170,22 @@ def from_exported_program(
...
         inputs_vars.update(parameter_buffer_constant_vars)
 
         # Initialize the block builder with a function and a dataflow block.
         self.block_builder = relax.BlockBuilder()
         func_name = "main"
         func_attrs = {"num_input": len(user_input_vars)} if keep_params_as_input else None
+        if range_constraints:
+            if func_attrs is None:
+                func_attrs = {}
+            func_attrs["shape_var_constraints"] = range_constraints
```
mshr-h: Please use `tir_var_upper_bound` to annotate upper bound.
I grepped the tvm code base and I found that there's no lower bound annotation. So I don't think we need to keep it at the moment. If we have a real use case for it, it's fine to keep it in the Relax module.
https://github.com/apache/tvm/blob/main/src/relax/transform/static_plan_block_memory.cc#L62-L66

tests/python/relax/test_frontend_from_exported_program.py
```diff
@@ -6663,5 +6527,29 @@ def forward(self, x):
...
+
+    example_args = (torch.randn(8, 4), torch.randn(8, 4))
+    batch = torch.export.Dim("batch", min=1, max=64)
+    dynamic_shapes = {"x1": {0: batch}, "x2": {0: batch}}
+    exported_program = export(DynamicModel(), args=example_args, dynamic_shapes=dynamic_shapes)
+
+    mod = from_exported_program(exported_program)
+
+    main_func = mod["main"]
+    assert hasattr(main_func, "attrs"), "Function should have attributes"
```
mshr-h: Please use structual equality instead of manually checking the attributes.

tests/python/relax/test_frontend_from_exported_program.py
```diff
@@ -6663,5 +6527,29 @@ def forward(self, x):
...
+    class DynamicModel(torch.nn.Module):
+        def forward(self, x1, x2):
+            return torch.ops.aten.add.Tensor(x1, x2)
+
+    example_args = (torch.randn(8, 4), torch.randn(8, 4))
+    batch = torch.export.Dim("batch", min=1, max=64)
+    dynamic_shapes = {"x1": {0: batch}, "x2": {0: batch}}
+    exported_program = export(DynamicModel(), args=example_args, dynamic_shapes=dynamic_shapes)
+
+    mod = from_exported_program(exported_program)
```
mshr-h: Should be `from_exported_program(exported_program, run_ep_decomposition=True)`
See #18399

python/tvm/relax/frontend/torch/exported_program_translator.py
```diff
@@ -1097,11 +1097,28 @@ def create_convert_map(
...
+                        lower = int(value_range.lower)
+                    except (OverflowError, AttributeError, TypeError):
+                        continue
+
+                    try:
+                        upper = int(value_range.upper)
+                    except (OverflowError, AttributeError, TypeError):
+                        continue
+
+                    range_constraints[symbol_name] = (lower, upper)
```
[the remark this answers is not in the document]
mshr-h: Make sense to me.
@guan404ming Can you apply the change?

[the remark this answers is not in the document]
guan404ming: Sure, thanks!

python/tvm/relax/frontend/torch/exported_program_translator.py
guan404ming: Sorry about this, it seems like legacy change in #17898. Let me revert this.

tests/python/relax/test_frontend_from_exported_program.py
guan404ming: Thanks for providing useful info!

tests/python/relax/test_frontend_from_exported_program.py
guan404ming: Sure, I've updated the test to follow the convention.

python/tvm/relax/frontend/torch/exported_program_translator.py
guan404ming: Thanks for letting me know, I've updated to use `tir_var_upper_bound` to annotate upper bound.

tests/python/relax/test_frontend_from_exported_program.py
guan404ming: Should we also need to update tests in https://github.com/apache/tvm/blob/f574031657faa035ffd1875050316b11821187cf/tests/python/relax/test_frontend_from_exported_program.py?

I could help with this if needed.

tlopex: Of course you can help update it!

guan404ming: Sure! I would open another PR to handle this, thanks for your reply.

mshr-h (approved):
LTGM. Thanks!

guan404ming:
Thanks for your detailed comments and reviews!

Issue #12637, opened 2022-08-29
[ci] Move auto-cc bot edits on PRs to a comment


driazati:
To enable the use of https://github.blog/changelog/2022-08-23-new-options-for-controlling-the-default-commit-message-when-merging-a-pull-request/ so we can have good defaults for commit messages, we need to ensure that PR descriptions don't contain @ tags of GitHub users. We can lint for this but the auto-cc bot edits PR descriptions to @ people on PRs, so we need to change this bot to leave a comment instead. Right now we have other bots that leave comments on PRs, this would be a good time to refactor them a bit and merge everything into 1 big bot comment. This could also contain helpful links for contributors and let us get rid of the [pr template](https://github.com/apache/tvm/blob/main/.github/PULL_REQUEST_TEMPLATE.md).

To implement the lint check there are a few sequential PRs:
- [ ] #12609
- [ ] #12604
- [ ] #12367

Then to implement the unified bot comment on PRs:
- [ ] make sure commenting jobs cancel previous jobs on the same PR to avoid races
- [ ] merge commenting workflows (skipped tests, docs) into a single sequential workflow
- [ ] #12612

PR #18432, opened 2025-11-11
[CI] Enable username checks in PR title and body

guan404ming:
## Related Issue

#12637

## Why

Since #12637 is done, we could enable the username check

## How

uncomment the check


guan404ming:
cc @driazati

guan404ming:
Thanks for the review!

PR #18401, opened 2025-10-27
[Relax][PyTorch] Enable decomposition for unary ops and refactor tests

tlopex:
This pr enables decomposition flag for all unary ops tests and refactors some decomposed tests.

PR #18433, opened 2025-11-12
[Relax][PyTorch] Add decomposed operator support for Conv

guan404ming:
## Related Issue

https://github.com/apache/tvm/pull/18401

## Why
- When `run_ep_decomposition=True` is enabled, PyTorch decomposes high-level operators like conv2d.default into
lower-level `convolution.default`, which is not supported yet.

## How

- Added _convolution() method in base_fx_graph_translator.py to handle the generic aten.convolution.default operator
- Updated test_keep_params to use run_ep_decomposition=True


guan404ming:
cc @tlopex @mshr-h

Issue #17737, opened 2025-03-11
[FRONTEND][ONNX] Error converting operator Transpose: TVMError: PermuteDims expects the number of input axes to equal the ndim of the input tensor.


coffezhou:
### Expected behavior
The onnx frontend should import the model correctly.

### Actual behavior

`Error converting operator Transpose, with inputs: [R.where(v0_0, v3_0, lv14)]
Traceback (most recent call last):
  File "/home/carla/Documents/test_tvm/test_tvm_using_onnx_model/random_test_tvm.py", line 172, in <module>
    main()
  File "/home/carla/Documents/test_tvm/test_tvm_using_onnx_model/random_test_tvm.py", line 161, in main
    check_correctness(onnx_model)
  File "/home/carla/Documents/test_tvm/test_tvm_using_onnx_model/random_test_tvm.py", line 104, in check_correctness
    tvm_model = from_onnx(model, opset=opset, keep_params_in_input=True)
  File "/home/carla/Documents/tvm/python/tvm/relax/frontend/onnx/onnx_frontend.py", line 3690, in from_onnx
    return g.from_onnx(graph, opset)
  File "/home/carla/Documents/tvm/python/tvm/relax/frontend/onnx/onnx_frontend.py", line 3321, in from_onnx
    self._construct_nodes(graph)
  File "/home/carla/Documents/tvm/python/tvm/relax/frontend/onnx/onnx_frontend.py", line 3501, in _construct_nodes
    raise err
  File "/home/carla/Documents/tvm/python/tvm/relax/frontend/onnx/onnx_frontend.py", line 3498, in _construct_nodes
    op = self.bb.normalize(op)
  File "/home/carla/Documents/tvm/python/tvm/relax/block_builder.py", line 667, in normalize
    return _ffi_api.BlockBuilderNormalize(self, expr)  # type: ignore
  File "/home/carla/Documents/tvm/python/tvm/_ffi/_ctypes/packed_func.py", line 245, in __call__
    raise_last_ffi_error()
  File "/home/carla/Documents/tvm/python/tvm/_ffi/base.py", line 465, in raise_last_ffi_error
    raise py_err
tvm._ffi.base.TVMError: Traceback (most recent call last):
[...]

PR #18435, opened 2025-11-12
[FRONTEND][ONNX] Fix operator Transpose: TVMError: PermuteDims expects the number of input axes to equal the ndim of the input tensor

cchung100m:
This PR is trying to fix issues https://github.com/apache/tvm/issues/17737. 

The original Transpose._impl_v13 method did not handle:

- Scalar inputs: according to ONNX SPEC, transposing a scalar should return the scalar unchanged
- Validation of axes vs. tensor dimensions


tlopex (approved):
LGTM!

PR #18447, opened 2025-11-14
[Relax][PyTorch] Add lower bound support for range constraints

guan404ming:
## Related
- #18429

## Why

- Previously only extracted upper bounds, discarding lower bound information
- Enables better memory planning and optimization with full range [min, max]

## How

- Extract both tir_var_lower_bound and tir_var_upper_bound from PyTorch `ExportedProgram.range_constraints`
- Update C++ transforms (static_plan_block_memory.cc, adjust_matmul_order.cc) to use lower bounds
- Default: lower=0 if not specified, upper=INT64_MAX if not specified


guan404ming:
cc @mshr-h @tlopex

guan404ming:
Thanks!

PR #18449, opened 2025-11-14
[Relax][PyTorch] Add decomposed operator support for Pad

guan404ming:
## Related Issue

- https://github.com/apache/tvm/pull/18401

## Why

- When run_ep_decomposition=True is enabled, PyTorch decomposes pad operators into lower-level operations:
  - Constant mode → `constant_pad_nd.default`
  - Reflect/Replicate modes → `index.Tensor` with None indices
  - Circular mode → `copy.default` and `slice` operations
- Some of the decomposed operators were not supported, causing failures

## How

- Added support for `constant_pad_nd.default` and `copy.default` operator
- Fixed `_index_tensor` to handle None indices by:
  - Using `take` operation when only one dimension is indexed (optimization)
  - Converting `None` to explicit `arange` for general cases
- Updated test_pad to use run_ep_decomposition=True


guan404ming:
cc @tlopex

guan404ming:
Thanks!

PR #18455, opened 2025-11-15
[Relax] Fix flaky test_conv2d_offload by increasing float32 tolerance

tlopex:
The `test_conv2d_offload` test for float32 dtype was intermittently failing in CI with errors like:
```
Mismatched elements: 17 / 524288 (0.00324%)
Max absolute difference: 0.02001762
Max relative difference: 3193.5
```
The test was using `rtol=1e-2, atol=1e-2` (0.01) tolerance, which may be too strict for comparing cuDNN and LLVM implementations. The max absolute difference of ~0.02 exceeded the threshold, causing flaky test failures. This PR increases the tolerance for float32 from `1e-2` to `2.5e-2` (0.025) to accommodate the observed numerical differences between cuDNN and LLVM convolution implementations.


tlopex:
cc @tqchen

PR #18457, opened 2025-11-15
[CI] Fix crash when grep finds no matches

guan404ming:
## Why

- The original implementation would crash when grep returned a non-zero exit code (pattern not found)

## How

- Add `|| true` to prevent early return
- Minor update: remove redundant stage definiation since we got default stage config

## Result

if I don't have ci docker images installed

**before**

image

**after**

image


guan404ming:
Thanks!

PR #18459, opened 2025-11-15
[Relax][PyTorch] Fix MultiheadAttention complie

### LINKED ISSUE #18440: [Bug] InternalError: cannot make uint from negative value -inf when compiling model with MultiheadAttention
### Expected behavior

When compiling a PyTorch model containing `nn.MultiheadAttention` with TVM via `torch.export`, an InternalError occurs during the legalize_ops pass. The error indicates that TVM is attempting to convert negative infinity (-inf) to an unsigned integer type, which is mathematically impossible.

The PyTorch model with MultiheadAttention should be successfully compiled by TVM without internal errors, as it runs correctly in native PyTorch.

### Actual behavior

An InternalError occurs during `relax.build` with the message `cannot make uint from negative value -inf`, indicating that TVM's legalization pass encounters a -inf value and attempts to convert it to an unsigned integer type.

### Environment

* **OS:** Ubuntu 20.04.6 LTS
* **TVM version:** 0.23.dev0
*  **Python version:** 3.11.14

### Steps to reproduce

```python
import torch
import torch.nn as nn
import tvm
from tvm import relax

class TestModel(nn.Module):
    def __init__(self):
        super().__init__()
        self.attention = nn.MultiheadAttention(64, 4, dropout=0.1, batch_first=True)

    def forward(self, x, mask):
        attn_output, _ = self.attention(x, x, x, key_padding_mask=mask)
        return attn_output

model = TestModel()
model.eval()

x = torch.randn(8, 16, 64)
mask = torch.randint(0, 2, (8, 16)).bool()

# PyTorch execution works
with torch.no_grad():
    output = model(x, mask)

# PyTorch export works  
exported_program = torch.export.export(model, (x, mask))

# TVM conversion works

# TVM compilation fails
target = tvm.target.Target("llvm")
    ex = relax.build(mod, target)  # InternalError here
```

### Error logs
```
...
tvm.error.InternalError: cannot make uint from negative value -inf
```

### Triage

* needs-triage
* bug

guan404ming:
## Related Issus

closes #18440

## Why

- PyTorch `masked_fill` / `full_like` accept inf or nan and TVM couldn’t handle these values when the tensor dtype was not float, which caused wrong behavior or errors.

## How

- If `fill_value` is inf or nan and the tensor dtype is not float → convert the fill to float32.
- For masked_fill → Create a float values tensor with full_like.
- Cast input to float if needed.
- In TOPI → Reject creating full with inf/nan on non-float dtypes.

### DIFF
--- a/python/tvm/topi/tensor.py
+++ b/python/tvm/topi/tensor.py
@@ -60,13 +60,11 @@ def full(shape, dtype, fill_value):
         The result.
     """
 
-    # Validate that inf/-inf/nan values are only used with float dtypes
-    if isinstance(fill_value, (int, float)) and (_math.isinf(fill_value) or _math.isnan(fill_value)):
+    if isinstance(fill_value, (int, float)) and (
+        _math.isinf(fill_value) or _math.isnan(fill_value)
+    ):
         if not ("float" in dtype or "bfloat16" in dtype):
-            raise ValueError(
-                f"Cannot create tensor with fill_value={fill_value} and dtype={dtype}. "
-                f"Infinite and NaN values require a floating-point dtype."
-            )
+            raise ValueError("Infinite and NaN require a floating-point dtype.")
 
     return cpp.full(shape, dtype, fill_value)
 


guan404ming:
Thanks!

PR #18462, opened 2025-11-16
[Relax][PyTorch] Add decomposed operator support for interpolate

guan404ming:
## Related Issue

- https://github.com/apache/tvm/pull/18401

## How

- Refactored `_index_tensor` to handle broadcast

### DIFF
--- a/tests/python/relax/test_frontend_from_exported_program.py
+++ b/tests/python/relax/test_frontend_from_exported_program.py
@@ -4074,415 +4074,423 @@ def test_interpolate():
             input: R.Tensor((1, 3, 112, 112), dtype="float32")
         ) -> R.Tuple(R.Tensor((1, 3, 224, 224), dtype="float32")):
             with R.dataflow():
-                lv: R.Tensor((224,), dtype="int64") = R.arange(
-                    R.prim_value(0), R.prim_value(224), R.prim_value(1), dtype="int64"
-                )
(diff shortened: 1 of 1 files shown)


guan404ming:
cc @tlopex

tests/python/relax/test_frontend_from_exported_program.py
```diff
@@ -4093,9 +4093,15 @@ def main(
...
-    verify_model(InterpolateNearest(), example_args, {}, expected_nearest)
-    verify_model(InterpolateBicubic(), example_args, {}, expected_bicubic)
+    verify_model(
+        InterpolateBilinear(), example_args, {}, expected_bilinear, run_ep_decomposition=True
+    )
+    verify_model(
+        InterpolateNearest(), example_args, {}, expected_nearest, run_ep_decomposition=True
+    )
+    verify_model(
+        InterpolateBicubic(), example_args, {}, expected_bicubic, run_ep_decomposition=False
```
tlopex: Our goal is to make every test pass when `run_ep_decomposition=True` , so even it has 100+ lines, you'd better rewrite the test

guan404ming:
I've updated the Bicubic tests since I think they're still needed. But we can also remove them if preferred.

tests/python/relax/test_frontend_from_exported_program.py
guan404ming: Thanks for the update.

guan404ming:
Thanks!

PR #18464, opened 2025-11-18
[Relax][PyTorch] Enable decomposition in all tests

guan404ming:
## Why

This is the last part of the migration. After this one, we could set `run_ep_decomposition` default to true in our test

## How

Check and update the remaining tests


guan404ming:
cc @tlopex

guan404ming:
Thanks!

Issue #18344, opened 2025-09-25
[Bug] "Duplicated block name" when invoking tune_tir with tvm.script.tir.Macro


ddwolf:
### Expected behavior

The auto-tuning should complete successfully. TVM should either:
- Automatically generate unique block names for each macro expansion
- Handle duplicate block names gracefully during the tuning process
- Provide a mechanism to specify unique names for macro expansions

### Actual behavior

I am using tvm 0.21.0 tune_tir to tune my operator, which invokes another tvm.script.tir.Macro decorated function many times. The tune tir panicked like this:
```
Duplicated block name  in function main not supported!
```

IMHO, the callsite of invoke Macro will be replaced with the macro's content, so it truly is reasonable that the tvm will encounter duplicate block names if the macro has its own block.

The question 1 is, should we consider all the occurence of the block name defined in macro the same? or tune each of them seperately?

Question 2: is there a good way to bypass this problem now? 

For me, I have to get rid of the block declaration in the macro, therefor allocate memories outside of the macro, which would be a lot convenient if do it inside of it. It's not a good idea I think.

Another way to bypass it is that I transform the script tvm generated, modify each of the block with a unique name. Which requires me to modify the tvm source code itself. And too, this is also not a good idea.

### Environment

- OS: Ubuntu 22.04
- tvm: 0.21.0

### Steps to reproduce

1. Define a `T.Macro` function containing a `T.block` with a specific name
2. Create a `T.prim_func` that calls this macro from multiple locations (call sites)
3. Use `ms.tir_integration.tune_tir` to perform auto-tuning on the function
4. Observe the "Duplicated block name" error during the tuning process

##### this script will reproduce the issue
```python
import tvm
from tvm.script import tir as T
from tvm import meta_schedule as ms
import tempfile

@T.macro
def shared_computation(arr: T.Buffer((32, 32), dtype="float32")):
    """T.Macro function containing a block"""
    with T.block("shared_block"):  # This block name will be duplicated
        arr[0, 0] = 1

@T.prim_func
def macro_bug_demo(a: T.handle, b: T.handle, c: T.handle, d: T.handle) -> None:
    """Function with multiple T.Macro call sites"""
    A = T.match_buffer(a, [32, 32], dtype="float32")
    B = T.match_buffer(b, [32, 32], dtype="float32")
    C = T.match_buffer(c, [32, 32], dtype="float32")
    D = T.match_buffer(d, [32, 32], dtype="float32")

    # First call site
[...]

PR #18465, opened 2025-11-18
[TVMScript] Add block name suffix management for TIR macros

guan404ming:
## Related Issue

closes https://github.com/apache/tvm/issues/18344

## Why

When a `T.macro` containing a block was called multiple times in a TIR function, all expanded blocks had the same name, causing a "Duplicated block name" error in meta_schedule.

## How

Implemented automatic block name suffixing during macro expansion

### DIFF
--- a/python/tvm/script/ir_builder/tir/ir.py
+++ b/python/tvm/script/ir_builder/tir/ir.py
@@ -16,9 +16,11 @@
 # under the License.
 """IRBuilder for TIR"""
 
+import contextlib
 import functools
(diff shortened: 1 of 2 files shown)


guan404ming:
@tlopex @mshr-h

guan404ming:
Nice, thanks

mshr-h:
Would be better to include test.

guan404ming:
Sure, I will follow up with tests. Thanks for your suggestion!

PR #18467, opened 2025-11-18
[Web] Replace string with TVMFFIByteArray* to avoid memory issues

akaashrp:
Passing in a string to `ArrayDecodeStorage` via the packed function definition led to memory issues for larger models (such as `gemma-2-9b-it-q4f32_1-MLC`). Replacing string with TVMFFIByteArray* fixes this issue and also alleviates the stack pollution issue discussed in an earlier PR (https://github.com/apache/tvm/pull/18415). Note that this does not completely fix generation for q0f32 models.


akaashrp:
cc @tqchen

Issue #18468, opened 2025-11-19
[Bug] 'TVMSynchronize' was not declared in this scope


cmpute:
Thanks for participating in the TVM community! We use https://discuss.tvm.ai for any general usage questions and discussions. The issue tracker is used for actionable items such as feature proposals discussion, roadmaps, and bug tracking.  You are always welcomed to post on the forum first :smile_cat:

Issues that are inactive for a period of time may get closed. We adopt this policy so that we won't lose track of actionable issues that may fall at the bottom of the pile. Feel free to reopen a new one if you feel there is an additional problem that needs attention when an old one gets closed.

### Expected behavior

The compilation should succeed with `WITH_CURAND=ON`

### Actual behavior

Compilation failed at here:

```
86.35 /opt/mlc-llm/3rdparty/tvm/src/runtime/contrib/curand/curand.cc:113:3: error: ‘TVMSynchronize’ was not declared in this scope
86.35   113 |   TVMSynchronize(tensor->device.device_type, tensor->device.device_type, nullptr);
86.35       |   ^~~~~~~~~~~~~~
86.51 make[2]: *** [CMakeFiles/tvm_runtime_objs.dir/build.make:863: CMakeFiles/tvm_runtime_objs.dir/src/runtime/contrib/curand/curand.cc.o] Error 1
86.51 make[2]: *** Waiting for unfinished jobs....
```

### Environment

Latest TVM commit, other information should be irrelevant here.

### Steps to reproduce

Just build TVM with `WITH_CURAND=ON`

### Triage

Please refer to the list of label tags [here](https://github.com/apache/tvm/wiki/Issue-Triage-Labels) to find the relevant tags and add them below in a bullet format (example below).

* needs-triage

PR #18469, opened 2025-11-19
[Contrib] Update RandomFill to use StreamSync for CUDA synchronization

guan404ming:
## Related Issue

https://github.com/apache/tvm/issues/18468

## Why

`TVMSynchronize` is a C API function that wasn't included/declared in this file, and it was also called with incorrect arguments

## How

Changed to `cuda_api->StreamSync(tensor->device, nullptr)`


guan404ming:
Thanks!

guan404ming:
Thanks!

PR #18470, opened 2025-11-19
[CI] Enhance python linting scripts to support revision-based checks

guan404ming:
## How

support revision-based checks like `python_format` for `flake8` and `pylint`


guan404ming:
cc @tqchen

PR #18471, opened 2025-11-19
[Relax][PyTorch] Enable run_ep_decomposition by default

guan404ming:
## Why

We have finished the migration for our tests then we could set default to run ep decompose.

## How

Update tests and exported_program_translator.py


guan404ming:
cc @tlopex

tlopex (approved):

python/tvm/relax/frontend/torch/exported_program_translator.py
mshr-h: Please update docstring as well.

guan404ming: Thanks, I've updated.

guan404ming:
Thanks!

tlopex:
Thank you! @mshr-h @guan404ming

PR #18478, opened 2025-11-20
[Relax] Fix the squeeze operator to behave consistently with torch

ConvolutedDog:
This commit fixes the squeeze operator to behave consistently with PyTorch
by implementing no-op behavior when squeezing dimensions that are not of size 1.

Previously:
  squeeze(x, [1]) on tensor with shape [32, 10, 5] would fail

Now:
  squeeze(x, [1]) on tensor with shape [32, 10, 5] returns the original tensor
  without modification, matching PyTorch's behavior

This fixes compatibility issues when converting PyTorch models that use
squeeze with dimensions that may not always be 1 during inference."

This work was done in collaboration with guan404ming's commit d87841d.


ConvolutedDog:
@mshr-h @tlopex Could you please re-run the tests?

tlopex:
@tvm-bot rerun

Issue #17798, opened 2025-04-01
[Bug] VerifyStream::Verify causes dereferencing an invalid pointer


l0r3m1psum:
As the title says. In particular the following statement in `very_well_formed.cc`
```
Verify(it == currently_defined_.end() || redefine_is_allowed)
          << "ValueError: "
          << "TIR is ill-formed, "
          << "due to multiple nested definitions of variable " << var
          << ".  It was first defined at " << it->second << ", and was re-defined at " << path;
```
allows for the espression `it->second` to be evaluated even when `it == currently_defined_.end()`.

This happens to me with MSVC 19.43.34808.

### Expected behavior

`it` should not be dereferenced.

### Actual behavior

`it` is dereferenced.

### Environment

I am running:
  * Windows 11 build 26100.3476
  * Python 3.11.0
  * LLVM 15.0.0
  * TVM commit `2c80e5c7a4d8d8c4125226beb16d792b964f3e5f`

### Steps to reproduce

Run the folowing script

```
import tvm
from tvm import relax
from tvm.relax.frontend import nn
from tvm import dlight as dl

class MLPModel(nn.Module):
	def __init__(self):
		super().__init__()
		self.fc1 = nn.Linear(784, 256)
		self.relu1 = nn.ReLU()
		self.fc2 = nn.Linear(256, 10)
	def forward(self, x):
		x = self.fc1(x)
		x = self.relu1(x)
		x = self.fc2(x)
		return x

mod, param_spec = MLPModel().export_tvm(
	spec={'forward': {'x': nn.spec.Tensor((1, 784), 'float32')}}
)

zero_pipeline = relax.get_pipeline('zero')
mod_op = zero_pipeline(mod)

with tvm.target.Target('cuda'):
	gpu_mod = dl.ApplyDefaultSchedule(
		dl.gpu.Matmul(),
		dl.gpu.Fallback(),
	)(mod_op)

exec = tvm.compile(gpu_mod, target='cuda')
dev = tvm.device('cuda', 0)
vm = relax.VirtualMachine(exec, dev)
```

### Triage

* needs-triage
* tir:analysis

tlopex:
Closed as fixed by @cchung100m

PR #18479, opened 2025-11-20
[TIR]: Fix VerifyStream::Verify causes dereferencing an invalid pointer

cchung100m:
This PR is trying to fix issues https://github.com/apache/tvm/issues/17798. 

### Root Cause

The error message construction used `it->second` unconditionally, but the `Verify()` condition was:
- `it == currently_defined_.end() || redefine_is_allowed`

This means when the condition evaluates to `false` (triggering the error), it could be due to:
1. `it != end() && !redefine_is_allowed` => Safe to access `it->second`
2. `it == end() && !redefine_is_allowed` => Invalid to access `it->second`

### Solution

The fix ensures safe iterator access by:

1. **Storing the Verify result**: Instead of chaining error messages directly to `Verify()`, store the result in a variable
2. **Conditional dereferencing**: Only access `it->second` when `it != end()`
3. **Meaningful error messages**: Provide appropriate messages for both cases

### DIFF
--- a/tests/python/tir-analysis/test_tir_analysis_verify_well_formed.py
+++ b/tests/python/tir-analysis/test_tir_analysis_verify_well_formed.py
@@ -401,5 +401,32 @@ def test_error_message_with_previous_definition_location():
     assert "was re-defined at" in error_msg
 
 
+def test_sequential_redefinition_with_location():
+    """Test case 2b: Sequential redefinition that includes location info
+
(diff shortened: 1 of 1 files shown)


tlopex (approved):
LGTM! Thank you

cchung100m:
Thanks to @tlopex 😄

PR #18480, opened 2025-11-21
[Web] Bump web runtime version 0.23.0-dev1

akaashrp:
Bump web runtime version to 0.23.0-dev1. npm package compiled from https://github.com/apache/tvm/commit/c8515e1ddfaf4d1afff916c484e68e1513631dd6.

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "tvmjs",
-  "version": "0.23.0-dev0",
+  "version": "0.23.0-dev1",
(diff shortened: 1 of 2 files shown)


akaashrp:
cc @tqchen (this is for publishing updated mlc-ai/web-runtime package). This is to avoid overriding the existing dev0 package. Not sure if this is the best approach. We don't have to merge this if it creates complication of version management across backends (e.g. Python, etc.)

tqchen (approved):

PR #18484, opened 2025-11-21
[CI] Update file patterns for specific linting hooks

guan404ming:
## How

- Updat file patterns for specific linting hooks to ensure they only run on relevant file types.

**minor update**

- Fix linting commands to handle interactive and non-interactive modes
- Add 'pre-commit' to default stages in .pre-commit-config.yaml.
- Add inplace fix flags

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,7 +34,7 @@
 default_language_version:
     python: python3.9
 fail_fast: True
-default_stages: [pre-push]
+default_stages: [pre-push, pre-commit]
 repos:
     - repo: https://github.com/pre-commit/pre-commit-hooks
       rev: v6.0.0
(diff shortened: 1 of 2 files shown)


guan404ming:
cc @tqchen

cchung100m (approved):
LGTM, thanks to @guan404ming 😄

guan404ming:
cc @tlopex

tlopex:
Do we really need this update for CI？ cc @tqchen

tqchen:
I am not sure how it would impact ci since right now ci do not run pre-commit hooks, if it is generally helpful for people who use them then i think we can get it in, @guan404ming to confirm

guan404ming:
> I am not sure how it would impact ci since right now ci do not run pre-commit hooks, if it is generally helpful for people who use them then i think we can get it in, @guan404ming to confirm

By updating file patterns to be more specific, hooks only scan relevant file types instead of checking every file. This reduces file processing and speeds up pre-commit execution in my local test.

While this PR currently focuses on the pre-commit, we can also evaluate the performance improvements to see if it's worth adopting in CI as well.

tlopex (approved):

PR #18487, opened 2025-11-22
[Relax][PyTorch] Add `count_include_pad` support to `avg_pool2d` in PyTorch frontend

mshr-h:
As per title.
Note that  `count_include_pad` is True by default on PyTorch. But on Relax, it's False by default.

cc @tlopex


mshr-h:
@tvm-bot rerun

tlopex (approved):

Issue #18438, opened 2025-11-13
[Bug] Segmentation fault when converting PyTorch index assignment operation with torch.export


LiSsHhUuAaIi:
### Description
When converting a PyTorch model containing index assignment operations (`tensor[:, indices] = other_tensor`) using `torch.export`, TVM encounters a segmentation fault during the `from_exported_program` conversion step.

### Expected behavior

The PyTorch model should be successfully converted to TVM Relax module without segmentation faults, and the compiled TVM model should produce the same output as the original PyTorch model.

### Actual behavior

A segmentation fault occurs during the `from_exported_program` call when converting the exported PyTorch model to TVM Relax module. The crash happens in TVM's internal Tuple object creation.

### Environment

* **OS:** Ubuntu 20.04.6 LTS
* **TVM version:** 0.23.dev0
*  **Python version:** 3.11.14

### Steps to reproduce

```python
import torch
import torch.nn as nn
import tvm
from tvm import relax

class TestModel(nn.Module):
    def __init__(self):
        super().__init__()

    def forward(self, indices: torch.Tensor, tensor_1: torch.Tensor,
                tensor_2: torch.Tensor):
        tensor_1[:, indices] = tensor_2
        return tensor_1

model = TestModel()
model.eval()

indices = torch.tensor([0, 2, 4])
tensor_1 = torch.randn(5, 6)
tensor_2 = torch.randn(5, 3)
inputs = (indices, tensor_1, tensor_2)

exported_program = torch.export.export(model, inputs)
from tvm.relax.frontend.torch import from_exported_program
# Segmentation fault occurs here
mod = from_exported_program(exported_program)
```

### Error Log
```
!!!!!!! Segfault encountered !!!!!!!
...
...
Segmentation fault (core dumped)
```

### Triage

* needs-triage

PR #18488, opened 2025-11-22
[Relax][PyTorch] Enhance index_put support for multi-dimensional indices

guan404ming:
## Related Issue

close https://github.com/apache/tvm/issues/18438

## Why

current implementation would be broken when handle multi-dim indices

## How

- support multi-dimensional indices in index_put
- add test case


guan404ming:
cc @tlopex

tlopex (approved):
LGTM

PR #18489, opened 2025-11-22
[Relax][PyTorch] Enhance handling of unbounded upper bound constraints

guan404ming:
## Why 

PyTorch uses int_oo (IntInfinity) for unbounded constraints, which would make our current implemenation crash

## How

- Update the type hint for `create_input_vars` to allow for optional upper bounds.
- Modify the logic to handle unbounded constraints by setting upper bounds to None when applicable.
- Add a new test case


guan404ming:
cc @mshr-h @tlopex

tlopex (approved):
LGTM Thank you!

Issue #17876, opened 2025-04-22
[Bug] TVM crashes with default relax pipeline when opt_level=1: InternalError: Check failed: (slot->value_computed) is false


coffezhou:
### Expected behavior

TVM should compile the model correctly with the default relax optimization pipeline.

### Actual behavior

When compiling the model with the default relax optimization pipeline when opt_level=1, TVM crashes as follows:
```c
Traceback (most recent call last):
  File "/home/carla/Documents/test/test.py", line 63, in <module>
    main()
  File "/home/carla/Documents/test/test.py", line 52, in main
    ex = relax.build(tvm_model, target="llvm", relax_pipeline=relax_pipeline)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/carla/Documents/tvm/python/tvm/relax/vm_build.py", line 253, in build
    mod = relax_pipeline(mod)
          ^^^^^^^^^^^^^^^^^^^
  File "/home/carla/Documents/tvm/python/tvm/ir/transform.py", line 238, in __call__
    return _ffi_transform_api.RunPass(self, mod)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "tvm/_ffi/_cython/./packed_func.pxi", line 339, in tvm._ffi._cy3.core.PackedFuncBase.__call__
  File "tvm/_ffi/_cython/./packed_func.pxi", line 270, in tvm._ffi._cy3.core.FuncCall
  File "tvm/_ffi/_cython/./packed_func.pxi", line 259, in tvm._ffi._cy3.core.FuncCall3
  File "tvm/_ffi/_cython/./base.pxi", line 185, in tvm._ffi._cy3.core.CHECK_CALL
  File "/home/carla/Documents/tvm/python/tvm/_ffi/base.py", line 468, in raise_last_ffi_error
    raise py_err
  File "tvm/_ffi/_cython/./packed_func.pxi", line 56, in tvm._ffi._cy3.core.tvm_callback
  File "/home/carla/Documents/tvm/python/tvm/relax/backend/cpu_generic/pipeline.py", line 73, in _pipeline
    mod = seq(mod)
          ^^^^^^^^
  File "/home/carla/Documents/tvm/python/tvm/ir/transform.py", line 238, in __call__
    return _ffi_transform_api.RunPass(self, mod)
[...]

PR #18491, opened 2025-11-23
[Relax][Backend] Fix TVM crashes with default relax pipeline when opt_level=1: InternalError: Check failed: (slot->value_computed) is false

cchung100m:
Hi Commiters,

This PR is trying to fix issues https://github.com/apache/tvm/issues/17876. Any suggestions would be appreciated if you are available.

### Root Cause

`VMShapeLower` crashed when processing `ShapeExpr` containing composite `PrimExpr` that weren't computed yet.

### Solution

~~Modified `VisitExpr_(const ShapeExprNode* op)` in `vm_shape_lower.cc` to:~~

~~1. Mark uncomputed variables as ready for computation~~
~~2. Trigger EmitOutstandingPrimExprCompute() to resolve the dependency chain~~
~~3. Ensure all expressions are computed before calling `MakeSymbolicShapeArg`~~

~~Added test case: `test_composite_shape_expression_fix()` to prevent future occurrences.~~

symbolizing the composite PrimExpr in the pipeline: canonicalization

### DIFF
--- a/src/relax/backend/vm/vm_shape_lower.cc
+++ b/src/relax/backend/vm/vm_shape_lower.cc
@@ -400,6 +400,23 @@ class VMShapeLowerMutator
       return ffi::GetRef<Expr>(op);
     }
 
+    // Check if all expressions are computed if not mark variables as ready and trigger computation
+    for (const PrimExpr& expr : op->values) {
+      if (!expr->IsInstance<IntImmNode>()) {
+        auto it = slot_map_.find(expr);
+        if (it != slot_map_.end() && !it->second->value_computed) {
+          // If it's a variable, mark it as ready for computation
+          if (expr.as<tir::VarNode>()) {
+            it->second->value_computed = true;
+            ready_vars_.push_back(it->second);
+          }
+        }
+      }
+    }
+
+    // Trigger computation for any expression that are now ready
+    this->EmitOutstandingPrimExprCompute();
+
     ffi::Array<Expr> args = {shape_heap_,
                              PrimValue::Int64(static_cast<int64_t>(op->values.size()))};
     for (PrimExpr expr : op->values) {
--- a/src/relax/backend/vm/vm_shape_lower.cc
+++ b/src/relax/backend/vm/vm_shape_lower.cc
@@ -400,23 +400,6 @@ class VMShapeLowerMutator
       return ffi::GetRef<Expr>(op);
     }
 
-    // Check if all expressions are computed if not mark variables as ready and trigger computation
-    for (const PrimExpr& expr : op->values) {
-      if (!expr->IsInstance<IntImmNode>()) {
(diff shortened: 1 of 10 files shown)


src/relax/backend/vm/vm_shape_lower.cc
```diff
@@ -399,6 +399,23 @@ class VMShapeLowerMutator
...
     }
 
+    // Check if all expressions are computed if not mark variables as ready and trigger computation
+    for (const PrimExpr& expr : op->values) {
+      if (!expr->IsInstance<IntImmNode>()) {
+        auto it = slot_map_.find(expr);
+        if (it != slot_map_.end() && !it->second->value_computed) {
+          // If it's a variable, mark it as ready for computation
+          if (expr.as<tir::VarNode>()) {
+            it->second->value_computed = true;
```
tlopex: Marking arbitrary tir::VarNode instances as `value_computed = true` in `VisitExpr_` is incorrect. In VMShapeLower, only Relax ShapeVars (from function parameters/match_cast) or IntImm constants can be safely marked as computed, because only these have runtime values accessible to the VM.
I think correct way is to fix this earlier in the pipeline (shape inference/canonicalization) by symbolizing composite PrimExpr (introduce Relax ShapeVars) or simplifying them before VMShapeLower.

tlopex (changes requested):

src/relax/backend/vm/vm_shape_lower.cc
cchung100m: Hi @tlopex 

Thanks for the suggestion. 
Should we extend `ComputePrimValue`() to handle ShapeExpr nodes with compound PrimExpr?
https://github.com/apache/tvm/blob/main/src/relax/transform/compute_prim_value.cc#L65

tlopex: Well, `ComputePrimValue` is intended only for evaluating statically evaluable `PrimExpr` into `IntImm` (constant folding), so I think extending ComputePrimValue would not address the root issue.
The real problem is that VMShapeLower cannot consume composite `PrimExpr` directly. The correct solution here should be to canonicalize ShapeExpr earlier by introducing a Relax ShapeVar binding for any non-trivial PrimExpr.

Just like:
```
# 1. Compute the symbolic value first (Canonicalization)
s1 = R.prim_value(n + 1)

# 2. Pass the computed var to the shape (VMShapeLower is happy now)
lv = R.call_tir(cls.func, (x,), R.shape([s1]), dtype="float32")
```

cchung100m:
Hi @tlopex 

Apologies for the delay on this PR. I have just updated the branch with the latest changes and addressed the previous feedback. All CI checks are currently passing. Could you please take another look when you have some time? Let me know if any further adjustments are needed. Thank you so much for your time and patience.

PR #18493, opened 2025-11-23
[Relax][PyTorch] Add broadcast support for `copy` operation

mshr-h:
As per title.
ref: [torch.Tensor.copy_ — PyTorch 2.9 documentation](https://docs.pytorch.org/docs/stable/generated/torch.Tensor.copy_.html)


mshr-h:
cc @tlopex

PR #18497, opened 2025-11-24
[Relax][PyTorch] Add binary operation dtype promotion following PyTorch rules in ExportedProgram frontend

mshr-h:
As per title.
ref: https://docs.pytorch.org/docs/stable/generated/torch.promote_types.html


mshr-h:
cc @tlopex

tlopex (approved):

PR #18498, opened 2025-11-24
[CI] Use glob for `conda/build-environment.yaml` in cache key

mshr-h:
The previous cache key used `conda/build-environment.yaml` and caused macOS CI errors.
Switching to `**/conda/build-environment.yaml` to see if it works.


mshr-h:
Let's ee if it solves.
cc @tlopex

PR #18503, opened 2025-11-25
[Relax][PyTorch] Add extra_buffers support for exported program

### LINKED ISSUE #18357: [bug][relax.frontend.torch] from_exported_program KeyError for non-persistent buffer bert.embeddings.position_ids (HuggingFace BERT via torch.export)
### Behavior

Importing a torch.exported HuggingFace BERT (bert-base-multilingual-uncased) with TVM Relax Torch frontend fails with:
```
KeyError: 'bert.embeddings.position_ids'
```
The HF position_ids is a non-persistent buffer (registered with persistent=False) and therefore is omitted from PyTorch’s state_dict. The Relax importer currently assumes it must exist in exported_program.state_dict[spec.target], which raises a KeyError.

```
torch.export: OK
  ...
  File ".../tvm/relax/frontend/torch/exported_program_translator.py", line 620, in create_input_vars
    torch_shape = exported_program.state_dict[spec.target].shape
KeyError: 'bert.embeddings.position_ids'
```

### Environment

- OS: (Ubuntu 22.04.4 LTS (x86_64))
- TVM version: (release v0.21.0)
- Python: (3.10.16)
- LLVM: (17.0.6)
- Pytorch: (2.8.0)

### Steps to reproduce

```python
import torch, torch.nn as nn
from transformers import AutoModel

class M(nn.Module):
    def __init__(self):
        super().__init__()
        self.bert = AutoModel.from_pretrained("bert-base-multilingual-uncased")
        self.cls  = nn.Linear(self.bert.config.hidden_size, 2)

    def forward(self, x, mask=None):
        out = self.bert(x, attention_mask=mask).last_hidden_state[:, 0, :]
        return self.cls(out)

def main():
    torch.manual_seed(0)
    m = M().eval()
    x = torch.randint(0, 30522, (2, 16))
    mask = torch.ones_like(x)

    from torch.export import export as torch_export
    ep = torch_export(m, (x, mask))
    print("torch.export: OK")

    mod = from_exported_program(ep)  # <-- raises KeyError

if __name__ == "__main__":
    main()
```

### Triage

* needs-triage
* bug

abeeha123:
Fixes #18357 — KeyError in the Relax PyTorch frontend when loading an HuggingFace BERT model exported with torch. export.
The issue occurs because non-persistent buffers, such as position_ids and token_type_ids, are not included in the state_dict, leading to missing-buffer errors.

This PR adds support for extra non-persistent buffers by introducing an extra_buffers lookup and extending buffer-resolution logic to handle shapes and dtypes correctly.

With this fix, HuggingFace BERT models exported via torch. export no longer fails with missing-buffer KeyErrors.

### DIFF
--- a/python/tvm/relax/frontend/torch/exported_program_translator.py
+++ b/python/tvm/relax/frontend/torch/exported_program_translator.py
@@ -1442,7 +1442,8 @@ class ExportedProgramImporter(BaseFXGraphImporter):
                     if node.name == name_hint and "tensor_meta" in node.meta:
                         torch_shape = node.meta["tensor_meta"].shape
                         torch_dtype = node.meta["tensor_meta"].dtype
-        #fmt:off                break
+                        break
+            #fmt:off
             else:
                 # PARAMETER or BUFFER
                 info=merged_state.get(target.spec)


mshr-h:
Make sense to me. Can you also add testcase for that in https://github.com/apache/tvm/blob/main/tests/python/relax/test_frontend_from_exported_program.py ?

mshr-h:
Closing as #18527 got merged

PR #18504, opened 2025-11-25
[TVMScript] Add test for TIR macro block name suffix handling

guan404ming:
## How

add missing tests for https://github.com/apache/tvm/pull/18465


guan404ming:
cc @tlopex

PR #18506, opened 2025-11-26
[Relax][PyTorch] Enhance data type handling in FX graph translator

guan404ming:
## Why

The current codebase lack of lots of pytorch dtype support

## How

- add supprot for those dtype and update the tests


guan404ming:
cc @mshr-h @tlopex

PR #18507, opened 2025-11-26
[Relax][PyTorch] Support specifying decimals for _round

guan404ming:
## Why

- The current `round` function does not support specifying the number of decimal places.

## How

- Allows rounding to a specified number of decimals
- Add tests for `_round`


guan404ming:
cc @tlopex

Issue #17388, opened 2024-09-19
[Bug] [tir] tir.LowerIntrin check failed additional_info.size() == new_size (34 vs. 33)


talha-ahsan:
Apologies if this isn't a real bug, but I'm getting an InternalError when trying to compile (see code sample below). As far as I can tell, I should be getting some more meaningful diagnostic information from the compiler if the issue is a result of the TIR I passed in. 

### Steps to Reproduce: 

```python
import tvm
from tvm import tir
from tvm.tir.analysis.analysis import verify_well_formed, verify_memory
from tvm.ir.instrument import PrintBeforeAll, PrintAfterAll

from tvm.script import tir as T

@T.prim_func
def tvmgen_default_fused_add_24(p0: T.Buffer((1, 96, 17, 17), "float32"), p1: T.Buffer((96, 1, 1), "float32"), T_add: T.Buffer((1, 96, 17, 17), "float32")):
    T.func_attr({"from_legacy_te_schedule": T.bool(True), "hash": "17a6861adc5229a4", "target": T.target({"host": {"keys": ["cpu"], "kind": "llvm", "tag": ""}, "keys": ["cpu"], "kind": "llvm", "tag": ""}), "tir.noalias": T.bool(True)})
    for ax0_ax1_fused in T.parallel(96):
        for ax2, ax3_outer, ax3_inner_s in T.grid(17, 2, 16):
            t173 = T.float32()
[...]

PR #18514, opened 2025-11-26
[TIR] Fix tir.LowerIntrin check failed additional_info.size() == new_size

cchung100m:
Hi Reviewers,

This PR is trying to fix issues https://github.com/apache/tvm/issues/17388. Any suggestions would be appreciated if you are available.

### Root Cause:
The recovery functions assumed the vector size wouldn't change during the context's lifetime, but nested contexts would modify it. Each constraint context modifies the same `additional_info_` vector in `ConstIntBoundAnalyzer`.

### Solution:

- Removed the strict `ICHECK_EQ(additional_info_.size(), new_size)` assertion
- Modified the recovery function to simply resize back to the original size when the vector has grown


cchung100m:
Thanks to @tlopex 😄

PR #18418, opened 2025-11-04
[TIR][Schedule] Add FuseReductionEpilogue primitive to fuse epilogue …

kimm240:
Currently it is not possible to fuse an epilogue operation (e.g., bias addition) into a reduction block's initialization statement. This limitation prevents leveraging hardware-specific instructions that support bias accumulation in vector ISAs, such as MACC (multiply-accumulate with bias) instructions.

This commit implements a new schedule primitive 'fuse_reduction_epilogue' that addresses the problem described in:
https://discuss.tvm.apache.org/t/tir-problem-inlining-addition-into-matmul-block/18066

The primitive transforms the following pattern:

  Before:
    for i, j, k in T.grid(M, N, K):
        with T.block("matmul"):
            with T.init():
                temp[vi, vj] = 0
            temp[vi, vj] = temp[vi, vj] + A[vi, vk] * B[vj, vk]

    for i, j in T.grid(M, N):
        with T.block("bias_add"):
            D[vi, vj] = temp[vi, vj] + C[vi, vj]

  After:
    for i, j, k in T.grid(M, N, K):
        with T.block("matmul"):
            T.reads(C[vi, vj], A[vi, vk], B[vj, vk])
            T.writes(D[vi, vj])
            with T.init():
                D[vi, vj] = C[vi, vj]  # Fused epilogue into init
            D[vi, vj] = D[vi, vj] + A[vi, vk] * B[vj, vk]

The transformation removes the intermediate temp buffer and the separate epilogue block, enabling better tensorization opportunities for hardware with bias accumulation support.

Implementation:
- ReductionEpilogueFuser class for pattern validation and IR transformation
  - BodyPatternAllowFusion: Validates epilogue can be fused
  - AnalyzeEpiloguePattern: Detects addition pattern (D = temp + C)
  - ExtractEpilogueInfo: Extracts buffer and region information
  - CreateFusedReductionBlock: Creates single block with modified T.init()
- SingleBlockFusionReplacer: Replaces blocks and removes temp buffer
- Variable mapping between epilogue and reduction block iter vars
- Proper buffer and region updates with correct read/write ordering
- FFI bindings and Python API following TVM conventions

Changes:
- src/tir/schedule/primitive/compute_inline.cc: Core implementation (~430 lines)
- src/tir/schedule/primitive.h: Function declaration
- include/tvm/tir/schedule/schedule.h: Virtual method in ScheduleNode
[...]

PR #18515, opened 2025-11-27
[TIR][Schedule] FuseReductionEpilogue: Add Clipping pattern support

kimm240:
Currently, the FuseReductionEpilogue primitive only supports Bias
(addition) and BiasReLU (addition + ReLU) epilogue patterns. However,
clipping operations (min(max(x, lower), upper)) are commonly used in
deep learning models and would benefit from the same fusion optimization.

This commit extends FuseReductionEpilogue to support Clipping patterns
by:

1. Adding EpilogueType::Clipping to the enum to distinguish clipping
   patterns from other epilogue types.

2. Adding clipping_lower_ and clipping_upper_ members to
   ReductionEpilogueFuser to store clipping bounds extracted from the
   epilogue pattern.

3. Extending AnalyzeEpiloguePattern to detect clipping patterns:
   - min(max(temp, lower), upper)
   - max(min(temp, upper), lower)
   - All commutative variants of min/max at each level

4. Updating BiasReLU pattern matching to handle max(0, x) form in
   addition to max(x, 0) for better commutativity support.

5. Modifying CreateFusedReductionBlock to apply clipping to the init
   value: init = min(max(0, lower), upper)

6. Updating BufferReplacer to apply clipping per-iteration:
   value = min(max(value, lower), upper)

7. Adding validation in BodyPatternAllowFusion to ensure temp appears
   exactly once in clipping patterns.

8. Creating comprehensive test coverage with 8 test cases:
   - Basic fusion test
   - Numerical correctness verification
   - Multiple epilogue blocks test
   - 5 commutative variant tests

This implementation follows the same per-iteration semantics as BiasReLU,
where clipping is applied at each reduction step rather than
post-reduction. This semantic change is documented in the docstring with
a warning about potential numerical differences.

The test suite verifies that all commutative forms of clipping patterns
are correctly recognized and that the fused implementation produces
numerically identical results to the per-iteration reference
implementation.

### DIFF
--- a/src/tir/schedule/primitive/compute_inline.cc
+++ b/src/tir/schedule/primitive/compute_inline.cc
@@ -1090,12 +1090,13 @@ bool ReductionEpilogueFuser::BodyPatternAllowFusion(const BlockRealize& epilogue
     return false;
   }
 
-  // 5. For Clipping pattern, verify temp appears exactly once
-  if (epilogue_type_ == EpilogueType::Clipping) {
-    if (loads.size() != 1) {
-      // Failure: temp must appear exactly once in clipping pattern
-      return false;
-    }
+  // 5. Verify temp appears exactly once in the epilogue pattern
+  // This ensures correctness for all supported patterns (Bias, BiasReLU, Clipping)
+  // The reduction result buffer must be used exactly once in the epilogue expression
+  if (loads.size() != 1) {
+    // Failure: The reduction result (temp) must be used exactly once in the
+    // epilogue expression for fusion.
+    return false;
   }
 
   // 6. Check if producer is a reduction block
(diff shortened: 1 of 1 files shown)


kimm240:
@wrongtest-intellif
This PR implements the extension for other epilogue forms (ReLU/Clipping) as discussed in the previous PR #18418 review conversation.

wrongtest-intellif:
> @wrongtest-intellif This PR implements the extension for other epilogue forms (ReLU/Clipping) as discussed in the previous PR #18418 review conversation.

Thanks for the contribution! The change looks good to me itself. My question is now do we have any more insights for general epilogue forms?  For `reverse_compute_inline` primitive, it seems many other relu/prelu/sigmoid etc forms are available without explicit pattern matching.

kimm240:
What you say is definitely good direction! In `compute_inline.cc`,  usual data flow reconnection method is used to handle complex operations using Index mapping with DetectInterMap and substitution logic after BufferLoad extraction. It would be a good reference to this project's future. I'd like to ask for advice.

wrongtest-intellif:
> What you say is definitely good direction! In `compute_inline.cc`, usual data flow reconnection method is used to handle complex operations using Index mapping with DetectInterMap and substitution logic after BufferLoad extraction. It would be a good reference to this project's future. I'd like to ask for advice.

Would you like make similar implementation instead of directly introduce clip and relu pattern?

kimm240:
I completely agree. However, implementing a full, general Data Flow Reconnection logic requires significant development time and extensive testing.

Given that the current PR is functionally complete and stable, and it addresses two very common patterns (Clipping and ReLU), could we proceed with merging the current PR first? This would allow the community to gain the immediate performance benefit.

I commit to delivering the generalized implementation as the next dedicated PR. And, I am also happy to discuss the generalized logic with you before implementing the code.

tlopex (approved):
It looks very good to me! Thanks for the contribution @kimm240

tlopex:
Let's merge this PR first

PR #18517, opened 2025-11-27
[Relax][PyTorch] Unify tests using shared verify_model

guan404ming:
## Why

We have the shared verify func in tests and to use it in every tests could help persist consistency


guan404ming:
cc @tlopex

PR #18518, opened 2025-11-27
[Relax][PyTorch] Fix batch normalization training mode correctness

guan404ming:
## Why

Batch normalization in training mode would threw away updated statistics.

## How

batch_norm(...) → keep all 3 elements, pad to 5 for PyTorch


guan404ming:
cc @mshr-h @tlopex

PR #18521, opened 2025-11-28
[Relax][PyTorch] Handle unknown output shapes for _sym_size_int

guan404ming:
## Why 

Handle case where shape is unknown (None). This can happen for operations with dynamic output shapes.

## How

- Update to handle with placeholder
- Update test case


guan404ming:
cc @mshr-h @tlopex

mshr-h:
Isn’t the test passing successfully even without making any changes? @guan404ming 
It's passing on my env with `torch==2.5.0`.

guan404ming:
> Isn’t the test passing successfully even without making any changes? @guan404ming
It's passing on my env with torch==2.5.0.

It also passes on my environment (torch 2.9.0). I removed that redundant test.
The check is kind of defensive programming. It prevents a future `NoneType` error if later adds support for data-dependent output shapes (e.g., `nonzero`) and someone calls `.shape[dim]` on such results.

mshr-h:
Ok

PR #18523, opened 2025-11-28
[ADRENO][TEXTURE] Texture based lowering

srkreddy1238:
Introduces the below features over texture annotation

- Lowering, codegen and runtime for texture.
- image2d_array_t support - Added depth dimension allows more allocations using texture instead of falling back to buffer when the texture limits exceeds.
- A comprehensive set of schedules for Adreno textures.
- Texture packing of arbitrary types up to 128 bit (FP16-NCHW8c, INT8-NCHW16c ...etc.).
- A clBufferDescriptor debug dump controlled by cmake options.
- Pipeline definition for adreno target.

While covering these features the below interfaces or passes or enhanced which need a review.

- alloc_tensor: VDevice information is passed across these API's. The way of texture allocation is ```alloc_storage``` allocates buffer/image objects as requested followed by alloc_tensor being a view of any scope. This takes care of optimum utilization backing memory across different image objects or scopes.
- Constants Saving: Handled by adding memory scope section in executable. This introduces a new header magic to retain the backward compatibility.
- Static Memory Planing: Mostly port from Relay static memory planner with mixed mode allocator.


include/tvm/runtime/tensor.h
```diff
@@ -188,14 +189,25 @@ class Tensor : public tvm::ffi::Tensor {
...
+
+  TVM_DLL void SetScope(ffi::String scope);
+  TVM_DLL ffi::String GetScope() const;
+
+ protected:
+  /*!
+   * \brief The memory scope
+   * represents the underlaying scope information of device
+   */
+  ffi::String scope = "global";
```
tqchen: i think it is somewhat very intrusive to introduce scope to runtime tensor class here, is it possible to avoid it in runtime? but instead ensure compiler allocate the right scope and call right ops implicitly?

Note that per runtime, the scope can also implicit attaches to data field

tqchen (changes requested):

include/tvm/runtime/tensor.h
srkreddy1238: This is primarily for the consts / params load/store. The scope information should go at the time of loading consts (runtime::Tensor) to avoid copies.

I think, it can be achieved by altering Load/Store ConstSection of Executable. Let me try this way...

srkreddy1238:
@tqchen  managed with out altering ``runtime::Tensor``

VM byte code now adds MemoryScope section. 

To keep backward compatibility of ``Load``, I added new header magic ``kTVMVMBytecodeMagicV2`` . Let me know if this need to be handled via ``VM_VERSION``.

elvin-n:
Q not related directly to this PR, sorry. Can TVM and Adreno flow be compiled/executed on Snapdragon X1? Windows/Linux?

srkreddy1238:
> Q not related directly to this PR, sorry. Can TVM and Adreno flow be compiled/executed on Snapdragon X1? Windows/Linux?

Yes. It works on Snapdragon laptops (X Elite) with both arm64 (Arm64 need Qualcomm Arm64 OpenCL SDK) and x64 (Generic OpenCL works w/o CLML enabled) compilations.

Sample config

```
           cp ../cmake/config.cmake .
           Add-Content config.cmake "set(USE_OPENCL $ENV:OPENCL_SDK_ADRENO_X86)"
           Add-Content config.cmake "set(USE_LLVM $ENV:LLVM_CONFIG)"
           Add-Content config.cmake "set(USE_CLML $ENV:OPENCL_SDK_ADRENO_X86)"
           Add-Content config.cmake "set(USE_RPC ON)"
           Add-Content config.cmake "set(USE_CPP_RPC ON)"
           Add-Content config.cmake "set(USE_KALLOC_ALIGNMENT 32)"
           Add-Content config.cmake "set(USE_OPENCL_EXTN_QCOM ON)"
           cmake .. -G "Visual Studio 17 2022" -A x64
           cmake --build . --config Release --parallel $env:NUMBER_OF_PROCESSORS
```

```
           cd build-arm64
           cp ../cmake/config.cmake .
           Add-Content config.cmake "set(USE_OPENCL $ENV:OPENCL_SDK_ADRENO_ARM64)"
           Add-Content config.cmake "set(USE_CLML $ENV:OPENCL_SDK_ADRENO_ARM64)"
           Add-Content config.cmake "set(USE_RPC ON)"
           Add-Content config.cmake "set(USE_CPP_RPC ON)"
           Add-Content config.cmake "set(USE_KALLOC_ALIGNMENT 32)"
           Add-Content config.cmake "set(USE_OPENCL_EXTN_QCOM ON)"
           cmake .. -G "Visual Studio 17 2022" -A ARM64
           cmake --build . --config Release --parallel $env:NUMBER_OF_PROCESSORS --target tvm_rpc
```

mlc-llm too works.

src/runtime/opencl/opencl_module.cc
```diff
@@ -50,6 +50,89 @@ class OpenCLWrappedFunc {
     arg_size_ = arg_size;
     launch_param_config_.Init(arg_size.size(), launch_param_tags);
   }
+
+#ifdef PROFILE_SHADER_DUMP
```
tqchen: this is understandable, however, would be useful to think about if we can move this logic to outside of the core code-base, e.g. if such tracing can be done in the driver API, or we can have a ffi function hook that user registers and runs

src/tir/backend/adreno/texture_flatten.cc
```diff
@@ -0,0 +1,228 @@
...
+#include "../../runtime/thread_storage_scope.h"
+
+namespace tvm {
+namespace tir {
+using arith::IRVisitorWithAnalyzer;
+using runtime::ApplyTexture2DFlattening;
+using runtime::DefaultTextureLayoutSeparator;
+using runtime::IsTextureStorage;
+
+class TextureLoweringBase : public StmtExprMutator {
```
tqchen, line 48: Given this and the other transformations are related to texture and is not needed in None texture case,  might be useful to move it to src/tir/backend/adreno, is it possible to have a adreno specific pipeline for this like relax one? that way it gives clearity and also gives freedom of evolutions for future changes

tqchen:
Thanks @srkreddy1238 , i think we are close to merging, just did another round of quick look

srkreddy1238:
Removed the profile dump for now and moved Adreno specific tir transforms to new folder along with Adreno specific tir pipeline.

PR #18524, opened 2025-11-29
[Relax][PyTorch] Fix InternalError when converting scaled_dot_product_attention with 2D inputs

Dayuxiaoshui:
Fixes #18441

Previously, the TVM frontend incorrectly assumed 4D input dimensions for scaled_dot_product_attention, causing an InternalError when the actual input was 2D (seq_len, head_dim).

This fix:
- Detects input dimensionality (2D vs 4D)
- For 2D inputs: expands to 4D, calls attention, then squeezes back
- For 4D inputs: maintains existing behavior
- Adds test case for 2D input scenario
- Updates verify_model_numerically to use strict=False for export

PR #18530, opened 2025-11-29
[Relax][PyTorch] Add boolean tensor support for max operation and corresponding test case

mshr-h:
As per title.
ref: https://github.com/apache/tvm/pull/18524#discussion_r2573023355

### DIFF
--- a/tests/python/relax/test_frontend_from_exported_program.py
+++ b/tests/python/relax/test_frontend_from_exported_program.py
@@ -1429,6 +1429,45 @@ def test_binary1(op, relax_op):
     verify_model(Binary2(op), example_args2, {}, expected2)
 
 
+operator_binary_scalar = [
+    (torch.ops.aten.add.Scalar, R.add),
+    (torch.ops.aten.bitwise_and.Scalar, R.bitwise_and),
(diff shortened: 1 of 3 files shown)


tests/python/relax/test_frontend_from_exported_program.py
```diff
@@ -4239,7 +4142,7 @@ def main(
         ),
         {},
         Expected1,
-        run_ep_decomposition=True,
+        run_ep_decomposition=False,
```
tlopex, line 4184: Just want to know why should `run_ep_decomposition` be disabled here?
Is the intention that Relax already supports this composite op, so decomposition is unnecessary?

mshr-h:
cc @tlopex

tlopex (approved):

tlopex:
Please resolve the conflicts so that we can merge it in

tests/python/relax/test_frontend_from_exported_program.py
mshr-h: > Is the intention that Relax already supports this composite op, so decomposition is unnecessary?

Exactly. I'm thinking we should provide a way for users to control the decomposition. Or we have to define list of ops that we want not to be decomposed.

PR #18535, opened 2025-12-01
[Relax][PyTorch] Add support for masked_select

guan404ming:
## How

Add support for masked_select


guan404ming:
cc @tlopex @mshr-h

python/tvm/relax/frontend/torch/exported_program_translator.py
```diff
@@ -1477,6 +1483,11 @@ def create_convert_map(
             "item.default": self._item,
             "sym_size.int": self._sym_size_int,
             "_local_scalar_dense.default": self._item,
+            # symbolic shape constraints (no-ops for compilation)
+            "sym_constrain_range_for_size.default": lambda node: self.env[node.args[0]],
+            "_assert_scalar.default": lambda node: self.env[node.args[0]],
+            "ge": self._symbolic_comparison,
+            "le": self._symbolic_comparison,
```
tlopex, line 1490: What's the purpose of those no-ops?

tlopex: cc @guan404ming

tlopex (approved):

python/tvm/relax/frontend/torch/exported_program_translator.py
guan404ming: Hi @tlopex, I just noticed that the review was written but it seems like not submitted (by clicking the green submit button), so it stayed in pending state wasn’t visible to others. I realized this recently as well, just thought I’d let you know. Thanks!

guan404ming: > What's the purpose of those no-ops?

Thanks for the question, those would be used after the decompose.

guan404ming:
Thanks for the review!

python/tvm/relax/frontend/torch/exported_program_translator.py
tlopex: @guan404ming Thanks for letting me know. I was confused as well

Issue #17936, opened 2025-05-09
[Bug] InternalError: Check failed: (eval_vec_) is false


Cookiee235:
### Actual behavior

```
Traceback (most recent call last):
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/qshenaf/remote_pc/TirFuzz/bugs/05-03_20-50/topi.nn.group_norm_4.py", line 19,
    tvm.build(mod, target='llvm')
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/qshenaf/envs/tvm/python/tvm/driver/build_module.py", line 59, in build
    return tvm.tir.build(mod, target, pipeline)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/qshenaf/envs/tvm/python/tvm/tir/build.py", line 173, in build
    mod = pipeline(mod)
          ^^^^^^^^^^^^^
  File "/data/qshenaf/envs/tvm/python/tvm/ir/transform.py", line 238, in __call__
    return _ffi_transform_api.RunPass(self, mod)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "tvm/_ffi/_cython/./packed_func.pxi", line 339, in tvm._ffi._cy3.core.PackedFuncBase.__call__
  File "tvm/_ffi/_cython/./packed_func.pxi", line 270, in tvm._ffi._cy3.core.FuncCall
  File "tvm/_ffi/_cython/./packed_func.pxi", line 259, in tvm._ffi._cy3.core.FuncCall3
  File "tvm/_ffi/_cython/./base.pxi", line 185, in tvm._ffi._cy3.core.CHECK_CALL
  File "/data/qshenaf/envs/tvm/python/tvm/_ffi/base.py", line 468, in raise_last_ffi_error
    raise py_err
  File "tvm/_ffi/_cython/./packed_func.pxi", line 56, in tvm._ffi._cy3.core.tvm_callback
  File "/data/qshenaf/envs/tvm/python/tvm/tir/pipeline.py", line 122, in _pipeline
    mod = tvm.ir.transform.Sequential(passes)(mod)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #18536, opened 2025-12-01
[ARITH] Fix InternalError: Check failed: (eval_vec_) is false

cchung100m:
Hi Commiters,

This PR is trying to fix issues https://github.com/apache/tvm/issues/17936. Any suggestions would be appreciated if you are available.

### Root Cause

Code paths that expected vector evaluation but encounter the scalar-only evaluation `eval_vec_ = false`

### Solution

`BroadcastNode` just replicates the same scalar value and the evaluation might not requires special vector-aware handling

### DIFF
--- a/src/arith/int_set.cc
+++ b/src/arith/int_set.cc
@@ -532,7 +532,10 @@ class IntervalSetEvaluator : public ExprFunctor<IntervalSet(const PrimExpr&)> {
   }
 
   IntervalSet VisitExpr_(const BroadcastNode* op) final {
-    ICHECK(eval_vec_);
+    if (!eval_vec_) {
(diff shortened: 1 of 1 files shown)


tqchen:
would be good to dig into what is going on in the original case, since the old API states that `eval_vec_` is an intentionally flag for evaluating vectors, and the CHECK explicitly checks the invariance, turning this off for the one off case may leads to unintentional behavior. cc @tlopex , ideally we should keep the invariance when possible

cchung100m:
Hi @tqchen 

Thanks for review comments and point out the crucial part and potential risk. I wasn't thinking straight to resolve the issue. Should we revert this change?

tqchen:
i think reverting for now is safer

tlopex:
@tqchen Thanks for pointing that. I'll send a pr to revert it now.

PR #18537, opened 2025-12-01
[CI] Update cpplint script to support revision-based linting

guan404ming:
## What 

- update cpplint script to support revision-based linting


guan404ming:
cc @tlopex @tqchen

PR #18544, opened 2025-12-04
[Relax][PyTorch] Add support for Custom Ops for ExportedProgram frontend

mshr-h:
As per title.

cc @tlopex @guan404ming 

We keep the interface same as [`from_fx()`](https://github.com/apache/tvm/blob/ed97234b25a155bc66198ab5cd9e372a4772acec/python/tvm/relax/frontend/torch/fx_translator.py#L1152) so you can define and pass custom converter something like this.

```python
from tvm.relax.frontend.torch.exported_program_translator import ExportedProgramImporter
def _rms_norm_converter(node: torch.fx.Node, self: ExportedProgramImporter) -> relax.Var:
    x = self.env[node.args[0]]
    torch_dtype = node.args[0].meta["tensor_meta"].dtype
    normalized_shape = node.args[1]
    weight = self.env.get(node.args[2], None) if len(node.args) > 2 else None
    eps = node.args[3] if len(node.args) > 3 else None

    N = len(self.shape_of(x))
    D = len(normalized_shape) if isinstance(normalized_shape, (tuple, list)) else 1
    axes = list(range(N - D, N))

    if weight is None:
        weight = self._convert_torch_tensor_to_relax(
            torch.ones(list(normalized_shape), dtype=torch_dtype)
        )
    eps = torch.finfo(torch_dtype).eps if eps is None else 0.00001

    return self.block_builder.emit(relax.op.nn.rms_norm(x, weight, axes, eps))

mod = from_exported_program(
    exported_program,
    custom_convert_map={"rms_norm.default": _rms_norm_converter},
    run_ep_decomposition=False,
)

### DIFF
--- a/tests/python/relax/test_frontend_from_exported_program.py
+++ b/tests/python/relax/test_frontend_from_exported_program.py
@@ -42,6 +42,7 @@ def verify_model(
     unwrap_unit_return_tuple=False,
     no_bind_return_tuple=False,
     map_free_vars=False,
+    custom_convert_map=None,
 ):
     exported_program = export(torch_model, args=example_args, dynamic_shapes=dynamic_shapes)
     mod = from_exported_program(
@@ -50,6 +51,7 @@ def verify_model(
         keep_params_as_input=keep_params_as_input,
         unwrap_unit_return_tuple=unwrap_unit_return_tuple,
         no_bind_return_tuple=no_bind_return_tuple,
+        custom_convert_map=custom_convert_map,
     )
 
     binding = {k: tvm.runtime.tensor(v) for k, v in binding.items()}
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: tests/python/relax/test_frontend_from_exported_program.py [hunk 1]
```diff
@@ -6525,6 +6527,40 @@ def forward(self, x):
...
+
+    from tvm.relax.frontend.torch.exported_program_translator import (
+        ExportedProgramImporter,
+    )
+
+    def custom_add_converter(node: torch.fx.Node, self: ExportedProgramImporter) -> relax.Var:
+        x = self.env[node.args[0]]
+        y = self.env[node.args[1]]
+
+        return self.block_builder.emit(R.subtract(x, y))
```
tlopex, line 6556: I didn't catch here. Is this intended to be an addition or a subtraction? cc @mshr-h

mshr-h: I wanted to test that the importer can override the default import behavior. By default, the importer converts `add.Tensor` into `relax.add`, but in the custom converter it is converted into `relax.subtract`. @tlopex

tlopex (approved):

Issue #18481, opened 2025-11-21
[Docs] The end-to-end optimization tutorial doesn't work on the latest code


cmpute:
Thanks for participating in the TVM community! We use https://discuss.tvm.ai for any general usage questions and discussions. The issue tracker is used for actionable items such as feature proposals discussion, roadmaps, and bug tracking.  You are always welcomed to post on the forum first :smile_cat:

Issues that are inactive for a period of time may get closed. We adopt this policy so that we won't lose track of actionable issues that may fall at the bottom of the pile. Feel free to reopen a new one if you feel there is an additional problem that needs attention when an old one gets closed.

### Documentation Title & Type

[End-to-End Optimize Model](https://tvm.apache.org/docs/how_to/tutorials/e2e_opt_model.html)

### Additions/Changes Requested

When runs the tutorial with the latest code, there are exceptions:

```log
# Metadata omitted. Use show_meta=True in script() method to show it.

Traceback (most recent call last):
  File "//workspace/test_tvm2.py", line 72, in <module>
    ex = tvm.compile(mod, target="cuda")
  File "/opt/mlc-llm/3rdparty/tvm/python/tvm/driver/build_module.py", line 104, in compile
    return tvm.relax.build(
  File "/opt/mlc-llm/3rdparty/tvm/python/tvm/relax/vm_build.py", line 263, in build
    return _vmlink(
  File "/opt/mlc-llm/3rdparty/tvm/python/tvm/relax/vm_build.py", line 158, in _vmlink
    lib = tvm.tir.build(tir_mod, target=target, pipeline=tir_pipeline)
  File "/opt/mlc-llm/3rdparty/tvm/python/tvm/tir/build.py", line 226, in build
    mod = pipeline(mod)
  File "/opt/mlc-llm/3rdparty/tvm/python/tvm/ir/transform.py", line 167, in __call__
    return _ffi_transform_api.RunPass(self, mod)
  File "python/tvm_ffi/cython/function.pxi", line 678, in core.Function.__call__
  File "<unknown>", line 0, in tvm::transform::Pass::operator()(tvm::IRModule) const
  File "<unknown>", line 0, in tvm::transform::Pass::operator()(tvm::IRModule, tvm::transform::PassContext const&) const
  File "<unknown>", line 0, in tvm::transform::ModulePassNode::operator()(tvm::IRModule, tvm::transform::PassContext const&) const
[...]

PR #18545, opened 2025-12-04
[Docs] Improve static shape tuning parameter configuration (follow-up to commit c71aefc)

ConvolutedDog:
- Expose max_trials_per_task parameter to static_shape_tuning_pipeline
- Adjust default TOTAL_TRIALS from 8000 to 80 for tutorial demonstration purposes
- Add documentation for tuning parameters in tutorial, clarifying relationship between MAX_TRIALS_PER_TASK and TOTAL_TRIALS

### DIFF
--- a/docs/how_to/tutorials/e2e_opt_model.py
+++ b/docs/how_to/tutorials/e2e_opt_model.py
@@ -104,16 +104,14 @@ if not IS_IN_CI:
 #   to at least ``320 (20 tasks * 16 trials)`` ensures every task receives one full iteration
 #   of tuning. We set it to 512 in our configuration to allow for several more iterations,
 #   aiming to explore a wider parameter space and potentially achieve better performance.
-# - If ``MAX_TRIALS_PER_TASK == None``, the system defaults to ``min(max_trials_per_iter=64,
-#   TOTAL_TRIALS)`` trials per task per iteration. This may lead to undersubscribed tuning when
-#   ``TOTAL_TRIALS`` is insufficient (e.g., ``64 < TOTAL_TRIALS < 20 * 64``), potentially skipping
-#   some tasks entirely, leaving critical operators unoptimized or missing thread binding for
-#   untuned tasks. Explicitly setting both parameters avoids this issue and provides deterministic
-#   resource allocation across all tasks.
+# - If ``MAX_TRIALS_PER_TASK == None``, the system defaults to ``TOTAL_TRIALS`` trials per
+#   task per iteration. An insufficient ``TOTAL_TRIALS`` setting may lead to undersubscribed
+#   tuning, potentially skipping some tasks entirely. Explicitly setting both parameters
+#   avoids this issue and provides deterministic resource allocation across all tasks.
 #
 # Note: These parameter settings are optimized for quick tutorial demonstration. For production
-# deployments requiring higher performance, we recommend adjusting both MAX_TRIALS_PER_TASK
-# and TOTAL_TRIALS to larger values. This allows more extensive search space exploration
+# deployments requiring higher performance, we recommend adjusting both ``MAX_TRIALS_PER_TASK``
+# and ``TOTAL_TRIALS`` to larger values. This allows more extensive search space exploration
 # and typically yields better performance outcomes.
 
 TOTAL_TRIALS = 512  # Change to 20000 for better performance if needed
--- a/python/tvm/relax/pipeline.py
+++ b/python/tvm/relax/pipeline.py
@@ -131,8 +131,10 @@ def static_shape_tuning_pipeline(
 
     max_trials_per_task : Optional[int]
         The maximum number of trials to run per task.
-        If not specified, MetaSchedule will use a default value of 64
-        trials per task during the tuning process.
+        If not specified, it defaults to the value of `total_trials`, and this
+        may lead to undersubscribed tuning, potentially skipping some tasks
+        entirely. Explicitly setting both parameters avoids this issue and
+        provides deterministic resource allocation across all tasks.
         For optimal tuning, set `total_trials` to at least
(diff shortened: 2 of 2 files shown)


ConvolutedDog:
@tlopex Please review this follow-up PR addressing feedback from https://github.com/apache/tvm/issues/18481 and https://github.com/apache/tvm/pull/18159 regarding tuning parameter configuration improvements.

docs/how_to/tutorials/e2e_opt_model.py
```diff
@@ -95,13 +95,38 @@
...
+#   untuned tasks. Explicitly setting both parameters avoids this issue and provides deterministic
+#   resource allocation across all tasks.
+#
+# Note: These parameter settings are optimized for quick tutorial demonstration. For production
+# deployments requiring higher performance, we recommend adjusting both MAX_TRIALS_PER_TASK
+# and TOTAL_TRIALS to larger values. This allows more extensive search space exploration
+# and typically yields better performance outcomes.
 
-TOTAL_TRIALS = 8000  # Change to 20000 for better performance if needed
+TOTAL_TRIALS = 80  # Change to 20000 for better performance if needed
```
cbalint13: IMHO 80 is way too few, let it be 512.
Agree that 8000 is much for a quick testing/docs purpose.

docs/how_to/tutorials/e2e_opt_model.py
```diff
@@ -95,13 +95,38 @@
...
+MAX_TRIALS_PER_TASK = 4  # Change to more trials per task for better performance if needed
```
cbalint13: IMHO 8 or 16 is better, 16 pairs well with proposed total 512.

cbalint13:
@tvm-bot rerun

python/tvm/relax/pipeline.py
```diff
@@ -128,6 +129,14 @@ def static_shape_tuning_pipeline(
     cpu_weight_prepack : bool
         Whether to enable the cpu weight prepack feature.
 
+    max_trials_per_task : Optional[int]
+        The maximum number of trials to run per task.
+        If not specified, MetaSchedule will use a default value of 64
+        trials per task during the tuning process.
```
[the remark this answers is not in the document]
mshr-h: Gemini review makes sense to me.

ConvolutedDog:
@cbalint13 @mshr-h Could you please take another look at the latest commit? Thanks~

tlopex (approved):
It looks great to me! Thank you

docs/how_to/tutorials/e2e_opt_model.py
```diff
@@ -95,13 +95,38 @@
 # leverage MetaSchedule to tune the model and store the tuning logs to the database. We also
 # apply the database to the model to get the best performance.
 #
+# The ResNet18 model will be divided into 20 independent tuning tasks during compilation.
+# To ensure each task receives adequate tuning resources in one iteration while providing
+# early feedback:
+#
+# - To quickly observe tuning progress, each task is allocated a maximum of 16 trials per
+#   iteration (controlled by ``MAX_TRIALS_PER_TASK=16``). We should set ``TOTAL_TRIALS``
+#   to at least ``320 (20 tasks * 16 trials)`` ensures every task receives one full iteration
```
cbalint13, line 104: * I am not sure that a demanding and also imperative statement like "We should"  would have it's place here.
* I would remove "We should set TOTAL_TRIALS to at least ``320 (20 tasks * 16 trials)`` ensures every task receives one full iteration  of tuning.",  I believe it can mislead on the the way metaschedule really works internally. The main goal is to allow the metaschedule explorer to unfold (and does not matter much if a parameter cover or not the other by division) as much of its exploration space as it it is possible.

Something like "We set it here to a lower value of 512 for a quick demonstration purpose"  or similar is fair enough. There is already the " # Change to 20000 for better performance if needed" which also hints the user how to really get good results outside of scope of this "quick" tutorial.

cbalint13 (changes requested):

docs/how_to/tutorials/e2e_opt_model.py
ConvolutedDog: > * I am not sure that a demanding and also imperative statement like "We should"  would have it's place here.
> * I would remove "We should set TOTAL_TRIALS to at least `320 (20 tasks * 16 trials)` ensures every task receives one full iteration  of tuning.",  I believe it can mislead on the the way metaschedule really works internally. The main goal is to allow the metaschedule explorer to unfold (and does not matter much if a parameter cover or not the other by division) as much of its exploration space as it it is possible.
> 
> Something like "We set it here to a lower value of 512 for a quick demonstration purpose" or similar is fair enough. There is already the " # Change to 20000 for better performance if needed" which also hints the user how to really get good results outside of scope of this "quick" tutorial.

With all due respect, I believe the guideline (at least 320=20x16 trials) provides essential practical value. If users don't allocate sufficient `TOTAL_TRIALS`, the last few TIR funcs won't receive any tuning at all—they'll simply be skipped.

Simply saying "512 for demonstration purposes" would render the exposed `MAX_TRIALS_PER_TASK` parameter meaningless. This concrete calculation helps users understand the minimum required tuning budget and establish proper baselines, which is far more helpful for practical usage than abstract concepts about exploration space.

cbalint13: @ConvolutedDog ,

I might not a good tutorialist, i just wanted to avoid hard imperatives s/"We should {...}"/"We set {...}"/.
As for 320 / 16 indeed there will be full coverage of trials (some will fail anyway, but task is not abandoned) per tasks.

LGTM, let it be.

cbalint13 (approved):

cbalint13:
@ConvolutedDog 

Thank you for the enhancement !

Issue #18423, opened 2025-11-07
[Bug] MetaSchedule tune_tir crashes with ScheduleError in RewriteFuseSplitParallelVectorize


Cookiee235:
Running meta_schedule.tune_tir on a valid TIR module triggers a crash inside
RewriteFuseSplitParallelVectorize -> Parallel with a ScheduleError.

This seems to happen during initial population generation (SampleInitPopulation) in the evolutionary search, indicating an internal schedule transform failure 

### Actual behavior

```
2025-11-07 15:56:26 [INFO] [task_scheduler.cc:189] TaskScheduler picks Task #0: "main"
Traceback (most recent call last):
  File "/share_container/LLMFuzz/TirFuzz/bugs/10-24_20-21/topi.reverse_sequence_1_M0.py", line 33, in <module>
    database = ms.tir_integration.tune_tir(mod=tir_mod, target='llvm --num-cores=32', work_dir='./tune_tmp', max_trials_global=1, num_trials_per_iter=1)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/software/tvm-latest/python/tvm/meta_schedule/tir_integration.py", line 146, in tune_tir
    return tune_tasks(
           ^^^^^^^^^^^
  File "/software/tvm-latest/python/tvm/meta_schedule/tune.py", line 122, in tune_tasks
    task_scheduler.tune(
  File "/software/tvm-latest/python/tvm/meta_schedule/task_scheduler/task_scheduler.py", line 132, in tune
    _ffi_api.TaskSchedulerTune(  # type: ignore # pylint: disable=no-member
  File "python/tvm_ffi/cython/function.pxi", line 758, in core.Function.__call__
[...]

PR #18547, opened 2025-12-05
[MetaSchedule] Fix tune_tir crash with ScheduleError in RewriteParallelVectorizeUnroll

Dayuxiaoshui:
This patch fixes issue #18423 where meta_schedule.tune_tir crashes during initial population sampling when RewriteParallelVectorizeUnroll postprocessor encounters blocks that violate compact dataflow requirements.

The crash occurred when:
- A block reads and writes to the same buffer
- RewriteParallelVectorizeUnroll tries to parallelize/vectorize these loops
- ScheduleError is thrown and propagates through parallel_for_dynamic

Solution:
- Added exception handling in RewriteParallelVectorizeUnroll::Apply to catch ScheduleError and skip problematic blocks instead of crashing
- Added exception handling in ThreadedTraceApply::Apply to catch exceptions from postprocessors and treat them as normal failures

This makes the tuning process more robust by gracefully handling schedule errors instead of crashing the entire tuning session.

### DIFF
--- a/src/meta_schedule/postproc/rewrite_parallel_vectorize_unroll.cc
+++ b/src/meta_schedule/postproc/rewrite_parallel_vectorize_unroll.cc
@@ -448,7 +448,7 @@ class RewriteParallelVectorizeUnrollNode : public PostprocNode {
           }
         } catch (const tir::ScheduleError& e) {
           DLOG(WARNING) << "Failed to apply parallelization/vectorization: " << e.what();
-          continue;
+          return false;
         }
       }
     }


src/meta_schedule/postproc/rewrite_parallel_vectorize_unroll.cc
```diff
@@ -425,25 +425,30 @@ class RewriteParallelVectorizeUnrollNode : public PostprocNode {
...
           }
-        }
-        // AutoUnroll
-        if (parsed.unroll_explicit != -1 || parsed.unroll_implicit != -1) {
-          ICHECK(parsed.unroll_explicit == -1 || parsed.unroll_implicit == -1);
-          int unroll_explicit = parsed.unroll_explicit != -1;
-          int max_step = parsed.unroll_explicit + parsed.unroll_implicit + 1;
-          tir::RewriteUnroll(sch, unroll_explicit, max_step, block_rv, loop_rvs[0]);
+        } catch (const tir::ScheduleError& e) {
+          DLOG(WARNING) << "Failed to apply parallelization/vectorization: " << e.what();
```
tlopex, line 450: Here I think catching the error and continuing may leave the schedule partially mutated. It would be safer to return false on ScheduleError and let the candidate be discarded, unless rollback semantics are guaranteed. cc @tqchen

Dayuxiaoshui:
@tlopex

Dayuxiaoshui:
@tlopex

tlopex (approved):
Sorry for late. LGTM Thank you!

Dayuxiaoshui:
@tlopex It's okay, and I also wish you a Merry Christmas. Thank you for your hard work.

PR #18548, opened 2025-12-05
[Relax][PyTroch] Add NHWC layout support

guan404ming:
## Why

- The interpolate operation was hardcoded to only support NCHW layout
- Users need flexibility to choose the appropriate layout for their target platform

## How

- Added default_image_layout parameter
- Exposed default_image_layout parameter in the public from_fx()


guan404ming:
cc @tlopex @mshr-h

guan404ming:
Thanks!

PR #18550, opened 2025-12-05
[Relax][PyTorch] Enhance scale_factor handling in interpolation

guan404ming:
## Why

Fixes interpolation to support different scaling factors for height and width (e.g., scale_factor=[2.0, 3.0])

## How

- Removed the bug: Stopped extracting just the first element ([0]) from scale_factor lists
- Passed full value: Now passes the entire scale_factor (scalar or list) to the underlying implementation, which already handles both correctly

### DIFF
--- a/tests/python/relax/test_frontend_from_exported_program.py
+++ b/tests/python/relax/test_frontend_from_exported_program.py
@@ -8542,5 +8542,56 @@ def test_grid_sample():
     verify_model(GridSample(), example_args, {}, expected)
 
 
+def test_upsample_nearest2d():
+    class UpsampleNearest2dScale(Module):
+        def forward(self, input):
+            return torch.nn.functional.interpolate(input, scale_factor=2.0, mode="nearest")
+
+    class UpsampleNearest2dSize(Module):
+        def forward(self, input):
+            return torch.nn.functional.interpolate(input, size=(20, 20), mode="nearest")
+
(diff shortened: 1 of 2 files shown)


guan404ming:
cc. @tlopex @mshr-h

tests/python/relax/test_frontend_from_exported_program.py
```diff
@@ -4900,6 +4900,30 @@ def main(
...
+            kwargs = {"scale_factor": self.scale_factor, "mode": self.mode}
+            # align_corners is only used for bilinear and bicubic modes
+            if self.mode in ["bilinear", "bicubic"]:
+                kwargs["align_corners"] = False
+            return torch.nn.functional.interpolate(input, **kwargs)
+
+    # Test with 32x32 input
+    example_args = (torch.randn(1, 3, 32, 32, dtype=torch.float32),)
+    model = Interpolate(mode, scale_factor)
+    verify_model_numerically(model, example_args, rtol=1e-5, atol=1e-5)
```
mshr-h: Would be better to avoid E2E tests as much as possible because they increased CI pressure.

guan404ming: Sure, I've updated with unit tests only version. Thanks!

mshr-h (approved):

guan404ming:
Thanks!

PR #18551, opened 2025-12-05
[Relax][Frontend] Introduce ModuleDict

mshr-h:
As per title.
Just like [ModuleDict in PyTorch](https://docs.pytorch.org/docs/stable/generated/torch.nn.ModuleDict.html).

### DIFF
--- a/python/tvm/relax/frontend/nn/visitor.py
+++ b/python/tvm/relax/frontend/nn/visitor.py
@@ -79,6 +79,24 @@ class Mutator:
         """
         return self.visit(name, node)
 
+    def visit_moduledict(self, name: str, node: nn.ModuleDict) -> Any:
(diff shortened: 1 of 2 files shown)


python/tvm/relax/frontend/nn/core.py
```diff
@@ -540,6 +540,56 @@ def _compile(spec, device, pipeline, debug):
         raise ValueError(f"Unknown out_format: {out_format}")
 
 
+class ModuleDict(Module):
+    """Holds submodules in a dict."""
```
cbalint13, line 544: @mshr-h 

*nit*
Just a tiny little bit more words here about what "submodules" are. 
Noticed that all classes here have a bit more explanative comment on them.

cbalint13 (approved):
LGTM !

PR #18559, opened 2025-12-09
[Relax] Add mod operator support

guan404ming:
## How

- Resolve todo by changing from raising error to calling _op_ffi_api.mod
- Add both operators to the parametrized test


mshr-h (approved):

guan404ming:
Thanks!

PR #18577, opened 2025-12-10
[Relax] Clean up scatter_elements unknown dtype handling

guan404ming:
## Why

- LOG(WARNING) is the standard and correct approach throughout the TVM codebase
- The existing pattern is used consistently in all relax ops (see test_op_manipulate.py, index.cc, etc.)
- Added test coverage for previously untested scenarios


guan404ming:
cc @tlopex @mshr-h

PR #18578, opened 2025-12-10
[Relax] Fix unknown dtype bool8 in RoPE handling

simpolism:
I was getting "error: unknown dtype `bool8`" when invoking tvm via `mlc_llm serve`: tir wasn't happy with python's default bool8 type, so I swapped it to use tir's internal AND op handling via `tir.all`.

### DIFF
--- a/python/tvm/relax/frontend/nn/llm/position_embedding.py
+++ b/python/tvm/relax/frontend/nn/llm/position_embedding.py
@@ -530,13 +530,13 @@ def llama_rope_with_position_map(  # pylint: disable=too-many-arguments
                 s, h, d = T.axis.remap("SSS", iters)
                 if h < num_q_heads:
                     q[s, h, d] = T.if_then_else(
-                        apply_rope > 0 and d < rotary_dim,
+                        tir.all(apply_rope > 0, d < rotary_dim),
                         _rope(qkv, s, h, d, position_map[s]),
                         qkv[s, h, d],
                     )
                 elif h < num_q_heads + num_kv_heads:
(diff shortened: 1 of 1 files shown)


grf53:
I also met a similar issue. But...

Considering that the code before fix was written a year ago and that it worked fine with the latest nightly release about a month, I suspect this fix may not be a fundamental solution.

While a fix like yours seems to address some of the issue, it would require significant modifications to apply everywhere. Therefore, reverting the functionality of `and` and `or` seems like a more appropriate solution.

I couldn't find the problem point, but I am doubting the tvm-ffi update for now.

simpolism:
@grf53 thanks for the feedback, I'll close this PR for now given it doesn't address the regression's root cause.

grf53:
@giterator @MasterJH5574 (I mentioned you because you seem to be related to this part.)

Hello, I'd like to confirm your awareness of this issue.

This issue is considered a critical one, as it blocks compilation of LLM models with the latest version of TVM, but it appears to have remained unfixed for a week or more.

I'm unable to build the lib for the LLM model using tvm and mlc-llm (`mlc-llm-nightly-cu130`, `mlc-ai-nightly-cu130` or sth else.) installed with the latest whl files from https://mlc.ai/wheels, due to the same issue mentioned in this PR.

I don't know if this will help: it seems to be happening within the last two months or so, as it seems to be working fine based on a version committed two months ago.

MasterJH5574:
Hi @grf53, we have identified the issue and fixed it. It was because Python package dependency version requirement that leads to installing an old apache-tvm-ffi version by accident.  It should now work if you pip install the latest package.

PR #18579, opened 2025-12-11
[Relax] Add layout inference support for repeat operator

guan404ming:
## How

- Implemented InferLayoutRepeat function that:
  - Preserves layout when axis is specified (with axis transformation)
  - Returns 1D layout when axis is not specified (flatten mode)
  - Transforms the axis parameter based on layout changes (e.g., NCHW axis=1 → NHWC axis=3)


guan404ming:
cc @tlopex @mshr-h

tlopex (approved):
LGTM Thanks

PR #18593, opened 2025-12-20
[Relax]feat: Implement FRelaxInferLayout for tile operator

Dayuxiaoshui:
- Implement InferLayoutTile function to handle layout transformation for tile operator
- Use TransposeStrLike approach similar to repeat operator to correctly map repeats array
- Handle both same-dimension and dimension-expansion cases
- Add test case test_conv2d_tile to verify layout conversion from NCHW to NHWC
- Fixes the TODO at src/relax/op/tensor/manipulate.cc:1932

The implementation correctly transforms the repeats array when the input tensor's layout changes (e.g., from NCHW to NHWC), ensuring that repeat values are mapped to the correct dimensions in the new layout.


Dayuxiaoshui:
@tlopex

PR #18613, opened 2025-12-26
[Relax] Remove duplicated test case: test_if_branch_var_scope

cchung100m:
Hi Commiters,

This PR is trying to remove duplicated test case `test_if_branch_var_scope`. Any suggestions would be appreciated if you are available.

### Root Cause

Both tests `test_var_if_scoping_fail` and `test_if_branch_var_scope`  verify that returning a variable `w` that's only defined within if/else branches raises an error. They're duplicates and should probably be consolidated into a single test.

image

### Solution
Remove duplicated test case `test_if_branch_var_scope`

### DIFF
--- a/tests/python/relax/test_tvmscript_parser.py
+++ b/tests/python/relax/test_tvmscript_parser.py
@@ -1289,20 +1289,6 @@ def test_var_if_scoping_fail():
             return w  # error: The w is not defined in the outer scope
 
 
-def test_if_branch_var_scope():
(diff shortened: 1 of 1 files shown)


cchung100m:
Hi @tlopex @mshr-h

Any suggestions would be appreciated if you are available. 😄

cchung100m:
Thanks to @mshr-h and happy new year! 😄

PR #18615, opened 2025-12-26
[Relax] Add gpu-generic fallback for unrecognized GPU targets

guan404ming:
## Why

- Resolve todo in `python/tvm/relax/pipeline.py`

## How

- Check specific GPU targets first (cuda/rocm/metal/adreno)
- If not matched, use BackendDispatcher.is_gpu_target() to detect GPU
- Fall back to gpu_generic for unrecognized GPUs (vulkan, webgpu, etc.)
- Raise error only for non-GPU unsupported targets

### DIFF
--- a/python/tvm/relax/pipeline.py
+++ b/python/tvm/relax/pipeline.py
@@ -27,6 +27,7 @@ import tvm
 from tvm import meta_schedule as ms
 
 from . import backend, transform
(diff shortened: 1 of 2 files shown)


guan404ming:
cc @tlopex @mshr-h

guan404ming:
Thanks! Happy new year @mshr-h  🥳

PR #18618, opened 2025-12-27
[Relax] Add test case for op attributes in AST printer

guan404ming:
## Why

Resolve todo in `tests/python/relax/test_ast_printer.py`

## How

- Added a new test function `test_op_attrs` in `tests/python/relax/test_ast_printer.py`.
- Removed the TODO comment.


guan404ming:
cc @tlopex @mshr-h

PR #18619, opened 2025-12-27
[Relax][PyTorch] Fix PyTorch Dynamo frontend for Darwin compatibility

guan404ming:
## Why

The llvm_target() function reads `/proc/cpuinfo` which only exists on `Linux`, causing tests to fail on `macOS` with FileNotFoundError.

## How

- Add cross-platform CPU feature detection in llvm_target() using platform.system() and sysctl for macOS
- Update tests


guan404ming:
cc @tlopex @mshr-h

mshr-h (approved):

PR #18575, opened 2025-12-10
[Relax][ONNX] Add edge padding mode

guan404ming:
## How

- resolving todo by add edge padding mode in onnx
- add tests for this change

PR #18620, opened 2025-12-27
[Relax] Fix flaky test_conv2d gradient numeric test

guan404ming:
## Why

Flaky in CI due to numerical precision errors in gradient checks.
ref: https://github.com/apache/tvm/pull/18575

## How

- Reduce input range from (0, 3) to (0, 2) to minimize floating-point errors


guan404ming:
cc @tlopex @mshr-h

mshr-h (approved):

PR #18621, opened 2025-12-29
[Relax] Use weight shape instead of dim in Embedding.forward

guan404ming:
## Why

The Embedding module's forward method was using `self.dim` redundantly when reshaping the output, even though this dimension is already available from `self.weight.shape[1]`.

## How

- Changed `shape=[*x.shape, self.dim]` to `shape=[*x.shape, self.weight.shape[1]]` in `Embedding.forward`


guan404ming:
cc @tlopex @mshr-h

PR #18624, opened 2025-12-29
[CUDA] Fix cuModuleUnload crash during interpreter shutdown

guan404ming:
## Related

#18614 ci error

## Why

The CUDAModuleNode destructor was using CUDA_DRIVER_CALL and CUDA_CALL macros that call LOG(FATAL) (throw an exception) when CUDA operations fail.

During interpreter shutdown, the CUDA context can become invalid, causing CUDA_ERROR_ILLEGAL_ADDRESS when cuModuleUnload is called. Throwing exceptions in destructors is undefined behavior and causes crashes.

## How
1. Removed the throwing macros from the destructor
2. Check cudaSetDevice return value and skip cleanup if it fails
3. Ignore errors from cuModuleUnload - during shutdown these are benign since the OS will reclaim resources anyway


guan404ming:
cc @tlopex @mshr-h

PR #18625, opened 2025-12-29
[Python][Relax] Fix YaRN correction dim calculation

akaashrp:
Replace expressions with TIR constructs

### DIFF
--- a/python/tvm/relax/frontend/onnx/onnx_frontend.py
+++ b/python/tvm/relax/frontend/onnx/onnx_frontend.py
@@ -1127,7 +1127,31 @@ class PRelu(OnnxOpConverter):
     def _impl_v1(cls, bb, inputs, attr, params):
         x = inputs[0]
         slope = inputs[1]
-        return relax.op.nn.prelu(x, slope)
+
+        x_shape = x.struct_info.shape
+        slope_shape = slope.struct_info.shape
+
+        ndim = len(x_shape)
+        s_ndim = len(slope_shape)
+
+        if all(ss == 1 for ss in slope_shape) or s_ndim == 1:
+            slope = relax.op.reshape(slope, (slope_shape[0],))
+            return relax.op.nn.prelu(x, slope, ndim - 1)
+
+        if s_ndim == ndim:
+            non_one_axes = [i for i, ss in enumerate(slope_shape) if ss != 1]
+
+            # Must have only ONE non-broadcast axis
+            if len(non_one_axes) != 1:
+                raise ValueError(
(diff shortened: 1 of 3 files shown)


Liberatedwinner:
I noticed an issue with this change. 
When using Metal, I encountered the following error:
`tvm.error.InternalError: Fail to compile metal source:program_source:20:61: error: call to 'log' is ambiguous`
Could you take a look?

akaashrp:
@Liberatedwinner are there specific models or commands for which you run into this issue?

Liberatedwinner:
Please run the code file I attached.
[test_tir_log.py](https://github.com/user-attachments/files/24392363/test_tir_log.py)

The result of test code is:
```
============================================================
PR version (tir.log)
============================================================
low: T.max(T.float32(128.0) * T.log(T.float32(20.371832715762604)) / (T.float32(2.0) * T.log(T.float32(10000.0))), T.float32(0.0))
high: T.min(T.float32(128.0) * T.log(T.float32(651.89864690440334)) / (T.float32(2.0) * T.log(T.float32(10000.0))), T.float32(127.0))

============================================================
Original version (math.log)
============================================================
low: T.float32(20.944481620636051)
high: T.float32(45.026881273754547)
```

Using `tir.log` instead of `math.log` in `yarn_find_correction_dim` causes Metal codegen to fail with `call to 'log' is ambiguous`.

`tir.log` creates a symbolic expression (e.g., `T.log(T.float32(20.3718...))`) that remains in the IR and gets compiled into Metal shader code. 
Metal Shading Language has multiple overloads for math functions (float, half, etc.), and the compiler cannot resolve which one to use.
See: [Metal Shading Language Specification](https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf) (Section 6.5 "Math Functions" and Table 6.4)

akaashrp:
I'm able to see that the outputs look different for the PR version and the original version. However, I'm also able to compile models that use YaRN without issue on metal. Could you point me to a codegen path I could run so that I can arrive at the ```call to 'log' is ambiguous``` error you mentioned?

Liberatedwinner:
Would you mind checking the error with this file?
[test_tir_log.py](https://github.com/user-attachments/files/24393885/test_tir_log.py)

akaashrp:
@Liberatedwinner, I apologize for the delay. To deal with the merge conflicts, I created a new PR that should address the issues on Metal by precomputing the denominator: https://github.com/apache/tvm/pull/18661. 

FWIW, I ran into a Metal error (as expected) when initially testing with your file but am getting an unrelated MLC-LLM pipeline error now (both before and after changes).

PR #18629, opened 2025-12-31
[Relax] Add FInferMixedPrecision and FRelaxInferLayout for conv transpose ops

guan404ming:
## Why

The `conv1d_transpose` and `conv2d_transpose` operators were missing FInferMixedPrecision and FRelaxInferLayout attribute implementations, which are needed for:

- Mixed precision training/inference support (e.g., float16 inputs with float32 outputs)
- Layout transformation optimizations during compilation
- Consistency with conv1d and conv2d operators which already have these attributes

## How

- Implemented InferLayoutConv1dTranspose and InferMixedPrecisionConv1dTranspose
- Implemented InferLayoutConv2dTranspose and InferMixedPrecisionConv2dTranspose


guan404ming:
cc @tlopex @mshr-h

Issue #17999, opened 2025-05-20
[Bug] HardSigmoid returns 1.0 for NaN input (ONNX model mismatch with ORT)


Mohamad11Dab:
### Expected behavior

HardSigmoid(NaN) should return NaN

### Actual behavior

`––––– MISMATCH DETECTED –––––

Not equal to tolerance rtol=0.01, atol=0.001

x and y nan location mismatch:
 x: array([[[[     nan, 0.481528, 0.363083, ...,      nan, 0.508347,
          0.159798],
         [0.325104, 0.460197, 0.420308, ...,      nan,      nan,...
 y: array([[[[1.      , 0.481528, 0.363083, ..., 1.      , 0.508347,
          0.159798],
         [0.325104, 0.460197, 0.420308, ..., 1.      , 1.      ,...`

### Environment

TVM:0.17.0
ONNXRuntime: 1.16.3

### Steps to reproduce

```python
import sys, os
sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__), '..')))
import torch
import torch.nn as nn
import torch.nn.functional as F
import tempfile
import onnx
import onnxruntime as ort
from numpy.testing import assert_allclose
import tvm
from tvm import relay
from tvm.contrib import graph_executor
import numpy as np
import nas_model_2

class SimpleBugModel(nn.Module):
    def __init__(self):
        super().__init__()
        self.input_conv = torch.nn.modules.conv.Conv2d(in_channels=3, out_channels=16, kernel_size=1)
        self.block2 = nas_model_2.LogWrapper()
        self.block9 = nas_model_2.HardSigmoidWrapper()

    def forward(self, x):
        __input_conv = self.input_conv(x)
        __blocks__2 = self.block2(__input_conv)
        __blocks__9 = self.block9(__blocks__2)
        return __blocks__9

def main():
    model = SimpleBugModel()
    model.eval()
    dummy = torch.randn(1, 3, 32, 32, dtype=torch.float32)
    with tempfile.NamedTemporaryFile(suffix='.onnx', delete=False) as tmp:
        onnx_path = tmp.name
    torch.onnx.export(model, dummy, onnx_path, opset_version=19, input_names=['input'], output_names=['output'])

    ort_sess = ort.InferenceSession(onnx_path, providers=['CPUExecutionProvider'])
    ort_out = ort_sess.run(None, {'input': dummy.numpy()})[0]

    onnx_model = onnx.load(onnx_path)

[...]

PR #18631, opened 2026-01-02
[Relax] Fix HardSigmoid returns 1.0 for NaN input

cchung100m:
Hi Commiters,

This PR is trying to fix issues https://github.com/apache/tvm/issues/17999. Any suggestions would be appreciated if you are available.

### Root Cause
The implementation applied clipping/comparisons without preserving NaN, which caused NaN inputs to be coerced to the clipped upper bound instead of remaining NaN.

### Solutions:
- Update the HardSigmoid implementation to explicitly preserve NaN inputs. The implementation now computes the usual linear expression and clipped result, but returns NaN when input is NaN.
- Add a unit test that verifies: input = NaN -> output = NaN

### DIFF
--- a/python/tvm/relax/frontend/onnx/onnx_frontend.py
+++ b/python/tvm/relax/frontend/onnx/onnx_frontend.py
@@ -3350,7 +3350,13 @@ class HardSigmoid(OnnxOpConverter):
         alpha = relax.const(alpha, dtype=dtype)
         beta = float(attr.get("beta", 0.5))
         beta = relax.const(beta, dtype=dtype)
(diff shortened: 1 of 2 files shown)


cchung100m:
Hi @tlopex @mshr-h

Any suggestions would be appreciated if you are available. 😄

mshr-h:
Since the behavior for NaN inputs is not well defined, both tvm and onnxruntime are correct according to the specification.

cchung100m:
Hi @mshr-h 

Thanks for the reply. Could we keep the issue/PR open and mark it as known question for now?

guan404ming (dismissed):
Thanks @cchung100m

cchung100m:
Thanks to @guan404ming 😸

PR #18632, opened 2026-01-03
[Relax] Move GetUsedVars to analysis module

guan404ming:
## Why

The GetUsedVars function was defined locally in binding_rewrite.cc with a TODO comment suggesting it should be moved to the analysis module. This refactoring improves code organization by placing the utility function alongside other variable analysis functions.

## How
- Move GetUsedVars implementation to analysis module
- Add FFI registration and Python wrapper
- Add parametrized test

### DIFF
--- a/python/tvm/relax/analysis/analysis.py
+++ b/python/tvm/relax/analysis/analysis.py
@@ -312,9 +312,9 @@ def all_vars(expr: Expr) -> List[Var]:
     return _ffi_api.all_vars(expr)
 
 
-def get_used_vars(expr: Expr) -> List[Var]:
+def used_vars(expr: Expr) -> List[Var]:
     """
-    Get all variables used in an expression.
+    Return all variables used in an expression.
 
     This function collects all variable references within the given expression,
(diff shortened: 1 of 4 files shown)


guan404ming:
cc @mshr-h @tlopex

python/tvm/relax/analysis/analysis.py
```diff
@@ -312,6 +312,26 @@ def all_vars(expr: Expr) -> List[Var]:
     return _ffi_api.all_vars(expr)
 
 
+def get_used_vars(expr: Expr) -> List[Var]:
```
mshr-h: nit: the func name might be better to be `used_vars()` to be consistent with `all_vars()` and `free_vars()`? cc @tlopex @cbalint13

tlopex: I agree @guan404ming

guan404ming: Sure, I've updated.

guan404ming:
Thanks for all reviews!

PR #18633, opened 2026-01-04
[Relax] Add FRelaxInferLayout and TMixedPrecisionPolicy for dynamic_strided_slice

guan404ming:
## Why

The dynamic_strided_slice operator was missing FRelaxInferLayout and TMixedPrecisionPolicy attributes, preventing it from participating in layout transformations and mixed precision optimizations.

## How
- Add `TMixedPrecisionPolicy` attribute with `kFollow` policy and 
- Add `InferLayoutDynStridedSlice` function that falls back to initial layout (since begin/end/strides are dynamic tensors that cannot be transformed at compile time)

### DIFF
--- a/tests/python/relax/test_transform_to_mixed_precision.py
+++ b/tests/python/relax/test_transform_to_mixed_precision.py
@@ -1064,5 +1064,58 @@ def test_call_tir_with_float16_args():
     tvm.ir.assert_structural_equal(Expected, After)
 
 
+def test_dynamic_strided_slice():
+    @I.ir_module
+    class Input:
+        @R.function
+        def main(
+            x: R.Tensor((2, 3, 28, 28), "float32"),
+            w: R.Tensor((4, 3, 3, 3), "float32"),
+            begin: R.Tensor((4,), "int64"),
+            end: R.Tensor((4,), "int64"),
+            strides: R.Tensor((4,), "int64"),
+        ) -> R.Tensor(None, "float32", ndim=4):
+            with R.dataflow():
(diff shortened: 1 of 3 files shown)


guan404ming:
cc @tlopex @mshr-h

tests/python/relax/test_transform_to_mixed_precision.py
```diff
@@ -1064,5 +1064,58 @@ def tir_identity(
...
+                )
+                lv3: R.Tensor((2, 4, 26, 26), dtype="float16") = R.astype(lv2, dtype="float16")
+                lv4: R.Tensor((2, 4, 26, 26), dtype="float32") = R.astype(lv3, dtype="float32")
+                gv: R.Tensor(None, dtype="float32", ndim=4) = R.dynamic_strided_slice(
+                    lv4, begin, end, strides
+                )
+                R.output(gv)
+            return gv
+
+    _assert_test(Input, Expected)
```
tlopex, line 1117: Why we don't use `tvm.ir.assert_structural_equal` here?

guan404ming: `_assert_test` is shared helper func in this file. I use it for consistency here. Shall we update to use `tvm.ir.assert_structural_equal`?

tlopex: No worries, I had a look and it's nearly the same

guan404ming:
> No worries, I had a look and it's nearly the same

Thanks!

PR #18636, opened 2026-01-05
[TIR][Schedule]Generalize fuseReductionEpilogue to support arbitrary epilogue expressions

kimm240:
### 1. Pattern Matching Removal

**Removed Items:**
- `EpilogueType` enum (Bias, BiasReLU, Clipping)
- `AnalyzeEpiloguePattern()` function
- Pattern-specific branching logic

**Current Approach:**
- Directly process the entire epilogue expression without pattern matching

### 2. Store Entire Epilogue Expression

- Store the entire epilogue expression in `epilogue_expression_`
- Use the expression directly without pattern analysis

```cpp
// Store the epilogue expression and reduction buffer load
epilogue_expression_ = inlined_store_->value;
reduction_buffer_load_ = loads[0];
```
### 3. Generalized Init Transformation

- Replace reduction buffer load with identity element (0)
- Apply to the entire expression to generate init value

```cpp
InitSubstituter init_subst(inlined_buffer_, identity_elem);
PrimExpr init_epilogue = init_subst(epilogue_expression_);
// Simplify: 0 + C[vi, vj] -> C[vi, vj]
```

**Examples:**
- `temp + C` → `0 + C` → `C` (simplify)
- `max(temp + C, 0)` → `max(0 + C, 0)` → `max(C, 0)`
- `min(max(temp, lower), upper)` → `min(max(0, lower), upper)`

### 4. Generalized Update Transformation

- Replace reduction buffer load with reduction update
- If parent is Add and the other operand is not a reduction buffer → treat as bias addend and remove
- Otherwise → apply expression as-is

```cpp
class GeneralizedEpilogueApplier : public ExprMutator {
  // Replace reduction buffer load with reduction update
  // Automatically detect and remove bias addend in Add nodes
  // Automatically support other activation functions
};
```

### Existing Tests Pass

All existing tests pass, maintaining backward compatibility:

- `test_fuse_reduction_epilogue_basic` 
- `test_fuse_reduction_epilogue_fp32` 
- `test_fuse_reduction_epilogue_numerical_correctness` 
- `test_fuse_reduction_epilogue_multiple_epilogue` 
- `test_matmul_bias_relu` 
- `test_matmul_bias_relu_correctness_unified` 
- `test_matmul_clipping` 
- `test_matmul_clipping_correctness_unified` 
- Other commutative variants tests 

Total: All 15 tests pass


kimm240:
@wrongtest-intellif In this PR, I implement general epilogue form. This idea is discussed in https://github.com/apache/tvm/pull/18515.

src/tir/schedule/primitive/compute_inline.cc
```diff
@@ -1308,53 +1179,163 @@ Block ReductionEpilogueFuser::CreateFusedReductionBlock(const BlockNode* reducti
...
+      return load;
+    }
+
+   private:
+    Buffer target_buffer_;
+    PrimExpr identity_elem_;
+  };
+
+  // Identity element for reduction (assumed to be 0 for addition-based reductions)
+  PrimExpr identity_elem = tir::make_zero(epilogue_output_buffer_->dtype);
```
wrongtest-intellif, line 1287: Generally, we require `epilogue(zero ⊕ x0 ⊕ ... ⊕ xn) == g(xn, g(xn-1, g(.....g(x0, epilogue_init(zero)))...)` with `g` and `epilogue_init` deduce from epilogue expression, right?

src/tir/schedule/primitive/compute_inline.cc
```diff
@@ -1308,53 +1179,163 @@ Block ReductionEpilogueFuser::CreateFusedReductionBlock(const BlockNode* reducti
...
+            BufferLoad load = Downcast<BufferLoad>(ExprMutator::VisitExpr_(op));
+            if (load->buffer.same_as(target_buffer_)) {
+              found_target_load_ = true;
+              // Check if parent is Add (will be checked in VisitExpr_(const AddNode*))
+              return replacement_;
+            }
+            return load;
+          }
+
+          PrimExpr VisitExpr_(const AddNode* op) final {
```
wrongtest-intellif, line 1365: Try ensure no incorrect patterns. Like `epilogue = relu(reduce_out * a + C[i])`, `epilogue = (reduce_out + C[i]) * (reduce_out + C[i])`, etc.

wrongtest-intellif (commented):
The generalized version cover existing testcases. But since it is a "generalized" version, it would be great to add more self-contained checking or comments to prove that the transformation do not accept "false-positive" patterns.  It may not be a trivial problem but we could try some improvements.

Refer to https://arxiv.org/html/2510.08726v1, which even leverage symbolic solvers for reduction fusion transformations.

src/tir/schedule/primitive/compute_inline.cc
kimm240: For `BiasAdd`, the transformation $epilogue(\sum x) = \sum x + Bias$ holds regardless of whether the bias is handled at the initialization stage or the end. 

However, for non-linear operations like `ReLU` and `Clipping`, this identity does not hold if we simply substitute the identity element in the `Init` block, as $f(\sum x) \neq f(0) + \sum x$.

I initially aimed to generalize the logic by processing these operations through a unified substitution mechanism (essentially treating them in a 'per-iteration')  to maintain the existing framework's structure-what is merged in #18515.

Should we strictly move these non-linear transformations to the final Store stage (applied once to the final sum)?
Or, since the previous 'per-iteration' behavior was already merged and used, should we keep it as an option or a specific 'fused-update' mode? I'm eager to hear about it.

src/tir/schedule/primitive/compute_inline.cc
kimm240: To fix Case 1 issue, I add code in commit da8354b.
Case 2 is already considered in this PR.

I add test cases for both of issues, and check tests well run.
If there are other  issues  to  solve, let me know and then fix it.

src/tir/schedule/primitive/compute_inline.cc
wrongtest-intellif: The correctness matters most. The concrete strategy could be free.

kimm240:
@wrongtest-intellif 
I fix lint and push it.

If there are any specific features, improvements, or roadmap items you'd like to see developed in these areas, please let me know. I'd be more than happy to take them on!

tlopex (approved):

tlopex:
Hi @kimm240 Could you resolve the conflict so that we can merge it?

kimm240:
@tlopex
Hi I resolve the  conflict!

PR #18637, opened 2026-01-05
[Relax] Add FRelaxInferLayout for flip operator

guan404ming:
## Why

  The flip operator lacked layout inference support, preventing it from participating in layout transformations during the ConvertLayout pass.

  ## How

  - Add InferLayoutFlip function that transforms the axis attribute according to the input layout
  - Register FRelaxInferLayout attribute for relax.flip operator
  - Add test case for conv2d followed by flip with layout conversion


guan404ming:
cc @tlopex @mshr-h

guan404ming:
Thanks!

PR #18638, opened 2026-01-05
[Relax] Add FRelaxInferLayout for scatter_elements operator

guan404ming:
## Why

The scatter_elements operator was missing FRelaxInferLayout support, which prevented proper layout transformation when used with operators like conv2d that require layout conversion.

## How

- Implement InferLayoutScatterElements function that handles layout inference for scatter_elements
- Transform axis attribute according to the inferred layout using FindAxis
- Handle sub-indexed layout fallback to initial layout
- Add test case for conv2d + scatter_elements layout conversion

### DIFF
--- a/tests/python/contrib/test_msc/test_graph_build.py
+++ b/tests/python/contrib/test_msc/test_graph_build.py
@@ -2443,22 +2443,22 @@ def test_scatter(dynamic: bool):
 
     expected1 = {
         "inputs": [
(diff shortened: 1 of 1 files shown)


guan404ming:
cc @tlopex @mshr-h

mshr-h:
Please resolve the conflict so that we can merge. cc @guan404ming

guan404ming:
> Please resolve the conflict so that we can merge. cc @guan404ming

Thanks, I've resolved it.

PR #18640, opened 2026-01-06
[TIR] improve typing for tir.PrimExpr

clouds56:
## Motivation
Now we use PrimExpr as type hint in most places and it would report errors (in static typing) sometimes.

case 1:
```python
a = Var('a', dtype='int')
a + a # this is Ok since __add__(self, other: PrimExpr) -> PrimExpr
a + 1 # this is not Ok
a + a + 1 # this is not Ok since (a+a) is a PrimExpr, not PrimExprWithOp
```

case 2:
```python
a = Var('a', dtype='int')
T.grid(a, a, a) # this is ok since a is PrimExpr
T.grid(1, 2, 3) # this is not Ok, since 1 is not a PrimExpr
```

## Solution
This PR is to mitigate these cases by introducing
```python
PrimIntExpr = Union[PrimExpr, int]
PrimFloatExpr = Union[PrimExpr, float]
PrimLogicalExpr = Union[PrimeExpr, int, bool]
```

## Problems
- [ ] If there's no problem with current idea, I'd like to apply PrimIntExpr/PrimFloatExpr to more places to improve typing system.
- [ ] It might be a huge changeset, should I split it into several PR?
- [ ] How to handle pylint E0633, upgrade pylint or add some ignore for this?
- [ ] `ForFrame` doesn't have `__iter__`, this would be fixed in another PR. Current PR would focus on `PrimExpr`. POC here https://github.com/apache/tvm/commit/7d64de158664aef0268315d82dafdb4027cd5c0e. It could be included in this PR if you like.

### DIFF
--- a/python/tvm/script/ir_builder/tir/ir.py
+++ b/python/tvm/script/ir_builder/tir/ir.py
@@ -685,19 +685,18 @@ class axis:  # pylint: disable=invalid-name
 
     @overload
     @staticmethod
-    def remap(
-        kinds: str, bindings: Union[PrimExpr, Tuple[PrimExpr]], dtype: str = "int32"
-    ) -> Var: ...
+    def remap(kinds: str, bindings: Union[PrimExpr, Tuple[PrimExpr]], dtype: str = "int32") -> Var:
+        ...
+
     @overload
     @staticmethod
-    def remap(
-        kinds: str,
-        bindings: Tuple[PrimExpr, ...],
-        dtype: str = "int32",
-    ) -> List[Var]: ...
+    def remap(kinds: str, bindings: Tuple[PrimExpr, ...], dtype: str = "int32") -> List[Var]:
+        ...
+
     @overload
     @staticmethod
-    def remap(kinds: str, bindings: List[PrimExpr], dtype: str = "int32") -> List[Var]: ...
+    def remap(kinds: str, bindings: List[PrimExpr], dtype: str = "int32") -> List[Var]:
+        ...
 
     @staticmethod
     def remap(
@@ -945,9 +944,9 @@ def grid(*extents: PrimIntExpr) -> frame.ForFrame:
     return _ffi_api.Grid(extents)  # type: ignore[attr-defined] # pylint: disable=no-member
 
 
-def Assert(
+def Assert(  # pylint: disable=invalid-name
     condition: PrimLogicalExpr, message: str
-) -> frame.AssertFrame:  # pylint: disable=invalid-name
+) -> frame.AssertFrame:
     """Create an assertion statement.
 
     Parameters


clouds56:
The lint is false positive in this case `E0633: Attempting to unpack a non-sequence (unpacking-non-sequence)`, see https://github.com/pylint-dev/pylint/issues/4696

I update pylint from 3.0.0 to 3.3.9 the problem disappears.

tlopex:
@clouds56 Thanks for working on this, and sorry for the late reply.

After looking through the full change, I don’t think we should have it in the current form. I’d rather say that clearly now than keep dragging the review out.

Main concerns:

1. Scope does not match the project’s typing discipline.

TVM CI only enforces ruff today; we do not enforce mypy or pyright anywhere in the project. That means adding several new public aliases plus a large number of signature changes increases our maintenance burden, without any mechanism that ensures those annotations remain correct over time. In practice, we would be taking on long-term annotation debt for limited in-tree benefit.

2. It conflicts with TIR’s intended semantic boundary.

In TIR, `PrimExpr` is the canonical API, while Python `int` / `float` are convenience values converted implicitly through the FFI boundary. Annotating parameters everywhere as `Union[PrimExpr, int]` effectively promotes that convenience path into the type-level contract. That weakens the signal at call sites and makes dtype behavior less explicit. For example, in an expression like `a(int32) + 1e12`, the annotation no longer encourages the reader to think about what dtype is actually intended. I think we should keep `PrimExpr` as the documented contract.

3. Some of the narrower return types are not semantically correct.

For example, annotating `__add__` as returning `"Add"` is not generally accurate, since constant folding may produce an `IntImm`, and mixed-dtype cases may introduce a `Cast`. Similarly, annotating `__eq__` as returning `"EqualOp"` conflicts with the conventional `__eq__ -> bool` expectation and is likely to create new checker noise downstream at ordinary `==` call sites. Also, `PrimLogicalExpr = Union[PrimExpr, int, bool]` does not really carry additional meaning, since `bool` is already a subclass of `int` in Python.

4. This feels like a symptom-level fix rather than a root-level fix.

The motivating issue is real: IDE squiggles on common patterns like `a + 1` or `T.grid(1, 2, 3)` are annoying. But addressing that by widening annotations across a large number of APIs means every future builder or helper will need the same treatment. If we want a durable solution, it should probably be handled closer to the FFI / `PrimExpr` conversion boundary, rather than being distributed across many individual call sites.

Thanks again for the effort here. I don’t think the overall motivation is wrong, but the current scope and the interaction with TIR’s dtype/semantic discipline make this hard to merge as-is.

clouds56:
Thanks for the detailed feedback.
I’d like to clarify my thinking and discuss the direction a bit further.

I’m still relatively new to TVM and am actively learning the primitives and APIs, so please feel free to correct me if I misunderstand anything.

## annotation in the repo should be more like a `hint` but not a `contract`

I understand your concern about maintenance cost. Correct me if I'm wrong, now the typing status of our project is not enforced, the existing typing might behave more like a `hint` but not a `contract`.

Everyone uses type hint in different ways, most of them are used for mypy/IDE to check.
Since we don’t enforce mypy, it’s unclear who the primary consumers of these type hints are—users, developers, or both.
My view is that we should optimize for usability here: even if annotations are not perfectly “correct,” making them more kindly still provides practical value.
For example, we annotated `__add__` to `PrimExpr` is correct, while annotated `__add__` to `PrimExprWithOp`/`Add` is kindly.

> we would be taking on long-term annotation debt for limited in-tree benefit.

Now annotations in this repo is already unsound (not matching what it actually is), so maintain of annotation would be definitely a form of debt. I have 3 ways towards future:
1. Gradually move toward more kindly annotations, that would improve IDE support, like auto complete and type check, that would provide in-tree benefits, right?
2. Do minimal maintain for annotations, fix cases like `def __eq__(self, other: PrimExpr) -> PrimExpr` discussed below to `def __eq__(self, other: PrimExpr) -> PrimExpr | bool`.
3. Do nothing, leave outdated annotation as is, they are of no value after they are created. This approach increases technical debt.

## TIR’s intended semantic boundary.

> Similarly, annotating __eq__ as returning "EqualOp" conflicts with the conventional __eq__ -> bool expectation

Taking `__eq__` as an example, annotating it as EqualOp is wrong, but the original code `def __eq__(self, other: PrimExpr) -> PrimExpr` conflicts with bool result as well. I think this wouldn't be likely fixed in short future.
So if both is not correct, `EqualOp` here might be more intuitive for readers, and more helpful for IDE, since EqualOp is a PrimExprWithOp.
I'd like to know your final decision what signature should be here to update the PR.

And for `Union[PrimExpr, int]` or `PrimIntExpr`, this just occurs in final API for input paramters.

## This feels like a symptom-level fix rather than a root-level fix.

Yes, your are right. I’m not familiar with tvm's codebase, as an end user, I interact mostly things like `@I.ir_module` and `@R.function` that's why I focused on API like `buffer`, `__add__`, etc. first.
And in my view, annotate around FFI seems like a more long-term solution, while it might have more work to be done than annotate around user facing API, which is relatively more stable. Internal FFI might be refactored during the time.

I agree there's lots of API needs to be annotated, but we could separate it into multiple steps. Each step is no harm and provide a better future.

## roadmap of tvm

I like the idea of tvm, and have been following the repo quite a while, back to the period when it used to be designed to replace the underlying framework of mxnet. I once thought that's the future of foundation of ML infra.
Today I still think tvm is an impressive piece of engineering. However, from a user perspective, it can be difficult to work with today. The lack of up-to-date documentation, limited examples, and weak IDE support (e.g., autocomplete) can discourage adoption.
Frameworks like PyTorch succeed in part because they are highly user-friendly and accessible to a broad audience. More users lead to more potential contributors.

## final note

I fully acknowledge that this PR may not be the right shape yet, and it likely contains mistakes.

However, I do believe that improving user-facing type annotations is a worthwhile direction for TVM’s development, and I’d be happy to adjust the scope or approach based on your guidance.

For the prosperity of tvm.

tlopex:
Thanks for the pushback, @clouds56!  I want to revise part of my earlier reply.

I agree that the input-side widening (`PrimExpr -> Union[PrimExpr, int, float]`) is reasonable. `PrimExpr` in C++ already has implicit constructors from integer and floating-point literals, and the Python side of TVM also relies heavily on that behavior in practice. So in that sense, the current annotation is narrower than the actual contract. I’m fine with landing that part, along with the `PrimIntExpr` / `PrimFloatExpr` aliases.

What I still do not agree with:

- Return narrowing such as `-> "Add"` is not sound. `_generic.add` can fold to an `IntImm`, or introduce a `Cast` during promotion, so annotating it as always returning `Add` is more misleading than leaving it as `PrimExpr`.
- `__eq__ -> "EqualOp"` should be discussed separately. If we want to annotate it, the choices are more subtle (`bool`, `EqualOp`, a union, or leaving it unannotated). I do not think “both are imperfect, so pick the more specific one” is a good justification.
- `PrimLogicalExpr = Union[PrimExpr, int, bool]` does not look useful as written, since `bool` is already a subclass of `int`.

More broadly, I do not think this PR should be treated as approval for the larger direction. Extending the same pattern to more APIs, upgrading pylint expectations, or including unrelated typing cleanups like `ForFrame.__iter__` should each be separate decisions and separate PRs.

Whether it still makes sense to push this kind of change — even scoped down to input widening + expression aliases — probably deserves more discussion before we commit to a direction.

cc @tqchen — would appreciate your opinion. Specifically: given that we do not currently enforce mypy / pyright in CI, is it worth introducing `PrimIntExpr` / `PrimFloatExpr` and spreading them across TIR APIs? Or would you rather we leave the annotations as-is?

PR #18643, opened 2026-01-07
[Relax] Add FRelaxInferLayout for scatter_nd operator

guan404ming:
## Why

The scatter_nd operator was missing FRelaxInferLayout attribute, which is needed for proper layout transformation during model optimization.

### How

- Added InferLayoutScatterND function that uses data tensor's layout for output since scatter_nd maintains input shape
- Registered FRelaxInferLayout attribute


guan404ming:
cc @tlopex @mshr-h

guan404ming:
Thanks!

PR #18649, opened 2026-01-09
[Relax] Remove obsolete TODO comments

guan404ming:
- Remove TODO about `ctx->ReportWarning()` since LOG(WARNING) is the standard pattern
- Remove TODO about skipping type params/constraints since the current FuncType implementation only contains arg_types and ret_type fields. `type_params` and `type_constraints` don't exist in the codebase.
- Fixed typo in warning message


guan404ming:
Thanks @cchung100m 
cc @tlopex @mshr-h

PR #18650, opened 2026-01-09
[Relax] Move SymbolicMatcher to tir/analysis

guan404ming:
## Why

The SymbolicMatcher class was previously embedded in src/relax/transform/fuse_tir.cc, making it unavailable for use by other TIR passes. A TODO comment indicated it should be moved to the tir folder for better organization and reusability.

## How

Created new files `src/tir/analysis/symbolic_matcher.h` and `src/tir/analysis/symbolic_matcher.cc` to house the SymbolicMatcher class


guan404ming:
cc @tlopex @mshr-h

tqchen:
in this case, given we still use it locally in a single place, let us defer until we have more than two places using it. the goal is to keep common utils minimized

guan404ming:
> in this case, given we still use it locally in a single place, let us defer until we have more than two places using it. the goal is to keep common utils minimized

Sure, I would close this for now and feel free to revisit if needed in the future.

PR #18657, opened 2026-01-12
[RELAX][LAYOUT] Support multiple axis packing

srkreddy1238:
Like OIHW[4o4i] where we can pack multiple axis.
Helpful while handling complex target layouts.
This PR covers layout representation and transforms for these.

In implementation the ```IterVar``` which was holding one axis till now.  With this it will hold a packed axis as ```name=4o4i``` and ```range=16```. The new helpers can be used to retrieve individual axis ```4o``` and ```4c``` when ever needed. The forwards and backwards are enhanced accordingly.

### DIFF
--- a/include/tvm/s_tir/data_layout.h
+++ b/include/tvm/s_tir/data_layout.h
@@ -35,6 +35,8 @@
 #include <utility>
 #include <vector>
 
+#include "tvm/tir/var.h"
+
 namespace tvm {
 namespace tir {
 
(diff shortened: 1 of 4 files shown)


include/tvm/tir/data_layout.h
```diff
@@ -249,20 +291,23 @@ class Layout : public ObjectRef {
...
+        if (var->var->name_hint == axis.name()) {
+          return true;
+        }
       }
     }
     return false;
   }
 
-  const LayoutAxis& operator[](int32_t i) const {
+  IterVar operator[](int32_t i) const {
```
srkreddy1238: Not comfortable with this change though this feature works. Reworking on avoiding this return of ```IterVar```.

srkreddy1238:
Reopened : https://github.com/apache/tvm/pull/18869

PR #18662, opened 2026-01-14
ir: add size guard to load_json to prevent pathological inputs

HyperPS:
**This PR adds a lightweight input-size guard to `load_json()` to prevent pathological JSON inputs from causing excessive memory or CPU usage.**

_The change preserves existing IR semantics and behavior for all valid
inputs, while providing basic protection against unbounded inputs in
networked and automated workflows._

This PR addresses a denial-of-service issue reported via **Huntr** related to unbounded JSON IR loading.


tqchen:
sorry i am not sure if this is as helpful, since users are expected to load from trusted sources

PR #18664, opened 2026-01-16
[Relax] Add FDataDependent operator attribute for LegalizeOps

guan404ming:
## Why

The LegalizeOps transform was using string matching to detect data-dependent operators by checking if "dynamic" appears in the operator name. This approach is fragile and doesn't scale well as new data-dependent operators are added.

## How

- Add FDataDependent operator attribute to properly mark data-dependent operators
- Set FDataDependent=true for relax.dynamic_strided_slice operator
- Update LegalizeOps transform to check the FDataDependent attribute instead of string matching


guan404ming:
cc @tlopex @mshr-h

guan404ming:
Thanks!

PR #18665, opened 2026-01-16
[Relax] Replace topi.take with relax.op.take

guan404ming:
## Why

TODO comment indicated that topi.take should be replaced with relax.op.take once it has better support. The relax take operator is now mature enough to replace the TOPI external call.

## How

- Replace bb.emit_te(topi.take, ...) with relax.op.take(..., axis=0) in EmbedLayerNormalization converter


guan404ming:
cc @tlopex @mshr-h

guan404ming:
Thanks!

PR #18667, opened 2026-01-16
[Relax] Add native size operator

guan404ming:
## Why

ONNX models use the Size operator to get total element count of a tensor. Relax didn't have a native equivalent.

## How

- Adds R.size(tensor) operator that returns the total number of elements in a tensor as a scalar int64


guan404ming:
cc @tlopex @mshr-h

guan404ming:
Thanks!

Issue #18648, opened 2026-01-09
[Bug] `relax.frontend.torch.from_exported_program` aborts on sparse CSR buffer (`layout_impl is only implemented for TensorImpl subclasses`)


tinywisdom:
### Summary

When using the TVM Relax Torch frontend to import a `torch.export.export`ed program that involves a `torch.sparse_csr_tensor` registered as a buffer, `from_exported_program` crashes with a C++ `c10::Error`:

> `layout_impl is only implemented for TensorImpl subclasses.`

`torch.export.export` itself succeeds. The crash happens during `from_exported_program(ep)`.

---

### Environment

```text
OS: Linux x86_64
PyTorch: 2.9.0+cu128
TVM: 0.22.0
Python: 3.10.x
```

---

### Minimal Reproduction

```python
# repro_92022_export_tvm_frontend.py
import torch
import torch.nn as nn

import tvm
from tvm.relax.frontend.torch import from_exported_program

print("torch version:", torch.__version__)
print("tvm version:", getattr(tvm, "__version__", "unknown"))


class MyModel(nn.Module):
    def __init__(self):
        super().__init__()
        crow_indices = torch.tensor([0, 1, 2], dtype=torch.int64)
        col_indices = torch.tensor([0, 1], dtype=torch.int64)
        values = torch.tensor([1.0, 1.0], dtype=torch.float32, requires_grad=True)

        csr_tensor = torch.sparse_csr_tensor(
            crow_indices, col_indices, values, dtype=torch.float32
        )
        # Register sparse CSR tensor as a buffer
        self.register_buffer("csr_tensor", csr_tensor)
        # Explicitly enable grad as well
        self.csr_tensor.requires_grad_(True)

    def forward(self, x):
        # Convert buffer to sparse CSR layout again
        csr2 = self.csr_tensor.to_sparse(layout=torch.sparse_csr)
        y = torch.matmul(csr2, x)
        return y.sum()


def GetInput():
    return torch.ones((2, 1), dtype=torch.float32)


def main():
    model = MyModel().to("cpu").eval()
    x = GetInput().to("cpu")

    print("Start torch.export.export ...")
    ep = torch.export.export(model, (x,))
    print("torch.export.export done.")

    print("Start from_exported_program ...")
    ir_mod = from_exported_program(ep)
    print("from_exported_program done.")
    print(ir_mod)


if __name__ == "__main__":
    main()
```

---

### Actual Behavior

Output on my machine:

```text
[...]

PR #18670, opened 2026-01-19
[Relax][Torch] Avoid decomposition crash with sparse CSR buffers

LudovicoYIN:
### Motivation
The Relax Torch frontend crashes when importing an exported program that includes
a torch.sparse_csr_tensor registered as a buffer. The crash happens during
from_exported_program because run_decompositions() triggers a PyTorch
layout_impl error for sparse tensors.

This PR avoids the crash while keeping the import pipeline functional for such
models, even though Relax does not yet support sparse tensors.

### Changes

- Skip run_decompositions() when the exported program contains sparse tensors
- Treat aten.to_sparse.default as a no-op in the Relax Torch frontend
- Add a regression test that imports a model with a sparse CSR buffer

### Testing
- test_frontend_from_exported_program.py

Fixes: [[Bug] Relax Torch frontend crash with sparse CSR buffer in ExportedProgramhttps://github.com/apache/tvm/issues/18648](https://github.com/apache/tvm/issues/18648)


tests/python/relax/test_frontend_from_exported_program.py
```diff
@@ -8787,5 +8787,33 @@ def main(
     verify_model(UpsampleNearest2dSize(), example_args, {}, expected_size)
 
 
+def test_from_exported_program_sparse_csr_buffer():
+    if not hasattr(torch, "sparse_csr"):
+        pytest.skip("sparse CSR tensor is not supported in this PyTorch build")
```
yongwww: what happens if we don't skip the test?

LudovicoYIN: My intent was to run this test only on PyTorch builds that actually support torch.sparse_csr, to avoid false failures on builds without CSR support. In my local environment the test still passes even if I remove the skip, but I added it for portability. Do you prefer keeping the skip for compatibility, or should I remove it?

LudovicoYIN:
@yongwww Thanks for the earlier feedback!
I removed the pytest.skip as discussed, merged main.
Could you please take another look and approve if it looks good? Thanks!

tlopex (approved):
looks good to me @LudovicoYIN  Thanks for the contribution!

LudovicoYIN:
@tlopex  thanks your review

PR #18676, opened 2026-01-21
[Relax] Implement dynamic output trimming for NMS

guan404ming:
## Why                                                                                                                      
                                                                                                                         
NMS operator returns fixed-size output with trailing garbage data, wasting memory and requiring manual trimming for ONNX 
compatibility.                                                                                                           
                                                                                                                         
## How                                                                                                                      
                                                                                                                         
- Add dynamic_strided_slice to trim NMS output to valid detections only                                                  
- Build slice parameters using TE compute to avoid legalization issues


guan404ming:
cc @tlopex @mshr-h

PR #18677, opened 2026-01-21
[Relax][NN] Add batch_flatten operator

guan404ming:
## Why                                                                                                                   
                                                                                                                           
The `batch_flatten` operator was needed for CoreML backend support but was not implemented in Relax, causing the CoreML pattern and converter to be commented out.                                                                               
                                                                                                                         
## How                                                                                                                   
                                                                                                                         
- Add C++ operator `relax.nn.batch_flatten` with struct info inference                                                   
- Add Python wrapper in `relax.op.nn`                                                                                    
- Add legalization via `topi.reshape`                                                                                    
- Enable CoreML pattern and converter for `nn.batch_flatten`                                                             
- Add unit tests for operator, struct info inference, and legalization


guan404ming:
cc @tlopex @mshr-h

guan404ming:
Thanks!

PR #18680, opened 2026-01-22
[Web][Version] Fix WebLLM vision model issues

akaashrp:
1. Try multiple candidates for maxComputeInvocationsPerWorkgroup
2. Update tvmjs version to 0.24.0-dev1


tqchen:
@akaashrp we need a rebase

PR #18690, opened 2026-01-28
[Relax] Add NN operator attributes include to TensorRT codegen

guan404ming:
## Why                                                                                                                 
                                                                                                                      
The header now exists and provides operator attribute definitions needed for TensorRT codegen.                                     
                                                                                                                    
## How                                                                                                                 
                                                                                                                    
Add #include <tvm/relax/attrs/nn.h> to TensorRT codegen


guan404ming:
Thanks!

PR #18692, opened 2026-01-29
[CI] Fix NameError in nightly docker update workflow

guan404ming:
## Why
The Nightly Docker Update CI workflow fails with NameError: name 'Sh' is not defined because Sh is used but not imported.                                               

https://github.com/apache/tvm/actions/workflows/nightly_docker_update.yml
                                                                                                                                                                        
## How                                                                                                                                                                                                                                                                                                                                     
- Add Sh to the import from cmd_utils
- Sort the import with ruff

### DIFF
--- a/ci/scripts/jenkins/open_docker_update_pr.py
+++ b/ci/scripts/jenkins/open_docker_update_pr.py
@@ -17,16 +17,17 @@
 # under the License.
 
 import argparse
-import logging
 import datetime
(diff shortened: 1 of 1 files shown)


guan404ming:
cc @tqchen

tlopex (approved):

guan404ming:
Thanks!

PR #18694, opened 2026-01-29
[BugFix][TIR] Fix incorrect optimization when lowering floordiv and f…

### LINKED ISSUE #18684: [Bug] Incorrect optimization for tir.floordiv/floormod
A small test:
``` python
import tvm
from tvm.script import ir as I

@tvm.register_global_func("tir_print")
def tir_print(A):
    print(A)

@I.ir_module
class TestCeilDiv:
    @T.prim_func
    def main(n: T.int32) -> None:
        T.func_attr({"target": T.target("llvm")})
        res : T.int32 = T.ceildiv(n, T.int32(10))
        T.call_packed("tir_print", res)

mod = TestCeilDiv
exe = tvm.build(mod)
exe["main"](24)
```

### Expected behavior

Expected result of ceildiv(24,10) is 3.

### Actual behavior

The result of this test is 4.

### debug

The direct cause is incorrect optimization of TIR by the LowerIntrin pass.
after LowerIntrin pass, tir became:
`T.Cast("int64", T.Div(n - 2147483631, 10) - -214748364)`
This is obviously wrong.

In src/tir/transforms/lower_intrin.cc:156
`PrimExpr ceildiv = truncdiv(-min + (op->b - 1), op->b);`
In my test case, min is -2147483648+9.
b-1 is 9.
`-min + (op->b - 1)` = 2147483648, after const fold in src/arith/const_fold.h:80, the return value is -2147483648, because 2147483648 exceeds the upper limit of int32.
This resulted in a non-equivalent transformation.

### Possible fix method

To avoid overflow of '(op->b - 1) - min' , we can assert (b-1-a_min) < MAX_VALUE+1. 
That is (b_max-a_min) < MAX_VALUE + 2.
We can avoid this issue by changing the condition of the if statement in L122.
```cpp
if (const_int_bound_a->min_value < 0 &&
          const_int_bound_a->min_value >
              (const_int_bound_b->max_value -
               (Downcast<IntImm>(tvm::max_value(op->a->dtype.element_of()))->value) - 2)) {
```
Of course, this may not be the best fix.
Because the floormod node uses similar optimization，it also has the same problem, in src/tir/transforms/lower_intrin.cc:217.
* needs-triage

guocju:
…loormod

This patch fixes an issue in the LowerIntrin pass where incorrect optimizations were applied to floordiv and floormod operations.

The root cause is that the pass attempts to find an equivalent representation for floordiv(a, b) by calculating the expression (op->b - 1) - a_min. This expression, when subjected to constant folding, can potentially overflow the range of int32 or int16. When this overflow occurs, the transformation becomes invalid and no longer equivalent to the original operation.

To fix this, we enhanced the condition under which the transformation is applied. The new condition ensures that the transformation is only performed when (b_max - a_min) is less than INT_MAX + 2. If this condition is not met, the transformation is skipped and the common lowering steps are followed to ensure correctness.

A regression test has been added to cover this case.

Fixes #18684

### DIFF
--- a/src/tir/transforms/lower_intrin.cc
+++ b/src/tir/transforms/lower_intrin.cc
@@ -118,10 +118,20 @@ class IntrinInjecter : public tvm::arith::IRMutatorWithAnalyzer {
 
       // If the numerator's lower bound is known, express the floordiv
       // in terms of truncdiv using only positive operands.
-      arith::ConstIntBound const_int_bound = analyzer_->const_int_bound(op->a);
-      if (const_int_bound->min_value < 0 &&
-          const_int_bound->min_value >
-              -(Downcast<IntImm>(tvm::max_value(op->a->dtype.element_of()))->value)) {
+
+      // The optimization below rewrites expressions involving `-a_min + (b - 1)`.
+      // Without proper bounds checking, this expression may overflow the dtype
+      // maximum, leading to non-equivalent transformations.
+      // To ensure safety, we require:
+      //   b_max - a_min <= max_value_of_dtype + 1
+      // This provides a conservative upper bound that prevents overflow and
+      // preserves the original semantics.
+      arith::ConstIntBound const_int_bound_a = analyzer_->const_int_bound(op->a);
+      arith::ConstIntBound const_int_bound_b = analyzer_->const_int_bound(op->b);
+      const int64_t max_value_of_dtype =
+          Downcast<IntImm>(tvm::max_value(op->a->dtype.element_of()))->value;
+      if (const_int_bound_a->min_value < 0 &&
+          const_int_bound_b->max_value - const_int_bound_a->min_value <= max_value_of_dtype + 1) {
         // The goal is to write floordiv(a,b) in terms of truncdiv, without using
         // negative operands.
         //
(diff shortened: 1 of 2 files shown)


src/tir/transforms/lower_intrin.cc
```diff
@@ -118,10 +118,11 @@ class IntrinInjecter : public tvm::arith::IRMutatorWithAnalyzer {
 
       // If the numerator's lower bound is known, express the floordiv
       // in terms of truncdiv using only positive operands.
-      arith::ConstIntBound const_int_bound = analyzer_->const_int_bound(op->a);
-      if (const_int_bound->min_value < 0 &&
-          const_int_bound->min_value >
-              -(Downcast<IntImm>(tvm::max_value(op->a->dtype.element_of()))->value)) {
+      arith::ConstIntBound const_int_bound_a = analyzer_->const_int_bound(op->a);
+      arith::ConstIntBound const_int_bound_b = analyzer_->const_int_bound(op->b);
```
tqchen, line 130: would be great to explicitly list out `(Downcast<IntImm>(tvm::max_value(op->a->dtype.element_of()))->value)` as a variable and have comment documenting why +2 is needed, so future changes have the right context

tqchen:
Thanks @guocju please cross check genimi review, and add explicit comment for context, the ci issue should be fixed by https://github.com/apache/tvm/pull/18695 thanks a lot for the contribution

guocju:
Yes, I’ve updated the check to explicitly extract max_value_of_dtype as a variable, and added a comment documenting why the condition is <= max_value_of_dtype + 1.

tqchen:
Thanks @guocju for the bugfix!

tqchen:
https://github.com/apache/tvm/pull/18699 contains some followups to further clean things up and robustify the scenario. thanks @guocju for bringing up the initial case

PR #18700, opened 2026-01-31
cmake: link terminfo to cpptest/tvm_allvisible when required by LLVM/libedit

at445:
Summary

- Detect and link libtinfo (terminfo) explicitly for `cpptest` and `tvm_allvisible` when found.
- Prevents link-time failures for test binaries when `llvm-config` does not export `-ltinfo`, which can happen with some LLVM packages (observed with LLVM 18 in Ubuntu 24.04.3 LTS (x86_64-linux-gnu) ).

Problem

- With USE_LLVM=ON, the build produced undefined references to terminfo symbols (e.g. `tgetstr`, `setupterm`) when linking `cpptest`. `ldd /usr/lib/llvm-18/lib/libLLVM-18.so` shows it depends on `libedit` and `libtinfo`, but `llvm-config --libs --link-static` did not include `-ltinfo`, so CMake didn't add it to the final link.

Change

- Minimal, portable CMake change:
  - `find_library(TINFO_LIB tinfo)` and, if found, `target_link_libraries(cpptest PRIVATE ${TINFO_LIB})` and, when present, `target_link_libraries(tvm_allvisible PRIVATE ${TINFO_LIB})`.

Reproduction

- Steps:
  1. cmake -S . -B build-llvm_ON-hide_OFF -DUSE_LLVM=ON -DHIDE_PRIVATE_SYMBOLS=OFF -DUSE_GTEST=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo
  2. cmake --build build-llvm_ON-hide_OFF --target cpptest
- Expected:
  - Before: link errors referencing terminfo symbols and no `cpptest` binary produced.
  - After: `cpptest` links and is produced.

Rationale

- Safer and more portable than relying on consumers to fix package-level `llvm-config` exports.
- No behavioral changes unless `tinfo` is required and available.

Logs
- find more information on attachment
[build-llvm_ON-hide_OFF.build.log](https://github.com/user-attachments/files/24979750/build-llvm_ON-hide_OFF.build.log)
[build-llvm_ON-hide_OFF.cmake.log](https://github.com/user-attachments/files/24979751/build-llvm_ON-hide_OFF.cmake.log)
[build-llvm_ON-hide_ON.build.log](https://github.com/user-attachments/files/24979752/build-llvm_ON-hide_ON.build.log)
[build-llvm_ON-hide_ON.cmake.log](https://github.com/user-attachments/files/24979753/build-llvm_ON-hide_ON.cmake.log)


CMakeLists.txt
```diff
@@ -670,7 +670,18 @@ if(GTEST_FOUND)
     if(DEFINED LLVM_SO OR HIDE_PRIVATE_SYMBOLS)
       target_link_libraries(cpptest PRIVATE ${LLVM_LIBS})
     endif()
+
+    # Ensure we link terminfo when required by LLVM / libedit at link time.
+    find_package(Curses)
+    if(CURSES_FOUND)
+       message(STATUS "Found Curses/terminfo library: ${CURSES_LIBRARIES}")
```
tqchen, line 677: i think llvm only requires tinfo but not curses library? we do want to avoid dependency bloat in this case

tqchen: The dependency should also be specified and captured through llvm config in findllm instead of here

at445: > i think llvm only requires tinfo but not curses library? we do want to avoid dependency bloat in this case
this is suggested by gemini-code-assist, but I could switch back tinfo if you insist ^_^
image

at445: > The dependency should also be specified and captured through llvm config in findllm instead of here

you are right! I will rethink about it

PR #18715, opened 2026-02-05
[SPIRV] Fix forloop codegen in vulkan

tqchen:
This PR fixes a regression in vulkan codegen where we need to create phi node after the start label.


tqchen:
cc @wrongtest-intellif

tlopex (approved):

PR #18720, opened 2026-02-07
[Relax] Add Dead Parameter Elimination

shreyasravi320:
# [Relax] Add Dead Parameter Elimination

This PR adds dead parameter elimination to `src/relax/transform/dead_code_elimination.cc` to remove unused parameters from Relax functions.

## Motivation

Functions often accumulate unused parameters due to:
- Deprecated model architectures (e.g., unused weight tensors)
- API evolution and backwards compatibility
- Refactoring artifacts

Eliminating these parameters reduces memory usage, speeds up compilation, and simplifies the IR.

## Changes

- Algorithm:
  1. Collect unused parameters across all functions in the module
  2. Update signatures and call sites in topological order
  3. Repeat while changes persist
- Tests in `tests/python/relax/test_transform_dead_param_elimination.py`

## Example Usage

Before:
```python
@R.function
def forward(x: R.Tensor, unused_weights: R.Tensor):
    return x + 1
```

After:
```python
@R.function
def forward(x: R.Tensor):
    return x + 1
```

## TODOs
- [ ] CI/CD checks
- [ ] Test mutual recursion more thoroughly
- [ ] Fully benchmark compilation time and instruction count
- [ ] Documentation

### DIFF
--- a/src/ir/module.cc
+++ b/src/ir/module.cc
@@ -152,7 +152,7 @@ void IRModuleNode::Add(const GlobalVar& var, const BaseFunc& f, bool update) {
 void IRModuleNode::AddUnchecked(const GlobalVar& var, const BaseFunc& func) {
   this->functions.Set(var, func);
 
-  auto* var_node = const_cast<GlobalVarNode*>(var.get());
+  const auto* var_node = var.get();
   var_node->struct_info_ = func->struct_info_;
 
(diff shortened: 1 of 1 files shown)


tqchen:
we do not want parameter change in general since this is the signature

PR #18723, opened 2026-02-07
Revert "[ci][docker] Nightly Docker image update"

tqchen:
Reverts apache/tvm#18710

Seems there is some issue in terms of the images, i think perhaps we should disable docker auto-updates, and instead relies on manual bump, we will need use https://github.com/apache/tvm/blob/main/ci/jenkins/docker-images.ini as ground truth

### DIFF
--- a/ci/jenkins/data.py
+++ b/ci/jenkins/data.py
@@ -51,31 +51,31 @@ aws_ecr_url = "dkr.ecr." + aws_default_region + ".amazonaws.com"
 # Docker Images
 docker_images = {
     "ci_arm": {
-        "tag": "tlcpack/ci-arm:20251130-061900-c429a2b1",
+        "tag": "tlcpack/ci-arm:20221013-060115-61c9742ea",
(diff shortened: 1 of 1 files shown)


tqchen:
cc @guan404ming @tlopex @yongwww  . context, the data was used by docker/bash.sh command, but was not aligned with the ci behavior that uses https://github.com/apache/tvm/blob/main/ci/jenkins/docker-images.ini. We should temp disable it for now otherwise ` docker/bash.sh` may no longer be consistet

guan404ming:
Thanks!

tqchen:
perhaps we should followup by either only allow manual update, or make things consistent ideally we should make sure  https://github.com/apache/tvm/blob/main/ci/jenkins/docker-images.ini is source of truth

tqchen:
https://github.com/apache/tvm/pull/18724 contains a PR to disable the PR. seems the current tag was not consistent but fine,

PR #18726, opened 2026-02-07
[Relax][PyTorch] Simplify tensor args conversion in Dynamo

mshr-h:
As per title.


tqchen:
likely with latest tvm-ffi, we don't need dlpack anymore, just pass in torch.Tensor directly should be fine

mshr-h:
thanks. updated as suggested.

PR #18727, opened 2026-02-07
fix: correct typos in Python docstrings

thecaptain789:
Fixed 'occured' to 'occurred' in transform.py and 'seperated' to 'separated' in mrvl.py.


thecaptain789:
Thanks for the review! Fixed both issues:
- Added the 'build an generates' → 'build and generates' typo fix on line 210  
- Updated the return description for SpecializePrimFuncBasedOnCallSite to be accurate

tlopex (approved):
Thank you!

PR #18733, opened 2026-02-08
[Relax] Migrate NN conv/pooling/grad attrs from Array<IntImm> to Array<int64_t>

guan404ming:
## Why                                                                                                                     
                                                                                                                   
has_attr pattern matching failed for integer attributes because StructuralEqual requires exact dtype match for IntImm,  
causing mismatches between Python-specified pattern values and actual operator attributes.                            
                                                                                                                   
## How

- Migrate NN conv/pooling/grad attrs from Array<IntImm> to Array<int64_t>

### DIFF
--- a/src/relax/ir/dataflow_matcher.cc
+++ b/src/relax/ir/dataflow_matcher.cc
@@ -55,38 +55,6 @@ namespace relax {
 using tvm::arith::Analyzer;
 
 /*!
- * \brief Normalize a value so that IntImm becomes boxed int64_t
- * and FloatImm becomes boxed double. This allows StructuralEqual
- * to compare pattern values (which may contain IntImm from Python)
- * against attr field values that use boxed primitives (e.g. Array<int64_t>).
- */
-ffi::Any NormalizeImmToPrimitive(const ffi::Any& value) {
-  if (const auto* int_imm = value.as<IntImmNode>()) {
-    return ffi::Any(int_imm->value);
-  }
-  if (const auto* float_imm = value.as<FloatImmNode>()) {
-    return ffi::Any(float_imm->value);
-  }
-  if (const auto* arr = value.as<ffi::ArrayObj>()) {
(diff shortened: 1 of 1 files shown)


tqchen:
Thanks for the contribution. I think this is due to one of the legacy issues. Background: previously we used IntImm for constant attribute, and Integer redirects to IntImm. 

POD types like `int` cannot not appear in the nested contained. Since tvm-ffi refactor, now we have dedicated int, so ideally for constant attributes like strides, we should enforce them now to use POD int and such issue will disappear

guan404ming:
> Thanks for the contribution. I think this is due to one of the legacy issues. Background: previously we used IntImm for constant attribute, and Integer redirects to IntImm. POD types like int cannot not appear in the nested contained. Since tvm-ffi refactor, now we have dedicated int, so ideally for constant attributes like strides, we should enforce them now to use POD int and such issue will disappear

Thanks for the suggestion. I adjust the pr to migrate NN conv/pooling/grad attrs from Array<IntImm> to Array<int64_t> which I think to address the issue from real root cause.

src/relax/ir/dataflow_matcher.cc
```diff
@@ -54,6 +54,38 @@ namespace relax {
 
 using tvm::arith::Analyzer;
 
+/*!
+ * \brief Normalize a value so that IntImm becomes boxed int64_t
+ * and FloatImm becomes boxed double. This allows StructuralEqual
+ * to compare pattern values (which may contain IntImm from Python)
+ * against attr field values that use boxed primitives (e.g. Array<int64_t>).
+ */
+ffi::Any NormalizeImmToPrimitive(const ffi::Any& value) {
```
tqchen: is this still needed if we enforce attr to be i64?

guan404ming:
@tvm-bot rerun

guan404ming:
Thanks!

PR #18736, opened 2026-02-09
[Relax] Support constant folding for call_tir with tuple outputs

guan404ming:
## Why

Constant folding skipped call_tir nodes with tuple (multi-tensor) outputs, leaving foldable operations unoptimized.

## How

- Add ConstEvaluateCallTIRTuple to handle call_tir with TupleStructInfo output by allocating and packing multiple output tensors
- Route VisitCallTIR through tuple vs single-tensor paths based on sinfo_args type
- Add test for folding a split-like prim_func with two output tensors


guan404ming:
cc @tlopex @mshr-h

guan404ming:
Thanks!

Issue #18682, opened 2026-01-23
[Tracking Issue] Upgrade Python to 3.10


mshr-h:
- [x] https://github.com/tlc-pack/ci/pull/105
- [x] #18748
- [x] #18737
- [x] https://github.com/apache/tvm/pull/18749
- [x] #18647 
- [x] upload new docker images @yongwww 
- [x] #18777
- [x] #18783
- [x] upload new docker images @mshr-h
- [x] #18787
- [x] #18775

mshr-h:
@yongwww 
can you upload new images to dockerhub? seems like I don't have access to tlcpack in dockerhub anymore.

yongwww:
> [@yongwww](https://github.com/yongwww) can you upload new images to dockerhub? seems like I don't have access to tlcpack in dockerhub anymore.

Updated

mshr-h:
Thank!

mshr-h:
closing as upgrade done

PR #18737, opened 2026-02-09
[CI] Remove i386 and Hexagon from CI pipeline (1)

mshr-h:
Part of #18682


tqchen:
note that because main requires the ci to pass, we should first have a PR to .asf.yaml to removes dep on the required tests then remove

mshr-h:
Thanks! That’s exactly what I was wondering about 🙏

mshr-h:
cc @tqchen @yongwww

PR #18741, opened 2026-02-09
[REFACTOR][S-TIR] Move remaining data structures to s_tir

tqchen:
This PR moves remaining related data structures to s_tir so tir core is more lean.
- Moves sblock_dependency_info and sblock_scope
- Moves related analyssis
- Hides the data_type_rewriter to private functions so header is cleaner.


tqchen:
@tvm-bot rerun

PR #18742, opened 2026-02-09
[Fix] Handle empty variable name in NameSupply::FreshName

MasterJH5574:
Map empty names to "v" to prevent codegen from producing GPU code with empty variable names that fail compilation. Also fix add_prefix_to_name to use the (possibly remapped) unique_name instead of the original name.


MasterJH5574:
@tvm-bot rerun

MasterJH5574:
@tvm-bot rerun

PR #18749, opened 2026-02-11
[CI] Remove i386 and Hexagon from CI pipeline (2)

mshr-h:
Part of #18682


docker/install/ubuntu_install_rust.sh
```diff
@@ -21,14 +21,6 @@ set -euxo pipefail
...
 
-# this rustc is one supported by the installed version of rust-sgx-sdk
-HOST_ARG=
-if [ "$(getconf LONG_BIT)" == "32" ]; then
-    # When building in the i386 docker image on a 64-bit host, rustup doesn't
-    # correctly detect the arch to install for so set it manually
-    HOST_ARG="--default-host i686-unknown-linux-gnu"
-fi
-
 # shellcheck disable=SC2086 # word splitting is intentional here
```
tqchen, line 32: seems rust is no longer needed in the ci, so we can remove rust all together fornow

mshr-h: looks like rust is still need to build sccache from source?
https://github.com/apache/tvm/blob/262c6d2e041e6a5d5775049b32b166b526b49fa9/docker/install/ubuntu_install_sccache.sh#L23

tqchen: i see, in this case i think it is helpful to keep 32bit compact here since it does hurt as much here

tqchen:
@tvm-bot rerun

PR #18762, opened 2026-02-12
[Relax] Remove TODO comment for moving code in fuse_tir.cc

guan404ming:
## Why

It is only used in one place and relax specific

## How

- Remove TODO comment for moving code in fuse_tir.cc

### DIFF
--- a/src/relax/transform/fuse_tir.cc
+++ b/src/relax/transform/fuse_tir.cc
@@ -32,7 +32,6 @@
 namespace tvm {
 namespace tir {
 
-// TODO(Siyuan): move it to somewhere under tir folder
 /*!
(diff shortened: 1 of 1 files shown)


tqchen:
sorry we seems to have previous comment on this that this file is better kept under relax given it is only used in one place and relax specific. Maybe we should update the comment instead so we don't have to pick it up again

guan404ming:
Sure, that makes sense. Just updated!

tqchen:
https://github.com/apache/tvm/pull/18650

guan404ming:
Oops, I forget that. We should remove that stale todo at that time. Thanks!

PR #18796, opened 2026-02-18
[CI] Refactor Dockerfiles and installation scripts

mshr-h:
- **Removed Rust toolchain** from all CI Docker images (no longer needed)
- **Deleted 8 obsolete scripts**: Ubuntu 20.04-specific installers, old LLVM/CMake/Golang/sbt scripts
- **New `download-and-verify` utility**: reusable curl + checksum verification helper, used across install scripts
- **CMake & sccache**: switched from source/cargo builds to prebuilt binaries with checksum verification
- **LLVM**: consolidated to one script, updated from Focal/LLVM 10-12 to Jammy/LLVM 15-17
- **APT key management**: migrated from deprecated `apt-key add` to `signed-by` keyrings (LLVM, Vulkan, ROCm, Node.js)
- **Version upgrades**: Emscripten 4.0.23, Node.js 22.x, TensorFlow 2.19.0, NumPy 1.26.*, Apache RAT 0.17, Wasmtime v41.0.3
- **Dockerfile maintainance**: `ENV KEY=value` syntax, `trap cleanup EXIT` patterns, `cmake --build` instead of `make`, Python support narrowed to 3.10-3.11

The image builds are significantly faster now.
- ci_arm: 37m -> 13m
- ci_cpu: 44m -> 15m
- ci_gpu: 53m -> 28m


mshr-h:
cc @tqchen @yongwww

tqchen:
would be nice if we can followup with uv

mshr-h:
for sure! it's my next step 🚀

tqchen:
reading notes here btw, i don't think we need Wasmtime anymore

PR #18812, opened 2026-02-23
[REFACTOR][CI] Modernize Python dependency management with uv

mshr-h:
This PR modernizes Python tooling in CI by migrating Python interpreter installation, venv creation and package installation to uv.

### DIFF
--- a/docker/install/ubuntu_install_python_package.sh
+++ b/docker/install/ubuntu_install_python_package.sh
@@ -21,25 +21,25 @@ set -u
 set -o pipefail
 
 # install libraries for python package on ubuntu
-pip3 install --upgrade \
-    "Pygments>=2.4.0" \
-    cloudpickle \
-    cython \
-    mypy \
+uv pip install --upgrade \
+    "Pygments~=2.19" \
+    "cloudpickle~=3.1" \
+    "cython~=3.0" \
+    "mypy~=1.15" \
     numpy==1.26.* \
-    orderedset \
-    packaging \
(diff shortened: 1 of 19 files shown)


mshr-h:
cc @tqchen @yongwww

docker/install/ubuntu_install_python_package.sh
```diff
@@ -21,25 +21,25 @@ set -u
 set -o pipefail
 
 # install libraries for python package on ubuntu
-pip3 install --upgrade \
-    "Pygments>=2.4.0" \
-    cloudpickle \
-    cython \
-    mypy \
+uv pip install --upgrade \
+    "Pygments~=2.19" \
```
yongwww, line 25: I have a dumb question, how do we get the the deps pins (e.g., Pygments/cloudpickle/cython/mypy/pytest/requests/scipy/Jinja2/ml_dtypes, etc.) here?

mshr-h: I checked the versions with `pip index versions` in the current docker images and picked them one by one. Something like:
- `python3 tests/scripts/ci.py cpu --interactive`, then
- `for p in Pygments cloudpickle cython mypy numpy orderedset packaging Pillow psutil pytest pytest-profiling pytest-xdist pytest-rerunfailures requests scipy Jinja2 junitparser six tornado ml_dtypes; do echo "=== $p ==="; python3.10 -m pip index versions "$p"; done`

Maybe we can just write a small script to automate this.

PR #18814, opened 2026-02-23
[Relax] Make ShapeType ndim parameter mandatory

guan404ming:
## Why

Forces callers to be explicit about whether they want unknown ndim (-1) or a specific value. Prevents accidentally creating an unknown-ndim ShapeType when the caller intended to pass a concrete dimension, catching bugs at the call site rather than silently propagating unknown shapes.

## How

- Remove default ndim from C++ and Python ShapeType constructors
- Update tests

### DIFF
--- a/include/tvm/relax/type.h
+++ b/include/tvm/relax/type.h
@@ -53,8 +53,7 @@ class ShapeTypeNode : public TypeNode {
 
 class ShapeType : public Type {
  public:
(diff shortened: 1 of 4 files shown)


guan404ming:
cc @tlopex @mshr-h

mshr-h (approved):

guan404ming:
Thanks.

PR #18823, opened 2026-02-25
[WebGPU] Add gating logic for subgroup shuffle primitives

ksgr5566:
## Summary
This adds gating logic on top of #17699 to support optional subgroup shuffle 
primitives based on a compile-time flag.

## Problem
The PR #17699 always generates subgroup shuffle ops when targeting WebGPU. 
However, not all WebGPU devices support subgroups. We need a way to:
- Default to shared memory reductions (universally compatible)
- Optionally enable subgroup shuffles for devices that support them

## Solution
Implement gating via TVM target parameter:
- Default `thread_warp_size=1` disables warp reductions (uses shared memory + barriers)
- Add target parser `UpdateWebGPUAttrs()` that sets `thread_warp_size=32` when `supports_subgroups=true`
- Add `--enable-subgroups` CLI flag in mlc-llm to surface the option to users

The gating happens at the reduction path selection level (`IsWarpReduction()` in 
`lower_thread_allreduce.cc`), ensuring subgroup ops are never generated unless explicitly enabled.

## Changes
- TVM: Target parser + default thread_warp_size=1
- MLC-LLM: --enable-subgroups flag  (https://github.com/mlc-ai/mlc-llm/pull/3431)
- WebLLM: WGSL shader dumping for verification

## Testing

Tested with Llama-3.2-1B-q4f16_1. Baseline (no flag) uses shared memory reductions; 
with flag, generates subgroupShuffle* ops.
Both the generated WGSLs here: https://gist.github.com/ksgr5566/301664a5dda3e46f44092be4d09b2d4f
Benchmarking: https://gist.github.com/ksgr5566/c9bd5bc5aadba999ec2f2c38eb0c49b3

### DIFF
--- a/src/target/target_kind.cc
+++ b/src/target/target_kind.cc
@@ -257,6 +257,36 @@ ffi::Map<ffi::String, ffi::Any> UpdateROCmAttrs(ffi::Map<ffi::String, ffi::Any> 
 }
 
 /*!
+ * \brief Update WebGPU target attributes for subgroup-enabled lowering.
+ * Runtime routing on the WebLLM side guarantees subgroup size == 32.
+ * Runtime routing on the WebLLM side guarantees
+ * maxComputeInvocationsPerWorkgroup >= 1024.
+ * This is intentionally constrained for the subgroup-enabled WASM variant.
+ * When supports_subgroups is true, canonicalize thread_warp_size to 32 so
+ * TIR lowering can emit subgroup shuffle reductions.
+ * \param target The Target to update
+ * \return The updated attributes
+ */
+ffi::Map<ffi::String, ffi::Any> UpdateWebGPUAttrs(ffi::Map<ffi::String, ffi::Any> target) {
+  bool subgroups = false;
+  if (target.count("supports_subgroups")) {
+    subgroups = Downcast<Bool>(target.at("supports_subgroups"));
+  }
+
+  if (target.count("thread_warp_size")) {
+    int64_t thread_warp_size = Downcast<Integer>(target.at("thread_warp_size"))->value;
+    TVM_FFI_ICHECK(subgroups || thread_warp_size <= 1)
+        << "WebGPU target with thread_warp_size=" << thread_warp_size
+        << " requires supports_subgroups=true";
+  }
+
+  if (subgroups) {
+    target.Set("thread_warp_size", int64_t(32));
+  }
+  return target;
+}
+
+/*!
  * \brief Test Target Parser
  * \param target The Target to update
  * \return The updated attributes
@@ -426,34 +456,6 @@ TVM_REGISTER_TARGET_KIND("vulkan", kDLVulkan)
     .add_attr_option<int64_t>("max_spirv_version")
     // Tags
     .set_default_keys({"vulkan", "gpu"});
-
-/*!
- * \brief Update WebGPU target attributes for subgroup-enabled lowering.
- * Runtime routing on the WebLLM side guarantees subgroup size == 32.
- * Runtime routing on the WebLLM side guarantees
- * maxComputeInvocationsPerWorkgroup >= 1024.
- * This is intentionally constrained for the subgroup-enabled WASM variant.
- * When supports_subgroups is true, canonicalize thread_warp_size to 32 so
- * TIR lowering can emit subgroup shuffle reductions.
- */
-ffi::Map<ffi::String, ffi::Any> UpdateWebGPUAttrs(ffi::Map<ffi::String, ffi::Any> target) {
-  bool subgroups = false;
-  if (target.count("supports_subgroups")) {
-    subgroups = Downcast<Bool>(target.at("supports_subgroups"));
-  }
-
-  if (target.count("thread_warp_size")) {
-    int64_t thread_warp_size = Downcast<Integer>(target.at("thread_warp_size"))->value;
-    TVM_FFI_ICHECK(subgroups || thread_warp_size <= 1)
-        << "WebGPU target with thread_warp_size=" << thread_warp_size
-        << " requires supports_subgroups=true";
-  }
-
-  if (subgroups) {
-    target.Set("thread_warp_size", int64_t(32));
-  }
-  return target;
-}
 
 TVM_REGISTER_TARGET_KIND("webgpu", kDLWebGPU)
     .add_attr_option<int64_t>("max_num_threads", refl::DefaultValue(256))
--- a/web/emcc/webgpu_runtime.cc
+++ b/web/emcc/webgpu_runtime.cc
@@ -40,8 +40,6 @@
 #include "../../src/runtime/metadata.h"
 #include "../../src/runtime/workspace_pool.h"
 #include "../../src/support/bytes_io.h"
-#include "3rdparty/tvm-ffi/src/ffi/extra/json_parser.cc"
-#include "3rdparty/tvm-ffi/src/ffi/extra/json_writer.cc"
 
 namespace tvm {
 namespace runtime {


src/s_tir/transform/lower_thread_allreduce.cc
```diff
@@ -510,7 +510,9 @@ class ThreadAllreduceBuilder final : public StmtExprMutator {
         //
         // The former may cause dead lock as there is a divergent
         // branch with a warp sync call inside.
-        PrimExpr other = WarpShuffle(builtin::tvm_warp_shuffle_down(), mask_buffer, val, offset);
+        bool cast_offset_to_uint = target_->kind->name == "webgpu";
```
MasterJH5574: For changes in `src/s_tir/transform/lower_thread_allreduce.cc`, could you please add a unit test in `tests/python/tir-transform/test_tir_transform_lower_thread_all_reduce.py` to test the expected behavior?

MasterJH5574:
Also there are failing tests. Please take a look and resolve them.

yongwww:
@tvm-bot rerun

src/s_tir/transform/lower_thread_allreduce.cc
ksgr5566: added

src/target/source/codegen_webgpu.cc
```diff
@@ -118,7 +121,9 @@ void CodeGenWebGPU::InitFuncState(const PrimFunc& f) {
   }
 }
 
-CodeGenWebGPU::CodeGenWebGPU(Target target) : target_(target) {}
+CodeGenWebGPU::CodeGenWebGPU(Target target) : target_(target) {
+  enable_subgroups_ = target_->GetAttr<Bool>("supports_subgroups").value_or(Bool(false));
```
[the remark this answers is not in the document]
ksgr5566: supports_subgroups flag is the single intentional entry point, and UpdateWebGPUAttrs already sets thread_warp_size=32 when it's true. Manually setting thread_warp_size without supports_subgroups isn't a supported configuration.

src/target/target_kind.cc
```diff
@@ -424,8 +424,27 @@ TVM_REGISTER_TARGET_KIND("vulkan", kDLVulkan)
...
+ * \brief Update WebGPU target attributes based on subgroup support.
+ * When supports_subgroups is true, set thread_warp_size to 32 so that
+ * TIR lowering uses warp-level shuffle reductions instead of shared memory.
+ */
+TargetJSON UpdateWebGPUAttrs(TargetJSON target) {
+  if (target.count("supports_subgroups")) {
+    bool subgroups = Downcast<Bool>(target.at("supports_subgroups"));
+    if (subgroups) {
+      target.Set("thread_warp_size", int64_t(32));
+    }
```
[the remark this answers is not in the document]
ksgr5566: supports_subgroups flag is the single intentional entry point, and UpdateWebGPUAttrs already sets thread_warp_size=32 when it's true. Manually setting thread_warp_size without supports_subgroups isn't a supported configuration.

src/target/target_kind.cc
```diff
@@ -424,8 +424,27 @@ TVM_REGISTER_TARGET_KIND("vulkan", kDLVulkan)
...
+  return target;
+}
+
 TVM_REGISTER_TARGET_KIND("webgpu", kDLWebGPU)
     .add_attr_option<int64_t>("max_num_threads", 256)
+    .add_attr_option<bool>("supports_subgroups", false)
+    // thread_warp_size=1: is_subwarp_reduction and is_multiwarp_reduction returns false, so no subgroup ops are emitted.
+    .add_attr_option<int64_t>("thread_warp_size", 1)
+    .set_target_parser(UpdateWebGPUAttrs)
     .set_default_keys({"webgpu", "gpu"});
```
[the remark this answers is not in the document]
ksgr5566: added

src/target/target_kind.cc
```diff
@@ -427,8 +427,28 @@ TVM_REGISTER_TARGET_KIND("vulkan", kDLVulkan)
...
 
+/*!
+ * \brief Update WebGPU target attributes based on subgroup support.
+ * When supports_subgroups is true, set thread_warp_size to 32 so that
+ * TIR lowering uses warp-level shuffle reductions instead of shared memory.
+ */
+ffi::Map<ffi::String, ffi::Any> UpdateWebGPUAttrs(ffi::Map<ffi::String, ffi::Any> target) {
+  if (target.count("supports_subgroups")) {
+    bool subgroups = Downcast<Bool>(target.at("supports_subgroups"));
+    if (subgroups) {
```
akaashrp: Could you add a comment stating the following:
1. Runtime routing on the WebLLM side guarantees subgroup size == 32
2. Runtime routing on the WebLLM side guarantees maxComputeInvocationsPerWorkgroup >= 1024
3. This is intentionally constrained for the subgroup-enabled WASM variant

src/s_tir/transform/lower_thread_allreduce.cc
akaashrp, line 759: I think we should add a check of the following form here since WebGPU does not support f64, i64, or u64:
```
if (target_->kind->name == "webgpu" &&
    std::any_of(types.begin(), types.end(), [](DataType ty) {
      if (ty.is_float() && ty.bits() > 32) return true;              // f64
      if ((ty.is_int() || ty.is_uint()) && ty.bits() > 32) return true;  // i64/u64
      return false;
    })) {
  return false;
}

```

src/s_tir/transform/lower_thread_allreduce.cc
```diff
@@ -719,11 +723,11 @@ class ThreadAllreduceBuilder final : public StmtExprMutator {
   bool IsWarpReduction(const std::vector<DataType>& types, int group_extent, int reduce_extent,
                        int contiguous_reduce_extent) {
     if ((target_->kind->name != "cuda") && (target_->kind->name != "rocm") &&
-        (target_->kind->name != "metal")) {
+        (target_->kind->name != "metal") && (target_->kind->name != "webgpu")) {
       return false;
     }
```
akaashrp, line 748: Add a check here of the following form:
```
if (target_->kind->name == "webgpu" && !supports_subgroups_) {
  return false;
}
```

This is to avoid scenarios where a target such as `{"kind":"webgpu","thread_warp_size":32,"supports_subgroups":false}` would still emit subgroup ops, but the WGSL would not contain `enable subgroups;`.

src/target/source/codegen_webgpu.cc
```diff
@@ -120,7 +123,9 @@ void CodeGenWebGPU::InitFuncState(const PrimFunc& f) {
...
```
akaashrp, line 127: Let's add the following check here:
```
Bool supports_subgroups = target_->GetAttr<Bool>("supports_subgroups").value_or(Bool(false));
int64_t thread_warp_size = target_->GetAttr<Integer>("thread_warp_size", 1).value()->value;
if (thread_warp_size > 1 && !supports_subgroups) {
  LOG(FATAL) << "WebGPU target has thread_warp_size=" << thread_warp_size
             << " but supports_subgroups is false.";
}
enable_subgroups_ = supports_subgroups;
```

src/s_tir/transform/lower_thread_allreduce.cc
```diff
@@ -700,6 +700,10 @@ class ThreadAllreduceBuilder final : public StmtExprMutator {
   // Emit warp shuffle  calls.
   PrimExpr WarpShuffle(const Op& op, ffi::Optional<Buffer> mask_buffer, PrimExpr val,
                        PrimExpr delta_or_lane) {
+    // WebGPU's WGSL requires u32 for subgroupShuffle lane/delta arguments.
```
akaashrp: I believe you can move this to DispatchWebGPUShuffle by casting call->args[2] to UInt(32)

src/s_tir/transform/lower_thread_allreduce.cc
ksgr5566: CodeGenWebGPU::PrintType already handles this. I think adding one more check here is redundant.

src/s_tir/transform/lower_thread_allreduce.cc
ksgr5566: Instead of adding another WebGPU-specific branch in here, I fixed the invariant at target construction in target_kind.cc. Now that target shape is rejected up front, so this scenario never comes up in lowering/codegen. I also added a test for that in test_target_target.py.

src/target/source/codegen_webgpu.cc
ksgr5566: I added validation in UpdateWebGPUAttrs so that combination is rejected up front

akaashrp:
cc @tqchen for review

akaashrp:
@MasterJH5574 would appreciate it if you could take a look. Thanks!

MasterJH5574:
BTW just a quick note.  @ksgr5566 you can resolve all addressed comments, which helps make the PR page cleaner.

src/target/target_kind.cc
```diff
@@ -427,8 +427,41 @@ TVM_REGISTER_TARGET_KIND("vulkan", kDLVulkan)
...
+/*!
+ * \brief Update WebGPU target attributes for subgroup-enabled lowering.
+ * Runtime routing on the WebLLM side guarantees subgroup size == 32.
+ * Runtime routing on the WebLLM side guarantees
+ * maxComputeInvocationsPerWorkgroup >= 1024.
+ * This is intentionally constrained for the subgroup-enabled WASM variant.
+ * When supports_subgroups is true, canonicalize thread_warp_size to 32 so
+ * TIR lowering can emit subgroup shuffle reductions.
+ */
+ffi::Map<ffi::String, ffi::Any> UpdateWebGPUAttrs(ffi::Map<ffi::String, ffi::Any> target) {
```
MasterJH5574: Can we move this function to line ~258, since we have a lot of UpdateXXXAttrs there?

web/emcc/webgpu_runtime.cc
```diff
@@ -40,6 +40,8 @@
 #include "../../src/runtime/metadata.h"
 #include "../../src/runtime/workspace_pool.h"
 #include "../../src/support/bytes_io.h"
+#include "3rdparty/tvm-ffi/src/ffi/extra/json_parser.cc"
+#include "3rdparty/tvm-ffi/src/ffi/extra/json_writer.cc"
```
MasterJH5574: Just curious: are these two lines related to this PR, or they are for some other purpose?

MasterJH5574:
@ksgr5566 @akaashrp And I have a question that, is "subgroup support" completely decided by a flag that users set when compile a kernel, or it can be automatically detected?

web/emcc/webgpu_runtime.cc
ksgr5566: Without this I encountered a failure at at WebGPU runtime in the browser, when loading the compiled model WASM.

Upon further digging I found that the break came from the refactor that replaced `dmlc::JSONWriter` with `tvm::ffi::json::Stringify` in `webgpu_runtime.cc`. dmlc was header-only, but `tvm::ffi::json` splits declaration and implementation, so in the WASM unity build the `Stringify` implementation was not compiled into the WebGPU runtime unit and showed up as a missing runtime symbol.

These two includes are there to pull the JSON implementation into the same WASM compilation unit.

ksgr5566:
> @ksgr5566 @akaashrp And I have a question that, is "subgroup support" completely decided by a flag that users set when compile a kernel, or it can be automatically detected?

Subgroup support is decided by the user’s compile flag. Automatic device detection is not part of this compile path. It is only possible during WebGPU runtime in the browser, and WebGPU can then select the relevant compiled WASM.

web/emcc/webgpu_runtime.cc
ksgr5566: Removing as https://github.com/apache/tvm/pull/18944 takes care of it

akaashrp:
@MasterJH5574 comments should be fixed now. Just to add to what @ksgr5566 said, automatic device detection is not really possible since browser limits and features are only observed during runtime. The best we can do at the moment is to compile separate WASMs depending on capabilities and route in WebLLM or the tvm/web runtime depending on browser limits. For now, we prefer to route in WebLLM to avoid enforcing subgroup-enabled WASMs on users until we get better speedups.

PR #18833, opened 2026-02-26
[TVMScript][Relax] Use tir.SizeVar for shape variables

tqchen:
TVMScript parser now uses `tir.SizeVar` (known non-negative) instead of `tir.Var` for symbolic variables defined in size contexts (R.Tensor shapes, R.Shape), while keeping `tir.Var` for R.Prim arguments. This aligns TVMScript parsing with Relax transformations that call `Analyzer::MarkGlobalNonNegValue`, fixing shape inference differences between parsing and transformation passes


tqchen:
Closing: the SizeVar approach is too fragile — too many passes create new symbolic Var objects (fuse_ops, lambda_lift, canonicalize_bindings, bind_symbolic_vars, etc.) without preserving SizeVar, causing StructuralEqual mismatches. The existing local kernel shape reasoning (MarkGlobalNonNegValue during transformations) is sufficient.

PR #18867, opened 2026-03-03
[ADRENO] Revive and consolicate Adreno features

srkreddy1238:
Enable opencl target for gpu tests.
Consolidates all Adreno tests under tests/python/relax/backend/adreno
Changes to CLML corresponding to recent changes on json codegen/runtime. 
Docker specification for Adreno (ci_gpu + Android SDK, Gradle).


tqchen:
the consolidation looks good, maybe we can land those, then ci separately?

PR #18868, opened 2026-03-04
[Relax][TFLite] Introduce TensorFlow Lite frontend

srkreddy1238:
Verified for entire range of classification nets
Quantization is disabled at the moment
There exists few unsupported ops in conversion maps which is need to be mapped in future when relax op inventory grows.


mshr-h:
Sorry for the late review. Thanks. LGTM.

PR #18869, opened 2026-03-04
[RELAX][LAYOUT] Support multiple axis paching

srkreddy1238:
Like OIHW[4o4i] where we can pack multiple axis.
Helpful while handling complex target layouts.
This PR covers layout representation and transforms for these.


guan404ming (approved):
Thanks

tlopex (approved):
Thank you!

PR #18871, opened 2026-03-04
Batched GPU dispatch and object caching for WebGPU runtime

mitiskuma:
## Summary
- Batch compute dispatches into a single GPUCommandEncoder, flushing on sync/readback instead of per-dispatch submit to reduce JS↔GPU transition overhead during LLM decode
- Cache uniform buffers (FIFO/512), bind groups (FIFO/256), shape tuples, and pool MAP_READ staging buffers to eliminate redundant GPU object creation
- Fix padding self-assignment bug in `deviceCopyToGPU`


web/src/webgpu.ts
```diff
@@ -727,17 +812,37 @@ export class WebGPUContext {
...
-        compute.setBindGroup(0, this.device.createBindGroup({
-          layout: bindGroupLayout,
-          entries: bindGroupEntries
-        }));
+        // Bind group cache: build key from pipeline identity + buffer ptrs + uniform key
+        let bgCacheKey = pipelineId;
+        for (let i = 0; i < bufferArgIndices.length; ++i) {
+          bgCacheKey += ":" + args[bufferArgIndices[i]];
+        }
+        bgCacheKey += ":" + uniformKey;
```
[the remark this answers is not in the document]
mitiskuma: This is largely theoretical nonsense. TVM shader names are generated by the compiler and don't contain colons. Even if they did, an attacker would need to control shader names at compile time, which means they already have full code execution. This isn't a real security issue it's a cache key collision edge case that can't happen in practice with TVM-generated code.

web/src/runtime.ts
```diff
@@ -1674,8 +1679,17 @@ export class Instance implements Disposable {
...
+    if (cached !== undefined) {
+      return cached;
+    }
     const shapeArray = shape.map((value) => new Scalar(value, "int"));
-    return this.ctx.makeShapeTuple(...shapeArray);
+    const tuple = this.ctx.makeShapeTuple(...shapeArray);
+    // Detach from scope so the cached object survives across scopes.
+    this.detachFromCurrentScope(tuple);
+    this.shapeTupleCache.set(key, tuple);
+    return tuple;
```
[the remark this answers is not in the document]
mitiskuma: The number of unique shapes in an LLM model is small and bounded (typically a few dozen at most), so this cache won't grow unboundedly in practice. Adding FIFO eviction here would be over-engineering for a problem that doesn't occur in real workloads.

tqchen:
thanks for the PR. Given GPU is async, i think it is relatively ok and desirable to keep runtime simple and not containing too smart of a caching in this case. This being said, lazy submission could make sense, although would be good to see the performance measurement. Since eager also have its own merit as host compute can run in parallel with gpu compute

mitiskuma:
@tqchen thanks for the review.
Here are the benchmark results comparing Apache TVM (main) vs this branch, using the same web-llm-baseline code, only the TVM web runtime differs:
  - 48 t/s (baseline) vs 143 t/s (optimized) ~3x speedup on the same model, same web-llm-baseline, only TVM runtime differs
- TTFT dropped from 181ms to 110-130ms
- Same output, deterministic (temperature=0), also testes with different temp, similar % improvements.

Regarding keeping the runtime simple: the caching is fully contained within the WebGPU layer and doesn't affect the API surface. The caches are bounded (FIFO eviction at 512 uniform buffers and 256 bind groups), so there's no unbounded memory growth.
On eager vs lazy submission.. you're right that eager submission allows host compute to overlap with GPU compute. However, during LLM decode the CPU work between dispatches is mostly FFI/binding overhead with very little meaningful host computation to overlap. The 3x speedup suggests that the per-dispatch JS-to-GPU transition cost is the dominant bottleneck here, and batching those submissions is a net win. That said, I'm open to making the flush strategy configurable if that would be preferred.

tqchen:
Thanks for the note, seems keeping lazy is something that is helpful here! In that case i think converting to lazy by default makes sense. I think the lazy launch and staging buffer are reasonably less controversial and we could land them if we find benefit. 

Caching and strategy of caching is something that is interesting and would be good to disentangle a bit.
 
Do we know if we only do lazy but not caching would have similar impact? Also it is interesting to learn about model used, since 142 t/s sounds like a reasonably small model. just to understand the state

mitiskuma:
I think we'd lose a 10-15% of the performances without caching, I can get bencharmarks if you think it's something to explore. meanwhile I get you few models comparison vs base:
Qwen3 0.6B | base: 50t/s vs improved: 150t/s
Qwen3 1.7B | base: 49t/s vs improved: 117t/s
Qwen3 4B | base: 34t/s vs improved:   69t/s

tqchen:
thanks @mitiskuma , mainly from design pov:

- A0: i think we should be able to get  lazy + staging buffer in, these are great changes and i think can get polished clean version, always lazy is good if we find that is the best approach. 
- A1: The caching logic i think can needs a bit deliberation, main in terms of how we ideally structure it. If we have something like an explicit CacheState and clearly think through some of the hash collison scenario to make them robust, i think we can also bring some in. 

From the operational pov, likely splitting the two into two PRs would make sense. Likely we can focus on landing A0 soon, then working on A1 if we find it being useful

mitiskuma:
@tqchen thank you, let me know how to proceed. if you prefer handling on your end, or if you want me to remove the caching, keep lazy.
and handle that caching on a follow up PR.

mitiskuma:
@tqchen ok after more benchmarking we lose about 50% without caching. The main issue is that with lazy submission, each dispatch in a batch needs its own uniform buffer and bind group. We can't reuse a shared one because queue.writeBuffer executes immediately while the compute passes are deferred in the encoder.
We recover some of the loss (~60-70%) by pooling uniform buffers per dispatch index and reusing them across flushes. But createBindGroup() per dispatch is still expensive and that's where the remaining gap comes from.
So the caching isn't just a nice to have optimization, it's closely tied to making lazy submission performant imo.
The uniform buffer cache avoids per-dispatch allocation, and the bind group cache avoids redundant createBindGroup calls for repeated kernel signatures (which is common as you know in transformer inference where the same kernels run every token). Would it work to keep the caching in this PR since it's a correctness requirement for lazy submission to not regress? Or would you prefer we land lazy submission with the pool-based workaround (accepting ~35% regression vs full caching)?

tqchen:
I think starting with a uniform pool + lazy seems to be a good first step as that strictly improves over what we had. My understanding is we can always flush the queue when we run out of uniform pool. If you can send a PR that starts with uniform pool + lazy it would be great.

I get how uniform buffer cache helps to reduce things further, and would be happy we explore that as a followup PR. Maybe one idea is that we have introduce a CacheState class that provides the relevant functionalities (wrt to cache key computation with comments on correctness, invalidation scenarios).

Thanks @mitiskuma for looking into this. I think the two phase approach could give us more confidence here and also have a bit more thinkings to make the CacheState also robust.

mitiskuma:
I agree with you. let me know if you want to discuss CacheState elsewhere, I'm open to it.
let me know what you think about the uniform buffer pool, pretty straightforward.

web/src/runtime.ts
```diff
@@ -954,6 +956,9 @@ export class Instance implements Disposable {
   dispose(): void {
     this.deviceLostIsError = false;  // prevent dispose to trigger device.lost error
     // order matters
+    // dispose cached shape tuples before ctx
+    this.shapeTupleCache.forEach((obj) => obj.dispose());
```
tqchen: consider introduce cache_state.ts which include a LRUCache that is generic, with the folliwing method
- get(key, lambda constructor) that allows LRU cache with key, and if not hit, use lambda to construct, and do eviction if exceeding LRU limit and return the result
- invalidate() clear all existing cache items
- needEviction(key) => poll and return whether eviction is needed, since we may need to have followup items like flush


CacheState class can contain members shape_cache, uniform_cache.
- compute_shape_key()
- compute_uniform_key()

at each variable document:

- Why we need the cache
- Rule of invalidation: e.g. shape cache do not need invalidation, uniform cache needs invalidation during any de-allocation

web/src/webgpu.ts
```diff
@@ -402,10 +401,21 @@ export class WebGPUContext {
...
   private bufferTableFreeId: Array<number> = [];
-  private podArgStagingBuffers: Array<GPUBuffer> = [];
   private canvasRenderManager?: CanvasRenderManager = undefined;
-  // number of pod arg staging buffers
-  private maxNumPodArgsStagingBuffers = 2;
+  // Pool of MAP_READ staging buffers to avoid per-copy create/destroy overhead
+  private readStagingBufferPool: Array<{ buffer: GPUBuffer; size: number }> = [];
+  private maxReadStagingBuffers = 4;
+  // Pending mapAsync promise from the last GPU→CPU copy (faster than onSubmittedWorkDone)
+  private pendingRead: Promise<void> | null = null;
```
tqchen: clarity: pendingRead=> pendingGPUToCPUCopy

web/src/webgpu.ts
```diff
@@ -426,26 +436,54 @@ export class WebGPUContext {
...
 
+
   /**
    * Wait for all pending GPU tasks to complete
    */
   async sync(): Promise<void> {
-    await this.device.queue.onSubmittedWorkDone();
+    // Flush any batched compute passes before waiting on the queue.
+    this.flushCommands();
+    if (this.pendingRead) {
```
tqchen: semantics clairification, if there is pendingGPUToCPUCopy and we have extra operations in queue after that, will we sync to these operations? I know that from effect pov it does not matter, but from the semantics it maybe helpful to clarify, e.g. one approach would be we only use pendingGPUToCPUCopy iff the reads is the last op. In that case, maybe we need some trackings

web/src/webgpu.ts
```diff
@@ -695,9 +704,28 @@ export class WebGPUContext {
...
-        // push pod buffer
+        // Each dispatch in a batch needs its own uniform buffer since the
+        // encoder defers submission. We maintain a reusable pool indexed by
+        // dispatch position — buffers grow as needed but are never destroyed.
         const sizeOfI32 = 4;
-        const podArgBuffer = this.getPodArgsBuffer((podArgIndices.length + 1) * sizeOfI32);
+        const bufBytes = (podArgIndices.length + 1) * sizeOfI32;
+        const dispatchIdx = this.pendingDispatchCount++;
+        let podArgBuffer: GPUBuffer;
+        if (dispatchIdx < this.uniformBufferPool.length &&
```
tqchen: seems we are misisng when dispatchIdx runs out of `this.uniformBufferPool.length`, at whihc pt we should flush (seems to be missing here)?. This part of code should be a subfunctio. getUniformFromPool. We need to have comment on state transition impact if we flush (e.g. if there were stale state that was created before this pt)

web/src/webgpu.ts
```diff
@@ -426,26 +436,54 @@ export class WebGPUContext {
     this.device = device;
   }
 
+  /**
+   * Flush all pending compute passes by finishing and submitting the
+   * accumulated command encoder. Call before any operation that reads
```
tqchen: list the operatios

- read back GPU data
- wait on queue
- dealloc

web/src/webgpu.ts
```diff
@@ -880,17 +916,51 @@ export class WebGPUContext {
     );
   }
 
+  /**
+   * Get a MAP_READ staging buffer from the pool, or create one if none fits.
+   * Uses first-fit-by-size: returns the first pooled buffer >= nbytes.
+   */
+  private getReadStagingBuffer(nbytes: number): GPUBuffer {
+    for (let i = 0; i < this.readStagingBufferPool.length; i++) {
```
tqchen, line 957: getOrCreateReadStagingBuffer

web/src/webgpu.ts
```diff
@@ -880,17 +916,51 @@ export class WebGPUContext {
...
+      size: nbytes,
+      usage: GPUBufferUsage.MAP_READ | GPUBufferUsage.COPY_DST,
+    });
+  }
+
+  /**
+   * Return a MAP_READ staging buffer to the pool for reuse.
+   * Evicts the smallest buffer if the pool is full.
+   */
+  private returnReadStagingBuffer(buf: GPUBuffer): void {
```
tqchen: recycleReadStagingBuffer

tqchen:
Thanks @mitiskuma , I did a round of review through the current code, mainly on readability and clarity. I think we should be able to land it then use it as basis for next one

web/src/runtime.ts
```diff
@@ -1674,8 +1679,17 @@ export class Instance implements Disposable {
...
+      return cached;
+    }
     const shapeArray = shape.map((value) => new Scalar(value, "int"));
-    return this.ctx.makeShapeTuple(...shapeArray);
+    const tuple = this.ctx.makeShapeTuple(...shapeArray);
+    // Detach from scope so the cached object survives across scopes.
+    this.detachFromCurrentScope(tuple);
+    this.shapeTupleCache.set(key, tuple);
+    return tuple;
   }
```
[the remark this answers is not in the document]
mitiskuma: Shape tuples are internal objects managed by the runtime. users never manually dispose them.   
  They're detached from scope specifically so the cache owns their lifecycle. The LRU eviction in the new code also handles disposal. Not a real risk but probably worth noting.

web/src/webgpu.ts
```diff
@@ -903,11 +973,15 @@ export class WebGPUContext {
...
+      const data = gpuTemp.getMappedRange(0, nbytes);
       this.memory.storeRawBytes(to, new Uint8Array(data));
-      gpuTemp.destroy();
+      this.returnReadStagingBuffer(gpuTemp);
     });
+    // Chain with any existing pending read so sync() awaits all of them.
+    this.pendingRead = this.pendingRead
+      ? this.pendingRead.then(() => readPromise)
+      : readPromise;
   }
```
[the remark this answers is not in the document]
mitiskuma: 1. Out-of-order writes to same address: Not a real concern, TVM never issues overlapping readbacks to the same Wasm address concurrently. Each deviceCopyFromGPU targets a different destination.
2. Rejected promise breaking the chain: This is valid. If mapAsync rejects (device loss), the chain breaks. However, device loss is already fatal, device.lost triggers shutdown. So this is a theoretical concern that doesn't manifest in practice.

tqchen:
pls fix lint, we can aim to land it, would be good to cross check e2e perf and correctness of this PR, cc @akaashrp 


thanks @mitiskuma

PR #18880, opened 2026-03-06
[TOPI] Reject non-float inputs for inverse unary math ops

### LINKED ISSUE #18729: [Bug] topi.acos accepts integer input and crashes in LowerIntrin with InternalError
### Actual behavior

Calling `opi.acos` with an integer-typed input tensor (e.g. int16) leads to a `tvm.error.InternalError` during LowerIntrin / IntrinInjecter in the TIR build pipeline.
Instead of reporting a user-facing dtype validation error (or inserting an implicit cast), TVM crashes while constructing an IntImm<int16> with a value outside the valid range.

This appears to be missing dtype validation in TOPI for floating-point–only math intrinsics.

### Environment
tvm:

### Steps to reproduce
```python
import tvm
from tvm import te, topi, tir
import numpy as np

def model_func(x):
    return topi.acos(x)

x = te.placeholder((1, 128, 256), dtype="int16", name="x")
out = model_func(x)

func = te.create_prim_func([x, out]).with_attr(
    "target", tvm.target.Target("llvm")
)

sch = tir.Schedule(func)
np_input = np.random.uniform(-1, 1, size=(1, 128, 256)).astype("int16")
with tvm.transform.PassContext(opt_level=0):
    mod = tvm.build(sch.mod, target="llvm")

```

### Triage
* lowering
* TOPI
* needs-triage

LudovicoYIN:
## Summary

Reject non-float inputs for inverse trigonometric and hyperbolic unary ops in TOPI.

## Changes

- add a shared floating-point dtype check for inverse unary math ops in TOPI
- apply the check to `topi.acos`, `topi.acosh`, `topi.asin`, `topi.asinh`, and `topi.atanh`
- add TE tests covering integer-input rejection for these ops
- add regression tests covering successful LLVM build for both `float32` and `bfloat16`

## Validation

- `tests/python/te/test_te_create_primfunc.py -k 'topi_float_unary'`
- local repro now fails early with a clear `TypeError` for integer inputs
- local regression check confirms the valid `float32` and `bfloat16` paths still compile with LLVM

## Issue

Fixes #18729

### DIFF
--- a/python/tvm/topi/math.py
+++ b/python/tvm/topi/math.py
@@ -235,6 +235,7 @@ def acosh(x):
     y : tvm.te.Tensor
         The result.
     """
+    x = _require_float_tensor("acosh", x)
     return te.compute(x.shape, lambda *i: te.acosh(x(*i)))
 
 
@@ -270,6 +271,7 @@ def asinh(x):
     y : tvm.te.Tensor
         The result.
     """
+    x = _require_float_tensor("asinh", x)
     return te.compute(x.shape, lambda *i: te.asinh(x(*i)))
 
 
@@ -304,6 +306,7 @@ def atanh(x):
     y : tvm.te.Tensor
         The result.
     """
+    x = _require_float_tensor("atanh", x)
     return te.compute(x.shape, lambda *i: te.atanh(x(*i)))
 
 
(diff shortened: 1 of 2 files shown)


LudovicoYIN:
I extended the same float-only validation to `topi.acosh`, `topi.asinh`, and `topi.atanh` as well, and expanded the tests to cover both rejection of integer inputs and successful build for `float32` and `bfloat16`.

PR #18884, opened 2026-03-06
[FIX][LLVM] Use isCPUStringValid for mcpu validation instead of enumerating processor descriptions

gnguralnick:
## Summary

Fix false rejection of `apple-m1`, `apple-m2`, and `apple-m3` as LLVM CPU names when building TVM with LLVM 22+.

## Behavior

After following the [installation from source instructions](https://tvm.apache.org/docs/install/from_source.html) and building against LLVM 22, every `import tvm` produces spurious error messages:

```
Error: Using LLVM 22.1.0 with `-mcpu=apple-m1` is not valid in `-mtriple=arm64-apple-macos`, using default `-mcpu=generic`
Error: Using LLVM 22.1.0 with `-mcpu=apple-m2` is not valid in `-mtriple=arm64-apple-macos`, using default `-mcpu=generic`
```

These are triggered by the Metal target tag registrations in `python/tvm/target/tag_registry/metal.py`, which use `apple-m1` and `apple-m2` as the host `-mcpu`. The CPUs are silently downgraded to `generic`.

## Root cause

LLVM 22 reorganized its AArch64 processor table. `apple-m1` through `apple-m3` are now CPU **aliases** — fully valid and accepted by `createTargetMachine` and `isCPUStringValid()`, but no longer returned by `MCSubtargetInfo::getAllProcessorDescriptions()`.

TVM's `LLVMTargetInfo` constructor validates `-mcpu` by enumerating `getAllProcessorDescriptions()` and checking membership, so it misses alias-only names.

## Fix

Replace the enumeration-based check with a new `IsValidCPU()` method that uses `MCSubtargetInfo::isCPUStringValid()`, which correctly handles both primary names and aliases. This API has been available since at least LLVM 7, well before TVM's minimum supported version.

## Validation

- Built and tested on macOS (Apple Silicon) with LLVM 22.1.0
- `python -c "import tvm; print(tvm.__file__)"` produces clean output with no error messages

### DIFF
--- a/src/target/llvm/llvm_instance.cc
+++ b/src/target/llvm/llvm_instance.cc
@@ -443,12 +443,18 @@ llvm::TargetMachine* LLVMTargetInfo::GetOrCreateTargetMachine(bool allow_missing
 bool LLVMTargetInfo::IsValidCPU(const std::string& cpu) const {
   auto llvm_instance = CreateLLVMTargetInstance(triple_, true);
   if (!llvm_instance) return false;
-  auto tm = CreateLLVMTargetMachine(llvm_instance, triple_, "", "");
-  if (!tm) return false;
-  const auto* MCInfo = tm->getMCSubtargetInfo();
-  // Use isCPUStringValid which correctly handles CPU aliases (e.g. apple-m1
-  // in LLVM 22+) that don't appear in getAllProcessorDescriptions().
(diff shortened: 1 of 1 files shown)


tqchen:
thanks, please check if the gemini suggestion makes sense

tqchen:
thanks @gnguralnick

PR #18890, opened 2026-03-08
[runtime][vm] Add LoRA adapter metadata to paged KV cache

MagellaX:
## Summary

This PR adds minimal runtime support for LoRA adapter metadata in paged KV cache state.

This is intended as a small runtime building block for downstream multi-LoRA serving work discussed here:
- https://github.com/mlc-ai/mlc-llm/issues/3446
- https://github.com/mlc-ai/mlc-llm/pull/3281

## Changes

- add `lora_adapter_id` to `Sequence`
- add runtime KV-cache APIs to:
  - set sequence adapter id
  - get sequence adapter id
  - get current-batch adapter ids after `BeginForward`
- make `ForkSequence` inherit the parent adapter id
- populate current-batch adapter ids during `BeginForward`
- add Python runtime tests covering:
  - default adapter id
  - set/get
  - fork inheritance
  - current-batch ordering

## Scope

This is runtime-state plumbing only.

This PR does not yet add:
- frontend request plumbing
- LoRA runtime execution in model ops
- multi-adapter batching logic
- specialized LoRA kernels

## Validation

Validated on a fresh `apache/tvm` base with a local source build with Modal...

### DIFF
--- a/src/runtime/vm/paged_kv_cache.cc
+++ b/src/runtime/vm/paged_kv_cache.cc
@@ -1221,12 +1221,7 @@ class PagedAttentionKVCacheObj : public AttentionKVCacheObj {
   }
 
   ffi::Shape GetCurrentLoraAdapterIds() final {
-    std::vector<ffi::Shape::index_type> adapter_ids;
-    adapter_ids.reserve(current_lora_adapter_ids_host_.size());
-    for (size_t i = 0; i < current_lora_adapter_ids_host_.size(); ++i) {
-      adapter_ids.push_back(current_lora_adapter_ids_host_[i]);
-    }
-    return ffi::Shape(std::move(adapter_ids));
+    return current_lora_adapter_ids_host_.as_int_tuple();
   }
 
   ffi::Shape DisaggPrepareRecv(int64_t seq_id, int append_length) final {


MagellaX:
@babusid lmk your thoughts!!!

babusid:
> @babusid lmk your thoughts!!!

I've been mulling over what the best architecture for Multi-LoRA support might look like, and I don't think the original breakdown I gave you was the best. Sorry about that.

While the S-LoRA paper does use paged memory blocks to handle the adaptor weights, given that TVM's construct is explicitly a `PagedKVCache`, I don't think overloading it to hold adaptor weights is the cleanest. 

I'm wondering if we're not better served by implementing a separate memory pool construct, in the same vein as PagedKVCache.cc or RNNState.cc, designed specifically for holding LoRA adaptor weights at runtime. 

That said, by restructuring the efforts, we might be able to keep things strictly downstream (not completely sure on this last point).

Let me discuss a bit more internally with @MasterJH5574 and others, and see if I can better formalize an idea / design.
I also welcome your thoughts on the topic.

MagellaX:
> > @babusid lmk your thoughts!!!
> 
> I've been mulling over what the best architecture for Multi-LoRA support might look like, and I don't think the original breakdown I gave you was the best. Sorry about that.
> 
> While the S-LoRA paper does use paged memory blocks to handle the adaptor weights, given that TVM's construct is explicitly a `PagedKVCache`, I don't think overloading it to hold adaptor weights is the cleanest.
> 
> I'm wondering if we're not better served by implementing a separate memory pool construct, in the same vein as PagedKVCache.cc or RNNState.cc, designed specifically for holding LoRA adaptor weights at runtime.
> 
> That said, by restructuring the efforts, we might be able to keep things strictly downstream (not completely sure on this last point).
> 
> Let me discuss a bit more internally with @MasterJH5574 and others, and see if I can better formalize an idea / design. I also welcome your thoughts on the topic.

yeah, after digging through S-LoRA/Punica again and also looking at how the actual serving stacks handle this, I think you’re probably right....

The consistent pattern seems to be adapter identity may need to flow with request / cache correctness, but adapter weights themselves usually live in a separate adapter cache / pool, not inside the KV-cache object. S-LoRA’s unified paging reads more like a lower-level allocator/kernel strategy than the right API boundary. TRT-LLM is especially explicit about this since it keeps a LoRA cache distinct from the KV cache, and vLLM / SGLang also seem to converge on separate adapter management with adapter ids still affecting cache correctness.

So my current leaning is to avoid pushing actual LoRA weights into `PagedKVCache`, and instead re-scope around a separate `LoRAPool` / adapter-pool runtime object, with adapter id threaded separately. Then if unified paging still makes sense later, it can sit underneath both KV and adapter storage rather than overloading the KV abstraction.

MasterJH5574:
@MagellaX Thank you so much for the contributions!  My overall read is that we probably need to first establish end-to-end LoRA serving flow, with runnable tests and real commands, before upstreaming parts.  The main reason is that we don't want to iterate over the implementations for too many times in the mainline repo without seeing end-to-end effects.

MagellaX:
> @MagellaX Thank you so much for the contributions! My overall read is that we probably need to first establish end-to-end LoRA serving flow, with runnable tests and real commands, before upstreaming parts. The main reason is that we don't want to iterate over the implementations for too many times in the mainline repo without seeing end-to-end effects.

yeah totally agree with that!!! i think with this PR the main useful outcome was clarifying the runtime boundary a bit, but I agree the next step should be downstream first, not upstream first. I’ll focus on getting a minimal end-to-end LoRA serving path working in MLC with real runnable commands, tests, and a clear single-adapter flow, and then only upstream the smallest TVM pieces that are actually required by that working path. That should make the design much easier to evaluate and avoid churning upstream abstractions too early.

MagellaX:
let me know!!!

PR #18907, opened 2026-03-16
[Relax][Frontend][Torch] Fix squeeze to return no-op for non-unit dims

### LINKED ISSUE #18442: [Bug] InternalError: Squeeze dimension check too strict compared to PyTorch behavior
### Description
When converting a PyTorch model containing `squeeze` operation on a dimension that is not 1, TVM fails with an InternalError. PyTorch's `squeeze` operation silently ignores dimensions that are not 1, but TVM performs strict checking and requires the dimension to be exactly 1.

### Expected behavior
The PyTorch model with `squeeze` on non-1 dimensions should be successfully converted to TVM Relax module, matching PyTorch's behavior of silently ignoring such dimensions.

### Actual behavior
An InternalError occurs during `from_exported_program` conversion with the message `Squeeze expects the input tensor shape values at the given axis positions to be all 1. However, the tensor shape at axis 1 is T.int64(10) which is not 1.`, indicating that TVM's squeeze implementation is stricter than PyTorch's.

### Environment

* **OS:** Ubuntu 20.04.6 LTS
* **TVM version:** 0.23.dev0
*  **Python version:** 3.11.14

### Steps to reproduce

```python
import torch
import torch.nn as nn
import tvm
from tvm import relax

class SqueezeModel(nn.Module):
    def __init__(self):
        super().__init__()

    def forward(self, x):
        return x.squeeze(1)  # This works in PyTorch even if dim=1 != 1

model = SqueezeModel()
model.eval()

# Create tensor where dim=1 is not 1
x = torch.randn(32, 10, 5)  # shape [32, 10, 5]

# PyTorch execution works (squeeze is ignored when dim != 1)
with torch.no_grad():
    output = model(x)
print(f"PyTorch output shape: {output.shape}")  # [32, 10, 5]

# PyTorch export works  
exported_program = torch.export.export(model, (x,))

# TVM conversion fails
mod = from_exported_program(exported_program)  # InternalError here
```

### Error Log
```
  File "test.py", line 33, in <module>
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  ...

tvm.error.InternalError: Squeeze expects the input tensor shape values at the given axis positions to be all 1. However, the tensor shape at axis 1 is T.int64(10) which is not 1. If it is symbolic, please use MatchCast to cast it to 1 before doing Squeeze.
```

### Triage

* needs-triage
* bug
* frontend: pytorch

Aharrypotter:
### What problem does this PR solve?

Fixes #18442

When using `squeeze(dim)` with a dimension that is not size 1, PyTorch silently returns the original tensor unchanged (a no-op). The TVM Relax PyTorch frontend was not replicating this behavior — it forwarded `dim` unconditionally to `relax.op.squeeze`, which raised an `InternalError` in TVM ≤ 0.23 and still emits a redundant no-op squeeze op on the current main branch.

**Reproduction:**
\`\`\`python
import torch, torch.nn as nn

class SqueezeModel(nn.Module):
    def forward(self, x):
        return x.squeeze(1)  # dim=1 has size 10, not 1

x = torch.randn(32, 10, 5)
ep = torch.export.export(SqueezeModel().eval(), (x,))
mod = from_exported_program(ep)  # previously: InternalError
\`\`\`

Two bugs fixed in `BaseFXGraphImporter._squeeze`:
1. **Single-dim case**: no check on the actual dimension size before passing to `relax.op.squeeze`
2. **List/tuple dims case**: existing filter only checked array bounds, not whether each axis is actually size 1

**Fix**: for statically-known non-unit dimensions, return the input tensor directly. Dynamic/symbolic dimensions pass through conservatively as before.

### Tests

- `Squeeze2` (no-arg `squeeze()`): expected axis filter corrected from `[0,1,2,3]` to `[1,3]`
- `Squeeze3` (`squeeze(2)` on non-unit dim): now expects no squeeze op emitted
- `Squeeze4` (new): exact repro of #18442 — `squeeze(1)` on shape `(32,10,5)`

### Checklist
- [x] PR title starts with a category tag
- [x] All existing and new tests pass
- [x] Fixes a confirmed bug (linked issue: #18442)

### DIFF
--- a/python/tvm/relax/frontend/torch/base_fx_graph_translator.py
+++ b/python/tvm/relax/frontend/torch/base_fx_graph_translator.py
@@ -2133,18 +2133,34 @@ class BaseFXGraphImporter(metaclass=abc.ABCMeta):
         if dim is None:
             dim = node.kwargs.get("dims", None)
 
-        # If dims is a list, filter out axes where dimension is not 1
-        # This is needed because PyTorch decomposition may pass all axes
+        shape = self.shape_of(x)
+        ndim = len(shape)
+
         if isinstance(dim, list | tuple) and len(dim) > 0:
-            shape = self.shape_of(x)
-            # Filter to only include axes where the dimension is 1
+            # PyTorch decomposition may pass all axes. Filter to only include axes
+            # where the dimension is statically known to be 1, or is symbolic (dynamic).
+            # Axes where the dimension is statically known to be != 1 are silently
+            # ignored, matching PyTorch's behavior.
             valid_dims = []
             for d in dim:
-                axis = d if d >= 0 else len(shape) + d
-                if axis < len(shape):
-                    valid_dims.append(d)
-            # If no valid dims, use None to squeeze all size-1 dimensions
-            dim = valid_dims if valid_dims else None
+                axis = d if d >= 0 else ndim + d
+                if 0 <= axis < ndim:
+                    dim_val = shape[axis]
+                    # Include axis if it's dynamic (not a static integer) or is size 1
+                    if not isinstance(dim_val, tir.IntImm) or dim_val.value == 1:
+                        valid_dims.append(d)
+            # If no axes will actually be squeezed, return the original tensor (PyTorch no-op)
+            if not valid_dims:
+                return x
+            dim = valid_dims
+        elif dim is not None:
+            # Single-dim case: mimic PyTorch behavior — if the dimension is statically
+            # known to not be 1, squeezing is a no-op, so return the original tensor.
+            axis = dim if dim >= 0 else ndim + dim
+            if 0 <= axis < ndim:
+                dim_val = shape[axis]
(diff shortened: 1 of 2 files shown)


tlopex:
Hi @Aharrypotter Thanks for the contribution!
I think this bug is already fixed, there is no error with the code in the reproduction.

tlopex:
cc @mshr-h @guan404ming

Aharrypotter:
> Hi @Aharrypotter Thanks for the contribution! I think this bug is already fixed, there is no error with the code in the reproduction.

 Thanks for the context! After looking at #18478, I can see it already fixes both the single-dim and list/tuple cases at the C++ level (manipulate.cc + topi/transform.h), so this PR is indeed redundant. I'll close it.
  Appreciate the quick feedback.

PR #18912, opened 2026-03-18
[TIR] Handle Bind in LowerDeviceKernelLaunch

MasterJH5574:
DeviceInfoCollector did not track Bind statements, so when CSE (or
any other pass) inserted a Bind before a thread_extent AttrStmt, the
collected extent referenced a locally-bound variable instead of
function parameters.  LowerDeviceKernelLaunch then produced dangling
references in the host function.

Fix: record Bind definitions in DeviceInfoCollector and inline them
when extracting thread_extent values and dynamic shared memory sizes.

### DIFF
--- a/src/tirx/transform/lower_device_kernel_launch.cc
+++ b/src/tirx/transform/lower_device_kernel_launch.cc
@@ -104,6 +104,17 @@ class DeviceInfoCollector : public StmtVisitor {
     return extent.value();
   }
 
+  void VisitStmt_(const BindNode* op) final {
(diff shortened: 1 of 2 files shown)


tqchen:
would be good to cross check and deep deeper to see if lowe `LowerDeviceKernelLaunch` should be made aware of the Binds

MasterJH5574:
@tqchen @tlopex I just updated the PR. It turns out that the LowerDeviceKernelLaunch pass is the right place to fix the issue, and the CSE changes have been reverted. The PR description and title have already been updated as well.

PR #18914, opened 2026-03-19
[Vulkan] Avoid explicit layout decoration on non-interface allocations

kabu1204:
SPIR-V codegen currently emits `ArrayStride` and `Offset` decorations for non-interface allocations in `GetStructArrayType()`. That is correct for descriptor-backed interface blocks, but not for static workgroup allocations. 

I hit this while bringing up tilelang vulkan shared memory allocation path: the vulkan validation rejected shaders that used shared memory lowered through this path:
```
tvm.error.InternalError: Check failed: res == SPV_SUCCESS (-10 vs. 0) :
index=44 error:[VUID-StandaloneSpirv-None-10684] Invalid explicit layout decorations on type for operand '25[%_ptr_Workgroup__struct_24]'
  %A_shared = OpVariable %_ptr_Workgroup__struct_24 Workgroup
```

FIX: This PR keeps layout decoration for interface blocks, and skips for non-interface allocations such as static shared/workgroup memory. A new compile-only test is added for this.

One possible concern is that there's already a pre-existing test using `fetch_to_shared`.

### DIFF
--- a/tests/python/codegen/test_target_codegen_vulkan.py
+++ b/tests/python/codegen/test_target_codegen_vulkan.py
@@ -23,7 +23,6 @@ import pytest
 
 import tvm
 import tvm.testing
-from tvm import te, tir
 from tvm.script import ir as I
 from tvm.script import tir as T
 from tvm.script.ir_builder import IRBuilder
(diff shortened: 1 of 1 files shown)


tests/python/codegen/test_target_codegen_vulkan.py
```diff
@@ -515,6 +516,28 @@ def kernel():
     vulkan_codegen(Module, target)
 
 
+@tvm.testing.requires_vulkan(support_required="compile-only")
+def test_codegen_static_shared_memory():
+    """The codegen should accept static shared/workgroup allocations."""
+
+    A = te.placeholder((128,), name="A", dtype="float32")
```
tqchen: please use tvmscript instead of schedule for tests

kistenklaus:
#18915 is related, but also points out another vulkan codegen bug.

PR #18917, opened 2026-03-20
[TOPI] Fix strided_slice_with_axes to handle negative axis values

mshr-h:
Negative axis values (e.g., `axes=[-1]`) in `strided_slice_with_axes` were used directly as array indices without normalization, causing an `IndexError` during `LegalizeOps`.

This PR normalizes negative axes to positive equivalents before passing them to `StridedSliceCanonicalizeBegin`, `StridedSliceOutputShape`, and the compute lambda.


mshr-h:
cc @tlopex @guan404ming

PR #18918, opened 2026-03-20
[Relax][PyTorch] Fix _slice and _expand for dynamic shapes in PyTorch ExportedProgram frontend

mshr-h:
Fixes two issues when translating PyTorch models with dynamic shapes:

1. **_slice**: Resolve `fx.Node` references in start/end/step arguments and detect identity slices where the symbolic end equals the tensor dimension (avoids redundant `strided_slice` ops).

2. **_expand**: Fall back to FX node metadata when `shape_of()` returns `None` for tensors with unknown shapes.


mshr-h:
cc @tlopex @guan404ming

tlopex (approved):

PR #18925, opened 2026-03-23
fix: add safety warning to pickle_memoize cache loading

scruge1:
## Summary
`pickle_memoize` loads cached pickle files via `pickle.load()` without any integrity verification or user warning. If an attacker can write to the cache directory, they can inject malicious pickle payloads that execute arbitrary code on next load.

## Fix
Adds a `UserWarning` when loading pickle cache files to alert users about the security risk.

## Related
Huntr security vulnerability report (CWE-502: Deserialization of Untrusted Data)

### DIFF
--- a/python/tvm/contrib/pickle_memoize.py
+++ b/python/tvm/contrib/pickle_memoize.py
@@ -23,6 +23,7 @@ import functools
 import os
 import pathlib
 import sys
+import warnings
 
 try:
     import cPickle as pickle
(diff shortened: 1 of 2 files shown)


tqchen:
given pickle is used as part of our code and our security model assumes trusted data for pickle memoize that is only used for trusted data in testcase, we would prefer not having a warning in such case (just like pickle itself won't have warnings).

scruge1:
Understood — makes sense that pickle_memoize operates within TVM's trusted-data security model. Closing this PR. Thank you for the review.

scruge1:
Understood — the revert in #18926 is appropriate given TVM's trusted-data security model. Thank you for the consideration.

PR #18927, opened 2026-03-25
feat(meta_schedule): expand CUDA unroll steps for SM70 optimization

jianhua1724:
## Motivation
Expand CUDA unroll search space to support optimal loop unrolling sizes for SM70 (V100) GPUs.
Add missing critical unroll steps: 32, 128, 256.

## Changes
Only modify 1 line:
Extend `unroll_max_steps` in `ScheduleRule::DefaultCUDA()`
From: {0, 16, 64, 512, 1024}
To:   {0, 16, 32, 64, 128, 256, 512, 1024}

## Benefits
- 5%~15% performance improvement on SM70 (V100)
- No breaking changes
- Backward compatible with all CUDA architectures
- Only expand search space, no logic modification

## Test
Compiled and tested on SM70 (V100), no regression.

### DIFF
--- a/src/s_tir/meta_schedule/schedule_rule/schedule_rule.cc
+++ b/src/s_tir/meta_schedule/schedule_rule/schedule_rule.cc
@@ -166,7 +166,7 @@ ffi::Array<ScheduleRule> ScheduleRule::DefaultCUDA() {
       ScheduleRule::ParallelizeVectorizeUnroll(
           /*max_jobs_per_core=*/-1,
           /*max_vectorize_extent=*/-1,
-          /*unroll_max_steps=*/ffi::Array<Integer>{0, 16, 64, 512, 1024},
+          /*unroll_max_steps=*/ffi::Array<Integer>{0, 16, 32, 64, 128, 256, 512, 1024},
           /*unroll_explicit=*/true),
       ScheduleRule::AutoBind(
           /*max_threadblocks=*/256,


jianhua1724:
> Thanks for the contribution. The motivation makes sense, but we need concrete data before merging a change to the default search space that affects all CUDA users:
> 
> 1. Which model/workload did you benchmark? Please share actual latency numbers (before vs after) on V100.
> 2. The search space grows from 5 to 8 candidates. Did you measure the tuning time increase?
> 3. "No regression" on other architectures — does this mean you tested on non-SM70 GPUs, or just that the code compiles?


Thank you for your reply. I will answer your three questions as follows:
My scenario is to accelerate the model training of algorithm engineers on the V100 GPU (and inference acceleration will be supported later).
The computationally heavy network structure in the model is a feed-forward network with fixed input shapes, such as the following:

batch_size = 1024
input_size = 256
hidden_size = 512
output_size = 256
class DeepFFN(nn.Module):
    def __init__(self):
        super().__init__()
        self.fc1 = nn.Linear(input_size, hidden_size, dtype=torch.float16)
        self.relu = nn.ReLU()
        self.fc2 = nn.Linear(hidden_size, output_size, dtype=torch.float16)

    def forward(self, x):
        return self.fc2(self.relu(self.fc1(x)))


I defined three operators to replace PyTorch's native implementation for training acceleration. 
The three operators are: one for the forward propagation of the network, and two for the backward propagation logic (the backward propagation is complex, so I split it into two operators).
Then I used TVM's MetaSchedule to compile the three operators and finally replaced PyTorch's native implementation.

1.Test data:
1.1 Speed difference of a single operator: 
The following is the test data of the forward propagation operator in the two versions:
Version with "unroll_max_steps=ffi::Array<Integer>{0, 16, 64, 512, 1024}":

2026-04-13 12:33:33 [INFO] [task_scheduler.cc:326]
 ID | Name |      FLOP | Weight | Speed (GFLOPS) | Latency (us) | Weighted Latency (us) | Trials | Done
--------------------------------------------------------------------------------------------------------
  0 | main | 537657344 |      1 |     14873.9572 |      36.1476 |               36.1476 |    100 |    Y
--------------------------------------------------------------------------------------------------------
Total trials: 100
Total latency (us): 36.1476


Version with "unroll_max_steps=ffi::Array<Integer>{0, 16, 32, 64, 128, 256, 512, 1024}":

 ID | Name |      FLOP | Weight | Speed (GFLOPS) | Latency (us) | Weighted Latency (us) | Trials | Done
--------------------------------------------------------------------------------------------------------
  0 | main | 537657344 |      1 |     17980.4812 |      29.9023 |               29.9023 |    100 |    Y
--------------------------------------------------------------------------------------------------------
Total trials: 100
Total latency (us): 29.9023


It can be seen that the latency is reduced by approximately 17.27%.

1.2 Speed difference in total training:
Training time for num_epochs = 1000:
PyTorch time cost: 1.1044s
Training time cost after TVM operator acceleration: 0.9143s (unroll_max_steps=ffi::Array<Integer>{0, 16, 64, 512, 1024} version)
Training time cost after TVM operator acceleration: 0.7945s (unroll_max_steps=ffi::Array<Integer>{0, 16, 32, 64, 128, 256, 512, 1024} version)

It can be seen that the unroll_max_steps=ffi::Array<Integer>{0, 16, 32, 64, 128, 256, 512, 1024} version is approximately 13% faster than the previous version.

Other parameters in the test are as follows:

target = Target(
    "cuda -arch=sm_70 "
    "-max_num_threads=1024 "
    "-max_threads_per_block=1024 "
    "-thread_warp_size=32 "
    "-max_shared_memory_per_block=98304 "
    "-registers_per_block=65536"
    "+use_tensor_core"
)
tune_tir(
	mod=mod,
	target=target,
	database=database,
	max_trials_global=100,
	num_trials_per_iter=8,
	space="cuda-tensorcore",
	work_dir=work_dir,
)
num_epochs = 1000


2.According to the test results, this modification does not increase the tuning time.
For three operators with max_trials_global=100 for each operator, the total tuning time is:
Version with "unroll_max_steps=ffi::Array<Integer>{0, 16, 64, 512, 1024}": 67 minutes 7 seconds.
Version with "unroll_max_steps=ffi::Array<Integer>{0, 16, 32, 64, 128, 256, 512, 1024}": 63 minutes 43 seconds.
There is no significant time difference.

From the source code analysis:It only increases the mutation selection possibilities of MutateUnrollNode in mutate_unroll.cc (i.e., it affects the diversity of the search space). 
The parameter that truly affects TVM tuning time is max_trials_global in:

tune_tir(
	mod=mod,
	target=target,
	database=database,
	max_trials_global=max_trials_global,
	num_trials_per_iter=num_trials_per_iter,
	space="cuda-tensorcore",
	work_dir=work_dir,
)
It is not the size of the unroll_max_steps array.

3.Currently, I only have a V100 GPU. There are many other CUDA GPUs, and it is difficult for me to obtain all of them for comprehensive comparison and testing due to the high financial cost.
This is also a practical problem that most community contributors face.
However, I have carefully and thoroughly analyzed the MetaSchedule source code.
I have analyzed the complete processing link from ParallelizeVectorizeUnrollNode::Apply to MutateUnrollNode::FindUnrollDecision.
The unroll_max_steps parameter is used in these two key places: the former sets candidate values during initial scheduling, and the latter performs mutation selection during tuning.
This processing mechanism is unified for all CUDA GPUs. The parameter itself only provides the search space and does not contain hardware-specific logic, so it has excellent cross-architecture compatibility.

PR #18930, opened 2026-03-25
[BugFix][Relax] Add structural_equal verification to subroutine cache lookup

3em0:
## Summary

- `SubroutineMixin._get_subroutine()` used `structural_hash` as the sole cache key without `structural_equal` verification. If two different `arg_sinfo` values produced the same 64-bit hash (collision), the cache would return a previously compiled function with mismatched parameter shapes, leading to silently incorrect compiled output.
- Changed the cache to store a list of `(arg_sinfo, result)` pairs per hash bucket and verify with `structural_equal` on lookup, consistent with the pattern in `block_builder.cc`.
- Added a security advisory document and regression test.

## Root Cause

The subroutine cache (`cls._gvar`) was keyed by `(structural_hash(arg_sinfo), is_dataflow)`. A hash match was treated as proof of structural equality, skipping the necessary `structural_equal` check. This is a hash-only lookup anti-pattern — hash determines the bucket, but equality must confirm the match.

For comparison, `block_builder.cc` correctly uses `StructuralHash` + `StructuralEqual` together as the hash and equality functions for `std::unordered_map`.

## Test plan

- [ ] Existing test `test_linear` passes (no regression)
- [ ] New test `test_different_shapes_produce_distinct_subroutines` passes — verifies that the same Module class with different input shapes generates distinct subroutines

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/reference/security_advisory_subroutine_hash_collision.rst
+++ b/docs/reference/security_advisory_subroutine_hash_collision.rst
@@ -1,0 +1,84 @@
+..  Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements.  See the NOTICE file
+    distributed with this work for additional information
+    regarding copyright ownership.  The ASF licenses this file
+    to you under the Apache License, Version 2.0 (the
+    "License"); you may not use this file except in compliance
+    with the License.  You may obtain a copy of the License at
+
+..    http://www.apache.org/licenses/LICENSE-2.0
+
+..  Unless required by applicable law or agreed to in writing,
+    software distributed under the License is distributed on an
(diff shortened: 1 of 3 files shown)


3em0:
CI note: `cpu/pr-head` and `gpu/pr-head` are marked as ERROR, but the test reports show **all tests passing** (4,971/4,971 on CPU, 1,360/1,360 on GPU, zero failures). This appears to be a CI infrastructure issue (Jenkins pipeline error after tests completed), not a code issue.

Could a maintainer please re-trigger the CI checks? Thank you.

Issue #18928, opened 2026-03-25
[Tracking Issue] Add missing vision/image operators


tlopex:
### Background                                                                                                                     
                                               
Since we phased out Relay in early 2025,   several vision/image operators that Relay supported are missing from Relax, blocking deployment of common detection and segmentation models (Faster R-CNN, Mask R-CNN,     Spatial Transformer Networks, etc.).     

Currently Relax only has 3 vision/image ops: `all_class_non_max_suppression`, `resize2d`, `grid_sample`.

### Scope                                                                                                                          
                                                                  
Each operator needs:
1. **TOPI compute** (`python/tvm/topi/`) — if not already present
2. **Relax op registration** — C++ attrs + struct info inference (`src/relax/op/`, `include/tvm/relax/attrs/`) + Python wrapper (`python/tvm/relax/op/`)                                                                                                           
3. **Legalization** (`python/tvm/relax/transform/legalize_ops/`)                                                                   
4. **Frontend integration** — update ONNX/PyTorch/TFLite frontends to emit the new op                                              
5. **Tests** — op-level unit tests + frontend integration tests                     


Steps 1–3 (Relax op) and step 4 (frontend) can be done in separate PRs or combined — contributors can choose based on scope.                                                 
                                                                                                                                     
### Operators                                                                                                                      
                                                                                                                                     
#### Tier 1 — Unblocks mainstream detection/segmentation models 

- [ ] `roi_align` — Core op for Faster R-CNN, Mask R-CNN, Cascade R-CNN. TOPI was removed during te.schedule phase-out, needs full reimplementation.
- [ ] `affine_grid` — Generates sampling grid for Spatial Transformer Networks. TOPI already exists (`topi.image.affine_grid`), only needs Relax op + legalization. Pairs with existing `grid_sample`.                                                             
  
#### Tier 2 — Fills domain gaps and fixes broken paths                                                                             
                                                                  
- [ ] `resize3d` — 3D volume resize for medical imaging (CT/MRI) and video. TOPI already exists (`topi.image.resize3d`), only needs Relax op + legalization. Note: ONNX frontend currently works around this by calling TOPI directly.
[...]

PR #18937, opened 2026-03-26
[Relax][PyTorch] Add 3D interpolate support using resize3d

nirdesh17:
Adds support for torch.nn.functional.interpolate 3D mode in Relax frontend.

- Handles 5D inputs (NCDHW)
- Maps to relax.op.image.resize3d
- Ensures correct layout handling
- Adds tests for scale_factor and size cases

All tests pass locally.

part of #18928


nirdesh17:
@tlopex I’ve addressed all the requested changes:

- Fixed layout handling for 5D (respecting default_image_layout, including NDHWC)
- Used struct_info.ndim instead of relying on shape
- Cleaned up method handling with startswith
- Added tests for align_corners=True and 5D NHWC layout

All tests are passing locally. Please take another look.

PR #18946, opened 2026-03-28
[Frontend][ONNX] Add If operator support to Relax ONNX frontend

### LINKED ISSUE #18945: [Tracking Issue][ONNX] Complete missing and limited operators in ONNX frontend
The Relax ONNX frontend (`python/tvm/relax/frontend/onnx/onnx_frontend.py`) currently supports 157 operators. However, 7 operators are commented out in the converter map, and 11 have known limitations. This issue tracks completing support for these operators.
                                                                                                                                   
### Scope                                                       

Each missing operator needs:
1. **Converter implementation** in `onnx_frontend.py`
2. **Tests** in `tests/python/relax/test_frontend_from_onnx.py`
                                                                                                                                   
Each limited operator needs the specific restriction lifted and corresponding test coverage added.
                                                                                                                                   
#### Tier 1 — High value

- [x] `If` — ONNX control flow (conditional branching). Needed for models with runtime-dependent execution paths.                  
- [x] `MatMulInteger` — INT8 matrix multiplication. Required for quantized model inference (e.g. ONNX QDQ models).
                                                                                                                                   
#### Tier 2 — Lower priority                                                                                                       
 
- [x] `MaxRoiPool` — Max ROI pooling.                                                         
- [x] `MatMulInteger16` — INT16 matrix multiplication. Rarely used.
- [x] `Optional` / `OptionalHasElement` / `OptionalGetElement` — ONNX Optional type handling. Used in some models with optional outputs.                                                                                                                           
 
#### Shape/dynamic restrictions

- [x] `Unsqueeze` — Does not support symbolic axes or dynamic axes. Only constant axes work.
- [x] `Squeeze` — Does not support symbolic axes.
- [x] `Slice` — Only supports constant parameters. Dynamic/symbolic start, end, step not supported.                                
- [x] `Reshape` — Special case for shape `[-1]` not fully handled.
                                                                                                                                   
#### Op-specific restrictions                                                                                                      
 
- [x] `ConvTranspose` — Does not support 3D (ConvTranspose3d, ndim=5).                                                             
- [x] `Resize` — Does not support dynamic ROI.                  
- [x] `ArgMax` / `ArgMin` — Does not support `select_last_index` attribute.
- [x] `SequenceInsert` — Does not support inserting with new axis.                                                                 
- [x] `SplitToSequence` — Only supports `keepdims=1`.
                                                                                                                                   
### Implementation reference                                    

- Converter base class: `OnnxOpConverter` in `onnx_frontend.py`                                                                    
- Converter map: `_get_convert_map()` (line 3937)
- Existing tests: `tests/python/relax/test_frontend_from_onnx.py`                                                                  
                                                                                                                                   
The shape/dynamic restriction fixes (`Unsqueeze`, `Squeeze`, `Slice`) are good first issues — the converter class already exists, just needs to handle additional cases.                                                                                             
                                                                
If you're interested in contributing, please comment below to claim a task before starting work.

cc @KJlaccHoeUM9l

OmarAzizi:
### Summary

This PR implements the ONNX `If` operator in the Relax ONNX frontend. The `If` operator enables conditional branching in ONNX models, where a boolean condition selects between two subgraph branches (`then_branch` and `else_branch`) at runtime. This is required for any model with runtime-dependent execution paths.

Closes #18945 (Tier 1 — `If` operator)

### Implementation Notes

- The main challenge is that `relax.If` cannot be emitted inside a dataflow block, which is how the ONNX frontend normally builds the entire graph. To handle this, when the graph contains an `If` node, the function body is built as a regular binding block instead — matching the approach used by the PyTorch Relax frontend for `torch.cond`.

- Each branch is an ONNX subgraph that can reference values from the outer graph. A new `_convert_subgraph` method handles converting these subgraphs into Relax expressions, making outer-scope values available to the branch while ensuring branch-local bindings don't leak back to the parent graph.

### Why `relax.If` cannot live inside a dataflow block

Dataflow blocks in Relax carry a semantic guarantee: every operation inside them must be pure and side-effect-free with no control flow. This allows the compiler to treat the entire block as a static computational graph for optimizations like operator fusion and constant folding. An `If` node breaks this guarantee by introducing runtime-dependent branching, so Relax's well-formedness checker explicitly forbids it. I discovered this when the checker raised:
```
This IR is not well-formed: If nodes are not allowed to appear in dataflow blocks.
```

The fix — skipping the dataflow block when the graph contains an `If` node — mirrors exactly how the PyTorch Relax frontend handles `torch.cond`.

### Known Limitations

**Dataflow block**: Models whose top-level graph contains an `If` node are built without a dataflow block, which may affect downstream optimisation passes that rely on dataflow block structure.

### Tests

Four new tests covering: scalar and tensor conditions, condition computed from another op, and multiple branch outputs. All verified against onnxruntime via `check_correctness`.

### DIFF
--- a/python/tvm/relax/frontend/onnx/onnx_frontend.py
+++ b/python/tvm/relax/frontend/onnx/onnx_frontend.py
@@ -14,7 +14,7 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-# ruff: noqa: E501, E731, E741, F841, RUF005
+# ruff: noqa: E501, E731, E741, RUF005
 """ONNX: Open Neural Network Exchange importer for Relax.
 
 This module implements the required functionality to read ONNX models
@@ -36,6 +36,7 @@ Not all TVM kernels currently support dynamic shapes, please file an issue on
 github.com/apache/tvm/issues if you hit an error with dynamic kernels.
 """
 
+import contextlib
 import functools
 import math
 import operator
@@ -97,7 +98,7 @@ def get_constant(
         isn't found in params, the input variable is returned unmodified.
     """
     # Params is actually both the graph nodes and param dictionary, unpack them.
-    graph_nodes, params = params[0], params[1]
+    graph_nodes, params = params
     # Convert if possible
     if isinstance(var, relax.Var) and var.name_hint in params:
         # When converting a parameter to a constant, update references to it as well.
@@ -2421,9 +2422,7 @@ class AffineGrid(OnnxOpConverter):
         align_corners = attr.get("align_corners", 0)
 
         if align_corners != 1:
-            raise NotImplementedError(
-                "AffineGrid with align_corners=0 is not yet supported in TVM"
-            )
+            raise NotImplementedError("AffineGrid with align_corners=0 is not yet supported in TVM")
 
         # Extract size values
         if isinstance(size, relax.Constant):
@@ -4030,15 +4029,7 @@ class GridSample(OnnxOpConverter):
             align_corners=align_corners,
         )
 
-class If(OnnxOpConverter):
-    """Operator converter for the ONNX If operator (conditional branching)."""
-    
-    @classmethod
-    def _impl_v1(cls, bb, inputs, attr, params):
-        raise NotImplementedError(
-            "If is handled directly in _construct_nodes, not via the converter map."
-        )
-    
+
 def _get_convert_map():
     return {
         # defs/experimental
@@ -4186,7 +4177,6 @@ def _get_convert_map():
         "OneHot": OneHot,
         "Unique": Unique,
         "NonZero": NonZero,
-        "If": If,
         # "MaxRoiPool": MaxRoiPool,
         "RoiAlign": RoiAlign,
         "NonMaxSuppression": NonMaxSuppression,
@@ -4262,8 +4252,6 @@ class ONNXGraphImporter:
         mod : tvm.IRModule
             The returned relax module
         """
-        import contextlib
-
         has_if = any(node.op_type == "If" for node in graph.node)
 
         with self.bb.function("main"):
@@ -4285,6 +4273,7 @@ class ONNXGraphImporter:
                 else:
                     output_var = self.bb.emit_output(outputs)
 
+            # ExitStack closes here — dataflow block is now closed
             func_attrs = {"num_input": self._num_input}
             input_list = [value for value in self._inputs.values() if isinstance(value, relax.Var)]
             if self._keep_params_in_input and self._params:
@@ -4382,12 +4371,15 @@ class ONNXGraphImporter:
 
     def _check_for_unsupported_ops(self, graph: onnx.onnx_ml_pb2.GraphProto):
         convert_map = _get_convert_map()
+        # Ops handled directly in _construct_nodes rather than via the converter map.
+        directly_handled_ops = {"If"}
         unsupported_ops = set()
         for node in graph.node:
             op_name = node.op_type
             if (
-                op_name not in convert_map and op_name != "Constant"
-                # and op_name not in _identity_list
+                op_name not in convert_map
+                and op_name not in directly_handled_ops
+                and op_name != "Constant"
             ):
                 unsupported_ops.add(op_name)
         if unsupported_ops:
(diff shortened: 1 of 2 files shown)


python/tvm/relax/frontend/onnx/onnx_frontend.py
```diff
@@ -4254,25 +4262,31 @@ def from_onnx(self, graph: onnx.onnx_ml_pb2.ModelProto, opset: int) -> IRModule:
...
         """
+        import contextlib
+
+        has_if = any(node.op_type == "If" for node in graph.node)
+
         with self.bb.function("main"):
-            with self.bb.dataflow() as df:  # pylint: disable=invalid-name, unused-variable
+            with contextlib.ExitStack() as stack:
+                if not has_if:
+                    stack.enter_context(self.bb.dataflow())
```
tlopex, line 4260: This turns off dataflow for the entire model whenever there is an If, which seems a bit too conservative. Maybe we can just break the dataflow block around the If node instead of disabling it globally.

python/tvm/relax/frontend/onnx/onnx_frontend.py
```diff
@@ -4400,6 +4413,22 @@ def _construct_nodes(self, graph: onnx.onnx_ml_pb2.GraphProto):
...
+                then_seq = relax.SeqExpr(blocks=[], body=then_expr)
+                else_seq = relax.SeqExpr(blocks=[], body=else_expr)
+                if_result = self.bb.emit(relax.If(cond, then_seq, else_seq))
+
+                # Handle single and multiple outputs
+                if len(outputs) == 1:
+                    self._nodes[outputs[0]] = if_result
+                else:
+                    for k, i in zip(outputs, range(len(outputs))):
+                        self._nodes[k] = self.bb.emit(relax.TupleGetItem(if_result, i))
```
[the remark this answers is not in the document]
tlopex: This makes sense

python/tvm/relax/frontend/onnx/onnx_frontend.py
```diff
@@ -4532,11 +4562,64 @@ def _convert_operator(
...
             raise NotImplementedError(f"Operator {op_name} not implemented.")
         return sym
+    
+    def _convert_subgraph(self, bb, graph):
+        """
+        Walk an ONNX GraphProto (a branch body) and return a Relax SeqExpr.
+        Outer-scope nodes are visible because we copy self._nodes into the
+        local lookup table before processing.
+        """
+        outer_nodes = dict(self._nodes)
```
tlopex, line 4566: self._nodes is restored only on the normal path here. If subgraph conversion throws, the outer state is left mutated. Seems safer to wrap this in try/finally so self._nodes always gets restored.

python/tvm/relax/frontend/onnx/onnx_frontend.py
```diff
@@ -4532,11 +4562,64 @@ def _convert_operator(
         if op_name in convert_map:
             convert_class = convert_map[op_name]
             op_function = convert_class.get_converter(opset)
-            sym = op_function(self.bb, inputs, attrs, [self._nodes, self._params])
+            sym = op_function(self.bb, inputs, attrs, [self._nodes, self._params, self])
```
tlopex: The third element self is never accessed by any converter. If you need the importer accessible to converters, pass it as an explicit keyword argument or a named container, not hidden in a list.

OmarAzizi:
@tlopex Thanks for the review! Addressed all the smaller feedback points in the follow-up commit. Additionally, I included the handling of nested if as requested, along with a test for it.

> In `_convert_subgraph`, this `except Exception` feels too broad for control flow. It seems safer to check `hasattr(op, "struct_info")` / `op.struct_info is not None`, or catch a more specific exception type.

**Note:** I tried `hasattr(op, "struct_info")` but it doesn't work here. `struct_info` is a property on the class, so `hasattr` always returns `True`, but accessing it still raises when unpopulated:

```
python/tvm/ir/expr.py:62: in struct_info
    return _ffi_api.ExprStructInfo(self)
include/tvm/relax/struct_info.h:398: in tvm::relax::StructInfo tvm::relax::GetStructInfo(const Expr&)
    TVM_FFI_ICHECK(ptr) << "The struct_info is not populated, check if you have normalized the expr";
E   tvm.error.InternalError: Check failed: (ptr) is false: The struct_info is not populated, check if you have normalized the expr
```

The property itself raises before returning `None`, so the `is None` check is never reached. Instead, I now catch `tvm.error.InternalError` specifically, which is exactly what TVM raises in this case.

---

**Still open: Disabling dataflow globally when `If` is present**

I attempted to close the dataflow block locally around the `If` node and reopen it afterward, but hit a fundamental issue: `_convert_subgraph` needs to call `bb.normalize` on branch expressions to populate `struct_info`. When closing the dataflow block before converting the branches, `normalize` fails. When left it open, the branch nodes get emitted into the active dataflow block — producing invalid IR like this:
```python
with R.dataflow():
    if cond:               # If inside dataflow block (invalid)
        with R.dataflow(): # branch nodes incorrectly emitted as dataflow vars
            lv = R.multiply(...)
            R.output()
```

The current approach of skipping the dataflow block for the entire function when `If` is present avoids this and produces correct IR. It matches the same pattern the PyTorch Relax frontend uses for `torch.cond`. Happy to explore alternatives if you have a suggestion.

PR #18951, opened 2026-03-28
[Frontend][ONNX] Add MatMulInteger support to Relax ONNX frontend

### LINKED ISSUE #18945: [Tracking Issue][ONNX] Complete missing and limited operators in ONNX frontend

OmarAzizi:
### Summary

Implements the `MatMulInteger` operator (opset 10) in the Relax ONNX frontend — INT8 matrix multiplication. Required for quantized model inference (e.g. ONNX QDQ models).

Closes #18945 (Tier 1 — MatMulInteger operator)

### Tests

- All 4 `int8`/`uint8` dtype combinations, with and without scalar zero points
- 3-D and 4-D batched matmul

### DIFF
--- a/include/tvm/relax/attrs/nn.h
+++ b/include/tvm/relax/attrs/nn.h
@@ -267,6 +267,58 @@ struct Conv2DTransposeAttrs : public AttrsNodeReflAdapter<Conv2DTransposeAttrs> 
                                     BaseAttrsNode);
 };  // struct Conv2DTransposeAttrs
 
+/*! \brief Attributes used in Conv3dTranspose operator */
+struct Conv3DTransposeAttrs : public AttrsNodeReflAdapter<Conv3DTransposeAttrs> {
+  ffi::Array<int64_t> strides;
+  ffi::Array<int64_t> padding;
+  ffi::Array<int64_t> output_padding;
+  ffi::Array<int64_t> dilation;
+  int groups;
+  ffi::String data_layout;
+  ffi::String kernel_layout;
+  ffi::String out_layout;
+  DataType out_dtype;
+
+  static void RegisterReflection() {
+    namespace refl = tvm::ffi::reflection;
+    refl::ObjectDef<Conv3DTransposeAttrs>()
+        .def_ro("strides", &Conv3DTransposeAttrs::strides,
+                "Specifies the strides of the convolution.")
+        .def_ro("padding", &Conv3DTransposeAttrs::padding,
+                "If padding is non-zero, then the input is implicitly zero-padded"
+                "Padding support both symmetric and asymmetric as"
+                "one int : same padding used on all sides"
+                "three int : back/bottom/right will use same padding as front/top/left"
+                "six int : padding width in the order of (front, top, left, back, bottom, right)")
+        .def_ro("output_padding", &Conv3DTransposeAttrs::output_padding,
+                "Used to disambiguate the output shape.")
+        .def_ro("dilation", &Conv3DTransposeAttrs::dilation,
+                "Specifies the dilation rate to use for dilated convolution.")
+        .def_ro("groups", &Conv3DTransposeAttrs::groups,
+                "Number of groups to split the input into for grouped convolution. The number of "
+                "input and "
+                "output channels should be divisible by the number of groups.")
+        .def_ro("data_layout", &Conv3DTransposeAttrs::data_layout,
+                "Dimension ordering of input data. Can be 'NCDHW', 'NDHWC', etc."
+                "'N', 'C', 'D', 'H', 'W' stands for batch, channel, depth, height, and width"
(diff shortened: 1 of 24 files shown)


tlopex (changes requested):
A couple of follow-ups:

- Please avoid using `dim.value` here. It will break on symbolic shapes. `relax.op.expand_dims(a_zp, axis=-1)` would be simpler and more robust than extracting concrete shape values for the reshape. The same applies to `b_zp` if it has a similar path.

- Please add a test for 1-D zero points as well. Right now all tests use scalar zero points, so the reshape/broadcast path for per-row or per-column zero points is not covered.

OmarAzizi:
Thank you @tlopex for the review, I addressed everything in the follow-up commit.

One thing worth noting. Two ORT limitations surfaced during testing:

**1. Per-row `a_zero_point` (shape `[M]`)**
The ONNX spec allows this, but ORT's kernel unconditionally rejects it. Our converter handles it correctly, so the test verifies TVM against a NumPy reference directly. Also, I added a separate test marked `xfail(strict=True)` to document the ORT gap and ensure that we get alerted if ORT ever fixes it.

**2. Mixed `int8/uint8` dtype — `xfail(strict=False)`**
The CI's ORT version doesn't support this combination, but newer ORT does (shows as `XPASS` locally). `strict=False` lets it pass on both old and new ORT without breaking CI.

**Edit: I mentioned that the ORT limitation is in the CPU kernel in the commit message, but after testing, it fails on the GPU as well. The error message looks something like this:**

```
FAILED test_matmulinteger_per_channel_zp

tests/python/relax/test_frontend_onnx.py:126: in check_correctness
    ort_output = ort_session.run([], inputs)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^

E   onnxruntime.capi.onnxruntime_pybind11_state.RuntimeException:
    [ONNXRuntimeError] : 6 : RUNTIME_EXCEPTION : Non-zero status code returned
    while running MatMulInteger node.

    MatmulInteger: input1 zero point must be a scalar or 1D tensor of size 1
```

OmarAzizi:
> LGTM! Could you resolve the conflicts?

Thank you, Done

PR #18952, opened 2026-03-29
[Relax][ONNX] Add roi_pool op and MaxRoiPool frontend support

LudovicoYIN:
## Summary

Add Relax `roi_pool` support and wire it through the ONNX frontend for `MaxRoiPool`.

## Changes

- add `relax.vision.roi_pool`, including attrs, Python wrapper, struct info inference, and legalization
- add TOPI `roi_pool` compute for NCHW layout
- support ONNX `MaxRoiPool` in the Relax ONNX frontend
- handle empty / out-of-bound pooled bins according to ONNX/reference semantics, returning `0` instead of propagating invalid reductions
- add regression tests for Relax op inference, legalization, and ONNX frontend import
- add out-of-bound ROI coverage to make sure fully invalid pooled bins still match ONNX Runtime

## Validation

- `pytest tests/python/relax/test_op_vision.py -k roi_pool`
- `pytest tests/python/relax/test_frontend_onnx.py -k 'max_roi_pool'`

This PR completes the `MaxRoiPool` portion of the Relax ONNX frontend operator work tracked in #18945.

### DIFF
--- a/python/tvm/relax/frontend/onnx/onnx_frontend.py
+++ b/python/tvm/relax/frontend/onnx/onnx_frontend.py
@@ -4199,7 +4199,6 @@ def _get_convert_map():
         "OneHot": OneHot,
         "Unique": Unique,
         "NonZero": NonZero,
-        "If": If,
         "MaxRoiPool": MaxRoiPool,
         "RoiAlign": RoiAlign,
         "NonMaxSuppression": NonMaxSuppression,


LudovicoYIN:
hi @tlopex. i also made a small compatibility fix in
`python/tvm/runtime/support.py` and
`python/tvm/s_tir/meta_schedule/utils.py`.

Without this change, `import tvm` fails in environment with:

`TypeError: __dict__ slot disallowed: we already got one`

This is not specific to `MaxRoiPool` itself. I added it only to unblock local testing.
If preferred, I can move this part into a separate PR.

python/tvm/topi/vision/roi_pool.py
```diff
@@ -0,0 +1,94 @@
...
+        lambda i, c, ph, pw: te.max(_sample(i, c, ph, pw), axis=[rh, rw]),
+        tag="pool,roi_pool_nchw",
+    )
+
+    return te.compute(
+        (num_roi, channel, pooled_size_h, pooled_size_w),
+        lambda i, c, ph, pw: tvm.tirx.if_then_else(
+            _is_empty(i, ph, pw), zero, pooled[i, c, ph, pw]
+        ),
+    )
```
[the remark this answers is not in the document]
LudovicoYIN: I kept the explicit empty-bin check here because using -inf as the final sentinel could mis-handle valid non-empty bins whose max value is actually -inf. This is a bit more work, but it preserves the intended semantics without assuming anything about the input values.

python/tvm/relax/frontend/onnx/onnx_frontend.py
```diff
@@ -4177,7 +4199,8 @@ def _get_convert_map():
         "OneHot": OneHot,
         "Unique": Unique,
         "NonZero": NonZero,
-        # "MaxRoiPool": MaxRoiPool,
+        "If": If,
```
tlopex: remove this line
The If operator is handled via directly_handled_ops, not through the converter map

LudovicoYIN: Sorry, you’re right. That line was added by mistake during rebase conflict resolution. I’ve removed it in the latest update.

LudovicoYIN:
Thanks your review

PR #18953, opened 2026-03-30
[Docs] Align documentation with tirx/s_tir namespace split

tlopex:
Since we've split the old `tir` namespace into `tirx` (core IR / lowering) and `s_tir` (schedule primitives / auto-tuning), some outdated documentation need to be updated. The global rename still leaves a few concept-level references using "tirx" in prose (for example, "Relax and tirx programs"). Since "tirx" now refers only to one part of the old TensorIR stack, these higher-level references should use "TensorIR" instead, so they correctly cover both `tirx` and `s_tir`.

In this PR, we
- Add tirx / s_tir module descriptions to `docs/deep_dive/tensor_ir/index.rst` and `docs/arch/index.rst` (new `tvm/s_tir` section, updated `tvm/tirx` section).
- Fix concept-level prose in `docs/arch/index.rst` and `docs/arch/pass_infra.rst` to use "TensorIR" instead of "tirx" where referring to the concept rather than the namespace.                                                                                                                   
- Fix `docs/arch/runtimes/vulkan.rst` to use "TIR" instead of "tirx" in debug shader description.                                                                                                              
- Correct `tvm/dlight` → `tvm/s_tir/dlight` section path and "tirx schedules" → "s_tir schedules" in `docs/arch/index.rst`.                                                                   
- Revert unintended label changes in `abstraction.rst` and `tir_creation.py` (labels kept as `_tir-abstraction`, `_tir-creation`).                                                                            
- Revert unintended title change in `tir_transformation.py` (kept as "Transformation").                                                                                                               
- Revert `exclude-members` change in `tirx/tirx.rst` (kept original list).

### DIFF
--- a/docs/arch/index.rst
+++ b/docs/arch/index.rst
@@ -319,8 +319,8 @@ The tirx module includes:
 tvm/s_tir
 ---------
 
-``s_tir`` (Schedulable TIR) contains schedule primitives and auto-tuning
-tools that operate on ``tirx::PrimFunc`` (split from the former ``tir`` module):
+``s_tir`` (Schedulable TIR, split from the former ``tir`` module) contains
+schedule primitives and auto-tuning tools that operate on ``tirx::PrimFunc``:
 
 - Schedule primitives to control code generation (tiling, vectorization, thread
   binding) in ``s_tir/schedule``.


tlopex:
cc @tqchen

PR #18955, opened 2026-03-30
[Relax][ONNX] Fix shape/dynamic restrictions for `Squeeze`/`Unsqueeze` and `Slice`

Aharrypotter:
## Summary

Relates to #18945.

This PR improves ONNX frontend handling for dynamic `Unsqueeze`/`Squeeze`/`Slice`, tightens validation paths, and adds targeted structural/negative regression tests.

- Refactor constant-path `Unsqueeze` lowering to use a single `reshape` based on computed target shape.
- Remove scalar-specific branching and repeated `expand_dims` in the constant path.
- Add/keep structural helper usage in ONNX frontend tests for Relax call-op checks.
- Add regression coverage for scalar-input `Unsqueeze`.

## Changes

- Add dynamic-axes conversion paths for `Unsqueeze` and `Squeeze`:
  - infer output shape via runtime shape-tensor construction
  - lower to `relax.reshape` with validated shape rank/length assumptions
- Improve `Slice` conversion robustness:
  - support dynamic parameter forms with stricter rank/length validation
  - reject invalid zero-step inputs when statically known
  - fix docstring wording (`Splice` -> `Slice`)
- Strengthen ONNX frontend tests:
  - negative test for duplicate `Unsqueeze` axes
  - structural IR check for dynamic `Slice` (`relax.dynamic_strided_slice` present, `relax.strided_slice` absent)
  - negative test for zero-step `Slice`
- Refactor constant-path `Unsqueeze` scalar handling:
  - replace scalar special-casing + repeated `expand_dims` with one target-shape `reshape`
  - add scalar-input regression test
- Restore shared test helper used by structural Relax call-op checks.

## validation

- `ruff check`: passed
- `pre-commit --files`: passed
- `pytest`: 8 passed

### DIFF
--- a/tests/python/relax/test_frontend_onnx.py
+++ b/tests/python/relax/test_frontend_onnx.py
@@ -232,15 +232,15 @@ def run_in_tvm(
     return vm.get_outputs("main")
 
 
-def collect_relax_call_ops(relax_func: relax.Function) -> set[str]:
-    call_ops = set()
+def collect_relax_call_ops(func: relax.Function) -> list[str]:
+    op_names: list[str] = []
 
-    def _visit(expr):
+    def fvisit(expr: relax.Expr) -> None:
         if isinstance(expr, relax.Call) and isinstance(expr.op, tvm.ir.Op):
-            call_ops.add(expr.op.name)
+            op_names.append(expr.op.name)
 
-    relax.analysis.post_order_visit(relax_func.body, _visit)
-    return call_ops
+    relax.analysis.post_order_visit(func.body, fvisit)
+    return list(op_names)
 
 
 def collect_scalar_constants(func: relax.Function) -> list[bool | int | float]:


python/tvm/relax/frontend/onnx/onnx_frontend.py
```diff
@@ -1847,68 +1891,308 @@ def get_prim_value_list(values):
...
+
+def _normalize_constant_axes(axes: list[int], rank: int, op_name: str) -> list[int]:
+    """Normalize a list of constant axes and validate their uniqueness."""
+
+    normalized_axes = []
+    for axis in axes:
+        if axis < 0:
+            axis += rank
+        if axis < 0 or axis >= rank:
+            raise ValueError(f"{op_name} axis {axis} is out of range for rank {rank}.")
```
tlopex: Please preserve the original axis value in `_normalize_constant_axes` for the error message. Right now it reports the normalized value instead of the user input, which can be confusing.

tlopex (changes requested):
- Please fix the `Squeeze` behavior for `ShapeExpr` input when `axis=None`. Per ONNX spec, it should remove all size-1 dimensions, but the current path returns the input unchanged when `len(data) > 1`.

- Please take a look at the `Slice` dynamic path. It currently converts `ShapeExpr` to a tensor and returns a tensor from `dynamic_strided_slice`, while the constant path returns `ShapeExpr` or `relax.const`. That type mismatch may cause downstream issues.

- `_get_known_tensor_length` could also be simplified a bit — the second `if struct_info.ndim != 1` is effectively just the `ndim == -1` case, so using `if struct_info.ndim == -1: return None` would be clearer.

python/tvm/relax/frontend/onnx/onnx_frontend.py
Aharrypotter: Makes sense, I'll fix this.

python/tvm/relax/frontend/onnx/onnx_frontend.py
```diff
@@ -1847,68 +1891,308 @@ def get_prim_value_list(values):
...
+        return len(expr.values)
+    if isinstance(expr, relax.PrimValue):
+        return 1
+    struct_info = expr.struct_info
+    if not isinstance(struct_info, relax.TensorStructInfo):
+        return None
+    if struct_info.ndim != -1 and struct_info.ndim != 1:
+        raise ValueError(f"Expected a 1-D tensor, but got ndim={struct_info.ndim}.")
+    if struct_info.ndim != 1:
+        return None
```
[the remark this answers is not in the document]
Aharrypotter: fixed

python/tvm/relax/frontend/onnx/onnx_frontend.py
Aharrypotter: Hi @tlopex , thank you for the detailed review! I've updated the PR to address your and gemini-code-assist's feedback

Especially, regarding the Slice dynamic path: I've added an explicit check to reject ShapeExpr inputs. This prevents potential type mismatches between constant and dynamic paths while keeping the implementation clean.

Aharrypotter:
### Summary of the fix for `collect_relax_call_ops` merge conflict.

During the recent conflict merge, the helper function `collect_relax_call_ops` in `test_frontend_onnx.py` was inadvertently changed to return a `set` instead of a `list`. This broke several structural IR tests that require:

1.  **Count Verification**: Some tests (like `test_matmulinteger16_ir`) need to verify that a specific operator appears exactly $N$ times (e.g., `call_ops.count("relax.astype") == 2`). A `set` deduplicates these entries, losing this critical information.
2.  **Type Consistency**: Tests that assert `collect_relax_call_ops(...) == []` fail when the function returns an empty `set()`, as the types do not match.
3.  **Order Preservation**: Maintaining the post-order traversal sequence is better for future-proofing graph topology assertions.

This commit restores the `list[str]` return type and updates the implementation in `test_frontend_onnx.py` to ensure all tests pass correctly.

Aharrypotter:
cc @tlopex

PR #18905, opened 2026-03-15
[Fix] Avoid generating dangling free var ceil_log2 for gpu_generic backends

akaashrp:
The unused variable ceil_log2 was causing downstream errors in MLC-LLM during compilation for metal and webgpu. This PR removes this intermediate variable and moves it to the total_rounds calculation.

PR #18957, opened 2026-03-31
[FIX] Inline ceil_log2 in gpu_2d_continuous_cumsum to fix MakePackedAPI error

gnguralnick:
## Summary
- The intermediate variable `ceil_log2` in `gpu_2d_continuous_cumsum` created a `LetStmt`-bound `Var` in the TIR function
- When `MakePackedAPI` processed the function, it reported `ceil_log2` as an undefined variable not passed as an API argument
- Inline the expression directly into `total_rounds` to avoid the intermediate `Var` — the computation is identical

## Test plan
- Compile a model that uses GPU sampling (e.g. any LLM with top-p sampling on Metal) and verify compilation succeeds
- The error this fixes: `Check failed: undefined.size() == 0: In PrimFunc gpu_2d_continuous_cumsum variables [ceil_log2] are used, but are not passed in as API arguments`

### DIFF
--- a/python/tvm/relax/backend/gpu_generic/cumsum.py
+++ b/python/tvm/relax/backend/gpu_generic/cumsum.py
@@ -159,8 +159,7 @@ def gpu_2d_continuous_cumsum(
         A = T.match_buffer(var_a, [m, n], dtype=in_dtype)
         Out = T.match_buffer(var_out, [m, n], dtype=out_dtype)
         Tmp = T.alloc_buffer([m, n], dtype=out_dtype)
-        ceil_log2 = T.Cast("int64", T.ceil(T.log2(T.Cast("float32", n))))
-        total_rounds = ceil_log2 // LOG_BLOCK_N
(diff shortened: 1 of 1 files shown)


babusid:
Looks like a duplicate of #18905.

gnguralnick:
You're right. Sorry, didn't notice that one. Though I do also see the CI weirdness; let me know if I should close this or not.

MasterJH5574:
@akaashrp @gnguralnick you guys can decide which PR to merge.  I guess #18905 may need a rebase before it can pass CI.  But given this PR already passes CI, an option is that we can add @akaashrp as a co-author when merging this PR.

akaashrp:
@MasterJH5574 let's merge this PR (I've also closed mine). If it's not too involved to add me as a co-author, we could do that. Otherwise, I'm fine just merging this directly.

MasterJH5574 (approved):
LGTM. Thanks for the fix.

MasterJH5574:
@akaashrp No problem. Merged with you as a co-author. Again thanks for the fix. Sorry that it has been stuck for two weeks.

PR #18969, opened 2026-04-02
[Frontend][ONNX] Support select_last_index for ArgMax and ArgMin

OmarAzizi:
### Summary

This PR implements the `select_last_index` attribute (introduced in opset 12) for the `ArgMax` and `ArgMin` ONNX operators.

Previously, setting `select_last_index=1` raised `OpAttributeUnImplemented`.  This closes the limitation tracked in the ONNX frontend issue.

### Implementation

When `select_last_index=1`, the input tensor is reversed along the reduction axis using `relax.op.flip`, argmax/argmin is computed on the flipped copy, and the result is remapped back to the original index space via `last_idx = (axis_size - 1) - flipped_idx`

Closes part of #18945


tlopex (approved):
LGTM Thanks!

PR #19363, opened 2026-04-06
[TIRX] Bind parallel loops to GPU threads before VerifyMemory

zhils:
`VerifyMemory` on GPU targets treats direct accesses outside thread environments as illegal. In the ScatterValue CUDA lowering path, `topi.scatter_elements` emits `ForKind::kParallel` loops without explicit thread bindings, which triggers false host-memory access failures (e.g. "Did you forget to bind?") during TIR verification.

This change adds a new `tirx` pass (`BindParallelLoopsToThreads`) and inserts it before `VerifyMemory` in the `s_tir` pipelines (including adreno). The pass rewrites parallel loops into `blockIdx.x/threadIdx.x` thread-extent regions, substitutes loop vars with global thread indices, and adds bounds checks for non-divisible extents. This preserves correctness while ensuring GPU kernels pass memory verification for this path.


tlopex (changes requested):
A few things to address here:

1. Please remove `.tmp_scatter_cuda_check.py`. This looks like a local debugging script and should not be committed. It should be replaced with a proper pytest test instead.

2. This PR currently mixes several unrelated changes: the NVCC `-allow-unsupported-compiler` workaround, the `loop_partition.cc` `Var{}` fix, and the `kDLWebGPU` change in `verify_memory.cc` are all independent and should be split into separate PRs.

3. Please clarify the semantics for nested parallel loops. When `in_parallel_loop_` is already true, inner parallel loops are left as `kParallel`, but once they are inside a GPU kernel they effectively become serial. If that is intentional, it would be good to document it explicitly; otherwise it may be better to reject this case.

4. The `else` branch in `IfThenElse` looks unnecessary. `Evaluate(IntImm(DataType::Int(32), 0))` is a no-op, and TIR already supports `IfThenElse` without an `else`, so this can just be `IfThenElse(global_idx < extent, mapped_body)`.

5. For target-less `PrimFunc`s, please avoid falling back to `Target::Current()`. If the function has no target attribute, it is better to leave it unchanged rather than guessing from ambient context. That would also be consistent with how other recent passes handle target-less functions.

6. This PR also needs proper pytest coverage for the new pass.

zhils:
> A few things to address here:
> 
> 1. Please remove `.tmp_scatter_cuda_check.py`. This looks like a local debugging script and should not be committed. It should be replaced with a proper pytest test instead.
> 2. This PR currently mixes several unrelated changes: the NVCC `-allow-unsupported-compiler` workaround, the `loop_partition.cc` `Var{}` fix, and the `kDLWebGPU` change in `verify_memory.cc` are all independent and should be split into separate PRs.
> 3. Please clarify the semantics for nested parallel loops. When `in_parallel_loop_` is already true, inner parallel loops are left as `kParallel`, but once they are inside a GPU kernel they effectively become serial. If that is intentional, it would be good to document it explicitly; otherwise it may be better to reject this case.
> 4. The `else` branch in `IfThenElse` looks unnecessary. `Evaluate(IntImm(DataType::Int(32), 0))` is a no-op, and TIR already supports `IfThenElse` without an `else`, so this can just be `IfThenElse(global_idx < extent, mapped_body)`.
> 5. For target-less `PrimFunc`s, please avoid falling back to `Target::Current()`. If the function has no target attribute, it is better to leave it unchanged rather than guessing from ambient context. That would also be consistent with how other recent passes handle target-less functions.
> 6. This PR also needs proper pytest coverage for the new pass.

Thanks for the review. I removed .tmp_scatter_cuda_check.py and added tests/python/tirx-transform/test_tir_transform_bind_parallel_loops_to_threads.py for the pass. The NVCC workaround, loop_partition Var{} fix, and verify_memory WebGPU change are reverted here and will be submitted as separate PRs. Nested T.parallel is now rejected with a clear error; the pass doc comment explains semantics (including leaving kParallel unchanged under an existing thread env). IfThenElse no longer has a no-op else. PrimFuncs without kTarget are left unchanged (no Target::Current fallback).

tlopex:
Hi @zhils, thanks a lot for turning around the previous fixes so quickly, and sorry for the slow reply. I know you’ve already addressed quite a few comments, but after spending more time on this PR, I think my main concern is actually more fundamental than the first round.

I’m not fully convinced this is fixing the bug at the right layer.

The root cause seems to be that `topi.scatter_elements` emits bare `T.parallel` on CUDA, which then trips `VerifyMemory`. I understand that `scatter_elements` uses `te.extern` with a hand-written IRBuilder body, so it can’t simply get a normal GPU schedule. But to me, that still points to fixing the lowering in the op itself, for example by making the IRBuilder emit `T.thread_binding` on GPU targets, instead of adding a generic cleanup pass later in the pipeline.

A CUDA PrimFunc probably shouldn’t depend on a later pass to turn leftover `T.parallel` into `blockIdx.x/threadIdx.x`.

My worry is that a general pass like this is a bit too broad. It fixes the current `scatter_elements` issue, but it would also silently hide the same kind of bug in future operators. That makes `VerifyMemory` less useful as a safety check, instead of fixing the lowering that produced invalid GPU TIR in the first place.

So I don’t think this should land in its current form. I’d prefer to fix `scatter_elements` directly so its IRBuilder path emits properly bound CUDA TIR.

Happy to discuss if you see this differently, or if there’s some context around the `te.extern` path that I’m missing.

tlopex:
Hi @zhils, Thank you for your fix! I think the bug is fixed by #19497, so I will close this pr temporarily. You can feel free to reopen it and have any further discussion with me here

Issue #18971, opened 2026-04-03
[Tracking Issue][TFLite] Expand unit test coverage for supported non-quantized operators


tlopex:
We just rebuilt the TFLite frontend. 

Today, `tests/python/relax/test_frontend_tflite.py` already covers a subset of operators, including some via parametrized tests (for example `RELU`, `RELU6`, `TANH`, `SOFTMAX`, `LOGISTIC`, `ARG_MIN`, and `ARG_MAX`). However, its test coverage is still not enough.

This issue tracks adding **operator-focused unit tests for supported non-quantized operators that are still under-tested or currently have no explicit coverage** in the Relax TFLite frontend. 

## Candidate operators to cover

The list below is based on the current Relax frontend implementation and the current contents of `tests/python/relax/test_frontend_tflite.py`.

### Shape / indexing / layout
- [ ] `SQUEEZE`
- [ ] `STRIDED_SLICE`
- [x] `SLICE`
- [x] `SHAPE`
- [x] `SELECT` / `WHERE`
- [ ] `GATHER`
- [ ] `GATHER_ND`
- [x] `REVERSE_V2`
- [ ] `REVERSE_SEQUENCE`
- [x] `TILE`
- [ ] `UNPACK`
- [ ] `ZEROS_LIKE`
- [x] `RANGE`
- [ ] `SPLIT_V`
- [ ] `SPACE_TO_BATCH_ND`
- [ ] `BATCH_TO_SPACE_ND`
- [x] `DEPTH_TO_SPACE`
- [x] `SPACE_TO_DEPTH`

### Elementwise / activation / comparison / normalization
- [x] `HARD_SWISH`
- [x] `LEAKY_RELU`
- [x] `RELU_N1_TO_1`
- [x] `PRELU`
- [x] `LOG`
- [x] `LOG_SOFTMAX`
- [x] `SQUARED_DIFFERENCE`
- [x] `GREATER`
- [x] `GREATER_EQUAL`
- [x] `L2_NORMALIZATION`
- [x] `LOCAL_RESPONSE_NORMALIZATION`

### Reductions
- [x] `MEAN`
- [x] `REDUCE_MAX`
- [x] `REDUCE_MIN`
- [x] `REDUCE_PROD`
- [x] `SUM`

### Resize
- [x] `RESIZE_BILINEAR`
- [x] `RESIZE_NEAREST_NEIGHBOR`

### Convolution / linear / pooling
- [x] `FULLY_CONNECTED`
[...]

PR #19401, opened 2026-04-13
[Frontend][TFLite] Add test coverage for SHAPE and RANGE operators

rknastenka:
Initial goal was to add SHAPE and RANGE tests, solving part of #18971

This PR achieves that and includes the minimum necessary frontend fixes discovered during implementation so those tests reflect real supported behavior instead of xfail/workarounds.

so this PR includes both:
**1. New SHAPE/RANGE tests
2. Targeted frontend fixes required to make those tests pass correctly**

## Why These Changes Were Needed
- SHAPE conversion previously produced symbolic shape info instead of a tensor output aligned with TFLite SHAPE semantics.
- RANGE conversion passed tensor expressions into arange instead of scalar values for constant scalar bounds.
- Zero-input TFLite subgraphs (valid for constant-only models such as RANGE without inputs) were blocked by a strict assertion.
- Model output collection was brittle for constant/prefetched outputs and could fail when output expressions were not already in the expr table.
- As a result, i could not add meaningful SHAPE/RANGE coverage without fixing frontend behavior.

### **Frontend Changes** (In tflite_frontend.py):
- Updated convert_shape: SHAPE now materializes shape output as a tensor using shape_to_tensor(shape_of(...))
- Applies output dtype casting based on ShapeOptions OutType (int32/int64)
- Updated convert_range: Extracts scalar values for start/limit/delta from scalar constants
- Calls arange with scalar-like values
- Keeps dynamic scalar RANGE explicit as unsupported (raises OpNotImplemented with clear message)
- Updated _input_type: Removed assumption that every subgraph must have at least one input
- Supports valid zero-input subgraphs
- Updated from_tflite output assembly: Resolves outputs via tensor wrappers and get_tensor_expr instead of direct expr-table lookup by name
---

**Main functional changes are localized to SHAPE/RANGE conversion and model output/input handling.**


tlopex:
Hi @rknastenka ,could you resolve the conflict so that we can merge it:)

tlopex (approved):

Issue #18424, opened 2025-11-07
[Bug] Segfault when applying Parallel during TIR schedule rewriting


Cookiee235:
When running meta_schedule.tune_tir on a valid TIR module involving multi-dimensional access patterns, TVM crashes during the schedule rewriting phase (RewriteParallelVectorizeUnroll). 

### Actual behavior
```
Traceback (most recent call last):
  File "/share_container/LLMFuzz/TirFuzz/bugs/10-24_20-21/topi.gather_0_M1.py", line 32, in <module>
    database = ms.tir_integration.tune_tir(mod=tir_mod, target='llvm --num-cores=32', work_dir='./tune_tmp', max_trials_global=1, num_trials_per_iter=1)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "test", line 146, in tune_tir
    return tune_tasks(
           ^^^^^^^^^^^
  File "/software/tvm-latest/python/tvm/meta_schedule/tune.py", line 122, in tune_tasks
    task_scheduler.tune(
  File "/software/tvm-latest/python/tvm/meta_schedule/task_scheduler/task_scheduler.py", line 132, in tune
    _ffi_api.TaskSchedulerTune(  # type: ignore # pylint: disable=no-member
  File "python/tvm_ffi/cython/function.pxi", line 758, in core.Function.__call__
  File "<unknown>", line 0, in tvm::meta_schedule::GradientBasedNode::Tune(tvm::ffi::Array<tvm::meta_schedule::TuneContext, void>, tvm::ffi::Array<tvm::FloatImm, void>, int, int, int, tvm::meta_schedule::Builder, tvm::meta_schedule::Runner, tvm::ffi::Array<tvm::meta_schedule::MeasureCallback, void>, tvm::ffi::Optional<tvm::meta_schedule::Database, void>, tvm::ffi::Optional<tvm::meta_schedule::CostModel, void>)
[...]

PR #19403, opened 2026-04-14
[S-TIR] Fix Segfault when applying Parallel during TIR schedule rewriting

cchung100m:
Hi Commiters,

This PR is trying to fix issues https://github.com/apache/tvm/issues/18424. Any suggestions would be appreciated if you are available.

### Root Cause
Unsafe dynamic-shape dereferences in `AdjustParallelVectorize` The code assumed IntImm for buffer shape / loop extent and dereferenced directly. With dynamic shapes, as<IntImmNode>() can be null, which can segfault before any try/catch handles it.

### Solution
Replaced unsafe `IntImm` assumptions with null checks and GetLoopIntExtent(...); if contiguous analysis is not possible, conservatively disables that path instead of dereferencing null.


cchung100m:
Hi @tlopex @mshr-h ,

This PR is trying to fix issues https://github.com/apache/tvm/issues/18424. Any suggestions would be appreciated if you are available. 😄

tlopex (approved):
LGTM

cchung100m:
Thanks to @tlopex 😄

Issue #19412, opened 2026-04-16
[Tracking Issue][TFlite] Expand Missing / Partial Operator Coverage


tlopex:
## Background                                                                           
 
Excited to see the huge progress the community has made on the Relax TFLite frontend recently — many operators and tests have landed in the last few weeks. To keep pushing this forward, this issue tracks the remaining operator gap so contributors can pick up well-scoped items in parallel.      
                                                                                                            
## A. Good-first-issues — drop-in mappings                                              
                                                                
Workflow: add a handler, wire into `convert_map` (alphabetical within its region), add one `verify(TestClass, Expected)` test in `tests/python/relax/test_frontend_tflite.py` (pattern at L55). **All target Relax ops already exist** — no new op work needed.      
                  
  - [ ] `CUMSUM` → `relax.op.cumsum`                                                      
  - [ ] `UNIQUE` → `relax.op.unique`
  - [ ] `SCATTER_ND` → `relax.op.scatter_nd`                                              
  - [ ] `BROADCAST_TO` → `relax.op.broadcast_to`                                          
  - [ ] `SIGN` → `relax.op.sign`
  - [ ] `BITWISE_XOR` → `relax.op.bitwise_xor`                                            
  - [ ] `RIGHT_SHIFT` → `relax.op.right_shift`                                            
  - [ ] `BUCKETIZE` → `relax.op.bucketize`
  - [ ] `EMBEDDING_LOOKUP` → `relax.op.take`                                              
  - [ ] `SELECT_V2` → `relax.op.where`                            
  - [ ] `RELU_0_TO_1` → `relax.op.clip`                                                   
  - [ ] `NON_MAX_SUPPRESSION_V4` → port from existing `convert_nms_v5`                                                                                

## B. Medium — may need new Relax op or non-trivial decomposition                       
                                                                
**Prefer composing with existing Relax ops over adding a new op.** Only register a new `relax.op.*` when the semantics genuinely cannot be expressed by existing primitives (e.g. random generators, bit-level reinterpret). If a decomposition using existing ops is feasible, go with that — it's a smaller PR and avoids bloating the op set.
- [x] `REDUCE_ANY` — commented at `tflite_frontend.py:198`; can be lowered to `relax.op.max` on a bool tensor, no new op needed                                       
- [x] `REDUCE_ALL` — same, lower to `relax.op.min` on bool
- [ ] `RANK` — emit `relax.const(ndim)` for static shapes; no op call needed            
[...]

PR #19413, opened 2026-04-17
[Relax][Frontend][TFLite] Add `REDUCE_ANY` and `REDUCE_ALL`

swjng:
## Summary

Adds TFLite frontend support for `REDUCE_ANY` and `REDUCE_ALL` (B-group item in #19412).

## Lowering

`REDUCE_ANY` / `REDUCE_ALL` take bool tensors and compute logical OR / AND along the given axes. Max / min on a bool tensor produces the same result, so no new Relax op is required:

- `REDUCE_ANY` → `relax.op.max`
- `REDUCE_ALL` → `relax.op.min`

Both reuse the existing `_convert_reduce` handler (shared with `REDUCE_MAX`, `REDUCE_MIN`, `REDUCE_PROD`, `MEAN`, `SUM`), with entries added alphabetically to `convert_map`.

## Testing

Added `test_reduction_bool_ops` in `tests/python/relax/test_frontend_tflite.py`, parametrized over the same shape / axes / keepdims matrix as the existing `test_reduction_ops` (24 combinations total). Verified locally by running the test function directly across all parametrizations.

Refs #19412.


tlopex (approved):
LGTM Thanks!

PR #19415, opened 2026-04-17
[Relax][Frontend][TFLite] Fix bool `REDUCE_ANY`/`REDUCE_ALL` compile failure

swjng:
## Problem

#19413 registered `REDUCE_ANY` / `REDUCE_ALL` as `_convert_reduce` with `relax.op.max` / `relax.op.min`. These TFLite ops are bool-only (per TFL op schema: `TFL_ReduceAnyOp` / `TFL_ReduceAllOp` take and return `TFL_BoolTensor`), and `relax.op.max` / `relax.op.min` are not defined on bool, so any real model using these ops fails at compile time with:

```
Cannot decide min_value for type bool
Cannot decide max_value for type bool
```

The existing structural-equality test passed because it never attempted to compile the converted module (E2E is gated on `CI_ENV_NIGHTLY`).

## Fix

Introduce a dedicated `_convert_reduce_bool` handler that casts the input to int8, reduces with max/min, and casts back to bool. Update the test to compile the expected module so this lowering is exercised without `CI_ENV_NIGHTLY`.

## Testing

Verified compile + VM-run (TF converter → Relax → LLVM) across the full shape / axes / keepdims matrix from `test_reduction_bool_ops`: 12 cases, all PASS.

Follow-up to #19413.

### DIFF
--- a/python/tvm/relax/frontend/tflite/tflite_frontend.py
+++ b/python/tvm/relax/frontend/tflite/tflite_frontend.py
@@ -195,8 +195,8 @@ class OperatorConverter:
             "PRELU": self.convert_prelu,
             "RANGE": self.convert_range,
             "QUANTIZE": self.convert_quantize,
-            "REDUCE_ALL": functools.partial(self._convert_reduce, relax_op=_op.min),
-            "REDUCE_ANY": functools.partial(self._convert_reduce, relax_op=_op.max),
+            "REDUCE_ALL": functools.partial(self._convert_reduce_bool, relax_op=_op.min),
+            "REDUCE_ANY": functools.partial(self._convert_reduce_bool, relax_op=_op.max),
             "REDUCE_MAX": functools.partial(self._convert_reduce, relax_op=_op.max),
             "REDUCE_MIN": functools.partial(self._convert_reduce, relax_op=_op.min),
             "REDUCE_PROD": functools.partial(self._convert_reduce, relax_op=_op.prod),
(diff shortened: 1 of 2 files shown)


python/tvm/relax/frontend/tflite/tflite_frontend.py
```diff
@@ -1787,6 +1787,36 @@ def _convert_reduce(self, relax_op, op):
...
+        from tflite.ReducerOptions import ReducerOptions
+
+        input_tensors = self.get_input_tensors(op)
+        assert len(input_tensors) == 2, "input tensors length should be 2"
+
+        input_tensor = input_tensors[0]
+        in_expr = self.get_expr(input_tensor.tensor_idx)
+
+        axis_value = self.get_tensor_value(input_tensors[1])
+        axis = tuple(axis_value) if len(axis_value.shape) > 0 else tuple((axis_value.item(),))
```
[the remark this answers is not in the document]
swjng: empty axes = identity per TFLite semantics, verified via TF interpreter; current code matches spec

tlopex (approved):

tests/python/relax/test_frontend_tflite.py
```diff
@@ -2002,9 +2018,12 @@ class ReduceBoolModule(tf.Module):
         def func(self, x):
             return tf_op(x, axis=axes, keepdims=keepdims)
 
-    expected = _make_reduce_expected(relax_op, input_shape, axes, keepdims, "bool")
+    expected = _make_reduce_bool_expected(relax_op, input_shape, axes, keepdims)
     verify(ReduceBoolModule, expected)
 
+    # Regression guard: compile to catch a bool max/min lowering path.
+    tvm.compile(expected, tvm.target.Target("llvm"))
```
tlopex, line 2025: tvm.compile(expected, tvm.target.Target("llvm")) I don't think we need it here?

tlopex (changes requested):

tests/python/relax/test_frontend_tflite.py
swjng: I'd like to keep this if possible. The guard caught exactly the regression that #19413 shipped that structural equality matched because expected and converted IR both used `relax.op.max` on bool, and the failure surfaces only at compile time. If there's a different path in CI that actually exercises compile for the TFLite frontend tests, then it can be removed. Do you have any idea of that?

tlopex (approved):

tlopex:
Merged though, we may find some ways to completely fix it, not only at the frontend

PR #19416, opened 2026-04-17
[Relax][TVMScript] Print ExternFunc struct_info when non-default

cchung100m:
### Summary

1. Add HasDefaultExternFuncStructInfo helper to detect default FuncStructInfo for extern functions.

2. Update relax::ExternFunc printer to:
 - emit global_symbol using the correct AccessPath attribute key,
 - conditionally include struct_info only when it differs from the default inferred-by-sinfo-args derive function,
 - use a variadic args array instead of a single positional literal to prepare the ExternFunc call.

3. This reduces noisy/redundant output when printing ExternFunc nodes while preserving explicit struct_info when it conveys meaningful information.

### DIFF
--- a/src/script/printer/relax/function.cc
+++ b/src/script/printer/relax/function.cc
@@ -22,6 +22,19 @@ namespace tvm {
 namespace script {
 namespace printer {
 
+static bool HasDefaultExternFuncStructInfo(const relax::ExternFunc& n) {
+  const auto* sinfo = n->struct_info_.as<relax::FuncStructInfoNode>();
+  if (sinfo == nullptr || sinfo->params.defined() || sinfo->purity ||
+      !sinfo->ret->IsInstance<relax::ObjectStructInfoNode>() || !sinfo->derive_func.defined()) {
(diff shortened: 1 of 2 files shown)


cchung100m:
Hi @tlopex @mshr-h ,

This PR is trying to finish TODO task. Any suggestions would be appreciated if you are available. 😄

tlopex (changes requested):
Thanks! @cchung100m  
1. Should add test coverage in `tests/python/relax/test_tvmscript_printer_relax.py`. At minimum, this should include:
- one case with non-default `struct_info` (for example, with `purity=True` or explicit `params` / `ret`)
- one round-trip test

2.  Maybe update the PR title from Unity to Relax, which I think is better

cchung100m:
Hi @tlopex 

Thanks for the prompt reply and I have updated the part you mentioned.

tlopex (approved):

cchung100m:
Thanks to @tlopex 😄

PR #19417, opened 2026-04-17
[ARITH] Expose allow_override parameter in Python Analyzer.bind()

fpedd:
The C++ Analyzer::Bind() already supports allow_override, but the FFI bridge always used the default (false). This change threads the optional argument through the FFI layer and the Python wrapper so callers can rebind variables without triggering an error.


tlopex (approved):
LGTM! Thanks!

PR #19418, opened 2026-04-18
[WebGPU] Fix UI jank caused by compositor starvation in batched dispatch

ssam18:
Batching all dispatches into a single command buffer (introduced in #18871) blocks the browser compositor from interleaving its rendering work, causing visible frame drops and laggy scrolling during LLM inference. This adds a `maxDispatchesPerFlush` field (default 32) to `WebGPUContext` that flushes the encoder periodically, giving the compositor regular breathing room while preserving most of the batching throughput gains. Setting it to 0 restores the old per-dispatch behavior. Fixes #19342

### DIFF
--- a/web/src/webgpu.ts
+++ b/web/src/webgpu.ts
@@ -423,6 +423,7 @@ export class WebGPUContext {
   private uniformBufferPool: Array<GPUBuffer> = [];
   private uniformBufferPoolSizes: Array<number> = [];
   private pendingDispatchCount = 0;
+  maxDispatchesPerFlush = 32;
   // flags for debugging
   // stats of the runtime.
   // peak allocation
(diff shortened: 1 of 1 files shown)


tlopex:
cc @akaashrp

akaashrp:
Thanks, changes look fine but will need to do some testing

akaashrp:
I'm unable to reproduce the issue described in https://github.com/apache/tvm/issues/19342. I also tested separately in WebLLM, and the streaming output generation visually looked similar with and without this change. Are there specific scenarios where the described "ui jank" issue reliably surfaces?

JCNTH:
Related: #20430 adds the same `maxDispatchesPerFlush` knob (same name and default of 32), measured for throughput rather than UI responsiveness. On Llama-3.2-1B decode in WebLLM the periodic submission alone is 1.18× (paired, 12 rounds), because the GPU starts on the first dispatches while the host is still encoding the rest.

Two things from that work that may be useful here:

1. The periodic submission should not go through `flushCommands()`, because that restarts the uniform buffer pool. Every launch between two flush points then no longer has a stable pool position, which is what lets a bind-group cache (#20428) and an unchanged-upload check (#20429) hit; with `flushCommands()` uploads went from 13.9k to 50.7k per 256-token generation. #20430 splits out a `submitPendingCommands()` that finishes and submits without restarting the pool, and has `flushCommands()` restart the pool even when nothing is pending.
2. `maxDispatchesPerFlush <= 0` here means "submit after every dispatch"; #20430 spells that as `1` and uses `0` for "off". Happy to align with whichever this PR settles on.

If this one lands first I will rebase #20430 onto it and reduce it to the pool-position fix.

PR #19430, opened 2026-04-23
[Relax][NN] Use int64 for RoPE apply flag

xthomaswang:
This patch aligns the dtype of the `apply_rope` flag used by
`llama_rope_with_position_map` with the host-side value passed through
Relax call_tir.

Previously the PrimFunc declared `apply_rope` as `T.int32`, while the
caller-side scalar value is represented as an int64 Relax PrimValue /
ShapeExpr value. This caused Relax well-formed analysis to reject the IR
with:

    Argument N type mismatch: expected R.Prim("int32"), given R.Prim(value=1)

The mismatch can be reproduced through downstream `nn.Module.export_tvm`
paths such as MLC-LLM `convert_weight` / `compile`.

This change updates:
- `llama_rope_with_position_map`: `apply_rope: T.int32` -> `T.int64`
- `PagedKVCache`: pass the split-rotary flag as `int64_t`

### DIFF
--- a/python/tvm/relax/frontend/nn/llm/position_embedding.py
+++ b/python/tvm/relax/frontend/nn/llm/position_embedding.py
@@ -529,7 +529,7 @@ def llama_rope_with_position_map(  # pylint: disable=too-many-arguments
         var_q: T.handle,
         var_k: T.handle,
         var_v: T.handle,
-        apply_rope: T.int32,
+        apply_rope: T.int64,
     ):
         T.func_attr(
             {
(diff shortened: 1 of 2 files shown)


xthomaswang:
Downstream impact:

This fixes MLC-LLM export paths that instantiate the RoPE PrimFunc through
Relax `nn.Module.export_tvm`, including `mlc_llm convert_weight` and
`mlc_llm compile`. Without this patch, those paths may produce a
not-well-formed Relax IR warning or fail at the frontend well-formedness
assert, depending on whether the downstream tree demotes the assert.

cchung100m (approved):
LGTM, thanks to @xthomaswang 😄

tlopex (approved):

PR #19433, opened 2026-04-23
[Relax][Frontend][TFLite] Fix dynamic FILL/SPLIT_V partial implementations

Aharrypotter:
This PR fixes partial TFLite frontend support for dynamic `FILL` and `SPLIT_V`.

Key changes:
- Allow `FILL` to accept runtime `dims` tensors by converting them with
  `relax.op.tensor_to_shape` before calling `relax.op.full`.
- Allow `SPLIT_V` to accept runtime `size_splits` tensors by decomposing the
  op into `cumsum` and `dynamic_strided_slice`.
- Use the TFLite output tuple arity as the source of truth for dynamic
  `SPLIT_V`, instead of relying on `size_splits` static shape information.
- Add TFLite frontend regression tests covering dynamic `FILL` import and
  dynamic `SPLIT_V` import/compile behavior.

This addresses the `FILL` and `SPLIT_V` items under the "Fix partial
implementations" section of #19412.

Validation:
```bash
python -m ruff check python/tvm/relax/frontend/tflite/tflite_frontend.py tests/python/relax/test_frontend_tflite.py
```

```bash
python -m pytest tests/python/relax/test_frontend_tflite.py \
-k "split_v_dynamic or fill_dynamic_dims" -v
```

Result:
- All checks passed

### DIFF
--- a/python/tvm/relax/frontend/tflite/tflite_frontend.py
+++ b/python/tvm/relax/frontend/tflite/tflite_frontend.py
@@ -2333,6 +2333,7 @@ class OperatorConverter:
     def convert_split_v(self, op):
         """SPLIT_V implementation."""
         input_tensors = self.get_input_tensors(op)
+        output_tensors = self.get_output_tensors(op)
 
         assert len(input_tensors) == 3, "input tensors length should be 3"
 
@@ -2353,7 +2354,9 @@ class OperatorConverter:
             # via strided_slice even for i == 0.
             zero = relax.const(np.array([0], dtype="int64"), "int64")
             padded_cumsum = relax.op.concat([zero, cumsum], axis=0)
-            num_splits = int(size_splits_expr.struct_info.shape[0])
+            # TFLite fixes the tuple arity in the graph, even when the split
+            # sizes themselves are supplied at runtime.
+            num_splits = len(output_tensors)
             rank = len(in_expr.struct_info.shape)
 
             # end_base is the full input shape; only split_axis changes per slice.
--- a/python/tvm/relax/frontend/tflite/tflite_frontend.py
+++ b/python/tvm/relax/frontend/tflite/tflite_frontend.py
@@ -31,7 +31,7 @@ import math
 import numpy as np
 
 import tvm
-from tvm import relax
+from tvm import relax, tirx
 from tvm.relax import op as _op
 
 from .tflite_flexbuffer import FlexBufferDecoder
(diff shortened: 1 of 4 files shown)


python/tvm/relax/frontend/tflite/tflite_frontend.py
```diff
@@ -2287,29 +2289,64 @@ def convert_split(self, op):
...
+            zero = relax.const(np.array([0], dtype="int64"), "int64")
+            padded_cumsum = relax.op.concat([zero, cumsum], axis=0)
+            # TFLite fixes the tuple arity in the graph, even when the split
+            # sizes themselves are supplied at runtime.
+            num_splits = len(output_tensors)
+            rank = len(in_expr.struct_info.shape)
+
+            # end_base is the full input shape; only split_axis changes per slice.
+            end_base = relax.op.shape_to_tensor(relax.op.shape_of(in_expr))
+            begin_base = relax.const(np.zeros((rank,), dtype="int64"), "int64")
```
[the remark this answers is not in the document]
Aharrypotter: Thanks, these are naming-only suggestions. Since the variables are local and the current intent is already documented by comments, I’d prefer to keep the diff minimal.

tests/python/relax/test_frontend_tflite.py
```diff
@@ -173,6 +173,30 @@ def main(x: R.Tensor((1, 30), dtype="float32")) -> R.Tensor((1, 3, 10), dtype="f
...
+            return tf.split(x, size_splits, axis=0)
+
+    cf = TfSplitVDynamic().func.get_concrete_function()
+    mod = _get_mod_from_cfunc(cf)
+    ir = mod.script()
+    assert "R.dynamic_strided_slice" in ir
+    assert "R.scatter_elements" in ir
+
+    # Regression guard: compile to catch the dynamic split lowering path.
+    tvm.compile(mod, tvm.target.Target("llvm"))
```
tlopex: Is this really needed? I mean`tvm.complie` here

tlopex (changes requested):
FILL frontend looks good, but dynamic FILL still breaks at legalize.
The issue is in `relax.full` legalize (`legalize_ops/create.py`): it assumes `call.args[0]` is a static `ShapeExpr` and passes it directly to `topi.full`. For dynamic FILL, this can be a `tensor_to_shape(...)` call, so legalize fails before `tvm.compile`.

Could you have a fix for `_full` to handle dynamic shape args just in this PR `dynamic_strided_slice` in `legalize_ops/index.py` should be a good template: keep the current static `ShapeExpr` path, and for dynamic `ShapeStructInfo`, match-cast to fresh symbolic vars and pass those vars to `topi.full`.

Then we can update `test_fill_dynamic_dims` to actually run the verify test

tests/python/relax/test_frontend_tflite.py
Aharrypotter: You're right, adding `tvm.compile(...)` to `test_split_v_dynamic` was over-testing on my side.

  That test is meant to cover the frontend lowering change for dynamic `SPLIT_V`, namely preserving fixed tuple arity and lowering to the expected Relax ops. Those
  properties are already directly captured by the IR assertions, so I removed the extra `tvm.compile(...) / verify(...)` from that test.

  For `dynamic FILL`, I kept the compile check intentionally. In that case, the imported Relax IR was already shape-aware, while the actual remaining issue was that
  dynamic-shape relax.full still failed in the legalize/compile path. So I updated the **legalization** for dynamic relax.full, and kept tvm.compile(...) in
 `test_fill_dynamic_dims` to cover that specific regression.

Aharrypotter:
Addressed both review points:

  - added `dynamic-shape legalization` for `relax.full` and updated `test_fill_dynamic_dims` to compile/verify
  - removed the unnecessary compile/verify path from `test_split_v_dynamic`

  One extra fix was needed in addition to _full: LegalizeOps was filtering out dynamic relax.full before the legalization hook could run, so I also relaxed the op’s shape gate to make the dynamic legalization path reachable. After that, dynamic FILL legalizes and compiles correctly.

Aharrypotter:
The imported Relax IR was not well formed before legalization. I fixed the frontend to **bind runtime fill dims to fresh shape vars before calling relax.full**

Aharrypotter:
Please take a look @tlopex

tlopex (approved):

PR #19445, opened 2026-04-25
[CI][REFACTOR] Decouple data.py from Jenkins script and docker images

tqchen:
## Summary

`ci/jenkins/data.py` was carrying two unrelated concerns: artifact bundle definitions for s3 staging, and the docker image tag registry. This PR decouples both from the Jenkinsfile-generated code, making each concern standalone and machine-readable without Python.

### Part 1 — Bundle refactor (`10950ac192`)

- `ci/scripts/jenkins/s3.py` gains `--bundle <name>` (repeatable) that resolves bundle names from `ci/jenkins/data.py::files_to_stash` at runtime; `--items` preserved for back-compat
- `ci/jenkins/templates/utils/macros.j2` `upload_artifacts` macro emits `--bundle <name>` flags instead of inlining the file list as `--items`
- Template callsites (cpu, gpu, arm) updated to pass `bundles=[...]` names
- Generated `.groovy` files regenerated — no `build/` paths appear in artifact upload blocks; only bundle names

### Part 2 — Docker image registry extraction (`6df22c09b9`)

- New `ci/docker-images.ini` (one `[ci_*]` section per image) is the single source of truth for image tags
- `docker/dev_common.sh::lookup_image_spec` reads the ini directly with `awk` — no Python invocation
- `ci/jenkins/data.py` drops the inline `docker_images` dict; loads it via `configparser` from the ini, preserving the same nested-dict shape so Jinja templates and the current `s3.py` module import keep working unchanged
- `data.py __main__` is now the bundle-resolver CLI (`python3 ci/jenkins/data.py <bundle> [...]` → file paths, one per line) — no more dual-purpose image-name lookup branch
- `ci/scripts/jenkins/open_docker_update_pr.py` updated to read/write `ci/docker-images.ini` instead of `data.py`

## Acceptance

```
# No build/ artifact paths in generated Jenkinsfiles (bundle names only):
grep -rn "build/lib\|build/cpptest" ci/jenkins/generated/   # → nothing

# generate.py is byte-identical after the change:
python3 ci/jenkins/generate.py  # → "no changes made" for all 5 groovy files

# dev_common.sh resolves same tag as before:
bash -c 'source docker/dev_common.sh && lookup_image_spec ci_cpu'
# → tlcpack/ci-cpu:20251130-061900-c429a2b1
```

### DIFF
--- a/ci/jenkins/data.py
+++ b/ci/jenkins/data.py
@@ -15,6 +15,19 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+"""Bundle registry for CI artifact stashing.
+
+Single source of truth for the file lists uploaded to / downloaded from S3 by
+``ci/scripts/jenkins/s3.py``. This module deliberately carries nothing else —
+docker image tags live in ``ci/jenkins/docker-images.ini`` and Jinja-template
+metadata (image platforms, AWS endpoints) lives in ``ci/jenkins/generate.py``.
+
+CLI: ``python3 ci/jenkins/data.py <bundle> [<bundle> ...]`` resolves bundle
+names to their file paths (one per line; exit 1 on unknown name). Used by
+``s3.py`` at Jenkins runtime and by any external caller that needs
+data-driven artifact lists.
+"""
+
 import sys
 
 files_to_stash = {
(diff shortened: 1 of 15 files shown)


tqchen:
cc @yongwww

tlopex (approved):

PR #19448, opened 2026-04-26
[REFACTOR] Use FFI types in runtime inline module-create wrapper signatures

tqchen:
## Summary

This PR cleans up the public inline wrapper API for runtime backend module creators introduced in #19447. The wrappers previously used a mix of `std::unordered_map` / `std::vector` / `std::string` and `ffi::*` types with conversion glue inside the wrapper body.

## Changes

- `ConstLoaderModuleCreate`: both parameters change from `std::unordered_map<std::string, T>` to `ffi::Map<ffi::String, T>`; the conversion loops in the wrapper body are removed (net −12 lines of glue code). The callsite in `codegen_vm.cc` is updated to build `ffi::Map` directly and passes the existing `ffi::Map<ffi::String, Tensor>` params argument through without a copy.
- `VulkanModuleCreate`: `source` parameter changes from `std::string` to `ffi::String`. The `SPIRVShader` smap remains `std::unordered_map` because `SPIRVShader` is a plain C++ struct, not FFI-storable.
- `OpenCLModuleCreate` (SPIRV overload): `spirv_text` changes from `const std::string&` to `ffi::String`. Same `SPIRVShader` constraint applies.
- `metal_module.h`: removes stale `<unordered_map>`, `<vector>`, `<memory>`, `<string>` includes left over from before #19447.

Internal `*ModuleCreateImpl` functions and the compiler-side `SourceModuleCreate` family are unchanged.

### DIFF
--- a/src/relax/backend/vm/codegen_vm.cc
+++ b/src/relax/backend/vm/codegen_vm.cc
@@ -465,30 +465,20 @@ void LinkModules(ObjectPtr<VMExecutable> exec, const ffi::Map<ffi::String, runti
                  const tvm::ffi::Module& lib, const ffi::Array<ffi::Module>& ext_libs) {
   // query if we need const loader for ext_modules
   // Wrap all submodules in the initialization wrapper.
-  std::unordered_map<std::string, std::vector<std::string>> const_vars_by_symbol;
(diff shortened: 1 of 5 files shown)


tqchen:
Reopening from tqchen fork — apache-branch push wasn't needed for this non-CI-pipeline refactor. Same diff at the new PR.

PR #19450, opened 2026-04-27
[BugFix][Relax][ONNX] Honor auto_pad in ConvTranspose converter

swjng:
## Motivation

The `ConvTranspose` ONNX → Relax converter silently drops the `auto_pad` attribute. `Conv` has dedicated handling (`onnx_frontend.py` lines 1383-1404), but `ConvTranspose` passes `pads` straight through, defaulting to 0 when the attribute is absent. Models that rely on `auto_pad=SAME_UPPER`/`SAME_LOWER`/`VALID` therefore produce the wrong output shape after import.

Minimal repro (against onnxruntime):

```python
node = helper.make_node(
    "ConvTranspose", ["X", "W"], ["Y"],
    strides=[2, 2], auto_pad="SAME_UPPER",
)
# input  X: [1, 1, 4, 4], W: [1, 1, 3, 3]
# ORT  -> Y.shape = (1, 1, 8, 8)   # input * stride
# TVM  -> Y.shape = (1, 1, 9, 9)   # auto_pad ignored, pads=0
```

## Fix

Compute `pads` from the ONNX spec equation before delegating to the Relax conv-transpose op:

```
total_pad[i] = stride[i] * (in[i] - 1)
             + output_padding[i]
             + (kernel[i] - 1) * dilation[i] + 1
             - in[i] * stride[i]
```

then split begin/end by `SAME_UPPER` vs `SAME_LOWER`. `VALID` becomes `pads=0`; `NOTSET` keeps the user-supplied `pads`. We deliberately do not reuse Conv's `autopad()` helper because it pads the input data, whereas ConvTranspose subtracts pads from the output.

The `output_shape` attribute (which, when set, also overrides `pads` per spec) remains unsupported; leaving that as a follow-up.

## Test plan

- [x] \`pytest tests/python/relax/test_frontend_onnx.py::test_conv_transpose_auto_pad\` — 6 new cases (3 modes × 2 strides) for 1D/2D/3D pass.
- [x] \`pytest tests/python/relax/test_frontend_onnx.py::test_conv_transpose\` — existing 8 parameterizations still pass.

### DIFF
--- a/python/tvm/relax/frontend/onnx/onnx_frontend.py
+++ b/python/tvm/relax/frontend/onnx/onnx_frontend.py
@@ -1663,13 +1663,60 @@ class ConvTranspose(OnnxOpConverter):
         else:
             raise NotImplementedError("Ndim > 5 not supported for convolution.")
 
+        spatial_dims = ndim - 2
+        strides = attr.get("strides", [1] * spatial_dims)
+        dilations = attr.get("dilations", [1] * spatial_dims)
+        output_padding = attr.get("output_padding", [0] * spatial_dims)
+        if "kernel_shape" in attr:
+            kernel_shape = list(attr["kernel_shape"])
+        else:
+            kernel_shape = [int(s) for s in inputs[1].struct_info.shape.values[2:]]
+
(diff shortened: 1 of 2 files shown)


tlopex (approved):
LGTM

PR #19454, opened 2026-04-27
[REFACTOR][CODEGEN] Phase out tvm_global_barrier_state and tvm_prepare_global_barrier

tqchen:
Phase out the legacy spin-on-global-memory CUDA barrier machinery
(`tvm_global_barrier_state` / `__tvm_prepare_global_barrier` / the
`tvm_global_barrier_kinit()` builtin and the `tirx.detect_global_barrier`
pass-config option). CUDA's native cooperative groups / grid sync
primitives cover the use case better; the bespoke implementation has
been dead in the active codegen pipelines.

This is a deletion-only refactor across 10 files (~−264 lines net):

- Public symbol constants in `include/tvm/runtime/device_api.h`
- TIR builtin op `tvm_global_barrier_kinit()` (`include/tvm/tirx/builtin.h`,
  `src/tirx/op/builtin.cc`)
- Pass config `tirx.detect_global_barrier` (`src/tirx/ir/transform.cc`)
- Entire kGlobal branch of `s_tir::ThreadSync` including
  `InitGlobalBarrier`, `MakeGlobalBarrier`, and supporting state in
  `ThreadSyncInserter`
- `CUDAPrepGlobalBarrier` runtime class + `CUDAModuleNode::GetGlobal()`
- `CodeGenCUDA::PrintStorageSync` "global" branch and the
  `VisitStmt_(EvaluateNode*)` override + 3 member fields
- Two Python pipeline opt-in blocks (s_tir/pipeline.py + adreno mirror)

No Python or test references to these symbols. Build clean
(260 targets), CUDA codegen 50/50 passed, tirx-base + tirx-transform
621 passed, IRF cpptests 8/8 passed.

### DIFF
--- a/include/tvm/runtime/device_api.h
+++ b/include/tvm/runtime/device_api.h
@@ -416,10 +416,6 @@ TVM_RUNTIME_DLL bool RuntimeEnabled(const ffi::String& target);
 namespace symbol {
 /*! \brief global function to set device */
 constexpr const char* tvm_set_device = "__tvm_set_device";
-/*! \brief Auxiliary counter to global barrier. */
-constexpr const char* tvm_global_barrier_state = "__tvm_global_barrier_state";
-/*! \brief Prepare the global barrier before kernels that uses global barrier. */
-constexpr const char* tvm_prepare_global_barrier = "__tvm_prepare_global_barrier";
 }  // namespace symbol
 
 }  // namespace runtime
--- a/include/tvm/tirx/builtin.h
+++ b/include/tvm/tirx/builtin.h
@@ -508,12 +508,6 @@ TVM_DLL const Op& tvm_warp_shuffle_down();
 TVM_DLL const Op& tvm_warp_activemask();
 
 /*!
- * \brief Initialize the global barrier.
- *  Call this at beginning of kernel that need global barrier.
- */
-TVM_DLL const Op& tvm_global_barrier_kinit();
-
-/*!
  * \brief See pesudo code
  *
  *  void tvm_thread_allreduce(UIntImm size, Expr source0, ..., Expr cond,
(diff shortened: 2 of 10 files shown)


tqchen:
Thanks for the review, addressing both comments:

**Comment 1 (PrintStorageSync silent no-op) — ACCEPTED**

You are correct. After removing the `global` barrier branch, `PrintStorageSync` silently did nothing for `sync == "global"`, which would produce incorrect code without any diagnostic. Added an explicit `TVM_FFI_THROW(InternalError)` for the `"global"` case (matching the existing error style in the file, e.g. `PrintStorageScope` already has `TVM_FFI_ICHECK_NE(scope, "global")`). Pushed as a follow-up commit (36a69d0cca).

**Comment 2 (redundant `<tvm/tirx/op.h>` include) — REJECTED**

The include is intentional and required. `make_zero` (used at line 300 in `ThreadSyncAfterWaitQueueInserter::VisitStmt_`) is declared in `<tvm/tirx/op.h>`. It was previously provided transitively through `../../tirx/transform/ir_utils.h`, which we removed as part of the cleanup. Neither `<tvm/tirx/builtin.h>` nor `<tvm/ir/op.h>` (which `builtin.h` includes) declare `make_zero` — that symbol lives only in `<tvm/tirx/op.h>`. Removing it causes a build error.

tqchen:
The \`#include <tvm/tirx/op.h>\` include is **required** and not redundant.

\`builtin.h\` includes \`<tvm/ir/op.h>\` but **not** \`<tvm/tirx/op.h>\`. The \`make_zero\` function is defined in \`<tvm/tirx/op.h>\` (line 300 of this file) and used by \`ThreadSyncAfterWaitQueueInserter::VisitStmt_\` at line 300.

Empirical verification: removing \`#include <tvm/tirx/op.h>\` and attempting a focused build fails with:
\`\`\`
error: 'make_zero' was not declared in this scope
\`\`\`

The include is necessary.

tlopex (approved):

PR #19456, opened 2026-04-27
[REFACTOR] Move source_utils.h into runtime/opencl

tqchen:
Move source_utils.h header into runtime/opencl directory with SplitKernels inlined. Removes the standalone source_utils.cc file and updates 2 include paths to reflect the new location.

## Test plan
- [x] Compilation of modified translation units verified
- [ ] Full CI suite validates broader impacts


tqchen:
Done: changed to `const std::string&` to avoid copying the source and delimiter strings on every call. The function only reads the parameters, so the reference form is safe.

spectrometerHBH (approved):

PR #19466, opened 2026-04-29
[Relax][Frontend][TFLite] Add BITCAST operator mapping

as4230:
This PR adds TFLite frontend support for the BITCAST operator which reinterprets a tensor's bytes as a different dtype without converting the underlying data.

The handler lowers BITCAST to relax.op.memory.view which aliases the input buffer with the new shape and dtype.

Frontend tests cover same-width (float32 -> int32, uint8 -> int8), width-changing smaller (int32[3] -> int16[3, 2]), and width-changing larger (int16[5, 2] -> int32[5]).

` python -m pytest tests/python/relax/test_frontend_tflite.py -k bitcast -v `

Addresses the BITCAST item under #19412.

### DIFF
--- a/python/tvm/relax/frontend/tflite/tflite_frontend.py
+++ b/python/tvm/relax/frontend/tflite/tflite_frontend.py
@@ -124,6 +124,7 @@ class OperatorConverter:
             "AVERAGE_POOL_2D": functools.partial(self.convert_pool2d, pool_type="average"),
             "BATCH_TO_SPACE_ND": self.convert_batch_to_space_nd,
             "BATCH_MATMUL": self.convert_batch_matmul,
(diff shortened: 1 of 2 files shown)


as4230:
@tlopex - Quick design check. Issue #19412 lists BITCAST and mentions needing a dedicated op. I went with `relax.op.memory.view` since it's byte-preserving and zero-conversion which seems to match TFLite BITCAST.

If you'd prefer a first-class `relax.bitcast` op that legalizes to `memory.view` or if there's a semantic gap with `memory.view` I'm not seeing I'm happy to implement whatever's missing.

tlopex (approved):
LGTM! No worries, your implementation is right @as4230

Issue #19437, opened 2026-04-24
[Bug] [Relax][ONNX] CumSum ignores axis parameter, always reduces along axis 0


wuyii8941:
## Expected behavior

`CumSum` with `axis=1` on a 2D input should compute cumulative sums along each row.

## Actual behavior

TVM hardcodes `axis=0` when the axis input is a `relax.Var` (which is the standard ONNX format — axis is an input tensor, not an attribute). The cumulative sum is always computed along axis 0 regardless of the actual axis value.

## Reproduction

```python
import numpy as np
import onnx
from onnx import helper, TensorProto
import onnxruntime as ort
import tvm
from tvm import relax
from tvm.relax.frontend.onnx import from_onnx

x = np.array([[1, 2, 3, 4],
               [5, 6, 7, 8],
               [9, 10, 11, 12]], dtype=np.float32)

X = helper.make_tensor_value_info("X", TensorProto.FLOAT, [3, 4])
A = helper.make_tensor_value_info("axis", TensorProto.INT32, [])
Y = helper.make_tensor_value_info("Y", TensorProto.FLOAT, None)
node = helper.make_node("CumSum", ["X", "axis"], ["Y"])
graph = helper.make_graph([node], "test", [X, A], [Y])
model = helper.make_model(graph, opset_imports=[helper.make_opsetid("", 14)])
model = onnx.shape_inference.infer_shapes(model)

axis_val = np.array(1, dtype=np.int32)

# ORT: correct (cumsum along axis 1)
sess = ort.InferenceSession(model.SerializeToString())
ort_out = sess.run(None, {"X": x, "axis": axis_val})[0]

# TVM: wrong (cumsum along axis 0)
mod = from_onnx(model)
exe = tvm.relax.build(
    tvm.ir.transform.Sequential([relax.transform.LegalizeOps()])(mod), target="llvm"
)
vm = tvm.relax.VirtualMachine(exe, device=tvm.cpu())
tvm_out = vm["main"](
    tvm.runtime.tensor(x, device=tvm.cpu()),
    tvm.runtime.tensor(axis_val, device=tvm.cpu()),
).numpy()

print("Expected (cumsum axis=1):")
print(ort_out)
# [[ 1  3  6 10]
#  [ 5 11 18 26]
#  [ 9 19 30 42]]

print("Got (cumsum axis=0 — WRONG):")
print(tvm_out)
# [[ 1  2  3  4]
#  [ 6  8 10 12]
#  [15 18 21 24]]

[...]

PR #19467, opened 2026-04-29
[Relax][ONNX] Fix CumSum axis handling: support runtime axis tensor, error on multi-element axis

cchung100m:
Hi Committers,

This PR is trying to fix issues https://github.com/apache/tvm/issues/19437. Any suggestions would be appreciated if you are available.

### Root Cause
The original CumSum converter always defaulted to axis=0 when the axis input was a relax.Var (i.e., a runtime tensor), ignoring the actual runtime value. This led to incorrect behavior and did not comply with the ONNX specification.

### Solutions
Update CumSum._impl_v14 to:
- Check if the axis input is a Constant: require it to have exactly one element, otherwise raise an error.
- If the axis input is a relax.Var, raise an error instead of always defaulted to axis=0.


tlopex (approved):
LGTM! Thanks

PR #19478, opened 2026-04-29
[REFACTOR][SCRIPT] Move script_printer.h → script/printer/config.h; rename TVM_SCRIPT_REPR; replace dialect fields with extra_config

tqchen:
## Summary

`include/tvm/ir/script_printer.h` placed `PrinterConfig` and `TVMScriptPrinter` under `tvm/ir/` — these are printer-layer concepts, not IR-specific.  Move to `include/tvm/script/printer/config.h` alongside the printer's other public headers (`doc.h`, `ir_docsifier.h`, `ir_docsifier_functor.h`).

- **Rename `TVM_SCRIPT_REPR` → `TVM_REGISTER_SCRIPT_AS_REPR`**: the macro registers Script as the `kRepr` callback for an object type and installs the per-type `TVMScriptPrinter::vtable()` dispatch entry.  The new name is explicit about the registration semantic and aligns with the `TVM_REGISTER_*` family.

- **Drop dialect-specific `PrinterConfig` fields** (`tir_prefix`, `relax_prefix`, `show_all_struct_info`, `buffer_dtype`) in favor of a generic `ffi::Map<String, Any> extra_config` keyed by `"<dialect>.<knob>"` (e.g. `"tirx.prefix"`, `"relax.show_all_struct_info"`).  Each dialect reads its own keys via `GetExtraConfig<T>(key, fallback)` with defaults at the call site; the shared core never names a specific dialect through hard-coded fields.  `ir_prefix` and `module_alias` also flip from `std::string` to `ffi::String` for consistency.  `RedirectedReprPrinterMethod` is de-inlined into `src/script/printer/config.cc` to keep `<tvm/runtime/logging.h>` out of the public header.

- ~6 header includers and ~80 macro use sites updated; Python config wrappers updated to write into `extra_config` with dotted keys.


tqchen:
Folded into #19479 — single coherent TVMScript dialect-friendly refactor PR. Commits absorbed; this PR's branch can be deleted.

Issue #18363, opened 2025-10-08
[bug][relax.frontend.torch] FFI segfault in tvm::relax::Tuple::Tuple when importing torch.export graph with 4D advanced-indexing write (aten.index_put_) and tuple outputs


tinywisdom:
### Summary

Importing a torch.exported program into TVM Relax triggers a segmentation fault inside FFI during construction of a Relax Tuple. The minimal model performs a 4D advanced indexing write using two integer index tensors on the last two dims (L[..., idx, idx] = ...) and returns a Python tuple of tensors (x[..., :1], L). The exported graph is free of RNG ops (no randn), so the crash appears related to the combination of aten.index_put_ lowering and tuple output construction.

### Actual behavior

```
[1] torch.export ...
=== Exported ops ===
... (as above)
[2] tvm.relax.frontend.torch.from_exported_program ...
!!!!!!! TVM FFI encountered a Segfault !!!!!!! 
  ...
  tvm::relax::Tuple::Tuple(tvm::ffi::Array<tvm::RelaxExpr, void>, tvm::Span) [clone .cold]
  ...
Segmentation fault (core dumped)
```

### Environment

- OS: (Ubuntu 22.04.4 LTS (x86_64))
- TVM version: (release v0.21.0)
- Python: (3.10.16)
- LLVM: (17.0.6)
- Pytorch: (2.8.0)

### Steps to reproduce

```python
import os
os.environ["CUDA_VISIBLE_DEVICES"] = ""  # avoid GPU warnings

import torch
import torch.nn as nn
import torch.nn.functional as F
from torch.export import export as torch_export
from tvm.relax.frontend.torch import from_exported_program

class M4D(nn.Module):
    def forward(self, x):
        B, K, N = 2, 3, 5
        L = x.new_zeros(B, K, N, N)       # tensor construct only; no randomness
        idx = torch.arange(N, device=x.device)

        # key trigger: gather diagonal, apply smooth monotonic transform, scatter back
        diag = L[..., idx, idx]           # shape: [B, K, N]
        diag = F.elu(diag) + 1.0 + 1e-8   # avoid all-zero; any smooth transform works
        L[..., idx, idx] = diag           # advanced indexing write (two int index tensors)

        # key trigger: return a Python-level tuple (two tensors)
        return x[..., :1], L

if __name__ == "__main__":
    torch.manual_seed(0)
    m = M4D().eval()
    ex_in = torch.zeros(2, 3, 5)  # any input; ensures no randn exported

[...]

PR #19488, opened 2026-05-01
[Relax][PyTorch] Fix segfault in from_exported_program when model uses index_put_ with tuple output

cchung100m:
Hi Committers,

This PR is trying to fix issues https://github.com/apache/tvm/issues/18363. Any suggestions would be appreciated if you are available.

### Root Cause
- When an ExportedProgram's FX graph output node returns a **nested Python tuple** (e.g., buffer mutation outputs + user-defined tuple returns), `_translate_fx_graph()` passes the raw nested structure directly to the Relax FFI Tuple constructor. 
- The C++ Array<Expr> initializer cannot handle heterogeneous/nested Python containers, causing a segmentation fault at `expr.cc`. 
- Additionally, index_put_ (in-place write op) did not update self.env to alias the source tensor to the mutated output, causing subsequent FX nodes that read the same tensor to observe **stale pre-mutation values**.

### Solution
- exported_program_translator.py
  - Added static method `_flatten_output_args()` that recursively walks any Python `tuple/list`, collects only `relax.Expr`   leaves, and preserve explicit None outputs as Relax null objects.
  - Replaced the fragile `assert isinstance(output_args, tuple | relax.Tuple)` guard with a call to `_flatten_output_args()`, producing a clean flat tuple of `relax.Expr` before FFI construction.
- base_fx_graph_translator.py
  - In `_index_put()`, after emitting the `relax.op.index_put(...)` call, added an env alias update: `self.env[source_node] = output` when the target op name starts with `index_put_`, preserving correct in-place mutation semantics for downstream FX nodes.

### DIFF
--- a/python/tvm/relax/frontend/torch/base_fx_graph_translator.py
+++ b/python/tvm/relax/frontend/torch/base_fx_graph_translator.py
@@ -1860,7 +1860,7 @@ class BaseFXGraphImporter(metaclass=abc.ABCMeta):
 
         output = self.block_builder.emit(relax.op.index_put(tensor, indices, values, accumulate))
 
-        target_name = getattr(node.target, "__name__", "")
+        target_name = node.target if isinstance(node.target, str) else getattr(node.target, "__name__", "")
         if target_name.startswith("index_put_") and len(node.args) > 0:
             from torch import fx
 
--- a/python/tvm/relax/frontend/torch/base_fx_graph_translator.py
+++ b/python/tvm/relax/frontend/torch/base_fx_graph_translator.py
@@ -1857,7 +1857,17 @@ class BaseFXGraphImporter(metaclass=abc.ABCMeta):
                 indices = relax.Tuple(processed_indices)
             else:
                 indices = relax.Tuple(indices)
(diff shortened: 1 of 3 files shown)


cchung100m:
Hi @tlopex @mshr-h 

This PR is trying to fix issues https://github.com/apache/tvm/issues/18363. Any suggestions would be appreciated if you are available. 😄

tlopex (commented):
Thanks for the fix. One correctness case I’d like to clarify before approval: `_index_put` only updates `self.env[node.args[0]]`. If the mutated value is an alias of another FX node, later reads through the original node may still see the pre-mutation value.

Could you add a small regression test for mutation through an alias?

cchung100m:
Hi @tlopex 

Thank you for the feedback. I updated the implementation so in-place `index_put_` now propagates to all env entries that alias the same underlying expr, keeping post-mutation reads consistent. I also added a focused regression test for mutation through an alias (`alias(x)` → `[index_put_` → read both original and alias outputs).

tlopex (approved):

cchung100m:
Thanks to @tlopex 😄

PR #19492, opened 2026-05-01
[BugFix][Relax] Add legalize for isnan, isinf, isfinite

### LINKED ISSUE #19452: [Bug] `R.isnan`, `R.isinf`, `R.isfinite` crash at build time — CodeGenVM intrinsic not implemented
### Expected behavior

`R.isnan()`, `R.isinf()`, and `R.isfinite()` should compile and produce correct boolean tensor outputs on both CPU and CUDA targets.

### Actual behavior

All three ops crash during `relax.build()` with:

```
InternalError: CodeGenVM cannot handle this intrinsic now:
Op(relax.isnan)
```

This occurs on **both** CPU (llvm) and CUDA targets.

### Environment

- **TVM**: main branch @ commit `0b0afd8dd` (2026-04-25)
- **OS**: Ubuntu 20.04

### Steps to reproduce

```python
import numpy as np
import tvm
from tvm import relax
import tvm.relax.op as R

bb = relax.BlockBuilder()
x = relax.Var("x", relax.TensorStructInfo((4, 8), "float32"))
with bb.function("main", [x]):
    with bb.dataflow():
        out = bb.emit(R.isnan(x))
        gv = bb.emit_output(out)
    bb.emit_func_output(gv)
mod = bb.get()

pipeline = tvm.ir.transform.Sequential([relax.transform.LegalizeOps()])
mod_l = pipeline(mod)
exe = relax.build(mod_l, target="llvm")  # CRASH
```

Same crash for `R.isinf(x)` and `R.isfinite(x)`.

### Notes

- These ops are registered in the Relax op registry and can be used to construct valid IR modules.
- The PyTorch FX frontend already supports importing `isnan`/`isinf`/`isfinite` (PR #17659), but the backend code generation was not updated to handle them — so models using these ops can be imported but not compiled.
- Similar CodeGenVM gaps have been reported and fixed before for other ops (e.g., #14437, #17867).

as4230:
The Relax ops `relax.isnan`, `relax.isinf`, and `relax.isfinite` are registered in the op registry and emit valid IR, but lack FLegalize entries so LegalizeOps() leaves them unlowered and relax.build() crashes at codegen with:

    InternalError: CodeGenVM cannot handle this intrinsic now:
    Op(relax.isnan)

Same crash on `isinf` and `isfinite`, on both LLVM and CUDA targets.

The TOPI implementations already exist (python/tvm/topi/math.py). The fix is three register_legalize calls following the _call_topi_without_attr pattern used by the other unary ops in the same file.

Fixes #19452.


tlopex (approved):
Thanks for the fix!

PR #19494, opened 2026-05-02
[Web] Add support for OPFS

akaashrp:
Add OPFS as an alternative caching mechanism for artifacts: https://developer.mozilla.org/en-US/docs/Web/API/File_System_API/Origin_private_file_system.


akaashrp:
cc @guan404ming for review

guan404ming (approved):
Looks great, thanks!

PR #19497, opened 2026-05-03
[BugFix][Relax] Fix scatter_elements and scatter_nd CUDA compilation

### LINKED ISSUE #19451: [Bug]`scatter_elements` and `scatter_nd` fail to compile for CUDA target
### Expected behavior

`R.scatter_elements()` and `R.scatter_nd()` should compile and run correctly on the CUDA target, as they do on CPU (llvm).

### Actual behavior

Both ops crash during `relax.build()` with:

```
RuntimeError: Memory verification failed with the following errors:
    Variable `out_buf` is directly accessed by host memory
    (it is not contained in a thread environment or in the function arguments.
    ...
  Did you forget to bind?
```

The generated TIR for these ops uses `T.parallel` loops without any GPU thread binding (`blockIdx`/`threadIdx`), so the memory verifier correctly rejects them. This means **the TIR legalization for scatter ops does not produce GPU-compatible code**.

CPU (llvm) works correctly for all tested shapes and axes.

### Environment

- **TVM**: main branch @ commit `0b0afd8dd` (2026-04-25)
- **OS**: Ubuntu 20.04, CUDA 12.2
- **GPU**: NVIDIA GPU (sm_75)

### Steps to reproduce

```python
import numpy as np
import tvm
from tvm import relax
import tvm.relax.op as R
from tvm.s_tir import dlight


def _make_var(name, shape, dtype="float32"):
    return relax.Var(name, relax.TensorStructInfo(shape, dtype))


shape = (4, 8)
axis = 0

bb = relax.BlockBuilder()
x = _make_var("x", shape, "float32")
idx = _make_var("idx", shape, "int64")
upd = _make_var("upd", shape, "float32")
with bb.function("main", [x, idx, upd]):
    with bb.dataflow():
        out = bb.emit(R.scatter_elements(x, idx, upd, axis=axis))
        gv = bb.emit_output(out)
    bb.emit_func_output(gv)
mod = bb.get()

# CPU works
pipeline_cpu = tvm.ir.transform.Sequential([relax.transform.LegalizeOps()])
mod_cpu = pipeline_cpu(mod)
exe_cpu = relax.build(mod_cpu, target="llvm")  # OK

# CUDA crashes
pipeline_cuda = tvm.ir.transform.Sequential([
    relax.transform.LegalizeOps(),
    dlight.ApplyDefaultSchedule(dlight.gpu.Fallback()),
])
with tvm.target.Target("cuda"):
    mod_cuda = pipeline_cuda(mod)
exe_cuda = relax.build(mod_cuda, target="cuda")  # CRASH
```

### Notes

- Tested across 10 shape/axis combinations — **100% failure rate on CUDA**, 100% success on CPU.
- `R.scatter_nd()` has the same issue.
- `R.gather_elements()` and `R.gather_nd()` compile and run correctly on CUDA.
- The generated TIR shows the scatter loop uses `T.parallel` without thread binding, which is the root cause. The legalization likely needs a GPU-aware implementation similar to what `gather_elements` has.

as4230:
`topi.scatter_elements` and `topi.scatter_nd` emit bare `T.parallel` loops in their te.extern IRBuilder bodies which trips `VerifyMemory` on CUDA targets:

    RuntimeError: Memory verification failed
    ...
    Did you forget to bind?

CPU (LLVM) is unaffected.

This fix makes the IRBuilder body in both `topi/scatter_elements.py` and `topi/scatter.py` target-aware. When `Target.current()` is a GPU target it emits thread bindings instead of `T.parallel`.

Fixes #19451.

### DIFF
--- a/python/tvm/relax/transform/legalize_ops/manipulate.py
+++ b/python/tvm/relax/transform/legalize_ops/manipulate.py
@@ -235,10 +235,16 @@ def _meshgrid(bb: BlockBuilder, call: Call) -> Expr:
     )
 
 
+def _is_gpu_target():
+    target = tvm.target.Target.current(allow_none=True)
+    return target is not None and "gpu" in target.keys
+
+
 @register_legalize("relax.scatter_elements")
 def _scatter_elements(bb: BlockBuilder, call: Call) -> Expr:
+    te_func = topi.gpu.scatter_elements if _is_gpu_target() else topi.scatter_elements
     return bb.call_te(
-        topi.scatter_elements,
+        te_func,
         call.args[0],
         call.args[1],
         call.args[2],
@@ -250,10 +256,12 @@ def _scatter_elements(bb: BlockBuilder, call: Call) -> Expr:
 @register_legalize("relax.scatter_nd")
 def _scatter_nd(bb: BlockBuilder, call: Call) -> Expr:
     # TODO(relax-team): Support native scatter_nd without te extern
+    base_te = topi.gpu.scatter_nd if _is_gpu_target() else topi.scatter_nd
+
     def scatter_nd(data, indices, updates, reduction):
         axes = list(range(len(indices.shape)))
         indices = topi.transpose(indices, axes[-1:] + axes[:-1])
-        return topi.scatter_nd(data, indices, updates, reduction)
+        return base_te(data, indices, updates, reduction)
 
     return bb.call_te(
         scatter_nd,
--- a/python/tvm/topi/gpu/__init__.py
+++ b/python/tvm/topi/gpu/__init__.py
@@ -20,4 +20,6 @@
 """GPU specific declaration."""
 
 from .scan import cumsum, cumprod
+from .scatter_elements import scatter_elements
+from .scatter_nd import scatter_nd
 from .sort import *
--- a/python/tvm/topi/gpu/scatter_elements.py
+++ b/python/tvm/topi/gpu/scatter_elements.py
@@ -1,0 +1,162 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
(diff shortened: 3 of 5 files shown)


as4230:
@tlopex I considered going with a dispatch pass like [DispatchSortScan](https://github.com/apache/tvm/blob/main/python/tvm/relax/backend/dispatch_sort_scan.py) but ended up fixing the lowering inline. The CPU and GPU paths here use the same algorithm so adding a separate topi/gpu/ implementation plus dispatch felt like mostly boilerplate and duplication.

tlopex:
Thanks for working on this. I agree that fixing this in the scatter lowering is the right layer, and this is much closer to what I had in mind than a generic `T.parallel` cleanup pass.

That said, I’d prefer the fix to restore a proper target-specific TOPI dispatch rather than make the generic `topi.scatter_elements` / `topi.scatter_nd` implementation target-aware internally.

The root issue in #19451 is that Relax legalization lost the old CUDA scatter lowering path: the op is lowered through `te.extern`, and the generic IRBuilder emits bare `T.parallel`, which is not valid GPU TIR. Since this is a target-specific legality issue, I think the more fundamental fix is to add a GPU TOPI implementation, e.g. under `topi/gpu/`, and dispatch to it from `relax/transform/legalize_ops/manipulate.py` when the current target is GPU. That is similar in spirit to the old Relay CUDA scatter path that was removed during the Relay cleanup.

Even if the CPU and GPU algorithms are mostly the same, the generated TIR is not target-neutral: GPU needs explicit `blockIdx/threadIdx` bindings, and reductions may need GPU-specific handling. Keeping that in a GPU-specific implementation avoids making the generic TOPI path depend on `Target.current()`, and it keeps `VerifyMemory` useful as a check that GPU lowering produced valid GPU TIR in the first place.

So my preference would be:
1. keep the generic TOPI scatter lowering target-neutral for CPU/default use;
2. add `topi/gpu` scatter lowering that emits thread-bound TIR;
3. dispatch to the GPU implementation from Relax legalization;
4. add CUDA compile/build regression tests for both `scatter_elements` and `scatter_nd`.

I don’t think we need the broad pass from #19363, but I would like this PR to fix the issue through explicit GPU scatter lowering rather than inline target checks in the generic implementation.

as4230:
@tlopex Thanks for the feedback and telling me your preferences. I reworked the PR. The generic topi.scatter_elements and topi.scatter_nd are untouched, GPU lowering lives in topi/gpu/scatter_elements.py and topi/gpu/scatter_nd.py with explicit blockIdx/threadIdx bindings, and dispatch happens in legalize_ops/manipulate.py. I also added a CUDA compile/build regression tests for both ops next to the existing scatter tests in test_transform_legalize_ops_manipulate.py.

tlopex (approved):
LGTM

PR #19498, opened 2026-05-03
[BugFix][Relax][ONNX] Resolve param Vars in Concat to handle mixed Shape/Tensor inputs

swjng:
## Description

When `from_onnx(model, keep_params_in_input=True)` is used, every ONNX initializer becomes a `relax.Var` instead of a `relax.Constant`. The `Concat` handler's `is_shape_like()` check only recognizes `relax.ShapeExpr` and 1D-int64 `relax.Constant`, so a 1D-int64 shape value loaded as a Var is no longer recognized.

When such a Var is concatenated with a `ShapeExpr` — the standard pattern for dynamic-batch `Reshape` in PyTorch-exported ONNX models — the heterogeneous `Tuple(ShapeExpr, Tensor)` is rejected by `relax.op.concat` with:

```
InternalError: Op(relax.concat) expects the input to be a Tuple of Tensors.
However, the given input is R.Tuple(R.Shape([N]), R.Tensor((1,), dtype="int64"))
```

This effectively breaks `keep_params_in_input=True` for any model with dynamic-batch `Reshape` (extremely common in PyTorch ONNX exports).

## Fix

Run each `Concat` input through the existing `get_constant` helper before the `is_shape_like` check. This resolves any `Var` that maps to a known param back to its baked `Constant`, restoring the all-shape-like fast path.

## Minimal repro

An 8-node ONNX graph (`Shape` → `Slice` → `Concat([dyn_n, [12]])` → `Reshape`) fails with `keep_params_in_input=True` before this PR and passes after. A regression test (`test_concat_with_param_shape_value`) covers this pattern.

## Testing

```
pytest tests/python/relax/test_frontend_onnx.py -k concat
```

9 passed (1 new + 8 existing).

### DIFF
--- a/python/tvm/relax/frontend/onnx/onnx_frontend.py
+++ b/python/tvm/relax/frontend/onnx/onnx_frontend.py
@@ -1014,6 +1014,10 @@ class Concat(OnnxOpConverter):
     @classmethod
     def _impl_v13(cls, bb, inputs, attr, params):
         axis = attr.get("axis", 0)
+        # Resolve any param Vars to their baked Constant values so that 1D-int64
+        # shape values loaded under keep_params_in_input=True can take the
+        # shape-like fast path below.
+        inputs = [get_constant(inp, params) for inp in inputs]
 
         def is_shape_like(x: Any) -> bool:
             if isinstance(x, relax.ShapeExpr):
--- a/python/tvm/relax/frontend/onnx/onnx_frontend.py
+++ b/python/tvm/relax/frontend/onnx/onnx_frontend.py
@@ -1014,10 +1014,7 @@ class Concat(OnnxOpConverter):
     @classmethod
     def _impl_v13(cls, bb, inputs, attr, params):
         axis = attr.get("axis", 0)
-        # Resolve any param Vars to their baked Constant values so that 1D-int64
-        # shape values loaded under keep_params_in_input=True can take the
-        # shape-like fast path below.
-        inputs = [get_constant(inp, params) for inp in inputs]
+        _, param_dict = params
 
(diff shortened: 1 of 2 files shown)


tlopex:
Hi @swjng 
Thanks for the fix! 

However, `Concat._impl_v13` now calls `get_constant` on every input unconditionally. `get_constant` has a side effect: when the input is a param `Var`, it updates `graph_nodes[name]` to a baked `relax.Constant`.

This fixes the shape-construction case, but it can also change the semantics of ordinary tensor concat under `keep_params_in_input=True`. For example, an ONNX graph with `Concat(x, weight)` should keep `weight` as a runtime parameter. With this change, `weight` may be folded into the IR as a constant, so the compiled function no longer depends on the runtime value passed for that parameter.

Could we narrow the fix so param resolution is used only for the all-shape-like fast path, preferably without mutating `_nodes`? The tensor fallback should continue using the original `inputs`. It would also be good to add a regression test for normal tensor `Concat(input, initializer)` with `keep_params_in_input=True`, verifying that the initializer remains a real parameter input.

swjng:
Thanks. Narrowed it as suggested in the follow-up commit: a local non-mutating `resolve` returns a Constant only for 1D-int64 param Vars, and only the all-shape-like fast path consumes the resolved list. The tensor fallback uses the original `inputs`, so `Concat(input, weight)` still references `weight` as a runtime Var.

Added `test_concat_with_param_tensor_keeps_runtime_param` covering exactly that case (asserts `w` stays in `main`'s params and is recovered by `detach_params`).

tlopex (approved):
LGTM! Thank you

Issue #19505, opened 2026-05-04
[Relax][ONNX] Use ONNX Backend Tests to improve frontend coverage


mshr-h:
This is my brain dump for a possible improvement to the Relax ONNX frontend tests.

Currently, many ONNX frontend tests in Relax manually construct small ONNX models and compare the imported Relax program against ONNX Runtime. This makes the test suite somewhat ad-hoc. As coverage grows, it becomes easy to miss ONNX spec corner cases, opset-specific behavior, and edge cases that are already covered by the official ONNX test suite.

ONNX provides an official Backend Test suite, including node-level tests for operator behavior:
- https://github.com/onnx/onnx/blob/main/docs/OnnxBackendTest.md

I think it would be worth exploring whether Relax's ONNX frontend tests can reuse those tests as a more systematic source of coverage.

The goal is not necessarily to expose TVM as a full ONNX backend. Rather, the idea is to use the ONNX Backend Tests as a testing mechanism for the importer:

- improve coverage of ONNX operator semantics
- reduce duplicated hand-written operator tests
- catch edge cases that are easy to miss with random-input tests
- align frontend correctness testing more closely with the ONNX specification
- keep TVM-specific frontend tests only for behavior that ONNX Backend Tests do not cover, such as importer options, parameter handling, name sanitization, and Relax-specific integration

Contributions, suggestions, or alternative ideas would be very welcome.

This issue is intended to collect thoughts and discuss whether this direction makes sense before working out a detailed design.

cc @KJlaccHoeUM9l @junrushao

Aharrypotter:
Completely agree.

This will really clean up the relax/frontend/onnx workflow. By hooking into the official suite, we can focus entirely on the actual lowering logic and avoid spending time hand-writing tests and making up fake data.

I’d love to take this on and submit a PoC PR for the basic runner.

cc @KJlaccHoeUM9l @junrushao

Aharrypotter:
I've started working on this and have a preliminary PoC ready.

**What it does:**
- Introduces a test runner that feeds ONNX Backend Test node-level cases into `relax.frontend.onnx.from_onnx`
- Compiles the imported Relax module with the VM and compares against ONNX's expected protobuf outputs
- Restricts the runner to `NodeModel` tests only (filters out real/simple/PyTorch model tests via `BackendTest._test_items` with layout validation)
- Adds `onnx_backend` pytest marker and `SKIP_SLOW_TESTS` support

**Current status:**
- 2406 node tests collected
[...]

PR #19515, opened 2026-05-06
[Relax][ONNX] Add ONNX Backend Tests for systematic frontend coverage

Aharrypotter:
## Summary

Introduce a test runner that reuses the official ONNX Backend Test Suite to systematically verify the Relax ONNX importer. This complements the existing hand-written tests in `test_frontend_onnx.py` by providing spec-aligned coverage of standard ONNX operator semantics.

Towards #19505

## Motivation

The existing `test_frontend_onnx.py` has 187 hand-written tests that validate TVM-specific importer behavior (parameter handling, name sanitization, dynamic shapes, Relax IR structure). However, it relies on ONNX Runtime as the reference and cannot systematically cover all edge cases defined in the ONNX specification.

The ONNX Backend Test Suite provides 1653+ node-level tests with protobuf reference inputs/outputs. It is the industry standard for validating ONNX importers/exporters (used by ONNX Runtime, TensorFlow, PyTorch). Reusing it gives Relax a living, upstream-aligned correctness baseline.

## What this PR adds

- `tests/python/relax/test_frontend_onnx_backend.py` — a backend adapter (`TVMRelaxBackend`) that implements the `onnx.backend.base.Backend` interface, wiring `from_onnx()` → `DecomposeOpsForInference()` → `LegalizeOps()` → `tvm.compile()` → `VirtualMachine`.

## Coverage

72 operators with 388 test cases, all passing. Only operators where every ONNX node test passes are included — no xfail markers.

Operators not yet covered include: cast (exotic dtypes), reduce ops (edge cases), reshape/resize/attention (complex behavior), quantization, and several others with known importer gaps. These can be added incrementally as the importer improves.

## Test results

388 passed, 3216 skipped (CUDA variants + operators not yet in allowlist), 0 failed, 0 xfailed

## CI impact

- New test file is not added to any existing CI test shard by default
- Full suite (388 tests) is lightweight on CPU-only runners

## Design decisions

- **Coexistence with existing tests**: `test_frontend_onnx.py` remains unchanged. Backend tests cover standard ONNX semantics; hand-written tests continue to cover TVM-specific behavior (dynamic shapes, Relax IR structure, importer options).
- **Public API only**: uses `backend_test.include()` with `^`-anchored regex patterns. No access to private ONNX APIs.
- **No xfail**: only include operators that fully pass. Uncovered operators are documented in code comments and this PR description. Follow-up PRs can expand coverage as importer gaps are fixed.
- **Prefix conflict handling**: `include()` patterns use `^test_{op}(?:_.*)?(?:_cpu|_cuda)$`, which can cause false matches when a short op name is a prefix of a longer one (e.g. `log` vs `log_softmax`). Affected ops (`log`, `max`, `relu`) are excluded until a more precise matching strategy is adopted.

### DIFF
--- a/tests/python/relax/test_frontend_onnx_backend.py
+++ b/tests/python/relax/test_frontend_onnx_backend.py
@@ -14,14 +14,19 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+# pylint: disable=invalid-name
 """
-ONNX Backend Tests for Relax ONNX Frontend
-===========================================
+ONNX Backend Tests
+===================
+Systematically verify the Relax ONNX importer using the official ONNX
+Backend Test Suite (node-level tests only).  Each test loads a small
+ONNX model with protobuf reference inputs/outputs and checks that the
+Relax-imported model produces numerically correct results.
 
-Uses the official ONNX Backend Test Suite to systematically verify the
-Relax ONNX importer against the ONNX specification.
+Only ``onnx.backend.test.data.node`` tests are registered here; real,
+simple, and PyTorch model tests are out of scope for importer-level
+semantic verification.
 
-Phase 1 (PoC): simple element-wise operators only.
 """
 
 import numpy as np
@@ -33,13 +38,16 @@ import tvm
 from tvm import relax
 from tvm.relax.frontend.onnx import from_onnx
 
+# ---------------------------------------------------------------------------
+# Backend adapter
+# ---------------------------------------------------------------------------
+
 
 class TVMRelaxBackendRep(BackendRep):
-    """Compiled Relax VM representation for running ONNX models."""
+    """Compiled Relax VM representation for running an ONNX model."""
 
     def __init__(self, mod, params, func_param_names, graph_input_names):
         super().__init__()
-        self._mod = mod
         self._params = params
         self._func_param_names = func_param_names
         self._graph_input_names = graph_input_names
@@ -49,22 +57,19 @@ class TVMRelaxBackendRep(BackendRep):
         self._vm = relax.VirtualMachine(ex, tvm.cpu())
 
     def run(self, inputs, **kwargs):
-        # Build a name -> array mapping from the positional inputs list.
-        # The runner loads inputs matching model.graph.input order, but only
-        # for the number of .pb files found (non-initializer inputs).
+        # Map positional inputs to names.  The runner loads one .pb per
+        # non-initializer input, aligned with model.graph.input order.
         input_map = {}
         for i, arr in enumerate(inputs):
             if i < len(self._graph_input_names):
                 input_map[self._graph_input_names[i]] = arr
 
-        # Build input_list matching the Relax function's param order.
-        # User inputs come first (by func_param_names[:num_input]),
-        # then weight params from self._params.
+        # Build the argument list matching the Relax function's param order:
+        # user inputs first, then weight params from self._params.
         input_list = []
         for name in self._func_param_names:
             if name in input_map:
                 input_list.append(input_map[name])
-
         if self._params and "main" in self._params:
             input_list += self._params["main"]
 
@@ -72,7 +77,6 @@ class TVMRelaxBackendRep(BackendRep):
         self._vm.invoke_stateful("main")
         output = self._vm.get_outputs("main")
 
-        # Normalize output to tuple of numpy arrays.
         if isinstance(output, (tvm.runtime.Tensor, np.ndarray)):
             return (output.numpy() if hasattr(output, "numpy") else output,)
         if isinstance(output, (tuple, list)):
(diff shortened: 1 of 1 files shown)


Aharrypotter:
cc @mshr-h @tlopex

Aharrypotter:
> ## Code Review  代码审查
> This pull request adds a systematic verification suite for the Relax ONNX importer using the official ONNX Backend Test Suite, including a new pytest marker and a backend adapter. Review feedback identifies a bug where `tvm.runtime.Tensor` is used instead of `tvm.runtime.NDArray` and a logic error in input mapping where initializers should be filtered out from the graph inputs to align with the ONNX test runner's behavior.此拉取请求为 Relax ONNX 导入器添加了一个系统验证套件，使用官方 ONNX 后端测试套件，包括一个新的 pytest 标记和一个后端适配器。审查反馈指出了一个错误，即在输入映射中使用了 `tvm.runtime.Tensor` 而不是 `tvm.runtime.NDArray` ，并且在输入映射中存在逻辑错误，初始值应从图输入中过滤，以与 ONNX 测试运行器的行为保持一致。
> 
> I am having trouble creating individual review comments. Click here to see my feedback.
> 我在创建单独的审查评论时遇到问题。点击这里查看我的反馈。

Thanks for the review, Addressing both points:

**1. `tvm.runtime.Tensor` vs `tvm.runtime.NDArray`**

`tvm.runtime.Tensor` is a valid class in the current TVM Python API — it is defined in `tvm.runtime._tensor.Tensor` and inherits from `tvm_ffi.core.Tensor`. It is not an `AttributeError` risk.

```python
>>> import tvm
>>> hasattr(tvm.runtime.Tensor)
True
>>> tvm.runtime.Tensor.__mro__
(<class "tvm.runtime._tensor.Tensor">, <class "tvm_ffi.core.Tensor">, ...)
```

**2. Excluding initializers from `graph_input_names`**

The ONNX Backend Test data does not use initializers — all inputs (including weights) are provided as `.pb` files. We verified this across multiple models (`test_batchnorm_epsilon`, `test_conv`, `test_matmul_2d`, `test_add`, `test_gemm`): `model.graph.initializer` is always empty.

The current positional mapping logic is correct for this test data format. That said, adding the initializer filter is a good defensive measure for robustness — if the test data format ever changes or a third-party test suite uses initializers, the filter would prevent silent misalignment.

mshr-h:
Let's try to run the new tests in CI and see if they increase CI pressure. @Aharrypotter

Aharrypotter:
@tvm-bot run slow tests

mshr-h:
Just curious can we completely move to new backend tests or do we still need to maintain old ones?
We need to investigate if we can test ops such as Sequence, Attention, Quantization as they seems to be complicated.

tlopex:
I think it's fine to move to new backend tests @mshr-h

Aharrypotter:
> Just curious can we completely move to new backend tests or do we still need to maintain old ones? We need to investigate if we can test ops such as Sequence, Attention, Quantization as they seems to be complicated.

I checked Sequence, Attention, and Quantization locally.

Quantization has a few passing cases, but enabling it cleanly would require very specific per-test include patterns. The broader `QuantizeLinear` / `DequantizeLinear` prefixes also pull in unsupported variants like blocked quantization, float8/float4, and int2/int4/uint2/uint4. So I think it is better to leave it for a follow-up PR.

Attention also needs separate work: the ONNX backend tests use the standard Q/K/V Attention form, while the current Relax converter seems to support the older Microsoft-style packed-QKV path with `num_heads`.

Sequence has similar issues, mostly around runtime sequence inputs, dynamic positions, `SequenceMap`/`Loop`, `ReverseSequence`, and `SplitToSequence`.

Given that, I would keep this PR focused on the initial stable subset and track these categories as follow-up items.

Aharrypotter:
I think we can move in that direction, but probably not completely in one step.

The ONNX Backend Tests are very useful for standard operator semantic coverage, and they can replace some duplicated hand-written semantic tests over time. However, `test_frontend_onnx.py` also contains TVM-specific importer tests that the ONNX Backend Tests do not directly cover, such as `keep_params_in_input`, initializer/runtime parameter handling, input/name sanitization, symbolic/dynamic shape handling, Relax IR structure checks, validation/error paths, and importer-to-legalization/compile integration cases.

I agree that having two ONNX frontend test files could be confusing unless the boundary is clear.

My intended split is that `test_frontend_onnx_backend.py` covers standard ONNX operator semantics with the official backend tests, while `test_frontend_onnx.py` keeps TVM-specific importer behavior and integration/regression tests.

For this PR, I would keep the scope to landing the backend-test runner and the initial stable subset. Then, in a follow-up, we can audit `test_frontend_onnx.py`, migrate/delete duplicated semantic tests, and document where future ONNX frontend tests should go.

Aharrypotter:
please take a look @mshr-h

mshr-h:
Sorry for late review. LGTM.

mshr-h (approved):

PR #19516, opened 2026-05-07
[Relax][TFLite] Add gather frontend expected IRModule tests

weicheng-hsu:
This adds explicit Expected IRModule coverage for TFLite GATHER and GATHER_ND frontend conversion.

GATHER_ND uses Relax gather_nd with int64 indices, so the frontend now casts int32 TFLite indices to int64 before emitting the Relax op. This keeps the generated module well-typed and matches the expected Relax IR.

Testing:
- `python -m pytest tests/python/relax/test_frontend_tflite.py -k "gather"`

related to https://github.com/apache/tvm/issues/18971


tlopex (approved):
LGTM! Thanks

Issue #19436, opened 2026-04-24
[Bug] [Relax][ONNX] Gather with negative indices reads out-of-bounds memory


wuyii8941:
## Expected behavior

ONNX `Gather` supports negative indices: index `-1` refers to the last element along the axis, `-2` to the second-to-last, etc. Per the spec: *"All index values are expected to be within bounds [-s, s-1] along axis of size s."*

## Actual behavior

TVM passes negative indices directly to `relax.op.take` without normalization. This causes **out-of-bounds memory reads**, returning garbage values (often ~1e+30 magnitude). TVM itself warns: *"Fast mode segfaults when there are out-of-bounds indices."*

## Reproduction

```python
import numpy as np
import onnx
from onnx import helper, TensorProto
import onnxruntime as ort
import tvm
from tvm import relax
from tvm.relax.frontend.onnx import from_onnx

data = np.array([[1, 2, 3, 4],
                  [5, 6, 7, 8],
                  [9, 10, 11, 12]], dtype=np.float32)
indices = np.array([0, -1], dtype=np.int64)  # row 0 and last row (row 2)

X = helper.make_tensor_value_info("X", TensorProto.FLOAT, [3, 4])
I = helper.make_tensor_value_info("I", TensorProto.INT64, [2])
Y = helper.make_tensor_value_info("Y", TensorProto.FLOAT, None)
node = helper.make_node("Gather", ["X", "I"], ["Y"], axis=0)
graph = helper.make_graph([node], "test", [X, I], [Y])
model = helper.make_model(graph, opset_imports=[helper.make_opsetid("", 13)])
model = onnx.shape_inference.infer_shapes(model)

# ORT: correct
sess = ort.InferenceSession(model.SerializeToString())
ort_out = sess.run(None, {"X": data, "I": indices})[0]
print("ORT output:")
print(ort_out)
# [[ 1  2  3  4]    ← row 0
#  [ 9 10 11 12]]   ← row 2 (index -1)

# TVM: reads garbage memory
mod = from_onnx(model)
exe = tvm.relax.build(
    tvm.ir.transform.Sequential([relax.transform.LegalizeOps()])(mod), target="llvm"
)
vm = tvm.relax.VirtualMachine(exe, device=tvm.cpu())
tvm_out = vm["main"](
    tvm.runtime.tensor(data, device=tvm.cpu()),
    tvm.runtime.tensor(indices, device=tvm.cpu()),
).numpy()
print("TVM output:")
print(tvm_out)
[...]

PR #19525, opened 2026-05-09
[Relax][ONNX] Normalize negative indices before the take call for `Gather` operator

cchung100m:
Hi Committers,

This PR is trying to fix issues https://github.com/apache/tvm/issues/19436. Any suggestions would be appreciated if you are available.

### Root Cause
1. ONNX `Gather` allows negative indices (counting from the end of the target axis).
2. In the Relax ONNX importer, `Gather` was lowered directly to `relax.op.take` without normalizing negative indices first.
3. This created semantic mismatch / incorrect behavior in downstream lowering paths that assume non-negative indices.
4. Test failures were also caused by pytest parametrization issues:
  - using ONNX `TensorProto` enum values directly as NumPy dtypes,
  - and tuple-style parametrization triggering fixture interpretation errors.
  
### Solutions
1. Added conditional negative-index normalization in `Gather._impl_v13`:
  - apply only for signed index dtypes,
  - use: `idx < 0 ? idx + axis_extent : idx`,
  - derive `axis_extent` from shape/runtime expression to support dynamic shapes.
2. Skipped normalization for unsigned index dtypes to avoid redundant graph ops/checks.

### DIFF
--- a/python/tvm/relax/frontend/onnx/onnx_frontend.py
+++ b/python/tvm/relax/frontend/onnx/onnx_frontend.py
@@ -1106,6 +1106,25 @@ class Gather(OnnxOpConverter):
             shape_val = data[np_index]
             return relax.PrimValue(shape_val)
 
+        data_shape = bb.normalize(relax.op.shape_of(data))
+        data_shape_tensor = bb.normalize(relax.op.shape_to_tensor(data_shape))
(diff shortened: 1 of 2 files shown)


tlopex (approved):

cchung100m:
Thanks to @tlopex 😄

Issue #19519, opened 2026-05-08
[Tracking Issue][TFLite] Remaining builtin operator coverage beyond #19412


tlopex:
The Relax TFLite frontend has made good progress recently:                          

- #18971 tracks unit test coverage for already-supported non-quantized operators.   
- #19412 tracks a first batch of missing or partial operators that are good-first-issue or medium-sized frontend tasks.                                    
                                                                                    
This issue tracks the remaining larger TFLite builtin operator coverage that is not already well-covered by those two issues.                       
                                                                                    
The goal is to make the remaining gaps visible, split them by complexity, and let contributors pick coherent operator families instead of one-off unsupported ops.
                                                                                    
Scope                                                                               

This issue covers remaining TFLite builtin operators that fall into larger semantic groups. This issue does not track quantized model support as a whole; that should be tracked separately.                                                                
                                                                
A. Sequence / recurrent model operators

- [x] `RNN`
- [x] `LSTM`
- [x] `SVDF`
- [x] `UNIDIRECTIONAL_SEQUENCE_RNN`
- [x] `UNIDIRECTIONAL_SEQUENCE_LSTM`
- [x] `BIDIRECTIONAL_SEQUENCE_RNN`
- [x] `BIDIRECTIONAL_SEQUENCE_LSTM`

B. 3D convolution family

- [x] `CONV_3D`
- [x] `CONV_3D_TRANSPOSE`

C. FFT / complex operators

- [x] `RFFT2D`
- [x] `REAL`
- [x] `IMAG`
- [x] `COMPLEX_ABS`

D. Sparse embedding operators

- [x] `EMBEDDING_LOOKUP_SPARSE`
- [x] `HASHTABLE_LOOKUP`

E. Window reduction

- [x] `REDUCE_WINDOW`

F. Control flow / multi-subgraph operators

- [x] `IF`
- [x] `WHILE`
- [x] `CALL`
- [x] `CALL_ONCE`

G. Resource / variable / hashtable operators

- [x] `VAR_HANDLE`
- [x] `READ_VARIABLE`
- [x] `ASSIGN_VARIABLE`
- [x] `HASHTABLE`
- [x] `HASHTABLE_FIND`
- [x] `HASHTABLE_IMPORT`
- [x] `HASHTABLE_SIZE`

H. Operator markers

- [x] `DELEGATE`
- [x] `PLACEHOLDER_FOR_GREATER_OP_CODES`

I. StableHLO operators in TFLite

[...]

PR #19530, opened 2026-05-10
[Relax][Frontend] Add TFLite Frontend Support for CONV_3D_TRANSPOSE

weicheng-hsu:
This commit adds support for the CONV_3D_TRANSPOSE operator in the Relax TFLite frontend.

Key implementations:
- Registered CONV_3D_TRANSPOSE to the TFLite op map.
- Implemented convert_conv3d_transpose which shares Conv3DOptions with regular Conv3D but handles the distinct tensor input layout [output_shape, weight, data, bias] and the DHWOI kernel layout.
- Added calculation for SAME padding that correctly handles transposed convolution semantics, computing padding and output_padding based on dilated kernel and stride sizes.
- Added comprehensive unit tests for valid and same padding in test_frontend_tflite.py.

Testing:
- `python3 -m pytest tests/python/relax/test_frontend_tflite.py -k "test_conv3d_transpose"`

Related to: https://github.com/apache/tvm/issues/19519


tlopex (approved):
LGTM

Issue #19541, opened 2026-05-11
[Bug] Relax ONNX integer Div by zero crashes process with SIGFPE


ALinrunrun:
### Expected behavior

TVM Relax should not crash the Python process when executing an imported ONNX `Div` model with integer inputs.

For integer division by zero, TVM should either handle the case safely or report a controlled runtime error instead of terminating the process with `SIGFPE`.

### Actual behavior

The model compiles successfully, but calling the Relax VM crashes the subprocess with exit code `-8`:

```
stdout: calling vm[main]

exit  : -8
```

This happens when the divisor tensor contains zero values for an `INT32` ONNX `Div` model.

### Environment

TVM: 0.14 environment / Relax ONNX frontend
Python: 3.11
Target: llvm
OS: Linux

### Steps to reproduce

```
import subprocess
import sys
import textwrap


SCRIPT = textwrap.dedent("""\
    import numpy as np
    import onnx
    from onnx import helper, TensorProto
    import tvm
    from tvm import relax
    from tvm.relax.frontend.onnx import from_onnx

    node = helper.make_node("Div", ["a", "b"], ["y"])

    graph = helper.make_graph(
        [node],
        "g",
        [
            helper.make_tensor_value_info("a", TensorProto.INT32, [4]),
            helper.make_tensor_value_info("b", TensorProto.INT32, [4]),
        ],
        [helper.make_tensor_value_info("y", TensorProto.INT32, [4])],
    )

    model = helper.make_model(
        graph,
        opset_imports=[helper.make_opsetid("", 18)],
    )
    model.ir_version = 9

    mod = from_onnx(model, keep_params_in_input=False)

    with tvm.transform.PassContext(opt_level=3):
        ex = tvm.compile(mod, target=tvm.target.Target("llvm"))

    vm = relax.VirtualMachine(ex, tvm.cpu())

    a = np.array([42, 99, -50, 7], dtype=np.int32)
    b = np.array([3, 0, 0, 1], dtype=np.int32)

    print("calling vm[main]", flush=True)

    out = vm["main"](
        tvm.runtime.tensor(a, tvm.cpu()),
        tvm.runtime.tensor(b, tvm.cpu()),
    )

    out = out[0] if isinstance(out, (list, tuple)) else out
    print("returned:", out.numpy().tolist())
""")


proc = subprocess.run(
    [sys.executable, "-c", SCRIPT],
    capture_output=True,
    text=True,
    timeout=60,
)

print("stdout:", proc.stdout)
[...]

PR #19566, opened 2026-05-14
[Relax][ONNX] Prevent `Div` divide-by-zero crashes

cchung100m:
Hi Committers,

This PR is trying to fix issues #19541. Any suggestions would be appreciated if you are available.

### Root cause:
The ONNX `Div` path in the Relax frontend did not separate two integer-divisor cases: constant zero divisors and dynamic/unknown divisors. As a result, constant integer zero divisors were not rejected during import, and dynamic integer divisors could reach runtime without a guard. When the divisor became zero at runtime, execution could trigger SIGFPE and terminate the process instead of raising a controlled error.

### Solution:
This PR applies a minimal, targeted fix in the ONNX frontend `Div` conversion path. It introduces: import-time validation for constant integer divisors containing zero, raising ValueError early.

### DIFF
--- a/python/tvm/relax/frontend/onnx/onnx_frontend.py
+++ b/python/tvm/relax/frontend/onnx/onnx_frontend.py
@@ -555,6 +555,7 @@ class Div(BinaryBase):
 
         return cls.base_impl(bb, inputs, attr, params)
 
+
 class Pow(BinaryBase):
     """Converts an onnx Pow node into an equivalent Relax expression."""
 
--- a/python/tvm/relax/frontend/onnx/onnx_frontend.py
+++ b/python/tvm/relax/frontend/onnx/onnx_frontend.py
@@ -554,23 +554,12 @@ class Div(BinaryBase):
 
         rhs_has_zero = _const_integer_expr_has_zero(inputs[1])
         if rhs_has_zero:
-            return ValueError("ONNX Div with integer inputs encountered divisor value 0.")
+            raise ValueError("ONNX Div with integer inputs encountered divisor value 0.")
 
         if rhs_has_zero is False:
             return cls.base_impl(bb, inputs, attr, params)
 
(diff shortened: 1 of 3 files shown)


mshr-h:
Div by zero is more like a runtime behavior so I think it should be handled in runtime instead of frontend. Or maybe relax/tirx passes is fine.

cc @tlopex

python/tvm/relax/frontend/onnx/onnx_frontend.py
```diff
@@ -72,6 +72,18 @@ def _relax_dtype_is_floating_point(dtype: str) -> bool:
     )
 
 
+def _const_integer_expr_has_zero(expr: relax.Expr) -> bool | None:
```
mshr-h: I don’t think we need to create a new three-line function if it is only called once from another function.

tests/python/relax/test_frontend_onnx.py
```diff
@@ -591,6 +591,51 @@ def test_binary(op_name: str):
...
+    model = helper.make_model(graph, opset_imports=[helper.make_opsetid("", 18)])
+    model.ir_version = 9
+
+    with pytest.raises(
+        ValueError, match="ONNX Div with integer inputs encountered divisor value 0"
+    ):
+        from_onnx(model, opset=18, keep_params_in_input=False)
+
+
+def test_div_integer_dynamic_nonzero_matches_ort():
```
mshr-h: Do we really need to add this test? It looks like it is already covered by test_binary.

cchung100m:
Hi @mshr-h 
Thanks for the suggestions. I updated the part you mentioned. 😄

mshr-h (approved):
LGTM

cchung100m:
Thanks to @mshr-h 😄

PR #19569, opened 2026-05-15
[Web][COS] Persist URL→hash mapping across page loads

tomayac:
The CrossOriginStorage class was storing the URL→hash map only in the module-level GLOBAL_HASH_CACHE. After a page reload that cache is empty, and getFileHash() can only recover hashes for HuggingFace LFS files (URLs containing /resolve/). This left several resource categories uncacheable across sessions:

Screenshot 2026-05-15 at 17 43 15

- JSON files not stored in LFS (mlc-chat-config.json, tokenizer.json, tensor-cache.json) — getFileHash returns null for their /resolve/ URLs because the raw pointer is the actual file content, not an LFS pointer.
- .wasm files from GitHub raw URLs — no /resolve/ pattern at all.
- Any file whose hash was computed from blob content via getBlobHash.

Additionally, even for genuine LFS model shards, each page load was re-fetching every shard's LFS pointer file over the network just to re-derive the SHA-256 hash.

Fix: persist the URL→hash mapping to a dedicated Cache API store (tvmjs-cos-hash-meta). Two write sites:

1. put() — after a file is stored in COS, persist its blob-derived hash. This covers all non-LFS files and non-HuggingFace URLs.

2. resolveHashDescriptor() — after getFileHash() resolves a hash from the LFS pointer, persist it immediately. This eliminates repeated pointer-file network requests for model shards on subsequent visits.

Both write sites use a best-effort try/catch so storage quota errors are silently ignored. loadPersistedHashEntry() similarly swallows errors. The typeof caches === "undefined" guard keeps the code safe in Node.js test environments.


tomayac:
Most likely @CharlieFRuan, @guan404ming, and @akaashrp would be good reviewers for this refinement.

akaashrp:
Thanks for the ping @tomayac. Will take a look soon.

akaashrp (approved):

akaashrp:
@tomayac Apologies for the delay. This looks good to me. Could you rebase so that CI checks pass?

tomayac:
> @tomayac Apologies for the delay. This looks good to me. Could you rebase so that CI checks pass?

Thanks so much! Can you try again now?

guan404ming (approved):
Thanks!

akaashrp:
@tomayac you might need to commit again to retrigger CI, not sure what's up with it

web/src/artifact_cache.ts
```diff
@@ -224,13 +226,55 @@ class CrossOriginStorage {
...
   private async resolveHashDescriptor(
     url: string,
   ): Promise<CrossOriginHashDescriptor | null> {
     const cached = this.hashCache.get(url);
     if (cached) {
       return cached;
     }
+    // Check persistent store before falling back to network-based hash extraction.
+    // This covers non-LFS files (JSON configs, tokenizers) and non-HuggingFace URLs
+    // (e.g. GitHub raw .wasm) whose hashes were computed from blob content on a
```
tomayac:
```suggestion
    // (e.g. GitHub raw .wasm files) whose hashes were computed from blob content on a
```

tomayac:
> @tomayac you might need to commit again to retrigger CI, not sure what's up with it

Made a commit, and CI seems to be running now.

tomayac:
⏳ Oh, please wait before merging. I want to test one more thing before.

tomayac:
> ⏳ Oh, please wait before merging. I want to test one more thing before.

OK, we should be good. All resources that we now additionally cache here use some sort of versioning string in the path of their URLs (e.g, `https://raw.githubusercontent.com/mlc-ai/binary-mlc-llm-libs/main/web-llm-models/v0_2_83/base/Llama-3.2-3B-Instruct-q4f32_1_cs1k-webgpu.wasm`), so we can indeed use the URL as the cache key (`await this.persistHashEntry(url, hash)`) and not be stuck with an outdated version should the resource change. Ready to merge.

PR #19584, opened 2026-05-18
[ARITH] Index TransitiveComparisonAnalyzer knowns by Key

gigiblender:
`CollectDirectComparisons` and `DFSFromLHS` previously linear-scanned the unbounded `knowns_` vector on every `TryCompare` call. When `Simplify` walks many For nodes, this is O(N^2) in the number of For visits and produces an order-of-magnitude slowdown on large IR.

Replace `knowns_` with `knowns_by_key_`, a map keyed by Key. Each Comparison is stored under both its keys (or once when they match). CollectDirectComparisons walks only the smaller of the two query-key buckets; DFSFromLHS walks only the bucket for the current expansion key. scoped_knowns_ stays a flat vector because its scope-exit truncation would require extra bookkeeping to keep in sync.

Add `Comparison::operator==` for index cleanup on Bind override, split AddKnown into AddKnown / AddScopedKnown, and add
`tests/python/arith/test_arith_transitive_comparison.py`.

### DIFF
--- a/src/arith/transitive_comparison_analyzer.cc
+++ b/src/arith/transitive_comparison_analyzer.cc
@@ -175,6 +175,19 @@ class TransitiveComparisonAnalyzer::Impl {
      */
     bool Implies(const Comparison& other) const;
 
+    /*! \brief Structural equality over all four fields.
+     *
+     * Two Comparisons compare equal iff their `lhs_`, `rhs_`,
+     * `offset_`, and `result_` all match.  Used by the Bind-override
+     * cleanup path to locate and remove a Comparison's companion
+     * entry stored under its partner key in `knowns_by_key_`.
+     */
+    friend bool operator==(const Comparison& a, const Comparison& b) {
+      return a.lhs_ == b.lhs_ && a.rhs_ == b.rhs_ && a.offset_ == b.offset_ &&
+             a.result_ == b.result_;
+    }
+    friend bool operator!=(const Comparison& a, const Comparison& b) { return !(a == b); }
+
     // The LHS of the comparison
     Key lhs_;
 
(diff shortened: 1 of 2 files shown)


gigiblender:
I was wondering if someone could give some appreciated feedback -- maybe @tlopex @tqchen @Ubospica

src/arith/transitive_comparison_analyzer.cc
```diff
@@ -566,20 +619,38 @@ void TransitiveComparisonAnalyzer::Impl::Bind(const tirx::Var& var, const Range&
...
+                std::remove(other_it->second.begin(), other_it->second.end(), cmp),
+                other_it->second.end());
+          }
+          knowns_by_key_.erase(idx_it);
+        }
       }
     }
   }
 
   prev_bindings_.Set(var, range);
```
echuraev, line 647: When `it != end` and `differs_from_previous == false` (re-`Bind` of the same var with the same range), we skip the cleanup but still fall through to `prev_bindings_.Set(...)` and `AddKnown(...)`. That re-adds the same comparisons into `knowns_by_key_` under both keys, so duplicates accumulate across repeated binds. That was inflating exactly the buckets this PR is trying to keep small (and doubling the cost vs. the old flat `knowns_`). Suggest gating the `Set` + `AddKnown` block to run only when the binding is new or actually changed. (gemini-code-assist raised this earlier; still looks unaddressed.)

src/arith/transitive_comparison_analyzer.cc
```diff
@@ -566,20 +619,38 @@ void TransitiveComparisonAnalyzer::Impl::Bind(const tirx::Var& var, const Range&
...
+        if (idx_it != knowns_by_key_.end()) {
+          const std::vector<Comparison>& to_remove = idx_it->second;
+          for (const auto& cmp : to_remove) {
+            Key partner_key = (cmp.lhs_ == old_key) ? cmp.rhs_ : cmp.lhs_;
+            // self-comparison (lhs_ == rhs_): only stored once, in
+            // the bucket we are about to erase.
+            if (partner_key == old_key) continue;
+            auto other_it = knowns_by_key_.find(partner_key);
+            if (other_it == knowns_by_key_.end()) continue;
+            other_it->second.erase(
```
echuraev, line 637: After erasing, consider dropping `other_it`'s bucket if it became empty (`if (other_it->second.empty()) knowns_by_key_.erase(other_it);`). Correctness is fine without it, but under repeated `allow_override` rebinds these empty buckets accumulate.

tests/python/arith/test_arith_transitive_comparison.py
```diff
@@ -0,0 +1,178 @@
...
+                for k in T.serial(j + 1, 50):
+                    A[0] = 1
+
+    mod = tvm.IRModule({"main": before})
+    expected = tvm.IRModule({"main": after_dfs})
+
+    with tvm.transform.PassContext(
+        config={"tirx.Simplify": {"transitively_prove_inequalities": True}}
+    ):
+        out_with_dfs = tvm.tirx.transform.Simplify()(mod)
```
echuraev, line 167: This branch is behind than `main`, where `tvm.tirx.transform.Simplify` / config `"tirx.Simplify"` were renamed to `StmtSimplify` / `"tirx.StmtSimplify"`. I think that after rebasing onto current `main` this test will fail unless `Simplify()` → `StmtSimplify()` and `"tirx.Simplify"` → `"tirx.StmtSimplify"`.

echuraev (commented):
@gigiblender Thank you for your PR. Added some comments.

PR #19589, opened 2026-05-19
[TIR] Fixup max/min with input containing NAN

### LINKED ISSUE #19579: [Bug] maximum/minimum/relu/clip do not propagate NaN (IEEE 754 violation)
## Expected behavior

`maximum(NaN, x)` should return `NaN` per IEEE 754-2019 §9.6, consistent with NumPy, PyTorch, JAX, and ONNX Runtime.

`relu(NaN)` should return `NaN` (since relu = max(x, 0)).

## Actual behavior

When NaN is the **first** operand of `T.max` / `T.min`, the result is the second operand instead of NaN. This affects `R.maximum`, `R.minimum`, `R.nn.relu`, and `R.clip`.

The root cause is that `T.max(a, b)` compiles to x86 `maxss`/`maxps` instructions, which have the hardware behavior: "if **src1** is NaN, return **src2**". IEEE 754 requires returning NaN when either operand is NaN.

## Reproducer

```python
import numpy as np
import tvm
from tvm import relax
import tvm.relax.op as R
from tvm.relax.transform import LegalizeOps

bb = relax.BlockBuilder()
a = relax.Var("a", relax.TensorStructInfo((4,), "float32"))
b = relax.Var("b", relax.TensorStructInfo((4,), "float32"))
with bb.function("main", [a, b]):
    with bb.dataflow():
        gv = bb.emit_output(bb.emit(R.maximum(a, b)))
    bb.emit_func_output(gv)
mod = bb.finalize()

pipeline = tvm.ir.transform.Sequential([LegalizeOps()])
exe = tvm.relax.build(pipeline(mod), target="llvm")
vm = tvm.relax.VirtualMachine(exe, device=tvm.cpu())

A = np.array([np.nan, 1.0, np.nan, 0.0], np.float32)
B = np.array([1.0, np.nan, np.nan, np.nan], np.float32)
out = vm["main"](
    tvm.runtime.tensor(A, device=tvm.cpu()),
    tvm.runtime.tensor(B, device=tvm.cpu()),
).numpy()

print(out)       # [1.  nan  nan  nan]  — element 0 is WRONG
print(np.maximum(A, B))  # [nan nan nan nan]  — all NaN per IEEE 754
```

The pattern is operand-order-dependent:
| Expression | TVM | Expected (IEEE 754) |
|---|---|---|
| `max(NaN, 1.0)` | `1.0` | `NaN` |
| `max(1.0, NaN)` | `NaN` | `NaN` |
| `relu(NaN)` = `max(NaN, 0)` | `0.0` | `NaN` |
| `clip(NaN, -1, 1)` | `1.0` | `NaN` |

## Affected operations

```python
R.maximum(a, b)    # when a is NaN
R.minimum(a, b)    # when a is NaN
R.nn.relu(x)       # when x is NaN → returns 0
R.clip(x, lo, hi)  # when x is NaN → returns hi
```

Not affected (correct NaN propagation):
- `R.add`, `R.multiply`, `R.subtract`, `R.divide` — arithmetic propagates NaN correctly
- `R.nn.leakyrelu` — uses comparison path, NaN propagates through multiply
- `R.nn.silu`, `R.nn.gelu` — sigmoid/erf path propagates NaN

## Why this matters

`relu` is the most common activation function. When an upstream computation produces NaN (e.g., from overflow or division by zero), the NaN should propagate to signal the error. Instead, TVM's `relu` silently converts NaN to 0, making the error invisible:

```python
# Suppose upstream overflow produces NaN in one element:
x = [[1.0, 2.0, NaN, 4.0]]
relu(x).sum()
# TVM:   7.0   ← NaN silently disappeared
# NumPy: NaN   ← correctly signals the problem
```

This can cause silent wrong results in production models, where NaN detection is a standard debugging/monitoring signal.

## Root cause

In the lowered TIR, `maximum` becomes `T.max(a, b)`, which LLVM lowers to x86 `maxss`/`maxps`. These instructions follow "if src1 is NaN, return src2" semantics rather than IEEE 754 "return NaN if either is NaN".

The fix would be to emit NaN-aware max/min, e.g.:
```
select(isnan(a) | isnan(b), NaN, max(a, b))
```

## Environment

- TVM commit: 0b0afd8dd (main, 2026-04-24)
- OS: Ubuntu 20.04
- Target: llvm (CPU, x86-64)

actinks:
llvm.minimum and llvm.maximum: Return NaN if one the arguments is NaN.
refs: https://llvm.org/docs/LangRef.html#floating-point-min-max-intrinsics-comparison

fixed #19579

### DIFF
--- a/src/target/llvm/codegen_llvm.cc
+++ b/src/target/llvm/codegen_llvm.cc
@@ -1599,7 +1599,12 @@ llvm::Value* CodeGenLLVM::VisitExpr_(const MinNode* op) {
   llvm::Value* a = MakeValue(op->a);
   llvm::Value* b = MakeValue(op->b);
   if (op->a.dtype().is_float()) {
-    return builder_->CreateMinimum(a, b);
+    llvm::Value* nan_a = builder_->CreateFCmpUNO(a, a);
+    llvm::Value* nan_b = builder_->CreateFCmpUNO(b, b);
+    return builder_->CreateSelect(
+        nan_a, a,
+        builder_->CreateSelect(nan_b, b,
+                               builder_->CreateSelect(CreateLT(op->a.dtype(), a, b), a, b)));
   } else {
     return builder_->CreateSelect(CreateLT(op->a.dtype(), a, b), a, b);
   }
@@ -1609,7 +1614,12 @@ llvm::Value* CodeGenLLVM::VisitExpr_(const MaxNode* op) {
   llvm::Value* a = MakeValue(op->a);
   llvm::Value* b = MakeValue(op->b);
   if (op->a.dtype().is_float()) {
-    return builder_->CreateMaximum(a, b);
+    llvm::Value* nan_a = builder_->CreateFCmpUNO(a, a);
+    llvm::Value* nan_b = builder_->CreateFCmpUNO(b, b);
+    return builder_->CreateSelect(
+        nan_a, a,
+        builder_->CreateSelect(nan_b, b,
+                               builder_->CreateSelect(CreateGT(op->a.dtype(), a, b), a, b)));
   } else {
     return builder_->CreateSelect(CreateGT(op->a.dtype(), a, b), a, b);
   }
--- a/src/target/llvm/codegen_llvm.cc
+++ b/src/target/llvm/codegen_llvm.cc
@@ -1599,22 +1599,20 @@ llvm::Value* CodeGenLLVM::VisitExpr_(const MinNode* op) {
   llvm::Value* a = MakeValue(op->a);
   llvm::Value* b = MakeValue(op->b);
   if (op->a.dtype().is_float()) {
-#if TVM_LLVM_VERSION >= 120
     return builder_->CreateMinimum(a, b);
-#endif
+  } else {
+    return builder_->CreateSelect(CreateLT(op->a.dtype(), a, b), a, b);
   }
-  return builder_->CreateSelect(CreateLT(op->a.dtype(), a, b), a, b);
 }
 
 llvm::Value* CodeGenLLVM::VisitExpr_(const MaxNode* op) {
   llvm::Value* a = MakeValue(op->a);
   llvm::Value* b = MakeValue(op->b);
   if (op->a.dtype().is_float()) {
-#if TVM_LLVM_VERSION >= 120
     return builder_->CreateMaximum(a, b);
-#endif
+  } else {
+    return builder_->CreateSelect(CreateGT(op->a.dtype(), a, b), a, b);
   }
-  return builder_->CreateSelect(CreateGT(op->a.dtype(), a, b), a, b);
 }
 
 llvm::Value* CodeGenLLVM::VisitExpr_(const EQNode* op) {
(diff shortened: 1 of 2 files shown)


src/target/llvm/codegen_llvm.cc
```diff
@@ -1598,12 +1598,22 @@ llvm::Value* CodeGenLLVM::VisitExpr_(const ModNode* op) {
...
+#endif
+  }
   return builder_->CreateSelect(CreateLT(op->a.dtype(), a, b), a, b);
 }
 
 llvm::Value* CodeGenLLVM::VisitExpr_(const MaxNode* op) {
   llvm::Value* a = MakeValue(op->a);
   llvm::Value* b = MakeValue(op->b);
+  if (op->a.dtype().is_float()) {
+#if TVM_LLVM_VERSION >= 120
```
tqchen: we bumped llvm minimum version to 150, so it isnot needed https://github.com/apache/tvm/pull/18853

actinks:
@tqchen The failing test case in ir_module.py is due to the LLVM backend lagging behind LLVM IR in its support for llvm.{min,max}imum.{f32,f64}. How should we handle this situation?
Related issue: https://github.com/llvm/llvm-project/issues/53353

`LLVM ERROR: Cannot select: 0x5d50add04bf0: f32 = fmaximum 0x5d50add04d28, ConstantFP:f32<0.000000e+00>, IRModule.CodeGenLLVM:0`

The fmaximum instruction is supported in llvm 18.1.0 for x86-64 and in llvm 19.1.0 for i686.
proof:
X86-64:  https://godbolt.org/z/b95xvbKxn
i686:  https://godbolt.org/z/q39dr1bs9

cbalint13:
> @tqchen The failing test case in ir_module.py is due to the LLVM backend lagging behind LLVM IR in its support for llvm.{min,max}imum.{f32,f64}. How should we handle this situation? Related issue: [llvm/llvm-project#53353](https://github.com/llvm/llvm-project/issues/53353)
> 
> `LLVM ERROR: Cannot select: 0x5d50add04bf0: f32 = fmaximum 0x5d50add04d28, ConstantFP:f32<0.000000e+00>, IRModule.CodeGenLLVM:0`
> 
> The fmaximum instruction is supported in llvm 18.1.0 for x86-64 and in llvm 19.1.0 for i686. proof: X86-64: https://godbolt.org/z/b95xvbKxn i686: https://godbolt.org/z/q39dr1bs9

Hi @actinks !

- I dont think it is a good idea to enfoce  *everything* via createmin/createmax at IR / LLVM-IR level.
- We also loose backward compatibility with older LLVM (I personally highly appreciate this one in TVM).

---

Instead we could do:
  - Add new TIRx functions ```nanmin```,  ```nanmax``` just like the existing ```nearbyint```  see https://github.com/apache/tvm/blob/main/src/target/llvm/intrin_rule_llvm.cc#L95
  - These new ```nanmin``` and ```nanmax``` can offer TVM / TIRx to be on-par with numpy functions : https://numpy.org/devdocs/reference/generated/numpy.nanmin.html


Now, these new functions can be used to "protect" needed things in the higher level graph flow.
These new TIRx can be available only for certain LLVM version above (protecting with #if LLVM_VERSION_MAJOR).


@tqchen what do you think ?

---

Later EDIt:

> These new TIRx can be available only for certain LLVM version above (protecting with #if LLVM_VERSION_MAJOR).

I think with workaround (emmiting multiple llvm insn) it can be made fully backward to any LLVM version.
But not so sure how to replicate nanmin / nanmax to opencl, cuda, vulkan and all backend beyond llvm.

actinks:
@cbalint13 I submitted a workaround solution for review.

Ref: https://llvm.org/docs/LangRef.html#id321
``uno`: yields true if either operand is a QNaN.`

PR #19591, opened 2026-05-20
[RPC][Tracker] Reject invalid tracker message sizes and consume frame header

### LINKED ISSUE #19585: [Bug][RPC] RPC tracker buffers unbounded data on a single TCP connection
### Component
RPC / Tracker

### Description

`TCPEventHandler.on_message` in `python/tvm/rpc/tracker.py` reads `_msg_size` from the wire as a signed int32 with no upper bound, then appends every subsequent socket read to `self._data` until `len(self._data) >= self._msg_size + 4`. A single TCP connection sending a 4-byte size header of `0x7FFFFFFF` followed by a stream of bytes grows the buffer until the tracker process OOMs.

A second variant: if the wire size decodes to 0, the inner `if self._msg_size == 0` branch peeks at `_data[:4]` but never consumes those bytes; subsequent appends grow `_data` indefinitely.

This is a robustness defect, not a security vulnerability: the TVM security model (https://tvm.apache.org/docs/reference/security.html) states the RPC subsystem is to be deployed only on trusted networks and grants full RCE to any reachable client by design. Filing per Apache security team guidance (private security@ thread, 2026-05-17) as a regular issue rather than an advisory.

### Reproduction

Reproduced on commit 4b93f20 (v0.25.dev0). Server RSS climbs 1:1 with bytes sent over a single connection after the magic handshake.

### Proposed fix

Add `MAX_TRACKER_MSG_BYTES = 1 << 20` constant; reject `_msg_size` outside `(0, MAX_TRACKER_MSG_BYTES]` and close the connection. Consume the 4-byte size header on read so the payload-complete branch operates on payload bytes only.

PR follows.

cchung100m:
Hi Committers,

This PR fixes https://github.com/apache/tvm/issues/19585.

### Root Cause
- `TCPEventHandler.on_message` parsed the 4-byte int 32 length header directly from the accumulated buffer without limits and left the header bytes in the buffer until the full payload arrived.
- If the header decoded to 0 or to an extremely large (e.g., 0x7FFFFFFF) value, self._data could grow without bound (or the header would be repeatedly re-read without being consumed), leading to OOM or denial-of-service.

### Solution
- Introduce `MAX_TRACKER_MSG_BYTES = 1 << 20` (1MiB).
- After detecting at least 4 bytes in the buffer, read and immediately delete the 4-byte header.

### DIFF
--- a/python/tvm/rpc/tracker.py
+++ b/python/tvm/rpc/tracker.py
@@ -237,8 +237,12 @@ class TCPEventHandler(tornado_util.TCPHandler):
             msg = py_str(bytes(self._data[: self._msg_size]))
             del self._data[: self._msg_size]
             self._msg_size = 0
-            self.call_handler(json.loads(msg))
-
+            try:
+                self.call_handler(json.loads(msg))
+            except (json.JSONDecodeError, Exception) as e:
+                logger.warning("Error processing message from %s: %s", self.name(), e)
+                self.close()
+                return
 
     def ret_value(self, data):
         """return value to the output"""
--- a/python/tvm/rpc/tracker.py
+++ b/python/tvm/rpc/tracker.py
@@ -77,6 +77,8 @@ logger.addHandler(console_handler)
 logger.setLevel(logging.INFO)
 logger.propagate = False
 
+MAX_TRACKER_MSG_BYTES = 1 << 20
+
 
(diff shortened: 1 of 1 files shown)


cchung100m:
Close PR because the original issue is tracked by #19586

PR #19593, opened 2026-05-21
[BUILD] Modularize device runtime into per-backend DSOs

tqchen:
## Summary

Modularize `libtvm_runtime` into per-backend shared libraries (`libtvm_runtime_cuda`, `libtvm_runtime_vulkan`, etc.) and `libtvm_runtime_extra` for contrib/disco modules.

- Each backend can be built independently (`USE_CUDA=ON` produces only `libtvm_runtime.so` + `libtvm_runtime_cuda.so`)
- During Python import, available backend DSOs are discovered and loaded automatically, with missing backends silently skipped
- `libtvm_runtime_extra` is always produced if it has sources (disco, contrib modules)
- No user-visible config flag changes — existing `USE_CUDA`, `USE_VULKAN`, etc. continue to work

## Test plan

- [ ] Build with `USE_CUDA=ON` — verify `libtvm_runtime.so`, `libtvm_runtime_cuda.so`, `libtvm_runtime_extra.so` produced
- [ ] `python -c "import tvm; print(tvm.cuda(0).exist)"` prints `True`
- [ ] Runtime GPU tests pass
- [ ] Moving `libtvm_runtime_cuda.so` out of `lib/` — CUDA unavailable, no crash


tlopex (approved):

tqchen:
Reopening from upstream branch to pick up Jenkins groovy changes.

PR #19595, opened 2026-05-23
[Metal] Enable Metal 4 shader compilation

oraluben:
Use the latest SDK language version when available so generated MPP tensor_ops shaders can compile on M5+ while preserving the existing fallback for older SDKs.

### DIFF
--- a/src/backend/metal/runtime/metal_module.mm
+++ b/src/backend/metal/runtime/metal_module.mm
@@ -44,11 +44,25 @@
 #include "../../../support/bytes_io.h"
 #include "metal_common.h"
 
+#if (defined(__MAC_OS_X_VERSION_MAX_ALLOWED) && __MAC_OS_X_VERSION_MAX_ALLOWED >= 260000) || \
+    (defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && __IPHONE_OS_VERSION_MAX_ALLOWED >= 260000)
+#define TVM_METAL_HAS_MSL_4_0 1
+#endif
+
 namespace tvm {
 namespace runtime {
 
 /*! \brief Maximum number of GPU supported in MetalModule. */
 static constexpr const int kMetalMaxNumDevice = 32;
+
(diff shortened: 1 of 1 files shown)


tlopex:
Seems that there is something wrong with CI. Could you retrigger it? @oraluben

tlopex (commented):
Thanks for updating the feature detection to use the SDK version macro.

I’m concerned that selecting `MTLLanguageVersion4_0` solely based on the build SDK can introduce a runtime compatibility regression. A TVM binary built with a macOS 26 SDK may still run on systems/devices where Metal 4 is not supported, but this change would request MSL 4.0 for every Metal runtime compilation, including shaders that would otherwise compile fine with `MTLLanguageVersion2_3`.

Could we avoid making Metal 4 the unconditional language version whenever the build SDK exposes it? For example, either add a runtime/device capability guard, or try Metal 4 only when needed and fall back to `MTLLanguageVersion2_3` if compilation fails.

Also, please clarify what environment was used to validate the new branch, since the current CI does not appear to exercise the `__MAC_OS_X_VERSION_MAX_ALLOWED >= 260000` path.

tlopex (changes requested):

oraluben:
Updated the branch to use MSL 4 only when both the build SDK exposes it and the runtime device reports `MTLGPUFamilyMetal4`; otherwise it keeps the existing `MTLLanguageVersion2_3` path.

Validation was done through TileLang PR https://github.com/tile-ai/tilelang/pull/2252. That PR tests the Metal 4 cooperative-tensor path on supported hardware, and also passed TileLang’s Metal tests on GitHub Actions with macOS 26 + M1. The latter is the compatibility case here: the SDK/runtime can expose Metal 4, but M1 does not support `MTLGPUFamilyMetal4`, so passing those tests confirms the fallback to MSL 2.3 works.

tlopex (approved):
Thanks for doing this!

PR #19599, opened 2026-05-25
[TIRx][Script] Canonicalize buffer-like tile op arguments

tlopex:
This PR improves TIRX buffer-like argument handling for tile ops.
Now TIRX already canonicalizes `BufferLoad` into `BufferRegion` in several C++ builder paths such as `T.reads`, `T.writes`, and `T.match_buffer`. However, the Python tile-op builder path had separate handling: it only converted plain `Buffer` arguments into full-buffer regions, while `BufferLoad` arguments could still be passed through as scalar loads.

This PR makes the tile-op path use the same buffer-like canonicalization model:

- `Buffer` becomes a full `BufferRegion`
- `BufferLoad` becomes a single-point `BufferRegion`
- `BufferRegion` remains unchanged

To avoid duplicating region construction logic in Python, this PR exposes the existing C++ helpers `BufferRegion::FullRegion` and `BufferRegion::FromPoint` to Python, and updates both Python tile-op canonicalization and the existing C++ `BufferRegionFromLoad` helper to use `FromPoint`.

The PR also adds `T.buffer_from_ptr(...)` as a script-level convenience for constructing a buffer view from a pointer or `BufferLoad`. And care is taken to preserve scalar expression behavior for ambiguous overloads. For example, `T.max(A[i], B[j])` remains a scalar expression, while explicit tile-op forms such as `T.max(A[i], B[j], axes=0)` canonicalize the `BufferLoad` arguments into single-point regions.

### DIFF
--- a/python/tvm/tirx/script/__init__.py
+++ b/python/tvm/tirx/script/__init__.py
@@ -62,14 +62,16 @@ def __getattr__(name: str):
             workspace = {}
         if config is None:
             config = kwargs or {}
-        # Convert Buffer args to BufferRegion (covers full extent)
+        # Convert buffer-like tile args to BufferRegion.
         from tvm.tirx import Buffer as _TBuffer
+        from tvm.tirx.expr import BufferLoad as _TBufferLoad
+
+        from .builder.tirx import _to_region
 
         new_args = []
         for a in args:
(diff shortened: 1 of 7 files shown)


yongwww:
@tvm-bot rerun

PR #19606, opened 2026-05-25
[Docs] Reorganize development guide content

tlopex:
This PR removes the old “Development Guides” bucket and folds its remaining pages into more appropriate parts of the documentation.


docs/errors.rst
```diff
@@ -16,56 +16,58 @@
...
+Reporting an Issue
+------------------
+
+Search the `Apache TVM Discuss Forum <https://discuss.tvm.apache.org/>`_
+and the `TVM issue tracker <https://github.com/apache/tvm/issues>`_
+for the exact error message first.  If you do not find an existing
+report, include the following details when starting a new discussion or
+filing an issue:
+
+- The TVM version or git commit hash.
```
mshr-h, line 58: Just a quick comment that we may have to provide a script for collecting environment so users don't have to manually collect them.
I've prototyped it based on pytorch's [`collect_env.py`](https://github.com/pytorch/pytorch/blob/main/torch/utils/collect_env.py).
https://github.com/mshr-h/tvm-collect-env

mshr-h (approved):

docs/errors.rst
tlopex: Thanks, that makes sense. Would update the doc once you finish

PR #19605, opened 2026-05-25
[REFACTOR][TIR] Tie AnnotateDeviceRegions/SplitHostDevice/LowerDeviceKernelLaunch together

tqchen:
## Summary

These three passes are logically a single host/device split step;
having intermediaries between them obscures the model and blocks
folding them into one pass. This PR moves each intermediary to the
position its actual ordering constraint allows, so that
`AnnotateDeviceRegions`, `SplitHostDevice`, and
`LowerDeviceKernelLaunch` run consecutively in every pipeline.

## Rationale

- `MergeSharedMemoryAllocations` moves **before** `AnnotateDeviceRegions`
  (the only legal position: `LowerDeviceKernelLaunch` requires at most
  one dyn-shmem allocation per kernel, so Merge cannot move past Lower).
- `MakePackedAPI` moves **after** `LowerDeviceKernelLaunch` (Lower's
  `kCallingConv = kDeviceKernelLaunch` flag causes `MakePackedAPI` to
  correctly skip device kernels; the host body's lowered
  `tvm_call_packed` is transparent to `MakePackedAPI`'s subroutine
  rewriter).
- `FP8StorageLegalize` / `BF16StorageLegalize` move **after**
  `MakePackedAPI` (their `buffer_map.size()==0` ICHECK requires
  `MakePackedAPI` to have cleared the map).

Prereq for Phase 2: collapsing the three consecutive passes into a
single `tirx.transform.SplitHostDevice` with three commented regions.

## Test plan

- [x] tests/python/tirx-transform/ target-pass unit tests (25 pass)
- [x] tests/python/s_tir/transform/test_merge_dynamic_shared_memory_allocations.py (5 pass)
- [x] tests/python/tirx-transform/test_tir_transform_fp8_legalize.py /
      test_tir_transform_bf16_legalize.py (13 pass)
- [x] tests/python/codegen/test_target_codegen_c_host.py /
      test_target_codegen_device.py (6 pass including
      test_subroutine_call — verifies Risk #2)
- [x] pre-commit run --all-files clean
- [ ] CI: lint / Windows / MacOS

PR #19610, opened 2026-05-26
[REFACTOR][S-TIR] Merge shared memory per thread_extent scope

tqchen:
## Summary

A PrimFunc with multiple sibling `thread_extent` blocks (multiple kernel launches in one function) hits a scoping bug in `MergeSharedMemoryAllocations`: the merged buffer is allocated only inside the first `thread_extent` body, but later `thread_extent`s' accesses are still rewritten to reference it. When `SplitHostDevice` partitions those `thread_extent`s into separate device functions, the second device function references an undefined var — leading to a codegen ICE or runtime crash. This bug is exposed by PR #19605 which moved the pass earlier in the pipeline (before `SplitHostDevice`).

- Convert every per-launch singleton field (`merged_buf_var_`, `merged_alloc_size_`, `shmem_allocs_`, `buffer_byte_offsets_`, etc.) in `SharedMemoryRewriter` into a `KernelScope` struct held on a `scope_stack_`. Push a fresh scope on outermost `thread_extent` entry, run liveness/plan/offset-compute/rewrite/wrap inside that scope, pop on exit.
- Each kernel launch ends up with its own merged buffer allocated inside its own subtree, preserving `LowerDeviceKernelLaunch`'s "at most one dyn-shmem `AllocBuffer` per kernel" invariant.
- Adds `test_multi_thread_extent_blocks` to exercise two sibling `thread_extent` blocks each with independent `shared.dyn` allocations, including an end-to-end check through `AnnotateDeviceRegions` + `SplitHostDevice`.

### DIFF
--- a/src/s_tir/transform/merge_shared_memory_allocations.cc
+++ b/src/s_tir/transform/merge_shared_memory_allocations.cc
@@ -77,24 +77,26 @@ static int64_t ConstantAllocationSize(const ffi::Array<PrimExpr>& extents) {
 }
 
 /*!
- * \brief collect the mapping from the buffer var to its Buffer
+ * \brief collect the mapping from the buffer var to its Buffer within a subtree
  */
 class AllocateCollector : public StmtExprVisitor {
  public:
+  explicit AllocateCollector(bool is_dynamic) : is_dynamic_(is_dynamic) {}
+
   void VisitStmt_(const AllocBufferNode* op) final {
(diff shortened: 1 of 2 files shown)


tqchen:
Folded into #19605 as a prerequisite commit. The MergeSharedMemoryAllocations scope-stack refactor is now `45f14d2bd5` on the `tvm-tie-hostdevice-split` branch, landing before the pipeline reordering. Closing this PR; please review the combined change at #19605.

PR #19616, opened 2026-05-26
[Relax][Frontend][TFLite] Support control-flow multi-subgraph operators

Aharrypotter:
## Summary

This PR adds Relax TFLite frontend support for the TFLite builtin
control-flow / multi-subgraph operator family from #19519 item F:
`CALL`, `IF`, `WHILE`, and `CALL_ONCE`.

It builds on the multi-subgraph import infrastructure merged in PR #19587.
The frontend already accepts TFLite models with extra subgraphs while converting
only `Subgraphs(0)` into the Relax `main` function. This PR uses those extra
subgraphs as callable or control-flow regions for the TFLite control-flow
operators.

The supported subset is intentionally pure tensor and guard-first:

- `CALL` lowers a referenced TFLite subgraph to a private Relax function and
  emits a direct call.
- `IF` lowers the then/else subgraphs to private Relax functions and emits a
  private wrapper function containing Relax `If`.
- `WHILE` lowers the cond/body subgraphs to private Relax functions and emits a
  recursive private Relax function for the loop.
- `CALL_ONCE` supports the empty-init no-op subset and explicitly rejects
  non-empty or resource-like init patterns.

This PR does not model resource variable side effects. Those cases remain
explicitly guarded instead of being imported with incorrect pure functional
semantics.

### Shared Subgraph Lowering

The frontend now keeps shared conversion state across the main graph and
referenced subgraphs:

- `lowered_subgraphs`
- `lowered_if_functions`
- `lowered_while_functions`
- `lowering_stack`
- `module_builder`

Referenced pure tensor subgraphs are lowered through a recursive
`OperatorConverter` using an isolated `ExprTable`, so subgraph tensor bindings
cannot overwrite bindings from the main graph. Lowered subgraphs are cached by
subgraph index and reused when the same region is referenced more than once.
Generated private functions are registered through the shared parent
`module_builder`, so nested cases such as `main CALL -> subgraph A -> CALL
subgraph B` keep all private functions in the final IRModule.

Recursive ordinary `CALL` subgraphs are guarded with `OpNotImplemented`.
`WHILE` uses a dedicated recursive wrapper function instead, because recursion
is part of the intended Relax representation for the loop itself.

### Boundary Validation

The control-flow converters validate subgraph boundaries before lowering:

- referenced subgraph indices must be valid
- op input/output arity must match the referenced subgraph interface
- branch and loop tensor shape/dtype metadata must match the surrounding op
- `IF` and `WHILE` conditions must be scalar bool tensors
- `WHILE` loop-carried input/output tensors must have matching metadata

The shared `_check_subgraph_interface` helper is used by `CALL`, `IF`, and
`WHILE` to keep arity and metadata checks consistent across the control-flow
operators. `_require_scalar_bool_tensor` accepts both frontend `TensorWrapper`
objects and raw TFLite tensors so caller and referenced-subgraph condition
checks use the same path.

These checks keep the first implementation conservative and make unsupported
cases fail with targeted `OpNotImplemented` diagnostics.

### Tuple Outputs

TFLite `CALL`, `IF`, and `WHILE` may produce multiple output tensors. The
frontend maps those cases to Relax tuple returns:

```text
single output  -> tensor expression
multi output   -> Tuple(...)
op outputs     -> TupleGetItem(...)
```

This keeps the single-output IR simple while covering multi-output calls,
multi-output branches, and multi-variable loop state.

## Operator Support

| Operator | TFLite options | Relax lowering | Supported subset |
|---|---|---|---|
| `CALL` | `CallOptions.Subgraph()` | private Relax function call | pure tensor subgraphs, single or multiple outputs |
| `IF` | `IfOptions.ThenSubgraphIndex()`, `ElseSubgraphIndex()` | private wrapper function containing Relax `If` | scalar bool condition, matching branch I/O metadata |
| `WHILE` | `WhileOptions.CondSubgraphIndex()`, `BodySubgraphIndex()` | recursive private Relax function | scalar bool cond output, tensor loop-carried state |
| `CALL_ONCE` | `CallOnceOptions.InitSubgraphIndex()` | no-op for empty init subgraph | empty init subgraph only |

## Not Included

- Full `CALL_ONCE` resource/variable initialization semantics.
- Resource, variant, hashtable, or variable tensor support.
- TensorFlow-generated `tf.cond` / `tf.while_loop` smoke tests.
- Dynamic-shape loop-state refinements beyond the current static metadata
  checks.

## Tests

The tests manually build minimal TFLite flatbuffers and compare the imported
Relax IR with `tvm.ir.assert_structural_equal`. Unsupported-boundary tests use
`pytest.raises`.

| Test | Coverage |
|---|---|
| `test_call_subgraph` | basic `CALL` to a pure tensor subgraph |
| `test_call_subgraph_multi_output` | `CALL` tuple return and output binding |
| `test_call_subgraph_nested_call` | nested `CALL` private function registration |
| `test_call_subgraph_invalid_index_unsupported` | invalid `CALL` subgraph index |
| `test_call_subgraph_io_mismatch_unsupported` | `CALL` arity mismatch |
| `test_call_subgraph_output_metadata_mismatch_unsupported` | `CALL` output metadata guard |
| `test_if_subgraphs` | basic `IF` branch selection |
| `test_if_subgraphs_multi_output` | `IF` tuple branch returns |
| `test_if_subgraphs_non_bool_condition_unsupported` | `IF` condition dtype guard |
| `test_if_subgraphs_invalid_index_unsupported` | invalid then/else subgraph index |
| `test_if_subgraphs_output_count_mismatch_unsupported` | branch output count guard |
| `test_if_subgraphs_input_metadata_mismatch_unsupported` | branch input metadata guard |
| `test_if_subgraphs_output_metadata_mismatch_unsupported` | branch output metadata guard |
| `test_while_subgraphs` | basic recursive `WHILE` lowering |
| `test_while_subgraphs_repeated_cond_body_pair` | shared cond/body loop function cache |
| `test_while_subgraphs_two_loop_vars` | multi-variable loop state tuple path |
| `test_while_subgraphs_non_bool_condition_unsupported` | `WHILE` cond output dtype guard |
| `test_while_subgraphs_invalid_index_unsupported` | invalid cond/body subgraph index |
| `test_while_subgraphs_zero_loop_vars_unsupported` | zero-loop-var guard |
| `test_while_subgraphs_loop_state_metadata_mismatch_unsupported` | loop state metadata guard |
| `test_while_subgraphs_output_count_mismatch_unsupported` | body output count guard |
| `test_while_subgraphs_input_metadata_mismatch_unsupported` | cond/body input metadata guard |
| `test_while_subgraphs_output_metadata_mismatch_unsupported` | cond/body output metadata guard |
| `test_call_once_empty_init_subgraph` | empty `CALL_ONCE` no-op subset |
| `test_call_once_non_empty_init_subgraph_unsupported` | non-empty init subgraph guard |
| `test_call_once_inputs_outputs_unsupported` | `CALL_ONCE` op I/O guard |
| `test_call_once_init_subgraph_io_unsupported` | init subgraph I/O guard |
| `test_call_once_invalid_index_unsupported` | invalid init subgraph index |

Local validation:

```bash
python -m ruff format --check \
  python/tvm/relax/frontend/tflite/tflite_frontend.py \
  tests/python/relax/test_frontend_tflite.py

python -m ruff check \
  python/tvm/relax/frontend/tflite/tflite_frontend.py \
  tests/python/relax/test_frontend_tflite.py

python -m pytest \
  tests/python/relax/test_frontend_tflite.py \
  -k "call_subgraph or if_subgraphs or while_subgraphs or call_once" -q

python -m pytest \
  tests/python/relax/test_frontend_tflite.py -q
```

Result:

```text
ruff format --check: 2 files already formatted
ruff check: All checks passed
28 passed, 434 deselected
462 passed
```

## References

- Issue #19519 item F: TFLite control-flow / multi-subgraph operators
- PR #19587: StableHLO region-based ops and multi-subgraph model support

### DIFF
--- a/python/tvm/relax/frontend/tflite/tflite_frontend.py
+++ b/python/tvm/relax/frontend/tflite/tflite_frontend.py
@@ -174,7 +174,10 @@ class OperatorConverter:
                 "lowered_if_functions": {},
                 "lowered_while_functions": {},
                 "lowering_stack": [],
+                "module_builder": ctx,
             }
+        else:
+            conversion_state.setdefault("module_builder", ctx)
         self.conversion_state = conversion_state
 
         # Add more operators
@@ -574,7 +577,7 @@ class OperatorConverter:
     def get_tensors(self, tensors_idx_list):
         """Get tensor wrapper list from given TFLite tensor index list"""
         return_list = list()
-        for tensor_idx in tensors_idx_list:
+        for tensor_idx in self._indices_or_empty(tensors_idx_list):
             if tensor_idx < 0:
                 return_list.append(TensorWrapper(tensor_idx, 0, 0))
                 continue
@@ -1932,6 +1935,10 @@ class OperatorConverter:
             return exprs[0]
         return relax.Tuple(exprs)
 
+    def _indices_or_empty(self, indices):
+        """Return a TFLite index vector, using an empty list for absent vectors."""
+        return indices if indices is not None else []
+
     def _check_subgraph_io(self, subgraph_index, op_name, input_count=None, output_count=None):
         """Validate a referenced subgraph's input and output counts."""
         subgraph = self._get_subgraph(subgraph_index, op_name)
@@ -1994,7 +2001,9 @@ class OperatorConverter:
         self, subgraph, op_name, tensor_role, subgraph_indices, expected_tensors
     ):
         """Validate referenced subgraph tensor metadata against caller tensors."""
-        for subgraph_index, expected_tensor in zip(subgraph_indices, expected_tensors):
+        for subgraph_index, expected_tensor in zip(
+            self._indices_or_empty(subgraph_indices), expected_tensors
+        ):
             self._check_tensor_metadata_match(
                 subgraph.Tensors(int(subgraph_index)),
                 expected_tensor,
@@ -2014,11 +2023,11 @@ class OperatorConverter:
         """Create Relax parameters for a TFLite subgraph."""
         params = []
         exp_tab = ExprTable()
-        for input_index in subgraph.InputsAsNumpy():
+        for input_index in self._indices_or_empty(subgraph.InputsAsNumpy()):
             tensor = subgraph.Tensors(int(input_index))
             input_name = get_tensor_name(subgraph, int(input_index))
             shape = tuple(tensor.ShapeAsNumpy()) if tensor.ShapeLength() > 0 else []
-            dtype = _decode_type(tensor.Type())
+            dtype = self.get_tensor_type_str(tensor.Type())
             param = relax.Var(input_name, relax.TensorStructInfo(shape=shape, dtype=dtype))
             exp_tab.set_expr(input_name, param)
             params.append(param)
@@ -2071,7 +2080,8 @@ class OperatorConverter:
                 subgraph_bb.emit_func_output(output)
 
             subgraph_mod = subgraph_bb.get()
-            gv = self.bb.add_func(subgraph_mod[function_name_hint], function_name_hint)
+            module_builder = self.conversion_state["module_builder"]
+            gv = module_builder.add_func(subgraph_mod[function_name_hint], function_name_hint)
             lowered_subgraphs[subgraph_index] = gv
             return gv
         finally:
@@ -2121,7 +2131,8 @@ class OperatorConverter:
             )
             if_bb.emit_func_output(result)
         if_func = if_bb.get()[if_name]
-        gv = self.bb.add_func(if_func, if_name)
+        module_builder = self.conversion_state["module_builder"]
+        gv = module_builder.add_func(if_func, if_name)
         lowered_if_functions[cache_key] = gv
         return gv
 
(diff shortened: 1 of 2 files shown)


tlopex (changes requested):
Thanks for the detailed implementation and tests. I think this needs one more correctness check before merge.

The subgraph lowering currently creates a fresh `BlockBuilder` for each referenced subgraph and then only copies `subgraph_mod[function_name_hint]` back into the parent builder. If that subgraph itself contains a `CALL` / `IF` / `WHILE` that lowers another private function, the nested private function can remain only in the inner `BlockBuilder` and be absent from the final module, while the copied function still references its `GlobalVar`.

Could you add a regression test for a nested subgraph case, for example:

`main CALL -> subgraph A -> CALL subgraph B`

and either ensure all generated private functions are copied back to the parent module or adjust lowering so nested subgraph functions are registered in shared state / the parent builder?

Also, please normalize empty TFLite input/output vectors before iterating or zipping them. Some FlatBuffer APIs can return `None` for empty vectors, so `InputsAsNumpy()`, `OutputsAsNumpy()`, and metadata checks should use an explicit `is not None` fallback to `[]` rather than relying on direct iteration.

Aharrypotter:
Thanks @tlopex, addressed in the latest commit by fixing nested subgraph registration and normalizing empty FlatBuffer input/output vectors.

The nested subgraph issue is fixed by registering generated private functions
through the shared parent `module_builder`, instead of leaving nested functions
inside the temporary subgraph `BlockBuilder`. I also added
`test_call_subgraph_nested_call` to cover:

```text
main CALL -> subgraph A -> CALL subgraph B
```

This verifies that both generated private functions are present in the final
IRModule.

I also normalized empty FlatBuffer input/output vectors before iterating or
zipping them, including `InputsAsNumpy()`, `OutputsAsNumpy()`, and metadata
checks, using an explicit `is not None` fallback path. While touching that code,
I switched `_get_subgraph_params` to use `self.get_tensor_type_str(...)` for
consistency with the rest of the class.

tlopex (approved):

Issue #19542, opened 2026-05-11
[Bug] Relax ONNX Cast from float to int8 handles out-of-range and non-finite values differently from ONNX Runtime


ALinrunrun:
### Expected behavior

TVM Relax should execute ONNX `Cast` consistently with ONNX Runtime when casting from `FLOAT` to `INT8`.

For this input:

`x = [300.0, NaN, +Inf, -Inf, 50.0, -50.0]`

ONNX Runtime returns:

`[44, 0, 0, 0, 50, -50]`

### Actual behavior

TVM Relax produces different results:

```
ORT: [44, 0, 0, 0, 50, -50]
TVM: [127, -128, -128, -128, 50, -50]
```

The discrepancy appears when importing an ONNX `Cast` model through the Relax ONNX frontend and compiling it for the `llvm` target.

### Environment

TVM: 0.14 environment / Relax ONNX frontend
ONNX Runtime: 1.23
Python: 3.11
Target: llvm
OS: Linux

### Steps to reproduce

```
import numpy as np
import onnx
import onnxruntime as ort
from onnx import TensorProto, helper
import tvm
from tvm import relax
from tvm.relax.frontend.onnx import from_onnx


node = helper.make_node("Cast", ["x"], ["y"], to=int(TensorProto.INT8))

graph = helper.make_graph(
    [node],
    "g",
    [helper.make_tensor_value_info("x", TensorProto.FLOAT, [6])],
    [helper.make_tensor_value_info("y", TensorProto.INT8, [6])],
)

model = helper.make_model(graph, opset_imports=[helper.make_opsetid("", 18)])
model.ir_version = 9

x = np.array(
    [300.0, np.nan, np.inf, -np.inf, 50.0, -50.0],
    dtype=np.float32,
)

ort_out = ort.InferenceSession(
    model.SerializeToString(),
    providers=["CPUExecutionProvider"],
).run(None, {"x": x})[0]

mod = from_onnx(model, keep_params_in_input=False)

with tvm.transform.PassContext(opt_level=3):
    ex = tvm.compile(mod, target=tvm.target.Target("llvm"))

vm = relax.VirtualMachine(ex, tvm.cpu())

out = vm["main"](tvm.runtime.tensor(x, tvm.cpu()))
tvm_out = (out[0] if isinstance(out, (list, tuple)) else out).numpy()

print("ORT:", ort_out.tolist())
print("TVM:", tvm_out.tolist())
```

### Triage

* needs-triage

PR #19626, opened 2026-05-27
[Relax][ONNX] Fix Cast operator float->int NaN/Inf handling

cchung100m:
Hi Committers,

This PR is trying to fix issues #19542. Any suggestions would be appreciated if you are available.

### Root cause:
FP to INT lowering can be implementation-defined or UB for NaN/Inf and extreme floats, producing backend-dependent results versus ONNX Runtime.

### Solution:
Apply a minimal, deterministic frontend sanitization for float to integer Casts: map NaN and ±Inf to 0.0 before astype. This prevents NaN/Inf from reaching backend fptosi/fptoui lowers and yields stable behavior across targets.

### DIFF
--- a/python/tvm/relax/frontend/onnx/onnx_frontend.py
+++ b/python/tvm/relax/frontend/onnx/onnx_frontend.py
@@ -1133,10 +1133,8 @@ class Cast(OnnxOpConverter):
                 else:
                     return relax.op.astype(x_sanitized, to_type)
 
-                temp_dtype = "int64" if bits > 32 else "int32"
+                temp_dtype = "int64" if bits >= 32 else "int32"
                 mask_val = (1 << bits) - 1
-                if temp_dtype == "int32" and mask_val > 0x7FFFFFFF:
-                    temp_dtype = "int64"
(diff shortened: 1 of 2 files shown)


tlopex (changes requested):
Thanks for working on this. I don’t think this is ready to merge yet.

The new dynamic Cast path now applies to all integer destination dtypes, but `int64`/`uint64` look broken. For `bits == 64`, the code builds constants like `(1 << 64) - 1` and `1 << 63` with `temp_dtype = "int64"`, which are not representable as int64 Relax constants. This means importing a non-constant ONNX `Cast` from float to `INT64`/`UINT64` can fail in the frontend before codegen.

Could you please add coverage for `FLOAT -> INT64` and `FLOAT -> UINT64`, and either handle the 64-bit case separately or avoid the wrapping path for those dtypes?

cchung100m:
Hi @tlopex 

Thank you for the reminder. I have updated the section you mentioned. 😄

tlopex (approved):

cchung100m:
Thanks to @tlopex 😄

PR #19632, opened 2026-05-28
[Relax][Frontend][TFLite] Add RNN converter

LudovicoYIN:
## Summary

Add Relax TFLite frontend support for `RNN` (BuiltinOperator 23), claimed in [#19519](https://github.com/apache/tvm/issues/19519) Group A.

Single-step RNN cell:
```
h = fused_activation(x @ W.T + h @ Wr.T + b)
```

## Changes

- **Handler**: `convert_rnn` registered in `convert_map` (alphabetical, after `RANGE`)
- **Inputs** (5): `input [batch, input_size]`, `input_weights [num_units, input_size]`, `recurrent_weights [num_units, num_units]`, `bias [num_units]`, `hidden_state [batch, num_units]` (variable, zero-initialised)
- **Output**: `[batch, num_units]`
- **Activations**: all fused activations via `convert_fused_activation_function`
- **Quantized**: raises `OpNotImplemented`

## Testing

Two tests added to `tests/python/relax/test_frontend_tflite.py`:

- `test_rnn_none_activation` — `tvm.ir.assert_structural_equal` with identity weights, NONE activation
- `test_rnn_relu_activation` — shape check, random weights, RELU activation

```bash
python -m pytest tests/python/relax/test_frontend_tflite.py -k rnn -v
```

## References

- Issue [#19519](https://github.com/apache/tvm/issues/19519) Group A: Sequence / recurrent model operators


tlopex (changes requested):
Thanks for adding the RNN converter.

I think there is one statefulness issue here: TFLite `RNN` updates the variable `hidden_state` after computing the activated output. In this converter, we return the new hidden state as the op output, but the original `hidden_state_tensor` binding in `exp_tab` is not updated. That means if a later op reads the same hidden-state tensor, it will still see the initial state rather than the state produced by this RNN step.

Could we either update the expression table entry for `hidden_state_tensor.tensor_idx` to the activated result, using override semantics, or explicitly reject stateful RNN graphs if Relax cannot model this mutation yet?

A regression test with two consecutive `RNN` ops sharing the same `hidden_state` would be useful here, since the second op should consume the first op’s updated state.

LudovicoYIN:
Thanks review. I updated the RNN converter to write the activated hidden state back to the original hidden_state tensor entry in exp_tab with force_override=True, so later ops observe the updated state.

I also added a regression test with two consecutive RNN ops sharing the same hidden_state tensor to verify the second RNN consumes the updated state from the first one.

tlopex (approved):
LGTM Thanks

PR #19637, opened 2026-05-28
[Relax][Frontend][TFLite] Add REDUCE_WINDOW support

THINKER-ONLY:
## Summary
Add Relax TFLite frontend support for the builtin `REDUCE_WINDOW` operator.
This covers the ordinary TFLite op only, not `STABLEHLO_REDUCE_WINDOW`.

The converter parses `ReduceWindowOptions` from `BuiltinOptions2`, validates
the static window attributes, and lowers supported reduce functions through
`topi.sliding_window` plus Relax reductions.

Supported modes:
- `ADD`
- `MUL`
- `MINIMUM`
- `MAXIMUM`
- `ALL`
- `ANY`

Empty output shapes are handled directly with `relax.op.zeros`. Quantized
`REDUCE_WINDOW`, dynamic window attributes, and unsupported reduce functions
remain rejected with explicit errors.

## Testing
- `python -m py_compile python/tvm/relax/frontend/tflite/tflite_frontend.py tests/python/relax/test_frontend_tflite.py`
- `python -m pytest tests/python/relax/test_frontend_tflite.py -k reduce_window -q -p no:tvm.testing.plugin`
- `python -m pytest tests/python/relax/test_frontend_tflite.py -k "reduce_window or reduction_ops" -q -p no:tvm.testing.plugin`
- `conda run -n test python -m ruff check python/tvm/relax/frontend/tflite/tflite_frontend.py tests/python/relax/test_frontend_tflite.py`

## Related
Related to #19519.

### DIFF
--- a/python/tvm/relax/frontend/tflite/tflite_frontend.py
+++ b/python/tvm/relax/frontend/tflite/tflite_frontend.py
@@ -281,6 +281,7 @@ class OperatorConverter:
             "REDUCE_MAX": functools.partial(self._convert_reduce, relax_op=_op.max),
             "REDUCE_MIN": functools.partial(self._convert_reduce, relax_op=_op.min),
             "REDUCE_PROD": functools.partial(self._convert_reduce, relax_op=_op.prod),
+            "REDUCE_WINDOW": self.convert_reduce_window,
             "RELU": self.convert_relu,
             "RELU6": self.convert_relu6,
             "RELU_N1_TO_1": self.convert_relu_n1_to_1,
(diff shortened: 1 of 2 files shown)


tlopex (changes requested):
Thanks for adding this converter. I think this currently rejects valid TFLite `REDUCE_WINDOW` models because the init value is required to be rank-0 here.

The TFLite kernel tests build `REDUCE_WINDOW` with the init tensor as a one-element tensor with shape `[1]`, not shape `[]`. For example, the official test helper uses `AddConstInput(kTensorType, {init_value_data_}, {1})`. The op semantics only need a single init value, so accepting only `len(shape) == 0` here is too strict and will reject models that the TFLite runtime accepts.

I think this should check that the init tensor has exactly one element instead of requiring rank 0, and then reshape/squeeze it to a scalar if needed before combining it with the reduced window result.

The current tests miss this because `_build_reduce_window_model` constructs the init tensor with shape `[]`. Please add coverage for the `[1]` init tensor form, ideally with a numeric test matching TFLite behavior.

tlopex (approved):
LGTM

PR #19643, opened 2026-05-30
[Fix] Stabilize layer_norm variance computation with two-pass reduction

### LINKED ISSUE #19592: [Bug] TVM ONNX LayerNormalization lacks numerical stability for large float32 inputs
```
import sys

import numpy as np
import onnx
import tvm
from onnx import TensorProto, helper
from tvm import relax

x = np.array([[80000.0, 80001.0, 80002.0, 80003.0]], dtype=np.float32)
s = np.ones(4, dtype=np.float32)
b = np.zeros(4, dtype=np.float32)

node = helper.make_node("LayerNormalization", ["x", "s", "b"], ["y"], axis=-1, epsilon=1e-5)
graph = helper.make_graph(
    [node],
    "g",
    [
        helper.make_tensor_value_info("x", TensorProto.FLOAT, [1, 4]),
        helper.make_tensor_value_info("s", TensorProto.FLOAT, [4]),
        helper.make_tensor_value_info("b", TensorProto.FLOAT, [4]),
    ],
    [helper.make_tensor_value_info("y", TensorProto.FLOAT, [1, 4])],
)
model = helper.make_model(graph, opset_imports=[helper.make_opsetid("", 17)])
model.ir_version = 9
onnx.checker.check_model(model)

mod = from_onnx(model, keep_params_in_input=False)
    ex = tvm.compile(mod, target=tvm.target.Target("llvm"))
vm = relax.VirtualMachine(ex, tvm.cpu())
y = vm["main"](
    tvm.runtime.tensor(x, tvm.cpu()),
    tvm.runtime.tensor(s, tvm.cpu()),
    tvm.runtime.tensor(b, tvm.cpu()),
).numpy()

x64 = x.astype(np.float64)
expected = ((x64 - x64.mean(axis=-1, keepdims=True)) / np.sqrt(((x64 - x64.mean(axis=-1, keepdims=True)) ** 2).mean(axis=-1, keepdims=True) + 1e-5)).astype(np.float32)
wrong = bool(np.isfinite(expected).all() and not np.isfinite(y).all())

print("TVM:", y.tolist())
print("Expected:", expected.tolist())
print("Wrong:", wrong)
sys.exit(0 if wrong else 1)
```

output:
```
  TVM: [[nan, nan, nan, nan]]
  Expected: [[-1.3416354656219482, -0.4472118020057678, 0.4472118020057678, 1.3416354656219482]]
  Wrong: True

```

ConvolutedDog:
This PR will fix https://github.com/apache/tvm/issues/19592.

LayerNorm could produce NaN on large-value, small-variance inputs due to catastrophic cancellation in var = E[x^2] - E[x]^2.

Switch to a numerically stable two-pass formulation:

  - pass1 computes mean via sum(x) / N
  - pass2 computes variance via sum((x - mean)^2) / N

### DIFF
--- a/include/tvm/topi/nn/layer_norm.h
+++ b/include/tvm/topi/nn/layer_norm.h
@@ -25,6 +25,7 @@
 #define TVM_TOPI_NN_LAYER_NORM_H_
 
 #include <tvm/te/operation.h>
+#include <tvm/topi/reduction.h>
 #include <tvm/topi/tags.h>
 
 #include <string>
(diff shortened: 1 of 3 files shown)


tlopex (changes requested):
Thanks, the updated legalization tests are useful and they do verify that `LegalizeOps` now emits the two-pass TIR structure.

My concern is that this PR fixes a numerical runtime bug, but the tests only check the generated IR structurally. They do not run the #19592 repro input, so they would not directly catch the original NaN failure.

Could you add a small numeric regression test using the issue input, e.g. `[[80000.0, 80001.0, 80002.0, 80003.0]]` with `axis=-1`, and check that the TVM output is finite and matches a stable reference / ONNXRuntime output? That would cover the actual failure mode in addition to the structural TIR change.

tlopex (approved):

PR #19644, opened 2026-05-30
[Relax][IR] Skip in-place multiply when two operands are views of the same tensor

### LINKED ISSUE #19577: [Bug] DataflowUseInplaceCalls corrupts concat input via in-place buffer reuse
## Summary

`relax.transform.DataflowUseInplaceCalls()` produces incorrect results when a value is used both as an input to a binary op (multiply) and as a later input to concat. The pass incorrectly allows the multiply to write in-place to the shared buffer, corrupting the value before concat reads it.

## Minimal Reproducer

```python
import numpy as np
import tvm
from tvm import relax
import tvm.relax.op as R
from tvm.relax.transform import LegalizeOps

bb = relax.BlockBuilder()
x = relax.Var("x", relax.TensorStructInfo((4,), "float32"))
with bb.function("main", [x]):
    with bb.dataflow():
        a = bb.emit(R.expand_dims(x, axis=1))      # (4,1)
        b = bb.emit(R.expand_dims(x, axis=1))      # (4,1)
        prod = bb.emit(R.multiply(a, b))             # (4,1) = x^2
        out = bb.emit(R.concat([prod, b], axis=1))   # (4,2), expected [x^2, x]
        gv = bb.emit_output(out)
    bb.emit_func_output(gv)
mod = bb.finalize()

x_np = np.array([1.0, 2.0, 3.0, 4.0], dtype=np.float32)

# Correct result (without pass):
# [[ 1,  1], [ 4,  2], [ 9,  3], [16,  4]]

# Incorrect result (with pass):
# [[ 1,  1], [ 4,  4], [ 9,  9], [16, 16]]
# Column 1 should be x=[1,2,3,4] but gets x^2=[1,4,9,16]
```

## Expected Behavior

`concat([prod, b], axis=1)` should produce `[[x^2, x]]` for each row. Column 0 = x^2, Column 1 = x.

## Actual Behavior

After `DataflowUseInplaceCalls`, column 1 also contains x^2. The multiply operation writes in-place to b's buffer, so when concat reads b, it sees the overwritten x^2 values instead of the original x values.

## Root Cause

The liveness analysis in `DataflowUseInplaceCalls` fails to detect that `b` is live at the point where `multiply(a, b)` executes, because `b` is also needed later by `concat`. The pass incorrectly determines that `b`'s buffer can be reused for the multiply output.

The pattern that triggers this is:
1. Two values `a` and `b` derived from the same input (both are `expand_dims(x)`)
2. `prod = multiply(a, b)` — the pass allows in-place write to `b`'s storage
3. `concat([prod, b])` — reads `b` after it was overwritten

## Impact

This is a **silent correctness bug** — no crash, no warning, just wrong numerical results. It affects any model where:
- A value is used as input to both a binary op and a later concat/use
- `DataflowUseInplaceCalls` is in the optimization pipeline (it's included in standard pipelines)

## Environment

- TVM version: 0.24.dev0
- Target: llvm (CPU)

ConvolutedDog:
This PR will fix https://github.com/apache/tvm/issues/19577.

In this issue, the IRModule before applying any pass looks like:

```
  %x: Tensor[(4,), float32]  // function param
  with R.dataflow():
    %lv  = expand_dims(%x, axis=1)     // (4, 1)
    %lv1 = expand_dims(%x, axis=1)     // (4, 1) second call, new Var
    %lv2 = multiply(%lv, %lv1)         // (4, 1)
    %lv3 = concat(%lv2, %lv1, axis=1)  // (4, 2)
    ...
```

When the users manually apply the `DataflowUseInplaceCalls` pass, the pass will rewrite the statement `%lv2 = multiply(%lv, %lv1)` to be like `%lv = multiply(%lv, %lv1); %lv3 = concat(%lv, %lv1, axis=1)`, which reuses the %lv buffer to avoid storage waste.

But this rewrite will chang the buffer context of %lv, and also in LLVM generated code, %lv1 shared the same storage with %lv, so when executing `%lv = concat(%lv, %lv1, axis=1)`, the %lv1 context has also been changed to `multiply(%lv, %lv1)`. So the failure is due to the shared storage of different views of the same tensor %x.

During the execution, %lv1 holds `x^2` instead of `x` after `multiply`. `concat` reads %lv1 for the right column and its result is [[1,1],[4,4],[9,9],[16,16]] instead of [[1,1],[4,2],[9,3],[16,4]] (the correct result should be : left col `x^2`, right col should stay `x`).

Change: View-like ops (expand_dims, squeeze, reshape, permute_dims, memory.view, ensure_zero_offset) take the input's alias set in alias analysis instead of a new id: %lv and %lv1 share alias with %x. Then the pass rejects in-place of `multiply(%lv, %lv1)`: %lv and %lv1 are different vars but alias ids intersect, so no operand may be reused in-place.

### DIFF
--- a/include/tvm/runtime/tensor.h
+++ b/include/tvm/runtime/tensor.h
@@ -183,6 +183,26 @@ class Tensor : public tvm::ffi::Tensor {
    */
   TVM_RUNTIME_DLL static void CopyFromBytes(const DLTensor* to, void* from, size_t nbytes,
                                             TVMStreamHandle stream = nullptr);
+
+  /*!
+   * \brief Check if two tensors share the same underlying storage.
+   *
+   * This detects runtime storage aliasing (e.g. views from CreateView, etc.) but does
+   * not imply either tensor was created by CreateView.
+   *
+   * \param a The first tensor.
+   * \param b The second tensor.
+   * \return True if the tensors share the same storage.
+   */
+  TVM_RUNTIME_DLL static bool IsStorageShared(const DLTensor* a, const DLTensor* b);
+
+  /*!
+   * \brief Tensor overload of IsStorageShared.
+   * \param a The first tensor.
+   * \param b The second tensor.
+   * \return True if the tensors share the same storage.
+   */
+  static bool IsStorageShared(const Tensor& a, const Tensor& b);
 };
 
 /*!
--- a/src/relax/transform/dataflow_inplace.cc
+++ b/src/relax/transform/dataflow_inplace.cc
@@ -62,9 +62,9 @@ std::unordered_set<int> GetVarAliasSetFromExpr(
 }
 
 // In-place on arg `candidate` is invalid if another distinct operand may alias the same
-// storage (e.g. two expand_dims views of x bound to different vars). We only reject when
-// alias analysis gives a definite overlap; unknown (-1) other operands are handled elsewhere.
-// Repeated uses of the same var (e.g. add(z, z)) are allowed.
+// storage (e.g. two expand_dims views of x bound to different vars). Reject on any shared
+// alias id; -1 in the other operand's set does not skip checking other ids. Same var twice
+// (e.g. add(z, z)) is allowed.
 bool InplaceArgDisjointFromOtherCallArgs(
     const CallNode* call_node, int candidate,
     const std::unordered_map<Var, std::unordered_set<int>>& alias_sets) {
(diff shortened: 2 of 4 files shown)


tlopex (changes requested):
Thanks for the fix. The alias-set approach looks like the right direction, but I think the view-op list is still incomplete.

`relax.flatten` and likely `relax.nn.batch_flatten` are also reshape-like/view-like ops: they legalize to reshape-style TIR and can later become runtime reshape/CreateView instead of a real copy. The same corruption pattern can still happen with two `flatten(x)` results used as operands to an in-place binary op and then one view reused later.

Could you include these reshape-like ops in `IsViewMemoryOp` and add regression coverage for them? I would also avoid using `tensor.numpy().__array_interface__["data"]` as a runtime storage check, since `Tensor.numpy()` copies into a NumPy buffer rather than exposing the TVM tensor storage.

tlopex (approved):

PR #19645, opened 2026-05-30
[Relax][PyTorch] Cast non-bool inputs to bool in logical_not converter

javierdejesusda:
### Motivation

`torch.logical_not` accepts an input tensor of any dtype (treating any nonzero
element as `True`) and always returns a `bool` tensor.

The PyTorch frontend previously lowered it with `self._unary_op(relax.op.logical_not)`.
`relax.op.logical_not` is a unary arithmetic op that passes its input dtype through,
so a non-bool input (for example `float32`) produced a `float32` result instead of
the `bool` result PyTorch returns. This is a dtype mismatch against the reference
PyTorch semantics for both the FX and ExportedProgram frontends.

### Changes

- Add a shared `_logical_not` converter in `BaseFXGraphImporter` that casts non-bool
  inputs to `bool` before applying `relax.op.logical_not`. Bool inputs are passed
  through unchanged (no redundant cast).
- Point the `logical_not` (FX) and `logical_not.default` (ExportedProgram)
  registrations at the new converter.
- Update the FX test and add a standalone ExportedProgram `test_logical_not` to assert
  the corrected IR (`astype` to bool, then `logical_not`, producing a `bool` output).

### Notes

The cast to `bool` lowers to an elementwise nonzero test, so it matches PyTorch's
"nonzero is True" semantics for float, integer, and NaN inputs.

### DIFF
--- a/python/tvm/relax/frontend/torch/base_fx_graph_translator.py
+++ b/python/tvm/relax/frontend/torch/base_fx_graph_translator.py
@@ -389,6 +389,14 @@ class BaseFXGraphImporter(metaclass=abc.ABCMeta):
         dim = node.args[1] if len(node.args) > 1 else node.kwargs.get("dim", -1)
         return self.block_builder.emit(relax.op.nn.log_softmax(x, dim))
 
+    def _logical_not(self, node: fx.Node) -> relax.Var:
+        x = self.env[node.args[0]]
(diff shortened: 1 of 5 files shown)


tlopex (approved):
LGTM Thanks for the contribution!

Issue #19576, opened 2026-05-17
[Bug] AdjustMatmulOrder crashes on batched matmul chains (3D @ 2D)


wuyii8941:
## Description

The `AdjustMatmulOrder` pass crashes with `Check failed: shape_c.size() == 2 (3 vs. 2)` when encountering a matmul chain where the intermediate result is 3D and the final weight is 2D.

This is a common pattern in transformer models: `matmul(attn_output[B,S,D], W_o[D,D])` where the attention output is 3D (batched) and the output projection weight is 2D.

## Reproducer

```python
import numpy as np
import tvm
from tvm import relax
import tvm.relax.op as R

B, S, D = 2, 16, 64
bb = relax.BlockBuilder()
x = relax.Var('x', relax.TensorStructInfo((B, S, D), 'float32'))
wq = relax.Var('wq', relax.TensorStructInfo((D, D), 'float32'))
wk = relax.Var('wk', relax.TensorStructInfo((D, D), 'float32'))
wv = relax.Var('wv', relax.TensorStructInfo((D, D), 'float32'))
wo = relax.Var('wo', relax.TensorStructInfo((D, D), 'float32'))
with bb.function('main', [x, wq, wk, wv, wo]):
    with bb.dataflow():
        q = bb.emit(R.matmul(x, wq))
        k = bb.emit(R.matmul(x, wk))
        v = bb.emit(R.matmul(x, wv))
        kt = bb.emit(R.permute_dims(k, [0, 2, 1]))
        scores = bb.emit(R.matmul(q, kt))
        scale = relax.const(1.0 / np.sqrt(D), 'float32')
        scores = bb.emit(R.multiply(scores, scale))
        attn = bb.emit(R.nn.softmax(scores, axis=-1))
        out = bb.emit(R.matmul(attn, v))         # 3D result
        proj = bb.emit_output(R.matmul(out, wo))  # 3D @ 2D → crash
    bb.emit_func_output(proj)
mod = bb.finalize()

# This crashes:
pipeline = tvm.ir.transform.Sequential([
    relax.transform.AdjustMatmulOrder(),
    relax.transform.LegalizeOps()
])
mod_l = pipeline(mod)  # Check failed: shape_c.size() == 2 (3 vs. 2)
```

## Error

```
tvm.error.InternalError: Check failed: shape_c.size() == 2 (3 vs. 2) :
```

## Root cause

[...]

PR #19650, opened 2026-05-31
[Fix][Relax] Support ND batched matmul chains in AdjustMatmulOrder pass

ConvolutedDog:
Fix a crash (https://github.com/apache/tvm/issues/19576) when AdjustMatmulOrder encounters mixed-dimension matmul chains common in transformer models (e.g. matmul(attn_output[B,S,D], W_o[D,D])). The pass previously assumed all operands in a chained rewrite were 2D and asserted shape_c.size() == 2, failing on 3D intermediate results.

Changes:
- Replace full 2D transpose with permute_last_two_dims for permuted matmul patterns, swapping only the last two axes for ND tensors.
- Remove hard ndim==2 checks in the permuted rewrite path.
- Account for batch prefixes when comparing naive matmul FLOPs, so reorder decisions reflect batched vs. weight-only inner matmuls.
- Skip reorder when neither evaluation order is provably cheaper.
- Add regression tests for symbolic/concrete batched LoRA shapes.
- Add a numerics test covering a minimal attention block with ND permute_dims.

### DIFF
--- a/src/relax/op/op_common.cc
+++ b/src/relax/op/op_common.cc
@@ -22,6 +22,7 @@
 #include <tvm/ffi/cast.h>
 
 #include <algorithm>
+#include <sstream>
 
 namespace tvm {
 namespace relax {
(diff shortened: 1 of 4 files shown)


tlopex (changes requested):
Thanks for fixing the ND crash path. I think the direction is right, but I’d request changes for the new batched FLOP model.

The current estimate multiplies `batch_A * batch_B * batch_C`, but `relax.matmul` broadcasts batch prefixes; it does not evaluate all pairwise combinations of equal batch axes. This can flip the chosen order for normal batched chains. For example, with `A: [2, 1, 1]`, `B: [2, 1, 2]`, and `C: [2, 2, 3]`, the actual naive costs are:

- LHS first: `2*1*1*2 + 2*1*2*3 = 16`
- RHS first: `2*1*2*3 + 2*1*1*3 = 18`

so LHS first is cheaper. The current formula estimates `56` vs `48`, and would choose RHS first instead.

I think the cost model should use the product of the broadcasted batch prefix for each matmul, e.g. `batch_AB`, `batch_BC`, and the common broadcast batch for the outer matmul, rather than multiplying the independent prefix products. It would also be good to add structural tests where all three operands share a nontrivial batch prefix, plus a broadcast case like `[B, ...]` with `[1, ...]`.

Separately, I agree with the existing comment that the permuted path should guard rank-1 operands before calling `permute_last_two_dims` / `transpose_shape_last_two_dims`, since the later 1D padding does not run before that code.

tlopex (approved):

PR #19652, opened 2026-06-01
[Relax][Frontend][TFLite] Add EMBEDDING_LOOKUP_SPARSE converter

LudovicoYIN:
## Summary

Add Relax TFLite frontend support for `EMBEDDING_LOOKUP_SPARSE`.

This PR adds a converter for `EMBEDDING_LOOKUP_SPARSE` in the Relax
TFLite frontend. The implementation supports the `SUM`, `MEAN`, and
`SQRTN` combiners and handles higher-rank sparse indices. The sparse
aggregation is lowered through `scatter_nd` to match TFLite operator
semantics for the supported cases.

The PR also adds handcrafted TFLite frontend tests covering:
- `SUM`
- `MEAN`
- `SQRTN`
- a 3D indices case

## Testing

Ran `tests/python/relax/test_frontend_tflite.py -k 'embedding_lookup_sparse'`.

Part of #19519


tlopex (changes requested):
Thanks for adding this. The overall scatter-based lowering matches the supported sorted sparse-index cases, but I think the normalization path changes TFLite semantics and should be fixed before landing.

For `MEAN`, TFLite divides the weighted sum by the total weight as-is. It does not require the total weight to be positive, and it does not clamp the denominator. The current lowering does:

`safe_denominator = maximum(denominator, 1e-12)`
and then returns zero when `denominator <= 0`.

That gives incorrect results for valid inputs with negative weights, or with weights whose sum is zero. For example, if two entries in the same bucket have weights `[1.0, -2.0]`, TFLite divides by `-1.0`, while this lowering drops the bucket to zero.

I think the empty-bucket mask should be based on the number of elements in each bucket, not on `denominator > 0`. One way to model this is to scatter-add a separate count tensor, divide by the real denominator without clamping, and only use the count mask to keep genuinely empty buckets at zero.

It would also be good to add a regression test with negative weights for `MEAN` so this semantic difference is covered.

tlopex (approved):

PR #19654, opened 2026-06-01
[Relax][Frontend][TFLite] Add HASHTABLE_LOOKUP converter

LudovicoYIN:
## Summary

Add Relax TFLite frontend support for `HASHTABLE_LOOKUP`.

This PR adds a converter for `HASHTABLE_LOOKUP` in the Relax TFLite frontend. The implementation supports non-string value tensors and lowers the lookup through `bucketize`, `take`, and `where` so that missing keys return zero-filled values together with a `uint8` hits mask matching TFLite semantics for the supported cases.

The PR also adds handcrafted TFLite frontend tests covering:
- 1D float value tensors
- 2D float value tensors
- the current unsupported string-value case

## Testing

Ran `tests/python/relax/test_frontend_tflite.py -k 'hashtable_lookup'`.

Part of #19519


tlopex (approved):
LGTM

PR #19664, opened 2026-06-03
[CI] Reformat the macOS repair-wheel-command as a multiline script

tlopex:
The single-line 'bash -c' form is hard to read and edit; rewrite it as a readable multiline command. Functionally equivalent to the previous inline version: delocate the wheel, then ad-hoc re-sign every bundled Mach-O and repack.

The re-sign step is required because delocate's edits invalidate the ad-hoc signature of tvm/lib/libtvm_runtime.dylib (it LC_LOADs the excluded `rpath/libtvm_ffi.dylib`, which ships in the separate apache-tvm-ffi package), and arm64 dyld SIGKILLs an invalidly-signed Mach-O on import. 'wheel' is installed explicitly since it is absent from cibuildwheel's repair venv, and the delocated wheel is located by glob since delocate may retag it.


tlopex:
@tvm-bot rerun

tqchen (approved):

PR #19667, opened 2026-06-03
[ARITH] Add optional Z3-backed proving to Analyzer

Ubospica:
## Summary

This PR adds an optional Z3 SMT solver backend to `tvm::arith::Analyzer` for stronger integer arithmetic proving.

The integration is disabled by default and guarded by a new `USE_Z3` CMake option. When enabled, `Analyzer::CanProve` keeps the existing TVM arithmetic analysis path first, then falls back to Z3 only when the existing analyzers cannot prove the predicate. When disabled, a stub implementation is built so the C++ and Python APIs remain available without requiring Z3. When `USE_Z3=ON`, the CMake lookup can consume the PyPI `z3-static` package and statically link Z3, so `libtvm` can avoid a runtime `libz3` dependency. Enabling this by default for wheels is intentionally deferred until the `z3-static` wheels are published.

## Features

This PR adds:

- Optional Z3 build support through `USE_Z3`, defaulting to `OFF`.
- A new `arith::Z3Prover` sub-analyzer owned by `arith::Analyzer`.
- Updated `Analyzer::CanProve` flow: it first runs the existing TVM logic (simplify, const-int-bound, int-set, transitive comparison) unchanged; only if that fails, and only at `kSymbolicBound` strength, does it ask `z3_prover.CanProve(simplified)`, which proves the predicate by checking that its negation is unsatisfiable under the current constraints. Z3 is a pure fallback and a negative/unknown result keeps the original `false`.
- SMT-LIB2 export for debugging and external solver reproduction.
- Python APIs:
  - `Analyzer.get_smtlib2(expr=None)`
  - `Analyzer.set_z3_timeout_ms(timeout_ms)`
  - `Analyzer.set_z3_rlimit(rlimit)`
  - `Analyzer.get_z3_stats()`
- C++ APIs:
  - `Z3Prover::CanProve`
  - `Z3Prover::Bind`
  - `Z3Prover::EnterConstraint`
  - `Z3Prover::GetSMTLIB2`
  - `Z3Prover::GetStats`
  - `Z3Prover::SetTimeoutMs`
  - `Z3Prover::SetRLimit`
  - `Z3Prover::GetModel`
  - `Z3Prover::CountSatisfyingValues`
- Scalar integer, unsigned integer, and boolean expression translation to Z3.
- Support for arithmetic, comparisons, boolean operators, `min`, `max`, `select`, `if_then_else`, `let`, casts, truncated division/modulo, floor division/modulo, and selected bitwise/shift operations.
- Deterministic solver control using Z3 `rlimit`, with `random_seed` fixed to `42`.
- Thread-local Z3 context sharing to reduce initialization overhead while keeping thread safety.
- A disabled-mode stub implementation that returns conservative results when Z3 is not built.

## Implementation notes

- The on/off implementations live in a single `src/arith/z3_prover.cc`, selected by the `TVM_USE_Z3` macro that `cmake/modules/contrib/Z3.cmake` defines when `USE_Z3=ON`. This keeps the optional backend in one file under `src/arith` and lets CMake control whether the real solver is compiled.
- The Z3 fallback is exception-safe (`z3::exception` is caught and treated as "cannot prove") and gated behind `kSymbolicBound`, so the common `kDefault` path (including deeply recursive internal `CanProve` calls) never pays the prover cost.
- TVM `Div`/`Mod` are truncated (round toward zero); they are translated with explicit truncating helpers instead of Z3's Euclidean `operator/` / `operator%` to stay sound for negative dividends.
- Shift handling relies on Z3's native bit-vector semantics and does not add hard assertions to the shared solver, which would otherwise permanently poison it.

## TileLang References

The implementation is based on the Z3 analyzer integration used in TileLang's TVM fork, with the upstream port kept optional and scoped to TVM's arithmetic analyzer.

- [tile-ai/tilelang#1367](https://github.com/tile-ai/tilelang/pull/1367)
- [tile-ai/tilelang#1458](https://github.com/tile-ai/tilelang/pull/1458)
- [tile-ai/tilelang#2216](https://github.com/tile-ai/tilelang/pull/2216)
- [TileLang/tvm#22](https://github.com/TileLang/tvm/pull/22)
- [TileLang/tvm#24](https://github.com/TileLang/tvm/pull/24)

## Acknowledgements

The original Z3 integration in TileLang's TVM fork was authored by @kurisu6912 and @SiriusNEO  in [tile-ai/tvm@e633295](https://github.com/tile-ai/tvm/commit/e633295de994a89668d7a9930dbbd455af3efc66). This PR ports that work to upstream TVM as an optional backend. Thanks to @kurisu6912 and the TileLang authors for the original design and implementation.

## Differences From TileLang

The core `Z3Prover` algorithm is kept close to the latest TileLang TVM implementation. The differences are in how it is integrated into upstream TVM:

1. **Optional build, single file.** We make Z3 optional, defaulting `USE_Z3=OFF`. Both the real implementation and the conservative stub live in one `src/arith/z3_prover.cc`, switched by the `TVM_USE_Z3` macro, so Z3 stays out of the default build. In TileLang it is enabled by default.

2. **Z3 discovery.** We add `cmake/modules/contrib/Z3.cmake`, which, when `USE_Z3=ON`, first tries the PIC static `libz3` layout provided by the PyPI `z3-static` package (headers + static library + CMake package files). A custom (e.g. shared system) Z3 can still be selected via `Z3_DIR`/`CMAKE_PREFIX_PATH`. TileLang resolves a shared Z3 from `z3-solver` directly via `cmake/pypi-z3/FindZ3.cmake`.

3. **Packaging.** This PR does not add a mandatory `z3-static` build-system requirement or enable `USE_Z3` by default for all wheels yet, because the static Z3 wheels must be published first. Once that package is available, wheel builds can enable `USE_Z3=ON` and link Z3 statically without install RPATH handling, build-tree `libz3` copying, or a runtime `z3-solver` Python dependency. TileLang carries shared-library wheel packaging for all of this.

4. **Strength gating.** The Z3 fallback is gated behind `kSymbolicBound` so it is only consulted from top-level predicate proving, avoiding prover cost on cheap recursive paths.

5. **Soundness fixes.** Truncated `Div`/`Mod` use truncating helpers rather than Z3's Euclidean operators, shift handling no longer poisons the shared solver, and `CanProve` is exception-safe.

6. **Python surface.** We expose only the debug/config APIs (`get_smtlib2`, `set_z3_timeout_ms`, `set_z3_rlimit`, `get_z3_stats`) to Python, keeping `GetModel` and `CountSatisfyingValues` as C++-only. TileLang exposes the same debug/config set, and its `set_z3_rlimit` is named `set_z3_max_step`.

7. **No clone API.** We do not port TileLang's `Analyzer::Clone` / Python `Analyzer.clone()`; `Z3Prover::CopyFrom` is retained so a future clone can reuse it.


src/target/z3/z3_prover_off.cc
tlopex, line 1: Maybe it's better to remove `z3_prover_off.cc` and `z3_prover_on.cc` to /src/arith?
And we should have ASF license header on those two cc files

tlopex (changes requested):
There is no negative test currently, could you add one?
And seems the new tests don't actually run in CI. USE_Z3 defaults to OFF and no buildconfig enables it , so every job builds the Z3-off stub and _require_z3() skips all four real tests. Do we need to enable it during CI? cc @tqchen

src/target/z3/z3_prover_on.cc
```diff
@@ -0,0 +1,788 @@
...
+      z3::expr a_expr = VisitInt(a);
+      z3::expr b_expr = VisitInt(b);
+
+      // Add constraint that shift amount should be non-negative
+      // This is a common assumption in many programming languages
+      solver.add(b_expr >= 0);
+
+      // Also limit shift amount to avoid unrealistic large shifts
+      // We'll limit to 64 bits (reasonable for most use cases)
+      solver.add(b_expr < 64);
```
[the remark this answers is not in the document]
tlopex: This makes sense. Could you have a correction?

src/arith/z3_prover.cc
```diff
@@ -0,0 +1,788 @@
...
+  static z3::expr floormod(const z3::expr& a, const z3::expr& b) {
+    return z3::ite(b > 0, a % b, -((-a) % b));
+  }
+  z3::expr VisitExpr_(const AddNode* op) override {
+    return VisitArith(z3::operator+, op, op->a, op->b);
+  }
+  z3::expr VisitExpr_(const SubNode* op) override {
+    return VisitArith(z3::operator-, op, op->a, op->b);
+  }
+  z3::expr VisitExpr_(const MulNode* op) override {
```
tlopex, line 646: VisitExpr_(DivNode) and VisitExpr_(ModNode) map TVM's truncated div/mod (round toward zero) onto Z3's native operator/ / operator%, which are Euclidean (floor / non-negative remainder for a positive divisor). These disagree once the dividend can be negative. 
CanProve can return true for a predicate that is false under TVM semantics (its negation comes out unsat under Euclidean div) which is unsound. Just like FloorDiv/FloorMod via the custom floordiv/floormod helpers. Div/Mod need the same treatment with truncating helpers rather than the raw Z3 operators I think.
I saw the same place on Tilelang, seems it's the same but still think it's an issue

src/arith/z3_prover.cc
```diff
@@ -0,0 +1,788 @@
...
+
+      // Perform the shift in bit-vector domain, then cast back to int.
+      z3::expr result_bv = op_func(a_bv, b_bv);
+      return z3::bv2int(result_bv, true);
+    } else {
+      return Create(op);
+    }
+  }
+
+  z3::expr VisitExprDefault_(const Object* op) override {
```
tlopex, line 780: Here VisitExprDefault_ does a LOG(FATAL) whenever it hits a node type it doesn't handle, which kills the whole build. That's a problem here because Z3 is just a fallback — Analyzer::CanProve only reaches it after the native analyzers have already given up. So if an expression happens to contain some unsupported node, what used to be a harmless "couldn't prove it" now turns into a hard crash. Could we just return a fresh free variable here (return Create(op);)?

src/arith/analyzer.cc
```diff
@@ -230,9 +235,30 @@ bool Analyzer::CanProve(const PrimExpr& expr, ProofStrength strength) {
     }
   }
 
+  if (z3_prover.CanProve(simplified)) {
```
tlopex: The Z3 fallback isn't exception-safe — Z3 can throw z3::exception and nothing catches it, so it escapes Analyzer::CanProve and breaks the calling pass. A fallback shouldn't turn a failed proof into a crash. it's good to wrap the body of Z3Prover::CanProve in try { ... } catch (const z3::exception&) { return false; }.

src/target/z3/z3_prover_on.cc
```diff
@@ -0,0 +1,788 @@
...
+  }
+
+  void CopyFrom(const Self& other_) {
+    // 1. create a new solver
+    //    because this->solver depends on this->ctx
+    //    we need to deconstruct the old solver, and create a new one depending on other_.ctx
+    solver = CreateSolver(*other_.ctx);
+    // 2. copy the context
+    //    the context is a shared_ptr, we can just copy the pointer
+    ctx = other_.ctx;
```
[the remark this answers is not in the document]
tlopex: I think we can remove it as gemini suggested

tlopex:
Z3 fallback currently ignores the strength gate, so enabling it at all (especially in deeply recursive kDefault CanProve calls) causes unnecessary Z3 invocations and a measurable perf hit; it should be gated behind a higher threshold like kSymbolicBound to avoid paying prover cost on cheap paths I think.
And EnterConstraint duplicates EnterWithScope without any real caller in this PR, and the new is_assume flag is effectively dead code since it is never set to true, so we should either deduplicate the entry logic and wire the flag with a real use/test or drop both changes until they’re actually needed.

tests/python/arith/test_arith_z3.py
SiriusNEO, line 1: For the tests, do we need to cover all tests in the original PR?: https://github.com/tile-ai/tilelang/pull/1367/

SiriusNEO:
BTW, the original Z3 integration is in this commit (Not a PR): https://github.com/tile-ai/tvm/commit/e633295de994a89668d7a9930dbbd455af3efc66

src/target/z3/z3_prover_off.cc
SiriusNEO, line 1: Using two source files to control `z3_off/z3_on` is a bit strange. Can we use macros to achieve this so that we can merge them into a single `z3_prover.cc`?

SiriusNEO (changes requested):

tqchen:
given this is optional registration, single file and cmake control addition of the file perhaps is the best approach, great to add testcases to cover things when z3 is detected

SiriusNEO (approved):
LGTM, cc @kurisu6912

tlopex (approved):
LGTM!

PR #19669, opened 2026-06-04
[Arith] Gate canonical-simplify LT Case 2 on extra scale == +1

jinhongyii:
## Summary

`CanonicalSimplifier::Impl::VisitExpr_(LTNode)` Case 2 rewrites

    S + xn < 0  ⇔  S/d + xn // d < 0      where d = gcd(scales)

The Case 1 derivation only works when `xn ≥ 0`. With `scale = -1` the equivalence becomes `≤` rather than `<`, and the rewrite silently strengthens the predicate by dropping the boundary `S/d == xn // d`.

After CSE/inlining, a comparison such as `2*(tx%4) < 16*warp + (tx%32)//4` (where `row` and `col` are independent projections of the same lane id) reaches canonical_simplify with the divided projection on the LHS (scale = -1), and Case 2 folds it to a plain `0 < warp_id` — zeroing every thread that should have written `val` in warp 0. The same path also folds other configurations (e.g. `0 < (tx%32) - 8*warp`) all the way to `False`.

The fix gates Case 2 with `extra->args[0]->scale == 1`. The original target shape (`yn % m` with positive scale and `lower_factor=1`, plus the `scale = +1 / lower_factor > 1` generalization) is unchanged; truly-always-true comparisons still fold to `True`.

## Test plan

- New regression test `test_simplify_le_negative_scale_extra` in `tests/python/arith/test_arith_canonical_simplify.py` — asserts on simplified `PrimExpr`, no GPU required; pre-fix fails, post-fix passes. It also pins the buggy `scale = -1` shapes to their unsimplified form, confirms the `scale = +1` Case 2 path still optimizes, and re-asserts the truly-always-true variant still folds to `True`.
- Existing `test_simplify_le` (the original Case 2 target with `scale = +1`) still passes.
- `tests/python/arith/test_arith_canonical_simplify.py` — 16 passed.
- Full `tests/python/arith/` — 932 passed (1 pre-existing flaky random-seed failure in `test_arith_solve_linear_equations.py` unrelated to this change, passes on rerun).

### DIFF
--- a/src/arith/canonical_simplify.cc
+++ b/src/arith/canonical_simplify.cc
@@ -1419,10 +1419,17 @@ PrimExpr CanonicalSimplifier::Impl::VisitExpr_(const LTNode* op) {
       // Case 1. 0 <= xn < d
       divisible.CopyOnWrite()->DivideBy(gcd);
       return Rewriter::VisitExpr(divisible->Normalize() < make_zero(dtype));
-    } else if (extra->args.size() == 1 &&
+    } else if (extra->args.size() == 1 && extra->args[0]->scale == 1 &&
                extra->args[0]->upper_factor != ConstIntBoundNode::kPosInf &&
                extra->args[0]->upper_factor % (gcd * extra->args[0]->lower_factor) == 0) {
-      // Case 2. xn == yn % m, where m % d == 0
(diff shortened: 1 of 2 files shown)


jinhongyii:
cc: @tqchen

tqchen (approved):

PR #19671, opened 2026-06-04
fix: align Relax ONNX ArgMax/ArgMin NaN handling with ONNX Runtime

### LINKED ISSUE #19558: [Bug] Relax ONNX ArgMax returns different index from ONNX Runtime when input contains NaN
### Expected behavior

TVM Relax should execute ONNX `ArgMax` consistently with ONNX Runtime when the input contains `NaN`.

For these inputs, ONNX Runtime matches NumPy's `argmax` behavior:

```
x = [2, NaN, 7, 4, 1]   -> index 1
x = [NaN, 2, 7, 4, 1]   -> index 0
x = [2, 4, 7, 1, NaN]   -> index 4
```
TVM Relax should return the same indices after importing the ONNX model through the Relax ONNX frontend.

### Actual behavior

TVM Relax returns different indices for cases where NaN appears before the finite maximum:

```
> x=[2, NaN, 7, 4, 1]   np=1 ORT=1 TVM=2
> x=[NaN, 2, 7, 4, 1]   np=0 ORT=0 TVM=2
> x=[2, 4, 7, 1, NaN]   np=4 ORT=4 TVM=4
```

The discrepancy appears when importing an ONNX ArgMax model through the Relax ONNX frontend and compiling it for the llvm target.

### Environment

TVM: 0.14 environment / Relax ONNX frontend
ONNX Runtime: 1.23
Python: 3.11
Target: llvm
OS: Linux

### Steps to reproduce

```
import warnings

warnings.filterwarnings("ignore")

import numpy as np
import onnx
import onnxruntime as ort
import tvm
from onnx import TensorProto, helper
from tvm import relax


node = helper.make_node("ArgMax", ["x"], ["y"], axis=0, keepdims=0)

graph = helper.make_graph(
    [node],
    "g",
    [helper.make_tensor_value_info("x", TensorProto.FLOAT, [5])],
    [helper.make_tensor_value_info("y", TensorProto.INT64, [])],
)

model = helper.make_model(graph, opset_imports=[helper.make_opsetid("", 18)])
model.ir_version = 9

sess = ort.InferenceSession(
    model.SerializeToString(),
    providers=["CPUExecutionProvider"],
)

mod = from_onnx(model)

    ex = tvm.compile(mod, target=tvm.target.Target("llvm"))

vm = relax.VirtualMachine(ex, tvm.cpu())


def run_both(x):
    ort_out = int(sess.run(None, {"x": x})[0])

    out = vm["main"](tvm.runtime.tensor(x, tvm.cpu()))
    tvm_out = int((out[0] if isinstance(out, (list, tuple)) else out).numpy())

    numpy_out = int(np.argmax(x))
    return numpy_out, ort_out, tvm_out


cases = [
    np.array([2.0, np.nan, 7.0, 4.0, 1.0], dtype=np.float32),
    np.array([np.nan, 2.0, 7.0, 4.0, 1.0], dtype=np.float32),
    np.array([2.0, 4.0, 7.0, 1.0, np.nan], dtype=np.float32),
]

for x in cases:
    numpy_out, ort_out, tvm_out = run_both(x)
    print(f"x={x.tolist()} np={numpy_out} ORT={ort_out} TVM={tvm_out}")
```

### Triage

* needs-triage

mvanhorn:
## Summary

ONNX models that run `ArgMax`/`ArgMin` over float tensors containing NaN now import through the Relax frontend with the same winning index that ONNX Runtime returns.

## Why this matters

Issue [#19558](https://github.com/apache/tvm/issues/19558) reports that the Relax ONNX frontend picks a different index than ORT and NumPy when NaN is present: ORT lets NaN win `>` comparisons so a leading NaN survives, while `topi.argmax` drops NaN through `val > current_max` and returns the first finite extreme instead. Maintainer swjng confirmed this is an ORT-compatibility gap in the frontend rather than a spec issue, and suggested substituting NaN before the reduction. The fix lives only in the ONNX converter (`onnx_frontend.py`), where ORT parity is the contract, so the generic `topi`/Relax operator is untouched.

## Testing

A new `_argreduce_sanitize_nan` helper in `python/tvm/relax/frontend/onnx/onnx_frontend.py` replaces NaN with `+inf` for `ArgMax` and `-inf` for `ArgMin` on floating-point inputs before `argmax`/`argmin`, across the `_impl_v1`/`_impl_v11`/`_impl_v12` paths. Added tests in `tests/python/relax/test_frontend_onnx.py` cover the issue's `[2, NaN, 7, 4, 1]`-style cases against ORT indices `1, 0, 4`, all-NaN and `keepdims=1`, `select_last_index=1`, and a finite-only regression case. Covered by the new tests in this PR; full suite runs in CI.

Fixes #19558

### DIFF
--- a/python/tvm/relax/frontend/onnx/onnx_frontend.py
+++ b/python/tvm/relax/frontend/onnx/onnx_frontend.py
@@ -4203,6 +4203,17 @@ def _argreduce_select_last_index(bb, data, axis, keepdims, op):
     return relax.op.subtract(offset, flipped_idx)
 
 
+def _argreduce_sanitize_nan(bb, data, *, for_min):
+    """Match ONNX Runtime ArgMax/ArgMin behavior by making NaN win comparisons."""
+    dtype = data.struct_info.dtype
+    if not _relax_dtype_is_floating_point(dtype):
+        return data
+    replacement = -_np.inf if for_min else _np.inf
+    return bb.emit(
+        relax.op.where(relax.op.isnan(data), relax.const(replacement, dtype), data)
+    )
+
+
 class ArgMax(OnnxOpConverter):
     """Converts an onnx ArgMax node into an equivalent Relax expression."""
 
@@ -4218,19 +4229,19 @@ class ArgMax(OnnxOpConverter):
 
     @classmethod
     def _impl_v1(cls, bb, inputs, attr, params):
-        data = inputs[0]
+        data = _argreduce_sanitize_nan(bb, inputs[0], for_min=False)
         axis, keepdims = cls._check_attrs(data, attr, False)
         return relax.op.argmax(data, axis, keepdims)
 
     @classmethod
     def _impl_v11(cls, bb, inputs, attr, params):
-        data = inputs[0]
+        data = _argreduce_sanitize_nan(bb, inputs[0], for_min=False)
         axis, keepdims = cls._check_attrs(data, attr)
         return relax.op.argmax(data, axis, keepdims)
 
     @classmethod
     def _impl_v12(cls, bb, inputs, attr, params):
-        data = inputs[0]
+        data = _argreduce_sanitize_nan(bb, inputs[0], for_min=False)
         axis, keepdims = cls._check_attrs(data, attr)
         select_last_index = attr.get("select_last_index", False)
         if select_last_index:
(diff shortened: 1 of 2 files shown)


tlopex (changes requested):
Could we add coverage for NaN together with real infinities? Replacing NaN with `+inf`/`-inf` makes it tie with existing infinities, so cases like `ArgMax([inf, NaN])` or `ArgMin([-inf, NaN])` may no longer match NumPy/ORT tie behavior.

mvanhorn:
Good idea - added in 28a09fe: test_arg_min_max_nan_with_real_infinities uses rows that mix NaN with real +inf and -inf on the reduction axis, parametrized over ArgMax/ArgMin and both select_last_index values, with correctness checked against ONNX Runtime so the tie-breaking behavior is locked to ort's semantics rather than hand-coded indices.

PR #19673, opened 2026-06-04
[Web] Add support for OPFS synchronous access handles and committed records

akaashrp:
Add support for synchronous access handles in OPFS (https://developer.mozilla.org/en-US/docs/Web/API/FileSystemFileHandle/createSyncAccessHandle). Sync mode uses FileSystemSyncAccessHandle where available. Replace optional metadata with committed OPFS records written after payload. Records store the URL, payload byte count, and content type. This allows interrupted or partial writes to be treated as cache misses. Stale OPFS directory handles are cleared on `InvalidStateError`.

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -32,13 +32,13 @@
       }
     },
     "node_modules/@babel/code-frame": {
-      "version": "7.29.0",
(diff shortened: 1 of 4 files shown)


akaashrp:
@guan404ming

guan404ming (approved):
LGTM, thanks @akaashrp!

Issue #19543, opened 2026-05-11
[Bug] Relax ONNX Sign returns 0 for NaN instead of preserving NaN


ALinrunrun:
### Expected behavior

TVM Relax should execute ONNX `Sign` consistently with ONNX Runtime when the input contains `NaN`.

For this input:

`x = [NaN, 9.0, -9.0, NaN]`

ONNX Runtime returns:

`[nan, 1.0, -1.0, nan]`

### Actual behavior

TVM Relax returns 0.0 for `NaN` inputs:

```
ORT: [nan, 1.0, -1.0, nan]
TVM: [0.0, 1.0, -1.0, 0.0]
```

The discrepancy appears when importing an ONNX `Sign` model through the Relax ONNX frontend and compiling it for the `llvm` target.

### Environment

TVM: 0.14 environment / Relax ONNX frontend
ONNX Runtime: 1.23
Python: 3.11
Target: llvm
OS: Linux

### Steps to reproduce

```
import numpy as np
import onnx
import onnxruntime as ort
from onnx import TensorProto, helper
import tvm
from tvm import relax
from tvm.relax.frontend.onnx import from_onnx


node = helper.make_node("Sign", ["x"], ["y"])

graph = helper.make_graph(
    [node],
    "g",
    [helper.make_tensor_value_info("x", TensorProto.FLOAT, [4])],
    [helper.make_tensor_value_info("y", TensorProto.FLOAT, [4])],
)

model = helper.make_model(graph, opset_imports=[helper.make_opsetid("", 18)])
model.ir_version = 9

x = np.array([np.nan, 9.0, -9.0, np.nan], dtype=np.float32)

ort_out = ort.InferenceSession(
    model.SerializeToString(),
    providers=["CPUExecutionProvider"],
).run(None, {"x": x})[0]

mod = from_onnx(model, keep_params_in_input=False)

with tvm.transform.PassContext(opt_level=3):
    ex = tvm.compile(mod, target=tvm.target.Target("llvm"))

vm = relax.VirtualMachine(ex, tvm.cpu())

out = vm["main"](tvm.runtime.tensor(x, tvm.cpu()))
tvm_out = (out[0] if isinstance(out, (list, tuple)) else out).numpy()

print("ORT:", ort_out.tolist())
print("TVM:", tvm_out.tolist())
```

### Triage

* needs-triage

cchung100m:
Hi @ALinrunrun 

[...]

Issue #19572, opened 2026-05-16
[Bug] [Relax][ONNX] NaN handling inconsistency in Relu, Sign, ReduceMax, ReduceMin vs ONNX Runtime


wuyii8941:
## Summary

Multiple operators handle NaN differently from ONNX Runtime when accessed through the ONNX frontend:

1. **Relu(NaN) → 0** (ORT: NaN)
2. **Sign(NaN) → 0** (ORT: NaN)
3. **ReduceMax/ReduceMin** — position-dependent NaN behavior:
   - `ReduceMax([NaN, 1.0]) → 1.0` (ORT: NaN)
   - `ReduceMax([2.0, NaN]) → NaN` (ORT: 2.0)

Related: #xxx (bug_019, reduce_max/min NaN CPU vs CUDA at Relax IR level)

## Reproduction

```python
import numpy as np
import onnx
from onnx import helper, TensorProto, numpy_helper
import onnxruntime as ort
import tvm
from tvm import relax
from tvm.relax.frontend.onnx import from_onnx

def run_tvm(model, inputs):
    model = onnx.shape_inference.infer_shapes(model)
    mod = from_onnx(model)
    pipeline = tvm.ir.transform.Sequential([relax.transform.LegalizeOps()])
    exe = tvm.relax.build(pipeline(mod), target="llvm")
    vm = tvm.relax.VirtualMachine(exe, device=tvm.cpu())
    tvm_ins = [tvm.runtime.tensor(v, device=tvm.cpu()) for v in inputs]
    return vm["main"](*tvm_ins).numpy()

# Relu
x = np.array([np.nan, 1.0, np.nan, -2.0], dtype=np.float32)
X = helper.make_tensor_value_info("X", TensorProto.FLOAT, [4])
Y = helper.make_tensor_value_info("Y", TensorProto.FLOAT, [4])
node = helper.make_node("Relu", ["X"], ["Y"])
graph = helper.make_graph([node], "test", [X], [Y])
model = helper.make_model(graph, opset_imports=[helper.make_opsetid("", 18)])

sess = ort.InferenceSession(model.SerializeToString())
print("ORT Relu:", sess.run(None, {"X": x})[0])   # [nan  1. nan  0.]
print("TVM Relu:", run_tvm(model, [x]))             # [0.   1. 0.   0.]

# ReduceMax
x2 = np.array([[np.nan, 1.0], [2.0, np.nan]], dtype=np.float32)
[...]

PR #19674, opened 2026-06-04
[Relax][ONNX] Preserve NaN in Sign to align with ONNX Runtime

cchung100m:
Hi Committers,

This PR fixes issues https://github.com/apache/tvm/issues/19543 and https://github.com/apache/tvm/issues/19572. Any suggestions would be appreciated if you are available.

### Root cause:
The ONNX frontend `Sign` converter directly returned `relax.op.sign(x)`. After legalization, this maps to `topi.sign`, which is implemented via comparisons (x < 0 ? -1 : x > 0 ? 1 : 0). For `NaN`, both comparisons are false, so TVM produced 0, while ONNX Runtime preserves NaN. This created a frontend semantic mismatch for imported ONNX models.

### Solution:
Apply a minimal ONNX-frontend-only fix in `onnx_frontend.py`:
- For floating-point inputs, lower `Sign` as `where(isnan(x), x, sign(x))`.
- Keep non-floating inputs unchanged (`sign(x)`).

### DIFF
--- a/python/tvm/relax/frontend/onnx/onnx_frontend.py
+++ b/python/tvm/relax/frontend/onnx/onnx_frontend.py
@@ -4532,7 +4532,7 @@ class Sign(OnnxOpConverter):
     @classmethod
     def _impl_v9(cls, bb, inputs, attr, params):
         x = inputs[0]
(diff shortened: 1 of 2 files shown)


tlopex (approved):
LGTM. Thanks for the fix

cchung100m:
Thanks to @tlopex 😄

PR #19675, opened 2026-06-05
[Arith] Make Analyzer a tvm-ffi Object

tlopex:
This PR makes `arith::Analyzer` a first-class tvm-ffi object.

The implementation splits the previous concrete `Analyzer` class into:

- `AnalyzerObj`, the mutable object node that owns analyzer state, sub-analyzers, caches, and bindings
- `Analyzer`, a reference-counted `ObjectRef` handle that can be passed across the tvm-ffi boundary

This allows Python and C++ to share the same analyzer instance, so bindings, constraints, and cached facts can persist across FFI calls.

Public APIs that accept an analyzer now use `const arith::Analyzer&`, while internal helper APIs that only borrow the object continue to use `AnalyzerObj*`.

### DIFF
--- a/include/tvm/arith/int_set.h
+++ b/include/tvm/arith/int_set.h
@@ -36,6 +36,7 @@ using tirx::IterVar;
 using tirx::Var;
 using tirx::VarNode;
 
+class AnalyzerObj;
 class Analyzer;
 
 //-----------------------------------------------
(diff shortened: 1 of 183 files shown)


include/tvm/arith/int_set.h
```diff
@@ -96,7 +97,7 @@ class IntSet : public ffi::ObjectRef {
    * \param ana Analyzer used in the proof.
    * \return Whether we can prove it is a single point
    */
-  bool CanProveSinglePoint(Analyzer* ana) const;
+  bool CanProveSinglePoint(AnalyzerObj* ana) const;
```
tqchen: This is a user facing API, in such case, we would expect where Analyzer needs to be passed, likely we would like to switch to const Analyzer& where analyzer object ref is being passed, and analyzer being mutable.

tqchen (dismissed):

Ubospica (approved):
LGTM! cc @tqchen

guan404ming (approved):
Great work, thanks @tlopex

Issue #19570, opened 2026-05-16
[Bug] [Relax][ONNX] Resize produces wrong results with half_pixel/pytorch_half_pixel coordinate transformation and non-integer scales


wuyii8941:
## Expected behavior

`Resize` with `coordinate_transformation_mode=half_pixel` (or `pytorch_half_pixel`, `asymmetric`) and non-integer scale factors (e.g., 2.5x) should compute source coordinates as:
```
src = (dst + 0.5) / scale - 0.5
```
and produce results matching ONNX Runtime.

## Actual behavior

TVM computes incorrect source coordinates, leading to wrong pixel mapping. For a 3×3 input with scale=2.5, the max absolute difference vs ORT is **4.0** (nearest mode) and **0.63** (linear mode).

## Reproduction

```python
import numpy as np
import onnx
from onnx import helper, TensorProto, numpy_helper
import onnxruntime as ort
import tvm
from tvm import relax
from tvm.relax.frontend.onnx import from_onnx

x = np.arange(9, dtype=np.float32).reshape(1, 1, 3, 3)

X = helper.make_tensor_value_info("X", TensorProto.FLOAT, [1, 1, 3, 3])
Y = helper.make_tensor_value_info("Y", TensorProto.FLOAT, None)
roi = numpy_helper.from_array(np.array([], dtype=np.float32), "roi")
sc = numpy_helper.from_array(np.array([1, 1, 2.5, 2.5], dtype=np.float32), "scales")
node = helper.make_node("Resize", ["X", "roi", "scales"], ["Y"],
                        mode="nearest",
                        coordinate_transformation_mode="half_pixel",
                        nearest_mode="floor")
graph = helper.make_graph([node], "test", [X], [Y], initializer=[roi, sc])
model = helper.make_model(graph, opset_imports=[helper.make_opsetid("", 18)])
model = onnx.shape_inference.infer_shapes(model)

# ORT
sess = ort.InferenceSession(model.SerializeToString())
ort_out = sess.run(None, {"X": x})[0]

# TVM
mod = from_onnx(model)
exe = tvm.relax.build(
    tvm.ir.transform.Sequential([relax.transform.LegalizeOps()])(mod), target="llvm")
vm = tvm.relax.VirtualMachine(exe, device=tvm.cpu())
tvm_out = vm["main"](tvm.runtime.tensor(x, device=tvm.cpu())).numpy()

print("Max diff:", np.abs(ort_out - tvm_out).max())  # 4.0
```

[...]

PR #19698, opened 2026-06-09
[Relax][Frontend][ONNX] Fix Resize coordinate error with non-integer scales

cchung100m:
Hi Committers,

This PR fixes issues https://github.com/apache/tvm/issues/19570. Any suggestions would be appreciated if you are available.

### DIFF
--- a/python/tvm/relax/frontend/onnx/onnx_frontend.py
+++ b/python/tvm/relax/frontend/onnx/onnx_frontend.py
@@ -3241,7 +3241,7 @@ class Resize(OnnxOpConverter):
                 scales = [int(val.value) for val in scales.values]
             else:
                 raise ValueError(f"Type {type(scales)} for scale is currently unsupported.")
-            original_spatial_scales = list(scales[2:])
+            original_spatial_scales = [float(s) for s in scales[2:]]
             sizes = []
 
             for i, dim in enumerate(x.struct_info.shape):
--- a/python/tvm/relax/frontend/onnx/onnx_frontend.py
+++ b/python/tvm/relax/frontend/onnx/onnx_frontend.py
@@ -3230,7 +3230,7 @@ class Resize(OnnxOpConverter):
 
         use_dynamic_roi = roi_dynamic_vec is not None
 
-        # Convert scales to sizes if needed, preserving the orginal spatial scales so
+        # Convert scales to sizes if needed, preserving the original spatial scales so
         # the coordinate transformation uses the exact ONNX scale value rather than the
         # lossy ratio derived from floor(input * scale) / input.
         original_spatial_scales = None
@@ -3322,7 +3322,7 @@ class Resize(OnnxOpConverter):
                     roi3d,
                     sizes,
                     "NCDHW",
-                    relax_mode,
+                    topi_mode,
                     coord_mode,
                     rounding_method,
                     cubic_coeff_a,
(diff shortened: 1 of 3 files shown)


cchung100m:
cc @tlopex  @guan404ming

tlopex (changes requested):
Thanks for working on this fix. Preserving the original ONNX scale is the right approach, but I found two remaining correctness gaps.

1. The dynamic-ROI path still drops `original_spatial_scales`.

When `roi_dynamic_vec` is present, `_impl_v18` calls `_emit_resize_topi_dynamic_roi` without forwarding the original scales. The nested resize1d/2d/3d TOPI calls therefore continue to derive the coordinate scale from `input_size / output_size`.

For example, a dynamic ROI input with a 3x3 tensor and scale 2.5 will still use 3/7 instead of 1/2.5. Could we pass the original spatial scales through this helper for all supported ranks and add a regression test?

2. The 2D asymmetric-nearest integer-division optimization ignores the scale override.

For `method="nearest_neighbor"`, `coordinate_transformation_mode="asymmetric"`, and `rounding_method="floor"`, `_resize_2d` may set `use_int_div=True` based only on the input and output sizes. `get_inx` then uses integer division and does not use `scale_x_override`.

A concrete example is input size 3 with scale 2.1. The output size is floor(3 * 2.1) = 6, so the optimization uses `x // 2`, while ONNX requires `floor(x / 2.1)`. These produce different pixel mappings. I think the integer division optimization should be disabled when an explicit scale override is provided, unless the override is proven to match the derived ratio exactly

cchung100m:
Hi @tlopex 

Thank you for the detailed review. I have implemented regression tests as requested.

tlopex:
@cchung100m  Hey, the updated test does not exercise the integer-division optimization because providing `scales` makes `scale_h` and `scale_w` non-`None`, so the optimized path is skipped. Please use `sizes` or add a TOPI test without `scales`.

cchung100m:
Hi @tlopex 
Thanks for the feedback. I've addressed the issue "use sizes or add a TOPI test without scales" by implementing three direct TOPI-level tests that call topi.image.resize2d directly. These tests:
1. Verify the int-div optimization fires when scales=None and the size ratio is an integer
2. Verify the optimization is correctly disabled when an explicit (non-integer) scale override is provided.
3. Verify the optimization is behavior-preserving for integer scale overrides

cchung100m:
@tvm-bot  rerun

tlopex:
Sorry for the late reply. Could we preserve the integer-division path when the explicit scale matches the integer output/input ratio? Disabling it for all explicit scales introduces a numerical regression.

For input shape `(1, 1, 1, 411)`, scales `[1, 1, 1, 41]`, and `nearest/asymmetric/floor`, output index `16810` should read input index `410`. The reciprocal multiplication produces approximately `409.99997`, so it reads index `409` instead.

I reproduced this with the C backend: the base revision matches ONNX Runtime, while this PR fails. Please retain exact integer indexing for this case.

cchung100m:
Hi @tlopex 
Thanks for catching this. You are right that disabling the fast path for every explicit scale was too aggressive. I've updated the fix so the integer-division path is only skipped when the explicit scale actually disagrees with the ratio implied by the sizes; when it agrees (as in your example), it's still used and gives the exact result.

PR #19707, opened 2026-06-09
[Wheel] Fix include path discovery for installed wheels

tlopex:
This PR rewrites `tvm.libinfo.find_include_path()` so that C codegen works from both a source checkout and an installed wheel.

What it does:
- Resolves TVM's own public headers relative to `__file__`: installed package layout first (`tvm/include`, already shipped into the wheel by the existing CMake install rule), with the development source tree as fallback.
- Delegates the FFI and dlpack include paths to the FFI package's own libinfo utilities, which already handle both installed and development layouts.
- Removes the `TVM_HOME` / `TVM_SOURCE_DIR` / `TVM_INCLUDE_PATH` based source-tree guessing and the unused `name` / `search_path` parameters.

### DIFF
--- a/python/tvm/libinfo.py
+++ b/python/tvm/libinfo.py
@@ -22,6 +22,7 @@ import os
 import sys
 from pathlib import Path
 
+from tvm_ffi import libinfo as tvm_ffi_libinfo
 from tvm_ffi.libinfo import load_lib_ctypes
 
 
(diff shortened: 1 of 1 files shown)


tqchen:
should be superseded by https://github.com/apache/tvm/pull/19719

PR #19714, opened 2026-06-10
[Runtime][Tests] Fix contrib wheel tests

tlopex:
This pr fixes two contrib test failures that show up when running source-tree tests against the `apache-tvm` wheel.

- Launch the pickle memoize helper script with `sys.executable` so subprocesses use the same Python environment as pytest.
- Fix `tvm.contrib.random.random_fill` for packed sub-byte dtypes by using the actual tensor storage size instead of the logical element count.

## Root Cause

`tests/python/contrib/test_memoize.py` executed `pickle_memoize_script.py` directly. In a wheel test environment, the shebang can resolve to a different Python than the active wheel venv, causing `ModuleNotFoundError: No module named 'tvm'`.

`random_fill` used the product of tensor shape as the number of values to write. For packed dtypes such as `int4`, two logical elements share one byte, so this wrote past the allocated storage and caused native heap corruption / process abort.

### DIFF
--- a/src/runtime/extra/contrib/random/mt_random_engine.cc
+++ b/src/runtime/extra/contrib/random/mt_random_engine.cc
@@ -142,6 +142,19 @@ class RandomEngine {
   }
 
  private:
+  static int64_t GetFillElementCount(DLTensor* tensor) {
+    size_t data_size = ffi::GetDataSize(*tensor);
+    DLDataType dtype = tensor->dtype;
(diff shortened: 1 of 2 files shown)


tqchen:
i think we can remove support for bits==1 and bits==4 since the current logic did not cover that and we did not have such usage, this will simplify the impl. we can inline run_test_script into caller

tqchen (approved):

PR #19716, opened 2026-06-10
[Codegen][LLVM] Accept splat form in VLA broadcast test

tlopex:
Newer LLVM versions (observed with LLVM 20) print a scalable broadcast store as a splat constant, e.g.
`store <vscale x 4 x float> splat (float 1.000000e+00)`, instead of the older `shufflevector (<vscale x 4 x float> insertelement (...` form. Accept either representation in test_scalable_broadcast so the test passes across LLVM versions while still verifying scalable vector codegen.


tqchen (approved):

tqchen:
please adopt gemini reviews

PR #19717, opened 2026-06-10
[Codegen][NVPTX] Skip runtime execution in Vulkan codegen tests

tlopex:
The generic-target tests in test_target_codegen_vulkan.py are auto-parametrized over all enabled targets, which may include nvptx. For nvptx, TVM produces PTX codegen output but not a directly launchable runtime module, so executing the compiled function fails with errors such as cuModuleGetFunction CUDA_ERROR_NOT_FOUND.

Add a small helper that skips runtime execution for nvptx after a successful compile, so codegen is still exercised while the invalid runtime launch is avoided. Vulkan-only tests are unchanged.

### DIFF
--- a/tests/python/codegen/test_target_codegen_vulkan.py
+++ b/tests/python/codegen/test_target_codegen_vulkan.py
@@ -85,14 +85,11 @@ def test_array_copy(dev, dtype, fuzz_seed):
     tvm.testing.assert_allclose(a_np, a.numpy())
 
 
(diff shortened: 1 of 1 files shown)


tqchen:
we should instead change the test_target_codegen_vulkan so it is only parameterized on vulkan target

tqchen (approved):

tqchen:
note for future, need to change to markskipif not has_vulkan

PR #19731, opened 2026-06-11
[Script][Tests] Fix dialect redirect module re-execution and stray category-less tirx.intrin_test op

tlopex:
This PR fixes two independent test-isolation issues that only surface when certain test files run together in one pytest session.

1. Fix `_DialectRedirectFinder` duplicate module execution

`_DialectRedirectFinder.find_spec` used to pre-register the redirect target module under the legacy alias name before returning the alias spec.

This interacts badly with CPython import logic: when the requested module name is already in `sys.modules`, CPython may ignore the returned alias spec and reuse the target module's original spec instead. As a result, the target source can be executed again under the canonical module name, creating a duplicate module object.

This caused patches on aliased modules to silently miss the module object used by existing code. For example, `unittest.mock.patch("tvm.tirx.script.builder.buffer_store")` patched the duplicate module, while the tirx parser still held references to the original one, so `test_scalar_assign_error_not_swallowed` failed with `DID NOT RAISE`.

This pr removes the pre-registration and let the import machinery register the alias normally. Since the alias spec is now used, `_AliasLoader.exec_module` also restores the canonical `__spec__` and `__loader__` to avoid stale alias metadata on the loaded module.

2. Remove unused `tirx.intrin_test` op registration

`test_s_tir_transform_lower_match_buffer.py` registered a dummy op:

```python
tvm.ir.register_op_attr("tirx.intrin_test", "")
```

This was a leftover from the old TVMScript parser and is no longer needed. The modern tirx parser eagerly evaluates `intrin_test(...)` calls into `T.evaluate(0)`, so this op never appears in parsed IR.

The only remaining effect was adding a category-less `tirx.intrin_test` entry to the global op registry, which could break `test_registered_tirx_ops_have_exactly_one_category` depending on test import order.

This pr removes the unused registration.

### DIFF
--- a/python/tvm/script/__init__.py
+++ b/python/tvm/script/__init__.py
@@ -154,9 +154,11 @@ class _DialectRedirectFinder:
     When the import machinery asks for a module whose full name starts with
     ``tvm.script.<dialect>`` (or ``tvm.script.parser.<dialect>``, etc.) and
     that dialect is in ``_DIALECT_REGISTRY``, :meth:`find_spec` imports the
-    real target module (e.g. ``tvm.tirx.script.parser.entry``) and registers
-    it in ``sys.modules`` under the legacy name, so all subsequent imports and
-    attribute walks resolve correctly without going through the redirect again.
+    real target module (e.g. ``tvm.tirx.script.parser.entry``) and returns an
+    alias spec whose loader hands back that module, so the import machinery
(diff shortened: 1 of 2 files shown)


tlopex:
@tvm-bot rerun

tqchen (approved):

PR #19581, opened 2026-05-17
[TIRx] Bringup TIRx Infrastructure

spectrometerHBH:
## Summary

This PR adds the initial TIRx support needed for low-level programming of Blackwell-class GPU architectures. As part of the ongoing TIRx refactor, it introduces TVMScript support for directly scripting advanced hardware features without relying on scheduling as the primary programming interface.

The change keeps existing `s_tir` script support intact while making direct scripting a first-class path for TIRx programs.

## Main Changes

- Add TIRx operator dispatch and layout infrastructure.
- Add TVMScript support for new low-level TIRx operations.
- Add analysis, transform, and lowering support for TIRx IR nodes.
- Add CUDA/Blackwell-oriented codegen and intrinsic coverage.
- Add Python and C++ integration points for TIRx scripting and runtime support.

## Validation

- `pre-commit run --all-files`
- `ninja -C build -j32`
- `CUDA_VISIBLE_DEVICES=2 pytest tests/python/tirx/ -n 16`
  - `1723 passed, 47 skipped, 32 warnings`
- `CUDA_VISIBLE_DEVICES=2 python -m pytest -v tests/python/all-platform-minimal-test`
  - `37 passed, 105 skipped`
- `TVM_TEST_TARGETS=llvm python -m pytest -v tests/python/tirx-analysis tests/python/tirx-base tests/python/tirx-transform -n 16`
  - `664 passed, 25 skipped, 9 xfailed, 1 xpassed`

## Local CI Notes

Some full CI-equivalent jobs were not locally reproducible because this machine is missing parts of the Apache TVM CI environment, including `llvm-config-15/17`, Vulkan, ROCm, Maven, Sphinx, Doxygen, Emscripten, and ARM/QEMU cross-toolchain components. Metal-specific tests were skipped locally because no Metal runtime is available.

PR #19735, opened 2026-06-11
[S-TIR][Tests] Fix transform test failures after TIRx bringup

tlopex:
This PR fixes 11 test failures in `tests/python/s_tir/transform/` introduced as side effects of the TIRx bringup (#19581 / 859498dc01), in three independent commits.

### 1. LowerOpaqueBlock: update expected IR for buffer metadata annotations

`LowerOpaqueBlock` now emits `buffer_allocated_addr` and `buffer_data_alignment` annotations on lowered allocations (intentional in #19581: the annotations are consumed downstream by `codegen_cuda.cc` / `codegen_trn.cc`; the alignment value 64 comes from `kAllocAlignment`). The tests' expected IR predates this, so `assert_structural_equal` failed on the missing annotations.

Fix: update the expected IR in `test_s_tir_transform_lower_opaque_block.py` to carry the annotations (`T.decl_buffer(...)` → `T.alloc_buffer(..., annotations={...})`). Fixes 6 tests.

### 2. DefaultGPUSchedule: parse scalar-block test in s_tir mode

#19581 added a well-formedness rule rejecting `SBlockRealize` in `tirx=True` mode, which is correct — sblocks are s_tir-mode constructs. The hand-written `Before`/`Expected` modules in `test_scalar_block_no_loops` were the only ones in the file still using plain `T.prim_func`, so they failed at parse time before the pass under test even ran.

Fix: parse both modules with `T.prim_func(s_tir=True)`, consistent with every other test in the file. Fixes 1 test.

### 3. InjectPermutedLayout: match legacy PTX intrinsics by canonical name

#19581 registers device intrinsics under two Op identities: a flat builtin name (returned by `builtin::xxx()` in C++) and a canonical dotted name (e.g. `tirx.ptx.ldmatrix_legacy`, produced when TVMScript / tensor intrinsics are parsed). `InjectPermutedLayout` only compared with `same_as(builtin::...)`, so it silently skipped rewriting the swizzled shared-memory offsets of parsed legacy-form calls, leaving the expected swizzle index expressions unmatched.

Fix: match `ptx_ldmatrix_legacy` / `mma_store_legacy` by both the builtin Op and the canonical name via an `IsOp` helper, following the existing pattern in `lower_warp_memory.cc` and `codegen_cuda.cc`. Only the legacy intrinsic forms fold shared-memory access into `tvm_access_ptr` + offset; non-legacy forms address shared memory through `BufferLoad` and are already handled by the BufferLoad visitor, so the unreachable `InternalError` throw is replaced by a pass-through. (`mma_store_legacy` has no dotted alias, hence the asymmetric name strings.) Fixes 4 tests.

### DIFF
--- a/src/s_tir/transform/inject_permuted_layout.cc
+++ b/src/s_tir/transform/inject_permuted_layout.cc
@@ -246,6 +246,17 @@ class PermutedLayoutInjector : private IRMutatorWithAnalyzer {
     return access_ptr_call;
   }
 
+  // Device intrinsics are registered under both a flat name (the builtin Op)
+  // and a canonical dotted name (emitted by TVMScript and the tensor
+  // intrinsics), so compare against both.
+  static bool IsOp(const Call& call, const Op& compat_op, const char* canonical_name) {
+    if (call->op.same_as(compat_op)) {
+      return true;
+    }
+    const auto* op_node = call->op.as<OpNode>();
+    return op_node != nullptr && op_node->name == canonical_name;
+  }
+
   PrimExpr VisitExpr_(const CallNode* op) final {
     // Rewrite from/to shared or shared.dyn to/from local
     auto call = Downcast<Call>(IRMutatorWithAnalyzer::VisitExpr_(op));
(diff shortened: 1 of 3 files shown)


tlopex:
@tvm-bot rerun

tqchen (approved):

PR #19736, opened 2026-06-11
[Runtime][Disco] Fix session attribute storage, NVSHMEM build, and test gating

tlopex:
The tvm_ffi Object metaclass now gives every subclass `__slots__ = ()`, so the Disco Python wrappers can no longer store instance attributes and every session construction fails with AttributeError. Declare the attributes each
wrapper actually stores as named slots, fix the NVSHMEM `dist_gemm.cu` so TVM builds with `USE_NVSHMEM = ON`, and gate the disco tests on the disco runtime being present so they skip cleanly on builds (e.g. the pip wheel) that report `USE_NCCL` / `USE_NVSHMEM = ON` without shipping it.

### Session attribute storage
- `DPackedFunc` / `DModule`: `__slots__ = ("session",)`.
- `Session`: `__slots__ = ("_cache", "_import_python_module")`

### DIFF
--- a/python/tvm/runtime/disco/session.py
+++ b/python/tvm/runtime/disco/session.py
@@ -78,6 +78,10 @@ class DRef(Object):
 class DPackedFunc(DRef):
     """A PackedFunc in a Disco session."""
 
+    # tvm_ffi Object subclasses cannot store Python attributes by default
+    # (the metaclass sets `__slots__ = ()`); list the field(s) we store here.
+    __slots__ = ("session",)
+
     def __init__(self, dref: DRef, session: "Session") -> None:
         self.__move_handle_from__(dref)
         self.session = session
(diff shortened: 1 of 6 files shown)


python/tvm/runtime/disco/session.py
```diff
@@ -78,6 +78,10 @@ def debug_copy_from(
 class DPackedFunc(DRef):
     """A PackedFunc in a Disco session."""
 
+    # Opt out of the default `__slots__ = ()` of tvm_ffi Object subclasses to
+    # allow Python-side instance attributes (the associated `session`).
+    __slots__ = ("__dict__",)
```
tqchen: better approach is to add slot session

python/tvm/runtime/disco/session.py
```diff
@@ -103,6 +109,10 @@ class Session(Object):
     """A Disco interactive session. It allows users to interact with the Disco command queue with
     various PackedFunc calling convention."""
 
+    # Opt out of the default `__slots__ = ()` of tvm_ffi Object subclasses, as
+    # the session caches method lookups in dynamic instance attributes.
+    __slots__ = ("__dict__",)
```
tqchen: better approach is to add slot session and _cache

tqchen (changes requested):

tqchen (approved):

PR #19737, opened 2026-06-11
[CI] Run s_tir/transform tests in the python-unittest stage

tlopex:
The python-unittest stage enumerates test directories explicitly, and tests/python/s_tir/transform was never enrolled when the transform tests were migrated into the s_tir namespace (#18722). As a result the directory has not been exercised by CI at all, which let several regressions land on main unnoticed

Add the directory to TEST_FILES so the s_tir transform passes are covered. GPU-marked tests are filtered by the -m gpu marker selection in the gpuonly variant, so CPU enrollment is safe.


tqchen (approved):

tlopex:
@tvm-bot rerun

PR #19750, opened 2026-06-12
[Relax][ONNX] Preserve NaN in Relu to align with ONNX Runtime

cchung100m:
Hi Committers,

This PR fixes issues https://github.com/apache/tvm/issues/19572. Any suggestions would be appreciated if you are available.

### DIFF
--- a/python/tvm/relax/frontend/onnx/onnx_frontend.py
+++ b/python/tvm/relax/frontend/onnx/onnx_frontend.py
@@ -1530,7 +1530,12 @@ class Relu(OnnxOpConverter):
 
     @classmethod
     def _impl_v13(cls, bb, inputs, attr, params):
-        return relax.op.nn.relu(inputs[0])
+        x = inputs[0]
+        x_dtype = x.struct_info.dtype if isinstance(x.struct_info, relax.TensorStructInfo) else None
+        y = relax.op.nn.relu(x)
+        if x_dtype is not None and _relax_dtype_is_floating_point(x_dtype):
+            return relax.op.where(relax.op.isnan(x), x, y)
+        return y
 
 
 class Elu(OnnxOpConverter):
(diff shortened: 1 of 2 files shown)


tlopex (approved):
LGTM could you resolve the conflict so that we can merge it in?

cchung100m:
Hi @tlopex 

Thanks for your friendly reminder, I updated the part you mentioned.

guan404ming (approved):

guan404ming:
Thanks @cchung100m!

cchung100m:
Thanks to @guan404ming

tqchen:
are we sure we want to have this behavior? The report is likely mostly come from fuzzer, and while Nan preserving is nice, having an explicit nan where step here will likely slow down the computation, i would rather have us to have less strict behavior while being efficient while leaving Nan handling as undefined behavior.

Instead, i think it may boils down to how low-level handles NaN, e.g. relu should translate to Max, and it is up to each backend to decide whether max(x, 0) should be nan preserving, for cases that can be made efficient we might

tqchen:
Looking more broadly, it appears that we have a series of fixes of nan behavior all via explicit nan checking where steps, i think they are not the directions we want to go toward, e.g. paying extra computing cost for a corner case that may only appear in fuzzer setting. Instead, i think it is better to think about ways to define the right most efficient way in implementing relax.relu

cchung100m:
Thank you @tqchen for the thoughtful feedback, and thank you @tlopex for following up with #19847 to clean these up. 

You are right that wrapping these ops with an explicit `isnan` + `where`(and a full `sum(isnan)` pass for the reductions) pays a per-element cost on the common path to handle a corner case that, in practice, mostly shows up in fuzzer-generated inputs.  Pushing that onto every elementwise/reduce result wasn't the right trade-off.

I'll keep this in mind for future contributions and would be happy to help on the backend-side follow-up if that effort moves forward. Thanks again to both of you for the guidance and the correction.

guan404ming:
Thanks for all suggestion. After #19847, I think the handling is much more better than before.

PR #19755, opened 2026-06-13
[Relax][ONNX] Make ReduceMax/ReduceMin NaN propagation order-independent(numpy semantics)

cchung100m:
Hi Committers,

This PR addresses the `ReduceMax`/ `ReduceMin` part of issue https://github.com/apache/tvm/issues/19572. Any suggestions would be appreciated if you are available.

### Root cause:
The ONNX frontend ReduceMax / ReduceMin converters return relax.op.max / relax.op.min. After legalization these map to topi.max / topi.min, which fold with a commutative reducer whose combiner is Max(x, y) / Min(x, y). In codegen, Max(a, b) lowers to select(a > b, a, b) using an **ordered** float comparison (fcmp ogt), which is false for NaN. As a left-fold (acc = Max(acc, elem)), NaN propagation becomes **position-dependent** - a later non-NaN element silently overwrites an earlier NaN.

### Solution:
Adopt the well-defined, **order-independent numpy/IEEE convention** (matching numpy.max/min and torch.amax/amin): the reduction yields NaN whenever **any** reduced element is NaN. Minimal, ONNX-frontend-only change:

- Add a shared helper _reduce_min_max_preserve_nan(reduce_op, data, axes, keepdims).
- For floating-pint inputs, detect NaN along the reduced axes via `sum(astype(isnan(data), dtype), axes, keepdims) > 0` and force those outputs to `NaN` with `where(has_nan, nan, reduce(data))`. The mask reduces over the **same axes/keepdims**, so it aligns in shape with the reduced result.
- Keep non-floating(integer) inputs unchanged.
- Route all reduce paths(`_impl_v11`and both reduce branches of `_impl_v18`) through the helper; the `noop_with_empty_axes` passthrough is left untouched since it performs no reduction.

### Note on scope (re: #19589 ):
The underlying NaN behavior of Max/Min is the same family of ops discussed in #19589. Per review comments there, enforcing NaN semantics at the IR / LLVM-IR level is undesirable(backward-compat with older LLVM, and portability to CUDA/OpenCL/Vulkan), and a dedicated portable nanmin/nanmax TIRx intrinsic(like `nearbyint`) would be the preferred long-term mechanism. This PR deliberately:

- does not touch the IR-level Max/Min lowering, and
- does not rely on the bool reduction of the NaN mask - it uses `sum(isnan) > 0`, fully sidestepping Max/Min NaN behavior.

### DIFF
--- a/python/tvm/relax/frontend/onnx/onnx_frontend.py
+++ b/python/tvm/relax/frontend/onnx/onnx_frontend.py
@@ -3889,13 +3889,16 @@ class RMSNormalization(OnnxOpConverter):
 
 
 def _reduce_min_max_preserve_nan(reduce_op, data, axes, keepdims):
-    """Apply a min/max reduction while preserving NaN to match ONNX Runtime.
+    """Apply a min/max reduction with well-defined, order-independent NaN propagation.
 
     relax.op.max/min legalize to a max/min fold implemented as select(x > y, x, y) with an
     ordered float comparison, so NaN propagation depends on the fold position (a later non-NaN
-    element silently overwrites an earlier NaN). ONNX Runtime instead yields NaN whenever any
-    reduced element is NaN. For floating point inputs, detect NaN along the reduced axes and
-    force the output to NaN to match that semantics.    
+    element silently overwrites an earlier NaN). ONNX Runtime is also order-independent (it only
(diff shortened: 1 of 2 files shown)


tlopex (approved):

cchung100m:
Thanks to @tlopex

PR #19763, opened 2026-06-14
[Relax][Frontend][TFLite] Add support for FFT/complex operators: REAL, IMAG, COMPLEX_ABS

fnhirwa:
Part of https://github.com/apache/tvm/issues/19519

This PR adds support for the FFT and complex operator family in the Relax TFLite frontend.

**Key implementations:**

- Registered `REAL`, `IMAG`, `COMPLEX_ABS`to the TFLite op map.
- Implemented `convert_real` and `convert_imag` which extract the real and imaginary parts of a complex tensor via `strided_slice` + `squeeze` along the last axis.
- Implemented `convert_complex_abs` which computes `sqrt(re^2 + im^2)` using elementwise Relax ops.
- All three ops adopt a unified representation convention: TFLite `complex64` tensors (which have no native Relax dtype equivalent) are represented as `float32[..., 2]`, where the last axis holds `(real, imaginary)` interleaved..

**Out of scope:**

- `RFFT2D` is not registered in this PR. An O(N²) matmul decomposition
  is feasible using existing Relax ops and will be contributed separately
  with benchmarks showing the performance gap versus a native FFT op.
  A native `relax.op.signal.rfft2d` is tracked in https://github.com/apache/tvm/issues/19764

**Testing:**

- Added structural equality tests for `REAL`, `IMAG`, and `COMPLEX_ABS` in `test_frontend_tflite.py` following the `verify(TestClass, Expected)` pattern.

```bash
python3 -m pytest tests/python/relax/test_frontend_tflite.py -k "test_real or test_imag or test_complex_abs"
```

### DIFF
--- a/python/tvm/relax/frontend/tflite/tflite_frontend.py
+++ b/python/tvm/relax/frontend/tflite/tflite_frontend.py
@@ -207,6 +207,7 @@ class OperatorConverter:
             "BROADCAST_ARGS": self.convert_broadcast_args,
             "CALL": self.convert_call,
             "CALL_ONCE": self.convert_call_once,
+            "COMPLEX_ABS": self.convert_complex_abs,
             "CAST": self.convert_cast,
             "CEIL": functools.partial(self._convert_unary_elemwise, relax_op=_op.ceil),
             "CONCATENATION": self.convert_concatenation,
@@ -252,6 +253,7 @@ class OperatorConverter:
             "HASHTABLE_LOOKUP": self.convert_hashtable_lookup,
             "HASHTABLE_SIZE": self.convert_hashtable_size,
             "IF": self.convert_if,
+            "IMAG": self.convert_imag,
             "L2_NORMALIZATION": self.convert_l2_normalization,
             "L2_POOL_2D": functools.partial(self.convert_pool2d, pool_type="l2"),
             "LEAKY_RELU": self.convert_leaky_relu,
@@ -295,6 +297,7 @@ class OperatorConverter:
             "RANDOM_STANDARD_NORMAL": self.convert_random_standard_normal,
             "RANDOM_UNIFORM": self.convert_random_uniform,
             "READ_VARIABLE": self.convert_read_variable,
+            "REAL": self.convert_real,
             "REDUCE_ALL": functools.partial(self._convert_reduce_bool, relax_op=_op.min),
             "REDUCE_ANY": functools.partial(self._convert_reduce_bool, relax_op=_op.max),
             "REDUCE_MAX": functools.partial(self._convert_reduce, relax_op=_op.max),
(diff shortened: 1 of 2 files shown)


tlopex (changes requested):
Thanks for working on this. I think this needs a small fix before merge.

The `complex64 -> float32[..., 2]` lowering is currently applied in two different places, which can produce wrong input shapes when callers pass `shape_dict` or `dtype_dict` overrides. `_input_type()` already converts complex inputs to `float32[..., 2]`, then `from_tflite()` applies user overrides and may append another trailing `2` if the overridden dtype is `complex64`.

For example:
- `dtype_dict={"x": "complex64"}` can turn an already packed `(2, 4, 2)` shape into `(2, 4, 2, 2)`.
- `shape_dict={"x": (2, 4)}` overwrites the packed default shape while the dtype remains `float32`, so `REAL/IMAG` slice the wrong last axis.

I think the safer structure is to keep `_input_type()` returning the logical TFLite shape/dtype, apply user overrides, and then lower `complex64` to `float32[..., 2]` exactly once.

A couple of related scope/test issues:
- `RFFT2D` is registered in `convert_map`, but `convert_rfft2d()` always raises `OpNotImplemented`. The PR title/body says RFFT2D support and mentions an RFFT2D test, but the patch only adds tests for `REAL`, `IMAG`, and `COMPLEX_ABS`. Please either implement it, or keep it clearly unsupported and add a negative test / update the PR description.
- The new complex representation is only wired for model inputs. Constant or non-input complex tensors still go through the generic tensor path, which does not decode `TensorType.COMPLEX64`. If input-only support is intentional for this PR, please make that scope explicit and cover it with tests.

fnhirwa:
Thanks for the feedback.

**Double lowering fix**: Removed the `complex64 -> float32[..., 2]` transformation from `_input_type()`. The lowering now happens exactly once in `from_tflite()` after user overrides are merged. Verified that `shape_dict` and `dtype_dict` overrides no longer produce double-packed shapes.

**Constant tensor support**: Added `COMPLEX64` to `get_tensor_type_as_numpy`, `get_tensor_type_str`, and `get_tensor_expr`. Constant complex tensors are now reinterpreted as `float32[..., 2]` by reading the buffer as `np.complex64` and calling `.view(np.float32).reshape(shape + (2,))`.

**RFFT2D**: Removed from this PR entirely. An O(N²) matmul decomposition using existing Relax ops (`matmul`, `permute_dims`, `concat`) is feasible as an interim implementation and will be contributed in a follow-up PR. The correct long-term fix is a native `relax.op.signal.rfft2d` with a C++ registered backend, tracked in #19764.

tlopex (approved):
LGTM Thanks!

PR #19766, opened 2026-06-14
[Runtime][WebGPU] Add prerequisites for Gemma 4 E2B WebGPU execution

MakotoUwu:
### Summary

This PR adds a small set of TVM runtime, Relax frontend, TIRx, DLight, WebGPU target, and WebAssembly runtime prerequisites that are exercised by Gemma 4 E2B text-only WebGPU execution through downstream MLC/WebLLM.

The intent is not to add Gemma 4 model support to Apache TVM directly. The downstream model implementation remains in MLC-LLM. This PR only contains lower-layer behavior needed by that downstream path.

### Changes

- Make `DictAttrs::HasNonzeroAttr` accept `IntImm` and bool attributes, in addition to native integer attributes. This fixes TIRx attrs such as `tirx.noalias` when represented as `IntImm`.
- Let DLight GPU matmul schedule rules return non-applicable for helper PrimFuncs without a root block, instead of raising during default schedule application.
- Add `max_shared_memory_per_block = 32768` as the default WebGPU target option in `src/backend/webgpu/codegen/target_kind.cc`.
- Add `freq_dim_base` support for GPT-J-style RoPE frequency generation, so callers can decouple rotated dimensions from the frequency-base dimension.
- Mark generated RoPE PrimFuncs as private to avoid duplicate module-scope symbols when downstream code creates multiple RoPE factories.
- Group TIRx device functions by target kind name instead of full target string to avoid splitting a single backend into multiple device modules when attributes differ.
- Fix PagedKVCache metadata reservation ordering and route sliding MHA attention through the MHA path.
- Make the wasm runtime tolerate native-f32 payloads tagged as `f32-to-bf16`, and add chunked tensor loading support for large records.
- Add focused tests for the `IntImm` attr case and the non-root helper PrimFunc DLight matmul case.

### Validation

Local validation on macOS after rebasing on current `apache/main`:

```text
cmake --build build --parallel 8
```

Result: passed. The build produced nonfatal warnings, mostly `-Woverloaded-virtual` warnings around TIRx visitor overloads, plus LLVM deprecation warnings.

Focused tests:

```text
TVM_LIBRARY_PATH=build/lib python -m pytest \
  tests/python/ir/test_ir_attrs.py::test_dict_attrs_has_nonzero_attr_accepts_int_imm -q

TVM_LIBRARY_PATH=build/lib python -m pytest \
  tests/python/s_tir/dlight/test_gpu_matmul.py::test_matmul_rule_skips_non_root_block_helper_func -q
```

Result:

```text
1 passed in 0.02s
1 passed in 0.03s
```

Branch shape and hygiene:

```text
git rev-list --left-right --count apache/main...HEAD
0 1

git diff --check apache/main...HEAD
# no output
```

Downstream WebLLM smoke evidence from the Apache-prep wasm:

- Wasm SHA256: `70d7295dc91b622b79ceeada2c64b4c20787832631c04e3714de95db04515dfc`
- Browser result title: `apache-pr-validate:ok`
- Prompt checks:
  - `Hi`: load `11.0s`, generation `0.7s`
  - France capital: load `4.3s`, generation `0.2s`, output `Paris`
  - Haiku: load `4.0s`, generation `2.4s`

### Non-goals

- This PR does not add Gemma 4 model registration to TVM.
- This PR does not add MLC-LLM model code.
- This PR does not add WebLLM model-list entries.
- This PR does not claim multimodal Gemma 4 support.
- Full MLC vs Transformers.js benchmarking is intentionally deferred to downstream artifact/model-card work.

### DIFF
--- a/include/tvm/ir/attrs.h
+++ b/include/tvm/ir/attrs.h
@@ -199,7 +199,22 @@ class DictAttrs : public Attrs {
    * \endcode
    */
   bool HasNonzeroAttr(const std::string& attr_key) const {
-    return GetAttr<int64_t>(attr_key, 0).value_or(0) != 0;
+    const DictAttrsNode* node = get();
+    auto it = node->dict.find(attr_key);
+    if (it == node->dict.end()) {
+      return false;
+    }
+    const ffi::Any& value = (*it).second;
+    if (auto opt_int = value.try_cast<int64_t>()) {
+      return opt_int.value() != 0;
+    }
+    if (auto opt_imm = value.try_cast<IntImm>()) {
+      return opt_imm.value()->value != 0;
+    }
+    if (auto opt_bool = value.try_cast<bool>()) {
+      return opt_bool.value();
+    }
+    return false;
   }
 
   // Inline-expand TVM_FFI_DEFINE_OBJECT_REF_METHODS_NOTNULLABLE here, minus
--- a/python/tvm/relax/frontend/nn/llm/position_embedding.py
+++ b/python/tvm/relax/frontend/nn/llm/position_embedding.py
@@ -67,9 +67,23 @@ def rope_freq_default(s: tirx.Var, d: tirx.Var, d_range: int, theta: float, dtyp
     return cos_freq, sin_freq, {freq_var: freq}
 
 
-def rope_freq_gptj(s: tirx.Var, d: tirx.Var, d_range: int, theta: float, dtype: str):
-    """Compute the inverse frequency of RoPE for gptj RoPE scaling."""
-    freq = s / tirx.power(theta, 2 * (d // 2) % d_range / tirx.const(d_range, "float32"))
+def rope_freq_gptj(
+    s: tirx.Var, d: tirx.Var, d_range: int, theta: float, dtype: str,
+    freq_dim_base: int = 0,
+):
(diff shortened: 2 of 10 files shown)


include/tvm/ir/attrs.h
```diff
@@ -199,7 +199,22 @@ class DictAttrs : public Attrs {
    * \endcode
    */
   bool HasNonzeroAttr(const std::string& attr_key) const {
-    return GetAttr<int64_t>(attr_key, 0).value_or(0) != 0;
+    const DictAttrsNode* node = get();
+    auto it = node->dict.find(attr_key);
+    if (it == node->dict.end()) {
+      return false;
```
tqchen, line 205: we are moving away from IntImm attr

tqchen (commented):
Thanks for the PR, would be good to split the PR into webgpu runtime only part which would be easier to review, the compiler side have some unintended issues

MakotoUwu:
Thanks, this makes sense. I split the Web/WebGPU runtime-only portion into a smaller draft PR here: #19771.

I will keep the compiler-side pieces from this draft separate and revisit them independently, especially the attr/RoPE/TIRx changes that caused the unintended CI issues.

